PR #11532, opened 2024-03-11
feat(suite-desktop): use node-bridge as primary transport

mroz22:
Another step in old bridge deprecation:
- ensure feature parity between old and new bridge
- make new bridge run by default in desktop app

As of now, suite-desktop e2e tests already use the new node bridge and it is also possible already to force the new bridge using `--bridge-node` flag. However suite-desktop app in production would still start trezord-go binary on startup. 

  - trezord-go is no longer started, trezor-suite starts node-bridge module which exposes the same interface as the old bridge instead
  - old bridge is stil runable using flags `--bridge-legacy` and `--bridge-legacy-dev` (for emulator support) 

issues: 
- #12163

PR #12235, opened 2024-05-02
suite-desktop: make node bridge accept ilogger

mroz22:
part of #11532

At the moment we use two major logging mechanisms [`Log` ](https://github.com/trezor/trezor-suite/blob/develop/packages/utils/src/logs.ts#L12 )(connect, transport) and [`ILogger`](https://github.com/trezor/trezor-suite/blob/develop/packages/suite-desktop-core/src/libs/logger.ts#L43) used by suite-desktop. Maybe there are more of them. 

Since node-bridge is being added to suite-desktop, these 2 mechanisms start to mix. The main motivation of this PR is to unify log output of suite-desktop (see screenshots below). 

before: 
image

after:
image

How far should we go now? 
 
- [x] - easy: overwrite `ILogger` properties which are needed in runtime to match `Log`
- [ ]  - advanced: create some converter fn in suite-desktop which takes `ILogger` and returns `Log` without any ts-ignores
- [ ] - pro: unify `ILogger` and `Log` under single interface

### DIFF
--- a/packages/suite-desktop-core/src/modules/bridge.ts
+++ b/packages/suite-desktop-core/src/modules/bridge.ts
@@ -48,6 +48,16 @@ const getBridgeInstance = () => {
             port: 21325,
             api: bridgeNodeTest ? 'udp' : 'usb',
             assetPrefix: '../build/node-bridge',
+            // passing down ILogger where Log is expected.
+            // @ts-expect-error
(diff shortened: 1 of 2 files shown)


marekrjpolak (approved):
Working nicely :+1:

Issue #12157, opened 2024-04-25
(Draft) New Connect release procedure


MiroslavProchazka:
### Background
To improve the stability of our releases and address any potential bugs that may impact our users, we have transitioned from continuous releases to date-anchored releases. These releases experience thorough testing by our QA team and are made available in early access to 3rd party developers well ahead of their stable release, this will help us discover and eliminate possible bugs way ahead of Stable release, thus contributing to safer and more reliable environment. 

### Stable vs Beta releases

We do 2 types of release builds - `Stable` and `Beta`

`Stable release` once in two months (last week of the month)

`Canary release` are being released continuously (when we decide)

### Versioning System

`Canary release` - 9.1.x-canary (9.1.12-canary)
Bumping just the _patch_ version with "canary" tag at the end

`Stable release` - 9.x.0 (e.g., 9.2.0)
Bumping _minor_ version

New _major_ version is released when there are breaking changes on API level. Previous major version will be maintained for 12 months and after this period it can be taken down without notice.

Current major version 9 is being developed on top of develop branch. When there is a new major version a new branch should be created (v-<major-version>) and mantained separately.

Furthermore we will do split versioning for specific parts of connect
npm package web
npm package node
npm package webextension
content of trezor.io (popup + iframe + core).

### Release process
Each Canary version will be deployed to https://connect.trezor.io/canary/9/trezor-connect.js.
Each Stable version will be deployed to https://connect.trezor.io/9/trezor-connect.js.

We will be using Canary releases as well for any hot-fixes, if such situation occurs. In that case any developer or third party will be able to use this hot-fix version instead of the Stable stream until the hot-fix become Stable release, or he simply switch back to the main release stream.
We would also welcome any developer or third party to try our Canary releases and give us feedback on new features and latest changes.

### Release flow
`Canary`
**Continuous development** - *Day x-y*
**Freeze** - *Day 1*
**Build preparation** - *Day 1*
**QA testing** - *Day 1-2
**Canary release preparation + deployment** -  *Day 3*
**Post release testing** - *Day 3
**Connect explorer update*** - *Day 3*

`Stable`
**Freeze** - *Day 1
**NPM package check** - Day 1*
**Build preparation and deploy to staging** - *Day 2* 
***Communication to partners and connect explorer update** - Day 2*
**QA testing** - *Day 3-7* 
**Beta release preparation + deployment** -  *Day 7*
**Post release testing** - *Day 7*

### External communication
[...]

Issue #11918, opened 2024-04-08
Migrate Connect relase from GitLab to GiHub actions.


karliatto:
We have 2 main sections we need to totally move all the CI release process.

### NPM releases
* [x] beta deploy connect dependencies
* [x] beta deploy connect v9
* [x] deploy production connect dependencies
* [x] deploy production connect v9

### connect.trezor.io release
* [x] Create a template to build and release connect that can be used for multiple environments
* [x]  test release process to dev.suite.sldev.cz
* [x] release to connect.trezor.io/9

Related to https://github.com/trezor/trezor-suite/issues/12157

karliatto:
Connect v9 from develop branch was deployed with actionhttps://github.com/trezor/trezor-suite/actions/runs/8596271618 to https://dev.suite.sldev.cz/connect-release/9/ we should change the URL and paths as well as add credentials and environment to the action https://dev.suite.sldev.cz/connect-release/9/

PR #12243, opened 2024-05-03
ci(scripts): update packages when moving to stable from canary

karliatto:
## Description
Updating the scripts that identify which packages need to be updated now when we have `stable` and `canary` releases there are 2 things we need to check extra:
1. When doing a `canary` release when last one was also `canary` we need to check if changes where happening in that tag because that contained the last version of the code.
2. When doing an `stable` release when last one was `canary` we need to release the package also to make sure the package is released with the `latest` tag which is the NPM tag that we use for `stable` releases in NPM.

You can easily test it by running the script `ci/scripts/check-npm-dependencies.js` with the required arguments.
## Related Issue

Related https://github.com/trezor/trezor-suite/issues/11918
Related https://github.com/trezor/trezor-suite/issues/12157

### DIFF
--- a/ci/scripts/check-npm-dependencies.js
+++ b/ci/scripts/check-npm-dependencies.js
@@ -1,6 +1,8 @@
 /**
- * node check-npm-dependencies.js <package_name>
- */
+ * node check-npm-dependencies.js <package_name> <semver>
+ *
+ * This script checks the dependencies for the specified package to ensure they match expected criteria.
+ * It is used to verify the output of ci/scripts/connect-release-init-npm.js. */
 const { promisify } = require('util');
 const fs = require('fs');
 const path = require('path');
@@ -13,8 +15,16 @@ const packagesPath = path.join(rootPath, 'packages');
 
 const args = process.argv.slice(2);
 
-if (args.length < 1) throw new Error('Check npm dependencies requires 1 parameter: package name');
-const [packageName] = args;
+if (args.length < 2)
+    throw new Error('Usage: node check-npm-dependencies.js <package_name> <semver>');
+
+const [packageName, semver] = args;
+const allowedSemvers = ['patch', 'minor', 'prerelease'];
+if (!allowedSemvers.includes(semver)) {
+    throw new Error(`Provided semver '${semver}' must be one of ${allowedSemvers.join(', ')}`);
+}
+
+const deploymentType = semver === 'prerelease' ? 'canary' : 'stable';
 
 (async () => {
     const packages = await readdir(packagesPath, {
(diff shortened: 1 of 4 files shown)


ci/scripts/check-npm-dependencies.js
```diff
@@ -13,8 +15,16 @@ const packagesPath = path.join(rootPath, 'packages');
 
 const args = process.argv.slice(2);
 
-if (args.length < 1) throw new Error('Check npm dependencies requires 1 parameter: package name');
-const [packageName] = args;
+if (args.length < 2)
+    throw new Error('Usage: node check-npm-dependencies.js <package_name> <semver>');
+
+const [packageName, semver] = args;
+const allowedSemvers = ['patch', 'minor', 'prerelease'];
```
mroz22, line 22: note:
we should unify terms, `prerelease` `beta` `canary` all of these meaning more or less the same.

mroz22 (commented):
good. I am just starting to question whether we should use the term `canary` which I proposed initially. 

### Pros:

- canary IMHO bears some additional meaning. it is a superset of beta but additionally it denotes to the intention of continuous integration testing by 3rd parties.

### cons:
- in privacy oriented businesses, canary has typical another meaning https://trezor.io/transparency/canary.txt. These are 2 totally different context but seeing the word canary in connection with trezor could trigger some of the users.
- we historically did use word beta and we already have beta releases in npm.

karliatto:
> good. I am just starting to question whether we should use the term `canary` which I proposed initially.
> ### Pros:
> 
>     * canary IMHO bears some additional meaning. it is a superset of beta but additionally it denotes to the intention of continuous integration testing by 3rd parties.
> 
> 
> ### cons:
> 
>     * in privacy oriented businesses, canary has typical another meaning https://trezor.io/transparency/canary.txt. These are 2 totally different context but seeing the word canary in connection with trezor could trigger some of the users.
> 
>     * we historically did use word beta and we already have beta releases in npm.

Yes, for now canary is used only as a name in our actions so no facing out side world, but I think we could easily change it with only benefits, at least more clarity. In NPM when we do a canary release we use `beta` tag, and when doing an stable release we use `latest`.

We want to discuss as well if when moving form beta to stable in NPM we want to change the version number or just change the channel (beta, latest) I am going to do a bit more research but it looks like if we want to keep consistency of package version number in latest tag we have to add `-beta.0` to the end on the version number.

Without `-beta.0` at the end of the version number we would do multiple beta releases for example 9.2.4, 9.2.5, 9.2.6 and then if we want to do an stable release we have to go to 9.2.7 or make more confusion using one of the version that was used for beta as stable (if that is even possible in NPM).

ci/scripts/check-npm-dependencies.js
karliatto: Yes, it is nice to unify, but it is complex since `yarn bump` uses `prerelease` NPM uses `beta` and we in the current approach of releases looks like we are using canary.

karliatto:
Regarding this PR, we need to have this PR either if we go with one approach of versioning or another of the described here https://github.com/trezor/trezor-suite/issues/12157
So I do not see any blocker right now to merge this PR.

mroz22 (approved):

PR #12249, opened 2024-05-03
feat(connect): update protobuf messages

peter-sanderson:
## Description

Result of running `yarn update-protobuf main`

## Related Issue

Resolve

### DIFF
--- a/packages/protobuf/src/messages.ts
+++ b/packages/protobuf/src/messages.ts
@@ -1734,8 +1734,16 @@ export type ResetDevice = {
     backup_type?: Enum_BackupType;
 };
 
+export type Slip39Group = {
+    member_threshold: number;
+    member_count: number;
+};
+
 // BackupDevice
-export type BackupDevice = {};
+export type BackupDevice = {
+    group_threshold?: number;
+    groups?: Slip39Group[];
+};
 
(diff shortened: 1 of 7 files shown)


packages/protobuf/src/messages.ts
```diff
@@ -1734,8 +1734,16 @@ export type ResetDevice = {
...
+export type Slip39Group = {
+    member_threshold: number;
+    member_count: number;
+};
+
 // BackupDevice
-export type BackupDevice = {};
+export type BackupDevice = {
+    group_threshold?: number;
+    groups: Slip39Group[];
```
peter-sanderson: this is reauired, it it ok? @mroz22

mroz22: hmm, if so, it would be probably better to pass empty array as a default? I am not sure how to do it in some sane way. Maybe @martykan got some idea? 🤔

martykan: BackupDevice is a management restricted method, so I think it's not a big deal if there is a breaking change, since it shouldn't be used outside of Suite

mroz22 (approved):
looks good. more descriptive commit message would be nice since we use them to generate changelogs.

PR #12252, opened 2024-05-03
chore(suite): remove cross-fetch, we dont need it as we use node18 everywhere

### LINKED ISSUE #10146: Remove cross-fetch lib as we use Node18 everywhere
- remove `cross-fetch`
- BEWARE of Timeout Issue. SEE: https://github.com/trezor/trezor-suite/pull/11820
- beware of https://github.com/trezor/trezor-suite/pull/10137

peter-sanderson:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/10146


mroz22:
doesn't it break bridge communication for long lasting `/call` calls?

martykan:
Yeah it had to be reverted last time (after a bunch of painful troubleshooting) 
https://github.com/trezor/trezor-suite/pull/11820

peter-sanderson:
Ouch, lets wait for Node 19 then.

marekrjpolak:
> Ouch, lets wait for Node 19 then.

And what will 19 change? What we need is unlimited timeout by default.

peter-sanderson:
@marekrjpolak I was following on Komars message on Slack that we may try again when we upgrade node. 

> We can try removing cross-fetch again when we upgrade node.

PR #12258, opened 2024-05-03
fix(connect): add favicon to log.html

### LINKED ISSUE #12257: missing favicon in connect.trezor.io/9/log.html
missing favicon in connect.trezor.io/9/log.html

image

mroz22:
resolve #12257

maybe it would be better to put a static file there, this is IMHO ok too.


martykan:
In popup.html we just use
```
<link rel="icon" type="image/png" href="images/favicon.png" />
```
wouldn't that be nicer?

Edit: fixed link that GH added

martykan (approved):

PR #12260, opened 2024-05-03
Top banner position

marekrjpolak:
## Description

Very quick and dirty fix of top banner position. Definitely must be polished before potentially accepted.

Should resolve this 
image
and this
image
.

### DIFF
--- a/packages/suite/src/components/suite/banners/Banner.tsx
+++ b/packages/suite/src/components/suite/banners/Banner.tsx
@@ -67,7 +67,7 @@ const Wrapper = styled.div<{ $variant: BannerVariant }>`
     padding: ${spacingsPx.xs} ${spacingsPx.sm};
     ${typography.highlight}
     border-radius: ${borders.radii.sm};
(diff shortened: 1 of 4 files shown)


MiroslavProchazka:
cc @jvaclavik  - can you polish it and merge it, please?

marekrjpolak:
Or just tell me what's wrong with this solution, please. If the problem is that not whole main column scrolls, it's also possible to place the banner above the main column header and let it scroll away (which it currently does anyway).

marekrjpolak:
Irrespective of banner currently being reverted to the original state, I've split my draft into two parts:

The first commit just moves the banner above the main column header, but it's still part of AppWrapper and it can scroll away; it's kinda clean and uncontroversial imo.

The second commit changes the layout of the main column header as before, so the banner doesn't scroll unlike the rest of the main column.

But still, I don't know all requirements of yours, so they're just proposals.

jvaclavik (approved):

PR #11443, opened 2024-03-05
fix(connect): fix build of connect-web

Nodonisko:
## Related Issue

Resolve #11442

Issue #11442, opened 2024-03-05
Web app build fails with `@trezor/connect-web@9.2.0`


gabrielKerekes:
**Describe the bug**
Updating `@trezor/connect-web` to version `9.2.0` results in build failure due to failed module resolution. Namely:
```
 ⨯ ./node_modules/@trezor/connect-web/lib/channels/serviceworker-window.js:4:19
Module not found: Can't resolve '@trezor/connect-common/src/messageChannel/abstract'
```

Version 9.1.6 worked without problems.

More info as well as steps to reproduce can be found in [this Solana wallet adapter issue](https://github.com/anza-xyz/wallet-adapter/issues/918).

In addition to the repo linked in the Wallet Adapter issue I was also able to reproduce the issue when updating `@trezor/connect-web` to `9.2.0` in [one of my own test repos](https://github.com/gabrielKerekes/solana-connect-test) and in NuFi wallet as well.

I feel like the issue might stem from changes introduced here - https://github.com/trezor/trezor-suite/commit/7febd10cf0aab51090190b7e8a4810e9562ddbc3#diff-c34fda44918aae434921e87b60b832f12afedea27dac76ba67b693f779099d6dL21.

I tried reverting these changes locally and `yarn link`ing the package to NuFi and NuFi build suddenly passed. I tried introducing the changes again and re-linking to properly isolate the issue, but I wasn't able to re-link for some reason 🙈

mroz22:
just to clarify - this happens only after updating npm package, right? doesnt affect runtime without any update on your side?

gabrielKerekes:
> doesnt affect runtime without any update on your side?

True.

However, it also affects new apps integrating with Solana wallet adapter since connect is listed as `"@trezor/connect-web": "^9.1.6"` i.e. version 9.2.0 gets pulled for apps which are newly integrating wallet adapter and it breaks their build.

PR #12271, opened 2024-05-06
chore(connect): smaller build of content-script.js

mroz22:
this change by @Nodonisko https://github.com/trezor/trezor-suite/pull/11443

made content-script.js bloat a bit by bundling some unwanted code into it. this little change here removes about 80 lines from the resulting bundle, not a big deal but let's keep it clean.


mroz22:
Aha, but now I see that Nodonisko was actually fixing this issue here https://github.com/trezor/trezor-suite/issues/11442

PR #12274, opened 2024-05-06
fix(connect-popup): wait for POPUP.LOADED in webextension

martykan:
## Description

1. Always use `popupPromise` in PopupManager
2. Wait for `popupPromise` before sending `POPUP.INIT`
3. Fire `onLoad` in popup even in case the document is already loaded

### DIFF
--- a/packages/connect-webextension/src/contentScript.ts
+++ b/packages/connect-webextension/src/contentScript.ts
@@ -1,4 +1,5 @@
 import { WindowServiceWorkerChannel } from '@trezor/connect-web/src/channels/window-serviceworker';
+import { POPUP } from '@trezor/connect/src/events/popup';
 
 /**
  * communication between service worker and both webextension and popup manager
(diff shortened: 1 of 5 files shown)


packages/connect-webextension/src/contentScript.ts
```diff
@@ -15,7 +16,7 @@ channel.init().then(() => {
     // once script is loaded. send information about the webextension that injected it into the popup
     window.postMessage(
         {
-            type: 'popup-content-script-loaded',
+            type: POPUP.CONTENT_SCRIPT_LOADED,
```
mroz22, line 46: this slightly increases bundle size of content-script.js 🗡️ 
it would like take all the imports from popup.ts with it 
```
import { UI_EVENT } from './ui-request';
import type { TransportInfo } from './transport';
import type { ConnectSettings, SystemInfo } from '../types';
import type { MessageFactoryFn } from '../types/utils';
```

martykan: After it's minified the impact on output size is totally negligible, only +24 bytes.

mroz22 (approved):
lets give it a shot

PR #12275, opened 2024-05-06
ci(connect-explorer-nextra): allow manual ci run

martykan:
## Description

Add `on: workflow_dispatch` to allow manually running CI. 
Useful for example when we want to deploy from the develop branch during the day.


karliatto (approved):
:+1:

PR #12282, opened 2024-05-06
fix(github): typo in group name in GitHub Actions

karliatto:
## Description
Typo in group name in GitHub Actions.

I might be wrong but I think the `$$` double dollar sign is not required, but it could be some GitHub Action feat I am not aware of.


vdovhanych (approved):
Yeah, double $$ shouldn't be required.

PR #12296, opened 2024-05-07
chore(github): update ::set-output in GH Actions

karliatto:
## Description
Github is deprecating `set-output` https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

So this PR updates to using `GITHUB_OUTPUT`  instead.


vdovhanych (approved):
wanted to do this myself thanks 😀

PR #12299, opened 2024-05-07
fix(connect-popup): webextension mv2 loading issue

martykan:
## Description

Content script may not be ready to pass the `POPUP.BOOTSTRAP` message when it's sent from `popup.html`, which would cause the entire loading process to fail.

Mitigation:
- Add 100ms delay to `popup.html` before sending `POPUP.BOOTSTRAP` and loading `popup.js` to ensure content script executes first
- Increase popup open timeout from 3s to 5s to give it more time to load on slower networks

E2E change: 
- Run both tests in an independent context - before a failure of MV2 would affect MV3.

### DIFF
--- a/packages/connect-popup/src/static/popup.html
+++ b/packages/connect-popup/src/static/popup.html
@@ -3241,18 +3241,18 @@
                     peer: '@trezor/connect-web',
                 },
             };
-            if (window.opener) {
-                window.opener.postMessage(message, '*');
-            } else {
-                window.postMessage(message, '*');
-            }
-
-            const popupScript = document.createElement('script');
-            popupScript.setAttribute('type', 'text/javascript');
-            popupScript.setAttribute('src', '<%= htmlWebpackPlugin.files.js[0] %>');
-            popupScript.setAttribute('async', 'true');
             setTimeout(() => {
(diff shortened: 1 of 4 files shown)


packages/connect-popup/src/static/popup.html
```diff
@@ -3241,18 +3241,18 @@ <h3>Enter your passphrase on <span class="device-name"></span></h3>
...
-                window.opener.postMessage(message, '*');
-            } else {
-                window.postMessage(message, '*');
-            }
-
-            const popupScript = document.createElement('script');
-            popupScript.setAttribute('type', 'text/javascript');
-            popupScript.setAttribute('src', '<%= htmlWebpackPlugin.files.js[0] %>');
-            popupScript.setAttribute('async', 'true');
             setTimeout(() => {
```
mroz22, line 3244: please add a comment why 🙏

mroz22 (approved):

packages/connect-popup/src/static/popup.html
martykan: There is a comment from before on the line below, is it enough or should I add more information?

mroz22: ah, its ok, sorry, missed that

PR #12300, opened 2024-05-07
feat(suite-native): new device switcher

### LINKED ISSUE #11892: View only - wallet switcher
[Figma](https://www.figma.com/file/mmDnYtctfjnreWhTjLfeQH/Passphrase?type=design&node-id=2146-52545&mode=design&t=byABEIZniotrg0l6-0)

Update wallet switcher visual to show view only devices correctly

- [x] Generic device icon changed to model-specific
- [x] Show current device connection state
- [x] Show/hide device selector on button tap
- [x] Show all devices including connection state (view only) 
- [x] Make device selector slide out/in animated
- [x] Show all wallets for current device
- [x] Hide` Open passphrase` button behind feature flag and make it disabled during discovery
- [x] New connected state for View only settings harmonised with new design
- [x] New visual for current device/wallet when closed manager
 
image

These enhancements will be done in followup issue #12421
- Show wallet fiat amount
- Scrolling wallets to fit size

vytick:
Update wallet switcher visual to show view only devices correctly

 - [x] Generic device icon changed to model-specific
 - [x] Show current device connection state
 - [x] Show/hide device selector on button tap
 - [x] Show all devices including connection state (view only)
 - [x] Make device selector slide out/in animated
 - [x] Show all wallets for current device
 - [x] Hide Open passphrase button behind feature flag and make it disabled during discovery
 - [x] New connected state for View only settings harmonised with new design
 - [x] New visual for current device/wallet when closed manager
 
 These two tasks will be dealt with in following PR so this one is not even bigger
 -  Show wallet fiat amount
 -  Scrolling wallets to fit size

## Related Issue

Part of #11892

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/b7858d48-e537-48d2-882b-e3a84c280a1d


vytick:
@SocketSecurity ignore-all

suite-common/suite-utils/src/device.ts
```diff
@@ -409,6 +409,20 @@ export const getPhysicalDeviceUniqueIds = (devices: Device[]) =>
 export const getPhysicalDeviceCount = (devices: Device[]) =>
     getPhysicalDeviceUniqueIds(devices).length;
 
+export const getSortedDevicesWithoutInstances = (
```
juriczech, line 412: Please create a selector for this. 

Also what does it mean without instances? The name seems confusing. You're just excluding device who's id you're taking as param, or am I missing something? 🤔

suite-native/device-manager/src/components/AddHiddenWalletButton.tsx
```diff
@@ -1,12 +1,32 @@
...
 import { useDeviceManager } from '../hooks/useDeviceManager';
 
+const contentStyle = prepareNativeStyle(utils => ({
+    paddingHorizontal: utils.spacings.medium,
+    paddingVertical: 12,
+    alignItems: 'center',
+    height: 44,
+    gap: utils.spacings.small,
+    backgroundColor: utils.colors.backgroundSurfaceElevation1,
+    borderWidth: 1,
```
juriczech:
```suggestion
    borderWidth: utils.borders.widths.small,
```

suite-native/device-manager/src/components/AddHiddenWalletButton.tsx
```diff
@@ -21,8 +41,17 @@ export const AddHiddenWalletButton = () => {
...
+            style={applyStyle(flexStyle)}
+            testID="@device-manager/passphrase/add"
+        >
+            <HStack style={applyStyle(contentStyle)}>
+                <Text variant="hint" style={applyStyle(flexStyle)}>
+                    <Translation id="deviceManager.deviceButtons.addHiddenWallet" />
+                </Text>
+                <Icon name="chevronRight" color="iconDefault" size="mediumLarge" />
+            </HStack>
+        </Pressable>
```
juriczech: JSX & styles seem almost identical to `DeviceInfoButton`. I would create a component or at least wrapper component

suite-native/device-manager/src/components/DeviceManagerModal.tsx
```diff
@@ -1,47 +1,63 @@
...
+import { selectDevice } from '@suite-common/wallet-core';
+import { AcquiredDevice } from '@suite-common/suite-types';
 
-import { DeviceSwitch } from './DeviceSwitch';
 import { useDeviceManager } from '../hooks/useDeviceManager';
+import { DeviceItemContent } from './DeviceItemContent';
 
 type DeviceManagerModalProps = {
     children: ReactNode;
+    customSwitchRighView?: ReactNode;
```
juriczech:
```suggestion
    customSwitchRightView?: ReactNode;
```

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { selectDevice, selectDevices } from '@suite-common/wallet-core';
+import { VStack } from '@suite-native/atoms';
+import { AcquiredDevice } from '@suite-common/suite-types';
+import { getDeviceInstances } from '@suite-common/suite-utils';
+
+import { WalletItem } from './WalletItem';
+
+type WalletListProps = {
+    onSelectedDevice?: (device: AcquiredDevice) => void;
```
juriczech: Please make this prop required. Why would it be undefined?
```suggestion
    onSelectedDevice: (device: AcquiredDevice) => void;
```

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const WalletList = ({ onSelectedDevice }: WalletListProps) => {
+    const device = useSelector(selectDevice) as AcquiredDevice;
+    const allDevices = useSelector(selectDevices);
+    const wallets = getDeviceInstances(device, allDevices);
```
juriczech: Please create a merged selector for this

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+    extend: {
+        condition: isSelected,
+        style: { borderWidth: 2, borderColor: utils.colors.borderSecondary },
+    },
+}));
+
+type WalletItemProps = {
+    device: AcquiredDevice;
+    isSelectable?: boolean;
+    onPress?: () => void;
```
juriczech: Undefined is not handled but also shouldn't be needed, so please make this required

```suggestion
    onPress: () => void;
```

suite-native/device-manager/src/components/DeviceItemContent.tsx
```diff
@@ -1,82 +1,201 @@
...
-import { useActiveColorScheme } from '@suite-native/theme';
+import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
+
+import { DeviceModelIcon } from './DeviceModelIcon';
+
+type DeviceItemContentVariant = 'simple' | 'walletDetail';
 
 type DeviceItemContentProps = {
-    deviceId?: TrezorDevice['id'];
+    device?: AcquiredDevice;
```
juriczech: Why did you change this? Prefer sending ID as a prop and use selector here to select the device

suite-native/device-manager/src/components/DeviceItemContent.tsx
```diff
@@ -1,82 +1,201 @@
-import React from 'react';
+import React, { ReactNode } from 'react';
```
juriczech: There is 5 components in this file. Could this be separated a little better please?

suite-native/device-manager/src/components/DeviceItemContent.tsx
```diff
@@ -1,82 +1,201 @@
...
 
-    const isPortfolioTrackerDevice = deviceId === PORTFOLIO_TRACKER_DEVICE_ID;
+    const isPortfolioTrackerDevice = device?.id === PORTFOLIO_TRACKER_DEVICE_ID;
+
+    const deviceHeader =
+        (isPortfolioTrackerDevice ? deviceName : deviceLabel) ??
+        translate('deviceManager.defaultHeader');
+
+    const trimmedHeader =
+        deviceHeader.length > 24 ? `${deviceHeader.slice(0, 24)}...` : deviceHeader;
```
juriczech: Why don't use use this? 
https://reactnative.dev/docs/text#ellipsizemode

suite-native/device-manager/src/components/PortfolioTrackerDeviceManagerContent.tsx
```diff
@@ -23,8 +24,18 @@ type NavigationProp = StackToStackCompositeNavigationProps<
     RootStackParamList
 >;
 
+const contentStyle = prepareNativeStyle(utils => ({
+    paddingHorizontal: utils.spacings.medium,
+
+    spacing: utils.spacings.medium,
+
+    paddingBottom: utils.spacings.medium,
+    marginTop: utils.spacings.medium,
```
juriczech, line 33: Why does this have `marginTop` but not `DeviceManagerContent` which is rendered in the same conditional?

suite-native/device-manager/src/components/DeviceManagerModal.tsx
```diff
@@ -1,47 +1,63 @@
...
+const deviceSwitchWrapperStyle = prepareNativeStyle<{ insets: EdgeInsets }>(
     (utils, { insets }) => ({
-        paddingTop: Math.max(insets.top, utils.spacings.small),
+        paddingTop: insets.top + utils.spacings.large,
         backgroundColor: utils.colors.backgroundSurfaceElevation1,
         borderBottomLeftRadius: MANAGER_MODAL_BOTTOM_RADIUS,
         borderBottomRightRadius: MANAGER_MODAL_BOTTOM_RADIUS,
+        borderWidth: utils.borders.widths.small,
+        borderColor: utils.colors.borderElevation0,
+        zIndex: 20,
```
juriczech, line 41: Why is there zIndex here and on the animation as well?

juriczech (changes requested):
Mostly just nitpicks, but what seems strange is the need to type cast devices and optionality of some of the component props.

juriczech (changes requested):
Mostly nitpicks, but it seems strange to me that you're type casting the device type and also the optionality of some of these props seems strange to me...

suite-common/suite-utils/src/device.ts
vytick: This returns all devices that are not passphrase devices (instances) ordered. You can exclude current device so you can get list of all devices other than the current one.

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: So things slide one under the another. Devices should slide under the rounded borders of header and wallets should slide under the devices with rounded borders

suite-native/device-manager/src/components/PortfolioTrackerDeviceManagerContent.tsx
vytick: in DeviceManagerContent the children have it (DeviceList and WalletList)

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: very good spot!

suite-common/suite-utils/src/device.ts
PeKne: Will desktop ever potentionally use this? If not, move it to suite-native please.

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const physicalDevices = useSelector(selectPhysicalDevices);
+    const portfolioDevice = useSelector((state: DeviceRootState) =>
+        selectDeviceById(state, PORTFOLIO_TRACKER_DEVICE_ID),
+    );
+    const { utils } = useNativeStyles();
+    const opacity = useSharedValue(0);
+    const height = useSharedValue(0);
+    const [isShown, setIsShown] = useState(false);
+
+    const allDevices = [...physicalDevices, portfolioDevice].filter(
```
PeKne: could not be just used selector `selectDevices` that selects all devices from state?

PeKne: I agree with @juriczech that all this device selecting logic should be hidden a selector.

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,165 @@
...
+const SPACING = 8;
+const BUTTON_HEIGHT = 48;
+const SEPARATOR_HEIGHT = 18;
+const VERTICAL_PADDING = 16;
+
+const buttonWrapperStyle = prepareNativeStyle(utils => ({
+    paddingHorizontal: utils.spacings.medium,
+}));
+
+export const DeviceList = ({ isVisible, onSelectedDevice }: DeviceListProps) => {
```
PeKne: nit: `onDeviceSelect` or `onSelectDevice`

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -1,90 +1,88 @@
...
+                        deviceButtonState={isDeviceListVisible ? 'open' : 'closed'}
+                        onDeviceButtonTap={toggleIsDeviceListVisible}
+                    />
+                )
+            }
+            onClose={() => setIsDeviceListVisible(false)}
+        >
+            <VStack spacing="medium">
+                <DeviceList
+                    isVisible={isDeviceListVisible || isPortfolioTrackerDevice}
```
PeKne: It is counterintuitive that the `isVisible` prop might be true even if `isDeviceListVisible` would be false. The names of these two values are almost the same. Could you refactor this a little bit?

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -1,90 +1,88 @@
...
+                    onSelectedDevice={handleSelectDevice}
+                />
+                {!isPortfolioTrackerDevice && <WalletList onSelectedDevice={handleSelectDevice} />}
```
PeKne: two completely same condtional renders next to each other. Remove one of them and move its content to the other one

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -1,90 +1,88 @@
...
 
-type NavigationProp = StackToStackCompositeNavigationProps<
-    RootStackParamList,
-    RootStackRoutes.AppTabs,
-    RootStackParamList
->;
+const deviceButtonsStyle = prepareNativeStyle(utils => ({
+    width: '100%',
+    paddingHorizontal: utils.spacings.medium,
+    paddingBottom: utils.spacings.medium,
```
PeKne, line 28: nit:
```suggestion
    padding: utils.spacings.medium,
```

suite-native/device-manager/src/components/DeviceItemContent.tsx
```diff
@@ -1,82 +1,201 @@
...
-type DeviceItemIconProps = Pick<DeviceItemContentProps, 'deviceId'>;
-
-const DeviceItemIcon = ({ deviceId }: DeviceItemIconProps) => {
-    const activeColorScheme = useActiveColorScheme();
+type DeviceItemIconProps = Pick<DeviceItemContentProps, 'device' | 'iconSize'>;
 
-    // TODO: when we enable remember mode, icon representing disconnected device have to be handled.
-    const connectedDeviceIcon: IconName =
-        activeColorScheme === 'standard' ? 'trezorConnectedLight' : 'trezorConnectedDark';
+type SwitchContentType = {
```
PeKne:
```suggestion
type SwitchContent = {
```

suite-native/device-manager/src/components/DeviceModelIcon.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { DeviceModelInternal } from '@trezor/connect';
+
+const icons = {
+    T1B1: 'trezorT1B1',
+    T2T1: 'trezorT2T1',
+    T2B1: 'trezorT2B1',
+    T3T1: 'trezorT3T1',
+} as const satisfies Record<DeviceModelInternal, IconName>;
+
+type DeviceModelIconProps = { deviceModel?: DeviceModelInternal; size?: IconSize | number };
```
PeKne: Why is deviceModel optional? If it would be required, you chan remove the check of `undefined` few lines below.

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+            >
+                <HStack>
+                    <Icon
+                        name={device.useEmptyPassphrase ? 'standardWallet' : 'password'}
+                        size="mediumLarge"
+                    />
+                    <Text>{walletNameLabel}</Text>
+                </HStack>
+                {isSelectable && (
+                    <Radio value="" onPress={_ => onPress?.()} isChecked={isSelected} />
```
PeKne: nit: no inline functions please

suite-native/device-manager/src/components/DeviceManagerModal.tsx
```diff
@@ -1,47 +1,63 @@
...
+const deviceManagerModalWrapperStyle = prepareNativeStyle(utils => ({
+    backgroundColor: utils.colors.backgroundSurfaceElevation0,
+    borderBottomLeftRadius: MANAGER_MODAL_BOTTOM_RADIUS,
+    borderBottomRightRadius: MANAGER_MODAL_BOTTOM_RADIUS,
+}));
+
+const deviceSwitchWrapperStyle = prepareNativeStyle<{ insets: EdgeInsets }>(
     (utils, { insets }) => ({
-        paddingTop: Math.max(insets.top, utils.spacings.small),
+        paddingTop: insets.top + utils.spacings.large,
```
PeKne, line 35: Why did you decided to add these two values? The padding might be huge on some specific screens.

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,165 @@
...
+            opacity.value = withTiming(0, { duration: 300 });
+            timer = setTimeout(() => {
+                setIsShown(isVisible);
+            }, 300);
+        }
+
+        return () => clearTimeout(timer);
+    }, [height, isVisible, opacity, otherDevices.length]);
+
+    const listStyle = useAnimatedStyle(
```
PeKne: nit: 
```suggestion
    const listAnimatedStyle = useAnimatedStyle(
```

so it is clear that this style contains animation

PeKne: Also please divide this style into two. One simple style that contains all static values and the animated styles that contains dynamic animated values. And then merge them at he place where you apply it. It is a good practice while using reanimated.

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,165 @@
...
+            payload: { action: 'connectDeviceButton' },
+        });
+    };
+
+    //to hide/show the list with animation
+    useEffect(() => {
+        let timer: ReturnType<typeof setTimeout>;
+
+        if (isVisible) {
+            const h =
```
PeKne, line 108: this is very long expression. Could you split it into multiple for better readability please?

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,165 @@
...
+            setIsShown(isVisible);
+        } else {
+            height.value = withDelay(100, withTiming(0, { duration: 200 }));
+            opacity.value = withTiming(0, { duration: 300 });
+            timer = setTimeout(() => {
+                setIsShown(isVisible);
+            }, 300);
+        }
+
+        return () => clearTimeout(timer);
```
PeKne: No need for an extra timer. You can attach a callback to the previous `withTiming` call which will be executed right after the animation finishes. https://docs.swmansion.com/react-native-reanimated/docs/animations/withTiming/#callback-

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const { setIsDeviceManagerVisible } = useDeviceManager();
+    const device = useSelector(selectDevice);
+    const physicalDevices = useSelector(selectPhysicalDevices);
+    const portfolioDevice = useSelector((state: DeviceRootState) =>
+        selectDeviceById(state, PORTFOLIO_TRACKER_DEVICE_ID),
+    );
+    const { utils } = useNativeStyles();
+    const opacity = useSharedValue(0);
+    const height = useSharedValue(0);
+    const [isShown, setIsShown] = useState(false);
```
PeKne, line 75: Why is `isShown` needed at all? Does not it just reflect the prop value of `isVisible`?

PeKne (changes requested):
.

suite-native/device-manager/src/components/DeviceList.tsx
vytick: good spot. I don't know what I was thinking about here 🙈

suite-native/device-manager/src/components/DeviceManagerContent.tsx
vytick: paddingTop is not wanted here, therefore horizontal + bottom were used.

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: this is to have consistent background color. 
deviceManagerModalWrapperStyle sets it and deviceSwitchWrapperStyle layouts content and needs to have offset from the top

Screenshot 2024-05-10 at 10 51 58

suite-native/device-manager/src/components/DeviceManagerContent.tsx
PeKne: my mistake

suite-native/device-manager/src/components/DeviceList.tsx
vytick: This is needed so the height animation of removal finishes correctly and list stops rendering just after it finished

suite-native/device-manager/src/components/DeviceItemContent.tsx
vytick: Selecting by id can return different device than expected. This is because all passphrased devices have the same device id as the non passphrased device for which you opened the passphrase.

suite-native/device-manager/src/components/DeviceList.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceList.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceList.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/WalletItem.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceModelIcon.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceItemContent.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceManagerContent.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceManagerContent.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceList.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceItemContent.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceItemContent.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/WalletItem.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/WalletList.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/WalletList.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/AddHiddenWalletButton.tsx
vytick: 5825a46c34

suite-native/device-manager/src/components/AddHiddenWalletButton.tsx
vytick: 5825a46c34

suite-common/suite-utils/src/device.ts
vytick: 5825a46c34

suite-native/device-manager/src/components/DeviceList.tsx
vytick: 5825a46c34

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -866,3 +866,17 @@ export const selectDeviceState = (state: DeviceRootState) => {
...
 };
+
+export const selectDeviceInstances = memoize((state: DeviceRootState) => {
+    const device = selectDevice(state) as AcquiredDevice;
+    const allDevices = selectDevices(state);
+
+    return getDeviceInstances(device, allDevices);
+});
+
+export const selectDevicesWithoutSelectedAndInstances = memoize((state: DeviceRootState) => {
```
PeKne: nit: selectInstaclessUnselectedDevices?

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,158 @@
...
+    zIndex: 10,
+    ...utils.boxShadows.small,
+}));
+
+export const DeviceList = ({ isVisible, onSelectDevice }: DeviceListProps) => {
+    const { applyStyle } = useNativeStyles();
+    const navigation = useNavigation<NavigationProp>();
+    const { setIsDeviceManagerVisible } = useDeviceManager();
+    const device = useSelector(selectDevice);
+    const otherDevices = useSelector(selectDevicesWithoutSelectedAndInstances);
```
PeKne: `otherDevices` is not descriptive enough in context of this component. Could you name it so it is obvious what does this value hold?

suite-native/device-manager/src/types.ts
```diff
@@ -0,0 +1,21 @@
...
+    headerTextVariant?: TypographyStyle;
+    iconSize?: IconSize;
+    variant?: DeviceItemContentVariant;
+};
+
+export type SwitchContentProps = {
+    header: ReactNode;
+    subHeader?: ReactNode;
+    isPortfolioTrackerDevice: boolean;
+} & Omit<DeviceItemContentProps, 'variant'>;
```
PeKne:
```suggestion
} & Omit<DeviceItemContentProps, 'variant' | 'iconSize'>;
```

suite-native/device-manager/src/components/DeviceItem/SimpleDeviceItemContent.tsx
```diff
@@ -0,0 +1,61 @@
...
+    overflow: 'visible',
+}));
+
+export const SimpleDeviceItemContent = ({
+    device,
+    headerTextVariant,
+    isPortfolioLabelDisplayed,
+    header,
+    isPortfolioTrackerDevice,
+}: SwitchContentProps) => {
```
PeKne: `SwitchContentProps` define `subheader` prop that is never used here. It should not be possible to send a irelevant prop to a component. Please fix this.

suite-native/device-manager/src/components/DeviceManagerModal.tsx
PeKne: this PR does not fix it right? There is no change :D.

I was afraind of come devices that have some "dynamic island-like" feauture, taht it gonna take too much space on the top. Anyway lets keep it, if you are confident it gonna work 👍.

suite-native/device-manager/src/components/DeviceList.tsx
PeKne: In that case, I believe that you should use [entering/exiting](https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/entering-exiting-animations) animation and remove the `isShown` state. 

It is even possible to create a custom entering/exiting animation. [See this.](https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/custom-animations)

suite-native/device-manager/src/components/DeviceManagerModal.tsx
PeKne: The background color does not behave as expected on my localhost :/. 

Screenshot 2024-05-13 at 9 18 11

suite-native/device-manager/src/components/DeviceList.tsx
vytick: I've tried that. But it is suboptimal, because the element size goes away immediately and walletList (that is below) jumps up right away not waiting for animation to finish.

PeKne: @juriczech Did not you have simillar problem? Were you able to solve it?

suite-common/wallet-core/src/device/deviceReducer.ts
vytick: 6cdafc4705

suite-native/device-manager/src/components/DeviceItem/SimpleDeviceItemContent.tsx
vytick: 4650a6feb3

suite-native/device-manager/src/types.ts
vytick: 4650a6feb3

suite-native/device-manager/src/components/DeviceList.tsx
vytick: 4a27b715ac

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -1,90 +1,91 @@
...
+                <DeviceList
+                    isVisible={isChangeDeviceRequested || isPortfolioTrackerDevice}
+                    onSelectDevice={handleSelectDevice}
+                />
+                {!isPortfolioTrackerDevice && (
+                    <>
+                        <WalletList onSelectDevice={handleSelectDevice} />
+                        <HStack style={applyStyle(deviceButtonsStyle)}>
+                            <DeviceInfoButton />
+                            {isPassphraseFeatureEnabled && !isDiscoveryActive && (
```
matejkriz: We can't add passphrase if device is not connected. Ideally I would imagine to keep the button here but if user click it without connected device, show some reasonable warning guiding user to connect Trezor... Hiding this button would be of course easier solution, not sure how much worse from UX perspective.

matejkriz: Why the button is hidden while discovery is active? Why not to stop discovery if user want to open passhprase wallet?

vytick: That is what I was told by @juriczech and made sense to me. 

And I am not sure if the discovery will be triggered again if some accounts were already discovered and device was remembered. How does this work? 🤔

juriczech: I think it could work since we can jump the device calls queue and then discovery should continue on the previous device, but I haven't tested this yet. Anyway, I would go with this as a first step and revisit this in separate issue with passphrase finalising issues if you're okay with that

vytick: i'll hide this when device is disconnected and add issue to resolve this later after we discuss the best solution with product

suite-native/device-manager/src/components/DeviceList.tsx
juriczech: Not sure. @vytick if you'd like, we can have a call and you could give me more context so I'd know for sure but it's difficult to say if it's the same issue..

suite-native/device-manager/src/components/DeviceManagerContent.tsx
matejkriz: sure thing, let's revisit as part of passhprase https://github.com/trezor/trezor-suite/issues/12361

suite-native/device-manager/src/components/DeviceItemContent.tsx
matejkriz: Oh, we should not select by device.id in this case, we should use device.state which is unique per wallet

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { selectDevice } from '@suite-common/wallet-core';
+
+const walletItemStyle = prepareNativeStyle<{ isSelected: boolean }>((utils, { isSelected }) => ({
+    paddingHorizontal: utils.spacings.medium,
+    justifyContent: 'space-between',
+    alignItems: 'center',
+    height: 60,
+    gap: 12,
+    backgroundColor: utils.colors.backgroundSurfaceElevation1,
+    borderWidth: 1,
```
matejkriz: nit: utils.borders.widths.small

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+    borderRadius: 12,
+    borderColor: utils.colors.borderElevation1,
+    extend: {
+        condition: isSelected,
+        style: { borderWidth: 2, borderColor: utils.colors.borderSecondary },
```
matejkriz: nit: utils.borders.widths.large

vytick: 736ca59fbd

suite-native/device-manager/src/components/WalletItem.tsx
vytick: 736ca59fbd

suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { TrezorDevice } from '@suite-common/suite-types';
+
+import { DeviceModelIcon } from '../DeviceModelIcon';
+
+type DeviceItemIconProps = {
+    deviceId: TrezorDevice['id'];
+    iconSize: IconSize;
+};
+
+export const DeviceItemIcon = ({ deviceId, iconSize }: DeviceItemIconProps) => {
```
matejkriz, line 18: There used to be trezor icon if no device connected/remembered and [according to the design](https://www.figma.com/file/mmDnYtctfjnreWhTjLfeQH/Passphrase?type=design&node-id=2382-11343&mode=design&t=pYULMai1iFohCY72-4) it should still be there, right?

image


vs 

image

matejkriz: maybe I'm locally in some bad state, but this is how it looks to me with no device
image
but I would expect 
image

vytick: This is according to this design:
https://www.figma.com/file/mmDnYtctfjnreWhTjLfeQH/Passphrase?type=design&node-id=2325-66709&mode=design&t=DYCejprDAWSvtNQ4-0

vytick: The icon should be there. i'll check where it went

vytick: 1db74c5c20

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: That is a Status bar. I'll check what can I do about that

vytick: c784246629

suite-native/device-manager/src/components/DeviceItemContent.tsx
vytick: b11b9d623c

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -866,3 +866,22 @@ export const selectDeviceState = (state: DeviceRootState) => {
 
     return device?.state ?? null;
 };
+
+export const selectDeviceInstances = memoize((state: DeviceRootState) => {
```
juriczech, line 862: Just curious. Is there specific reason to memoize or not?

suite-native/device-manager/src/components/DeviceAction.tsx
```diff
@@ -0,0 +1,22 @@
...
+    alignItems: 'center',
+    height: 44,
+    gap: utils.spacings.small,
+    backgroundColor: utils.colors.backgroundSurfaceElevation1,
+    borderWidth: utils.borders.widths.small,
+    borderRadius: 12,
+    borderColor: utils.colors.borderElevation1,
+}));
+
+export const DeviceAction = ({ children }: { children: ReactNode }) => {
```
juriczech: Hmm, I wonder if the Pressable could be part of this component as well..

suite-native/device-manager/src/components/DeviceInfoButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+}));
+
+export const DeviceInfoButton = () => {
+    const { applyStyle } = useNativeStyles();
+    const navigation = useNavigation<NavigationProp>();
+    const { setIsDeviceManagerVisible } = useDeviceManager();
+    const [isPassphraseFeatureEnabled] = useFeatureFlag(FeatureFlag.IsPassphraseEnabled);
+    const selectedDevice = useSelector(selectDevice);
+    const isDiscoveryActive = useSelector(selectIsDeviceDiscoveryActive);
+    const showWithFlex =
```
juriczech: The name is not indicating very well why these properties are dependant on the flex style props. Could you elaborate more so we can figure out a better / more descriptive name so it's more clear when reading this component?

suite-native/device-manager/src/components/DeviceItem/DeviceItem.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export const DeviceItem = ({ id: deviceItemId, onPress }: DeviceItemProps) => {
+    const { applyStyle } = useNativeStyles();
+
+    const device = useSelector((state: DeviceRootState) => selectDeviceById(state, deviceItemId));
+
+    return (
+        <Pressable onPress={onPress}>
+            <HStack style={applyStyle(deviceItemWrapperStyle)}>
+                <DeviceItemContent deviceState={device?.state} />
```
juriczech: Remove selector for device as we don't need it here, send device state to this component and use that as prop here please. Selecting device by id seems redundant...

suite-native/device-manager/src/components/DeviceItem/DeviceItem.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { HStack } from '@suite-native/atoms';
+import { DeviceRootState, selectDeviceById } from '@suite-common/wallet-core';
+import { Icon } from '@suite-common/icons';
+import { TrezorDevice } from '@suite-common/suite-types';
+
+import { DeviceItemContent } from './DeviceItemContent';
+
+type DeviceItemProps = {
+    id: TrezorDevice['id'];
+    onPress?: () => void;
```
juriczech: Why is `onPress` optional? You should always have a press function here if this component gets rendered

suite-native/device-manager/src/components/DeviceItem/DeviceItemContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+    selectDeviceByState,
+    DeviceRootState,
+    PORTFOLIO_TRACKER_DEVICE_ID,
+} from '@suite-common/wallet-core';
+import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
+
+import { DeviceItemIcon } from './DeviceItemIcon';
+import { SimpleDeviceItemContent } from './SimpleDeviceItemContent';
+import { WalletDetailDeviceItemContent } from './WalletDetailDeviceItemContent';
+import { DeviceItemContentProps } from '../../types';
```
juriczech: Please move the props to the component. It should be defined in the file where it's used...

suite-native/device-manager/src/components/DeviceItem/DeviceItemContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+    isPortfolioLabelDisplayed = true,
+    headerTextVariant = 'body',
+    iconSize = 'large',
+    variant = 'simple',
+}: DeviceItemContentProps) => {
+    const { translate } = useTranslate();
+    const { applyStyle } = useNativeStyles();
+
+    const device = useSelector((state: DeviceRootState) =>
+        deviceState !== undefined ? selectDeviceByState(state, deviceState) : undefined,
```
juriczech: Make deviceState required parameter

suite-native/device-manager/src/components/DeviceItem/DeviceItemContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+    ) : (
+        <Translation
+            id="deviceManager.wallet.defaultPassphrase"
+            values={{ index: device?.walletNumber }}
+        />
+    );
+
+    return (
+        <HStack alignItems="center" spacing="medium" style={applyStyle(contentWrapperStyle)}>
+            <DeviceItemIcon deviceId={device?.id} iconSize={iconSize} />
```
juriczech:
```suggestion
            <DeviceItemIcon deviceId={deviceState} iconSize={iconSize} />
```

suite-native/device-manager/src/components/DeviceItem/DeviceItemContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <Translation id="deviceManager.wallet.standard" />
+    ) : (
+        <Translation
+            id="deviceManager.wallet.defaultPassphrase"
+            values={{ index: device?.walletNumber }}
+        />
+    );
+
+    return (
+        <HStack alignItems="center" spacing="medium" style={applyStyle(contentWrapperStyle)}>
```
juriczech: Please apply these style props in the style or let's just ignore this in all reviews :)

suite-native/device-manager/src/components/DeviceItem/SimpleDeviceItemContent.tsx
```diff
@@ -0,0 +1,69 @@
...
+export const SimpleDeviceItemContent = ({
+    deviceState,
+    headerTextVariant,
+    isPortfolioLabelDisplayed,
+    header,
+    isPortfolioTrackerDevice,
+}: SimpleDeviceItemContentProps) => {
+    const { applyStyle } = useNativeStyles();
+    const device = useSelector((state: DeviceRootState) =>
+        deviceState !== undefined ? selectDeviceByState(state, deviceState) : undefined,
```
juriczech: Make deviceState required

suite-native/device-manager/src/components/DeviceItem/SimpleDeviceItemContent.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <>
+            <Text
+                variant={headerTextVariant}
+                ellipsizeMode="tail"
+                numberOfLines={1}
+                style={applyStyle(headerStyle)}
+            >
+                {header}
+            </Text>
+            {device?.id && (
```
juriczech: Handle optional device above JSX. `if (!device) return null` or return `<ErrorMessage {...} />`

suite-native/device-manager/src/components/DeviceItem/WalletDetailDeviceItemContent.tsx
```diff
@@ -0,0 +1,67 @@
...
+    deviceState,
+    headerTextVariant,
+    isPortfolioLabelDisplayed,
+    header,
+    subHeader,
+    isPortfolioTrackerDevice,
+}: SwitchContentProps) => {
+    const { applyStyle } = useNativeStyles();
+    const device = useSelector((state: DeviceRootState) =>
+        deviceState !== undefined ? selectDeviceByState(state, deviceState) : undefined,
```
juriczech: Device state should be required

suite-native/device-manager/src/components/DeviceItem/WalletDetailDeviceItemContent.tsx
```diff
@@ -0,0 +1,67 @@
...
+                <Text
+                    variant={headerTextVariant}
+                    ellipsizeMode="tail"
+                    numberOfLines={1}
+                    style={applyStyle(headerTextStyle)}
+                >
+                    {header}
+                </Text>
+                {!isPortfolioTrackerDevice && device && (
+                    <ConnectionDot isConnected={!!device?.connected} />
```
juriczech: selectIsDeviceConntected, also handle optional device above render, same for the optional few lines below

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        <Box style={applyStyle(buttonWrapperStyle)}>
+                            <Button colorScheme="tertiaryElevation1" onPress={handleConnectDevice}>
+                                <Translation id="deviceManager.connectButton.another" />
+                            </Button>
+                        </Box>
+                    </>
+                )}
+            </VStack>
+        </Animated.View>
+    ) : null;
```
juriczech: nitpick: I know it's not mandatory, but I think it's much more readable to do something like `if (!isShown) return null` and then render the component.

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -1,90 +1,93 @@
...
+                    onSelectDevice={handleSelectDevice}
+                />
+                {!isPortfolioTrackerDevice && (
+                    <>
+                        <WalletList onSelectDevice={handleSelectDevice} />
+                        <HStack style={applyStyle(deviceButtonsStyle)}>
+                            <DeviceInfoButton />
+                            {isPassphraseFeatureEnabled &&
+                                !isDiscoveryActive &&
+                                !!device?.connected && <AddHiddenWalletButton />}
```
juriczech: Device shouldn't be undefined here.

suite-native/device-manager/src/components/DeviceManagerModal.tsx
```diff
@@ -1,47 +1,62 @@
...
+import { selectDevice } from '@suite-common/wallet-core';
 
-import { DeviceSwitch } from './DeviceSwitch';
 import { useDeviceManager } from '../hooks/useDeviceManager';
+import { DeviceItemContent } from './DeviceItem/DeviceItemContent';
 
 type DeviceManagerModalProps = {
     children: ReactNode;
+    customSwitchRightView?: ReactNode;
+    onClose?: () => void;
```
juriczech, line 17: Should be required

suite-native/device-manager/src/components/DeviceSwitch.tsx
```diff
@@ -46,7 +47,8 @@ export const DeviceSwitch = () => {
     const areAllDevicesDisconnectedOrAccountless = useSelector(
         selectAreAllDevicesDisconnectedOrAccountless,
     );
-    const deviceId = useSelector(selectDeviceId);
+    const device = useSelector(selectDevice);
```
juriczech: selectDeviceState, make sure it's defined in JSX and send further

suite-native/device-manager/src/components/DevicesToggleButton.tsx
```diff
@@ -0,0 +1,26 @@
+import { Button, Text } from '@suite-native/atoms';
+import { Translation } from '@suite-native/intl';
+
+type DeviceButtonState = 'open' | 'closed';
+
+type DevicesToggleButtonProps = {
+    deviceButtonState: DeviceButtonState;
+    onDeviceButtonTap?: () => void;
```
juriczech: Make this required

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+    },
+}));
+
+type WalletItemProps = {
+    device: TrezorDevice;
+    isSelectable?: boolean;
+    onPress: () => void;
+};
+
+export const WalletItem = ({ device, isSelectable = true, onPress }: WalletItemProps) => {
```
juriczech: Optional parameters with default value should be last items in the spread according to code style

juriczech (changes requested):

suite-common/wallet-core/src/device/deviceReducer.ts
vytick: getDeviceInstances returns new array that would cause rerenders

suite-native/device-manager/src/components/DeviceAction.tsx
vytick: 🤔 it might. I'll think about how to do that

suite-native/device-manager/src/components/DeviceInfoButton.tsx
vytick: For some occasions we want this element be full width and for some no. I might redo this if I put pressable into the DeviceAction element

suite-native/device-manager/src/components/DeviceItem/DeviceItem.tsx
vytick: good spot

suite-native/device-manager/src/components/DeviceInfoButton.tsx
juriczech: I understand, but IMO it should be clearer why this should be affected. Is it because sometimes it's rendered side by side with something etc. It's a nit, but when I would come to refactor this, it would be a pain to find out what this means

suite-native/device-manager/src/components/DeviceItem/DeviceItemContent.tsx
vytick: icon is not based on certain wallet but instead physical model, so there is no need to pass state instead of id which I also need here in this component.

juriczech: Okay, then please just get rid of the `?` in `device?.id`

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: it feels better to handle optionality here than having to write 
`onClose={() => {}}>`
in `PortfolioTrackerDeviceManagerContent`

suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
matejkriz: Thanks for the icon fix, but the expanded state of device switcher is still not according to the design with no device 

This is how it looks in current state:
image

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -0,0 +1,176 @@
...
+    };
+
+    //to hide/show the list with animation
+    useEffect(() => {
+        if (isVisible) {
+            const hasNotSelectedInstancelessDevices = notSelectedInstancelessDevices.length > 0;
+
+            const paddingsHeight = VERTICAL_PADDING * 2;
+            const otherDevicesHeight =
+                notSelectedInstancelessDevices.length * (ITEM_HEIGHT + ITEMS_SPACING);
```
juriczech: nit: I don't see any benefit using these consts that are hardcoded when you can use theming sizes. 

```suggestion
                notSelectedInstancelessDevices.length * (ITEM_HEIGHT + utils.spacings.small);
```

juriczech: Most of them are only used here and it's not tied to specific paddings or other calculations, so I'd rather do just theming sizes here, WDYT?

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -1,90 +1,98 @@
...
+const deviceButtonsStyle = prepareNativeStyle(utils => ({
+    width: '100%',
+    paddingHorizontal: utils.spacings.medium,
+    paddingBottom: utils.spacings.medium,
+}));
 
 export const DeviceManagerContent = () => {
-    const navigation = useNavigation<NavigationProp>();
+    const { applyStyle } = useNativeStyles();
+    const [isChangeDeviceRequested, setIsChangeDeviceRequested] = useState(false);
```
juriczech, line 33: nit: How about this naming? Feel free to resolve if you think the currenct name is better

```suggestion
    const [isDeviceListDisplayed, setIsDeviceListDisplayed] = useState(false);
```

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const WalletList = ({ onSelectDevice }: WalletListProps) => {
+    const wallets = useSelector(selectDeviceInstances);
+
+    return (
+        <VStack paddingHorizontal="medium" paddingTop="large">
+            {wallets.map(device => {
+                if (!device.state) {
+                    return null;
+                }
```
juriczech, line 21: Not all remembered wallets will have `device.state`. Am I missing something? 🤔

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -0,0 +1,34 @@
...
+type WalletListProps = {
+    onSelectDevice: (device: TrezorDevice) => void;
+};
+
+export const WalletList = ({ onSelectDevice }: WalletListProps) => {
+    const wallets = useSelector(selectDeviceInstances);
+
+    return (
+        <VStack paddingHorizontal="medium" paddingTop="large">
+            {wallets.map(device => {
```
juriczech:
```suggestion
            {devices.map(device => {
```

or 

```suggestion
            {wallets.map(wallet => {
```

suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
vytick: I probably dont understand what state is this. There is always a device. There is either portfolio device selected or a real device.

suite-native/device-manager/src/components/WalletList.tsx
vytick: device.state is string | undefined

I dont expect state to be undefined here, but for sure (because type allows it) I rather handle this state

suite-native/device-manager/src/components/DeviceManagerContent.tsx
vytick: the list might be forced to be visible in case it is portfolio tracker and Peta suggested not to use the naming you propose. I'll keep this one here

suite-native/device-manager/src/components/WalletList.tsx
vytick: I'll name everything device

juriczech (approved):
THANKS FOR RESOLVING ALL OF THE COMMENTS 🎉 

Tested on emulator on android and works well with remember and passphrase as well. 

Really nice animation, well done. 

Next time, it would be better for review and development to split into smaller PR's but maybe it was unexpected to be this big 😮

suite-native/device-manager/src/components/DeviceManagerModal.tsx
PeKne: thanks. [statusBarTranslucent](https://reactnative.dev/docs/modal#statusbartranslucent-android) is defined only for Android os. Did you test this on iOS?

suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
matejkriz: @vytick try it on Android...

vytick: Gotcha. I have updated the visual partially in 21677d7493 (trezor logo is even in opened switcher now, but this is where I can get. 


Screenshot 2024-05-14 at 22 38 46


The design does not technically make a sense, since even in this very moment, portfolio tracker device is already selected, we just show it differently in the switcher because there are no accounts. This means that I dont know what should happen if user clicks on portfolio device there. Lets talk this tmrw.

Screenshot 2024-05-14 at 22 41 23

matejkriz: design changed 👍 resolving

matejkriz:
RESOLVED
~~I just spotted missing padding below `Connect another device button` in case of multiple devices~~

image

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: yes, you can check screenshots that are from iOS

PeKne (approved):
👏 Thanks for addressing my comments.

vytick:
@matejkriz Good catch with layout glitch! fixed in a5016a6d9c

PR #12313, opened 2024-05-07
Connect/update messages haptic/brightness

mroz22:
updating messages to support https://github.com/trezor/trezor-firmware/pull/3802/files 

but the firmware PR doesn't appear to be really finished, so draft only.


MiroslavProchazka:
Will resolve https://github.com/trezor/trezor-suite/issues/12315

PR #12316, opened 2024-05-08
chore(suite-native): button has ReactElement not only icon

vytick:
Make Button to accept not only IconName but also a ReactElement for both sides.

This will let us insert animated elements and I will be able to use that in Device switcher [Change button animation](https://www.figma.com/proto/mmDnYtctfjnreWhTjLfeQH/Passphrase?page-id=0%3A1&type=design&node-id=2146-55968&viewport=5976%2C4395%2C1.31&t=8i3EiJVzLmaibGWX-1&scaling=min-zoom&starting-point-node-id=2146%3A55968&show-proto-sidebar=1&mode=design&fuid=1293095861312443248).

Example of custom and also animated element:

https://github.com/trezor/trezor-suite/assets/2011829/1253af02-a3d9-4e48-920a-e9cb96cac62e

### DIFF
--- a/suite-native/atoms/src/Button/Button.tsx
+++ b/suite-native/atoms/src/Button/Button.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useState } from 'react';
+import { ReactElement, ReactNode, useState } from 'react';
 import { Pressable, PressableProps } from 'react-native';
 import Animated from 'react-native-reanimated';
 
@@ -6,12 +6,16 @@ import { MergeExclusive } from 'type-fest';
 
 import { Color, TypographyStyle, nativeSpacings } from '@trezor/theme';
 import { NativeStyleObject, prepareNativeStyle, useNativeStyles } from '@trezor/styles';
-import { Icon, IconColor, IconName, IconSize } from '@suite-common/icons';
+import { Icon, IconColor, IconName, IconSize, icons } from '@suite-common/icons';
 
 import { Text } from '../Text';
 import { useButtonPressAnimatedStyle } from './useButtonPressAnimatedStyle';
 import { TestProps } from '../types';
 import { HStack } from '../Stack';
+
+// Using ReactElement instead of ReactNode to exclude string and have type check on IconName
+// and also because string needs to be rendered in the <Text> element anyway
+export type ButtonAccessory = IconName | ReactElement;
 
 export type ButtonSize = 'extraSmall' | 'small' | 'medium' | 'large';
 export type ButtonColorScheme =
@@ -35,13 +39,19 @@ export type ButtonProps = Omit<PressableProps, 'style' | 'onPressIn' | 'onPressO
     size?: ButtonSize;
     style?: NativeStyleObject;
     isDisabled?: boolean;
-} & MergeExclusive<{ iconLeft?: IconName }, { iconRight?: IconName }> &
+} & MergeExclusive<{ viewLeft?: ButtonAccessory }, { viewRight?: ButtonAccessory }> &
     TestProps;
 
 type ButtonIconProps = {
     iconName: IconName;
-    color: IconColor;
-    buttonSize: ButtonSize;
+    color?: IconColor;
+    size?: ButtonSize;
+};
+
+type ButtonAccessoryViewProps = {
+    element: ButtonAccessory;
+    iconColor?: IconColor;
+    iconSize?: ButtonSize;
 };
 
 type BaseButtonColorScheme = {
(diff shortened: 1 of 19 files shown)


suite-native/atoms/src/Button/Button.tsx
```diff
@@ -35,7 +40,7 @@ export type ButtonProps = Omit<PressableProps, 'style' | 'onPressIn' | 'onPressO
     size?: ButtonSize;
     style?: NativeStyleObject;
     isDisabled?: boolean;
-} & MergeExclusive<{ iconLeft?: IconName }, { iconRight?: IconName }> &
+} & MergeExclusive<{ viewLeft?: ButtonAccessoryType }, { viewRight?: ButtonAccessoryType }> &
```
PeKne: Why did you change the name fro icon to view? the icon seems to be more explanatory to me. We still want to send there only icons (even though they can be animated now). 

But maybe this is just my personal preference.

suite-native/atoms/src/Button/Button.tsx
```diff
@@ -1,18 +1,23 @@
...
+import { Icon, IconColor, IconName, IconSize, icons } from '@suite-common/icons';
 
 import { Text } from '../Text';
 import { useButtonPressAnimatedStyle } from './useButtonPressAnimatedStyle';
 import { TestProps } from '../types';
 import { HStack } from '../Stack';
 
+// Using ReactElement instead of ReactNode to exclude string and have type check on IconName
+// and also because string needs to be rendered in the <Text> element anyway
+export type ButtonAccessoryType = IconName | ReactElement | undefined | null;
```
PeKne: Why is `undefined` and `null` needed here?

PeKne: I think that type suffix is redundant here.
```suggestion
export type ButtonAccessory = IconName | ReactElement | undefined | null;
```

suite-native/atoms/src/Button/Button.tsx
```diff
@@ -44,6 +49,12 @@ type ButtonIconProps = {
     buttonSize: ButtonSize;
 };
 
+type ButtonAccessoryViewType = {
```
PeKne:
```suggestion
type ButtonAccessoryView = {
```

suite-native/atoms/src/Button/Button.tsx
```diff
@@ -278,11 +307,15 @@ export const Button = ({
                 ]}
             >
                 <HStack alignItems="center">
-                    {iconLeft && icon}
+                    <ButtonAccessoryView element={viewLeft} iconColor={iconColor} iconSize={size} />
```
PeKne: Why was the conditional rendering removed? IMO it is more readable like that.

suite-native/atoms/src/Button/Button.tsx
```diff
@@ -227,9 +238,32 @@ export const ButtonIcon = ({ iconName, color, buttonSize }: ButtonIconProps) =>
...
+export const ButtonAccessoryView = ({
+    element,
+    iconColor = 'iconDefault',
+    iconSize = 'medium',
+}: ButtonAccessoryViewType) => {
+    if (isIconName(element)) {
+        if (!iconColor || !iconSize) return null;
+
+        return (
+            <ButtonIcon iconName={element as IconName} color={iconColor} buttonSize={iconSize} />
```
PeKne: Change the `isIconName` to a type guard (`const isIconName = (value: ButtonAccessoryType) : is IconName => ...`) more here: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards. 

Then you can remove the type assertion here.

```suggestion
            <ButtonIcon iconName={element} color={iconColor} buttonSize={iconSize} />
```

suite-native/atoms/src/Button/Button.tsx
```diff
@@ -227,9 +238,32 @@ export const ButtonIcon = ({ iconName, color, buttonSize }: ButtonIconProps) =>
...
+
+// ButtonAccessoryView renders either a ButtonIcon or a provided custom element
+// iconColor and iconSize are only used when element is an IconName
+export const ButtonAccessoryView = ({
+    element,
+    iconColor = 'iconDefault',
+    iconSize = 'medium',
+}: ButtonAccessoryViewType) => {
+    if (isIconName(element)) {
+        if (!iconColor || !iconSize) return null;
```
PeKne: This condition will be never used since the iconColor and IconSize have default values.

suite-native/atoms/src/Button/Button.tsx
```diff
@@ -227,9 +238,32 @@ export const ButtonIcon = ({ iconName, color, buttonSize }: ButtonIconProps) =>
...
+export const isIconName = (element: ButtonAccessoryType) =>
+    G.isNotNullable(element) &&
+    typeof element === 'string' &&
+    icons[element as IconName] !== undefined;
+
+// ButtonAccessoryView renders either a ButtonIcon or a provided custom element
+// iconColor and iconSize are only used when element is an IconName
+export const ButtonAccessoryView = ({
+    element,
+    iconColor = 'iconDefault',
```
PeKne: What about "hiding" these defaults inside of IconButton component? It should be its responsible IMO. Just idea.

PeKne (changes requested):

suite-native/atoms/src/Button/Button.tsx
vytick: I changed the name because we can put anything there now, not only icon

suite-native/atoms/src/Button/Button.tsx
vytick: I put it there to handel undefined | null inside. But i will change it to conditional rendering as you suggested for the readibility

vytick: 0c52c10957

suite-native/atoms/src/Button/Button.tsx
vytick: 0c52c10957

suite-native/atoms/src/Button/Button.tsx
vytick: 0c52c10957

suite-native/atoms/src/Button/Button.tsx
vytick: 0c52c10957

suite-native/atoms/src/Button/Button.tsx
vytick: 0c52c10957

suite-native/atoms/src/Button/Button.tsx
vytick: 0c52c10957

suite-native/atoms/src/Button/Button.tsx
PeKne: the `Type` suffix is still there.

vytick: good point. updated in 705cf82150

PeKne (approved):
Thanks for addressing my request 🙌

PR #12321, opened 2024-05-09
fix(suite): device rotating image

tomasklim:
## Description

- fix missing rotation of new device in wallet select modal

## Screenshots:

https://github.com/trezor/trezor-suite/assets/33235762/c37de18f-95c1-434c-9099-449f31f0378c

### DIFF
--- a/packages/components/src/components/ConfirmOnDevice/RotateDeviceImage.tsx
+++ b/packages/components/src/components/ConfirmOnDevice/RotateDeviceImage.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { DeviceAnimation } from '../animations/DeviceAnimation';
 import { DeviceModelInternal } from '@trezor/connect';
 import { Image } from '../Image/Image';
+import styled from 'styled-components';
 
 interface DeviceImageProps {
     deviceModel?: DeviceModelInternal;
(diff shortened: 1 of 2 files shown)


packages/components/src/components/ConfirmOnDevice/RotateDeviceImage.tsx
```diff
@@ -11,6 +12,11 @@ interface DeviceImageProps {
     animationWidth?: string;
 }
 
+const StyledImage = styled(Image)`
+    /* do not apply the darkening filter in dark mode on device images */
```
jvaclavik, line 16: Is the comment correct? AFAIK we don't handle dark mode here

jvaclavik (approved):

packages/components/src/components/ConfirmOnDevice/RotateDeviceImage.tsx
tomasklim: yes, in dark mode there is some filter over images
Screenshot 2024-05-09 at 14 52 23

tomasklim: it is undefined for light mode

Issue #12317, opened 2024-05-08
Ugly Backup failed modal


bosomt:
**Describe the bug**
Ugly `Backup failed` modal

**Steps to reproduce:**
1. Connect device that needs backup
2. Navigate to `Settings/Device`
3. Start backup
4. Disconnect device

**Info:**
 - Suite version: desktop 24.5.1 (a94f65e0a34ae30fc78759aabec64c0f99055a1c)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.5.1 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.7.0 regular (revision null)
 - Transport: BridgeTransport 2.0.33

**Screenshots:**
<img width="690" alt="Screenshot 2024-05-08 at 15 50 52" src="https://github.com/trezor/trezor-suite/assets/31506317/384772c8-a52a-4510-8d27-e63dae06a181">

MiroslavProchazka:
Resolved by https://github.com/trezor/trezor-suite/pull/12325

STew790:
**QA OK**
As per https://github.com/trezor/trezor-suite/pull/12325#issuecomment-2102926320

**Info:**
 - Suite version: desktop 24.5.2 (1ca8b8063fd7028d1fe242db41caa31c0a1f7d5a)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.5.2 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: Trezor T2T1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)
 - Transport: BridgeTransport 2.0.33

PR #12325, opened 2024-05-09
fix(suite): Fix backup modal

jvaclavik:
## Description

not great not terrible 🫣

## Related Issue

Resolve 

## Screenshots:
image

### DIFF
--- a/packages/components/src/components/modals/Modal/Modal.tsx
+++ b/packages/components/src/components/modals/Modal/Modal.tsx
@@ -141,6 +141,7 @@ const Body = styled.div<{ $isWithoutTopPadding: boolean }>`
     padding-top: ${({ $isWithoutTopPadding }) => $isWithoutTopPadding && 0};
     display: flex;
     flex-direction: column;
+    flex: 1;
 `;
 
 const Description = styled.div`
--- a/packages/suite/src/views/backup/index.tsx
+++ b/packages/suite/src/views/backup/index.tsx
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 
-import { Paragraph, Button, Image } from '@trezor/components';
+import { Paragraph, Button, Image, Flex } from '@trezor/components';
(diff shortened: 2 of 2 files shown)


MiroslavProchazka:
Resolves [#12317](https://github.com/trezor/trezor-suite/issues/12317)

komret (approved):
It's fine, but I have some suggestions regarding `Flex`.

1. I don't think it should be exported as it is equivalent to `Columns` and there should ideally only be one way to do stuff.
2. The names are a bit confusing and it takes me some time to wrap my head around it every time. I'd rename `Columns` -> `Row`, `Rows` -> `Column`.

jvaclavik:
> It's fine, but I have some suggestions regarding `Flex`.
> 
> 1. I don't think it should be exported as it is equivalent to `Columns` and there should ideally only be one way to do stuff.
> 2. The names are a bit confusing and it takes me some time to wrap my head around it every time. I'd rename `Columns` -> `Row`, `Rows` -> `Column`.

1. We can export just Row and Column and theoretically not export Flex, this could fix this issue. But sometimes you don't care about the direction (like here - centering the content). WDYT?
2. Agree, it's probably more clear if we change it this way in singular. Will fix it in different PR

komret:
I think you care about direction so that you know whether you should `justify-content` or `align-items` to center it so I would not export `Flex`.

bosomt:
Looks like this now

image


**Info:**
 - Suite version: web 24.5.0 (115c7183cf7da0011fc6f8e594f279c28656fac0)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:125.0) Gecko/20100101 Firefox/125.0
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)

PR #12334, opened 2024-05-10
Connect fw deprecations

marekrjpolak:
## Description

Require firmware newer than 1.8.1/2.1.0, so some deprecated code branches could be removed.


mroz22 (approved):
👍

PR #12336, opened 2024-05-10
chore(github): rename workflows to match convention

karliatto:
## Description

Rename workflows to match convention and change some titles to be more standard.


.github/workflows/test-connect.yml
```diff
@@ -1,4 +1,4 @@
-name: "[Test] connect"
+name: "[Test] connect unit"
```
mroz22: these tests arent pure unit tests. I would describe them as e2e tests of connect core.

mroz22 (approved):
whatever brings us closer to a single convention I wholeheartedly approve

vdovhanych (approved):

mroz22 (approved):

PR #12338, opened 2024-05-10
feat(suite-native): reauthorize device

### LINKED ISSUE #12207: Passphrase: Handle "Start over" and "retry
Implement start over and retry functionality

image
image

juriczech:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12207

### DIFF
--- a/suite-common/wallet-core/src/device/deviceThunks.ts
+++ b/suite-common/wallet-core/src/device/deviceThunks.ts
@@ -275,7 +275,12 @@ export const acquireDevice = createThunk(
  */
 export const authorizeDevice = createThunk(
     `${DEVICE_MODULE_PREFIX}/authorizeDevice`,
-    async (_, { dispatch, getState, extra }): Promise<boolean> => {
+    async (
+        { shouldIgnoreDeviceState }: { shouldIgnoreDeviceState: boolean } | undefined = {
+            shouldIgnoreDeviceState: false,
+        },
+        { dispatch, getState, extra },
+    ): Promise<boolean> => {
         const {
             selectors: { selectCheckFirmwareAuthenticity },
             actions: { openModal },
@@ -287,7 +292,9 @@ export const authorizeDevice = createThunk(
         const isDeviceReady =
             device.connected &&
             device.features &&
-            !device.state &&
+            // Should ignore device state serves as a variant to call "reauthorize" device. For example in passphrase mode
+            // mobile has retry button which starts passphrase flow on the same device instance to override device state.
+            (!device.state || shouldIgnoreDeviceState) &&
             device.mode === 'normal' &&
             device.firmware !== 'required';
         if (!isDeviceReady) return false;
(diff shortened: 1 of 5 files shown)


suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -345,6 +352,13 @@ export const authorizeDevice = createThunk(
     },
 );
 
+export const reAuthorizeDevice = createThunk(
```
PeKne:
```suggestion
export const reAuthorizeDeviceThunk = createThunk(
```

suite-native/module-passphrase/src/passphraseThunks.ts
```diff
@@ -56,3 +57,14 @@ export const verifyPassphraseOnEmptyWalletThunk = createThunk(
         return fulfillWithValue(true);
     },
 );
+
+export const retryPassphraseAuthentication = createThunk(
```
PeKne:
```suggestion
export const retryPassphraseAuthenticationThunk = createThunk(
```

suite-common/wallet-core/src/device/deviceThunks.ts
PeKne: Is this thunk needed? what about calling `authorizeDevice` thunk directly from `suite-native retryPassphraseAuthentication`

suite-native/module-passphrase/src/passphraseThunks.ts
```diff
@@ -56,3 +57,14 @@ export const verifyPassphraseOnEmptyWalletThunk = createThunk(
...
+    `${PASSPHRASE_MODULE_PREFIX}/retryPassphraseAuthentication`,
+    (_, { dispatch, getState }) => {
+        const device = selectDevice(getState());
+
+        if (!device) return;
+        dispatch(deviceActions.removeButtonRequests({ device }));
```
PeKne, line 67: Is it necessary to delete all button request? the `removeButtonrequests` action has second argument `buttonRequestCode`. Isn't it enough to delete only the passphrase request? Not sure just thinking if this call will not clash with other device button request features.

PeKne (commented):
few question. Otherwise looks good 👍

suite-common/wallet-core/src/device/deviceThunks.ts
juriczech: Good point, it's redundant. Fixed and rebased.

suite-native/module-passphrase/src/passphraseThunks.ts
juriczech: Fixed

suite-native/module-passphrase/src/passphraseThunks.ts
juriczech: It's actually not necessary per say, but it's a precaution in case there would be some hanging button requests. Not sure whether this can happen but it certainly doesn't hurt. Since we're about to re-authorize the device, let's bomb any button requests available at this moment to have a clean state when we start the authorization process. I can of course remove it, it's just in case kind of thing here..not sure.

PeKne: Ok, let's leave it for now. If this issue occurs, we can change it later.

PeKne (approved):

PR #12340, opened 2024-05-10
fix(connect-popup): update text in selectAccount

martykan:
## Description

Rename account types to be the same as in Suite:

- Accounts -> Default accounts (SegWit)
- SegWit accounts -> Legacy SegWit accounts

Removed warning about support in Trezor Wallet

## Screenshots

Screenshot 2024-05-10 at 17 17 03

### DIFF
--- a/packages/connect-popup/src/static/popup.html
+++ b/packages/connect-popup/src/static/popup.html
@@ -670,16 +670,14 @@
             <div class="select-account p2wpkh">
                 <div>
                     <div class="tabs">
-                        <div data-tab="p2wpkh" class="tab-selection p2wpkh">Accounts</div>
+                        <div data-tab="p2wpkh" class="tab-selection p2wpkh">
+                            Default accounts (SegWit)
+                        </div>
                         <div data-tab="p2tr" class="tab-selection p2tr">Taproot accounts</div>
-                        <div data-tab="p2sh" class="tab-selection p2sh">Segwit accounts</div>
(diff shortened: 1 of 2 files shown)


packages/connect-popup/src/static/popup.html
```diff
@@ -670,16 +670,12 @@ <h3></h3>
             <div class="select-account p2wpkh">
                 <div>
                     <div class="tabs">
-                        <div data-tab="p2wpkh" class="tab-selection p2wpkh">Accounts</div>
+                        <div data-tab="p2wpkh" class="tab-selection p2wpkh">Default accounts</div>
```
mroz22: maybe something like 
```
Segwit accounts (default)
```
?

or whatever, lets wait for @Hannsek

Hannsek: "Default accounts (SegWit)" seems fine. 👍🏻

Don't know from the top of my head how we handle SegWit/Segwit in Suite. Please use whatever is in Suite.

martykan: I'm using the same label as in Suite which doesn't say SegWit, but maybe it's better to mention it

Hannsek: Here is a different environment for the users, so let's mention it.

Hannsek: However I was talking about whether we use SegWit (big "W") or Segwit (small "w").

martykan: I think big W is standard

Hannsek (approved):

PR #12342, opened 2024-05-10
chore(suite): rework transport section in debug settings

### LINKED ISSUE #12288: Put bridge switcher into the debug menu

mroz22:
### Debug transport section reworked

Since I have encountered some confusions by @bosomt  on how transport sections actually works I decided to rework it slightly. 

This PR also resolves #12288.

Changes: 
- removed the "Processes" component from the debug menu. It has never been used for anything but toggling bridge process. The original ambition to have a common control panel for processes was not fulfilled. 
- added "Transport backend" section in place of "Processes". It has the same features but adds also ability to change the "startOnStartup"  bridge settings and it newly implements the ability to opt-in for the legacy bridge as requested in  #12288
- what previously was named "Transports" section has now become "Transport clients" section to bring more clarity here.  Also, descriptions of specific transport clients have been added.

Todo: 
- shouldn't suite-desktop-core become dependency of desktop-api in order to avoid types duplication?
- shoudln't we add some common channel for store updates and data retrieval instead particular calls such as the here added `bridge/get-status` etc? This might be related to what @marekrjpolak is doing right now (the redux loading thing)

image


marekrjpolak (approved):
It looks like it does what it should. Add automatic app reload or don't, and :shipit:

mroz22:
hmm, lets merge it for now. making it more user-friendly and fool-proof could be useful but that can't be done as a followup

PR #12343, opened 2024-05-10
fix(suite): unify decimal/integer errors

tomasklim:
## Description

- integer validation in staking/unstaking should not be there, it is ignored anyway
- i think it is better to say that token does not allow decimals `Maximum {decimals} decimals allowed` instead of `Amount is not an integer`
- I am not sure what error should be there in case of satoshies but I unified it with exchange section

## Screenshots:

Screenshot 2024-05-10 at 21 45 31


marekrjpolak (commented):
The integer check doesn't take `decimals` into account from now on, because there are always non-zero decimals in every coin and token? And the only case when we really require integer is when satoshis are set?

tomasklim:
There are tokens which have decimals zero, but it should work with this logic, shouldnt it?

I see no reason to tell user that token X allows only 2 decimal places and that token Y allow only integers, it is easier to tell that it allows 0 decimals?

marekrjpolak (approved):
> I see no reason to tell user that token Y allows only 2 decimal places and that token Y allow only integers, it is easier to tell that it allows 0 decimals?

Oooh, there's also `validateDecimals`! I haven't noticed. Approved.

PR #12356, opened 2024-05-13
fix(suite): fix copy

peter-sanderson:
## Related Issue

Resolve

### DIFF
--- a/packages/suite/src/views/onboarding/steps/SelectBackupType.tsx
+++ b/packages/suite/src/views/onboarding/steps/SelectBackupType.tsx
@@ -7,6 +7,7 @@ import {
     Text,
     Warning,
     useElevation,
+    variables,
 } from '@trezor/components';
 import {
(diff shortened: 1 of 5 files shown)


packages/suite/src/views/onboarding/steps/SelectBackupType.tsx
```diff
@@ -182,17 +200,57 @@ type FloatingSelectionsProps = {
...
+};
+
+const UpgradableToMultiTag = () => {
+    const { isMobileLayout } = useLayoutSize();
+
+    return (
+        <Badge
+            variant="tertiary"
+            inline
+            margin={{ left: spacings.xxs }}
```
jvaclavik, line 227: btw I think it's fine use just `left: 4`, it doesn't allow other numbers except spacing values.

jvaclavik (approved):

PR #12360, opened 2024-05-13
fix(connect-explorer): use connect-web for dev server

martykan:
## Description

Use `connect-web` along with `connect-explorer` dev server. 
Previously `connect-popup` was used, but it's dev server doesn't include iframe. 

This also adds support for `PORT` and `NO_HTTPS` env variables for configuring the `connect-web` dev server


martykan:
@SocketSecurity ignore-all

mroz22 (approved):

PR #12031, opened 2024-04-16
Feat/view only persistance

matejkriz:
## Description

- Persist remembered device data in redux-persist.
- Persist also Portfolio Tracker device because it also acts as remembered device.
- Preselect most recently connected remembered device on app start or Portfolio tracker device.

Persisted data:
- list of devices with `remembered === true` (selectedDevice is not remembered, devices are sorted by timestamp instead) 
- graph setting (timeframe) for accounts from remembered devices
- `accounts` and `transactions` related to remembered devices

We may also consider persisting the discovery state.

## Related Issue

Resolve [#11852](https://github.com/trezor/trezor-suite/issues/11852)

PR #12363, opened 2024-05-13
feat(suite-native): forget device when view only disabled

### LINKED ISSUE #12362: View only: forget device immediately when view only disabled
If user disable View-only for disconnected device, it should be ejected right away. Currently it's only removed from persist storage, but it remains active in app until reload.

matejkriz:
## Description

When disabling view-only, device (wallet) has to be not only removed from persistent store (handled in https://github.com/trezor/trezor-suite/pull/12031) but also from current state.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12362

### DIFF
--- a/suite-native/module-settings/src/components/ViewOnly/WalletRow.tsx
+++ b/suite-native/module-settings/src/components/ViewOnly/WalletRow.tsx
@@ -2,7 +2,7 @@ import { useDispatch } from 'react-redux';
 
 import { Button, HStack, Text } from '@suite-native/atoms';
 import { Translation } from '@suite-native/intl';
-import { toggleRememberDevice } from '@suite-common/wallet-core';
+import { deviceActions, toggleRememberDevice } from '@suite-common/wallet-core';
 import { useAlert } from '@suite-native/alerts';
 import { useToast } from '@suite-native/toasts';
 import { Icon } from '@suite-common/icons';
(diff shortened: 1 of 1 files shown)


vytick (approved):

suite-native/module-settings/src/components/ViewOnly/WalletRow.tsx
```diff
@@ -42,6 +42,10 @@ export const WalletRow = ({ device }: { device: AcquiredDevice }) => {
             icon: 'check',
         });
         dispatch(toggleRememberDevice({ device }));
+
+        if (device.remember) {
+            dispatch(deviceActions.forgetDevice(device));
```
vytick, line 47: should we forget device that is connected?

matejkriz: definitely not 🙉 thanks for the catch

matejkriz: fixed

matejkriz: https://github.com/trezor/trezor-suite/pull/12363/commits/6f60b5d78f25036254bbab57d5c0b8629426fc66

vytick (approved):
🚢

PR #12364, opened 2024-05-13
fix(suite-native): view only enabling for no biometrics allowed

vytick:
If device does not have available biometrics, do not wait for it to be resolved before showing EnableViewOnlyBottomSheet for the device


juriczech (approved):

matejkriz (changes requested):
It breaks the e2e test, please fix the test 🙏

matejkriz:
> It breaks the e2e test, please fix the test 🙏

Test fixed and global retries count for mobile E2E tests lowered on 2 so we don't need to wait so long and waste resources. If it fails 2 times, makes not much sense to retry more. cc @PeKne

matejkriz (approved):

PeKne:
@matejkriz  Ok, let's see if the tests gonna be flakier now. But 5 retries was definitely overkill. 👍

vytick:
Thanks for fixing the tests!

PR #12365, opened 2024-05-14
ci(github): exec helper script to throw when error

karliatto:
## Description
`exec` helper script to throw when error when it fails so the workflow is not continuing and it does appears like failed workflow.


vdovhanych (approved):
Yeah this is better

PR #12366, opened 2024-05-14
ci(github): test if non protected branches are pushable form workflow

karliatto:
## Description

Test if non protected branches are pushable form workflow


vdovhanych:
we do not need this, we tested it with unprotecting the branch

PR #12370, opened 2024-05-14
refactor(suite-native): passphrase screen

juriczech:
## Description
Unify screen styles across passphrase flow. No screenshots or issue attached as this is just simple refactor and results are identical to previous version, just more maintainable and with less code.

### DIFF
--- a/suite-native/module-passphrase/src/components/PassphraseScreen.tsx
+++ b/suite-native/module-passphrase/src/components/PassphraseScreen.tsx
@@ -1,0 +1,23 @@
+import { ReactNode } from 'react';
+
+import { Screen } from '@suite-native/navigation';
+import { useNativeStyles } from '@trezor/styles';
+
+import { PassphraseScreenHeader } from './PassphraseScreenHeader';
+
+type PassphraseScreenWrapperProps = {
+    children: ReactNode;
+};
+
+export const PassphraseScreenWrapper = ({ children }: PassphraseScreenWrapperProps) => {
+    const { utils } = useNativeStyles();
+
+    return (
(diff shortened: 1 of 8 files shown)


suite-native/module-passphrase/src/components/PassphraseScreen.tsx
```diff
@@ -0,0 +1,21 @@
+import { Screen } from '@suite-native/navigation';
+import { useNativeStyles } from '@trezor/styles';
+
+import { PassphraseScreenHeader } from './PassphraseScreenHeader';
+
+type PassphraseScreenProps = {
+    children: React.ReactNode;
+};
+
+export const PassphraseScreen = ({ children }: PassphraseScreenProps) => {
```
vytick: nit: What about something like this?

```suggestion
export const PassphraseScreenWrapper = ({ children }: PassphraseScreenProps) => {
```

vytick (approved):
Thanks for simplifying the code 🚀 

I have one naming nit otherwise lfg

suite-native/module-passphrase/src/components/PassphraseScreen.tsx
juriczech: Fixed

PeKne: filename was not changed.

PeKne: Could you also rename the other component for consistency? 🙏

eg. `PassphraseContentScreen` --> `TitledPassphraseScreenWrapper`

PR #12371, opened 2024-05-14
ci(github): js script GITHUB_TOKEN check

karliatto:
## Description
Checking and making sure the script is using the proper GITHUB_TOKEN

### DIFF
--- a/.github/workflows/release-connect-npm-init.yml
+++ b/.github/workflows/release-connect-npm-init.yml
@@ -44,6 +44,7 @@ jobs:
           git config --global user.name "trezor-ci"
           git config --global user.email "${{ secrets.TREZOR_BOT_EMAIL }}"
           gh config set prompt disabled
+          gh api /user --jq .login
           node ./ci/scripts/connect-release-init-npm.js ${{ github.event.inputs.semver }}
         env:
           GITHUB_TOKEN: ${{ secrets.TREZOR_BOT_TOKEN }}
--- a/ci/scripts/connect-release-init-npm.js
+++ b/ci/scripts/connect-release-init-npm.js
@@ -47,6 +47,16 @@ const findIndexByCommit = (commitArr, searchString) =>
     commitArr.findIndex(commit => commit.includes(searchString));
 
 const initConnectRelease = async () => {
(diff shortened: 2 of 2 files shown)


.github/workflows/release-connect-npm-init.yml
```diff
@@ -44,6 +44,7 @@ jobs:
           git config --global user.name "trezor-ci"
           git config --global user.email "${{ secrets.TREZOR_BOT_EMAIL }}"
           gh config set prompt disabled
+          gh api /user --jq .login
```
karliatto, line 47: I think this should log the user that is being used here.

ci/scripts/connect-release-init-npm.js
```diff
@@ -47,6 +47,16 @@ const findIndexByCommit = (commitArr, searchString) =>
     commitArr.findIndex(commit => commit.includes(searchString));
 
 const initConnectRelease = async () => {
+    console.log('Using GitHub Token:', process.env.GITHUB_TOKEN ? 'Yes' : 'No');
+
+    if (process.env.GITHUB_TOKEN) {
+        // Making sure we use the proper GITHUB_TOKEN
+        exec('gh', ['auth', 'setup-git']);
+        exec('gh', ['config', 'set', '-h', 'github.com', 'oauth_token', process.env.GITHUB_TOKEN]);
+    } else {
```
karliatto, line 56: It was not required before to do anything like this in the script to be able to use the proper user but maybe since we have made different configurations might be required now?

.github/workflows/release-connect-npm-init.yml
vdovhanych: Yes, but for the GH tool it's used correctly. It seems like it's not used correctly for the running script. 

the user that created the pull request was in fact trezor-ci so maybe my assumption was bad about the wrong credentials.

vdovhanych (approved):

.github/workflows/release-connect-npm-init.yml
karliatto: I see.
Is it possible that now some sort of environment is required? 
I am trying to think in terms of auth, what have changed, since this workflow with almost same script has been working for long time and now it does not work.

PR #12378, opened 2024-05-14
Refactor/send form redux cleanup

PeKne:
Batch of send form redux logic refactoring. I recommend to review this PR commit by commit.

## Description
- [refactor(suite): signed transaction data stored in redux send form state](https://github.com/trezor/trezor-suite/pull/12378/commits/b8fe6faa83f61c7260a2410229ab2d7b270fbf3d): signed transaction unserialized data stored in redux state so it does not have to be passed between thunks as an argument

- [refactor(suite): send form common thunks are rejected on fail](https://github.com/trezor/trezor-suite/pull/12378/commits/b3047003876ab1f5901378065c8c9805bb01f558): suite-common sendFormThunks utilize redux-toolkit `rejectWithValue` for better error handling
- [refactor(suite): send form precomposed transaction arguments naming unified](https://github.com/trezor/trezor-suite/pull/12378/commits/2492c1dc2aec14f0e4913ec2b8ed5b3e07e85c6c): naming of `precomposedForm` arguments unified for better readability
- [refactor(suite): unify send form precomposed-transaction types naming](https://github.com/trezor/trezor-suite/pull/12378/commits/ccb9c1574f02a7d0c7ae036db3d6a4fd4c7b7347): There are two types of a successful precomposedTransaction. One for Cardano and other for rest of the networks. This commit makes the naming of these types clearer.
- [refactor(suite): split pushSendFormTransactionThunk into sub-thunks](https://github.com/trezor/trezor-suite/pull/12378/commits/b84f20bf29c9af3adfde81221c30a402a25b64d2): Bulky pushSendFormTransactionThunk was split into smaller sub-thunks. Some of theme were needed only by desktop, these were moved to `packages/suite` package.
- [refactor(suite): convert send form drafts amount units refactored](https://github.com/trezor/trezor-suite/pull/12378/commits/d381866d6d9799b9aa4dddee4a17da33e671e29a)
- [refactor(suite): redundant precomposedForm state removed from sendFormReducer](https://github.com/trezor/trezor-suite/pull/12378/commits/aea18e1c1128afaacd945b9c9150054987ca931f)

### DIFF
--- a/packages/suite/src/actions/wallet/send/sendFormThunks.ts
+++ b/packages/suite/src/actions/wallet/send/sendFormThunks.ts
@@ -114,7 +114,7 @@ export const signAndPushSendFormTransactionThunk = createThunk(
         // this action is blocked by modalActions.preserve()
         dispatch(modalActions.preserve());
 
-        const { serializedTx, signedTransaction } = await dispatch(
+        const { serializedTx } = await dispatch(
             signTransactionThunk({
                 formValues,
                 transactionInfo: enhancedTxInfo,
@@ -137,7 +137,6 @@ export const signAndPushSendFormTransactionThunk = createThunk(
             // push tx to the network
             return dispatch(
                 pushSendFormTransactionThunk({
-                    signedTransaction,
                     selectedAccount,
                 }),
             ).unwrap();
--- a/packages/suite/src/actions/wallet/send/sendFormThunks.ts
+++ b/packages/suite/src/actions/wallet/send/sendFormThunks.ts
@@ -14,10 +14,10 @@ import {
     pushSendFormTransactionThunk,
     replaceTransactionThunk,
     selectDevice,
-    selectPrecomposedSendForm,
     selectSendFormDrafts,
     signTransactionThunk,
     sendFormActions,
+    selectSendFormDraftByAccountKey,
 } from '@suite-common/wallet-core';
 import { isCardanoTx } from '@suite-common/wallet-utils';
 import { MetadataAddPayload } from '@suite-common/metadata-types';
@@ -28,7 +28,7 @@ import {
     selectSelectedAccountKey,
     selectIsSelectedAccountLoaded,
 } from 'src/reducers/wallet/selectedAccountReducer';
-import { findLabelsToBeMovedOrDeleted } from '../moveLabelsForRbfActions';
+import { findLabelsToBeMovedOrDeleted, moveLabelsForRbfAction } from '../moveLabelsForRbfActions';
 import { selectMetadata } from 'src/reducers/suite/metadataReducer';
 import * as metadataLabelingActions from 'src/actions/suite/metadataLabelingActions';
 import * as modalActions from 'src/actions/suite/modalActions';
@@ -95,19 +95,15 @@ const updateRbfLabelsThunk = createThunk(
     (
         {
             labelsToBeEdited,
-            precomposedTx,
+            precomposedTransaction,
             txid,
         }: {
             labelsToBeEdited: RbfLabelsToBeUpdated;
-            precomposedTx: PrecomposedTransactionFinal;
+            precomposedTransaction: PrecomposedTransactionFinal;
             txid: string;
         },
-        { dispatch, extra },
+        { dispatch },
     ) => {
-        const {
-            thunks: { moveLabelsForRbfAction },
-        } = extra;
-
         dispatch(
             moveLabelsForRbfAction({
                 toBeMovedOrDeletedList: labelsToBeEdited,
@@ -120,7 +116,7 @@ const updateRbfLabelsThunk = createThunk(
         // this will update transaction details (like time, fee etc.)
         dispatch(
             replaceTransactionThunk({
-                precomposedTx,
+                precomposedTransaction,
                 newTxid: txid,
             }),
         );
@@ -132,63 +128,61 @@ const applySendFormMetadataLabelsThunk = createThunk(
     (
         {
             selectedAccount,
-            precomposedTx,
+            precomposedTransaction,
             txid,
         }: {
             selectedAccount: Account;
-            precomposedTx: GeneralPrecomposedTransactionFinal;
+            precomposedTransaction: GeneralPrecomposedTransactionFinal;
             txid: string;
         },
         { dispatch, getState },
     ) => {
         const metadata = selectMetadata(getState());
-        if (metadata.enabled) {
-            const precomposedForm = selectPrecomposedSendForm(getState());
-            let outputsPermutation: number[];
-            if (isCardanoTx(selectedAccount, precomposedTx)) {
-                // cardano preserves order of outputs
-                outputsPermutation = precomposedTx?.outputs.map((_o, i) => i);
-            } else {
-                outputsPermutation = precomposedTx?.outputsPermutation;
-            }
 
-            const synchronize = getSynchronize();
+        if (!metadata.enabled) return;
 
-            precomposedForm?.outputs
-                // create array of metadata objects
-                .map((formOutput, index) => {
-                    const { label } = formOutput;
-                    // final ordering of outputs differs from order in send form
-                    // outputsPermutation contains mapping from @trezor/utxo-lib outputs to send form outputs
-                    // mapping goes like this: Array<@trezor/utxo-lib index : send form index>
-                    const outputIndex = outputsPermutation.findIndex(p => p === index);
-                    const outputMetadata: Extract<MetadataAddPayload, { type: 'outputLabel' }> = {
-                        type: 'outputLabel',
-                        entityKey: selectedAccount.key,
-                        txid,
-                        outputIndex,
-                        value: label,
-                        defaultValue: '',
-                    };
+        const formDraft = selectSendFormDraftByAccountKey(getState(), selectedAccount.key);
 
-                    return outputMetadata;
-                })
-                // filter out empty values AFTER creating metadata objects (see outputs mapping above)
-                .filter(output => output.value)
-                // propagate metadata to reducers and persistent storage
-                .forEach((output, index, arr) => {
-                    const isLast = index === arr.length - 1;
+        const outputsPermutation = isCardanoTx(selectedAccount, precomposedTransaction)
+            ? precomposedTransaction?.outputs.map((_o, i) => i) // cardano preserves order of outputs
+            : precomposedTransaction?.outputsPermutation;
 
-                    synchronize(() =>
-                        dispatch(
-                            metadataLabelingActions.addAccountMetadata({
-                                ...output,
-                                skipSave: !isLast,
-                            }),
-                        ),
-                    );
-                });
-        }
+        const synchronize = getSynchronize();
+
+        formDraft?.outputs
+            // create array of metadata objects
+            .map((formOutput, index) => {
+                const { label } = formOutput;
+                // final ordering of outputs differs from order in send form
+                // outputsPermutation contains mapping from @trezor/utxo-lib outputs to send form outputs
+                // mapping goes like this: Array<@trezor/utxo-lib index : send form index>
+                const outputIndex = outputsPermutation.findIndex(p => p === index);
+                const outputMetadata: Extract<MetadataAddPayload, { type: 'outputLabel' }> = {
+                    type: 'outputLabel',
+                    entityKey: selectedAccount.key,
+                    txid,
+                    outputIndex,
+                    value: label,
+                    defaultValue: '',
+                };
+
+                return outputMetadata;
+            })
+            // filter out empty values AFTER creating metadata objects (see outputs mapping above)
+            .filter(output => output.value)
+            // propagate metadata to reducers and persistent storage
+            .forEach((output, index, arr) => {
+                const isLast = index === arr.length - 1;
+
+                synchronize(() =>
+                    dispatch(
+                        metadataLabelingActions.addAccountMetadata({
+                            ...output,
+                            skipSave: !isLast,
+                        }),
+                    ),
+                );
+            });
     },
 );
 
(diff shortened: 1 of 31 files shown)


suite-common/wallet-core/src/send/sendFormActions.ts
```diff
@@ -28,15 +28,15 @@ const storePrecomposedTransaction = createAction(
...
+        precomposedTransaction: GeneralPrecomposedTransactionFinal;
     }) => ({
         payload,
     }),
 );
 
 const storeSignedTransaction = createAction(
     `${SEND_MODULE_PREFIX}/store-signed-transaction`,
-    (payload: FormSignedTx) => ({
+    (payload: { serializedTx: SerializedTx; signedTx?: BlockbookTransaction }) => ({
```
vytick, line 40: will it work if the txn is from nonblockbook blockchains like solana, ripple or cardano?

packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -80,44 +90,144 @@ export const importSendFormRequestThunk = createThunk(
...
+            precomposedTx: PrecomposedTransactionFinal;
+            txid: string;
+        },
+        { dispatch, extra },
+    ) => {
+        const {
+            thunks: { moveLabelsForRbfAction },
+        } = extra;
+
+        if (labelsToBeEdited !== undefined) {
```
vytick: can labelsToBeEdited be undefined? It looks like no. If so, can you, please, update type?

packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -80,44 +90,144 @@ export const importSendFormRequestThunk = createThunk(
...
+                .map((formOutput, index) => {
+                    const { label } = formOutput;
+                    // final ordering of outputs differs from order in send form
+                    // outputsPermutation contains mapping from @trezor/utxo-lib outputs to send form outputs
+                    // mapping goes like this: Array<@trezor/utxo-lib index : send form index>
+                    const outputIndex = outputsPermutation.findIndex(p => p === index);
+                    const outputMetadata: Extract<MetadataAddPayload, { type: 'outputLabel' }> = {
+                        type: 'outputLabel',
+                        entityKey: selectedAccount.key,
+                        txid, // txid becomes available, use it
```
vytick: what does this comment mean? txid should always be there, shouldnt it?

vytick (commented):
I am not deep in the send here, but it lgtm :) I have a few questions though

suite-common/wallet-core/src/send/sendFormActions.ts
PeKne: Yes, `signedTx` is optional argument. Only BTC networks utilize it.

packages/suite/src/actions/wallet/send/sendFormThunks.ts
PeKne: fixed in e3583e3c18

packages/suite/src/actions/wallet/send/sendFormThunks.ts
PeKne: fixed in e3583e3c18

suite-common/wallet-utils/src/sendFormUtils.ts
```diff
@@ -204,10 +203,7 @@ export const getFeeUnits = (networkType: NetworkType) => {
     return 'sat/B';
 };
 
-export const getFee = (
-    networkType: NetworkType,
-    tx: PrecomposedTransactionFinal | TxFinalCardano,
-) => {
+export const getFee = (networkType: NetworkType, tx: GeneralPrecomposedTransactionFinal) => {
```
juriczech, line 206: nit: I know it's just 2 params, but I think since this will be called from redux, it might be worth doing these functions with named parameters for better readability. Now there are random names and you could call `getFee(someIrrelevantName, differentStupidName)` and it could be confusing there. 

Feel free to ignore.

suite-common/wallet-core/src/transactions/transactionsThunks.ts
```diff
@@ -105,20 +106,22 @@ export const replaceTransactionThunk = createThunk<ReplaceTransactionThunkParams
...
 interface AddFakePendingTransactionParams {
-    transaction: Transaction;
     precomposedTx: PrecomposedTransactionFinal;
     account: Account;
 }
 
 export const addFakePendingTxThunk = createThunk<AddFakePendingTransactionParams>(
     `${TRANSACTIONS_MODULE_PREFIX}/addFakePendingTransaction`,
-    ({ transaction, precomposedTx, account }, { dispatch, getState }) => {
+    ({ precomposedTx, account }, { dispatch, getState, rejectWithValue }) => {
```
juriczech: nit: Please use full name for better readability

```suggestion
    ({ precomposedTransaction, account }, { dispatch, getState, rejectWithValue }) => {
```

packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -80,44 +90,142 @@ export const importSendFormRequestThunk = createThunk(
...
+
+        // notification from the backend may be delayed.
+        // modify affected transaction(s) in the reducer until the real account update occurs.
+        // this will update transaction details (like time, fee etc.)
+        dispatch(
+            replaceTransactionThunk({
+                precomposedTx,
+                newTxid: txid,
+            }),
+        );
```
juriczech: I wonder if we should batch these 2 actions
https://react-redux.js.org/api/batch

packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -130,17 +238,52 @@ export const signAndPushSendFormTransactionThunk = createThunk(
...
+
+            if (isRejected(pushResponse)) {
+                return pushResponse.payload as Unsuccessful;
+            }
+
+            const result = pushResponse.payload;
+            const { txid } = result.payload;
+
+            if (isRbf && rbfLabelsToBeEdited) {
+                await dispatch(
```
juriczech: Redundant `await`. Wonder if we should just batch the dispatch of those two actions here as the thunk seems redundant, but if you feel that it will grow with more functionality, it's okay to keep it I guess.

packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -80,44 +90,142 @@ export const importSendFormRequestThunk = createThunk(
...
+        }: {
+            labelsToBeEdited: RbfLabelsToBeUpdated;
+            precomposedTx: PrecomposedTransactionFinal;
+            txid: string;
+        },
+        { dispatch, extra },
+    ) => {
+        const {
+            thunks: { moveLabelsForRbfAction },
+        } = extra;
```
juriczech: Please avoid extra dependencies here as it can be used straight away by simple dispatch.

packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -80,44 +90,142 @@ export const importSendFormRequestThunk = createThunk(
...
+            txid,
+        }: {
+            selectedAccount: Account;
+            precomposedTx: GeneralPrecomposedTransactionFinal;
+            txid: string;
+        },
+        { dispatch, getState },
+    ) => {
+        const metadata = selectMetadata(getState());
+        if (metadata.enabled) {
```
juriczech:
```suggestion
        if (!metadata.enabled) return rejectWithValue('something')
```

juriczech: And then there's no need for indentation below.

packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -80,44 +90,142 @@ export const importSendFormRequestThunk = createThunk(
...
+            const formDraft = selectSendFormDraftByAccountKey(getState(), selectedAccount.key);
+            let outputsPermutation: number[];
+            if (isCardanoTx(selectedAccount, precomposedTx)) {
+                // cardano preserves order of outputs
+                outputsPermutation = precomposedTx?.outputs.map((_o, i) => i);
+            } else {
+                outputsPermutation = precomposedTx?.outputsPermutation;
+            }
```
juriczech: This could be a simple util function that returns `outputsPermutation: number[];` without the need for variable. Would be much nicer IMO

packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -130,17 +238,52 @@ export const signAndPushSendFormTransactionThunk = createThunk(
         }
 
         // Open a deferred modal and get the decision
-        const decision = await dispatch(
+        const isPushConfirmed = await dispatch(
             modalActions.openDeferredModal({ type: 'review-transaction' }),
         );
-        if (decision) {
+        if (isPushConfirmed) {
```
juriczech: Same here, redundant indentation

packages/suite/src/hooks/wallet/form/useCompose.ts
```diff
@@ -240,7 +240,7 @@ export const useCompose = <TFieldValues extends FormState>({
             const result = await dispatch(
                 signAndPushSendFormTransactionThunk({
                     formValues: values,
-                    transactionInfo: composedTx,
+                    precomposedTransaction: composedTx,
                     selectedAccount,
                 }),
             ).unwrap();
```
juriczech: I know it was named in the past, but can you rename `composedTx` defined few lines above to be `precomposedTransaction`?

packages/suite/src/hooks/wallet/useCoinmarketRecomposeAndSign.ts
```diff
@@ -149,7 +149,7 @@ export const useCoinmarketRecomposeAndSign = () => {
             return dispatch(
                 signAndPushSendFormTransactionThunk({
                     formValues,
-                    transactionInfo: composedToSign,
+                    precomposedTransaction: composedToSign,
```
juriczech: Same here please :pray:

packages/suite/src/hooks/wallet/useSendForm.ts
```diff
@@ -268,7 +268,7 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
             const result = await dispatch(
                 signAndPushSendFormTransactionThunk({
                     formValues: values,
-                    transactionInfo: composedTx,
+                    precomposedTransaction: composedTx,
```
juriczech: Also here

bosomt:
See screenshots, same transaction, `RBF on` but after tx is signed modal reports `RBF off`

Screenshot 2024-05-17 at 11 28 58

Screenshot 2024-05-17 at 11 29 07


**Info:**
 - Suite version: desktop 24.6.0 (a6fe7cdcce320ec50ffd18cb118f07c19fab8b3e)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.6.0 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)

packages/suite/src/actions/wallet/send/sendFormThunks.ts
PeKne: The refered documentation says that if you are using React 18 (we do), the call of `batch()` is redundant. The newer version of React take care of this under the hood. So I would leave it as it is.

suite-common/wallet-core/src/transactions/transactionsThunks.ts
PeKne: fixed 7b2265b66b

packages/suite/src/actions/wallet/send/sendFormThunks.ts
PeKne: await removed 7b2265b66b

packages/suite/src/actions/wallet/send/sendFormThunks.ts
PeKne: 7b2265b66b

packages/suite/src/actions/wallet/send/sendFormThunks.ts
PeKne: fixed 7b2265b66b. But the `rejectWithValue` was not used, just `return` called. It is because this is a suite desktop  thunk that is called i multiple places in `useSendForm` and it expects a return shape then would `rejectWithValue` created. Then I would have to refactor way more then I wish in this PR.

I believe that this thunks will have to be revisited by desktop team anyway if they decide to do the desktop send form rewrite. So I would leave it like that for now.

packages/suite/src/actions/wallet/send/sendFormThunks.ts
PeKne: I got your concern about the unnecessary `let`. Since it is only three lines of code and the new function would not be used anywhere else, I decided to change it to ternary operator call instead, Fixed by 7b2265b66b.

packages/suite/src/actions/wallet/send/sendFormThunks.ts
PeKne: fixed 7b2265b66b, same coment as above applies.

packages/suite/src/hooks/wallet/form/useCompose.ts
PeKne: 7b2265b66b

packages/suite/src/hooks/wallet/useCoinmarketRecomposeAndSign.ts
PeKne: 7b2265b66b

packages/suite/src/hooks/wallet/useSendForm.ts
PeKne: 7b2265b66b

PeKne:
@bosomt The problem should be fixed by: [f635113](https://github.com/trezor/trezor-suite/pull/12378/commits/f635113d422ec705ce7b29cfa525f588f980a1d5)

Please test it again 🙏.

bosomt:
**QA OK**

yes fixed  and overall OK , i think we can merge it to develop

**Info:**
 - Suite version: web 24.6.0 (31b59dc33073b61623beb11c4b6347e10d4625e0)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:126.0) Gecko/20100101 Firefox/126.0
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)

juriczech (approved):

suite-common/wallet-utils/src/sendFormUtils.ts
PeKne: Oh, I forgot to answer to this no. I would rather not touch this. If I change the parameteres of this function, I think I should also change the parameters of all the other functions of this utils file. ALso this function is impoted to many places, so it would create a huge diff. 

I got your point and agree with you, but I would rather avoid this.

marekrjpolak (commented):
Looks good in general, a lot of work has been done here.

The only thing I'm not quite happy about is the first commit: I prefer passing required arguments explicitly between the stages of a single process, rather than using the global state for it. That way:
- data validation is performed by type checking in development time instead of runtime checks
- functions (mostly thunks in this case) have as few side effects as possible while having as few dependencies as possible at the same time
- less code is needed to pass an argument than to store it into the state, read it somewhere else and validate that it is in fact there
- less effort is needed to understand the code if it's directly visible what the function needs in order to work and what it produces without looking into its implementation (that's why pure functions are so beautiful)

In this case it seems that `signedTx` and `serializedTx` are both stored at the beginning of `signAndPushSendFormTransactionThunk` (or its subroutine) and needed somewhere at the end of it, therefore I don't see a reason for using global state other than hiding the complexity of passing the parameters, which makes the overall complexity bigger instead of smaller imo.

I'm not saying it cannot be merged as it is, but I'll try to look at it a bit more if you don't mind.

marekrjpolak (approved):
After call I agree that as a prerequisite for send functionality in native app this may make sense. :+1:  :shipit:

PR #12411, opened 2024-05-15
feat(transport): make signal required param

mroz22:
- [cf97b8d](https://github.com/trezor/trezor-suite/pull/12411/commits/cf97b8df95c5787eb7cecb3ecf7124ce70e9d80c) might be interesting for @Nodonisko or maybe @matejkriz. We don't really want to force implementors of connect, who need to pass a custom transport to call its constructor. so here I made changes that connect accepts transport class constructor instead of its instance. should be ok, tested it.
- other than that this is fixing `Jest did not exit one second after the test run has completed.` in `yarn workspace @trezor/transport test:unit` reported today by @szymonlesisz 

Followups: 

- pass abortController.signal from connect core to device list. This required slightly [more changes in core ](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/core/index.ts#L1035) so  I suggest doing it in a followup.

### DIFF
--- a/packages/connect/src/device/DeviceList.ts
+++ b/packages/connect/src/device/DeviceList.ts
@@ -78,42 +78,30 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> {
 
         const transportLogger = initLog('@trezor/transport', debug);
 
+        const transportCommonArgs = {
+            messages: this.messages,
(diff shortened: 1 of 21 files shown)


martykan (approved):
LGTM

PR #12422, opened 2024-05-15
tests(connect): add DeviceList.init tests

szymonlesisz:
## Description

Prerequisite for multiple future fixes. like https://github.com/trezor/trezor-suite/pull/12411 or multi-transport.

adding `DeviceList` unit tests where multiple cases have unexpected behavior (commented within test case)

PR #12423, opened 2024-05-15
Fix/transport stopped flag

szymonlesisz:
## Description

Based on https://github.com/trezor/trezor-suite/pull/12422

- Transport should be stopped before calling Transport.init or in case when Transport.init throws error
- AbstractApiTransport.acquire listenPromise condition check

### DIFF
--- a/packages/transport/src/transports/abstract.ts
+++ b/packages/transport/src/transports/abstract.ts
@@ -93,7 +93,7 @@ export abstract class AbstractTransport extends TypedEmitter<{
     /**
      * once transport has been stopped, it does not emit any events
      */
-    protected stopped = false;
+    protected stopped = true;
     /**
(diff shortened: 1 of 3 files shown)


mroz22:
please rebase 🙏

packages/transport/src/transports/abstract.ts
```diff
@@ -93,7 +93,7 @@ export abstract class AbstractTransport extends TypedEmitter<{
     /**
      * once transport has been stopped, it does not emit any events
      */
-    protected stopped = false;
+    protected stopped = true;
```
szymonlesisz, line 96: not sure if `stopped` is the best name, `inited` ? `running`?

mroz22: I would go for `dispose` and `disposed`

mroz22 (approved):

PR #12424, opened 2024-05-15
Feat/view only analytics

### LINKED ISSUE #11890: View only - analytics
Send events when:
User clicks on Enable
User clicks on Skip
When user changes the settings
View-only enabled/disabled when app is opened

matejkriz:
## Description

Adding analytics for View-only feature and an easy way of debugging analytics reporting.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/11890

## Screenshots:
image


suite-native/analytics/src/analytics.ts
```diff
@@ -11,5 +11,10 @@ export const analytics = new Analytics<SuiteNativeAnalyticsEvent>({
 
 if (isDebugEnv()) {
     // Do not send analytics in development
-    analytics.report = () => {};
+    analytics.report = (event: Event) => {
+        if (process.env.EXPO_PUBLIC_IS_ANALYTICS_LOGGER_ENABLED) {
```
vytick, line 15: i this condition (and env variable) needed when we only log this while `isDebugEnv()` ?

vytick (approved):
lgtm 🚀 

I have one nit Q, but feel free to ignore and resolve if you decide it does not matter

suite-native/analytics/src/analytics.ts
matejkriz: Yes, I think it is still useful - mostly you don't care about analytics during development and you don't want console spammed by it...

PR #12425, opened 2024-05-15
chore(connect): remove DataManager dependency from DeviceList constru…

mroz22:
as discussed with @szymonlesisz

### DIFF
--- a/packages/connect/src/device/DeviceList.ts
+++ b/packages/connect/src/device/DeviceList.ts
@@ -16,7 +16,7 @@ import { ERRORS } from '../constants';
 import { DEVICE, TransportInfo } from '../events';
 import { Device } from './Device';
 import type { ConnectSettings, Device as DeviceTyped } from '../types';
-import { DataManager } from '../data/DataManager';
+
 import { getBridgeInfo } from '../data/transportInfo';
 import { initLog } from '../utils/debug';
 import { resolveAfter } from '../utils/promiseUtils';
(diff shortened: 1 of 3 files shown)


packages/connect/src/device/DeviceList.ts
```diff
@@ -62,27 +61,24 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> {
 
     transportFirstEventPromise: Promise<void> | undefined;
 
-    constructor() {
+    constructor({
+        transports = ['BridgeTransport'],
```
martykan: This fallback won't work anymore if you pass an empty array. 
But I'm not sure if that can really happen at this point, maybe it's not important

martykan (approved):

packages/connect/src/device/DeviceList.ts
mroz22: ha, good point, I ll put it back. didn't want to make any runtime changes like this one

mroz22: f093d5e706c84493c8c81b1963075faa4a2cf537

packages/connect/src/device/DeviceList.ts
```diff
@@ -62,13 +61,16 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> {
 
     transportFirstEventPromise: Promise<void> | undefined;
 
-    constructor() {
+    constructor({
+        transports,
+        debug,
+        messages,
+    }: {
+        transports?: ConnectSettings['transports'];
```
szymonlesisz: maybe it wold be better to accept all settings here, like `constructor({ settings, messages })` 

DataManager is still used down in the code to get some other settings (priority, pendingTransportEvent) and this way it will be easier to get rid of it completely

mroz22: yes, but in this case I would need to assign these settings to `this`. So this is something I wanted to discuss or maybe do in a followup.

mroz22: removed in [2d3a470](https://github.com/trezor/trezor-suite/pull/12425/commits/2d3a47014e1412620b398d8518fbbf0ca234eea9)

packages/connect/src/device/__tests__/DeviceList.test.ts
```diff
@@ -94,27 +103,37 @@ const createDeviceList = () => {
 };
 
 describe('DeviceList', () => {
-    it('constructor throws error on unknown transport (string)', async () => {
-        await loadDataManager({ transports: ['FooBarTransport'] });
+    beforeAll(async () => {
+        // todo: I don't get it. If we pass empty messages: {} (see getDeviceListParams), tests behave differently.
```
szymonlesisz, line 102: probably because tests relay on real `Initialize > Features || Success` protobuf messages encoding/decoding

szymonlesisz (approved):
cool

PR #12440, opened 2024-05-16
ci: fix sldev build where git lfs checkout was missing

mroz22:
@bosomt @STew790 please test if fw update works for you on this branch. 

also are we sure that it will work in production builds?

### DIFF
--- a/.github/workflows/build-suite-web.yml
+++ b/.github/workflows/build-suite-web.yml
@@ -31,6 +31,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          lfs: true
 
       - name: Configure aws credentials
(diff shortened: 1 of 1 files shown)


vdovhanych:
This will eat all of the lfs bandwidth for us in a day or so :) That was the reason we got rid of it for this in the first place.

mroz22:
ok in that cases @trezor/qa should be informed 👍

mroz22:
@MiroslavProchazka maybe we can yet again start pulling firmwares for web suite from data.trezor.io. 
the only downside for this is that you won't be able to test newly added firmware binaries in suite web on sldev.

+ it would be nice to add some better errors to UI - validate fw binary slightly before sending it into device.

PR #12443, opened 2024-05-16
feat(suite-native): delete view only feature flag completely

### LINKED ISSUE #12013: View only [last] - delete feature flag or add migration
when we publish it, we need to either delete the feature flag completely or add a migration to make it appear to existing users

**Notes for QA**
View only feature is now turned on for everyone, even for those updating from 24.4 version.

matejkriz:
## Description

It is no more needed, and by this change it is also removed from persist store.

Acceptance criteria for **QA**:
- View only feature is now turned on for everyone, even for those updating from 24.4 version

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12013

### DIFF
--- a/suite-native/device/tsconfig.json
+++ b/suite-native/device/tsconfig.json
@@ -2,9 +2,7 @@
     "extends": "../../tsconfig.base.json",
     "compilerOptions": { "outDir": "libDev" },
     "references": [
-        {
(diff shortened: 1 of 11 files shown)


juriczech (approved):
LGTM

PR #12444, opened 2024-05-16
ci(github): connect release reorganization

karliatto:
## Description
This PR changes the process to release connect to be like below:

1. Create a non-protected branch to update changelogs and bump versions
2. Merging the previous branch to develop
3. Triggering new workflow that will from specific commit from develop branch create the `release/connect/...` branch and will check what versions are different in package.json and NPM and it will trigger the workflows to release those with different versions
4. The same workflow will trigger the relesase to staging (without permissions since it is just staging)
5. Then we need to manually trigger the production release that will need to trigger by an authorized member and approved by other one.
6. Release is done.

## Release process
So the process would be:
1. Run workflow `.github/workflows/release-connect-bump-versions.yml`
    * It will create a branch (non protected) with version bumped and changelogs most of it automated
    * This branch can be updated and modify until we get satisfying changelog
    * Once the changelogs and versions are fine we merge it to `develop`
2. Run workflow `.github/workflows/release-connect-init.yml` using the commit hash of the previously merged `bump-versions/connect-${version}`.
    * It will create the branch `release/connect/<version>`
    * Automatically trigger github/workflows/release-connect-v9-staging.yml  over the new created branch that will release to staging without any other approval.
    * Automatically trigger .github/workflows/release-connect-v9-production.yml over the new created release branch that will need approval from other member to release to production connect.trezor.io/9 for all the deployment types and only to connect.trezor.io/9.x.x for `canary` releases.
    * Automatically trigger .github/workflows/release-connect-npm-init.yml over new created release branch that will require approval to release all the npm dependencies.

## Related Issue

Related https://github.com/trezor/trezor-suite/issues/12157


.github/workflows/release-connect-bump-versions.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: yarn install
+
+      - name: Set git for trezor-ci
+        run: |
+          git config --global user.name "trezor-ci"
+          git config --global user.email "${{ secrets.TREZOR_BOT_EMAIL }}"
+
+      - name: Check dependencies to update
+        run: |
+          node ./ci/scripts/connect-bump-versions.js ${{ github.event.inputs.semver }}
```
mroz22: not really related to this PR, but I am thinking if we can't possibly convert our scripts to typescript and run them using `yarn tsx`. 

using pure node is probably only needed if  we don't want to install dependencies. And in this case, since we are using pure node, we maybe don't really need to do `yarn install` on line 36?

.github/workflows/release-connect-init.yml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Extract connect version
+        id: set-version
+        run: echo "version=$(node ./ci/scripts/get-connect-version.js)" >> $GITHUB_OUTPUT
+
+  create-push-release-branch:
+    needs: [extract-version]
+    name: "Create release branch for version ${{ needs.extract-version.outputs.version }}"
+    runs-on: ubuntu-latest
+    outputs:
+      branch_name: ${{ steps.push-branch.outputs.branch_name }}
```
mroz22, line 44: this + line 74 make `${{ needs.create-push-release-branch.outputs.branch_name }}` available for subsequent jobs, am I right?

.github/workflows/release-connect-npm-init.yml
```diff
@@ -3,21 +3,50 @@ name: "[Release] connect npm init"
...
+  extract-version-from-package-json:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.set-version.outputs.version }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Number of commits to fetch. 0 indicates all history for all branches and tags.
+          fetch-depth: 0
```
mroz22: does it really need full history here?

.github/workflows/release-connect-bump-versions.yml
```diff
@@ -0,0 +1,45 @@
...
+          - minor
+          - prerelease
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  bump-versions:
+    if: github.repository == 'trezor/trezor-suite'
```
mroz22: I noticed that not all the jobs here have this check. should they have it? or is it enough only for the first job to have it?

mroz22 (commented):
it looks nice. will get back to it later to read through it more carefully mostly to make sure that I understand everything.

.github/workflows/release-connect-bump-versions.yml
karliatto: It is good idea to use typescript, I will experiment with that. 
But anyway doing `yarn install` is still required since we use for example library `semver` in some of the helper functions used in this script.

.github/workflows/release-connect-bump-versions.yml
karliatto: Good point. I actually think that `if: github.repository == 'trezor/trezor-suite'` should be only used in workflows that are run automatically like the tests ones, to avoid running when running in forks. But this one has to be triggered manually so I would say it is not necessarily.  @vdovhanych could you confirm it?

.github/workflows/release-connect-init.yml
karliatto: Yes.

.github/workflows/release-connect-bump-versions.yml
vdovhanych: Yes, we can get rid of it for release workflows

.github/workflows/release-connect-npm-init.yml
karliatto: You are right, not required, removed.

.github/workflows/release-connect-npm-init.yml
```diff
@@ -36,15 +61,21 @@ jobs:
...
 
       - name: Build dependencies
-        # Running `build:libs` requires action to checkout with `submodules: true`.
         run: yarn build:libs
 
       - name: Run @trezor/connect create npm release branch
         run: |
-          git config --global user.name "trezor-ci"
-          git config --global user.email "${{ secrets.TREZOR_BOT_EMAIL }}"
-          gh config set prompt disabled
```
karliatto, line 46: @vdovhanych getting rid of "trezor-ci" git user here, since I think this was totally unnecessary.

.github/workflows/release-connect-init.yml
```diff
@@ -0,0 +1,92 @@
...
+          # Checkout the specified commit
+          ref: ${{ github.event.inputs.commit_sha }}
+
+      - name: Create and push new branch
+        env:
+          BRANCH_NAME: "release/connect/${{ needs.extract-version.outputs.version }}"
+        run: |
+          echo ${{ env.BRANCH_NAME }}
+          git checkout -b ${{ env.BRANCH_NAME }}
+          git push origin ${{ env.BRANCH_NAME }}
```
karliatto: @vdovhanych as agreed I am not adding here the "trezor-ci" bot credential and name since this branch is going to be created by the user that triggers the action.

.github/workflows/release-connect-bump-versions.yml
karliatto: @mroz22 I make the scripts to use `tsx` I just added some simple typing for now but can keep typing it further.

ci/scripts/connect-bump-versions.ts
```diff
@@ -28,35 +30,12 @@ const getGitCommitByPackageName = (packageName, maxCount = 10) =>
...
-        exec('gh', ['auth', 'setup-git']);
-        exec('gh', ['config', 'set', '-h', 'github.com', 'oauth_token', process.env.GITHUB_TOKEN]);
-    } else {
-        throw new Error('Missing GITHUB_TOKEN');
-    }
-
+const bumpConnect = async () => {
     const checkResult = await checkPackageDependencies('connect', deploymentType);
 
     const update = checkResult.update.map(package => package.replace('@trezor/', ''));
```
mroz22: here I am seeing this in my VS code. not sure if its a big issue, just letting you know 
image

mroz22: also on line 52 
image

it looks this script is not really checked by ts?

mroz22 (approved):
the new system sounds good to me. I just pointed few nitpicks around the review, not really important ones.

Maybe please wait also for @vdovhanych and his second approval 🙏

.github/workflows/release-connect-bump-versions.yml
```diff
@@ -0,0 +1,37 @@
+name: "[Release] connect bump versions"
```
vdovhanych, line 1: This will run and create a new unprotected branch with updated versions and create a pull request to the protected branch?

.github/workflows/release-connect-init.yml
vdovhanych: How is this supposed to work? Its trying to push te code to protected branch. It wont work like this, we don't want to have automation rights to push commits to protected branch.

.github/workflows/template-check-connect-version-match.yml
```diff
@@ -0,0 +1,37 @@
+name: Check connect version match with branch
```
vdovhanych: This should probably be called `[Template] ...`

vdovhanych: Also i would probably convert this to action as this is something that is quite static  and wont be changing that often (but its super low priority)

.github/workflows/release-connect-npm-init.yml
vdovhanych: Ok

.github/workflows/release-connect-init.yml
```diff
@@ -0,0 +1,92 @@
...
+        uses: actions/checkout@v4
+        with:
+          # Ensure the full commit history is available is required to get specific `ref`.
+          fetch-depth: 0
+          # Checkout the specified commit
+          ref: ${{ github.event.inputs.commit_sha }}
+
+      - name: Create and push new branch
+        env:
+          BRANCH_NAME: "release/connect/${{ needs.extract-version.outputs.version }}"
```
vdovhanych, line 62: This is where what branch its pushing to is specified. W probably want to push the code to `create-push-release-branch:` branch that was created in that job at the line 46.

.github/workflows/release-connect-init.yml
```diff
@@ -0,0 +1,92 @@
...
+        run: |
+          echo ${{ env.BRANCH_NAME }}
+          git checkout -b ${{ env.BRANCH_NAME }}
+          git push origin ${{ env.BRANCH_NAME }}
+          echo "branch_name=${{ env.BRANCH_NAME }}" >> $GITHUB_OUTPUT
+
+  trigger-v9-staging-release:
```
vdovhanych: Im not sure if this will do what we want, if we trigger the workflows like this it will still be possible to approve the run by one person. I trigger release and i can approve it myself because the deploy was requested by the `github-actions` "user". This is valied for all following trigger jobs

.github/workflows/release-connect-npm-init.yml
```diff
@@ -36,15 +61,21 @@ jobs:
...
-          gh api /user --jq .login
-          node ./ci/scripts/connect-release-init-npm.js ${{ github.event.inputs.semver }}
+          yarn tsx ./ci/scripts/connect-release-npm-init.ts ${{ github.event.inputs.deploymentType }} ${{ env.BRANCH_NAME }}
+        env:
+          BRANCH_NAME: release/connect/${{ needs.extract-version.outputs.version }}
+
+  trigger-npm-release-connect:
```
vdovhanych: same here with te triggering the action with github actions.

.github/workflows/release-connect-npm-init.yml
```diff
@@ -36,15 +61,21 @@ jobs:
...
         run: yarn build:libs
 
       - name: Run @trezor/connect create npm release branch
         run: |
-          git config --global user.name "trezor-ci"
-          git config --global user.email "${{ secrets.TREZOR_BOT_EMAIL }}"
-          gh config set prompt disabled
-          gh api /user --jq .login
-          node ./ci/scripts/connect-release-init-npm.js ${{ github.event.inputs.semver }}
+          yarn tsx ./ci/scripts/connect-release-npm-init.ts ${{ github.event.inputs.deploymentType }} ${{ env.BRANCH_NAME }}
```
vdovhanych: This again tries to push the code to the protected branch. Or I'm missing something?

ci/scripts/connect-bump-versions.ts
karliatto: Those are types issues, because it was `js` and just made it `ts` so I am typing it a bit more to satisfy those.

.github/workflows/release-connect-bump-versions.yml
karliatto: This will create a new unprotected branch with update versions and create a pull request to `develop`.
Then we merge it to develop and once it is in develop we run the other workflow [.github/workflows/release-connect-init.yml](https://github.com/trezor/trezor-suite/pull/12444/files/cf2d60fa1b195f90ad1f0602988524c321fab783#diff-5d3b3dd645ad50a3cd17a1ab1ff0c12c48f4c39e0a328c4a13adadfa9602e779) to create the protected branch `release/connect/...`

.github/workflows/template-check-connect-version-match.yml
karliatto: Yes, I made new action `.github/actions/check-connect-version-match/action.yml`

.github/workflows/release-connect-init.yml
karliatto: I see, I think you are right.
I am researching if there is a way of using the same user that triggered the action as the one that trigger it.

.github/workflows/release-connect-npm-init.yml
karliatto: No, this is not pushing code to any protected branch. This is triggering the workflow that will initialize the release of NPM.

It requires to have a script that checks what are the dependencies that need to be updated.

Please look at the section `## Release process` in the description of this PR, I am trying to explain there the process. And let' me know if you have doubts or ideas.

.github/workflows/release-connect-init.yml
karliatto: I think I have an idea, what if instead of running those workflows using `gh` I will make those workflows templates that will run in the original workflow [release-connect-init.yml](https://github.com/trezor/trezor-suite/pull/12444/files#diff-5d3b3dd645ad50a3cd17a1ab1ff0c12c48f4c39e0a328c4a13adadfa9602e779)  ?

Then it will all be run by the user that is triggering the initial workflow, right?

karliatto: Now it seams even more right in my mind.

karliatto: Maybe it is problematic with "approval" since the approval are per workflow not per job. Right?
And in my mind this is design to approve each workflow at a different moment.

karliatto: Or we can do the process a bit more manual and trigger those workflows manually  at least for now.

karliatto: Ok, for now I suggest get rid of the automation of triggering all the workflows from one and trigger them manually until we decide how to do it in order to achieve that each of the releases part has to be trigger by one authorize user and approve by other.

Therefore I added this fixup to remove all those trigger... jobs https://github.com/trezor/trezor-suite/pull/12444/commits/3029015c5037b1aef0f990ebcf3b77ee0b1cac18

@vdovhanych if you agree I will squash that fixup and we can try manually then we will see.

.github/workflows/release-connect-bump-versions.yml
vdovhanych: ok

.github/workflows/release-connect-init.yml
vdovhanych: What about this still?

.github/workflows/release-connect-npm-init.yml
vdovhanych: Ok

.github/workflows/release-connect-npm-init.yml
vdovhanych: I dont think this is resolved

karliatto: You are right, just getting rid of that file completely since this time we are going to run those workflows manually https://github.com/trezor/trezor-suite/pull/12444/commits/95ffe13723f84948396df291ac8c6022c967a6c8

.github/workflows/release-connect-init.yml
karliatto: This will now be done by the "trezor-ci".

vdovhanych (approved):

PR #12445, opened 2024-05-16
feat(fiat-services): prevent duplicate parallel requests -> speed up Suite loading

martykan:
## Description

I noticed that the QA wallet with all accounts enabled is loading very slowly when opening Suite, about 1m20s.
This is mostly due to loading fiat rates during account discovery.
However, it seems that a lot of the requests are duplicated, which is unnecessary and slows down the loading process.

To resolve this, the PR adds `ParallelRequestsCache`, a class which uses Deferreds to prevent duplicate requests.
To use it, the function is wrapped in `ParallelRequestsCache.cache` with the first argument being the key for the request.
This is applied to all fiat rate functions - `fetchCurrentFiatRates`, `fetchLastWeekFiatRates`, and `getFiatRatesForTimestamps`.

From my tests this cuts down the loading time to about 30-40 seconds, so approximately in half.

### DIFF
--- a/suite-common/fiat-services/src/cache.ts
+++ b/suite-common/fiat-services/src/cache.ts
@@ -1,0 +1,29 @@
+import { Deferred, createDeferred } from '@trezor/utils';
+
+// Cache for parallel requests
+// It's used to prevent multiple requests for the same data
+export class ParallelRequestsCache {
+    private promises: Record<string, Deferred<any>> = {};
+
+    async cache<T>(keys: (string | number | undefined)[], fn: () => Promise<T>): Promise<T> {
+        const key = keys.join('-');
+        if (this.promises[key]) {
+            // Cache hit
+            return this.promises[key].promise;
+        }
+
+        this.promises[key] = createDeferred();
+        try {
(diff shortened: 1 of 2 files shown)


AdamSchinzel (approved):
Awesome! With `coin-definitions` we should rarely hit 404 (on Cardano they are currently commented out) but it's nice improvement so why not have it

bosomt:
**QA OK**

i have not compared it our wallet loads really fast now

**Info:**
 - Suite version: desktop 24.6.0 (1ec4c5bfc5652f5d58f159ba6160e2e47f525e3d)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.6.0 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)

PR #12448, opened 2024-05-16
Small actions improvementes and migration to using more deployment environments

vdovhanych:
- **chore(ci): rename suite-native workflows and configure new environment**
- **chore(ci): update connect-release action to upload artifacts for non prod builds only**
- **chore(ci): configure connect release workflows to use new environments**
- **chore(ci): configure release definitions workflow to use new environment**
- **chore(ci): configure message release workflow to use new environments**
- **chore(ci): rename files to fit current naming convention**


.github/actions/release-connect/action.yml
```diff
@@ -102,6 +102,7 @@ runs:
         aws s3 sync --delete tmp_build_directory/ "${DEPLOY_PATH}"
 
     - name: Build connect-examples webextension
+      if: inputs.buildArtifacts == 'true'
```
karliatto, line 113: Don't you have to define the `inputs` in the section `inputs` at the top of the action? https://github.com/trezor/trezor-suite/pull/12448/files#diff-997d2c0811918f861e8d73a45635dc4a734155cc91e3b16b8fcd6e7b632e8d70R3

Also I think if you define it as `boolean` you don't need to treat it as a string. As we are doing in https://github.com/trezor/trezor-suite/blob/develop/.github/workflows/template-connect-test-params.yml#L22

Or is it for some reason I am overlooking better to do it like that?

.github/workflows/release-connect-v9-staging.yml
```diff
@@ -50,7 +50,7 @@ jobs:
   # This job deploys to staging-connect.trezor.io/9.x.x
   deploy-staging-semantic-version:
     needs: [extract-version, check-version-match]
-    environment: production
+    environment: staging-connect
```
karliatto, line 53: Cool, so if I understand well this new environment is not going to required approval from 2 authorized members right?

karliatto (changes requested):

.github/workflows/release-connect-v9-staging.yml
vdovhanych: Yes we can better manage the rules for different environments :)

.github/actions/release-connect/action.yml
vdovhanych: yes youre right, i like to use it as string instead fo bool because you are fixed in how the value is compared to bool where it can be `0` or `1` and `true` or `false` in different envs.

vdovhanych: fixed

karliatto (approved):
LGTM

matejkriz (approved):

PR #12452, opened 2024-05-16
fix(suite-native): device switcher pimping

vytick:
visual pimping based on [design QA](https://www.figma.com/design/mmDnYtctfjnreWhTjLfeQH/Passphrase?node-id=2699-11367&t=eH0teHpn36YuUFvn-0)

https://github.com/trezor/trezor-suite/assets/2011829/3bc29a3a-02aa-4158-b8b0-94d7d4f2147b

### DIFF
--- a/suite-native/atoms/src/TextDivider.tsx
+++ b/suite-native/atoms/src/TextDivider.tsx
@@ -1,5 +1,6 @@
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 import { Translation, TxKeyPath } from '@suite-native/intl';
+import { Color } from '@trezor/theme';
 
 import { Box } from './Box';
 import { HStack } from './Stack';
@@ -8,11 +9,13 @@ import { Text } from './Text';
 type TextDividerProps = {
     title: TxKeyPath;
     horizontalMargin?: number;
+    lineColor?: Color;
+    textColor?: Color;
 };
 
-const separatorStyle = prepareNativeStyle<{ horizontalMargin?: number }>(
-    (utils, { horizontalMargin }) => ({
-        backgroundColor: utils.colors.borderElevation0,
+const separatorStyle = prepareNativeStyle<{ horizontalMargin?: number; color: Color }>(
+    (utils, { horizontalMargin, color }) => ({
+        backgroundColor: utils.colors[color],
         height: utils.borders.widths.small,
         flex: 1,
         // We want the separator to be full width, but we need to offset it by the parent padding
@@ -25,18 +28,23 @@ const separatorTitleStyle = prepareNativeStyle(utils => ({
     paddingVertical: utils.spacings.extraSmall,
 }));
 
-export const TextDivider = ({ title, horizontalMargin = 0 }: TextDividerProps) => {
+export const TextDivider = ({
+    title,
+    horizontalMargin = 0,
+    lineColor = 'borderElevation1',
+    textColor = 'textDefault',
+}: TextDividerProps) => {
     const { applyStyle } = useNativeStyles();
 
     return (
         <HStack alignItems="center">
-            <Box style={applyStyle(separatorStyle, { horizontalMargin })} />
+            <Box style={applyStyle(separatorStyle, { horizontalMargin, color: lineColor })} />
             <Box style={applyStyle(separatorTitleStyle)}>
-                <Text variant="label">
+                <Text variant="label" color={textColor}>
                     <Translation id={title} />
                 </Text>
             </Box>
-            <Box style={applyStyle(separatorStyle, { horizontalMargin })} />
+            <Box style={applyStyle(separatorStyle, { horizontalMargin, color: lineColor })} />
         </HStack>
     );
 };
--- a/suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
+++ b/suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
@@ -1,6 +1,6 @@
 import { useSelector } from 'react-redux';
 
-import { Icon, IconSize } from '@suite-common/icons';
+import { Icon } from '@suite-common/icons';
 import {
     DeviceRootState,
     PORTFOLIO_TRACKER_DEVICE_ID,
@@ -12,20 +12,21 @@ import { DeviceModelIcon } from '../DeviceModelIcon';
 
 type DeviceItemIconProps = {
     deviceId: TrezorDevice['id'];
-    iconSize: IconSize;
 };
 
-export const DeviceItemIcon = ({ deviceId, iconSize }: DeviceItemIconProps) => {
+const ICON_SIZE = 28;
+
+export const DeviceItemIcon = ({ deviceId }: DeviceItemIconProps) => {
     const deviceModel = useSelector((state: DeviceRootState) =>
         selectDeviceModelById(state, deviceId),
     );
 
     if (deviceId === PORTFOLIO_TRACKER_DEVICE_ID) {
-        return <Icon name="database" color="iconDefault" size={iconSize} />;
+        return <Icon name="database" color="iconDefault" size={ICON_SIZE} />;
     }
     if (deviceModel !== null) {
-        return <DeviceModelIcon deviceModel={deviceModel} size={iconSize} />;
+        return <DeviceModelIcon deviceModel={deviceModel} size={ICON_SIZE} />;
     }
 
-    return <Icon name="trezor" color="iconDefault" size={iconSize} />;
+    return <Icon name="trezor" color="iconDefault" size={ICON_SIZE} />;
 };
--- a/suite-native/device-manager/src/components/DeviceItem/WalletDetailDeviceItemContent.tsx
+++ b/suite-native/device-manager/src/components/DeviceItem/WalletDetailDeviceItemContent.tsx
@@ -1,7 +1,7 @@
 import { useSelector } from 'react-redux';
 import { ReactNode } from 'react';
 
-import { HStack, Text, Box } from '@suite-native/atoms';
+import { HStack, Text } from '@suite-native/atoms';
 import { Translation } from '@suite-native/intl';
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 import { DeviceRootState, selectDeviceByState } from '@suite-common/wallet-core';
@@ -15,7 +15,6 @@ export type WalletDetailDeviceItemContentProps = {
     headerTextVariant?: TypographyStyle;
     header: ReactNode;
     subHeader?: ReactNode;
-    isPortfolioLabelDisplayed?: boolean;
     isPortfolioTrackerDevice: boolean;
 };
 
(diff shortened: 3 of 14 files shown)


suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
```diff
@@ -12,7 +12,7 @@ import { DeviceModelIcon } from '../DeviceModelIcon';
 
 type DeviceItemIconProps = {
     deviceId: TrezorDevice['id'];
-    iconSize: IconSize;
+    iconSize: IconSize | number;
 };
 
 export const DeviceItemIcon = ({ deviceId, iconSize }: DeviceItemIconProps) => {
```
PeKne: nit: What about setting the size as default, so if the component is used elsewhere, we don't have to always set the size via prop?
```suggestion
export const DeviceItemIcon = ({ deviceId, iconSize = 28 }: DeviceItemIconProps) => {
```

suite-native/atoms/src/TextDivider.tsx
```diff
@@ -12,7 +12,7 @@ type TextDividerProps = {
 
 const separatorStyle = prepareNativeStyle<{ horizontalMargin?: number }>(
     (utils, { horizontalMargin }) => ({
-        backgroundColor: utils.colors.borderElevation0,
+        backgroundColor: utils.colors.borderElevation1,
```
PeKne: Don't this break other occurences across the app? Maybe this should be configurable by a prop.

suite-native/device-manager/src/components/DeviceItem/WalletDetailDeviceItemContent.tsx
```diff
@@ -58,20 +57,12 @@ export const WalletDetailDeviceItemContent = ({
...
-                        <Text variant="label" color="textSubdued">
-                            {subHeader}
-                        </Text>
-                    </HStack>
+            <Text variant="hint" color="textSubdued">
+                {isPortfolioTrackerDevice && (
+                    <Translation id="deviceManager.status.portfolioTracker" />
                 )}
-            </Box>
+                {!isPortfolioTrackerDevice && <>{subHeader}</>}
```
PeKne: is the <></> fragment needed? It should not be if I'm not mistaken

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -78,19 +79,19 @@ export const DeviceManagerContent = () => {
...
         >
-            <VStack spacing="medium">
+            <VStack spacing={12}>
                 <DeviceList
                     isVisible={isChangeDeviceRequested || isPortfolioTrackerDevice}
                     onSelectDevice={handleSelectDevice}
                 />
                 {!isPortfolioTrackerDevice && (
-                    <>
+                    <View style={{ gap: 12, paddingTop: 24 }}>
```
PeKne: do not use Inline styles
```suggestion
                    <VStack spacing={12} paddingTop="large" />
```

suite-native/device-manager/src/components/DeviceManagerModal.tsx
```diff
@@ -79,27 +79,30 @@ export const DeviceManagerModal = ({
...
-                        </ScreenHeaderWrapper>
+                        <Pressable onPress={handleClose}>
+                            <ScreenHeaderWrapper>
+                                <HStack
+                                    justifyContent="space-between"
+                                    alignItems="center"
+                                    spacing="medium"
+                                    flex={1}
+                                >
+                                    <Box style={{ flexShrink: 1 }}>
```
PeKne:
```suggestion
                                    <Box flexShrink={1}>
```

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -13,8 +13,12 @@ type WalletListProps = {
 export const WalletList = ({ onSelectDevice }: WalletListProps) => {
     const devices = useSelector(selectDeviceInstances);
 
+    if (devices.length < 2) {
```
PeKne, line 17: explain this by a comment please.

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -142,7 +152,11 @@ export const DeviceList = ({ isVisible, onSelectDevice }: DeviceListProps) => {
 
     return (
         <Animated.View style={[listAnimatedStyle, applyStyle(listStaticStyle)]}>
-            <VStack>
+            <View
+                style={applyStyle(contentWrapperStyle, {
```
PeKne: Could no this be styled few lines lower in the conditional render? IMO it would be more readable since this style is applied under the same condition as the conditional render. I would prefer to isolate the conditional changes to single place.

PeKne (changes requested):

suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
vytick: changed in f72721d0be

suite-native/atoms/src/TextDivider.tsx
vytick: I have made colors default and configurable Also adjusted in other occurrences
f72721d0be

suite-native/device-manager/src/components/DeviceItem/WalletDetailDeviceItemContent.tsx
vytick: f72721d0be

suite-native/device-manager/src/components/DeviceManagerContent.tsx
vytick: f72721d0be

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: f72721d0be

suite-native/device-manager/src/components/WalletList.tsx
vytick: f72721d0be

suite-native/device-manager/src/components/DeviceList.tsx
vytick: good thinking. changed in f72721d0be

suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
PeKne: I was ok with leaving the prop there, just assigning a default to it. But this is also good. When we need to adjust size later for different use cases, we can add the prop later.

suite-native/device-manager/src/components/DeviceManagerModal.tsx
```diff
@@ -79,27 +79,30 @@ export const DeviceManagerModal = ({
...
+                        <Pressable onPress={handleClose}>
+                            <ScreenHeaderWrapper>
+                                <HStack
+                                    justifyContent="space-between"
+                                    alignItems="center"
+                                    spacing="medium"
+                                    flex={1}
+                                >
+                                    <Box flexShrink={1}>
+                                        {deviceState && (
```
PeKne, line 91: nit: Should not the deviceState conditionally render also the parent Box? If it will be false, then there will be a unnecessary empty Box. Is it expected?

suite-native/device-manager/src/components/DeviceManagerModal.tsx
```diff
@@ -79,27 +79,30 @@ export const DeviceManagerModal = ({
...
-                        </ScreenHeaderWrapper>
+                        <Pressable onPress={handleClose}>
+                            <ScreenHeaderWrapper>
+                                <HStack
+                                    justifyContent="space-between"
+                                    alignItems="center"
+                                    spacing="medium"
+                                    flex={1}
+                                >
+                                    <Box flexShrink={1}>
```
PeKne: nit: Should not the deviceState conditionally render also the parent Box? If it will be false, then there will be a unnecessary empty Box. Is it expected?
```suggestion
                                    { deviceState & <Box flexShrink={1}>
```

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -140,42 +137,42 @@ export const DeviceList = ({ isVisible, onSelectDevice }: DeviceListProps) => {
         return null;
     }
 
+    const hasNotSelectedDevicesToShow = notSelectedInstancelessDevices.length > 0;
```
PeKne: Please change the name of this prop. It is confusing. I read it as "It has not any device to show" --> there are no such devices.

But it actually should be true if there are some devices. 

What about:
```suggestion
    const hasUnselectedDevices = notSelectedInstancelessDevices.length > 0;
```

or something similar.

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -93,25 +94,21 @@ export const DeviceList = ({ isVisible, onSelectDevice }: DeviceListProps) => {
         if (isVisible) {
             const hasNotSelectedInstancelessDevices = notSelectedInstancelessDevices.length > 0;
```
PeKne: You are defining just the same variable few lines under this useEffect. Remove it here, and define it only once.

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -140,42 +137,42 @@ export const DeviceList = ({ isVisible, onSelectDevice }: DeviceListProps) => {
...
-
-                {A.isEmpty(notSelectedInstancelessDevices) ? (
+            <Box>
+                {notSelectedInstancelessDevices.map(d => {
+                    if (d.state === undefined) {
+                        return null;
+                    }
+
+                    return (
+                        <DeviceItem
```
PeKne: nit: Could be simplified to omit the nullish check at the beginning:
```suggestion
                        {d.state & <DeviceItem
```

PeKne (commented):
Few more minor things. Then we are ready to go 🏁.

suite-native/device-manager/src/components/DeviceItem/DeviceItemIcon.tsx
vytick: Yeah. That was exactly my thinking when I made it default first. The prop was obsolete after other changes in previous PR

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: f312d14b01

suite-native/device-manager/src/components/DeviceList.tsx
vytick: f312d14b01

suite-native/device-manager/src/components/DeviceList.tsx
vytick: f312d14b01

suite-native/device-manager/src/components/DeviceList.tsx
vytick: f312d14b01

PeKne (approved):

PR #12453, opened 2024-05-16
Handle conversion of non-checksummed eth address

### LINKED ISSUE #7407: Automatic conversion of non-checksummed address
**Design:**
convert-checksum

**Spec:**
As a user if I type or paste an Ethereum address in invalid format, Trezor Suite needs to convert the address to a supported format automatically and inform me about it.

When I paste or manually enter valid non-checksummed Ethereum address to the Ethereum Send address field, Trezor Suite converts it to checksummed address

I need to be informed that such automatic conversion happened below the Send address field

**Acceptance criteria**

- Trezor Suite checks the address and converts it to a checksummed version automatically
- Below the recipient’s address on the left side I see blue double check mark which indicates the address was checked as valid and also converted to the checksummed address
- Next to the double check mark I see blue text “Converted to checksum address”
- On the right side I see “Learn more” button which redirects to the KB article ([https://trezor.io/support/a/ethereum-convert-checksum-address](https://trezor.io/support/a/ethereum-convert-checksum-address))
- When I paste or manually enter invalid Ethereum address to the Ethereum Send address field, Trezor Suite shows an error the same way as it does now

enjojoy:
## Description

Based on the conversation in the issue, I've decided to do the following:

1. Check whether the address is a valid ethereum address (already implemented)
2. Check whether the address is used via blockbook
    a. If the address is used but unchecksummed or incorrectly checksummed, then Suite will automatically convert the address to the correct checksummed form and inform the user:
image
    b. if the address haven't been used yet and is not checksummed at all, offer to checksum with a button:
image

...WOP

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/7407

### DIFF
--- a/packages/components/src/components/form/BottomText.tsx
+++ b/packages/components/src/components/form/BottomText.tsx
@@ -10,14 +10,14 @@ import { InputState } from './inputTypes';
 export const BOTTOM_TEXT_MIN_HEIGHT = 26; // 1 line of text + top padding
 
 const slideDown = keyframes`
-    from {
-        opacity: 0;
-        transform: translateY(-2px);
-    }
-    to {
-        opacity: 1;
-        transform: translateY(0);
-    }
+from {
+    opacity: 0;
+    transform: translateY(-2px);
+}
+to {
+    opacity: 1;
+    transform: translateY(0);
+}
 `;
 
 export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: boolean }>`
@@ -36,25 +36,31 @@ interface BottomTextProps {
     inputState?: InputState;
     isDisabled?: boolean;
     icon?: IconName;
+    withIcon?: boolean;
     children: ReactNode;
 }
-
 export const BottomText = ({
     inputState,
     isDisabled,
-    icon = 'warningCircle',
     children,
+    icon,
+    withIcon,
 }: BottomTextProps) => {
     const theme = useTheme();
+    const iconMap = {
+        error: 'warningCircle',
+        warning: 'warningCircle',
+    } as const satisfies Record<InputState, IconName>;
 
     const iconColor: Color | CSSColor = isDisabled
         ? 'iconDisabled'
         : getInputStateTextColor(inputState, theme);
 
+    const iconName = icon || (inputState && iconMap[inputState]);
+
     return (
         <Container $inputState={inputState} $isDisabled={isDisabled}>
-            {icon && <Icon name={icon} size="medium" color={iconColor} />}
-
+            {withIcon && iconName && <Icon name={iconName} size="medium" color={iconColor} />}
             {children}
         </Container>
     );
--- a/packages/components/src/components/form/Input/Input.tsx
+++ b/packages/components/src/components/form/Input/Input.tsx
@@ -203,7 +203,11 @@ const Input = ({
             </InputWrapper>
 
             {bottomText && (
-                <BottomText inputState={inputState} isDisabled={isDisabled}>
+                <BottomText
+                    withIcon={inputState && ['error', 'warning'].includes(inputState)}
+                    inputState={inputState}
+                    isDisabled={isDisabled}
+                >
                     {bottomText}
                 </BottomText>
             )}
--- a/packages/suite/src/components/wallet/InputChecksumInfo.tsx
+++ b/packages/suite/src/components/wallet/InputChecksumInfo.tsx
@@ -1,0 +1,75 @@
+import styled, { css } from 'styled-components';
+import { useTheme } from 'styled-components';
+import { TooltipSymbol } from '../suite';
+import { spacingsPx, borders } from '@trezor/theme';
+import { TranslationKey, Translation } from '../suite/Translation';
+import { useTranslation } from 'src/hooks/suite';
+import { Icon } from '@suite-common/icons/src/webComponents';
+
+const Wrapper = styled.div`
+    display: flex;
+    align-items: center;
+    gap: ${spacingsPx.xs};
+    width: 100%;
+    justify-content: space-between;
+    background-color: ${({ theme }) => theme.backgroundAlertBlueSubtleOnElevation1};
+    border-radius: ${borders.radii.xxs};
+    position: relative;
+    margin: ${spacingsPx.xxs} 0 ${spacingsPx.sm} 0;
+    padding: ${spacingsPx.xs} ${spacingsPx.sm};
+    ${({ theme }) => css`
+        ::before {
+            content: '';
+            position: absolute;
+            top: -${spacingsPx.xs};
+            left: 20px;
+            transform: translateX(-50%);
+            width: 0;
+            height: 0;
+            border-left: 8px solid transparent;
+            border-right: 8px solid transparent;
+            border-bottom: 8px solid ${theme.backgroundAlertBlueSubtleOnElevation1};
+        }
+    `};
+`;
+
+const Container = styled.div`
+    display: flex;
+    flex-direction: row;
+    margin: 0;
+    padding: 0;
+`;
+
+export const Divider = styled.div`
+    height: inherit;
+    align-self: stretch;
+    width: 2px;
+    background-color: ${({ theme }) => theme.backgroundSurfaceElevation1};
+    margin: -${spacingsPx.xs} ${spacingsPx.xs};
+`;
+
+interface InputChecksumInfoProps {
+    message?: TranslationKey;
+    tooltipContent?: TranslationKey;
+}
+export const InputChecksumInfo = ({ message, tooltipContent }: InputChecksumInfoProps) => {
+    const { translationString } = useTranslation();
+    const theme = useTheme();
+
+    return (
+        <Wrapper>
+            <Container>
+                <Icon name="doubleCheck" size="medium" color={theme.textAlertBlue} />
+                <Divider />
+                {message && <Translation id={message} />}
+            </Container>
+            {tooltipContent && (
+                <TooltipSymbol
+                    content={translationString(tooltipContent)}
+                    iconColor={theme.textAlertBlue}
+                    icon="QUESTION_FILLED"
+                />
+            )}
+        </Wrapper>
+    );
+};
--- a/packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
+++ b/packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
@@ -8,6 +8,7 @@ import { scanOrRequestSendFormThunk } from 'src/actions/wallet/send/sendFormThun
 import { useDevice, useDispatch, useTranslation } from 'src/hooks/suite';
 import { useSendFormContext } from 'src/hooks/wallet';
 import { getProtocolInfo } from 'src/utils/suite/protocol';
+import TrezorConnect from '@trezor/connect';
 import {
     isAddressValid,
     isAddressDeprecated,
@@ -20,7 +21,7 @@ import { PROTOCOL_TO_NETWORK } from 'src/constants/suite/protocol';
 import { notificationsActions } from '@suite-common/toast-notifications';
 
 import type { Output } from '@suite-common/wallet-types';
-import { InputError } from 'src/components/wallet';
+import { InputError, InputChecksumInfo } from 'src/components/wallet';
 
 const Container = styled.div`
     position: relative;
@@ -51,6 +52,7 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
     const [addressDeprecatedUrl, setAddressDeprecatedUrl] = useState<
         AddressDeprecatedUrl | undefined
     >(undefined);
+    const [hasAddressChecksummed, setHasAddressChecksummed] = useState<boolean | undefined>();
     const dispatch = useDispatch();
     const { device } = useDevice();
     const {
@@ -66,7 +68,6 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
         setDraftSaveRequest,
     } = useSendFormContext();
     const { translationString } = useTranslation();
-
     const { descriptor, networkType, symbol } = account;
     const inputName = `outputs.${outputId}.address` as const;
     // NOTE: compose errors are always associated with the amount.
@@ -81,7 +82,13 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
     const address = watch(inputName);
     const options = getDefaultValue('options', []);
     const broadcastEnabled = options.includes('broadcast');
-    const inputState = getInputState(addressError);
+    const getInputErrorState = () => {
+        if (addressError) {
+            return getInputState(addressError);
+        }
+
+        return undefined;
+    };
 
     const handleQrClick = useCallback(async () => {
         const uri = await dispatch(scanOrRequestSendFormThunk()).unwrap();
@@ -89,7 +96,6 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
         if (typeof uri !== 'string') {
             return;
         }
-
         const protocol = getProtocolInfo(uri);
 
         if (protocol) {
@@ -142,10 +148,12 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                 break;
             case 'checksum':
                 return {
-                    onClick: () =>
+                    onClick: () => {
                         setValue(inputName, toChecksumAddress(address), {
                             shouldValidate: true,
-                        }),
+                        });
+                        setHasAddressChecksummed(true);
+                    },
                     text: translationString('TR_CONVERT_TO_CHECKSUM_ADDRESS'),
                 };
 
@@ -163,7 +171,10 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
     };
 
     const { ref: inputRef, ...inputField } = register(inputName, {
-        onChange: () => composeTransaction(amountInputName),
+        onChange: () => {
+            composeTransaction(amountInputName);
+            setHasAddressChecksummed(false);
+        },
         required: translationString('RECIPIENT_IS_NOT_SET'),
         validate: {
             deprecated: (value: string) => {
@@ -198,9 +209,30 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                 }
             },
             // eth addresses are valid without checksum but Trezor displays them as checksummed
-            checksum: (value: string) => {
-                if (networkType === 'ethereum' && !checkAddressCheckSum(value)) {
-                    return translationString('RECIPIENT_IS_NOT_VALID');
+            checksum: async (address: string) => {
+                if (networkType === 'ethereum') {
+                    const params = {
+                        descriptor: address,
+                        coin: symbol,
+                    };
+
+                    //1.If the address is used but unchecksummed, then Suite will automatically convert the address to the correct checksummed form and inform the user as described in the OP.
+                    const result = await TrezorConnect.getAccountInfo(params); //TODO handle error
+                    if (result.success) {
+                        if (result.payload.history.total !== 0 && !checkAddressCheckSum(address)) {
+                            setValue(inputName, toChecksumAddress(address), {
+                                shouldValidate: true,
+                            });
+                            setHasAddressChecksummed(true);
+                        }
+                        //2. if the address is not checksummed at all and not found in blockbook
+                        //offer to checksum it with a button
+                        if (result.payload.history.total == 0) {
+                            if (address === address.toLowerCase()) {
+                                return translationString('TR_ETH_ADDRESS_NOT_USED_NOT_CHECKSUMMED');
+                            }
+                        }
+                    }
                 }
             },
             rippleToSelf: (value: string) => {
@@ -222,10 +254,28 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
     });
     const addressBottomText = isAddressWithLabel ? addressLabelComponent : null;
 
+    const getBottomText = () => {
+        if (addressError) {
+            return (
+                <InputError message={addressError.message} button={getValidationButtonProps()} />
+            );
+        }
+        if (hasAddressChecksummed) {
+            return (
+                <InputChecksumInfo
+                    message="TR_CHECKSUM_CONVERSION_INFO"
+                    tooltipContent="TR_ETH_ADDRESS_CHECKSUMMED_INFO"
+                />
+            );
+        }
+
+        return addressBottomText;
+    };
+
     return (
         <Container>
             <Input
-                inputState={inputState}
+                inputState={getInputErrorState()}
                 innerAddon={
                     metadataEnabled && broadcastEnabled ? (
                         <MetadataLabelingWrapper>
@@ -287,16 +337,7 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                         />
                     ) : undefined
                 }
-                bottomText={
-                    addressError ? (
-                        <InputError
-                            message={addressError.message}
-                            button={getValidationButtonProps()}
-                        />
-                    ) : (
-                        addressBottomText
-                    )
-                }
+                bottomText={getBottomText()}
                 data-test={inputName}
                 defaultValue={addressValue}
                 maxLength={formInputsMaxLength.address}
--- a/suite-common/icons/assets/icons/doubleCheck.svg
+++ b/suite-common/icons/assets/icons/doubleCheck.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16">
+  <path stroke="#171717" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="m9.25 5.25-5.5 5.5L1 8m14-2.75-5.5 5.5-1.463-1.463"/>
+</svg>
--- a/packages/suite/src/components/wallet/index.tsx
+++ b/packages/suite/src/components/wallet/index.tsx
@@ -1,6 +1,7 @@
 import { WalletLayout } from './WalletLayout/WalletLayout';
 import { WalletSubpageHeading } from './WalletLayout/WalletSubpageHeading';
 import { InputError } from './InputError';
+import { InputChecksumInfo } from './InputChecksumInfo';
 import { AccountExceptionLayout } from './AccountExceptionLayout';
 import { DiscoveryProgress } from './DiscoveryProgress';
 import { UtxoAnonymity } from './UtxoAnonymity';
@@ -16,6 +17,7 @@ export {
     DiscoveryProgress,
     withSelectedAccountLoaded,
     InputError,
+    InputChecksumInfo,
     AccountExceptionLayout,
     UtxoAnonymity,
     Pagination,
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -2016,6 +2016,10 @@ export default defineMessages({
         description: 'Placeholder in seed input asking user to pay attention to his device',
         id: 'TR_CHECK_YOUR_DEVICE',
     },
+    TR_CHECKSUM_CONVERSION_INFO: {
+        defaultMessage: 'Converted to checksum',
+        id: 'TR_CHECKSUM_CONVERSION_INFO',
+    },
     TR_CLEAR_ALL: {
         defaultMessage: 'Clear all',
         description: 'Clear form button',
@@ -2186,6 +2190,18 @@ export default defineMessages({
     TR_EDIT: {
         defaultMessage: 'Edit',
         id: 'TR_EDIT',
+    },
+    TR_ETH_ADDRESS_CHECKSUMMED_INFO: {
+        defaultMessage: 'Your address was automatically checksummed. Learn more...',
+        id: 'TR_ETH_ADDRESS_CHECKSUMMED_INFO',
+    },
+    TR_ETH_ADDRESS_CONVERTED_CHECKSUM: {
+        defaultMessage: 'Address converted to checksum format.',
+        id: 'TR_ETH_ADDRESS_CONVERTED_CHECKSUM',
+    },
+    TR_ETH_ADDRESS_NOT_USED_NOT_CHECKSUMMED: {
+        defaultMessage: 'Your address was not used and is not checksummed, please double-check.',
+        id: 'TR_ETH_ADDRESS_NOT_USED_NOT_CHECKSUMMED',
     },
     TR_NEEDS_ATTENTION_BOOTLOADER: {
         defaultMessage: 'Trezor is in Bootloader mode.',
--- a/suite-common/icons/assets/icons/info.svg
+++ b/suite-common/icons/assets/icons/info.svg
@@ -1,5 +1,5 @@
 <svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" fill="none" viewBox="0 0 16 16">
-  <path stroke="#0078AC" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M8 14A6 6 0 1 0 8 2a6 6 0 0 0 0 12"/>
-  <path stroke="#0078AC" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M7.5 7.5H8V11h.5"/>
-  <path fill="#1F83AD" d="M7.95 5.5a.75.75 0 1 0 0-1.5.75.75 0 0 0 0 1.5"/>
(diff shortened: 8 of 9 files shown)


enjojoy:
I'm not sure the double checkmark icon is present in the icons package, I need to add it there as said in the icons [readme](https://github.com/trezor/trezor-suite/blob/develop/suite-common/icons/README.md), so I can use it in the info bottom text, but I'm not sure I have an access to figma. Could you help me to do that please/give access to figma? @AdamSchinzel

enjojoy:
WOP

image
image

enjojoy:
If the address is not checksummed and haven't been used before, but valid (text message to be adjusted)

image

packages/components/src/components/form/InputStyles.ts
```diff
@@ -50,6 +52,8 @@ export const getInputStateBgColor = (
             return theme.backgroundAlertYellowSubtleOnElevation1;
         case 'error':
             return theme.backgroundAlertRedSubtleOnElevation1;
+        case 'info':
+            return theme.backgroundAlertBlueSubtleOnElevation1; //TODO: should we highlight the automatically chacksummed address?
```
AdamSchinzel: No need to highlight it and also see [this comment](https://github.com/trezor/trezor-suite/pull/12453#discussion_r1609646378) (I would remove this).

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +206,56 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                //1.If the address is used but unchecksummed or incorrectly checksummed, then Suite will automatically convert the address to the correct checksummed form and inform the user as described in the OP.
+                if (networkType === 'ethereum') {
+                    const params = {
+                        descriptor: value,
+                        coin: 'ethereum',
+                    };
+                    const result = await TrezorConnect.getAccountInfo(params); //TODO handle error
+                    if (result.success && result.payload.history.total !== 0) {
+                        if (!checkAddressCheckSum(value)) {
+                            console.log('used but not checksummed');
```
AdamSchinzel: Probably leftover

packages/components/src/components/form/BottomText.tsx
```diff
@@ -39,21 +69,24 @@ interface BottomTextProps {
...
 
     const iconColor: Color | CSSColor = isDisabled
         ? 'iconDisabled'
         : getInputStateTextColor(inputState, theme);
 
     return (
         <Container $inputState={inputState} $isDisabled={isDisabled}>
-            {icon && <Icon name={icon} size="medium" color={iconColor} />}
+            {iconMap[inputState] && (
+                <Icon name={iconMap[inputState]} size="medium" color={iconColor} />
```
AdamSchinzel: You can define `iconMap[inputState]` as const and then use on both places here.

Plus you should be getting some type errors right now. To fix them I suggest:
```
const iconMap = {
        error: 'warningCircle',
        warning: 'warningCircle',
        info: 'doubleCheck',
} as const satisfies Record<InputState, IconName>;

...

{inputState && iconMap[inputState] && (
       <Icon name={iconMap[inputState]} size="medium" color={iconColor} />
)}
```

packages/components/src/components/form/BottomText.tsx
```diff
@@ -39,21 +69,24 @@ interface BottomTextProps {
     children: ReactNode;
 }
 
-export const BottomText = ({
-    inputState,
-    isDisabled,
-    icon = 'warningCircle',
-    children,
-}: BottomTextProps) => {
+export const BottomText = ({ inputState, isDisabled, children }: BottomTextProps) => {
```
AdamSchinzel: Remove icon from `BottomTextProps` if you don't use it anymore

packages/components/src/components/form/BottomText.tsx
```diff
@@ -39,21 +69,24 @@ interface BottomTextProps {
...
         : getInputStateTextColor(inputState, theme);
 
     return (
         <Container $inputState={inputState} $isDisabled={isDisabled}>
-            {icon && <Icon name={icon} size="medium" color={iconColor} />}
+            {iconMap[inputState] && (
+                <Icon name={iconMap[inputState]} size="medium" color={iconColor} />
+            )}
+
+            {inputState == 'info' ? <Divider /> : undefined}
```
AdamSchinzel:
```suggestion
            {inputState === 'info'  && <Divider />}
```

packages/components/src/components/form/BottomText.tsx
```diff
@@ -30,6 +30,36 @@ export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: bo
         $isDisabled ? theme.textDisabled : getInputStateTextColor($inputState, theme)};
     ${typography.label}
     animation: ${slideDown} 0.18s ease-in-out forwards;
+    background-color: ${({ $inputState, theme }) =>
+        $inputState == 'info' ? getInputStateBgColor($inputState, theme, 1) : undefined};
+    border-radius: 4px;
```
AdamSchinzel:
```suggestion
    border-radius: ${borders.radii.xxs};
```

packages/components/src/components/form/BottomText.tsx
```diff
@@ -30,6 +30,36 @@ export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: bo
...
+
+    ${({ $inputState, theme }) =>
+        $inputState === 'info' &&
+        `
+        margin: ${spacingsPx.sm} 0;
+        padding: ${spacingsPx.xs} ${spacingsPx.sm};
+        ::before {
+            content: '';
+            position: absolute;
+            top: -8px;
```
AdamSchinzel:
```suggestion
            top: -${spacingsPx.xs};
```

packages/components/src/components/form/BottomText.tsx
```diff
@@ -30,6 +30,36 @@ export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: bo
         $isDisabled ? theme.textDisabled : getInputStateTextColor($inputState, theme)};
     ${typography.label}
     animation: ${slideDown} 0.18s ease-in-out forwards;
+    background-color: ${({ $inputState, theme }) =>
+        $inputState == 'info' ? getInputStateBgColor($inputState, theme, 1) : undefined};
```
AdamSchinzel: I think you don't need to modify `getInputStateBgColor` for info case since you are basically hardcoding the output color anyway.

So instead I would prefer to just put `backgroundAlertBlueSubtleOnElevation1` fot both background-color and border-bottom

packages/suite/src/components/wallet/InputInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+import styled from 'styled-components';
+import { useTheme } from 'styled-components';
+import { Button } from '@trezor/components';
+import { LearnMoreButton } from '../suite/LearnMoreButton';
+import { TooltipSymbol } from '../suite';
+import { getInputStateTextColor } from '@trezor/components';
+const Wrapper = styled.div`
+    display: flex;
+    align-items: center;
+    gap: 8px;
```
AdamSchinzel:
```suggestion
    gap: ${spacingsPx.xs};
```

AdamSchinzel: Also put an empty line between imports and `Wrapper`

packages/suite/src/components/wallet/InputInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+    width: 100%;
+    justify-content: space-between;
+`;
+
+interface InputInfoProps {
+    button?: { onClick: MouseEventHandler<HTMLButtonElement>; text: string } | { url: string };
+    message?: string;
+    tooltip?: { content: string };
```
AdamSchinzel: You don't need to put content inside an object. Just do `tooltip?: string;`

suite-common/icons/assets/icons/doubleCheck.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16">
+  <path stroke="#0078AC" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="m9.25 5.25-5.5 5.5L1 8m14-2.75-5.5 5.5-1.463-1.463"/>
```
AdamSchinzel: Don't put the icon here already with the color, use classic black color not the blue color.

```suggestion
  <path stroke="#171717" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="m9.25 5.25-5.5 5.5L1 8m14-2.75-5.5 5.5-1.463-1.463"/>
```

packages/suite/src/components/wallet/InputInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+const Wrapper = styled.div`
+    display: flex;
+    align-items: center;
+    gap: 8px;
+    width: 100%;
+    justify-content: space-between;
+`;
+
+interface InputInfoProps {
+    button?: { onClick: MouseEventHandler<HTMLButtonElement>; text: string } | { url: string };
```
AdamSchinzel: I know that you put this here for future usage and because you want to approach it in generic way but since you don't use this at all I think you can remove it completely

packages/suite/src/components/wallet/InputInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+    message?: string;
```
AdamSchinzel:
```suggestion
    message: TranslationKey;
```

packages/suite/src/components/wallet/InputInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+    tooltip?: { content: string };
+}
+export const InputInfo = ({ button, message, tooltip }: InputInfoProps) => {
+    const theme = useTheme();
+    const textColor = getInputStateTextColor('info', theme);
+
+    return (
+        <Wrapper>
+            {message}
```
AdamSchinzel:
```suggestion
            <Translation id={message} />
```

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -222,10 +277,28 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                <InputError message={addressError.message} button={getValidationButtonProps()} />
+            );
+        }
+
+        if (addressInfoContent) {
+            const mess = 'Converted to Checksum'; //TODO: add translation
+
+            const tooltip = { content: 'Info about checksum conversion' };
+
+            return <InputInfo message={mess} tooltip={tooltip} />;
```
AdamSchinzel: You can put both values right inside the component no need to put them in separated consts

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +206,56 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                 }
             },
             // eth addresses are valid without checksum but Trezor displays them as checksummed
-            checksum: (value: string) => {
-                if (networkType === 'ethereum' && !checkAddressCheckSum(value)) {
-                    return translationString('RECIPIENT_IS_NOT_VALID');
+            used_checksum: async (value: string) => {
```
AdamSchinzel: I don't really get the whole part of this plus you are not returning anything so types are complaining. Can you elaborate what it does and why you even use localStorage etc.

AdamSchinzel (changes requested):

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
AdamSchinzel: Also using `addressInfoContent` seems to me to not be really needed since you can return some error as a string and if it matches `addressError` then in `getInputState` you can return `info` (so you don't even need to modify input state to void).

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +206,56 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                        }
+                    }
+                }
+            },
+            checksum: async (value: string) => {
+                if (networkType === 'ethereum') {
+                    const params = {
+                        descriptor: value,
+                        coin: 'ethereum',
+                    };
```
tomasklim:
```suggestion
            checksum: async (address: string) => {
                if (networkType === 'ethereum') {
                    const params = {
                        descriptor: address,
                        coin: symbol,
                    };
```

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +206,56 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                    const result = await TrezorConnect.getAccountInfo(params);
+                    //2. if address is not checksummed at all and not found in blockbook
+                    //offer to checksum width a button
+                    if (result.success && result.payload.history.total == 0) {
+                        if (value == value.toLowerCase()) {
```
tomasklim:
```suggestion
                        if (address === address.toLowerCase()) {
```
what if it is not checksummed nor lowercase? e.g. its all uppercase 😀  or mixed case but not in the correct checksummed way

tomasklim (changes requested):
Needs some polish but proposed solution looks good

packages/components/src/components/form/InputStyles.ts
enjojoy: removed

packages/components/src/components/form/BottomText.tsx
enjojoy: Yep, I agree and fixed

packages/components/src/components/form/BottomText.tsx
enjojoy: thank you, fixed
https://github.com/trezor/trezor-suite/pull/12453/commits/090f4526bb963c42a2aafa667b013c68f1d1d5d7

packages/suite/src/components/wallet/InputInfo.tsx
enjojoy: oki, yep I thought it might be useful in the future, but we can add it later in case we need it. removed

packages/suite/src/components/wallet/InputInfo.tsx
enjojoy: true, i changed it to tooltipContent and now passing only a string

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
enjojoy: > I don't really get the whole part of this plus you are not returning anything so types are complaining. Can you elaborate what it does and why you even use localStorage etc.

I will remake the solution, it was my first iteration that worked, but i agree that it doesn't look good. I won't use localStorage either.

suite-common/icons/assets/icons/doubleCheck.svg
enjojoy: fixed, also  on the info icon

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
enjojoy: If it's all uppercase, it can be treated the same way as all lowercase. But in case of mixed case (but wrong checkdummed), it should be just invalid, because it means that the address was affected by a typo. By automatically checksumming such an address we put the user under a risk of using a non-existent address. Hence the user should double-check and fix it himself.

enjojoy: If it's incorrectly checksummed in mixed case it's catched much earlier than the checksum validation. So no matter if the account is used, the uncorrectly checksummed address will be not valid. Here: 

            valid: (value: string) => {
                if (!isAddressValid(value, symbol)) {
                    return translationString('RECIPIENT_IS_NOT_VALID');
                }
            },

enjojoy: If it's all upprercase it's a bit tricky. If the x in 0x is uppercase too, it's invalid, if the x is lowercase but the rest of the address is all uppercase, it gets converted to checksum. Is that the desired behavior?

packages/components/src/components/form/BottomText.tsx
```diff
@@ -30,30 +30,63 @@ export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: bo
...
-    icon = 'warningCircle',
-    children,
-}: BottomTextProps) => {
+export const BottomText = ({ inputState, isDisabled, children }: BottomTextProps) => {
     const theme = useTheme();
+    const iconMap = {
+        error: 'warningCircle',
+        warning: 'warningCircle',
+        info: 'doubleCheck',
+    } as const satisfies Record<InputState, IconName>;
```
jvaclavik: We can use default icon for each input state, but we should be able to choose any icon with different prop.

Please take a look on component Warning (`packages/components/src/components/Warning/Warning.tsx`) and refactor it in the same style.

Screenshot 2024-06-07 at 14 24 11

Screenshot 2024-06-07 at 14 25 58

packages/components/src/components/form/BottomText.tsx
```diff
@@ -30,30 +30,63 @@ export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: bo
...
+            top: -${spacingsPx.xs};
+            left: 20px;
+            transform: translateX(-50%);
+            width: 0;
+            height: 0;
+            border-left: 8px solid transparent; 
+            border-right: 8px solid transparent; 
+            border-bottom: 8px solid ${theme.backgroundAlertBlueSubtleOnElevation1}; 
+        }
+        `}
```
jvaclavik: I believe this shouldn't be here, because `BottomText` is generic component in the design system. 

I suggest – if you want to use some custom component – just pass `bottomText`: (as custom ReactNode not string) to `Input` and style it on your own outside of the design system.

Something like:

```tsx
<Input 
  ...
  bottomText={<CustomComponent/>}
/>
```

packages/components/src/components/form/BottomText.tsx
```diff
@@ -30,30 +30,63 @@ export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: bo
...
+    align-self: stretch;
+    width: 2px;
+    background-color: ${({ theme }) => theme.backgroundSurfaceElevation1};
+    margin: -${spacingsPx.xs} ${spacingsPx.xs};
 `;
 
 interface BottomTextProps {
     inputState?: InputState;
     isDisabled?: boolean;
-    icon?: IconName;
```
jvaclavik, line 38: Why do you need it? We want to keep this prop here.

packages/components/src/components/form/BottomText.tsx
```diff
@@ -30,30 +30,63 @@ export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: bo
...
         : getInputStateTextColor(inputState, theme);
 
     return (
         <Container $inputState={inputState} $isDisabled={isDisabled}>
-            {icon && <Icon name={icon} size="medium" color={iconColor} />}
+            {inputState && iconMap[inputState] && (
+                <Icon name={iconMap[inputState]} size="medium" color={iconColor} />
+            )}
+
+            {inputState === 'info' && <Divider />}
```
jvaclavik: This shouldn't be dependent on the same. Also if you need it create BottomText as a custom component or add generic prop `hasDivider`.

jvaclavik (changes requested):

packages/components/src/components/form/BottomText.tsx
enjojoy: I see what you mean. I will rather create a new custom component and pass it into BottomText & leave BottomText untouched besides the Warning-like refactoring, so we can get rid of the icon when needed & have default re writable icons.

packages/components/src/components/form/BottomText.tsx
enjojoy: Returned

packages/components/src/components/form/BottomText.tsx
jvaclavik: Sounds good 👍

packages/components/src/components/form/BottomText.tsx
enjojoy: This behavior is introduced now in BottomText component

```
 let iconName;

    if (icon) {
        iconName = icon;
    } else if (inputState && iconMap[inputState]) {
        iconName = iconMap[inputState];
    }

    return (
        <Container $inputState={inputState} $isDisabled={isDisabled}>
            {withIcon && iconName && <Icon name={iconName} size="medium" color={iconColor} />}
            
```

packages/components/src/components/form/BottomText.tsx
enjojoy: This is removed and a custom component is introduced

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -222,10 +286,29 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+        }
+
+        return addressBottomText;
+    };
+
     return (
         <Container>
             <Input
-                inputState={inputState}
+                inputState={inputState()}
```
tomasklim: nit:
```suggestion
                getInputState={inputState()}
```

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -222,10 +286,29 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+            return (
+                <InputError message={addressError.message} button={getValidationButtonProps()} />
+            );
+        }
+
+        if (addressChecksumInfo) {
+            return (
+                <InputChecksumInfo
+                    message="TR_CHECKSUM_CONVERSION_INFO"
+                    tooltipContent="Info about checksummed address"
```
tomasklim: Untranslatable
```suggestion
                    tooltipContent="SOME_TRANSLATION_KEY"
```

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +226,45 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                            setValue(inputName, toChecksumAddress(address), {
+                                shouldValidate: true,
+                            });
+                        }
+                    }
+                    //2. if address is not checksummed at all and not found in blockbook
+                    //offer to checksum width a button
+                    if (result.success && result.payload.history.total == 0) {
+                        if (address == address.toLowerCase()) {
+                            return 'Not used and not checksummed at all';
```
tomasklim: Untranslatable

packages/components/src/components/form/BottomText.tsx
```diff
@@ -32,29 +40,35 @@ export const Container = styled.div<{ $inputState?: InputState; $isDisabled?: bo
...
         : getInputStateTextColor(inputState, theme);
 
+    let iconName;
+
+    if (icon) {
+        iconName = icon;
+    } else if (inputState && iconMap[inputState]) {
+        iconName = iconMap[inputState];
+    }
+
```
tomasklim:
```suggestion
    const iconName = icon || (inputState && iconMap[inputState]);
```

packages/components/src/components/form/BottomText.tsx
```diff
@@ -9,6 +9,14 @@ import { InputState } from './inputTypes';
 
 export const BOTTOM_TEXT_MIN_HEIGHT = 26; // 1 line of text + top padding
 
+interface BottomTextProps {
```
tomasklim: nit: We put interfaces above components

packages/components/src/components/form/Input/Input.tsx
```diff
@@ -198,7 +198,11 @@ const Input = ({
             </InputWrapper>
 
             {bottomText && (
-                <BottomText inputState={inputState} isDisabled={isDisabled}>
+                <BottomText
+                    withIcon={inputState === 'error' || inputState === 'warning'}
```
tomasklim: nit:
```suggestion
                    withIcon={['error', 'warning'].includes(inputState)}
```

packages/suite/src/components/wallet/InputChecksumInfo.tsx
```diff
@@ -0,0 +1,74 @@
...
+    gap: ${spacingsPx.xs};
+    width: 100%;
+    justify-content: space-between;
+    background-color: ${({ theme }) => theme.backgroundAlertBlueSubtleOnElevation1};
+    border-radius: ${borders.radii.xxs};
+    position: relative;
+    margin: ${spacingsPx.xxs} 0 ${spacingsPx.sm} 0;
+    padding: ${spacingsPx.xs} ${spacingsPx.sm};
+    ${({ theme }) =>
+        `
```
tomasklim:
```suggestion
    ${({ theme }) => css`
```

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -47,10 +48,32 @@ interface AddressProps {
...
+
+    return { lastTwoValues, addValue };
+};
+
 export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
     const [addressDeprecatedUrl, setAddressDeprecatedUrl] = useState<
         AddressDeprecatedUrl | undefined
     >(undefined);
+    type AddressChecksumInfo = { text: string } | undefined;
+    const [addressChecksumInfo, setAddressChecksumInfo] = useState<AddressChecksumInfo>(undefined);
```
tomasklim: Why wrapping `text` into object?
`type` should not be inside component
```suggestion
const [addressChecksumInfo, setAddressChecksumInfo] = useState<string | undefined>();
```

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +226,45 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                    const result = await TrezorConnect.getAccountInfo(params); //TODO handle error
+                    if (result.success && result.payload.history.total !== 0) {
+                        if (!checkAddressCheckSum(address)) {
+                            setValue(inputName, toChecksumAddress(address), {
+                                shouldValidate: true,
+                            });
+                        }
+                    }
+                    //2. if address is not checksummed at all and not found in blockbook
+                    //offer to checksum width a button
```
tomasklim:
```suggestion
                    // 2. If the address is not checksummed at all and not found in Blockbook,
                    // offer to checksum it with a button.
```

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +226,45 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                    if (result.success && result.payload.history.total == 0) {
+                        if (address == address.toLowerCase()) {
```
tomasklim: use `===` + you can wrap it into ` if (result.success) {`
```suggestion
                    if (result.success) {
                        if (result.payload.history.total !== 0 && !checkAddressCheckSum(address)) {
                            setValue(inputName, toChecksumAddress(address), {
                                shouldValidate: true,
                            });
                        }
                  
                        if (result.payload.history.total === 0 && address === address.toLowerCase()) {
```

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +226,45 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                        updatedValues[1] &&
+                        updatedValues[0] !== updatedValues[1] &&
+                        toChecksumAddress(updatedValues[0]) === updatedValues[1]
+                    ) {
+                        setAddressChecksumInfo({
+                            text: 'Address converted to checksum format.',
+                        });
+                    } else {
+                        setAddressChecksumInfo(undefined);
+                    }
```
tomasklim: Why is this required? What if we just put 
this
```
                 setAddressChecksumInfo({
                            text: 'Address converted to checksum format.',
                        });
```
here

```
if (result.success && result.payload.history.total !== 0) {
                        if (!checkAddressCheckSum(address)) {
                            setValue(inputName, toChecksumAddress(address), {
                                shouldValidate: true,
                            });
                            setAddressChecksumInfo({
                                text: 'Address converted to checksum format.',
                            });
                        }
                    }
```
and reset `addressChecksumInfo` `onChange` of address if it is set

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -198,9 +226,45 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                    //they are not the same but the second one is checksum of the first one, in this case show the info that the address was automatically checksummed
+                    if (
+                        updatedValues &&
+                        updatedValues[0] &&
+                        updatedValues[1] &&
+                        updatedValues[0] !== updatedValues[1] &&
+                        toChecksumAddress(updatedValues[0]) === updatedValues[1]
+                    ) {
+                        setAddressChecksumInfo({
+                            text: 'Address converted to checksum format.',
```
tomasklim: This string is not used at all, why not calling this `addressChecksumInfo` -> `hasAddressBeenChecksummed` boolean?

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -47,10 +48,32 @@ interface AddressProps {
...
+            }
+
+            return newValues;
+        });
+
+        return newValues;
+    };
+
+    return { lastTwoValues, addValue };
+};
```
tomasklim: I think this hook is not needed, info in some comment below. Moreover, it should not be part of this file

tomasklim (changes requested):
I think commits are little bit chaotic. Could you please reset them and commit it again in less / better described commits?

I dont understand the `useLastTwoValues` logic but maybe I am missing something

packages/components/src/components/form/Input/Input.tsx
enjojoy: nice!!

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
enjojoy: I'm not sure I understand why this rename is needed

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
enjojoy: I realized this text is not used, it will be a boolean

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
enjojoy: Yep, just did this myself too :D

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
enjojoy: You are actually right. 

The problem was when the address was changed through toChecksumAddress(), it also triggered the validation logic again and the information bubble disappeared. Thats why I introduced that non very elegant logic. To be completely honest I didn't know that onChange is triggered only when the input is changed through the keyboard or paste, but not when it changed programatically with the toChecksum...() func.  This solves everything and that logic is not needed.

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
enjojoy: yes, removed

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
tomasklim: you are right, i wanted to comment it like this

```suggestion
                inputState={getInputState()}
```

tomasklim: ah there is conflict in names.

However, `inputState` is a function so it should have name which stresses that it is a function and not a variable

enjojoy: makes sense, I will think of another name

enjojoy: Renamed to getInputErrorState()

PR #12469, opened 2024-05-17
fix(mobile): wrong graph after device switch and then refetch

### LINKED ISSUE #11231: Graph: Last touched point on graph is persisted through refresh + balance is taken from different device when view-only enabled
**Describe the bug**
Last touched point on graph is persisted through refresh
When user has two devices in view-only and refreshes the dashboard, the balance is refreshed from the different device - see the video. 

**Info:**

-   Mobile Suite Version: 24.3.1
-   OS: Android, 13

**How to reproduce**
Balance persisted on graph:
1. Choose a point on the graph that is different from the portfolio balance
2. Pull down to refresh
3. Note the old selection price represented as the portfolio balance

Balance persisted from second device
1. Have two devices in view-only
2. refresh graph on dashboard on one
3. The balance from the other one appears

**Expected behaviour**
Current portfolio balance to be visible

**Screenshots**
https://github.com/trezor/trezor-suite/assets/216917/20dd4053-e130-4738-8291-5533bfc27715
[

https://github.com/trezor/trezor-suite/assets/84183817/2a07fcf5-6405-4e9b-8867-f94b5b3992cf

](url)

**Additional context**

Nodonisko:
## Description

`useImperativeHandle` is tricky, there was issue with memoization in one of then so I found way how to remove it and only pass ref through which solved that.

## Related Issue

Resolve #11231

### DIFF
--- a/suite-native/module-home/src/screens/HomeScreen/HomeScreen.tsx
+++ b/suite-native/module-home/src/screens/HomeScreen/HomeScreen.tsx
@@ -13,9 +13,10 @@ import { Box } from '@suite-native/atoms';
 
 import { BiometricsBottomSheet } from './components/BiometricsBottomSheet';
 import { EmptyHomeRenderer } from './components/EmptyHomeRenderer';
-import { PortfolioContent, PortfolioContentRef } from './components/PortfolioContent';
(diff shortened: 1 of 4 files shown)


juriczech (approved):
LGTM

PR #12472, opened 2024-05-20
chore(connect): fw-update: add error message when binary is too small

mroz22:
opening this instead of https://github.com/trezor/trezor-suite/pull/12440
the goal is to give a more descriptive error in case connect downloads something which is apparently not a valid firmware. In that case it doesn't make sense to send this invalid binary into the device. Also since we decided not to make fw update possible in feature branches, I believe it will be useful to find out whats going on when QA reports this "bug" in the future. 

image


mroz22:
not ready yet. which job does the build for dev servers? did I need to mark it "ready to review" to trigger it?

marekrjpolak (approved):

PR #12477, opened 2024-05-20
fix(suite-native): useEmptyPassphrase on undefined device instance

### LINKED ISSUE #12471: Passphrase: passphrase entry screen after PIN entry with feature flag disabled
**Describe the bug**
Passhprase screen is presented even with passphrase feature flag disabled after PIN entry.

**Info:**

-   Mobile Suite Version: develop-24.5.0/3e14473a21e65de47dcc6d1c1b3432e5ae5962b0
-   OS: android
- TT

**How to reproduce**
(see the video, trezor locked after finished discovery)

1. Wait for finished discovery
1. Do not enable view only
1. Do not enable passhprase feature flag
1. Lock the device (long press on screen for TT)
1. Exit Suite Lite
1. Open Suite Lite again
1. App ask for PIN entry on device
1. Enter the PIN on device
1. Passhprase screen is opened

**Expected behavior**
Discovery of standard wallet should start, no request for passhprase

**Screenshots**
https://github.com/trezor/trezor-suite/assets/3729633/782c7860-1799-490e-944d-5fd32a5726c2

juriczech:
## Description
After locking,  `features.passphrase_protection` returns `null`, which sets the `deviceInstance` to undefined, which results in settings `useEmptyPassphrase` to false, resulting in triggering the passphrase flow.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12471

### DIFF
--- a/suite-common/wallet-core/src/device/deviceReducer.ts
+++ b/suite-common/wallet-core/src/device/deviceReducer.ts
@@ -70,7 +70,7 @@ const merge = (device: AcquiredDevice, upcoming: Partial<AcquiredDevice>): Trezo
 
 const getShouldUseEmptyPassphrase = (device: Device, deviceInstance?: number): boolean => {
     if (!device.features) return false;
-    if (isNative() && typeof deviceInstance === 'number' && deviceInstance === 1) {
+    if (isNative() && (!deviceInstance || deviceInstance === 1)) {
         // On mobile, if device has instance === 1, we always want to use empty passphrase since we
         // connect & authorize standard wallet by default. Other instances will have `usePassphraseProtection` set same way as web/desktop app.
         return true;


suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -70,7 +70,10 @@ const merge = (device: AcquiredDevice, upcoming: Partial<AcquiredDevice>): Trezo
 
 const getShouldUseEmptyPassphrase = (device: Device, deviceInstance?: number): boolean => {
     if (!device.features) return false;
-    if (isNative() && typeof deviceInstance === 'number' && deviceInstance === 1) {
+    if (
+        isNative() &&
+        ((typeof deviceInstance === 'number' && deviceInstance === 1) || !deviceInstance)
+    ) {
```
matejkriz: Maybe it can be simplified like this?
```suggestion
    if (isNative() && (!deviceInstance || deviceInstance === 1)) {
```

and it could use some comment why `!deviceInstace` is added

juriczech: Good catch! Fixed and rebased

matejkriz (approved):
Please update the commit message and then feel free to merge.

Issue #12454, opened 2024-05-17
View only: discovery interruption


matejkriz:
**Describe the bug**
Interrupted discovery never continues.

**Info:**

-   Mobile Suite Version: 24.5.0 6516c718b79596480fd136a80440748af31247d8
-   OS: android

**How to reproduce**
- Disconnect device during discovery, discovery looks like still loading, but get stucked.
- When the device is plugged in again, it somewhat continues (discover more BTC accounts) but not properly (no more coins discovered)
- restart the app, discovery never runs again for the view-only device, no more coins discovered.

**Expected behavior**
~~Discovery should continue if device is connected. App has to indicate somehow incomplete discovery.~~
current mitiagion/workaround:
View only is not offered until discovery is completed. It is also not possible to enable it in settings while discovery is running. See the PR https://github.com/trezor/trezor-suite/pull/12516

STew790:
**QA OK**
Discovery is interrupted and doesn't continue. View-only is enabled only after discovery finishes.

24.5.1 319a98a

PR #12479, opened 2024-05-20
Fix view only persistance

matejkriz:
## Description

Fixes and improvements that I have come across while researching https://github.com/trezor/trezor-suite/issues/12454
I believe they make sense on their own. It's more deletions than additions 🎉


suite-native/graph/src/graphMiddleware.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { removeAccountGraphTimeframe } from './slice';
+
+export const graphMiddleware = createMiddleware((action, { next, dispatch }) => {
+    if (deviceActions.forgetDevice.match(action)) {
+        const deviceState = action.payload.state;
+        if (deviceState) {
+            dispatch(removeAccountGraphTimeframe({ deviceState }));
+        }
+    }
```
juriczech: Does this have to be middleware? You could just use this action in the reducer and change state

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -784,23 +784,16 @@ export const selectDeviceFirmwareVersion = memoize((state: DeviceRootState) => {
...
 
     return devices.filter(device => device.id !== PORTFOLIO_TRACKER_DEVICE_ID);
 });
 
 export const selectIsNoPhysicalDeviceConnected = (state: DeviceRootState) => {
-    const devices = selectDevices(state);
+    const devices = selectPhysicalDevices(state);
 
-    return devices.length === 1 && devices[0].id === PORTFOLIO_TRACKER_DEVICE_ID;
+    return !devices.filter(device => device.connected).length;
```
juriczech: For me Array.some would be better for readability, but it's just a nit

suite-native/graph/src/graphMiddleware.ts
matejkriz: great point! 6827addbe681dc197ce7761978dfd5b8249cf20b

suite-common/wallet-core/src/device/deviceReducer.ts
matejkriz: 8a8749319f375d0330bf6d462b88755f13445153

juriczech: Even better 👏

juriczech (approved):

PR #12484, opened 2024-05-20
refactor(suite): device instance always defined

juriczech:
## Description
This seems like some legacy code for working with multiple wallet instances. We might not need this anymore and can always use device instance 1 for newly connected device. 

QA will test this and if it doesn't break anything in passphrase flow, then we can use this simplification.


juriczech:
QA Task cc @trezor/qa   https://www.notion.so/satoshilabs/Test-passphrase-wallets-adding-removing-etc-1ec4757314d34ec5b522809c913859c4?pvs=4

PR #12493, opened 2024-05-21
test(desktop): adding nightly gha flow

ondracja:
- adding a scheduled nighlty flow that launches desktop playwright tests at midnight

### DIFF
--- a/.github/workflows/test-suite-desktop-nightly.yml
+++ b/.github/workflows/test-suite-desktop-nightly.yml
@@ -1,0 +1,72 @@
+name: "[Test] nightly suite-desktop e2e"
+# this action builds a desktop version of Suite and runs basic test suite for it at midnight
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-suite-desktop-nightly.yml
```diff
@@ -0,0 +1,72 @@
+name: "[Build/Test] suite-desktop nightly e2e tests"
```
vdovhanych: I would only name it `name: "[Test] nightly suite-desktop e2e`

vdovhanych (commented):
Otherwise ok :)

.github/workflows/test-suite-desktop-nightly.yml
ondracja: Changed.

vdovhanych (approved):

PR #12508, opened 2024-05-21
feat(suite-native): TextInput secure text does not prompt autofill

### LINKED ISSUE #12383: Passphrase secure input
Current input is using `secureEntry` prop from react native's input and that gets caught by password managers so we'll need to implement some kind of `SecureInput` component, that will reuse lots of stuff from our custom Input implementation from atoms, but the value and onChange logic will have to be changed so that value gets changed for `*` after user types something. But the stored value should actually be normal string. 

**Acceptance criteria**
- same styles as our current input, ideally as much reused code between these components as possible
- reusable component in `suite-native/atoms`
- no use of `secureEntry` and test that it doesn't get picked up by password managers
- works with react-hook-form

PeKne:
## Description
If the `TextInputField` atom receive `secureTextEntry` prop, it adjust another native `TextInput` props to disable interaction with autofill and password managers.
## Related Issue

Resolve #12383 

Tested on physical devices of both platforms.

### DIFF
--- a/suite-native/forms/src/fields/TextInputField.tsx
+++ b/suite-native/forms/src/fields/TextInputField.tsx
@@ -20,8 +20,30 @@ export interface FieldProps extends AllowedTextInputFieldProps, AllowedInputWrap
     valueTransformer?: (value: string) => string;
 }
 
+const SECURE_TEXT_PROPS: Partial<FieldProps> = {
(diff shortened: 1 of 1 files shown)


juriczech (approved):
LGTM, thanks 💪

PR #12510, opened 2024-05-21
fix(connect): waiting for handshake

martykan:
## Description

This PR fixes issues with handshakes that users encountered on Adalite. 
The main cause of the issue was incorrect handling of `CANCEL_POPUP_REQUEST`, which didn't actually close the popup. This resulted in a problem where the popup is open but the call (not requiring a popup) would be already finished. 
This should also fix issues relating to fallback in Firefox. 

Testing with: https://adalite.io/?trezor-connect-src=https://dev.suite.sldev.cz/connect/fix/connect-waiting-for-handshake-debug/
(but it needs upgrading `@trezor/connect-web` to properly test the fix)

### DIFF
--- a/packages/connect-iframe/src/index.ts
+++ b/packages/connect-iframe/src/index.ts
@@ -93,6 +93,7 @@ const handleMessage = async (event: MessageEvent<CoreRequestMessage>) => {
             }
             // reassign to current MessagePort
             [_popupMessagePort] = event.ports;
+            _popupMessagePort.onmessage = message => handleMessage(message);
         }
 
         if (!_core) {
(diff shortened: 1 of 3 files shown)


packages/connect-iframe/src/index.ts
```diff
@@ -309,6 +309,8 @@ const init = async (payload: IFrameInit['payload'], origin: string) => {
             _popupMessagePort.onmessage = message => handleMessage(message);
         } catch (error) {
             // tell the popup to use MessageChannel fallback communication (thru IFRAME.LOADED > POPUP.INIT)
+            // TODO!!!
+            console.warn('BroadcastChannel error', error);
```
mroz22: do we have analytics at this point? at least sometimes yes I guess. we could use it to report errors, at least numbers of failed inits. and maybe we should finally put sentry here.

martykan: This actually doesn't seem to be a point of failure. At first I thougt it's unhandled based on the comments, but connect-popup fallbacks to MessageChannel automatically

mroz22 (approved):

PR #12513, opened 2024-05-21
fix(suite-native): hide device switcher connect btn in discovery

### LINKED ISSUE #12512: Connect another device button in device switcher during discovery puts app into bad state temporarily
**Describe the bug**
You might get locked in connect device screen while other discovery is running.

1. When the discovery is running, click button in device switcher -> you get to connect device screen. 
2. If you then click close, you are prompted to insert pin on device even in case you dont have it set. 
3. Also you are unable to leave the screen until the discovery ends.`

https://github.com/trezor/trezor-suite/assets/2011829/8658a840-0ed6-4896-9aee-681841394576

Correct behaviour:
Until discovery ends, we dont show Connect another device button in device switcher

vytick:
Hide Connect button in Device switcher during discovery

Resolve #12512

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/bc09e827-1f42-47af-98ad-f71848f708ca

### DIFF
--- a/suite-native/device-manager/src/components/DeviceList.tsx
+++ b/suite-native/device-manager/src/components/DeviceList.tsx
@@ -17,7 +17,11 @@ import {
     StackToStackCompositeNavigationProps,
 } from '@suite-native/navigation';
 import { analytics, EventType } from '@suite-native/analytics';
-import { selectDevice, selectInstacelessUnselectedDevices } from '@suite-common/wallet-core';
+import {
+    selectDevice,
+    selectInstacelessUnselectedDevices,
+    selectIsDeviceDiscoveryActive,
+} from '@suite-common/wallet-core';
 import { Button, Box, TextDivider, VStack } from '@suite-native/atoms';
 import { TrezorDevice } from '@suite-common/suite-types';
 import { Translation } from '@suite-native/intl';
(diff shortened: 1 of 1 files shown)


suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -51,6 +60,27 @@ const buttonWrapperStyle = prepareNativeStyle(utils => ({
     paddingTop: utils.spacings.small,
 }));
 
+const ConnectButtonComponent = ({ isDividerVisible, onPress }: ConnectButtonComponentProps) => {
```
PeKne: Everything is a component :D
```suggestion
const ConnectButton = ({ isDividerVisible, onPress }: ConnectButtonProps) => {
```

PeKne (commented):
just a nit

suite-native/device-manager/src/components/DeviceList.tsx
vytick: Yeah I know, but it this felt better than just ConnectButton, since it might be more than just button. I couldnt come with better name than this one :)

vytick: but if it feels good for you, I'll change it :)

vytick: e6ffd60283

PeKne: Now it is only a button so it should be called like this IMO. When we add something to the component, then it should be renamed.

PeKne (approved):

PR #12514, opened 2024-05-21
WIP Ugly workaround for interrupted discovery

### LINKED ISSUE #12454: View only: discovery interruption

matejkriz:
## Description

View only device is remembered only when discovery is completed.

## Related Issue

Partially resolve https://github.com/trezor/trezor-suite/issues/12454

### DIFF
--- a/packages/suite/src/actions/firmware/__tests__/firmwareActions.test.ts
+++ b/packages/suite/src/actions/firmware/__tests__/firmwareActions.test.ts
@@ -1,5 +1,5 @@
 import { testMocks } from '@suite-common/test-utils';
-import { prepareFirmwareReducer, State as DeviceState } from '@suite-common/wallet-core';
+import { prepareFirmwareReducer, DeviceState } from '@suite-common/wallet-core';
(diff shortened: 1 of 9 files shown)


matejkriz:
closing in favour of https://github.com/trezor/trezor-suite/pull/12516

PR #12516, opened 2024-05-22
Prompt remember only after discovery ended

vytick:
Offer UI for View Only enabling

Temporary solution for [#12454](https://github.com/trezor/trezor-suite/issues/12454)

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/c470f616-5306-41e5-994c-f4122f7caefc

### DIFF
--- a/suite-native/module-settings/src/components/ViewOnly/WalletRow.tsx
+++ b/suite-native/module-settings/src/components/ViewOnly/WalletRow.tsx
@@ -1,8 +1,14 @@
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 
-import { Button, HStack, Text } from '@suite-native/atoms';
+import { Button, HStack, Loader, Text } from '@suite-native/atoms';
 import { Translation } from '@suite-native/intl';
-import { deviceActions, toggleRememberDevice } from '@suite-common/wallet-core';
+import {
+    DeviceRootState,
+    DiscoveryRootState,
+    deviceActions,
+    selectIsDiscoveryActiveByDeviceState,
+    toggleRememberDevice,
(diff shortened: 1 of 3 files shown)


suite-native/module-settings/src/components/ViewOnly/WalletRow.tsx
```diff
@@ -90,6 +97,8 @@ export const WalletRow = ({ device }: { device: TrezorDevice }) => {
         });
     };
 
+    const showToggleButton = selectedDevice?.state !== device.state || !isDiscoveryActive;
```
matejkriz: Why this part of the condition? `selectedDevice?.state !== device.state ||` Why not only `!isDiscoveryActive`?

vytick: So you can toggle View only for other devices (those not connected).

vytick: Otherwise it would remove the button for all wallets in case this wallet was in the discovery process

matejkriz: I see, what about 
```suggestion
    const showToggleButton = device.remember || !isDiscoveryActive;
```
?

I think it is simpler and even more correct - with passhprase you can have multiple devices with unfinished discovery.

vytick: This probably wont work every time:
- connect device, wait for discovery
- remember device and disconnect it
- connect another device, wait until discovery starts
- switch to portfolio manager before discovery ended
- go to view only settings
- see this:

matejkriz (approved):
It works nice 👍 Thanks for the idea ❤️

PR #12518, opened 2024-05-22
test(suite-native): Import an already imported XPUB

bosomt:
Automatized tests of sanity test Import an already imported XPUB

https://www.notion.so/satoshilabs/Import-an-already-imported-XPUB-c15497dd8e3340618c17029745f11aab?pvs=4

### DIFF
--- a/suite-native/app/e2e/pageObjects/accountImportActions.ts
+++ b/suite-native/app/e2e/pageObjects/accountImportActions.ts
@@ -23,6 +23,19 @@ class AccountImportActions {
         await scrollUntilVisible(by.text(accountName));
     }
 
+    async importSameAccount({
+        networkSymbol,
+        xpub,
+    }: {
+        networkSymbol: string;
+        xpub: string;
+        accountName: string;
+    }) {
+        await this.selectCoin({ networkSymbol });
+        await this.submitXpub({ xpub, isValid: true });
+        await detoxExpect(element(by.text('Coin already synced'))).toBeVisible();
+    }
+
     async selectCoin({ networkSymbol }: { networkSymbol: string }) {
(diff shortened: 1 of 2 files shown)


suite-native/app/e2e/pageObjects/accountImportActions.ts
```diff
@@ -23,6 +23,19 @@ class AccountImportActions {
         await scrollUntilVisible(by.text(accountName));
     }
 
+    async importSameAccount({
```
ondracja, line 26: this method is unnecessary, pls use `importAccount` for this test's purposes.

suite-native/app/e2e/pageObjects/accountImportActions.ts
```diff
@@ -23,6 +23,19 @@ class AccountImportActions {
...
+        networkSymbol,
+        xpub,
+    }: {
+        networkSymbol: string;
+        xpub: string;
+        accountName: string;
+    }) {
+        await this.selectCoin({ networkSymbol });
+        await this.submitXpub({ xpub, isValid: true });
+        await detoxExpect(element(by.text('Coin already synced'))).toBeVisible();
```
ondracja, line 36: While onging expects are fine inside a page action method (eg if you want to assert/wait for something during the test), this is the final assert of the test which should `always` be in the `IT` part of the test, not iside the logic

ondracja (changes requested):

Nodonisko (changes requested):
I think we have this scenarion already covered in `suite-native/app/e2e/tests/invalidAccountsImport.test.ts`

bosomt:
@Nodonisko you are right, i have not noticed it.
@ondracja closing as duplicity

PR #12520, opened 2024-05-22
Transport prefixed path

mroz22:
- types tightening 
- path is now prefixed with "type" 

image

image


mroz22:
waiting for #12574 (where I cherrypicked the first 2 commits from here).

we decided that node-bridge will keep track of path prefixes only locally.

Issue #11562, opened 2024-03-13
Discovery improvements


shenkys:
Enable to restart/continue discovery if it was not finished

shenkys:
Design done [here](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Bugfixes?node-id=259-2068&t=gZI3UoFGeSkVq5WN-11)

STew790:
**QA NOK**
Waiting for this fix https://github.com/trezor/trezor-suite/pull/13855

vytick:
#13855 merged

STew790:
**QA OK**
Issue was fixed, otherwise ok.

**Info:**
 24.8.1 f2f56aa

PR #12521, opened 2024-05-22
Store also first empty accounts after discovery

matejkriz:
## Description

By creating invisible empty discovered accounts, user can see it once he sent some TX there.

We don't repeat discovery on every Trezor connect yet and with view-only feature it can happen more often, that user will send some TX on new account and will never see it in Suite Lite until disable view-only and reconnect.

As opposed to Suite Desktop, in mobile there is a batched discovery so it can happen that there are 2 invisible empty accounts for some coins.

This is not a final solution at all, just a small improvement as a first step – discovery resume/continue needs to be implemented https://github.com/trezor/trezor-suite/issues/11562

## Acceptance criteria (notes for QA)
- If user receive some funds on first empty account, it should appear in Suite Lite automatically.
- If user receive some funds on the very first account of some coin, that wasn't visible in Suite Lite, it should appear automatically.
- User should be able to add new account/coin until reaching 10 accounts per type limit and only 1 empty account limit (it's not possible to add new account if the previous one is empty and already visible)
- It could also affect visibility of empty accounts on desktop, would be nice to double check adding new coin and account there too.

### DIFF
--- a/suite-common/wallet-core/src/accounts/accountsReducer.ts
+++ b/suite-common/wallet-core/src/accounts/accountsReducer.ts
@@ -167,6 +167,10 @@ export const selectDeviceAccounts = (state: AccountsRootState & DeviceRootState)
     return selectAccountsByDeviceState(state, device.state);
 };
 
+export const selectVisibleDeviceAccounts = memoize((state: AccountsRootState & DeviceRootState) =>
+    selectDeviceAccounts(state).filter(account => account.visible),
+);
+
 export const selectDeviceAccountsForNetworkSymbolAndAccountType = memoizeWithArgs(
     (
         state: AccountsRootState & DeviceRootState,
@@ -363,7 +367,7 @@ export const selectAccountsSymbols = memoize(
 );
 
 export const selectIsDeviceAccountless = (state: AccountsRootState & DeviceRootState) =>
-    pipe(selectDeviceAccounts(state), A.isEmpty);
+    pipe(selectVisibleDeviceAccounts(state), A.isEmpty);
 
 // Selected device has no accounts and no active discovery. It can be empty portfolio device.
 export const selectIsEmptyDevice = (
--- a/suite-common/wallet-core/src/accounts/accountsReducer.ts
+++ b/suite-common/wallet-core/src/accounts/accountsReducer.ts
@@ -243,6 +243,16 @@ export const selectDeviceAccountsByNetworkSymbol = memoizeWithArgs(
     },
 );
 
+export const selectVisibleDeviceAccountsByNetworkSymbol = memoizeWithArgs(
+    (state: AccountsRootState & DeviceRootState, networkSymbol: NetworkSymbol | null) =>
+        selectDeviceAccountsByNetworkSymbol(state, networkSymbol).filter(
+            account => account.visible,
+        ),
+    {
+        size: Object.keys(networks).length,
+    },
+);
+
 export const selectVisibleNonEmptyDeviceAccountsByNetworkSymbol = (
     state: AccountsRootState & DeviceRootState,
     networkSymbol: NetworkSymbol | null,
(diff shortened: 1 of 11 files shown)


juriczech (approved):

vytick (changes requested):
The empty accounts are visible in My Assets and Receive tabs.

In case we hide them there, make sure that Add account works. Because now add new account wont add anything, since there is an empty account for each asset.

How should it work if we add first empty account?  Possibly discover the next one if under limit of 10?

vytick:
One more thought. Does this also solve situation where user in Suite adds new account of network that was not previously discovered on Lite app and receives txn there? Or in that case he wont see it?

matejkriz:
> It is WIP, because the newly added account is not visible on Assets/Recieve screens until you perform search. I guess it's a memoization issue, need to investigate more.

It was really a memoization issue caused by usage of `selectEthereumAccountsTokensWithFiatRates` selector, probably related to https://github.com/dai-shi/proxy-memoize/issues/96. As a workaround, I removed the filter for tokens with fiat rates completely, as I believe it has to change anyway - we have to make tokens without fiat rates accessible somehow later.

I'll also prepare update of proxy-memoize, but it has caused us a lot of problems before, so I'll prepare it in a special PR for testing.

matejkriz:
> One more thought. Does this also solve situation where user in Suite adds new account of network that was not previously discovered on Lite app and receives txn there? Or in that case he wont see it?

It should solve this case too, needs to be tested.

matejkriz:
Back to draft, empty wallet is not handled yet.

matejkriz:
> Back to draft, empty wallet is not handled yet.

fixed in e99f18ebe85321f82c1e52e0c8daf7c805491127

suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
```diff
@@ -297,37 +275,39 @@ export const useAddCoinAccount = () => {
...
+        }
+
+        const account =
+            firstHiddenEmptyAccount ||
+            (await dispatch(
+                addAndDiscoverNetworkAccountThunk({
+                    network,
+                    deviceState: device.state,
+                }),
+            ).unwrap());
```
matejkriz, line 310: Temporary solution, I would expect removal of addAndDiscoverNetworkAccountThunk as part of https://github.com/trezor/trezor-suite/issues/11562

vytick: Would it make sense to discover another hidden empty account here even in case this one is made visible?

vytick: hmm, this probably does not make a sense unless the account is not empty.

But we can check regularly, that there exists at least one empty account of each coin-accountType pair and if not, run discovery until we find it.

matejkriz: That is basically how it wokrks on desktop, but let's leave it please for another PR, this one is already bigger than expected 🙈

vytick: for sure :)

vytick (approved):
Great job. I’ve tested it and everything works as expected. 👌

suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -243,6 +247,16 @@ export const selectDeviceAccountsByNetworkSymbol = memoizeWithArgs(
     },
 );
 
+export const selectVisibleDeviceAccountsByNetworkSymbol = memoizeWithArgs(
```
juriczech, line 250: It might be worth mentioning in comment why this selector needs to be memoized

matejkriz: Here it returns derived data, so I can't think of any useful comment to write about this.🤷

bosomt:
Probably not QA OK

1. sent tx 
2. i see two accounts in desktop app
3. dev Android app still shows only one account after 5 minutes of waiting 


image

telegram-cloud-photo-size-4-5798940160614842845-y

24.6.0 d9c3a31

bosomt:
New account appeared after 10 minutes 🎉 
Not sure if its ok or not :)

bosomt:
**QA OK**

Retested on Litecoin and remembered wallet and new account appeared in seconds after i sent transaction from desktop app

24.6.0 https://github.com/trezor/trezor-suite/commit/d9c3a311e901901161ce239f9cfeae9a1df6c487

PR #12523, opened 2024-05-22
feat(suite-native) Device switcher with fiat balance

### LINKED ISSUE #13099: Passphrase: Add fiat balances to wallet swichter
- [x] add fiat rates for each wallet
- [x] different visual for single wallet per device

[Figma](https://www.figma.com/design/mmDnYtctfjnreWhTjLfeQH/Passphrase?node-id=2146-52545&t=vhP0Zy4NrG8fYeOU-0)

vytick:
- [x] add fiat balance for each wallet
- [x] if there is a single wallet for device, show different UI

## Related Issue

Resolve #13099

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/4be57958-86ff-4b3e-992f-39342046fc02


suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -7,36 +7,52 @@ import { Icon } from '@suite-common/icons';
...
+            },
+        ],
+    }),
+);
 
 export const WalletItem = ({ deviceState, onPress, isSelectable = true }: WalletItemProps) => {
     const { applyStyle } = useNativeStyles();
     const device = useSelector((state: any) => selectDeviceByState(state, deviceState));
     const selectedDevice = useSelector(selectDevice);
+    const fiatBalance = useSelector((state: any) =>
```
matejkriz, line 53: For the fiat value on dashboard we use `getTotalFiatBalance` that includes tokens, right? So it can lead to mismatch. Is that intentional?

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -69,7 +85,14 @@ export const WalletItem = ({ deviceState, onPress, isSelectable = true }: Wallet
                     />
                     <Text variant="callout">{walletNameLabel}</Text>
                 </HStack>
-                {isSelectable && <Radio value="" onPress={onPress} isChecked={isSelected} />}
+                <HStack alignItems="center" spacing={12}>
+                    <FiatAmountFormatter
```
matejkriz, line 96: Higher numbers does not fit on the screen and Radio is then out of the box. We should handle that. Maybe by using ellipsis for wallet label if it's too long?

vytick: ca1e203d68

Nodonisko:
If we can wait with merging this for few days I will be addressing this in my PR that's already draft so we can use that utility functions

Nodonisko (approved):

suite-native/device-manager/src/components/WalletItem.tsx
matejkriz: nice 👍

PR #12527, opened 2024-05-23
feat(message-system): Updating message system warnings

MiroslavProchazka:
Adding a new condition to address unofficial firmware in Trezor devices.

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
-    "timestamp": "2024-05-15T00:00:00+00:00",
-    "sequence": 54,
+    "timestamp": "2024-05-23T00:00:00+00:00",
+    "sequence": 55,
     "actions": [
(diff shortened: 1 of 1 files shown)


suite-common/message-system/config/config.v1.json
```diff
@@ -700,6 +700,14 @@
...
                             "vendor": "*"
+                        },
+                        {
+                            "model": "T2T1",
+                            "firmware": ">2.6.0",
+                            "bootloader": "*",
+                            "variant": "*",
+                            "firmwareRevision": "88e1f8c7a5c7615723664c64b0a25adc0c409dee",
+                            "vendor": "*"
                         }
```
matejkriz, line 719: please duplicate the same condition with `"model": "T"`  for older Suites

matejkriz (approved):

matejkriz (changes requested):
we have to update sequence number

matejkriz (approved):

PR #12530, opened 2024-05-23
feat(suite): Add passphrase banner to dashboard

### LINKED ISSUE #12396: Add promo banner to dashboard (view-only)
**Ondra will design all banners first so we can unify it to common component.**

Screenshot 2024-05-15 at 10 30 25

jvaclavik:
## Description

image

## Related Issue

Resolve #12396

### DIFF
--- a/packages/suite/src/views/dashboard/components/DashboardPassphraseBanner.tsx
+++ b/packages/suite/src/views/dashboard/components/DashboardPassphraseBanner.tsx
@@ -1,0 +1,58 @@
+import { selectDevice } from '@suite-common/wallet-core';
+import { Warning, H3, Text, Button, IconButton, Row, Column } from '@trezor/components';
+import { goto } from 'src/actions/suite/routerActions';
+import { setFlag } from 'src/actions/suite/suiteActions';
+import { Translation } from 'src/components/suite';
+import { SettingsAnchor } from 'src/constants/suite/anchors';
+import { useDiscovery, useDispatch, useSelector } from 'src/hooks/suite';
+import { selectSuiteFlags } from 'src/reducers/suite/suiteReducer';
+
+export const DashboardPassphraseBanner = () => {
+    const dispatch = useDispatch();
+    const { isDashboardPassphraseBannerVisible } = useSelector(selectSuiteFlags);
+    const device = useSelector(selectDevice);
+    const { isDiscoveryRunning } = useDiscovery();
+
+    if (
+        isDashboardPassphraseBannerVisible === false ||
+        device?.useEmptyPassphrase === true ||
+        isDiscoveryRunning === true ||
+        isDiscoveryRunning === undefined
+    )
+        return null;
+
+    const handleClose = () => {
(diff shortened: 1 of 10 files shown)


packages/suite/src/views/dashboard/components/DashboardPassphraseBanner.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const device = useSelector(selectDevice);
+    const { isDiscoveryRunning } = useDiscovery();
+
+    if (
+        isDashboardPassphraseBannerVisible === false ||
+        device?.useEmptyPassphrase === true ||
+        isDiscoveryRunning === true ||
+        isDiscoveryRunning === undefined
+    )
+        return null;
```
peter-sanderson, line 29: it's multi-line `{}` :pray:

packages/suite/src/views/dashboard/components/DashboardPassphraseBanner.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export const DashboardPassphraseBanner = () => {
+    const dispatch = useDispatch();
+    const { isDashboardPassphraseBannerVisible } = useSelector(selectSuiteFlags);
+    const device = useSelector(selectDevice);
+    const { isDiscoveryRunning } = useDiscovery();
+
+    if (
+        isDashboardPassphraseBannerVisible === false ||
+        device?.useEmptyPassphrase === true ||
```
peter-sanderson, line 24: I think the logic here shall be different :thinking: maybe?

This way the banner will be visible in EVERY passphrase wallet regardles if user uses the default one as well. Is this intended?

I think, condition shall be: if ALL your devices are `useEmptyPassphrase===false`  the we shall show banner? Not sure if this is enough, maybe lets discuss this.

packages/suite/src/views/settings/SettingsDebug/ViewOnlySettings.tsx
```diff
@@ -47,6 +51,23 @@ export const ViewOnlySettings = () => {
                     />
                 </ActionColumn>
             </SectionItem>
+
+            <SectionItem>
+                <TextColumn title="Set isDashboardPassphraseBannerVisible" />
+                <ActionColumn>
+                    <Checkbox
+                        isChecked={isDashboardPassphraseBannerVisible}
+                        onClick={() => {
```
peter-sanderson, line 60: nit: separate to handleOnClicnt or something

peter-sanderson (changes requested):
Lets talk ;)

packages/suite/src/views/dashboard/components/DashboardPassphraseBanner.tsx
jvaclavik: I don't think that it should be only when all have `useEmptyPassphrase===false`

But we need @Hermez-cz for decide this

Anyway, we can merge it this way and update it later. I also added another condition so the banner will not be visible anymore when your default wallet is already passphrase.

peter-sanderson (approved):

PR #12533, opened 2024-05-23
fix(suite): card component bug

### LINKED ISSUE #12532: Card component bug
### Info
The `margin` and `maxWidth` properties don't work correctly in the _Card_ component. It works only if the `label` parameter is provided.

adderpositive:
### Info
The `margin` and `maxWidth` properties don't work correctly in the _Card_ component. It works only if the `label` parameter is provided.


jvaclavik (approved):
Nice catch!

peter-sanderson (approved):
:clap: :bow: :+1:

PR #12539, opened 2024-05-23
Fix and refactor selectors for empty device states

### LINKED ISSUE #12524: Switch to portfolio tracker sent user to empty dashboard instead of sync coin
1. Open the app
3. Connect device
4. Enable view-only
5. Disconnect device
6. Try to switch from disconnected device to portfolio tracker
7. You are redirected to "empty dashboard"

Video: 
https://github.com/trezor/trezor-suite/assets/84183817/5f081569-4731-4b4f-a45e-316a7fca7cc1

Version: 24.5.0, Last commit hash: 5e5d8a3

matejkriz:
## Description

The root cause of the bug was in `selectIsNoPhysicalDeviceConnected` that has been changed in https://github.com/trezor/trezor-suite/pull/12479/commits/3b0a4ca969faf970aef52f3b3c6d7a0041c8fa1b to do what the name says and fix bug with connecting trezor while there is already something remembered.

We discussed the names thoroughly with @vytick  and this is the best we could come up with.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12524

### DIFF
--- a/suite-native/device/src/hooks/useDetectDeviceError.tsx
+++ b/suite-native/device/src/hooks/useDetectDeviceError.tsx
@@ -153,7 +153,10 @@ export const useDetectDeviceError = () => {
     ]);
 
     useEffect(() => {
-        // Hide the error alert on disconnect of the device
+        // Hide the error alert when the device is disconnected.
+        // Device with error can't be view-only.
+        // Edge case: If user has connected two devices simultaneously,
+        // it will not hide the alert.
         if (isNoPhysicalDeviceConnected) {
             hideAlert();
         }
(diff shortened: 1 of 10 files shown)


suite-native/device/src/hooks/useDetectDeviceError.tsx
```diff
@@ -151,11 +149,4 @@ export const useDetectDeviceError = () => {
         showAlert,
         handleDisconnect,
     ]);
-
```
matejkriz, line 154: fixme

juriczech (approved):
LGTM. Mucho readable IMO. GG

PR #12540, opened 2024-05-23
Fix: connect Core initialization

szymonlesisz:
## Description

Something to discuss with @marekrjpolak and @mroz22 

Solving problems:
- multiple synchronous initializations at once
- throttle `DeviceList` events to be dispatched **after** Core successful initialization
- first step to `Core` grooming
- invalid condition when iframe is loaded (throws Manifest is missing)

Whats changed:
- created `coreManager` - manager for `Core` state (loaded/loading/init)
- moved `initCore` function to `Core` class
- added some unit tests for Core/coreManager
- moved `createTestTransport` mock to shared definitions

### DIFF
--- a/packages/connect-iframe/src/index.ts
+++ b/packages/connect-iframe/src/index.ts
@@ -20,7 +20,7 @@ import {
     CoreRequestMessage,
     CoreEventMessage,
 } from '@trezor/connect';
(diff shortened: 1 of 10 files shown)


marekrjpolak (approved):
I have some ideas for little improvements, but for now it's good as it is.

PR #12548, opened 2024-05-24
fix(suite-native): accordion icon handler

juriczech:
## Related Issue

Resolve


matejkriz (approved):
LGTM, haven't test it

PR #12550, opened 2024-05-24
transport bridge throttle listen

mroz22:
node-bridge bug: 

1. connect a dongle with 2 devices
2. suite sees 2 devices
3. disconnect the dongle
4. in maybe 30% of cases only 1 device disconnects

reason:
1. connect (bridge transport) learns about changes in connected devices via /listen subscriptions. 
2. there is an active subscription
3. USB layer reports disconnected device
4. subscription in 2 is resolved
5. The other device is reported as disconnected before a new subscription is created => client never learns about this change

proposed solution:
- throttle resolving of the subscriptions. the first reported event is handled right away but should there be another event before 500ms have passed, wait till 500ms have passed since the first event.


packages/transport-bridge/tests/http.test.ts
```diff
@@ -89,50 +111,195 @@ describe('http', () => {
...
-        describe('POST', () => {
-            it('/', async () => {
-                const url = trezordNode.server!.getRouteAddress('/')!;
-                const response = await bridgeApiCall({
-                    url,
-                    method: 'POST',
-                });
-                if (!response.success) {
-                    throw new Error(response.error);
+        describe.only('listen', () => {
```
marekrjpolak: only

marekrjpolak (approved):

packages/transport-bridge/tests/http.test.ts
mroz22: done

PR #12567, opened 2024-05-24
chore(suite-native): longer e2e timeout for onboarding test

vytick:


matejkriz (approved):
It can be a temporary workaround, but not a solution...

PeKne:
It is extraordinary that a timeout this high is needed. I will try to debug it this week. There has to be some smelly cause of this.

PR #12575, opened 2024-05-26
fix(staking): add base converting bignumber to string

dev-pvl:
## Description

Fixed an issue where BigNumber.js automatically formats the string as exponential. According to the documentation calling toString with a base argument, e.g. toString(10), will always return normal notation.

## Related Issue

Resolve


AdamSchinzel:
Related https://github.com/trezor/trezor-suite/pull/12576

AdamSchinzel:
Closed in favor of #12576

PR #12576, opened 2024-05-26
chore: BigNumber.toString(10) to prevent output as exponential

martykan:
## Description

This change ensures that all `toString` calls on `BigNumber` objects include base 10 as a parameter. By doing this, we prevent numbers from being displayed in exponential form.

- added unit and e2e test for ETH staking with balances >= 1000 ETH

Screenshot 2024-05-27 at 15 37 45 (1)

### DIFF
--- a/packages/utils/src/bigNumber.ts
+++ b/packages/utils/src/bigNumber.ts
@@ -1,0 +1,9 @@
+import BN from 'bignumber.js';
+
+export const BigNumber = BN.clone({
+    // Set to the maximum value to avoid scientific notation
+    EXPONENTIAL_AT: 1e9,
+});
+
+export type BigNumber = BN;
+export type BigNumberValue = BN.Value;
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -40,3 +40,4 @@ export * from './typedEventEmitter';
 export * from './urlToOnion';
 export * from './logs';
 export * from './logsManager';
+export * from './bigNumber';
--- a/packages/blockchain-link-utils/package.json
+++ b/packages/blockchain-link-utils/package.json
@@ -23,8 +23,7 @@
         "@mobily/ts-belt": "^3.13.1",
         "@solana/web3.js": "^1.91.6",
         "@trezor/env-utils": "workspace:*",
-        "@trezor/utils": "workspace:*",
-        "bignumber.js": "^9.1.2"
+        "@trezor/utils": "workspace:*"
     },
     "devDependencies": {
         "@trezor/blockchain-link-types": "workspace:*",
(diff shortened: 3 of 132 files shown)


tomasklim (changes requested):
I am worried that someone will forget to add `10` and the same issue will occur again later.

It is now already missing in e.g. `blockfrost.ts` 

Can we modify the BigNumber config to a really big number?

```
BigNumber.config({ EXPONENTIAL_AT: 1e+50 });
```

or modify `toString` prototype to have `base=10` by default?

```
const originalToString = BigNumber.prototype.toString;

BigNumber.prototype.toString = (base = 10) => originalToString.call(this, base);
```

martykan:
Yeah, it's true that this is not that maintainable. My idea to solve this was by making an Eslint plugin, since I already have a script which I used to make these changes. 

But based on your ideas, I think the best solution would probably be to create a new wrapper package which we would import instead of BigNumber, there would use `BigNumber.clone()` to create our instance and set the config on it. Or override the prototype.

packages/utils/src/bigNumber.ts
```diff
@@ -0,0 +1,9 @@
+import BN from 'bignumber.js';
+
+export const BigNumber = BN.clone({
```
Nodonisko, line 3: I am not sure if I like BN in utils package, because it was meant to be zero dependency. But probably fine.

martykan: I wanted to avoid making a new package just for this, so utils seemed like the best place for it.

mroz22: hmm, thats right, I sort of forgot about this zero-dependency rule when I advised @martykan to do it like this. Maybe an option would be to list this bignumber as a peer dep, but it is probably even more confusing

packages/utils/src/index.ts
```diff
@@ -40,3 +40,4 @@ export * from './typedEventEmitter';
 export * from './urlToOnion';
 export * from './logs';
 export * from './logsManager';
+export * from './bigNumber';
```
mroz22, line 43: maybe not exporting this via index could help prevent some accidental unwanted bundling of it? just thinking

martykan: Yeah, we can definitely do that, would that solve things?

mroz22: not sure, lets summon @Nodonisko

packages/utils/src/bigNumber.ts
martykan: Seems like it's possible to mark the peer dep as optional, I think that could work

packages/utils/src/index.ts
Nodonisko: Do not overcomplicate things, I thinked it though and it's probably fine to keep it as it is. We already have this dependency in nearly every package anyway and no-one uses out utils lib without rest of our libs.

MiroslavProchazka:
@Nodonisko @mroz22 can you please do the final review and approve/add comment? Thanks!

mroz22 (approved):

tomasklim (approved):

PR #12599, opened 2024-05-27
ci(github): deploy connect-explorer with relative paths

karliatto:
## Description
deploy connect-explorer with relative paths

### DIFF
--- a/.github/actions/release-connect/action.yml
+++ b/.github/actions/release-connect/action.yml
@@ -89,8 +89,8 @@ runs:
       shell: bash
       env:
         NODE_ENV: "production"
-        __TREZOR_CONNECT_SRC: https://${{ inputs.serverHostname }}/${{ inputs.serverPath }}/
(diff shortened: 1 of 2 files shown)


martykan (approved):

karliatto:
Some tests are failing and it is because connect-explorer still uses the connect src from build time https://github.com/trezor/trezor-suite/blob/develop/packages/connect-explorer/next.config.mjs#L39

I am not sure if we can change this, I see 2 solutions
* to use "TREZOR_CONNECT_SRC" in the tests that will use settings with the problem that you always have to use the settings for configuring for right env
* build it also for production instead of copying from staging.

:thinking:

martykan:
What if we made `TREZOR_CONNECT_SRC` work with relative paths too?

karliatto:
> What if we made `TREZOR_CONNECT_SRC` work with relative paths too?

Also, other option, we could modify it here https://github.com/trezor/trezor-suite/blob/139214b98bb9203e5c5cd4de310d168443329ea0/packages/connect-explorer/src/actions/trezorConnectActions.ts

vdovhanych (approved):

PR #12602, opened 2024-05-27
fix(suite): no staking dashboard when staking tx was 25 txs ago

tomasklim:
## Description

- If the staking transaction was 25 transactions ago, the user was unable to see the staking dashboard, which meant he could not unstake or claim his ETH. It was possible to fix this by loading transaction list, for example, by going to other pages in his tx list or by triggering search on the account. However, that is not smart


mroz22 (approved):
I am not very much into staking but it sort of makes sense

martykan (approved):

PR #12622, opened 2024-05-28
Fix/transport bridge udp

szymonlesisz:
## Description

- trezor-bridge use enumerate timeout if UdpApi is used (same as in UdpTransport)
- UdpTransport move enumeration timeout initialization from contructor
- trezor-bridge: handle argument of `node ./packages/transport-bridge/dist/bin.js udp`

### DIFF
--- a/packages/transport/src/transports/udp.ts
+++ b/packages/transport/src/transports/udp.ts
@@ -8,6 +8,7 @@ import { SessionsBackground } from '../sessions/background';
 
 export class UdpTransport extends AbstractApiTransport {
     public name = 'UdpTransport' as const;
+    private enumerateTimeout: ReturnType<typeof setTimeout> | undefined;
 
     constructor(params: AbstractTransportParams) {
         const { messages, logger, signal } = params;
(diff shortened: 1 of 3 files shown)


mroz22 (approved):

szymonlesisz:
converting to draft, issue with too often enumeration

packages/transport/src/transports/udp.ts
```diff
@@ -30,12 +31,27 @@ export class UdpTransport extends AbstractApiTransport {
...
+            if (!this.listening) return;
+
+            this.enumerateTimeout = setTimeout(() => {
                 this.enumerate().promise.finally(enumerateRecursive);
             }, 500);
         };
         enumerateRecursive();
+
+        return super.listen();
+    }
```
mroz22, line 47: just a side note - I will probably need to do something like this also for usb. unfortunatelly, onconnect event in windows does not return serialNumber which we rely on. 

so my rough plan for now is
- on api level, introduce abstract listen member. its task will be to make api start emitting events. 
- for udp, it would be this code here
- for usb, it would be probably listening to onconnect events and doing enumeration after they are detected.

mroz22 (approved):
as long as it helps with your usecase, feel free to merge it. 👍

PR #12627, opened 2024-05-29
POC Ledger derivation path for Solana and EVM

### LINKED ISSUE #8591: Ledger Live derivation path in Suite
**Problem**
Ledger users with a legacy Ethereum derivation path cannot access funds in Trezor Suite.

as an MVP without optimalization, discover accounts/addresess with this derivation path: `m/44'/coin_type'/0'/account`

Fws supporting this feature: **1.12.1**, **2.6.0**

### LINKED ISSUE #12234: Start discovering Ledger derivation path for SOL
With great sadness, I have to ask for implementation of Ledger Solana derivation path (`"m/44'/501'/i'`) during discovery phase.

tomasklim:
## Description

- discovery is done for Solana ledger derivation path and EVM Ledger Live and EVM Ledger Legacy derivation paths
- if you enable debug mode, you can add those accounts, otherwise they are visible only if they have some balance

TODO:
- it will report that not all coins have discovery done
  - that is caused by incorrect calculation of total accounts because first derivation path of ledger and trezor accounts in EVM is same so I skip it (by hack for now). If you add ledger account, it will have immediately number 2
- details in SOL, CARDANO and EVM accounts if more acc available?
- fw account type restriction
- improve the weird `accountUtils.ts`
- Sending to Ledger account / legacy account in send form.
- maybe some tag in account name?

## Related Issue

Resolves #8591

## Screenshots:
Screenshot 2024-06-02 at 22 43 20
Screenshot 2024-06-02 at 22 43 42

All accounts
Screenshot 2024-05-29 at 9 54 20

### DIFF
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AccountTypeSelect/AccountTypeDescription.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AccountTypeSelect/AccountTypeDescription.tsx
@@ -11,17 +11,18 @@ const Info = styled(Paragraph)`
 `;
 
 interface AccountTypeDescriptionProps {
-    bip43Path: Network['bip43Path'];
+    network: Network;
     hasMultipleAccountTypes: boolean;
 }
 
 export const AccountTypeDescription = ({
-    bip43Path,
+    network,
     hasMultipleAccountTypes,
 }: AccountTypeDescriptionProps) => {
     if (!hasMultipleAccountTypes) return null;
-    const accountTypeUrl = getAccountTypeUrl(bip43Path);
-    const accountTypeDesc = getAccountTypeDesc(bip43Path);
+
+    const accountTypeUrl = getAccountTypeUrl(network);
+    const accountTypeDesc = getAccountTypeDesc(network);
 
     return (
         <>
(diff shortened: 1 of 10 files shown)


MiroslavProchazka:
Please restrict the Ledger legacy discovery just for FW from v 2.6.0

STew790:
QA findings can be found here https://www.notion.so/satoshilabs/New-SOL-and-ETH-derivation-paths-532b6436b69445d29756ad2a24d5c8dd

MiroslavProchazka:
@STew790 , have you validated the derivation paths on Ledger? Do we know if we are discovering the right addresses?

tomasklim:
@Hermez-cz
1. do you agree to add account type here?
Screenshot 2024-06-03 at 20 52 38
2.
- for accounts which are not default, we should show at least some tag what account type it is
- should I add details tab to Ethereum, Solana and Cardano for non-default accounts or also for default ones?
- how about adding some banner that the account is not "official"?
- Ledger Ethereum n.1 account is same as Trezor n.1, thats why it is skipped, how to communicate it, to not confuse users?
Screenshot 2024-06-03 at 20 54 47

tomasklim:
https://www.figma.com/design/o6eb4wUMjSu5xMb34423vh/Tokens?node-id=4341-13003&t=ymVL5AzFo1ExeyGM-1 (edited)

PR #12636, opened 2024-05-29
@trezor/env-utils dependency fixes

martykan:
## Description

- Move `publicKey` from `@suite-common/wallet-constants` into `@trezor/env-utils` to avoid dependency on private package
- Make `expo-constants` optional peer dependency


packages/env-utils/package.json
```diff
@@ -28,16 +28,18 @@
...
     "peerDependencies": {
+        "expo-constants": "*",
         "expo-localization": "*",
         "react-native": "*",
         "tslib": "^2.6.2"
     },
     "peerDependenciesMeta": {
+        "expo-constants": {
+            "optional": true
+        },
```
mroz22, line 42: does this work for all the packages? npm, yarn, pnpm? just curious because this is the first time I see it.

martykan: I think it has pretty good support and it's already used here for `expo-localization` and `react-native`

mroz22 (approved):
good with me

Issue #12624, opened 2024-05-28
Push a message via banner to users with fw version <1.7.0 about updating their fw


Hannsek:
Banner copy: `Please update device firmware now: Your Trezor will not be recognized or usable in new versions of Trezor Suite.`

MiroslavProchazka:
Do we want to put there some CTA button?

Hannsek:
yes, that would be nice, didn't know we can tho. Let's put there a button with deeplink to settings. Simmilar to what we have in _backup failed_ banner.

MiroslavProchazka:
Resolved by https://github.com/trezor/trezor-suite/actions/runs/9287745344


https://github.com/trezor/trezor-suite/assets/152851523/043608dd-c489-4ae6-87a0-fbbcd0c1fc20

STew790:
**QA OK**
Tested with FW 1.5.0.

![image](https://github.com/trezor/trezor-suite/assets/48477147/fe2aab60-9530-4563-9eaf-c23a4235cca2)

**Info:**
 - Suite version: web 24.8.0 (f65530492c89854af6495e39b50e3780d9950565)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: Trezor T1B1 1.5.0 regular (revision 6b74139b4530a4687b4a317b8b08f4329704efc4)
 - Transport: BridgeTransport 2.0.33

PR #12637, opened 2024-05-29
Messaging banner for T1 devices with older than 1.7.0 FW

MiroslavProchazka:
Resolving https://github.com/trezor/trezor-suite/issues/12624

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
-    "timestamp": "2024-05-23T00:00:00+00:00",
-    "sequence": 55,
+    "timestamp": "2024-05-29T00:00:00+00:00",
+    "sequence": 56,
     "actions": [
         {
             "conditions": [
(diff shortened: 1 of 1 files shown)


MiroslavProchazka:
I had to adjust the file several times to pass the message-system validate config, so the section "Files changed" might not be the best place to check the PR. Please checkout to `Node-bridge-transition-messaging` verify the code.

suite-common/message-system/config/config.v1.json
```diff
@@ -795,44 +795,23 @@
...
-                        "web": ">=22.6.2"
+                        "web": ">=21.7.1"
                     },
                     "devices": [
                         {
                             "model": "1",
-                            "firmware": "<1.11.1",
+                            "firmware": "<1.7.0",
                             "bootloader": "*",
                             "variant": "regular",
```
matejkriz: variant should be `*` (affects also bitcoin-only FW...)

suite-common/message-system/config/config.v1.json
```diff
@@ -795,44 +795,23 @@
...
-                        {
-                            "zec": true
-                        }
-                    ],
                     "environment": {
-                        "desktop": ">=22.6.2",
+                        "desktop": ">=21.7.1",
                         "mobile": "!",
-                        "web": ">=22.6.2"
+                        "web": ">=21.7.1"
```
matejkriz: why from 21.7.1 Suite version? Is it the version where FW 1.7.0 has been introduced?

matejkriz:
Please add it in https://www.notion.so/satoshilabs/Secure-message-system-5ade512cbd38481f9d5f65f03dbfd9a8?pvs=4 🙏

suite-common/message-system/config/config.v1.json
MiroslavProchazka: It is one of the first Suite released, so I wanted to make sure that practically everyone will receive it

matejkriz: Then you can use `*` for any version or remove that environment condition completely.

matejkriz (approved):
Consider removing the environment condition, but otherwise it looks good to me and it can stay as it is. I'm not sure what would internal-link do in mobile app, so it's safer to keep it excluded, or test it.

PR #12641, opened 2024-05-29
fix(suite): refactor redundant select-code

peter-sanderson:


packages/suite/src/views/onboarding/steps/SelectBackupType.tsx
```diff
@@ -328,10 +328,121 @@ const DividerWrapper = styled.div`
...
+                        <Translation id="TR_ONBOARDING_BACKUP_TYPE_12_WORDS_DEFAULT_NOTE" />
+                    </Text>
+                )}
+            </OptionWithContent>
+
+            <OptionWithContent
+                onSelect={onSelect}
+                selected={selected}
+                value="24-words"
+                tags={<>{defaultType === '24-words' && <DefaultTag />}</>}
```
jvaclavik, line 396: Why do you need `Fragment`? Can we use just this?

```jsx
tags={defaultType === '24-words' ? <DefaultTag /> : null}
```

and btw `tags` evoke to me passing some array, which is not happening, but probably it's subjective

jvaclavik (approved):
much better!

packages/suite/src/views/onboarding/steps/SelectBackupType.tsx
peter-sanderson: lets solve this in follow up PR -> https://github.com/trezor/trezor-suite/pull/12652

PR #12642, opened 2024-05-29
feat(suite):add date pending transactions

eddymonk:
## Description
I added the date above pending transactions and grouped them by date as the normal one. So right now we have pending={date}

## Screenshots:
image

## Problems
Seems like there is an error while we store the transactions in the Redux store. Right now sometimes I see a bug where the tx.blockTime becomes the same for some of them. I don't know how my changes could affect that but I also know that we don't have it in production the same bug, not sure about it. Do you have some suggestions?

### DIFF
--- a/suite-common/wallet-utils/src/transactionUtils.ts
+++ b/suite-common/wallet-utils/src/transactionUtils.ts
@@ -101,7 +101,7 @@ export const groupTransactionsByDate = (
                 const key =
                     !isTxPending && item.blockTime && item.blockTime > 0
                         ? keyFormatter(new Date(item.blockTime * 1000))
-                        : 'pending';
+                        : `pending-${keyFormatter(new Date(item.blockTime ? item.blockTime * 1000 : new Date()))}`;
                 const prev = r[key] ?? [];
 
                 return {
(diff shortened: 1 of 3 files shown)


tomasklim:
@eddymonk could you please provide examples of the "problems"?

tomasklim:
Does it make sense to merge the sections like this?
Screenshot 2024-05-29 at 18 06 03

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -101,7 +101,7 @@ export const groupTransactionsByDate = (
                 const key =
                     !isTxPending && item.blockTime && item.blockTime > 0
                         ? keyFormatter(new Date(item.blockTime * 1000))
-                        : 'pending';
+                        : `pending-${keyFormatter(new Date(item.blockTime ? item.blockTime * 1000 : new Date()))}`;
```
tomasklim, line 104: not a strong opinion

How about changing this method that it will return 

```
{
    [key: string]: {pending: WalletAccountTransaction[]; success: WalletAccountTransaction[];}
}

or 

{
  pending: {
      [key: string]: WalletAccountTransaction[];
  };
  success: {
      [key: string]: WalletAccountTransaction[];
  };
}
```

instead of 
```
{
    [key: string]: WalletAccountTransaction[];
}
```
with the "hacky" pending date

tomasklim:
Please use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

Please link resolved issue

komret:
Thanks @tomasklim but let's not link the issue just yet, this is a recruitment assignment 😅 I will review it when I am back.

PR #12655, opened 2024-05-30
feat(suite-native): when device error show correct alert

### LINKED ISSUE #12603: Error while connecting Trezor appears as an uninitialized device
# Bug

From time to time there is some error while connecting Trezor (we can't reproduce it consistently) but app present that error as uninitialized device:
image

<details>
  <summary>Example of device state in that case</summary>

  ```js
{
selectedDevice: {
  available: false,
  buttonRequests: [],
  connected: true,
  error: "Initialize failed: unexpected error ",
  label: "Unreadable device",
  metadata: {},
  name: "Trezor",
  path: "21095CA051AA872F01416C1C",
  ts: 1716801174952,
  type: "unreadable",
  useEmptyPassphrase: true,
}
}
  ```
</details>

# Solution
We need to show general error instead and navigate user to reconnect the device. See screenshots below and follow [figma link](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Bugfixes?node-id=482-4228&t=IYUGy4yCtouxmmha-4).

- Underlying error should be reported to Sentry if analytics are allowed.
- `Reconect device` button should disconnect the unreadable device and navigate to `ConnectAndUnlockDevice` screen.
- `Contact support` button should lead to https://trezor.io/learn/c/trezor-suite-lite#open-chat

image
image

[Figma](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Bugfixes?node-id=482-4228&t=IYUGy4yCtouxmmha-4)

vytick:
If there is a device error, show general error alert instead of uninitialized device

Resolve #12603

### DIFF
--- a/suite-native/device/src/hooks/useDetectDeviceError.tsx
+++ b/suite-native/device/src/hooks/useDetectDeviceError.tsx
@@ -1,6 +1,9 @@
 import { useCallback, useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
+import * as Sentry from '@sentry/react-native';
+import { useNavigation } from '@react-navigation/native';
+
 import { analytics, EventType } from '@suite-native/analytics';
 import {
     acquireDevice,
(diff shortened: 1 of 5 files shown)


suite-native/device/src/hooks/useDetectDeviceError.tsx
```diff
@@ -37,6 +57,8 @@ export const useDetectDeviceError = () => {
 
     const isDeviceFirmwareSupported = useSelector(selectIsDeviceFirmwareSupported);
 
+    const hasDeviceError = selectedDevice?.error !== undefined;
```
juriczech: nit: `!!selectedDevice?.error`

juriczech: Also please create a selector and also use the returned value in useEffect deps

juriczech (approved):
LGTM

vytick:
@SocketSecurity ignore-all

suite-native/device/src/hooks/useDetectDeviceError.tsx
vytick: 10aacf0fcc

suite-native/device/src/hooks/useDetectDeviceError.tsx
```diff
@@ -152,6 +183,36 @@ export const useDetectDeviceError = () => {
         handleDisconnect,
     ]);
 
+    useEffect(() => {
+        if (hasDeviceError) {
```
matejkriz: I'm not sure if `hasDeviceError` is a right condition for such a general error, I don't know if there is always some error 🤔  I feel like `selectIsConnectedDeviceUninitialized` is not robust enough and should not return true if there is such an error (if device.features are undefined)

suite-native/device/src/hooks/useDetectDeviceError.tsx
```diff
@@ -1,6 +1,9 @@
 import { useCallback, useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
+import * as Sentry from '@sentry/react-native';
```
matejkriz, line 4: nitpick: I would prefer to import only used method 
```suggestion
import { captureException } from '@sentry/react-native';
```
but there is probably no real difference

suite-native/device/src/hooks/useDetectDeviceError.tsx
matejkriz: We should be aware of other cases, see for example https://github.com/trezor/trezor-suite/blob/develop/suite-common/suite-utils/src/device.ts#L31 So far "no-seed" alert is the default that catches all those cases, I believe that this new more generic error would be more appropriate.

matejkriz (approved):
I didn't manage to break it 👍 Please double check the failing e2e tests. And please prepare the followup issue to refactor error alerts from effects to something more sustainable  and easier to think about, this is very fragile and difficult to think through.

suite-native/device/src/hooks/useDetectDeviceError.tsx
matejkriz: after 4d008a03ad92d85eae17e1439135106a1a77facf it should work, but it should be refactored soon, selectIsConnectedDeviceUninitialized is confusing and it works only because unacquired device is handled with higher priority

vytick:
followup issue:
https://github.com/trezor/trezor-suite/issues/12678

PR #12677, opened 2024-05-31
feat(suite-native): passphrase on receive

### LINKED ISSUE #12132: Passphrase on receive
On remembered passphrase device, when reconnected, device won't ask for passphrase until receive address is requested. Handle passphrase request in that flow.

Image

https://www.figma.com/design/mmDnYtctfjnreWhTjLfeQH/Passphrase?node-id=2895-5652&t=WScMaNsEhAqbs34a-0

juriczech:
## Description
Handle passphrase request during receive flow on remembered passphrase device when reconnected.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12132

## Screenshots:
https://github.com/trezor/trezor-suite/assets/36101761/c04574bb-a72f-40f7-a2af-dfc8f56a21f5


suite-native/intl/src/en.ts
```diff
@@ -727,6 +728,10 @@ export const en = {
             subtitle: 'Go to your device and confirm you’d like to enable passphase.',
             cancelledError: 'Passphrase enabling cancelled.',
         },
+        passphraseFeatureUnlock: {
+            title: 'Enter to passphrase to continue',
```
matejkriz:
```suggestion
            title: 'Enter passphrase to continue',
```

juriczech: Good catch! Commited the changes

PeKne:
Screenshot 2024-06-03 at 15 13 51

The error message is not user friendly. But it will be fixed in a followup right? If yes, ignore this.

PeKne:
It is kind of weird that the adress is showed for about a second before hidden behind the passhprase input screen. Could we handle this somehow better?  


https://github.com/trezor/trezor-suite/assets/26143964/315f14cd-acb7-426e-9c50-c8132b12e064

juriczech:
@PeKne Yeah, I know. It's untranslated message from connect which is used in all error states in receive flow. I'd like to revisit this separetely together with add coin

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
```diff
@@ -0,0 +1,68 @@
+import { useCallback, useEffect } from 'react';
```
PeKne: What about simplifying this hook name to `usePassphrase` or something else shorter?

suite-native/receive/src/hooks/useAccountReceiveAddress.tsx
```diff
@@ -121,9 +143,9 @@ export const useAccountReceiveAddress = (accountKey: AccountKey) => {
         }
 
         return false;
-    }, [accountKey, freshAddress, dispatch, showToast, navigation, showAlert]);
+    };
 
-    const handleShowAddress = useCallback(async () => {
+    const handleShowAddress = async () => {
```
PeKne, line 148: Why was this `useCallback` removed? Since the function is returned from this hook, every run of it will cause a rerender in any component/hook that uses `handleShowAddress`

suite-native/receive/src/hooks/useAccountReceiveAddress.tsx
```diff
@@ -104,13 +119,20 @@ export const useAccountReceiveAddress = (accountKey: AccountKey) => {
...
+                    primaryButtonVariant: 'redBold',
+                    primaryButtonTitle: <Translation id="generic.buttons.tryAgain" />,
                     onPressPrimaryButton: () => {
-                        TrezorConnect.cancel();
-                        navigation.goBack();
-                        setIsUnverifiedAddressRevealed(false);
+                        handleCancel();
+                        // Handle show address calls verifyAddressOnDevice and here we want to trigger it again so there is cyclic dependency
+                        // between these calls.
+                        // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
PeKne, line 128: why this eslint-disable?

suite-native/receive/src/hooks/useAccountReceiveAddress.tsx
```diff
@@ -104,13 +119,20 @@ export const useAccountReceiveAddress = (accountKey: AccountKey) => {
...
+                        handleShowAddress();
```
PeKne, line 129: Do we really need to run `handleshowAdress`? What about running just `verifyAddressOnDevice directly?`

suite-native/module-passphrase/src/components/PassphraseScreenHeader.tsx
```diff
@@ -23,6 +23,11 @@ import {
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 import { Translation } from '@suite-native/intl';
 import { cancelPassphraseAndSelectStandardDeviceThunk } from '@suite-native/passphrase';
+import TrezorConnect from '@trezor/connect';
+
+type PassphraseScreenHeaderProps = {
+    shouldRemoveCurrentPassphraseWallet: boolean;
```
PeKne: I would add onClose to the prop name for better understandability. Something like:
```suggestion
    isWalletRemovedOnClose: boolean;
```

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
```diff
@@ -0,0 +1,68 @@
...
+    RootStackRoutes,
+    StackToStackCompositeNavigationProps,
+} from '@suite-native/navigation';
+import {
+    deviceActions,
+    selectDevice,
+    selectDeviceButtonRequestsCodes,
+} from '@suite-common/wallet-core';
+
+type NavigationProp = StackToStackCompositeNavigationProps<
```
PeKne: Will theme navigation props work for any screen? Now it works, because it is only used on a `ReceiveModalScreen` that is placed in RootStack. Not sure if it would work for example for navigation from `ReceiveStack` or `SendStack`. Could you verify that, please? If it is not gonna work, then the navigation handler should be passed to this hook as an argument.

suite-native/receive/src/hooks/useAccountReceiveAddress.tsx
juriczech: It's explained in the comment. Is it still unclear?

suite-native/receive/src/hooks/useAccountReceiveAddress.tsx
juriczech: I believe it's better to call `handleShowAddress` to really trigger try again. This way, the analytics event and appropriate setters are triggered on succesfully displaying the address.

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
juriczech: I think this is more explanatory and won't be confused with `useHandleDeviceRequestsPassphrase`

suite-native/receive/src/hooks/useAccountReceiveAddress.tsx
juriczech: Since it's not called from any useEffect that would trigger re-rendering of this hook or the function itself. The `useCallback` is redundant. The only time the hook is run is when the component gets mounted which always creates new functions anyways, because the previous ones wouldn't have been available anyway. 
I believe the `useCallback` section from [this article](https://kentcdodds.com/blog/usememo-and-usecallback) applies here.

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
juriczech: These are just typings, but the navigation works the same across the whole app. Wherever you call `navigation.goBack()`, it will do that as long as navigation is available and there is a screen to go back to. Since it's currently used primarily on the RootStack, it's typed that way and should there be need to extend this type, we can do `RootStack | AnyOtherStack`, but since all other screens, that are not directly in RootStack are subset of RootStack, the typing should be future proof. Good thought though

juriczech:
@PeKne I created an issue for the little bug and would rather solve it separetely. It's not considered release blocker now.
https://github.com/trezor/trezor-suite/issues/12706

suite-native/module-passphrase/src/components/PassphraseScreenHeader.tsx
juriczech: https://github.com/trezor/trezor-suite/pull/12677/commits/5dde9a246eb412fc29d8f12f75138271543c371b

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
juriczech: Btw. If `navigation.goBack` wouldn't work, it throws runtime error so then it's pretty clear what failed and can be debugged. But it's same as everywhere in our navigation logic

suite-native/module-passphrase/src/components/PassphraseScreenHeader.tsx
```diff
@@ -47,13 +52,20 @@ export const PassphraseScreenHeader = () => {
...
 
     const handleClose = () => {
-        dispatch(cancelPassphraseAndSelectStandardDeviceThunk());
-        navigation.navigate(RootStackRoutes.AppTabs, {
-            screen: AppTabsRoutes.HomeStack,
-            params: {
-                screen: HomeStackRoutes.Home,
-            },
-        });
+        if (isWalletRemovedOnClose) {
```
matejkriz: This is not handled on `PassphraseEnterOnTrezorScreen`, right? 

What about refactoring `cancelPassphraseAndSelectStandardDeviceThunk()` to not do both? (I mean extract TrezorConnect.cancel() from it). 

OR I believe it should be possible to distinguish those two states from the redux state - remove the wallet on cancel only if the wallet was never discovered or something like that...

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
```diff
@@ -0,0 +1,68 @@
...
+    PassphraseStackRoutes,
+    RootStackParamList
+>;
+
+// When device is in view only mode and is reconnected, passphrase is not requested after reconnecting
+// but when the app interacts with the device (e.g. requesting address / validating transaction, etc).
+// To make the passphrase working on this feature, just use this hook in the component that will interact with the device
+// and add the features button request code to the useEffect that redirects back to the flow
+// after successfull passphrase submission.
+export const useFeatureRequestsPassphrase = () => {
```
matejkriz: Why is this not handled globally?

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const dispatch = useDispatch();
+
+    const navigation = useNavigation<NavigationProp>();
+
+    const device = useSelector(selectDevice);
+    const buttonRequestCodes = useSelector(selectDeviceButtonRequestsCodes);
+
+    useEffect(() => {
+        if (
+            buttonRequestCodes.includes('ButtonRequest_Address') &&
```
matejkriz: This looks connected to receive address, can't work for adding new account or xpub verification and so on.

juriczech: Correct. It will have a condition for all "success" states of the actions that will require this functionality. In my follow up PR, it will be extracted to better named constant that will make it more clear.

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
juriczech: It's different treatment to other `ui-request_passphrase` events so the idea was to keep it scoped together with the functionality that requires this. I'd rather have two simple hooks with single purpose than one overcomplicated hook with conditions that would be hard to read and understand. If it was all handled within one `ui-request_passphrase` handler, I'm not sure what would the condition be to decide whether to navigate to passphrase form for wallet "creation" or this UI.

suite-native/module-passphrase/src/components/PassphraseScreenHeader.tsx
juriczech: Nice catch! Missed that completely 😮 

I originally did that as part of the thunk, but in the end all that remained was `TrezorConnect.cancel()` for this case so I moved it to the component instead. But I guess since you suggested it, it might actually be better implementation in the end. Let me know what you think and I might redo it

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
```diff
@@ -0,0 +1,68 @@
...
+            buttonRequestCodes.includes('ButtonRequest_Other') &&
+            navigation.canGoBack()
+        ) {
+            navigation.goBack();
```
matejkriz: I don't think this is a good place to navigate back. Closing passphrase screen can be hopefully handled inside passhprase screen? This seems to be fragile, not sure on what screen it happens and what is back. It might navigate back to PIN matrix screen.

matejkriz: TT, passphrase on device, I got stucked with this screen after passphrase confirmation on the device and device also ask me to verify address on this screen
image

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }, [buttonRequestCodes, device, dispatch, navigation]);
+
+    const handleFeatureRequestsPassphrase = useCallback(() => {
+        navigation.navigate(RootStackRoutes.PassphraseStack, {
+            screen: PassphraseStackRoutes.PassphraseFeatureUnlockScreen,
+        });
+    }, [navigation]);
+
+    useEffect(() => {
+        TrezorConnect.on('ui-request_passphrase', () => {
```
matejkriz: Please use constant instead of string literal
```suggestion
        TrezorConnect.on(UI.REQUEST_PASSPHRASE, () => {
```

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
matejkriz: demonstration of goBack on PIN screen

https://github.com/trezor/trezor-suite/assets/3729633/e8e93f82-50cc-4dfa-8098-96122418d4f6

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
matejkriz: I'm afraid this is more complicated than global handler and it is easy to miss handling passhprase while adding new feature... 

The condition can be if the selected device is already discovered or new device.

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
juriczech: https://github.com/trezor/trezor-suite/pull/12677/commits/b6d6469c0d616015be50aefe21b6a1c44d101ac0

suite-native/passphrase/src/useFeatureRequestsPassphrase.tsx
juriczech: After mutual discussion, I moved the UI.REQUEST_PASSPHRASE listener to one global hook that will handle these redirects.
https://github.com/trezor/trezor-suite/pull/12677/commits/b6d6469c0d616015be50aefe21b6a1c44d101ac0

PR #12681, opened 2024-05-31
Chore/reduce mobile app size

Nodonisko:
This should reduce mobile JS bundle size for about ~15%. 

1. I moved export transactions utils back to desktop because it's used only there and PDF lib that it is using is included in our mobile JS bundle. 
2. Removed unused date-fns locales.
3. Removed desktop EN transations from mobile app

## Related Issue

Resolve


matejkriz (approved):
nice 👏 
image

matejkriz (commented):
@Nodonisko not sure if `EXPO_UNSTABLE_ATLAS=true` is needed? If so, would be nice to put it in `.env.debug` file

Nodonisko:
I think it won't work if it's put into .env file because this file is loaded into Metro after it starts, but not sure.

PR #12691, opened 2024-06-03
fix(connect-web): check if chrome.tabs really available in webextension env

martykan:
## Description

This change adds checks to PopupManager to check if chrome.tabs API is really available. 
The reason is that when used in [offscreen](https://developer.chrome.com/docs/extensions/reference/api/offscreen) environment, which is used by Metamask for example, these APIs cannot be used. 
Metamask had to use a workaround with a patch that pretends the environment is web, which is not ideal (https://github.com/MetaMask/metamask-extension/blob/07f1eb412350d32231cadfda00d1347acb28749c/.yarn/patches/%40trezor-connect-web-npm-9.2.2-a4de8e45fc.patch)

### DIFF
--- a/packages/connect-web/src/popup/index.ts
+++ b/packages/connect-web/src/popup/index.ts
@@ -73,7 +73,7 @@ export class PopupManager extends EventEmitter {
         this.origin = getOrigin(settings.popupSrc);
         this.logger = logger;
 
-        if (this.settings.env === 'webextension') {
+        if (this.isWebExtensionWithTab()) {
             this.channel = new ServiceWorkerWindowChannel<CoreEventMessage>({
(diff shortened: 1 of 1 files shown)


mroz22 (approved):

mroz22:
maybe introducing a new env "offscreen" https://github.com/trezor/trezor-suite/blob/develop/packages/connect-web/src/connectSettings.ts#L5
would that make sense? it feels maybe a little cleaner than this

martykan:
I feel like that would complicate things, we would also need to check for `offscreen` everywhere where we handle `webextension`. Offscreen is mostly the same, it allows communication through `chrome.runtime` APIs, just not window/tab management, which is only done in PopupManager

PR #12694, opened 2024-06-03
feat(crypto-icons): download crypto icons from Coingecko

FreeWall:
## Description

- download crypto icons from Coingecko and upload it to aws s3
- icons resized into two sizes (24px, 48px), webp format
- naming: `[coingeckoId].webp` or `[platform]_[contract_address].webp`

## Related Issue

Resolve


tomasklim (approved):
LGTM. I dont understand the github script.

.github/workflows/release-crypto-icons.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Download crypto icons
+        run: |
+          yarn install
+          cd suite-common/icons
+          yarn download-crypto-icons
+
+      - name: Upload crypto icons
+        run: |
+          aws s3 sync suite-common/icons/files/cryptoIcons/ s3://data.trezor.io/cryptoIcons/ --cache-control no-cache
```
vdovhanych: I don't like the path we chose to deploy to. It's better to put it elsewhere. Also, we definitely don't want to set any cache for the images.

```suggestion
          aws s3 sync suite-common/icons/files/cryptoIcons s3://data.trezor.io/suite/icons/coins
```

.github/workflows/release-crypto-icons.yml
```diff
@@ -0,0 +1,44 @@
...
+      AWS_CLOUDFRONT_ID: E1ERY5K2OTKKI1
+      COINGECKO_API_KEY: ${{ secrets.COINGECKO_API_KEY }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          # role-to-assume: # TODO:
```
vdovhanych: this should be set to
```suggestion
          role-to-assume: arn:aws:iam::538326561891:role/gh_actions_suite_production_icons
```

.github/workflows/release-crypto-icons.yml
```diff
@@ -0,0 +1,44 @@
...
+    # Runs at midnight UTC every Monday
+    - cron: "0 0 * * *"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  release-crypto-icons:
+    environment: ${{ github.event.inputs.environment }}
```
vdovhanych: This will not work. You need to set the environment directly. There is no input for the schedule trigger or workflow dispatch. i will create environment `production-icons`

```suggestion
    environment: production-icons
```

.github/workflows/release-crypto-icons.yml
```diff
@@ -0,0 +1,44 @@
+name: "[Release] crypto icons"
```
vdovhanych: I don't like the name crypto icons lets name it `[Release] suite coin icons`

.github/workflows/release-crypto-icons.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Download crypto icons
+        run: |
+          yarn install
+          cd suite-common/icons
+          yarn download-crypto-icons
+
+      - name: Upload crypto icons
+        run: |
+          aws s3 sync suite-common/icons/files/cryptoIcons/ s3://data.trezor.io/cryptoIcons/ --cache-control no-cache
+          aws cloudfront create-invalidation --distribution-id ${AWS_CLOUDFRONT_ID} --paths '/cryptoIcons/*'
```
vdovhanych: this should be

```suggestion
          aws cloudfront create-invalidation --distribution-id ${AWS_CLOUDFRONT_ID} --paths '/suite/icons/coins/*'
```

vdovhanych (commented):
Please take a look.

How long will it run for? Will it be faster after the initial download and upload?

.github/workflows/release-crypto-icons.yml
FreeWall: fixed in https://github.com/trezor/trezor-suite/pull/12694/commits/8489b3c8d033ea65d640e0e8d9a361864c24ab55

.github/workflows/release-crypto-icons.yml
FreeWall: fixed in https://github.com/trezor/trezor-suite/pull/12694/commits/8489b3c8d033ea65d640e0e8d9a361864c24ab55

.github/workflows/release-crypto-icons.yml
FreeWall: fixed in https://github.com/trezor/trezor-suite/pull/12694/commits/8489b3c8d033ea65d640e0e8d9a361864c24ab55

.github/workflows/release-crypto-icons.yml
FreeWall: fixed in https://github.com/trezor/trezor-suite/pull/12694/commits/8489b3c8d033ea65d640e0e8d9a361864c24ab55

.github/workflows/release-crypto-icons.yml
FreeWall: fixed in https://github.com/trezor/trezor-suite/pull/12694/commits/8489b3c8d033ea65d640e0e8d9a361864c24ab55

FreeWall:
> Please take a look.
> 
> How long will it run for? Will it be faster after the initial download and upload?

I implemented max time limit of 60 minutes for this script, which should run every day.
We want to regularly update all icons and in order to check the coin image, request to coingecko needs to be made first, so it won't be faster next time.

vdovhanych:
Can we implement it so it does not download the image we already have for a certain time period? I feel like it's wasting API limits and CI resources if we download whatever number of images every day. Otherwise, it should be fine.

FreeWall:
What are the limits suitable for you? I think I can lower the "limit per minute" from 60 to 30, so the whole cycle will complete in ~10 days.

vdovhanych:
I dont really care if its 60 or 30 i was just asking if we download the same images every single time we run the script and how long does it need to run for the images to be uploaded, if we run it every day do we need to download all of the imagease again or just any new ones? 

i just dont want to download stuff if we already have it...

FreeWall:
Icons of specific coins/tokens may change during time, so we need to download the existing ones again as well.
About the upload, I'm not really sure how long it will take, it's about 8000 files (1 KB each) per github action run. If the upload is bottleneck, I could implement some file content comparison of existing icons, but it may increase CI resources too.

vdovhanych (approved):
Let's run it once and see what happens. Please remove the cron trigger from the job and only run it manually for now, but I believe this is a sub-par solution to this problem, and we need to figure out a better solution for sure!

FreeWall:
ok, cron trigger removed

ffrediani:
Have all the relevant token icons downloaded already ? I can't find them, only the network/coin ones.

PR #12698, opened 2024-06-03
fix(suite): fix condition for wipe warning

komret:
Fix condition in https://github.com/trezor/trezor-suite/pull/12697 😅


peter-sanderson (approved):
image

PR #12702, opened 2024-06-03
ci(github): improve release connect npm process

karliatto:
## Description

* identify the NPM packages that need to be released by checking the version numbers locally and in NPM, since locally they were previously changed by bump-connect-version workflow.
* It also identifies from the versions the `deploymentType`
* And with that information it performs some sanity checks where it checks that all the packages are going to be released with same `deploymentType` not mixed.
* It also checks that connect version is the same as in the release branch name `release/connect/<version>`.


ci/scripts/determine-deployment-type.ts
```diff
@@ -0,0 +1,12 @@
+const semver = require('semver');
+
+const version = process.argv[2];
+
+let deploymentType;
+if (semver.prerelease(version)) {
+    deploymentType = 'canary';
+} else {
+    deploymentType = 'stable';
+}
```
mroz22: maybe some sort of validation wouldn't hurt here? I  mean, in case process.argv[2] is whatever bogus (foo, undefined). you get deployment type 'stable'. 

maybe semver.valid could do the thing

.github/workflows/release-connect-npm.yml
```diff
@@ -1,27 +1,84 @@
 name: "[Release] Connect NPM"
 on:
   workflow_dispatch:
-    inputs:
-      packages:
-        description: 'Array string with names of the packages to deploy. (example: ["blockchain-link-utils","blockchain-link-types","analytics"])'
```
mroz22, line 6: so we no longer need to list them manually?

mroz22 (approved):
looks ok.

ci/scripts/determine-deployment-type.ts
```diff
@@ -0,0 +1,14 @@
...
+const version = process.argv[2];
+
+let deploymentType;
+if (semver.prerelease(version)) {
+    deploymentType = 'canary';
+} else if (semver.minor(version) || semver.major(version)) {
+    deploymentType = 'stable';
+} else {
+    throw new Error(`Invalid version: ${version}`);
+}
```
mroz22, line 12: 👍  safer

vdovhanych:
Can you add a description of the release process now? If we go ahead and merge this how will the process look if we go from develop up until we release it?

karliatto:
> Can you add a description of the release process now? If we go ahead and merge this how will the process look if we go from develop up until we release it?

I have documented in notion. Is that enough for you?

PR #12716, opened 2024-06-04
Ci: migrate utils tests to GitHub

karliatto:
## Description

Migrate utils test from GitLab to GitHub with new name `misc` so it is not confused with utils package.


.github/workflows/test-utils.yml
```diff
@@ -0,0 +1,101 @@
+name: "[test] utils"
```
mroz22: maybe `[test] misc` so that it won't get confused with utils package?

karliatto: Yes, makes sense. 7074867e645225c67728114e1b707ae9bb21f992

mroz22 (approved):

.github/workflows/test-misc.yml
```diff
@@ -0,0 +1,101 @@
...
+      - run: ./packages/connect/e2e/test-yarn-install.sh
+
+  test-unit:
+    runs-on: ubuntu-latest
+    if: github.repository == 'trezor/trezor-suite'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: true
```
mroz22: does test unit require submodules?

karliatto: You are right, they are not needed. Thanks! :+1:

Issue #12618, opened 2024-05-28
Passphrase UX revamp


juriczech:
Most critical UX issues on passphrase flow that should be addressed before release and would be confusing to the user and make the whole flow less fluent if not implemented. For example bottom sheets displayed in wrong screens, wrong redirects, useless steps in flow, etc.

UX Issues linked [here](https://www.figma.com/design/mmDnYtctfjnreWhTjLfeQH/Passphrase?node-id=2896-19290&t=R6WoQx6TcqxRCVpT-11)

- [x] Description to be done after design check by @shenkys 
- [x] Cancel error 🔴
- [x] Confirm empty scrolling 🔴 (fixed https://github.com/trezor/trezor-suite/pull/12737)
- [x] Missing "reveal" eye icon 🔴 (fixed by https://github.com/trezor/trezor-suite/pull/12727)

juriczech:
@trezor/qa Will be tested by @shenkys

PR #12727, opened 2024-06-04
feat(suite-native): toggleable secute text passphrase input

PeKne:
## Description
- SecureTextInputField component created
  - this componet behaves as a TextIputField but allows to hide/show text content with a toggle
## Related Issue

Related to #12618

## Screenshots:

https://github.com/trezor/trezor-suite/assets/26143964/b87994db-db6c-47a7-b62b-7d4faaa5e585


PeKne:
Commit [3661121](https://github.com/trezor/trezor-suite/pull/12727/commits/3661121c8010f79b34a6eda29aa6278b9a634db6) changes the `eyeSlash` from solid to outline:

https://github.com/trezor/trezor-suite/assets/26143964/4777e078-feb0-4329-92f1-efdf59bb4b4b

suite-native/forms/src/fields/SecureTextInputField.tsx
```diff
@@ -0,0 +1,48 @@
...
+type ToggleSecureTextIconProps = {
+    onPress: () => void;
+    isTextHidden: boolean;
+};
+
+const ToggleSecureTextIcon = ({ onPress, isTextHidden }: ToggleSecureTextIconProps) => {
+    const iconName: IconName = isTextHidden ? 'eye' : 'eyeSlash';
+
+    return (
+        <Pressable onPress={onPress}>
```
juriczech, line 19: I would probably prefer using TouchableOpacity, but it's inside an input, so it's debatable. Feel free to leave it as it is, but some visual feedback onPress would be nicer, WDYT?

juriczech (approved):
LGTM, just wondering about the visual feedback on the icon, however that's not too important right now, I think.

Edit: I tested this on iOS (also tested max length), not android, but I see you tested Android so should be fine to merge.

suite-native/forms/src/fields/SecureTextInputField.tsx
PeKne: I originally used `TouchableOpacity` but changed it to `Pressable` before pushing it. It felt weird that the eye button was animated, but the text changing to dots and back was not. So I decided to remove the animation to unify it. It feels somehow more natural, but it is a matter of taste for sure. Let's leave it as it is. I don't mind changing it later if Design or Product will complain.

PR #12682, opened 2024-05-31
chore: extend logging for transport-bridge

mroz22:
- add bambilion of logs on transport, transport-bridge level
- few other minor changes related to logging
image

PR #12728, opened 2024-06-04
fix(transport-bridge): windows

### LINKED ISSUE #12560: Device not detected after reboot to bootloader
**Describe the bug**

Device not detected after reboot to bootloader, im not happen to reproduce it with 100% rate.
It happens before install of firmware, after reboot /after firmware installation/
Reconnect of device or reload or restart of app does not help /i will create follow up issue/

I was only able to reproduce it on Windows computer and not on macARM computer.
Same behavior when connected directly or via hub.
Same behavior when connected external mouse and keyboard.

Info:
 - Suite version: desktop 24.6.0 (b6ccc949b30d2c22cfc9a3e59bfe0459a32dc693)
 - Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.6.0 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: Win32
 - Screen: 1920x1080
 - Device: Trezor T2B1 2.7.0 regular (revision null)

**Screenshots:**
telegram-cloud-photo-size-4-5947173913089916918-y

### LINKED ISSUE #12562: Acquire error initalise failed: unexpected error when empty device
**Describe the bug**

- i am getting `Acquire error initalise failed: unexpected error` error lots of times when trying to acquire wiped device mostly
- Including 400 bad request errors/responses from node bridge
- only quit and start of whole app/bridge helps /reconnect of device or reload of app does not help/
- its same behavior if you run new app as source of bridge and you use production web wallet
- i will try reproduce it with new bridge that will save log file
- it happens on **mac** and on **win** machine

**Steps to reproduce:**
1 Wipe device and reconnect it
1 Update device that includes wiping of device
1 ???

**Info:**
Suite version: desktop 24.6.0 (https://github.com/trezor/trezor-suite/commit/b6ccc949b30d2c22cfc9a3e59bfe0459a32dc693)
    Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.6.0 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
    OS: Win32
    Screen: 1920x1080
    Device: Trezor T2B1 2.7.0 regular (revision null)
    Transport: BridgeTransport 3.0.0

**Screenshots:**
telegram-cloud-photo-size-4-5947173913089916915-x
telegram-cloud-photo-size-4-5947173913089916916-y

mroz22:
problem: 

In windows, node usb transport does not return serialNumber in connect event. Or to be more specific - it does return the serialNumber but only until device was worked with. After connecting device, doing for example account discovery - when I reconnect the device again, it does not have serialNumber anymore. 

This is a big problem since serialNumber is used to distinguish between multiple simultaneously connected devices on transport api layer. Practically this issue can cause api layer to delete a different device from the one that was physically disconnected. 

Commits: 
- [chore(transport): add more logs to usb](https://github.com/trezor/trezor-suite/pull/12728/commits/90594776062dc422e9c3b1ed7bf470b5026f1527)
  - just adding some logs, could be useful.
  - also it is possible to get libusb logs, `usb` [package supports this](https://github.com/node-usb/node-usb?tab=readme-ov-file#usbsetdebuglevellevel--int). these are the logs that are available via the "download detailed log" button in the old bridge status page. The question is, how do I redirect these logs into a file?  
- [chore(transport): introduce listen in abstractApi](https://github.com/trezor/trezor-suite/pull/12728/commits/3289f1da4b74dc0714dcf701ac431dbae6426b91)
  - In my opinion this is a little cleaner approach (see the removed code from core.ts in transport-bridge). 
- [chore(transport): createDevices only if there are some nonHid devices…](https://github.com/trezor/trezor-suite/pull/12728/commits/5b5b94e40abce5e672d4f341984fbcea8f79418c) 
  - a small optimization, it is not necessary to call createDevices function in case the array is empty - there is no need to emit transport-interface-change event in case there is in fact no change that could be consumed by higher layers.
- [chore(transport): log connected hid device](https://github.com/trezor/trezor-suite/pull/12728/commits/d48a27e83118e56e96e30ad70092c576be29fe70)
  - at the moment, node-bridge can't handle old hid devices. this is a known limitation and I belive product is fine with this.  However I also belive that we should propagate this information upwards to be able to show "unreadable device screen" with appropriate troubleshooting tips. Lets address this in a followup. 
  - another option would be to implement hid api as well.[ should be possible using node-hid package. ](https://www.npmjs.com/package/node-hid)
- [chore(transport): change public method to private](https://github.com/trezor/trezor-suite/pull/12728/commits/eeab7cc948fffaaa8f8dfcc4097a51761f459c17)
  - tiny change. 
- [fix(transport): make device.serialNumber available also for windows](https://github.com/trezor/trezor-suite/pull/12728/commits/3a3751133d60fe89c02246acddcd1acdb0cca56d)
  - this commit ensures that serialNumber is always awailable even for Windows.
  - unasnwered question: should we run this code on all platforms (Linux, Win, Mac) or limit it only for Win? What is better? Use the same code everywhere which is easier to test and maybe has less maintanance overhead but it comes at the cost of doing unnecessary work on some of the platforms?

Followups: 
- [ ] propagate unreadable hid device upwards.

Issues and PRs:
- might fix #12562
- might fix #12560
- based on #12682

### DIFF
--- a/packages/transport-bridge/src/core.ts
+++ b/packages/transport-bridge/src/core.ts
@@ -36,6 +36,9 @@ export const createApi = (apiArg: 'usb' | 'udp' | AbstractApi, logger?: Log) => 
                       usbInterface: new WebUSB({
                           allowAllDevices: true, // return all devices, not only authorized
                       }),
+
+                      // todo: possibly only for windows
+                      forceReadSerialOnConnect: true,
                   });
     } else {
         api = apiArg;
--- a/packages/transport-bridge/src/core.ts
+++ b/packages/transport-bridge/src/core.ts
@@ -41,6 +41,8 @@ export const createApi = (apiArg: 'usb' | 'udp' | AbstractApi, logger?: Log) => 
         api = apiArg;
     }
 
+    api.listen();
+
     // whenever low-level api reports changes to descriptors, report them to sessions module
     api.on('transport-interface-change', descriptors => {
         logger?.debug(`core: transport-interface-change ${JSON.stringify(descriptors)}`);
@@ -116,22 +118,6 @@ export const createApi = (apiArg: 'usb' | 'udp' | AbstractApi, logger?: Log) => 
 
         return enumerateDoneResponse;
     };
-
-    let listening = true;
-    let enumerateTimeout: ReturnType<typeof setTimeout> | undefined;
-
-    if (api instanceof UdpApi) {
-        // same as UdpTransport listen, set timeout to ping udp devices
-        const enumerateRecursive = () => {
-            if (!listening) return;
-
-            enumerateTimeout = setTimeout(() => {
-                enumerate().finally(enumerateRecursive);
-            }, 500);
-        };
-
-        enumerateRecursive();
-    }
 
     const acquire = async (
         acquireInput: Omit<AcquireInput, 'previous'> & { previous: Session | 'null' },
@@ -241,11 +227,6 @@ export const createApi = (apiArg: 'usb' | 'udp' | AbstractApi, logger?: Log) => 
     };
 
     const dispose = () => {
-        listening = false;
-        if (enumerateTimeout) {
-            clearTimeout(enumerateTimeout);
-            enumerateTimeout = undefined;
-        }
         api.dispose();
     };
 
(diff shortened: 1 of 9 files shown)


marekrjpolak (approved):
Apart from one verbally shared nitpick, LGTM.

mroz22:
force pushed to satisfy commits atomicity

PR #12739, opened 2024-06-05
feat(suite): for old FW the single-share is not available

### LINKED ISSUE #12701: Onboarding - restrict using of SLIP 39 extendable seed only for devices with newer FW
Please add a condition that only devices with firmware version 2.7.1 (TT/TS3) or later can use the new Slip 39 extendable seed wallet backup.

All devices with firmware version 2.7.0 and older will not be able to do this (fields will be deactivated).  Please add a tooltip on hover with the following copy:

“Please update your device firmware to enable the new Single-share / Multi-share Backup feature.”

peter-sanderson:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12701

## Screenshots:

image

### DIFF
--- a/packages/suite/src/views/onboarding/steps/SelectBackupType/OptionWithContent.tsx
+++ b/packages/suite/src/views/onboarding/steps/SelectBackupType/OptionWithContent.tsx
@@ -1,4 +1,4 @@
-import { Radio, Row, variables, Text, Icon, useElevation } from '@trezor/components';
+import { Radio, Row, variables, Text, Icon, useElevation, Tooltip } from '@trezor/components';
 import { BackupType } from '../../../../reducers/onboarding/onboardingReducer';
 import { ReactNode, forwardRef } from 'react';
 import { Translation } from 'src/components/suite';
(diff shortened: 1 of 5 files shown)


packages/suite/src/views/onboarding/steps/SelectBackupType/OptionWithContent.tsx
```diff
@@ -160,4 +183,6 @@ export const OptionWithContent = ({
             </OptionText>
         </Option>
     );
+
+    return tooltip !== undefined ? <Tooltip content={tooltip}>{inner}</Tooltip> : inner;
```
jvaclavik: I would add `hasArrow`

jvaclavik (approved):

packages/suite/src/views/onboarding/steps/SelectBackupType/OptionWithContent.tsx
peter-sanderson: Maybe, shall we consider this to be default?

jvaclavik: 💯

PR #12750, opened 2024-06-05
fix(react-native-usb): implement opened WebUSBDevice state and closing device when leaving the app

### LINKED ISSUE #11896: Handle connection to Trezor Suite web and connected Trezor

___

___

image

**Original design**
image

image
___

matejkriz:
## Description

One more way to fix https://github.com/trezor/trezor-suite/issues/11896 🙉  This version keep the current permission flow - when you connect Trezor, Android ask you to choose if you want to open the app.

This PR unblock Trezor when Suite Lite goes into the background.

Defining `opened` state properly, Suite Lite is no longer blocking Trezor from usage in other app so it's possible to onboard via Suite Web even if Suite Lite is installed.

New UI for uninitialized device has to be implemented and I would like to use some changes from https://github.com/trezor/trezor-suite/pull/12177 to make permission handling even better in followup PR.

## Related Issue

Part of https://github.com/trezor/trezor-suite/issues/11896

## Screenshots

https://github.com/trezor/trezor-suite/assets/3729633/e99b4c3e-372c-490a-ae02-d0715bbeda94

### DIFF
--- a/packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
+++ b/packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
@@ -20,6 +20,8 @@ import expo.modules.kotlin.exception.CodedException
 import expo.modules.core.errors.ModuleDestroyedException
 import java.nio.ByteBuffer
 
+// Convert Android USBDevice to JS compatible WebUSBDevice
+typealias WebUSBDevice = Map<String, Any?>
 
 const val ON_DEVICE_CONNECT_EVENT_NAME = "onDeviceConnect"
 const val ON_DEVICE_DISCONNECT_EVENT_NAME = "onDeviceDisconnect"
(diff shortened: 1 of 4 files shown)


packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
```diff
@@ -362,6 +377,6 @@ fun getWebUSBDevice(device: UsbDevice): WebUSBDevice {
             "serialNumber" to device.serialNumber,
             //"configuration" to device.configuration,
             //"configurations" to device.configurations,
-            "opened" to false,
+            "opened" to opened,
```
Nodonisko: Why not use `hasOpenedConnection(device.deviceName)` here instead of passing it as param?

Nodonisko: I see it's not method but standalone function, then we should make it method

matejkriz: I like standalone functions more :-) but of course I can make it a method here, let's do it.

Nodonisko: I think it doesn't make sense to pass information manually which could only lead to bugs.

matejkriz: 7e3b6dec61d6956cee803793fd15e4d6e26b32de

Nodonisko (approved):
LGTM 🎉

PR #12752, opened 2024-06-06
ci(github): fix bump-version-connect script to find out packages to release and produce right changelog

karliatto:
## Description

This PR fixes 2 things:

1. The script `connect-bump-versions.ts` was producing wrong results in CI because it was building the packages to compare with the ones in NPM without building the libs so the result were always different and it wrongly wanted always to release all the packages.
2. The script `connect-bump-versions.ts`  did not have all the git commit history so when producing the CHANGELOG for the packages to release it was only able to find the last commit, producing wrong CHANGELOGs

### DIFF
--- a/.github/workflows/release-connect-bump-versions.yml
+++ b/.github/workflows/release-connect-bump-versions.yml
@@ -23,6 +23,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ secrets.TREZOR_BOT_TOKEN }}
+          # Fetch all commit history because Script connect-bump-versions.ts needs them to produce CHANGELOG for packages.
+          fetch-depth: 0
+          # `ref` makes sure that we checkout the branch we are running workflow on.
+          ref: ${{ github.head_ref }}
+          # `submodules` are required to run `yarn build:libs`.
+          submodules: true
 
       - name: Setup node
         uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/release-connect-bump-versions.yml
```diff
@@ -23,6 +23,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ secrets.TREZOR_BOT_TOKEN }}
+          # Fetch all commit history because Script connect-bump-versions.ts needs them to produce CHANGELOG for packages.
+          fetch-depth: 0
+          # `ref` makes sure that we checkout
+          ref: ${{ github.head_ref }}
+          # `submodules` are required to run `yarn build:libs`
+          submodules: true
```
mroz22, line 31: hm true, good you put the comment!

.github/workflows/release-connect-bump-versions.yml
```diff
@@ -23,6 +23,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ secrets.TREZOR_BOT_TOKEN }}
+          # Fetch all commit history because Script connect-bump-versions.ts needs them to produce CHANGELOG for packages.
+          fetch-depth: 0
+          # `ref` makes sure that we checkout
```
mroz22: makes sure that we checkout <something>? is the sentence finished ? : P

mroz22 (approved):

.github/workflows/release-connect-bump-versions.yml
karliatto: Ups, you are right, I just completed the sentence. Thanks!

PR #12753, opened 2024-06-06
ci(gitlab): get rid of GitLab CI

karliatto:
## Description
Get rid of unused code in build and deploy GitLab CI.


mroz22 (approved):
afuera. 

also the other CI_JOB env variables in docker files are probably useless. Or does github use the same variables names?

mroz22:
btw it breaks gitlab now

vdovhanych:
we use CI_JOB for some tests in github 

```

      - name: Run connect popup test
        env:
          URL: https://${{ inputs.DEV_SERVER_HOSTNAME }}/connect/${{ steps.extract_branch.outputs.branch }}/
          # skip settings page, this url is set at build time anyway
          #TREZOR_CONNECT_SRC: https://${{ inputs.DEV_SERVER_HOSTNAME }}/connect/${{ steps.extract_branch.outputs.branch }}/
          CI_COMMIT_BRANCH: ${{ steps.extract_branch.outputs.branch }}
          CI_JOB_NAME: ${{ inputs.test-name }}-${{ github.run_attempt }}          
```
we should change it to something else or use native value from github

mroz22 (changes requested):
uf thanks. I didn't notice we re-create these variables. thought they were abandoned

vdovhanych (changes requested):
Yes we should either keep it or change it to something else.

karliatto:
ok guys you are right @mroz22 @vdovhanych fixed!

Thanks!

vdovhanych (changes requested):
Also please fix this error from gitlab 

```
connect-popup legacy npm package manual: [9.0.10, browser-support.test] job: undefined need: connect-web build
```
We don't use it but its should be valid configuration

vdovhanych:
Let's completely remove Gitlab, as it's no longer being used at all since last week.

karliatto:
> Let's completely remove Gitlab, as it's no longer being used at all since last week.

Something like https://github.com/trezor/trezor-suite/pull/12753/commits/5d524e1d251c64ccfae6f0d837df83c07c1c2833  ?

What about moving the current `ci/scripts` to a new one `.github/scripts` ? https://github.com/trezor/trezor-suite/pull/12753/commits/59d30fbbd536663a35b6b82bd82c0412df5fedad

vdovhanych:
I don't know if I like the scripts in the `.github` folder. If we're to move it, I would put it in the `scripts/ci/`

karliatto:
> I don't know if I like the scripts in the `.github` folder. If we're to move it, I would put it in the `scripts/ci/`

@vdovhanych that makes a lot of sense. Thanks! Done https://github.com/trezor/trezor-suite/pull/12753/commits/9a0e40a3cdc31656d82f7ee5fa89d15f681797a7

scripts/ci/tsconfig.json
Nodonisko, line 1: I think this file is not necessary to move, also it should be kept in original location too. So if needed here please copy it intead of moving.

karliatto: Ok, I get that it can used the the tsconfig from scripts but why should it be kept in original location too? This PR is getting rid of the original location.

Removing the unnecessary file: 66f9ca51b499f9c5b0d0a176b040ab0414475bde

scripts/ci/check-npm-dependencies.ts
```diff
@@ -16,15 +16,17 @@ const packagesPath = path.join(rootPath, 'packages');
 const args = process.argv.slice(2);
 
 if (args.length < 2)
-    throw new Error('Usage: yarn tsx check-npm-dependencies.ts <package_name> <semver>');
+    throw new Error('Usage: yarn tsx check-npm-dependencies.ts <packageName> <semanticVersion>');
 
-const [packageName, semver] = args;
+const [packageName, semanticVersion] = args;
```
karliatto, line 21: Changing names in this files since the ts rules applied in this path are complaining about it having same name as the package `semver` that is used in imported helper.

scripts/ci/tsconfig.json
mroz22: ok, so it was removed because it was not needed in the end right?

mroz22 (approved):
I am good with it. not sure how about others? maybe @vdovhanych has something to say? if no, he should remain silent forever ⛪ 

also @karliatto please observ CI after merging this if it doesn't misbehave in some way. we are not running everything everytime..

vdovhanych (approved):
I'm fine with it, but I'm with @mroz22 on this. We should watch it a bit more carefully after we merge it.

vdovhanych (commented):
Also maybe you can completely remove gitlab-ci.yml file its not needed at all now.

scripts/ci/tsconfig.json
karliatto: Right, it was required when it was in `ci/scripts`, but now it is not required in `scripts/ci/` becase there is already tsconfig in `scripts/` and it is using that one.

vdovhanych (approved):

karliatto:
Merging this. I will keep close attention to test after merging it.
Unfortunately there are some tests failing, but the same are failing in schedule nightly job for other reason, so not related to this changes. 

- https://github.com/trezor/trezor-suite/actions/runs/9786394952
- https://github.com/trezor/trezor-suite/actions/runs/9786359584

PR #12755, opened 2024-06-06
feat(connect-explorer): verify connectSrc works

### LINKED ISSUE #12671: connect-explorer: validate connectSrc in settings page
it happily accepts incorrect urls 
I would probably do also an async check that validates if the selected enpoint is available, does it make sense? 🤔 

image

martykan:
## Description

Adds `connectSrc` validation to Connect Explorer. This is done by checking if `js/core.js` loads as a script. 

Also add a fallback to default for `connectSrc` when loading core in popup. 

## Related Issue

Resolve #12671

### DIFF
--- a/packages/connect-explorer/src/actions/trezorConnectActions.ts
+++ b/packages/connect-explorer/src/actions/trezorConnectActions.ts
@@ -17,6 +17,7 @@ export type TrezorConnectAction =
     | { type: typeof DEVICE.DISCONNECT; device: TrezorConnectDevice }
     | { type: typeof ACTIONS.ON_CHANGE_CONNECT_OPTIONS; payload: ConnectOptions }
     | { type: typeof ACTIONS.ON_HANDSHAKE_CONFIRMED }
+    | { type: typeof ACTIONS.ON_INIT_ERROR; payload: string }
     | {
           type: typeof ACTIONS.ON_CHANGE_CONNECT_OPTION;
           payload: { option: Field<any>; value: any };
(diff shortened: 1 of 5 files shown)


packages/connect-explorer/src/actions/trezorConnectActions.ts
```diff
@@ -43,6 +44,23 @@
     return !/^(?:[a-z]+:)?\/\//i.test(path);
 };
 
+const testLoadingScript = (src: string) => {
+    return new Promise((resolve, reject) => {
+        const script = document.createElement('script');
+        script.src = src;
```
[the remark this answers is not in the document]
martykan: This is technically true, but this is the exact purpose of the `connectSrc` field, so this is not an issue.

packages/connect-explorer/src/actions/trezorConnectActions.ts
```diff
@@ -117,11 +151,12 @@ export const init =
         try {
             await TrezorConnect.init(connectOptions);
         } catch (err) {
-            console.log('ERROR', err);
+            dispatch({ type: ACTIONS.ON_INIT_ERROR, payload: err.message });
 
             return;
         }
 
+        dispatch({ type: ACTIONS.ON_INIT_ERROR, payload: undefined });
```
szymonlesisz: why dispatch error on successful init?

in TrezorConnectAction payload is typed as string yet here is undefined

martykan: True, I could get rid of this and reset the error in `ON_CHANGE_CONNECT_OPTIONS` reducer

mroz22 (approved):
lgtm

PR #12760, opened 2024-06-06
fix(suite): jumping underline under tab

tomasklim:
## Screenshots:
The underline should be big as under coins under each tab
Screenshot 2024-06-06 at 13 24 18
But it was not the case sometimes
Screenshot 2024-06-06 at 13 24 14
Now it is fixed
Screenshot 2024-06-06 at 13 24 01


jvaclavik (approved):
nice catch

Issue #12731, opened 2024-06-04
connect-webextension proxy - revise desing


mroz22:
- I believe that [init](https://github.com/trezor/trezor-suite/blob/develop/packages/connect-webextension/src/proxy/index.ts#L45) should be hadled implicitly inside `call`
- does manifest method make sense here? 
- does cancel method do everything it should?
- if init was not call, and channel not set, call method throws on the first attempt to access channel, this is caught, but in the catch block later on obviously not. https://github.com/trezor/trezor-suite/blob/develop/packages/connect-webextension/src/proxy/index.ts#L95
- add default `connectSrc`. it is not set and 3rd parties have to do stuff like this https://github.com/RabbyHub/Rabby/blob/59e06afc722e69c6f609654c1f5ccba1f710710f/src/background/service/keyring/eth-trezor-keyring/trezor-offscreen-bridge.ts#L30
- lib/index.js does not contain inlined content-script. so simply importing from @trezor/connect-webextension forces you to use version where you have to vendorize content script. 

related to #12717

karliatto:
Regarding the proxy:
After talking about this with @martykan we came to the conclusion that the proxy is out of the scope of TrezorConnect and we should focus on making connect-webextension work nicely in service worker with popup with ES modules.

karliatto:
> I believe that [init](https://github.com/trezor/trezor-suite/blob/develop/packages/connect-webextension/src/proxy/index.ts#L45) should be hadled implicitly inside call

[init](https://github.com/trezor/trezor-suite/blob/develop/packages/connect-webextension/src/proxy/index.ts#L45) requires settings in functions arguments and we do not have that in call method if init has not been called before.

> lib/index.js does not contain inlined content-script. so simply importing from @trezor/connect-webextension forces you to use version where you have to vendorize content script.

Should be fixed by this PR https://github.com/trezor/trezor-suite/pull/12761

> add default connectSrc. it is not set and 3rd parties have to do stuff like this https://github.com/RabbyHub/Rabby/blob/59e06afc722e69c6f609654c1f5ccba1f710710f/src/background/service/keyring/eth-trezor-keyring/trezor-offscreen-bridge.ts#L30

> if init was not call, and channel not set, call method throws on the first attempt to access channel, this is caught, but in the catch block later on obviously not. https://github.com/trezor/trezor-suite/blob/develop/packages/connect-webextension/src/proxy/index.ts#L95

Isn't `init` required in all the TrezorConnect implementation?

[...]

PR #12761, opened 2024-06-06
fix(connect-webextension): use bundled build as package main

martykan:
## Description

Use bundled build as package main. 
This is because only the bundle has inlined content-script.

Also update `connect-webextension` to use ES Modules to test this change. 

## Related Issue

Related to #12731

### DIFF
--- a/packages/connect-explorer/src-webextension/background/serviceWorker.ts
+++ b/packages/connect-explorer/src-webextension/background/serviceWorker.ts
@@ -1,3 +1,9 @@
 /// <reference lib="webworker" />
 
-importScripts('vendor/trezor-connect-webextension.js');
+import TrezorConnect, { DEVICE_EVENT } from '@trezor/connect-webextension';
+
+// Example use of TrezorConnect
+// Without this, the import would be removed by Webpack tree-shaking
+TrezorConnect.on(DEVICE_EVENT, (event: any) => {
+    console.log('DEVICE_EVENT', event);
+});
--- a/packages/connect-explorer/src-webextension/manifest.json
+++ b/packages/connect-explorer/src-webextension/manifest.json
@@ -6,7 +6,8 @@
         "default_popup": "extension-popup.html"
     },
     "background": {
-        "service_worker": "serviceWorker.bundle.js"
+        "service_worker": "serviceWorker.bundle.js",
+        "type": "module"
     },
     "permissions": ["scripting"],
     "host_permissions": ["http://*/*", "https://*/*"]
--- a/packages/connect-webextension/package.json
+++ b/packages/connect-webextension/package.json
@@ -20,10 +20,8 @@
         "serviceworker",
         "webextension"
     ],
-    "main": "src/index.ts",
-    "publishConfig": {
-        "main": "lib/index.js"
-    },
+    "main": "build/trezor-connect-webextension.js",
+    "types": "src/index.ts",
     "files": [
         "build/trezor-connect-webextension.js",
         "build/trezor-connect-webextension.min.js",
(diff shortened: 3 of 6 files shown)


packages/connect-webextension/package.json
```diff
@@ -22,7 +22,7 @@
     ],
     "main": "src/index.ts",
     "publishConfig": {
-        "main": "lib/index.js"
+        "main": "build/trezor-connect-webextension.js"
```
mroz22: how about the minified version? does using unminified help with debugging?

packages/connect-explorer/src-webextension/manifest.json
```diff
@@ -6,7 +6,8 @@
         "default_popup": "extension-popup.html"
     },
     "background": {
-        "service_worker": "serviceWorker.bundle.js"
+        "service_worker": "serviceWorker.bundle.js",
+        "type": "module"
```
mroz22, line 10: nice, if it works, than great job. how about connet-explorer webextension build? 
do we want to address this as well? https://github.com/trezor/trezor-suite/blob/develop/packages/connect-explorer/src-webextension/background/serviceWorker.ts#L3

martykan: This is for the connect-explorer webextension build

packages/connect-webextension/package.json
martykan: I will check. The minified version is actually not that minified - no renaming variables or formatting, probably it just has pruned unused code. It might make sense to only ship that

martykan: Seems that the minified version doesn't actually work. Not even with importScripts

martykan: I would probably just remove the minified version

mroz22: ok. makes sense.

martykan:
It seems to be somewhat flaky, I will be investigating...

karliatto:
I just don't want to forget that if we release this it should go (as per our rules) to minor version not a patch.
Since it modifies the type of files we provide in the release. So in version 9.3.1.

martykan:
@SocketSecurity ignore-all

packages/connect-explorer/src-webextension/background/serviceWorker.ts
```diff
@@ -1,3 +1,12 @@
 /// <reference lib="webworker" />
 
-importScripts('vendor/trezor-connect-webextension.js');
+// in Trezor monorepo we need to specify the path since main points to `src/index.ts`
+// in a published NPM package it points to `build/trezor-connect-webextension.js`
```
karliatto: Wouldn't make sense here to point main to `build/trezor-connect-webextension.js` ? 
Since this package we do not use in any other way?

martykan: True, that would make things cleaner. I'll change it

mroz22:
I would make this minor version and give it a rich changelog record 👍

packages/connect-explorer/src-webextension/background/serviceWorker.ts
```diff
@@ -1,3 +1,7 @@
 /// <reference lib="webworker" />
 
-importScripts('vendor/trezor-connect-webextension.js');
+import TrezorConnect, { DEVICE_EVENT } from '@trezor/connect-webextension';
+
+TrezorConnect.on(DEVICE_EVENT, (event: any) => {
+    console.log('DEVICE_EVENT', event);
+});
```
karliatto, line 9: If I understood well, we are adding it here just so Webpack does not get rid of the previous import.

Maybe adding a comment here about why we add here this log could be useful. Since we are not doing it for the obvious reason which is having a log but for the reason that we do not want Webpack to do tree-shaking with it.

:question:

mroz22: I am not sure this is the case. Isn't it only to test that both named and default exports work as expected?

martykan: Yes it's actually because Webpack was getting rid of it, at least with our config

karliatto: What about adding some comments about why this log is here?
Not only for other people looking at this but for our future us, so we don't forget about the reason.

karliatto (approved):
Looks good to me.

PR #12763, opened 2024-06-06
fix(connect-ui): install bridge title align

### LINKED ISSUE #12673: connect-popup - install bridge page broken UI (misaligned header)
image

karliatto:
## Description

Fix install bridge title alignment.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12673

## Screenshots:

image

### DIFF
--- a/packages/connect-ui/src/components/View.tsx
+++ b/packages/connect-ui/src/components/View.tsx
@@ -8,6 +8,7 @@ const Title = styled(H3)`
     font-size: 19px;
     color: #333;
     font-weight: ${variables.FONT_WEIGHT.DEMI_BOLD};
+    text-align: center;
 `;
 
 const StyledP = styled(Paragraph)`


packages/connect-ui/src/views/Transport.tsx
```diff
@@ -7,9 +8,16 @@ import imageSrc from '../images/man_with_laptop.svg';
 
 export type TransportEventProps = Extract<UiEvent, { type: 'ui-no_transport' }>;
 
+const Title = styled(H3)`
```
martykan: Why not change this in the View component directly? I think we always want it centered

karliatto: I was afraid of modifying something unexpectedly but you are right, actually before it was all centered, just it was broken by some changes in H3 component.

So there there is fixup 92d0ec0651c791e5c430b262a95766321d527dbe

martykan (approved):
👍

PR #12766, opened 2024-06-06
chore(suite): setup ev certificate signing for windows app

vdovhanych:
This adds functionality to sign the Windows app using a new EV signing certificate.
This is done based on the [electron-builder documentation](https://www.electron.build/tutorials/code-signing-windows-apps-on-unix#signing-windows-app-on-maclinux-using-jsign) that suggests doing the signing this way if you want to do it on Linux/macOS 

Related with - #12767


packages/suite-desktop-core/scripts/sign-windows.ts
```diff
@@ -0,0 +1,18 @@
+exports.default = async function (configuration) {
```
komret, line 2: Shouldn't be async? See Files changed for other linter warnings as well.

komret (changes requested):

packages/suite-desktop-core/scripts/sign-windows.ts
vdovhanych: Yeah was just fixing it :) I took the script from electron documentation i guess its not really good one.. :D

komret (approved):

packages/suite-desktop-core/scripts/sign-windows.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    execSync(
+        `java -jar ../suite-desktop-core/scripts/jsign-6.0.jar --keystore ../suite-desktop-core/scripts/hardwareToken.cfg --storepass '${TOKEN_PASSWORD}' --storetype PKCS11 --tsaurl http://timestamp.digicert.com --alias "${CERTIFICATE_NAME}" "${configuration.path}"`,
+        {
+            stdio: 'inherit',
+        },
+    );
+}
+
+export default signWindows;
```
matejkriz:
```suggestion
// eslint-disable-next-line import/no-default-export
export default signWindows;
```

matejkriz:
@komret this needs to be cherry-picked later in release/24.6 (only, not in 24.5)

PR #12767, opened 2024-06-06
Hotfix/release 24 5 intermediary

matejkriz:
## Description

- SatoshiLabs, s.r.o. renamed to Trezor Company s.r.o. so there is a change in publisher name of Win signing certificate and we need a version of Suite signed by old certificate but accepting new certificate during auto-update.
- Because of new publisher name in Win signing certificate, we have to switch to a new auto-update provider so users on older Suite versions get always this version at first and then they can update on latest Suite signed by new certificate.
- Signature file is downloaded from the same URL as installation file (which is fine, public key is bundled inside Suite)
- EAP version is distributed in different bucket/folder
- Changelog has to be bundled with installation files, no longer downloads from GitHub.


matejkriz:
The lint issue can be ignored here, it's totally unrelated to the changes and fixed in develop.

matejkriz:
closing the PR in favour of https://github.com/trezor/trezor-suite/pull/12788, keeping branch as it is based on release/24.5

PR #12769, opened 2024-06-06
Feat/transport api abortable

szymonlesisz:
## Description

Abort transport apis using AbortSignal

### DIFF
--- a/packages/transport-bridge/src/core.ts
+++ b/packages/transport-bridge/src/core.ts
@@ -12,20 +12,20 @@ import { Session } from '@trezor/transport/src/types';
 import { Log } from '@trezor/utils';
 import { AbstractApi } from '@trezor/transport/src/api/abstract';
 
-const abortController = new AbortController();
-export const sessionsBackground = new SessionsBackground({ signal: abortController.signal });
-
-export const sessionsClient = new SessionsClient({
-    requestFn: args => sessionsBackground.handleMessage(args),
-    registerBackgroundCallbacks: () => {},
-});
-
-sessionsBackground.on('descriptors', descriptors => {
-    sessionsClient.emit('descriptors', descriptors);
-});
-
 export const createApi = (apiArg: 'usb' | 'udp' | AbstractApi, logger?: Log) => {
     let api: AbstractApi;
+
+    const abortController = new AbortController();
+    const sessionsBackground = new SessionsBackground({ signal: abortController.signal });
+
+    const sessionsClient = new SessionsClient({
+        requestFn: args => sessionsBackground.handleMessage(args),
+        registerBackgroundCallbacks: () => {},
+    });
+
+    sessionsBackground.on('descriptors', descriptors => {
+        sessionsClient.emit('descriptors', descriptors);
+    });
 
     if (typeof apiArg === 'string') {
         api =
@@ -269,5 +269,6 @@ export const createApi = (apiArg: 'usb' | 'udp' | AbstractApi, logger?: Log) => 
         send,
         receive,
         dispose,
+        sessionsClient,
     };
 };
--- a/packages/transport-bridge/src/core.ts
+++ b/packages/transport-bridge/src/core.ts
@@ -52,7 +52,15 @@ export const createApi = (apiArg: 'usb' | 'udp' | AbstractApi, logger?: Log) => 
         sessionsClient.enumerateDone({ descriptors });
     });
 
-    const writeUtil = async ({ path, data }: { path: string; data: string }) => {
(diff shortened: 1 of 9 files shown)


packages/transport/src/types/apiCall.ts
```diff
@@ -13,6 +13,7 @@ export type ErrorGeneric<ErrorType> = ErrorType extends AnyError
     ? {
           success: false;
           error: ErrorType;
+          message?: string;
```
szymonlesisz: to discuss with @mroz22

packages/transport-bridge/src/core.ts
```diff
@@ -104,11 +112,15 @@ export const createApi = (apiArg: 'usb' | 'udp' | AbstractApi, logger?: Log) =>
...
+    const enumerate = async ({ signal }: { signal: AbortSignal }) => {
         await sessionsClient.enumerateIntent();
 
-        const enumerateResult = await api.enumerate();
+        const enumerateResult = await api.enumerate(signal);
         if (!enumerateResult.success) {
+            await sessionsClient.enumerateDone({
+                descriptors: [],
+            });
+
```
mroz22: should be possible to remove lines 116 and 120-123. 
concurrent enumeration has now been allowed for some time and only now I am removing some left over redundant code in https://github.com/trezor/trezor-suite/pull/12954

mroz22: Yes, it seems to work. So I suggest merging #12954 first and dropping [3861988](https://github.com/trezor/trezor-suite/pull/12769/commits/3861988743306fd275fb3521a9e6e61af9bfe7af)

packages/transport/src/types/apiCall.ts
mroz22: soo, you found a situation where there was a typed error based on this list https://github.com/trezor/trezor-suite/blob/feat/tranport-api-abortable/packages/transport/src/errors.ts#L1 
and some loosely typed string error message at the same time?

mroz22: hmm yes, this type should be produced either by 
https://github.com/trezor/trezor-suite/blob/feat/tranport-api-abortable/packages/transport/src/utils/result.ts#L12
or by
https://github.com/trezor/trezor-suite/blob/feat/tranport-api-abortable/packages/transport/src/utils/result.ts#L15

and both these functions have error message.

mroz22: maybe something like this? https://github.com/trezor/trezor-suite/pull/13044

packages/transport/src/api/udp.ts
```diff
@@ -25,25 +25,39 @@ export class UdpApi extends AbstractApi {
                 typeof ERRORS.INTERFACE_DATA_TRANSFER | typeof ERRORS.UNEXPECTED_ERROR
             >
         >(resolve => {
+            signal?.addEventListener('abort', () => {
+                resolve(
+                    this.error({
+                        error: ERRORS.INTERFACE_DATA_TRANSFER,
+                        message: ERRORS.ABORTED_BY_SIGNAL,
+                    }),
```
mroz22: I would rather return only
```
 this.error({
      error: ERRORS.ABORTED_BY_SIGNAL,
 }),
```
we can discuss

mroz22: hmm still thinking about it. its not a big deal. just leaving this note here. maybe it can stay as it is. not a strong opinion.

packages/transport-bridge/src/http.ts
```diff
@@ -154,6 +164,7 @@ export class TrezordNode {
                 parseBodyJSON,
                 (req, res) => {
                     res.setHeader('Content-Type', 'text/plain');
+                    // TODO: how to abort subscription?
```
mroz22: hmm imho aborting a single request should not clear subscriptions since they might have been created by another client

packages/transport-bridge/tests/http.test.ts
```diff
@@ -253,7 +356,7 @@ describe('http', () => {
                 return { server, client, changeDescriptorsOnApi, onListenResolvedSpy };
             };
 
-            it('api emitting events and listen correctly reporting', async () => {
+            it.skip('api emitting events and listen correctly reporting', async () => {
```
mroz22: skip.

packages/transport-bridge/src/core.ts
mroz22: well, not dropping it all together actually, there are imho valuable changes here. may I force push this commit?

mroz22 (approved):

PR #12772, opened 2024-06-06
fix(connect): preserve DEVICE.ACQUIRED listeners after unsuccessful workflow

szymonlesisz:
## Description

Currently unacquired => acquired works only if first attempt is successful, otherwise all listeners are removed and DEVICE.ACQUIRED is never fired again


mroz22:
yep, I already noticed this edgecase too and arrived the same conclusion that it should be fixed something like this. Still I would like to summon @marekrjpolak to take a look 🙏

marekrjpolak (approved):
Haven't tested it, but seems reasonable and uncontroversial.

PR #12773, opened 2024-06-06
fix(transport-bridge): pass type to listen,stable stringify comparison

mroz22:
the new lib - we already had it long tie ago https://github.com/trezor/trezor-link/blob/master/src/lowlevel/withSharedConnections.ts#L14 and it served the very sume purpose that I am adding it now for. Previously, webusb ran neverending enumeration cycle and only not so long ago (1year maybe) we switched to events. 

Subscriptions for legacy vs node are slightly different. 

Legacy bridge

image
image

Node Bridge

image
image


szymonlesisz (approved):
fixed for me :+1:

PR #12785, opened 2024-06-07
chore(connect-webextension): stop publishing proxy

karliatto:
## Description

Stop publishing connect-webextension proxy that was initially created only for internal use in order to be able to support connect-explorer with webextension so we can use it to run popup tests with webextension package.

This proxy to allow communication between foreground (UI) of the webextension and service worker is out of the scope of TrezorConnect and each implementer should handle it by themself.

This will require a minor release not just patch since it changes the files we publish. Along with https://github.com/trezor/trezor-suite/pull/12761

## Related Issue

Related https://github.com/trezor/trezor-suite/issues/12731

### DIFF
--- a/packages/connect-webextension/package.json
+++ b/packages/connect-webextension/package.json
@@ -28,7 +28,8 @@
         "build/trezor-connect-webextension.js",
         "build/trezor-connect-webextension.min.js",
         "build/content-script.js",
-        "lib/"
+        "lib/",
+        "!lib/proxy"
     ],
     "scripts": {
         "lint:js": "yarn g:eslint '**/*.{ts,tsx,js}'",


mroz22:
aren't rabby ppl using it?

karliatto:
> aren't rabby ppl using it?

It does not look like they are using it based on their code github.com/RabbyHub/Rabby

But I asked them to double check it in case I missed anything. https://satoshilabs.slack.com/archives/C06Q0B1T56U/p1718004605082949


EDIT: Confirmed. Rabbby webextension is not using proxy.

karliatto:
If we get rid of proxy in NPM I think we should modify `packages/connect-examples/webextension-mv3-sw` to not use it, and maybe extend that example to be more useful for developers with more comments.

karliatto:
> aren't rabby ppl using it?

Con

martykan (approved):

Issue #12236, opened 2024-05-02
Autoupdate not working in EAP when a newer tag exists


komret:
Mobile release tags break Suite autoupdate for users in Early Access Program while the tag for mobile app is newer than for desktop, even if it is not marked as `Latest` or `Pre-release`. This is because Suite accesses https://github.com/trezor/trezor-suite/releases.atom, not https://github.com/trezor/trezor-suite/releases/latest when EAP is on. This is possibly an Electron bug. Needs to be investigated.

![image](https://github.com/trezor/trezor-suite/assets/42465546/511b678a-8dbe-44e8-8df4-dccf58647a3c)

MiroslavProchazka:
Is kinda solved for now by deleting the Suite Lite release tag and whole release from GH.

matejkriz:
But we still want to find a way how to publish other things than Suite Desktop without interfering with the auto-update.

matejkriz:
This is no longer an issue with new auto-update provider https://github.com/trezor/trezor-suite/pull/12788

PR #12788, opened 2024-06-07
Feat/new auto update provider

### LINKED ISSUE #4692: Improve autoupdate download speed and stability
As some noticed the download speed during autoupdate is somewhat unsatisfactory. Most likely (though I haven't found clear numbers) Github throttles the download speed. It would be great to download binaries directly from our S3 bucket instead of Github.

Note that we if agree upon https://github.com/trezor/webwallet-data/issues/34 we can simply replace the github link with data.trezor.io (see the issue for details) and that's it. I hope that is easy to do in the autoupdater.

matejkriz:
## Description

- SatoshiLabs, s.r.o. renamed to Trezor Company s.r.o. so there is a change in publisher name of Win signing certificate and we need a version of Suite signed by old certificate but accepting new certificate during auto-update.
- Because of new publisher name in Win signing certificate, we have to switch to a new auto-update provider so users on older Suite versions get always this version at first and then they can update on latest Suite signed by new certificate.
- Signature file is downloaded from the same URL as installation file (which is fine, public key is bundled inside Suite)
- EAP version is distributed in different bucket/folder
- Changelog is no longer fetched from GitHub, has to be inside latest.yml (latest-mac.yml, latest-linux.yml, latest-linux-arm64.yml) files.
  - It is automatically copied there during build if you put it in packages/suite-desktop/releaseNotes/release-notes.md file (or platform specific release-notes-mac.md release-notes-linux.md or release-notes-windows.md).
  - It is technically posible to change the release notes later manually. We would like to provide a github action for it later.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/4692
Maybe it can solve https://github.com/trezor/trezor-suite/issues/3433 (at least decrease number of issues related to auto-update in Sentry)

This is still opened issue https://github.com/trezor/trezor-suite/issues/12236
in a new auto-update flow, EAP on older Suite versions will be broken (user will not see any new update until EAP turned off)


matejkriz:
@komret This is supposed to be cherry-picked to both release/24.6 and release/24.5

tomasklim (approved):
LGTM

packages/suite-desktop-core/src/modules/auto-updater.ts
```diff
@@ -237,8 +231,8 @@ export const init: Module = ({ mainWindow, store }) => {
         mainWindow.webContents.send('update/allow-prerelease', value);
         const settings = store.getUpdateSettings();
         store.setUpdateSettings({ ...settings, allowPrerelease: value });
-
-        autoUpdater.allowPrerelease = value;
+        feedURL = allowPrerelease ? defaultFeedURL.preRelease : defaultFeedURL.latest;
```
matejkriz: bug 🐞

PR #12809, opened 2024-06-10
fix(token-definitions): unify cardano addresses

tomasklim:
## Description

- strip encoded name in cardano `policy id`

## Related Issue

Resolve

### DIFF
--- a/suite-common/token-definitions/package.json
+++ b/suite-common/token-definitions/package.json
@@ -17,6 +17,7 @@
         "@suite-common/redux-utils": "workspace:*",
         "@suite-common/wallet-config": "workspace:*",
         "@suite-common/wallet-types": "workspace:*",
+        "@trezor/blockchain-link-utils": "workspace:*",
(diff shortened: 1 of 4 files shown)


suite-common/token-definitions/package.json
```diff
@@ -23,6 +23,7 @@
         "jws": "^4.0.0"
     },
     "devDependencies": {
+        "@trezor/blockchain-link-utils": "workspace:*",
```
AdamSchinzel: This is on purpose? I think i should be inside `dependencies`

tomasklim: Why, it is used in scripts, not in the app. I will move it but there has to be something wrong with configuration

tomasklim: it is used in the same place as `ajv` and that dep can be in devDependencies so i dont know

AdamSchinzel (approved):
LGTM

PR #12813, opened 2024-06-10
fix(suite): copy in onboarding

peter-sanderson:


komret:
It's OK but let's not merge it until Crowdin sync.

komret:
Also validation is failing on formatting.

peter-sanderson:
@Hermez-cz I think this is no longer relevant correct? I see there is conflict in code and it seems copy was already adjusted.

PR #12834, opened 2024-06-11
feat(suite): label device select button in firmware installation

martykan:
## Description

Add a label to the button that appears in WebUSB mode to select device when going from or to bootloader.

## Screenshots:

Before
Screenshot 2024-06-11 at 12 54 22

After
Screenshot 2024-06-11 at 14 48 31

### DIFF
--- a/packages/suite/src/components/suite/WebUsbButton.tsx
+++ b/packages/suite/src/components/suite/WebUsbButton.tsx
@@ -1,20 +1,30 @@
 import TrezorConnect from '@trezor/connect';
 import { ButtonProps, Button } from '@trezor/components';
-import { Translation } from './Translation';
+import { Translation, TranslationKey } from './Translation';
 
-export const WebUsbButton = (props: Omit<ButtonProps, 'children'>) => (
+interface WebUsbButtonProps extends Omit<ButtonProps, 'children' | 'icon'> {
+    translationId?: TranslationKey;
+    icon?: ButtonProps['icon'] | false;
+}
+
+export const WebUsbButton = ({
+    translationId = 'TR_CHECK_FOR_DEVICES',
+    icon = 'SEARCH',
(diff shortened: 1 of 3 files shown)


packages/suite/src/components/suite/WebUsbButton.tsx
```diff
@@ -1,20 +1,25 @@
...
-import { Translation } from './Translation';
+import { Translation, TranslationKey } from './Translation';
 
-export const WebUsbButton = (props: Omit<ButtonProps, 'children'>) => (
+interface WebUsbButtonProps extends Omit<ButtonProps, 'children' | 'icon'> {
+    translationId?: TranslationKey;
+    icon?: ButtonProps['icon'] | 'NONE';
+}
+
+export const WebUsbButton = (props: WebUsbButtonProps) => (
```
komret: I'd define the default values for props here.

packages/suite/src/components/suite/WebUsbButton.tsx
```diff
@@ -1,20 +1,25 @@
 import TrezorConnect from '@trezor/connect';
 import { ButtonProps, Button } from '@trezor/components';
-import { Translation } from './Translation';
+import { Translation, TranslationKey } from './Translation';
 
-export const WebUsbButton = (props: Omit<ButtonProps, 'children'>) => (
+interface WebUsbButtonProps extends Omit<ButtonProps, 'children' | 'icon'> {
+    translationId?: TranslationKey;
+    icon?: ButtonProps['icon'] | 'NONE';
```
komret: I'd rather use 'false' instead of `'NONE'` because `'NONE'` could eventually be an icon name. Alternatively, you could add a `noIcon` prop.

komret (changes requested):

komret (approved):

PR #12838, opened 2024-06-11
ci(github): build webextension in staging when releasing for proper testing

karliatto:
Build webextension in staging when releasing for proper testing.

### DIFF
--- a/.github/workflows/release-connect-v9-staging.yml
+++ b/.github/workflows/release-connect-v9-staging.yml
@@ -66,6 +66,8 @@ jobs:
           awsRegion: "eu-central-1"
           serverHostname: "staging-connect.trezor.io"
           serverPath: ${{ needs.extract-version.outputs.version }}
+          buildArtifacts: "true"
+          uploadArtifacts: "true"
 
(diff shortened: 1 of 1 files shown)


vdovhanych (commented):
why do we need artifacts for the prod builds?

karliatto:
> why do we need artifacts for the prod builds?

When building staging QA should be testing webextension with the staging connect src in order to be as close to production as possible.

vdovhanych:
> > why do we need artifacts for the prod builds?
> 
> When building staging QA should be testing webextension with the staging connect src in order to be as close to production as possible.

Yes i understand but this only builds the webextension for artifacts and uploads it there but is not deployed anywhere else

karliatto:
> > > why do we need artifacts for the prod builds?
> > 
> > 
> > When building staging QA should be testing webextension with the staging connect src in order to be as close to production as possible.
> 
> Yes i understand but this only builds the webextension for artifacts and uploads it there but is not deployed anywhere else

Right, it is upload to GitHub so QA can download and test locally.

vdovhanych (approved):
Ok np then, lets merge.

PR #12871, opened 2024-06-12
feat(suite): Implement ResizableBox component

### LINKED ISSUE #11681: Create ResizableBox component
## Description

Create a generic component `ResizableBox` into the [Trezor design system](https://github.com/trezor/trezor-suite/tree/develop/packages/components/src/components) for drag & drop resizing elements.

- It should follow basic UI patterns used for this type of component. 
- whole side should be draggable, it should be highlighted on hover and should use appropriate cursor and highlighting (check screenshot below).
- It should be implemented with Javascript (ideally without external dependencies)
- create storybook with example

## Screenshots

Screenshot 2024-03-21 at 16 03 57

image

https://github.com/trezor/trezor-suite/assets/858321/06d9bae7-8380-4204-8f14-aa08ccae1b88

## Interface

Props:
```typescript
type ResizableBoxProps = {
  children: React.ReactNode;
  directions: Array<"top" | "left" | "right" | "bottom">; 
  isLocked?: boolean; // if true, it's not possible to resize the component (default = false)
  width?: number;
  minWidth?: number;
  maxWidth?: number;
  height?: number;
  minHeight?: number;
  maxHeight?: number;
}
```

Usage:
```tsx
<ResizableBox width={250} minHeight={60} maxHeight={400} directions=["right"]>
  <Sidebar 
    ...
  />  
</ResizableBox>
```

hstastna:
## Description

This PR implements new `ResizableBox` component into the Trezor design system for drag & drop resizing elements. The component should be easily resizable from any side. Storybook with examples included.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/11681

## Screenshots:
Video of the actual result in storybook:

https://github.com/trezor/trezor-suite/assets/13417815/423c6f28-ac63-448d-98c1-77d0e2a0242c

### DIFF
--- a/packages/components/src/components/ResizableBox/ResizableBox.tsx
+++ b/packages/components/src/components/ResizableBox/ResizableBox.tsx
@@ -1,0 +1,283 @@
+import styled from 'styled-components';
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { createCooldown } from '@trezor/utils';
+
+type Direction = 'top' | 'left' | 'right' | 'bottom';
+type Directions = Array<Direction>;
+
+export type ResizableBoxProps = {
+    children: React.ReactNode;
+    directions: Directions;
+    isLocked?: boolean;
+    width?: number;
+    minWidth?: number;
+    maxWidth?: number;
+    height?: number;
+    minHeight?: number;
+    maxHeight?: number;
+};
+
+type ResizersProps = {
+    $width?: number;
+    $minWidth?: number;
+    $maxWidth?: number;
+    $height?: number;
+    $minHeight?: number;
+    $maxHeight?: number;
+
+    $highlightDirection?: Direction | null;
+    $isResizing?: boolean;
+};
+
+const MINIMAL_BOX_SIZE = 1;
+const REACTIVE_AREA_WIDTH = 15;
+
+const Resizers = styled.div<ResizersProps>(
+    ({
+        $width,
+        $minWidth,
+        $maxWidth,
+        $height,
+        $minHeight,
+        $maxHeight,
+        $highlightDirection,
+        $isResizing,
+        theme,
+    }) => `
+        ${$width ? `width: ${$width}px;` : 'width: auto;'};
+        ${$minWidth && `min-width: ${$minWidth}px;`};
+        ${$maxWidth && `max-width: ${$maxWidth}px;`};
+        ${$height ? `height: ${$height}px;` : 'height: auto;'};
+        ${$minHeight && `min-height: ${$minHeight}px;`};
+        ${$maxHeight && `max-height: ${$maxHeight}px;`};
+        box-sizing: border-box;
+        position: relative;
+
+        ${$isResizing && `user-select: none;`};
+
+        &::after {
+            content:'';
+            position: absolute;
+            top: 0;
+            left: 0;
+            display:block;
+            width: 100%;
+            height: 100%;
+            pointer-events: none;
+            ${
+                $highlightDirection !== null
+                    ? `border-${$highlightDirection}: 3px solid ${theme.borderFocus};`
+                    : ''
+            };
+        }
+    `,
+);
+
+const TopHandler = styled.div`
+    position: absolute;
+    width: 100%;
+    height: ${REACTIVE_AREA_WIDTH}px;
+    cursor: ns-resize;
+    top: ${`-${REACTIVE_AREA_WIDTH / 2}px`};
+`;
+
+const BottomHandler = styled.div`
+    position: absolute;
+    width: 100%;
+    height: ${REACTIVE_AREA_WIDTH}px;
+    cursor: ns-resize;
+    bottom: ${`-${REACTIVE_AREA_WIDTH / 2}px`};
+`;
+
+const LeftHandler = styled.div`
+    position: absolute;
+    width: ${REACTIVE_AREA_WIDTH}px;
+    height: 100%;
+    bottom: 0;
+    cursor: ew-resize;
+    left: ${`-${REACTIVE_AREA_WIDTH / 2}px`};
+`;
+
+const RightHandler = styled.div`
+    position: absolute;
+    width: ${REACTIVE_AREA_WIDTH}px;
+    height: 100%;
+    bottom: 0;
+    cursor: ew-resize;
+    right: ${`-${REACTIVE_AREA_WIDTH / 2}px`};
+`;
+
+const Child = styled(Resizers)`
+    position: relative;
+    width: 100%;
+    height: 100%;
+    overflow: auto;
+`;
+
+const ensureMinimalSize = (size: number): number =>
+    size < MINIMAL_BOX_SIZE ? MINIMAL_BOX_SIZE : size;
+
+const getMinResult = (min: number, result: number) => (result > min ? result : min);
+
+const getMaxResult = (max: number | undefined, result: number) =>
+    max === undefined || result < max ? result : max;
+
+export const ResizableBox = ({
+    children,
+    directions,
+    isLocked = false,
+    width,
+    minWidth = 0,
+    maxWidth,
+    height,
+    minHeight = 0,
+    maxHeight,
+}: ResizableBoxProps) => {
+    const resizableBoxRef = useRef<HTMLDivElement>(null);
+
+    const [newX, setNewX] = useState<number>(0);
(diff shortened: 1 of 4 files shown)


packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -0,0 +1,281 @@
...
+const ensureMinimalSize = (size: number): number => (size < 1 ? 1 : size);
+
+// get the final width/height when resizing according to the minWidth/minHeight
+const getMinResult = (min: number, result: number) => (result > min ? result : min);
+
+// get the final width/height when resizing according to the maxWidth/maxHeight
+const getMaxResult = (max: number | undefined, result: number) =>
+    max === undefined || result < max ? result : max;
+
+export const ResizableBox = ({
```
jvaclavik, line 125: Please export this component in `packages/components/src/index.ts`

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -0,0 +1,281 @@
+import styled from 'styled-components';
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { createCooldown } from '@trezor/utils';
+
+type Directions = 'top' | 'left' | 'right' | 'bottom';
+
+export type ResizableBoxProps = {
+    children: React.ReactNode;
+    directions: Array<Directions>;
+    isLocked?: boolean; // if true, it's not possible to resize the component
```
jvaclavik: nit: Please delete this comment 🙏  I think the name is self-explanatory.

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -0,0 +1,281 @@
...
+            ${
+                $highlightDirection !== undefined
+                    ? `border-${$highlightDirection}: 3px solid ${theme.borderFocus};`
+                    : ''
+            };
+        }
+    `,
+);
+
+const TopDiv = styled.div`
```
jvaclavik: nit: To understand purpose of these components I'd rename it to `TopHandler` etc. `Div` doesn't say much about the cause.

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+// make sure the final width or height will be at least 1px while resizing, to make sure the box doesn't disappear
+const ensureMinimalSize = (size: number): number => (size < 1 ? 1 : size);
+
+// get the final width/height when resizing according to the minWidth/minHeight
+const getMinResult = (min: number, result: number) => (result > min ? result : min);
+
+// get the final width/height when resizing according to the maxWidth/maxHeight
+const getMaxResult = (max: number | undefined, result: number) =>
+    max === undefined || result < max ? result : max;
```
jvaclavik: I believe it isn't necessary to have comments here, because these functions are really easy to read. I'd try not to overuse comments for basic functionality.

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -0,0 +1,281 @@
...
+}: ResizableBoxProps) => {
+    const resizableBoxRef = useRef<HTMLDivElement>(null);
+
+    const [newX, setNewX] = useState<number>(0);
+    const [newY, setNewY] = useState<number>(0);
+    const [newWidth, setNewWidth] = useState<number>(width || minWidth);
+    const [newHeight, setNewHeight] = useState<number>(height || minHeight);
+    const [isResizing, setIsResizing] = useState<boolean>(false);
+    const [isHovering, setIsHovering] = useState(false);
+    const [direction, setDirection] = useState<Directions>();
```
jvaclavik: Not an issue, but I suggest to always type an initial value.

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -0,0 +1,281 @@
...
+    maxHeight,
+}: ResizableBoxProps) => {
+    const resizableBoxRef = useRef<HTMLDivElement>(null);
+
+    const [newX, setNewX] = useState<number>(0);
+    const [newY, setNewY] = useState<number>(0);
+    const [newWidth, setNewWidth] = useState<number>(width || minWidth);
+    const [newHeight, setNewHeight] = useState<number>(height || minHeight);
+    const [isResizing, setIsResizing] = useState<boolean>(false);
+    const [isHovering, setIsHovering] = useState(false);
```
jvaclavik: To keep consistency I suggest add type also here, you did it in the previous line.

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+    const handleMouseOverDirection = (direction: Directions) => {
+        if (isResizing === false) {
+            setIsHovering(true);
+            setDirection(direction);
+        }
+    };
+
+    const highlightDirection = () =>
+        isHovering === true || isResizing === true ? direction : undefined;
```
jvaclavik: Btw this is not an issue, just my suggestion how to distinguish between `null` and `undefined`:

`
null and undefined: These are two special values. They’re special because there’s a lot of things you can’t do with them — they often cause errors. Usually, null represents that some value is missing intentionally, and undefined represents that a value is missing unintentionally. However, when to use either is left to the programmer. They exist because sometimes it’s better for an operation to fail than to proceed with a missing value.
`

https://overreacted.io/what-is-javascript-made-of/

packages/components/src/components/ResizableBox/ResizableBox.tsx
hstastna: Done :)

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -140,8 +138,8 @@ export const ResizableBox = ({
     const [newWidth, setNewWidth] = useState<number>(width || minWidth);
     const [newHeight, setNewHeight] = useState<number>(height || minHeight);
     const [isResizing, setIsResizing] = useState<boolean>(false);
-    const [isHovering, setIsHovering] = useState(false);
-    const [direction, setDirection] = useState<Directions>();
+    const [isHovering, setIsHovering] = useState<boolean>(false);
+    const [direction, setDirection] = useState<Directions>(null);
```
jvaclavik: Please use it this way:

```
type Directions = 'top' | 'left' | 'right' | 'bottom';

...

const [direction, setDirection] = useState<Directions | null>(null);
```

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -30,6 +28,9 @@ type ResizersProps = {
     $isResizing?: boolean;
 };
 
+const minimalBoxSize = 1;
```
jvaclavik: I suggest to use uppercase with _ for non-changable consts 

```
const MINIMAL_BOX_SIZE = 1;
const REACTIVE_AREA_WIDTH = 15;
```

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -2,12 +2,12 @@ import styled from 'styled-components';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { createCooldown } from '@trezor/utils';
 
-type Directions = 'top' | 'left' | 'right' | 'bottom';
+type Directions = 'top' | 'left' | 'right' | 'bottom' | null;
```
jvaclavik: I would rename it to singular:

`Directions -> Direction`

`Directions` should be in case of `Array<Direction>`

hstastna: Good point, thanks!

packages/components/src/components/ResizableBox/ResizableBox.tsx
hstastna: I tried exactly this one but from some reasons, it was not possible.. (and hence I added null to Directions). Will add more details in a moment..

hstastna: There was some another typo, already fixed!

jvaclavik (approved):
good job!

PR #12887, opened 2024-06-13
chore(connect): grooming in Device.state

szymonlesisz:
## Description

1. grooming in Device.state

- `internalState` and `externalState` joined into array of `DeviceState`
- `internalState` renamed to `state.sessionId`
- `externalState` renamed to `state.staticSessionId`
- added `Device._state` field, future form of Device.state

This is not a breaking change. connect accept both params: device.state as `string` and device.state as `DeviceState` object. Returned `_state` should replace `state` in the next major release 

2. unify names of Device fields

add prefix to distinguish `sessionId` (DeviceState) from `transportSession` (Session)

- `activitySessionID` -> `transportSession`
- `keepSession` -> `keepTransportSession`

### DIFF
--- a/packages/connect/src/api/getDeviceState.ts
+++ b/packages/connect/src/api/getDeviceState.ts
@@ -1,6 +1,7 @@
 // origin: https://github.com/trezor/connect/blob/develop/src/js/core/methods/GetDeviceState.js
 
 import { AbstractMethod } from '../core/AbstractMethod';
+import { ERRORS } from '../constants';
 
 export default class GetDeviceState extends AbstractMethod<'getDeviceState'> {
     init() {
@@ -8,8 +9,11 @@ export default class GetDeviceState extends AbstractMethod<'getDeviceState'> {
     }
 
     run() {
-        return Promise.resolve({
-            state: this.device.getExternalState(),
-        });
+        const state = this.device.getState();
+        if (!state?.staticSessionId) {
+            throw ERRORS.TypedError('Runtime', 'Device state not set');
+        }
+
+        return Promise.resolve({ state: state.staticSessionId, _state: state });
     }
 }
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -113,23 +113,19 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
     featuresNeedsReload = false;
 
-    acquirePromise?: ReturnType<Transport['acquire']> = undefined;
-    releasePromise?: ReturnType<Transport['release']> = undefined;
-
-    runPromise?: Deferred<void>;
+    // variables used in one workflow: acquire -> transportSession -> commands -> run -> keepTransportSession -> release
+    private acquirePromise?: ReturnType<Transport['acquire']>;
+    private releasePromise?: ReturnType<Transport['release']>;
+    private runPromise?: Deferred<void>;
+    private transportSession?: Session | null;
+    keepTransportSession = false;
+    private commands?: DeviceCommands;
 
     loaded = false;
 
     inconsistent = false;
 
     firstRunPromise: Deferred<boolean>;
-
-    activitySessionID?: Session | null;
-
-    commands?: DeviceCommands;
-
-    keepSession = false;
-
     instance = 0;
 
     // DeviceState list [this.instance]: DeviceState | undefined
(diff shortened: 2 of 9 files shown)


packages/connect/src/device/DeviceList.ts
```diff
@@ -434,7 +434,8 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> {
                     path: d.path,
                     product: d.product,
                 };
-                this.devices[d.path].activitySessionID = d.session;
+                // TODO: is this ok? transportSession should be set only as result of acquire/release
+                // this.devices[d.path].transportSession = d.session;
```
szymonlesisz, line 510: controversy here

mroz22: yes, I would also like to have this deleted. it was there for a reason I don't remember now but I will check on it.

packages/connect/src/types/device.ts
```diff
@@ -29,6 +29,15 @@ export enum FirmwareType {
     Regular = 'regular',
 }
 
+export type DeviceState = {
+    // sessionId:
+    // - T1B1 < 1.9.0 passphrase string received from UI
+    // - T2T1 < 2.3.0 result of Deprecated_PassphraseStateRequest message
```
mroz22: @Hannsek are we allowed to stop supporting these firmwares? In the way that connect would reject everything except firmware update?

Hannsek: Yes, I'll just write it to some slack channel.

mroz22:
for reference - old passphrase handling https://docs.trezor.io/trezor-firmware/common/communication/passphrase-redesign-migration.html

packages/connect/src/types/device.ts
```diff
@@ -29,6 +29,15 @@ export enum FirmwareType {
...
+    // - otherwise Features.session_id
+    sessionId?: string;
+    external?: string; // 1st address TODO: rename to fingerprint?
```
mroz22: - staticSessionId
- persistentSessionId
- publicId
- staticPublicId

🤔

packages/connect/src/types/device.ts
mroz22: these fws are already deprecated and their handling code has been removed. probably source of conflicts in this PR

packages/connect/src/types/device.ts
```diff
@@ -47,6 +56,7 @@ export type KnownDevice = {
     color?: string;
     status: DeviceStatus;
     mode: DeviceMode;
+    // state?: DeviceState; // TODO: breaking change, sessionId could be sensitive (legacy T1 passphrase)
```
mroz22: `_state`?

packages/connect/src/device/DeviceList.ts
mroz22: I will test couple of edgecase scenarios without it. If everything goes well and tests are green, I guess we can get rid of it.

packages/connect/src/types/device.ts
szymonlesisz: staticSessionId it is

packages/connect/src/device/Device.ts
```diff
@@ -112,28 +113,23 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
 
-    acquirePromise?: ReturnType<Transport['acquire']> = undefined;
-    releasePromise?: ReturnType<Transport['release']> = undefined;
-
-    runPromise?: Deferred<void>;
+    // variables used in one workflow: acquire -> transportSession -> commands -> run -> keepTransportSession -> release
+    private acquirePromise?: ReturnType<Transport['acquire']>;
+    private releasePromise?: ReturnType<Transport['release']>;
+    private runPromise?: Deferred<void>;
+    private transportSession?: Session | null;
```
szymonlesisz, line 120: should be public?

mroz22: they seem to be used privately, or at least some of them

packages/connect/src/api/getDeviceState.ts
```diff
@@ -1,15 +1,19 @@
...
     }
 
     run() {
-        return Promise.resolve({
-            state: this.device.getExternalState(),
-        });
+        const state = this.device.getState();
+        if (!state?.staticSessionId) {
+            throw ERRORS.TypedError('Runtime', 'Device state not set');
+        }
```
mroz22, line 15: there is change here, right? previously it would return undefined?
```javascript
 device.getExternalState() {
        return this.externalState[this.instance];
```

packages/connect/src/device/DeviceList.ts
mroz22: it looks ok.  I am still a bit paranoic but yes, lets get rid of it.

mroz22 (approved):

packages/connect/src/device/Device.ts
szymonlesisz: not now, but it will be used in future modules, i've rebased few parallel [branches](https://github.com/trezor/trezor-suite/blob/d4f90fcf6dcc893e17edf8ffc12a7c278baaaaf5/packages/connect/src/device/prompts.ts#L32) where i need transportSession

packages/connect/src/api/getDeviceState.ts
szymonlesisz: exactly, but it shouldn't

PR #12889, opened 2024-06-13
feat(suite-native): handle pin by state, not button requests

juriczech:
## Description
Remove PIN button requests in favor of using redux state as the primary trigger for pin actions and states. This PR shouldn't have any new functionality and aims to refactor pin logic 1:1 to match the version on develop. 

## Related Issue

Resolve


suite-native/device-authentication/src/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+import { UI } from '@trezor/connect';
+
+export const isPinAction = (action: AnyAction) =>
+    action.type === UI.REQUEST_BUTTON &&
+    (action.payload.code === 'ButtonRequest_PinEntry' ||
+        action.payload.code === 'PinMatrixRequestType_Current');
+
+export const isCloseWindowActionOrNonPinAction = (action: AnyAction) =>
+    action.type === UI.CLOSE_UI_WINDOW ||
+    (action.type === UI.REQUEST_BUTTON && !isPinAction(action));
```
matejkriz: This looks too complex to me, I would prefer to separate CLOSE_UI_WINDOW condition from request button condition for better readability.

juriczech: Good point. I already have something like this in a cleanup commit that I'm preparing :) 
image

juriczech: Separated. You can check here (I force pushed, sorry):
https://github.com/trezor/trezor-suite/blob/2ad1bf6c528f77d5567665cb766ea0bd007f163b/suite-native/device-authentication/src/utils.ts

matejkriz: From my point of view, it didn't improve readability. And it's wrong. So I guess it's really too complex... 🙂 

I would prefer to delete `isCloseWindowActionOrNonPinAction` completely change the matcher in deviceAuthenticationSlice.ts

juriczech: That can be troublesome since this is hacky that we dispatch events from connect. These actions are not created with action creator that would fulfill the conditions to use `isAnyOf` etc. That would result in having 3 different matchers just for ending pin flow and then there would be more for passphrase as well. I would prefer to make these utils more readable instead, if that could work. 

What do you think of this? https://github.com/trezor/trezor-suite/pull/12889/commits/34f49c2959868e0c62f7cb23db62fd3884232352

image

Note: I'm considering writing some kind of test for this as well.

juriczech: Updated conditions in slice here: [ff2980a](https://github.com/trezor/trezor-suite/pull/12889/commits/ff2980a3ea0386747c503e51b9591a275032c280)

suite-native/device-authentication/package.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "@suite-native/device-authentication",
```
matejkriz: nitpick: I think authorization is more appropriate term than authentication here. Authentication is process of identity verification. Authorization is verifying permission to do something.

suite-native/device-authentication/src/utils.ts
```diff
@@ -0,0 +1,5 @@
+import { AnyAction } from '@reduxjs/toolkit';
+
+export const isPinButtonRequestCode = (action: AnyAction) =>
+    action.payload.code === 'ButtonRequest_PinEntry' ||
+    action.payload.code === 'PinMatrixRequestType_Current';
```
matejkriz: Would be nice to keep the comment why two button requests and what is the difference.

suite-native/device/src/middlewares/buttonRequestMiddleware.ts
```diff
@@ -17,8 +18,10 @@ export const prepareButtonRequestMiddleware = createMiddlewareWithExtraDeps(
             );
         }
 
-        if (action.type === UI.REQUEST_BUTTON) {
+        // Pin is not handled by buttonRequest but redux state within @suite-native/device-authentication
+        if (action.type === UI.REQUEST_BUTTON && !isPinButtonRequestCode(action)) {
```
matejkriz: Is there any reason why to exclude pin related button requests from the buttonRequest property of the device?

suite-native/module-connect-device/src/components/PinOnDevice.tsx
```diff
@@ -66,12 +58,6 @@ export const PinOnDevice = ({ deviceModel }: PinOnDeviceProps) => {
         }
     }, [hasDeviceRequestedPin, handleSuccess]);
 
-    useEffect(() => {
-        TrezorConnect.on(UI_REQUEST.CLOSE_UI_WINDOW, handleSuccess);
-
-        return () => TrezorConnect.off(UI_REQUEST.CLOSE_UI_WINDOW, handleSuccess);
-    }, [handleSuccess]);
-
```
matejkriz, line 74: this is nice simplification  👍

matejkriz (changes requested):
Connecting screen is missing while connecting TT, see the video

https://github.com/trezor/trezor-suite/assets/3729633/53cc7d41-9eb4-4471-8af1-ac05e43db984

suite-native/device-authentication/package.json
```diff
@@ -0,0 +1,16 @@
...
+    "version": "1.0.0",
+    "private": true,
+    "license": "See LICENSE.md in repo root",
+    "sideEffects": false,
+    "main": "src/index",
+    "scripts": {
+        "lint:js": "yarn g:eslint '**/*.{ts,tsx,js}'",
+        "type-check": "yarn g:tsc --build"
```
PeKne: please rebase and add `depcheck` script please 🙏. You can get inspired in any other package. It is already merged in develop.

matejkriz:
> Connecting screen is missing while connecting TT, see the video
> 
>  Screenrecorder-2024-06-19-15-09-21-142.mp4

Now I see it's not related to this PR, it's broken on develop too, I believe it's because of this condition: https://github.com/trezor/trezor-suite/blob/develop/suite-native/device/src/hooks/useHandleDeviceConnection.ts/#L60

suite-native/device-authentication/src/deviceAuthenticationSlice.ts
```diff
@@ -0,0 +1,53 @@
...
+    hasDeviceRequestedPin: false,
+};
+
+export const deviceAuthenticationSlice = createSlice({
+    name: 'deviceAuthentication',
+    initialState: deviceAuthenticationInitialState,
+    reducers: {},
+    extraReducers: builder => {
+        builder
+            .addMatcher(
```
PeKne: I think there is no need for using `addMatcher`, you could use `addCase` instead:

```suggestion
            .addCase(UI.REQUEST_PIN, ...
```

Please change it for the other actions as well.

PeKne (commented):
Just two nits. Good job, I believe this is a step in the good direction 👏

suite-native/device/src/middlewares/buttonRequestMiddleware.ts
juriczech: Hmm. I was thinking that this button request will be unused in suite-native and if we switch to this state management, it would just be a clutter. Elsewhere, it could interfere with some other logic. No hard opinion on this, but I just think it would be redundant to keep it.

suite-native/device-authentication/src/deviceAuthenticationSlice.ts
juriczech: Good catch https://github.com/trezor/trezor-suite/pull/12889/commits/74554a00bc14a59b3ef8a21be79a5b4910d3490b

suite-native/device-authentication/src/utils.ts
juriczech: https://github.com/trezor/trezor-suite/pull/12889/commits/74554a00bc14a59b3ef8a21be79a5b4910d3490b

suite-native/device-authentication/package.json
juriczech: Yeah, you're right https://github.com/trezor/trezor-suite/pull/12889/commits/eb7933f48c2ce956ad89638d90a70dd99c388628

suite-native/device/src/middlewares/buttonRequestMiddleware.ts
matejkriz: The status quo is that any REQUEST_BUTTON is added as buttonRequest to the device. It was like that before we started using buttonRequests array for PIN handling. I would prefer to not exclude them here. It's nice at least for debugging.

suite-native/device-authentication/package.json
matejkriz: rename the files and folders too, please

suite-native/device/src/middlewares/buttonRequestMiddleware.ts
juriczech: https://github.com/trezor/trezor-suite/pull/12889/commits/d12344e0a9cb6d9a3dcdb4ee3eea5c6674ceaf09

suite-native/device-authentication/src/deviceAuthenticationSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+    name: 'deviceAuthentication',
+    initialState: deviceAuthenticationInitialState,
+    reducers: {},
+    extraReducers: builder => {
+        builder
+            .addCase(UI.REQUEST_PIN, state => {
+                state.hasDeviceRequestedPin = true;
+            })
+            .addCase(UI.REQUEST_BUTTON, (state, action) => {
+                if (isPinButtonRequestCode(action)) {
```
vytick:
```suggestion
state.hasDeviceRequestedPin = isPinButtonRequestCode(action);
```

vytick: I think this would make it more readable. Otherwise great job :)

suite-native/device-authentication/package.json
juriczech: [0c96f73](https://github.com/trezor/trezor-suite/pull/12889/commits/0c96f732fc4311383f775f6fa6c3112a0ae2766a)

suite-native/device-authentication/package.json
juriczech: [d947394](https://github.com/trezor/trezor-suite/pull/12889/commits/d94739451111f6544b19592ee56289ee31411928)

suite-native/device-authentication/src/deviceAuthenticationSlice.ts
juriczech: In a follow up PR, there will be more stuff that will be related to passphrase. I've played with this but want to keep this simple as it can become really confusing otherwise. I'll revisit your suggestion when I plug passphrase into this to see whether it can be simplified.

matejkriz (approved):

matejkriz (approved):

Issue #12421, opened 2024-05-15
Passphrase: Wallet switcher enhancements


vytick:
Better prepare Wallet switcher for passphrase by showing fiat amount for each wallet, make content scrollable in case of many wallets and update visual in case of single wallet per device

[Figma here](https://www.figma.com/design/mmDnYtctfjnreWhTjLfeQH/Passphrase?node-id=2146-54343&t=IUU4Nb1aSTFhpLPl-0)

- [x] make device switcher content scrollable when there are more wallets
- [x] `Change` button animation between `⌄` and `^` icon state
<img src="https://github.com/trezor/trezor-suite/assets/2011829/6234c3aa-24a2-4f9e-9118-6c6624780dc3" width=300 />

matejkriz:
waiting for transaction cache and scrolling design

vytick:
mindnote: my discussion about this with Sheny:
https://satoshilabs.slack.com/archives/D06CTSHJ34Y/p1717008785611149

STew790:
**QA NOK**
I can scroll through the menu but I don't see the fiat. Is there any flag for this or should it just be there by default?

![image](https://github.com/trezor/trezor-suite/assets/48477147/cd28c47d-45ff-466a-b386-38939536685d)

24.6.0 814ab5d

vytick:
@STew790 I have split this issue last week to make PR smaller and fiat balance is being added in the next task: #13099

STew790:
**QA OK**
I see, otherwise it is OK then, approving.

PR #12898, opened 2024-06-14
feat(suite-native): animated device change button

vytick:
Make change button toggle animated

Part of #12421

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/a6f30b39-24d3-456d-840c-0458f15990f0

### DIFF
--- a/suite-native/device-manager/src/components/DevicesToggleButton.tsx
+++ b/suite-native/device-manager/src/components/DevicesToggleButton.tsx
@@ -1,26 +1,41 @@
-import { Button, Text } from '@suite-native/atoms';
+import Animated, { useAnimatedStyle, withTiming } from 'react-native-reanimated';
+
+import { Button, ButtonIcon, Text } from '@suite-native/atoms';
 import { Translation } from '@suite-native/intl';
 
-type DeviceButtonState = 'open' | 'closed';
-
 type DevicesToggleButtonProps = {
-    deviceButtonState: DeviceButtonState;
+    isOpened: boolean;
     onDeviceButtonTap: () => void;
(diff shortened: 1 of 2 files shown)


suite-native/device-manager/src/components/DevicesToggleButton.tsx
```diff
@@ -8,19 +11,45 @@ type DevicesToggleButtonProps = {
...
+                }),
+            },
+        ],
+    }));
+
+    useEffect(() => {
+        'worklet';
+
+        isOpen.value = deviceButtonState === 'open';
+    }, [deviceButtonState, isOpen]);
```
juriczech: Can you reuse the `deviceButtonState` instead of creating new value using `useSharedValue`? 

If not, I'd prefer to extend the functionality of the `onDeviceButtonTap` by creating and extending function which would handle this functionality. No reason to add a `useEffect` for this IMO. 

Something like this maybe:
```
const handleOnDeviceButtonTap = () => {
    onDeviceButtonTap()
    // Other things you want to add to this function and need to be in this component
}
```

You might also consider sending the style directly to this component as a prop. However I think the useEffect here is not very pretty.

PeKne: I agree with @juriczech. The `isOpen` and `useEffect` are redundant here. You can access the `deviceButtonState` directly in the `useAnimatedStyle`.

vytick: 8f6bff1956

PeKne (approved):

PR #12914, opened 2024-06-17
chore/depcheck-enabled

PeKne:
## Description
- b0fad8d81d12c0c75d07d95383ba3bc9db473a1e:
  - `yarn depcheck` command for finding unused package.json dependencies added
  - is also run as part of `yarn validate`
- 7fc3beff44e2690310a4f96f84e116b994cc2822:
  - unused dependencies removed (there were more than one hundred of them) Screenshot 2024-06-17 at 8 34 42

## Screenshots:
Screenshot 2024-06-17 at 8 26 28


packages/protobuf/package.json
```diff
@@ -35,11 +36,9 @@
     },
     "dependencies": {
         "@trezor/schema-utils": "workspace:*",
-        "long": "^4.0.0",
         "protobufjs": "7.2.6"
     },
     "devDependencies": {
-        "protobufjs-cli": "^1.1.2",
```
martykan: We need this, it's used in a shell script (`protobuf-build.sh`)

martykan (changes requested):

packages/blockchain-link-types/package.json
```diff
@@ -25,7 +26,6 @@
         "socks-proxy-agent": "6.1.1"
     },
     "devDependencies": {
-        "ripple-lib": "^1.10.1",
```
peter-sanderson, line 28: :man_dancing:

yarn.lock
```diff
@@ -508,6 +508,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/parser@npm:^7.24.4":
```
peter-sanderson: I think this shall not be here. Maybe `yarn dedupe`?

Also we shall add this to pipeline as well (if dedupe depdupes something -> fail)

yarn.lock
```diff
@@ -19048,6 +18964,39 @@ __metadata:
   languageName: node
   linkType: hard
 
+"depcheck@npm:^1.4.7":
+  version: 1.4.7
+  resolution: "depcheck@npm:1.4.7"
+  dependencies:
+    "@babel/parser": "npm:^7.23.0"
+    "@babel/traverse": "npm:^7.23.2"
+    "@vue/compiler-sfc": "npm:^3.3.4"
```
peter-sanderson: Sadly, It has a quite a lot of dependencies :(

There is probably nothing we can do about it, you can ignore my comment here.

peter-sanderson (commented):
I don't feel experienced enough to approve this, but I really like this. :clap:

Nodonisko (changes requested):
I think we should add this to CI, probably to "Other checks"

.depcheckrc.json
```diff
@@ -0,0 +1,14 @@
...
+    "ignore-patterns": ["libDev"],
+    "ignores": [
+        "babel*",
+        "karma*",
+        "@babel/*",
+        "@playwright/*",
+        "webpack*",
+        "worker-loader",
+        "node-libs-browser",
+        "rimraf"
```
szymonlesisz: i would add `tslib` as well - it's marked as "peerDependency" in connect and other libs because it was constantly removed as unused

PeKne: added in commit: [70b9074](https://github.com/trezor/trezor-suite/pull/12914/commits/70b90749b22d908b7f8860f0972b528bf2dbfdf1)

packages/protobuf/package.json
PeKne: fixed in commit: [70b9074](https://github.com/trezor/trezor-suite/pull/12914/commits/70b90749b22d908b7f8860f0972b528bf2dbfdf1)

yarn.lock
PeKne: fixed in commit [3161f27](https://github.com/trezor/trezor-suite/pull/12914/commits/3161f274b321cf362c0e2defad9c4945547e8943)

good call :) 
Screenshot 2024-06-18 at 7 38 40

martykan (approved):

Nodonisko (approved):

PR #12919, opened 2024-06-17
Docs/update bump versions pr description

karliatto:
## Description

1. Update ci/scripts/check-npm-dependencies.ts to work with TS now that `helpers` are also in TS
2. Update the description in the bump-version/connect PR to be more accurate at what we do now docs/releases/connect-bump-version.md


ci/scripts/check-npm-dependencies.ts
martykan, line 1: Usage help text is not updated (L19)
Other than that LGTM

karliatto: Good catch! Updated https://github.com/trezor/trezor-suite/pull/12919/commits/1e4a87724daefd738d034293cc9995f2909bdc20

martykan (approved):

PR #12501, opened 2024-05-21
chore(suite): token improvements vol. 1

AdamSchinzel:
## Description

First wave of improvements to tokens section:

- Show/hide tokens
- Blur urls inside token symbol and name
- Table redesign
- Including tokens subheading in receive modal

[Design](https://www.figma.com/design/o6eb4wUMjSu5xMb34423vh/Tokens?node-id=4084%3A8416&t=xXUHelGSSoAKr65m-1)

## Related Issue

Resolve #12812
Resolve #12811
Resolve #12810
Resolve #12846
Resolve #12841

PR #12923, opened 2024-06-17
Fix/tokens nitpicks

### LINKED ISSUE #12905: EVM Tokens / Send form select
Send form does not respect Tokens / Hidden / Unregosnised

Screenshot 2024-06-16 at 19 24 30

AdamSchinzel:
## Description

Resolved some nitpicks from #12501 and one issue with tokens select (done some important refactor in `getTokens` so it was better to have in the same PR).

## Related Issue

Resolves #12905

### DIFF
--- a/packages/suite/src/components/suite/FormattedCryptoAmount.tsx
+++ b/packages/suite/src/components/suite/FormattedCryptoAmount.tsx
@@ -13,7 +13,7 @@ import {
 } from '@suite-common/wallet-utils';
 import { isSignValuePositive } from '@suite-common/formatters';
 import { selectLanguage } from 'src/reducers/suite/suiteReducer';
-import { BlurUrls } from 'src/views/wallet/tokens/common/UrlBlur';
+import { BlurUrls } from 'src/views/wallet/tokens/common/BlurUrls';
 
 const Container = styled.span`
     max-width: 100%;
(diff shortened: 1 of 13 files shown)


packages/suite/src/components/suite/FormattedCryptoAmount.tsx
```diff
@@ -90,7 +90,7 @@ export const FormattedCryptoAmount = ({
         <Container className={className}>
             {!!signValue && <Sign value={signValue} />}
             <Value data-test={dataTest}>{formattedValue}</Value>
-            {formattedSymbol && <BlurUrls text={' ' + formattedSymbol} />}
+            {formattedSymbol && <BlurUrls text={'\u00A0' + formattedSymbol} />}
```
tomasklim, line 93: I like using `&nbsp;` but ok

packages/suite/src/views/wallet/send/components/Outputs/components/Amount/components/TokenSelect.tsx
```diff
@@ -49,35 +47,31 @@ export const buildTokenOptions = (
...
-    if (tokens) {
-        const unknownTokens: Option['options'] = [];
-        const hasCoinDefinitions = getNetworkFeatures(symbol).includes('coin-definitions');
-
-        tokens.forEach(token => {
-            if (new BigNumber(token?.balance || '').eq('0')) {
-                return;
-            }
+    if (accountTokens) {
+        const tokens = getTokens(accountTokens, symbol, isDebug, coinDefinitions);
```
tomasklim: does it make sense to have `isDebug` here? i dont think so

tomasklim (approved):
LGTM, just I would remove `isDebug` from token select if you agree

packages/suite/src/views/wallet/send/components/Outputs/components/Amount/components/TokenSelect.tsx
AdamSchinzel: Removed it and set to `false` by default

PR #12928, opened 2024-06-17
fix(suite-native): error on failed initStore

juriczech:
## Description
If there is any error on `initStore()`, the app doesn't display any information about it and just hangs on splash screen. This leads to troublesome debugging. Wrapping this inside a try / catch will still not work, but at least there will be error displayed to provide some information on what's wrong.

## Related Issue

Resolve

### DIFF
--- a/suite-native/state/src/StoreProvider.tsx
+++ b/suite-native/state/src/StoreProvider.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useState, ReactNode } from 'react';
 import { Provider } from 'react-redux';
 
+import * as Sentry from '@sentry/react-native';
 import { EnhancedStore } from '@reduxjs/toolkit';
 import { Persistor, persistStore } from 'redux-persist';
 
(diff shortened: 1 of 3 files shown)


suite-native/state/src/StoreProvider.tsx
```diff
@@ -18,10 +18,14 @@ export const StoreProvider = ({ children }: StoreProviderProps) => {
...
-            const freshPersistor = persistStore(freshStore);
-            setStore(freshStore);
-            setStorePersistor(freshPersistor);
+            try {
+                const freshStore = await initStore();
+                const freshPersistor = persistStore(freshStore);
+                setStore(freshStore);
+                setStorePersistor(freshPersistor);
+            } catch (error) {
+                console.error('Init store error:', error);
```
Nodonisko, line 28: Add Sentry pls, we definitely want to know if this happen in prod

juriczech: Good point. Added

suite-native/state/src/StoreProvider.tsx
```diff
@@ -18,10 +19,15 @@ export const StoreProvider = ({ children }: StoreProviderProps) => {
...
+                Sentry.captureException(new Error(error));
```
Nodonisko: I think you can just do `Sentry.captureException(error as Error);`

juriczech: It's already with correct type, my bad
https://github.com/trezor/trezor-suite/pull/12928/commits/01ec4fbe77a6845bb528fbc819a1ef5bfbbbdca6

Nodonisko (approved):

PR #12930, opened 2024-06-17
fix(suite): Replace PinButton component background values with new ones

hstastna:
## Description

Replace `PinButton` component's older margin, border and background color values with the new ones from `theme`. Use `mapElevationToBorder` or `mapElevationToBackground` helpers when applicable.

## Related Issue

This is one of the suggested "Future improvements" (medium priority).

## Screenshots:
**Before:**

https://github.com/trezor/trezor-suite/assets/13417815/2379e830-6bd9-4aab-859b-9e077bc2f71d

**After:**

https://github.com/trezor/trezor-suite/assets/13417815/45c81164-4039-40ad-bc0b-47540e0fcc03

### DIFF
--- a/packages/components/src/components/buttons/PinButton/PinButton.tsx
+++ b/packages/components/src/components/buttons/PinButton/PinButton.tsx
@@ -1,30 +1,38 @@
+import {
+    Elevation,
+    borders,
+    mapElevationToBackground,
+    mapElevationToBorder,
+    spacingsPx,
+} from '@trezor/theme';
 import { ButtonHTMLAttributes } from 'react';
 import styled from 'styled-components';
+import { useElevation } from '../../ElevationContext/ElevationContext';
 
-const Button = styled.button`
+const Button = styled.button<{ $elevation: Elevation }>`
     max-width: 100px;
     height: 60px;
     transition: all 0.2s;
     position: relative;
     cursor: pointer;
-    margin: 4px;
+    margin: ${spacingsPx.xxs};
 
     width: 100%;
 
-    border-radius: 5px;
-    border: 1px solid ${({ theme }) => theme.BG_GREY};
-    background: ${({ theme }) => theme.BG_GREY};
+    border-radius: ${borders.radii.xxs};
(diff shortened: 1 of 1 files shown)


packages/components/src/components/buttons/PinButton/PinButton.tsx
```diff
@@ -1,30 +1,37 @@
...
     height: 60px;
     transition: all 0.2s;
     position: relative;
     cursor: pointer;
-    margin: 4px;
+    margin: ${spacingsPx.xxs};
 
     width: 100%;
 
     border-radius: 5px;
```
jvaclavik: border-radius: ${borders.radii.xxs};

packages/components/src/components/buttons/PinButton/PinButton.tsx
```diff
@@ -33,18 +40,22 @@ const Button = styled.button`
         content: ' ';
         position: absolute;
         border-radius: 100%;
```
jvaclavik, line 43: border-radius: ${borders.radii.full};

peter-sanderson (approved):

packages/components/src/components/buttons/PinButton/PinButton.tsx
hstastna: That is 4px, not 5 as originally. So I wasn't changing it. Is that really what we want?

packages/components/src/components/buttons/PinButton/PinButton.tsx
hstastna: Hmm, do we really want to change 100% to 100px?

packages/components/src/components/buttons/PinButton/PinButton.tsx
jvaclavik: Yes, it's fine

packages/components/src/components/buttons/PinButton/PinButton.tsx
jvaclavik: OK, let's keep it

PR #12943, opened 2024-06-18
fix(ci): COMMIT_MESSAGE format

martykan:
## Description

Change COMMIT_MESSAGE format from %B (raw body) to %s (subject).
https://git-scm.com/docs/pretty-formats#Documentation/pretty-formats.txt-emsem

This solves the problem when the body has multiple lines, which causes a failure, eg. here: https://github.com/trezor/trezor-suite/actions/runs/9460182885/job/26058521948?pr=12817


vdovhanych (approved):
Cool :)

PR #12954, opened 2024-06-18
chore(transport): remove enumeration intent signaling

mroz22:
A follow-up after https://github.com/trezor/trezor-suite/pull/12494

enumerating transport api should never require exclusive access by a single client, so no locks should be in place.


szymonlesisz (approved):
seems reasonable

PR #12956, opened 2024-06-18
fix(suite-native): fix initial selection of portfolio tracker device

matejkriz:
## Description
On first run of the app, Wallet Switcher is empty.

This bug is only in develop, not released on production.

Portfolio device is not yet created on initial run of the app when initDevices() is called from initActions. Creating imported device before dispatching initDevices would have to be blocking and so prolonging initial loading of the app. That is why the device is selected later from the middleware.

## Screenshots:
fixing this:
Screenshot_2024-06-18-11-03-10-837_io trezor suite debug (1)

### DIFF
--- a/suite-native/device/src/middlewares/deviceMiddleware.ts
+++ b/suite-native/device/src/middlewares/deviceMiddleware.ts
@@ -12,6 +12,7 @@ import {
     selectDeviceThunk,
     selectAccountsByDeviceState,
     createDeviceInstanceThunk,
+    createImportedDeviceThunk,
 } from '@suite-common/wallet-core';
 import { FeatureFlag, selectIsFeatureFlagEnabled } from '@suite-native/feature-flags';
 import { clearAndUnlockDeviceAccessQueue } from '@suite-native/device-mutex';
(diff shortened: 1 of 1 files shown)


juriczech:
I tried to test it but there was circular call of `selectDeviceThunk`. For some reason it happens because of `isAnyOf` and I don't really understand why that it, yet. However using `thunk.match(action) || otherThunk.match(action)` seems to fix it. I pushed the fix, however feel free to drop the commit if it's not correct.

matejkriz:
> I tried to test it but there was circular call of `selectDeviceThunk`. For some reason it happens because of `isAnyOf` and I don't really understand why that it, yet. However using `thunk.match(action) || otherThunk.match(action)` seems to fix it. I pushed the fix, however feel free to drop the commit if it's not correct.

It was correct, but I fixed that usage of isAnyOf - it wasn't called as a function in the initial commit: `isAnyOf(someAction) vs isAnyOf(someAction)(action)`

juriczech (approved):

Nodonisko (approved):

PR #12961, opened 2024-06-19
fix(connect): don't save state with legacy passphrase

martykan:
## Description

As @szymonlesisz alerted me, T1 with firmware <1.9.0 saves the passphrase in internal state. 
We don't want that to end up in plaintext in localStorage, so this adds a check for legacy T1 before saving.


karliatto (approved):
Looks legit. 
I cannot fully test it since I do not have a device with lower FW than "1.9.0".

PR #12963, opened 2024-06-19
chore: refactor transactions fetching

Nodonisko:
This PR refactor some parts of transactions fetching to make it more reliable, efficient and provide more detailed information about loading/errors per account.

## Related Issue

Resolve

### DIFF
--- a/suite-common/wallet-core/src/transactions/transactionsThunks.ts
+++ b/suite-common/wallet-core/src/transactions/transactionsThunks.ts
@@ -1,4 +1,4 @@
-import { createThunk } from '@suite-common/redux-utils';
+import { createSingleInstanceThunk, createThunk } from '@suite-common/redux-utils';
 import {
     Account,
     AccountKey,
@@ -10,7 +10,6 @@ import {
     enhanceTransaction,
     findAccountsByAddress,
     findTransactions,
-    getAccountTransactions,
     getPendingAccount,
     getRbfParams,
     isTrezorConnectBackendType,
@@ -18,14 +17,19 @@ import {
     tryGetAccountIdentity,
 } from '@suite-common/wallet-utils';
 import { blockbookUtils } from '@trezor/blockchain-link-utils';
-import TrezorConnect from '@trezor/connect';
+import TrezorConnect, { AccountInfo } from '@trezor/connect';
+import { getTxsPerPage } from '@suite-common/suite-utils';
 
 import { accountsActions } from '../accounts/accountsActions';
 import { selectAccountByKey, selectAccounts } from '../accounts/accountsReducer';
 import { selectBlockchainHeightBySymbol } from '../blockchain/blockchainReducer';
 import { selectSendSignedTx } from '../send/sendFormReducer';
 import { TRANSACTIONS_MODULE_PREFIX, transactionsActions } from './transactionsActions';
-import { selectTransactions } from './transactionsReducer';
+import {
+    selectAccountTransactionsWithNulls,
+    selectAreAllAccountTransactionsLoaded,
+    selectTransactions,
+} from './transactionsReducer';
 
 /**
  * Replace existing transaction in the reducer (RBF)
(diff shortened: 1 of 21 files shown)


suite-common/wallet-core/src/transactions/transactionsThunks.ts
```diff
@@ -315,32 +297,90 @@ export const fetchTransactionsThunk = createThunk(
...
+
+            if (signal.aborted) {
+                throw new Error('Aborted');
+            }
+
+            if (result === 'ALREADY_FETCHED') {
+                if (areAllTransactionsAlreadyFetched) {
+                    // If we previously fetched all transactions, we are now quite sure that we have all new transactions fetched.
+                    break;
+                } else if (!areAllTransactionsAlreadyFetched && account.backendType === 'ripple') {
```
vytick: `!areAllTransactionsAlreadyFetched` is not needed here because that is ensured by not falling into previous condition

suite-native/storage/src/transforms/walletTransforms.ts
```diff
@@ -40,12 +40,17 @@ export const walletPersistTransform = createTransform<InboundState, OutboundStat
             devicesStatesNotRemembered,
         );
 
+        const transactionFetchStatusDetail = filterObjectKeys(
+            inboundState.transactions?.fetchStatusDetail,
+            // shouldn't this be AccountKey instead of deviceState? Ask Matej
```
vytick: what did he said? :)

suite-common/wallet-core/src/transactions/transactionsThunks.ts
```diff
@@ -226,60 +230,43 @@ export const addFakePendingCardanoTxThunk = createThunk(
...
-                });
-                await promise;
-            }
-
-            return;
-        }
+        const txsForPage = transactions.slice(startIndex, stopIndex).filter(tx => !!tx.txid); // filter out "empty" values
+        // always fetch first page because there might be new transactions
+        const isFirstPage = page === 1;
+        const isPageAlreadyFetched = txsForPage.length === perPage;
```
vytick, line 266: if there was only 1 txn for this account and perPage is > 1, than this would be false even if that txn was fetched already.

suite-common/wallet-core/src/transactions/transactionsThunks.ts
```diff
@@ -226,60 +230,43 @@ export const addFakePendingCardanoTxThunk = createThunk(
...
-            page: number;
-            perPage: number;
-            noLoading?: boolean;
-            recursive?: boolean;
-        },
-        { dispatch, getState, signal },
+        { accountKey, page, perPage, forceRefetch }: FetchTransactionsPageThunkParams,
+        { dispatch, getState },
     ) => {
+        // console.log('fetchTransactionsPageThunk', accountKey, page, perPage, forceRefetch);
```
vytick, line 251: obsolete comment

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
```diff
@@ -104,49 +113,53 @@ const renderSectionHeader = ({ section: { monthKey } }: RenderSectionHeaderParam
...
-            await dispatch(fetchAndUpdateAccountThunk({ accountKey }));
+            await Promise.allSettled([
+                dispatch(fetchAndUpdateAccountThunk({ accountKey })),
+                fetchTransactions(),
+            ]);
         } catch (e) {
             // Do nothing
         }
         // It's usually too fast so loading indicator only flashes for a moment, which is not nice
         setTimeout(() => setIsRefreshing(false), 1500);
```
vytick, line 161: can you handle timeout in destructor, please?

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
```diff
@@ -104,49 +113,53 @@ const renderSectionHeader = ({ section: { monthKey } }: RenderSectionHeaderParam
...
     const handleOnRefresh = useCallback(async () => {
         try {
             setIsRefreshing(true);
-            await dispatch(fetchAndUpdateAccountThunk({ accountKey }));
+            await Promise.allSettled([
+                dispatch(fetchAndUpdateAccountThunk({ accountKey })),
+                fetchTransactions(),
+            ]);
         } catch (e) {
             // Do nothing
```
vytick, line 158: do we want to send this error to sentry?

suite-native/storage/src/transforms/walletTransforms.ts
Nodonisko: @matejkriz check this pls

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
Nodonisko: I don't think we want, this could happen quite often because of network issues, BE down etc. nothing we are interested in.

suite-common/wallet-core/src/transactions/transactionsThunks.ts
Nodonisko: Good catch, but I still thinking about keeping it because it's somehow easier to understand condition then.

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
Nodonisko: I think it's not necessary here, it should not produce any incosistent state and it will be too much work to handle it :D

suite-common/wallet-core/src/transactions/transactionsThunks.ts
Nodonisko: Good catch but not sure how to handle it better (if needed). It will be issue only on first page (which is refeteched always anyway) and on the last page.

suite-common/wallet-core/src/transactions/transactionsThunks.ts
vytick: I actually think it makes it harder to read. Or at least it made it harder for me :) . otherwise it is mostly comments and no complex logic

Nodonisko: Ok I will remove it then.

suite-common/wallet-core/src/transactions/transactionsThunks.ts
vytick: Hmm, for the first page only if it is a last page as well. You are right, this is probably unnecessary to change, as it will happen only once per account max.

vytick (approved):
LGTM, txns load as expected, nice job 🚀  

I'll leave up to you whether you want to wait for more people to review. Approving

Issue #12937, opened 2024-06-18
Tooltip arrow is incorrectly positioned


AdamSchinzel:
**Describe the bug**
We recently introduces tooltip with array by default and now the arrow position is broken everywhere.

**How to reproduce**

1. Go to dashboard
1. Hover on portfolio title
1. See the arrow position

**Expected behavior**
Arrow is position correctly or it's not there at all.

**Screenshots**
![Screenshot 2024-06-18 at 10 38 01 AM](https://github.com/trezor/trezor-suite/assets/66002635/06f58e67-cd13-4976-aed4-b95552d96f5d)

**Additional context**
For now we can just bring back `hasArrow = false` in `Tooltip` component.

AdamSchinzel:
Original issue here - #12860

PR #12964, opened 2024-06-19
chore(suite): remove arrow from all tooltips

AdamSchinzel:
## Description

Removed arrow from tooltips since it renders incorrectly.

## Related Issue

Related #12937

## Screenshots:
Screenshot 2024-06-19 at 11 39 50 AM

### DIFF
--- a/packages/suite/src/views/suite/SwitchDevice/DeviceItem/EjectButton.tsx
+++ b/packages/suite/src/views/suite/SwitchDevice/DeviceItem/EjectButton.tsx
@@ -31,7 +31,7 @@ export const EjectButton = ({ setContentType, dataTest }: EjectButtonProps) => {
 
     return (
         <EjectContainer $elevation={elevation}>
-            <Tooltip hasArrow cursor="pointer" content={<Translation id="TR_EJECT_HEADING" />}>
+            <Tooltip cursor="pointer" content={<Translation id="TR_EJECT_HEADING" />}>
                 <Icon
                     data-test={`${dataTest}/eject-button`}
                     icon="EJECT"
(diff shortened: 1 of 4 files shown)


peter-sanderson (approved):

peter-sanderson:
Thanks for the fix.

packages/suite/src/views/suite/SwitchDevice/DeviceItem/EjectButton.tsx
```diff
@@ -31,7 +31,7 @@ export const EjectButton = ({ setContentType, dataTest }: EjectButtonProps) => {
 
     return (
         <EjectContainer $elevation={elevation}>
-            <Tooltip hasArrow cursor="pointer" content={<Translation id="TR_EJECT_HEADING" />}>
```
peter-sanderson, line 34: Not sure if it shall be removed from all places where it is used explicitly. 

If its a place where you saw the render is ugly, its fine, but you just removed it everywhere I think somewhere it was intentionally used. 

Definitely you should check the visual of those places if it really was broken

AdamSchinzel: True. I just checked it briefly like here where it was definitely good to remove it.
Screenshot 2024-06-19 at 11 46 06 AM

Since it renders every time to the same place I think It can't be correct anywhere.

jvaclavik (approved):

packages/suite/src/views/suite/SwitchDevice/DeviceItem/EjectButton.tsx
jvaclavik: Doesn't matter here, it's going to be removed soon

Issue #12541, opened 2024-05-23
Refactor Device Switcher animation


vytick:
Refactor device switcher open/close animation in style of accordion https://github.com/trezor/trezor-suite/pull/12534 to make code more sustainable

vytick:
I've briefly tried to refactor and still run into issue with measuring the component height. It still kept returning 0 for height.

matejkriz:
Now it's not responsive at all so different font size breaks it 
<img width="349" alt="image" src="https://github.com/user-attachments/assets/1bae6c9f-571f-47f5-9f06-7171df9a6752">

matejkriz:
Wasn’t this resolved by https://github.com/trezor/trezor-suite/pull/20561 @BrantalikP ?

BrantalikP:
yep, that one has been fixed along https://github.com/trezor/trezor-suite/pull/20561

BrantalikP:
I briefly recoll that changing the wrapper from `Fragment` to `Box` did the [trick](https://github.com/trezor/trezor-suite/pull/20561/files#diff-1f193c62cbc357f695fb735d13881a111b38ee6332bd688c199dbdfd115c4169R41)

PR #12967, opened 2024-06-19
feat(suite-native): device switcher can scroll

vytick:
When user has too many wallets or devices, the content of device switcher detail starts scrolling

## Related Issue

Part of #12421 

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/681b7b1e-ec2e-41c6-aa6b-b072f56acef0

### DIFF
--- a/suite-native/device-manager/src/components/DeviceList.tsx
+++ b/suite-native/device-manager/src/components/DeviceList.tsx
@@ -1,12 +1,6 @@
 import { useSelector } from 'react-redux';
-import Animated, {
-    runOnJS,
-    useAnimatedStyle,
-    useSharedValue,
-    withDelay,
-    withTiming,
-} from 'react-native-reanimated';
-import { useEffect, useState } from 'react';
+import Animated, { useAnimatedStyle, useSharedValue, withTiming } from 'react-native-reanimated';
+import { View } from 'react-native';
 
 import { useNavigation } from '@react-navigation/native';
 
@@ -47,6 +41,9 @@ type ConnectButtonProps = {
     onPress: () => void;
 };
 
+const ANIMATION_DURATION = 300;
+
+const DEVICE_LIST_TOP_MARGIN = 300; // This is so the bg is consistent even if the device list is dragged down while scrolling
 const ITEM_HEIGHT = 66;
 const BUTTON_HEIGHT = 48;
 const BUTTON_PADDING_TOP = 8;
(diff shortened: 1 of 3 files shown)


PeKne:
I just tested that. The animation looks very smooth good job 👏. But maybe the duration is too high? :D 

https://github.com/trezor/trezor-suite/assets/26143964/b98cb9d5-690f-4c06-b7ef-beaabed9bd45

vytick:
> I just tested that. The animation looks very smooth good job 👏. But maybe the duration is too high? :D
> 
🙈 9548b4bd82

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -122,81 +121,69 @@ export const DeviceList = ({ isVisible, onSelectDevice }: DeviceListProps) => {
...
+        const h =
+            MANAGER_MODAL_BOTTOM_RADIUS + //top margin for radius
+            PADDING_TOP +
+            otherDevicesHeight + // all other device
+            separatorHeight + // separator if there is one
+            buttonHeight +
+            PADDING_BOTTOM -
+            MANAGER_MODAL_BOTTOM_RADIUS;
+
+        height.value = isVisible ? h : 0;
```
PeKne, line 150: I know it was already like this before creating this PR. But When I looked at it now again, I realized it should be better to use `useDerivedValue`(https://www.reanimated2.com/docs/core/useDerivedValue) for both `height` and `opacity`. Then there would be no need for this effect and the animation would stay within the reanimated API borders.

IMO it will work just the same, but the readability will improve.

suite-native/device-manager/src/components/DeviceList.tsx
```diff
@@ -122,81 +121,69 @@ export const DeviceList = ({ isVisible, onSelectDevice }: DeviceListProps) => {
...
-        isDiscoveryActive,
+        opacity,
+        utils.borders.widths.small,
     ]);
 
     const listAnimatedStyle = useAnimatedStyle(
         () => ({
-            opacity: opacity.value,
-            height: height.value,
+            opacity: withTiming(opacity.value, { duration: ANIMATION_DURATION }),
```
PeKne, line 154: Please move the withTiming to the place where is the value set. So to the `useDerivedValue` or the `useEffect`, depends how you will resolve my previous comment.

suite-native/device-manager/src/components/DeviceManagerModal.tsx
```diff
@@ -28,6 +28,7 @@ const deviceManagerModalWrapperStyle = prepareNativeStyle(utils => ({
     backgroundColor: utils.colors.backgroundSurfaceElevation0,
     borderBottomLeftRadius: MANAGER_MODAL_BOTTOM_RADIUS,
     borderBottomRightRadius: MANAGER_MODAL_BOTTOM_RADIUS,
+    maxHeight: '80%',
```
PeKne: why 80%? clarify this with a comment please.

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -63,6 +76,10 @@ export const DeviceManagerContent = () => {
         return null;
     }
 
+    // offset and the height of the content header
+    const scrollViewTopOffset = insets.top + utils.spacings.large + 86;
```
PeKne: `utils.spacings.large + 86` = `100`. Please create some const at the top of the file with descriptive name so the comment is not needed. 

e.g. const HEADER_HEIGHT = 100, or something like this.

suite-native/device-manager/src/components/DeviceList.tsx
vytick: based on the docs `useDerivedValue` should be used to derive sharedValue from other sharedValues. That is not what is being done here. Here I assign value based on other stuff

vytick: Anyway. I have tried to use this and it works

suite-native/device-manager/src/components/DeviceManagerContent.tsx
vytick: 10e7175793

suite-native/device-manager/src/components/DeviceManagerModal.tsx
vytick: 10e7175793

suite-native/device-manager/src/components/DeviceList.tsx
vytick: 10e7175793

suite-native/device-manager/src/components/DeviceList.tsx
vytick: 10e7175793

PeKne: Probably I did not make myself clear. The `useDerivedValue` is a hook that returns the shared value out of it. so iMO it would make sense create two derived values, each for `height` and `opacity`. 

```
const opacity = useDerivedValue(()=>{
  return  withTiming(isVisible ? 1 : 0, { duration: ANIMATION_DURATION });
});

const height = useDerivedValue(() => {
    const otherDevicesHeight = notSelectedInstancelessDevices.length * ITEM_HEIGHT;

        const separatorHeight =
            hasUnselectedDevices && !isDiscoveryActive
                ? 2 * SEPARATOR_VERTICAL_PADDING + SEPARATOR_HEIGHT
                : 0;
        const buttonHeight = isDiscoveryActive ? 0 : BUTTON_PADDING_TOP + BUTTON_HEIGHT;
        const h =
            MANAGER_MODAL_BOTTOM_RADIUS + //top margin for radius
            PADDING_TOP +
            otherDevicesHeight + // all other device
            separatorHeight + // separator if there is one
            buttonHeight +
            PADDING_BOTTOM -
            MANAGER_MODAL_BOTTOM_RADIUS;

     return withTiming(isVisible ? h : 0, { duration: ANIMATION_DURATION });

});
```

Like this, you don't need to set both values in a single useEffect. Hmu on Slack if you have any further questions or concers.

vytick: As we agreed on the call, I will move the calculation into useAnimatedStyle, which will make more readable and the end game is to get rid of it in #12541

vytick: https://github.com/trezor/trezor-suite/commit/bd4de936f5cb1388b9aaa6049db16f4d6c78cc1c

PeKne (approved):

PR #12240, opened 2024-05-03
native send flow: review & send

PeKne:
## Description
- 07f1426e0bf1812995a3fceb06dcac8dbd949780: reusable transaction review utils and types moved from desktop package to suite-common
- a9ac15c2a1c5b45df23be1f1e6132b84c253ed6a: send transaction on-device review and send added

## Related Issue

Closes #10867 
Closes #10868 

## Screenshots:

https://github.com/trezor/trezor-suite/assets/26143964/0c837f81-c0c8-4267-ad4f-5c4490882b4c

PR #12968, opened 2024-06-19
Feat/native/review and send

### LINKED ISSUE #10867: Send coins - inputs review screen
## Description
After filling an input form, user should be able to validate these values on the device screen.

## Acceptance criteria
- [x] After submiting a valid send form user is redirected to another screen where he can review filled inputs
- [x] After user goes through the device process, the screen renders simple message saying that verification was successful

### LINKED ISSUE #10868: Send coins - send transaction
## Description
After validating the send values by device, user should be able to create transaction and send it.

## Acceptance criteria
- [ ] after the device values verification, user is able to click `Send` button
- [ ] button click triggers on device send flow
- [ ] after successful device send flow, the user is in app notified that send was successful and redirected back to the account screen

PeKne:
Enables basic send form submission of BTC testnets (TEST and REGTEST).  While reviewing this, please focus on the design of the whole flow and code architecture. Reviewing UI or some styling/translation nits is not necessary at this stage, because this implementation does not reflect the final design. The purpose of this PR is solely to prepare the codebase so it can send the transaction. 

Also there are some steps of the send flow that are not implemented yet but will be revisited in follow up PRs (eg. fee selection, multiple recipients...)

## Description
The PR is divided into two commits:
- [5d693e5](https://github.com/trezor/trezor-suite/pull/12968/commits/5d693e5119f6ee36bdaf3184523b6be4766d113a)- **moves transaction outputs utils from `packages` to `suite-common` to be reused in native**. There are no new things implemented only moving stuff around.
- [f6df275](https://github.com/trezor/trezor-suite/pull/12968/commits/f6df275bfd83f9f9696708e44d2bc787b6780d67) -  **transaction review & send**
  - Review screen reflecting on-device confirmation steps added
  - basic error handling implemented
  - on successful on-device verification, the user can click the `send transaction` button and broadcast the transaction to the blockchain
  - note: sorry for making this PR so big, but most of it is just UI so I hope it's not that big a deal. In case you need a better explanation of this PR, ping me on Slack
 
## Related Issues

Resolve #10867
Resolve #10868

## Screenshots:

https://github.com/trezor/trezor-suite/assets/26143964/6757c098-fa24-406f-b7dd-1a0df8acb83a


packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
```diff
@@ -60,25 +63,28 @@ export const TransactionReviewModalContent = ({
     const { account } = selectedAccount;
     const { precomposedTx, serializedTx } = txInfoState;
 
+    const precomposedForm: FormState | StakeFormState | undefined = isStakeState(txInfoState)
+        ? txInfoState.precomposedForm
+        : txInfoState.drafts[account?.key ?? ''];
+
+    const decreaseOutputId = precomposedTx?.useNativeRbf
+        ? precomposedForm?.setMaxOutputId
+        : undefined;
```
juriczech: Curious: Why was this moved above the check of non-null account? Now instead of `account.key`, you are doing `account?.key ?? ""` with no added benefit? If it's needed for the selector, there should be a selector for this for the logic to be hidden there. It's a small thing, but I think you should create a selector for this to hide this logic there.

suite-common/wallet-types/src/transaction.ts
```diff
@@ -243,3 +243,41 @@ export type TransactionFiatRateUpdatePayload = {
 export type TransactionType = Pick<WalletAccountTransaction, 'type'>['type'];
 
 export type ExportFileType = 'csv' | 'pdf' | 'json';
+
+export type ReviewOutput =
```
juriczech, line 247: I would move this to send package as explained in earlier comment

suite-common/wallet-utils/src/reviewTransactionUtils.ts
```diff
@@ -0,0 +1,333 @@
...
+    FormState,
+    GeneralPrecomposedTransactionFinal,
+    ReviewOutputState,
+} from '@suite-common/wallet-types';
+import { StakeFormState, ReviewOutput } from '@suite-common/wallet-types';
+import { TrezorDevice } from '@suite-common/suite-types';
+
+import { getShortFingerprint, isCardanoTx } from './cardanoUtils';
+
+export const getOutputState = (index: number, buttonRequestsCount: number): ReviewOutputState => {
```
juriczech: Similar here: It should be in send package and I'd also consider better naming for these functions. `getOutputState` in regards to wallet-utils is not really explanatory enough.

packages/suite/src/types/wallet/transaction.ts
```diff
@@ -1,39 +1,3 @@
...
-      }
-    | {
-          type: 'reduce-output';
-          label: string;
-          value: string;
-          value2: string;
-          token?: undefined;
-      };
-
 export type {
```
juriczech: IMO this file no longer serves it's purpose. You've stripped the only reason there was to leave this file as is and now it's just re-exporting other types. Wouldn't it be better to just move the whole file and rename it? Git might treat it as just moving + renaming since changes would be minimal (removing re-export) and it could still be shared code.

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -923,3 +924,37 @@ export const selectInstacelessUnselectedDevices = memoize((state: DeviceRootStat
 
     return deviceUtils.getSortedDevicesWithoutInstances(allDevices, device?.id);
 });
+
+export const selectSendFormReviewButtonRequestsCount = (
```
juriczech: This could be added to send form reducer maybe? Is there a reason why you couldn't import `DeviceRootState` or some selectors from this package in there? IMO it would be better.

packages/suite/src/types/wallet/transaction.ts
juriczech: This file no longer has any purpose. Please delete it because now it's just re-exporting types which shouldn't happen / shouldn' be needed.

suite-native/module-send/src/components/ReviewOutputItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    return (
+        <Text numberOfLines={1} adjustsFontSizeToFit>
+            {value}
+        </Text>
+    );
+};
+
+// TODO: adjust UI to the design when ready
```
juriczech: I know it's nitpick, but please let's not add this type of comments everywhere. This really is redundant. If it's in some design and issue, it will be done and without linking PR it's just clutter.

suite-native/module-send/src/components/ReviewOutputItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const ReviewOutputItem = ({
+    reviewOutput,
+    networkSymbol,
+}: {
+    networkSymbol: NetworkSymbol;
+    reviewOutput: StatefulReviewOutput;
+    status: ReviewOutputState;
+}) => {
```
juriczech: `ReviewOutputItemProps` at the top of the file please

suite-native/module-send/src/components/ReviewOutputItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+    switch (state) {
+        case 'active':
+            return 'warning';
+        case 'success':
+            return 'success';
+        default:
+            return 'error';
+    }
+};
+
```
juriczech:
```
const alertBoxVariantMap = {
    active: "warning",
    success: "success",
} as const satisfies Record<ReviewOutputState, AlertBoxVariant>
```

Then use in component could look like `alertBoxVariantMap[output.state] ?? 'error'`

This seems better and less verbose to me.

suite-native/module-send/src/components/SendOutputFields.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const amountTransformer = (value: string) =>
+    value
+        .replace(/[^0-9\.]/g, '') // remove all non-numeric characters
+        .replace(/^\./g, '') // remove '.' symbol if it is not preceded by number
+        .replace(/(?<=\..*)\./g, '') // keep only first appearance of the '.' symbol
+        .replace(/(?<=^0+)0/g, ''); // remove all leading zeros except the first one
+
+export const SendOutputFields = () => {
+    const { control } = useContext(FormContext);
```
juriczech: Why aren't you using `useFormContext`?

suite-native/module-send/src/components/SendOutputsForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { SendOutputFields } from './SendOutputFields';
+
+type SendFormProps = {
+    accountKey: AccountKey;
+};
+
+// TODO: this data structure will be revisited in a follow up PR
+const constructFormDraft = ({ outputs }: SendOutputsFormValues): FormState => ({
+    outputs:
+        outputs?.map(({ address, amount }) => ({
```
juriczech: Isn't `outputs.map()` sufficient since it's array and if it's empty array, you don't need to do that?

suite-native/module-send/src/components/SendOutputsForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const {
+        handleSubmit,
+        control,
+        getValues,
+        formState: { isValid, isSubmitting },
+    } = form;
+    const watchedFormValues = useWatch({ control });
+
+    const storeFormDraftIfValid = useCallback(async () => {
+        await dispatch(
```
juriczech: Redundant async / await?

suite-native/module-send/src/components/SendOutputsForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+        },
+    });
+
+    const {
+        handleSubmit,
+        control,
+        getValues,
+        formState: { isValid, isSubmitting },
+    } = form;
+    const watchedFormValues = useWatch({ control });
```
juriczech, line 94: No need to pass `control`, since it's optional when using form provider.

suite-native/module-send/src/components/SendOutputsForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+        );
+    }, [accountKey, dispatch, getValues]);
+
+    useEffect(() => {
+        if (watchedFormValues && isValid) debounce(storeFormDraftIfValid);
+    }, [isValid, storeFormDraftIfValid, watchedFormValues, debounce]);
+
+    const handleNavigateToReviewScreen = handleSubmit(async () => {
+        // Ignore another submit button clicks until is the submit not finished.
+        if (isSubmitting) return;
```
juriczech: You should disable the button if `isSubmitting` is true.

suite-native/module-send/src/components/SendOutputsForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+        // TODO: navigate to SendFeeScreen instead, when ready
+        // issue: https://github.com/trezor/trezor-suite/issues/10871
+        navigation.navigate(SendStackRoutes.SendReview, { accountKey });
+        const response = await dispatch(
+            onDeviceTransactionReviewThunk({
+                accountKey,
+                formState: constructFormDraft(getValues()),
```
juriczech: You don't need to `getValues()`, `handleSubmit` passes these as a parameter to the anonymous function so they can be used here.

suite-native/module-send/src/components/SendOutputsForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+            }),
+        );
+
+        if (isRejected(response)) {
+            // TODO: display error message based on the error code
+            showToast({ variant: 'error', message: 'Something went wrong', icon: 'closeCircle' });
+
+            navigation.navigate(SendStackRoutes.SendAccounts);
```
juriczech, line 124: I believe this should be `navigation.goBack`. We should count with send module being opened from multiple places and if we navigate back, we might want to go back where user came from instead of specific screen (we do that in a lot of other places)

suite-native/module-send/src/screens/SendOutputsScreen.tsx
```diff
@@ -45,7 +45,7 @@ export const SendFormScreen = ({
                         />
                     </HStack>
                 </VStack>
-                <SendForm accountKey={accountKey} />
+                <SendOutputsForm accountKey={accountKey} />
```
juriczech, line 57: I believe it should either be as it was. We're not sending outputs per say, we're just sending transaction which has outputs, no? This name is pretty confusing IMO

suite-native/module-send/src/screens/SendReviewScreen.tsx
```diff
@@ -0,0 +1,115 @@
...
+        if (isRejected(sendResponse)) {
+            // TODO: error handling
+        }
+
+        showToast({ variant: 'success', message: 'Transaction sent', icon: 'check' });
+
+        navigation.dispatch(navigateToAccountDetail({ accountKey }));
+    };
+
+    const handleGoBack = () => {
```
juriczech: `handleCancel` might be better name? Is there an actual redirect happening?

suite-native/module-send/src/screens/SendReviewScreen.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const signedTransaction = useSelector(selectSendSignedTx);
+    const account = useSelector((state: AccountsRootState) =>
+        selectAccountByKey(state, accountKey),
+    );
+
+    if (!account) return;
+
+    const handleSendTransaction = async () => {
+        // TODO: error handling
+        if (!signedTransaction) return;
```
juriczech: I believe this should be in outer scope (similar to account condition) to return null (or better yet - ErrorMessage from atoms) if no transaction was found by selector? This function should be called just when there is a transaction to be sent IMO and we just handle reject from sending it.

suite-native/navigation/src/components/GoBackIcon.tsx
```diff
@@ -5,10 +5,11 @@ import { IconButton } from '@suite-native/atoms';
 import { CloseActionType } from '../navigators';
 
 type GoBackIconProps = {
+    onPress?: () => void;
```
juriczech: I would advise against this if you're sending onPress which doesn't go back, then it shouldn't be in GoBack component :D

juriczech (changes requested):
Overall good job! Just some things I'd rather get right from the beginning rather than creating code that will be refactored. I'm sorry if this sounds the wrong way, but IMO in the future, we shouldn't create PR this size with saying most of it will change. If I'm reviewing some code, I'd like to get it right to build on top of it or don't do it in a PR like this. Becuase then it might never be changed and if we don't review it then it's a recipe for unmaintainability. However it's just some things for discussion so let's see if they're important or not. Feel free to huddle me if you want to further discuss them.

juriczech:
This pull request can be closed as redundant now? https://github.com/trezor/trezor-suite/pull/12240

packages/suite/src/types/wallet/transaction.ts
PeKne: duplicate comment, closing.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
PeKne: The thing is that this cannot be moved to selector in context of this file because of the way how is this component designed. The whole state of this module is served a as a component prop (yes, you heard me right 😩) and it can be either `SendState` or `StakeState`. These states are very similar, but not the same 😤. So if I would create a selector in `SendReducer` that we use also in mobile, it would broke the desktop send form for staking. 

I agree that it is not ideal, but I'm sure I don't want to refactor this as part of this PR. Because it would be huge refactor across many packages of desktop suite.

packages/suite/src/types/wallet/transaction.ts
PeKne: fixed in a883ca4bae. I already remaped the imports in the original PR, i just forgot to remove this leftover 🙈.

suite-common/wallet-core/src/device/deviceReducer.ts
PeKne: moved a883ca4bae

suite-native/module-send/src/components/ReviewOutputItem.tsx
PeKne: I removed it. But you would be suprised that I got some comment in the older PR, asking me if it is the final solution :D. For this reason I included this comment.

suite-common/wallet-utils/src/reviewTransactionUtils.ts
PeKne: Named changed for better context. a883ca4bae

suite-native/module-send/src/components/ReviewOutputItem.tsx
PeKne: fixed 3fd9ef1399

suite-native/module-send/src/components/ReviewOutputItem.tsx
PeKne: fixed 3fd9ef1399

suite-native/module-send/src/components/SendOutputFields.tsx
PeKne: I wasn't aware we have it 👍 , fixed: 3fd9ef1399

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: fixed by making the otputs required in yup validation schema: 3fd9ef1399

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: This hook is not called inside of the form context (the `Form` is defined in this file), For this reason, the `control` argument is needed.

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: fixed 3fd9ef1399

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: good point, fixed 3fd9ef1399

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: The information from design is that it should behave this way. If user clicks this button, he goes back to the send account selection.

suite-native/module-send/src/screens/SendOutputsScreen.tsx
PeKne: It has to be like this. Because in the next PR will be added `SendFeesForm`.

suite-native/module-send/src/screens/SendReviewScreen.tsx
PeKne: renamed 3fd9ef1399

suite-native/navigation/src/components/GoBackIcon.tsx
PeKne: You are right. The GoBackIcon changes reverted: 3fd9ef1399

suite-native/module-send/src/screens/SendReviewScreen.tsx
PeKne: Most of the lifecycle of this component, there should be no `signedTransaction`. The component first show the review steps and only if user confirms all of them, the signed transaction is created. For this reason, it can not be handled in out of this function.

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: good catch. Fixed by: 3b8cbb289f. I had to change typing of `useAsyncDebounce` to allow to send simple synchronous function as a argument.

juriczech: I woul also rename the hook to `useDebounce` as it doesn't only accept async functions now. Or maybe separate into 2 different functions. However I myself prefer the first option.

suite-native/navigation/src/components/GoBackIcon.tsx
juriczech: But you're still not doing any navigation. Am I missing something?

suite-native/module-send/src/screens/SendReviewScreen.tsx
juriczech: But the button that triggers this function already checks that `signedTransaction` is present. So I think you can just remove this.

suite-native/module-send/src/components/SendOutputsForm.tsx
juriczech: Correct me if I'm wrong, but you're only navigating to this screen from SendAccounts - meaning navigation go back would do the same thing. 

Benefit of using `navigation.goBack` is that it would also work if we add new route from which you can navigate to send form. If user clicks back in that case, we would want him to be redirected back, not to send accounts.

suite-common/wallet-utils/src/reviewTransactionUtils.ts
```diff
@@ -0,0 +1,336 @@
...
+};
+
+type ConstructOutputsParams = {
+    precomposedTx: GeneralPrecomposedTransactionFinal;
+    decreaseOutputId: number | undefined;
+    account: Account;
+    precomposedForm: FormState | StakeFormState;
+};
+
+const constructOldFlow = ({
```
juriczech, line 88: Can you please add comments to this old vs new flow? What does that even mean? 😆 Rename these please, if possible.

Nodonisko (commented):
Seems that @juriczech already pointed out most of issues, only one thing that I have in mind is that we should call `updateFeeInfoThunk` in the begging of send flow because on mobile we don't update fee with every block like on desktop.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
juriczech: I thought that's what the `isStakeState` is for and you could use it in the selector. Could you elaborate?

suite-native/module-send/src/screens/SendReviewScreen.tsx
PeKne: Yes, it does, but Typescript does not know this. I have to assert that the value is present to make the linter happy.

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: I don't agree. The name `useAsyncDebounce` is telling you, that it makes the call "asynchronously debounce" rather than it accepts only the async functions.

suite-native/module-send/src/screens/SendReviewScreen.tsx
juriczech: Of course, I didn't realise. In that case, I'd consider putting the text and button in a new component. It will be better to have it there because: 
- you can send signedTransaction as prop that is required
- it would accept accountKey, so you will have access to account

I believe that is better and you can keep this function without redundant condition, that is already checked in JSX.

suite-native/module-send/src/components/SendOutputsForm.tsx
juriczech: I think that's not correct. The async means that the debounce returns a promise. Since it doesn't always return promise anymore, it's not correct and is just redundant. I doubt that it means "asynchronously debounce", since debounce is always like that, that's why it's called `debounce` and not `asynchronous debounce` 😄 I went a little bit far with the metaphor here, but had to prove a point because I think we're looking at this differently. 

Since this is pretty easy to rename (pretty quick if you use the rename symbol in VS code), can we do that to keep this hook name simple and to the point? 

If this was a complicated refactor, it's obviously not too important and could be ignored, but I think we spent more time on this already than if we would just rename it. 

Happy to discuss more if you want though.

suite-native/module-send/src/screens/SendReviewScreen.tsx
PeKne: Fixed in [affdc58](https://github.com/trezor/trezor-suite/pull/12968/commits/affdc5880efe6a9efaddd99b05b742b2a753ebbb) by moving the button to the separate component.

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: fixed 96cbb1d01abbc546dc377addafef8a1deb874148

suite-common/wallet-types/src/transaction.ts
PeKne: This can not be moved to the `wallet-core` package, since these types are also needed in `wallet-utils` package.

I still review all the types that are here nad found some that should be elsewhere. Some of theme were moved to wallet-core, so even to the packages/suite (since there will be ever used only by desktop). 

Changes are in commit: [e94fba7](https://github.com/trezor/trezor-suite/pull/12968/commits/e94fba7de6abf55db1824d38b13db5ba85a33a09). I highly encourage to review this commit in isolation.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
PeKne: Special selector created here: [05f5b4b](https://github.com/trezor/trezor-suite/pull/12968/commits/05f5b4bafada99e14c5cecad39deaa76eab5f064)

Not much optimistic that this solved anything, but the code looks a bit better at least.

suite-native/module-send/src/components/SendOutputsForm.tsx
PeKne: This will be revisited in the error handling follow-up PR. There will be no direct navigation in this callback, but the error state will be set. Then the module will be able to detect this error in a useEffect and handle it according to its context (in this case it will trigger the navigation action).

Issue: https://github.com/trezor/trezor-suite/issues/13125

suite-native/navigation/src/components/GoBackIcon.tsx
PeKne: The navigation will be implemented in the useEffect, that will react to change of the send form error state.

Issue created here: https://github.com/trezor/trezor-suite/issues/13125

packages/suite/src/hooks/wallet/form/useCompose.ts
```diff
@@ -5,20 +5,18 @@ import { FeeLevel } from '@trezor/connect';
...
-    ComposeActionContext,
-    SendContextValues,
     PrecomposedTransaction,
     PrecomposedTransactionCardano,
     PrecomposedLevels,
     PrecomposedLevelsCardano,
 } from '@suite-common/wallet-types';
 import { COMPOSE_ERROR_TYPES } from '@suite-common/wallet-constants';
 import { selectSelectedAccount } from 'src/reducers/wallet/selectedAccountReducer';
+import { SendContextValues, UseSendFormState } from '../../../types/wallet/sendForm';
```
juriczech, line 19: 'src/types/wallet/sendForm' - please also update in all other usages of this commit

suite-common/wallet-types/src/transaction.ts
juriczech: Please fix the imports to `src/types/wallet/sendForm` (comment visible in commit)

packages/suite/src/hooks/wallet/form/useCompose.ts
PeKne: ouch, poorly configured auto-import... Fixed: cba8fb0741

juriczech (approved):

PeKne:
@Nodonisko the `updateFeeInfoThunk` call added ✅

PR #12969, opened 2024-06-19
feat(suite): Add cursor: not-allowed to disabled components

### LINKED ISSUE #12945: Add cursor:not-allowed to disable components

Components:
```
Input
Button
Radio
Checkbox
AutoscalingInput
Range
Select
SelectBar (?)
Switch
Textarea

ButtonGroup
IconButton
PinButton
TextButton
TooltipButton
```

hstastna:
## Description

Add `cursor: not-allowed` for cursor to look nicer to all interactive/input components in our design system, when they are disabled.

Additionally, add args in storybook for components that are missing them so one could easily change the values of selected component's props in storybook.

- Input
- Button, ButttonGroup
- Radio
- Checkbox

TODO (in followup PRs):
- [ ] AutoscalingInput
- [ ] Range
- [ ] Select
- [ ] SelectBar (?)
- [ ] Switch
- [ ] Textarea
- [ ] IconButton
- [ ] PinButton
- [ ] TextButton
- [ ] TooltipButton

## Related Issue

## Screenshots:
**Before:**
`Input` component cursor not so nice + missing some of the controls in the storybook:
input_before
`Button` component (+ missing some of the controls, same for other components):
button_before
`Radio`:
radio_before
`Checkbox:`
check_before

**After:**
`Input` component cursor nicer + more controls in the storybook:
input_after
`Button` component:
button_after
`Radio`:
radio_after
`Checkbox:`
check_after

### DIFF
--- a/packages/components/src/components/buttons/Button/Button.tsx
+++ b/packages/components/src/components/buttons/Button/Button.tsx
@@ -48,8 +48,7 @@ export const ButtonContainer = styled.button<ButtonContainerProps>`
     &:disabled {
         background: ${({ theme }) => theme.BG_GREY};
         color: ${({ theme }) => theme.textDisabled};
-        pointer-events: none;
-        cursor: default;
+        cursor: not-allowed;
     }
 
     ${withFrameProps}
@@ -142,6 +141,7 @@ export const Button = ({
             $hasIcon={!!icon || isLoading}
             $elevation={elevation}
             {...rest}
+            onClick={isDisabled ? undefined : rest?.onClick}
             {...makePropsTransient(frameProps)}
         >
             {!isLoading && icon && IconComponent}
--- a/packages/components/src/components/form/Checkbox/Checkbox.tsx
+++ b/packages/components/src/components/form/Checkbox/Checkbox.tsx
@@ -92,17 +92,22 @@ export const CheckContainer = styled.div<{ $variant: CheckboxVariant }>`
         background 0.1s,
         box-shadow 0.1s ease-out;
 
-    input:checked + && {
+    input:checked + & {
         background: ${({ theme, $variant }) => theme[variantStyles[$variant].backgroundChecked]};
         border-color: ${({ theme, $variant }) => theme[variantStyles[$variant].borderChecked]};
     }
 
-    input:disabled:not(:checked) + && {
+    input:disabled + & {
+        cursor: not-allowed;
+        pointer-events: auto;
+    }
+
+    input:disabled:not(:checked) + & {
         background: ${({ theme, $variant }) => theme[variantStyles[$variant].backgroundDisabled]};
         border-color: ${({ theme, $variant }) => theme[variantStyles[$variant].borderDisabled]};
     }
 
-    input:disabled:checked + && {
+    input:disabled:checked + & {
         background: ${({ theme, $variant }) =>
             theme[variantStyles[$variant].backgroundDisabledChecked]};
         border-color: ${({ theme, $variant }) =>
@@ -160,7 +165,7 @@ export type CheckboxProps = {
 export const Checkbox = ({
     variant = 'primary',
     isChecked,
-    isDisabled,
+    isDisabled = false,
     labelAlignment = 'right',
     onClick,
     'data-test': dataTest,
@@ -171,7 +176,10 @@ export const Checkbox = ({
     const theme = useTheme();
 
     const handleKeyUp = (event: KeyboardEvent<HTMLElement>) => {
-        if (event.code === KEYBOARD_CODE.SPACE || event.code === KEYBOARD_CODE.ENTER) {
+        if (
+            !isDisabled &&
+            (event.code === KEYBOARD_CODE.SPACE || event.code === KEYBOARD_CODE.ENTER)
+        ) {
             onClick(event);
         }
     };
@@ -184,7 +192,7 @@ export const Checkbox = ({
         <Container
             $isDisabled={isDisabled}
             $labelAlignment={labelAlignment}
-            onClick={onClick}
+            onClick={isDisabled ? undefined : onClick}
             onKeyUp={handleKeyUp}
             data-test={dataTest}
             className={className}
--- a/packages/components/src/components/form/Radio/Radio.tsx
+++ b/packages/components/src/components/form/Radio/Radio.tsx
@@ -54,7 +54,7 @@ const RadioIcon = styled(CheckContainer)`
         transition: background 0.1s;
     }
 
-    input:checked + && {
+    input:checked + & {
         background: ${({ theme }) => theme.backgroundSurfaceElevation0};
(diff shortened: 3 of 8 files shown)


packages/components/src/components/form/Checkbox/Checkbox.tsx
```diff
@@ -180,11 +188,15 @@ export const Checkbox = ({
...
+    const onClickHandler = (event: SyntheticEvent) => {
+        if (isDisabled !== true) return onClick(event);
+    };
+
     return (
         <Container
             $isDisabled={isDisabled}
             $labelAlignment={labelAlignment}
-            onClick={onClick}
+            onClick={onClickHandler}
```
jvaclavik:
```
onClick={ isDisabled ? undefined : onClick }
```


`isDisabled` should be set to `false` by default (line 168): 

image

packages/components/src/components/form/Checkbox/Checkbox.tsx
```diff
@@ -97,6 +97,11 @@ export const CheckContainer = styled.div<{ $variant: CheckboxVariant }>`
         border-color: ${({ theme, $variant }) => theme[variantStyles[$variant].borderChecked]};
     }
 
+    input:disabled + && {
```
jvaclavik: could it work with `input:disabled + &`?

In general could you try to remove these `&&` in whole file? I think they works as overriding which is not good.

packages/components/src/components/form/Checkbox/Checkbox.tsx
```diff
@@ -171,7 +176,10 @@ export const Checkbox = ({
     const theme = useTheme();
 
     const handleKeyUp = (event: KeyboardEvent<HTMLElement>) => {
-        if (event.code === KEYBOARD_CODE.SPACE || event.code === KEYBOARD_CODE.ENTER) {
+        if (
+            isDisabled !== true &&
```
jvaclavik: please use `isDisabled === false` and `isDisabled` should be set to `false` by default

packages/components/src/components/form/Radio/Radio.tsx
```diff
@@ -64,6 +64,10 @@ const RadioIcon = styled(CheckContainer)`
         }
     }
 
+    input:disabled + && {
```
jvaclavik: Also, can't we use just `input:disabled + &&`?

packages/components/src/components/form/Radio/Radio.tsx
```diff
@@ -115,14 +119,21 @@ export const Radio = ({
...
     };
 
+    const onClickHandler = (event: SyntheticEvent) => {
+        if (isDisabled !== true) return onClick(event);
+    };
+
     return (
         <Container
-            onClick={onClick}
+            onClick={onClickHandler}
```
jvaclavik: `onClick={ isDisabled ? undefined : onClick }`

packages/components/src/components/form/Checkbox/Checkbox.tsx
hstastna: I did this because in this component, there was possible that `isDisabled` could be undefined. I wasn't changing the original default values even when I noticed some inconsistencies regarding the components. Of course if we decide to have `isDisabled` set to `false` by default, it is definitely better to check against `false` here.

packages/components/src/components/form/Radio/Radio.tsx
hstastna: Hmm, I am not sure I understand..

packages/components/src/components/form/Checkbox/Checkbox.tsx
hstastna: 🆗

packages/components/src/components/form/Checkbox/Checkbox.tsx
hstastna: Yeah, as we were talking about this, I'll figure that out.

packages/components/src/components/form/Radio/Radio.tsx
jvaclavik: Sorry, I mean try to use `input:disabled + &`

packages/components/src/components/form/Checkbox/Checkbox.tsx
jvaclavik: I believe `isDisabled={undefined}` should behave like false, so let's have it more strict and use only boolean values.

packages/components/src/components/buttons/Button/Button.tsx
```diff
@@ -48,8 +48,7 @@ export const ButtonContainer = styled.button<ButtonContainerProps>`
     &:disabled {
         background: ${({ theme }) => theme.BG_GREY};
         color: ${({ theme }) => theme.textDisabled};
-        pointer-events: none;
-        cursor: default;
+        cursor: not-allowed;
```
peter-sanderson, line 51: I suggest completely not use this mechanism and rely solely on `onClick={ isDisabled ? undefined : onClick }`

Is there anything this is useful? @hstastna / @jvaclavik (Genuine question, I am oldchool grumpy dev who hates CSS :D)

hstastna: I am already about to use `onClick={ isDisabled ? undefined : onClick }` as it's definitely better as we make sure `onClick` isn't called when component disabled.
And what `cursor: not-allowed;` does is that we just can see a different, nicer cursor for disabled component when hovering on it.

packages/components/src/components/form/Checkbox/Checkbox.tsx
```diff
@@ -171,7 +176,10 @@ export const Checkbox = ({
     const theme = useTheme();
 
     const handleKeyUp = (event: KeyboardEvent<HTMLElement>) => {
-        if (event.code === KEYBOARD_CODE.SPACE || event.code === KEYBOARD_CODE.ENTER) {
+        if (
+            isDisabled === false &&
```
peter-sanderson: nit: `isDisabled === false` not needed as there is default value and it cannot be `undefined` => `!isDisabled` is ok

packages/components/src/components/form/Radio/Radio.tsx
```diff
@@ -109,20 +111,23 @@ export const Radio = ({
...
-    isDisabled,
+    isDisabled = false,
     onClick,
     'data-test': dataTest,
     children,
 }: RadioProps) => {
     const handleKeyUp = (event: KeyboardEvent<HTMLElement>) => {
-        if (event.code === KEYBOARD_CODE.SPACE || event.code === KEYBOARD_CODE.ENTER) {
+        if (
+            isDisabled === false &&
```
peter-sanderson: nit: `!isDisabled` same here

peter-sanderson (approved):

packages/components/src/components/form/Checkbox/Checkbox.tsx
hstastna: True! Thanks!

PR #12970, opened 2024-06-19
fix(suite): break lines correctly for legacy addresses

### LINKED ISSUE #12272: Difference address on Trezor One and Suite with Legacy account.
There is a difference between displayed address on Trezor One and Suite with Legacy account.

Last letter of the first line on Suite  (j) should be on the second line just like on the Trezor.

image

komret:
## Description

Trezor One has varying line length when it comes to legacy addresses. This fixes it and also adds some missing test cases for TS5.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12272

## Screenshots:
Screenshot 2024-06-19 at 13 48 17

### DIFF
--- a/packages/suite/src/components/suite/DeviceDisplay/__tests__/parseTextToPagesAndLines.test.ts
+++ b/packages/suite/src/components/suite/DeviceDisplay/__tests__/parseTextToPagesAndLines.test.ts
@@ -32,6 +32,36 @@ describe(parseTextToPagesAndLines.name, () => {
             },
         },
         {
+            it: 'Trezor One: Legacy Address',
+            input: {
+                deviceModel: DeviceModelInternal.T1B1,
+                text: '1PmVvr5DNVYJygtDT7J312qmxpa5pceu9E',
+            },
+            expected: {
+                pages: [
+                    {
+                        rows: [{ text: '1PmVvr5DNVYJygtDT' }, { text: '7J312qmxpa5pceu9E' }], // line breaks after 17 characters
+                    },
+                ],
+                hasNextIcon: false,
+            },
+        },
+        {
+            it: 'Trezor One: Legacy SegWit testnet Address',
+            input: {
+                deviceModel: DeviceModelInternal.T1B1,
+                text: '2NEPdNcUiLHLT7brh6AJaMV3AE5XG36HtsP',
+            },
+            expected: {
+                pages: [
+                    {
+                        rows: [{ text: '2NEPdNcUiLHLT7brh6' }, { text: 'AJaMV3AE5XG36HtsP' }], // line breaks after 18 characters
+                    },
+                ],
+                hasNextIcon: false,
+            },
+        },
+        {
             it: 'Trezor One: XPUB',
             input: {
                 deviceModel: DeviceModelInternal.T1B1,
@@ -91,7 +121,7 @@ describe(parseTextToPagesAndLines.name, () => {
             },
         },
 
-        // --------------- Trezor Safe3 ---------------
+        // --------------- Trezor Safe 3 ---------------
 
         // No case for 'Trezor Safe 3: Address', it is handled by different logic, as chunks are used
         {
@@ -121,10 +151,6 @@ describe(parseTextToPagesAndLines.name, () => {
                 hasNextIcon: true,
             },
         },
-
-        // --------------- Trezor Safe3 ---------------
-
-        // No case for 'Trezor Safe 3: Address', it is handled by different logic, as chunks are used
         {
             it: 'Trezor Safe 3 XPUB Taproot',
             input: {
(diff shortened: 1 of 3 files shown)


packages/suite/src/components/suite/DeviceDisplay/__tests__/parseTextToPagesAndLines.test.ts
```diff
@@ -158,6 +184,62 @@ describe(parseTextToPagesAndLines.name, () => {
                 hasNextIcon: true,
             },
         },
+
+        // --------------- Trezor Safe 5 ---------------
+
+        // No case for 'Trezor Safe 5: Address', it is handled by different logic, as chunks are used
+        {
+            it: 'Trezor Safe 5 XPUB',
```
peter-sanderson, line 192: :heart:

packages/suite/src/components/suite/DeviceDisplay/parseTextToPagesAndLines.ts
```diff
@@ -52,10 +53,26 @@ export const parseTextToPagesAndLines = ({
...
     text,
 }: ParseTextToLinesParams): ParseTextToLinesResult => {
+    const getCharsPerRow = (): number => {
+        // T1B1 has varying row length depending on address type.
+        if (deviceModel === DeviceModelInternal.T1B1) {
+            switch (text.length) {
+                // legacy and legacy segwit accounts:
+                case 34:
+                    return 17;
+                // legacy segwit testnet accounts:
```
peter-sanderson: I suggest to rewrite it like this: 

```ts
const isLegacySegwitAccount = text.length === 34

if (isLegacySegwitAccount) { return 17}

const isLegacySegwitTestnetAccount = text.length === 35

if (isLegacySegwitTestnetAccount) { return 18 }
```

This way, we dont need comments and code is self-explanatory

packages/suite/src/components/suite/DeviceDisplay/parseTextToPagesAndLines.ts
```diff
@@ -52,10 +53,26 @@ export const parseTextToPagesAndLines = ({
     deviceModel,
     text,
 }: ParseTextToLinesParams): ParseTextToLinesResult => {
+    const getCharsPerRow = (): number => {
+        // T1B1 has varying row length depending on address type.
+        if (deviceModel === DeviceModelInternal.T1B1) {
+            switch (text.length) {
+                // legacy and legacy segwit accounts:
```
peter-sanderson: `legacy and legacy` do I miss something or is it a typo?

packages/suite/src/components/suite/DeviceDisplay/parseTextToPagesAndLines.ts
```diff
@@ -52,10 +53,26 @@ export const parseTextToPagesAndLines = ({
     deviceModel,
     text,
 }: ParseTextToLinesParams): ParseTextToLinesResult => {
+    const getCharsPerRow = (): number => {
+        // T1B1 has varying row length depending on address type.
+        if (deviceModel === DeviceModelInternal.T1B1) {
+            switch (text.length) {
```
peter-sanderson, line 59: Is the text length enough to assume the type of the (bitcoin?) address? What about other coins?

packages/suite/src/components/suite/DeviceDisplay/parseTextToPagesAndLines.ts
komret: `legacy` and `legacy segwit`

peter-sanderson: Maybe  legacy accounts and legacy segwit accounts? So its more clear :pray: (in case comment stays)

peter-sanderson (approved):

packages/suite/src/components/suite/DeviceDisplay/parseTextToPagesAndLines.ts
komret: I can't help it, I love `switch` :D

peter-sanderson: Switch is fine, but the comments that can be replaced by proper variable naming aren't :D

packages/suite/src/components/suite/DeviceDisplay/parseTextToPagesAndLines.ts
komret: 469d534d6992e348a57ed5e34a1d0d70b1e01bbe

packages/suite/src/components/suite/DeviceDisplay/parseTextToPagesAndLines.ts
komret: It possibly doesn't matter whether the text is an address or not and the device always breaks lines like this. Not sure. tested with Litecoin and it works, too.

PR #12971, opened 2024-06-19
fix(suite): token receive disable button when locked

AdamSchinzel:
## Description
Receive button in tokens was not disabled when device when the device is locked or awaiting auth confirmation. That allowed user to click on the button even when discovery was not finished or click multiple times on the button before actually the auth modal was visible.

## Screenshots:
https://github.com/trezor/trezor-suite/assets/66002635/12d371fe-74c9-4a56-81d2-b75ea91c2575


tomasklim (approved):
LGTM

PR #12972, opened 2024-06-19
fix(connect-explorer): load changelog dynamically on client

martykan:
## Description

Loads the latest changelog from Github on the client instead of statically during build. 
Before loading the changelog itself, we also get the latest versions on NPM, to find the latest release branch (beta or prod). This ensures we always show the latest changelog. 

The tradeoff is that it's not loaded instantly and not indexed by search engines, however continually updating it via CI would be more complicated.

Preview: https://dev.suite.sldev.cz/connect/fix/connect-explorer-load-changelog-dynamically/changelog/

### DIFF
--- a/packages/connect-explorer/src/pages/changelog.mdx
+++ b/packages/connect-explorer/src/pages/changelog.mdx
@@ -2,19 +2,27 @@
 icon: LOG
 ---
 
-import { useData } from 'nextra/data';
-import { serialize } from 'next-mdx-remote/serialize';
-import { MDXRemote } from 'next-mdx-remote';
 import rehypeSectionize from '@hbsnow/rehype-sectionize';
 import remarkGfm from 'remark-gfm';
 import remarkGemoji from 'remark-gemoji';
+import useSWR from 'swr';
+import Markdown from 'react-markdown';
+import { useMDXComponents } from '@trezor/connect-explorer-theme';
+import { Callout } from 'nextra/components';
 
 import { spacings } from '@trezor/theme';
 import { CollapsibleBox } from '@trezor/components';
+import { isNewer } from '@trezor/utils/src/versionUtils';
 
-export const getStaticProps = ({ params }) => {
+export const loadNPMVersions = tag => {
+    return fetch(`https://registry.npmjs.org/@trezor/connect/${tag}`)
+        .then(res => res.json())
+        .then(data => data.version);
+};
+
+export const loadChangelog = branch => {
     return fetch(
-        `https://raw.githubusercontent.com/trezor/trezor-suite/${process.env.COMMIT_HASH}/packages/connect/CHANGELOG.md`,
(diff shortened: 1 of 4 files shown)


packages/connect-explorer/src/pages/changelog.mdx
```diff
@@ -2,19 +2,20 @@
...
+import { Callout } from 'nextra/components';
 
 import { spacings } from '@trezor/theme';
 import { CollapsibleBox } from '@trezor/components';
 
-export const getStaticProps = ({ params }) => {
+export const loadData = () => {
     return fetch(
-        `https://raw.githubusercontent.com/trezor/trezor-suite/${process.env.COMMIT_HASH}/packages/connect/CHANGELOG.md`,
+        `https://raw.githubusercontent.com/trezor/trezor-suite/develop/packages/connect/CHANGELOG.md`,
```
mroz22: hmm does this mean that changes in develop are affecting production?

martykan: The problem behind this PR is that when we create a canary release, the [changelog deployed on prod]( https://connect.trezor.io/9/changelog/) doesn't update. There it still has the old canary release

mroz22: I see, but I still don't like that changing develop changes production. Don't we have some kind of predictable branching pattern that could handle this? What does @karliatto think?

martykan: We would have to cherrypick the changelog changes from canary into the current prod release branch. Another option would be to pull the current package versions from NPM, not the changelog.

karliatto: Right, the other option is to use this endpoint from NPM https://registry.npmjs.org/@trezor/connect

You get:
```
response."dist-tags": {
    beta: "9.2.5-beta.1"
    latest: "9.2.4"
}
```

The good thing about this approach is that you always get what is real production and canary versions.

martykan: Ok, that solves the NPM packages, but then there's still `connect.trezor.io/9`

karliatto: You are right, but in practice it should always be the same as NPM. 

With the current release process it should always be the same latest canary version in NPM `@trezor/connect` as in connect.trezor.io/9.x.x, of course it is possible that some day it fails because we do not complete the release for some reason, but that would some error or similar and we should fix it.

mroz22 (approved):

packages/connect-explorer/src/pages/changelog.mdx
martykan: So for now I will load the latest version from NPM and from that determine the latest branch where I will get CHANGELOG.md from Github

PR #12973, opened 2024-06-19
fix(ci): yarn dedupe check

martykan:
## Description

Adds `yarn dedupe --check` to other checks in validation CI job

### DIFF
--- a/.github/workflows/check-code-validation.yml
+++ b/.github/workflows/check-code-validation.yml
@@ -124,6 +124,8 @@ jobs:
         run: yarn verify-project-references
       - name: Detect unused dependencies
         run: yarn depcheck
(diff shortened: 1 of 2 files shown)


.github/workflows/check-code-validation.yml
```diff
@@ -124,6 +124,8 @@ jobs:
         run: yarn verify-project-references
       - name: Detect unused dependencies
         run: yarn depcheck
+      - name: Yarn Dedupe check
```
peter-sanderson, line 127: :clap: :clap: :clap:

peter-sanderson (approved):
Great! (cc @PeKne I think we talked about this, so FYI)

PR #12974, opened 2024-06-19
feat(suite): prefill token send form

### LINKED ISSUE #12820: EVM tokens/ Send form adjustment for tokens
after clicking the 'Send' button next to the token, the user should be taken to the send form, where the token would be prefilled

Screenshot 2024-06-11 at 16 59 52

Screenshot 2024-06-11 at 17 00 06

AdamSchinzel:
## Description

When you click send in tokens table it will redirect you to send form with that token prefilled for you.

## Related Issue

Resolve #12820

### DIFF
--- a/packages/suite/src/constants/suite/routes.ts
+++ b/packages/suite/src/constants/suite/routes.ts
@@ -15,6 +15,7 @@ type RouteParamsTypes = {
     accountIndex: number;
     accountType: NonNullable<Network['accountType']>;
     cancelable: boolean;
+    contractAddress?: string;
 };
 
 type ExtractType<T extends keyof RouteParamsTypes> = {
--- a/packages/suite/src/utils/suite/router.ts
+++ b/packages/suite/src/utils/suite/router.ts
@@ -49,7 +49,7 @@ export const getApp = (url: string) => {
 const validateWalletParams = (url: string): CommonWalletParams => {
     const [, hash] = stripPrefixedURL(url).split('#');
     if (!hash) return;
-    const [symbol, index, type] = hash.split('/').filter(p => p.length > 0);
+    const [symbol, index, type, contractAddress] = hash.split('/').filter(p => p.length > 0);
     if (!symbol || !index) return;
     const network = NETWORKS.find(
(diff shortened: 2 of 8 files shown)


packages/suite/src/constants/suite/routes.ts
```diff
@@ -15,6 +15,7 @@ type RouteParamsTypes = {
     accountIndex: number;
     accountType: NonNullable<Network['accountType']>;
     cancelable: boolean;
+    contractAddressPrefill?: string;
```
AdamSchinzel: Was thinking about some better name for this but didn't come up with anything.

tomasklim: how about just `contractAddress`?

tomasklim: there is symbol, accountType, i think it is fine to use contractAddress

packages/suite/src/utils/suite/router.ts
```diff
@@ -49,7 +49,7 @@ export const getApp = (url: string) => {
 const validateWalletParams = (url: string): CommonWalletParams => {
     const [, hash] = stripPrefixedURL(url).split('#');
     if (!hash) return;
-    const [symbol, index, type] = hash.split('/').filter(p => p.length > 0);
+    const [symbol, index, contractAddress, type] = hash.split('/').filter(p => p.length > 0);
```
tomasklim: I think that `contractAddress` has to be 4th as it is optional

tomasklim: - `#/btc/0/segwit` 
- `web/accounts/tokens#/eth/1/ledger`

tomasklim (changes requested):

packages/suite/src/utils/suite/router.ts
AdamSchinzel: Yes but there was a problem that `normal` type was excluded on purpose from the hash so I removed this and put the `contractAddress` as the last item.

This should also fix the failing unit tests.

tomasklim (approved):

PR #12975, opened 2024-06-19
Feat/group pending transactions by date

komret:
Group pending transactions by date. Otherwise, only time is displayed which is confusing when the transaction has been in a pending state for a long time.

Screenshot 2024-06-20 at 17 17 08

image

### DIFF
--- a/packages/suite/src/views/wallet/transactions/TransactionList/TransactionsGroup/PendingGroupHeader.tsx
+++ b/packages/suite/src/views/wallet/transactions/TransactionList/TransactionsGroup/PendingGroupHeader.tsx
@@ -1,0 +1,14 @@
+import { Translation } from 'src/components/suite';
+import { ColPending, HeaderWrapper } from './CommonComponents';
+
+type PendingGroupHeaderProps = { txsCount: number };
+
+export const PendingGroupHeader = ({ txsCount }: PendingGroupHeaderProps) => {
+    return (
+        <HeaderWrapper>
+            <ColPending data-test="@transaction-group/pending/count">
+                <Translation id="TR_PENDING_TX_HEADING" values={{ count: txsCount }} /> • {txsCount}
+            </ColPending>
+        </HeaderWrapper>
+    );
+};
--- a/suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
+++ b/suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
@@ -169,7 +169,7 @@ export const TransactionList = ({
         const [pendingTxs, confirmedTxs] = arrayPartition(transactions, isPending);
         const accountTransactionsByMonth = groupTransactionsByDate(confirmedTxs, 'month');
         accountTransactionsByMonth['pending'] = [
-            ...accountTransactionsByMonth['pending'],
+            ...accountTransactionsByMonth['no-blocktime'],
             ...pendingTxs,
         ];
         const transactionMonthKeys = Object.keys(accountTransactionsByMonth) as MonthKey[];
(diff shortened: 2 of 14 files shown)


komret:
We encountered some (unrelated) weird stuff when testing this, but given this is testnet and very edge-casey, I am not going to investigate further.

This behaviour can (for now) be observed on the second default testnet account with tbtc2 backend:

1. The date and time of pending transactions changes over time. The have been made sometime in May, but the date keeps updating, usually it is the same day or the day before. Looks like a blockbook bug. Reported here: https://github.com/trezor/blockbook/issues/1118.
Screenshot 2024-07-31 at 9 53 38

2. When attempting to bump fee, the modal says the output amount is _reduced_, but in fact it is increased and the transaction cannot be sing due to  _Transaction signing error: Increasing original output amounts is not supported._ There is some bug in the algorithm, but bear in mind these are chained transactions, sent to myself, possibly to a change address, and there was a fee spike on testnet at the time the original ones were broadcast.
Screenshot 2024-07-31 at 9 48 31

packages/suite/src/views/wallet/transactions/TransactionList/TransactionsGroup/PendingGroupHeader.tsx
```diff
@@ -0,0 +1,14 @@
+import { Translation } from 'src/components/suite';
+import { ColPending, HeaderWrapper } from './CommonComponents';
+
+type PendingGroupHeaderProps = { txsCount: number };
+
+export const PendingGroupHeader = ({ txsCount }: PendingGroupHeaderProps) => {
```
peter-sanderson, line 6: ultra-nit: inline function

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
```diff
@@ -73,6 +74,23 @@ const sectionListContainerStyle = prepareNativeStyle(utils => ({
...
+    if (b === 'pending') return 1;
+
+    const dateA = new Date(a);
+    const dateB = new Date(b);
+
+    return dateA.getTime() - dateB.getTime();
+};
+
+const sortPendingTransactions = (a: WalletAccountTransaction, b: WalletAccountTransaction) => {
+    if (a.blockTime === undefined) return -1;
```
peter-sanderson, line 90: what if both are undefined? I think there should be also:

```ts
if (a.blockTime === undefined && b.blockTime === undefined) return 0
``` 
as first condition

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
```diff
@@ -73,6 +74,23 @@ const sectionListContainerStyle = prepareNativeStyle(utils => ({
     paddingVertical: utils.spacings.small,
 }));
 
+const sortKeysPendingFirst = (a: string, b: string) => {
+    if (a === 'pending') return -1;
```
peter-sanderson, line 79: what if both are `pending`? it shall `return 0` I guess

peter-sanderson (approved):

packages/suite/src/views/wallet/transactions/TransactionList/TransactionsGroup/PendingGroupHeader.tsx
komret: I'd keep this. We may need to add some logic later.

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
matejkriz: I believe the output will be the same, but you are right, let's be explicit, will fix 👍

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
matejkriz: 52579935427dce97d9ef1257f571407e46f5bdbc

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
matejkriz: 52579935427dce97d9ef1257f571407e46f5bdbc

PR #12977, opened 2024-06-19
fix(suite): passphrase missmatch resets the flow now

peter-sanderson:
Addresses the weird behavior when user mismatches the passphrase. Now it resets the flow instead of being stucked in inconsistent state

### DIFF
--- a/packages/connect-web/src/index.ts
+++ b/packages/connect-web/src/index.ts
@@ -156,8 +156,8 @@ export class TrezorConnectDynamicImpl implements ConnectFactoryDependencies {
         return this.getTarget().requestWebUSBDevice();
     }
 
-    public cancel() {
-        return this.getTarget().cancel();
+    public cancel(error?: string) {
+        return this.getTarget().cancel(error);
     }
 
     public dispose() {
--- a/suite-common/wallet-core/src/device/deviceThunks.ts
+++ b/suite-common/wallet-core/src/device/deviceThunks.ts
@@ -420,6 +420,11 @@ export const authConfirm = createThunk(
                 const settings = extra.selectors.selectSuiteSettings(getState());
                 dispatch(deviceActions.forgetDevice({ device, settings }));
 
+                if (settings.isViewOnlyModeVisible) {
+                    const newDevice = selectDeviceSelector(getState());
+                    dispatch(deviceActions.selectDevice(newDevice));
+                }
+
                 return;
             }
             dispatch(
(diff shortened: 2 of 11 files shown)


suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -428,14 +428,22 @@ export const authConfirm = createThunk(
...
         }
 
         if (response.payload.state !== device.state) {
+            console.log('___authConfirm: ', response.payload.state, '!==', device.state);
+
             dispatch(notificationsActions.addToast({ type: 'auth-confirm-error' }));
-            dispatch(deviceActions.receiveAuthConfirm({ device, success: false }));
+
+            const settings = extra.selectors.selectSuiteSettings(getState());
+            dispatch(deviceActions.forgetDevice({ device, settings }));
```
peter-sanderson, line 445: :point_up:

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -407,13 +407,12 @@ export const startDiscoveryThunk = createThunk(
                 await dispatch(initMetadata(false));
             }
 
-            dispatch({
-                type: startDiscovery.type,
```
peter-sanderson, line 411: this is some old way of dispatching actions and it was hard to find the usages when used this way, lets use actionCreator instead

packages/connect-web/src/index.ts
```diff
@@ -156,8 +156,8 @@ export class TrezorConnectDynamicImpl implements ConnectFactoryDependencies {
         return this.getTarget().requestWebUSBDevice();
     }
 
-    public cancel() {
-        return this.getTarget().cancel();
+    public cancel(error?: string) {
```
peter-sanderson, line 159: cc @martykan :scream:  :grin:

peter-sanderson: Todo: would be great to have somehow test for this

martykan: Oops, missed that one. But it's strange TS didn't catch it, the interface should be typed correctly

peter-sanderson: Its optional, thats probably it. 

I am more surprised that no test had failed.... We have a logic in suite that relies on the passage of errors through this.

peter-sanderson: Here for example: https://github.com/trezor/trezor-suite/pull/12977/files#diff-2580f7331a6878d62ed30bb9de2c33adbf4566b5d67d8cf33f5705f48af51930R415

(this is how I found it)

jvaclavik (approved):

Issue #12976, opened 2024-06-19
Connecting screen not displayed 


juriczech:
Connect screen is not being present when device that has enabled passphrase is connected

Video recording can be seen at this PR
https://github.com/trezor/trezor-suite/pull/12889#pullrequestreview-2128190853

STew790:
**QA OK**
Connecting screen shows correctly.

**Info:**

24.6.0 https://github.com/trezor/trezor-suite/commit/c747f2a42759f6804364b7bfbe32010124172b76

PR #12979, opened 2024-06-19
feat(react-native-usb): request usb permission when app goes foreground

matejkriz:
## Description

One more improvement towards https://github.com/trezor/trezor-suite/issues/11896 

This keeps the current dialog to select which app to open whenever Trezor is connected. But on top of that adding request permission if you then go to another Suite (or some potential 3rd party app using react-native-usb), so you can give USB permissions to multiple apps, see the video for the demonstration.

> Currently, the Suite app requests permissions for any USB device connected while the app is open. I attempted to filter this but encountered difficulties passing the ID via the constructor and decided to postpone it for later. It's not a major issue if you cancel or grant permissions for any USB device to Suite, and it seems like a very edge case to me.

## Related Issue

Related to https://github.com/trezor/trezor-suite/issues/11896

## Screenshots:
Demonstration using multiple Suite environments (production and debug versions of the app), but imagine some 3rd party app instead of one of those.
Ignore missing Connecting screen in debug, it's this known issue https://github.com/trezor/trezor-suite/issues/12976

https://github.com/trezor/trezor-suite/assets/3729633/c4381c25-ce1a-4f50-a46d-72eaaaefd208

### DIFF
--- a/packages/react-native-usb/android/src/main/AndroidManifest.xml
+++ b/packages/react-native-usb/android/src/main/AndroidManifest.xml
@@ -17,5 +17,10 @@
                 <action android:name="android.hardware.usb.action.USB_DEVICE_DETACHED" />
             </intent-filter>
         </receiver>
+        <receiver android:name=".ReactNativeUsbPermissionReceiver" android:exported="false">
+            <intent-filter>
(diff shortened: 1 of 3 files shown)


Nodonisko (approved):
cool stuff

PR #12984, opened 2024-06-20
Uninitialized device directed to suite web

### LINKED ISSUE #11896: Handle connection to Trezor Suite web and connected Trezor

matejkriz:
## Description

Design simplifed after discussions with @shenkys. This is only temporary screen till we have onboarding.

ListItems can be reusable for Passphrase flow, but let's do it in separate PR. There is already ListItem in atoms but that can be moved to module-settings because it's not used anywhere else. cc @juriczech 

## Related Issue

Resolve #11896

## Notes for @trezor/qa 
- Please test also that onboarding via web on Android phone is possible without major issues.

## Screenshots:
image

https://github.com/trezor/trezor-suite/assets/3729633/0e7f28e7-6c1a-4326-aa88-5f5c84a0eb8f


suite-native/device/src/hooks/useDetectDeviceError.tsx
```diff
@@ -130,19 +134,23 @@ export const useDetectDeviceError = () => {
...
+
                     analytics.report({
                         type: EventType.UnsupportedDevice,
                         payload: { deviceState: 'noSeed' },
                     });
                 },
+                secondaryButtonTitle: <Translation id="generic.buttons.cancel" />,
+                onPressSecondaryButton: () => {
+                    handleDisconnect();
+                },
```
juriczech:
```suggestion
                onPressSecondaryButton: handleDisconnect,
```

suite-native/intl/src/en.ts
```diff
@@ -247,18 +247,27 @@ export const en = {
         },
     },
     moduleDevice: {
-        IncompatibleDeviceModalAppendix: {
+        incompatibleFirmwareModalAppendix: {
```
juriczech, line 247: Why capitalize first letter? I'd prefer to keep it consistent with camelcase, unless we want to change it everywhere?

suite-native/device/src/components/UninitializedDeviceModalAppendix.tsx
```diff
@@ -0,0 +1,88 @@
...
+} & (
+    | {
+          iconName: IconName;
+          iconNumber?: never;
+      }
+    | {
+          iconName?: never;
+          iconNumber: number;
+      }
+);
```
juriczech: I believe you could use this. IMO it's a little bit more readable when looking at the props. https://github.com/sindresorhus/type-fest/blob/main/source/require-exactly-one.d.ts

suite-native/device/src/components/UninitializedDeviceModalAppendix.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const UninitializedDeviceModalAppendix = () => {
+    return (
+        <VStack>
+            <Text variant="callout">
+                <Translation id="moduleDevice.noSeedModal.appendix.title" />
+            </Text>
+            <VStack spacing="medium" paddingTop="large">
+                <ListItem
+                    iconNumber={1}
+                    translationKey="moduleDevice.noSeedModal.appendix.lines.1"
```
PeKne, line 71: Should not be here a number instead of a icon? Same in the next item.

PeKne: Now it diverges from the design since the check icons are in every item.
Screenshot 2024-07-02 at 7 59 00

suite-native/device/src/components/UninitializedDeviceModalAppendix.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const iconBackgroundStyle = prepareNativeStyle<{ color: Color }>((utils, { color }) => ({
+    backgroundColor: utils.colors[color],
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: utils.spacings.extraLarge,
+    height: utils.spacings.extraLarge,
+    borderWidth: color ? utils.borders.widths.small : 0,
+    borderColor: utils.colors.borderElevation1,
+    borderRadius: 12,
```
PeKne, line 20: Nit: since multiple style props are depending on the `color` I suggest grouping the in a single condition and then remove it form the main style body:


```suggestion
    borderRadius: 12,
    extend: {
        condition: G.isNotNullable(color),
        style: {
            backgroundColor: utils.colors[color],
            borderWidth: utils.borders.widths.small,
            borderColor: utils.colors.borderElevation1,
        },
    },
```

PeKne:
Tested on Model T emulator and works as expected. I was able to navigate to the trezor.io and insert the all-seed to the unitialized Trezor 👏.

suite-native/intl/src/en.ts
matejkriz: It was PascalCase and I made it camelCase...

suite-native/device/src/components/UninitializedDeviceModalAppendix.tsx
matejkriz: That is exactly what I was looking for, but I couldn't remember the name, thanks.

matejkriz: 44bef1e263469b088aaff93b170d0e6f43dc980c

suite-native/device/src/hooks/useDetectDeviceError.tsx
matejkriz: 44bef1e263469b088aaff93b170d0e6f43dc980c

suite-native/intl/src/en.ts
juriczech: Hah, sorry 😄 I guess my brain was still half asleep and switched the colors in my head 🤔

juriczech (approved):
LGTM

suite-native/device/src/components/UninitializedDeviceModalAppendix.tsx
```diff
@@ -0,0 +1,86 @@
...
+    iconBackgroundColor = 'backgroundTertiaryDefaultOnElevation1',
+    translationKey,
+}: ListItemProps) => {
+    const { applyStyle } = useNativeStyles();
+
+    return (
+        <HStack spacing={12} alignItems="center">
+            <Box style={applyStyle(iconBackgroundStyle, { color: iconBackgroundColor })}>
+                {iconNumber && <Text color={iconColor}>{iconNumber}</Text>}
+                {!iconNumber && iconName && (
```
vytick: nit(resolve if you feel like that): these should be mutually exclusive based on RequireExactlyOne  props, shouldnt they? 
```suggestion
                {iconName && (
```

vytick (commented):
Nice work! :)

suite-native/device/src/components/UninitializedDeviceModalAppendix.tsx
matejkriz: nice, thank you for the tip 👍 c50a89424eda2e436bb3840da71c1f713838e8a1

suite-native/device/src/components/UninitializedDeviceModalAppendix.tsx
matejkriz: I tend to not rely on typescript in cases like this and validate whenever it makes sense, but here it makes not much sense and I wouldn't have written the condition if I hadn't written it before the type... 6dddbfb38f53c2ff0a882ba5f8079197ecb44cf9

PeKne (approved):

matejkriz:
> Tested on Model T emulator and works as expected. I was able to navigate to the trezor.io and insert the all-seed to the unitialized Trezor 👏.

P.S.: With the emulator you don't actually test the changed permissions logic, because emulator doesn't use react-native-usb package at all. The real flow can be tested only on real device.

vytick (approved):
🚀

PR #12987, opened 2024-06-20
feat(connect-examples): new webextension example for mv3 sw and ES6 modules in TypeScript

karliatto:
## Description

New webextension example for mv3 sw and ES6 modules in TypeScript.

This will make it more clear to developers using ES6 modules and Typescript and closer to a realistic realworld application scaffolding. 
At the same time it provides a way for us to test webextension package with ES6 modules.

### DIFF
--- a/packages/connect-examples/README.md
+++ b/packages/connect-examples/README.md
@@ -2,17 +2,18 @@
 
 A collection of example implementations of @trezor/connect in various environments. See individual READMEs for instructions:
 
-| Environment                                                           | package                                                           | Popup |
-| --------------------------------------------------------------------- | ----------------------------------------------------------------- | ----- |
-| [Node.js](./node)                                                     | [@trezor/connect](../connect/README.md)                           | No    |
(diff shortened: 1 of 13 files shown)


martykan:
Looks good, just a minor detail, it should probably be added to `connect-examples` README and maybe also linked in `connect-webextension` README

karliatto:
> Looks good, just a minor detail, it should probably be added to `connect-examples` README and maybe also linked in `connect-webextension` README

I added here to connect-example README [e6378a6](https://github.com/trezor/trezor-suite/pull/12987/commits/e6378a67e7c39aeb8b4a98dee97e2fac40c25871)
And in connect-webextension I added it by removing connect-explorer since I think that connect-explorer webextension can be very confusing for developers, but you might have different opinion, let me know if so.  [b0855c6](https://github.com/trezor/trezor-suite/pull/12987/commits/b0855c643da7ea7ee10609cb74e8c5844d20cffd)

martykan (approved):
Great

PR #12991, opened 2024-06-20
Tor with Snowflakes as experimental feature

karliatto:
## Description

In some countries ISPs have started blocking Trezor servers, in the same countries  using Tor directly is not possible, it is blocked at the ISP level as well. 

Tor snowflakes (https://snowflake.torproject.org/) is "pluggable" transport of Tor that helps user to circumvent  those prohibitions. 

* Make Tor run with snowflakes as a proxy integrated in Suite
* Repository with all the binaries https://www.torproject.org/download/tor/
* There could be a feature that will download the binary for the architecture and use it.

image

Proper documentation about what is required to run it will be added to Trezor Knowledge Base but if you are reviewing it now you can follow steps bellow to get the binary.

```
* Use it with Debian
      - Run "apt install snowflake-client"
      - Then use the path to the binary "/usr/bin/snowflake-client"

      or:
      - Download it from https://packages.debian.org/bookworm/amd64/snowflake-client/download
      - Install it like sudo dpkg -i snowflake-client_2.5.1-1+b3_amd64.deb
      - Then use the path to the binary "/usr/bin/snowflake-client"

* Use it with macOS:
      - Run "brew install snowflake"
      - Then use the path to the binary "/opt/homebrew/bin/snowflake-client"

* Use it with Windows:

      - Download Tor Browser from [here](https://www.torproject.org/dist/torbrowser/13.5.1/tor-browser-windows-x86_64-portable-13.5.1.exe).
      - Install it. By default, it installs (extracts) on the Desktop.
      - The binary is located in "C:\Users\USERNAME\Desktop\Tor Browser\Browser\TorBrowser\Tor\PluggableTransports"
```

### DIFF
--- a/packages/node-utils/src/tests/checkFileExists.test.ts
+++ b/packages/node-utils/src/tests/checkFileExists.test.ts
@@ -1,0 +1,49 @@
+import fs from 'fs';
+import path from 'path';
+
+import { checkFileExists } from '../checkFileExists';
+
+const existingFilePath = path.join(__dirname, 'test-file.txt');
+const nonExistingFilePath = path.join(__dirname, 'non-existing-file.txt');
+const directoryPath = path.join(__dirname, 'test-directory');
+
+beforeAll(async () => {
+    await fs.promises.writeFile(existingFilePath, 'Test content');
+});
+
+afterAll(async () => {
+    try {
+        await fs.promises.unlink(existingFilePath);
+    } catch (error) {
+        console.error(error);
+    }
+    try {
+        await fs.promises.rmdir(directoryPath);
+    } catch (error) {
+        console.error(error);
+    }
+});
+
+describe('checkFileExists', () => {
+    it('should return true for an existing file', async () => {
+        const exists = await checkFileExists(existingFilePath);
+        expect(exists).toBe(true);
+    });
+
+    it('should return false for a non-existing file', async () => {
+        const exists = await checkFileExists(nonExistingFilePath);
+        expect(exists).toBe(false);
+    });
+
+    it('should return false for a directory path', async () => {
+        await fs.promises.mkdir(directoryPath);
+        const exists = await checkFileExists(directoryPath);
+        expect(exists).toBe(false);
+    });
+
+    it('should handle invalid paths gracefully', async () => {
+        const invalidPath = path.join(__dirname, 'invalid-path', 'file.txt');
+        const exists = await checkFileExists(invalidPath);
+        expect(exists).toBe(false);
+    });
+});
--- a/packages/request-manager/e2e/interceptor.test.ts
+++ b/packages/request-manager/e2e/interceptor.test.ts
@@ -11,7 +11,6 @@ const port = 38835;
 const controlPort = 35527;
 const processId = process.pid;
 const snowflakeBinaryPath = '';
-const shouldUseSnowflake = false;
 
 // 1 minute before timeout, because Tor might be slow to start.
 jest.setTimeout(60000);
@@ -47,9 +46,8 @@ describe('Interceptor', () => {
             controlPort,
             torDataDir,
             snowflakeBinaryPath,
-            shouldUseSnowflake,
         });
-        const torParams = torController.getTorConfiguration(processId, false);
+        const torParams = await torController.getTorConfiguration(processId);
         // Starting Tor process from binary.
         torProcess = torRunner({
             torParams,
--- a/packages/request-manager/e2e/interceptor.test.ts
+++ b/packages/request-manager/e2e/interceptor.test.ts
@@ -10,6 +10,8 @@ const host = '127.0.0.1';
 const port = 38835;
 const controlPort = 35527;
 const processId = process.pid;
+const snowflakeBinaryPath = '';
+const shouldUseSnowflake = false;
 
 // 1 minute before timeout, because Tor might be slow to start.
 jest.setTimeout(60000);
@@ -28,7 +30,7 @@ describe('Interceptor', () => {
     let torController: TorController;
     let torIdentities: TorIdentities;
 
-    const torSettings = { running: true, host, port };
+    const torSettings = { running: true, host, port, snowflakeBinaryPath };
 
     const INTERCEPTOR = {
         handler: () => {},
@@ -44,8 +46,10 @@ describe('Interceptor', () => {
             port,
             controlPort,
             torDataDir,
+            snowflakeBinaryPath,
+            shouldUseSnowflake,
         });
-        const torParams = torController.getTorConfiguration(processId);
+        const torParams = torController.getTorConfiguration(processId, false);
         // Starting Tor process from binary.
         torProcess = torRunner({
             torParams,
--- a/packages/request-manager/src/controller.ts
+++ b/packages/request-manager/src/controller.ts
@@ -2,6 +2,7 @@ import { EventEmitter } from 'events';
 import path from 'path';
 
 import { createTimeoutPromise } from '@trezor/utils';
+import { checkFileExists } from '@trezor/node-utils';
 
 import { TorControlPort } from './torControlPort';
 import {
@@ -66,14 +67,13 @@ export class TorController extends EventEmitter {
         }, this.bootstrapSlowThreshold);
     }
 
-    public getTorConfiguration(
+    public async getTorConfiguration(
         processId: number,
-        useSnowflake?: boolean,
         snowflakeBinaryPath?: string,
-    ): string[] {
+    ): Promise<string[]> {
         const { torDataDir } = this.options;
         const controlAuthCookiePath = path.join(torDataDir, 'control_auth_cookie');
-        const snowflakeLogPath = path.join(this.options.torDataDir, 'snowflake.log');
+        const snowflakeLogPath = path.join(torDataDir, 'snowflake.log');
 
         // https://github.com/torproject/tor/blob/bf30943cb75911d70367106af644d4273baaa85d/doc/man/tor.1.txt
         const config: string[] = [
@@ -136,7 +136,13 @@ export class TorController extends EventEmitter {
             this.options.torDataDir,
         ];
 
-        if (useSnowflake && snowflakeBinaryPath && snowflakeBinaryPath.trim() !== '') {
+        let existsSnowflakeBinary = false;
+        if (snowflakeBinaryPath && snowflakeBinaryPath.trim() !== '') {
+            // If provided snowflake file does not exists, do not use it.
+            existsSnowflakeBinary = await checkFileExists(snowflakeBinaryPath);
+        }
+
+        if (existsSnowflakeBinary) {
             // Snowflake is a WebRTC pluggable transport for Tor (client)
             // More info:
             // https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/tree/main/client
--- a/packages/request-manager/src/controller.ts
+++ b/packages/request-manager/src/controller.ts
@@ -66,11 +66,17 @@ export class TorController extends EventEmitter {
         }, this.bootstrapSlowThreshold);
     }
 
-    public getTorConfiguration(processId: number): string[] {
-        const controlAuthCookiePath = path.join(this.options.torDataDir, 'control_auth_cookie');
+    public getTorConfiguration(
+        processId: number,
+        useSnowflake?: boolean,
+        snowflakeBinaryPath?: string,
+    ): string[] {
+        const { torDataDir } = this.options;
+        const controlAuthCookiePath = path.join(torDataDir, 'control_auth_cookie');
+        const snowflakeLogPath = path.join(this.options.torDataDir, 'snowflake.log');
 
         // https://github.com/torproject/tor/blob/bf30943cb75911d70367106af644d4273baaa85d/doc/man/tor.1.txt
-        return [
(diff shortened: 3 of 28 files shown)


mroz22 (commented):
are the client files automatically checked out to git-lfs? 
I guess they are but I am not very skilled with git-lfs so maybe i am doing something wrong but I can't see it there 
```git lfs ls-files --all | grep snowflake```

**Comment by karliatto** [comment]
> are the client files automatically checked out to git-lfs? I guess they are but I am not very skilled with git-lfs so maybe i am doing something wrong but I can't see it there `git lfs ls-files --all | grep snowflake`

@mroz22 you are right it was not in lfs but I just added it. Thanks!

File: packages/request-manager/src/controller.ts [hunk 1]
```diff
@@ -66,11 +66,17 @@ export class TorController extends EventEmitter {
...
-    public getTorConfiguration(processId: number): string[] {
-        const controlAuthCookiePath = path.join(this.options.torDataDir, 'control_auth_cookie');
+    public getTorConfiguration(
+        processId: number,
+        useSnowflake?: boolean,
+        snowflakeBinaryPath?: string,
+    ): string[] {
+        const { torDataDir } = this.options;
+        const controlAuthCookiePath = path.join(torDataDir, 'control_auth_cookie');
+        const snowflakeLogPath = path.join(this.options.torDataDir, 'snowflake.log');
```
marekrjpolak: nit:
```suggestion
        const snowflakeLogPath = path.join(torDataDir, 'snowflake.log');
```

packages/suite/src/views/settings/SettingsGeneral/TorSnoflake.tsx
```diff
@@ -0,0 +1,87 @@
+import { Input, Switch } from '@trezor/components';
```
marekrjpolak: nit: typo in file name

packages/suite/src/views/settings/SettingsGeneral/TorSnoflake.tsx
```diff
@@ -0,0 +1,87 @@
...
+            }
+        });
+
+        desktopApi.getTorSettings().then(result => {
+            if (result.success) {
+                setTorSettings(result.payload);
+            } else {
+                setError(result.error);
+            }
+        });
```
marekrjpolak: Rebase issue?

packages/suite/src/views/settings/SettingsGeneral/TorSnoflake.tsx
```diff
@@ -0,0 +1,87 @@
...
+                                shouldUseSnowflake: !torSettings.shouldUseSnowflake,
+                            })
+                        }
+                    />
+                </ActionColumn>
+            </SectionItem>
+            <SectionItem data-test="@settings/general/tor/snowflake-binary-path">
+                <TextColumn
+                    title="Path for Tor Snowflake"
+                    description="This should be the path of Snowflake in your system."
```
marekrjpolak: Missing translations, but it's probably kinda ok for experimental features.

marekrjpolak (commented):
I'm not quite sure about separate `shouldUseSnowflake` bool and **required** `snowflakeBinaryPath` string.

First, I'm not sure whether it makes sense to have binary path set but the whole snowflake turned off, or to have it turned on without valid binary path.

Second, what happened to me: I already had `torSettings` in main (=> no fallback to default). When I opened Suite settings, I received these stored settings, without the flag and binary path. I tried to turn snowflake on (which could work even with `shouldUseSnowflake` undefined), but the message was blocked by `suite-desktop-api` because it's invalid without `snowflakeBinaryPath`, which was never set anywhere yet.

Wouldn't it make sense to have it as an input with save button (and therefore also not sending every typed character to desktop-api, for example as in Device/Customization/Device name) and either have a string there (= snowflake on) or not (snowflake off)? And if not, at least make the `snowflakeBinaryPath` optional, instead of having `''` as a default value?

(Note: I still haven't tried whether it works or not as a whole, these were just first observations of mine)

packages/suite/src/views/settings/SettingsGeneral/TorSnoflake.tsx
karliatto: Yes, it should be fixed now https://github.com/trezor/trezor-suite/pull/12991/commits/35d42e1dda383e2601304466c42ea2d0b9bb7c76

packages/suite/src/views/settings/SettingsGeneral/TorSnoflake.tsx
karliatto: This text is not used anymore in the new version https://github.com/trezor/trezor-suite/pull/12991/commits/35d42e1dda383e2601304466c42ea2d0b9bb7c76#diff-3227a1b9781609288bd6371e791b3b03d4f459f1f231763b6891cb774da430e6

packages/suite/src/views/settings/SettingsGeneral/TorSnoflake.tsx
karliatto: Thanks! Fixed https://github.com/trezor/trezor-suite/pull/12991/commits/35d42e1dda383e2601304466c42ea2d0b9bb7c76#diff-3227a1b9781609288bd6371e791b3b03d4f459f1f231763b6891cb774da430e6

packages/request-manager/src/controller.ts
karliatto: Right! dd4ab4d899c527a21728300159673622a222b18d

karliatto:
> I'm not quite sure about separate `shouldUseSnowflake` bool and **required** `snowflakeBinaryPath` string.
> 
> First, I'm not sure whether it makes sense to have binary path set but the whole snowflake turned off, or to have it turned on without valid binary path.
> 
> Second, what happened to me: I already had `torSettings` in main (=> no fallback to default). When I opened Suite settings, I received these stored settings, without the flag and binary path. I tried to turn snowflake on (which could work even with `shouldUseSnowflake` undefined), but the message was blocked by `suite-desktop-api` because it's invalid without `snowflakeBinaryPath`, which was never set anywhere yet.
> 
> Wouldn't it make sense to have it as an input with save button (and therefore also not sending every typed character to desktop-api, for example as in Device/Customization/Device name) and either have a string there (= snowflake on) or not (snowflake off)? And if not, at least make the `snowflakeBinaryPath` optional, instead of having `''` as a default value?
> 
> (Note: I still haven't tried whether it works or not as a whole, these were just first observations of mine)

I made it in "2 steps", path and enable, because putting myself in the foot of the final user I thought it was easier to understand.  I am still not sure so if you really think it will be better with only path and when path is there then it means it has to use snowflake let me know.

But I made couple of changes:
1. Follow your suggestion and implemented it similar to Device name so it does send it to desktop-api when enable
2. Add a check that it has to be a "full path"
3. Add a check in request-manager that checks if the file exists and it is actually a file otherwise it defaults to normal Tor without it.

packages/utils/src/isFullPath.ts
```diff
@@ -0,0 +1,13 @@
+export const isFullPath = (path: string) => {
+    // Full Unix-like and Windows paths
+    const fullPathPattern = /^(\/|([a-zA-Z]:\\))/;
+
+    // Check to avoid matching URLs
+    const urlPattern = /^(https?:\/\/|www\.)/;
```
marekrjpolak: It seems that the intersection of these two regexes is empty, therefore the second one is not needed, is it possible?

packages/node-utils/src/tests/checkFileExists.test.ts
```diff
@@ -0,0 +1,48 @@
+import { checkFileExists } from '../checkFileExists';
```
marekrjpolak: Eslint warnings:
``` There should be at least one empty line between import groups ```
``` `../checkFileExists` import should occur after import of `path` ```

File: packages/request-manager/e2e/interceptor.test.ts [hunk 7]
```diff
@@ -44,8 +45,9 @@ describe('Interceptor', () => {
             port,
             controlPort,
             torDataDir,
+            snowflakeBinaryPath,
         });
-        const torParams = torController.getTorConfiguration(processId);
+        const torParams = await torController.getTorConfiguration(processId, false);
```
marekrjpolak: The second param probably shouldn't be bool anymore.

packages/request-manager/src/controller.ts
```diff
@@ -2,6 +2,7 @@ import { EventEmitter } from 'events';
 import path from 'path';
 
 import { createTimeoutPromise } from '@trezor/utils';
+import { checkFileExists } from '@trezor/node-utils';
```
marekrjpolak, line 5: Eslint error:
```'@trezor/node-utils' should be listed in the project's dependencies. Run 'npm i -S @trezor/node-utils' to add it```

File: packages/utils/src/isFullPath.ts [hunk 9]
```diff
@@ -0,0 +1,13 @@
...
+
+    // Check for full path patterns, excluding URLs
+    const isFullPath = fullPathPattern.test(path);
```
marekrjpolak: Eslint error:
```'isFullPath' is already declared in the upper scope on line 1 column 14.```

File: packages/suite-desktop-core/src/libs/processes/TorProcess.ts [hunk 10]
```diff
@@ -1,13 +1,8 @@
 import { TorController, TOR_CONTROLLER_STATUS } from '@trezor/request-manager';
 
-import { BaseProcess, Status } from './BaseProcess';
+import { TorConnectionOptions } from '@trezor/request-manager/src/types';
```
marekrjpolak, line 2: Eslint warning:
```There should be no empty line within import group```

marekrjpolak (commented):
I'm not sure whether it's issue of my setup, Tor network or something else, but when I set the correct snowflake binary path, Tor never connects and I'm not able to turn it off until Suite restart.

Also, I don't know how to turn the snowflake off now. 
image

packages/request-manager/src/controller.ts
karliatto: https://github.com/trezor/trezor-suite/pull/12991/commits/c53bdc9fcf78fcf83c266fad4f8d5792eb01e73e

packages/request-manager/e2e/interceptor.test.ts
karliatto: https://github.com/trezor/trezor-suite/pull/12991/commits/c53bdc9fcf78fcf83c266fad4f8d5792eb01e73e

packages/node-utils/src/tests/checkFileExists.test.ts
karliatto: 32c6d71fcced0dd60bedb4477e70344ba211d51a

packages/suite-desktop-core/src/libs/processes/TorProcess.ts
karliatto: 449369ec9a3e6ac35f219dfd1f76a3833003e5c1

packages/utils/src/isFullPath.ts
karliatto: What about a simplification like https://github.com/trezor/trezor-suite/pull/12991/commits/6a40e46a6c52c8df049c4d8d45434be4adc8651a  ?

packages/utils/src/isFullPath.ts
karliatto: Let's simplify it like https://github.com/trezor/trezor-suite/commit/6a40e46a6c52c8df049c4d8d45434be4adc8651a

karliatto:
> Also, I don't know how to turn the snowflake off now.

Good point, what about when changing from some path to empty string it will display "Disable Snowflake" in the button to save it like there 589e43f8dd8cc5e0237accc96bd35b84815e02b5  ?

image

packages/utils/src/isFullPath.ts
marekrjpolak: Excellent!

Ondra-Zik:
> Good poin

Also, maybe I'm commenting a bit late, so it's okay if it is not possible, but could there be a file picker so you are not forced to copy the path from file explorer?

marekrjpolak:
> could there be a file picker

Now I'm just genuinely amazed that it didn't cross my mind until you mentioned it @Ondra-Zik-SL :sweat_smile:

marekrjpolak (approved):
I can confirm that it works and I believe it's good enough for experimental features. Good job!

I agree with @Ondra-Zik-SL that the file picker is nice-to-have, but it's definitely not necessary for my approval.

karliatto:
> > Good poin
> 
> Also, maybe I'm commenting a bit late, so it's okay if it is not possible, but could there be a file picker so you are not forced to copy the path from file explorer?

That is a good point, I didn't think about it neither. :sweat_smile: I will merge with current approach and will iterate over the file picker in a follow-up.

Thanks both @marekrjpolak and @Ondra-Zik-SL for the feedback!

karliatto:
@Ondra-Zik-SL it is actually not possible to use the `<input type="file">` from HTML due to browser security features it does not allow you to get the full path, just the file object that can be used for uploading for example, but not for our purposes. Instead the browser replaces the path with default "C:\fakepath\" for historical reasons. More details in link below:

https://html.spec.whatwg.org/multipage/input.html#fakepath-srsly

PR #12993, opened 2024-06-20
fix:(blockchain-link): omit solana foreign tx effects

PeterBenc:
## Description
 
Previously, transaction history and tx details showed positive transaction effects also on other addresses. This could have been confusing for users, as a foreign address was shown right next to users address.

## Changes

Omit token transaction effects if sender nor receiver is the users account
Omit SOL positive transaction effects on foreign addresses for transactions other than sent

## Notes
I haven't been able to test this e2e yet, I don't have device on me and the emulator is fighting me as always. I will do it tomorrow. Until then, you can ideally test yourself and you can still check the unit tests I added which test all 3 places where the changes need to be made.

### DIFF
--- a/packages/blockchain-link-utils/src/solana.ts
+++ b/packages/blockchain-link-utils/src/solana.ts
@@ -251,6 +251,12 @@ export const getTargets = (
                 return false;
             }
 
+            // exclude effects on foreign addresses for tx types other then sent, otherwise it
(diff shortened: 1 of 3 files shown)


tomasklim:
Closing in favor of https://github.com/trezor/trezor-suite/pull/12996

PR #12996, opened 2024-06-21
fix:(blockchain-link): omit solana foreign tx effects

### LINKED ISSUE #10915: "Receiving" SOL when receiving token
**Describe the bug**
It looks like that I got some Solana together with a token but it is not true

**Info:**

-   Suite Version: 24.2.0

**How to reproduce**

1. Send solana tokens

**Expected behavior**
It should not be visible

**Screenshots**

Screenshot 2024-01-30 at 14 32 16
Screenshot 2024-01-30 at 14 32 30
Screenshot 2024-01-30 at 14 32 44

tomasklim:
Omit token transaction effects if sender not receiver is the account Omit SOL positive transaction effects on foreign addresses for transactions other than sent

origin https://github.com/trezor/trezor-suite/pull/12993
closes #10915

### DIFF
--- a/packages/blockchain-link-utils/src/solana.ts
+++ b/packages/blockchain-link-utils/src/solana.ts
@@ -251,6 +251,12 @@ export const getTargets = (
                 return false;
             }
 
+            // Exclude effects on foreign addresses for tx types other than sent, otherwise it
+            // leads to the foreign address being displayed next to user's own address which might lead to confusion.
+            if (txType !== 'sent' && effect.address !== accountAddress) {
+                return false;
+            }
+
             // count in only positive effects, for `sent` tx they gonna be represented as negative, for `recv` as positive
             return effect.amount.isGreaterThan(0);
         })
@@ -356,7 +362,13 @@ export const getDetails = (
     txType: Transaction['type'],
 ): Transaction['details'] => {
     const senders = effects.filter(({ amount }) => amount.isNegative());
-    const receivers = effects.filter(({ amount }) => amount.isPositive());
+
+    // include positive effects only on accountAddress for tx types other then sent, otherwise it
+    // leads to foreign address being displayed next to users own address which might lead to confusion
+    const receivers = effects.filter(
+        ({ amount, address }) =>
+            amount.isPositive() && (txType !== 'sent' ? address === accountAddress : true),
+    );
 
     const getVin = ({ address, amount }: { address: string; amount?: BigNumber }, i: number) => ({
         txid: transaction.transaction.signatures[0].toString(),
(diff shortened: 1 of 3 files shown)


packages/blockchain-link-utils/src/solana.ts
```diff
@@ -251,6 +251,12 @@ export const getTargets = (
                 return false;
             }
 
+            // exclude effects on foreign addresses for tx types other then sent, otherwise it
+            // leads to the foreign address being displayed next to users own address which might lead to confusion
```
komret:
```suggestion
            // Exclude effects on foreign addresses for tx types other than sent, otherwise it
            // leads to the foreign address being displayed next to user's own address which might lead to confusion.
```

packages/blockchain-link-utils/src/solana.ts
```diff
@@ -356,7 +362,11 @@ export const getDetails = (
     txType: Transaction['type'],
 ): Transaction['details'] => {
     const senders = effects.filter(({ amount }) => amount.isNegative());
-    const receivers = effects.filter(({ amount }) => amount.isPositive());
+    const receivers = effects
+        .filter(({ amount }) => amount.isPositive())
+        // include positive effects only on accountAddress for tx types other then sent, otherwise it
+        // leads to foreign address being displayed next to users own address which might lead to confusion
+        .filter(({ address }) => txType !== 'sent' ? address === accountAddress : true);
```
komret: The condition and comment are repeated, let's abstract it to a function?

packages/blockchain-link-utils/src/solana.ts
```diff
@@ -356,7 +362,11 @@ export const getDetails = (
     txType: Transaction['type'],
 ): Transaction['details'] => {
     const senders = effects.filter(({ amount }) => amount.isNegative());
-    const receivers = effects.filter(({ amount }) => amount.isPositive());
+    const receivers = effects
+        .filter(({ amount }) => amount.isPositive())
```
komret: I'd merge the two filters to avoid looping twice.

komret (changes requested):

packages/blockchain-link-utils/src/solana.ts
tomasklim: a2c1eeaea0fdcd908163b94f12cea63c8549706d

packages/blockchain-link-utils/src/solana.ts
tomasklim: a2c1eeaea0fdcd908163b94f12cea63c8549706d

komret (approved):

PR #12998, opened 2024-06-21
Feat/checksum

tomasklim:
run pipelines


packages/components/src/components/form/BottomText.tsx
```diff
@@ -10,14 +10,14 @@ import { InputState } from './inputTypes';
...
 const slideDown = keyframes`
-    from {
-        opacity: 0;
-        transform: translateY(-2px);
-    }
-    to {
-        opacity: 1;
-        transform: translateY(0);
-    }
+from {
```
peter-sanderson, line 13: why this change (blank space only?)?

packages/components/src/components/form/BottomText.tsx
```diff
@@ -36,25 +36,32 @@ interface BottomTextProps {
     inputState?: InputState;
     isDisabled?: boolean;
     icon?: IconName;
+    withIcon?: boolean;
     children: ReactNode;
 }
-
+//TODO: refactor as in Warning component
```
peter-sanderson, line 42: Why not do it? Shall be easy

packages/components/src/components/form/Input/Input.tsx
```diff
@@ -198,7 +198,11 @@ const Input = ({
             </InputWrapper>
 
             {bottomText && (
-                <BottomText inputState={inputState} isDisabled={isDisabled}>
+                <BottomText
+                    withIcon={inputState && ['error', 'warning'].includes(inputState)}
```
peter-sanderson, line 207: cc @jvaclavik this looks fishy WDYT?

packages/components/src/components/form/Input/Input.tsx
```diff
@@ -198,7 +198,11 @@ const Input = ({
             </InputWrapper>
 
             {bottomText && (
-                <BottomText inputState={inputState} isDisabled={isDisabled}>
+                <BottomText
```
peter-sanderson, line 206: I would assume that shall be just injectable component and you can send it through props and not just hardcoded `BottomText`

peter-sanderson: I think here we should have something like `bottomTextRendered()` which is a function that can render custom bottom text and we can pass `inputState` as a parameter of the `bottomTextRendered` to inject it

packages/components/src/components/form/Input/Input.tsx
peter-sanderson: anyway `withIcon` can be solved inside the `BottomText` same way as the `icon` itself is. 

You are already passing whole `inputState` so no need to solve it on this level.

jvaclavik: Yes, it looks fishy to me. Not sure if we want to do it for all inputs.

packages/suite/src/components/wallet/InputChecksumInfo.tsx
```diff
@@ -0,0 +1,75 @@
...
+`;
+
+const Container = styled.div`
+    display: flex;
+    flex-direction: row;
+    margin: 0;
+    padding: 0;
+`;
+
+export const Divider = styled.div`
```
peter-sanderson, line 43: We have a special component for the divider: 
`packages/components/src/components/Divider/Divider.tsx`

packages/suite/src/components/wallet/InputChecksumInfo.tsx
```diff
@@ -0,0 +1,75 @@
...
+    height: inherit;
+    align-self: stretch;
+    width: 2px;
+    background-color: ${({ theme }) => theme.backgroundSurfaceElevation1};
```
peter-sanderson, line 47: Just note: please use `elevation` mechanics and dont hardcode the elevation colors

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -81,15 +82,20 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+        return undefined;
+    };
 
     const handleQrClick = useCallback(async () => {
         const uri = await dispatch(scanOrRequestSendFormThunk()).unwrap();
 
         if (typeof uri !== 'string') {
             return;
         }
-
```
peter-sanderson, line 92: nit: I prefer having whitespecas to separate code, it imo makes it more readable

packages/suite/src/components/wallet/InputChecksumInfo.tsx
```diff
@@ -0,0 +1,75 @@
+import styled, { css } from 'styled-components';
+import { useTheme } from 'styled-components';
+import { TooltipSymbol } from '../suite';
+import { spacingsPx, borders } from '@trezor/theme';
+import { TranslationKey, Translation } from '../suite/Translation';
+import { useTranslation } from 'src/hooks/suite';
+import { Icon } from '@suite-common/icons/src/webComponents';
+
+const Wrapper = styled.div`
```
jvaclavik, line 9: Btw it would be useful to add screenshots to PRs like this

packages/components/src/components/form/BottomText.tsx
enjojoy: I think it was a linter conflict, removed to the previous state in here https://github.com/trezor/trezor-suite/pull/13623/commits/5e2a396fd96765f4330a71cf21ce68b46196a2f9

enjojoy:
Now working on it [here](https://github.com/trezor/trezor-suite/pull/13623)

packages/suite/src/components/wallet/InputChecksumInfo.tsx
enjojoy: Yes, It was just a copy of my PR that I created when I was outside of the org and couldn't run tests, to see that tests can run. 

In my initial PR there were screenshots, now I created a new one as a part of the org and will add screenshots there too.

packages/suite/src/components/wallet/InputChecksumInfo.tsx
enjojoy: Added [here](https://github.com/trezor/trezor-suite/pull/13623/commits/2bd1c33effa173297e4ce2e3361f0bf127583814)

I had to adjust the Divider component a little so it's more flexible for my purpose. If you'll have any further comments please add them to this [PR](https://github.com/trezor/trezor-suite/pull/13623)

packages/components/src/components/form/Input/Input.tsx
enjojoy: Agree, moved this logic inside BottomText https://github.com/trezor/trezor-suite/pull/13623/commits/ac001b0ffa37a4225922db6e3351fd1f1355d641

PR #13008, opened 2024-06-21
Fix passphrase missmatch refactoring

jvaclavik:
## Related Issue

Resolve

### DIFF
--- a/packages/suite/src/actions/wallet/addWalletThunk.ts
+++ b/packages/suite/src/actions/wallet/addWalletThunk.ts
@@ -1,0 +1,75 @@
+import { createThunk } from '@suite-common/redux-utils';
+import { TrezorDevice } from '@suite-common/suite-types';
+import { getDeviceInstances } from '@suite-common/suite-utils';
+import {
+    DEVICE_MODULE_PREFIX,
+    createDeviceInstanceThunk,
+    deviceActions,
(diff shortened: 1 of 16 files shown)


suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -20,7 +20,7 @@ import {
     getDeviceInstances,
     getFirstDeviceInstance,
 } from '@suite-common/suite-utils';
-import { AccountKey } from '@suite-common/wallet-types';
+import { AccountKey, WalletType } from '@suite-common/wallet-types';
```
peter-sanderson: maybe unused import?

packages/suite/src/actions/wallet/addWalletThunk.ts
```diff
@@ -0,0 +1,77 @@
...
+            await dispatch(
+                createDeviceInstanceThunk({
+                    device,
+                    useEmptyPassphrase: walletType === WalletType.STANDARD,
+                }),
+            );
+        };
+
+        const selectDeviceInstance = ({ device }: { device: TrezorDevice }) => {
+            if (walletType !== undefined) {
```
peter-sanderson: `walletType` cannot be undefined, useless condition I think

PR #13027, opened 2024-06-21
Staking copy fixes en

MiroslavProchazka:
## Related Issue

Resolve

### DIFF
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -8933,7 +8933,7 @@ export default defineMessages({
     TR_STAKE_UNSTAKING_TAKES: {
         id: 'TR_STAKE_UNSTAKING_TAKES',
         defaultMessage:
-            'Unstaking currently takes {count, plural, one {# day} few {# days} other {# days}}. Once completed, you can trade or send it.',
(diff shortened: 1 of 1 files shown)


packages/suite/src/support/messages.ts
```diff
@@ -8933,7 +8933,7 @@ export default defineMessages({
     TR_STAKE_UNSTAKING_TAKES: {
         id: 'TR_STAKE_UNSTAKING_TAKES',
         defaultMessage:
-            'Unstaking currently takes {count, plural, one {# day} few {# days} other {# days}}. Once completed, you can trade or send it.',
+            'Unstaking currently takes {count, plural, one {# day} other {# days}}. Once completed, you can trade or send it.',
```
tomasklim, line 8936: few should be in all strings
because 1 den, 2 dny, 5 dnů

MiroslavProchazka: It was not working with "few" parameter, Crowdin showed syntax error. Thats why we get rid of it

tomasklim: but it is valid for Czech, so it should me modified there

MiroslavProchazka: We will take care of it in Crowdin as well

tomasklim (approved):

PR #13031, opened 2024-06-21
feat(connect-web): automatic fallback to core in popup

### LINKED ISSUE #12488: Add webUSB as a secondary transport option to connect-web package
In `trezor/connect-web`:

if **Trezor Bridge** is running, use **core in iframe**

if **Trezor Bridge** is not running and/or **core in iframe** cannot be used, use **core in popup**

martykan:
## Description

Implement automatic fallback for core in popup. 

- New param `coreMode`
    - `auto` - automatically choose the best option based on the environment
    - `iframe` (default) - always use iframe, same way as before
    - `popup` - always use popup,
- Switch to popup on error (`Init_IframeBlocked`, `Init_IframeTimeout`, `Transport_Missing`, …)
- Checking for bridge state is done using a new event `TRANSPORT.GET_INFO`, which will trigger `Transport_Missing` during init if in auto mode

## Related Issue
Resolve #12488

## Followup
- Deprecate `useCoreInPopup` param
- Update documentation
- Use auto as default?

### DIFF
--- a/packages/connect-explorer/src/pages/settings.mdx
+++ b/packages/connect-explorer/src/pages/settings.mdx
@@ -29,6 +29,7 @@ export const Settings = () => {
         trustedHost: state.connect?.options?.trustedHost,
         connectSrc: state.connect?.options?.connectSrc,
         useCoreInPopup: state?.connect?.options?.useCoreInPopup,
+        coreMode: state?.connect?.options?.coreMode,
     }));
 
     const initError = useSelector(state => state.connect?.initError);
@@ -52,6 +53,17 @@ export const Settings = () => {
             type: 'checkbox',
             key: 'useCoreInPopup',
             value: connectOptions?.useCoreInPopup || false,
+        },
+        {
+            name: 'coreMode',
+            type: 'select',
+            key: 'coreMode',
+            value: connectOptions?.coreMode || 'iframe',
+            data: [
+                { value: 'auto', label: 'Auto' },
+                { value: 'iframe', label: 'Iframe' },
+                { value: 'popup', label: 'Popup' },
+            ],
         },
         {
             name: 'connectSrc',
--- a/packages/connect-web/src/impl/core-in-iframe.ts
+++ b/packages/connect-web/src/impl/core-in-iframe.ts
@@ -14,7 +14,6 @@ import {
     BLOCKCHAIN_EVENT,
     TRANSPORT,
     parseMessage,
-    createUiMessage,
     createErrorMessage,
     UiResponseEvent,
     CoreEventMessage,
@@ -174,6 +173,18 @@ export class CoreInIframe implements ConnectFactoryDependencies {
 
         await iframe.init(this._settings);
 
+        if (this._settings.coreMode === 'auto') {
+            // Checking bridge availability
+            const { promiseId, promise } = this._messagePromises.create();
+            this._log.debug('coreMode = auto, Checking bridge availability');
+            iframe.postMessage({ id: promiseId, type: TRANSPORT.GET_INFO });
+            const response = await promise;
+            this._log.debug('Bridge availability response', response);
+            if (response.payload === undefined) {
+                throw ERRORS.TypedError('Transport_Missing');
+            }
+        }
+
         // sharedLogger can be disable but it is enable by default.
         if (this._settings.sharedLogger !== false) {
             // connect-web is running in third-party domain so we use iframe to pass logs to shared worker.
@@ -193,19 +204,13 @@ export class CoreInIframe implements ConnectFactoryDependencies {
             if (!this._popupManager) {
                 this._popupManager = this.initPopupManager();
             }
-            this._popupManager.request();
 
             // auto init with default settings
             try {
                 await this.init(this._settings);
             } catch (error) {
                 if (this._popupManager) {
-                    // Catch fatal iframe errors (not loading)
-                    if (['Init_IframeBlocked', 'Init_IframeTimeout'].includes(error.code)) {
-                        this._popupManager.postMessage(createUiMessage(UI.IFRAME_FAILURE));
-                    } else {
-                        this._popupManager.clear();
-                    }
+                    this._popupManager.clear();
                 }
 
                 return createErrorMessage(error);
--- a/packages/connect-web/src/impl/core-in-popup.ts
+++ b/packages/connect-web/src/impl/core-in-popup.ts
@@ -130,26 +130,26 @@ export class CoreInPopup implements ConnectFactoryDependencies {
             await this._popupManager.request();
         }
 
-        await this._popupManager.channel.init();
-
-        if (this._settings.env === 'webextension') {
-            // In webextension we init based on the popup promise
-            // In core this is handled in the popup manager
-            await this._popupManager.popupPromise?.promise;
-
-            this._popupManager.channel.postMessage({
-                type: POPUP.INIT,
-                payload: {
-                    settings: this._settings,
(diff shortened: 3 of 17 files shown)


packages/connect-web/src/utils/proxy-event-emitter.ts
```diff
@@ -0,0 +1,89 @@
+import EventEmitter from 'events';
+
+export default class ProxyEventEmitter implements EventEmitter {
```
mroz22, line 7: maybe please add a comment why is this needed, I remember we were discussing this and it made sense back then but after 3 weeks I am completely lost :D

martykan: It can basically combine N event emitters into 1, for the different core in X implementations

packages/connect-explorer/src/pages/settings.mdx
```diff
@@ -52,6 +53,17 @@ export const Settings = () => {
             key: 'useCoreInPopup',
```
mroz22, line 54: maybe we could get rid of it already? I understand that this requires some work in tests...

packages/connect-web/src/utils/proxy-event-emitter.ts
```diff
@@ -0,0 +1,93 @@
...
+        return this.eventEmitters[0].listenerCount(eventName, listener);
+    }
+
+    rawListeners(eventName: string | symbol) {
+        return this.eventEmitters[0].rawListeners(eventName);
+    }
+
+    listeners(eventName: string | symbol) {
+        return this.eventEmitters[0].listeners(eventName);
+    }
```
mroz22, line 94: these are there only for types compatibility?

mroz22:
this can be probably merged separately? 6252066d173d09fd525fd71ae4601c02ccd87172 also do we have a way how to automatically detect bundling stuff we don't want to bundle? (I know I asked this some time ago but I forgot) :/

packages/connect-web/src/utils/proxy-event-emitter.ts
martykan: Mostly yes but some of these methods are actually used.

packages/connect-web/src/impl/core-in-iframe.ts
```diff
@@ -174,6 +173,18 @@ export class CoreInIframe implements ConnectFactoryDependencies {
...
+            console.log('Checking bridge availability'); // DEBUG
+            const { promiseId, promise } = this._messagePromises.create();
+            iframe.postMessage({ id: promiseId, type: TRANSPORT.GET_INFO });
+            const response = await promise;
+            console.log('Bridge availability response', response);
+            if (response.payload === undefined) {
+                throw ERRORS.TypedError('Transport_Missing');
+            }
+        }
+
```
mroz22: I am still thinking about the race-condition scenario. not sure if we were able to rule it out or not. the worst thing that can happen is that under mode `auto` we will switch to popup mode even when it is not necessary, right?

packages/connect-explorer/src/pages/settings.mdx
martykan: Yes, but I think we will actually have to leave it internally, for compatibility with older `connect-webextension` for example

packages/connect-web/src/impl/core-in-iframe.ts
```diff
@@ -193,19 +204,13 @@ export class CoreInIframe implements ConnectFactoryDependencies {
...
                 this._popupManager = this.initPopupManager();
             }
-            this._popupManager.request();
 
             // auto init with default settings
             try {
                 await this.init(this._settings);
             } catch (error) {
                 if (this._popupManager) {
-                    // Catch fatal iframe errors (not loading)
```
mroz22, line 203: cool, so this means that these errors are no longer fatal and we can retry usign a different way

packages/connect-web/src/impl/core-in-iframe.ts
martykan: Yes, nothing too bad should happen

martykan: But I actually changed `transportReconnect` to wait in case of `auto` mode

packages/connect-web/src/impl/core-in-popup.ts
```diff
@@ -130,26 +130,26 @@ export class CoreInPopup implements ConnectFactoryDependencies {
...
+                    },
+                });
+            }
 
-        await this._popupManager.handshakePromise?.promise;
+            await this._popupManager.handshakePromise?.promise;
 
-        // post message to core in popup
-        try {
+            // post message to core in popup
```
mroz22, line 152: what is the reason for this change?

packages/connect-web/src/utils/proxy-event-emitter.ts
```diff
@@ -0,0 +1,93 @@
+import EventEmitter from 'events';
+
+/**
+ * ProxyEventEmitter is an EventEmitter that allows to use multiple EventEmitters as one
+ * This is used in connect-web to allow switching between iframe and core-in-popup implementations
+ */
+export default class ProxyEventEmitter implements EventEmitter {
```
mroz22, line 7: cc @marekrjpolak 

this is a "utilitish code". at the moment I am not seeing any other usecase for it that would advocate moving this into packages/utils. still wanted to ping you about it if you have any opinion/idea

packages/connect/src/data/connectSettings.ts
```diff
@@ -137,6 +137,13 @@ export const parseConnectSettings = (input: Partial<ConnectSettings> = {}) => {
         settings.useCoreInPopup = input.useCoreInPopup;
     }
 
+    if (
```
mroz22, line 140: more like a note for me: 
btw defaults are set only on the 'front-end' level?

packages/connect-web/src/utils/proxy-event-emitter.ts
marekrjpolak: I'd leave it as it is for now, and potentially move it to `@trezor/utils` as soon as we want to use something similar for the 2nd time.

packages/connect-web/src/impl/core-in-popup.ts
martykan: We want to catch all errors and return `createErrorMessage`

packages/connect-web/src/utils/proxy-event-emitter.ts
mroz22: ok pleease add a comment that they have everything the same

packages/connect-web/src/impl/core-in-iframe.ts
```diff
@@ -174,6 +173,18 @@ export class CoreInIframe implements ConnectFactoryDependencies {
 
         await iframe.init(this._settings);
 
+        if (this._settings.coreMode === 'auto') {
+            // Checking bridge availability
+            console.log('Checking bridge availability'); // DEBUG
```
mroz22: lets replace console.log with  this._log.debug/info

mroz22 (approved):
good, just some left behind console.logs

Hannsek:
Why auto is not set as a default?

martykan:
To keep the default behavior the same as before, so we don't accidentally break anything. 
The goal is to make it default, but after testing it as opt-in

Hannsek:
I'd push for making it default asap, as I doubt that anyone will change it to auto manually

mroz22:
lets start experimenting with it for the next release

PR #13042, opened 2024-06-24
chore(connect): type only exports

martykan:
## Description

Enforce type only imports/exports to avoid unneeded schemas getting into bundles


mroz22:
where it was bundled to where we didnt want it to happen? also there is a stupid script that was supposed to catch problems like this. wasn't it enough? https://github.com/trezor/trezor-suite/blob/develop/packages/connect-web/scripts/check-inline-build-size.sh#L9

martykan:
The inline build is fine due to our tree shaking, but if you import `connect-web` as a module in some external project and use default webpack settings, you will get a quite big bundle. 
This is just some initial exploration on how to improve this, still WIP...

PR #13050, opened 2024-06-24
fix(connect-webextension): firefox mv3

martykan:
## Description

In Firefox, there are additional security protections that cause the following error with our content script: 

```
Not allowed to define cross-origin object as property on [Object] or [Array] XrayWrapper
```

This is solved by cloning objects that are passed between content script and the page. 

Followup: make sure our Explorer example works on Firefox

### DIFF
--- a/packages/connect-explorer/src-webextension/manifest.json
+++ b/packages/connect-explorer/src-webextension/manifest.json
@@ -7,6 +7,7 @@
     },
     "background": {
         "service_worker": "serviceWorker.bundle.js",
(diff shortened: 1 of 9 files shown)


mroz22 (approved):
uf, do you have some read on this?

karliatto:
@martykan I am curious about your comment `Followup: make sure our Explorer example works on Firefox` 
Does it mean that even with the changes proposed in this PR our MV3 does not work with Firefox?

mroz22 (approved):

PR #13059, opened 2024-06-25
fix(suite): Polish view only according to design review (2)

jvaclavik:
## Description

https://www.figma.com/design/wGshbwWuR7shAA1i1UBqR8/branch/1O5PSi8KzMpeVyFMrEMZeA/Remember-Mode?node-id=3371-39940&t=3snaAns3xx0dTNGd-0

## Related Issue

Resolve

### DIFF
--- a/packages/components/src/components/Passphrase/EnterOnTrezorButton.tsx
+++ b/packages/components/src/components/Passphrase/EnterOnTrezorButton.tsx
@@ -1,16 +1,15 @@
 import { FormattedMessage } from 'react-intl';
-import styled from 'styled-components';
+import styled, { useTheme } from 'styled-components';
 import { Image } from '../Image/Image';
 import { DeviceModelInternal } from '@trezor/connect';
 import { Card } from '../Card/Card';
 import { Icon } from '../assets/Icon/Icon';
-import { spacingsPx } from '@trezor/theme';
+import { Text } from '../typography/Text/Text';
+import { spacings } from '@trezor/theme';
+import { Row } from '../Flex/Flex';
 
-const Row = styled.div`
-    display: flex;
-    gap: ${spacingsPx.xs};
-    align-items: center;
-    justify-content: space-between;
+const Flex = styled.div`
+    flex: 1;
 `;
 
 interface EnterOnTrezorButtonProps {
@@ -20,19 +19,25 @@ interface EnterOnTrezorButtonProps {
 }
 
 export const EnterOnTrezorButton = ({ submit, value, deviceModel }: EnterOnTrezorButtonProps) => {
+    const theme = useTheme();
+
     return (
         <Card
             paddingType="small"
             onClick={() => submit(value, true)}
             data-test="@passphrase/enter-on-device-button"
         >
-            <Row>
+            <Row gap={spacings.lg} alignItems="center" justifyContent="space-between">
                 {deviceModel && <Image alt="Trezor" image={`TREZOR_${deviceModel}`} height={34} />}
-                <FormattedMessage
-                    id="TR_ENTER_PASSPHRASE_ON_DEVICE"
-                    defaultMessage="Enter passphrase on Trezor"
-                />
-                <Icon icon="ARROW_RIGHT" />
+                <Flex>
+                    <Text variant="tertiary">
+                        <FormattedMessage
+                            id="TR_ENTER_PASSPHRASE_ON_DEVICE"
+                            defaultMessage="or enter on Trezor"
+                        />
+                    </Text>
+                </Flex>
+                <Icon icon="ARROW_RIGHT" color={theme.iconSubdued} />
             </Row>
         </Card>
     );
--- a/packages/connect-ui/src/views/Passphrase.tsx
+++ b/packages/connect-ui/src/views/Passphrase.tsx
@@ -100,7 +100,7 @@ export const Passphrase = (props: PassphraseProps) => {
                         title={
                             <FormattedMessage
                                 id="TR_WALLET_SELECTION_HIDDEN_WALLET"
-                                defaultMessage="Hidden wallet"
+                                defaultMessage="Passphrase wallet"
                             />
                         }
                         description={
@@ -112,7 +112,7 @@ export const Passphrase = (props: PassphraseProps) => {
                         submitLabel={
                             <FormattedMessage
                                 id="TR_WALLET_SELECTION_ACCESS_HIDDEN_WALLET"
-                                defaultMessage="Access Hidden wallet"
+                                defaultMessage="Access Passphrase wallet"
                             />
                         }
                         type="hidden"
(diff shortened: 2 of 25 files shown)


packages/components/src/components/Passphrase/EnterOnTrezorButton.tsx
```diff
@@ -20,19 +19,25 @@ interface EnterOnTrezorButtonProps {
...
-                <FormattedMessage
-                    id="TR_ENTER_PASSPHRASE_ON_DEVICE"
-                    defaultMessage="Enter passphrase on Trezor"
-                />
-                <Icon icon="ARROW_RIGHT" />
+                <Flex>
+                    <Text variant="tertiary">
+                        <FormattedMessage
+                            id="TR_ENTER_PASSPHRASE_ON_DEVICE"
+                            defaultMessage="or enter on Trezor"
```
peter-sanderson, line 36: I am not sure if we need this `defaultMessage` here at all. To me it seems as relict of the past.

Maybe @komret would know? Can we just delete all places and rely on default messages in `messages.ts`M

suite-common/suite-types/src/device.ts
```diff
@@ -40,7 +40,7 @@ export interface ExtendedDevice {
     ts: number;
     buttonRequests: ButtonRequest[];
     metadata: DeviceMetadata;
-    walletNumber?: number; // number of hidden wallet intended to be used in UI
```
peter-sanderson, line 43: here from `hidden` to `passphrase`, but above you did from `passphrase` to `hidden`? why what is correct naming now?

peter-sanderson: Here for example: https://github.com/trezor/trezor-suite/pull/13059/files#diff-2db736617f53aec5e3acbfeddb95cf37f79cf3bbd8f217da1d81d91fff13bf16L99

packages/connect-ui/src/views/Passphrase.tsx
```diff
@@ -100,7 +100,7 @@ export const Passphrase = (props: PassphraseProps) => {
                         title={
                             <FormattedMessage
                                 id="TR_WALLET_SELECTION_HIDDEN_WALLET"
-                                defaultMessage="Hidden wallet"
+                                defaultMessage="Passphrase wallet"
```
peter-sanderson, line 103: I think important is to edit `messages.ts` and not in-code stuff,

packages/suite/src/components/suite/WebUsbButton.tsx
```diff
@@ -1,30 +1,54 @@
...
         </Button>
     </div>
 );
+
+export const WebUsbIconButton = ({
+    translationId = 'TR_CHECK_FOR_DEVICES',
+    size = 'tiny',
+    variant = 'primary',
+    ...rest
+}: WebUsbButtonProps) => {
```
peter-sanderson: nit: inline function

packages/components/src/components/Passphrase/EnterOnTrezorButton.tsx
komret: `defaultMessage` is necessary here because `@trezor/components` package doesn't have access to the `Translation` component from `@trezor/suite` and doesn't see the strings in `messages.ts`. But I am not sure how exactly it works and @mroz22 doesn't remember either. Surely it can be improved.

jvaclavik: So is it safe to delete the whole `defaultMessage`?

packages/suite/src/components/suite/WebUsbButton.tsx
peter-sanderson: I created a proposal for this: https://github.com/trezor/trezor-suite/issues/13075

packages/components/src/components/Passphrase/EnterOnTrezorButton.tsx
peter-sanderson: Thanks @komret got it. 

@jvaclavik no, we have to keep it here if I understand correctly. 

However, this seem that this component shall not be in component package  at all. Why its there? I see its used only in Suite. Lets move it there then. Or?

packages/connect-ui/src/views/Passphrase.tsx
peter-sanderson: @jvaclavik this is `connect` (explorer?) Not sure how what Komret wrote applies here. So feel free to ignore my comment

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseList.tsx
```diff
@@ -1,16 +1,16 @@
 import { spacingsPx, typography } from '@trezor/theme';
 import styled from 'styled-components';
 
-export const PassphraseList = styled.div`
+export const PassphraseList = styled.div<{ $gap?: 'small' | 'normal' }>`
```
peter-sanderson: nit: I prefer extra type definitions. 

(reason even in this case: it tends to break the VSCode code-colorization, dont know why but extracting type helps)

packages/suite/src/views/suite/SwitchDevice/DeviceItem/DeviceStatusText.tsx
```diff
@@ -17,21 +17,23 @@ type DeviceStatusVisible = {
...
-        {walletLabel ? (
-            <TruncateWithTooltip delayShow={TOOLTIP_DELAY_LONG}>
-                <WalletLabeling device={device} />
-            </TruncateWithTooltip>
-        ) : (
-            <Translation id={connected ? 'TR_CONNECTED' : 'TR_DISCONNECTED'} />
-        )}
-    </DeviceConnectionText>
-);
+const DeviceStatusVisible = ({ connected, walletLabel }: DeviceStatusVisible) => {
```
peter-sanderson: nit: inline arrow function

packages/suite/src/views/view-only/ViewOnlyTooltip.tsx
```diff
@@ -1,27 +1,50 @@
...
+const Icons = styled.div`
+    display: flex;
+    flex-direction: row;
+`;
+
+const IconContainer = styled.div`
+    border: solid 1px ${palette.lightWhiteAlpha400};
+    background-color: ${palette.darkGray300};
+    border-radius: ${borders.radii.full};
+    padding: 4px;
```
peter-sanderson, line 22: is this 4 connected to this: 

image


If yes lets create some constant that explains it.

Or use `spacings` constants? Not sure if it makes semantically sense here.

peter-sanderson (approved):
Just cosmetics :shipit:

suite-common/suite-types/src/device.ts
jvaclavik: That is not true 🤷‍♂️😀 it should be hidden -> passphrase everywhere, because hidden is deprecated now

peter-sanderson: yea... sorry :D

packages/suite/src/views/view-only/ViewOnlyTooltip.tsx
jvaclavik: Yes, it's correct and I think it's not necessary to use spacings here until we have minus values.

PR #13067, opened 2024-06-25
feat(suite-native): passphrase on device only when possible

### LINKED ISSUE #12703: Enter passphrase on Trezor not offered on T1
On passphrase form, `Enter on Trezor` should only be available for devices that enable this functionality and T1 is not one of them. So we should remove it for that device.

vytick:
Only let user insert passphrase on device when supported by actual device

## Related Issue

Resolve #12703

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/f3cfffc8-f481-47fd-9a0c-2192dcbe3bc0

### DIFF
--- a/suite-common/wallet-core/src/device/deviceReducer.ts
+++ b/suite-common/wallet-core/src/device/deviceReducer.ts
@@ -709,6 +709,12 @@ export const selectDeviceByState = (state: DeviceRootState, deviceState: string)
 export const selectDeviceUnavailableCapabilities = (state: DeviceRootState) =>
     state.device.selectedDevice?.unavailableCapabilities;
 
+export const selectDeviceCapabilities = (state: DeviceRootState) =>
+    selectDeviceFeatures(state)?.capabilities;
+
+export const selectHasDevicePassphraseEntryCapability = (state: DeviceRootState) =>
+    !!selectDeviceCapabilities(state)?.includes('Capability_PassphraseEntry');
+
 export const selectDeviceStatus = (state: DeviceRootState) => {
     const device = selectDevice(state);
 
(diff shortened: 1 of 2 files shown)


suite-native/module-passphrase/src/components/PassphraseForm.tsx
```diff
@@ -78,6 +78,10 @@ export const PassphraseForm = ({ inputLabel, onFocus }: PassphraseFormProps) =>
         reset();
     });
 
+    const offerPassphraseOnDevice = !!device?.features?.capabilities?.includes(
+        'Capability_PassphraseEntry',
+    );
```
juriczech: Please get features using `selectDeviceFeatures` or better yet, create a selector for this.

vytick: 7fecf6c399

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -709,6 +709,12 @@ export const selectDeviceByState = (state: DeviceRootState, deviceState: string)
 export const selectDeviceUnavailableCapabilities = (state: DeviceRootState) =>
     state.device.selectedDevice?.unavailableCapabilities;
 
+export const selectDeviceCapabilities = (state: DeviceRootState) =>
+    selectDeviceFeatures(state)?.capabilities;
+
+export const selectHasDevicePassphraseEntryCapability = (state: DeviceRootState) =>
```
juriczech, line 715:
```suggestion
export const selectDeviceHasPassphraseEntryCapability = (state: DeviceRootState) =>
```

I think this one is correct. But I'm not sure. Leaving it as a nitpick, nothing critical

juriczech: If you change it, please also change variable name in component as well

juriczech (approved):

suite-common/wallet-core/src/device/deviceReducer.ts
vytick: I think this form is better due to clarity of what is the selector result and it is in line with similar selectors that are also returning boolean like:

`selectIsDeviceBitcoinOnly`
`selectHasDeviceFirmwareInstalled`
`selectIsDeviceConnected`
`selectIsDeviceInViewOnlyMode`

I thought about the form you are suggesting before, but this one feels more spot on to me.

PR #13068, opened 2024-06-25
fix(suite-native): feature gate for view only setting

### LINKED ISSUE #13066: View-only settings page should be hidden on iOS
View-only settings section should not be visible at all on iOS.

Use IsDeviceConnectEnabled for it

vytick:
Gate View Only settings by isUsbDeviceConnectFeatureEnabled

## Related Issue

Resolve [#13066](https://github.com/trezor/trezor-suite/issues/13066)

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/5fecfc7f-76f6-4e68-9e85-e3f4295e63d9


matejkriz (approved):
I haven't tested it, but LGTM

PR #13073, opened 2024-06-25
feat(suite-native): passphrase loading has subtitle

### LINKED ISSUE #12719: Longer indication label for passphrase discovery
We need to indicate that the passphrase discovery might take a while. 

For this purpose, we'll add a smaler text below the label when loading. 

Link to figma [here](https://www.figma.com/design/mmDnYtctfjnreWhTjLfeQH/Passphrase?node-id=2350-73720&t=zSlSsyD2w6IXuXa3-11)

vytick:
Add subtitle to passphrase loading

## Related Issue

Resolve #12719

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/0ed828c9-92b2-4d8b-997b-822afd0fbdbd

### DIFF
--- a/suite-native/intl/src/en.ts
+++ b/suite-native/intl/src/en.ts
@@ -673,7 +673,10 @@ export const en = {
             title: 'Continue on Trezor',
             subtitle: 'Enter your passphrase on your Trezor',
         },
-        loadingTitle: 'Checking passphrase wallet for balances & transactions',
+        loading: {
+            title: 'Checking passphrase wallet for balances & transactions',
+            subtitle: 'This might take up to a minute.',
+        },
         confirmOnDevice: {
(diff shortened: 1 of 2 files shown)


suite-native/intl/src/en.ts
```diff
@@ -696,7 +699,7 @@ export const en = {
                 button: 'Try again',
             },
             confirmEmptyWalletSheet: {
-                title: 'Passphrase best practises',
+                title: 'Passphrase best practices',
```
juriczech, line 702: Good catch! ❤️

juriczech (approved):

suite-native/intl/src/en.ts
vytick: It was not me, but that VSC extension PeKne suggested :)

PR #13080, opened 2024-06-26
feat(coinmarket): best rate selecting

adderpositive:
## Info
A tag `Best rate` is added only to offers with the highest amount the client receives

### DIFF
--- a/packages/suite/src/hooks/wallet/coinmarket/offers/useCoinmarketSellOffers.ts
+++ b/packages/suite/src/hooks/wallet/coinmarket/offers/useCoinmarketSellOffers.ts
@@ -21,8 +21,9 @@ import {
 import { goto } from 'src/actions/suite/routerActions';
 import { useCoinmarketNavigation } from 'src/hooks/wallet/useCoinmarketNavigation';
 import {
+    addIdsToQuotes,
+    filterQuotesAccordingTags,
     getUnusedAddressFromAccount,
-    processSellAndBuyQuotes,
 } from 'src/utils/wallet/coinmarket/coinmarketUtils';
 import type { TradeSell } from 'src/types/wallet/coinmarketCommonTypes';
 import { useBitcoinAmountUnit } from 'src/hooks/wallet/useBitcoinAmountUnit';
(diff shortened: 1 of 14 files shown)


packages/suite/src/utils/wallet/coinmarket/coinmarketUtils.ts
```diff
@@ -273,3 +276,24 @@ export function processSellAndBuyQuotes<T extends CoinmarketTradeBuySellType>(
...
+    const getTypedAmount = (quote: CoinmarketTradeDetailType) => {
+        if (type === 'sell') {
+            return new BigNumber((quote as SellFiatTrade).fiatStringAmount ?? 0);
+        }
+
+        return new BigNumber((quote as BuyTrade | ExchangeTrade).receiveStringAmount ?? 0);
+    };
+
+    const bestRatedQuotes = quotes
+        ? quotes.toSorted((a, b) => getTypedAmount(b).minus(getTypedAmount(a)).toNumber())
```
FreeWall: is `toSorted` supported in all browsers suite supports?

adderpositive: `sort` will be safer! Thanks

packages/suite/src/hooks/wallet/coinmarket/offers/useCoinmarketSellOffers.ts
```diff
@@ -82,7 +83,9 @@ export const useCoinmarketSellOffers = ({
 
                 setInnerQuotes(
                     getFilteredSuccessQuotes<CoinmarketTradeSellType>(
-                        processSellAndBuyQuotes<CoinmarketTradeSellType>(allQuotes),
+                        filterQuotesAccordingTags<CoinmarketTradeSellType>(
+                            addIdsToQuotes<CoinmarketTradeSellType>(allQuotes, 'sell'),
+                        ),
                     ),
                 );
```
LukasRada: Please extract function result into const variables and use them accordingly.

packages/suite/src/utils/wallet/coinmarket/coinmarketUtils.ts
```diff
@@ -257,19 +261,51 @@ export const getDefaultCountry = (country: string = regional.unknownCountry) =>
...
+              }
+
+              // descending to rate for sell/exchange - higher rate more crypto/fiat client receives
+              return new BigNumber(b.rate ?? 0).minus(new BigNumber(a.rate ?? 0)).toNumber();
+          })
+        : null;
+    const bestRatedQuote = bestRatedQuotes?.[0];
+
+    return bestRatedQuote;
+};
```
LukasRada, line 327: Maybe move the logic to server? cc: @FreeWall

LukasRada (changes requested):

packages/suite/src/utils/wallet/coinmarket/coinmarketUtils.ts
FreeWall: Best rated quote (or `quotes` array in this case) depends on client-side filtering (payment method), so it's calculated on client-side.
In order to use logic on server side, we would need to calculate best rate for each payment method and pass the info to client somehow.

Issue #12487, opened 2024-05-20
Create a list of places in Suite, Suite Web, Connect popup where Trezor Bridge is mentioned


Hannsek:
Please **include pictures** and **describe in which situation** the message about Trezor Bridge is shown.

Please put your findings [here](https://www.notion.so/satoshilabs/Trezor-Bridge-deprecation-6080c7b16f1f41c6b2d8f8f38315cea4?pvs=4#b828f07e4cd746ecb32ddb8bfdf77ba5)

Hannsek:
Two columns (old | new):
**Connect popup**
![image](https://github.com/trezor/trezor-suite/assets/57008159/d1d0b9f0-9400-4c33-9dc5-2d96c658947a)
- delete _install bridge_ modal/notification

_Suite web_
![image](https://github.com/trezor/trezor-suite/assets/57008159/dd9960e5-5a9f-4bd8-83d4-120d0381f1a6)
![image](https://github.com/trezor/trezor-suite/assets/57008159/6da1ea88-757a-4b12-a508-7f3e6b4e6eab)
![image](https://github.com/trezor/trezor-suite/assets/57008159/fb02395e-b24e-47af-92b5-3d2ddd75420c)

_Suite desktop_
![image](https://github.com/trezor/trezor-suite/assets/57008159/9721f377-4508-4ade-8e7c-b92812d9f08b)
_note: copy of the last bullet point still in progress_

PR #13090, opened 2024-06-28
feat(suite-desktop-core): deeplink and basic styling to http-receiver…

mroz22:
Wherever we are promoting bridge download (see #12487) we will want to show a deeplink to open Suite instead. While doing some preliminary investigations I bumped into some related low-hanging fruit. 

Before:

image

After:

https://github.com/trezor/trezor-suite/assets/30367552/145cad57-3a81-4a2f-b112-d1500beb76a8

Leaving this to more UI-wise skilled ppl, maybe @komret or @jvaclavik.
This also affects Invity screens, so if someone please could take a look :pray:

### DIFF
--- a/packages/suite-desktop-core/src/libs/http-receiver.ts
+++ b/packages/suite-desktop-core/src/libs/http-receiver.ts
@@ -2,6 +2,7 @@ import * as url from 'url';
 
 import { xssFilters } from '@trezor/utils';
 import { HttpServer, allowReferers } from '@trezor/node-utils';
+import { trezorLogo } from '@suite-common/suite-constants';
 
 import { HTTP_ORIGINS_DEFAULT } from './constants';
 import { convertILoggerToLog } from '../utils/IloggerToLog';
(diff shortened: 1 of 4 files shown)


jvaclavik:
Looks good to me, maybe just one nitpick, I'd distinguish "Go back to trezor" somehow, because it's a link. `font-weight: bold` should be enough.

jvaclavik (approved):

packages/suite-desktop-core/src/libs/http-receiver.ts
```diff
@@ -21,16 +22,42 @@ interface Events {
...
+                        display: inline-flex;
+                        align-items: center;
+                    }
+                    a:hover {
+                      text-decoration: underline;
+                    }
+                </style>
             </head>
             <body>
+                <img style="margin-bottom:40px" src="data:image/png;base64, ${trezorLogo}" />
```
adderpositive: Missing `alt` attribute.

adderpositive (approved):

packages/suite-desktop-core/src/libs/http-receiver.ts
mroz22: added, rebased and forcepushed

PR #13091, opened 2024-06-28
Killing old bridge

### LINKED ISSUE #9356: Stop offering Bridge when possible
We agreed to push webusb where possible. We should no longer advertise Bridge as a substitute in webusb-capable browsers. We should suggest installing desktop version instead.

On Chrome:
image

To be honest I am not sure we should offer it in Firefox either, I would strongly recommend to download Suite Desktop for those that have nothing installed.

### LINKED ISSUE #12487: Create a list of places in Suite, Suite Web, Connect popup where Trezor Bridge is mentioned

mroz22:
Updated messages in Prerequisites Guide
Screenshot 2024-07-18 at 14 14 22
Replaced Bridge modal
Screenshot 2024-07-18 at 14 21 21
New modal that can be deep-linked into from web suite/Connect to start the bridge
Screenshot 2024-07-17 at 10 12 11

## Related issues
- Resolves #9356
- Resolves #12487

### DIFF
--- a/packages/connect-ui/src/components/Notification.tsx
+++ b/packages/connect-ui/src/components/Notification.tsx
@@ -3,7 +3,7 @@ import { useState } from 'react';
 import styled from 'styled-components';
 
 import { Icon, Button } from '@trezor/components';
-import { SUITE_FIRMWARE_URL, SUITE_BRIDGE_URL, SUITE_BACKUP_URL } from '@trezor/urls';
+import { SUITE_FIRMWARE_URL, SUITE_URL, SUITE_BACKUP_URL } from '@trezor/urls';
 
 const NotificationBox = styled.div`
     background-color: ${({ color }) => color};
@@ -134,9 +134,9 @@ export const BackupNotification = () => (
 export const BridgeUpdateNotification = () => (
     <Notification
         variant="warning"
-        header="New bridge is available"
-        body=""
-        cta={{ desc: 'Update bridge', url: SUITE_BRIDGE_URL }}
+        header="Current bridge is outdated"
+        body="Trezor Bridge is no longer supported. Switch to Trezor Suite desktop for the best experience."
+        cta={{ desc: 'Install Trezor Suite', url: SUITE_URL }}
     />
 );
 
--- a/packages/connect-ui/src/views/Transport.tsx
+++ b/packages/connect-ui/src/views/Transport.tsx
@@ -1,24 +1,58 @@
-import { Button, Link, Image } from '@trezor/components';
-import type { UiEvent } from '@trezor/connect';
-import { SUITE_BRIDGE_URL } from '@trezor/urls';
+import { FormattedMessage } from 'react-intl';
+
+import { Button, Image } from '@trezor/components';
+import { type UiEvent } from '@trezor/connect';
+import { SUITE_URL, SUITE_BRIDGE_DEEPLINK } from '@trezor/urls';
+import { useWindowFocus } from '@trezor/react-utils';
 
 import { View } from '../components/View';
 import imageSrc from '../images/man_with_laptop.svg';
 
 export type TransportEventProps = Extract<UiEvent, { type: 'ui-no_transport' }>;
 
-export const Transport = () => (
-    <View
-        title="Install Bridge"
-        description="The Bridge is a communication tool, which facilitates the connection between
-        your Trezor device and your internet browser."
-        image={<Image imageSrc={imageSrc} />}
-        buttons={
-            <Link variant="nostyle" href={SUITE_BRIDGE_URL} onClick={() => window.closeWindow()}>
-                <Button variant="primary" icon="EXTERNAL_LINK" iconAlignment="right">
-                    Install Bridge
-                </Button>
-            </Link>
-        }
-    />
-);
+export const Transport = () => {
+    const windowFocused = useWindowFocus();
+    const handleOpenSuite = () => {
+        location.href = SUITE_BRIDGE_DEEPLINK;
+
+        // fallback in case deeplink does not work
+        window.setTimeout(() => {
+            if (!windowFocused.current) return;
+
+            window.open(SUITE_URL);
+        }, 500);
+    };
+
+    return (
+        <View
+            title={
+                <FormattedMessage
+                    id="TR_NO_TRANSPORT"
+                    defaultMessage="Browser can't communicate with device"
+                />
+            }
+            description={
+                <FormattedMessage
+                    id="TR_BRIDGE_NEEDED_DESCRIPTION"
+                    defaultMessage="We recommend downloading and running the Trezor Suite desktop app in the background for the best experience. Alternatively, use a supported browser that is compatible with WebUSB"
+                />
+            }
+            image={<Image imageSrc={imageSrc} />}
+            buttons={
+                <>
+                    <Button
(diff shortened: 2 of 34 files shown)


martykan:
It would be nice if we got some design for the bridge popup, maybe with some icon/simple illustration

packages/connect-ui/src/components/Notification.tsx
```diff
@@ -134,9 +134,9 @@ export const BackupNotification = () => (
 export const BridgeUpdateNotification = () => (
     <Notification
         variant="warning"
-        header="New bridge is available"
-        body=""
-        cta={{ desc: 'Update bridge', url: SUITE_BRIDGE_URL }}
+        header="Current bridge is outdated"
+        body="Trezor Bridge is no longer supported. Switch to Trezor Suite desktop for the best experience."
+        cta={{ desc: 'Install Trezor Suite', url: SUITE_URL }}
```
mroz22, line 139: note to me: check conditions when this is rendered.

packages/connect-ui/src/views/Transport.tsx
```diff
@@ -1,24 +1,41 @@
...
-                <Button variant="primary" icon="EXTERNAL_LINK" iconAlignment="right">
-                    Install Bridge
-                </Button>
-            </Link>
-        }
-    />
-);
+export const Transport = () => {
+    const handleOpenSuite = () => {
+        // todo: can we somehow check if the app is installed (not running ) and present a deeplink in that case?
```
mroz22: this is probably not possible I guess :(

packages/suite/src/components/onboarding/ConnectDevicePromptManager/ConnectDevicePromptManager.tsx
```diff
@@ -64,6 +64,8 @@ export const ConnectDevicePromptManager = ({
     // todo: move to some hook or something
 
     if (!transport?.type) {
+        // todo: onboarding has a different NoTransport component?
+
```
mroz22: there is couple of semi-duplicated components. its a mess.

packages/connect-ui/src/views/Transport.tsx
martykan: I've seen this work on mobile. I have some ideas - try the deeplink, wait some period of time, then check via the HTTP server if it's alive. If not go to the download page. The only problem in this case is the necessary delay...

packages/connect-ui/src/components/Notification.tsx
martykan: It seems to be checking the version from here https://connect.trezor.io/9/data/bridge/releases.json

packages/connect-ui/src/views/Transport.tsx
mroz22: 🤔 what if I don't have the app registered for the deeplink but it is already installed? handling based on focus? until user focuses this page again, we show some spinner and expected that he is clicking somewhere in native dialogues?

packages/connect-ui/src/components/Notification.tsx
mroz22: yep, this is there because of some ancient bridge versions, right. 

this might be, maybe, the place to start experimenting with some copy or even prompt to uninstall.

packages/connect-ui/src/views/Transport.tsx
martykan: Focus sounds like it could be viable. On Mac it seems like the deeplinks work "automagically", not sure about other platforms.

mroz22:
we should probably also:

- implement openAtLogin https://www.electronjs.org/docs/latest/api/app#appsetloginitemsettingssettings-macos-windows
- if app was openedAtLogin - make sure that it is [not grabbing devices ](https://github.com/trezor/trezor-suite/blob/kill-bridge/packages/connect/src/device/DeviceList.ts#L198) as the first application. trezor-suite should be the first class citizen but not in this case.
- once user started actively using trezor-suite, it should become the first class citizen again.

packages/urls/src/urls.ts
```diff
@@ -15,6 +15,7 @@ export const SUITE_BACKUP_URL = 'https://suite.trezor.io/web/backup/';
 export const SUITE_BRIDGE_URL = 'https://suite.trezor.io/web/bridge/';
 export const SUITE_FIRMWARE_URL = 'https://suite.trezor.io/web/firmware/';
 export const SUITE_UDEV_URL = 'https://suite.trezor.io/web/udev/';
+export const SUITE_BRIDGE_DEEPLINK = 'trezorsuite://bridge-requested-by-a-3rd-party';
```
mroz22: this list was intended for urls accessible over the internet. there is a nightly check that those links return some non 4** response

martykan: Where should we put it? Maybe `src/constants/suite/protocol`?

mroz22: can every part that needs it import from there? If yes, it sounds good to me

martykan: Oh, probably not, we'll need it in Connect. So maybe a new file in urls package for deeplinks?

mroz22: good choice!

packages/connect-ui/src/views/Transport.tsx
```diff
@@ -1,24 +1,43 @@
...
+        window.setTimeout(() => {
+            if (!windowFocused.current) return;
+
+            window.open(SUITE_URL);
+        }, 500);
+    };
+
+    return (
+        <View
+            title="Device connection is not available"
```
mroz22: maybe lets make these lines translatable as well?

packages/suite/src/views/suite/bridge/index.tsx
```diff
@@ -31,200 +15,75 @@ const StyledButton = styled(Button)`
...
+                </Footer>
+            </Modal>
+        );
+    }
+
     return (
         <Modal
-            heading={<Translation id="TR_TREZOR_BRIDGE_DOWNLOAD" />}
-            description={<Translation id="TR_NEW_COMMUNICATION_TOOL" />}
+            heading="Bridge"
```
mroz22: make them translatable? I would suggest that.

mroz22 (commented):
looks good, how much we want to discuss/test it before merging? 

@Hannsek @MiroslavProchazka @bosomt ?

MiroslavProchazka:
> looks good, how much we want to discuss/test it before merging?
> 
> @Hannsek @MiroslavProchazka @bosomt ?

My suggestion: 
24.7.3 (next week) - 50% EAP 
24.8.x - 100% EAP 
24.9.x - 30% users 
24.10 - 100% 

Thoughts?

Hannsek:
I'd be more pushy in this. 
In 24.8.x let's put 100 % EAP and 10 % users ?

> looks good, how much we want to discuss/test it before merging?

Also I'd like to discuss the copy on Monday.

martykan (approved):

PR #13092, opened 2024-06-28
Fix/tokens

tomasklim:
## Description

- add `ETC` token definitions, they are already live
- remove `Add token` from footer of empty tokens bubble
  - not needed + it was there for all coins which is not what we want
- sort nav according to figma design, have same order for all coins
  - `Details` is less important 
  - `Tokens` are "always" there in case of non-btc-like coins
  - `Staking will also be "always" there in case of non-btc-like coins

## Screenshots:
Screenshot 2024-06-28 at 23 15 21
Screenshot 2024-06-28 at 23 15 36
Screenshot 2024-06-28 at 23 23 58


AdamSchinzel (approved):
LGTM

PR #13093, opened 2024-06-28
Feat/cardano definitions

### LINKED ISSUE #12840: Resolve Cardano token definitions
- [x] Unify token definitions to include only fingerprint or policy id
- [x] Enable `token-definitions` for cardano
- [x] Fix name and symbol
- [x] Add policy id, finger print to dropdown
- [x] Do migration

tomasklim:
## Description

- cardano tokens now include name and symbol https://satoshilabs.slack.com/archives/C0292ULQHGD/p1718026910965839
- token definitions are enabled
- token unit is policy id + encoded name, this is required for signing tx
  - for everything else - token definitions, fiat rates, explorer, policy id is required and is now saved to contract address (for unit it is change on one place, for other it would be change in whole codebase)
- fixed explorer urls, some of them were not working as redirect was incorrect

## Related Issue

Resolve #12840

## Screenshots:
Screenshot 2024-06-29 at 1 40 37
Screenshot 2024-06-29 at 16 38 40


suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -490,6 +491,7 @@ export const networks = {
         features: ['tokens', 'staking'],
         explorer: {
             tx: 'https://testnet-explorer.blockfrost.dev/transaction/',
```
martykan: We should probably replace these as well with https://preview.cexplorer.io/...
Some of the URLs are not working for me

packages/suite/src/storage/index.ts
```diff
@@ -4,7 +4,7 @@ import { migrate } from './migrations';
 
 import type { SuiteDBSchema } from './definitions';
 
-const VERSION = 47; // don't forget to add migration and CHANGELOG when changing versions!
+const VERSION = 46; // don't forget to add migration and CHANGELOG when changing versions!
```
martykan, line 7: I assume 46 and 47 haven't been released yet?

tomasklim: Yes, it wasnt, i bumped it down again

suite-common/wallet-config/src/networksConfig.ts
tomasklim: 5af2707436895a0a351002a5512222529275825f

martykan (approved):

PR #13102, opened 2024-07-01
fix(suite): adding input clear to the transaction test

ondracja:
- fixing the broken `Overview and transactions check` test by clearing the input after the search

### DIFF
--- a/packages/suite-web/e2e/support/pageObjects/accountsObject.ts
+++ b/packages/suite-web/e2e/support/pageObjects/accountsObject.ts
@@ -74,6 +74,7 @@ class AccountsPage {
                     .children()
                     .should('have.length', 2);
             });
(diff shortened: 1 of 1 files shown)


packages/suite-web/e2e/support/pageObjects/accountsObject.ts
```diff
@@ -74,6 +74,7 @@ class AccountsPage {
                     .children()
                     .should('have.length', 2);
             });
+        cy.getTestElement('@wallet/accounts/search-icon').clear();
```
mroz22, line 77: shouldn't application logic take care of this?

tomasklim:
Fixed here https://github.com/trezor/trezor-suite/issues/13104

PR #13109, opened 2024-07-01
refactor(suite): DeviceStatus prop drilling

jvaclavik:
## Related Issue

Resolve

### DIFF
--- a/packages/suite/src/components/suite/labeling/WalletLabeling.tsx
+++ b/packages/suite/src/components/suite/labeling/WalletLabeling.tsx
@@ -18,10 +18,12 @@ export const useWalletLabeling = () => {
     const { translationString } = useTranslation();
 
     const defaultAccountLabelString = useCallback(
-        ({ device }: { device: TrezorDevice }) =>
-            device.useEmptyPassphrase
-                ? translationString('TR_NO_PASSPHRASE_WALLET')
-                : translationString('TR_PASSPHRASE_WALLET', { id: device.walletNumber }),
+        ({ device }: { device: TrezorDevice }) => {
+            if (device.useEmptyPassphrase) return translationString('TR_NO_PASSPHRASE_WALLET');
+            if (!device.walletNumber) return undefined;
+
+            return translationString('TR_PASSPHRASE_WALLET', { id: device.walletNumber });
+        },
         [translationString],
     );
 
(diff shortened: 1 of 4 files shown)


packages/suite/src/components/suite/labeling/WalletLabeling.tsx
```diff
@@ -18,10 +18,12 @@ export const useWalletLabeling = () => {
     const { translationString } = useTranslation();
 
     const defaultAccountLabelString = useCallback(
-        ({ device }: { device: TrezorDevice }) =>
-            device.useEmptyPassphrase
-                ? translationString('TR_NO_PASSPHRASE_WALLET')
-                : translationString('TR_PASSPHRASE_WALLET', { id: device.walletNumber }),
+        ({ device }: { device: TrezorDevice }) => {
+            if (device.useEmptyPassphrase) return translationString('TR_NO_PASSPHRASE_WALLET');
+            if (!device.walletNumber) return undefined;
```
jvaclavik, line 23: Missing this leads to names like "Hidden wallet #"

Should be related to this: https://satoshilabs.slack.com/archives/G019WLX2P7B/p1719566758372549

packages/suite/src/views/suite/SwitchDevice/DeviceItem/DeviceStatusText.tsx
```diff
@@ -1,43 +1,59 @@
...
-);
-
-export const DeviceStatusText = ({
-    device,
-    onRefreshClick,
-    walletLabel,
-}: DeviceStatusTextProps) => {
-    const { connected } = device;
-    const deviceStatus = deviceUtils.getStatus(device);
+const DeviceStatusVisible = ({ connected }: DeviceStatusVisible) => {
```
jvaclavik: We use walletLabel only here and it doesn't make sense to pass it through multiple components. At the same time, it's even harder to read when logic in multiple places.

packages/suite/src/components/suite/labeling/WalletLabeling.tsx
peter-sanderson: nit: I really prefer `{}` everywhere

peter-sanderson (approved):

Issue #12982, opened 2024-06-19
Grab authenticity public keys from trezor/data repository


tsusanka:
Take public keys from trezor/data repository where we moved it as part of https://github.com/trezor/data/pull/79. That is the source of truth from now on and we should make a copy in Suite during build process.

PR #13114, opened 2024-07-01
Chore: add data to suite

karliatto:
## Description
- 04f5f4a1ae25b605e440ebf6214d0c4a91e4db9c - adds https://github.com/trezor/data as a submodule of Trezor Suite. It is a small repository so it should not be big deal, but let me know if you think otherwise
- 7d20d301367daca254342f299224f2bcdadf719e - script to update the keys automatically 

If this approach is accepted I will create a GitHub workflow that checks when trezor/data submodule is updated and create a PR with changes.

## Related Issue

Related but does not complete until we have the GitHub workflow automation https://github.com/trezor/trezor-suite/issues/12982

### DIFF
--- a/scripts/ci/check-connect-data.ts
+++ b/scripts/ci/check-connect-data.ts
@@ -1,0 +1,116 @@
+import fetch from 'cross-fetch';
+import fs from 'fs-extra';
+import path from 'path';
+import { commit } from './helpers';
+
+const { exec } = require('./helpers');
+
+const AUTHENTICITY_BASE_URL = 'https://data.trezor.io';
+const authenticityPaths = {
+    T2B1: {
+        authenticity: 'firmware/t2b1/authenticity.json',
+        authenticityDev: 'firmware/t2b1/authenticity-dev.json',
+    },
+    T3T1: {
+        authenticity: 'firmware/t3t1/authenticity.json',
+        authenticityDev: 'firmware/t3t1/authenticity-dev.json',
+    },
+};
+
+const ROOT = path.join(__dirname, '..', '..');
+const CONFIG_FILE_PATH = path.join(ROOT, 'packages/connect/src/data/deviceAuthenticityConfig.ts');
+
+type AuthenticityPathsKeys = keyof typeof authenticityPaths;
+
+const fetchJSON = async (url: string) => {
+    const response = await fetch(url);
+    if (!response.ok) {
+        throw new Error(`Failed to fetch ${url}: ${response.statusText}`);
+    }
+    return response.json();
+};
+
+const updateConfigFromJSON = async () => {
+    try {
+        const devicesKeys = Object.keys(authenticityPaths) as AuthenticityPathsKeys[];
+
+        // Import the current configuration object
+        let { deviceAuthenticityConfig } = require(CONFIG_FILE_PATH);
+
+        for (const deviceKey of devicesKeys) {
+            const { authenticity, authenticityDev } = authenticityPaths[deviceKey];
+            const authenticityUrl = `${AUTHENTICITY_BASE_URL}/${authenticity}`;
+            const authenticityData = await fetchJSON(authenticityUrl);
+            const authenticityDevUrl = `${AUTHENTICITY_BASE_URL}/${authenticityDev}`;
+            const authenticityDevData = await fetchJSON(authenticityDevUrl);
+
+            deviceAuthenticityConfig[deviceKey] = {
+                rootPubKeys: authenticityData.root_pubkeys,
+                caPubKeys: authenticityData.ca_pubkeys,
+                debug: {
+                    rootPubKeys: authenticityDevData.root_pubkeys,
+                    caPubKeys: authenticityDevData.ca_pubkeys,
+                },
+            };
+        }
+
+        const updatedConfigString = `
+            /** THIS FILE IS AUTOMATICALLY UPDATED by script ci/scripts/check-connect-data.ts  */
+            import { DeviceAuthenticityConfig } from './deviceAuthenticityConfigTypes';
+
+            /**
+             * How to update this config or check Sentry "Device authenticity invalid!" error? Please read this internal description:
+             * https://www.notion.so/satoshilabs/Device-authenticity-check-b8656a0fe3ab4a0d84c61534a73de462?pvs=4
+             */
+            export const deviceAuthenticityConfig: DeviceAuthenticityConfig = ${JSON.stringify(deviceAuthenticityConfig, null, 4)};
+          `;
+
+        await fs.writeFile(CONFIG_FILE_PATH, updatedConfigString);
+
+        await exec('yarn', ['prettier', '--write', CONFIG_FILE_PATH]);
+
+        console.log('Configuration updated successfully.');
+
+        console.log('Checking if there were changes.');
+        const changes = await exec('git', ['diff', CONFIG_FILE_PATH]);
+        if (changes.stdout !== '') {
+            console.log('There were changes in keys.');
+            // There were changes in CONFIG_FILE_PATH
+            const hashChanges = await exec('git', [
+                'log',
+                '-1',
+                '--pretty=format:"%H"',
+                '--',
+                CONFIG_FILE_PATH,
+            ]);
+            // We use the hash of the changes to create branch name to avoid use a branch that already exists.
+            const branchName = `chore/update-device-authenticity-config-${hashChanges.stdout.replace(/"/g, '')}`;
+            const commitMessage = 'chore(connect): update device authenticity config';
+            await exec('git', ['checkout', '-b', branchName]);
+            commit({
+                path: ROOT,
+                message: commitMessage,
+            });
+            await exec('git', ['push', 'origin', branchName]);
+
+            await exec('gh', [
+                'pr',
+                'create',
+                '--repo',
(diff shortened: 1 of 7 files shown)


mroz22:
its not that small, if you search for  `submodules: true` you get about 17 results. I'd either make actions/checkout@v4 checkout only selected submodule (trezor-common which is needed for type-check), and if this is not possible I would rather checkout this separately from some script maybe. 

and we may also ask @vdovhanych what is TOO BIG? 



```
du -h submodules/trezor-data/
8.0K	submodules/trezor-data//udev
 22M	submodules/trezor-data//firmware/t1b1
 13M	submodules/trezor-data//firmware/t2t1
 16K	submodules/trezor-data//firmware/t2b1
 22M	submodules/trezor-data//firmware/1
  0B	submodules/trezor-data//firmware/translations/t2t1
  0B	submodules/trezor-data//firmware/translations/t2b1
  0B	submodules/trezor-data//firmware/translations/t3t1
  0B	submodules/trezor-data//firmware/translations
1.6M	submodules/trezor-data//firmware/t3t1
 13M	submodules/trezor-data//firmware/2
 70M	submodules/trezor-data//firmware
4.0K	submodules/trezor-data//transparency
8.0K	submodules/trezor-data//security
4.0K	submodules/trezor-data//ci
 53M	submodules/trezor-data//bridge/2.0.30
 28M	submodules/trezor-data//bridge/2.0.24
 37M	submodules/trezor-data//bridge/2.0.25
 16K	submodules/trezor-data//bridge/cert
 37M	submodules/trezor-data//bridge/2.0.27
 42M	submodules/trezor-data//bridge/2.0.29
 42M	submodules/trezor-data//bridge/2.0.28
 37M	submodules/trezor-data//bridge/2.0.26
276M	submodules/trezor-data//bridge
128K	submodules/trezor-data//bootloader/t1b1
504K	submodules/trezor-data//bootloader/t2t1
104K	submodules/trezor-data//bootloader/t2b1
128K	submodules/trezor-data//bootloader/1
504K	submodules/trezor-data//bootloader/2
1.3M	submodules/trezor-data//bootloader
8.0K	submodules/trezor-data//.github/workflows
8.0K	submodules/trezor-data//.github
4.0K	submodules/trezor-data//registry
8.0K	submodules/trezor-data//dev
4.0K	submodules/trezor-data//suite/releases
  0B	submodules/trezor-data//suite/log/web
  0B	submodules/trezor-data//suite/log/desktop
  0B	submodules/trezor-data//suite/log/mobile
  0B	submodules/trezor-data//suite/log
4.0K	submodules/trezor-data//suite/definitions
8.0K	submodules/trezor-data//suite
376K	submodules/trezor-data//legal
  0B	submodules/trezor-data//connect/log
  0B	submodules/trezor-data//connect
348M	submodules/trezor-data/
```

tsusanka:
Two questions, not exactly related to the code.

1. Will / can you use also the firmware binaries? Because you build them inside Suite right? Or not really? (just asking if it would simplify some stuff on your side)
2. If not, does it make sense for you to be dependent on trezor/data? @matejcik is there any reasons we didn't put this to trezor-common? Would that be more appropriate place?

karliatto:
> Two questions, not exactly related to the code.
> 
>     1. Will / can you use also the firmware binaries? Because you build them inside Suite right? Or not really? (just asking if it would simplify some stuff on your side)

Yes, we include firmware binaries in suite but only latest version, it looks like trezor/data includes all the versions so I think current approach is better.

>     2. If not, does it make sense for you to be dependent on trezor/data? @matejcik is there any reasons we didn't put this to trezor-common? Would that be more appropriate place?

If we could have the files with those keys in trezor-common that would simplify things.

matejcik:
> 2\. If not, does it make sense for you to be dependent on trezor/data? @matejcik is there any reasons we didn't put this to trezor-common? Would that be more appropriate place?

The reason is that this data is (a) not a dependency of firmware and (b) in fact independent of firmware development.
We're trying to _reduce_ unrelated churn from the firmware monorepo, not add more of it :) If we added the keys to the monorepo, we'd get a completely random commit every few weeks that adds some keys to some datafile.

From my POV `trezor-data` is exactly the right place for data of this kind.

karliatto:
....
> 348M	submodules/trezor-data/
> 

I think this is too big to be added as submodule to trezor-suite for the purpose of the keys.

If we do not want to add those keys to trezor-common because it creates too many "random commits" every week as @matejcik stated in previous comment, then I can come up with different solution for this. 

I would try to retrieve the files with keys from GitHub directly instead of using trezor/data as submodule which is very inefficient.

vdovhanych:
Having `trezor/data` as a submodule is a terrible solution, and I'm glad you've come to the same conclusion. 

What is the issue with fetching the keys from data.trezor.io in the same way we fetch other stuff? We could also fetch it during the build to have it ready offline in the suite.

karliatto:
> 
> What is the issue with fetching the keys from data.trezor.io in the same way we fetch other stuff? We could also fetch it during the build to have it ready offline in the suite.

No issue with that, I just did not know that was the intended way to use trezor/data. That simplify things. Thanks!

matejcik:
I'll just raise the possibility of data.trezor.io getting attacked and this data replaced. Admittedly the risk is low so perhaps not a blocker for this solution -- plus, we should add signature over this data soon-ish anyway. Just so y'all know to keep it in mind.

matejcik:
that is:

> What is the issue with fetching the keys from data.trezor.io in the same way we fetch other stuff?

the data is unsigned and we don't want to rely on _customers_ to stay secure, so this is a no-go until signatures are added

> We could also fetch it during the build to have it ready offline in the suite.

this is very likely fine

karliatto:
> so this is a no-go until signatures are added

For now I propose to use data.trezor.io with an automation that will create a PR towards the suite repository with new keys to update the file `packages/connect/src/data/deviceAuthenticityConfig.ts`.

This way there will be human checking the update. Once we have the files signed we can proceed with a way to load those keys when building or similar with less human interaction.

vdovhanych:
How often are the keys updated? Is it with every FW update? If so, I see no reason to automate it; just add it with FW binaries, and that's it.

karliatto:
> How often are the keys updated? Is it with every FW update? If so, I see no reason to automate it; just add it with FW binaries, and that's it.

You can find more details about when new keys are created here https://www.notion.so/satoshilabs/Device-authenticity-check-b8656a0fe3ab4a0d84c61534a73de462

tsusanka:
> The reason is that this data is (a) not a dependency of firmware and (b) in fact independent of firmware development.
We're trying to reduce unrelated churn from the firmware monorepo, not add more of it :) If we added the keys to the monorepo, we'd get a completely random commit every few weeks that adds some keys to some datafile.

I mean on one side you are right, on the other hand, we are creating more work for others for a bit dubious reasons. What is wrong about one commit per month in FW monorepo? Will you run out of sha256 hashes? :) I'd argue that simplifying this for other teams has higher priority. Btw we already do have [udev](https://github.com/trezor/trezor-common/tree/master/udev) rules and some [keys](https://github.com/trezor/trezor-common/tree/master/keys) in `trezor-common`, where is the line between `common` and `data`?

We can solve this by fetching it from data.trezor.io, but as @matejcik says it creates the unnecessary dependency on the security of the website. I mean yes it is not a biggie as having the website compromised exactly in the moment of building production Suite is very low + we plan to introduce the signing, but I just do not see how it is better than `trezor-common`.

@matejcik are you "vetoing" having this in `trezor-common`? If yes, let's fetch it from data.trezor.io. If not, let's move it to `trezor-common`. And I think you should have the final say as the owner of the firmware monorepo.

matejcik:
i've spent at least 15 minutes on this reply already so instead i'll keep it short and to the point:

* fw team is _not the owner of this data_
* trezor-common, being _part of the monorepo_, is not a good warehouse for data produced in provisioning and consumed by Suite. i'd have less of an issue if trezor-common was a shared ownership stand-alone repository
* the main reason i put this data into trezor/data originally is **to publish it on data\.trezor\.io**, which trezor-common is obviously not doing. `trezorctl` is already looking for this data on a data\.trezor\.io url, the endgame here is that Suite will do the same (when online).

in general that's a **no** for trezor-common.

tsusanka:
Ok settled. @karliatto fetching from data.trezor.io during build process is the way to go, no submodules. And we will introduce some signing later on.

scripts/ci/check-connect-data.ts
```diff
@@ -0,0 +1,116 @@
...
+                '--body-file',
+                'docs/packages/connect/check-connect-data.md',
+                '--base',
+                'develop',
+                '--head',
+                branchName,
+            ]);
+        }
+    } catch (error) {
+        console.error(`Error updating configuration: ${error.message}`);
```
mroz22, line 112: just thinking how useful is to try catch this? also not sure what happens in the CI if we don't do it. This appears to me that if the script fails for any reason, we might not be seeing stac traces in CI log and without catching the error we might be able to see them? Just a nitpick and I am really not sure what is the advisable solution here, would need to run some experiments which might not be worth of doing :D

mroz22 (approved):
looks nice

scripts/ci/check-connect-data.ts
karliatto: If that fails it will log in the CI the error message and it will give us a clue about what happened.

PR #13116, opened 2024-07-01
feat(coinmarket): default fiat currencies

adderpositive:
## Info
After selecting a fiat currency, set the amount of fiat from a static list gets from Invity API.


packages/suite/src/hooks/wallet/coinmarket/form/useCoinmarketBuyForm.tsx
```diff
@@ -115,7 +114,7 @@ const useCoinmarketBuyForm = ({
               fiatInput:
                   draft.fiatInput && draft.fiatInput !== ''
                       ? draft.fiatInput
-                      : defaultFiatCurrencies.get('czk'),
+                      : buyInfo?.buyInfo?.defaultFiatCurrencies.get('czk'),
```
LukasRada: Please use suggestedFiatCurrency instead of czk.

LukasRada (changes requested):

PR #13138, opened 2024-07-02
fix(suite): Device status text

jvaclavik:
## Related Issue

Resolve 

## Screenshots:

1) only one standard wallet active and there are no other wallets:

image

2) passphrase wallet active (doesn't matter how many wallets are created):

image

3) standard wallet when there are other wallets:

image

in modal there is always a connection info:

image

image

### DIFF
--- a/packages/suite/src/components/suite/layouts/SuiteLayout/DeviceSelector/SidebarDeviceStatus.tsx
+++ b/packages/suite/src/components/suite/layouts/SuiteLayout/DeviceSelector/SidebarDeviceStatus.tsx
@@ -1,5 +1,5 @@
 import { MouseEventHandler } from 'react';
-import { acquireDevice, selectDevice } from '@suite-common/wallet-core';
+import { acquireDevice, selectDevice, selectDevices } from '@suite-common/wallet-core';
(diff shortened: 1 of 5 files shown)


marekrjpolak (approved):
LGTM

MiroslavProchazka (approved):

packages/suite/src/components/suite/layouts/SuiteLayout/DeviceSelector/SidebarDeviceStatus.tsx
```diff
@@ -36,13 +36,19 @@ export const SidebarDeviceStatus = () => {
     if (!selectedDevice || !selectedDeviceModelInternal) {
         return null;
     }
+    const instances = deviceUtils.getDeviceInstances(selectedDevice, devices);
+    const instancesWithState = instances.filter(i => i.state);
```
peter-sanderson, line 40: I am not really fan of one-letter variables, shall be `instance`

jvaclavik: Good point, I agree

PR #13147, opened 2024-07-02
fix(connect): proceed with fw install even when language data fails t…

mroz22:
…o download

tested it by setting connect to download language data from some bogus url. it went through. 

image

is there an issue?

### DIFF
--- a/packages/connect/src/core/onCallFirmwareUpdate.ts
+++ b/packages/connect/src/core/onCallFirmwareUpdate.ts
@@ -358,6 +358,8 @@ export const onCallFirmwareUpdate = async ({
                       language: targetLanguage,
                       version: device.firmwareRelease.release.version,
                       internal_model: device.features.internal_model,
+                  }).catch(() => {
+                      // silent, language data is not critical, it can be updated any time later and it indeed happens inside device.updateFeatures
                   })
                 : null;
 


komret (approved):

Hannsek:
Does user have to update the language manually or will it be automatic the next fw update?

mroz22:
it will update automatically under the hood when he connects device next time, or maybe even triggers next connect call

Hannsek:
Will the user have to confirm it? Confirm the installation of the language blob?

bosomt:
Tested on 24.6.3 latest release candidate

komret:
> Will the user have to confirm it? Confirm the installation of the language blob?

No, the update is silent.

Hannsek:
So the language of the firmware will be changed automatically during the next fw update? This is not nice. :/

komret:
No, it should never change. If an outdated blob is detected, it attempts to silently update to the same language. If there is no translation blob, nothing happens.

Not sure how much this was tested though.

Hannsek:
Not sure if the firmware accepts the blob silently. @matejcik?

matejcik:
what @komret says: Suite is always allowed to _silently_ update an outdated language blob to one matching the current language + version

PR #13157, opened 2024-07-03
fix(suite): Fix Divider in WalletInstance

jvaclavik:
## Related Issue

Resolve 

## Screenshots:

before
Screenshot 2024-07-03 at 09 46 14

after
Screenshot 2024-07-03 at 09 46 21


packages/suite/src/views/suite/SwitchDevice/types.ts
```diff
@@ -1 +1 @@
-export type ContentType = 'default' | 'ejectConfirmation' | 'disabling-view-only-ejects-wallet';
+export type ContentType = 'default' | 'eject-confirmation' | 'disabling-view-only-ejects-wallet';
```
jvaclavik, line 1: It's just about the consistency of naming

peter-sanderson (approved):

PR #13158, opened 2024-07-03
fix(suite-native): onboard and connect test

vytick:
Check for test id instead of a string

### DIFF
--- a/suite-native/app/e2e/tests/onboardAndConnect.test.ts
+++ b/suite-native/app/e2e/tests/onboardAndConnect.test.ts
@@ -46,7 +46,7 @@ describe('Go through onboarding and connect Trezor.', () => {
                 .toBeVisible()
                 .withTimeout(10000); // communication between connected Trezor and app takes some time.
 
-            await detoxExpect(element(by.text('Bitcoin'))).toBeVisible();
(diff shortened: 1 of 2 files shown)


suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraphHeader.tsx
```diff
@@ -60,31 +60,33 @@ export const PortfolioGraphHeader = () => {
...
-                                    />
-                                </>
-                            ) : (
-                                <Text>{' ' /* just placeholder to avoid layout shift */}</Text>
-                            )}
-                        </HStack>
-                    </>
-                )}
-            </VStack>
+            {!isLoading && (
```
PeKne: Won't this be as flaky as the previous version? The view with the specific `testID` will appear only after the discovery finishes, which is even later than the original check of discovered BTC account. 

Did you consider this? IMO it would be enough to check that discovery at least started (UI changed according to it).

vytick: [1d22544](https://github.com/trezor/trezor-suite/pull/13158/commits/1d22544511330f1f7da87d080ea90aa9629f9ffe)

Nodonisko:
FYI Every screen has testID so it might be useful here too

PeKne (approved):

PR #13175, opened 2024-07-03
chore(transport): remove unnecessary return type from Usb and Udp api

szymonlesisz:
## Description
Remove unnecessary type overload from transport api `read` methods.
Type correctness is already watched by AbstractApi['read']

### DIFF
--- a/packages/transport/src/api/udp.ts
+++ b/packages/transport/src/api/udp.ts
@@ -1,8 +1,12 @@
 import UDP from 'dgram';
 import { isNotUndefined } from '@trezor/utils';
 
-import { AbstractApi, AbstractApiConstructorParams, DEVICE_TYPE } from './abstract';
-import { ResultWithTypedError } from '../types';
+import {
+    AbstractApi,
+    AbstractApiAwaitedResult,
+    AbstractApiConstructorParams,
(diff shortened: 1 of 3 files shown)


packages/transport/src/api/udp.ts
```diff
@@ -37,14 +41,7 @@ export class UdpApi extends AbstractApi {
...
         const [hostname, port] = path.split(':');
 
-        return new Promise<
-            ResultWithTypedError<
-                undefined,
-                | typeof ERRORS.INTERFACE_DATA_TRANSFER
-                | typeof ERRORS.UNEXPECTED_ERROR
-                | typeof ERRORS.ABORTED_BY_SIGNAL
-            >
-        >(resolve => {
```
mroz22, line 47: not 100% sure by this change. in fact I am lying a bit here, this very promise can't return all of the errors specified by the type. but it is probably ok, this type is intended to be consumed by a higher layer so we probably don't need to be extremely specific here.

mroz22 (approved):

packages/transport/src/api/udp.ts
szymonlesisz: lgtm

PR #13177, opened 2024-07-03
fix(suite): add validation for IPC messages for Electron

peter-sanderson:
Addresses the https://www.electronjs.org/docs/latest/tutorial/security#17-validate-the-sender-of-all-ipc-messages

### DIFF
--- a/packages/ipc-proxy/package.json
+++ b/packages/ipc-proxy/package.json
@@ -17,7 +17,7 @@
     "browser": "src/proxy",
     "scripts": {
         "lint:js": "eslint '**/*.{ts,tsx,js}'",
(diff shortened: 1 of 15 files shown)


peter-sanderson:
Todo: figure out the correct URL origins to be allowed

Production suite `window.location`
- image

Dev suite:
- `http://localhost:8000/`

komret:
> Todo: figure out the correct URL origins to be allowed

It should be `http://localhost:8000/` for dev, otherwise `file:///index.html` - just checked that using logger on a built version. Localhost should not be allowed in the built/prod version if possible.

komret (approved):
👏

PR #13184, opened 2024-07-04
feat(mobile): use local transactions for graph

Nodonisko:
## Description

This PR moves transaction-cache functionallity to redux, because in the end it's simplest solution. There two next steps that should be done, we should add function for calculating balance changes of ETH + tokens and optionally we may look into optimilizing performance of this solution (it works nice for normal accounts for for accounts with lot of transactions it could be slow).

## Related Issue

Resolve


suite-common/graph/src/graphBalanceEvents.ts
```diff
@@ -107,18 +110,26 @@ export const getAccountMovementEvents = async ({
...
     startOfTimeFrameDate: Date | null;
     endOfTimeFrameDate: Date;
+    dispatch: ReturnType<typeof useDispatch>;
 }) => {
     const { coin, identity, descriptor } = account;
 
     const getBalanceHistory = async () => {
         if (getNetworkType(coin) === 'ripple') {
-            return TransactionCacheEngine.getAccountBalanceHistory({
+            const allTransactions = await dispatch(
```
matejkriz, line 124: Why is this applied for ripple only?

suite-common/graph/src/balanceHistoryUtils.ts
```diff
@@ -0,0 +1,138 @@
...
+            summary.txs += 1;
+            summary.received = amount.plus(receivedDrops);
+            summary.sent = amount.plus(sentDrops);
+        } else {
+            summaryMap.set(blockTime, {
+                time: blockTime,
+                txs: 1,
+                received: receivedDrops,
+                sent: sentDrops,
+                sentToSelf: new BigNumber(0),
```
matejkriz: With ripple is it not possible to send to self?

suite-common/graph/src/balanceHistoryUtils.ts
```diff
@@ -0,0 +1,138 @@
...
+            sentDrops = amount;
+        } else if (tx.type === 'recv') {
+            receivedDrops = amount;
+        }
+
+        if (summaryMap.has(blockTime)) {
+            const summary = summaryMap.get(blockTime)!;
+            summary.txs += 1;
+            summary.received = amount.plus(receivedDrops);
+            summary.sent = amount.plus(sentDrops);
```
matejkriz: I don't get this 🤔
 if for example `tx.type === 'sent'`:
 ```
sentDrops = amount
receivedDrops = 0
// =>
summary.sent = amount + amount // ? 
```

suite-common/graph/src/balanceHistoryUtils.ts
Nodonisko: Yes, from what I know ripple could have only one output.

suite-common/graph/src/balanceHistoryUtils.ts
Nodonisko: Yea, this is definitely mistake.

matejkriz:
already merged in another PR

PR #13167, opened 2024-07-03
fix(suite): Electron Security, Disable/Limit navigation

peter-sanderson:
To adhere to this recommendation https://www.electronjs.org/docs/latest/tutorial/security#13-disable-or-limit-navigation

To test this just add:

```ts
<button onClick={() => {
    location.assign('https://google.com');
}}>Click me!</button>
```

somewhere

PR #13185, opened 2024-07-04
Chore/disable navigation

komret:
## Description

Disable navigation in Electron as recommended in https://www.electronjs.org/docs/latest/tutorial/security#13-disable-or-limit-navigation


peter-sanderson:
Already merged in https://github.com/trezor/trezor-suite/pull/13167

PR #13186, opened 2024-07-04
fix(suite): Delete unused condition for enable passphrase

jvaclavik:
## Related Issue

Resolve

### DIFF
--- a/suite-common/wallet-core/src/device/deviceThunks.ts
+++ b/suite-common/wallet-core/src/device/deviceThunks.ts
@@ -114,27 +114,8 @@ export const createDeviceInstanceThunk = createThunk<
     { rejectValue: CreateDeviceInstanceError }
 >(
     `${DEVICE_MODULE_PREFIX}/createDeviceInstance`,
-    async (
-        { device, useEmptyPassphrase = false },
-        { dispatch, getState, rejectWithValue, fulfillWithValue },
-    ) => {
+    ({ device, useEmptyPassphrase = false }, { getState, rejectWithValue, fulfillWithValue }) => {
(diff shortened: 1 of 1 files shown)


packages/suite/src/views/suite/SwitchDevice/DeviceItem/AddWalletButton.tsx
```diff
@@ -79,7 +79,7 @@ export const AddWalletButton = ({ device, instances, onCancel }: AddWalletButton
                         >
                             <Row gap={spacings.xs}>
                                 <Translation id="TR_ADD_HIDDEN_WALLET" />{' '}
-                                {!isLocked && <HotkeyBadge hotkey={['ALT', 'KEY_P']} />}
```
peter-sanderson: :question:

suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -114,27 +114,11 @@ export const createDeviceInstanceThunk = createThunk<
     { rejectValue: CreateDeviceInstanceError }
 >(
     `${DEVICE_MODULE_PREFIX}/createDeviceInstance`,
-    async (
+    (
         { device, useEmptyPassphrase = false },
         { dispatch, getState, rejectWithValue, fulfillWithValue },
     ) => {
         if (!device.features) return rejectWithValue({ error: 'features-unavailable' });
-        if (!device.features.passphrase_protection) {
```
peter-sanderson, line 122: tbh I think this whole thunk is useless afrer we remove this

jvaclavik:
It's probably safer not to delete it right now

PR #13193, opened 2024-07-04
chore(suite): add empty lines around "class", "if", "while", "switch", "try"

peter-sanderson:
Example implementation for this Code Improvement proposal: https://github.com/trezor/trezor-suite/issues/13194


mroz22 (changes requested):
please use a better commit message. it is affecting also connect and libs where we use git log mesages to generate changelogs.

marekrjpolak (changes requested):
Rejected for excessive fascism.

Really, who needs this?

tomasklim (changes requested):
I hate even the one rule for space between method code and return

PR #13200, opened 2024-07-04
fix(suite): Let Sidebar update its height with window change or zoom

hstastna:
## Description

Sidebar's height was not updating correctly to its maximum (window size) even when there wasn't any maximum height explicitly set for it. Let the sidebar's height update accordingly when changing window size or zoom level. Make this behavior optional by adding new `updateWidthOnWindowResize` and `updateHeightOnWindowResize` props to `ResizableBox` component.

The issue was visible mainly in Chrome browser and in desktop Suite (probably also on Mac).

## Screenshots:
**Before:**

https://github.com/trezor/trezor-suite/assets/13417815/a3812367-0a29-4cb2-915f-13a2fc578abc

**After:**

https://github.com/trezor/trezor-suite/assets/13417815/d65578fc-de79-48c7-b074-a06a09025ba5

### DIFF
--- a/packages/components/src/components/ResizableBox/ResizableBox.stories.tsx
+++ b/packages/components/src/components/ResizableBox/ResizableBox.stories.tsx
@@ -38,13 +38,41 @@ export const ResizableBox: StoryObj<ResizableBoxProps> = {
         height: 100,
         minHeight: 50,
         maxHeight: 300,
+        updateWidthOnWindowResize: false,
+        updateHeightOnWindowResize: false,
     },
     argTypes: {
+        children: {
+            table: {
+                type: {
+                    summary: 'ReactNode',
+                },
+            },
+        },
         directions: {
             control: {
                 type: 'check',
             },
             options: ['top', 'left', 'right', 'bottom'],
+        },
+        isLocked: { control: 'boolean' },
+        width: { control: 'number' },
+        minWidth: { control: 'number' },
+        maxWidth: { control: 'number' },
+        height: { control: 'number' },
+        minHeight: { control: 'number' },
+        maxHeight: { control: 'number' },
+        updateWidthOnWindowResize: {
+            control: 'boolean',
+            description:
+                'Maximize the box width to the window width when resizing window or zooming',
+            defaultValue: false,
+        },
+        updateHeightOnWindowResize: {
+            control: 'boolean',
+            description:
+                'Maximize the box height to the window height when resizing window or zooming',
+            defaultValue: false,
         },
     },
 };
--- a/packages/components/src/components/ResizableBox/ResizableBox.tsx
+++ b/packages/components/src/components/ResizableBox/ResizableBox.tsx
@@ -16,6 +16,8 @@ export type ResizableBoxProps = {
     height?: number;
     minHeight?: number;
     maxHeight?: number;
+    updateWidthOnWindowResize?: boolean;
+    updateHeightOnWindowResize?: boolean;
     zIndex?: ZIndexValues;
 };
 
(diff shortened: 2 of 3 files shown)


packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -177,6 +180,7 @@ export const ResizableBox = ({
     height,
     minHeight = 0,
     maxHeight,
+    maxSizeDirections,
```
jvaclavik: I suggest to rename it to something more clear:

e.g.:
```
updateWidthOnWindowResize?: boolean;
updateHeightOnWindowResize?: boolean;
```

hstastna: I am not sure.. Do we really need 2 props if this behavior could sufficiently be handled only by one? It is about the same thing, just the direction is different. And the ResizableBox already has a lot of props..

jvaclavik: The thing is that when I read `maxSizeDirections ` I don't understand what it does. 

I couldn't figure out one prop for this two use cases without loosing context. And I think having many props is not an issue if you know what they do.

hstastna: Right, makes sense. Haven't figured out better namings so I used your suggestions. Thanks a lot, now the code looks better, more understandable.

packages/components/src/components/ResizableBox/ResizableBox.stories.tsx
```diff
@@ -38,6 +38,7 @@ export const ResizableBox: StoryObj<ResizableBoxProps> = {
         height: 100,
         minHeight: 50,
         maxHeight: 300,
+        maxSizeDirections: undefined,
```
jvaclavik: `updateWidthOnWindowResize`
`updateHeightOnWindowResize`

packages/components/src/components/ResizableBox/ResizableBox.stories.tsx
```diff
@@ -46,5 +47,11 @@ export const ResizableBox: StoryObj<ResizableBoxProps> = {
             },
             options: ['top', 'left', 'right', 'bottom'],
         },
+        maxSizeDirections: {
```
jvaclavik: same here

hstastna: Oops, right!

packages/components/src/components/ResizableBox/ResizableBox.stories.tsx
```diff
@@ -38,6 +38,8 @@ export const ResizableBox: StoryObj<ResizableBoxProps> = {
         height: 100,
         minHeight: 50,
         maxHeight: 300,
+        updateHeightOnWindowResize: false,
+        updateWidthOnWindowResize: false,
```
jvaclavik: Btw you could add these props also to `argTypes`

hstastna: Yes I could but do we need it? Storybook automatically infers the controls for props based on the component's propTypes or TypeScript definitions. I can see many components don't have all their props added both in `args` and `argTypes` objects..

hstastna: I've added anything missing to argTypes for ResizableBox in this PR, even unnecessary. But I think we should make clear if we want to do this for all the components or not, or when (and then to update them accordingly, of course).

jvaclavik (approved):

PR #13202, opened 2024-07-04
fix(suite-native): set added account visible before go to detail

### LINKED ISSUE #13196: Added account is not visible in My Assets.
**Describe the bug**
Added account is not visible in My Assets.

**Info:**

-   Mobile Suite Version: develop

**How to reproduce**

1. Go to My assets
2. Click on + (top right corner) with connected device after finished discovery
3. Chose bitcoin
4. Tap Continue with SegWit
5. New Bitcoin account detail appears (remember the #number)
6. tap X (top left) and app goes to My assets
7.  BTC account it #number is not present in the list

**Expected behavior**
in step 7. BTC account should be present

**Screenshots**

https://github.com/trezor/trezor-suite/assets/2011829/e8731c52-bf6e-4528-a53b-99f6640b058c

**Additional context**
- If user taps Change account type in step 4. and in next screen selects the same SegWit, it works correctly.
- For Coins that do not go through any kind of account type selection (ETH, ETC, DOGE) it works correctly

vytick:
`fetchTransactionsPageThunk` did use old data of account to fetch transaction. The account data changed during async `getAccountInfo` call, but this change was overwritten by using the old data in `accountsActions.updateAccount`.

This PR uses fresh account data even after getAccountInfo fetch

## Related Issue

Resolve #13196

## Screenshots:

https://github.com/trezor/trezor-suite/assets/2011829/76ee012e-eaf3-4cc0-a0c4-5b91375ad7c2

### DIFF
--- a/suite-common/wallet-core/src/transactions/transactionsThunks.ts
+++ b/suite-common/wallet-core/src/transactions/transactionsThunks.ts
@@ -281,8 +281,15 @@ export const fetchTransactionsPageThunk = createThunk(
             suppressBackupWarning: true,
         });
 
+        // Account might have changed during async getAccountInfo call, so we fetch current state
+        const currentAccount = selectAccountByKey(getState(), accountKey);
+
+        if (!currentAccount) {
+            throw new Error(`Account not found: ${accountKey}`);
+        }
+
         if (result && result.success) {
-            const updateAction = accountsActions.updateAccount(account, result.payload);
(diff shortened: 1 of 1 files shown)


matejkriz (changes requested):
Let's try to fix `fetchTransactionsPageThunk` instead 🙏

Nodonisko (commented):
Good catch

matejkriz (approved):

matejkriz (approved):
Works as expected 👌

PR #13206, opened 2024-07-04
Feat/autheticate multiple devices

komret:
## Description

Security check now supports multiple devices at the same time. Previously, only the first selected device was checked, which could have led to skipped device authenticity checks.

Scenarios:

1) Two devices with SE connected in fresh Suite:

https://github.com/trezor/trezor-suite/assets/42465546/3e4aa1e2-cc6f-4414-bd3d-e2934a8dda2a

2) Two devices connected, the first one selected doesn't have SE:

https://github.com/trezor/trezor-suite/assets/42465546/547b1200-0c8b-41ca-81c3-54c34351d590

3) Multiple devices selected, one other than the first one selected doesn't have SE - security check for that one is skipped. I don't think it is a big deal because it is super edge case and there would be no authenticity check anyway.

### DIFF
--- a/packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
+++ b/packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
@@ -2,7 +2,12 @@ import { useEffect, useState } from 'react';
 import styled, { useTheme } from 'styled-components';
 
 import { getConnectedDeviceStatus } from '@suite-common/suite-utils';
-import { deviceActions, selectDevice, selectDevices } from '@suite-common/wallet-core';
+import {
+    deviceActions,
+    selectDevice,
+    selectDeviceAuthenticity,
+    selectDevices,
+} from '@suite-common/wallet-core';
 import { Icon, Tooltip, variables, H2, useElevation } from '@trezor/components';
 import { DeviceModelInternal } from '@trezor/connect';
 import { TREZOR_RESELLERS_URL, TREZOR_URL } from '@trezor/urls';
@@ -300,7 +305,7 @@ const SecurityCheckContent = ({
 export const SecurityCheck = () => {
     const device = useSelector(selectDevice);
     const devices = useSelector(selectDevices);
-    const allDevicesAutheticity = useSelector(state => state.device.deviceAuthenticity);
+    const deviceAuthenticity = useSelector(selectDeviceAuthenticity);
     const dispatch = useDispatch();
     const [isAuthenticityCheckStep, setIsAuthenticityCheckStep] = useState(false);
     const { goToSuite } = useOnboarding();
@@ -316,7 +321,7 @@ export const SecurityCheck = () => {
                 device.features?.internal_model &&
                 isAuthenticationSupportedMap[device.features.internal_model] &&
                 device.id &&
-                !allDevicesAutheticity?.[device.id],
+                !deviceAuthenticity?.[device.id],
         );
 
         if (nextDeviceToAuthenticate !== undefined) {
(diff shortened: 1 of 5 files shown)


packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
```diff
@@ -119,6 +124,9 @@ const firmwareInstalledChecklist = [
     },
 ] as const;
 
+const isAuthenticityCheckSupported = (deviceModel: DeviceModelInternal) =>
+    [DeviceModelInternal.T2B1, DeviceModelInternal.T3T1].includes(deviceModel);
```
peter-sanderson: Here I would add a map `Record<DeviceModelInternal, boolean>` so when adding a new device we have to explicitly define if it has AUthenticity check.

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
```diff
@@ -289,25 +300,47 @@ export const SecurityCheckContent = ({
...
+    const goToSuiteOrNextDevice = () => {
+        const nextDeviceToAuthenticate = devices.find(
+            device =>
+                device.features?.internal_model &&
+                isAuthenticityCheckSupported(device?.features?.internal_model) &&
+                device.id &&
+                !deviceAuthenticity?.[device.id],
+        );
+
+        if (nextDeviceToAuthenticate) {
```
peter-sanderson: nit: `nextDeviceToAuthenticate !== undefined`

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
```diff
@@ -289,25 +300,47 @@ export const SecurityCheckContent = ({
...
     // Device A disconnects while on the first screen of the check.
     useEffect(() => {
-        if (!isAuthenticityCheckSupported) {
+        if (isDeviceAuthenticityCheck && !shouldAuthenticateSelectedDevice) {
             setIsDeviceAuthenticityCheck(false);
         }
-    }, [isAuthenticityCheckSupported]);
+    }, [isDeviceAuthenticityCheck, shouldAuthenticateSelectedDevice]);
 
     if (isDeviceAuthenticityCheck) {
```
peter-sanderson: I know the code was here before, but I was struggling to understand what does this mean. 

I suggest something like `should(Show|Display|Render)DeviceAuthenticityCheck`

peter-sanderson (approved):
Just nits.

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
komret: 9822eb26d306fd5f3bc4b83bf3b280221025dd70

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
komret: 9822eb26d306fd5f3bc4b83bf3b280221025dd70

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
komret: 9822eb26d306fd5f3bc4b83bf3b280221025dd70

bosomt:
**QA OK**

Connected TS5 and TS3 at once with emtpy Suite storage

**Info:**
 - Suite version: web 24.8.0 (89cf20ef5aa790a71e9a41310f865014bc7bbed4)
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.0 regular (revision dd4671a5104952ef505d28d1f9e94d1484b4607a)

PR #13214, opened 2024-07-04
Chore/native/use add coin returns error

### LINKED ISSUE #13163: Use add account receives an actual error
In a hook `useAddCoinAccount`, there is a function `addCoinAccount` which is calling `addAndDiscoverNetworkAccountThunk`, which then calls `fetchBundleDescriptorsThunk`. Instead of getting the actual error and passing it to the caller, it returns undefined so we cannot display why there was an error (e. g. passphrase incorrect, etc). 

This needs to be refactored so instead of plain error, we can display an actual message as to why this call failed.

vytick:
## Description
- addAndDiscoverNetworkAccountThunk now rejects or fulfils with value
- fixed typo in component name

## Related Issue

Resolve #13163

### DIFF
--- a/suite-native/module-add-accounts/src/components/AccountTypeDecisionBottomSheet.tsx
+++ b/suite-native/module-add-accounts/src/components/AccountTypeDecisionBottomSheet.tsx
@@ -10,7 +10,7 @@ const descStyle = prepareNativeStyle(utils => ({
     paddingBottom: utils.spacings.small,
 }));
 
-type AccountTypeDecisionBootomSheetProps = {
+type AccountTypeDecisionBottomSheetProps = {
     coinName: string;
     typeName: string;
     isVisible: boolean;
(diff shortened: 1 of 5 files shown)


suite-native/module-add-accounts/src/components/AccountTypeDecisionBottomSheet.tsx
```diff
@@ -10,7 +10,7 @@ const descStyle = prepareNativeStyle(utils => ({
     paddingBottom: utils.spacings.small,
 }));
 
-type AccountTypeDecisionBootomSheetProps = {
+type AccountTypeDecisionBottomSheetProps = {
```
juriczech, line 13: Good catch 😄 👏

juriczech (changes requested):
This is not working as expected. First we need to refactor `fetchBundleDescriptorsThunk` and it's return value from connect. The thunk returns empty array if no bundle was found (if there was error). We need to change that so that it doesn't return empty array but an error. Than we propagate the error to the callers. 

You've actually made matters worse a little bit since add coin on develop returns at least basic error, now it actually doesn't do anything and just hangs on a loading screen. 

If you'd like, maybe we could huddle and I might explain better looking at the code.

juriczech:
Could you also retest discovery so we know it hasn't been broken by this change? I believe, you're not handling rejected `fetchBundleDescriptorsThunk` here and now `deviceAccessResponse.payload` doesn't have to be an array but could be some error. 
https://github.com/trezor/trezor-suite/blob/fe647245bbaba774423476bb862b4b9a65e69823/suite-native/discovery/src/discoveryThunks.ts#L461

juriczech (approved):
LGTM. Great job 💪

Issue #12163, opened 2024-04-26
node-bridge "Segmentation fault: 11"


mroz22:
getting this error time to time 
```
Segmentation fault: 11
```

when running node-bridge. using mac arm, node 18. hope it will go away in node20 🙏

mroz22:
- node --version v20.12.2
- connect 1 device (model One in my case)
- run node-bridge `node packages/transport-bridge/dist/bin.js
- open 2 suites at once `open http://localhost:8000 http://localhost:8000`

get segmentation fault: 11 error

log:

```
@trezor/transport-bridge Handling request for POST /
@trezor/transport-bridge Handling request for POST /enumerate
@trezor/transport-bridge usb: enumerate
@trezor/transport-bridge usb: enumerate done. connected devices.length: 5. trezor old (hid) devices: 0. trezor devices: 1
@trezor/transport-bridge usb: creating device {"productName":"TREZOR","manufacturerName":"SatoshiLabs","serialNumber":"185B982B5F37F9D96706EC49","vendorId":4617,"productId":21441,"deviceVersionMajor":1,"deviceVersionMinor":0,"opened":false,"deviceProtocol":0,"deviceClass":0,"deviceSubclass":0,"usbVersionMajor":2,"usbVersionMinor":1,"usbVersionSubminor":0}
@trezor/transport-bridge usb: loadSerialNumber
@trezor/transport-bridge usb: loadSerialNumber done, serialNumber: 185B982B5F37F9D96706EC49
@trezor/transport-bridge http: sessionsClient reported descriptors: [{"path":"185B982B5F37F9D96706EC49","type":1,"product":21441,"session":null}]
@trezor/transport-bridge Handling request for POST /listen
@trezor/transport-bridge Handling request for POST /acquire/185B982B5F37F9D96706EC49/null
@trezor/transport-bridge usb: openDevice attempt 0
@trezor/transport-bridge usb: device.open
@trezor/transport-bridge usb: device.open done. device: {"productName":"TREZOR","manufacturerName":"SatoshiLabs","serialNumber":"185B982B5F37F9D96706EC49","vendorId":4617,"productId":21441,"deviceVersionMajor":1,"deviceVersionMinor":0,"opened":true,"deviceProtocol":0,"deviceClass":0,"deviceSubclass":0,"usbVersionMajor":2,"usbVersionMinor":1,"usbVersionSubminor":0}
@trezor/transport-bridge usb: device.selectConfiguration 1
[...]

PR #13220, opened 2024-07-07
transport: propagate hid devices descriptors

mroz22:
With the old bridge going out of service, we are no longer able to work with very old Trezor HID devices. We are only able to detect them but no communication is possible. 

To test this, you need an ancient device.
```
Features (126 bytes) {
    bootloader_hash: 32 bytes 0xc4c32539b4a025a8e753a4c46264285911a45fcb14f4718179e711b1ce990524,
    device_id: 'F7F53817FDC08323ECA22200',
    imported: False,
    initialized: True,
    label: 'My Trezor',
    major_version: 1,
    minor_version: 2,
    passphrase_protection: False,
    patch_version: 0,
    pin_protection: False,
    revision: 20 bytes 0xdf524b9f35fd5cdba14eaa2bf2d948e3dc75254a,
    vendor: 'bitcointrezor.com',
}
```

This PR unifies how these devices are communicated for both bridge and webusb transports. 

- c8712c05da7afaecb9bb5c730957dc546d29affa - transport now propagates descriptors even for hid devices to higher layers. connect then tries to work with these descriptors resulting into 'Unable to open device' error. which already results in creation of an 'Unreadable device'. 
- 719aa3962995396089e1a7085375a078846fe1ff - a couple of fixes were needed to make it behave the same way for node-bridge

While testing this for possible regressions I noticed a[ nasty bug](https://github.com/trezor/trezor-suite/issues/12163#issuecomment-2212429718) that we might need to address before releasing node-bridge publically

Screenshots of how it looks like in trezor suite now. cc @martykan since he is now addressing this #12487, cc @Hannsek since we might need some designs/copy. Those bullet points are not very accurate. 
image

image

### DIFF
--- a/packages/transport/src/api/usb.ts
+++ b/packages/transport/src/api/usb.ts
@@ -44,28 +44,26 @@ export class UsbApi extends AbstractApi {
     }
 
     public listen() {
-        this.usbInterface.onconnect = async event => {
+        this.usbInterface.onconnect = event => {
             this.logger?.debug(`usb: onconnect: ${this.formatDeviceForLog(event.device)}`);
             const [_hidDevices, nonHidDevices] = this.filterDevices([event.device]);
 
             _hidDevices.forEach(() => {
-                // hidDevices that do not support webusb. these are very very old. we used to emit unreadable
-                // device for these but I am not sure if it is still worth the effort.
+                // hidDevices that do not support webusb standard. emit them as normal descriptors. higher layers are responsible
+                // for displaying them as unreadable devices
                 this.logger?.error(
                     `usb: unreadable hid device connected. device: ${this.formatDeviceForLog(event.device)}`,
                 );
             });
-            if (nonHidDevices.length) {
-                await this.createDevices(nonHidDevices, this.abortController.signal)
-                    .then(newDevices => {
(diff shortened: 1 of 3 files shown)


Hannsek:
The bullet points are changed in [this comment](https://github.com/trezor/trezor-suite/issues/12487#issuecomment-2134804972).

packages/transport/src/api/usb.ts
```diff
@@ -395,10 +393,28 @@ export class UsbApi extends AbstractApi {
         return device.device;
     }
 
-    private createDevices(nonHidDevices: USBDevice[], signal?: AbortSignal) {
+    private async createDevices(
+        nonHidDevices: USBDevice[],
```
szymonlesisz: we can get rid of code duplicates from `enumetate` and `listen` functions by passing all the devices here and do filterDevices before `const loadedDevices = await Promise.all(`

mroz22: cool, added as a new commit 5d893eedb877d12d5e9ec17a85ad5c5aa7885a7c

szymonlesisz (approved):

PR #13224, opened 2024-07-08
fix(suite): apply disabled based on deviceLocked on a button that iss…

mroz22:
This "renew" button should imho be disabled on click. 

image

If not disabled, you mess up your app state (locks) and probably the only way out of this is to reload the application. 

image

image


jvaclavik (approved):
Good point to fix this 👍

PR #13229, opened 2024-07-08
feat(protobuf): update protobuf messages for newest version

### LINKED ISSUE #13065: Analytics for `optiga_sec`
Starting with the next firmware release v2.8.0 Trezor will include a new field `optiga_sec` in `Features`, which contains the value of Optiga's security event counter. This field appears in devices that have the Optiga secure element, i.e. TS3 and TS5. The possible values are integers from 0 to 255 or empty. Let's collect the value of this field from users who have opted in to providing analytics data. The value may change in time. So I propose that we collect the value from the first `Features` message that Suite receives from the device after it's connected.

Related to https://github.com/trezor/trezor-firmware/pull/3977.

peter-sanderson:
Adds a field to analytics for `optiga_sec` field. So we have reports when this counter starts to increase.

Resolves: https://github.com/trezor/trezor-suite/issues/13065

image

### DIFF
--- a/packages/protobuf/messages.json
+++ b/packages/protobuf/messages.json
@@ -673,12 +673,10 @@
                             "id": 3
                         },
                         "mask_public_key": {
-                            "rule": "required",
                             "type": "bytes",
                             "id": 4
                         },
                         "signature": {
-                            "rule": "required",
                             "type": "bytes",
                             "id": 5
                         }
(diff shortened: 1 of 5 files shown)


komret (approved):

karliatto:
Where do we set the required tests to merge?

This PR is modifying protobufs and there is a connect core api test failing https://github.com/trezor/trezor-suite/actions/runs/9837671273/job/27156442512 
I do not think the issue is in this PR, but I think in this case this should have been block by that test.

PR #13233, opened 2024-07-08
fix(suite): remove the disabling of the sandbox in Electron

peter-sanderson:
This PR enables Sandbox for the Browser process. We have done a testing for Tor & Sentry and it seems it works.

So we have not found a reason why to keep it disabled.

### DIFF
--- a/packages/suite-desktop-core/src/app.ts
+++ b/packages/suite-desktop-core/src/app.ts
@@ -37,9 +37,6 @@ const createMainWindow = (winBounds: WinBounds) => {
             webSecurity: !isDevEnv,
             allowRunningInsecureContent: isDevEnv,
             preload: path.join(__dirname, 'preload.js'),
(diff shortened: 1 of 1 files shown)


peter-sanderson:
@tomasklim The lines I am now removing were added by you, do you have any objections? :) => Can you recall any other reason we shall keep it or **do you know about some flow that we shall test?**

Test we did:

1. we build the app so we are in prod environment,
2. we enabled Tor 
3. and then we were triggering the (`throw new Error`) in the Debug section and checking if the error reaches the Sentry.


We hit some situation when Sentry error did not went through but it was not consistent an happend in non-sanboxed build as well, so we assume it is just Sentry blocking some exit-nodes.

tomasklim (commented):
I don't remember it sorry

komret (approved):
Let's see if anything breaks.

PR #13234, opened 2024-07-08
fix(suite): Make Eject icon bigger when asking to "Eject this wallet"

hstastna:
## Description

Explicitly set `iconSize` prop for `Button` icon to display _EJECT_ icon properly, when asking to "Eject this wallet?".

Usually, for "small" size the icon size is set default to 16px, but for this icon we need it slightly bigger - 20px do be displayed nicely. 

## Screenshots:
**Before:**
The eject icon looks cut off:
eject_before
**After:**

The eject icon displayed whole, as expected:
eject_after


peter-sanderson (approved):

jvaclavik (approved):
👏

PR #13237, opened 2024-07-08
fix(suite): Fix images in storybook 🥳

jvaclavik:
## Description

- fix imports for static files
- add Image storybook
- update icon selects in stories

## Related Issue

Resolve 

## Screenshots:

add image stories:
image

fix images:
before:
image

after:
image

### DIFF
--- a/packages/components/.storybook/main.js
+++ b/packages/components/.storybook/main.js
@@ -9,7 +9,7 @@ module.exports = {
         getAbsolutePath('@storybook/addon-viewport'),
         getAbsolutePath('@storybook/addon-actions'),
     ],
-    staticDirs: ['../public'],
+    staticDirs: ['../public', { from: '../../suite-data/files', to: '/static' }],
     framework: {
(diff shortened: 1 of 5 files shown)


packages/components/.storybook/main.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
         getAbsolutePath('@storybook/addon-viewport'),
         getAbsolutePath('@storybook/addon-actions'),
     ],
-    staticDirs: ['../public'],
+    staticDirs: ['../public', { from: '../../suite-data/files', to: '/static' }],
```
peter-sanderson, line 12: This goes outside of the package. Is it ok? Maybe we can ask @Nodonisko if this is ok, or how to do it correctly. :pray:

Nodonisko: Since this is basically static path in bundler settings and not direct dependency I think it's fine

peter-sanderson (approved):

PR #13238, opened 2024-07-08
chore(message-system): added ts5 compatibility suite version

MiroslavProchazka:
Cleanup of some old messages 
Bumped minimal recommended Suite version to 23.7.2 
Added message to inform user to download newer Suite for TS5 devices
Added T3T1 and TS5 to config.schema.v1.json


matejkriz (approved):
It looks good, but I haven't tested it.

STew790:
**QA OK**
Messages and message system is correct.

PR #13245, opened 2024-07-09
chore(suite): accept device update when changed from 'unacquired' to …

mroz22:
cherry-picking from #13227

you might get into a situation where device **becomes** unreadable and it can't be communicated with. it simply hangs somewhere in `device.transferOut` call. but since it wasn't unreadable some time ago, it might have been acquired by somebody and never released. 
This takes you to the screen where you have the "Use device here" button. Clicking the button triggers a connect call, which will fail (something I need yet to solve in #13277) and the device should be changed from unacquired to unreadable. 

this was not possible till now as suite would not accept changes of device that do not have features. 

discussed briefly with @szymonlesisz, he is ok with this change. May I assign someone, lets say @komret ?

### DIFF
--- a/suite-common/wallet-core/src/device/deviceReducer.ts
+++ b/suite-common/wallet-core/src/device/deviceReducer.ts
@@ -206,6 +206,27 @@ const changeDevice = (
     device: Device | TrezorDevice,
     extended?: Partial<AcquiredDevice>,
 ) => {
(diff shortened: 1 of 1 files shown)


komret (commented):
Can you add this to Suite Issues for QA to see this and also some reproduction steps for them?

mroz22:
image

PR #13251, opened 2024-07-09
feat(suite): Add cursor: not-allowed to some components when disabled

### LINKED ISSUE #12945: Add cursor:not-allowed to disable components

hstastna:
## Description

Add `cursor: not-allowed` for cursor to look nicer to some more of the interactive/input components in our design system, when being disabled.

Additionally, add args and argTypes in storybook for components that are missing them so one could easily change the values of selected component's props in storybook.

- `AutoscalingInput`
- `Range`
- `Select`
- `SelectBar`
- `Switch`

Also for `Button` component, add way to choose from all the available icons in storybook, incl. choosing the option "No icon" in storybook, as the _icon_ prop is optional.

For `AutoscalingInput`, add story for playing with its props, use the existing one to just show the examples.

TODO (in followup PR):
- [ ] `Textarea`
- [ ] `IconButton`
- [ ] `PinButton`
- [ ] `TextButton`
- [ ] `TooltipButton`

## Related Issue

## Screenshots:
**Before:**
Example: `Label` component
label_before

**After:**
Example: `Label` component: updated cursor when disabled + added more controls to storybook
label_after

Button component: now you can easily choose an icon if you want, or also not:
icons

### DIFF
--- a/packages/components/src/components/form/Select/Select.tsx
+++ b/packages/components/src/components/form/Select/Select.tsx
@@ -217,6 +217,13 @@ const Wrapper = styled.div<WrapperProps>`
         border: none;
         z-index: ${zIndices.base};
     }
+
+    ${({ $isDisabled }) =>
+        $isDisabled &&
+        css`
+            pointer-events: auto;
+            cursor: not-allowed;
+        `}
 `;
 
 const SelectLabel = styled(Label)`
@@ -272,7 +279,7 @@ export const Select = ({
     components,
     onChange,
     placeholder,
-    isDisabled,
+    isDisabled = false,
     'data-test': dataTest,
     ...props
 }: SelectProps) => {
--- a/packages/components/src/components/form/SelectBar/SelectBar.tsx
+++ b/packages/components/src/components/form/SelectBar/SelectBar.tsx
@@ -118,7 +118,8 @@ const Option = styled.div<{ $isSelected: boolean; $isDisabled: boolean }>`
         $isDisabled &&
         css`
             color: ${({ theme }) => theme.textDisabled};
-            pointer-events: none;
+            pointer-events: auto;
+            cursor: not-allowed;
         `}
 `;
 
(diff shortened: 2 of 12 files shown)


packages/components/src/components/form/SelectBar/SelectBar.tsx
```diff
@@ -216,7 +217,7 @@ export const SelectBar: <V extends ValueTypes>(props: SelectBarProps<V>) => JSX.
                 {options.map(option => (
                     <Option
                         key={String(option.value)}
-                        onClick={handleOptionClick(option)}
+                        onClick={handleOptionClick(isDisabled, option)}
```
peter-sanderson: I think `isDisabled` dont need to be passed as argument, it is already available in the handler from component props.

packages/components/src/components/form/Select/Select.tsx
```diff
@@ -217,6 +217,8 @@ const Wrapper = styled.div<WrapperProps>`
         border: none;
         z-index: ${zIndices.base};
     }
+
+    ${({ $isDisabled }) => $isDisabled && `pointer-events: auto; cursor: not-allowed;`};
```
peter-sanderson: This maybe works but I think it should be: 

```ts
    ${({ $isDisabled }) =>
        $isDisabled &&
        css`
            pointer-events: auto;
            cursor: not-allowed;
        `};
```

Or at least we use it this way everywhere else. Also with this 
```
css`...`
```
 auto-format works and it then places every CSS directive on separate line.

hstastna: Right! And it's more readable like this. Don't know why I've used that one line format here instead :)

packages/components/src/components/form/SelectBar/SelectBar.tsx
hstastna: I think I had a good reason for this. But I'll take a look again. Now seems that `isDisabled` really doesn't need to be there. Thanks!

peter-sanderson (approved):

PR #13255, opened 2024-07-09
fix(suite-native): unlock device mutex on callback rejection

vytick:
When `requestDeviceAccess` or `requestPrioritizedDeviceAccess` failed, device mutex was not unlocked causing next call not receiving the lock.

### DIFF
--- a/suite-native/device-mutex/src/requestDeviceAccess.ts
+++ b/suite-native/device-mutex/src/requestDeviceAccess.ts
@@ -5,34 +5,42 @@ import { DeviceAccessResponse } from './types';
 /**
  * Puts the callback to the end of the queue and waits for its turn to execute.
  */
-export const requestDeviceAccess = async <TParams extends unknown[], TReturnType>(
-    deviceCallback: (...args: TParams) => TReturnType,
-    ...callbackParams: TParams
-): DeviceAccessResponse<TReturnType> => {
-    const wasLockSuccessful = await deviceAccessMutex.lock();
+export const requestDeviceAccess = async <TParams extends unknown[], TReturnType>({
+    deviceCallback,
+    isPrioritized = false,
+    callbackParams = [] as unknown as TParams,
+}: {
+    deviceCallback: (...args: TParams) => TReturnType;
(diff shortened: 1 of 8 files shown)


PeKne:
Thanks for fixing this 👏. There is a [test file](https://github.com/trezor/trezor-suite/blob/44fa12a79d275c325e1dd10a1703baebba56e546/suite-native/device-mutex/src/tests/DeviceAccessMutex.test.ts) testing these functions. Could you please add a unit test of this "error-throwing" scenario? It seems to be quite crucial.

juriczech (changes requested):
I know you're not an author of this code, but since you're adding more code to this function, I cannot help it, these functions are IDENTICAL apart from the method which they are calling on the `deviceAccessMutex` class. 

This can be fixed by passing a parameter for prioritization. Shouldn't be a hard fix, you'll just have to fix the usages once you merge this. I recommend doing a parameter with default value. Preferably named parameters, since it might be boolean, so it's not just plain `true / false` when you call it. 

If you'd like to avoid changing the usages, you could also just change the first function to take an optional parameter with default `false` (so non-priority access) and the second one will just call the first one with `true` so that the usages are maintained and there are 2 different names - but the functionality is only defined in one place.

Another thing to consider - maybe the class method itself could take a parameter and handle locking according to priority.

Anyway, good job finding and fixing this 💪

Nodonisko:
I don't think it's necessarily problem that function repeat itself twice, but if it could be easily changed into one function let's do it.

I am writing this only to warn you to not fall into DRY trap, because some people tried to pushed us DRY everywhere and it was big mistake. Some duplication is not problem in most cases, but premature abstraction to avoid duplication is problem very often.

juriczech:
@Nodonisko Yeah, I considered this before writing the comment. I think in this case, it's worth the rewrite since it's really simple IMO. Generally, I agree it wouldn't be worth if there would be more code that's not same 👍

juriczech (approved):
LGTM 👏

vytick:
@PeKne @juriczech updated in 93349ec4d4

PR #13226, opened 2024-07-08
feat(suite): remove RBF option from the SendForm, make all Transactions with RBF enabled by default

peter-sanderson:
Resolves https://github.com/trezor/trezor-suite/issues/13160

PR #13268, opened 2024-07-10
fix(suite): replace RBF tx with locktime

szymonlesisz:
## Description

Since [rbf is always enabled](https://github.com/trezor/trezor-suite/pull/13226) it is possible to create rbf + locktime transaction.
There was a TODO for that in RBF related actions.

 
## Steps (with user-env)
- load regtest account
- send yourself some coins using user-env
- go to send form
- set locktime to network currentBlock or lower (probably 151? verify in regtest explorer)
- enable broadcasting (nitpick: this should be enabled since broadcasting is possible)
- send some coins anywhere
- go to dashboard, and hit "replace transaction"
- go thru process
- tx should be successfully replaced 

## Resolve

bug on current develop, replacement tx is rejected with missing `nLocKTime` error


peter-sanderson (approved):
:clap: Thanks!

PR #13269, opened 2024-07-10
fix(connect): handle missing language binary

matejkriz:
## Description

There is a new bug in Suite Lite - Trezor ask user to confirm language change when connected and by other interactions with the device too.
image

steps to reproduce:
You have to be on previous FW version (i.e. 2.7.2) in language other than EN
Update to 2.8.0 FW
Connect Trezor to Suite Lite

Downloading language binary on Suite Lite is not working at all and failed fetch handling was missing.

This only handles the error state, so it is a workaround solution for Suite Lite - user will not see Language change screen.

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -581,6 +581,10 @@ export class Device extends TypedEmitter<DeviceEvents> {
             internal_model: this.features.internal_model,
         });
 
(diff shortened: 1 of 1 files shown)


mroz22 (approved):

STew790:
**QA OK**
I was not prompted to change the language on TS3 or TS5.

24.7.1   79c6bf0

PR #13272, opened 2024-07-10
feat(suite): Save sidebar width to SuiteSettings to be persisted

### LINKED ISSUE #13055: Persisted sidebar width
It should be saved in `SuiteSettings` (`packages/suite/src/reducers/suite/suiteReducer.ts`) - it's synced with IndexDB

save it on mouse up, not every mouse move event

hstastna:
## Description

After resizing the sidebar on mouse up event, save the actual sidebar width so after disconnecting and reconnecting Trezor and visiting Suite again, the user can see their custom sidebar width and not the default one (280px).

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13055

## Screenshots:
Changing the default sidebar width (for example to maximum 400px):
change_width

**Before:**
After reconnecting, sidebar width back to the default:
width_before

**After:**
width_after

### DIFF
--- a/packages/components/src/components/ResizableBox/ResizableBox.tsx
+++ b/packages/components/src/components/ResizableBox/ResizableBox.tsx
@@ -19,6 +19,8 @@ export type ResizableBoxProps = {
     updateWidthOnWindowResize?: boolean;
     updateHeightOnWindowResize?: boolean;
     zIndex?: ZIndexValues;
+    onWidthResizeEnd?: (width: number) => void;
+    onHeightResizeEnd?: (height: number) => void;
 };
 
 type ResizerHandlersProps = {
@@ -182,6 +184,8 @@ export const ResizableBox = ({
     updateWidthOnWindowResize = false,
     updateHeightOnWindowResize = false,
     zIndex = zIndices.draggableComponent,
+    onWidthResizeEnd,
+    onHeightResizeEnd,
 }: ResizableBoxProps) => {
     const resizableBoxRef = useRef<HTMLDivElement>(null);
 
@@ -272,7 +276,15 @@ export const ResizableBox = ({
             }
         };
 
-        document.onmouseup = () => setIsResizing(false);
+        document.onmouseup = () => {
+            setIsResizing(false);
+            if (onWidthResizeEnd) {
+                onWidthResizeEnd(newWidth);
+            }
+            if (onHeightResizeEnd) {
+                onHeightResizeEnd(newHeight);
+            }
+        };
 
         window.onresize = () => {
             if (resizeCooldown() === true) {
@@ -292,6 +304,8 @@ export const ResizableBox = ({
         maxWidth,
         newHeight,
         newWidth,
+        onHeightResizeEnd,
+        onWidthResizeEnd,
         resizableBoxRef,
         resize,
         resizeCooldown,
--- a/packages/suite/src/actions/suite/suiteActions.ts
+++ b/packages/suite/src/actions/suite/suiteActions.ts
@@ -75,7 +75,8 @@ export type SuiteAction =
           payload: Partial<AutodetectSettings>;
       }
     | { type: typeof deviceActions.requestDeviceReconnect.type }
-    | { type: typeof SUITE.SET_EXPERIMENTAL_FEATURES; payload?: ExperimentalFeature[] };
+    | { type: typeof SUITE.SET_EXPERIMENTAL_FEATURES; payload?: ExperimentalFeature[] }
+    | { type: typeof SUITE.SET_SIDEBAR_WIDTH; payload: { width: number } };
 
 export const appChanged = createAction(
     SUITE.APP_CHANGED,
@@ -128,6 +129,11 @@ export const initialRunCompleted = () => (dispatch: Dispatch, getState: GetState
         dispatch(setFlag('initialRun', false));
     }
 };
+
+export const setSidebarWidth = (payload: { width: number }): SuiteAction => ({
+    type: SUITE.SET_SIDEBAR_WIDTH,
+    payload: { width: payload.width },
+});
 
 /**
  * Triggered by `@suite-support/OnlineStatus` or `@suite-native/support/OnlineStatus`
--- a/packages/suite/src/actions/suite/constants/suiteConstants.ts
+++ b/packages/suite/src/actions/suite/constants/suiteConstants.ts
@@ -31,3 +31,4 @@ export const LOCK_TYPE = {
 } as const;
 export const REQUEST_DEVICE_RECONNECT = '@suite/request-device-reconnect';
 export const SET_EXPERIMENTAL_FEATURES = '@suite/set-experimental-features';
+export const SET_SIDEBAR_WIDTH = '@suite/set-sidebar-width';
--- a/packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/Sidebar.tsx
+++ b/packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/Sidebar.tsx
@@ -5,8 +5,9 @@ import { Navigation } from './Navigation';
 import { AccountsMenu } from 'src/components/wallet/WalletLayout/AccountsMenu/AccountsMenu';
 import { QuickActions } from './QuickActions';
 import { ElevationUp, ResizableBox, useElevation } from '@trezor/components';
-import { SIDEBAR_WIDTH_NUMERIC } from 'src/constants/suite/layout';
 import { Elevation, mapElevationToBackground, mapElevationToBorder, zIndices } from '@trezor/theme';
+import { useActions, useSelector } from 'src/hooks/suite';
+import * as suiteActions from 'src/actions/suite/suiteActions';
 
 const Container = styled.nav<{ $elevation: Elevation }>`
     display: flex;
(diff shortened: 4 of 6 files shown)


packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -19,6 +19,7 @@ export type ResizableBoxProps = {
     updateWidthOnWindowResize?: boolean;
     updateHeightOnWindowResize?: boolean;
     zIndex?: ZIndexValues;
+    onSaveWidth?: (width: number) => void;
```
jvaclavik: Please update it to:
``` 
onResizeEnd?: (width: number, height: number) => void
```

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -272,7 +274,12 @@ export const ResizableBox = ({
             }
         };
 
-        document.onmouseup = () => setIsResizing(false);
+        document.onmouseup = () => {
+            setIsResizing(false);
+            if (onSaveWidth) {
+                onSaveWidth(newWidth);
```
jvaclavik: We should provide the same functionality also for height

packages/suite/src/actions/suite/suiteActions.ts
```diff
@@ -75,7 +75,8 @@ export type SuiteAction =
           payload: Partial<AutodetectSettings>;
       }
     | { type: typeof deviceActions.requestDeviceReconnect.type }
-    | { type: typeof SUITE.SET_EXPERIMENTAL_FEATURES; payload?: ExperimentalFeature[] };
+    | { type: typeof SUITE.SET_EXPERIMENTAL_FEATURES; payload?: ExperimentalFeature[] }
+    | { type: typeof SUITE.SET_SIDEBAR_WIDTH; payload: number };
```
jvaclavik: Could you update it to 
```
| { type: typeof SUITE.SET_SIDEBAR; payload: {width: number} };
```
?

I know it's not like this in other actions, but this kicked us multiple times, so we should use payload as an object, not direct value.

packages/components/src/components/ResizableBox/ResizableBox.tsx
hstastna: Could you please provide more context about the suggestion? I'd avoid using "end" in this prop naming.

packages/components/src/components/ResizableBox/ResizableBox.tsx
hstastna: That's new info for me. But ok. Will add it.

packages/components/src/components/ResizableBox/ResizableBox.tsx
jvaclavik: - `onResize`: function called when you are resizing the box (actually it doesn't matter if it's saving or not, it's not job for this component)
- `End`: It's action called on mouseup (end of mouse click)

packages/suite/src/actions/suite/suiteActions.ts
hstastna: Could you please provide more context also about this suggestion?

packages/components/src/components/ResizableBox/ResizableBox.tsx
jvaclavik: You can do it in one function:
```
onResizeEnd?: (width: number, height: number) => void
```

When we add some feature for width, we need the same feature also for height, otherwise it's not going to be consistent and predictable component.

packages/suite/src/actions/suite/suiteActions.ts
jvaclavik: When we add new data to payload we don't have to refactor all old usages if we use payload as an object.

packages/components/src/components/ResizableBox/ResizableBox.tsx
hstastna: Makes totally sense in case we want to let the component accept some function that is gonna simply be executed on mouse up, no matter what it is. But I think this is more specific case, added for specific purpose, and the function accepting specific prop(s). Hence we should be more specific also in naming.. if we don't decide we want to be super universal here. I'm not sure yet as both approaches make sense and have both pros and cons.

jvaclavik: I don't agree that this is the special use case. It's `onChange` function that is called once in the end of the click, nothing else. You don't have any special logic related to saving. What name would you suggest?

packages/components/src/components/ResizableBox/ResizableBox.tsx
hstastna: Makes sense. Just.. it's gonna be unused (and hence unnecessary). Because for the sidebar, it doesn't make sense to save its height to SuiteSettings or doing anything else with the height on mouse up that isn't already implemented in ResizableBox.

jvaclavik: Yes, height is going to be unused in case of sidebar and it's totally fine. But in our design system which should be generic, we don't want to have components with half functionality.

packages/suite/src/actions/suite/suiteActions.ts
hstastna: Right. Just this is not applicable for this specific case I am adding here because it doesn't make sense to provide anything else than a number for sidebar width. If we added more data to payload for this action in the future, we should also adjust the action's name to reflect its purpose.

jvaclavik: OK, it doesn't matter much here.

packages/components/src/components/ResizableBox/ResizableBox.tsx
hstastna: On the other hand it is not the best practice to add unnecessary functionality just in case it might be used sometimes, so we should be a bit careful in such situations as now. Fortunately this is a very small change and maybe possibly useful in the future so I don't mind.

packages/components/src/components/ResizableBox/ResizableBox.tsx
hstastna: It is specific because the function works with `newWidth` actually (or will work with the box dimensions once I push the update for the height). The name should reflect its use case, shouldn't be totally universal because there's so much other stuff we could do with the box on mouse up, for example to adjust color, highlighting, focus etc. But we don't want to work with those.

Not sure about the exact name yet. But using "end" isn't the best as it can also make one feel it is gonna be something related to some specific part of the box, like it's for css when "flex-end" value is used, for example. So I'd be careful to prevent potential confusions, maybe "onResizeFinish" would be good enough. Even better would be if it reflected that it has something to do with the dimensions of the box (and not color or z-index etc etc). The more specific we are, the more control we have over the behavior, the better usability of the component, less bugs etc.

jvaclavik: Please don't overthink stuff. I don't want to spend so much time on discussing naming all the time. If you think I'm wrong try to ask someone else in the team.

I still think `onResizeEnd` is valid here. In arguments, you are reflecting the state of the component in the moment when you ended the touch. If there will be more values than width and height, we will create an object for it. 
Also, `End` suffix is widely used across events like this. For example in official API for touch events: https://developer.mozilla.org/en-US/docs/Web/API/Element/touchend_event

hstastna: > Please don't overthink stuff. I don't want to spend so much time on discussing naming all the time. If you think I'm wrong try to ask someone else in the team.

Thank you for sharing your opinion. Without discussion or having different opinion, I'd learn nothing. I really appreciate you invest time in the discussion and provide ideas, thoughts, information.

> I still think onResizeEnd is valid here. In arguments, you are reflecting the state of the component in the moment when you ended the touch. If there will be more values than width and height, we will create an object for it.
Also, End suffix is widely used across events like this. For example in official API for touch events: https://developer.mozilla.org/en-US/docs/Web/API/Element/touchend_event

From this point of view, 'onResizeEnd' makes more sense to me now. And it's great to inspire there when naming functions. On the other hand, we're still adding slightly different thing to ResizableBox here, we're trying to get some data to save it in the Redux store. My original naming isn't the best but I'd rather not to mix Redux stuff with anything else. 'onResizeEnd' seems also to me too general comparing to that, general naming for a very specific use case, as it won't work for anything else than saving the width (and/or height) - this can be misleading. So I'm not sure..

To sum up, yes, naming can be hard and also a very subjective topic. Work in progress. And that's why it's good to discuss. But let's step away from it for a while to prevent spending "so much time on discussing naming all the time" and let's see if anything better comes to our minds ;)

packages/suite/src/actions/suite/suiteActions.ts
peter-sanderson: I recognize this will probably never change (I don't see situation when we want to add more fields), but I like consistency and just for this sake of consistency I would prefer to have it wrapped in `{}`,

packages/components/src/components/ResizableBox/ResizableBox.tsx
peter-sanderson: This component already handles `height` resize which we are currently not using. 

Either:
1. Lets remove height logic (its symmetrical logic to width logic) and its very easy to add later (just by copy-pasting the width).
2. Or lets threat it with same level of completeness as `width` and lets have the `onXXXX` handler for both.

As `width` and `height` never changes together (we don't have corner drag implemented, and I assume we dont plan to) -> therefore I suggest to have 2 callbacks: `onWidthXXX` and `onHeightXXX`



Subjectively I prefer option 2.

packages/components/src/components/ResizableBox/ResizableBox.tsx
peter-sanderson: > ...The name should reflect its use case...

In general I found very useful to name things base on *what* they are, rather then *how* are they used. Component is placed in the separate context (package) of abstraction and has its API which I feel shall be rather generic. Reason for this is to not leak the suite/our-current-usage domain into the component domain (which can be used elsewhere, for example in the `connect-explorer`).

peter-sanderson: Both `onResizeFinish` and `onResizeEnd` are totally fine for me.

jvaclavik (approved):

PR #13273, opened 2024-07-10
`DeviceList` revamp

marekrjpolak:
## Description

This is part of big Connect refactoring, this time related to `DeviceList`. The goal was to make it something like a singleton, instead of destroying and recreating it whenever a transport error appears. I've tried to make very small commits so it should be relatively easy to review them.

## Commits
- https://github.com/trezor/trezor-suite/pull/13273/commits/0d0d37befb5be799664315108ac6ec74433c6c18 - making as many device list's fields private as possible
- https://github.com/trezor/trezor-suite/pull/13273/commits/5cdca733f581a5c919c7e367864ffb36c86c62e9 - separate transports creation logic from device list's constructor, so it can be moved in the next commit
- https://github.com/trezor/trezor-suite/pull/13273/commits/7d6ad626b415b5d04d83a903dc2a3afa97dc9e48 - device list constructor now doesn't create transports (well, except the default Bridge one so there's always one) so it should never throw and can live forever; transport creation logic moved to `setTransports()` which can throw without the chance to recover
- https://github.com/trezor/trezor-suite/pull/13273/commits/c65f97a061d155447521f38f70166bbf9cbf2a36 - top-level transport reconnection in core index removed because it doesn't make sense on `setTransports` fail; the only reconnection happens on `TRANSPORT.ERROR`
- https://github.com/trezor/trezor-suite/pull/13273/commits/7d651e64fd91838e79b91353a719739f7158a837 - `TRANSPORT.UPDATE` handler separated into its own method
- https://github.com/trezor/trezor-suite/pull/13273/commits/76894b55644ebeaaa2005d82bc724ea53dc100d4 - isolate waiting for devices in case of `pendingTransportEvent` into a single, well-encapsulated method
- https://github.com/trezor/trezor-suite/pull/13273/commits/606f541f3d0974aca0708ef826a0e7357d54faea - separate init method into stages (select/initialize) which throw errors when necessary, and emit appropriate transport event at the end, so `waitForFirstTransportEvent` is no longer needed
- https://github.com/trezor/trezor-suite/pull/13273/commits/bcf8bd7cf679b9386fd267b77ed4c569372d5d7e - device penalization logic separated into `authPenaltyManager`
- https://github.com/trezor/trezor-suite/pull/13273/commits/552b22cdfad7d0f266fd06a28d42d4465282f4e7 - tiny code improvement (less lines, yaay)
- https://github.com/trezor/trezor-suite/pull/13273/commits/651ccf431619ae8d88b0595f4d7c480368610936 - transport instance is now propagated into the whole `onTransportUpdate` subtree instead of using the class instance one which may or may not be there, which should make it a bit more resilient; **this subtree should be somehow separated from the class in the future**
- https://github.com/trezor/trezor-suite/pull/13273/commits/4b4cbdc81c080c6b9fe76860af154210f1973dd2 - device list can now be connected (when there is active transport) or disconnected (when there isn't); methods requiring active transport should be called only on connected device list; device list connection status is now checked in connect core instead of its presence, so it's now created only once on init
- https://github.com/trezor/trezor-suite/pull/13273/commits/10dec95103d1c0af6fe80b350085815103530e70 - device list is now cleaned up from inside on transport error instead of disposed from outside (e.g. all the listeners are preserved); therefore it can be reused
- https://github.com/trezor/trezor-suite/pull/13273/commits/b87c55f94a5245d9c2783410f175708a1c32e287 - `resolveAfter` util now returns `reject` instead of `timeout`, so it can be aborted without memory leaking (as opposed to clearing the timeout and letting the promise hang forever)
- https://github.com/trezor/trezor-suite/pull/13273/commits/a4d982fe2a2ffd3dc53410e13b7ba3960dcb14eb - transport reconnect logic moved inside device list; in case of transport error, new init promise (including delay) is immediately created and awaitable from outside; **decide what to do when init is called before the previous init finished**
- https://github.com/trezor/trezor-suite/pull/13273/commits/16ad4ac79587edc7fc92af5c84ad7a3c50317cb3 - device list initialization methods are called directly instead of wrapping them inside `initDeviceList`, which should be a bit clearer
- https://github.com/trezor/trezor-suite/pull/13273/commits/14c68b01d0334647bdaea86333629e875a8cfea1 - unit test adjustments

**MUST BE TESTED THOROUGHLY**

### DIFF
--- a/packages/connect/src/core/index.ts
+++ b/packages/connect/src/core/index.ts
@@ -991,18 +991,7 @@ const initDeviceList = async (transportReconnect?: boolean) => {
         throw error;
     }
 
-    // if transport fails during app lifetime, try to reconnect
-    if (transportReconnect) {
-        _deviceList.once(TRANSPORT.ERROR, () => {
-            const { promise, reject } = resolveAfter(1000, null);
-            _deviceListInitReject = reject;
-            promise.then(() => {
-                initDeviceList(transportReconnect);
-            });
-        });
-    }
-
-    _deviceList.init(pendingTransportEvent);
+    _deviceList.init({ pendingTransportEvent, transportReconnect });
     await _deviceList.pendingConnection();
 };
 
@@ -1191,6 +1180,7 @@ const disableWebUSBTransport = async () => {
         // clean previous device list
         _deviceList.cleanup();
         //and init with new settings, without webusb
+        // TODO possible issue with new init not replacing the old one???
         await initDeviceList(settings.transportReconnect);
     } catch (error) {
         // do nothing
--- a/packages/connect/src/core/index.ts
+++ b/packages/connect/src/core/index.ts
@@ -964,10 +964,12 @@ const handleDeviceSelectionChanges = (interruptDevice?: DeviceTyped) => {
  */
 const initDeviceList = async (transportReconnect?: boolean) => {
     try {
-        _deviceList = new DeviceList({
-            settings: DataManager.getSettings(),
-            messages: DataManager.getProtobufMessages(),
-        });
+        const { debug, transports, priority, pendingTransportEvent } = DataManager.getSettings();
+        const messages = DataManager.getProtobufMessages();
+
+        _deviceList = new DeviceList({ debug, messages, priority });
+
+        _deviceList.setTransports(transports);
 
         _deviceList.on(DEVICE.CONNECT, device => {
             handleDeviceSelectionChanges();
@@ -1013,7 +1015,7 @@ const initDeviceList = async (transportReconnect?: boolean) => {
             postMessage(createTransportMessage(TRANSPORT.START, transportType)),
         );
 
-        _deviceList.init();
+        _deviceList.init(pendingTransportEvent);
         if (_deviceList) {
             await _deviceList.waitForTransportFirstEvent();
         }
--- a/packages/connect/src/core/index.ts
+++ b/packages/connect/src/core/index.ts
@@ -947,12 +947,40 @@ const handleDeviceSelectionChanges = (interruptDevice?: DeviceTyped) => {
     }
 };
 
-/**
- * Start DeviceList with listeners.
- * @param {boolean} transportReconnect
- * @returns {Promise<void>}
- * @memberof Core
- */
+const createDeviceList = (params: ConstructorParameters<typeof DeviceList>[0]) => {
+    const deviceList = new DeviceList(params);
+
+    deviceList.on(DEVICE.CONNECT, device => {
+        handleDeviceSelectionChanges();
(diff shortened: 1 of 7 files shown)


mroz22:
related https://github.com/trezor/trezor-suite/issues/12696

packages/connect/src/core/index.ts
```diff
@@ -963,75 +963,65 @@ const handleDeviceSelectionChanges = (interruptDevice?: DeviceTyped) => {
...
-            }
-        });
+    _deviceList.on(DEVICE.CHANGED, device => {
+        postMessage(createDeviceMessage(DEVICE.CHANGED, device));
+    });
 
-        _deviceList.on(TRANSPORT.START, transportType =>
-            postMessage(createTransportMessage(TRANSPORT.START, transportType)),
-        );
+    _deviceList.on(TRANSPORT.ERROR, error => {
```
mroz22: note: test reconnect to webusb when bridge dies

packages/connect/src/device/DeviceList.ts
```diff
@@ -329,14 +323,28 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> {
...
-        if (autoResolveTransportEventTimeout) {
+    private waitForDevices(deviceCount: number, autoResolveMs: number) {
+        const { promise, resolve } = createDeferred();
+        let transportStartPending = deviceCount;
+
+        autoResolveTransportEventTimeout = setTimeout(resolve, autoResolveMs);
+
+        const onDeviceConnect = () => {
+            transportStartPending--;
             clearTimeout(autoResolveTransportEventTimeout);
```
mroz22: lets fix this. clear autoresolve later

packages/connect/src/device/DeviceList.ts
```diff
@@ -254,73 +254,69 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> {
...
-                });
-            } else {
-                this.emit(TRANSPORT.START, this.getTransportInfo());
-            }
-            this.transport.handleDescriptorsChange(descriptors);
-            this.transport.listen();
-        } catch (error) {
-            // transport should never. lets observe it but we could even remove try catch from here
-            console.error('DeviceList init error', error);
+        // TODO handleDescriptorChange can emit TRANSPORT.UPDATE before TRANSPORT.START is emitted, check whether acceptable
```
mroz22, line 358: note check

mroz22 (approved):

packages/connect/src/device/DeviceList.ts
mroz22: i am not sure about this. lets bring it up again later

PR #13274, opened 2024-07-10
chore(ci): remove dependency on suite-desktop for suite web staging build

vdovhanych:
This removes dependency on suite desktop for suite-web staging build and deploy. It was used for old suite landing page which no longer exists.


komret (approved):

vdovhanych:
@matejkriz Is my assumption correct that we do not use the files from there in in any way?

matejkriz:
> @matejkriz Is my assumption correct that we do not use the files from there in in any way?

I haven't checked, but I think so too.

PR #13280, opened 2024-07-10
fix(suite): do not allow broadcast of the TX with locktime > blockheight

peter-sanderson:
image

### DIFF
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -6184,6 +6184,11 @@ export default defineMessages({
         id: 'BROADCAST_TOOLTIP',
         defaultMessage: 'Broadcast the transaction to the network.',
     },
+    BROADCAST_TOOLTIP_DISABLED_LOCKTIME: {
+        id: 'BROADCAST_TOOLTIP_DISABLED_LOCKTIME',
+        defaultMessage:
+            'A transaction with locktime higher than the current block/timestamp would be rejected by the network.',
+    },
     IMPORT_CSV: {
         id: 'IMPORT_CSV',
         description: 'item in dropdown menu',
--- a/packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
+++ b/packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
@@ -6,9 +6,12 @@ import { spacingsPx } from '@trezor/theme';
 import { Translation } from 'src/components/suite';
 import { useSendFormContext } from 'src/hooks/wallet';
 import { OpenGuideFromTooltip } from 'src/components/guide';
-import { Locktime } from './Locktime';
+import { Locktime } from './Locktime/Locktime';
 import { CoinControl } from './CoinControl/CoinControl';
 import { OnOffSwitcher } from '../OnOffSwitcher';
+import { selectNetworkBlockchainInfo } from '@suite-common/wallet-core';
+import { useSelector } from 'src/hooks/suite';
+import { canLocktimeTxBeBroadcast } from './Locktime/canLocktimeTxBeBroadcast';
 
 const Wrapper = styled.div`
     display: flex;
@@ -63,6 +66,8 @@ export const BitcoinOptions = () => {
         setDraftSaveRequest,
         resetDefaultValue,
         setValue,
+        watch,
+        network,
     } = useSendFormContext();
 
     const options = useWatch({
@@ -84,6 +89,15 @@ export const BitcoinOptions = () => {
         // Without this, this change may be lost which will result in UI glitch (closing the Coin Control UI)
         setDraftSaveRequest(true);
     };
+
+    const blockchain = useSelector(selectNetworkBlockchainInfo(network.symbol));
+
+    const locktime = watch('bitcoinLockTime');
+
+    const isBroadcastDisabled = !canLocktimeTxBeBroadcast({
+        locktime: locktime !== undefined ? Number(locktime) : undefined,
+        currentBlockHeight: blockchain.blockHeight,
+    });
 
     return (
         <Wrapper>
@@ -112,7 +126,17 @@ export const BitcoinOptions = () => {
                             </StyledButton>
                         </Tooltip>
                     )}
-                    <Tooltip content={<Translation id="BROADCAST_TOOLTIP" />} cursor="pointer">
+                    <Tooltip
+                        content={
+                            <Translation
+                                id={
+                                    isBroadcastDisabled
+                                        ? 'BROADCAST_TOOLTIP_DISABLED_LOCKTIME'
+                                        : 'BROADCAST_TOOLTIP'
+                                }
+                            />
+                        }
+                    >
                         <StyledButton
                             variant="tertiary"
                             size="small"
@@ -122,6 +146,7 @@ export const BitcoinOptions = () => {
                                 composeTransaction();
                             }}
                             data-test="broadcast-button"
+                            isDisabled={isBroadcastDisabled}
                         >
                             <Inline>
                                 <Translation id="BROADCAST" />
--- a/packages/suite/src/hooks/wallet/useSendFormFields.ts
+++ b/packages/suite/src/hooks/wallet/useSendFormFields.ts
@@ -111,18 +111,21 @@ export const useSendFormFields = ({
         return getValues(fieldName);
     };
 
-    const toggleOption = (option: FormOptions) => {
-        const enabledOptions = getValues('options') || [];
-        const isEnabled = enabledOptions.includes(option);
-        if (isEnabled) {
-            setValue(
-                'options',
-                enabledOptions.filter(o => o !== option),
-            );
-        } else {
-            setValue('options', [...enabledOptions, option]);
-        }
-    };
+    const toggleOption = useCallback(
+        (option: FormOptions) => {
+            const enabledOptions = getValues('options') || [];
+            const isEnabled = enabledOptions.includes(option);
+            if (isEnabled) {
+                setValue(
+                    'options',
+                    enabledOptions.filter(o => o !== option),
+                );
+            } else {
+                setValue('options', [...enabledOptions, option]);
+            }
+        },
+        [getValues, setValue],
+    );
 
     return {
         calculateFiat,
--- a/packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/Locktime/Locktime.tsx
+++ b/packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/Locktime/Locktime.tsx
@@ -4,8 +4,11 @@ import { NumberInput, Translation } from 'src/components/suite';
 import { useSendFormContext } from 'src/hooks/wallet';
 import { Card, Icon, IconButton } from '@trezor/components';
 import { getInputState, isInteger } from '@suite-common/wallet-utils';
-import { useTranslation } from 'src/hooks/suite';
+import { useSelector, useTranslation } from 'src/hooks/suite';
 import { spacingsPx } from '@trezor/theme';
+import { selectNetworkBlockchainInfo } from '@suite-common/wallet-core';
+import { useEffect } from 'react';
+import { canLocktimeTxBeBroadcast } from './canLocktimeTxBeBroadcast';
 
 const Label = styled.div`
     display: flex;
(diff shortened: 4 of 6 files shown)


packages/suite/src/support/messages.ts
```diff
@@ -6184,6 +6184,11 @@ export default defineMessages({
         id: 'BROADCAST_TOOLTIP',
         defaultMessage: 'Broadcast the transaction to the network.',
     },
+    BROADCAST_TOOLTIP_DISABLED_LOCKTIME: {
+        id: 'BROADCAST_TOOLTIP_DISABLED_LOCKTIME',
+        defaultMessage:
+            'You cannot broadcast transaction with locktime that is more then the next block.',
```
peter-sanderson: @Hermez-cz :pray: copy :pray:

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
```diff
@@ -112,7 +123,18 @@ export const BitcoinOptions = () => {
...
+                        content={
+                            <Translation
+                                id={
+                                    broadcastDisabled
+                                        ? 'BROADCAST_TOOLTIP_DISABLED_LOCKTIME'
+                                        : 'BROADCAST_TOOLTIP'
+                                }
+                            />
+                        }
+                        cursor="pointer"
```
komret: Looks like this is ignored.

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
```diff
@@ -9,6 +9,8 @@ import { OpenGuideFromTooltip } from 'src/components/guide';
 import { Locktime } from './Locktime';
 import { CoinControl } from './CoinControl/CoinControl';
 import { OnOffSwitcher } from '../OnOffSwitcher';
+import { selectNetworkBlockchainInfo } from '@suite-common/wallet-core';
+import { useSelector } from '../../../../../../hooks/suite';
```
komret:
```suggestion
import { useSelector } from 'src/hooks/suite';
```

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
```diff
@@ -85,6 +89,13 @@ export const BitcoinOptions = () => {
         setDraftSaveRequest(true);
     };
 
+    const blockchain = useSelector(selectNetworkBlockchainInfo(network.symbol));
+
+    const locktime = watch('bitcoinLockTime');
+
+    const broadcastDisabled =
```
komret: Please add a comment on why we are disabling this.

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/Locktime.tsx
```diff
@@ -24,22 +26,28 @@ export const Locktime = ({ close }: LocktimeProps) => {
...
-    const inputValue = getDefaultValue(inputName) || '';
+    const defaultInputValue = getDefaultValue(inputName) || '';
     const error = errors[inputName];
 
-    const handleLocktimeChange = () => {
-        if (!error) {
-            if (broadcastEnabled) toggleOption('broadcast');
+    const locktime = watch('bitcoinLockTime');
+
+    useEffect(() => {
```
komret, line 46: Does this have to be an effect? I'd prefer it to stay in the change handler, if possible. Effects make the code less predictable.

komret (changes requested):

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
peter-sanderson: Interesting. It is not ignored by the Tooltip component itself, but because Tooltip contains only `button` which has the `pointer` cursor it is always pointer. I will remove it from here.

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
peter-sanderson: I prefer relative paths as they are IMO better supported by IDEs and tooling and don't require any special configuration. 

Do we have any team consensus on this?

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/Locktime.tsx
peter-sanderson: Well, I hate those effects as well.  My understanding of ReactHookForm is that this is the way how to do it and using `onChange` is the bad way (as I cannot guarantee that the form state was already updated). Lets discuss this. I talked about this with @PeKne and we came to this conclusion.

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
```diff
@@ -85,6 +89,13 @@ export const BitcoinOptions = () => {
...
+        locktime !== undefined && Number(locktime) > blockchain.blockHeight + 1;
```
PeKne: could be simplified as 
```suggestion
      Number(locktime) > blockchain.blockHeight + 1;
```

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
PeKne: nit: boolean variables should ideally start with `is` or `has` prefix for better readability.
```suggestion
    const isBroadcastDisabled =
```

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
komret: There are some unspoken rules regarding import ordering (should be automated though) and relative vs. absolute imports which prefer absolute imports unless the file is in the same folder or subfolder. We can discuss at code improvements. I have no hard preference, but would like to be consistent.

peter-sanderson: okok, i have the other peference, but its not really strong. but it would be nice to have it somehow automated and autofixed. thinking about import formatting is annoying

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/Locktime.tsx
komret: I believe it is safe. If you need to access `locktime`, you can do so from `e.target.value`. Not tested.

peter-sanderson: Yes if you need just value from the field itself. But here you also touch `broadcastEnabled` both read & write (in this case it may work, but still, I think we should not do this in general as it can break things.

Also calling `composeTransaction` from onChange is IMO recipe for a disaster as this function has many™ consequences.

komret: I actually think having `composeTransaction` gives you more control over when it is called, which is preferable. `broadcastEnabled ` can also be safely access via `getValues`. Not sure about `getDefaultValue`. But I will try.

szymonlesisz:
im having doubts if it's really currentBlock+1.

bitcoin core code say it so:

https://github.com/bitcoin/bitcoin/blob/9adebe145557ef410964dd48a02f3d239f488cd0/src/validation.cpp#L151
```
const int nBlockHeight = active_chain_tip.nHeight + 1;
```

https://github.com/bitcoin/bitcoin/blob/9adebe145557ef410964dd48a02f3d239f488cd0/src/validation.cpp#L804-L809
```
    // Only accept nLockTime-using transactions that can be mined in the next
    // block; we don't want our mempool filled up with transactions that can't
    // be mined yet.
    if (!CheckFinalTxAtTip(*Assert(m_active_chainstate.m_chain.Tip()), tx)) {
        return state.Invalid(TxValidationResult::TX_PREMATURE_SPEND, "non-final");
    }
```


but when i try on regtest: currentBlock=151, customLocktime=152  im getting `error: -26: non-final`

probably core active_chain_tip.nHeight !== ours currentBlock ?

szymonlesisz:
re-validation is not working correctly:
- put value `100` > broadcast button is ON and active
- put value `1000` > broadcast button is OFF and disabled
- backspace to `100` > broadcast button is OFF and active (should be ON and active)

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
peter-sanderson: Do we have a consensus on this? I see both approaches in the code. I don't mind using either of them.

peter-sanderson:
> backspace to 100 > broadcast button is OFF and active (should be ON and active)

@szymonlesisz I am not sure about this. I cannot distinguish is user turned broadcast OFF before changing time-lock or not. It would require some extra state to keep the information that it was turned off by validation so we can re-enable it back when locktime is low again.

It seems to me as a big edge-case (given that time-lock feature itself is an edge-case already).

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
peter-sanderson: I am not really fun of implicit conversions. `Number(undefined` => `0`. Even it may have sense here, I prefer to be consistent and not doing them at all as they may be source of bugs and unintentional behavior.

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
PeKne: very vaguely described here: https://www.notion.so/satoshilabs/Naming-6f8117ecfcab4a198faf9743467abfd8?pvs=4#8fff06b677164732b10bff0011b83024

peter-sanderson:
> im having doubts if it's really currentBlock+1.

@szymonlesisz I have tested the same thing on the Testnet and it seems you are right. Current height was `2867417` and I tried `2867418` and I got the `non-final` error as well.

peter-sanderson:
I have therefore changed the condition from `<= height +1` to just `<= height`

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/Loctime/canBeLocktimeTxBroadcasted.ts
```diff
@@ -0,0 +1,27 @@
...
+type CanBeLocktimeTxBroadcastedParams = {
+    locktime: number | undefined;
+    currentBlockHeight: number;
+};
+
+/**
+ * Transaction with TimeLock > currentBlockHeight is considered Non-Standard.
+ * Such transaction would be rejected by the full-node's Mempool with error: "non-final".
+ */
+export const canBeLocktimeTxBroadcasted = ({
```
komret:
```suggestion
export const canLocktimeTxBeBroadcast = ({
```

peter-sanderson: fixed

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/BitcoinOptions.tsx
peter-sanderson: fixed

komret (approved):
Optional: talk to @jvaclavik about possible UI improvements for the Locktime section.

packages/suite/src/views/wallet/send/components/Options/BitcoinOptions/Locktime.tsx
peter-sanderson: solved in person, we leave it as it is

komret:
Also please change the commit message, it is no longer +1.

Issue #13261, opened 2024-07-10
Polish Welcome screen


jvaclavik:
1. Check main button color - should be `variant="primary"`

2. There are multiple responsive issues:

![image](https://github.com/trezor/trezor-suite/assets/858321/bafbe3b4-1a54-489a-ab5b-796b4feb534b)

![image](https://github.com/trezor/trezor-suite/assets/858321/b99c4d9e-6ea8-430a-a5c2-9519791fdf6f)

3. Verify design with Ondra

hstastna:
I think there are more places in Suite with the same overflow problem (especially with big zoom and small window size):
![wdyt](https://github.com/trezor/trezor-suite/assets/13417815/6e5a5810-39cc-41e0-9777-c9ab8b5b72fc)

matejkriz:
One more important for us now (as we will navigate users from Suite Lite to mobile Suite Web for onboarding) 🙏 :
<img width="342" alt="image" src="https://github.com/trezor/trezor-suite/assets/3729633/dc4199d0-cf40-4990-8331-376a35994d0b">

hstastna:
![image](https://github.com/user-attachments/assets/82af61f8-54b3-4545-bdcd-f1cfbebfe033)

jvaclavik:
One more screen. Can you please take a look @hstastna? 🙏 

![image](https://github.com/user-attachments/assets/9c94be04-ad70-4a6e-a596-52a4a5b3b740)

hstastna:
Need to adjust view only promo content related screen, hence reopening the issue.

bosomt:
**QA OK**


**Info:**
 - Suite version: web 24.8.0 (96da7b911c634b1506220e297f1e755ccd42394a)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.7.2 bitcoin-only (revision da75d8f4b67410b40a9cfd2954d183d81dd6e8e8)
 - Transport: BridgeTransport 2.0.33

Issue #12678, opened 2024-05-31
Refactor useDetectDeviceError


vytick:
Current state: 

In [useDetectDeviceError](https://github.com/trezor/trezor-suite/blob/develop/suite-native/device/src/hooks/useDetectDeviceError.tsx) there are already numerous useEffects showing alerts, that should be mutually exclusive and we have to  make sure the others dont trigger at the same device state. This is already hard to read and as we will be probably adding more status detection, it will be even harder with bigger complexity.

Expected state:

- Refactored code to use single decision tree to show only one alert at the time (e.g. switch if possible) with clear priority of alerts.
- Remove `selectIsConnectedDeviceUninitialized` and use `selectIsDeviceInitialized` instead.
- make `selectIsDeviceInitialized` only check for initialization and do not check seedles status or name the selector correctly

PR #13282, opened 2024-07-10
feat(suite-native): show better message if user is navigated back from web Suite onboarding after FW installation

### LINKED ISSUE #11896: Handle connection to Trezor Suite web and connected Trezor

matejkriz:
## Description

During onboarding via web Suite on phone, right after firmware installation, Trezor appears as a completely new device so new system dialog appears navigating user to Suite Lite app (it can be dismissed, but CTA is to open Suite Lite).

But onboarding is not finished at that point. We can detect that state (firmware is present) and provide better message.

(CTA button lead to new tab in browser (at least in Chrome), but I see no way how to lead to existing tab to continue there and user can open the app in that state even without having Suite Web already opened...)

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/11896 (last part: _4. New requirement: After FW installation via Web Suite_)

## Screenshots:
image

___

video: 
Notes for video:
- UI issues with responsivity reported in https://github.com/trezor/trezor-suite/issues/13261
- After the whole onboarding, Trezor had troubles to connect to Suite Lite right away - It is not consistent, it depends probably on state in Suite Web (i.e. running discovery), but reconnecting the device resolves it, so I don't see it as a blocker.

https://github.com/trezor/trezor-suite/assets/3729633/06e153ee-8f3e-4be4-9541-8ef5447e04e0

### DIFF
--- a/suite-native/analytics/src/events.ts
+++ b/suite-native/analytics/src/events.ts
@@ -170,7 +170,11 @@ export type SuiteNativeAnalyticsEvent =
     | {
           type: EventType.UnsupportedDevice;
           payload: {
-              deviceState: 'unsupportedFirmware' | 'noSeed' | 'bootloaderMode';
+              deviceState:
+                  | 'unsupportedFirmware'
+                  | 'noSeed'
+                  | 'bootloaderMode'
+                  | 'noSeedWithFirmware';
           };
       }
     | {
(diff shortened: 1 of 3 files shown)


matejkriz:
bumping priority of https://github.com/trezor/trezor-suite/issues/12678 🙉

Nodonisko (approved):
Did not tested, but seems OK to me.

Issue #12352, opened 2024-05-13
Balances mismatch


mnuky:
We have several reports from internal and support about following issues in the app:
1. Balance mismatch -The top value on the homepage is different than the value shown in the transactions list when I have only BTC for example
2. Amount doubled - Some users see exactly double the amount at the top value in homepage
3. Staking is not shown as part of the top value on the homepage

Proposal: 
1. **Partially decoupling the top value from the graph**. Meaning that when the user is not interacting with graph, we will show the total amount of assets, however, if the user interacts with the graph, this will display the value from the graph
3. **Include staking in top value and device switcher**, so it is only missing in the graph
4. Solana is included (already) in the top value, but not in the graph

- **After Discussion We are going for Solution Number 2**
    - Let’s try this partial decoupling
    - Include staking in top value and device switcher, so it is only missing in the graph
    - Solana is included (already) in the top value, but not in the graph
    - ~~Hide % if it is zero is in the graph view~~ % will not calculate agains first point of graph but agains first non-zero point of graph so now percentages kind of make sense even for "all" timeframe etc.
    - Show the current balance independently on the graph so even if there is an error in the graph, but we know the current value.

matejkriz:
I was able to reproduce balance mismatch on Cardano and Ethereum. I guess it's due to fiat rates inconsistencies.

matejkriz:
~~I'm able to see missing transaction too on `Bitcoin #1 (Legacy SegWit)` account, `allall`, passphrase `test`~~
I think I was rather mistaken and thought there were two send transactions, but there is only one.

matejkriz:
We don't know how to reproduce it consistently.

fang314:
related to #16141

STew790:
**QA OK**
As mentioned [here](https://github.com/trezor/trezor-suite/pull/16457#pullrequestreview-2564362461) as well, very nice improvement. I haven't seen any balance mismatch. 
Also, already released.

**Info**
25.1.3 dde03ca

PR #13287, opened 2024-07-10
feat(mobile): support ETH tokens in graphs

Nodonisko:
This PR shouldn't change anything for users right now. In follow-up PR we should enable ERC tokens across the app.

## Related Issue

Resolve


suite-common/graph/src/eth.go
matejkriz, line 1: What is this 😮?

suite-common/graph/src/graphDataFetching.ts
```diff
@@ -102,12 +135,20 @@ const getAccountBalanceHistory = async ({
     }
 
     const getBalanceHistory = async () => {
-        if (getNetworkType(coin) === 'ripple') {
-            return TransactionCacheEngine.getAccountBalanceHistory({
+        const networkType = getNetworkType(coin);
+        if (networkType === 'ripple' || networkType === 'ethereum') {
```
matejkriz, line 136: why not bitcoin?

suite-common/graph/src/eth.go
Nodonisko: That's blockbook implemenation I used it as reference for our implementation :D I will delete it...

suite-common/graph/src/graphDataFetching.ts
Nodonisko: I would not do it yet, as we discussed

suite-common/suite-types/src/messageSystem.ts
```diff
@@ -8,7 +8,7 @@
  */
 export type DateTime = string;
 export type Version = string | string[];
-export type Model = '1' | 'T' | 'T1B1' | 'T2T1' | 'T2B1' | 'Safe 3' | '';
+export type Model = '1' | 'T' | 'T1B1' | 'T2T1' | 'T2B1' | 'Safe 3' | 'T3T1' | '';
```
Nodonisko: Where is this from?

matejkriz: generated from `config.schema.v1.json`, resolved on develop....

suite-common/graph/src/graphDataFetching.ts
```diff
@@ -57,57 +65,85 @@ const getLatestAccountBalance = async ({
...
+const accountBalanceHistoryCache: Record<string, AccountBalanceHistoryWithTokens> = {};
 
 const getAccountBalanceHistory = async ({
-    coin,
-    identity,
-    descriptor,
+    accountItem,
     endOfTimeFrameDate,
     forceRefetch,
+    dispatch,
```
matejkriz, line 119: Could you please explain in code comment why this is not a thunk?

suite-common/graph/src/balanceHistoryUtils.ts
matejkriz, line 1: Could you please try to add references in code comments? Links to Blockbook or any other sources...

suite-common/graph/src/graphDataFetching.ts
```diff
@@ -163,24 +235,26 @@ const fiatRatesCache: Record<string, FiatRatesItem[]> = {};
...
 }: {
     timestamps: number[];
     coin: NetworkSymbol;
+    contractId?: TokenAddress;
     fiatCurrency: FiatCurrencyCode;
     forceRefetch?: boolean;
     isElectrumBackend: boolean;
 }) => {
-    const cacheKey = `${coin}-${fiatCurrency}-${JSON.stringify(timestamps)}`;
+    const cacheKey = `${coin}-${contractId}-${fiatCurrency}-${JSON.stringify(timestamps)}`;
```
matejkriz, line 252: contractId is possibly undefined, is it ok here?

suite-common/graph/src/types.ts
```diff
@@ -35,3 +49,34 @@ export type GroupedBalanceMovementEvent = {
...
+
+export type AccountHistoryMovementItem = {
+    time: number;
+    txs: number;
+    received: BigNumber;
+    sent: BigNumber;
+    sentToSelf: BigNumber;
+};
+
+export type AccountHistoryMovement = {
```
PeKne, line 71: nit:
```suggestion
export type AccountMovementHistory = {
```
Isn't it better naming? Since the object describes `history` rather than `historical movement`.

suite-common/graph/src/graphUtils.ts
```diff
@@ -34,7 +37,7 @@ export const getTimestampsInTimeFrame = (
     });
 
     // sometimes endOfRangeDate could be too close to recent time and that cause problems with blockbook
```
PeKne:
```suggestion
```
Please remove also the comment if it does not apply anymore.

suite-common/graph/src/graphBalanceEvents.ts
```diff
@@ -107,19 +112,27 @@ export const getAccountMovementEvents = async ({
...
+                    accountKey: account.accountKey,
+                }),
+            ).unwrap();
+
+            return getAccountHistoryMovementFromTransactions({
+                transactions: allTransactions,
                 coin,
-                descriptor,
-            });
+            }).main;
```
PeKne, line 137: Please add comment why only `main` is needed and tokens are not.

suite-common/graph/src/graphDataFetching.ts
Nodonisko: No issue at all, it will make unique string either way which is what we want here.

suite-common/graph/src/types.ts
Nodonisko: It describes historical movements in specific block so I think it's fine, but I don't have any strong opinion about it.

PeKne: If it describes movements only in a single block and not across the whole history, than it is fine I guess 👍.

suite-common/graph/src/balanceHistoryUtils.ts
Nodonisko: done

suite-common/graph/src/graphUtils.ts
Nodonisko: done

suite-common/graph/src/graphBalanceEvents.ts
Nodonisko: done, it's because we don't have UI for multiple movements in one graph, will be fixed in next PR

suite-common/graph/src/graphDataFetching.ts
matejkriz: 👌

suite-common/graph/src/graphUtils.ts
```diff
@@ -33,8 +36,7 @@ export const getTimestampsInTimeFrame = (
         numberOfPoints,
     });
 
-    // sometimes endOfRangeDate could be too close to recent time and that cause problems with blockbook
-    const intervalEndDate = fromUnixTime(getBlockbookSafeTime(getUnixTime(endOfTimeFrameDate)));
+    const intervalEndDate = fromUnixTime(getUnixTime(endOfTimeFrameDate));
```
matejkriz, line 39: How have you found out that `getBlockbookSafeTime` is no longer necessary?

Nodonisko: I think from @martinboehm quite long time ago

matejkriz (approved):
Not changing anything for the user right now but still it is a huge change and great step forward! 🎉

Let's see if it helps to resolve at least some issues from https://github.com/trezor/trezor-suite/issues/12352, it doesn't solve the balance mismatch in my case, but maybe missing transactions? We'll see.

PR #13292, opened 2024-07-11
Core Mode followup

martykan:
## Description

Follow-up to #13031

- [x] Update documentation
- [x] Deprecate useCoreInPopup param
- [x] Use auto as default

### DIFF
--- a/packages/connect-explorer/src/pages/methods/other/init.mdx
+++ b/packages/connect-explorer/src/pages/methods/other/init.mdx
@@ -15,7 +15,16 @@ export const InitSchema = Type.Object({
     lazyLoad: Type.Optional(Type.Boolean()),
     popup: Type.Optional(Type.Boolean()),
     transports: Type.Optional(
-        Type.Array(Type.Literal('BridgeTransport', 'WebUsbTransport', 'NodeUsbTransport')),
+        Type.Array(
+            Type.Union([
+                Type.Literal('BridgeTransport'),
+                Type.Literal('WebUsbTransport'),
+                Type.Literal('NodeUsbTransport'),
+            ]),
+        ),
+    ),
+    coreMode: Type.Optional(
+        Type.Union([Type.Literal('auto'), Type.Literal('iframe'), Type.Literal('popup')]),
     ),
 });
 
@@ -32,15 +41,35 @@ export const paramDescriptions = {
         'default `false`. Postpone iframe creation until TrezorConnect is called for the first time.',
     popup: 'default `true`. Projects running on trusted domains (trezor.io) are not required to use popup. For other domains this option is ignored.',
     transports:
-        'Array of transports that should be use. If not provided, TrezorConnect will chose a reasonable default based on your environment.',
+        'Array of transports that should be use. If not provided, TrezorConnect will choose a reasonable default based on your environment.',
+    coreMode:
+        `default \`auto\`.
+
+If set to \`popup\`, webUSB transport would work cross-origin on the web. The disadvantage of this mode is that Connect popup does not retain a stateful connection to the device, which means for example that device events are not sent to the application when the popup is not open.
+
+If set to \`iframe\`, TrezorConnect will use iframe with persistent connection for the communication between popup and 3rd party wallet.
+
+In the default \`auto\` mode, TrezorConnect will automatically choose \`iframe\` if bridge transport is available and \`popup\` otherwise, to allow for webUSB use.`
 };
 
 ## Init
 
-Initiate TrezorConnect
+To initiate TrezorConnect, please provide the manifest object with your email and appURL. This is crucial for us to identify your use of TrezorConnect and to contact you if necessary.
+If the manifest object is not set, TrezorConnect methods will not be available.
+
+It is also possible to customize more advanced options for TrezorConnect.
 
 ```javascript
-const result = await TrezorConnect.init(params);
+const result = await TrezorConnect.init({
+    manifest: {
+        appUrl: 'https://your.application.com',
+        email: 'your@email.com',
+    },
+    // More advanced options
+    debug: true,
+    lazyLoad: true,
+    coreMode: 'auto',
+});
 ```
 
 ### Params
--- a/packages/connect-web/src/index.ts
+++ b/packages/connect-web/src/index.ts
@@ -2,9 +2,10 @@ import { ConnectFactoryDependencies, factory } from '@trezor/connect/src/factory
 import { CoreInIframe } from './impl/core-in-iframe';
 import { CoreInPopup } from './impl/core-in-popup';
 import ProxyEventEmitter from './utils/proxy-event-emitter';
-import type { ConnectSettings, Manifest } from '@trezor/connect/src/types';
+import type { ConnectSettings, ConnectSettingsPublic, Manifest } from '@trezor/connect/src/types';
 import EventEmitter from 'events';
 import { CallMethodPayload } from '@trezor/connect/src/events';
+import { getEnv } from './connectSettings';
 
 type TrezorConnectType = 'core-in-popup' | 'iframe';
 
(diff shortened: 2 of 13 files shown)


packages/connect-explorer/src/pages/methods/other/init.mdx
```diff
@@ -33,14 +42,27 @@ export const paramDescriptions = {
     popup: 'default `true`. Projects running on trusted domains (trezor.io) are not required to use popup. For other domains this option is ignored.',
     transports:
         'Array of transports that should be use. If not provided, TrezorConnect will chose a reasonable default based on your environment.',
+    coreMode:
+        'default `iframe`. If set to `popup`, TrezorConnect will use popup for interaction with user. The advantage is that popup enables WebUSB to work cross-origin on the web. The disadvantage is that popup does not retain a stateful connection to the device, which can be a problem for some use-cases. If set to `auto` TrezorConnect will use iframe only if bridge is available.',
```
mroz22: maybe @Hannsek check on this. "will use popup for interaction with user" feels somewhat misleading to me, it will use popup regardless of the mode. The rests sounds good to me.

packages/connect-explorer/src/pages/methods/other/init.mdx
```diff
@@ -33,14 +42,27 @@ export const paramDescriptions = {
...
 };
 
 ## Init
 
-Initiate TrezorConnect
+Initiate TrezorConnect. It is important to set the manifest object with your email and appUrl, otherwise TrezorConnect methods won't be available.
+
+It is also possible to customize more advanced options for TrezorConnect.
```
mroz22: I would mention why we need this. Could @Hannsek step in please?

packages/connect-web/src/index.ts
```diff
@@ -55,9 +56,14 @@ export class TrezorConnectDynamicImpl implements ConnectFactoryDependencies {
     }
 
     public async init(settings: Partial<ConnectSettings> = {}) {
-        if (settings.useCoreInPopup || settings.coreMode === 'popup') {
+        const env = settings.env || getEnv();
+        if (settings.coreMode === 'iframe' || settings.popup === false || env === 'webextension') {
+            this.currentTarget = 'iframe';
+        } else if (settings.useCoreInPopup || settings.coreMode === 'popup') {
```
mroz22: `useCoreInPopup` is now deprecated but used in production right? I am just thinking - could we add a console.warn if this variable is set?

packages/connect-explorer/src/pages/methods/other/init.mdx
```diff
@@ -33,14 +42,27 @@ export const paramDescriptions = {
     popup: 'default `true`. Projects running on trusted domains (trezor.io) are not required to use popup. For other domains this option is ignored.',
```
Hannsek, line 42: Do we have to mention if it is not available for any external developers?

packages/connect-explorer/src/pages/methods/other/init.mdx
```diff
@@ -33,14 +42,27 @@ export const paramDescriptions = {
...
     transports:
         'Array of transports that should be use. If not provided, TrezorConnect will chose a reasonable default based on your environment.',
```
Hannsek:
```suggestion
        'Array of transports that should be use. If not provided, TrezorConnect will choose a reasonable default based on your environment.',
```

packages/connect-explorer/src/pages/methods/other/init.mdx
martykan: It's better to have it documented regardless imo

packages/connect-web/src/index.ts
martykan: Ok actually it shouldn't be deprecated, it should remain but as internal. I just did a refactor to separate the public/internal settings and I removed it from the init method. It was never publicly documented, so it should be safe

packages/connect-explorer/src/pages/methods/other/init.mdx
martykan: I just updated the text slightly

packages/connect-explorer/src/pages/methods/other/init.mdx
```diff
@@ -32,15 +41,28 @@ export const paramDescriptions = {
         'default `false`. Postpone iframe creation until TrezorConnect is called for the first time.',
     popup: 'default `true`. Projects running on trusted domains (trezor.io) are not required to use popup. For other domains this option is ignored.',
     transports:
-        'Array of transports that should be use. If not provided, TrezorConnect will chose a reasonable default based on your environment.',
+        'Array of transports that should be use. If not provided, TrezorConnect will choose a reasonable default based on your environment.',
+    coreMode:
+        'default `auto`. In `iframe` mode, the Connect core is loaded inside an iframe injected into the page. If set to `popup`, TrezorConnect will avoid using the iframe, instead loading the core in popup. The advantage is that popup enables WebUSB to work cross-origin on the web. The disadvantage is that popup does not retain a stateful connection to the device, which can be a problem for some use-cases. If set to `auto` TrezorConnect will automatically use `iframe` if bridge is available, `popup` otherwise.',
```
martykan: @mroz22 @Hannsek new text

packages/connect-explorer/src/pages/methods/other/init.mdx
Hannsek: I think that the naming "popup" itself is misleading a bit. If we won't change the naming I'd do something like:

If set to `popup`, webUSB transport would work cross-origin on the web. The disadvantage of this mode is that Connect popup does not retain a stateful connection to the device. 

If set to `auto`, TrezorConnect will use iframe with persistent connection for the communication between popup and 3rd party wallet only if nodeBridge is available.

packages/connect-explorer/src/pages/methods/other/init.mdx
Hannsek: something like this:

```suggestion
To initiate TrezorConnect, please provide the manifest object with your email and appURL. This is crucial for us to identify your use of TrezorConnect and to contact you if necessary.
If the manifest object is not set, TrezorConnect methods will not be available.

It is also possible to customize more advanced options for TrezorConnect.
```

packages/connect-explorer/src/pages/methods/other/init.mdx
Hannsek: > which can be a problem for some use-cases

Which problems exactly? 🤔

martykan: For example when you want to receive events about the device

Hannsek: I tried suggested some text here:
https://github.com/trezor/trezor-suite/pull/13292#discussion_r1677783635

Hannsek: I'd specify these things that won't be available when using `popup` mode.

martykan: Ok I will update it based on your suggestion

Hannsek: But please add there some warning that _some_ things won't be available so it is crystal clear. :)

Hannsek:
I'd also put somewhere a scheme how does it look in the individual modes. What do you think?

packages/connect/src/types/settings.ts
```diff
@@ -22,7 +22,13 @@ export interface ConnectSettings {
...
     lazyLoad?: boolean;
     interactionTimeout?: number;
-    // internal part, not to be accepted from .init()
+    coreMode?: 'auto' | 'popup' | 'iframe';
+    /* _extendWebextensionLifetime features makes the service worker in @trezor/connect-webextension stay alive longer */
+    _extendWebextensionLifetime?: boolean;
+}
+
+// internal part, not to be accepted from .init()
+export interface ConnectSettingsInternal {
```
mroz22, line 32: nice idea!

we do supply trustedHost from connect-explorer webextension setting page, on the other it is intended as internal. just a note.

mroz22 (approved):
looks okey

packages/connect/src/types/settings.ts
martykan: Ok so I'll move trustedHost to public to be safe

PR #13302, opened 2024-07-11
feat(suite): Add analytics to view only promo

### LINKED ISSUE #13300: Analytics: View only & passphrase
Please add analytics events to track:

- View only
    - Promotional frame shown
        - [Yes / Not) now selected
    - View only enabled in wallet switcher settings
    - View only disabled in wallet switcher settings
- Passphrase
    - Settings: Passphrase selected as primary wallet to load

Please add all new events to data team's [Overview](https://www.notion.so/satoshilabs/device-setup-completed-d565806f8de24d8ca8b5db8150721d0c?pvs=4) and [Changelog](https://www.notion.so/satoshilabs/Changelog-Suite-1551ab666b1943f080ff56ffc6896d12?pvs=4).

jvaclavik:
## Related Issue

Resolve #13300

### DIFF
--- a/packages/suite-analytics/src/types/events.ts
+++ b/packages/suite-analytics/src/types/events.ts
@@ -444,4 +444,10 @@ export type SuiteAnalyticsEvent =
           payload: {
               action: 'start' | 'done' | 'learn-more' | 'close-modal';
           };
+      }
+    | {
+          type: EventType.ViewOnlyPromo;
+          payload: {
+              wasAccepted: boolean;
+          };
       };
--- a/packages/suite-analytics/src/constants.ts
+++ b/packages/suite-analytics/src/constants.ts
@@ -57,6 +57,7 @@ export enum EventType {
     SwitchDeviceForget = 'switch-device/forget',
     SwitchDeviceRemember = 'switch-device/remember',
     SwitchDeviceEject = 'switch-device/eject',
+    ViewOnlyPromo = 'view-only-promo',
 
     SettingsDeviceCheckSeed = 'settings/device/check-seed',
     SettingsDeviceChangePinProtection = 'settings/device/change-pin-protection',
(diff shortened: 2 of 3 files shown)


packages/suite-analytics/src/types/events.ts
```diff
@@ -444,4 +444,10 @@ export type SuiteAnalyticsEvent =
           payload: {
               action: 'start' | 'done' | 'learn-more' | 'close-modal';
           };
+      }
+    | {
+          type: EventType.ViewOnlyPromoButton;
+          payload: {
+              action: 'view-only-enabled' | 'view-only-not-set';
```
peter-sanderson: Not sure about this, maybe it doesnt matter but I would just put boolean here and named id `viewOnlyEnabled: true | false`

peter-sanderson: Is there some specific request for this format?

jvaclavik: But it's not correct what's happening. You are setting the state to `true` or doing nothing. So it could be something like:
`viewOnlyEnabled: true | undefined`, but it's not easy to read

jvaclavik: Nothing except hiding the modal and redirection is happening when you click `No`:

image

peter-sanderson: But from the semantic of the promo banner you either 

1. accept what is offered to you
2. or you reject what is offered to you


there is no 3rd state => therefore I would suggest boolean

OFC naming may be hard

peter-sanderson (approved):

bosomt:
**QA OK**

image

**Info:**
 - Suite version: web 24.8.0 (01a00268f81a8b0ac72d0dcf357d7bbcf8b79cce)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:127.0) Gecko/20100101 Firefox/127.0
 - OS: MacIntel
 - Screen: 1512x982

Issue #13262, opened 2024-07-10
"Reset app" button in web Suite has inconsistent results


ondracja:
**Describe the bug**
When the user resets their app, the app sometimes ends back on dashboard and not on the analytics consent page. App restart happens somewhat though - if view-only mode was enabled, it is disabled after restarting. App is in a broken state - until it's refreshed, some parts don't work at all (eg activating coins). 

Problem sometimes happen right away (as seen on the video) or after second app restart (as described in steps to repro)

**Steps to reproduce:**
1. start web suite
2. click through initial run
3. navigate to `settings/application`
4. click on `reset app`
5. click through initial run again
6. navigate to `settings/application`
7. click on `reset app`
8. observe that you're back on Suite Dashboard

**Info:**
 - Suite version: web 24.8.0 (3bfe04e22b342e900b19b32adf675d1a3834a315)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 2560x1440
 - Device: Trezor T2T1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)
 - Transport: WebUsbTransport

**Expected result:**
No matter how many times I click `Reset app`, I always end up on the initial anayltics consent page.

**Actual result:**
I sometimes end up on dashboard page instead

**Video:**
Video from CI run that found this problem: 

https://github.com/trezor/trezor-suite/assets/9817938/d740c8cc-bccb-4d6b-8fee-85f25c8d9635

jvaclavik:
I wasn't able to replicate it

PR #13312, opened 2024-07-12
fix(suite): reset app bug

ondracja:
- to encourage TDD approach, I've created a PR with enabled test that was broken by issue #13262


packages/suite/src/views/view-only/ViewOnlyTooltip.tsx
```diff
@@ -77,7 +77,12 @@ export const ViewOnlyTooltip = ({ children }: ViewOnlyTooltipProps) => {
                             <Translation id="TR_VIEW_ONLY_TOOLTIP_DESCRIPTION" />
                         </Text>
                     </Content>
-                    <Button variant="tertiary" size="small" onClick={handleClose}>
+                    <Button
+                        variant="tertiary"
+                        size="small"
+                        onClick={handleClose}
+                        data-test="@viewOnlyTooltip/gotIt"
```
tomasklim, line 84: why mix of `kebab-case` and `camelCase`

tomasklim (changes requested):

HajekOndrej:
We have agreed with @mroz22 to close this, because after so much time it would be very hard to finish this due to conflicts and unclear purpose of the PR.

Issue #13313, opened 2024-07-12
Locktime doesn't change the send button correctly


ondracja:
**Describe the bug**
When the user tries to send a regtest transaction and sets a locktime, the submit button stays as `Send` until a higher threshold is reached. In  my testing, the threshold for change was `156`.

**Steps to reproduce:**
1. enable dev mode in settings to enable regtest
2. fill in the regtest send form
3. set a locktime with value `100` 
4. observe that the submit button stayed as `Review&Send`

**Info:**
 - Suite version: web 24.8.0 (3808da45bc3bda5b173a97f30d25cae631f1f127)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 2560x1440
 - Device: Trezor T2T1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)
 - Transport: BridgeTransport 2.0.33

**Expected result:**
After setting up locktime, button should change to `Sign`

**Actual result:**
After setting up locktime, button stays as `Review&Send` until a higher threshold is reached.

**Screenshots:**
<img width="1782" alt="image" src="https://github.com/user-attachments/assets/35a6c36e-533c-4747-8a4b-e3ba4843ff9b">
<img width="1473" alt="image" src="https://github.com/user-attachments/assets/1b13346e-43fd-42cf-99eb-85557bad5a57">

bosomt:
**QA OK**

Info:

    Suite version: web 24.8.0 (https://github.com/trezor/trezor-suite/commit/fe65e0963b951f8b8bf820916ca877d953964058)
    Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:128.0) Gecko/20100101 Firefox/128.0
    OS: MacIntel
    Screen: 1512x982
    Device: Trezor T3T1 2.8.0 regular (revision dd4671a5104952ef505d28d1f9e94d1484b4607a)
    Transport: BridgeTransport 2.0.33

PR #13314, opened 2024-07-12
fix(suite): locktime threshold bug

ondracja:
to encourage TDD approach, I've created a PR with enabled test that was broken by issue #13313


ondracja:
obsolete

PR #13321, opened 2024-07-12
feat(suite-native): incorrect passphrase copy

### LINKED ISSUE #13191: Error state copy for feature requests passphrase fail
If user wants to access data from device during view-only + passphrase, he's prompted to add passphrase. If passphrase is incorrect, we currently add just simple error. This task is to update copy to match design.

https://www.figma.com/design/mmDnYtctfjnreWhTjLfeQH/Passphrase?node-id=2895-5652&t=WScMaNsEhAqbs34a-0

juriczech:
## Description
Add correct copy to receive & add account when incorrect passphrase was used. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13191

## Screenshots:
image

### DIFF
--- a/suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
+++ b/suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
@@ -317,7 +317,16 @@ export const useAddCoinAccount = () => {
                 screen = AppTabsRoutes.ReceiveStack;
             }
 
-            showGeneralErrorAlert();
+            if (newAccountResult.payload === 'Passphrase is incorrect') {
+                showAlert({
+                    title: translate('modulePassphrase.featureAuthorizationError'),
+                    pictogramVariant: 'red',
+                    primaryButtonTitle: translate('generic.buttons.close'),
+                    primaryButtonVariant: 'redBold',
+                });
+            } else {
+                showGeneralErrorAlert();
+            }
             navigation.dispatch(
                 CommonActions.reset({
                     index: 0,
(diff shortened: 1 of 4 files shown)


suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
```diff
@@ -317,7 +317,16 @@ export const useAddCoinAccount = () => {
                 screen = AppTabsRoutes.ReceiveStack;
             }
 
-            showGeneralErrorAlert();
+            if (newAccountResult.payload === 'Passphrase is incorrect') {
```
PeKne, line 320: I would prefer to match this based on error type rather then string value. But I understand that would mean to refactor all the add new coin error types. So feel free to resolve this.

PeKne (approved):

suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
juriczech: Not add account types, but connect error types. This is from connect, which sends something like:
```
{ code: "Device_InvalidState", error: "Passphrase is incorrect" }
```

I agree this is not ideal, but right now it's our only option without having to change connect code.

matejkriz: It would be definitely better to use error code `Device_InvalidState`, not the message...

juriczech: I assume that the invalid state can be tied to multiple other cases that would trigger this alert even if it wasn't passphrase. I plan to check this with someone next week to improve the condition

PR #13323, opened 2024-07-14
fix(wallet-utils): allow tokens without symbol and valid balance

### LINKED ISSUE #13322: Suite does not display tokens without symbol
From [slack conversation](https://satoshilabs.slack.com/archives/C03SRP8PSS2/p1720429988564829)

image

Test token:
`0xa1Ac2d8a57b70ff9637521F76D5733260c6F0f5a`

tomasklim:
## Description

- do not filter out any tokens
- use formatter which is used also for solana and cardano in case there is missing symbol
- ignore nonsense `toLowerCase`, there is another issue for that. Symbols can be mixed case

## Related Issue

Resolve  #13322

## Screenshots:
Screenshot 2024-07-14 at 20 30 54

### DIFF
--- a/suite-common/wallet-utils/src/accountUtils.ts
+++ b/suite-common/wallet-utils/src/accountUtils.ts
@@ -33,6 +33,7 @@ import {
     HELP_CENTER_COINJOIN_URL,
     HELP_CENTER_TAPROOT_URL,
 } from '@trezor/urls';
+import { formatTokenSymbol } from '@trezor/blockchain-link-utils';
 
 import { toFiatCurrency } from './fiatConverterUtils';
 import { getFiatRateKey } from './fiatRatesUtils';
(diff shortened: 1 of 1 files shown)


suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -451,14 +452,16 @@ export const countUniqueCoins = (accounts: Account[]) => {
...
+        const symbol = formatTokenSymbol(t.symbol || t.contract);
+
+        return {
             ...t,
-            name: t.name || t.symbol,
-            symbol: t.symbol!.toLowerCase(),
-            balance: formatAmount(t.balance!, t.decimals),
-        }));
+            name: t.name || symbol,
+            symbol: symbol.toLowerCase(),
```
jvaclavik, line 461: nitpick:

You have this in the description:
```
ignore nonsense toLowerCase, there is another issue for that. Symbols can be mixed case
```

So why do you use `toLowerCase` here or am I missing something?

jvaclavik (approved):
One comment but approving 👍

suite-common/wallet-utils/src/accountUtils.ts
tomasklim: I expressed myself poorly. 

The lowercase does not make sense but I am keeping it there just to be sure that it does not break anything, but it will be removed soon

PR #13330, opened 2024-07-15
fix(crypto-icons): fix sort function

FreeWall:
Issue: in second run download process of icons doesn't continue where it left off


vdovhanych (approved):
Ok

PR #13333, opened 2024-07-15
fix(suite): Make Welcome screen more responsible for smaller screens

### LINKED ISSUE #13261: Polish Welcome screen

hstastna:
## Description

Make "Download desktop app" and "trezor.io" buttons/links more nicely visible also for smaller screens. Use `flex-flow: row wrap;` css style for it.

Additionally, use `spacingsPx` for margin sizes of the related components where possible instead of hardcoded values.

TODO:
more screens need the similar polishing. For more info see the related issue.

## Related Issue

## Screenshots:
**Before:**

https://github.com/user-attachments/assets/4d48af61-58fc-45c9-bd17-498c44262b81

**After:**

https://github.com/user-attachments/assets/d5e3fe9c-201c-499d-a732-9a27f9f90eb3

### DIFF
--- a/packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeLayout.tsx
+++ b/packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeLayout.tsx
@@ -26,7 +26,7 @@ import { GuideButton, GuideRouter } from 'src/components/guide';
 import { useGuide } from 'src/hooks/guide';
 import { MAX_ONBOARDING_WIDTH } from 'src/constants/suite/layout';
 import { NavSettings } from './NavSettings';
-import { Elevation, mapElevationToBackground } from '@trezor/theme';
+import { Elevation, mapElevationToBackground, spacingsPx } from '@trezor/theme';
 
 const Wrapper = styled.div`
     display: flex;
@@ -47,6 +47,7 @@ const Expander = styled.div`
     justify-content: center;
     align-items: center;
     flex: 1;
+    margin-top: 96px;
 `;
 
 const WelcomeWrapper = styled.div<{ $elevation: Elevation }>`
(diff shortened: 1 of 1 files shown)


packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeLayout.tsx
```diff
@@ -47,6 +47,7 @@ const Expander = styled.div`
     justify-content: center;
     align-items: center;
     flex: 1;
+    margin-top: calc(${spacingsPx.xxxxl} * 2);
```
peter-sanderson: Not sure maybe just place `96px`? If its not semantic. (WDYT @jvaclavik ?)

jvaclavik: I agree with using just `96px`

packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeLayout.tsx
```diff
@@ -69,18 +70,20 @@ const MotionWelcome = styled(motion.div)`
...
 
 const LinksContainer = styled.div`
-    position: absolute;
     bottom: 0;
     display: flex;
-    margin: 24px 0;
+    margin: ${spacingsPx.xl};
+    align-items: center;
+    flex-flow: row wrap;
+    gap: 14px;
```
jvaclavik: Maybe could we use `spacingsPx.md` here? Does it look good?

hstastna: Very little visual change of the gap, IMHO both options are ok, see 14px vs 16 px (difference almost unseen):
gap
I think we can safely use `spacingsPx.md` here.

jvaclavik (approved):

Issue #13230, opened 2024-07-08
Electrum does not detect connected device with bridge 3.0


bosomt:
**Describe the bug**
Electrum does not detect connected device with bridge 3.0.0
Bridge 2 works fine

**Steps to reproduce:**
1. Start Trezor Suite with Bridge 3.0
2. Start electrum-4.5.5
3. Try to detect connected device

**Info:**
 - Suite version: desktop 24.7.1 (88dbe0410ac5b6245f392ce13e8c30f0f8f549f2)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.7.1 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.0 regular (revision dd4671a5104952ef505d28d1f9e94d1484b4607a)
 - Transport: BridgeTransport 3.0.0

**Screenshots:**
![image](https://github.com/trezor/trezor-suite/assets/31506317/ad809fa9-b6f3-4f94-85b8-ec67302d6db7)
![image](https://github.com/trezor/trezor-suite/assets/31506317/3283531a-4e33-41c5-a8fc-cd26b89f85c7)

Hannsek:
Electrum use trezorlib for communication, no? 🤔

bosomt:
Looks like it https://github.com/spesmilo/electrum/pull/9118
But they dont detect any device not just TS5

mroz22:
trezorlib picks bridge as the primary choice if it is available. so it should work

bosomt:
same result for Exodus, 
im running 24.7.1 version with node bridge and latest Exodus app does not detect TS3 connected to computer

bosomt:
**QA OK**

tested with electrum version `electrum-4.5.3`
and with latest public version of exodus wallet app 24.29.2

**Info:**
 - Suite version: desktop 24.8.0 (68068a49c689e6441ffbf3d164ea26b7c9ef3886)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.8.0 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
[...]

PR #13334, opened 2024-07-15
Transport bridge configure

mroz22:
while investigating #13230 I found that packages/transport-bridge does not implement the [/configure endpoint](https://github.com/trezor/trezord-go/blob/db03d99230f5b609a354e3586f1dfc0ad6da16f7/server/api/api.go#L36)

we are not using it but could be that some legacy software does. so lets keep it compatible.


karliatto (approved):
utACK

PR #13349, opened 2024-07-15
fix(suite): Make final onboarding step more responsible

### LINKED ISSUE #13261: Polish Welcome screen

hstastna:
## Description

Also change the main button color - should be `variant="primary"`.

## Related Issue

## Screenshots:
**Before:**

https://github.com/user-attachments/assets/d10e4bc7-0cd8-4b0c-8285-f258bd3a7bf2

**After:**

https://github.com/user-attachments/assets/1ad10303-f914-4aae-87c3-2bc7e3ce1e9a

### DIFF
--- a/packages/suite/src/views/onboarding/steps/Final.tsx
+++ b/packages/suite/src/views/onboarding/steps/Final.tsx
@@ -66,6 +66,7 @@ const SetupActions = styled.div`
     border-bottom: 1px solid ${({ theme }) => theme.STROKE_GREY};
     width: fit-content;
     gap: ${spacingsPx.md};
+    flex-flow: row wrap;
 `;
 
 const EnterSuiteButton = styled(Button)`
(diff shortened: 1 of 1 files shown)


tomasklim:
Screenshot 2024-07-15 at 15 54 46
`Přejít do Suite` is aligned to left, the rest is centered

jvaclavik (approved):
much better!

hstastna:
> Screenshot 2024-07-15 at 15 54 46 `Přejít do Suite` is aligned to left, the rest is centered

Hmm, I think this is expected result, as it is aligned to the left side originally (also in bigger screens), and the window is still not so small, but we're gonna make sure with the designer, of course, who's not available in these days. Maybe some adjustments will follow up. But this PR is definitely some improvement of the actual behavior, so we merged that.

PR #13356, opened 2024-07-16
chore(wallet-core): lint warnings fix

mroz22:
noted some warnings in a PR 
image

shouldn't the lint checks be turned to error instead of warning?


tomasklim (approved):
thanks for fix

PR #13367, opened 2024-07-16
chore(deps): update trends libs

tomasklim:
## Description

- update non-bitcoin related deps


AdamSchinzel (commented):
`tsx` and `ajv` are not really trends libraries. What was the reason to upgrade/downgrade there? Maybe better in separate commit?

tomasklim:
`ajv` is used for token definitions. `tsx` was just cleaning because there was one occurence of different version

AdamSchinzel (approved):

Issue #5819, opened 2022-07-20
Trezor Suite not connecting from within Tails OS through TOR to Trezor Servers


NotwhatIwanttotell:
Trezor Suite does not connect to Trezor servers trough Tails OS

-   Remembered Wallet: yes
-   Suite Version: 22.7.2 Linux Appimage
-   Browser: Tails OS 
-   OS: Tails OS 5.2
-   Bridge:
sudo iptables -I OUTPUT -p tcp --dport 21325 -j ACCEPT -s localhost -d localhost
sudo iptables -I INPUT -p tcp --dport 21325 -j ACCEPT -s localhost -d localhost
-   Firmware Version: 2.5.1

Trezor Suite used to work very well from within Tails OS after running the "iptables" in a terminal and enabling traffic through the Tor network. Lately the suite does not connect to Trezor servers and its impossible to see the crypto addresses in the wallet.

![image](https://user-images.githubusercontent.com/83304971/179992364-65aba816-37e2-4d71-ad86-4020826e5b96.png)
![image](https://user-images.githubusercontent.com/83304971/179990873-bea78e70-2372-4df1-b973-f00dd1fa170a.png)
![image](https://user-images.githubusercontent.com/83304971/179991050-da7a3201-04b2-49fa-96c2-491c0a87a2b4.png)
![image](https://user-images.githubusercontent.com/83304971/179991191-ce5a1448-af54-4135-afd2-b35333816cd4.png)
![image](https://user-images.githubusercontent.com/83304971/179991643-5816347d-5338-4d0a-9ad7-5f49a6298a84.png)

The Suite tries to connect to Trezor servers but after approximately 30 seconds it fails. As you can see the Trezor T is connected (left upper corner of first screenshot), unfortunately the addresses are not visible.

Suggestion:
Make Trezor Suit loadable through Synaptic Package Manager. Make Trezor Suit work through Tails OS/Tor in the same way as Electrum and Featherwallet. If Electrum.appimage works from within Tails OS Trezor should be able as well.

NotwhatIwanttotell:
As both Trezor Suit and Tails OS has been updated recently I decided to run a test. Unfortunatelly the issue is still present. Trezor Suit can not connect to Trezor servers from within Tails OS.

Suite Version: 22.7.3 Linux Appimage
Browser: TOR Browser 11.5.1
OS: Tails OS 5.3

Tried bridge:
[...]

PR #13368, opened 2024-07-16
wip: extenal Tor daemon

karliatto:
## Description

This PR allows to run Trezor Suite with external Tor, in order to allow it to work in operating systems like Tails. In order for it to work it requires to set a custom configuration so Tor enables Control Port and Trezor Suite knows how to communicate with it.

## How to test it:
You need to create or update your torrc file in order to have the configurations required by Trezor Suite. Then you run Tor like: `tor -f /path/to/torrc`

1. Modify your `torrc` file to contain the configuration below. Modify the section `/path/to/your/tor/directory` to match your Tor path.

```jsx
AvoidDiskWrites 1
Log notice stdout
DormantCanceledByStartup 1
SocksPort 9050 ExtendedErrors KeepAliveIsolateSOCKSAuth
SocksTimeout 30
MaxCircuitDirtiness 1800
DataDirectory /path/to/your/tor/directory
ControlPort 35095
CookieAuthentication 1
CookieAuthFile /path/to/your/tor/directory/.tor/control_auth_cookie

```

1. Enable Tor external experimental feature in Trezor Suite.
2. Enable Tor External experimental feature.
3. After enable Tor External experimental feature in the Tor section there will be Tor Directory section and you have to set `/path/to/your/tor/directory` that you chose in your `torrc` .
4. Enable Tor in Trezor Suite (it already enabled then disable and enable it)

When running and doing discover with a device connected you should be able to make sure that all the connection is going throw Tor by running `watch -n 1 'lsof -i TCP | grep electron'` and checking that there is not connection going to the open internet. You should see TCP connection to `localhost:9050` and `localhost:35095` and  in your standalone Tor running some logs like `New control connection opened from 127.0.0.1.`.

### In Tails OS

1. Edit the torrc
```bash
    sudo vim /etc/tor/torrc
```
* Changes for Tails OS

```jsx
AvoidDiskWrites 1
Log notice stdout
DormantCanceledByStartup 1
SocksPort 9050 ExtendedErrors KeepAliveIsolateSOCKSAuth
SocksTimeout 30
MaxCircuitDirtiness 1800
DataDirectory /home/amnesia/.tor
ControlPort 9051
CookieAuthentication 1
CookieAuthFile /home/amnesia/.tor/control_auth_cookie
```
2. Restart tor service to use the modified torrc:
```bash
    sudo systemctl restart tor
```

### :exclamation: Issues with Tails 
* Tor ControlPort is blocked only for some applications so it cannot be used from Trezor Suite https://tails.net/contribute/design/

> To prevent this, access to the Tor control port is only granted to the root user, as well as to the members of the debian-tor group (via the control socket). A filtering proxy to the control port runs on port 9051 (i.e. the default Tor ControlPort), so for instance Torbutton still can perform safe commands like SIGNAL NEWNYM. It allows defining fine-grained lists of allowed commands (and their arguments) and events on a per-application basis, which can enforce rules like "this $user (e.g. amnesia) when running this $application (e.g. /usr/bin/onionshare) from this network namespace, can only run these commands (ADD_ONION etc.) and listen to these events (e.g. HS_DESC, which is expected after a successfull use of ADD_ONION)".

*

## Related Issue
Related https://github.com/trezor/trezor-suite/issues/5819

## Screenshots:
image


karliatto:
Closing since this was done in a different way in PR https://github.com/trezor/trezor-suite/pull/15642

PR #13370, opened 2024-07-16
feat(suite): add debug setting option to stop tor

karliatto:
## Description

Adding debug option to stop Tor.
When Tor is in bootstrap process there is no way to stop it in the current UI. When developing it is useful to be able to stop it so this allows it.


mroz22 (approved):
looks ok, maybe @komret wants to see it too?

komret (approved):

PR #13380, opened 2024-07-17
fix(suite): reflect hidden and shown tokens in txs history

### LINKED ISSUE #13039: EVM Tokens/ (un)hidden tokens are not reflected in transaction history
**Describe the bug**
When I unhide scam token, it is still marked as scam in tx history

**Info:**

-   Suite Version: develop

**Expected behavior**

- I would expect that unhidden tokens are not marked anymore
- I am not sure if I would expect that hidden tokens are now marked as "scam". Probably not
- I am not able to mark/unmark tokens with balance 0 because they are hidden, I would probably like to be able to hide/unhide tokens with balance 0 so there should probably be option to show 0 balance tokens using some filter

AdamSchinzel:
## Description

When user hides a token then transactions with this token will be blurred and marked as scammy. The opposite happens when user unhides an unverified or already hidden token.

## Related Issue

Resolve #13039

### DIFF
--- a/suite-common/wallet-utils/src/antiFraud.ts
+++ b/suite-common/wallet-utils/src/antiFraud.ts
@@ -21,9 +21,18 @@ export const getIsFakeTokenPhishing = (
     !transaction.tokens.some(tokenTx => {
         const isNftTx = isNftTokenTransfer(tokenTx);
         const definitions = isNftTx ? tokenDefinitions?.nft?.data : tokenDefinitions?.coin?.data;
+        const hide = isNftTx ? tokenDefinitions?.nft?.hide : tokenDefinitions?.coin?.hide;
+        const show = isNftTx ? tokenDefinitions?.nft?.show : tokenDefinitions?.coin?.show;
 
-        return isTokenDefinitionKnown(definitions, transaction.symbol, tokenTx.contract);
-    }); // all tokens are unknown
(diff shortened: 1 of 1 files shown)


tomasklim (approved):

Hannsek:
Sorry that I step in now, but I think this is not right as I would usually hide 0 balance tokens and not scammy tokens (those will be hidden automatically). So marking user-hidden tokens/transactions as scammy feels weird. My two cents.

bosomt:
**QA NOK**

see production vs dev, now all tokens says unhide

image

image

**Info:**
 - Suite version: web 24.8.0 (a707883081ddf9f3b75ececcb0e8a38732136988)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: WebUsbTransport

bosomt:
Please confirm that its related to this change so i can reopen connected issue or create new issue @AdamSchinzel

AdamSchinzel:
Good catch. It was caused by [this PR](https://github.com/trezor/trezor-suite/pull/13407), already fixed it, will create PR so no new issue needed. Thanks for reporting.

PR #13403, opened 2024-07-17
fix(suite): update the country select component on trade dashboard

### LINKED ISSUE #11462: Empower/Trade - unable to type country in the dropdown menu
In the Suite Trade section, users can select the country to display offers from. However, currently, users can only scroll down to choose the country, which is a time-consuming process. Therefore, it would be helpful to add a functionality that allows users to simply type the name of the country they want to search for, which will make the search much faster and more efficient.

image

seibei-iguchi:
## Description

Updated the country select component on trade dashboard so that is more obvious it can be used to type in the country. Fixed some visual issues. Could not fix the issue where the selected country can be deleted with backspace, should be supported by the underlying react-select component (via props `isClearable` and `backspaceRemovesValue`) but custom components probably prevent it to work properly.

### Before:

Screenshot 2024-07-17 at 20 37 56

### After:

Screenshot 2024-07-17 at 20 37 07

### DIFF
--- a/packages/components/src/components/form/Select/Select.tsx
+++ b/packages/components/src/components/form/Select/Select.tsx
@@ -170,7 +170,7 @@ const Wrapper = styled.div<WrapperProps>`
         flex-wrap: nowrap;
         min-width: ${({ $minValueWidth }) => $minValueWidth};
         justify-content: ${({ $isClean }) => ($isClean ? 'flex-end' : 'flex-start')};
-        padding: 0;
+        padding: 0 ${spacingsPx.xl} 0 0;
         border: none;
     }
 
@@ -179,16 +179,25 @@ const Wrapper = styled.div<WrapperProps>`
         display: flex;
         align-items: center;
         justify-content: ${({ $isClean }) => ($isClean ? 'flex-end' : 'flex-start')};
-        width: 100%;
         max-width: initial;
         color: ${({ $isDisabled, theme }) =>
             $isDisabled ? theme.textDisabled : theme.textDefault};
         border-style: none;
         transform: none;
-        margin-left: 0;
+        margin: 0;
 
         &:hover {
-            cursor: ${({ $isSearchable }) => $isSearchable && 'text'};
+            cursor: ${({ $isSearchable }) => ($isSearchable ? 'text' : 'pointer')};
+        }
+    }
+
+    .${reactSelectClassNamePrefix}__single-value + .${reactSelectClassNamePrefix}__input-container {
+        margin-left: 0;
+    }
+
+    .${reactSelectClassNamePrefix}__input-container {
+        &:hover {
+            cursor: ${({ $isSearchable }) => ($isSearchable ? 'text' : 'pointer')};
         }
     }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/form/Select/Select.tsx
```diff
@@ -179,13 +179,24 @@ const Wrapper = styled.div<WrapperProps>`
         display: flex;
         align-items: center;
         justify-content: ${({ $isClean }) => ($isClean ? 'flex-end' : 'flex-start')};
-        width: 100%;
+        width: ${({ $isSearchable }) => !$isSearchable && '100%'};
```
tomasklim: this will result into invalid css in case that isSearchable is false

packages/components/src/components/form/Select/Select.tsx
```diff
@@ -179,13 +179,24 @@ const Wrapper = styled.div<WrapperProps>`
...
         max-width: initial;
         color: ${({ $isDisabled, theme }) =>
             $isDisabled ? theme.textDisabled : theme.textDefault};
         border-style: none;
         transform: none;
-        margin-left: 0;
+        margin: 0;
+
+        &:hover {
+            cursor: ${({ $isSearchable }) => $isSearchable && 'text'};
```
tomasklim: same here

packages/components/src/components/form/Select/Select.tsx
seibei-iguchi: Noted, thanks. Had thought Styled would handle this and not render the property at all. I will fix it:)

packages/components/src/components/form/Select/Select.tsx
```diff
@@ -170,7 +170,7 @@ const Wrapper = styled.div<WrapperProps>`
         flex-wrap: nowrap;
         min-width: ${({ $minValueWidth }) => $minValueWidth};
         justify-content: ${({ $isClean }) => ($isClean ? 'flex-end' : 'flex-start')};
-        padding: 0;
+        padding: 0 25px 0 0;
```
jvaclavik: We try to use predefined spacings if possible.

Could you use it this way? 🙏 
`padding: 0 ${spacingsPx.xl} 0 0;`

packages/suite/src/views/wallet/coinmarket/buy/components/BuyForm/Footer/index.tsx
```diff
@@ -36,6 +36,7 @@ const Label = styled.div`
     padding-top: 1px;
     color: ${({ theme }) => theme.TYPE_LIGHT_GREY};
     font-weight: ${variables.FONT_WEIGHT.MEDIUM};
+    margin-right: 10px;
```
jvaclavik: please use spacings here🙏 

`margin-right: ${spacingsPx.xs}`

jvaclavik (commented):
Just few issues and ready to 🚢

packages/components/src/components/form/Select/Select.tsx
peter-sanderson: We try to user constants for spacing everywhere:  `${spacings.xl}` is `24px` can we use this here? 

If not and 25 is just needed in this case, no no big deal leaving it here.

packages/components/src/components/form/Select/Select.tsx
```diff
@@ -179,13 +179,24 @@ const Wrapper = styled.div<WrapperProps>`
...
         transform: none;
-        margin-left: 0;
+        margin: 0;
+
+        &:hover {
+            cursor: ${({ $isSearchable }) => $isSearchable && 'text'};
+        }
+    }
+
+    .${reactSelectClassNamePrefix}__single-value + .${reactSelectClassNamePrefix}__input-container {
```
peter-sanderson, line 194: @jvaclavik This component is full of the Dark Arts 🔮. WEN floating UI?

jvaclavik: The sooner, the better. Wen you have free week? 😀

peter-sanderson: Me? You have the biggest team now... :D

peter-sanderson (approved):

PR #13415, opened 2024-07-18
fix(suite): Make view only promo content screen more responsive

### LINKED ISSUE #13261: Polish Welcome screen

hstastna:
## Description

Use `spacingsPx` for some css padding/margin values instead, where possible, add some more small adjustments.

## Related Issue

## Screenshots:
**Before:**
promo_before

**After:**
promo_after

### DIFF
--- a/packages/suite/src/views/view-only/MacWindow.tsx
+++ b/packages/suite/src/views/view-only/MacWindow.tsx
@@ -1,4 +1,4 @@
-import { ElevationContext, useElevation } from '@trezor/components';
+import { ElevationContext, useElevation, variables } from '@trezor/components';
 import {
     Elevation,
     borders,
@@ -9,14 +9,22 @@ import {
 import { ReactNode } from 'react';
 import styled from 'styled-components';
 
+const MAC_WINDOW_HEIGHT = '184px';
+
 const Window = styled.div<{ $elevation: Elevation }>`
-    width: 258px;
-    height: 184px;
+    min-width: 258px;
+    height: ${MAC_WINDOW_HEIGHT};
     border: 1px solid ${mapElevationToBorder};
     background-color: ${mapElevationToBackground};
     border-bottom: 0;
-    border-top-left-radius: ${borders.radii.sm};
-    border-top-right-radius: ${borders.radii.sm};
+    border-radius: ${borders.radii.sm} ${borders.radii.sm} 0 0;
+    flex: 1 1 auto;
+
+    @media (max-width: ${variables.SCREEN_SIZE.SM}) {
+        height: calc(${MAC_WINDOW_HEIGHT} + ${spacingsPx.md});
+        border-radius: ${borders.radii.sm};
+        padding-bottom: ${spacingsPx.md};
+    }
 `;
 
 const WindowBar = styled.div<{ $elevation: Elevation }>`
--- a/packages/suite/src/views/view-only/ViewOnlyPromoContent.tsx
+++ b/packages/suite/src/views/view-only/ViewOnlyPromoContent.tsx
@@ -6,6 +6,7 @@ import {
     Image,
     Text,
     useElevation,
+    variables,
 } from '@trezor/components';
 import { FormattedCryptoAmount, Translation } from '../../components/suite';
 import styled from 'styled-components';
@@ -29,7 +30,7 @@ const StyledCard = styled(Card)`
     flex-direction: column;
     gap: ${spacingsPx.md};
     padding: ${spacingsPx.xxs};
-    width: 476px;
+    max-width: 476px;
 `;
 
 const Callout = styled.div`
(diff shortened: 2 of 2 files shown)


packages/suite/src/views/view-only/MacWindow.tsx
```diff
@@ -10,13 +10,21 @@ import { ReactNode } from 'react';
...
     height: 184px;
     border: 1px solid ${mapElevationToBorder};
     background-color: ${mapElevationToBackground};
     border-bottom: 0;
     border-top-left-radius: ${borders.radii.sm};
     border-top-right-radius: ${borders.radii.sm};
+    flex: 1 1 auto;
+
+    @media (max-width: ${variables.SCREEN_SIZE.SM}) {
+        height: calc(184px + ${spacingsPx.md});
```
jvaclavik: could you extract 184px to const?
like: 

```tsx
const WINDOW_HEIGHT = 184

const Window = styled.div<{ $elevation: Elevation }>`
    height: ${WINDOW_HEIGHT}px;
    ...

    @media (max-width: ${variables.SCREEN_SIZE.SM}) {
        height: calc(${WINDOW_HEIGHT}px + ${spacingsPx.md});
        ...

File: packages/suite/src/views/view-only/MacWindow.tsx [hunk 2]
```diff
@@ -10,13 +10,21 @@ import { ReactNode } from 'react';
...
+        border-bottom-left-radius: ${borders.radii.sm};
+        border-bottom-right-radius: ${borders.radii.sm};
```
jvaclavik: I think you don't have to duplicate these styles, because it's in upper scope

packages/suite/src/views/view-only/ViewOnlyPromoContent.tsx
```diff
@@ -67,6 +68,22 @@ const LargeDeviceImage = styled(DeviceImage)`
...
+    align-items: center;
+    justify-content: center;
+    padding: ${spacingsPx.zero} ${spacingsPx.md};
+    width: 100%;
+    overflow: hidden;
+
+    @media (max-width: ${variables.SCREEN_SIZE.SM}) {
+        padding-left: ${spacingsPx.lg};
+        padding-right: ${spacingsPx.lg};
+        padding-bottom: ${spacingsPx.sm};
```
jvaclavik: Wouldn't be easier to read something like this?

```ts
padding: 0 ${spacingsPx.lg} ${spacingsPx.sm};
```

jvaclavik (commented):
Looks much better. Few comments and ready to go

packages/suite/src/views/view-only/MacWindow.tsx
hstastna: This adds bottom radius as in this case the bottom of the `MacWindow` is not connected with the rest of the content, hence it looks kinda weird without the radius, so I've added it here conditionally, as we don't want it for bigger screens.
In the upper scope, there's only top radius added to the upper corners. But let me see if we can simplify this.

packages/suite/src/views/view-only/ViewOnlyPromoContent.tsx
hstastna: Yep, of course! Just forgot to do this before pushing the code here..

peter-sanderson:
`responsible` -> `responsive` ? :)

peter-sanderson (approved):
:clap: thanks for cleaning up my ugly CSS coding :bow: :grin:

packages/suite/src/views/view-only/ViewOnlyPromoContent.tsx
```diff
@@ -126,6 +146,18 @@ const ChartText = styled.div`
...
+        flex-flow: row wrap;
+    }
+`;
+
+const ViewOnlyButton = styled(Button)`
+    width: 50%;
+
+    @media (max-width: ${variables.SCREEN_SIZE.SM}) {
+        width: 100%;
+    }
```
jvaclavik: I'm thinking could we avoid overriding Button component? 🙏 We try to avoid overriding as much as possible, because it increases complexity and lowers predictability. 

I think it should be possible to do it here with following changes:
- `ButtonsContainer` should contain media queries logic and it should change `flex-direction`
- `Button` component should have `isFullWidth` prop which is the same like `width: 100%`

WDYT?

hstastna: Of course I do know we try to avoid overriding components as much as possible. I wouldn't do it if I saw some nicer ways to achieve the expected look and behavior.

Suggested `ButtonsContainer` changes could work, but without adjusting `Button` itself here I'm not sure if we can achieve the expectations easily. The thing is that...

...with `isFullWidth` the button takes too much horizontal space for big screens (when used with css flex props we need to make the screen more responsive). We'd still need to specify its width to avoid that and when we do this, using `isFullWidth` becomes unnecessary, so I've removed that.

But let's take some more time to figure this out...

jvaclavik (approved):
Nicely done! Works like a charm 👍

PR #13416, opened 2024-07-18
feat(suite): Adjust AssetShareIndicator to display any tokens icons

### LINKED ISSUE #13630: Network/Token icon
Prerequisite https://github.com/trezor/trezor-suite/pull/12694

**Description:**
Create component to show icon/logo of an asset.
Create another component which will add ring around the asset. This ring should have part filled using dominant color of the icon.

Screenshot 2024-08-02 at 22 25 38

In case file is not available, show first letter

Screenshot 2024-08-02 at 22 26 58

In case file nor text are available, show `?` instead of the letters.

_____ 
**Files:**

Assets which we have as a network are available offline in `packages/components/src/images/coins`
- naming: 
  - `[network].svg`
  - as it is svg, it can be scaled to whatever sizes

Assets available on `data.trezor.io`
- naming: 
  - in case of network `[coingecko_id].webp`
  - in case of token `[coingecko_id]--[contract_address].webp`
- icon display size: `24x24` 
- available sizes: 
  - `24x24` 
    - https://data.trezor.io/suite/icons/coins/bitcoin.webp
  - `48x48` 
    - `@2x` add to the name of file: `[platform]--[contract_address]@2x.webp`
    - https://data.trezor.io/suite/icons/coins/bitcoin@2x.webp

> https://github.com/trezor/trezor-suite/pull/13768 there is change from `_` separator to `--`

354698861-f7c9acbe-da6e-443c-b8e7-35af38ec9964

_____
**Beware of:**
- `networkSymbol` !== `coingecko_id`
  - `coingecko_id` is available in `networksConfig.ts` for each network except testnets
- to find out how `@2x` should be used, visit `Image.tsx`
- offline icon is preferred (just for networks) (those can have bigger size if we need it, please decide)
- in case network is `ada` please try fetch 2 urls, see `buildCoinUrls` in `coingecko.ts`
```
    if (ticker.symbol === 'ada') {
        const { policyId } = parseAsset(contract_address);

        return [`cardano_{policyId}`, `cardano_{contract_address}`];
    }
```

______
**Api:**
```tsx
// CoinLogo (can be deprecated in the future because we will have AssetLogo)
// ------------------------------------------
// we already use this component

type CoinLogoProps = {
  size: number; // can be any, we have it as SVG
  symbol: NetworkSymbol;
}

<CoinLogo symbol="btc" size={30} />

// AssetLogo 
// ------------------------------------------
// new component for rendering token logo
// icon is fetched from API 

type AssetLogoProps = {
  size: 20 | 24; // + inner implementation for @2x (srcSet on img)
  coingeckoId: string; // this is coingecko id, we can maybe also provide symbol and do translation inside?
  contractAddress?: string; 
  shouldTryToFetch?: boolean; // (by default true) we know that some images are not available
  placeholder: string; // letters when unavailable
}

<AssetLogo 
  coingeckoId="binance-smart-chain" 
  contractAddress="0x0203d275d2a65030889af45ed91d472be3948b92"
  size={24} 
  shouldTryToFetch={true} 
  placeholder="CATWIFHAT"
/>


// ShareIndicator
// ------------------------------------------
// - uses coin/token logos icons independently

type ShareIndicatorProps = {
  percentageShare: number; // 0-100
  image: ReactNode;
}

// ShareIndicator using coin logo
<ShareIndicator 
  percentageShare={90}
  image={<CoinLogo 
    symbol="btc" 
    size={30}
  />}
/>

// ShareIndicator using asset logo
<ShareIndicator 
  percentageShare={90}
  image={<AssetLogo 
      coingeckoId="binance-smart-chain" 
      symbol //
      contractAddress="0x0203d275d2a65030889af45ed91d472be3948b92"
      size={24} 
      shouldTryToFetch={true} 
      placeholder="CATWIFHAT"
  />}
/>
```

hstastna:
## Description

Make both existing `AssetShareIndicator` and `CoinLogo` components accept `coingeckoId`, `configcontractAddress`, ~~`quality`~~ props to construct the proper url to fetch the coin/token icon. Make sure the icon behind the generated url exists. Display alternative icon (new `DefaultCoinLogo` component) in case no icon picture exists (the beginning letter of the coin/token name or "?" if no any name/`coingeckoId` provided).

Also make `ProgressCircle` (part of `AssetShareIndicator`) recognize the dominant icon's color to use it for the circle's color (circle that shows the percentage amount of of total user's fiat amount of tokens).

Rename `CoinLogo` to `AssetLogo` and update all its occurrences.

Example usage with the indicator:
```
    <AssetShareIndicator
        coingeckoId="arbitrum-nova"
        contractAddress="0x6dcb98f460457fe4952e12779ba852f82ecc62c1"
        percentageShare={50}
    />
```

Example usage of the simple coin(token) logo:
```
    <AssetLogo
        coingeckoId="arbitrum-nova"
        contractAddress="0x6dcb98f460457fe4952e12779ba852f82ecc62c1"
    />
```

_How to run the test:_
go to packages/components and `yarn test`

TODOs:
- [x] detect the dominant color of the loaded icon to display the proper color of the progress circle
- [x] make sure the dominant color is calculated properly while transparent pixels are ignored, for webp format
- [x] construct the final url according to the related props such as `coingeckoId`, `configcontractAddress`, `quality` (see the Related Issue)
- [x] use the ability to construct the url in the related existing components to reuse them
- [x] if icon not found, display some default image instead, possibly with the beginning letter of the token name or "?"
- [x] rename `CoinLogo` to `AssetLogo` to reflect its usage more
- [x] separate `getDominantColor` function, setup jest and add tests for this function
- [x] add storybook for `AssetShareIndicator` component
- [x] update storybook for `CoinLogo` - renamed to `AssetLogo`
- [x] use only predefined four values for `AssetLogo` size, remove `quality` prop

## Related Issue

Resolve #13630

Note that the description of the mentioned issue has significantly changed since I started working on it. Hence this PR may change rapidly soon too.

## Screenshots:
Example: _arbitrum-nova_, small (16px) and medium (24px) icons without/with the circle:
icoons
No icon found/incorrect url for a specific coin/token:
a
No icon found, unknown coin/token name:
no-icon

Storybook: update `AssetLogo`, add `AssetShareIndicator` and also `DefaultAssetLogo`
asset


hstastna:
```
const alpha = imageData.data[i + 3];
    if (alpha === 0) {
        continue; // ignore completely transparent pixel
    }
```

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -24,15 +25,44 @@ const SvgWrapper = styled.div<Omit<CoinLogoProps, 'symbol'>>`
...
+        height: ${size}px;
+    `,
 );
+
+export const CoinLogo = ({ symbol, iconUrl, className, size = 32, ...rest }: CoinLogoProps) => {
+    const [isLoading, setIsLoading] = useState(true);
+
+    if (symbol) {
+        return (
+            <Wrapper className={className} size={size} {...rest}>
```
peter-sanderson: Not a big deal but `<Wrapper className={className} size={size} {...rest}>` is now duplicated

```ts

return (
    <Wrapper className={className} size={size} {...rest}>
        {symbol ? <ReactSVG .... /> : iconUrl && <> ... </>}
    </Wrapper>
)
```

Would remove the duplication.

Also whole 
```ts
 <ReactSVG
                    src={COINS[symbol]}
                    beforeInjection={svg => {
                        svg.setAttribute('width', `${size}px`);
                        svg.setAttribute('height', `${size}px`);
                    }}
                    loading={() => <span className="loading" />}
                />
```
Can be extracted as a sub-component with just `symbol` as a property.

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -1,18 +1,19 @@
...
-    symbol: CoinType;
+    symbol?: CoinType;
+    iconUrl?: string;
     className?: string;
     size?: number;
     index?: number;
 }
 
-const SvgWrapper = styled.div<Omit<CoinLogoProps, 'symbol'>>`
+const Wrapper = styled.div<Omit<CoinLogoProps, 'symbol'>>`
```
peter-sanderson: I am afraid this would create some react-errors. The custom props in the styled-component has to be prefixed with `$`

See: https://styled-components.com/docs/api#transient-props

We have some tooling for this `packages/components/src/utils/transientProps.ts`

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -24,15 +25,44 @@ const SvgWrapper = styled.div<Omit<CoinLogoProps, 'symbol'>>`
...
-        <ReactSVG
-            src={COINS[symbol]}
-            beforeInjection={svg => {
-                svg.setAttribute('width', `${size}px`);
-                svg.setAttribute('height', `${size}px`);
-            }}
-            loading={() => <span className="loading" />}
-        />
-    </SvgWrapper>
+const StyledIconImage = styled.img<CoinLogoProps>(
```
peter-sanderson: Same situation here with the transient Props

peter-sanderson: Event if this wont generate error the absence of `$` prefix will cause the `size` prop to be leaked into HTML DOM.

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
```diff
@@ -0,0 +1,87 @@
...
+const useAvatarUrl = ({ coingeckoId, quality }: { coingeckoId?: string; quality: Quality }) => {
+    const theme = useTheme();
+
+    const defaultCharacter = '?';
+    const size = QUALITY_SIZE[quality];
+    const firstCharacter = coingeckoId ? coingeckoId[0] : defaultCharacter;
+    const fontColor = theme.iconSubdued.replace('#', '');
+    const backgroundColor = theme.backgroundSurfaceElevation0.replace('#', '');
+
+    const avatarUrl = `https://ui-avatars.com/api/?size=${encodeURIComponent(size)}&name=${encodeURIComponent(firstCharacter)}&bold=true&font-size=0.75&color=${encodeURIComponent(fontColor)}&background=${encodeURIComponent(backgroundColor)}&rounded=true`;
```
peter-sanderson: I really don't like injecting external URLs into Suite.

peter-sanderson: cc @MiroslavProchazka  / @jvaclavik

peter-sanderson: If we just want some `[AB]` text in the round area, I think this is something we can easily style in the pure CSS.

**Why we need an external service to create a two colored letters in a circle with colored background?**

jvaclavik: I agree with @peter-sanderson

jvaclavik: We might create component for it in DS if we are going to use it more than once.

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
```diff
@@ -0,0 +1,87 @@
...
+    AssetShareIndicator,
+    AssetShareIndicatorProps,
+} from '../AssetShareIndicator/AssetShareIndicator';
+import { useTheme } from 'styled-components';
+
+export const UPDATED_ICONS_LIST_URL_BASE = 'https://data.trezor.io/suite/icons/coins/';
+
+const QUALITY = {
+    small: '',
+    big: '@2x',
```
jvaclavik: nitpick: what about `normal` instead of `small`. Because we can use also `@3x` etc.

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
```diff
@@ -0,0 +1,87 @@
...
+};
+
+const QUALITY_SIZE = {
+    small: 24,
+    big: 48,
```
jvaclavik: nitpick: Are we going to add also normal in between in the future? Otherwise I would also rename it to `normal`

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+    return fileName
+        ? `${UPDATED_ICONS_LIST_URL_BASE}${fileName}${QUALITY[quality]}${ICON_FILETYPE}`
+        : '';
+};
+
+const useAvatarUrl = ({ coingeckoId, quality }: { coingeckoId?: string; quality: Quality }) => {
+    const theme = useTheme();
+
+    const defaultCharacter = '?';
```
jvaclavik: nitpick: I would suggest to move it outside of scope and rename it as constant:

```
const DEFAULT_CHARACTER = "?"
```

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: Didn't create any react errors, but we can add `$` to custom props for sure!

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -24,15 +25,44 @@ const SvgWrapper = styled.div<Omit<CoinLogoProps, 'symbol'>>`
...
+    if (symbol) {
+        return (
+            <Wrapper className={className} size={size} {...rest}>
+                <ReactSVG
+                    src={COINS[symbol]}
+                    beforeInjection={svg => {
+                        svg.setAttribute('width', `${size}px`);
+                        svg.setAttribute('height', `${size}px`);
+                    }}
+                    loading={() => <span className="loading" />}
```
jvaclavik: Is there a reason why to use classname and not normal styled component?

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
hstastna: There are many more options to achieve the expected, it was just an idea while avoiding lots of css :)

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: Missed that, thanks!

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
hstastna: I don't know, this is just a first idea. We could also totally avoid using small/big/normal namings here and implement the expected differently.

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: I don't know the author's intentions, you are referring to the code that isn't mine. However it didn't look safe to just remove it.

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
jvaclavik: I think small/normal/big makes sense, it's kind of predictable when we try to use it everywhere

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -10,25 +10,109 @@ const Container = styled.div`
...
 `;
 
+const rgbToHexColor = (rgbColor: string) => {
+    const [r, g, b] = rgbColor.split('-').map(Number);
+    const hex = [r, g, b].map(x => x.toString(16).padStart(2, '0')).join('');
+
+    return '#' + hex;
+};
+
+const getDominantColor = (imageData: ImageData) => {
```
peter-sanderson: This is a nice pure function that can be separated and unit-tested.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -10,25 +10,109 @@ const Container = styled.div`
...
+    const colorMap: { [key: string]: number } = {};
+    let dominantKey = '';
+    let maxCount = 0;
+
+    for (let i = 0; i < imageData.data.length; i += 4) {
```
peter-sanderson: From code it is not  clear why `4` and why `3`, I suggest to use some well-named constants so it is obvious to the reader why those numbers are used

peter-sanderson: 4 -> pixel size? (r,b,g,alpha)?

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -10,25 +10,109 @@ const Container = styled.div`
...
+        const alpha = imageData.data[i + 3];
+        if (alpha !== 255) {
+            continue; // ignore pixel with transparency
+        }
+
+        const key = `${imageData.data[i]}-${imageData.data[i + 1]}-${imageData.data[i + 2]}`;
```
peter-sanderson: optional/opinionated: I would suggest to be a bit more specific `key` -> `imagePixelKey`?

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
hstastna: Just a note that later it was decided to use the terms "small" and "medium", 16px and 24px, and for both to use "@2x" suffix in the url to get a nicer good quality image for those sizes, at least for now.

And what about bigger sizes? That's discussion in progress.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -1,34 +1,128 @@
...
     const theme = useTheme();
+    const [dominantColor, setDominantColor] = useState<string | undefined>();
+
+    useEffect(() => {
+        if (!symbol && iconUrl) {
+            const image = new Image();
+
+            if (image) {
+                image.crossOrigin = 'anonymous';
+                image.onload = () => loadImageAndProcessColor(image, size * 2, setDominantColor); // process color from bigger image to get more accurate result for dominant color
```
jvaclavik: to consider: I think using event listener might be safer 🤷‍♂️

```
image.addEventListener('load', ()=>{...})

image.addEventListener('error', ()=>{...})
```

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -1,34 +1,128 @@
...
+}: ProgressCircleProps) => {
     const theme = useTheme();
+    const [dominantColor, setDominantColor] = useState<string | undefined>();
+
+    useEffect(() => {
+        if (!symbol && iconUrl) {
+            const image = new Image();
+
+            if (image) {
+                image.crossOrigin = 'anonymous';
```
jvaclavik: So does this work? 

Not sure if I like this approach, it should be probably set on the server.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -101,15 +195,31 @@ export const AssetShareIndicator = ({
     size = 32,
     percentageShare,
     index,
+    coingeckoId = '',
+    contractAddress,
+    quality = 'medium',
```
jvaclavik: Why do you call it `quality`? I think it's `size`. Or am I wrong?

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -101,15 +195,31 @@ export const AssetShareIndicator = ({
...
-            index={index}
-        />
-    </Container>
-);
+}: AssetShareIndicatorProps) => {
+    const iconUrl = useAssetUrl(coingeckoId, contractAddress);
+    const coinLogoSize = !symbol ? QUALITY_SIZE[quality] : size;
+
+    return (
+        <Container className={className} $size={coinLogoSize * 2}>
```
jvaclavik: Why is it *2? Is it because `@2x`?

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -1,38 +1,130 @@
...
+    <ReactSVG
+        src={$symbol}
+        beforeInjection={svg => {
+            svg.setAttribute('width', `${$size}px`);
+            svg.setAttribute('height', `${$size}px`);
+        }}
+    />
 );
+
+export const CoinLogo = ({
```
jvaclavik: Btw we might rename it somehow since it's not just for coins, WDYT?

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -1,38 +1,130 @@
...
+    size = 32,
+    coingeckoId = '',
+    contractAddress,
+    quality = 'medium',
+    ...rest
+}: CoinLogoProps) => {
+    const [isLoading, setIsLoading] = useState(true);
+
+    const iconUrl = useAssetUrl(coingeckoId, contractAddress);
+    const firstCharacter = coingeckoId && coingeckoId[0].toUpperCase();
```
jvaclavik: nit: 
```
    const firstCharacter = coingeckoId?.[0].toUpperCase();
```

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -1,38 +1,130 @@
...
+    const coinLogoSize = !symbol ? QUALITY_SIZE[quality] : size;
+
+    const logo = iconUrl ? (
+        <>
+            {isLoading && <span className="loading">Loading...</span>}
```
jvaclavik: To consider: What about showing `SkeletonCircle` instead of text?

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -1,38 +1,130 @@
...
+            <StyledIconImage
+                src={iconUrl}
+                $size={coinLogoSize}
+                onLoad={() => setIsLoading(false)}
+                onError={() => setIsLoading(false)}
```
jvaclavik: What about unify this function? 

Maybe something like this?
```
onLoad={hideLoading}
onError={hideLoading}
```

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -1,34 +1,128 @@
...
+
+const Container = styled.div<{ $size: number }>(
+    ({ $size }) => `
+        position: relative;
+        align-items: center;
+        display: flex;
+        justify-content: center;
+        width: ${$size}px;
+        height: ${$size}px;
+        border-radius: 50%;
```
AdamSchinzel:
```suggestion
        border-radius:  ${borders.radii.full};
```

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -1,34 +1,128 @@
...
+import { CoinLogo, CoinLogoProps, QUALITY_SIZE, useAssetUrl } from '../CoinLogo/CoinLogo';
+
+const Container = styled.div<{ $size: number }>(
+    ({ $size }) => `
+        position: relative;
+        align-items: center;
+        display: flex;
+        justify-content: center;
+        width: ${$size}px;
+        height: ${$size}px;
```
AdamSchinzel: nitpick: maybe it looks better to pass size just here like:

width: ${({ $size }) => $size}px;
height: ${({ $size }) => $size}px;

And not to all styles (applies to all occurrences inside this PR)

packages/components/src/components/assets/DefaultCoinLogo/DefaultCoinLogo.tsx
```diff
@@ -0,0 +1,28 @@
...
+type DefaultCoinLogoProps = {
+    coinFirstCharacter?: string;
+    size?: number;
+};
+
+const DefaultWrapper = styled.div<{ $size?: number }>(
+    ({ $size, theme }) => `
+      width: ${$size}px;
+      height: ${$size}px;
+      border-radius: 50%;
```
AdamSchinzel:
```suggestion
      border-radius:  ${borders.radii.full};
```

AdamSchinzel (dismissed):

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: For me it works fine, but not sure it will work for everybody. Yanas told me he'll try to reach Tali and ask what we can do regarding the server configuration.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: See the original GH issue. There's `quality`. However, these terms are related to each other so I do admit it can be confusing. Let's see what I can do to make it more clear in the code.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: Yes. There is also another way how to write this code to achieve expected.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: Where would you move this function? We can, but IMHO it's not necessary. I don't see any other possible use case for it.

peter-sanderson: Probably `components/utils/getDominantColor.ts`

The main reason is not re-usability, but to be able to write a test for it (it seems weird to export it from `tsx` component file)

peter-sanderson: It seems that there is `0` :exclamation: tests in components. And by adding a first one, you also "won" the work associated with it (add dependecy, setup it, etc...)

@jvaclavik I thinks its worth it to set it up anyway, but maybe in separate PR?

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: It took me a while to figure out. But yeah, it's correct. It's the information for `Container` which must be bigger than the logo itself. And we need to specify its size. But maybe it would be better to do the computation in its css? Would look better, I think.

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: Hmm maybe AssetLogo? WDYT? I think we should keep the name for this component simple.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
jvaclavik: I think it should be pretty easy to setup the testing here.

I prepared PR:
https://github.com/trezor/trezor-suite/pull/13575

Since we will merge it, it should work just with this command: 
`yarn nx run @trezor/components:test`

(and we might need also some new dependency)

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
jvaclavik: I was thinking about it once again because this seems to me confusing (personally I mixed it up with `size`), so I suggest following approach which should be more clear:

```
const QUALITY = {
    standard: '',
    double: '@2x',
    // triple: '@3x', // shows extensibility for the future
}
```

WDYT?

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
jvaclavik: OK, makes sense

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
jvaclavik: This is probably related to https://github.com/trezor/trezor-suite/pull/13416#discussion_r1698189810

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
jvaclavik: I agree with @AdamSchinzel here even though it's not 100% the same. In the future, we should change `borders.radii.full` to  `50%`. But we need to be sure we are not breaking anything.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
jvaclavik: I undestand that it should be bigger but how exactly do we know that it should be doubled?

jvaclavik: Ideally it should work without setting up exact size but it might lead to some visual issues

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
jvaclavik: AssetLogo sounds good 👍

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
hstastna: `standard` will very probably never be used as we've found the smallest possible quality to achieve nice sharp icon look is what you call `double`. I wanted to avoid adding code that would probably never be used. And for `triple` quality, I was told bigger icons than 24px are unfortunately not available, and it doesn't seem they will.

Are you ok with using the following?
```
const QUALITY = {
    double: '@2x',
}
```

IMHO unnecessary but it could make some parts of the code more understandable. And if needed, we always can expand this object in the future.

jvaclavik: In this case why do we need quality since we use just `@2x`? Let's delete `quality` prop then and component will be even simpler, WDYT?

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: However, these terms are related to each other.

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
hstastna: Not sure. Because we still need to distinguish between displaying smaller 16px and bigger 24px icon. Originally, this was meant kinda differently (smaller quality => smaller icon, bigger quality => bigger icon) but since we've found we need @2x (bigger quality) also for the smaller icon, the original approach is not quite true anymore.

jvaclavik: But relation between quality and size should be this:

```
finalResolution = imageSize * quality 
```

So if quality is always `@2x`, we don't need to change it. If we have two sizes of icons (16px and 24px), resolution should be like this:

```
// size = "small"
finalResolution = 16 * 2

// size = "medium"
finalResolution = 24 * 2
```

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: Ok, I will include setup and the test for `getDominantColor.ts` here. However I'm not sure what the purpose of the PR you've opened is. We could simply run `yarn test` in `packages/components` if we want to run the component's tests, after I add some basic setup.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: I had to figure out that the size of the whole component is icon size * 2. Additionally, the original implementation was buggy - the width and height were hardcoded in the Container component, and that lead to too big empty space around the whole component in case of using smaller icon (smaller than 24px) - which we need now. Now with the actual approach, we make sure the size will always be correct and corresponding with the size of the icon we decide to use.

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: Ok. There's gonna be many files changed, in this PR.. 😅🙈

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: This change leads to errors (the picture shows already renamed component to AssetLogo):
errrr
 
 I'll leave this as it was before.

jvaclavik: so `coingeckoId[0]` is `undefined`. I might be missing something but I think this is also not resolved in your previous code.

maybe this way it could work?

```
    const firstCharacter = coingeckoId?.[0]?.toUpperCase();

Comment (hstastna) [reply packages/components/src/components/assets/CoinLogo/CoinLogo.tsx]: Unfortunately this doesn't work. Not sure why. And in my previous code, it works for me. Mystery? Anyway this is such a nitpick that I'm not sure if we want to spend more time on this. Anyway it's definitely interesting. Maybe I have a bug somewhere else? :D Let me see.

File: packages/components/src/components/assets/AssetIcon/AssetIcon.tsx (hunk 5 above)
Comment (hstastna) [reply packages/components/src/components/assets/AssetIcon/AssetIcon.tsx]: I've decided to use `quality` as it is. Why to use both size and quality? Because `AssetShareIndicator` already uses `size` and we don't wanna mix with that, while keeping the original behavior of the component. With `quality` we make sure we choose one of the 2 sizes for the icon (16 and 24px) when not using `symbol` but using the generated url as the icon's source.

File: packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx (hunk 15 above)
Comment (hstastna) [reply packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx:207]: However... I realized we might break the actual usage of AssetShareIndicator in the code if using this change, occurrences of the component that use default 32px icon size and 48px size of the container - which naturally created some nice space around. Ideally, we should keep this while meet the new expectations. Not sure how I'm gonna solve it.

File: packages/components/src/components/assets/CoinLogo/CoinLogo.tsx (hunk 17 above)
Comment (jvaclavik) [reply packages/components/src/components/assets/CoinLogo/CoinLogo.tsx]: No problem, let's skip it

File: packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx [hunk 23]
```diff
@@ -101,15 +195,31 @@ export const AssetShareIndicator = ({
...
-            symbol={symbol}
-            size={size}
-            percentageShare={percentageShare}
-            index={index}
-        />
-    </Container>
-);
+}: AssetShareIndicatorProps) => {
+    const iconUrl = useAssetUrl(coingeckoId, contractAddress);
+    const coinLogoSize = !symbol ? QUALITY_SIZE[quality] : size;
```
peter-sanderson: Inverted logic (`!symbol`) is a bit confusion. Its easier to read and think about it when I don't need to negate stuff in my mind.

peter-sanderson: -> 

```ts
const coinLogoSize = symbol ? size : QUALITY_SIZE[quality];
```

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
```diff
@@ -1,38 +1,130 @@
...
+    coingeckoId = '',
+    contractAddress,
+    quality = 'medium',
+    ...rest
+}: CoinLogoProps) => {
+    const [isLoading, setIsLoading] = useState(true);
+
+    const iconUrl = useAssetUrl(coingeckoId, contractAddress);
+    const firstCharacter = coingeckoId && coingeckoId[0].toUpperCase();
+    const coinLogoSize = !symbol ? QUALITY_SIZE[quality] : size;
```
peter-sanderson: Why is this duplicated here as well. I is both in [outer component](https://github.com/trezor/trezor-suite/pull/13416/files#diff-ff9cb18bace51536407e61224e5497845491a828f4287cc20dc61f02d2d4a947R204) and here.

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: Awesome! How couldn't I have seen that earlier? :D

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
peter-sanderson: Do we need different sizes then 16/24? Can we unify it? **So it doesn't matter if we are using symbol or not.**

I suggest exploring the possibility of allowing only 16px and 24px and just leave it like this everywhere. 

Additionally I would use `*2` everywhere so 32 and 48. Therefore, in the components API we have only those two resolutions. We event may have 16, 24, 32 and 48 resolution available and live with those 4. Regardless of the `symbol` usage.

@hstastna @jvaclavik WDYT

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
```diff
@@ -101,15 +195,31 @@ export const AssetShareIndicator = ({
     size = 32,
     percentageShare,
     index,
+    coingeckoId = '',
+    contractAddress,
+    quality = 'medium',
     ...rest
-}: AssetShareIndicatorProps) => (
```
jvaclavik, line 105: I know it wasn't here before but could you please add storybook for this file? It's hard to test it without it.

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
jvaclavik: Sounds good to me. 

Ideally we should use just one prop for setting size without dependency to setting symbol. Otherwise component is not going to be predictable.

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: Yeah, this might not be obvious, but during working on this task I found we need the same functionality in both `AssetShareIndicator` and `CoinLogo` so we could use both components separately, as needed - according to our need of displaying the indicator together (`AssetShareIndicator`) with the logo or not (`CoinLogo`) while keeping the original existing usage of those components.

See: if we used this logic only in `CoinLogo`, it would be hard for the indicator to know the dominant color if using `CoinLogo` within `AssetShareIndicator`. And if we used the logic only in `AssetShareIndicator`, it would be impossible to display the icon without the indicator using the url we generate from the props - in other words - to use `CoinLogo` itself with the expected functionality. This is the expectation that came from Invity folks. Earlier I've prepared  `AssetShareIndicator` so it could be used universally - it could display or not display the indicator, but that's what folks didn't want. They wanted to be able to use the simpler `CoinLogo` with the expected functionality because they don't need the indicator. While we need the same with the indicator.

peter-sanderson: OK, I got that part. :+1: we need to use them independently.

However in the `AssetShareIndicator` the (https://github.com/trezor/trezor-suite/pull/13416/files#diff-ff9cb18bace51536407e61224e5497845491a828f4287cc20dc61f02d2d4a947R210) the `coinLogoSize` is passed into `size` and here it is again run thought same logic. 

At least I would pass just `size` and lets use `coinLogoSize` only for the `Container` in the `AssetShareIndicator`

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
hstastna: Please check the issue ticket or reach @MiroslavProchazka or @Hermez-cz for more discussion about that. The need for two specific sizes was requested while using `quality` prop (first it was 24 and 48px-for quality2x, later 16 and 24 while using 2x for both but we needed to distinguish between those 2 sizes somehow anyway). So I've implemented that. I believe there were many reasons behind what was expected to implement.

> Ideally we should use just one prop for setting size without dependency to setting symbol. Otherwise component is not going to be predictable.

This component works like this: if symbol defined, it will be working as before - using size etc. If props for url defined, quality prop will be taken into account and one of the 2 sizes for the icon will be chosen. With this approach we make sure nothing will be broken for existing usages of the components, just new functionality added. 

Here's another why: we cannot use bigger sizes for icons downloaded via that new generated url because the images are not big enough and it wouldn't look ok. Of course this was already tested and discussed before I made these changes.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: Done.

packages/components/src/components/assets/AssetShareIndicator/AssetShareIndicator.tsx
hstastna: Done.

hstastna:
When reviewing this PR, maybe it worth to check this discussion about sizes: https://satoshilabs.slack.com/archives/C04DD84P4A1/p1721989331148269

packages/components/src/components/assets/CoinLogo/CoinLogo.tsx
hstastna: We can do that but here are a few why-s, why I didn't do that:
- the code of `AssetShareIndicator` would IMHO look kinda weird as one would ask why we pass `size` to one sub-component while `coinLogoSize` to the other one, for the same prop => inconsistency. One would need to investigate more to understand <- unnecessary
- we still need to set the final `coinLogoSize` (or newly `assetLogoSize`) for `ProgressCircle` anyway, cannot avoid that

packages/components/src/components/assets/AssetIcon/AssetIcon.tsx
jvaclavik: I asked @MiroslavProchazka and @Hermez-cz and they didn't requested `quality` prop but size of the icon.

Before this PR we used `AssetShareIndicator` only once with `size={24}` so @peter-sanderson suggested that we could provide only predefined sizes (`16 | 24 | 32 | 48`) and in this case we don't have to care about `symbol`.
It will simplify API of the component and you could use it without studying inner logic of the component.

hstastna: I do agree simplification would be great, but I must disagree with the rest, as I was involved in some discussions earlier and have kinda different information, and the thing with sizes and qualities is more complicated. And this discussion is already too long, so to be simple, we must distinguish between sizes provided for symbol vs for the url. We cannot offer all to all. At least not for now. We don't have icons ready for that.

> Before this PR we used AssetShareIndicator only once with size={24} ...

That could be ok, but with AssetLogo not from various reasons.

PR #13421, opened 2024-07-18
fix(suite): Fix settings back button

### LINKED ISSUE #13417: Fix SUite returning to settings instead of Dashboard
Please make sure to fix the issue when Suite is returning to Settings page, even though you had to stay in the Dashboard - see following video.

https://github.com/user-attachments/assets/2e46e1cd-a288-4e83-acb2-6c5563f441be

jvaclavik:
## Related Issue

Resolve #13417


jvaclavik:
It's not so easy to fix this issue. It's happening because foreground screens in Suite don't have urls and it's much wider problem. Closing for now.

PR #13424, opened 2024-07-18
fix(suite-native): passphrase UI tweaks

juriczech:
## Description
Missing color definition for secondary button selected gray color as default. Should be red, so I'm defining it in the params.

Second commit changes revalidation mode of passphrase form. @shenkys Would like to bring this approach of validating simple 1 input forms only on submit (hiding the submit button if values are invalid). This brings inconsistency with other forms (xpub import, rename account), however it's been agreed that inconsistency is now okay with this for the purpose of having passphrase already implemented with this approach rather than re-doing other forms as well. He will create an issue to unify this.

## Screenshots:
image

https://github.com/user-attachments/assets/85e48b40-a88d-4f90-a89a-0a642bf1f064


PeKne (approved):

STew790:
**QA OK**

24.7.2 https://github.com/trezor/trezor-suite/commit/6975009cbe1e28d449bd1013ea19ac41f5f3d27a

PR #13433, opened 2024-07-19
test(e2e): adjusting test cleanup method for better stability

ondracja:
- omitting `Cache` folder from cleanup as there's sometimes an ongoing process that breaks the deletion in CI

### DIFF
--- a/packages/suite-desktop-core/e2e/support/common.ts
+++ b/packages/suite-desktop-core/e2e/support/common.ts
@@ -36,7 +36,13 @@ export const launchSuite = async (params: LaunchSuiteParams = {}) => {
     const localDataDir = await electronApp.evaluate(({ app }) => app.getPath('userData'));
 
     if (options.rmUserData) {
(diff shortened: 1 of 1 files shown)


packages/suite-desktop-core/e2e/support/common.ts
```diff
@@ -55,6 +55,16 @@ export const launchSuite = async (params: LaunchSuiteParams = {}) => {
 
     const window = await electronApp.firstWindow();
 
+    const filesToDelete = fse.readdirSync(localDataDir);
```
mroz22: lets reconcile with line 38-40

mroz22 (changes requested):

packages/suite-desktop-core/e2e/support/common.ts
ondracja: Moved to line 38.

mroz22 (approved):
ok, lets try if it is any better

PR #13434, opened 2024-07-19
fix(suite): removing obsolete security check copy

MiroslavProchazka:
Removing obsolete security check copy since we've get rid of the security check boxes a while ago. 

Before
image

After 
image

### DIFF
--- a/packages/suite/src/views/dashboard/components/PortfolioCard/components/EmptyWallet.tsx
+++ b/packages/suite/src/views/dashboard/components/PortfolioCard/components/EmptyWallet.tsx
@@ -35,15 +35,6 @@ const StyledImage = styled(Image)`
     margin-bottom: 24px;
 `;
 
-const SecurityItem = styled.div`
-    color: ${({ theme }) => theme.TYPE_LIGHT_GREY};
-    font-size: ${variables.FONT_SIZE.SMALL};
-
-    & + & {
-        margin-top: 12px;
-    }
-`;
-
 type EmptyWalletProps = HTMLAttributes<HTMLDivElement>;
 
 export const EmptyWallet = (props: EmptyWalletProps) => (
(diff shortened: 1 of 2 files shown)


packages/suite/src/views/dashboard/components/PortfolioCard/components/EmptyWallet.tsx
```diff
@@ -35,15 +35,6 @@ const StyledImage = styled(Image)`
     margin-bottom: 24px;
 `;
 
-const SecurityItem = styled.div`
```
MiroslavProchazka, line 38: cc @jvaclavik Should I keep these margins here or I can get rid of it?

jvaclavik: No 

image

jvaclavik (approved):

packages/suite/src/views/dashboard/components/PortfolioCard/components/EmptyWallet.tsx
hstastna: If `SecurityItem` component isn't used anywhere else in this file (looks like it isn't), you can remove this as you've already done. Looks like just a styled div, styled specifically for one specific sentence.

hstastna: If you are asking about 
```
   & + & {
        margin-top: 12px;
    }
```
this is a type of styling we are trying to get rid of from our codebase as it's less readable, produces various problems etc. This specifically styles siblings of an element. IMHO really no reason to keep it here. If we miss some margin somewhere close to this component, we should style it the other way.

bosomt:
**QA OK**

text is gone

**Info:**
 - Suite version: web 24.8.0 (76ee7c62a1b17ef35ed680a384b2b989aaeb00aa)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982

PR #13437, opened 2024-07-19
Fix/native/add new account visibility

### LINKED ISSUE #13435: Account #5 appears after adding account #3
See video.
Account #5 appears after adding account #3 
I was not able to reproduce it on 24.7.1
@STew790 managed to replicate this issue on BTC account with different seed

- Android 14
- Nothing phone 1
- version 24.7.2 7dd5a5f

image

https://github.com/user-attachments/assets/c61e91b1-c528-49cf-84f3-b253e191b2e6

matejkriz:
## Description

We've changed the adding new account flow to always discover one more account in addition to appearing when a transaction comes in on it. It was supposed to be hidden, but was always visible.

Two more edge cases handled - 
- make sure to change visibility of the empty account with the lowest index (previously it could happen to show the second one)
- do not show errors if the discovery fails, because user want to see the new account and he already see it - he doesn't care about background discovery of one extra account

## Related Issue

Resolve #13435

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -217,13 +217,25 @@ const addAccountByDescriptorThunk = createThunk(
             ...getAccountInfoDetailsLevel(bundleItem.coin),
         });
 
+        const accounts = selectDeviceAccountsForNetworkSymbolAndAccountType(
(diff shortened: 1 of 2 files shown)


vytick (approved):
great job 🚢 🇮🇹

PR #13444, opened 2024-07-22
feat(ci): unused translations auto cleanup

martykan:
## Description

Adds a `--cleanup` option to the `translations:list-unused` script, which automatically removes the unused messages from the file. 

The script also includes a `--pr` option, which will create a new branch with this change and submit a PR. 

This can be run from a new workflow in GitHub actions. The workflow is triggered manually to prevent too many PRs being created.

### DIFF
--- a/.github/workflows/bot-crowdin-sync.yml
+++ b/.github/workflows/bot-crowdin-sync.yml
@@ -1,6 +1,13 @@
 name: "[Bot] Crowdin translations update"
 
-on: [workflow_dispatch]
+on:
+  workflow_dispatch:
+    inputs:
+      remove_unused_translations:
+        description: "Confirm removal of unused translations. Check them first in misc tests pipeline."
+        required: true
+        default: false
+        type: boolean
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
@@ -32,6 +39,9 @@ jobs:
           git checkout -B ${{ env.BRANCH_NAME }}
           yarn workspace @trezor/suite translations:download --token=${{ secrets.CROWDIN_PERSONAL_TOKEN }}
           yarn workspace @trezor/suite translations:backport-en
+          if [ "${{ github.event.inputs.remove_unused_translations }}" = "true" ]; then
+            yarn workspace @trezor/suite translations:list-unused --cleanup
+          fi
           yarn workspace @trezor/suite translations:format
           yarn workspace @trezor/suite translations:extract
           cat packages/suite-data/files/translations/master.json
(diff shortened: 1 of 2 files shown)


.github/workflows/update-unused-messages.yml
```diff
@@ -0,0 +1,32 @@
+name: "[Check] remove unused messages and create PR"
+
+on:
+  workflow_dispatch:
```
mroz22: @karliatto would you please take over this PR and incorporate into as a nightly check similarly to what you are doing now?

mroz22 (approved):

.github/workflows/update-unused-messages.yml
vdovhanych: Is it necessary to add another action? You're also not using some naming convention for the file. We can add it to the same action but just add a trigger that checks if the action was scheduled to run to only do this instead of updating it.

martykan: It's originally part of `[test] misc` which runs nightly, but I didn't set that up to automatically create PRs, since it could potentially create duplicate PRs if the first one is not merged right away. 
Also I have another idea - maybe this could happen as a part of Crowdin sync?

vdovhanych: Yes, doing this every time you sync crowdin seems a bit better to me. 👍

karliatto: So what is the final idea? will it be run with the Crowding sync or with this workflow as well?

komret: Please add it to the Crowdin sync. It should run before the upload to Crowdin.

martykan: The upload also deletes from Crowdin? I'm just thinking whether it would be dangerous if someone ran it without checking the unused strings prior

komret: If a key does not exist in `messages.ts`, Crowdin sync deletes it from Crowdin and the translation JSONs. Can there be false positives from the script removing unused messages? If so, it has to be run separately.

martykan: Most likely not, except for the possibility of dynamic strings that are not marked as such.

.github/workflows/bot-crowdin-sync.yml
```diff
@@ -1,6 +1,10 @@
 name: "[Bot] Crowdin translations update"
 
-on: [workflow_dispatch]
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
```
vdovhanych: this is only for testing right? Or would you like to keep it like this?

martykan: Yeah just testing, it's in a separate commit

.github/workflows/update-unused-messages.yml
martykan: I added a checkbox as an input to confirm removing to make it safer

vdovhanych (approved):

Issue #13281, opened 2024-07-10
Suite silent update


komret:
- [x] Add an option to Experimental Features to enable silent app updates. This should work the same as enabling Password manager.
- [x] When the experimental feature is enabled, show a `SectionItem` to the Application section in `SettingsGeneral` called "Automatic Trezor Suite updates" with `description` "Trezor Suite automatically downloads the latest version in the background and installs it when restarting the app. This ensures you're always up-to-date with the latest features and security patches. Updates occur without requiring your permission." and a `Switch` which is on be default.
- [x] When the setting is on, users don't see the update modal when there is a new version available. Instead, the new version is automatically downloaded in the background. This should work the same as downloading in the background in the current setup.
- [ ] When the new version is downloaded, a notification similar to the one you see when sending a transaction is shown. it says "A new version of Trezor Suite was downloaded. Restart to install." It includes a button saying "Restart". That should work the same as restarting after a new version is downloaded in the current setup.
- [x] Next time Suite is launched, it is the new version.

To investigate: Once the new version is downloaded, is it possible to prevent its installation by toggling the switch in Application settings?

Note: experimental features don't have final copy and don't use translations.

[Notion spec](https://www.notion.so/satoshilabs/Silent-Automatic-Update-ba941c59eb1b4597aa89d248c81b9de2?d=caf0bf86ae6949a5bbc6e0037598184f#791cbcad38944acbb06e95211a8b5144) (may be outdated)

Hannsek:
A changelog should be accesible somewhere. I'd like to see a button in the notification with something like "what's new" and if you click on that you'd see a modal with the changelog… or something like that…

hynek-jina:
@Hannsek 
Changelog is accessible from Settings → Application.
User just needs to click on the version number
<img width="906" alt="Screenshot 2024-07-31 at 19 33 02" src="https://github.com/user-attachments/assets/3e61faef-dd64-4fdd-b7c9-ca683b0d76f9">

Hannsek:
I mean it should be accessible and _after_ the update (before it is in the settings and that is somewhat ok). Something like Brave or Signal update.

bosomt:
- `Automatic Trezor Suite updates` switch is available in Experimental section
- `Automatic Trezor Suite updates` switch is available after enabled via experimental section
- `Automatic Trezor Suite updates` switch disappears when you disable `Experimental features` section
- actual update will be tested with with 24.9.1 release candidate

[...]

PR #13446, opened 2024-07-22
feat(suite): Silent Update as Experimental Feature

peter-sanderson:
In this PR:

- I refactored a Experimental Features a bit so it supports more behavior 
- Added a silent update mechanics, that can be allowed as Experimental Feature. This silent update will **silently** downloads the binary a installs it on quit **without** asking the user. 

Related: https://github.com/trezor/trezor-suite/issues/13281


packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -87,7 +87,7 @@ export interface SuiteSettings {
     isDeviceAuthenticityCheckDisabled: boolean;
     addressDisplayType: AddressDisplayOptions;
     defaultWalletLoading: WalletType;
-    experimental?: ExperimentalFeature[];
+    experimental?: ExperimentalFeatureConfig['id'][];
```
marekrjpolak: There ain't no such thing as `ExperimentalFeatureConfig['id']`. Revert back to `ExperimentalFeature`?

packages/suite/src/constants/suite/experimental.ts
```diff
@@ -1,15 +1,23 @@
 import { TranslationKey } from '@suite-common/intl-types';
+import { isDesktop } from '@trezor/env-utils';
 
-export enum ExperimentalFeature {
-    PasswordManager = 'password-manager',
-}
+export type ExperimentalFeature = 'password-manager' | 'silent-update';
```
marekrjpolak: I thought using enums is encouraged now, so I'm actually glad for this change of yours, it's definitely more practical. :+1: 

But, given that `ExperimentalFeatureConfig` is now required for every feature, can we make `EXPERIMENTAL_FEATURES` single source of truth, maybe like this?
```
export const EXPERIMENTAL_FEATURES = {
   ...
} satisfies Record<string, ExperimentalFeatureConfig>;

// I've never used satisfies before (♥‿♥)

export type ExperimentalFeature = keyof typeof EXPERIMENTAL_FEATURES;
```

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
```diff
@@ -104,14 +108,21 @@ export const Experimental = () => {
                     }
                 />
                 <ActionColumn>
-                    <Switch isChecked={!!features} onChange={onSwitchExperimental} />
+                    <Switch
+                        isChecked={enabledFeatures !== undefined}
```
marekrjpolak, line 119: You must be fun at parties.

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
```diff
@@ -104,14 +108,21 @@ export const Experimental = () => {
...
+                        onChange={onSwitchExperimental}
+                    />
                 </ActionColumn>
             </SectionItem>
             <AnimatePresence>
-                {features && Object.keys(ExperimentalFeature).length && (
+                {enabledFeatures && experimentalFeatures.length > 0 && (
```
marekrjpolak, line 125: You really like unnecessary characters, don't you?

packages/suite/src/actions/suite/suiteActions.ts
```diff
@@ -75,7 +75,12 @@ export type SuiteAction =
           payload: Partial<AutodetectSettings>;
       }
     | { type: typeof deviceActions.requestDeviceReconnect.type }
-    | { type: typeof SUITE.SET_EXPERIMENTAL_FEATURES; payload?: ExperimentalFeature[] }
+    | {
+          type: typeof SUITE.SET_EXPERIMENTAL_FEATURES;
+          payload: {
+              enabledFeatures: ExperimentalFeature[] | undefined; // undefined means the experimental features are off as a whole
```
marekrjpolak, line 92: Is there a reason for this change, or do you just want to be consistent with none of the inconsistent actions above? :upside_down_face:

packages/suite/src/constants/suite/experimental.ts
```diff
@@ -1,15 +1,23 @@
...
 
-type FeatureIntlMap = Partial<Record<ExperimentalFeature, TranslationKey>>;
-
-export const ExperimentalFeatureTitle: FeatureIntlMap = {
-    [ExperimentalFeature.PasswordManager]: 'TR_EXPERIMENTAL_PASSWORD_MANAGER',
+export type ExperimentalFeatureConfig = {
+    title: TranslationKey;
+    description: TranslationKey;
+    isEnabled: () => boolean;
```
marekrjpolak: I agree that `title` and `description` probably should be there for every feature, but what about `isEnabled`? It seems unnecessary to be forced to write `() => true` everywhere. Unless most of the features will have a condition like this :thinking:

If not, one solution may be to switch it to `isDisabled` (or even better to `isHidden`, so it's not mistaken with actually enabling/disabling a feature) and make it optional?

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
peter-sanderson: I usually don't attend parties.

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
peter-sanderson: I do not like unnecessary characters. What I do like is precision and consistency.

packages/suite/src/actions/suite/suiteActions.ts
peter-sanderson: This allows future extension. We started doing this so future changes are not pain. For consistency reasons we are trying to use it everywhere in the same manner ( @jvaclavik  & @juriczech I think)

packages/suite/src/constants/suite/experimental.ts
peter-sanderson: > I thought using enums is encouraged now,

I don't know. My current understanding is that Enums are pain in Javascript. Do you have any source for this? :pray:

peter-sanderson: By doing that inference you lose the type-safety of `EXPERIMENTAL_FEATURES[feature]`. In your proposal this would satisfy any `string`. In current implementation only `'password-manager' | 'silent-update'` will.

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
marekrjpolak: It's perfectly consistent if you do it everywhere! Are you afraid of negative-length anti-arrays? :ghost:

packages/suite/src/constants/suite/experimental.ts
marekrjpolak: Nope?
image

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
peter-sanderson: > It's perfectly consistent if you do it everywhere!

Is a path ;)

> Are you afraid of negative-length anti-arrays? 👻

I am afraid of future changes, that change left part of the expression to something that would (by luck) satisfy the condition. Precision here reduces (not eliminates) the surface for such a mistake.

It's just defensive programming. 

I also like concept of atomic habits. So I don't need to think and pay attention to every detail and think about all possible future scenarios. I am just strict in comparison everywhere.

packages/suite/src/constants/suite/experimental.ts
peter-sanderson: Ah ok, you got me. Didnt know that satisfies works this way. Cool! Thanks!

packages/suite/src/actions/suite/suiteActions.ts
marekrjpolak: I personally can't see an advantage of doing it in case I'll need it someday instead of doing it when I need it, but whatever. :+1:

peter-sanderson: I see a difference here, but lets not prolong/complicate the discussion here. We can talk this more in person someday. :)

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
marekrjpolak: When empty or invalid js arrays will start to return truthy lengths, we'll have bigger problems than that. But whatever. :)

peter-sanderson: I meant that whole `experimentalFeatures.length` will be replaced with something. And instead of a type error you get nothing.

peter-sanderson: Here its pedantic, but in general this habit proven to me to be extremely useful. And because I don't want to think about when to do it and when not. I do it everywhere as it is super minor change to do and I do it automatically.

packages/suite/src/reducers/suite/suiteReducer.ts
peter-sanderson: fixed

packages/suite/src/constants/suite/experimental.ts
peter-sanderson: fixed

packages/suite/src/constants/suite/experimental.ts
peter-sanderson: I like `isDisabled` approach

peter-sanderson: fixed

packages/suite/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@trezor/suite",
-    "suiteVersion": "24.8.0",
+    "suiteVersion": "0.0.0",
```
peter-sanderson: **DO NOT FORGET ABOUT THIS** :exclamation:

packages/suite-desktop-core/src/modules/auto-updater.ts
```diff
@@ -94,6 +96,10 @@ export const init: Module = ({ mainWindow, store }) => {
 
         // Reset manual check flag
         isManualCheck = false;
+
+        if (isAutomaticUpdateEnabled) {
+            autoUpdater.downloadUpdate();
```
peter-sanderson: This is the place where magic happens. :point_up:

packages/suite-desktop-core/src/modules/auto-updater.ts
```diff
@@ -236,8 +242,27 @@ export const init: Module = ({ mainWindow, store }) => {
...
+        isAutomaticUpdateEnabled = value;
+
+        if (!isAutomaticUpdateEnabled) {
+            // In case
+            //      1) the auto-update was enabled
+            //      2) update is probably already downloaded
+            //      3) user wants do disable auto-update and PREVENT the downloaded update from installing
+            //
+            // We have to disable auto update so it won't get installed.
+            autoUpdater.autoInstallOnAppQuit = false;
```
peter-sanderson, line 266: This is important :point_up: . This is the mechanic how to prevent the suite from updating when it already downloaded but user wants to disable silent update!

packages/suite/src/actions/suite/desktopUpdateActions.ts
```diff
@@ -81,7 +82,8 @@ export const ready = (info: UpdateInfo) => (dispatch: Dispatch, getState: GetSta
 };
 
 export const installUpdate =
-    (shouldInstallOnQuit?: boolean) => (_: Dispatch, getState: GetState) => {
```
peter-sanderson, line 84: Boolean unnamed props makes code hard to read. Also I am not really happy about default values here. Especially when there is inverted (negated) logic.

packages/suite/src/views/settings/SettingsGeneral/AutomaticUpdate.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    const handleChange = () => {
+        const newValue = !isAutomaticUpdateEnabled;
+
+        desktopApi.setAutomaticUpdateEnabled(newValue);
+    };
+
+    return (
+        <SettingsSectionItem anchorId={SettingsAnchor.Analytics}>
```
komret: Incorrect `anchorId`.

packages/suite-desktop-core/src/modules/auto-updater.ts
```diff
@@ -236,8 +242,27 @@ export const init: Module = ({ mainWindow, store }) => {
...
     });
 
+    ipcMain.on('update/set-automatic-update-enabled', (_, value = true) => {
+        logger.info(SERVICE_NAME, `set-automatic-update-enabled: ${value ? 'true' : 'false'}`);
+
+        mainWindow.webContents.send('update/set-automatic-update-enabled', value);
+        const settings = store.getUpdateSettings();
+        store.setUpdateSettings({ ...settings, isAutomaticUpdateEnabled: value });
+        isAutomaticUpdateEnabled = value;
+
```
peter-sanderson, line 254: - [x] Todo: trigger version check + download right away when `isAutomaticUpdateEnabled` is set to `true`

packages/suite/src/views/settings/SettingsGeneral/AutomaticUpdate.tsx
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13446/commits/d0e723418fb64f2b2775b2013ab196857a1f6214

komret:
TODO: after manually closing and opening the app and installing the new version, notification should appear (currently it does not).

komret (approved):

PR #13448, opened 2024-07-22
feat(utils): scheduleAction with attempt failure handler

szymonlesisz:
## Description
`scheduleAction` with attempt failure handler, 

if `attemptFailureHandler` is set and it return/throw Error then break the attempts loop. 
Useful if you want to interrupt attempt loop  when attempt result is an unexpected/unfixable Error, like other than aborted by timeout/signal


marekrjpolak (approved):
Should work as intended. I've just made it a bit simpler, feel free to ignore the fixup. https://github.com/trezor/trezor-suite/pull/13448/commits/44f26e045beb334d87dd8915dcbb223a2bb962a6

PR #13450, opened 2024-07-22
Fix/tokens dropdown

AdamSchinzel:
## Description

Fixed label for hide/unhide token and added icons for send and receive actions.

## Screenshots:
Screenshot 2024-07-22 at 1 38 22 PM

### DIFF
--- a/packages/suite/src/views/wallet/tokens/common/TokenList.tsx
+++ b/packages/suite/src/views/wallet/tokens/common/TokenList.tsx
@@ -264,6 +264,7 @@ export const TokenList = ({
                                             options: [
                                                 {
                                                     label: <Translation id="TR_NAV_SEND" />,
+                                                    icon: 'SEND',
                                                     onClick: () => {
                                                         goToWithAnalytics('wallet-send', {
                                                             params: {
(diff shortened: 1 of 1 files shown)


packages/suite/src/views/wallet/tokens/common/TokenList.tsx
```diff
@@ -291,14 +293,23 @@ export const TokenList = ({
...
+                                                        <Translation
+                                                            id={
+                                                                tokenStatusType ===
+                                                                TokenManagementAction.SHOW
+                                                                    ? 'TR_UNHIDE_TOKEN'
+                                                                    : 'TR_HIDE_TOKEN'
+                                                            }
+                                                        />
+                                                    ),
                                                     icon: 'EYE_CLOSED',
```
tomasklim, line 306: how about eye open in case of `TR_UNHIDE_TOKEN`

AdamSchinzel: True but we have just closed or slashed.

Also I don't think this icon matches the button action since it says unhide but the icon is more like hide...
Screenshot 2024-07-23 at 4 58 42 PM

cc @Hermez-cz

tomasklim (approved):

PR #13452, opened 2024-07-22
fix(suite): zero value phishing tokens

AdamSchinzel:
## Description

Improve zero value phishing detection for tokens. Now transaction gets blurred if least one token has zero amount (not every token like before).

### DIFF
--- a/suite-common/wallet-utils/src/__fixtures__/antiFraud.ts
+++ b/suite-common/wallet-utils/src/__fixtures__/antiFraud.ts
@@ -104,6 +104,30 @@ export const getIsFakeTokenPhishingFixtures = [
         result: true,
     },
     {
+        testName: 'only zero-value fake tokens tx',
+        transaction: {
+            symbol: 'matic',
+            amount: '0',
+            tokens: [
+                { standard: 'ERC20', contract: '0xC', amount: '0' },
+                { standard: 'ERC20', contract: '0xD', amount: '0' },
+            ],
+        } as WalletAccountTransaction,
+        tokenDefinitions: {
+            coin: {
+                error: false,
+                isLoading: false,
+                data: ['0xa', '0xb'],
+            },
+            nft: {
+                error: false,
+                isLoading: false,
+                data: ['0xn', '0xf'],
+            },
+        } as TokenDefinitions,
+        result: true,
+    },
+    {
         testName: 'one fake, one legit token tx',
         transaction: {
             symbol: 'matic',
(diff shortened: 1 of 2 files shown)


suite-common/wallet-utils/src/antiFraud.ts
```diff
@@ -10,7 +10,7 @@ import { isNftTokenTransfer } from './transactionUtils';
 export const getIsZeroValuePhishing = (transaction: WalletAccountTransaction) =>
     new BigNumber(transaction.amount).isEqualTo(0) &&
     D.isNotEmpty(transaction.tokens) &&
-    transaction.tokens.every(token => new BigNumber(token.amount).isEqualTo(0));
+    transaction.tokens.some(token => new BigNumber(token.amount).isEqualTo(0));
 
 export const getIsFakeTokenPhishing = (
```
tomasklim, line 15: Just skip tokens with 0 balance here.
When 0, do not evaluate it as legit token

tomasklim (changes requested):

AdamSchinzel:
Maybe we should think it more though, closing it for now.

suite-common/wallet-utils/src/antiFraud.ts
```diff
@@ -8,9 +8,7 @@ import { getNetworkType } from '@suite-common/wallet-config';
 import { isNftTokenTransfer } from './transactionUtils';
 
 export const getIsZeroValuePhishing = (transaction: WalletAccountTransaction) =>
-    new BigNumber(transaction.amount).isEqualTo(0) &&
-    D.isNotEmpty(transaction.tokens) &&
-    transaction.tokens.every(token => new BigNumber(token.amount).isEqualTo(0));
+    new BigNumber(transaction.amount).isEqualTo(0) && D.isEmpty(transaction.tokens);
```
tomasklim: non needed change which breaks tests and also tx history
Screenshot 2024-07-23 at 13 48 43

suite-common/wallet-utils/src/antiFraud.ts
```diff
@@ -30,7 +28,8 @@ export const getIsFakeTokenPhishing = (
         return (
             (isTokenDefinitionKnown(definitions, transaction.symbol, tokenTx.contract) ||
                 isShown) &&
-            !isHidden
+            !isHidden &&
+            !new BigNumber(tokenTx.amount).isEqualTo(0)
```
tomasklim: ok
Screenshot 2024-07-23 at 14 03 57

but i would not complicate the return of this so i would just put condition right under `tokens.some...` 
```
 !transaction.tokens.some(tokenTx => {
        if (new BigNumber(tokenTx.amount).isEqualTo(0)) {
            return false;
        }

        const isNftTx = isNftTokenTransfer(tokenTx);
```

tomasklim (changes requested):
tests missing

suite-common/wallet-utils/src/__fixtures__/antiFraud.ts
```diff
@@ -127,6 +127,30 @@ export const getIsFakeTokenPhishingFixtures = [
         } as TokenDefinitions,
         result: false,
     },
+    {
+        testName: 'one fake, one zero-value legit token tx',
+        transaction: {
+            symbol: 'matic',
+            amount: '0',
+            tokens: [
+                { standard: 'ERC20', contract: '0xA', anmount: '1' },
```
tomasklim:
```suggestion
                { standard: 'ERC20', contract: '0xA', amount: '1' },
```

suite-common/wallet-utils/src/__fixtures__/antiFraud.ts
```diff
@@ -127,6 +127,30 @@ export const getIsFakeTokenPhishingFixtures = [
...
+                data: ['0xb'],
+            },
+            nft: {
+                error: false,
+                isLoading: false,
+                data: ['0xn', '0xf'],
+            },
+        } as TokenDefinitions,
+        result: true,
+    },
```
tomasklim, line 177: can you also provide other scenarios, such as:
- only zero legit tokens
- only zero non legit tokens
- legit balance 1, non-legit balance 0

so that it is covered and we do not need to return to it

maybe also one case with nft so we know that it does not break anything in case of non-ERC20

tomasklim (changes requested):

suite-common/wallet-utils/src/__fixtures__/antiFraud.ts
tomasklim: please type fixtures so that it does not occur again

not one by one but single record for whole fixture array, I accept `any` or `Partial` for some objects, but would be nice to avoid things like `anmount` 😀

suite-common/wallet-utils/src/__fixtures__/antiFraud.ts
AdamSchinzel: b590ae49034370ce7cdd33d150182c26138bd139

tomasklim (approved):

PR #13455, opened 2024-07-22
ci(script): check authenticity data branch name based on content

karliatto:
## Description

Use the content of the changes instead of the commit hash to create new branch with changes. If we use the commit hash it will always different since it is not based only on the content. This way if there is a PR open with same changes the script will fail when pushing the new branch (because it already exists and it not `-force`) and it will not create new PR everyday we do not merge the PR.

## Related Issue
Related to https://github.com/trezor/trezor-suite/pull/13420#issuecomment-2239496488


mroz22 (approved):
makes sense

PR #13458, opened 2024-07-22
fix(suite): on copy device display

### LINKED ISSUE #12988: Extra space on devices without address chunkification
**Describe the bug**
On old FWs chunkification is not supported, so we show address without it. 

I am not so smart user so I do not check address with device and I copy it right away from receive address modal. However, there are whitespaces :( that does not happen in case of chunkified address 

Screenshot 2024-06-19 at 17 14 09

AdamSchinzel:
## Description

Share `handleOnCopy` inside address display (chunked and paginated variants) so there are no extra whitespaces etc.

## Related Issue

Resolve #12988

### DIFF
--- a/packages/suite/src/components/suite/DeviceDisplay/DisplayChunks.tsx
+++ b/packages/suite/src/components/suite/DeviceDisplay/DisplayChunks.tsx
@@ -1,7 +1,7 @@
-import { ClipboardEvent } from 'react';
 import styled from 'styled-components';
 import { DeviceDisplayText } from './DeviceDisplayText';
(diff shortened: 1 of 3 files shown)


peter-sanderson:
Are we sure we want this? I would rather disable copy-pasting at all and force user to do device check first.

peter-sanderson:
@Hermez-cz :point_up: WDYT?

AdamSchinzel:
I have the same opinion, but the last time we discussed it, the product team wanted it that way.

tomasklim (approved):

PR #13461, opened 2024-07-23
feat(suite): Add shortcut for settings

jvaclavik:
## Related Issue

Resolve

### DIFF
--- a/packages/suite/src/components/suite/layouts/SuiteLayout/useAppShortcuts.tsx
+++ b/packages/suite/src/components/suite/layouts/SuiteLayout/useAppShortcuts.tsx
@@ -3,23 +3,30 @@ import { useEvent } from 'react-use';
 import { closeModalApp, goto } from 'src/actions/suite/routerActions';
 import { addWalletThunk } from 'src/actions/wallet/addWalletThunk';
 import { useDispatch, useSelector } from 'src/hooks/suite';
+import { KEYBOARD_CODE } from '@trezor/components';
 
 export const useAppShortcuts = () => {
     const selectedDevice = useSelector(selectDevice);
     const dispatch = useDispatch();
 
     useEvent('keydown', e => {
-        const { altKey } = e;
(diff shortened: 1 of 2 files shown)


packages/suite/src/components/suite/layouts/SuiteLayout/useAppShortcuts.tsx
```diff
@@ -29,5 +29,11 @@ export const useAppShortcuts = () => {
             );
             e.preventDefault();
         }
+
+        // press CMD + , to show Settings
+        if (metaKey && e.code === 'Comma' && selectedDevice) {
```
peter-sanderson: future idea: As keybinds are gonna grow, we may consider refactor those IFs into some map.

peter-sanderson: nit: `selectedDevice !== undefined`?

peter-sanderson (approved):

packages/suite/src/components/suite/layouts/SuiteLayout/useAppShortcuts.tsx
jvaclavik: > future idea: As keybinds are gonna grow, we may consider refactor those IFs into some map.

Yes, totally. We are going to improve shortcuts across the app, so this should be part of it.

packages/suite/src/components/suite/layouts/SuiteLayout/useAppShortcuts.tsx
```diff
@@ -29,5 +30,11 @@ export const useAppShortcuts = () => {
             );
             e.preventDefault();
         }
+
+        // press CMD + , to show Settings
+        if (metaKey && e.code === 'Comma' && isDeviceSelected) {
```
tomasklim: we have file for storing "code" `packages/components/src/constants/keyboardEvents.ts`

PR #13462, opened 2024-07-23
Fiat rates missing for transactions except first 5 in Portfolio tracker or view-only

matejkriz:
## Description

Steps to reproduce:
1. load portfolio tracker account or view only device with more than 5 transactions in history
2. restart the app
3. check transaction history of remembered account

You will see fiat rates for first 5 transactions only and zeros for the rest. Switching fiat currency fixes it until the next restart.

solution:
always check missing fiat rates after fetching transactions

## Screenshots:
image

## Notes for QA
- Check also receiving new transaction on that account. 
- Check switching fiat currency

### DIFF
--- a/suite-common/wallet-core/src/transactions/transactionsThunks.ts
+++ b/suite-common/wallet-core/src/transactions/transactionsThunks.ts
@@ -247,7 +247,6 @@ export const fetchTransactionsPageThunk = createThunk(
         { accountKey, page, perPage, forceRefetch }: FetchTransactionsPageThunkParams,
         { dispatch, getState },
     ) => {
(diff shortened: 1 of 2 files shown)


matejkriz:
This PR hasn't been merged, this was the preferred solution https://github.com/trezor/trezor-suite/pull/13463, so this serves only as an issue now. Sorry for the mess.

matejkriz:
will be released in 24.7.3

STew790:
**QA OK**
Fiat rates in transactions are ok now. I tried to change currency and that is ok as well.

**Info:**
24.7.3 https://github.com/trezor/trezor-suite/commit/1aefc81ad4e6850407907020bcc8058cfe93355a

PR #13463, opened 2024-07-23
fix(suite-native): always check missing fiat rates after fetching transactions via middleware

matejkriz:
## Description

This is an alternative solution of https://github.com/trezor/trezor-suite/pull/13462

It does the same, but checking for missing fiat rates is dispatched from middleware so it applies for desktop too.

## Related Issue

Resolve #13462


AdamSchinzel (approved):

vytick (approved):
🚀

PR #13464, opened 2024-07-23
fix(suite): update select input in buy and sell form to be searchable

### LINKED ISSUE #11462: Empower/Trade - unable to type country in the dropdown menu

seibei-iguchi:
Fix an attribute that has been possibly changed in a merge or rebase.


jvaclavik (approved):
LGTM!

PR #13465, opened 2024-07-23
Mobile README update

matejkriz:
## Description

Attempt to clarify first steps with suite-native by feedback from @komret. It wasn't clear that previously mentioned NPM scripts was meant to be called from `suite-native/app` folder. 

We have aliases in global package.json so I believe it's easier to use them.

BREAKING CHANGE: `yarn p` no longer call `prebuild:clean` but only `prebuild`

### DIFF
--- a/suite-native/app/README.md
+++ b/suite-native/app/README.md
@@ -8,41 +8,53 @@ Generally it's recommended to follow official [React Native environment setup](h
 
 1. [Android Guide](https://reactnative.dev/docs/set-up-your-environment?os=macos&platform=android)
 2. [iOS Guide](https://reactnative.dev/docs/set-up-your-environment?os=macos&platform=ios)
-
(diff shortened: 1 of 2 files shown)


vytick (approved):
🚢 🇮🇹

suite-native/app/README.md
```diff
@@ -8,41 +8,61 @@ Generally it's recommended to follow official [React Native environment setup](h
 
 1. [Android Guide](https://reactnative.dev/docs/set-up-your-environment?os=macos&platform=android)
 2. [iOS Guide](https://reactnative.dev/docs/set-up-your-environment?os=macos&platform=ios)
```
komret, line 10: nit: should not be an ordered list

komret (approved):

suite-native/app/README.md
matejkriz: 👍 maybe next time...

PR #13466, opened 2024-07-23
Update libs related to Connect

karliatto:
## Related Issue

Resolve


mroz22:
please lets do it one lib - one commit. including changes in yarn.lock 🙏 makes it easier to review.

karliatto:
> please lets do it one lib - one commit. including changes in yarn.lock 🙏 makes it easier to review.

Yes, I am doing like that and one by one letting all the tests run in CI after updating dependency to see if any breaks something.

karliatto:
> * @SocketSecurity ignore npm/bundle-name@4.1.0

It is dependency of a dependency so even it is a trivial package we would need to update it upstream in order to update it. So @SocketSecurity ignore npm/bundle-name@4.1.0

karliatto:
@SocketSecurity ignore npm/bundle-name@4.1.0

packages/address-validator/package.json
```diff
@@ -20,6 +20,6 @@
     },
     "devDependencies": {
         "jest-environment-jsdom": "29.7.0",
-        "rimraf": "^5.0.5"
+        "rimraf": "^6.0.1"
```
mroz22, line 23: todo later: maybe packages that are already defined in root package json shoudln't be present in particular packages package.jsons? cc @Nodonisko

mroz22 (approved):

packages/address-validator/package.json
komret: They shouldn't.

mroz22: ok, @karliatto please just rebase and let's merge it now. Removing root defined packages should be done as a followup.

PR #13467, opened 2024-07-23
chore(connect): add comments to UnavailableCapability type

mroz22:


packages/connect/src/types/device.ts
```diff
@@ -18,6 +18,12 @@ export type DeviceFirmwareStatus =
     | 'custom'
     | 'none';
 
+/**
+ * - no-capability - device firmware does not support this feature based on features.capabilities
+ * - no-support - device does not support this coin based on CoinInfo.support (https://github.com/trezor/trezor-common/blob/master/defs/support.json)
+ * - update-required - device needs to update firmware to support this feature
+ * - trezor-connect-outdated - TODO: is it still relevant? isn't it some remnant from times when there was dynamically fetched config.json?
```
szymonlesisz, line 25: this is for older releases case (major version)

used at least once in the past (branch/version v7) - disabling all methods for newest firmware (after passphrase redesign)

https://github.com/trezor/connect/commit/fe04d4b515d1cee80eee9d89769b01a9aa45ba7b

szymonlesisz: passphrase was redesigned in v8 and since we didnt want to backport it to v7 and below we just block those methods and kinda force 3rd party implementators to update connect package

PR #13472, opened 2024-07-23
Fix `<LoadingContent />` animation

marekrjpolak:
## Description

Sometimes when `<LoadingContent />` component is first rendered in active state and inactivated immediately after that, it creates a visual impression that something was just loading (even if it wasn't). Now the component always start inactive so there's time to revert the animation without anyone noticing.

Cc: @komret, who found the issue


komret:
It seems to me like a hacky solution and the problem is negligible so I'd rather keep it as is.

PR #13479, opened 2024-07-23
fix(connect-webextension): multiple content script extension conflicts

martykan:
## Description

Resolve the issue with multiple content scripts being injected at once by adding a URL paramter to popup.html, specifying the extension ID. 
There are also version and env URL parameters for futureproofing. 

The content script & popup.js then checks if their ID matches the extension ID in the URL parameter.

### DIFF
--- a/packages/connect-popup/src/index.tsx
+++ b/packages/connect-popup/src/index.tsx
@@ -360,6 +360,13 @@ const handleLogMessage = (event: MessageEvent<IFrameLogRequest>) => {
 
 const handleContentScriptLoaded = (data: PopupContentScriptLoaded) => {
     log.debug('content-script loaded', data.payload);
+    // Check if extension ID matches the popup URL
+    const urlParams = new URLSearchParams(window.location.search);
+    const targetExtensionId = urlParams.get('extension-id');
+    if (targetExtensionId && targetExtensionId !== data.payload.id) {
+        // Wrong extension ID, ignore it
+        return;
+    }
     setState({
         settings: {
             ...getState().settings!,
--- a/packages/connect-web/src/popup/index.ts
+++ b/packages/connect-web/src/popup/index.ts
@@ -5,6 +5,7 @@ import { createDeferred, Deferred, scheduleAction } from '@trezor/utils';
 import { POPUP, IFRAME, UI, CoreEventMessage, IFrameLoaded } from '@trezor/connect/src/events';
 import type { ConnectSettings } from '@trezor/connect/src/types';
 import { getOrigin } from '@trezor/connect/src/utils/urlUtils';
+import { CONTENT_SCRIPT_VERSION, VERSION } from '@trezor/connect/src/data/version';
 import { showPopupRequest } from './showPopupRequest';
 import { Log } from '@trezor/connect/src/utils/debug';
 
@@ -180,7 +181,8 @@ export class PopupManager extends EventEmitter {
     open() {
         const src = this.settings.popupSrc;
         this.popupPromise = createDeferred(POPUP.LOADED);
-        this.openWrapper(src);
+        const url = this.buildPopupUrl(src);
+        this.openWrapper(url);
 
         if (this.settings.useCoreInPopup) {
             // Timeout not used in Core mode, we can't run showPopupRequest with no DOM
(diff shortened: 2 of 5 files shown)


packages/connect-web/src/popup/index.ts
```diff
@@ -5,6 +5,7 @@ import { createDeferred, Deferred, scheduleAction } from '@trezor/utils';
 import { POPUP, IFRAME, UI, CoreEventMessage, IFrameLoaded } from '@trezor/connect/src/events';
 import type { ConnectSettings } from '@trezor/connect/src/types';
 import { getOrigin } from '@trezor/connect/src/utils/urlUtils';
+import { VERSION } from '@trezor/connect/src/data/version';
```
mroz22: just brainstorming. this is core VERSION. maybe we should introduce separate  version field for npm packages? which version we actually want to share?

packages/connect-webextension/src/contentScript.ts
```diff
@@ -1,81 +1,92 @@
...
-
-/*
- * Passing messages from popup to service worker
- */
-window.addEventListener('message', event => {
-    if (
-        event.data?.channel?.here === '@trezor/connect-webextension' ||
-        event.data?.type === POPUP.CONTENT_SCRIPT_LOADED
-    ) {
+function trezorContentScript() {
```
mroz22, line 5: don't we want to introduce in some separate versioning of content script?

lets say do something like:

```
CONTENT_SCRIPT_VERSION = 1
```

then we pass another param in query string such as SUPPORTED_CONTENT_SCRIPT_VERSION = 2 and this way we are able to detect version incompatibility between these elements and at least print an error to console?

martykan: Good idea, I was also thinking about something like this, we don't necessarily need it for this change since it's still compatible, but it might be useful in the future. 
This should only happen in situations when the extension authors updates the NPM package, has the content script copied manually and doesn't update it.

packages/connect-web/src/popup/index.ts
martykan: We keep all the main packages versions in sync so it shouldn't be an issue right?

packages/connect-popup/src/index.tsx
```diff
@@ -360,6 +360,18 @@ const handleLogMessage = (event: MessageEvent<IFrameLogRequest>) => {
 
 const handleContentScriptLoaded = (data: PopupContentScriptLoaded) => {
     log.debug('content-script loaded', data.payload);
+    // Check if extension ID matches the popup URL
+    const urlParams = new URLSearchParams(window.location.search);
+    const targetExtensionId = urlParams.get('extension-id');
+    console.log('popup targetExtensionId', targetExtensionId, 'data.payload.id', data.payload.id);
```
mroz22: forgotten console.log maybe?

packages/connect-popup/src/index.tsx
```diff
@@ -360,6 +360,18 @@ const handleLogMessage = (event: MessageEvent<IFrameLogRequest>) => {
...
+    if (targetExtensionId && targetExtensionId !== data.payload.id) {
+        return;
+    }
+    if (data.payload.contentScriptVersion !== CONTENT_SCRIPT_VERSION) {
+        console.error(
+            'Content script version mismatch! Make sure to update content script to the latest version from @trezor/connect-webextension.',
+        );
```
mroz22: couldn't this warning be added also to the client side? I think it would be more visible there for 3rd party devs. something like this line maybe? https://github.com/trezor/trezor-suite/blob/develop/packages/connect-webextension/src/contentScript.ts#L36

martykan: Good catch, actually it's a bad idea to put it here since the production version can be different.

packages/connect-webextension/src/contentScript.ts
```diff
@@ -1,81 +1,103 @@
...
-        window.postMessage(
-            {
-                type: POPUP.CONTENT_SCRIPT_LOADED,
-                payload: { ...chrome.runtime.getManifest(), id: chrome.runtime.id },
-            },
-            window.location.origin,
+    const targetContentScriptVersion = urlParams.get('cs-ver');
+    if (targetContentScriptVersion !== CONTENT_SCRIPT_VERSION.toString()) {
+        console.error(
+            'Content script version mismatch! Make sure to update content script to the latest version from @trezor/connect-webextension.',
```
mroz22: hmm but here it still would appear inside popup dev console and I doubt that 3rd party devs check it. Doesn't it make sense to put it to their main area of focus (where npm package is installed)? doesn't this content-script-loading event reach it as well?

martykan: Currently no, but we can make it happen

mroz22 (approved):

PR #13481, opened 2024-07-23
WIP: Chore: add t3b1 support suite

karliatto:
## Description
This is adding T3B1 support to suite. It is not completed, there are missing parts:

To have full support we will need:
- [ ]  support of coins to include T3B1, @matejcik https://github.com/trezor/trezor-firmware/blob/main/common/defs/support.json
- [ ] Add firmware in packages/connect-common/files/firmware/t3b1
-

### DIFF
--- a/packages/suite/src/middlewares/suite/suiteMiddleware.ts
+++ b/packages/suite/src/middlewares/suite/suiteMiddleware.ts
@@ -71,6 +71,7 @@ const suite =
         }
 
         if (deviceActions.connectDevice.match(action)) {
+            // TODO: should this apply also for DeviceModelInternal.T3B1 ???
             const isT3T1 =
                 action.payload?.device?.features?.internal_model === DeviceModelInternal.T3T1;
             const isT3T1DashboardPromoBannerActive =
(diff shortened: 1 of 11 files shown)


packages/suite/src/middlewares/suite/suiteMiddleware.ts
```diff
@@ -71,6 +71,7 @@ const suite =
         }
 
         if (deviceActions.connectDevice.match(action)) {
+            // TODO: should this apply also for DeviceModelInternal.T3B1 ???
```
karliatto, line 74: I would expect T3T1 to have similar feature set that T3B1 but I am not sure.

MiroslavProchazka: You mean similar to T2B1, right?

mroz22 (changes requested):
carlos, could you please try to make these places type-safe? in the meaning that unless all internal models are defined, typecheck will catch?

komret (changes requested):
T3B1 should mimic T2B1, not T3T1. It is the same device as T2B1, but with a different chip.

mroz22:
@komret doesn't your team want to take this from @karliatto? it looks like all the changes are to be done on the suite level mostly.

MiroslavProchazka:
> @komret doesn't your team want to take this from @karliatto? it looks like all the changes are to be done on the suite level mostly.

Yes, we will take over. Komár is already working on it.

komret:
Closing in favour of https://github.com/trezor/trezor-suite/pull/13544

PR #13483, opened 2024-07-23
chore: no longer recommend firefox as a supported browser

### LINKED ISSUE #12486: "Your browser is not supported" modal/page improvements - Suite web
Currently "Your browser is not supported" modal/page is displayed before accessing Suite web or Connect popup on any browser except Chrome and Firefox.
However, all features should work as expected on almost all Chromium-based browsers. 

- [ ] Create a better conditions for supported (Chromium-based) browsers so the modal doesn't has to be displayed for all non-Chrome/Firefox browsers

With the deprecation of support of Firefox browsers due to the deprecation of Trezor Bridge, 
- [x] remove Firefox from the list of supported browsers

image

martykan:
## Description

Remove Firefox download from the unsupported browser page (due to lack of WebUSB support)

## Screenshot

Screenshot 2024-07-23 at 17 19 59

## Related Issue

Resolve #12486


packages/suite-data/src/browser-detection/index.ts
```diff
@@ -140,7 +124,7 @@ window.addEventListener('load', () => {
     const updateFirefox = getMainHtml({
```
komret, line 124: Shouldn't Firefox be removed from `supportedBrowsers` and this function deleted?

komret (changes requested):
Note that by linking the issue with the _resolves_ keyword, merging this PR will close the issue despite it listing more tasks to do.

martykan:
The other part of the issue about supporting other Chromium-based browser should already be solved for a while. 
I personally use Arc browser, Brave and latest MS Edge also works

packages/suite-data/src/browser-detection/index.ts
martykan: We're not making Firefox unsupported with this change yet. We're just not recommending it anymore

komret (approved):

Issue #13179, opened 2024-07-03
connect: webusb: sync broken - bridge started after webusb


mroz22:
similar to #12002 but different reason 

1. start application A using webusb 
2. start suite-desktop using bridge
3. application A remains using webusb conflicting with suite-desktop using bridge

what is the best solution here?
1. if using a low priviledged transport (webusb), on any error check if a higher priviledged transport (node-bridge) works and if yes,   start using it?

Hannsek:
Can it be switched unnoticeably for the user?

How about THP? Is the transport type irrelevant or will user has to pair it again when switched to another transport type?

STew790:
**QA NOK**
When I try to:

1. connect device to webusb suite
2. start Suite desktop with nodebridge
3. neither of the apps "take over" the device and I can see this:
![Image](https://github.com/user-attachments/assets/d56c6e38-1d16-44bf-bd4a-a39effd49442)

**Info:**
 - Suite version: web 25.3.1 (bc43a98c2c2ac6eab179670fb723e3ec41b4ff31)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 3440x1440
 - Device: 
 - Transport: BridgeTransport 3.0.0-bundled.25.3.0

marekrjpolak:
a) I can't reproduce it accurately, only some similar cases.
b) It's probably fine to have it on one side, but I'll try to improve the message at least ("used by another app" is better)
c) It definitely seems like a bug to have it stuck on both sides.
Cc: @STew790

evgenysl:
1. Trezor used in web Trezor Suite
2. Open desktop Suite

<img width="1567" alt="Image" src="https://github.com/user-attachments/assets/57238ad0-9451-4a8e-a404-6b638614be43" />

3. Device is connected on refresh

bosomt:
**QA OK**

- already deployed to production
- follow up issue/PR in progress

PR #13484, opened 2024-07-23
Kill bridge followup

martykan:
## Description

Several changes to continue the process of removing references to Trezor Bridge


packages/suite/src/components/suite/DeviceInvalidModeLayout.tsx
```diff
@@ -1,83 +0,0 @@
-import { ReactNode } from 'react';
```
mroz22, line 1: these files here were never used / duplicates?

packages/suite/src/views/suite/device-unknown/index.tsx
```diff
@@ -1,7 +0,0 @@
-// TODO: remove whole file, replaced by @suite-components/PrerequisitesGuide/components/DeviceUnknown
```
mroz22, line 1: ahh, I see, there is even a comment saying that

martykan: Yeah 3 years ago 😄

packages/suite/src/hooks/suite/useOpenSuiteDesktop.ts
```diff
@@ -0,0 +1,26 @@
...
+import { SUITE_BRIDGE_DEEPLINK, SUITE_URL } from '@trezor/urls';
+import { isWebUsb } from 'src/utils/suite/transport';
+import { useSelector } from 'src/hooks/suite';
+
+export const useOpenSuiteDesktop = () => {
+    const transport = useSelector(state => state.suite.transport);
+    const windowFocused = useWindowFocus();
+    const handleOpenSuite = () => {
+        if (isWebUsb(transport)) {
+            TrezorConnect.disableWebUSB();
```
mroz22, line 12: hmm I am brainstorming (not related to this code here). The problem: https://github.com/trezor/trezor-suite/issues/13179

We at least used to have manual solution - disable webusb button. It was not perfect and I think we should address it now. The right solution would be to disable webusb somewhere under the hood based on either bridge polling, or when a connect call ends with (specific) error.

mroz22 (approved):

PR #13488, opened 2024-07-24
Fix/native/view only persist immediatelly

### LINKED ISSUE #13485: View only wallet showing "Your wallet is empty" even though it isn't
**Describe the bug**
When you connect a device and enable a view only feature for it and for the wallet and kill the app and start it again, it shows as if the device is empty, even though it had some accounts.

**Info:**

-   OS: Android

**How to reproduce**
It is not really 100% reproducible but it is tied with timing and more easy to reproduce on BTC only FW 
(according to @matejkriz )

1. Connect your device
1. Wait for the discovery and enable view only wallet
1. Kill the app 
1. disconnect trezor
1. start the app again
1. App is showing device connected but says "Your wallet is empty"

**Expected behavior**
Your wallet with your accounts is shown/loaded

**Screenshots**

https://github.com/user-attachments/assets/0d9116fa-3a2b-41b3-ac14-0bdcd8601774

matejkriz:
## Description

Persistence of accounts, transactions and graph in suite-native depends on device.remember state, but redux-persist is not checking for changes in other reducers. This is a workaround to update redux-persist state.

This is replacing part of state with it's shallow copy so there are no actual changes but it triggers the recalculation of redux-persist state. It can be optimized to change some smaller portion of state or we can introduce some extra reducer just to handle this, but I believe it's harmless this way as it is triggered only on user action (toggling remember the wallet).

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13485

### DIFF
--- a/suite-common/wallet-core/src/accounts/accountsReducer.ts
+++ b/suite-common/wallet-core/src/accounts/accountsReducer.ts
@@ -14,6 +14,7 @@ import {
     selectDevice,
     selectHasOnlyPortfolioDevice,
 } from '../device/deviceReducer';
+import { deviceActions } from '../device/deviceActions';
 import { DiscoveryRootState, selectIsDeviceDiscoveryActive } from '../discovery/discoveryReducer';
 
 export type AccountsState = Account[];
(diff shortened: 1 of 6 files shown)


vytick (approved):
lgtm 🚀

PR #13492, opened 2024-07-24
chore(utils): catch leaking listeners in unit tests

mroz22:
this makes unit tests in the utils package fail whenever `MaxListenersExceededWarning` should be printed


marekrjpolak (approved):
I've fixed the `scheduleAction` util so it correctly cleans listeners in this particular failing unit test. Nevertheless, I'm thinking about even stricter testing conditions. For now this is great ofc.

I also fixed some `eslint` complaints and removed duplicate line in config.

mroz22 (commented):
approved 🦭

PR #13493, opened 2024-07-24
transport - fixes and more testing scenarios

mroz22:
# 🐰 🕳️ 

### Commits
- [ca388c5](https://github.com/trezor/trezor-suite/pull/13493/commits/ca388c5a9c54d9817a3632382db5f0732711a247) - to be able to use the already existing test for testing node-bridge too, we need to move the tests to transport-bridge package. The reason is that now these tests must import from both transport and transport-bridge package and transport-bridge is not dependency of transport package.
- [9d9e904](https://github.com/trezor/trezor-suite/pull/13493/commits/9d9e9045ba1b4090625d60782ba43ef746313238) - cleanup in sessions background.
- [1fe634e](https://github.com/trezor/trezor-suite/pull/13493/commits/1fe634e8fe80aff9df00f6f418ec6d847564c4d7) - tests got adapted so that we can run them in 4 different setups (old/new, hw/emu). useful for detecting differences between implementations (there should be none) but the fact is that there are some at the moment, all of them documented here https://github.com/trezor/trezor-suite/blob/transport-meow/packages/transport-bridge/e2e/expect.ts#L1-L2  
- [d37cc1c](https://github.com/trezor/trezor-suite/pull/13493/commits/d37cc1c63fa54cec97f60b008f432c39438d4cd3) - CI now runs both emulator variants of the transport-birdge tests. I also made some changes in setup to make the tests faster. I believe that running jest test runner in a dedicated container is not necessary since they may be easily run in the base container. It saves maybe 2 minutes or so.
- [606203b](https://github.com/trezor/trezor-suite/pull/13493/commits/606203b60e5c56dfbcac796fec2d48f2f1f1ab83) - this only renames events.test.ts to multi-client.test.ts because this test is indeed focusing on how TransportBridge class is behaving in this scenario. 
- [0eeb8e6](https://github.com/trezor/trezor-suite/pull/13493/commits/0eeb8e6e7302ef9d63eacec6900f21313864bc1a) - this commit adds a new test documenting incorrect behavior of node-bridge in cases of concurrent enumeration. I believe that this issue comes from node USB lib or possibly from the underlying os USB drivers.
- [582b299](https://github.com/trezor/trezor-suite/pull/13493/commits/582b2991ede6591267e1908da0626866b6284172) - solution for the issue documented in the prev commit. we simply run enumeration in sequence one by one with the help of sessions background.
- [7473ced](https://github.com/trezor/trezor-suite/pull/13493/commits/7473cedfed9b32ecdc81545026311dd0fe09bab2) - disabling unrelated test, issue created https://github.com/trezor/trezor-suite/issues/13762 
- [83873b8](https://github.com/trezor/trezor-suite/pull/13493/commits/83873b8a1305bd5e083472e04208676baef03d62) - /read endpoint should not parse body as JSON as added in #13592. 

### Follow-ups:
- finish webusb tester #13769
- finish/review PR by @szymonlesisz https://github.com/trezor/trezor-suite/pull/13570 
- add path masking as done in the old bridge, to get rid of differences documented in  https://github.com/trezor/trezor-suite/blob/transport-meow/packages/transport-bridge/e2e/expect.ts#L1-L2  
- there is still one possible racecondition similar to the one fixed in 582b299.  The same async code execution might be triggered in onconnect event here https://github.com/trezor/trezor-suite/blob/develop/packages/transport/src/api/usb.ts#L50

### Issues:
- #13550 was found while testing this.
- https://github.com/trezor/trezor-suite/issues/12163#issuecomment-2212429718 is addressed by this.


packages/transport-test/e2e/controller.ts
```diff
@@ -0,0 +1,173 @@
...
+              ? () => this.waitForNumberOfDevices(0)
+              : () => {
+                    console.log('todo: ping local emu');
+
+                    return Promise.resolve(null);
+                };
+    }
+
+    private waitForNumberOfDevices = (expected: number) => {
+        this.logger.log(`waiting for ${expected} device to be connected`);
```
szymonlesisz: this.logger.log(
            `${env.USE_HW ? '[MANUAL ACTION REQUIRED]' : ''} waiting for ${expected} device to be connected`,
        );

packages/transport-test/e2e/tests/bridge.test.ts
```diff
@@ -93,35 +84,36 @@ describe('bridge', () => {
...
         const receiveResponse = await bridge.receive({ session }).promise;
+
         expect(receiveResponse).toMatchObject({
             success: true,
             payload: {
                 type: 'Failure',
                 message: {
                     code: 'Failure_ActionCancelled',
-                    message: 'Cancelled',
+                    message: USE_HW ? 'Action cancelled by user' : 'Cancelled',
```
szymonlesisz: this message is related to FW version, `Action cancelled by user'` is only a fallback added by DeviceCommands.

i got `Cancelled` message here while testing.

i guess testing `code: 'Failure_ActionCancelled'` should be enough

packages/transport-test/package.json
```diff
@@ -0,0 +1,27 @@
+{
```
szymonlesisz, line 1: some brief info in readme about the setup and how to run it would be nice

szymonlesisz (commented):
all 4 tests are passing on my machine

packages/transport-test/e2e/tests/bridge.test.ts
mroz22: right. it could be difference between T1 and TT.

szymonlesisz (approved):

Issue #13271, opened 2024-07-10
Implement fetching language blobs from internet


matejkriz:
[getLanguage](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/data/getLanguage.ts/#L3) Connect method is not working in suite-native.

We have [native version of `httpRequest`](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/utils/assets.native.ts/#L3) and our [native `getAssetByUrl` method](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/utils/assetUtils.ts/#L14) is only looking for specific local files, not handling HTTP requests at all.

I guess we should fix `httpRequest` method to handle actual HTTP requests too.

followup of https://github.com/trezor/trezor-suite/pull/13269

mroz22:
does  @trezor/suite-mobile team want to fix it or you are leaving it to us?

mroz22:
recent change related to this https://github.com/trezor/trezor-suite/pull/13495/files#diff-d46bdd3934851fa3c76c8f09857a15423c43d08b6563808b35f761e56cd17fc8R16

matejkriz:
@matejkriz found out if it's still needed or already fixed

matejkriz:
no issue at all

PR #13495, opened 2024-07-24
feat: Implement firmwareVersionCheck connect method to verify firmware revision

peter-sanderson:
Adds a check for firmware revision.

- If analytics are ON, we send this error to Sentry: 
image

- Check can be turned off the same way as Authenticity Check for newer devices
image


packages/connect/src/api/checkFirmwareAuthenticity.ts
```diff
@@ -58,7 +58,6 @@ export default class CheckFirmwareAuthenticity extends AbstractMethod<
                 });
 
             const { message } = result;
-            // const message = { hash: 'yolo' };
```
peter-sanderson, line 61: image

packages/connect-common/files/firmware/t1b1/releases.json
```diff
@@ -9,6 +9,7 @@
         "url_bitcoinonly": "firmware/t1b1/trezor-t1b1-1.12.1-bitcoinonly.bin",
         "fingerprint": "3c694191f5b66a65cb5bb209adbf113cb40209e644b77162ba996bb7ee8f382b",
         "fingerprint_bitcoinonly": "985fb6a8c87f7547fb810f6c4a8331ebf19c677445810358778eb21eca78a181",
+        "firmware_revision": "1eb0eb9d91b092e571aac63db4ebff2a07fd8a1f",
```
peter-sanderson: - [ ] TODO: DO NOT MERGE THIS

This has separate PR

packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -0,0 +1,71 @@
...
+    deviceVersion,
+    internalModel,
+    deviceRevision,
+    expectedRevision,
+}: CheckFirmwareRevisionParams): Promise<FirmwareRevisionCheckResult> => {
+    if (internalModel !== DeviceModelInternal.T2T1 && internalModel !== DeviceModelInternal.T1B1) {
+        return { success: true, error: null }; // For newer devices we are covered by secure element anyway
+    }
+
+    deviceRevision = 'yolo'; // TODO: remove this!!!
```
peter-sanderson: - [x] Dont forget remove this

packages/suite/src/views/settings/SettingsDevice/SettingsDevice.tsx
```diff
@@ -85,10 +87,8 @@ export const SettingsDevice = () => {
 
     const deviceModelInternal = device.features.internal_model;
 
-    const supportsDeviceAuthentication = [
-        DeviceModelInternal.T2B1,
-        DeviceModelInternal.T3T1,
```
peter-sanderson, line 91: The `T3B1` is missing here (most likely by error). => Refactored to use the type-safe map so we know to add this param or new device next time.

packages/connect/src/types/device.ts
```diff
@@ -51,6 +60,11 @@ export type KnownDevice = {
     features: PROTO.Features;
     unavailableCapabilities: UnavailableCapabilities;
     availableTranslations: string[];
+    authenticityChecks: {
+        firmwareRevision: FirmwareRevisionCheckResult;
+        // Maybe add FirmwareHashCheck result here?
+        // Maybe add AuthenticityCheck result here?
+    };
```
mroz22: yes, it makes sense to me. define logic for firmwareHashCheck and authenticityCheck similarly to firmwareRevision check so that it is available also internally in connect(not only on API), and save its result value with device object. 

searching for volunteers 🧑‍🏭

packages/suite/src/constants/suite/device.ts
```diff
@@ -44,3 +44,11 @@ export const CHARACTER_OFFSET_FOR_NEXT_ARROW: Record<DeviceModelInternal, number
...
     [DeviceModelInternal.T3T1]: 4,
 };
+
+export const SUPPORTS_DEVICE_AUTH: Record<DeviceModelInternal, boolean> = {
+    [DeviceModelInternal.T1B1]: false,
+    [DeviceModelInternal.T2T1]: false,
+    [DeviceModelInternal.T2B1]: true,
+    [DeviceModelInternal.T3B1]: true,
+    [DeviceModelInternal.T3T1]: true,
+};
```
mroz22: nitpick: just maybe naming could be better. isn't `SUPPORTS_DEVICE_AUTH`. too generic? meaning that there might be a device that supports firmware revision check but doesn't not support firmware hash checek.

packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    if (expectedRevision === undefined) {
+        if (deviceVersion.length !== 3) {
+            return failFirmwareRevisionCheck('firmware-version-unknown');
+        }
+
+        try {
+            const onlineRelease = await getOnlineReleaseMetadata({ deviceVersion, internalModel });
+
+            if (onlineRelease === undefined || onlineRelease.firmware_revision === undefined) {
```
mroz22: nitpick: maybe `if (!onlineRelease?.firmware_revision)`

packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    deviceRevision = 'yolo'; // TODO: remove this!!!
+
+    if (expectedRevision === undefined) {
+        if (deviceVersion.length !== 3) {
+            return failFirmwareRevisionCheck('firmware-version-unknown');
+        }
+
+        try {
+            const onlineRelease = await getOnlineReleaseMetadata({ deviceVersion, internalModel });
```
mroz22: sooo, there is no local first variant?

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -69,6 +71,21 @@ export const Preloader = ({ children }: PreloaderProps) => {
...
+        selectedDevice?.features &&
+        selectedDevice.connected === true &&
+        !isFirmwareRevisionCheckDisabled
+    ) {
+        const failedChecks = Object.values(selectedDevice.authenticityChecks).filter(
+            check => check.success === false,
+        );
+
+        if (failedChecks.length > 0) {
+            return <DeviceCompromised />;
```
peter-sanderson, line 87: - [x] Todo: send to Sentry/Analytics that this happened (check with Matěj, consult with Device Auth Check code how it does it)

packages/suite/src/components/suite/banners/SuiteBanners/SuiteBanners.tsx
MiroslavProchazka, line 1: We need to review those priorities against banners from config.v1.json, so they make sense altogether. Increasing the limit of priority above 100 is also on the table.

packages/connect/src/device/checkFirmwareRevision.ts
peter-sanderson: Yes there is. It is provided into the function through params from here:

image

packages/suite/src/constants/suite/device.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/8ff596a3791bd5e8ea78836d9ad10fed3135bd79

packages/connect/src/device/checkFirmwareRevision.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/995f935c1a7bc3f779f2005997d28bfaa3a6e919

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -69,6 +71,24 @@ export const Preloader = ({ children }: PreloaderProps) => {
         return <InitialLoading timeout={90} />;
     }
 
+    if (
+        router.route?.app !== 'settings' &&
+        selectedDevice?.features &&
+        selectedDevice.connected === true &&
+        !isFirmwareRevisionCheckDisabled
+    ) {
+        console.log('selectedDevice.authenticityChecks', selectedDevice.authenticityChecks);
```
szymonlesisz: forgotten console.log

packages/connect/src/types/device.ts
```diff
@@ -34,6 +34,20 @@ export enum FirmwareType {
...
+    | {
+          success: true;
+          error: null;
+      }
+    | {
+          success: false;
+          error: 'revision-mismatch' | 'firmware-version-unknown' | 'firmware-version-unknown';
+      }
+    | {
+          success: null;
```
szymonlesisz: i would not step out from the convention `success: boolean` that we use everywhere else.

im not a big fan of using `null` as a meaningful value and distinguishing between `null` and `undefined` because after a while nobody remembers why it's done like that

i think this could be be done like:

```
{
   success: false,
   error: 'cannot-perform-check-offline'
}
```

and evaluated in SuiteBanners and suite Preloader using specific error literal

```
} else if (
    !isFirmwareRevisionCheckDisabled &&
    device?.authenticityChecks.firmwareRevision?.error === 'cannot-perform-check-offline' 
) {

```

packages/connect/src/device/Device.ts
```diff
@@ -532,6 +538,16 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
         this._updateFeatures(message);
 
+        if (
+            // The check was not yet performed
+            this.authenticityChecks.firmwareRevision === null ||
+            // The check was performed, but outcome cannot be surely determined (Suite is offline)
+            // -> recheck on every getFeatures() until the result is known
+            this.authenticityChecks.firmwareRevision.success === null
+        ) {
+            await this.checkFirmwareRevision();
```
mroz22, line 552: note: what happens if we run this on some very slow internet? maybe there should be time limit.

packages/connect/src/device/checkFirmwareRevision.ts
mroz22: optimiziation idea: 
maybe we could pull loading of the latest list of releases somewhere higher so that it doesn't block device features update 🤔

packages/connect/src/types/firmware.ts
```diff
@@ -17,6 +17,7 @@ export type FirmwareRelease = {
     fingerprint: string;
     changelog: string | string[];
     changelog_bitcoinonly?: string | string[]; // Added later, may not be there for older releases
+    firmware_revision?: string;
```
szymonlesisz, line 20: this is confusing,
i dont see `firmware_revision` field in either online (https://data.trezor.io/firmware/t1b1/releases.json) or offline (data dir) releases jsons.


does it mean it will fail here or i missed some deeper logic behind it?

```
if (onlineRelease?.firmware_revision === undefined) {
    return failFirmwareRevisionCheck('firmware-version-unknown');
}
```

packages/connect/src/device/checkFirmwareRevision.ts
szymonlesisz: plus we can download them once and then update current releases stored in DataManager (from what i understand it would fetch new releases on each device connect event)

packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
```diff
@@ -0,0 +1,104 @@
...
+        version: [1, 12, 1],
+        bootloader_version: [1, 12, 1],
+        min_firmware_version: [1, 12, 0],
+        min_bootloader_version: [1, 12, 0],
+        url: 'firmware/t1b1/trezor-t1b1-1.12.1.bin',
+        url_bitcoinonly: 'firmware/t1b1/trezor-t1b1-1.12.1-bitcoinonly.bin',
+        fingerprint: '3c694191f5b66a65cb5bb209adbf113cb40209e644b77162ba996bb7ee8f382b',
+        fingerprint_bitcoinonly: '985fb6a8c87f7547fb810f6c4a8331ebf19c677445810358778eb21eca78a181',
+        firmware_revision: '1eb0eb9d91b092e571aac63db4ebff2a07fd8a1f',
+        changelog:
```
szymonlesisz: nitpick: since this i only a mock, you dont need to put real changelog here

packages/connect/src/device/checkFirmwareRevision.ts
peter-sanderson: I like the idea, but this shall happen only rarely (newer firmware in older suite). I expect this to be a super-edge-case and its not worth it to make more complicated code just to address it. WDYT?

packages/connect/src/device/Device.ts
peter-sanderson: When I added:

```ts
await new Promise(r => setTimeout(r, 60 * 1000));
```

It got stucked on this for 60 seconds:
image

- Then it begins to do discovery...
- and after a second or two you get Ghosts.

packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    return onlineReleases.find(onlineRelease => onlineRelease.version === deviceVersion);
+};
+
+const failFirmwareRevisionCheck = (
+    error: 'revision-mismatch' | 'firmware-version-unknown' | 'firmware-version-unknown',
+): FirmwareRevisionCheckResult => ({ success: false, error });
+
+export type CheckFirmwareRevisionParams = {
+    deviceVersion: VersionArray;
```
komret: I think `firmwareVersion` would be more appropriate, it isn't immediately clear what `deviceVersion` is.

packages/connect/src/types/device.ts
peter-sanderson: Good point :+1:

packages/connect/src/types/firmware.ts
peter-sanderson: It will be there on data.trezor.io. @MiroslavProchazka shall finish his PR https://github.com/trezor/data/pull/88 :grin:

peter-sanderson: If the FW release is not found locally AND its not found online it will show ghosts. 

So it is absolutely **crucial** to not forget to add `firmware_revision` to the data.trezor.io **before** we release new FW version.

packages/connect/src/types/device.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/7b2d5ae513b9c6c2be44bfeb57e4646ef54f8ae2

packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/7b2d5ae513b9c6c2be44bfeb57e4646ef54f8ae2

packages/suite/src/components/suite/Preloader/Preloader.tsx
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/7b2d5ae513b9c6c2be44bfeb57e4646ef54f8ae2

packages/suite/src/actions/suite/suiteActions.ts
```diff
@@ -286,6 +287,17 @@ export const deviceAutenticityOptOut = (payload: boolean) => (dispatch: Dispatch
     });
 };
 
+export const firmwareRevisionCheck =
```
komret: I'd rename to verb form as we usually do for functions, i.e. `checkFirmwareRevision`.

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
```diff
@@ -0,0 +1,41 @@
...
+                  ],
+                  authenticityChecks: device.authenticityChecks,
+              }
+            : {};
+
+        withSentryScope(scope => {
+            scope.setLevel('error');
+            scope.setTag('deviceAuthenticityError', 'device revision check failed');
+            captureSentryMessage(
+                `Device revision check failed! ${JSON.stringify(contextData)}`,
```
komret: _Firmware_ revision check failed!

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
```diff
@@ -0,0 +1,41 @@
...
+                scope,
+            );
+        });
+    }, [device]);
```
komret: The effect runs multiple times due to an object in the dependency array. The relevant properties should be dissected somehow.

packages/suite/src/components/suite/banners/SuiteBanners/SuiteBanners.tsx
```diff
@@ -53,6 +55,16 @@ export const SuiteBanners = () => {
     let priority = 0;
     if (device?.id && firmwareHashInvalid.includes(device.id)) {
         banner = <FirmwareHashMismatch />;
+        priority = 92;
+    } else if (
+        !isFirmwareRevisionCheckDisabled &&
+        !online &&
```
komret: Isn't this condition redundant? What if Suite goes online in the instant between the check and this?

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/FirmwareRevisionCheckModal.tsx
komret, line 1: I'd rename to `FirmwareRevisionOptOutModal` to match `DeviceAuhenticityOptOutModal`.

I also suggest creating a a separate component for shared logic between the two(plus possible `WipeDeviceModal` and some others. While at it, you could update to a more modern syntax (e.g. use `spacingsPx`) and clean up the styles (bold text everywhere looks lame, checkbox might look better aligned to the left).

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/UserContextModal.tsx
```diff
@@ -93,6 +94,8 @@ export const UserContextModal = ({
             return <WipeDeviceModal onCancel={onCancel} />;
         case 'device-authenticity-opt-out':
             return <DeviceAuthenticityOptOutModal onCancel={onCancel} />;
+        case 'device-firmware-revision':
```
komret:
```suggestion
        case 'firmware-revision-opt-out':
```

packages/suite/src/support/messages.ts
```diff
@@ -7163,7 +7163,7 @@ export default defineMessages({
     TR_DEVICE_AUTHENTICITY_OPT_OUT_MODAL_DESCRIPTION_1: {
         id: 'TR_DEVICE_AUTHENTICITY_OPT_OUT_MODAL_DESCRIPTION_1',
         defaultMessage:
-            "Only turn off the device check if you're fully aware of what you're doing and have clear reasons for doing so. If you're uncertain, contact Trezor Support for assistance.",
+            "Only turn off the firmware revision check if you're fully aware of what you're doing and have clear reasons for doing so. If you're uncertain, contact Trezor Support for assistance.",
```
komret: Seems unintended.

packages/suite/src/support/messages.ts
```diff
@@ -7173,7 +7173,65 @@ export default defineMessages({
     TR_DEVICE_AUTHENTICITY_OPT_OUT_MODAL_DESCRIPTION_3: {
         id: 'TR_DEVICE_AUTHENTICITY_OPT_OUT_MODAL_DESCRIPTION_3',
         defaultMessage:
-            'Trezor Support will never ask you to turn off the device check. This feature has been designed to ensure your security.',
+            'Trezor Support will never ask you to turn off the firmware revision check. This feature has been designed to ensure your security.',
```
komret: As above.

packages/connect/src/device/Device.ts
peter-sanderson: @mroz22 Finally I was able to test and simulate it.

image
image

With this setup:

image

packages/suite/src/components/suite/banners/SuiteBanners/SuiteBanners.tsx
```diff
@@ -53,6 +55,15 @@ export const SuiteBanners = () => {
     let priority = 0;
     if (device?.id && firmwareHashInvalid.includes(device.id)) {
         banner = <FirmwareHashMismatch />;
+        priority = 92;
+    } else if (
+        !isFirmwareRevisionCheckDisabled &&
+        device?.features &&
```
peter-sanderson, line 61: Intentionally not checking the `!online` here. Downloading the `releases.json` may fail (blocked, filtered, ..) regardless of being online. 

For example, when testing it and filtering it in the networks tab :)

packages/connect/src/types/firmware.ts
szymonlesisz: if its crucial then it should be covered by some pre-release e2e test

packages/connect/src/types/device.ts
```diff
@@ -34,6 +34,20 @@ export enum FirmwareType {
 // key = coin shortcut lowercase (ex: btc, eth, xrp) OR field declared in coins.json "supportedFirmware.capability"
 export type UnavailableCapabilities = { [key: string]: UnavailableCapability };
 
+export type FirmwareRevisionCheckResult =
+    | {
+          success: true;
+          error: null;
```
mroz22: btw isn't error: null kind of redundant?

packages/connect/src/device/checkFirmwareRevision.ts
mroz22: I think it is good enough now. it's an edgecase + it is timeoutable now, so even if it appears on the happy path (which it should not), it will not block it from execution too much.

packages/connect/src/types/firmware.ts
mroz22: also this PR should get the update of releases.json before merging, right?

peter-sanderson: @mroz22 already done https://github.com/trezor/trezor-suite/pull/13502

packages/connect/src/device/Device.ts
mroz22: fair enough

packages/suite/src/views/settings/SettingsDevice/SettingsDevice.tsx
```diff
@@ -86,10 +88,8 @@ export const SettingsDevice = () => {
...
-    ].includes(deviceModelInternal);
+    const supportsDeviceAuthentication = SUPPORTS_DEVICE_AUTHENTICITY_CHECK[deviceModelInternal];
+    const supportFirmwareRevisionCheck = !supportsDeviceAuthentication; // Older devices with no secure element
```
komret:
```suggestion
    const supportsFirmwareRevisionCheck = !supportsDeviceAuthentication; // Older devices with no secure element
```

Let's list the two devices to avoid some unexpected behaviour with new models.

packages/connect/src/types/device.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/e410e009138e31a4ed6414cf5cf3203547ea272c

komret (changes requested):

packages/connect/src/device/checkFirmwareRevision.ts
peter-sanderson: [31c4b55371](https://github.com/trezor/trezor-suite/pull/13495/commits/31c4b553717a4b4bd26d067457df72236801cbd9)

packages/suite/src/actions/suite/suiteActions.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/3d2414e313716a93e704f1da746fb49491d2616d

packages/suite/src/components/suite/banners/SuiteBanners/SuiteBanners.tsx
peter-sanderson: Yup, see: https://github.com/trezor/trezor-suite/pull/13495#discussion_r1704204027

packages/suite/src/views/settings/SettingsDevice/SettingsDevice.tsx
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/5b654e530778665277c72d2d093b910d4ccb91d8

packages/suite/src/support/messages.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/5b654e530778665277c72d2d093b910d4ccb91d8

packages/suite/src/support/messages.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/5b654e530778665277c72d2d093b910d4ccb91d8

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/UserContextModal.tsx
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/5b654e530778665277c72d2d093b910d4ccb91d8

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/FirmwareRevisionCheckModal.tsx
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/5b654e530778665277c72d2d093b910d4ccb91d8

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
peter-sanderson: https://github.com/trezor/trezor-suite/pull/13495/commits/5b654e530778665277c72d2d093b910d4ccb91d8

packages/connect/src/device/Device.ts
```diff
@@ -547,6 +564,34 @@ export class Device extends TypedEmitter<DeviceEvents> {
         }
     }
 
+    async checkFirmwareRevision() {
+        const firmwareVersion = this.getVersion();
+
+        if (firmwareVersion.length === 0 || !this.features) {
+            return; // This happens when device has no features (not yet connected)
```
mroz22, line 574: I think this shouldn't actually happen at all. but I understand this check is needed because of typescript, right?

packages/connect/src/utils/assets.ts
```diff
@@ -9,13 +9,37 @@ if (global && typeof global.fetch !== 'function') {
...
-export function httpRequest(url: string, type: 'text'): Promise<string>;
-export function httpRequest(url: string, type: 'binary'): Promise<ArrayBuffer>;
-export function httpRequest(url: string, type: 'json'): Promise<Record<string, any>>;
-export function httpRequest(url: any, type: any) {
-    const asset = getAssetByUrl(url);
+export function httpRequest(
+    url: string,
+    type: 'text',
+    options?: RequestInit,
+    skipLocalForceDownload?: boolean,
```
mroz22, line 16: side note: how is this change related with https://github.com/trezor/trezor-suite/issues/13271

mroz22 (approved):
connect part ack

packages/connect/src/device/Device.ts
peter-sanderson: image

komret (approved):

MiroslavProchazka (approved):

PR #13509, opened 2024-07-25
feat(suite): update LoadingContent component to use Spinner end state…

### LINKED ISSUE #13508: Loading content visuals are not consistent
The spinner component has an end animation for both success and error states which is not used in various context where a different solution is implemented instead. It should look and work the same everywhere.

seibei-iguchi:
### Before:
https://github.com/user-attachments/assets/688890b0-dd0d-405f-88de-0318dd10df15

### After:
https://github.com/user-attachments/assets/d1dde56a-337a-43c8-9ceb-abbb86467021


peter-sanderson (approved):

peter-sanderson (approved):
image

jvaclavik (approved):

PR #13513, opened 2024-07-25
chore(suite): remove duplicated file

komret:
## Description

Looks like this file should have been removed in https://github.com/trezor/trezor-suite/pull/12968 as its content has been moved to https://github.com/trezor/trezor-suite/blob/develop/suite-common/wallet-utils/src/reviewTransactionUtils.ts.


mroz22 (approved):
removed and still green? shipit

Issue #13526, opened 2024-07-26
Trezor popup shows UNDER browser extension window on Linux (Firefox)


superKalo:
When a browser extension opens its own popup (via `chrome.windows.create` as a `type: popup`) on Linux (Firefox), this window stays always on top. If the Trezor popup then gets requested within this extension popup, the Trezor popup appears under the extension popup. Making the Trezor popup hardly accessible (visible) for the user, if one doesn't think to move away the extension popup.

martykan:
Closing as legacy popup is being deprecated in favor of Trezor Suite integration

PR #13527, opened 2024-07-26
fix(connect-webextension): Trezor popup displaying UNDER browser extension popup on Linux (Firefox)

superKalo:
## Description

When a browser extension opens its own popup (via `chrome.windows.create` as a `type: popup`) on Linux (Firefox), this window stays always on top. If the Trezor popup then gets requested within this extension popup, the Trezor popup was appearing behind the extension popup, because the Trezor's one had the default type ("normal").

This made the Trezor popup hardly accessible (visible) for the user, if one doesn't think to move away the extension popup.

With this change, when a Trezor popup gets opened from a browser extension popup, it will appear on top of the extension popup.

## Related Issue

https://github.com/trezor/trezor-suite/issues/13526

### DIFF
--- a/packages/connect-web/src/popup/index.ts
+++ b/packages/connect-web/src/popup/index.ts
@@ -224,7 +224,7 @@ export class PopupManager extends EventEmitter {
                 // Request coming from extension popup,
                 // create new window above instead of opening new tab
                 if (currentWindow.type !== 'normal') {
-                    chrome.windows.create({ url }, newWindow => {
+                    chrome.windows.create({ url, type: 'popup' }, newWindow => {
                         chrome.tabs.query(
                             {
                                 windowId: newWindow?.id,


mroz22:
do you have a public reproduction where we could test it?

superKalo:
@mroz22 thanks for stepping in. It's a bit tricky, you can simulate it by:

0. On Linux, with Firefox.

1. Open the browser console in any extension and please run: `browser.windows.create({ url: "https://example.com/", type: 'popup' })` (in the browser console of an extension, so that the browser is available). This will open a browser window (type: popup) and will simulate the problem - this window stays always on top.

2. Now please trigger signing with Trezor within any extension popup (for example with MetaMask when signing a swap on UniSwap or signing a message via [Ambire's sig tool](https://sigtool.ambire.com/)). The Trezor popup will appear UNDER the `https://example.com/` window.

This is making the Trezor popup hardly accessible (visible) for the user, if one doesn't think to move away the `https://example.com/` window popup (which simulates an extension window popup).

With my changes, same scenario, the Trezor popup will appear as a popup now and on top of the `https://example.com/` window popup.

martykan:
Closing as legacy popup is being deprecated in favor of Trezor Suite integration

PR #13534, opened 2024-07-29
fix(suite): blur url 2.0

### LINKED ISSUE #13528: Invalid URL gets detected as a link and is blurred in token name
**Describe the bug**
If a user receives an Ethereum token where the token's name includes a string such as `2.0`, Trezor Suite mistakenly detects it as a valid link and incorrectly blurs it for phishing protection.

**Info:**
- Suite Version: desktop 24.7.2 (3282fe8)
- OS: Linux x86_64 

**How to reproduce**

1. Create an Ethereum account
2. Receive Lido Staked Ether (stETH) tokens
3. Navigate to tokens page
4. Observe token name being partially blurred

**Expected behavior**
`2.0` and other invalid URLs should not be detected as links, and should not be blurred in the tokens page.

**Screenshots**
Screenshot from 2024-07-28 15-21-54
Screenshot from 2024-07-28 15-21-44

tomasklim:
## Description

- 2.0 was incorrectly marked

## Related Issue

Resolve #13528 

## Screenshots:

unresolved because of special chars
Screenshot 2024-07-29 at 9 34 24

### DIFF
--- a/packages/utils/src/extractUrlsFromText.ts
+++ b/packages/utils/src/extractUrlsFromText.ts
@@ -1,0 +1,35 @@
+const URL_REGEX =
+    /\b(?:https?:\/\/|www\.)[a-zA-Z0-9-._~:/?#[\]@!$&'()*+,;=%]+\b|(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(?=\b|\s|$|\])/gi;
+
+export const extractUrlsFromText = (text: string) => {
+    const urls: string[] = [];
+    const textParts: string[] = [];
+    let lastIndex = 0;
+
+    const matches = [...text.matchAll(URL_REGEX)];
+
+    matches.forEach(match => {
+        const url = match[0];
+        const index = match.index !== undefined ? match.index : -1; // Ensure index is defined
+
+        // Capture text before the URL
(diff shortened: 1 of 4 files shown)


karliatto (changes requested):
There should be 2 commits in this PR.
The current one with prefix `fix(suite):....`
And second one with changes to `@trezor/utils` with prefix `feat(utils): ...`

The package  `@trezor/utils` is published to NPM and having commits with proper name and separated by package allow us to generate better automated CHANGELOGS when we do the releases.

tomasklim:
@karliatto done

packages/utils/src/extractUrlsFromText.ts
```diff
@@ -0,0 +1,35 @@
+const urlRegex =
```
AdamSchinzel: nitpick: maybe it will be better to have it like this:
```suggestion
const URL_REGEX =
```

AdamSchinzel (approved):
Nice tests

karliatto (approved):

packages/utils/src/extractUrlsFromText.ts
tomasklim: done

PR #13544, opened 2024-07-29
Support T3B1 in Suite

komret:
## Description

- Prepare Suite for T3B1.
- Use T3B1 as default over T2B1 (thus renaming the assets).
- Some improvements on the way (`isDeviceWithButtons` util).
- Make docs and comments more future-proof (avoid naming particular models).
- Fix white screen in onboarding with custom firmware.
- Run `yarn update-coins` so that it is synchronized with [trezor-common](https://github.com/trezor/trezor-common), no manual changes there.

TODO:
- Add T3B1 firmware binary and `releases.json` - waiting for FW.

How I created the single-frame hologram video for `DeviceAnimation` component:
- Downloaded the image from https://trezor.io/learn/a/authenticate-trezor-safe-5
- Turned it into a video by `ffmpeg -loop 1 -i input.webp -c:v libvpx-vp9 -t 5 -pix_fmt yuv420p output.webm`
- compressed by `ffmpeg -c:v libvpx-vp9 -i t1_success_dark.webm -vf scale=720:-1 t1_success_dark.webm` (https://github.com/trezor/trezor-suite/pull/6358)


tomasklim (dismissed):
Missing support for cardano and solana in `networksConfig.ts`

packages/suite/src/constants/suite/homescreens.ts
```diff
@@ -1,6 +1,6 @@
 import { DeviceModelInternal } from '@trezor/connect';
 
-const t2b1 = [
+const safe3 = [
```
tomasklim: I dont like `t2b1` nor `safe3`. Please come up with better naming

suite-common/message-system/config/config.v1.json
```diff
@@ -814,6 +814,63 @@
                     }
                 }
             }
+        },
+        {
+            "conditions": [
```
tomasklim, line 939: I would do this in separate PR, it is kinda dangerous as we can forget to bump version/release config,...

packages/suite/src/constants/suite/homescreens.ts
tomasklim: `homescreensT3B1` but definitely not just `safe3`,

komret:
> Missing support for cardano and solana in `networksConfig.ts`

Where exactly?

packages/suite/src/constants/suite/homescreens.ts
komret: 52b194bd423809cb1da0395b744c2436db8fc63d and 18f1c14180b6d6e3a46f79a8de21afb91e70120f

suite-common/message-system/config/config.v1.json
komret: Since we are waiting for FW... let's wait and see.

packages/connect-popup/src/static/popup.css
```diff
@@ -2370,8 +2370,8 @@ section > div .divider span:after {
 .select-device .wrapper .select-device-list button .trezor_icon.trezor_icon_t2t1 {
     background-image: url(./images/trezor-t2t1.png);
 }
-.select-device .wrapper .select-device-list button .trezor_icon.trezor_icon_t2b1 {
-    background-image: url(./images/trezor-t2b1.png);
+.select-device .wrapper .select-device-list button .trezor_icon.trezor_icon_t3b1 {
```
martykan, line 2374: This class name should stay, since T3B1 is already defined on L2376

packages/connect/src/data/deviceAuthenticityConfig.ts
```diff
@@ -53,6 +53,20 @@ export const deviceAuthenticityConfig: DeviceAuthenticityConfig = {
             ],
         },
     },
+    T3B1: {
```
mroz22, line 56: manual change? there is s script preparing this file in nightly I think 🤔

scripts/update-coins.sh
```diff
@@ -32,9 +32,12 @@ $SRC/cointool.py dump -p -d connect -o $DIST/coins.json \
     -E eth \
     -E erc20 \
 
+# Exclude coins that are not supported by any device model
+yarn tsx "$(dirname "${BASH_SOURCE[0]}")"/filterCoins.ts
+
 yarn prettier --write $DIST/coins.json
 
-$SRC/cointool.py dump -p -d connect -o $DIST/coins-eth.json \
+$SRC/cointool.py dump -p -o $DIST/coins-eth.json \
```
mroz22, line 40: btw don't eth coins need filtering too?

packages/connect-popup/src/static/popup.css
komret: fff902b2cd5f1151632a1a3d0a78e3da791d7201

scripts/update-coins.sh
komret: There is nothing it would filter out (yet). I could call it twice, not sure what is better.

mroz22: right, ok, I don't know. hope there will never be such scenario where one evm is supported but the other one not

packages/connect/src/data/deviceAuthenticityConfig.ts
komret: Yes, the script works, I just reordered it to keep it in the same order as in the rest of the codebase.

packages/suite/src/components/firmware/ReconnectDevicePrompt.tsx
```diff
@@ -241,6 +241,7 @@ export const ReconnectDevicePrompt = ({ onClose, onSuccess }: ReconnectDevicePro
                     : 'TR_SWITCH_TO_BOOTLOADER_HOLD_LEFT_BUTTON',
             [DeviceModelInternal.T2T1]: 'TR_SWITCH_TO_BOOTLOADER_SWIPE_YOUR_FINGERS',
             [DeviceModelInternal.T2B1]: 'TR_SWITCH_TO_BOOTLOADER_HOLD_LEFT_BUTTON',
+            [DeviceModelInternal.T3B1]: 'TR_SWITCH_TO_BOOTLOADER_HOLD_LEFT_BUTTON',
```
Lemonexe, line 250: This function could do with exhaustive typing, but it wasn't here before, no need to do it now.

suite-common/message-system/config/config.v1.json
```diff
@@ -877,6 +877,69 @@
...
+                {
+                    "environment": {
+                        "desktop": "<24.9.1",
+                        "mobile": "!",
+                        "web": "!"
```
Lemonexe: Why not target outdated web clients too? Cached old version is not very likely, as the client would have to cache `html`, which browsers generaly never cache. But if it's a matter of changing one line, why not?
We do it [for a similar message](https://github.com/trezor/trezor-suite/blob/5c2cb7d048201ff7f72f3d7335fa180109b4f1ba/suite-common/message-system/config/config.v1.json#L743-L745).
```suggestion
                        "web": "<24.9.1"
```

submodules/trezor-common
Lemonexe, line 1: Is this diff intentional? If yes, can't it be done in separate PR in `trezor-common`?

packages/suite/src/utils/suite/homescreen.ts
```diff
@@ -219,7 +221,11 @@ export const isValidImageHeight = (
 };
 
 export const isProgressiveJPG = (buffer: ArrayBuffer, deviceModelInternal: DeviceModelInternal) => {
-    if ([DeviceModelInternal.T2B1, DeviceModelInternal.T1B1].includes(deviceModelInternal)) {
+    if (
+        [DeviceModelInternal.T1B1, DeviceModelInternal.T2B1, DeviceModelInternal.T3B1].includes(
+            deviceModelInternal,
+        )
+    ) {
```
Lemonexe: This array feels very arbitrary, being defined in the `if`.
IMHO it should either be extracted, so it has a name such as `devicesSupportingProgressiveJpg` , or we can use `if (isDeviceWithButtons(deviceModelInternal))`

Though having buttons & supporting progressive JPG is not related, so idk we want to bind these two things together like that.

Same with all similar lines below.

matejkriz (commented):
Tested in mobile app with T2B1 and it works fine. Haven't tested with T3B1 yet.

packages/suite/src/views/settings/SettingsDevice/Homescreen.tsx
```diff
@@ -95,9 +95,11 @@ export const Homescreen = ({ isDeviceLocked }: HomescreenProps) => {
...
         <>
             <SettingsSectionItem anchorId={SettingsAnchor.Homescreen}>
-                {[DeviceModelInternal.T1B1, DeviceModelInternal.T2B1].includes(
-                    deviceModelInternal,
-                ) && (
+                {[
+                    DeviceModelInternal.T1B1,
+                    DeviceModelInternal.T2B1,
+                    DeviceModelInternal.T3B1,
+                ].includes(deviceModelInternal) && (
```
Lemonexe: Again, like in `homescreen.ts`, can we use `isDeviceWithButtons(deviceModelInternal)`

packages/connect/src/data/deviceAuthenticityConfig.ts
mroz22: just that the script might revert this change next time it runs maybe

packages/components/src/components/animations/DeviceAnimation.tsx
```diff
@@ -47,13 +49,29 @@ export const DeviceAnimation = forwardRef<HTMLVideoElement, DeviceAnimationProps
     ) => {
         const { THEME } = useTheme();
 
-        // T2B1, T3T1 animations are transparent
-        const theme = [DeviceModelInternal.T2B1, DeviceModelInternal.T3T1].includes(
-            deviceModelInternal,
-        )
+        // T3B1, T3T1 animations are transparent
```
Lemonexe:
```suggestion
        // T2B1, T3B1, T3T1 animations are transparent
```

Lemonexe (approved):
Codewise, I see no problems, just a few comments. 
I haven't tested the code, but in this case that needs to be done by QA, I don't have a T3B1 device.

packages/connect/src/data/deviceAuthenticityConfig.ts
komret: I checked, it shouldn't. It selects a key and rewrites the value. https://github.com/trezor/trezor-suite/blob/58da390d6157d7ee13d5ae8e773fe61a01582ab1/scripts/ci/check-connect-data.ts#L52

packages/components/src/components/animations/DeviceAnimation.tsx
komret: c52c2f051ed3b88af257f23f1a57a3fdd236dc8b

packages/suite/src/views/settings/SettingsDevice/Homescreen.tsx
komret: 42d003328df4bd298d813c40a2016f1527c6b438

packages/suite/src/utils/suite/homescreen.ts
komret: 42d003328df4bd298d813c40a2016f1527c6b438

suite-common/message-system/config/config.v1.json
komret: 6760d6c2abfe574314289160f84bae1e404b8f9b The message doesn't make much sense on web, but I guess it is better this way 👍

packages/suite/src/components/firmware/ReconnectDevicePrompt.tsx
komret: Agreed, I don't like this util. 42d003328df4bd298d813c40a2016f1527c6b438

packages/suite/src/components/firmware/ReconnectDevicePrompt.tsx
```diff
@@ -226,23 +227,28 @@ export const ReconnectDevicePrompt = ({ onClose, onSuccess }: ReconnectDevicePro
...
+                [DeviceModelInternal.T2B1]: 'TR_SWITCH_TO_BOOTLOADER_HOLD_LEFT_BUTTON',
+                [DeviceModelInternal.T3B1]: 'TR_SWITCH_TO_BOOTLOADER_HOLD_LEFT_BUTTON',
+                [DeviceModelInternal.T3T1]: 'TR_SWITCH_TO_BOOTLOADER_SWIPE_YOUR_FINGERS',
+            };
+
+            return switchToBootloaderMap[deviceModelInternal];
+        }
+
+        // Fallback. This should never happen.
+        return 'TR_SWITCH_TO_BOOTLOADER_HOLD_LEFT_BUTTON';
```
Lemonexe: Nit: I'd rather do early return to handle the `undefined` case. But that's just my opinion, I am not sure if we have a convention to prefer early returns.

packages/suite/src/constants/suite/device.ts
```diff
@@ -52,3 +52,11 @@ export const SUPPORTS_DEVICE_AUTHENTICITY_CHECK: Record<DeviceModelInternal, boo
...
     [DeviceModelInternal.T3T1]: true,
 };
+
+export const HAS_MONOCHROME_SCREEN: Record<DeviceModelInternal, boolean> = {
+    [DeviceModelInternal.T1B1]: true,
+    [DeviceModelInternal.T2T1]: false,
+    [DeviceModelInternal.T2B1]: true,
+    [DeviceModelInternal.T3B1]: true,
+    [DeviceModelInternal.T3T1]: false,
+};
```
Lemonexe, line 62: Perfect :ok_hand:

packages/suite/src/components/firmware/ReconnectDevicePrompt.tsx
komret: 06cd5504ff7681bebc9e5da5ebee1db05c068222 Yes, probably that's better. I also fixed a bug I created by the previous fixup and added some explanation to why it is done this way.

komret:
I added one last fixup (8ea3bf052fca355ba77f4cf36e25775b5bce0fe9) bumping the version is message-system because we are going to release T3B1 next month. Going to merge it now.

PR #13546, opened 2024-07-29
Chore: transport ArrayBuffer -> Buffer

szymonlesisz:
## Description

Forgotten todo, return `Buffer` instead of `ArrayBuffer` from transport api read (nodeusb/webusb)


mroz22 (approved):
lgtm, cc @matejkriz just in case connecting device stopped working in react-native

matejkriz (approved):
Thanks for the ping. It works fine in react-native 👌

PR #13555, opened 2024-07-30
fix(connect): longer timeout for getFeatures due to suite-native performance issue

### LINKED ISSUE #13456: Connecting device first and starting app second throws an error
**Describe the bug**
If I connect my device first and start the app via the prompt on Android, the app then shows a connection error and prompts me to "Reconnect device". I have to have already one view-only wallet in the app first. I also can't get out of this state unless I reconnect the device which the app don't exactly say I should do, unfortunately.

**Info:**

-   Mobile Suite Version: 24.6.0 https://github.com/trezor/trezor-suite/commit/ff75009f1076edbd6476b90c97fa899ca7807b48
-   OS: Android 12

**How to reproduce**

1. Connect your device
2. Wait for the discovery of the standard wallet and enable view only
3. Disconnect the device and kill the app
4. Connect the device and select to start the app when the Android prompt shows
5. A connection error is shown in the app

**Expected behavior**
My device correctly connects and discovery happens

**Screenshots**
image

matejkriz:
## Description

During app start of Suite Lite, get features could takes longer than 3s depending on view-only wallets and the phone. This is a workaround for now, we should investigate why it takes so long.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13456

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -6,6 +6,7 @@ import { DeviceCommands, PassphrasePromptResponse } from './DeviceCommands';
 import { PROTO, ERRORS, NETWORK } from '../constants';
 import { DEVICE, DeviceButtonRequestPayload, UI } from '../events';
 import { getAllNetworks } from '../data/coinInfo';
+import { DataManager } from '../data/DataManager';
 import { getFirmwareStatus, getRelease } from '../data/firmwareInfo';
 import {
     parseCapabilities,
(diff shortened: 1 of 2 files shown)


mroz22:
Hmm maybe it would be better to:

1. delete this timeout here.
2. even old bridge returns enough information from enumeration to be able to tell that this is an old HID device (see enumerate result: `[{"path":"1","vendor":21324,"product":1,"debug":false,"session":"5","debugSession":null}]
`. which is then used in this [map](https://github.com/trezor/trezor-suite/blob/develop/packages/transport/src/api/abstract.ts#L20)
3. introduce custom handling for this types of devices on connect level. we would not be reading the at all and we would emit unreadable device right away. 

on the other hand, old bridge still supports updating this device, sooo... 🤔

mroz22 (approved):

PR #13556, opened 2024-07-30
transport chores

mroz22:
cherry-pick non-controversial commits from #13493

### DIFF
--- a/packages/transport/e2e/tests/bridge.test.ts
+++ b/packages/transport/e2e/tests/bridge.test.ts
@@ -1,8 +1,6 @@
 import * as messages from '@trezor/protobuf/messages.json';
-// testing build. yarn workspace @trezor/transport build:lib is a required step therefore
 import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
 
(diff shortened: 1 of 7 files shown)


packages/transport/e2e/tests/bridge.test.ts
```diff
@@ -31,173 +29,159 @@ describe('bridge', () => {
         TrezorUserEnvLink.disconnect();
     });
 
-    // there might be more versions of bridge out there, see https://github.com/trezor/webwallet-data/tree/master/bridge
-    // but they are not available from trezor-user-env, see https://github.com/trezor/trezor-user-env/tree/master/src/binaries/trezord-go/bin
-    ['2.0.26', '2.0.27', undefined].forEach(bridgeVersion => {
```
mroz22, line 36: this is the only change, removing forEach. A bigger change, yes, but I don't think it makes sense to test legacy versions of legacy bridge anymore. And I am adding tests for node-bridge in  #13493 where only 1 legacy version of legacy bridge will still be tested

marekrjpolak (approved):
Seal of approval granted. :seal:

PR #13561, opened 2024-07-30
fix(connect): remove superfluous space from error message

### LINKED ISSUE #13560: UI glitch: connect: superfluous space
image

mroz22:
fix #13560 found while checking on #13555

Probably the most influential PR I ever made. 

image

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -375,7 +375,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
                 return Promise.reject(
                     ERRORS.TypedError(
                         'Device_InitializeFailed',
-                        `Initialize failed: ${error.message} ${
+                        `Initialize failed: ${error.message}${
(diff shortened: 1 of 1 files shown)


martykan (approved):
I didn't test it, but it should probably work I think

mroz22:
btw @Hannsek at the moment this is what prev 1.4.0 bl devices appear in suite after initial connection (in some cases). maybe they would deserve better handling. I will think about it a bit

Hannsek:
The idea is that there should be something like this:

image


When clicking on the link, the transport would either automatically switch to the old Trezor Bridge or will take to the the debug settings.

mroz22:
ok, please note it somewhere. at the moment, something like this is shown with the new bridge 

image

Hannsek:
by `Trezor Bridge` we mean the old or new?

mroz22:
:man-shrugging:

PR #13562, opened 2024-07-30
chore: remove unused libs, add used libs

tomasklim:
## Description

- remove unused libs
- add type-fest and bcrypto to deps as they are used


packages/connect-webextension/package.json
```diff
@@ -48,7 +48,6 @@
         "@babel/preset-typescript": "^7.23.3",
         "@trezor/node-utils": "workspace:*",
         "@trezor/trezor-user-env-link": "workspace:*",
-        "@types/chrome": "^0.0.260",
```
mroz22: you sure it is not used?

mroz22: my vscode sources types for this line from @types/chrome https://github.com/trezor/trezor-suite/blob/develop/packages/connect-web/src/channels/serviceworker-window.ts#L15

packages/suite-web/package.json
```diff
@@ -40,13 +40,11 @@
         "@trezor/transport": "workspace:*",
         "@trezor/trezor-user-env-link": "workspace:*",
         "@trezor/utils": "workspace:*",
-        "@types/chrome-remote-interface": "^0.31.14",
         "@types/cypress-image-snapshot": "^3.1.8",
         "@types/react": "18.2.79",
         "@types/react-dom": "18.2.19",
         "@types/react-router": "^5.1.20",
         "@types/react-router-dom": "^5.1.7",
-        "chrome-remote-interface": "^0.33.0",
```
mroz22, line 49: this used to be used in cypress tests. not sure if it is still active, haven't been maintaining it for a while.

packages/utxo-lib/package.json
```diff
@@ -41,6 +41,7 @@
     "dependencies": {
         "@trezor/utils": "workspace:*",
         "bchaddrjs": "^0.5.2",
+        "bcrypto": "5.4.0",
```
mroz22: normally eslint or some other tools yells when a dependency is missing, why didn't it catch this? 🤔

packages/suite-web/e2e/cypress.config.ts
```diff
@@ -1,6 +1,5 @@
 import { defineConfig } from 'cypress';
 
-// import CDP from 'chrome-remote-interface';
```
mroz22, line 3: hmm lines 41 and 176 are dead too. I think this was used to add hover ability to cypress. not sure why we commented this out? maybe @andrcj knows

packages/connect-webextension/package.json
tomasklim: returned back

packages/suite-web/package.json
tomasklim: it was commented out so i guess it is ready to be removed

packages/suite-web/e2e/cypress.config.ts
tomasklim: not sure if 176 is dead, `resetCRI` is used in `commands.ts`

packages/utxo-lib/package.json
tomasklim: my mistake, should not be added

packages/suite-web/package.json
mroz22: I guess so, cc @andrcj

packages/suite-web/e2e/cypress.config.ts
mroz22: cc @andrcj I don't really maintain this anymore

PR #13564, opened 2024-07-30
feat(analytics): start reporting font scale and pixel density

matejkriz:
## Description

Let's start reporting font scales and pixel density to be able to set proper priority for testing and developing for different font scales...


vytick (approved):
lgtm 🚀

PR #13568, opened 2024-07-30
chore(suite-native): cancel pin only if it was requested

matejkriz:
## Description

Add missing parenthesis and trigger connect cancel pin only if it was requested (ConnectDeviceScreenHeader is reused in multiple screens not related to PIN at all)

## Notes for QA @STew790
Please test X (cancel button) on different connecting device flows (T1 with PIN, any other Trezor model than T1, locked device during app use...)

### DIFF
--- a/suite-native/module-authorize-device/src/components/connect/ConnectDeviceScreenHeader.tsx
+++ b/suite-native/module-authorize-device/src/components/connect/ConnectDeviceScreenHeader.tsx
@@ -18,6 +18,7 @@ import { selectIsDeviceDiscoveryActive } from '@suite-common/wallet-core';
 import {
     cancelPassphraseAndSelectStandardDeviceThunk,
     selectIsCreatingNewPassphraseWallet,
+    selectDeviceRequestedPin,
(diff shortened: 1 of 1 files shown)


vytick (approved):
🚢 🇮🇹

STew790:
**QA NOK**
When I try to open passphrase with locked (but already connected) TS5, and then I try to cancel the PIN request, it keeps on requesting the PIN on the device and in the app (stuck in a loop)
If I cancel it on the device first and then in the app I get out of the loop

**Info**

24.8.0 5cd19fd

matejkriz:
I can't reproduce on 24.9.1, can you double check it @STew790, please?

https://github.com/user-attachments/assets/106e1d58-2fe1-4d3f-996b-b05cd4ca8e3e

STew790:
**QA OK**
Yes, it works correctly now for me as well.

**Info**
24.9.2 https://github.com/trezor/trezor-suite/commit/7a4ddc9a8137951fe69c7693d4ac9e70511a39b4

PR #13570, opened 2024-07-30
Fix/transport usb abort

szymonlesisz:
## Description

1. fix memory leak (MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners added to [AbortSignal]) in case when received/send message >= 10 chunks. i've tested it on DebugLinkGetState (require custom FW build) but i suppose it could be tested on some SignMessage (?)

2. fix endless read (device.transferIn) when aborted. testing scenario:
 - call transport.read
 - cancel transport read using abort signal
 - call transport.call('GetFeaures', {})
 
 
 NOTE: commit 2 needs more testing on different environments (windows, linux, mac, chromeos, webusb atd)

### DIFF
--- a/packages/transport/src/api/usb.ts
+++ b/packages/transport/src/api/usb.ts
@@ -125,7 +125,10 @@ export class UsbApi extends AbstractApi {
         }));
     }
 
-    private abortableMethod<R>(method: () => R, signal?: AbortSignal) {
+    private abortableMethod<R>(
+        method: () => R,
+        { signal, onAbort }: { signal?: AbortSignal; onAbort?: () => void },
+    ) {
         if (!signal) {
             return method();
         }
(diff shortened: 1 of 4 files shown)


packages/transport/src/api/usb.ts
```diff
@@ -119,32 +120,50 @@ export class UsbApi extends AbstractApi {
...
+        };
+
+        const dfd = createDeferred<R>();
+        const abortListener = () => {
+            abort();
+            dfd.reject(new Error(ERRORS.ABORTED_BY_SIGNAL));
+        };
+        signal?.addEventListener('abort', abortListener);
+
+        return Promise.race([method(), dfd.promise])
```
mroz22: note for me: this is probably place to serialize all promises to run in sequence. will explain why later

mroz22:
rebased and added a test in d98631823c7fc5852a35bceb580d053df29e0821 https://dev.suite.sldev.cz/transport-test/fix/transport-usb-abort/
 the problem is that it doesn't seem to work. calling reset does not seem to be clearing read buffer. please lets discuss this.

mroz22:
@trezor/qa I would like to draw your attention to our new webusb tester https://dev.suite.sldev.cz/transport-test/develop/
it is intended to test some barebones of our USB implementation easily in different operating systems and browsers. If you get all green here, it is very likely that there won't be any troubles on webusb connection layer.

mroz22 (approved):

PR #13571, opened 2024-07-30
feat(solana): validate solana amount with rent

PeterBenc:
## Description
 
- fetch necessary rent for non-empty solana accounts
- validate in send modal that remaining solana will be zero or more than rent 

## Related Issue

Resolve

### DIFF
--- a/docs/packages/suite/send.md
+++ b/docs/packages/suite/send.md
@@ -50,6 +50,7 @@
 -   AMOUNT_IS_LESS_THAN_RESERVE (XRP only: trying to send less XRP than required reserve to the empty account)
 -   AMOUNT_IS_NOT_IN_RANGE_DECIMALS (amount with invalid decimal places)
 -   AMOUNT_IS_NOT_INTEGER (ERC20 only: token doesn't accept decimal places)
+-   REMAINING_BALANCE_LESS_THAN_RENT (solana only: account has to keep a minimal balance equal to rent)
 
 ---
 
--- a/suite-common/wallet-core/src/send/sendFormSolanaThunks.ts
+++ b/suite-common/wallet-core/src/send/sendFormSolanaThunks.ts
@@ -6,6 +6,7 @@ import {
     ExternalOutput,
     PrecomposedTransaction,
     PrecomposedLevels,
+    Account,
 } from '@suite-common/wallet-types';
 import { createThunk } from '@suite-common/redux-utils';
 import {
@@ -36,6 +37,8 @@ const calculate = (
     availableBalance: string,
     output: ExternalOutput,
     feeLevel: FeeLevel,
+    decimals: number,
+    rent: number,
     token?: TokenInfo,
 ): PrecomposedTransaction => {
     const feeInLamports = feeLevel.feePerTx;
@@ -61,6 +64,19 @@ const calculate = (
 
         // errorMessage declared later
         return { type: 'error', error, errorMessage: { id: error } } as const;
+    }
+    const remainingSolBalance = new BigNumber(availableBalance).minus(totalSolSpent);
+
+    if (remainingSolBalance.isLessThan(rent) && remainingSolBalance.isGreaterThan(0)) {
+        const errorMessage = {
+            id: 'REMAINING_BALANCE_LESS_THAN_RENT' as const,
+            values: {
+                remainingSolBalance: formatAmount(remainingSolBalance, decimals),
+                rent: formatAmount(rent, decimals),
+            },
+        };
+
+        return { type: 'error', error: errorMessage.id, errorMessage } as const;
     }
 
     const payloadData: PrecomposedTransaction = {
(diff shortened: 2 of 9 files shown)


docs/packages/suite/send.md
```diff
@@ -50,6 +50,7 @@
 -   AMOUNT_IS_LESS_THAN_RESERVE (XRP only: trying to send less XRP than required reserve to the empty account)
 -   AMOUNT_IS_NOT_IN_RANGE_DECIMALS (amount with invalid decimal places)
 -   AMOUNT_IS_NOT_INTEGER (ERC20 only: token doesn't accept decimal places)
+-   REMAINING_BALANCE_LESS_THAN_RENT (solana only: account has to keep a minimal balance equal to rent)
```
mroz22, line 53: this is similar concept to what ripple has right? If yes, maybe there might be some reuse of components / copy. just a node for suite team

mroz22 (commented):
please @martykan check on the blockchain-link part 🙏 from my POV it looks okey but I am no solana guy

suite-common/wallet-core/src/send/sendFormSolanaThunks.ts
```diff
@@ -237,7 +253,14 @@ export const composeSolanaTransactionFeeLevelsThunk = createThunk<
 
         const resultLevels: PrecomposedLevels = {};
         const response = predefinedLevels.map(level =>
-            calculate(account.availableBalance, output, level, tokenInfo),
+            calculate(
+                account.availableBalance,
+                output,
+                level,
+                decimals,
+                (account as Extract<Account, { networkType: 'solana' }>).misc.rent!,
```
martykan: This looks kinda ugly to me, I would rather use a type guard

suite-common/wallet-core/src/send/sendFormSolanaThunks.ts
```diff
@@ -236,8 +259,18 @@ export const composeSolanaTransactionFeeLevelsThunk = createThunk<
...
+        assertIsSolanaAccount(account);
+
         const response = predefinedLevels.map(level =>
-            calculate(account.availableBalance, output, level, tokenInfo),
+            calculate(
+                account.availableBalance,
+                output,
+                level,
+                decimals,
+                account.misc.rent!,
```
martykan: And are we sure it can't be undefined?

PeterBenc: Yes, for Solana accounts, the misc property is always defined and the rent is too. Sadly the assertion has to be somewhere because the typing is not as precise as it could be, this is fragile I agree. 

We might want not to assert but fail gracefully to some default rent value so if this breaks, Solana still works, and only the rent-related logic would be "turned off". This makes sense to me, what do you think?

martykan: Yeah I think maybe just `?? 0` should be enough probably

MiroslavProchazka:
@PeterBenc please squash those fixup commits, otherwise it will not pass the commit message tests. Thx!

tomasklim (approved):
LGTM, not tested. Please @trezor/qa test thoroughly. You somehow need to make your account to pay rent

suite-common/wallet-core/src/send/sendFormSolanaThunks.ts
```diff
@@ -123,6 +139,13 @@ const fetchAccountOwnerAndTokenInfoForAddress = async (
     return [accountOwner, tokenInfo] as const;
 };
 
+function assertIsSolanaAccount(
+    account: Account,
+): asserts account is Extract<Account, { networkType: 'solana' }> {
+    if (account.networkType !== 'solana')
+        throw new Error(`Invalid network type. ${account.networkType}`);
+}
+
```
tomasklim, line 148:
```suggestion
const assertIsSolanaAccount = (
    account: Account,
): asserts account is Extract<Account, { networkType: 'solana' }> => {
    if (account.networkType !== 'solana') {
        throw new Error(`Invalid network type. ${account.networkType}`);
    }
}

```

arrow function + brackets, then please squash

PeterBenc: I would, but functions that return `asserts` has to be function declarations
https://blog.logrocket.com/assertion-functions-typescript/#:~:text=Learn%20more%20%E2%86%92-,Function%20declarations%20and%20expressions,-At%20the%20time

vytick (approved):

bosomt:
**QA OK**

https://github.com/trezor/trezor-suite/pull/13757#issuecomment-2286638027

MiroslavProchazka:
@tomasklim soo can we merge it? We still need to pass commit message thou - cc @PeterBenc

MiroslavProchazka:
Closing this as it's merged into repo via https://github.com/trezor/trezor-suite/pull/13757

PR #13572, opened 2024-07-31
feat(suite): show zero balance tokens

### LINKED ISSUE #13338: EVM Tokens/ Allow user to show 0 balance tokens
Prerequisite for #13039

Allow user to show 0 balance tokens to be able to operate with all tokens he has in tx history and if he want to trade it

Rabby also shows it under this button
Screenshot 2024-07-15 at 12 38 55

Design:
image

AdamSchinzel:
## Description

Show zero balance tokens instead of hiding them behind debug mode.

## Related Issue

Resolve #13338

## Screenshots:
Screenshot 2024-07-31 at 9 35 46 AM


packages/suite/src/utils/wallet/tokenUtils.ts
```diff
@@ -90,16 +90,41 @@ export const getTokens = (
 
         if (searchQuery && !isTokenMatchesSearch(token, query)) return;
 
+        const hasBalance = token.balance !== '0';
```
tomasklim: Please use BigNumber.gte(0), just to be sure

packages/suite/src/utils/wallet/tokenUtils.ts
```diff
@@ -90,16 +90,41 @@ export const getTokens = (
...
+
         if (isShown) {
```
tomasklim: how about something like this, but not strong opinion 

```
const hasBalance = token.balance !== '0';
const pushToArray = (arrayWithBalance, arrayWithoutBalance) => {
    if (hasBalance) {
        arrayWithBalance.push(token);
    } else {
        arrayWithoutBalance.push(token);
    }
};

if (isShown) {
    pushToArray(shownWithBalance, shownWithoutBalance);
} else if (hasCoinDefinitions && !isKnown) {
    pushToArray(unverifiedWithBalance, unverifiedWithoutBalance);
} else if (isHidden) {
    pushToArray(hiddenWithBalance, hiddenWithoutBalance);
} else {
    pushToArray(shownWithBalance, shownWithoutBalance);
}
```

tomasklim (commented):
Otherwise looks good. Please create test for `getTokens` it's getting really complicated

packages/suite/src/utils/wallet/tokenUtils.ts
AdamSchinzel: Looks cleaner 👍

tomasklim (approved):

PR #13575, opened 2024-07-31
feat(suite): Allow tests in DS

jvaclavik:
## Related Issue

Resolve


peter-sanderson (commented):
I would like to see at least one test (event super simple) to demonstrate it's working.

PR #13576, opened 2024-07-31
feat(connect): no transport auto retry

martykan:
## Description

Relatively simple implementation done by recursively retrying inside the `onCall` method. 
This will probably change with the multi transport refactor. 

Needs new Suite with the bridge requested modal, with older Suite the call conflicts with discovery.

### DIFF
--- a/packages/connect/src/core/index.ts
+++ b/packages/connect/src/core/index.ts
@@ -542,34 +542,35 @@ const onCallDevice = async (
 
     const shouldRetry = ['web', 'webextension'].includes(env);
     // find device
-    let device: Device;
-    try {
-        device = await initDevice(context, method.devicePath);
-    } catch (error) {
-        if (error.code === 'Transport_Missing') {
-            // wait for popup handshake
-            await waitForPopup(context);
-            // show message about transport
-            sendCoreMessage(createUiMessage(UI.TRANSPORT));
+    let tempDevice: Device | undefined;
+    while (!tempDevice) {
+        try {
+            tempDevice = await initDevice(context, method.devicePath);
+        } catch (error) {
+            if (error.code === 'Transport_Missing') {
+                // wait for popup handshake
+                await waitForPopup(context);
+                // show message about transport
+                sendCoreMessage(createUiMessage(UI.TRANSPORT));
 
-            // Retry initDevice again
-            // NOTE: this should change after multi-transports refactor, where transport will be always alive
-            if (deviceList.pendingConnection() && shouldRetry) {
-                while (deviceList.pendingConnection()) {
-                    await deviceList.pendingConnection();
+                // Retry initDevice again
+                // NOTE: this should change after multi-transports refactor, where transport will be always alive
+                if (deviceList.pendingConnection() && shouldRetry) {
+                    while (deviceList.pendingConnection()) {
+                        await deviceList.pendingConnection();
+                    }
+                    continue;
                 }
-
-                // call onCallDevice again recursively
-                return await onCallDevice(context, message, method);
+            } else {
+                // cancel popup request
+                sendCoreMessage(createPopupMessage(POPUP.CANCEL_POPUP_REQUEST));
             }
-        } else {
-            // cancel popup request
-            sendCoreMessage(createPopupMessage(POPUP.CANCEL_POPUP_REQUEST));
+            // TODO: this should not be returned here before user agrees on "read" perms...
+            sendCoreMessage(createResponseMessage(responseID, false, { error }));
+            throw error;
         }
-        // TODO: this should not be returned here before user agrees on "read" perms...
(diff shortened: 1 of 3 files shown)


packages/connect/src/core/index.ts
```diff
@@ -546,6 +558,17 @@ const onCall = async (context: CoreContext, message: IFrameCallMessage) => {
             await waitForPopup(context);
             // show message about transport
             sendCoreMessage(createUiMessage(UI.TRANSPORT));
+
+            // Retry onCallDevice again
+            // NOTE: this should change after multi-transports refactor, where transport will be always alive
```
mroz22: I believe that some kind of retry-logic will always make sense. It can always be possible that none transport will be available. And it will always be useful to be able to pause, solve the problem, come back and continue

packages/connect/src/core/index.ts
```diff
@@ -546,6 +558,17 @@ const onCall = async (context: CoreContext, message: IFrameCallMessage) => {
...
+            if (transportReconnect && (env === 'web' || env === 'webextension')) {
+                // wait 10s between reconnects
+                const reconnectTime = 10000 - (Date.now() - initDeviceTime);
+                await resolveAfter(reconnectTime).promise;
+
+                // call onCallDevice again recursively
+                return await onCallDevice(context, message, method);
```
mroz22: wouldnt it be better to remain hanging here on some promise, that would resolve after 
 ```
deviceList.on('transport-start')
```
is emitted and only after that it would retry the onCallDevice call?

File: packages/connect/src/core/index.ts [hunk 3]
```diff
@@ -621,7 +644,7 @@ const onCall = async (context: CoreContext, message: IFrameCallMessage) => {
     device.on(DEVICE.SAVE_STATE, (state: string) => {
         // Persist internal state only in case of core in popup
         // Currently also only for webextension until we asses security implications
-        if (useCoreInPopup && env === 'webextension') {
+        if (useCoreInPopup && env === 'webextension' && origin) {
```
mroz22, line 641: this is because in the onCall fn, origin is assigned as follows
```
    const origin = DataManager.getSettings('origin')!;
```
right?

mroz22:
I tested it and it worked. Left couple of comments. Also @marekrjpolak please if you could check on it 🙏

packages/connect/src/core/index.ts
martykan: Yes, in this method we're not doing that so it's possibly an undefined type.
Maybe it would be better to check that in onCall as well.

packages/connect/src/core/index.ts
martykan: Interesting idea, I will take a look at the emitted events

martykan: Just pushed the change to this, seems to be working well

mroz22 (approved):
approve, but @marekrjpolak please check whether it still fits the bigger picture.

marekrjpolak:
> approve, but @marekrjpolak please check whether it still fits the bigger picture.

I have some quick comments, but I'm still thinking how to join them together:

- I personally don't like recursion where unnecessary and actively try to transform it into cycles
- `deviceList.on(TRANSPORT.START, ...)` should be equivalent to `await deviceList.pendingConnection()` + `deviceList.isConnected()` which is already used somewhere
- `transportReconnect` should be somehow equivalent to `!deviceList.isConnected()` + `deviceList.pendingConnection()`, because whenever `deviceList` with `transportReconnect` disconnects, it immediately assigns new pending connection promise
- `Transport_Missing` from `initDevice` is explicitly thrown if `!deviceList.isConnected()` after `await deviceList.pendingConnection()` so it may be somehow propagated down instead of identified above?
- `onCallDevice` already started with checks like these (but for non-transportReconnect variant), could it be somehow unified?

I'll give it some time. I believe it can be solved simply with `deviceList.pendingConnection()` and `deviceList.isConnected()`.

marekrjpolak:
I've finally managed to finish my proposed changes (well, some of them). I'm not satisfied with them anyway, but I presume they gonna be changed with multitransport, so fuck it.

- https://github.com/trezor/trezor-suite/pull/13576/commits/ed98669d81fa110afe68ef47600eb0b7260d9db8 - practically the only **real** change is a bit crazy-looking `if (pendingConnection()) { while (pendingConnection()) { await pendingConnection() } }` branch, which should be the same as before:
  - if `transportReconnect` is active (because in that case deviceList sets next pending connection right after transport fails)
  - wait for every `pendingConnection` as long as there's any (because it's only set to undefined after successful connection)
- https://github.com/trezor/trezor-suite/pull/13576/commits/916b4e0b0f46e7549fb290350c3e87f6c5d35cfa - again should be the same as before, I just prefer cycles to recursion because the control flow is more obvious imo

Please @martykan check that it still satisfies the reason why you implemented it and that you're not too unhappy with it.

PR #13579, opened 2024-07-31
move brightness settings for T3T1/T2T1 outside debug mode

MiroslavProchazka:
Move brightness settings for t3t1 & T2T1 outside debug mode

image

### DIFF
--- a/packages/suite/src/views/settings/SettingsDevice/Brightness.tsx
+++ b/packages/suite/src/views/settings/SettingsDevice/Brightness.tsx
@@ -1,9 +1,8 @@
 import TrezorConnect from '@trezor/connect';
 import { analytics, EventType } from '@trezor/suite-analytics';
-import { SettingsSectionItem } from 'src/components/settings';
+import { SectionItem } from 'src/components/suite';
 import { ActionButton, ActionColumn, TextColumn, Translation } from 'src/components/suite';
-import { SettingsAnchor } from 'src/constants/suite/anchors';
-import { useDevice, useSelector } from '../../../hooks/suite';
+import { useDevice } from '../../../hooks/suite';
 
 interface DeviceLabelProps {
     isDeviceLocked: boolean;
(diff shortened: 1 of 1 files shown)


packages/suite/src/views/settings/SettingsDevice/Brightness.tsx
komret, line 1: There is an incorrect `anchorId` from previous changes. If you don't want to create a new one (probably not necessary for this section), use `SectionItem` instead of `SettingsSectionItem`.

komret (changes requested):

komret:
I am also missing the toast saying _Settings changed successfully_ which other options have, e.g. when changing homescreen.

packages/suite/src/views/settings/SettingsDevice/Brightness.tsx
```diff
@@ -3,7 +3,8 @@ import { analytics, EventType } from '@trezor/suite-analytics';
 import { SettingsSectionItem } from 'src/components/settings';
 import { ActionButton, ActionColumn, TextColumn, Translation } from 'src/components/suite';
 import { SettingsAnchor } from 'src/constants/suite/anchors';
-import { useDevice, useSelector } from '../../../hooks/suite';
+import { useDevice } from '../../../hooks/suite';
+import { applySettings } from 'src/actions/settings/deviceSettingsActions';
```
peter-sanderson: I think its is unused import.

peter-sanderson:
@MiroslavProchazka that toast is just calling

```ts
dispatch(notificationsActions.addToast({ type: 'settings-applied' }));
```

Normally, settings use `applySettings` thunk (that also pops this toast up). This setting however does not use `applySettings` and I did not noticed the toast is missing.

MiroslavProchazka:
Unfortunately we cannot get error toast message, since FW is somehow not supporting this action. Se we've decided with @komret to not include toast notification for now, until it will be fixed on FW side.

komret (approved):

bosomt:
**QA OK**

T3T1 and T2T1 available
T2B1 unavailable 

**Info:**
 - Suite version: web 24.8.0 (e0f5e835d94b657933b127f232422a7910c87954)
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.1 regular (revision 55473a0b4be7b22317cd0c401c938d430d86f29f)

PR #13580, opened 2024-07-31
feat(suite): token transaction history

### LINKED ISSUE #12825: EVM tokens/ Token transaction history
after clicking on transaction button, user should be taken on the transaction history screen, where transaction related only to that token would be filtered

AdamSchinzel:
## Description

Make possible to filter token transaction from token row.

For now Redux is used for both transaction history and send form prefill in favour of url params.

## Related Issue

Resolve #12825 

## Screenshots:
Screenshot 2024-07-31 at 5 31 56 PM
Screenshot 2024-07-31 at 5 34 44 PM


packages/suite/src/views/wallet/send/Outputs/Amount/TokenSelect.tsx
```diff
@@ -148,11 +149,15 @@ export const TokenSelect = ({ output, outputId }: TokenSelectProps) => {
     }, [outputId, tokenWatch, setAmount, getValues, account.networkType, isSetMaxActive]);
 
     useEffect(() => {
-        if (routerParams?.contractAddress) {
-            setValue(tokenInputName, routerParams.contractAddress, { shouldValidate: true });
+        if (sendFormPrefill) {
+            setValue(tokenInputName, sendFormPrefill, { shouldValidate: true });
```
tomasklim: I am missing `Clear all` button in send form when token is autofilled
Screenshot 2024-08-19 at 17 35 40

Screenshot 2024-08-19 at 17 31 19

packages/suite/src/views/wallet/transactions/TransactionList/TransactionListActions/TransactionListActions.tsx
```diff
@@ -71,6 +76,17 @@ export const TransactionListActions = ({
         setSearch('');
     }, [account.symbol, account.index, account.accountType, setSearch]);
 
+    useEffect(() => {
```
tomasklim, line 79: 👍 but fill is superslow. is it possible to make it faster?

tomasklim (approved):
LGTM, just solve clear button, please and check if it is possible to fill search faster 🙏

packages/suite/src/views/wallet/send/Outputs/Amount/TokenSelect.tsx
AdamSchinzel: Fixed

packages/suite/src/views/wallet/transactions/TransactionList/TransactionListActions/TransactionListActions.tsx
AdamSchinzel: When I was trying it the whole search was slow on like 7+ pages of transactions so coping the address manually into the input or clicking the button was quite the same.

If think this can be addressed in other PR 🙏

PR #13581, opened 2024-07-31
fix(transport): pass all options to bridgeApiCall (signal, timeout)

szymonlesisz:
## Description

`TrezorBridgeTransport` did not send `signal` and `timeout` options to `bridgeApiCall` therfore `fetch` api is never aborted/timeouted.

there is something wrong with eslint/ts-check here, it should notify about unused variables from the spread.


mroz22 (approved):
🙇

Issue #105, opened 2019-06-03
Add landing pages


vladimirvolek:
- [x] connect device
- [x] connect device with webusb
- [x] install bridge
- [x] version
- [x] firmware update in PR #423 
- [x] onboarding

mroz22:
done? maybe there are other pages (backup? #448, #95). Not sure if they are 'landing' though

PR #13582, opened 2024-07-31
fix(suite-native): open links via useOpenLink to catch errors

vytick:
Open links via useOpenLink and show toast on failed links.

Resolve [#105](https://github.com/trezor/trezor-suite-private/issues/105)

### DIFF
--- a/suite-native/link/src/useOpenLink.ts
+++ b/suite-native/link/src/useOpenLink.ts
@@ -6,19 +6,30 @@ import { useToast } from '@suite-native/toasts';
 export const useOpenLink = () => {
     const { showToast } = useToast();
 
+    const showErrorToast = useCallback(() => {
+        showToast({
+            variant: 'error',
+            icon: 'warningTriangle',
+            message: 'Unable to open the link',
+        });
+    }, [showToast]);
+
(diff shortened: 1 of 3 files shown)


suite-native/link/src/useOpenLink.ts
```diff
@@ -6,19 +6,29 @@ import { useToast } from '@suite-native/toasts';
...
+        showToast({
+            variant: 'error',
+            icon: 'warningTriangle',
+            message: 'Unable to open the link',
+        });
+    }, [showToast]);
+
     const handleOpenLink = useCallback(
         async (href: string) => {
+            const canOpenURL = await Linking.canOpenURL(href);
```
matejkriz: I would rather move it inside the `try` block too, even checking if canOpenURL can fail.

matejkriz (changes requested):
Cool, but there are imports `import { Linking } from 'react-native';` no longer used.

suite-native/link/src/useOpenLink.ts
vytick: 1f27a23501

matejkriz (approved):
🎉

PR #13588, opened 2024-08-01
fix(suite): account label behavior when changing in page header

### LINKED ISSUE #11940: Very long labels overlapping Send and Receive buttons
**Describe the bug**
Very long labels overlapping Send / Receive buttons

**Steps to reproduce:**
1. Create very long label
2. Try to send or receive coins
3. You have to scroll to see expected buttons

**Info:**
 - Suite version: desktop 24.4.1 (f4684c1893fe8b0f45c217837e2cbedaa6928d25)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.4.1 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)

**Screenshots:**

https://github.com/trezor/trezor-suite/assets/31506317/2f841fb6-5a01-4552-9cf0-28e31ad411d7

seibei-iguchi:
When renaming account label in the page header, the unsubmitted value was not kept when switching from the top and sticky positions of the header. It has been updated to use one component instead of two and the autoscaling input has been updated to handle changes in width based on style changes.

### Before
https://github.com/user-attachments/assets/4282641d-7b01-41ef-8b8b-e144ff700f30

### After
https://github.com/user-attachments/assets/4e3c12d1-48ab-42c0-844b-f0790d103b15


packages/components/src/components/AutoScalingInput/AutoScalingInput.tsx
```diff
@@ -18,8 +18,8 @@ type ApplyWidthParams = {
 const applyWidth = ({ width, minWidth, ref }: ApplyWidthParams) => {
     if (ref?.current?.style) {
         const borderSize = ref.current.offsetWidth - ref.current.clientWidth;
-
-        // First it needs to be set to minimum, this ensures that with of the input will be shrinking as text gets smaller
+        console.log('applyWidth')
```
peter-sanderson: note: dont forget to remove

packages/components/src/components/AutoScalingInput/AutoScalingInput.tsx
```diff
@@ -35,7 +35,7 @@ const createHandleOnChangeAndApplyNewWidth =
     (event: ChangeEvent<HTMLInputElement>) => {
         const { target } = event;
         const borderSize = target.offsetWidth - target.clientWidth;
-
+        console.log('onchange')
```
peter-sanderson: note: dont forget to remove

packages/components/src/components/AutoScalingInput/AutoScalingInput.tsx
```diff
@@ -60,32 +60,34 @@ const StyledInput = styled.input`
...
-        // Measure the size of the placeholder
+        // Measure the size of the placeholder 
         useEffect(() => {
             if (placeholderMeasureRef?.current?.style) {
-                setPlaceholderWidth(placeholderMeasureRef.current.scrollWidth ?? 0);
+                setCalculatedMin(Math.max(minWidth, placeholderMeasureRef.current.scrollWidth ?? 0));
+                console.log('fdfd', placeholderMeasureRef.current.scrollWidth)
             }
-        }, [setPlaceholderWidth]);
+        });
```
peter-sanderson: missing dependecy, :point_up: see github action error

packages/components/src/components/AutoScalingInput/AutoScalingInput.tsx
```diff
@@ -60,32 +60,34 @@ const StyledInput = styled.input`
...
+                : inputRef?.current?.scrollWidth ?? calculatedMin;
 
             applyWidth({
-                width,
+                width: calculatedMin,
                 minWidth: calculatedMin,
                 ref: inputRef,
             });
-        }, [value, inputRef, calculatedMin]);
+        }, [inputRef, calculatedMin]);
```
peter-sanderson, line 90: missing dependecy: `value`

packages/components/src/components/AutoScalingInput/AutoScalingInput.tsx
seibei-iguchi: As I wrote in the message in Slack, this is a complete mess and all of it will have to change. It works somehow currently but I'd like to understand why.

peter-sanderson: yep, I am just making notes so we dont forget anything

seibei-iguchi:
I've updated the autoscalinginput after a discussion with @peter-sanderson . I'd be happy to receive suggestion regarding naming the new so-far-stupidly-named prop that triggers the recalculation. Is it ok if I keep it as `any` or should I do state management somewhere along the way and cast it to `boolean`? I'm not sure myself.

packages/suite/src/components/suite/labeling/MetadataLabeling/definitions.ts
```diff
@@ -13,6 +13,7 @@ export interface Props {
     // override default behavior of metadata labeling element visible only on hover
     visible?: boolean;
     placeholder?: string;
+    updateFlag?: any;
```
jvaclavik, line 16: Could we be more specific about type here?

jvaclavik (approved):

jvaclavik:
> I've updated the autoscalinginput after a discussion with @peter-sanderson . I'd be happy to receive suggestion regarding naming the new so-far-stupidly-named prop that triggers the recalculation. Is it ok if I keep it as `any` or should I do state management somewhere along the way and cast it to `boolean`? I'm not sure myself.


What about rename it to `forceUpdate`? What exactly is the issue with the type? Why not have it as boolean?

seibei-iguchi:
Because I don't really care about the value itself (Boolean or not), only that it changes.

packages/suite/src/components/suite/labeling/MetadataLabeling/definitions.ts
peter-sanderson: I agree.

I was used to naming it something like  `xxxForceUpdateCounter` and having it as a (only increasing) number. That seems to me as obvious way how to communicate the intent. 

But I am open to any naming/type that is clear enough about the intent of this prop.

packages/components/src/components/AutoScalingInput/AutoScalingInput.tsx
```diff
@@ -71,21 +69,14 @@ export const AutoScalingInput = forwardRef<HTMLInputElement, Props>(
...
             }
-        }, [setPlaceholderWidth]);
+        }, [setPlaceholderWidth, value, updateFlag]);
 
         useEffect(() => {
-            const isValueEmpty = value === '' || value === null || value === undefined;
-
-            const width = isValueEmpty
-                ? calculatedMin
-                : Math.max(calculatedMin, inputRef?.current?.scrollWidth ?? calculatedMin);
```
peter-sanderson, line 81: This was here to cover the "empty" input (so it wont colapse to zero width). Is it working for this case?

seibei-iguchi: Yep, because the source of truth for width is now placeholderWidth along with the minWidth (max of those).

peter-sanderson (approved):
Haven't tested tho.

packages/suite/src/components/suite/labeling/MetadataLabeling/definitions.ts
seibei-iguchi: I'm not sure how to handle this, though. At this moment, the "flag" is a boolean, because I need the input to re-render when the header isScrolling (or not), but obviously, I don't care about the value itself and the input doesn't need to know what true or false means.

peter-sanderson: @adamhavel true, I am ok to keep it.

maybe one more suggestion: what about passing `forceUpdateDependecies: any[]` an the simply spreading them into dependecy array?

Not sure if this works? But maybe it can be a good (readable) pattern how to do those forced re-renders)

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 24.8.2 (2a555706122c837a581ed139cf65114ad85fe7cf)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.8.2 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982

PR #13592, opened 2024-08-01
Chore/transport bridge types

### LINKED ISSUE #13593: node-bridge - add validation of request params/body

szymonlesisz:
## Description

Some typescript black magic to get strongly typed request handlers with params and body validation.

Resolve: https://github.com/trezor/trezor-suite/issues/13593

Rules:
- params validator (ParamsValidatorHandler) needs to be defined as first handler on the list, otherwise `request.params` will be `unknown`
- body types are inherited from the next-to-last handler, if it is not defined then `request.body` will be `unknown`
- last handler (ResolveHandler) does not have `next` (3rd) argument

TODO:
- [x] params validators
- [x] body validators
- [x] parseBodyJSON on error (marked as TODO)

### DIFF
--- a/packages/transport-bridge/src/http.ts
+++ b/packages/transport-bridge/src/http.ts
@@ -7,7 +7,8 @@ import {
     allowOrigins,
     parseBodyJSON,
     parseBodyText,
-    Handler,
+    RequestHandler,
+    ParamsValidatorHandler,
     RequestWithParams,
     Response,
 } from '@trezor/node-utils';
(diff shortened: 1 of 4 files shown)


packages/transport-bridge/src/http.ts
```diff
@@ -357,29 +357,28 @@ export class TrezordNode {
 
                     const { ext } = path.parse(pathname);
 
-                    fs.exists(pathname, exist => {
-                        if (!exist) {
-                            // if the file is not found, return 404
-                            res.statusCode = 404;
-                            res.end(`File ${pathname} not found!`);
+                    const exist = fs.existsSync(pathname);
```
mroz22: imho blocking execution probably shouldn't be here. as long trezor-bridge does not run in a separate process, this could probably harm the smoothness of execution of the thread it is running in. I would switch it back to the async version. 

side note: I think we are going to run bridge in a separate thread so that the occasional segfaults don't crash suite-desktop if they happen. 

i can change it if you want

packages/transport-bridge/src/http.ts
```diff
@@ -181,47 +185,48 @@ export class TrezordNode {
...
             ]);
 
-            app.post('/listen', [
+            app.post<ApiPost<Descriptor[]>>('/listen', [
                 parseBodyJSON,
                 (req, res) => {
                     res.setHeader('Content-Type', 'text/plain');
                     this.listenSubscriptions.push({
-                        // todo: type in that if parseBodyText was called as previous handler, req.body is string. is that even possible
-                        // @ts-expect-error
```
mroz22: nice, or definitely nicer than having expect-errors. still the ultimate goal I failed to do was to set  req.body type dynamically based on middleware functions registered before the main handler. eg calling parseBodyText would turn req body from unknown to string.

szymonlesisz: i guess its done now

packages/transport-bridge/src/http.ts
szymonlesisz: resolved here: https://github.com/trezor/trezor-suite/pull/13606

mroz22 (approved):
looks good, thank you 🍀

PR #13595, opened 2024-08-01
fix(connect-web): don't fallback to core in popup with no webusb

martykan:
## Description

Slight afterthought with `coreMode: auto`. 

When using a browser that doesn't support webUSB or when it's explicitly disabled by settings, we shouldn't switch to popup since it doesn't provide any benefit. 

If there are iframe loading issues it will still switch. 

Also cleans up the error handling function.


mroz22 (approved):

Dave-Matthew:
good feature

PR #13596, opened 2024-08-01
refactor(suite): refactor frameProps

jvaclavik:
## Description

- do not allow all frame props for each DS component 
  - for example I can set that `Warning` component shouldn't have `width` prop but it can have `margin` 
- it automatically shows only all allowed frame props in storybook for each component
- add `height, width, maxHeight` to frame props
- delete bad usage of frame props in `TransactionReviewEvmExplanation`

not very related to frameprops:
- fix issue when user can't move with window on mac

## Related Issue

Resolve

### DIFF
--- a/packages/components/src/components/Badge/Badge.tsx
+++ b/packages/components/src/components/Badge/Badge.tsx
@@ -5,12 +5,16 @@ import { Icon } from '@suite-common/icons/src/webComponents';
 import { borders, Color, CSSColor, spacings, spacingsPx, typography } from '@trezor/theme';
 import { focusStyleTransition, getFocusShadowStyle } from '../../utils/utils';
 import type { UISize, UIVariant } from '../../config/types';
-import { FrameProps, TransientFrameProps, withFrameProps } from '../common/frameProps';
+import { FrameProps, withFrameProps } from '../common/frameProps';
+import { TransientProps } from '../../utils/transientProps';
+
+export const allowedFrameProps: (keyof FrameProps)[] = ['margin'];
+type AllowedFrameProps = Pick<FrameProps, (typeof allowedFrameProps)[number]>;
 
 type BadgeSize = Extract<UISize, 'tiny' | 'small' | 'medium'>;
 type BadgeVariant = Extract<UIVariant, 'primary' | 'tertiary' | 'destructive'>;
 
-export type BadgeProps = FrameProps & {
+export type BadgeProps = AllowedFrameProps & {
     size?: BadgeSize;
     variant?: BadgeVariant;
     isDisabled?: boolean;
@@ -31,7 +35,7 @@ type BadgeContainerProps = {
     $variant: BadgeVariant;
     $hasAlert: boolean;
     $inline: boolean;
-} & TransientFrameProps;
+} & TransientProps<AllowedFrameProps>;
 
 const mapVariantToBackgroundColor = ({ $variant, theme }: MapArgs): CSSColor => {
     const colorMap: Record<BadgeVariant, Color> = {
--- a/packages/components/src/components/Badge/Badge.tsx
+++ b/packages/components/src/components/Badge/Badge.tsx
@@ -5,10 +5,10 @@ import { Icon } from '@suite-common/icons/src/webComponents';
 import { borders, Color, CSSColor, spacings, spacingsPx, typography } from '@trezor/theme';
 import { focusStyleTransition, getFocusShadowStyle } from '../../utils/utils';
 import type { UISize, UIVariant } from '../../config/types';
-import { FrameProps, withFrameProps } from '../common/frameProps';
+import { FrameProps, FramePropsKeys, withFrameProps } from '../common/frameProps';
 import { TransientProps } from '../../utils/transientProps';
 
-export const allowedBadgeFrameProps: (keyof FrameProps)[] = ['margin'];
+export const allowedBadgeFrameProps: FramePropsKeys[] = ['margin'];
 type AllowedFrameProps = Pick<FrameProps, (typeof allowedBadgeFrameProps)[number]>;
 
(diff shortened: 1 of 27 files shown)


packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeLayout.tsx
```diff
@@ -145,32 +143,36 @@ const Left = () => {
...
-                                <Button
-                                    variant="tertiary"
-                                    icon="EXTERNAL_LINK"
-                                    iconAlignment="right"
-                                >
-                                    trezor.io
-                                </Button>
-                            </TrezorLink>
-                        </LinksContainer>
+                        <TrafficLightOffset>
```
jvaclavik, line 146: Without this component it wasn't possible to move with window while user on Mac was on the Welcome screen 🫣

packages/components/src/components/common/frameProps.tsx
```diff
@@ -11,12 +11,24 @@ type Margin = {
...
+    maxWidth?: string | number;
+    height?: string | number;
+    maxHeight?: string | number;
 };
 
 export type TransientFrameProps = TransientProps<FrameProps>;
 
-export const withFrameProps = ({ $margin, $maxWidth }: TransientFrameProps) => {
+const getValueWithUnit = (value: string | number) =>
+    typeof value === 'string' ? value : `${value}px`;
```
jvaclavik, line 24: We want to allow also string for values like `100%`

packages/components/src/components/common/frameProps.tsx
```diff
@@ -28,20 +40,54 @@ export const withFrameProps = ({ $margin, $maxWidth }: TransientFrameProps) => {
...
-                category: 'Frame props',
+export const getFramePropsStory = (allowedFrameProps: Array<keyof FrameProps>) => {
+    const argTypes = allowedFrameProps.reduce(
+        (acc, key) => ({
+            ...acc,
+            [key]: {
+                table: {
+                    category: 'Frame props',
+                },
+                type: key === 'margin' ? 'object' : 'number',
```
jvaclavik, line 69: In storybook we primarily show that values are in numbers, not strings, because it would be hard to understand that you need to type `10px` instead of 10. Also it wouldn't be predictable when we use numbers only in margin.

If you have better idea how to do it, let me know

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewEvmExplanation.tsx
```diff
@@ -29,16 +18,14 @@ export const TransactionReviewEvmExplanation = ({
     }
 
     return (
-        <Wrapper $margin={margin}>
```
jvaclavik, line 32: Good example how not use `frameProps`. It should available be only for internal usage in DS

seibei-iguchi: Can you please elaborate? I'm still not sure what's the issue here.

jvaclavik: It has zero additional value and it's easily replaceable with frame props in `Warning` component. Also I don't like that it's imported outside of `@trezor/component` API (`@trezor/components/src/components/common/frameProps`).

seibei-iguchi: Gotcha, thanks!

packages/components/src/components/typography/Heading/Heading.tsx
```diff
@@ -1,8 +1,11 @@
 import styled from 'styled-components';
 
 import { Color, TypographyStyle, typography } from '@trezor/theme';
-import { FrameProps, TransientFrameProps, withFrameProps } from '../../common/frameProps';
-import { makePropsTransient } from '../../../utils/transientProps';
+import { FrameProps, withFrameProps } from '../../common/frameProps';
+import { makePropsTransient, TransientProps } from '../../../utils/transientProps';
+
+export const allowedHeadingFrameProps: (keyof FrameProps)[] = ['margin'];
```
peter-sanderson: Can we export some type for  `(keyof FrameProps)` I fear we may end up repeating this in other places.

packages/components/src/components/Badge/Badge.tsx
```diff
@@ -5,12 +5,16 @@ import { Icon } from '@suite-common/icons/src/webComponents';
 import { borders, Color, CSSColor, spacings, spacingsPx, typography } from '@trezor/theme';
 import { focusStyleTransition, getFocusShadowStyle } from '../../utils/utils';
 import type { UISize, UIVariant } from '../../config/types';
-import { FrameProps, TransientFrameProps, withFrameProps } from '../common/frameProps';
+import { FrameProps, withFrameProps } from '../common/frameProps';
+import { TransientProps } from '../../utils/transientProps';
+
+export const allowedBadgeFrameProps: (keyof FrameProps)[] = ['margin'];
```
peter-sanderson: Here is the repeated usage of `(keyof FrameProps)` lets make it and type

seibei-iguchi (approved):

packages/components/src/components/common/frameProps.tsx
peter-sanderson: If you want to do some TS magic I think it is possible to type and percentage value by using [template-string typing](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html):

 ```ts
 type CssValue = `${number}%` | number | "auto" | "whatever..."
```
but I haven't tested it

peter-sanderson (approved):

PR #13608, opened 2024-08-02
fix(suite-native): fix assign to read-only property shadowOffset error

matejkriz:
## Description

In debug mode, biometrics was not working, always got this error:

`Cannot assign to read-only property 'shadowOffset'`

## Screenshots:
image


vytick (approved):
good spot 🚀

PR #13613, opened 2024-08-02
chore(utils): introduce new test for PEPE2.0 token

tomasklim:
## Description

- add new test just to make sure we do not break this token in the future 

## Related Issue

Resolve https://satoshilabs.slack.com/archives/CKZHV2G13/p1722593994191089


martykan (approved):
LGTM

PR #13615, opened 2024-08-02
connect explorer - method processing visual feedback

mroz22:
While working on #13493 I noticed a [particular connect-popup test failing ](https://github.com/trezor/trezor-suite/actions/runs/10215128459/job/28264136363?pr=13493#step:5:914) although I don't think I have made changes that may cause this. 
When debugging it locally, I observed a considerable delay between clicking the submit button in the connect-explorer and the connect popup opening.

This PR doesn't solve the issue. It just adds a spinner to the button to provide users with some basic UX feedback in case this happens (my first impression was that tests for some reason don' click the button).

Interestingly, nightly pipelines are mostly green https://github.com/trezor/trezor-suite/actions/workflows/test-connect-popup.yml?query=branch%3Adevelop while in all other branches they have way lower success rate https://github.com/trezor/trezor-suite/actions/workflows/test-connect-popup.yml. Could it be a race happening when runners are under stress?

### DIFF
--- a/packages/components/src/components/buttons/Button/Button.tsx
+++ b/packages/components/src/components/buttons/Button/Button.tsx
@@ -131,7 +131,7 @@ export const Button = ({
         />
     ) : null;
 
-    const Loader = <Spinner size={getIconSize(size)} />;
+    const Loader = <Spinner size={getIconSize(size)} dataTest={`${rest['data-test']}/spinner`} />;
 
     return (
         <ButtonContainer
--- a/packages/connect-explorer/src/components/Method.tsx
+++ b/packages/connect-explorer/src/components/Method.tsx
@@ -5,7 +5,7 @@ import { useState, useCallback, useEffect } from 'react';
 import styled, { useTheme } from 'styled-components';
 import { CopyToClipboard } from 'nextra/components';
 
-import { Button as TrezorButton, H3, Card } from '@trezor/components';
+import { Button as TrezorButton, ButtonProps, H3, Card } from '@trezor/components';
 import { spacingsPx } from '@trezor/theme';
 
 import type { Field, FieldWithBundle, FieldWithUnion } from '../types';
@@ -225,6 +225,26 @@ export const VerifyButton = ({ name, onClick }: VerifyButtonProps) => {
     return <Button onClick={() => onClick(verifyUrls[index])}>Verify response</Button>;
 };
 
+type SubmitButtonProps = {
+    onClick: ButtonProps['onClick'];
+    isFullWidth?: ButtonProps['isFullWidth'];
+    isLoading: ButtonProps['isLoading'];
+    text?: string;
+};
+
+const SubmitButton = ({ onClick, text, isFullWidth, isLoading }: SubmitButtonProps) => {
+    return (
+        <Button
+            onClick={onClick}
+            data-test="@submit-button"
+            isFullWidth={isFullWidth}
+            isLoading={isLoading}
+        >
+            {text || 'Submit'}
+        </Button>
+    );
+};
+
 export const Method = () => {
     const theme = useTheme();
     const { method } = useSelector(state => ({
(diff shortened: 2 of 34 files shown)


packages/components/src/components/buttons/Button/Button.tsx
```diff
@@ -131,7 +131,7 @@ export const Button = ({
         />
     ) : null;
 
-    const Loader = <Spinner size={getIconSize(size)} />;
+    const Loader = <Spinner size={getIconSize(size)} dataTest={rest['data-test']} />;
```
jvaclavik: This seems to me like we will have duplicit test id or inner component is too smart. 
In both cases it's not predictable what exactly is happening. 

I would suggest use it in `Spinner` like this:

```tsx
data-test={dataTest ? dataTest : `@spinner`}
```

instead of:

```tsx
data-test={dataTest ? `${dataTest}/spinner` : `@spinner`}
```


And of course it needs to specify it here:
```tsx
    const Loader = <Spinner size={getIconSize(size)} dataTest={`${rest['data-test']}/spinner`} />;
```

WDYT?

packages/connect-explorer/src/components/Method.tsx
```diff
@@ -225,6 +225,23 @@ export const VerifyButton = ({ name, onClick }: VerifyButtonProps) => {
     return <Button onClick={() => onClick(verifyUrls[index])}>Verify response</Button>;
 };
 
+type SubmitButtonProps = Omit<ButtonProps, 'children'> & {
+    text?: string;
+};
+
+const SubmitButton = ({ onClick, text, isFullWidth, isLoading }: SubmitButtonProps) => {
```
jvaclavik, line 235: You are using only 3 props from `ButtonProps` here. So I can use for example 
`<SubmitButton variant="tertiary" ... />` and I don't get any type error even though it's not working properly. 

What about specify type like this?

```tsx
type SubmitButtonProps = {
    onClick: ButtonProps["onClick"];
    isFullWidth: ButtonProps["isFullWidth"];
    isLoading: ButtonProps["isLoading"];
    text?: string;
};

```

packages/components/src/components/buttons/Button/Button.tsx
mroz22: this does not allow me to pass a distinguishable data-test to a spinner that is part of Button component. Both Button and Spinner rendered by the Button would have the same data-test

packages/connect-explorer/src/components/Method.tsx
mroz22: sounds right, fixed up

packages/components/src/components/buttons/Button/Button.tsx
mroz22: sounds right, fixed up

jvaclavik (approved):
LGTM!

packages/connect-explorer/src/components/ParamsTable.tsx
```diff
@@ -110,17 +110,17 @@ export const ParamsTable = ({ schema, topLevelSchema, descriptions }: ParamsTabl
     const common = { topLevelSchema: topLevelSchemaCurrent, descriptions };
     if (schema[Kind] === 'Union') {
         return schema.anyOf?.map((param: TSchema, i: number) => (
-            <>
+            <div key={i}>
                 {i > 0 && <h3>or</h3>}
                 <ParamsTable key={i} schema={param} {...common} />
```
mroz22, line 115: i could have removed he key from this line I guess

martykan:
We have a test in [popup-close](https://github.com/trezor/trezor-suite/actions/runs/10216255687/job/28267625116?pr=13615#logs) where we're doing multiple calls, should we allow clicking the button even when loading or rethink this test?

mroz22:
> We have a test in [popup-close](https://github.com/trezor/trezor-suite/actions/runs/10216255687/job/28267625116?pr=13615#logs) where we're doing multiple calls, should we allow clicking the button even when loading or rethink this test?

"solved" by forcing the click on disabled element

Issue #13303, opened 2024-07-11
BNB Smart Chain launch


tomasklim:
`BNB Smart Chain` 
- was added in #12880 in release `24.7.1` under `debug` mode.
- was moved in #13634 in release `24.8.1` to experimental features.
- will be released in `24.11.1`

```[tasklist]
### Launch list
- [ ] https://github.com/trezor/trezor-suite/issues/12903
- [ ] https://github.com/trezor/trezor-suite/issues/12901
- [ ] https://github.com/trezor/trezor-suite/issues/14672
```

PR #13634, opened 2024-08-04
Chore/bnb smart chain

tomasklim:
## Description

- when we were playing with fw repo with Komar, I found out that BNB is reported there as BNB and not BSC. I am not sure why I lived in conviction that it is BSC there. 

https://github.com/trezor/trezor-firmware/blob/7307c47b94ddc2f0df48e1fd359231ce5d881249/common/defs/ethereum/networks.json#L36C1-L37C1

- this PR renames all `bsc` symbols to `bnb`
- adds BNB into experimental features

@mroz22 the only thing I am not sure about is collision of BNB Smart Chain and Binance Chain in connect, both having symbol `bnb`. Binance Chain will probably be dead in 10 days anyway https://www.bnbchain.org/en/bnb-chain-fusion

## Related Issue

#13303

## Screenshots:
Screenshot 2024-08-04 at 22 15 54
Screenshot 2024-08-04 at 22 16 08
Screenshot 2024-08-04 at 22 16 26


mroz22:
It doesn't interfere in any way with TrezorConnect.binance* calls. At least with those that  we have tests for that such as: https://github.com/trezor/trezor-suite/blob/transport-meow/packages/connect/e2e/__fixtures__/binanceGetAddress.ts#L10

the only thing I am not sure about is whether there could be some interference with TrezorConnect.getAccountInfo for the old finance chain. But  I guess this was never  really used by anybody? If you would try to call TrezorConnect.getAccountInfo for old bnb chain, it would probably fail because there is no blockchain-link record here I guess? https://github.com/trezor/trezor-suite/blob/chore/bnb-smart-chain/packages/connect-common/files/coins.json#L2270

mroz22:
btw what is this? https://github.com/trezor/trezor-suite/blob/chore/bnb-smart-chain/packages/connect-common/files/coins.json#L2269 shouldn't it be killed too?

mroz22 (approved):
from the connect POV, this should be ok.

PR #13652, opened 2024-08-05
chore(connect-popup): allow on mobile

martykan:
## Description

Allow using Connect popup on mobile browsers & devices that support WebUSB, such as Chrome on Android. 
Adjust styling, use same font size as desktop (in my opinion the old mobile font was too small, these days phones have plenty of space and resolution).
Add test for methods on mobile.


mroz22 (commented):
good idea. those css changes seem to be affecting also non-mobile version but I guess in a desirable way?

martykan:
The non mobile version should look the same, the main change is that I made the font size the same on mobile

mroz22 (approved):

PR #13654, opened 2024-08-05
chore(suite): remove token count from string

komret:
https://github.com/trezor/trezor-suite/pull/13653#discussion_r1704185391


MiroslavProchazka (approved):
lgtm

PR #13657, opened 2024-08-05
feat(suite): Rework Welcome screen left sidebar, make it smaller

### LINKED ISSUE #13521: Welcome screen efficiency update
Please rework the welcome screen to new design. The aim is to get rid of the overly large left sidebar that has almost no content and make more space for the right content focused part.

Please leave out the e-shop icon for now, we will specify it in a separate issu later.

New design (unpolished WIP): https://www.figma.com/design/U7XRtJkKpGl3eVhxgiKts5/branch/wQYIG92l6vRVX72zMJUQtk/Connect-Trezor?m=auto&t=bb7ZqBeNyjWeVrIV-1

For polished icons and colors please use the left bar menu component: https://www.figma.com/design/Y2E8dcdNYrSMGIVRjYFXKp/%5BSuite%5D-Components?m=auto&t=XHwhA9Fthw8flxxh-1

hstastna:
## Description

Rework Welcome screen left sidebar, make it smaller, add appropriate icons according to the new design.

Additionally refactor `CheckIcon` used for indicating enabling/disabling Tor in Sidebar layout - prevent inheriting styles from `Icon` component, use it for Welcome screen too, create and use new related components.

Design [here](https://www.figma.com/design/U7XRtJkKpGl3eVhxgiKts5/Connect-Trezor?node-id=6050-9463&t=fxBQxL49QaQKhkDp-0)

TODO:
- [x] add correct style of the sidebar icons, on hover, also when active
- [x] add labels to icons, to Trezor icon add "Connect"
- [x] remove unnecessary animation
- [x] use default 84px width because of Mac
- [x] display correct page when clicking back on Trezor icon (we are anywhere except _Settings_)
- [x] when being on _Settings_, don't display unnecessary "Back" button
- [x] add footer icons: Hide/Show balances, enable/disable Tor, make icons work
- [ ] make the icon spinner when enabling/disabling Tor bigger
- [ ] test on Mac

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13521

## Videos:
**Before** (Desktop app):

https://github.com/user-attachments/assets/54e237ea-067e-4181-927b-22408b5b1a9d

**After** (Desktop app):

https://github.com/user-attachments/assets/71157e7e-ce97-4dc1-9015-fc401ea11a4c

### DIFF
--- a/packages/components/src/components/buttons/IconButton/IconButton.tsx
+++ b/packages/components/src/components/buttons/IconButton/IconButton.tsx
@@ -10,6 +10,9 @@ import { Tooltip } from '../../Tooltip/Tooltip';
 const IconButtonContainer = styled(ButtonContainer)`
     position: relative;
     padding: ${({ $size }) => getPadding($size, false)};
+    ${({ $borderRadius }) => $borderRadius && `border-radius: ${$borderRadius};`};
+    ${({ $width }) => $width && `width: ${$width};`};
+    ${({ $height }) => $height && `height: ${$height};`};
 `;
 
 const Label = styled.span<{ $isDisabled: boolean }>`
@@ -75,10 +78,13 @@ export const IconButton = ({
                 disabled={isDisabled || isLoading}
                 onClick={handleClick}
                 $isSubtle={isSubtle}
+                $borderRadius={rest.borderRadius}
+                $width={rest.width}
+                $height={rest.height}
                 {...rest}
             >
-                {!isLoading && icon && IconComponent}
                 {isLoading && Loader}
+                {!isLoading && icon && IconComponent}
 
                 {bottomLabel && <Label $isDisabled={isDisabled}>{bottomLabel}</Label>}
             </IconButtonContainer>
--- a/packages/components/src/components/buttons/buttonStyleUtils.ts
+++ b/packages/components/src/components/buttons/buttonStyleUtils.ts
@@ -7,10 +7,9 @@ import { hexToRgba } from '@suite-common/suite-utils';
 const SUBTLE_ALPHA = 0.12;
 const SUBTLE_ALPHA_HOVER = 0.2;
 
-export type ButtonVariant = Extract<
-    UIVariant,
-    'primary' | 'secondary' | 'tertiary' | 'info' | 'warning' | 'destructive'
->;
+export type ButtonVariant =
+    | Extract<UIVariant, 'primary' | 'secondary' | 'tertiary' | 'info' | 'warning' | 'destructive'>
+    | 'welcome';
 export type ButtonSize = Extract<UISize, 'large' | 'medium' | 'small' | 'tiny'>;
 export type IconAlignment = Extract<UIHorizontalAlignment, 'left' | 'right'>;
 
@@ -122,6 +121,13 @@ export const useVariantStyle = (
             text: theme.textOnRed,
             textSubtle: theme.textAlertRed,
         },
+        welcome: {
+            background: theme.backgroundSurfaceElevation1,
+            backgroundSubtle: theme.backgroundSurfaceElevationNegative,
+            backgroundHover: theme.backgroundSurfaceElevation1,
+            fill: 'black',
+            fillSubtle: theme.iconSubdued,
+        },
     };
 
     const colors = variantsColors[variant];
@@ -131,12 +137,16 @@ export const useVariantStyle = (
             ? colors.backgroundSubtle
             : colors.background};
         color: ${isSubtle && colors.textSubtle ? colors.textSubtle : colors.text};
+        ${colors.fill &&
+        colors.fillSubtle &&
+        `path { fill: ${isSubtle ? colors.fillSubtle : colors.fill}}`};
 
         &:hover,
         &:active {
             background: ${isSubtle && colors.backgroundSubtleHover
                 ? colors.backgroundSubtleHover
                 : colors.backgroundHover};
+            ${colors.fill && `path { fill: ${colors.fill}}`};
         }
     `;
 };
--- a/packages/suite/src/components/settings/SettingsLayout.tsx
+++ b/packages/suite/src/components/settings/SettingsLayout.tsx
@@ -16,7 +16,7 @@ type SettingsLayoutProps = {
     ['data-testid']?: string;
 };
 
-const SettingsHeader = () => {
+export const SettingsHeader = () => {
     const isDebugModeActive = useSelector(selectIsDebugModeActive);
 
     const dispatch = useDispatch();
--- a/packages/suite/src/components/suite/Preloader/Preloader.tsx
+++ b/packages/suite/src/components/suite/Preloader/Preloader.tsx
@@ -23,6 +23,8 @@ import { selectDevice } from '@suite-common/wallet-core';
 import { DeviceCompromised } from '../SecurityCheck/DeviceCompromised';
 import { RouterAppWithParams } from '../../../constants/suite/routes';
 import { Feature, selectIsFeatureDisabled } from '@suite-common/message-system';
+import { WelcomeSidebar } from '../layouts/WelcomeLayout/WelcomeSidebar';
+import styled from 'styled-components';
 
 const ROUTES_TO_SKIP_REVISION_CHECK: RouterAppWithParams['app'][] = [
     'settings',
@@ -30,6 +32,11 @@ const ROUTES_TO_SKIP_REVISION_CHECK: RouterAppWithParams['app'][] = [
     'firmware-type',
     'firmware-custom',
 ];
+
+const LeftWrapper = styled.div`
+    display: flex;
+    flex-direction: row;
+`;
 
 const getFullscreenApp = (route: AppState['router']['route']): FC | undefined => {
     switch (route?.app) {
@@ -147,7 +154,14 @@ export const Preloader = ({ children }: PreloaderProps) => {
 
     // if a device is not connected or initialized
     if (isLoggedOut) {
-        return <LoggedOutLayout>{children}</LoggedOutLayout>;
+        if (router.route?.app === 'settings') {
+            return (
+                <LeftWrapper>
+                    <WelcomeSidebar />
+                    <LoggedOutLayout>{children}</LoggedOutLayout>
+                </LeftWrapper>
+            );
+        } else return <LoggedOutLayout>{children}</LoggedOutLayout>;
     }
 
     // everything is set.
--- a/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/PageNames/SettingsName.tsx
+++ b/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/PageNames/SettingsName.tsx
@@ -1,13 +1,11 @@
 import { useState } from 'react';
 import styled from 'styled-components';
 import { desktopApi } from '@trezor/suite-desktop-api';
-import { IconButton } from '@trezor/components';
 import { spacingsPx } from '@trezor/theme';
 import { setDebugMode } from 'src/actions/suite/suiteActions';
 import { Translation } from 'src/components/suite';
 import { useDispatch, useSelector } from 'src/hooks/suite';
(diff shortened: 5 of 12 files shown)


packages/components/src/components/buttons/IconButton/IconButton.tsx
```diff
@@ -75,10 +78,13 @@ export const IconButton = ({
                 disabled={isDisabled || isLoading}
                 onClick={handleClick}
                 $isSubtle={isSubtle}
+                $borderRadius={rest.borderRadius}
```
jvaclavik, line 81: We should discuss with Ondra, because it doesn't look like typical IconButton component and at the same time we don't want to allow too much of variety for single use cases.

I checked the design and it looks like it's nav item which we have in the sidebar menu. I think we should create one component for these two use cases.

packages/components/src/components/buttons/buttonStyleUtils.ts
```diff
@@ -7,10 +7,9 @@ import { hexToRgba } from '@suite-common/suite-utils';
 const SUBTLE_ALPHA = 0.12;
 const SUBTLE_ALPHA_HOVER = 0.2;
 
-export type ButtonVariant = Extract<
-    UIVariant,
-    'primary' | 'secondary' | 'tertiary' | 'info' | 'warning' | 'destructive'
->;
+export type ButtonVariant =
+    | Extract<UIVariant, 'primary' | 'secondary' | 'tertiary' | 'info' | 'warning' | 'destructive'>
+    | 'welcome';
```
jvaclavik, line 12: Please do not add non-standard UI variants 🙏

packages/suite/src/components/settings/SettingsLayout.tsx
```diff
@@ -16,7 +16,7 @@ type SettingsLayoutProps = {
     ['data-testid']?: string;
 };
 
-const SettingsHeader = () => {
+export const SettingsHeader = () => {
```
jvaclavik, line 19: Do you need to export it? Looks like you don't use it anywhere else.

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -31,6 +33,11 @@ const ROUTES_TO_SKIP_REVISION_CHECK: RouterAppWithParams['app'][] = [
     'firmware-custom',
 ];
 
+const LeftWrapper = styled.div`
+    display: flex;
+    flex-direction: row;
+`;
+
```
jvaclavik, line 40: Could you please use `<Row>...</Row>` from `@trezor/components` instead of this? It allows us to have less duplicit components

packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/PageNames/SettingsName.tsx
```diff
@@ -47,20 +44,8 @@ export const SettingsName = () => {
...
-
     return (
         <Container>
-            {isLoggedOut && (
-                <IconButton
-                    icon="ARROW_LEFT"
-                    variant="tertiary"
-                    size="medium"
-                    onClick={handleBackClick}
-                    data-testid="@settings/menu/close"
```
jvaclavik, line 60: Btw this might break some e2e tests

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/TorContainerWithCheckIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+        return (
+            <TorToggleContainer
+                onMouseEnter={onTorContainerChange(true)}
+                onMouseLeave={onTorContainerChange(false)}
+                onFocus={onTorContainerChange(true)}
+                onBlur={onTorContainerChange(false)}
+            >
+                {children}
+                {isTorEnabled && (
+                    <CheckIcon isTorContainerHoveredOrFocused={isTorContainerHoveredOrFocused} />
```
jvaclavik, line 34: Could you be more specific about the name of the component? I don't understand what does it mean. 

nit: Also you use it only once, you can keet it in this file if it makes sense.

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/CheckIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+    left: 50%;
+    background: ${({ $background }) => $background};
+    border-radius: ${borders.radii.full};
+`;
+
+type CheckIconProps = {
+    isTorContainerHoveredOrFocused?: boolean;
+};
+
+export const CheckIcon = ({ isTorContainerHoveredOrFocused = false }: CheckIconProps) => {
```
jvaclavik, line 18: According to component name it looks very generic but according to props it looks like it's very specific. I suggest to rename it somehow to make it more readable what should happen here.

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/TorContainerWithCheckIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+                )}
+            </TorToggleContainer>
+        );
+    }
+
+    return <></>;
```
jvaclavik, line 40:
```suggestion
    return null;
```

packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeSidebar.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { selectIsDiscreteModeActive } from 'src/reducers/wallet/settingsReducer';
+import { TrafficLightOffset } from '../../TrafficLightOffset';
+import styled from 'styled-components';
+import { borders, spacingsPx } from '@trezor/theme';
+import { Translation } from '../../Translation';
+import { useCallback, useEffect, useState } from 'react';
+import { SettingsAnchor } from 'src/constants/suite/anchors';
+import { selectTorState } from 'src/reducers/suite/suiteReducer';
+import { useLocation } from 'react-router-dom';
+import { ButtonVariant } from '@trezor/components/src/components/buttons/buttonStyleUtils';
```
jvaclavik, line 14: Could you use just simple package here? 
`import ... from '@trezor/components'`

packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeSidebar.tsx
```diff
@@ -0,0 +1,155 @@
...
+    return (
+        <WelcomeWrapper>
+            <TrafficLightOffset>
+                <Column
+                    justifyContent="center"
+                    alignItems="center"
+                    height="100%"
+                    width={WELCOME_SIDEBAR_WIDTH}
+                >
+                    <LeftContainer>
```
jvaclavik, line 113: you are using `Column` here probably only for centering content but at the same time you use `LeftContainer` which is almost the same equivalent. Could you please merge it? Ideally to `Column`

packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeSidebar.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+const LeftContainer = styled.div`
+    padding: ${spacingsPx.xs};
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    height: 100%;
+`;
+
+const LeftWrapper = styled.div`
```
jvaclavik, line 40: Could you use `Column` instead of this component? 🙏

packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeSidebar.tsx
```diff
@@ -0,0 +1,155 @@
...
+`;
+
+type ExtraIconButtonProps = {
+    variant: ButtonVariant;
+    borderRadius: string;
+    width: number;
+    height: number;
+};
+
+const extraIconButtonProps = (): ExtraIconButtonProps => {
```
jvaclavik, line 53: I think this doesn't have to be a function, just normal const is fine.

packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeSidebar.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const location = useLocation();
+
+    const handleDiscreetModeClick = () => dispatch(setDiscreetMode(!isDiscreetModeActive));
+
+    const handleClickTrezor = () => {
+        setActivePage('trezor');
+        dispatch(goto('suite-start'));
+        const updatedUrl = removeSettingsFromUrl(window.location.pathname);
+        window.location.pathname = updatedUrl;
+        location.pathname = removeSettingsFromUrl(location.pathname);
```
jvaclavik, line 78: I think we shouldn't handle pathname this way. It's potentially dangerous and hard to understand why we do it. Couldn't we switch pages according to `activePage`?

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/CheckIcon.tsx
hstastna: Yes, of course. I just left the original name here for now, while figuring out the better name for the component.

packages/components/src/components/buttons/IconButton/IconButton.tsx
hstastna: Those 2 use cases - using the component in the sidebar and in the new welcome screen - use different styles (there are more differences), and the related component NavigationItem that seemed that it could be used for both is not ready for it. It was very ineffective to use it here, lots of differences with styling, colors, details, etc. With `IconButton` it's less code changes, more readable code, even if still not ideal.

I am open to better solutions but from what I can see now, we could:
- change the proposed design (can take time, not sure with Ondra's availability) so the actual components we have could be nicely used, or
- refactor NavigationItem so it could be easily used here too (can take some time, can potentially break what's used in the sidebar already), or
- create and use a totally new component for usage in Welcome screen (less risky and faster, however lots of code similar to IconButton or to NavigationItem anyway), or
- allow overwriting NavigationItem component's style (we don't want, but it's faster and less risky than refactoring or waiting for design changes), or
- to find some other solution I have no idea about yet.

packages/components/src/components/buttons/buttonStyleUtils.ts
hstastna: I knew you won't like it :) This button variant is really non-standard. The proposed style of the buttons/navigation items in the welcome screen isn't used anywhere else yet, unfortunately, even if it is similar to the style of nav item.

The design already exists for some time so I expected I could rely on it. Seemed that we really wanted a specific style for Welcome screen. I know this change here is kinda problematic also because we want to divide components into general ones and Suite specific ones. And this change violates this effort. However trying to use NavigationItem instead of button produced even more problems.. 

It seems to me some of our existing components lack more variety, properties, more functionality that prevents them to be nicely used for different screens and use cases without overwriting their styles or without using other not ideal ways of achieving the expected. And so the expected non-standard look is not easily achievable using our standard components.

packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeSidebar.tsx
hstastna: This is because of we are working with the foreground app here and the url isn't updated accordingly, We must do it manually to make things work as expected. Or to get rid of the foreground app. Or I can omit these lines but it's not gonna work well for some scenarios, if we don't find some other solution.

Working with `window` here is mainly for making it all work for web Suite also. Otherwise we would be fine with using just `useLocation` hook. 

Gonna continue looking into this, looking for some better way to handle all we need here.

packages/components/src/components/buttons/buttonStyleUtils.ts
hstastna: We should start another effort - to improve the existing components so they could be easily adaptable to what we want to achieve.

hstastna:
Didn't have chance to finish this task, closing the PR.

PR #13658, opened 2024-08-06
chore(suite): tpm visible only in debug

mroz22:
As per @MiroslavProchazka request. TPM is now visible only in debug. I did it in the simplest and fastest way possible and I guess it will not be exactly to @marekrjpolak taste. So we might want to revisit it later

image

image


marekrjpolak:
@mroz22 I believe @peter-sanderson dealt with similar thing in silent update PR? And while I had some issues with other things, this made sense there.

MiroslavProchazka (approved):

PR #13661, opened 2024-08-06
feat(mobile): graph for erc20 token prices

Nodonisko:
## Related Issue

Resolve 

## Screenshots:
Simulator Screenshot - iPhone 14 Pro - 2024-08-06 at 10 15 28


suite-common/graph/src/graphBalanceEvents.ts
```diff
@@ -7,12 +7,14 @@ import { NetworkSymbol, getNetworkType } from '@suite-common/wallet-config';
...
 import { AccountBalanceHistory as AccountMovementHistory } from '@trezor/blockchain-link';
 import { fetchAllTransactionsForAccountThunk } from '@suite-common/wallet-core';
+import { AccountKey, TokenAddress } from '@suite-common/wallet-types';
 
 import {
     BalanceMovementEvent,
     GroupedBalanceMovementEvent,
     AccountItem,
     AccountHistoryMovementItem,
+    AccountHistoryMovement,
```
vytick: not used

suite-native/graph/src/components/TransactionEventTooltip.tsx
```diff
@@ -1,15 +1,26 @@
...
 import { GroupedBalanceMovementEventPayload } from '@suite-common/graph';
 import { EventTooltipComponentProps } from '@suite-native/react-native-graph/src/LineGraphProps';
 import { SignValue } from '@suite-common/suite-types';
-import { NetworkSymbol } from '@suite-common/wallet-config';
+import { NetworkSymbol, getNetworkType } from '@suite-common/wallet-config';
+import { AccountKey, TokenAddress, TokenSymbol } from '@suite-common/wallet-types';
+import { AccountsRootState } from '@suite-common/wallet-core';
+import {
+    selectEthereumAccountTokenInfo,
+    selectEthereumAccountTokenSymbol,
```
vytick, line 22: not used

suite-native/graph/src/components/TransactionEventTooltip.tsx
```diff
@@ -71,8 +138,19 @@ export const TransactionEventTooltip = ({
         sent,
         receivedTransactionsCount,
         sentTransactionsCount,
+        tokenAddress,
+        accountKey,
     },
 }: TransactionEventTooltipProps) => {
+    console.log({
```
vytick: please, remove console.log

suite-native/module-accounts-management/src/components/AccountDetailCryptoValue.tsx
```diff
@@ -1,26 +1,47 @@
...
         <HStack spacing="small" flexDirection="row" alignItems="center" justifyContent="center">
-            <CryptoIcon symbol={networkSymbol} size="extraSmall" />
-            <CryptoAmountFormatter
-                value={value}
-                network={networkSymbol}
-                isBalance={isBalance}
-                adjustsFontSizeToFit
-            />
+            <CryptoIcon symbol={tokenAddress || networkSymbol} size="extraSmall" />
+            {console.log(decimals)}
```
vytick: console.log

vytick (commented):
Few nits:

suite-native/module-accounts-management/src/components/AccountDetailCryptoValue.tsx
Nodonisko: Fixed everything. I forget to push latest version, sorry.

suite-common/graph/src/balanceHistoryUtils.ts
```diff
@@ -11,11 +11,17 @@ https://github.com/trezor/blockbook/blob/0afcb3a002e9f818907f2e1398f5c2af72c9f1a
 
 type BlockTime = number;
 
+type GetAccountHistoryMovementItemParams = {
+    transactions: WalletAccountTransaction[];
+    from?: number;
```
vytick, line 16: nit(feel free to discard):

```suggestion
since?: number;
until?: number
```

vytick (approved):
LGTM 🥇

suite-common/graph/src/balanceHistoryUtils.ts
```diff
@@ -93,6 +103,10 @@ const getAccountHistoryMovementItemRipple = ({
             return;
         }
 
+        if ((from && blockTime < from) || (to && blockTime > to)) {
```
PeKne, line 106: Isn't this unnecessary code duplication? What about filtering the transaction only once in `getAccountHistoryMovementFromTransactions` above the switch instead of doing it in each network specific function separately?

some could be done for transactions where `!blockTime` = `false`.

suite-common/graph/src/graphBalanceEvents.ts
```diff
@@ -130,11 +144,17 @@ export const getAccountMovementEvents = async ({
...
                 transactions: allTransactions,
                 coin,
-            }).main;
+            });
+
+            if (tokenAddress) {
+                return movements.tokens[tokenAddress] ?? [];
+            }
+
+            // Return only main because we don't support tokens in that popups in graph yet
```
PeKne: This comment does not apply anymore

suite-common/graph/src/graphBalanceEvents.ts
```diff
@@ -119,7 +132,8 @@ export const getAccountMovementEvents = async ({
     endOfTimeFrameDate: Date;
     dispatch: ReturnType<typeof useDispatch>;
 }) => {
-    const { coin, identity, descriptor } = account;
+    const { coin, identity, descriptor, tokensFilter, accountKey } = account;
+    const tokenAddress = tokensFilter?.[0]; // This is only for graph on detail screen where we have always only one token
```
PeKne, line 136: How is it ensured that there will be only one token? Could not happen that ETH mainnet detail will show graph for the "first" token instead of ETH?

suite-native/graph/src/components/TransactionEventTooltip.tsx
```diff
@@ -19,6 +27,8 @@ type EventTooltipRowProps = {
     signValue: SignValue;
     value: number;
     networkSymbol: NetworkSymbol;
+    tokenAddress?: TokenAddress;
+    accountKey: AccountKey;
```
PeKne: nit: optional should be at the end.
```suggestion
    accountKey: AccountKey;
    tokenAddress?: TokenAddress;
```

suite-native/graph/src/components/TransactionEventTooltip.tsx
```diff
@@ -44,20 +54,74 @@ const TooltipCardStyle = prepareNativeStyle(utils => ({
...
+    tokenAddress: TokenAddress;
+    networkSymbol: NetworkSymbol;
+    value: number;
+}) => {
+    const tokenInfo = useSelector((state: AccountsRootState) =>
+        selectEthereumAccountTokenInfo(state, accountKey, tokenAddress),
+    );
+
+    const tokenSymbol = tokenInfo?.symbol?.toUpperCase() as TokenSymbol;
```
PeKne: nit `selectEthereumAccountTokenSymbol` might be used here.

suite-native/graph/src/components/TransactionEventTooltip.tsx
```diff
@@ -44,20 +54,74 @@ const TooltipCardStyle = prepareNativeStyle(utils => ({
...
+    const tokenDecimals = tokenInfo?.decimals;
+
+    if (!tokenSymbol || !tokenDecimals) {
+        return null;
+    }
+
+    // We might want to add support for other networks in the future.
+    if (getNetworkType(networkSymbol) === 'ethereum') {
```
PeKne, line 84: nit: What about placing this condition to the preceding if statement like:
```
  if (!tokenSymbol || !tokenDecimals || !getNetworkType(networkSymbol) === 'ethereum'){
    return null
  }
```

Then you could remove this `if` statemenet and also delete the `return null` at the end of this function.

suite-native/module-accounts-management/src/components/AccountDetailCryptoValue.tsx
```diff
@@ -1,26 +1,46 @@
...
+                    symbol={tokenSymbol}
+                    adjustsFontSizeToFit
+                />
+            ) : (
+                <CryptoAmountFormatter
+                    value={value}
+                    network={networkSymbol}
+                    isBalance={isBalance}
+                    adjustsFontSizeToFit
+                    decimals={decimals}
```
PeKne: Why are the decimals needed now? It is passed only to the CryptoAmountFormatter that is responsible for mainnet values. Inside of this component are  decimals already handled. What is the purpouse of passing it here again?

Should not be this value passed rather to the `EthereumTokenAmountFormatter`?

suite-native/module-accounts-management/src/components/AccountDetailGraphHeader.tsx
```diff
@@ -63,17 +79,36 @@ const FiatBalance = () => {
...
+        if (!tokenAddress || !account) return null;
+
+        // We might want to add support for other networks in the future.
+        if (getNetworkType(account.symbol) === 'ethereum') {
+            return selectEthereumAccountTokenInfo(state, accountKey, tokenAddress);
+        }
+
+        return null;
+    });
+    const tokenSymbol = tokenInfo?.symbol?.toUpperCase() as TokenSymbol;
```
PeKne: `selectEthereumAccountTokenSymbol`

suite-native/module-accounts-management/src/components/AccountDetailGraphHeader.tsx
```diff
@@ -63,17 +79,36 @@ const FiatBalance = () => {
     );
 };
 
-export const AccountDetailGraphHeader = ({ accountKey }: AccountBalanceProps) => {
+export const AccountDetailGraphHeader = ({ accountKey, tokenAddress }: AccountBalanceProps) => {
     const account = useSelector((state: AccountsRootState) =>
         selectAccountByKey(state, accountKey),
     );
+
+    const tokenInfo = useSelector((state: AccountsRootState) => {
```
PeKne: nit: What about creating a new selector of this and move it to some `...Selectors.ts` file?

suite-native/module-accounts-management/src/components/AccountDetailTokenHeader.tsx
```diff
@@ -16,7 +16,7 @@ import { AccountKey, TokenAddress } from '@suite-common/wallet-types';
 
 type AccountDetailTokenHeaderProps = {
     accountKey: AccountKey;
-    tokenContract: TokenAddress;
+    tokenContract?: TokenAddress;
```
PeKne: Technically correct, but kind of weird that token detail might have undefined `tokenContract`

suite-native/module-accounts-management/src/components/AccountDetailTokenHeader.tsx
```diff
@@ -16,7 +16,7 @@ import { AccountKey, TokenAddress } from '@suite-common/wallet-types';
...
 };
 
 export const AccountDetailTokenHeader = ({
```
PeKne: Is this component used anywhere now? It was removed from `TransactionListHeader` file and I can not find any other reference of it.

If it is unused, remove it please.

PeKne (commented):
Some questions,but generally looks good. GJ delivering this!

PeKne (changes requested):
.

suite-common/graph/src/balanceHistoryUtils.ts
Nodonisko: I agree this is better probably, but want to follow same API as blockbook

suite-common/graph/src/balanceHistoryUtils.ts
Nodonisko: It could be done that way but in the end I decided to do it for each function separately because it could save some work and calculations with BigNumber are quite slow. I am not sure about how big impact could be but since it's very simple few lines of code I don't mind some minor duplication.

suite-common/graph/src/graphBalanceEvents.ts
Nodonisko: There `tokenFilter = []` right now for mainnet so it will show only ETH without tokens, but I will address showing tokens movements in ETH graph in next PR.

suite-native/module-accounts-management/src/components/AccountDetailCryptoValue.tsx
Nodonisko: Sadly it doesn't work because when you use token balance from redux it's already formatted, but if you do calculations locally there are no decimals so you need to add them later.

This kind of sucks, but this is probably best solution that doesn't require big refactors.

suite-native/module-accounts-management/src/components/AccountDetailGraphHeader.tsx
Nodonisko: It would require new memoized selector but there is problem that memoization using proxy memoize is slow and expensive (my bad for choosing that lib, I want to write my own faster memo lib but not enough time 😅)

So this will actually call that expensive selector only when necessary.

suite-native/module-accounts-management/src/components/AccountDetailTokenHeader.tsx
Nodonisko: File is not used anymore, so I will delete it.

suite-native/graph/src/components/TransactionEventTooltip.tsx
Nodonisko: I think this version is better because there might be more than ETH tokens so you will only add more `else if` branches for that.

suite-native/module-accounts-management/src/components/AccountDetailTokenHeader.tsx
Nodonisko: deleted

suite-native/module-accounts-management/src/components/AccountDetailCryptoValue.tsx
PeKne: I still don't understand. This PR should only extend the functionality by displaying graph for token details, right?

So why is modified usage of `CryptoAmountFormatter` that should be rendered only on the mainnet (ETH, BTC etc.) screens? In case of token detail screen, only the `EthereumTokenAmountFormatter` should be rendered right?

So my question is: "Why is needed modification of the original functionality?"

Nodonisko: You are right. In the end I removed it entirely in 19fd4da132 because it now seems to work everywhere without it.

PeKne (approved):

Issue #4786, opened 2022-01-21
@trezor/utils package improvements


karliatto:
With the PR https://github.com/trezor/trezor-suite/pull/4666 we are adding new package `@trezor/utils`. The purpose of this issue is to track potential improvements to that package and discuss whether some utilities make sense to be moved there or not.

## Potential candidates for further improvements

- [ ] trezor-suite/packages/suite/src/utils/suite/parseUri.ts
- [x] trezor-suite/packages/suite/src/utils/wallet/validation.ts
- [x] trezor-suite/packages/suite/src/utils/wallet/promiseUtils.ts
- [x] trezor-suite/packages/suite/src/utils/wallet/ethUtils.ts
- [x] trezor-suite/packages/suite/src/utils/wallet/csvParser.ts
- [x] trezor-suite/packages/rollout/src/utils/version.ts #4870
- [x] trezor-suite/packages/transport/src/utils/semver-compare.ts #4897 
- [x] trezor-suite/packages/utxo-lib/src/crypto.ts
- [x] trezor-suite/packages/utxo-lib/src/bufferutils.ts
- [ ] connect-explorer/src/js/components/Response.tsx
- [ ] packages/transport/src/bridge/http.ts or parts of it to have a generic http utility
- [x] connect/deferred https://github.com/trezor/trezor-suite/pull/9353
- [x] connect/utils

szymonlesisz:
typescript utilities:
- [ ] trezor-suite/packages/suite/src/types/utils
- [ ] trezor-suite/packages/suite-desktop-api/src/methods
- [ ] connect/types/utils

mroz22:
@karliatto could you please go through the list once again? it is now pretty outdated. we could then refactor couple of those remaining and close this issue I guess

karliatto:
I think most of them have been moved to `@trezor/utils` or `@trezor/wallet-utils` or `@trezor/node-utils` with some exception that are just used in some package and probably does not make sense to move it to a more generic utils package at least until it is not required to be use somewhere else.

I moved this last one https://github.com/trezor/trezor-suite/pull/15856 and would I marked to resolve this issue done after that one is merged.

PR #13662, opened 2024-08-06
fix(suite-desktop-core): use checkFileExits form node-utils

karliatto:
## Description

Use `checkFileExists` from `@trezor/node-utils`.

## Related Issue

https://github.com/trezor/trezor-suite/issues/4786


mroz22 (approved):
looks ok

PR #13663, opened 2024-08-06
fix(suite): Fix QR code background in dark mode

### LINKED ISSUE #13083: QR code is inverted in dark mode
**Describe the bug**
When the app is in dark mode, the QR code is shown with inverted colors.
This is non-standard and while _some_ QR scanners can deal, not all of them, because they expect dark-on-light scheme.

**Info:**

-   Suite Version: web 24.6.2

**How to reproduce**

1. enable dark mode
2. go to a Receive tab
3. show an address + QR code

**Expected behavior**
Dark-mode QR code should not be inverted. It should be the same as light-mode QR code, surrounded by at least one cell wide white border.

**Screenshots**
image image

**Additional context**
reported on the forum: https://forum.trezor.io/t/qr-code-cannot-be-scanned-in-dark-mode-both-trezor-suite-desktop-and-web-based-are-affected/18111/2

Lemonexe:
## Description

There are two occurrences of `QrCode` component in suite:
- `PromoBanner` uses default `QrCode` colors, works as intended
- `ConfirmValueModal` used colors from theme to override `QrCode`, will use constant colors instead

## Related Issue

Resolve #13083

## Screenshots:

Dark mode before & after:

### ConfirmValueModal _unconfirmed_
image
image

### ConfirmValueModal _confirmed_
image
image

### PromoBanner
image
image

### DIFF
--- a/packages/suite/src/components/suite/modals/ReduxModal/ConfirmValueModal/ConfirmValueModal.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/ConfirmValueModal/ConfirmValueModal.tsx
@@ -1,5 +1,5 @@
 import { useEffect, ReactNode } from 'react';
-import styled, { useTheme } from 'styled-components';
+import styled from 'styled-components';
 
 import { notificationsActions } from '@suite-common/toast-notifications';
 import { Button, ConfirmOnDevice, ModalProps, Tooltip, variables } from '@trezor/components';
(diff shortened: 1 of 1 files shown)


packages/suite/src/components/suite/modals/ReduxModal/ConfirmValueModal/ConfirmValueModal.tsx
```diff
@@ -135,6 +133,9 @@ export const ConfirmValueModal = ({
         }
     }, [canConfirmOnDevice, dispatch, isConfirmed, modalContext, validateOnDevice]);
 
+    // light theme colors are used always for QR code
+    const qrCodeFgColor = addressConfirmed ? colorVariants.standard.iconDefault : colorVariants.standard.iconSubdued
```
jvaclavik: what about using color from `palette` directly? 

e.g.`palette.darkGray1000`?

Lemonexe: Done (though `lightGray1000` is the appropriate one)

jvaclavik (approved):
image

MiroslavProchazka (approved):
Ship it

PR #13669, opened 2024-08-06
chore(suite): don't run second process with autostart

martykan:
## Description

This changes the way the UI is launched when using autostart. 

Before this change, the Node bridge daemon and the UI would run on 2 separate processes independently, which wastes some memory.

After this change, only 1 process runs. When a start of a second process is attempted, the first process continues to initialize the UI if running in daemon mode. If the UI was already initialized, it will resume. 
Upon closing the app doesn't quit, but hides the UI to keep the node bridge running. 

Core functionality tested on MacOS, Win, Linux, but needs more thorough testing.
I did also test auto update on MacOS and it was working.

### DIFF
--- a/packages/suite-desktop-core/src/app.ts
+++ b/packages/suite-desktop-core/src/app.ts
@@ -22,6 +22,7 @@ import { init as initBridgeModule } from './modules/bridge';
 import { createInterceptor } from './libs/request-interceptor';
 import { hangDetect } from './hang-detect';
 import { Logger } from './libs/logger';
+import { MainWindowProxy } from './libs/main-window-proxy';
 
 // @ts-expect-error using internal electron API to set suite version in dev mode correctly
 if (isDevEnv) app.setVersion(process.env.VERSION);
@@ -170,24 +171,32 @@ const initUi = async ({
     const winBounds = store.getWinBounds();
     logger.debug('init', `Create Browser Window (${winBounds.width}x${winBounds.height})`);
 
-    const mainWindow = createMainWindow(winBounds);
-
-    app.on('second-instance', () => {
-        // Someone tried to run a second instance, we should focus our window.
-        app.dock?.show();
-        app.show();
-        if (!mainWindow.isVisible()) mainWindow.show();
-        if (mainWindow.isMinimized()) mainWindow.restore();
-        mainWindow.focus();
-    });
+    const mainWindowProxy = new MainWindowProxy();
 
     // init modules
     const interceptor = createInterceptor();
     const { loadModules, quitModules } = initModules({
-        mainWindow,
+        mainWindowProxy,
         store,
         interceptor,
         mainThreadEmitter,
+    });
+
+    app.on('second-instance', () => {
+        // Someone tried to run a second instance, we should focus our window.
+        logger.info('main', 'Second instance detected, focusing main window');
+        let mainWindow = mainWindowProxy.getInstance();
+        if (!mainWindow || mainWindow.isDestroyed()) {
+            logger.info('main', 'Main window destroyed, recreating');
+            mainWindow = createMainWindow(winBounds);
+            mainWindowProxy.setInstance(mainWindow);
+        }
+
+        app.dock?.show();
+        if (isMacOs()) app.show();
+        if (!mainWindow.isVisible()) mainWindow.show();
+        if (mainWindow.isMinimized()) mainWindow.restore();
+        mainWindow.focus();
     });
 
     // create handler for handshake/load-modules
@@ -212,7 +221,7 @@ const initUi = async ({
     // Tor module initializes separated from general `initModules` because Tor is different
     // since it is allowed to fail and then the user decides whether to `try again` or `disable`.
     const { onLoad: loadTorModule, onQuit: quitTorModule } = initTorModule({
-        mainWindow,
+        mainWindowProxy,
         store,
         interceptor,
         mainThreadEmitter,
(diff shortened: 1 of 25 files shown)


packages/suite-desktop-core/src/app.ts
```diff
@@ -158,7 +170,14 @@ const init = async () => {
         mainWindow.focus();
     });
 
-    app.on('before-quit', () => {
+    app.on('before-quit', event => {
```
mroz22: prerequisite https://github.com/trezor/trezor-suite/pull/13847

packages/suite-desktop-core/src/app.ts
```diff
@@ -128,18 +130,38 @@ const init = async () => {
 
     await app.whenReady();
 
-    // No UI mode with bridge only
+    // Load bridge module first, it is required in both UI and daemon mode
+    const { onLoad: loadBridgeModule, onQuit: quitBridgeModule } = initBridgeModule({ store });
+    await loadBridgeModule();
```
marekrjpolak, line 135: I'm just thinking whether it's ok to start Bridge module long before Tor module and request interceptor module.It probably is because it's only server-side code accesed only by localhost so there's nothing to intercept or route via Tor?

packages/suite-desktop-core/src/app.ts
```diff
@@ -128,18 +130,38 @@ const init = async () => {
...
+
+    // Daemon mode with no UI
     const { wasOpenedAtLogin } = app.getLoginItemSettings();
-    if (app.commandLine.hasSwitch('bridge-daemon') || wasOpenedAtLogin) {
+    const daemon = app.commandLine.hasSwitch('bridge-daemon') || wasOpenedAtLogin;
+    if (daemon) {
         logger.info('main', 'App is hidden, starting bridge only');
         app.dock?.hide(); // hide dock icon on macOS
-        app.releaseSingleInstanceLock(); // allow user to open new instance with UI
```
marekrjpolak, line 136: `releaseSingleInstanceLock()` not needed anymore?

packages/suite-desktop-core/src/app.ts
marekrjpolak: Also, what happens in ordinary run when Bridge succeeds to start e.g. on third trial, after 8 seconds, or not at all? Users won't see any UI until that?

packages/suite-desktop-core/src/hang-detect.ts
```diff
@@ -1,6 +1,7 @@
 import { BrowserWindow, dialog } from 'electron';
 
 import { validateIpcMessage } from '@trezor/ipc-proxy';
+import { ElectionIpcMainInvokeEvent } from '@trezor/ipc-proxy/src/proxy-handler';
```
marekrjpolak: Election what? :nauseated_face:  Cc: @peter-sanderson
image

packages/suite-desktop-core/src/app.ts
martykan: I believe bridge doesn't access the internet so Tor shouldn't be needed. 
It's true that if Bridge has some issue starting, it can delay the launch of the app. There is a timeout of 3 seconds and errors are silent, so it should start after that.

packages/suite-desktop-core/src/app.ts
martykan: Yes, this is intentional, we want only 1 process

packages/suite-desktop-core/src/hang-detect.ts
```diff
@@ -40,17 +43,22 @@ export const hangDetect = (
...
             clearTimeout(timeout);
-            // always resolve repeated handshakes from renderer (e.g. Ctrl+R)
-            ipcMain.handle('handshake/client', ipcEvent => {
-                validateIpcMessage(ipcEvent);
-
-                return Promise.resolve({});
-            });
+            try {
+                // always resolve repeated handshakes from renderer (e.g. Ctrl+R)
+                ipcMain.handle('handshake/client', handshakeHandler);
```
marekrjpolak, line 48: Now it's possible to enter this code block more than once? The logic behind it was that the main-side handshake runs only once (therefore handleOnce) while renderer-side handshakes are sent on every Ctrl+R and main just returns empty response. Is this still valid?

packages/suite-desktop-core/src/app.ts
marekrjpolak: Oh ok, I thought there's also 3 attempts. So the worst case is waiting 3 seconds for UI, right?

packages/suite-desktop-core/src/hang-detect.ts
martykan: Yes, for a new window the URL is not initially set, so we need to load the page through this

packages/suite-desktop-core/src/app.ts
martykan: Yes I think so

packages/suite-desktop-core/src/hang-detect.ts
peter-sanderson: :scream:

mroz22: 🏴 🟨

mroz22 (approved):

evgenysl:
Tested on Windows and Mac (both Intel and ARM)

Trezor Suite process is running on background as expected after PC restart:

suite background


Bridge is running and device is recognized.

bridge running

Bridge doesn't work in Firefox now, this will be fixed in new Connect version. 

I can start Trezor Suite app without any issues:

suite running

If I close Trezor Suite, process is still running on background :white_check_mark:

PR #13675, opened 2024-08-06
Chore/fix url type

komret:
## Description

Dynamic Urls must be typed `as const`, otherwise the `Url` type changes to `string`.

The tighter type helped me found a bug where an href was set to the URL key instead of value. This can be tested with address `1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa`. on Bitcoin Cash.

## Screenshots
Screenshot 2024-08-06 at 19 24 36
before:
Screenshot 2024-08-06 at 19 25 27

after:
Screenshot 2024-08-06 at 19 24 44

### DIFF
--- a/packages/suite/src/components/wallet/InputError.tsx
+++ b/packages/suite/src/components/wallet/InputError.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components';
 
 import { Button } from '@trezor/components';
 import { LearnMoreButton } from '../suite/LearnMoreButton';
+import { Url } from '@trezor/urls';
 
 const Wrapper = styled.div`
     display: flex;
@@ -11,7 +12,7 @@ const Wrapper = styled.div`
 `;
 
 interface InputErrorProps {
-    button?: { onClick: MouseEventHandler<HTMLButtonElement>; text: string } | { url: string };
+    button?: { onClick: MouseEventHandler<HTMLButtonElement>; text: string } | { url: Url };
     message?: string;
 }
 
--- a/packages/suite/src/components/wallet/InputError.tsx
+++ b/packages/suite/src/components/wallet/InputError.tsx
@@ -4,17 +4,21 @@ import styled from 'styled-components';
 import { Button } from '@trezor/components';
 import { LearnMoreButton } from '../suite/LearnMoreButton';
 import { Url } from '@trezor/urls';
+import { spacingsPx } from '@trezor/theme';
 
 const Wrapper = styled.div`
     display: flex;
     align-items: center;
-    gap: 8px;
+    gap: ${spacingsPx.xs};
 `;
 
-interface InputErrorProps {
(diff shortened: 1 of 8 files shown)


peter-sanderson (approved):

peter-sanderson (approved):

peter-sanderson (approved):

packages/suite/src/components/wallet/InputError.tsx
```diff
@@ -11,7 +12,7 @@ const Wrapper = styled.div`
 `;
 
 interface InputErrorProps {
-    button?: { onClick: MouseEventHandler<HTMLButtonElement>; text: string } | { url: string };
+    button?: { onClick: MouseEventHandler<HTMLButtonElement>; text: string } | { url: Url };
```
peter-sanderson: This is a bit funny. 

Why not just:

```suggestion
    button?: { onClick: MouseEventHandler<HTMLButtonElement>; text: string, url: Url };
```

But its a nit.

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
```diff
@@ -129,17 +127,19 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
     }, [amountInputName, composeTransaction, dispatch, inputName, setValue, symbol]);
 
     const getValidationButtonProps = ():
-        | { url: AddressDeprecatedUrl }
+        | { url: (typeof URLS)[NonNullable<ReturnType<typeof isAddressDeprecated>>] }
```
peter-sanderson: nit: but I suggest t export the type and make it just one object (remove the unnecessary union `|`)

peter-sanderson (approved):

packages/suite/src/views/wallet/send/components/Outputs/components/Address/index.tsx
komret: ea22b6a95f4d1a2f63cc7e0f44918313274bb183

packages/suite/src/components/wallet/InputError.tsx
komret: It is intended because you either want a URL or a click handler, never both at the same time. I made this more apparent in ea22b6a95f4d1a2f63cc7e0f44918313274bb183.

peter-sanderson: Sorry my bad, I had `&` in my mind instead of `|` for some reason. This makes sense, :+1:

bosomt:
**QA OK**

link is https://trezor.io/learn/a/glossary#CashAddr

**Info:**
 - Suite version: web 24.8.0 (55ddef6b9eab7465d26315474133a08e79071e95)
 - OS: MacIntel
 - Screen: 1512x982

PR #13681, opened 2024-08-06
feat(suite-native): account purge and discovery on coin enabling

### LINKED ISSUE #13547: Coin Enabling - implement settings screen update logic
- [x] When leaving screen (back press or swipe) run discovery for enabled networks
- [x] disabled coin accounts are removed from current and view only devices

vytick:
- [x] When leaving Coin enabling screen in settings, run discovery for enabled networks
- [x] disabled coin accounts are removed from current and view only devices at the same moment

## Related Issue

Resolve #13547

## Screenshots:
https://github.com/user-attachments/assets/2edbd194-e26a-4b27-826d-a662769d658c

### DIFF
--- a/suite-native/discovery/src/discoveryConfigSlice.ts
+++ b/suite-native/discovery/src/discoveryConfigSlice.ts
@@ -72,44 +72,64 @@ export const discoveryConfigSlice = createSlice({
     },
 });
 
-export const selectDiscoverySupportedNetworks = memoizeWithArgs(
-    (state: DeviceRootState, areTestnetsEnabled: boolean) =>
-        pipe(
-            selectDeviceSupportedNetworks(state),
-            networkSymbols => filterTestnetNetworks(networkSymbols, areTestnetsEnabled),
-            filterUnavailableNetworks,
-            filterBlacklistedNetworks,
-            sortNetworks,
-        ),
-    // for all areTestnetsEnabled states
-    { size: 2 },
-);
-
 export const selectAreTestnetsEnabled = (state: DiscoveryConfigSliceRootState) =>
     state.discoveryConfig.areTestnetsEnabled;
 
 export const selectDiscoveryInfo = (state: DiscoveryConfigSliceRootState) =>
     state.discoveryConfig.discoveryInfo;
 
+export const selectDiscoverySupportedNetworks = memoizeWithArgs(
+    (
+        state: DeviceRootState & DiscoveryConfigSliceRootState,
+        forcedAreTestnetsEnabled?: boolean,
+    ) => {
+        const areTestnetsEnabled = forcedAreTestnetsEnabled ?? selectAreTestnetsEnabled(state);
+
+        return pipe(
+            selectDeviceSupportedNetworks(state),
+            networkSymbols => filterTestnetNetworks(networkSymbols, areTestnetsEnabled),
+            filterUnavailableNetworks,
+            filterBlacklistedNetworks,
+            sortNetworks,
+        );
+    },
+    // for all areTestnetsEnabled states
+    { size: 2 },
+);
+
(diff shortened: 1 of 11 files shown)


PeKne:
Should not be graph hidden while is the re-run of discovery active? Like this it might to lead a frequent graph flashing and poor UX.

https://github.com/user-attachments/assets/a26af4e5-bd48-4e4e-bba2-cc7d60cbb426

suite-native/module-settings/src/screens/SettingsCoinEnablingScreen.tsx
```diff
@@ -1,9 +1,17 @@
...
 import { Screen, ScreenSubHeader } from '@suite-native/navigation';
 import { useTranslate } from '@suite-native/intl';
-import { DiscoveryCoinsFilter } from '@suite-native/coin-enabling';
+import { DiscoveryCoinsFilter, useCoinEnabling } from '@suite-native/coin-enabling';
 
 export const SettingsCoinEnablingScreen = () => {
     const { translate } = useTranslate();
+    const { applyDiscoveryChanges } = useCoinEnabling();
+
+    useEffect(() => {
```
PeKne: this useEffect could be included inside of the `useCoinEnabling` hook. Then it won't be necessary to call this hook here at all, since it is already included in the `DiscoveryCoinsFilter` that is part of this component's JSX.

suite-native/discovery/src/discoveryConfigSlice.ts
```diff
@@ -91,27 +91,39 @@ export const selectAreTestnetsEnabled = (state: DiscoveryConfigSliceRootState) =
...
+
+        return getNetworkSymbols(supportedNetworks);
+    },
+    { size: 2 },
+);
+
 export const selectEnabledDiscoveryNetworkSymbols = memoizeWithArgs(
     (
         state: DiscoveryConfigSliceRootState & DeviceRootState & FeatureFlagsRootState,
         areTestnetsEnabled: boolean,
```
PeKne: Why is `areTestnetsEnabled` parameter needed at all? Could not be selected from the `discoveryConfig.areTestnetsEnabled` state directly?

Then I think there would be no need for new selector `selectAllEnabledDiscoveryNetworkSymbols` But it would get handled automatically. Or maybe not? Let's discuss.

PeKne (commented):
Two concers, otherwise GJ 👍.

suite-native/discovery/src/discoveryConfigSlice.ts
vytick: `areTestnetsEnabled` can be forced from `prepareDiscoveryMiddleware` through `startDescriptorPreloadedDiscoveryThunk` in case we are just changing whether testnets are enabled. 


I'll rewrite this so it is not needed, if want to follow what is in state already.

suite-native/module-settings/src/screens/SettingsCoinEnablingScreen.tsx
vytick: Good idea. Now when I was thinking about it, I would rather have this logic in `DiscoveryCoinsFilter` than in `useCoinEnabling`. It would mean, that `useCoinEnabling` can be used without this side effect.

Wdyt?

suite-native/discovery/src/discoveryConfigSlice.ts
vytick: bfcac19e96

suite-native/module-settings/src/screens/SettingsCoinEnablingScreen.tsx
vytick: bfcac19e96

PeKne (approved):

PR #13695, opened 2024-08-07
test(suite): Unify data-testid attrs for E2E

### LINKED ISSUE #13247: Unify data-test
use `data-test-id` everywhere instead of `data-test` and `dataTest`. Right now it's dataTest in DS components and `data-test` everywhere else

It needs to be re-configurated also in tests:
- cypress: `getTestElement` in `packages/suite-web/e2e/support/utils/selectors.ts`
- playwright: `testIdAttribute` in `packages/suite-desktop-core/e2e/playwright.config.ts`

Lemonexe:
## Description

For Cypress and Playwright E2E tests, do following:

- 93634092192824e34222c2a56601680bd848b1c7 refactor all usages of `data-test` to `data-testid`
- 15bb29c4bb96998ba959aa2f48a9dafb7fc53deb when the value is passed through props, refactor prop names from `dataTest` or such to  `data-testid`

## Related Issue

Resolve #13247

### DIFF
--- a/packages/components/src/components/typography/typography.stories.tsx
+++ b/packages/components/src/components/typography/typography.stories.tsx
@@ -10,11 +10,11 @@ export default meta;
 export const All: StoryFn = () => (
     <>
         <StoryColumn>
-            <H1 data-test="heading-1">Heading 1</H1>
-            <H2 data-test="heading-2">Heading 2</H2>
+            <H1 data-testid="heading-1">Heading 1</H1>
+            <H2 data-testid="heading-2">Heading 2</H2>
         </StoryColumn>
         <StoryColumn>
-            <Paragraph data-test="paragraph-default">
+            <Paragraph data-testid="paragraph-default">
                 default
                 <br />
                 Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec a ante quis lectus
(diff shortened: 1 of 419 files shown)


packages/components/src/components/typography/typography.stories.tsx
```diff
@@ -10,75 +10,75 @@ export default meta;
 export const All: StoryFn = () => (
     <>
         <StoryColumn>
-            <H1 data-test="heading-1">Heading 1</H1>
-            <H2 data-test="heading-2">Heading 2</H2>
+            <H1 data-test-id="heading-1">Heading 1</H1>
```
peter-sanderson: @jvaclavik why we have `data-test-id` in storybook?

Lemonexe: I noticed it too, and was perplexed because those ids in `*.stories.tsx`  are not referenced anywhere. Seems useless, but I refactored it anyway for consistency.

peter-sanderson (approved):
I went through it all, ooph :)

suite-native/atoms/src/types.ts
```diff
@@ -1,6 +1,6 @@
 export type TestProps = {
     ['data-testid']?: never;
```
peter-sanderson, line 2: Can we remove all those variants and use just one? @matejkriz What is the reason for having such many?

Lemonexe: Good question, though this diff was my mistake, I am reverting it because I just duplicated `'data-test-id'`

packages/components/src/components/typography/typography.stories.tsx
komret: It doesn't make sense to me, let's drop it.

mroz22: I think that long time ago there were cypress tests configured to run also for the storybook.

mroz22:
are there any automated checks to force this kind of rule?

Lemonexe:
> are there any automated checks to force this kind of rule?

@mroz22  Not atm, maybe I could see if I can implement a custom eslint rule to just ban `data-test` and similar variants other than `data-test-id`. Though I think it's out of the scope for this PR.

jvaclavik (approved):
Good job! Let's ship it 👍

suite-native/atoms/src/types.ts
PeKne: React Native works only with the `testID` variant and nothing else. This type ensures that the TS shouts to you if you use any other variant than `testID`.

But IMO it is wrong. Even `date-testID` should be defined as `never`.

PR #13696, opened 2024-08-07
chore(suite-data): extend browser support

### LINKED ISSUE #12486: "Your browser is not supported" modal/page improvements - Suite web

martykan:
## Description

Extend browser support to MS Edge, Brave, Opera in Suite, since they are Chromium based.

## Related Issue

Resolve #12486

### DIFF
--- a/packages/connect/src/data/config.ts
+++ b/packages/connect/src/data/config.ts
@@ -130,6 +130,12 @@ export const config = {
             download: 'https://www.microsoft.com/en-us/edge',
             update: 'https://www.microsoft.com/en-us/edge',
         },
+        opera: {
+            // Chromium based
+            version: 95,
+            download: 'https://www.opera.com/download',
+            update: 'https://www.opera.com/download',
+        },
     },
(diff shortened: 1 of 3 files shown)


komret (commented):
Shouldn't https://github.com/trezor/trezor-suite/blob/develop/packages/suite-build/browserslist be updated?

martykan:
Perhaps yes, but I think it shouldn't affect anything since they're using Chromium engines

komret (approved):

mroz22 (approved):

PR #13699, opened 2024-08-07
Feat/transport bridge protocol message

szymonlesisz:
## Description

Prerequisite for `THP` (protocol v2)

Goal:
To be able to pass additional data (protocol name + protocol state in the future) in `/call`, `/post` and `/read` requests.
This also simplifies workflow a little. `TrezordNode` doesnt have to translate from protocol-bridge to protocol-v1 (received data) and from protocol-v1 to protocol-bridge (sent data)

format:
```
BridgeProtocolMessage = {
    protocol: 'v1',
    data: '000000',
}
```

- `TrezordNode` 
  - accept and validate JSON body (BridgeProtocolMessage) in `/call`, `/post` and `/read` requests
  - send new flag `protocolMessages` in getInfo `/` request
  - `procolMessages` flag is enabled by default 

- `TransportBridge`
  - if possible (if bridgeInfo.protocolMessage == true) send and receive body as JSON (BridgeProtocolMessage)
  
 - `TrezorConnect`
   - use protocol v1 as default

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -1,7 +1,7 @@
 // original file https://github.com/trezor/connect/blob/develop/src/js/device/Device.js
 import { versionUtils, createDeferred, Deferred, TypedEmitter } from '@trezor/utils';
 import { Session } from '@trezor/transport';
-import { TransportProtocol, v1 as v1Protocol, bridge as bridgeProtocol } from '@trezor/protocol';
+import { TransportProtocol, v1 as v1Protocol } from '@trezor/protocol';
 import { DeviceCommands, PassphrasePromptResponse } from './DeviceCommands';
 import { PROTO, ERRORS, NETWORK } from '../constants';
 import { DEVICE, DeviceButtonRequestPayload, UI } from '../events';
@@ -154,11 +154,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
     constructor(transport: Transport, descriptor: Descriptor) {
         super();
 
-        if (transport.name === 'BridgeTransport') {
-            this.protocol = bridgeProtocol;
-        } else {
-            this.protocol = v1Protocol;
-        }
+        this.protocol = v1Protocol;
 
         // === immutable properties
         this.transport = transport;
--- a/packages/transport-bridge/src/http.ts
+++ b/packages/transport-bridge/src/http.ts
@@ -13,6 +13,7 @@ import {
     Response,
 } from '@trezor/node-utils';
 import { Descriptor, Session } from '@trezor/transport/src/types';
+import { validateProtocolMessage } from '@trezor/transport/src/utils/bridgeProtocolMessage';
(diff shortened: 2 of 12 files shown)


mroz22 (commented):
nice, it seems to work nicely, I played around with it a bit, it doesn't seem to be breaking any backwards compatibility too. 

image

We are preparing next connect release right now, so please lets a wait until its branched out and after that this can be continued with.

packages/transport-bridge/src/http.ts
```diff
@@ -13,6 +13,7 @@ import {
     Response,
 } from '@trezor/node-utils';
 import { Descriptor, Session } from '@trezor/transport/src/types';
+import { validateProtocolMessage } from '@trezor/transport/src/utils/bridgeProtocolMessage';
```
mroz22, line 16: note: definition is part of b59e92a07134c752bef96d8f647ccbe582d5a4ad and not this commit. no problem with that, but maybe we should squash those commits before merging

packages/transport-bridge/tests/http.test.ts
```diff
@@ -111,6 +111,309 @@ describe('http', () => {
         await anotherInstance.stop();
     });
 
+    describe('BridgeProtocolMessage', () => {
+        const GET_FEATURES = '000000000000'; // GetFeatures message-in
```
mroz22: this is Initialize, not GetFeatures. am I nitpicking? :D

packages/connect/src/device/Device.ts
```diff
@@ -154,11 +154,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
     constructor(transport: Transport, descriptor: Descriptor) {
         super();
 
-        if (transport.name === 'BridgeTransport') {
-            this.protocol = bridgeProtocol;
-        } else {
-            this.protocol = v1Protocol;
-        }
+        this.protocol = v1Protocol;
```
mroz22, line 157: this doesn't seme to be changed anywhere on connect level - so It is a constant. and it is passed to transport calls for example here 
https://github.com/trezor/trezor-suite/blob/feat/transport-bridge-protocol-message/packages/connect/src/device/DeviceCommands.ts#L326
is it still needed? or is it supposed to be extended in the future with THP?

mroz22 (approved):
pushed a fixup with a typeguard. left couple of nitpicks. it works on my machine so approve

packages/transport-bridge/src/http.ts
szymonlesisz: typical chicken egg problem, i've reordered the commits.
now `BridgeTransport` implementation comes before `TrezordNode` but at least its atomic

packages/connect/src/device/Device.ts
szymonlesisz: right, it can be changed dynamically when we decide taht this is THP device and we want to use protocol v2

szymonlesisz:
resolving conflicts

PR #13700, opened 2024-08-07
feat(suite): Add Warning.Button

jvaclavik:
## Related Issue

Resolve 

## Screenshots:
image

### DIFF
--- a/packages/components/src/components/Warning/Warning.tsx
+++ b/packages/components/src/components/Warning/Warning.tsx
@@ -16,6 +16,7 @@ import {
 import { UIVariant } from '../../config/types';
 import { Button, ButtonProps, Row, TransientProps, useElevation } from '../..';
 import { FrameProps, FramePropsKeys, withFrameProps } from '../common/frameProps';
+import { useWarningContextContext, WarningContext } from './WarningContext';
 
 export const allowedWarningFrameProps: FramePropsKeys[] = ['margin'];
 type AllowedFrameProps = Pick<FrameProps, (typeof allowedWarningFrameProps)[number]>;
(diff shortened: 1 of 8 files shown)


packages/components/src/components/Warning/Warning.tsx
```diff
@@ -24,10 +26,14 @@ export type WarningVariant = Extract<
     'primary' | 'secondary' | 'info' | 'warning' | 'destructive' | 'tertiary'
 >;
 
+const DEFAULT_VARIANT = 'warning' as WarningVariant;
+
 export type WarningProps = AllowedFrameProps & {
     children: ReactNode;
+    extra?: ReactNode;
```
seibei-iguchi: Is this used?

packages/components/src/components/Warning/Warning.tsx
```diff
@@ -24,10 +26,14 @@ export type WarningVariant = Extract<
...
     className?: string;
     variant?: WarningVariant;
+    rightContent?: ReactNode;
```
seibei-iguchi, line 28: Isn't string sufficient?

packages/components/src/components/Warning/Warning.tsx
```diff
@@ -158,7 +165,29 @@ export const Warning = ({
...
+                    <WarningContext.Provider value={{ variant }}>
+                        {rightContent}
+                    </WarningContext.Provider>
+                )}
+            </Row>
         </Wrapper>
     );
 };
+
+const WarningButton = ({ children, ...rest }: ButtonProps) => {
```
seibei-iguchi: Shouldn't this be in its own file?

packages/components/src/components/Warning/Warning.tsx
jvaclavik: I missed that, thanks!

packages/components/src/components/Warning/Warning.tsx
jvaclavik: It is not because you can add there any component. Ideally predefined ones, but not necessarily.

packages/components/src/components/Warning/Warning.tsx
jvaclavik: It makes sense, I separated it to multiple files.

seibei-iguchi (approved):

PR #13701, opened 2024-08-07
chore(connect): if duplicated coins, let's check if all are unsupported

tomasklim:
## Description

- bnb smart chain is unsupported on T1 because of collision if BNB beacon chain
- it is going to dissapear really soon https://www.bnbchain.org/en/bnb-chain-fusion
- this "improvement" does not set coin to be unsupported in case there is at least one of duplicates supported

## Related Issue

Resolve


marekrjpolak (approved):
Probably makes sense, and shouldn't currently affect anything else than bnb.

PR #13706, opened 2024-08-08
test(desktop) : test stability improvement

ondracja:
- updating clicking function from discouraged `window.click()` to encouraged `window.locator.click();`
- making the function async as well since `.click()` returns a promise


Lemonexe (approved):
LGTM :rocket: 
Locator is more strict, less ambiguous than window.click, and [indeed recommended](https://github.com/microsoft/playwright/blob/236a8ac2ac00fe90701bdc3d6c36e840033c5a8e/packages/playwright-core/types/types.d.ts#L2006-L2025)..

PR #13714, opened 2024-08-08
Connect Core without global state

marekrjpolak:
Occasionally, unit/e2e tests were failing because of global state shared between two tests.

This PR moved top-level global variables from `connect/core/index` file one by one into `Core` class and propagates them through the functions as `CoreContext`. Therefore, two Core instances shouldn't interfere any more (at least on Connect level, on device/transport levels it's a whole different story).

Seems to be working, but still **must be carefully tested as some behavior could've (wrongfully) been dependent on the shared state**.


mroz22 (approved):
looks promising

PR #13717, opened 2024-08-08
chore(connect-common): add fw 2.8.1

komret:
## Description

Update FW binaries for T2T1 and T3T1 to 2.8.1

@Hannsek What do you think about the changelog UI? I think there should be fewer items in the changelog, this almost doesn't fit the screen.

## Screenshots:
💅 
Screenshot 2024-08-08 at 17 49 42

### DIFF
--- a/packages/connect-common/files/firmware/t2t1/releases.json
+++ b/packages/connect-common/files/firmware/t2t1/releases.json
@@ -1,6 +1,25 @@
 [
     {
         "required": false,
+        "version": [2, 8, 1],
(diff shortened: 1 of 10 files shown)


Hannsek:
I think that FW team is very productive! 😁

Also, "more smooth" is wrong. Rather please use "smoother".

MiroslavProchazka:
> Also, "more smooth" is wrong. Rather please use "smoother".

Fixed

MiroslavProchazka (approved):

PR #13502, opened 2024-07-25
Updating releases.json of additional property

MiroslavProchazka:
Updating releases.json of additional property. Needed for implementation of [this](https://github.com/trezor/trezor-suite-private/issues/104) 

To run check to validate the revisions are correct: `./packages/connect-common/scripts/check-firmware-revisions.sh`

image

PR #13726, opened 2024-08-09
feat(connect-common): Generalize firmware-check script for all devices, backfill revisionIds into releases

Lemonexe:
## Description

Followup to #13502

- Create a new `check-all-firmware-revisions.sh` script, which runs `check-firmware-revisions.sh` for each device.
  - _(understood as an existing folder in [connect-common firmware](https://github.com/trezor/trezor-suite/tree/develop/packages/connect-common/files/firmware))_
- Generalize `check-firmware-revisions.sh` to work on all devices
- Backfill revisionIds for T3T1 and T2B1 into their `releases.json` files

## CR instructions

Run the following (the yarn script does not seem to be part of CI)
```
yarn workspace @trezor/connect-common validate-releases.json
./packages/connect-common/scripts/check-firmware-revisions.sh t3t1
``` 

The following should error (invalid device name):
```
./packages/connect-common/scripts/check-firmware-revisions.sh t999b1234
```

## Related Issue

Part 1/3 of https://github.com/trezor/trezor-suite-private/issues/109

### DIFF
--- a/packages/connect-common/scripts/check-all-firmware-revisions.sh
+++ b/packages/connect-common/scripts/check-all-firmware-revisions.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+PARENT_PATH=$( cd "$(dirname "${BASH_SOURCE[0]}")" || exit ; pwd -P )
+
+cd "$PARENT_PATH/../files/firmware" || exit
+
+for dir in */ ; do
+    if [ -d "$dir" ]; then
+        if ! "$PARENT_PATH/check-firmware-revisions.sh" "${dir%/}" ; then
+            exit 1
+        fi;
+    fi
+done
--- a/packages/connect-common/files/firmware/t2b1/releases.json
+++ b/packages/connect-common/files/firmware/t2b1/releases.json
@@ -10,6 +10,7 @@
         "url_bitcoinonly": "firmware/t2b1/trezor-t2b1-2.8.0-bitcoinonly.bin",
         "fingerprint": "cf3ce230a69a681199f74cf6ac8c6c431f8fa7e0d0183437f93c5cc029fbd155",
         "fingerprint_bitcoinonly": "ae088439d44fc8643b8de28e0d7a8720cd3dbb247619f2742604bbe884542558",
+        "firmware_revision": "dd4671a5104952ef505d28d1f9e94d1484b4607a",
         "changelog": "* Removed CoSi functionality. \n* Increased Optiga read timeout to avoid spurious RSODs."
     },
     {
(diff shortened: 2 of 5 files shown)


packages/connect-common/scripts/check-all-firmware-revisions.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+PARENT_PATH=$( cd "$(dirname "${BASH_SOURCE[0]}")" || exit ; pwd -P )
+
+cd "$PARENT_PATH/../files/firmware" || exit
+
+for dir in */ ; do
```
peter-sanderson, line 7: this `cd` + glob is smart, I wanted to use it as well [here](https://github.com/trezor/data/pull/90/files#diff-174612db9a6f23fe0db7575ea6471cddf33039da8341189af284c0084d0ee632R3) but I did not found a way how to filter out (directly in glob) some results in an elegant way

Lemonexe: In `trezor/data` the case is more complicated because some dirs need to be filtered out, but here we can do it simpler :slightly_smiling_face:

packages/connect-common/scripts/check-all-firmware-revisions.sh
```diff
@@ -0,0 +1,11 @@
...
+    if [ -d "$dir" ]; then
+        "$PARENT_PATH/check-firmware-revisions.sh" "${dir%/}"
```
peter-sanderson: I think this wont propagate the bash return code. If you run this with some error will it produce and `1` when doing `echo $?`?

peter-sanderson: I think you have to explicitly check for it like this https://github.com/trezor/data/pull/90/files#diff-174612db9a6f23fe0db7575ea6471cddf33039da8341189af284c0084d0ee632R8

Lemonexe: Done :heavy_check_mark:

szymonlesisz (commented):
i've tested it locally and i feel that it works only partially,

it fails when i add some new release (1.13.0) to
```
./packages/connect-common/files/firmware/t1b1/releases.json
```
but it doesnt fail if i remove recent release (1.12.1) or somewhere in the middle (1.11.1)

is this expected that we allow releases.json to be outdated/different?

Lemonexe:
> i've tested it locally and i feel that it works only partially,

@szymonlesisz oh yes, that sounds like expected behavior.
The check works only one-way: if all entries in `releases.json` are valid.
It does **not** check it the other way: if all tags are in `releases.json.`

I believe it'd be impossible, as some releases are only for some devices, so if there is a tag `core/v2.6.0`, we can't expect it in every `releases.json`. That's what I think, but I'd have to check with @peter-sanderson, who originally wrote the script, I'm just expanding it.

vdovhanych (approved):

PR #13727, opened 2024-08-09
fix(suite): do not enforce revision-check for install firwmare in settings

peter-sanderson:
Allows to install firmware. It is a part of settings so we shall not re-trigger the compromised screen.

### DIFF
--- a/packages/suite/src/components/suite/Preloader/Preloader.tsx
+++ b/packages/suite/src/components/suite/Preloader/Preloader.tsx
@@ -21,6 +21,14 @@ import { WelcomeLayout } from '../layouts/WelcomeLayout/WelcomeLayout';
 import { ViewOnlyPromo } from 'src/views/view-only/ViewOnlyPromo';
 import { selectDevice } from '@suite-common/wallet-core';
 import { DeviceCompromised } from '../SecurityCheck/DeviceCompromised';
+import { RouterAppWithParams } from '../../../constants/suite/routes';
+
+const ROUTES_TO_SKIP_REVISION_CHECK: RouterAppWithParams['app'][] = [
+    'settings',
+    'firmware',
+    'firmware-type',
+    'firmware-custom',
+];
 
 const getFullscreenApp = (route: AppState['router']['route']): FC | undefined => {
     switch (route?.app) {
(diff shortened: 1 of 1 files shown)


packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -21,6 +21,13 @@ import { WelcomeLayout } from '../layouts/WelcomeLayout/WelcomeLayout';
 import { ViewOnlyPromo } from 'src/views/view-only/ViewOnlyPromo';
 import { selectDevice } from '@suite-common/wallet-core';
 import { DeviceCompromised } from '../SecurityCheck/DeviceCompromised';
+import { RouterAppWithParams } from '../../../constants/suite/routes';
+
+const ROUTES_TO_SKIP_REVISION_CHECK: RouterAppWithParams['app'][] = [
```
komret, line 26: Please add `firmware-type`.

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -72,7 +79,8 @@ export const Preloader = ({ children }: PreloaderProps) => {
     }
 
     if (
-        router.route?.app !== 'settings' &&
+        (router.route?.app === undefined ||
```
komret, line 83: Nit:

```suggestion
        (router.route?.app !== undefined &&
```

Isn't it more intuitive like this?

komret (changes requested):

komret (approved):

bosomt:
**QA OK**

i was able to install custom firmware and install official firmware even app identifies device as compromised

**Info:**
 - Suite version: desktop 24.8.1 (ff36df61a0f4df46dfc9fbff1ac26f00429a00dc)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.8.1 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.8.2 regular (revision 4b68431f7cc04f5e0ae041e4bf2ddc82bba79cd3)

PR #13742, opened 2024-08-09
Feat/native/send fees

### LINKED ISSUE #10871: Send coins - fee rate input
## Description
User should be able to specify the fee rate of the transaction. Under the amount input there will be rendered fee level options (low,normal,high).

For testnets there is always only normal level.

PeKne:
## Description
- new fee selection screen added to the send flow
- send flow enabled for any Bitcoin like account (including the mainnets!)

Please focus on the files related to the fees. The fee screen should be almost final (even design-wise). Do not investigate all the other screens (address and amount form, review screen etc.). These will be revisited in the follow up PRs.

## Related Issue

Resolve #10871

## Screenshots:
https://github.com/user-attachments/assets/957d80ac-c3ed-467e-95e1-c85a3ff00204


suite-native/formatters/src/hooks/useFiatFromCryptoValue.ts
```diff
@@ -16,15 +16,18 @@ type useFiatFromCryptoValueParams = {
     tokenDecimals?: number;
     historicRate?: number;
     useHistoricRate?: boolean;
+    isBalance?: boolean;
 };
 
 export const useFiatFromCryptoValue = ({
+    // TODO: is balance
```
vytick: unnecessary comment?

packages/connect/src/types/fees.ts
```diff
@@ -39,3 +39,5 @@ export const SelectFeeLevel = Type.Union([
         total: Type.String(),
     }),
 ]);
+
+export type FeeLevelLabel = FeeLevel['label'];
```
mroz22: probably not a big deal adding this, but lets keep in mind that with every new export from connect we have more complicated live in terms of keeping things backwards compatible

PeKne: I can move this to `suite-common` so the responsibility is on the side of the Suite. Would it be better for you?  `FeeLevel` is used there already anyway.

mroz22: I'd probably move it out of connect. not saying it can't be there, just that when it becomes part of public API, we should think twice whether it makes sense or not.

suite-native/module-send/src/components/FeeOption.tsx
```diff
@@ -0,0 +1,86 @@
...
+    feeKey: SendFeesFormValues['feeLevel'];
+    feeLevel: GeneralPrecomposedTransactionFinal;
+    networkSymbol: NetworkSymbol;
+}) => {
+    const { watch, setValue } = useContext(FormContext);
+    const selectedLevel = watch('feeLevel');
+
+    const isChecked = selectedLevel === feeKey;
+
+    const { label, timeEstimate } = feeLabelsMap[feeKey];
```
vytick, line 43: I am testing this flow and the app crashes  on btc testnet send attempt, on `custom` fee label which is not allowed for native but is inserted via `SendFeesScreen params`

suite-native/atoms/src/Radio.tsx
```diff
@@ -4,14 +4,14 @@ import { NativeStyleObject, prepareNativeStyle, useNativeStyles } from '@trezor/
 
 import { ACCESSIBILITY_FONTSIZE_MULTIPLIER } from './Text';
 
-type RadioValue = string | number;
-export interface RadioProps extends Omit<TouchableOpacityProps, 'style' | 'onPress'> {
-    value: RadioValue;
+// type RadioValue = string | number;
```
vytick: not needed here

suite-native/module-send/src/components/FeeOption.tsx
vytick: https://github.com/user-attachments/assets/a07ea84d-ed58-427d-8cae-0c2e97797e57

suite-native/module-send/src/components/FeesFooter.tsx
```diff
@@ -0,0 +1,79 @@
...
+    totalAmount: string;
+};
+
+const footerWrapperStyle = prepareNativeStyle(utils => ({
+    marginBottom: utils.spacings.small,
+}));
+
+const cardStyle = prepareNativeStyle(utils => ({
+    paddingTop: utils.spacings.small,
+    paddingBottom: utils.spacings.extraLarge + utils.spacings.small,
```
vytick: nit: I would prefer not to compose arbitrary numbers from predefined spacings, wdyt?

PeKne: 7f3034dd1d

suite-native/atoms/src/Radio.tsx
PeKne: 7f3034dd1d

suite-native/module-send/src/components/FeeOption.tsx
PeKne: good catch: 7f3034dd1d

suite-native/formatters/src/hooks/useFiatFromCryptoValue.ts
PeKne: 7f3034dd1d

packages/connect/src/types/fees.ts
PeKne: fixed by c8c1993b7b

suite-native/module-send/src/components/SendFeesForm.tsx
```diff
@@ -0,0 +1,97 @@
...
+    });
+    const { handleSubmit, control } = form;
+
+    const selectedFeeLevel = useWatch({ control, name: 'feeLevel' });
+    const feeLevelTransaction = feeLevels[selectedFeeLevel] as PrecomposedTransactionFinal;
+
+    if (!account) return;
+
+    const handleNavigateToReviewScreen = handleSubmit(async () => {
+        navigation.navigate(SendStackRoutes.SendReview, { accountKey });
```
vytick, line 61: I suggest to move this navigation after `isRejected` check of `signTransactionFeeLevelSigningThunk` or the app will navigate and crash in case `signTransactionFeeLevelSigningThunk` rejects.

vytick (changes requested):

suite-native/module-send/src/components/SendFeesForm.tsx
PeKne: Make sense on the first look, but unfortunately we can not do that. While is the `signTransactionFeeLevelSigningThunk` executing, the user needs to be already on the `SendReviewScreen` where he can see the individual steps of the review and their status. 

If the navigation would happen after the dispatch of `signTransactionFeeLevelSigningThunk` the Trezor device would start to display inputs for the review, but the app would not correspond with that.

In case it fails, it correctly returns to the beginning of the flow (account selection screen). I tested that. Or did it not work for you? If no, please let me know how to replicate that.

vytick (approved):
great job 🦾 🚀

suite-native/module-send/src/components/SendFeesForm.tsx
vytick: As we discussed the issue here was a validation that shouldnt let it pass and it fill be done in following PRs. Resolving

PR #13734, opened 2024-08-09
Feat/ledger derivations path evm

tomasklim:
## Description

- do discovery of ledger live accounts for all EVMs
- do discovery of ledger legacy accounts for ETH
- allow to add non-bitcoin coin account type when debug mode is on
- fix jumping of account label chevron
- skip first evm ledger account as it is same as trezor

- add account type in debug mode looks broken, the texts are wrong, but the logic works, there is another issue https://github.com/trezor/trezor-suite/issues/13736 to fix it 

## Related Issue

resolve https://github.com/trezor/trezor-suite/issues/8591
resolve #13733

## Screenshots:
non-debug
Screenshot 2024-08-09 at 15 49 19

debug

Screenshot 2024-08-09 at 15 49 08

Screenshot 2024-08-09 at 15 43 21

Screenshot 2024-08-09 at 16 24 13

PR #13748, opened 2024-08-10
fix(suite): add pading above network symbol

MiroslavProchazka:
Adding padding so the CoinLogo does not touch border of the modal

BTW maybe I've added the padding also in some other places by this change, please someone from usability team check it first before merging. Diki

Before
image

After
image

### DIFF
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
@@ -27,6 +27,7 @@ import { DeviceModelInternal } from '@trezor/connect';
 
 const StyledModal = styled(Modal)`
     width: 480px;
+    padding-top: 4px;
     text-align: left;
 `;
 


packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
```diff
@@ -27,6 +27,7 @@ import { DeviceModelInternal } from '@trezor/connect';
 
 const StyledModal = styled(Modal)`
     width: 480px;
+    padding-top: 4px;
     text-align: left;
```
tomasklim, line 31: I secretly removed the coin symbol https://github.com/trezor/trezor-suite/pull/13734 😡 

Sorry Míra, you are not going to get another commit

MiroslavProchazka: 🤬🤬🤬🤬🤬

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
```diff
@@ -27,6 +27,7 @@ import { DeviceModelInternal } from '@trezor/connect';
 
 const StyledModal = styled(Modal)`
     width: 480px;
+    padding-top: 4px;
```
hstastna, line 30: Not sure if this PR is still relevant, but we prefer using `spacingsPx` instead, if possible. So in your case, it would look like this:

`padding-top: ${spacingsPx.xxs}; `

Anyway, if we still needed to add padding to this modal, maybe I'd add a bit bigger padding, 4px seem to be too little. With `spacingsPx.xs` it could look better, IMHO, as everything else (like Select component, text, buttons) in the modal is far from the border of the modal :)

PR #13759, opened 2024-08-11
feat(suite): Add cursor: not-allowed to some more components when disabled

### LINKED ISSUE #12945: Add cursor:not-allowed to disable components

hstastna:
## Description

Add `cursor: not-allowed` for cursor to look nicer to some of the interactive/input components in our design system, when those being disabled.

Additionally, add `args` and `argTypes` in storybook for components that are missing them so one could easily change the values and see how the component changes.

- `Textarea`
- `IconButton`
- `TextButton`
- `TooltipButton`

## Related Issue

## Screenshots:
**Before:**
Example: `Textarea` component
textarea_before

**After:**
Example: `Textarea` component
textarea_after

### DIFF
--- a/packages/components/src/components/buttons/TooltipButton/TooltipButton.stories.tsx
+++ b/packages/components/src/components/buttons/TooltipButton/TooltipButton.stories.tsx
@@ -1,5 +1,8 @@
 import { Meta, StoryObj } from '@storybook/react';
 import { TooltipButton as TooltipButtonComponent } from './TooltipButton';
+import { getFramePropsStory } from '../../common/frameProps';
+import { allowedButtonFrameProps } from '../Button/Button';
+import { variables } from '../../../config';
 
 const meta: Meta = {
     title: 'Buttons/TooltipButton',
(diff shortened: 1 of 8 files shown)


packages/components/src/components/buttons/TooltipButton/TooltipButton.stories.tsx
```diff
@@ -1,5 +1,8 @@
 import { Meta, StoryObj } from '@storybook/react';
 import { TooltipButton as TooltipButtonComponent } from './TooltipButton';
+import { getFramePropsStory } from '../../common/frameProps';
+import { allowedButtonFrameProps } from '../Button/Button';
```
jvaclavik, line 4: I'm not sure if using allowed frame props from Button is a good idea. I would rather strictly separate these props to each component so I would add `allowedTooltipButtonFrameProps` to `TooltipButton`. 
WDYT?

packages/components/src/components/buttons/TooltipButton/TooltipButton.stories.tsx
```diff
@@ -10,7 +13,81 @@ export default meta;
...
+        iconAlignment: 'left',
+        title: 'Button title',
         tooltipContent: 'Example tooltip',
+        ...getFramePropsStory(allowedButtonFrameProps).args,
+    },
+    argTypes: {
+        children: {
+            table: {
+                type: {
+                    summary: 'ReactNode',
```
jvaclavik, line 30: I'm just curious, why do you need to specify the type?

packages/components/src/components/buttons/TooltipButton/TooltipButton.tsx
```diff
@@ -3,17 +3,17 @@ import styled, { useTheme } from 'styled-components';
 import { Icon } from '../../assets/Icon/Icon';
 import { Tooltip } from '../../Tooltip/Tooltip';
 import { Button, ButtonProps } from '../Button/Button';
+import { spacingsPx } from '@trezor/theme';
 
 const StyledButton = styled(Button)`
     position: relative;
-    padding-left: 32px;
-    padding-right: 32px;
+    padding-inline: ${spacingsPx.xxl};
```
jvaclavik, line 10: I didn't know that this exists, good to know 👍

packages/components/src/components/buttons/TooltipButton/TooltipButton.stories.tsx
hstastna: I don't think I do need, storybook can inherit many of them but earlier this summer I was told by you to specify type of each arg in storybook anyway, to be more clear with types and not just rely on storybook itself.

packages/components/src/components/buttons/TooltipButton/TooltipButton.tsx
hstastna: Yes, thanks, but beware, padding-inline is not 100% the same as using padding-left or right. Orientation of the screen matters when using it. However if both values are the same it does not matter and we can safely use it like this.

packages/components/src/components/buttons/TooltipButton/TooltipButton.stories.tsx
jvaclavik: That's true, I was just thinking what does it mean `summary: 'ReactNode'` here

packages/components/src/components/buttons/TooltipButton/TooltipButton.stories.tsx
hstastna: TooltipButton has the same props as `Button`, except `tooltipContent`:
```
export interface TooltipButtonProps extends ButtonProps {
    tooltipContent: ReactNode;
}
```
I found ok what I did because `TooltipButtonProps` is defined exactly like this ("...extends ButtonProps"), and I wanted to avoid repeating code. But I can add `allowedTooltipButtonFrameProps` to `TooltipButton` anyway. I just considered that unnecessary.

jvaclavik: You are right, it makes sense to keep it this way 👍

jvaclavik (approved):

packages/components/src/components/buttons/TooltipButton/TooltipButton.stories.tsx
hstastna: The `summary` field within the `type` object provides a brief description of the type. In this case, it indicates that the `children` property is expected to be of type `ReactNode`.

It's the way how to provide more specific, more custom type in storybook, different than just simple string or number or array. I find it better to use this, better than just writing "string" here instead (as it is incorrectly in some other places and IMHO we should fix it). This way helps Storybook generate more accurate and informative documentation for the component.

PR #13772, opened 2024-08-12
feat(coinmarket): Add DCA landing page (#13724)

### LINKED ISSUE #13724: DCA landing page in Trezor Suite
DCA has been removed from Suite recently and we'd like to offer users alternative in our Invity App which is more focused on this feature.

**Design**
https://www.figma.com/design/xSW1knAjH25SfOmqSdNioY/Trade?node-id=7987-27851&t=HV5JHw0xq1DT11aG-4

- DCA tab and this page will be only displayed to users in European Economic Area where is Invity App available

salax:
## Description

Add new DCA section to Trade.

## Related Issue

Resolve #13724

## Screenshots:
Screenshot 2024-08-12 at 15 56 18
Screenshot 2024-08-12 at 15 56 47


packages/suite/src/constants/wallet/coinmarket/EEA.ts
```diff
@@ -0,0 +1,34 @@
+// European Economic Area (EEA)
+
+export const EEACountryCodes = [
```
jvaclavik: just thinking if it is not a duplicate with this `packages/suite/src/constants/wallet/coinmarket/regional.ts`

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+const IconWrapper = styled.div`
+    display: flex;
+    flex-shrink: 0;
+    justify-content: center;
+    align-items: center;
+    width: 40px;
+    height: 40px;
+    border-radius: 50%;
+    background: ${palette.lightSecondaryLime500};
+    margin-right: ${spacings.sm}px;
```
jvaclavik: Btw you can use:

```tsx 
    margin-right: ${spacingsPx.sm};
```

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+`;
+
+const Subheading = styled(Paragraph)`
+    color: ${palette.lightSecondaryLime500};
+`;
+
+const WrappedText = styled(Paragraph)`
+    max-width: 200px;
+    text-align: center;
+`;
```
jvaclavik: Please don't override components from the design system (`@trezor/components`) 🙏 

(In the future we are going to prohibit it, first step is on the way: https://github.com/trezor/trezor-suite/pull/13621)

In your case here, I suggest to use `Text` component from `@trezor/components` instead of `Subheading`:

```tsx
<Text color={palette.lightSecondaryLime500}>...</Text>
```

and for WrappedText I suggest to create a wrapper component


**Why not override DS components:**
- it's really hard to refactor it when we (Usability team) change something crucial in the component
- developers don't require designers to create a consistent design because they can easily modify it themselves.
- we don't know the specific needs for individual components (it's easier to tweak a component than to tell us that something isn't working).

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+    padding: ${spacings.xxl}px;
+    border-radius: ${borders.radii.md};
+    color: ${colorVariants.standard.textOnPrimary};
+`;
+
+const Column2 = styled(Column)`
+    flex: 1 0 50%;
+    gap: ${spacings.md}px;
+    padding: ${spacings.xxl}px;
+`;
```
jvaclavik: Please don't override components here 🙏 

Looks like you are using Column only for `display:flex and flex-direction: column`, so just paste it here and delete overriding.

jvaclavik: Also when you use `Column` or `Row`, it's supposed to use it this way:

```tsx
<Row gap={spacings.sm} alignItems="flex-start">...</Row>
```

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+const StoreBadge = styled(Image)<{ $isLight: boolean }>`
+    ${({ $isLight }) => $isLight && `filter: invert(1);`}
+    transition: opacity 0.3s;
+    cursor: pointer;
+
+    &:hover {
+        opacity: 0.6;
+    }
```
jvaclavik, line 74: You can create wrapper and not override Image component here:

```tsx
const StoreBadge = styled.div`
    ${({ $isLight }) => $isLight && `filter: invert(1);`}
    transition: opacity 0.3s;
    cursor: pointer;
    &:hover {
        opacity: 0.6;
    }
`

...

<StoreBadge>
  <Image />
</StoreBadge>

File: packages/suite/src/views/wallet/coinmarket/DCA/index.tsx [hunk 6]
```diff
@@ -0,0 +1,189 @@
...
+`;
+
+const StoreSeparatorWrapper = styled.div`
+    height: 26px;
+`;
+
+const DCALanding = (props: UseCoinmarketProps) => {
+    const currentTheme = useSelector(state => state.suite.settings.theme.variant);
+
+    return (
```
jvaclavik, line 106: What about divide these JSX into multiple smaller components? You can keep them locally in this file, but it would be more readable IMHO.

Some compositions are repeating so these could be easily refactored.

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
```diff
@@ -34,6 +48,24 @@ const CoinmarketLayoutNavigation = () => {
                 />
             ) : null}
 
+            {isInEEA ? (
+                <>
+                    <SeparatorWrapper>
+                        <Divider
+                            orientation="vertical"
+                            strokeWidth={1}
+                            color="borderElevation0"
```
jvaclavik: Do you have to set this color here? Shouldn't it work without setting it? It should get elevated color

salax: yeah, without it, it's a different color `#EEEEEE` instead of `#E2E2E2`

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
```diff
@@ -1,17 +1,31 @@
...
+
 const CoinmarketLayoutNavigation = () => {
     const { device } = useDevice();
     const isBitcoinOnly = device?.firmwareType === FirmwareType.BitcoinOnly;
+    const country = useSelector(
+        state =>
+            state.wallet.coinmarket.buy.buyInfo?.buyInfo?.country ??
+            state.wallet.coinmarket.sell.sellInfo?.sellList?.country,
+    );
+    const isInEEA = useMemo(() => Boolean(country && regional.isInEEA(country)), [country]);
```
tomasklim: Maybe we should we hide it if tor is active

salax:
### New QR code
Screenshot 2024-08-20 at 11 28 34
Screenshot 2024-08-20 at 11 29 09

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    flex: 1 0 50%;
+    gap: ${spacingsPx.md};
+    padding: ${spacingsPx.xxl};
+`;
+
+const Layout = styled(CoinmarketLayout)`
+    & [class^='CoinmarketLayout__Content'] {
```
jvaclavik: Could you please refactor this solution? It's very hard to read and I'm not sure if it's working on production build because classnames are obfuscated there (it's not happening in dev build).

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
```diff
@@ -1,18 +1,39 @@
...
+
 const CoinmarketLayoutNavigation = () => {
     const { device } = useDevice();
     const isBitcoinOnly = device?.firmwareType === FirmwareType.BitcoinOnly;
 
+    const torStatus = useSelector(state => state.suite.torStatus);
+    const isTorEnabled = getIsTorEnabled(torStatus);
+    const country = useSelector(
+        state =>
+            state.wallet.coinmarket.buy.buyInfo?.buyInfo?.country ??
```
jvaclavik, line 28: this is interesting 🤔

`buyInfo?.buyInfo?`

packages/urls/src/urls.ts
```diff
@@ -117,3 +117,5 @@ export const CROWDIN_URL = 'https://crowdin.com/project/trezor-suite';
 
 export const TREZOR_SAFE_5_URL = 'https://trezor.io/trezor-safe-5';
 export const ESHOP_KEEP_METAL_SINGLE_SHARE_URL = 'https://trezor.io/trezor-keep-metal-single-share';
+
+export const COINMARKET_DOWNLOAD_INVITY_APP_URL = 'https://invity.onelink.me/yIY4/q7ltbnv0';
```
jvaclavik: If I would see this on hover in statusbar I would expect a phising website.

cc @marek-andrsk

jvaclavik (changes requested):

packages/urls/src/urls.ts
marek-andrsk: it allows us to see if users are actually downloading the app via Suite, which will be important for further iterations/improvements (or removal) of this page

MiroslavProchazka: Can we use a more reliable domain, such as trezor.io/sth? .onelink.me/xyz looks like a phishing site.
cc @Hermez-cz

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: yeah, it looks weird, but it's correct (I'm using already defined state)

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
salax: changed

packages/urls/src/urls.ts
Hermez-cz: Yes please, we need to avoid both looking phishy and/or teaching users it's OK.

packages/suite/src/constants/wallet/coinmarket/regional.ts
```diff
@@ -267,6 +267,44 @@ class Regional {
...
+        'MT',
+        'NL',
+        'NO',
+        'PL',
+        'PT',
+        'RO',
+        'SE',
+        'SI',
+        'SK',
+    ];
```
peter-sanderson: nit: `as const`

packages/suite/src/constants/wallet/coinmarket/regional.ts
```diff
@@ -267,6 +267,44 @@ class Regional {
 
             return l1.localeCompare(l2);
         });
+
+    // European Economic Area (EEA)
```
peter-sanderson: Can we link some source for this? So it can be updated in the future. Wikipage/etc..

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
```diff
@@ -1,18 +1,39 @@
...
+            state.wallet.coinmarket.sell.sellInfo?.sellList?.country,
+    );
+    const isInEEA = useMemo(
```
peter-sanderson: is this `useMemo` really needed? My understanding is that it comes with a price and this condition seems to be really simple.

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
```diff
@@ -1,18 +1,39 @@
...
+        () => Boolean(!isTorEnabled && country && regional.isInEEA(country)),
```
peter-sanderson: Why is it disabled when Tor is on?

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: it's not about the condition, but about searching country in the array (I know map, would be faster, but this array is quite small, so `useMemo` seemed like good ok idea)

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: if you use Tor, then we don't know in what country you are in

peter-sanderson: Ah, I thought it is selected somewhere. How is it detected? IP? What about EU resident on vacation? How does that work?

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
peter-sanderson: I agree with [this article](https://javascript.plainenglish.io/stop-using-usememo-now-e5d07d2bbf70). The loop over **static** array with 30ish items is not expensive calculation.

Its not a hill I want to die on, but I just want to bring to the attention here, that `useMemo` is IMO usually not needed.

packages/urls/src/urls.ts
Hermez-cz: Also, will there be an analytics event, so we can statistically monitor the CTR?

marek-andrsk: @Hermez-cz I don't think that's possible to embed into QR code, that's why we use Appsflyer link .. I found a way to use custom domain, processing this now, it will be via invity.io

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: ok, I removed it

packages/suite/src/constants/wallet/coinmarket/regional.ts
salax: that would blow up `includes` in the method `isInEEA` below:
```
Argument of type string is not assignable to parameter of type
'AT' | 'BE' | 'BG' | 'CY' | 'CZ' | 'DE' | 'DK' | 'EE' | 'ES' | 'FI' | 'FR' | 'GR' | 'HR' | 'HU' | 'IE' | 'IS' | 'IT' | 'LI' | 'LT' | 'LU' | 'LV' | 'MT' | 'NL' | 'NO' | 'PL' | 'PT' | 'RO' | 'SE' | 'SI' | 'SK'
```
Not sure how to solve it without casting. Do you?

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: AFAIK it's detected on Invity's BE and I think it doesn't matter too much for this simple promotion screen...

packages/urls/src/urls.ts
salax: link finally changed

LukasRada:
When we add translations to messages.ts we need to add these to en.json as well. Source: https://github.com/trezor/trezor-suite/pull/14007#discussion_r1741160270 (there is probably need to run some script or we just need to copy paste?) Cc: @MiroslavProchazka

salax:
translations copied and rabased

LukasRada (approved):

packages/suite/src/constants/wallet/coinmarket/EEA.ts
MiroslavProchazka: Is this solved?

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
MiroslavProchazka: Is this solved?

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
MiroslavProchazka: Is this solved?

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
MiroslavProchazka: Is this solved?

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
MiroslavProchazka: Is this solved?

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
MiroslavProchazka: Is this solved?

packages/suite/src/constants/wallet/coinmarket/EEA.ts
salax: yes

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
salax: yes

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
salax: yes

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
salax: yes

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: this would result in a different color than we want

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: yes

MiroslavProchazka (approved):
We need to test this properly and observe that it does not break anything. So I'm approving it with keeping in mind that you will observe develop after merge and react promptly in case of any issues.

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+const Column2 = styled.div`
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    flex: 1 0 50%;
+    gap: ${spacingsPx.md};
+    padding: ${spacingsPx.xxl};
+`;
```
jvaclavik: could you unify common css to abstract parent?

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+        display: flex;
+    }
+`;
+
+const Column1 = styled.div`
+    display: flex;
+    flex-direction: column;
+    flex: 1 0 50%;
+    gap: ${spacingsPx.xxl};
+    background-color: ${colorVariants.standard.backgroundPrimaryDefault};
```
jvaclavik: Why not to use colors directly from palette? You have both approaches in this file.

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
jvaclavik: I think we should still use `mapElevationToBorder`. We are going to update colors from figma soon. In this case we would need to refactor it.

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
```diff
@@ -1,18 +1,35 @@
...
-import { useDevice } from 'src/hooks/suite';
+import { useDevice, useSelector } from 'src/hooks/suite';
 import { FirmwareType } from '@trezor/connect';
 import CoinmarketLayoutNavigationItem from './CoinmarketLayoutNavigationItem';
+import { Divider } from '@trezor/components';
+import { spacings } from '@trezor/theme';
+import regional from 'src/constants/wallet/coinmarket/regional';
+import { getIsTorEnabled } from 'src/utils/suite/tor';
 
 const List = styled.div`
```
jvaclavik, line 10: nit: You can use <Row flexWrap="wrap" alignItems="center">...</Row>

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+                            <TrezorLink href={COINMARKET_DOWNLOAD_INVITY_APP_URL}>
+                                <StoreBadge $isLight={isLightTheme}>
+                                    <Image image="PLAY_STORE_TITLE" height={26} />
+                                </StoreBadge>
+                            </TrezorLink>
+                            <StoreSeparatorWrapper>
+                                <Divider
+                                    orientation="vertical"
+                                    strokeWidth={1}
+                                    color="borderElevation1"
```
jvaclavik: This should be elevated automatically according to `mapElevationToBorder`.

packages/suite/src/constants/wallet/coinmarket/regional.ts
jvaclavik: I think this should work:

```tsx
export const countryCodes = ['AT', 'BE', 'BG'] as const;
export type CountryCode = (typeof countryCodes)[number];
```

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
salax: because we want the same color in light and dark mode

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: What do you mean here?

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
salax: ok, removed

packages/suite/src/constants/wallet/coinmarket/regional.ts
salax: ok, changed

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
salax: done

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
salax: unified to use `palette`

packages/suite/src/views/wallet/coinmarket/DCA/index.tsx
salax: ok, removed

packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayoutNew/CoinmarketLayoutNavigation/CoinmarketLayoutNavigation.tsx
jvaclavik: Lol, what happened here :D don't worry about it

jvaclavik (approved):
LGTM!

PR #13783, opened 2024-08-13
feat(suite): Check firmware for all models

Lemonexe:
Check firmware revisionsIds also for T2B1 and T3T1

## Description

Although the new models have a hw authenticity chip, we would like to also check revision Ids also, as we do for T1B1 and T2T1, as an additional layer of security. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite-private/issues/109

## Screenshots:

Running T2B1 or T3T1 on latest `trezor-user-env`, but choosing `2-main`, gets you ghost (like all the other "main" versions):

For those models, in settings there is now both Turn off firmware revision check & device check:

### DIFF
--- a/packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
+++ b/packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
@@ -102,13 +102,17 @@ describe(checkFirmwareRevision.name, () => {
             expected: { success: false, error: 'cannot-perform-check-offline' },
         },
         {
-            it: "doesn't check anything for new versions with security chip (not implemented yet)",
-            params: createDevicePrams({
-                deviceRevision: '1eb0eb9d91b092e571aac63db4ebff2a07fd8a1f',
-                expectedRevision: 'cde8f31ec2ddcb7d35e36edbcf8a71dda983a9ea',
-                internalModel: DeviceModelInternal.T3B1,
-            }),
+            it: 'passes also for a different device',
+            params: createDevicePrams({ internalModel: DeviceModelInternal.T3T1 }),
             expected: { success: true },
+        },
+        {
+            it: 'fails also for a different device',
+            params: createDevicePrams({
+                internalModel: DeviceModelInternal.T3T1,
+                expectedRevision: '1234567890987654321',
+            }),
+            expected: { success: false, error: 'revision-mismatch' },
         },
     ])(`$it`, async ({ params, expected, httpRequestMock }) => {
         if (httpRequestMock !== undefined) {
--- a/packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
+++ b/packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
@@ -19,42 +19,39 @@ const ONLINE_RELEASES_JSON_MOCK: FirmwareRelease[] = [
     },
 ];
 
-const createDevicePrams = (
-    params: Partial<CheckFirmwareRevisionParams>,
-): CheckFirmwareRevisionParams => ({
+const DeviceNames = Object.values(DeviceModelInternal);
+
+type CreateDeviceParams = Omit<CheckFirmwareRevisionParams, 'internalModel'>;
+
+const createDeviceParams = (params: Partial<CreateDeviceParams>): CreateDeviceParams => ({
     deviceRevision: '1eb0eb9d91b092e571aac63db4ebff2a07fd8a1f',
     firmwareVersion: [2, 7, 2],
     expectedRevision: '1eb0eb9d91b092e571aac63db4ebff2a07fd8a1f',
-    internalModel: DeviceModelInternal.T1B1,
     ...params,
 });
 
-describe(checkFirmwareRevision.name, () => {
+describe.each(DeviceNames)(`${checkFirmwareRevision.name} for device %s`, internalModel => {
     it.each<{
         it: string;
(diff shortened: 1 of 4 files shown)


packages/suite/src/views/settings/SettingsDevice/SettingsDevice.tsx
```diff
@@ -89,7 +92,7 @@ export const SettingsDevice = () => {
     const deviceModelInternal = device.features.internal_model;
 
     const supportsDeviceAuthentication = SUPPORTS_DEVICE_AUTHENTICITY_CHECK[deviceModelInternal];
-    const supportsFirmwareRevisionCheck = !supportsDeviceAuthentication; // Older devices with no secure element
+    const supportsFirmwareRevisionCheck = SUPPORTS_FIRMWARE_REVISION_CHECK[deviceModelInternal];
```
Lemonexe: I could just remove the condition (always true), but I think keeping it granular is no harm done :slightly_smiling_face:

packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
```diff
@@ -102,13 +102,17 @@ describe(checkFirmwareRevision.name, () => {
             expected: { success: false, error: 'cannot-perform-check-offline' },
         },
         {
-            it: "doesn't check anything for new versions with security chip (not implemented yet)",
+            it: 'passes also for a different device',
```
peter-sanderson:
```suggestion
            it: 'passes also for a T3T1 device',
```

\+ maybe I would event consider to adding some test to test all know devices with a defined map of all devices (so Typescript force you to run such test for every device from `DeviceModelInternal` enum

packages/suite/src/constants/suite/device.ts
```diff
@@ -52,3 +52,11 @@ export const SUPPORTS_DEVICE_AUTHENTICITY_CHECK: Record<DeviceModelInternal, boo
     [DeviceModelInternal.T3B1]: true,
     [DeviceModelInternal.T3T1]: true,
 };
+
+export const SUPPORTS_FIRMWARE_REVISION_CHECK: Record<DeviceModelInternal, boolean> = {
```
peter-sanderson: :clap:

packages/suite/src/views/settings/SettingsDevice/SettingsDevice.tsx
peter-sanderson: I would just remove it. We want to do check for all devices into the future. So I think there is no reason to keep unnecessary code here.

packages/suite/src/constants/suite/device.ts
peter-sanderson: After reading file below I suggest to remove this at all together with the condition.

packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
Lemonexe: Done in new commit 6e294b875e946256e44f9cc74407ff9dc1f2c5a8.

We could list all devices manually here, and have it exhaustively typed so TS forces me to maintain it - that would work, but I propose a more straightforward way: simply iterate the test foreach `DeviceModelInternal` value.

packages/suite/src/views/settings/SettingsDevice/SettingsDevice.tsx
Lemonexe: Removed in amended commit 1e7ebc1bc5f15dc78ae41aa132c7d3d684dc017b

packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
```diff
@@ -19,42 +19,39 @@ const ONLINE_RELEASES_JSON_MOCK: FirmwareRelease[] = [
...
 ];
 
-const createDevicePrams = (
-    params: Partial<CheckFirmwareRevisionParams>,
-): CheckFirmwareRevisionParams => ({
+const DeviceNames = Object.values(DeviceModelInternal);
+
+type CreateDeviceParams = Omit<CheckFirmwareRevisionParams, 'internalModel'>;
+
+const createDeviceParams = (params: Partial<CreateDeviceParams>): CreateDeviceParams => ({
```
Lemonexe, line 26: fixed typo Prams -> Params

packages/connect/src/device/__tests__/checkFirmwareRevision.test.ts
```diff
@@ -19,42 +19,39 @@ const ONLINE_RELEASES_JSON_MOCK: FirmwareRelease[] = [
...
         httpRequestMock?: () => Promise<FirmwareRelease[]>;
-        params: CheckFirmwareRevisionParams;
+        params: CreateDeviceParams;
         expected: FirmwareRevisionCheckResult;
     }>([
         {
             it: 'passes when firmware revision is same as in static file',
-            params: createDevicePrams({
-                deviceRevision: '1eb0eb9d91b092e571aac63db4ebff2a07fd8a1f',
-                expectedRevision: '1eb0eb9d91b092e571aac63db4ebff2a07fd8a1f',
```
Lemonexe, line 43: since `createDeviceParams` assigns these as default values, I think it's needless to repeat them here, so I removed it, but I can put it back if you'd like to.

mroz22 (approved):

peter-sanderson (approved):

PR #13785, opened 2024-08-13
feat(suite-native): changing switch visuals to stand out more

vytick:
After the discussion with @shenkys the visual for switch has changed mostly in dark mode.
In unchecked state we add border radius and the thumb (circle) is white in both light and dark mode

Screenshot 2024-08-13 at 13 32 36

## Related Issue

Resolve 

## Screenshots:

https://github.com/user-attachments/assets/a096ce30-175a-4d0d-b66f-1ac72187d64a

### DIFF
--- a/suite-native/atoms/src/Switch.tsx
+++ b/suite-native/atoms/src/Switch.tsx
@@ -9,7 +9,7 @@ import Animated, {
 } from 'react-native-reanimated';
 
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
-import { Color } from '@trezor/theme';
+import { palette } from '@trezor/theme';
 
 import { ACCESSIBILITY_FONTSIZE_MULTIPLIER } from './Text';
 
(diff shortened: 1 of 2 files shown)


suite-native/atoms/src/Switch.tsx
```diff
@@ -9,46 +9,54 @@ import Animated, {
...
     isDisabled?: boolean; // Functionality of disabled works but styles are not implemented yet (waiting for design)
-    colorChecked?: Color;
-    colorUnchecked?: Color;
 };
 
-const SWITCH_CONTAINER_WIDTH = 44 * ACCESSIBILITY_FONTSIZE_MULTIPLIER;
-const SWITCH_CONTAINER_HEIGHT = 24 * ACCESSIBILITY_FONTSIZE_MULTIPLIER;
+const SWITCH_CONTAINER_BORDER_WIDTH = 1 * ACCESSIBILITY_FONTSIZE_MULTIPLIER;
+const SWITCH_CONTAINER_WIDTH =
+    44 + 2 * SWITCH_CONTAINER_BORDER_WIDTH * ACCESSIBILITY_FONTSIZE_MULTIPLIER;
```
PeKne: Should not it be?
```suggestion
    44 * ACCESSIBILITY_FONTSIZE_MULTIPLIER + 2 * SWITCH_CONTAINER_BORDER_WIDTH * ACCESSIBILITY_FONTSIZE_MULTIPLIER;
```

Same for `SWITCH_CONTAINER_HEIGHT`.

Was is tested with upscaled font?

suite-native/atoms/src/Switch.tsx
```diff
@@ -9,46 +9,54 @@ import Animated, {
...
+            },
+        },
+    ],
 }));
 
 const switchCircleStyle = prepareNativeStyle(utils => ({
     width: SWITCH_CIRCLE_SIZE,
     height: SWITCH_CIRCLE_SIZE,
-    backgroundColor: utils.colors.backgroundSurfaceElevation1,
+    backgroundColor: palette.darkGray1000,
```
PeKne, line 56: why is pallete used instead of color nickname?
```suggestion
    backgroundColor: utils.colors.backgroundNeutralBold,
```

PeKne (changes requested):

suite-native/atoms/src/Switch.tsx
vytick: This is because the color is the same for both light and dark scheme. `backgroundNeutralBold` makes it dark gray in light mode. 


We dont have any color token like that, because the change is not in the design system, so I directly used palette color which felt better than using `#ffffff` directly.

suite-native/atoms/src/Switch.tsx
vytick: good catch

vytick: 063e8ea369

vytick: enlarged fontsize

PeKne (approved):

PR #13788, opened 2024-08-13
fix(staking): staking tx in staking tab

### LINKED ISSUE #12827: Staking transactions in staking tab
Staking tab grabs the first page of TXs from blockbook and filters them out non-staking TXs. That means just 25 txs.

Active users with many general ETH transactions will not see staking transactions on the staking tab

As those solutions are still not available:
https://github.com/trezor/blockbook/issues/1101
https://github.com/trezor/trezor-suite/issues/11306

Let's implement temp solution.

Problem:
Heeelp I don't see my stake tx
Image

Because it is already on second page
Image

Everstake staking started in block/time: 18120102 / Sep-12-2023 11:58:23 AM UTC
https://etherscan.io/tx/0x746689576170f6bbfe64a0c8e7e1714978d6817705cf40c04b07d2a8d4ef0909

When user opens staking tab, let's fetch txs until ⏫ block/time.

dev-pvl:
## Description

Updated the logic to ensure that all staking transactions are correctly displayed, regardless of the number of general transactions the user has. This change ensures that active users can now see their staking transactions as expected.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12827

### DIFF
--- a/packages/suite/src/views/wallet/staking/components/EthStakingDashboard/components/Transactions.tsx
+++ b/packages/suite/src/views/wallet/staking/components/EthStakingDashboard/components/Transactions.tsx
@@ -2,17 +2,22 @@ import { TransactionList } from 'src/views/wallet/transactions/TransactionList/T
 import { useSelector } from 'src/hooks/suite';
 import {
     selectAccountStakeTypeTransactions,
+    selectAreAllTransactionsLoaded,
     selectIsLoadingAccountTransactions,
 } from '@suite-common/wallet-core';
 
 export const Transactions = () => {
     const selectedAccount = useSelector(state => state.wallet.selectedAccount);
+    const accountKey = selectedAccount.account?.key ?? '';
+
     const transactionsIsLoading = useSelector(state =>
-        selectIsLoadingAccountTransactions(state, selectedAccount.account?.key || null),
+        selectIsLoadingAccountTransactions(state, accountKey),
     );
-    const stakeTxs = useSelector(state =>
-        selectAccountStakeTypeTransactions(state, selectedAccount.account?.key || ''),
+    const areAllTransactionsLoaded = useSelector(state =>
+        selectAreAllTransactionsLoaded(state, accountKey),
     );
+
+    const stakeTxs = useSelector(state => selectAccountStakeTypeTransactions(state, accountKey));
 
     if (selectedAccount.status !== 'loaded' || stakeTxs.length < 1) {
         return null;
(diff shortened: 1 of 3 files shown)


dev-pvl:
I'd suggest this temporary solution since I've found a lot of uses of fetchAllTransactionsForAccountThunk.
I tried to enhance `TrezorConnect.getAccountInfo` with `from/to` params and used a `block/time: 18120102`, but for some reason, it doesn't work as expected. In this case, I always get `totalPages: -1` in the response, so it's impossible to get the next page. 

I just wanted to give you more context here for clarification. I used my oldest holesky transaction block/time [694288](https://holesky.etherscan.io/tx/0x441d3b3ddc58996a654751d08b047c0fb012f3111b9101c8fcd9be5aa621efde) (Jan-08-2024) for testing and got `totalPages: -1`. But when I used a newer block/time (July 10, 2024) `getAccountInfo` worked as expected. In my opinion, this from/to filtering is inconsistent.

tomasklim:
I think it works https://holesky1.trezor.io/api/address/0xb8FD7bBb411B44cB0B2B8b043A434D82aDFA0Ea7?from=694288
https://holesky1.trezor.io/api/address/0xb8FD7bBb411B44cB0B2B8b043A434D82aDFA0Ea7?from=694288&details=txs

tomasklim:
okay I see, with `pageSize` equal to 1, it returns totalPages equal to -1. I will check

https://holesky1.trezor.io/api/address/0xb8FD7bBb411B44cB0B2B8b043A434D82aDFA0Ea7?from=694288&details=txs&pageSize=1

dev-pvl:
Did you have a chance to take a look at this issue?

tomasklim:
`-1` is defined behaviour, Blockbook doesn't know how many txs is there. We have to go different direction

tomasklim:
@dev-pvl let's fetch all txs when user enters staking tab. In case of troubles we can resolve it later, let's not overengineer it if not needed.

packages/suite/src/views/wallet/staking/components/EthStakingDashboard/components/Transactions.tsx
```diff
@@ -1,19 +1,36 @@
...
     const transactionsIsLoading = useSelector(state =>
         selectIsLoadingAccountTransactions(state, selectedAccount.account?.key || null),
     );
     const stakeTxs = useSelector(state =>
         selectAccountStakeTypeTransactions(state, selectedAccount.account?.key || ''),
     );
 
+    const dispatch = useDispatch();
+
+    useEffect(() => {
```
tomasklim: I think this should be one component up and I would not show txs when they are still loading? Just show loader?

dev-pvl: Moved it up and the loader is showing now. https://github.com/trezor/trezor-suite/pull/13788/commits/7ac22c759f5f3a938288e4984245959758b1078f

tomasklim (approved):

PR #13789, opened 2024-08-13
test: test maintenance

ondracja:
- adjusting metadata test to be resilient to `firmware update` popup
- disabling part of the `db-migration` test due to cypress bug with aliases, will fix it in another PR.


MiroslavProchazka (approved):
Dunno if that is the correct way how to fix moved test element, but it is working - therefore I'm giving an approve. Works locally on my machine on a several attempts. 

https://github.com/user-attachments/assets/eae33bab-c75b-40e1-bdf1-f266cc85aa9f

PR #12693, opened 2024-06-03
Feat/message system staking

tomasklim:
## Description

- disable staking, unstaking, claiming separately
- show banner 

- disabled in dashboard and also in modals

## Related Issue

Resolves #12686

## Screenshots:
Screenshot 2024-06-03 at 11 56 03
Screenshot 2024-06-03 at 11 56 15
Screenshot 2024-06-03 at 11 57 46
Screenshot 2024-06-03 at 11 58 15

Screenshot 2024-06-03 at 12 24 13
Screenshot 2024-06-03 at 12 23 37
Screenshot 2024-06-03 at 12 22 02

PR #13793, opened 2024-08-14
feat(suite): Add message system toggle for revision check

Lemonexe:
Add a listener for `firmwarerevisioncheck` feature, which is not yet added to message config, so after this PR, the firmware revision check is  **still turned on**, this just prepares a method to shut it down quickly.

It is then turned off just like if user turns it off manually in Device settings.

## Description

When we release the firmware check, we need a way to quickly shut it down in case of problems - if we customer support gets a flurry of false positives.

#13794 will turn it off globally just like if user turns it off manually in Device settings.

## :warning: Contingency plan :scream:

Merge & deploy #13794 to turn off firmware revision check.

## Related Issue

https://github.com/trezor/trezor-suite-private/issues/104

### DIFF
--- a/packages/suite/src/components/suite/Preloader/Preloader.tsx
+++ b/packages/suite/src/components/suite/Preloader/Preloader.tsx
@@ -22,6 +22,7 @@ import { ViewOnlyPromo } from 'src/views/view-only/ViewOnlyPromo';
 import { selectDevice } from '@suite-common/wallet-core';
 import { DeviceCompromised } from '../SecurityCheck/DeviceCompromised';
 import { RouterAppWithParams } from '../../../constants/suite/routes';
+import { Feature, selectIsFeatureDisabled } from '@suite-common/message-system';
 
 const ROUTES_TO_SKIP_REVISION_CHECK: RouterAppWithParams['app'][] = [
     'settings',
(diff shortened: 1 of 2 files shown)


packages/suite/src/hooks/suite/useMessageSystemFirmwareRevisionCheck.ts
```diff
@@ -0,0 +1,10 @@
+import { Feature, selectIsFeatureDisabled } from '@suite-common/message-system';
```
tomasklim: please just use selector directly in component, we try to omit hooks, which can be replaced by selectors. You can create feature specific selector, if needed (do not think so)

Lemonexe: Will do. I took inspiration [in your PR](https://github.com/trezor/trezor-suite/pull/12693/commits/b7435db02aeb760bcab32c02d6ff44e2ada7fabe#diff-3f8c84c46f18b1531c7f9f16a33d7e454454f0683f0b08f2f24bf5e056425c31), but I agree that in this case it is unnecessary, since it only returns _one value_.

tomasklim (approved):
LGTM

PR #13796, opened 2024-08-14
refactor(suite): replace Box component with Warning

jvaclavik:
## Description

Btw `Warning` is not correct name and it's going to be renamed in the future.

## Related Issue

Resolve 

## Screenshots:

Before (approximately 😅):
image

After:
image

Before:
Screenshot 2024-08-14 at 10 06 26

After:
Screenshot 2024-08-14 at 10 14 03

Before:
no screenshot

After:
Screenshot 2024-08-14 at 09 49 58


packages/suite/src/components/wallet/WalletLayout/AccountBanners/CloseableBanner.tsx
```diff
@@ -1,39 +1,28 @@
...
     children: React.ReactNode;
 }
 
 export const CloseableBanner = ({ onClose, variant, title, children }: Props) => (
-    <StyledBox variant={variant}>
-        <Heading>
-            <H3>{title}</H3>
-            <Button variant="tertiary" onClick={onClose}>
+    <Warning
+        variant={variant}
```
hstastna, line 15: ~~Maybe we could add `withIcon` here too, as it is already in other change in this PR? I don't see any reason not to use it. But maybe I am missing something? :)~~
I can see we delete this prop in other PR.

hstastna (approved):

packages/suite/src/components/wallet/WalletLayout/AccountBanners/CloseableBanner.tsx
jvaclavik: I wanted to do it, but it's not very nice (icon would need to be positioned not to the center nor to the top but something between and we don't allow it for now).

PR #13802, opened 2024-08-14
fix(suite-web): change browser test for edge to safari

martykan:
## Description

Updated the browser support test: replaced the check for MS Edge with Safari. 

Since Edge is Chromium based and supported, it's no longer necessary to show the unsupported browser page, so we will use Safari for this test.


Lemonexe (approved):
LGTM

PR #13810, opened 2024-08-14
transport usb api - leaking abort listeners

mroz22:
- cherry-picked https://github.com/trezor/trezor-suite/pull/13570/commits originaly created by @szymonlesisz 
- added some more tests tweaking. now unit tests in transport package would fail should node emit "Error: MaxListenersExceededWarning detected." warning.

the question now - packages/transport/JestCustomEnv.js is now duplicated (the same file is in pacakges/utils). where do we put it instead?[ root?]([608b955](https://github.com/trezor/trezor-suite/pull/13810/commits/608b95595378c44be8f52aee3a3c135cbd8a8c18))

### DIFF
--- a/JestCustomEnv.js
+++ b/JestCustomEnv.js
@@ -1,3 +1,4 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
 const NodeEnvironment = require('jest-environment-node').default;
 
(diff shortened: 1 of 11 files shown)


szymonlesisz (approved):
:1st_place_medal:

PR #13815, opened 2024-08-15
feat(components): add new Modal component

seibei-iguchi:
## Screenshots:

Screenshot 2024-08-15 at 9 28 52


jvaclavik:
image

I tried to add longer children and it looks like there is no shadow until you scroll. It could be also bug in that shadow hook, could you take a look?

seibei-iguchi:
> I tried to add longer children and it looks like there is no shadow until you scroll. It could be also bug in that shadow hook, could you take a look?

It actually only breaks when changing the content dynamically (as in Storybook). If you reload the page with the longer content, it applies the shadow correctly.

packages/components/src/components/NewModal/NewModal.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+        <ModalComponent.Button variant="tertiary">Secondary Action</ModalComponent.Button>
+    </>
+);
+
+const meta: Meta = {
+    title: 'NewModal',
+    component: ModalComponent,
+    decorators: [
+        Story => (
+            <ThemeProvider theme={intermediaryTheme.light}>
```
jvaclavik, line 19: Are we going to be able to switch the theme somehow?

packages/components/src/components/NewModal/NewModal.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+        },
+        heading: {
+            control: 'text',
+        },
+        description: {
+            control: 'text',
+        },
+        onCancel: {
+            options: ['none', 'withCallback'],
+            mapping: { none: undefined, withCallback: () => console.log('close') },
```
jvaclavik: btw you could use storybook actions for these callbacks:


```tsx
import { action } from '@storybook/addon-actions';

...

mapping: { none: undefined, withCallback: action('onToggle') },

File: packages/components/src/components/NewModal/NewModal.tsx [hunk 3]
```diff
@@ -0,0 +1,208 @@
...
+import type { UISize } from '../../config/types';
+
+const NEW_MODAL_CONTENT_ID = 'modal-content';
+const MODAL_ELEVATION = 0;
+const ICON_SIZE = 40;
+
+type NewModalSize = Extract<UISize, 'large' | 'medium' | 'small'>;
+
+const getSize = (size: NewModalSize) => {
+    switch (size) {
```
jvaclavik: nit: we usually use mapping functions, but this is also fine

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+    width: ${({ $size }) => getSize($size)}px;
+    overflow: hidden;
+    background: ${mapElevationToBackground};
+    box-shadow: ${({ theme }) => theme.boxShadowElevated};
+`;
+
+const Header = styled.header`
+    display: flex;
+    align-items: center;
+    word-break: break-word;
```
jvaclavik, line 45: image

I believe close button should be on the top in case of long header

jvaclavik: nit: 

total edge case: 
image


when title is toooooo long it's not possible to reach other content.

I suggest to restrict title to two lines and add ellipsis. But I think we don't have component for it yet. Similar one could be `TruncateWithTooltip`

This is not necessary to fix.

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+const ScrollContainer = styled.div`
+    display: flex;
+    flex-direction: column;
+    flex: 1;
+    height: 100%;
+    overflow-y: auto;
+
+    ::-webkit-scrollbar {
+        display: none;
+    }
```
jvaclavik: Why are you hiding scrollbar? And why I can see it? I think I'm missing something 😅

image

packages/components/src/components/NewModal/NewModalBackdrop.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { useModalTarget } from './NewModalProvider';
+import type { UIHorizontalAlignment, UIVerticalAlignment } from '../../config/types';
+
+export type ModalAlignment = { x: UIHorizontalAlignment; y: UIVerticalAlignment };
+
+export type NewModalBackdropProps = {
+    onClick?: () => void;
+    children?: ReactNode;
+    alignment?: ModalAlignment;
+    padding?: number;
```
jvaclavik, line 15: What about adding here `SpacingValues`?

packages/components/src/components/NewModal/NewModalBackdrop.tsx
```diff
@@ -0,0 +1,61 @@
...
+const Wrapper = styled.div<{ $alignment: ModalAlignment; $padding: number }>`
+    position: absolute;
+    z-index: ${zIndices.modal};
+    inset: 0;
+    display: flex;
+    flex-direction: column;
+    padding: ${({ $padding }) => $padding}px;
+    overflow: auto;
+
+    /* backdrop-filter does not work in Firefox, use darker background instead */
```
jvaclavik: image

Are you sure?

btw could you add also `-webkit-backdrop-filter` for Safari?

packages/components/src/components/NewModal/NewModalBackdrop.tsx
```diff
@@ -0,0 +1,61 @@
...
+    backdrop-filter: blur(5px);
+    background: rgb(0 0 0 / 30%);
+
+    @supports not ((-webkit-backdrop-filter: blur(5px)) or (backdrop-filter: blur(5px))) {
```
jvaclavik: 👏

packages/components/src/components/NewModal/NewModalBackdrop.tsx
```diff
@@ -0,0 +1,61 @@
...
+        background: rgb(0 0 0 / 60%);
+    }
+
+    align-items: ${({ $alignment }) =>
+        // eslint-disable-next-line no-nested-ternary
+        $alignment.x === 'center' ? 'center' : $alignment.x === 'left' ? 'flex-start' : 'flex-end'};
+    justify-content: ${({ $alignment }) =>
+        // eslint-disable-next-line no-nested-ternary
+        $alignment.y === 'center' ? 'center' : $alignment.y === 'top' ? 'flex-start' : 'flex-end'};
```
jvaclavik: nit: I would rather use functions, but up to you

packages/components/src/components/modals/DialogModal/DialogModal.tsx
```diff
@@ -50,6 +50,9 @@ export interface DialogModalProps extends PickedModalProps {
     body?: ReactNode;
 }
 
+/**
+ * @deprecated Use NewModal.
```
jvaclavik, line 54: Nice! 👍

packages/components/src/utils/useScrollShadow.tsx
```diff
@@ -65,7 +65,8 @@ export const useScrollShadow = () => {
         }
     };
 
-    const { elevation } = useElevation();
+    const { parentElevation: elevation } = useElevation();
```
jvaclavik, line 68: 👍

packages/suite/src/views/suite/SwitchDevice/SwitchDevice.tsx
```diff
@@ -31,18 +32,20 @@ export const SwitchDevice = ({ cancelable, onCancel }: ForegroundAppProps) => {
...
-                        key={`${device.id}-${device.instance}`}
-                        device={device}
-                        instances={deviceUtils.getDeviceInstances(device, devices)}
-                        onCancel={onCancel}
-                        isFullHeaderVisible={index === 0}
-                    />
-                ))}
-            </Flex>
-        </SwitchDeviceRenderer>
+        <NewModal.Backdrop onClick={onCancel} alignment={{ x: 'left', y: 'top' }} padding={5}>
```
jvaclavik, line 35: I already love new modal 😀

jvaclavik (commented):
Really good job, Adam! 👍

🛳️🇮🇹

packages/components/src/components/NewModal/NewModal.tsx
jvaclavik: btw same issue for back button 

image

jvaclavik: Btw thinking about adding tooltip for close and back buttons. WDYT?

packages/components/src/components/NewModal/NewModal.stories.tsx
seibei-iguchi: Tried to do it but couldn't figure how. IMO we should skip the whole dual story preview thing and just make a theme switch in the Storybook.

packages/components/src/components/NewModal/NewModal.tsx
seibei-iguchi: This has been taken from the old modal, dunno

packages/components/src/components/NewModal/NewModalBackdrop.tsx
seibei-iguchi: I actually don't want that. It's for the device switch modal which requires specific padding not defined in the spacings.

packages/components/src/components/NewModal/NewModal.stories.tsx
jvaclavik: It's fine for now. In the future we could use 

image

https://storybook.js.org/docs/essentials/toolbars-and-globals

packages/components/src/components/NewModal/NewModalBackdrop.tsx
seibei-iguchi: Will do. Also taken verbatim from the old modal

packages/components/src/components/NewModal/NewModalBackdrop.tsx
jvaclavik: I see, it makes sense 👍

packages/components/src/components/NewModal/NewModal.tsx
jvaclavik: 🤔 I might delete it then

jvaclavik: WDYT?

packages/components/src/components/NewModal/NewModal.tsx
seibei-iguchi: From what I can tell from the Figma designs, it should be centered?

jvaclavik (approved):

packages/components/src/components/NewModal/NewModal.tsx
seibei-iguchi: Something like mapSizeToWidth?

packages/components/src/components/NewModal/NewModal.tsx
jvaclavik: I see, OK. This might be question for @Hermez-cz

seibei-iguchi: I believe it won't be such an issue if I restrict the title length

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -62,6 +53,13 @@ const HeadingContainer = styled.div`
     display: flex;
     flex-direction: column;
     flex: 1;
+    overflow: hidden;
+`;
+
+const Heading = styled(H3)`
```
jvaclavik, line 58: I'm not happy about this 🤔

Shouldn't we create some ellipsis prop directly in Heading? (Not necessarilu in this PR)

packages/components/src/components/NewModal/types.tsx
```diff
@@ -1,3 +1,7 @@
-import { UIVariant } from '../../config/types';
+import { UIVariant, UISize, UIHorizontalAlignment, UIVerticalAlignment } from '../../config/types';
 
 export type NewModalVariant = Extract<UIVariant, 'primary' | 'warning' | 'destructive'>;
+
+export type NewModalSize = Extract<UISize, 'large' | 'medium' | 'small'>;
```
jvaclavik: Btw this might be refactored in the future. Check this PR: https://github.com/trezor/trezor-suite/pull/13828

jvaclavik (approved):

packages/components/src/components/NewModal/NewModal.tsx
seibei-iguchi: Yeah, me neither. Due to how ellipsis work in css it can't be done any other way than directly styling the H3. I tried the Truncate component but it didn't work.

Issue #12074, opened 2024-04-19
Minor issues in Settings/Coins section


bosomt:
**Describe the bug**
Minor issues in Settings/Coins section
- `Activate coins` button is always active even device is disconnected, pressing it will end with forever loading state
- You are not allowed to change backends when device is disconnected
- `Connect device to change Crypto settings` message does not reflect that we renamed Coins section From Crypto back to Coins

**Info:**
 - Suite version: desktop 24.4.2 (22d7c411755389aef139a5c705da2dda7f3e4af1)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.4.2 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.7.0 regular (revision 45e8a842a31e62a6d43d7f6ccac62a45e1198ef0)
 - Transport: BridgeTransport 2.0.33

**Expected result:**

**Actual result:**
A clear and concise description of what actually happens.

**Screenshots:**
<img width="924" alt="image" src="https://github.com/trezor/trezor-suite/assets/31506317/46278d62-f17c-48bb-b1e4-ba23399a2ac3">

Lemonexe:
I took a look and can't reproduce the first issue - but maybe I am looking at wrong place..

EDIT: I was indeed looking at the wrong place, and both parts of the issue still stand.

Lemonexe:
Fully resolved by #13836

bosomt:
**QA OK**

- Activate coins button inactive ✅ 
- You are still not allowed to change backends when device is disconnected 🤷🏿‍♂️ , probably expected
- Connect device to change Coins settings ✅ 

**Info:**
 - Suite version: web 24.9.0 (5d77d96a2d5f48208e335047813c881415f6b31d)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:129.0) Gecko/20100101 Firefox/129.0
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.8.1 regular (revision 632b9561559b7ab6824bb7eeac072874e07b7b82)
 - Transport: BridgeTransport 2.0.32

PR #13818, opened 2024-08-15
fix(suite): Disable Activate Coins button when disconnected

Lemonexe:
When you select new coins but disconnect device before discovery, the button stays clickable but does not work.

## Description

- disable the button when there is discovery pending but device is disconnected
- add a tooltip
  - the tooltip is just my improvement idea - for consistency with the coin buttons
  - there are no design requirements AFAIK
  - ofc I am not a copywriter, so I chose an existing translation string 
  - if you think I was too bold choosing a tooltip message, I can remove it
- incidental fix of eslint warning for design system overloading

## Related Issue

Part 1/2 of #12074

## Screenshots:

discovery (Activate coins) button when disconnected and hovered on:
ActivateCoins disconnected

### DIFF
--- a/packages/suite/src/views/settings/SettingsCoins/SettingsCoins.tsx
+++ b/packages/suite/src/views/settings/SettingsCoins/SettingsCoins.tsx
@@ -6,7 +6,7 @@ import {
     startDiscoveryThunk,
     selectDeviceModel,
 } from '@suite-common/wallet-core';
-import { Button, motionEasing } from '@trezor/components';
+import { Button, motionEasing, Tooltip } from '@trezor/components';
 import { DeviceModelInternal } from '@trezor/connect';
 import type { Network } from 'src/types/wallet';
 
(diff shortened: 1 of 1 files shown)


packages/suite/src/views/settings/SettingsCoins/SettingsCoins.tsx
```diff
@@ -31,8 +31,9 @@ import { FirmwareTypeSuggestion } from './FirmwareTypeSuggestion';
 import { spacingsPx } from '@trezor/theme';
 import { selectSuiteFlags } from '../../../reducers/suite/suiteReducer';
 
-const StyledButton = styled(Button)`
```
Lemonexe, line 34: this was eslint warning for overloading, so I remade it as wrapper;
I needed a wrapper anyway for the `width: fit-content` (so that the tooltip is positioned correctly)

jvaclavik: Nice! Thank you for refactor overriding 🙏

jvaclavik (approved):

hstastna (approved):

PR #13820, opened 2024-08-15
fix(mobile): reverse order of transactions

Nodonisko:
## Description

Transactions was sorted from oldest on top but we want newest on top.

## Related Issue

Resolve


vytick (approved):
Good catch! 🚀

PR #13826, opened 2024-08-15
fix(suite): Fix button colors + subtle variant

jvaclavik:
## Description

- update button colors
- add `isSubtle` prop for less visible button
- buttons are not depending on elevation anymore 💪 

## Related Issue

Resolve 

## Screenshots:

image

image


seibei-iguchi:
Why were they dependent on elevation before? Does this also translate to IconButtons?

jvaclavik:
@adamhavel 

> Why were they dependent on elevation before? 

It was specified by designers and it already happened in few cases (tertiary button)

> Does this also translate to IconButtons?

Yes:

image

seibei-iguchi (approved):

PR #13830, opened 2024-08-15
chore(message-system): bump recommended and forced suite version

MiroslavProchazka:
Bumping  conditions for displaying Suite update banner through mesage system:

- Recommended to 24.6.3 (first stable version after intermediary)
- Forced to 24.10.1 (Suite release with support of TS3)

Tracked in Notion in [here](https://www.notion.so/satoshilabs/Bumping-recommended-and-forced-Suite-version-for-auto-update-df807653e6a941f98f5b08eeacef6446?pvs=4)

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
-    "timestamp": "2024-07-029T00:00:00+00:00",
-    "sequence": 60,
+    "timestamp": "2024-08-015T00:00:00+00:00",
+    "sequence": 61,
     "actions": [
         {
             "conditions": [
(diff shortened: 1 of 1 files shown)


suite-common/message-system/config/config.v1.json
```diff
@@ -245,14 +245,14 @@
             "conditions": [
                 {
                     "environment": {
-                        "desktop": "<23.7.2",
+                        "desktop": "<23.10.1",
                         "mobile": "!",
-                        "web": "<23.7.2"
+                        "web": "!"
```
Lemonexe, line 250: Is it intentional to exlude web?
When loading the web suite on production url, most clients will refetch the latest from server, but we can't rely on it 100%. Some custom browsers (or PWAs if we support that), may not always refetch, so it is possible to have an outdated client cached.
I'd guesstimate that it will be a very small minority of clients, but keeping the message for them does not cost us anything I think :slightly_smiling_face: 

But ofc idk what is desired..

Lemonexe (approved):
:rocket:  LGTM, with one small question

suite-common/message-system/config/config.v1.json
MiroslavProchazka: Hmm, that's interesting. @vdovhanych what do you think?

PR #13831, opened 2024-08-15
test(e2e): enabling node bridge in desktop tests

ondracja:
- reenabling node bridge usage instead of the legacy one

### DIFF
--- a/packages/suite-desktop-core/e2e/support/common.ts
+++ b/packages/suite-desktop-core/e2e/support/common.ts
@@ -25,9 +25,9 @@ export const launchSuite = async (params: LaunchSuiteParams = {}) => {
         args: [
             path.join(appDir, './dist/app.js'),
             `--log-level=${desiredLogLevel}`,
-            // '--bridge-node-test',
+            '--bridge-node-test',
             // uncomment to use legacy bridge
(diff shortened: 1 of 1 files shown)


packages/suite-desktop-core/e2e/support/common.ts
```diff
@@ -25,9 +25,9 @@ export const launchSuite = async (params: LaunchSuiteParams = {}) => {
         args: [
             path.join(appDir, './dist/app.js'),
             `--log-level=${desiredLogLevel}`,
-            // '--bridge-node-test',
+            '--bridge-node-test',
```
tomasklim, line 28: What should this do because I cannot find any results when searching for  `bridge-node-test` in the codebase

ondracja: I misunderstood the current state of node-bridge and its implementation in tests. I'm dropping this subject for the moment.

ondracja:
Not feasible at the moment.

tomasklim:
How about removing the commented-out code since it’s no longer relevant?

ondracja:
> How about removing the commented-out code since it’s no longer relevant?

removed in #13864

PR #13832, opened 2024-08-15
feat(message-system): add official lngs to message system

MiroslavProchazka:
- Adding Turkish and Brazilian Portuguese as language options in the message system banners.

- Adding also T3B1 to device list

I do not guarantee the accuracy of the translations; these will be sent to an agency for proofreading.


suite-common/message-system/config/config.v1.json
```diff
@@ -29,7 +29,9 @@
                     "hu": "A Coinjoin 2024. június 1-jén megszűnt a Trezor Suite-ban. Az Ön pénzeszközei továbbra is elérhetőek maradnak, és nincs szükség semmilyen intézkedésre.",
```
tomasklim, line 29: Screenshot 2024-08-15 at 22 08 39
please add official langs here and to required, remove community from required, thxs

tomasklim (changes requested):

suite-common/message-system/config/config.v1.json
MiroslavProchazka: Resolved in second commit

suite-common/message-system/schema/config.schema.v1.json
```diff
@@ -367,7 +368,7 @@
             "title": "Localization",
             "description": "A multilingual text localization.",
             "type": "object",
-            "required": ["en-GB", "en", "es", "cs", "ru", "ja"],
```
Lemonexe, line 370: Was it intentional to remove `"ru", "ja"` from the list, making them non-required?

Just making sure, because in PR description you wrote you are only adding :slightly_smiling_face:

MiroslavProchazka: Yes, these are community languages, so they are not required. Only official lngs should be imho.

tomasklim (approved):

PR #13834, opened 2024-08-15
chore(suite): define explorer queryString only if needed

tomasklim:
## Description

- does not make sense to define `queryString` for each network as it is used just for solana testnet

### DIFF
--- a/packages/suite/src/views/wallet/tokens/common/TokensList/TokenRow.tsx
+++ b/packages/suite/src/views/wallet/tokens/common/TokensList/TokenRow.tsx
@@ -115,7 +115,9 @@ const getTokenExplorerUrl = (network: Network, token: EnhancedTokenInfo) => {
 
     const contractAddress = network.networkType === 'cardano' ? token.fingerprint : token.contract;
 
-    return `${explorerUrl}${contractAddress}${network.explorer.queryString}`;
+    const queryString = network.explorer.queryString ?? '';
+
+    return `${explorerUrl}${contractAddress}${queryString}`;
 };
 
 interface TokenRowProps {
--- a/packages/suite/src/views/wallet/transactions/components/NoTransactions.tsx
+++ b/packages/suite/src/views/wallet/transactions/components/NoTransactions.tsx
@@ -10,7 +10,8 @@ interface NoTransactionsProps {
 
 export const NoTransactions = ({ account }: NoTransactionsProps) => {
     const network = getNetwork(account.symbol)!;
-    const explorerUrl = `${network.explorer.account}${account.descriptor}${network.explorer.queryString}`;
+
+    const explorerUrl = `${network.explorer.account}${account.descriptor}${network.explorer.queryString ?? ''}`;
 
     return (
         <AccountExceptionLayout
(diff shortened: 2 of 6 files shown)


suite-common/wallet-types/src/backend.ts
```diff
@@ -36,10 +36,10 @@ export interface Blockchain extends ConnectionStatus {
     explorer: {
         tx: string;
         account: string;
-        queryString: string;
-        nft?: string;
         address: string;
+        nft?: string;
         token?: string;
+        queryString?: string;
```
Lemonexe: This seems to duplicate the newly created `Explorer` type in `networksConfig`.
I'd suggest exporting `Explorer` and using it here; dependency tree will be unchanged

packages/suite/src/views/wallet/transactions/components/NoTransactions.tsx
```diff
@@ -10,7 +10,8 @@ interface NoTransactionsProps {
 
 export const NoTransactions = ({ account }: NoTransactionsProps) => {
     const network = getNetwork(account.symbol)!;
-    const explorerUrl = `${network.explorer.account}${account.descriptor}${network.explorer.queryString}`;
+
+    const explorerUrl = `${network.explorer.account}${account.descriptor}${network.explorer.queryString || ''}`;
```
Lemonexe: Same nit as above.
```suggestion
    const explorerUrl = `${network.explorer.account}${account.descriptor}${network.explorer.queryString ?? ''}`;
```

packages/suite/src/views/wallet/tokens/common/TokensList/TokenRow.tsx
```diff
@@ -115,7 +115,9 @@ const getTokenExplorerUrl = (network: Network, token: EnhancedTokenInfo) => {
 
     const contractAddress = network.networkType === 'cardano' ? token.fingerprint : token.contract;
 
-    return `${explorerUrl}${contractAddress}${network.explorer.queryString}`;
+    const queryString = network.explorer.queryString || '';
```
Lemonexe: Tiny nit :nerd_face: 
I think `??` is more suitable, because we expect only `undefined`, not falsy values in general.
```suggestion
    const queryString = network.explorer.queryString ?? '';
```

Lemonexe (changes requested):
Overall LGTM with just small nits, feel free to ignore them, but there is also a [TS error here](https://github.com/trezor/trezor-suite/blob/c59967ff4038143da2124b1bfd6c940321dea526/suite-common/wallet-core/src/blockchain/blockchainReducer.ts#L112), please take a look.

suite-common/wallet-types/src/backend.ts
tomasklim: great catch be6e52eed0335c23917d1842b2d96afcad787d45

packages/suite/src/views/wallet/tokens/common/TokensList/TokenRow.tsx
tomasklim: be6e52eed0335c23917d1842b2d96afcad787d45

packages/suite/src/views/wallet/transactions/components/NoTransactions.tsx
tomasklim: be6e52eed0335c23917d1842b2d96afcad787d45

Lemonexe (approved):
LGTM, thanks for addressing even the nitpicky nits 😄

jvaclavik (approved):

PR #13835, opened 2024-08-15
feat(suite-native): dashboard networks always in same order

### LINKED ISSUE #13766: Coin Enabling - ensure networks are in the same order all the time
No matter when you enabled the new network, still keep the same order of networks all the time.

vytick:
On dashboard always show networks in the same order no matter when the network wall enabled

## Related Issue

Resolve #13766

### DIFF
--- a/suite-native/assets/src/assetsSelectors.ts
+++ b/suite-native/assets/src/assetsSelectors.ts
@@ -1,4 +1,5 @@
 import { memoize } from 'proxy-memoize';
+import { A, F, pipe } from '@mobily/ts-belt';
 
 import { BigNumber } from '@trezor/utils/src/bigNumber';
 import { networks, NetworkSymbol } from '@suite-common/wallet-config';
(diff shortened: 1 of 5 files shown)


suite-native/assets/src/assetsSelectors.ts
```diff
@@ -58,6 +59,12 @@ export const selectDeviceAssetsWithBalances = memoize(
         const fiatCurrencyCode = selectFiatCurrencyCode(state);
 
         return deviceNetworksWithAssets
+            .sort((a, b) => {
```
Nodonisko: Sort of dangerous because it mutates original object, use A.sort or toSorted

Nodonisko (changes requested):

suite-native/assets/src/assetsSelectors.ts
vytick: 2513f11206

Nodonisko (approved):

vytick:
Bigger refactor of this file happens in https://github.com/trezor/trezor-suite/pull/13876 so we agreed with @Nodonisko to close this PR in favour of #13876 where this change will be included.

PR #13849, opened 2024-08-16
refactor(suite): Refactor banners

jvaclavik:
## Description

- Use existing component `Warning` instead of custom one `Banner`
- Banner was deleted
- add traffic lights for custom messages

## Related Issue

Resolve 

## Screenshots:

before
image

after
image

image

image

multiple banners
image

custom message:

welcome
image

dashboard
image


packages/suite/src/components/suite/banners/MessageSystemBanner.tsx
```diff
@@ -64,11 +55,31 @@ export const MessageSystemBanner = ({ message }: MessageSystemBannerProps) => {
...
-        <BannerOnTop
+        <Warning
+            icon
             variant={variant === 'critical' ? 'destructive' : variant}
-            body={content[language] || content.en}
-            action={actionConfig}
-            dismissal={dismissalConfig}
-        />
+            rightContent={
+                <Row gap={8}>
```
seibei-iguchi, line 62: I know this is handled by typescript, but I just want to make sure that we want to use it like this and not explicitly via spacingsPx?

seibei-iguchi (approved):

packages/suite/src/components/suite/banners/MessageSystemBanner.tsx
jvaclavik: I will fix it in next PR

PR #13853, opened 2024-08-16
chore(suite): Optimize webpack dev server

Lemonexe:
## Description

Configure webpack for dev server to **not** watch `node_modules`, as it creates unnecessary amount of file watchers, potentially degrading performance. This is suggested [by webpack](https://webpack.js.org/configuration/watch/#watchoptionsignored).

Specifically, this may sometimes cause issues on Linux Ubuntu, where there is default limit of 65536 file watchers in the `inotify` service, and running suite dev server may exceed it. Though even after this PR, [increasing the limit](https://stackoverflow.com/questions/55763428/react-native-error-enospc-system-limit-for-number-of-file-watchers-reached/55763478#55763478) may be necessary for linux devs. Our codebase is just that big :open_mouth:

## Related Issue

[Slack discussion](https://satoshilabs.slack.com/archives/G019WLX2P7B/p1720697642356569?thread_ts=1720686830.567509&cid=G019WLX2P7B) (for internal users only)

### DIFF
--- a/packages/suite-build/configs/dev.webpack.config.ts
+++ b/packages/suite-build/configs/dev.webpack.config.ts
@@ -21,6 +21,10 @@ const config: webpack.Configuration = {
         filename: 'js/[name].js',
         chunkFilename: 'js/[id].js',
     },
+    watchOptions: {
+        // reduce number of file watchers; for HMR it is not necessary to watch both source code & node_modules
+        ignored: /node_modules/,
+    },
     plugins: [
         new WebpackPluginServe({
             port: DEV_PORTS[project],


tomasklim:
How about our packages? isn't it considered as `node_modules`? e.g. importing stuff from `suite-common`?

Lemonexe:
> How about our packages? isn't it considered as `node_modules`? e.g. importing stuff from `suite-common`?

@tomasklim Hmm, I am not that knowledgable about the ínner workings of our package system, so I may be wrong, but this is how I understand it:
- I make a change in any file, let's say `package/react-utils/index.ts`
- `react-utils`  is rebuilt
- So there is now a corresponding change in `node_modules/@trezor/react-utils/index.ts` 
- In current codebase, both `packages/` and `node_modules/` are watched.
- With this change, only `packages/` is watched, but it's sufficient
  - I have verified that HMR still works :heavy_check_mark: ,  it is still triggered by the change in `packages/`  

**TL;DR:** So it seems there is no need to watch `node_modules` for `@trezor` packages, because AFAIK they cannot be modified without having been modified in `packages/` first, and that's still watched.
Thoughts? :thinking:

tomasklim (approved):
LGTM

komret (approved):
🚀

PR #13857, opened 2024-08-16
chore(suite): change url in security check fail

MiroslavProchazka:
Changing the URL for Device authenticity fail + Firmware revision fail events, which will start dedicated chat-bot flow asking user to fill in several information rightaway.

### DIFF
--- a/packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
+++ b/packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
@@ -2,7 +2,7 @@ import styled from 'styled-components';
 
 import { H2, Row, Text, useElevation } from '@trezor/components';
 import { Elevation, mapElevationToBorder, spacings, spacingsPx } from '@trezor/theme';
-import { TREZOR_SUPPORT_URL } from '@trezor/urls';
+import { TREZOR_SUPPORT_FW_CHECK_FAILED } from '@trezor/urls';
 
 import { Translation, TrezorLink } from 'src/components/suite';
 import { SecurityChecklist } from '../../../views/onboarding/steps/SecurityCheck/SecurityChecklist';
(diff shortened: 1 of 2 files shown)


Lemonexe:
Looks good to me, but as long as the new URL does not exist, I can't give approve.

peter-sanderson (approved):

MiroslavProchazka:
> Looks good to me, but as long as the new URL does not exist, I can't give approve.

The url will live in the upcoming days, the article is in the final phase. I wanted to add it to the upcoming suite release, so we dont need to cherry-pick it later. KB article can be published minutes before production release and it will work just fine,.

komret (changes requested):
Please prefix the constant name with `TREZOR_SUPPORT_` and move it lower in the file among the other support URLs.

We already have [HELP_CENTER_FIRMWARE_REVISION_CHECK](https://trezor.io/learn/a/trezor-firmware-revision-check), which is used in the banner. Shouldn't it also be replaced with this one?

Lemonexe (approved):
> The url will live in the upcoming days, the article is in the final phase. I wanted to add it to the upcoming suite release, so we dont need to cherry-pick it later. KB article can be published minutes before production release and it will work just fine,.

Alright, I just wanted to check we don't accidentally merge something not working :slightly_smiling_face:

MiroslavProchazka:
> We already have [HELP_CENTER_FIRMWARE_REVISION_CHECK](https://trezor.io/learn/a/trezor-firmware-revision-check), which is used in the banner. Shouldn't it also be replaced with this one?

No, The goal is to create a separate article with a step-by-step guide on what users should do if they encounter the Ghost modal. A dedicated chatbot flow will begin as soon as the user lands on this specific URL. We decided not to use the existing article because it might alarm users who are casually browsing the knowledge base. The new article will not be indexed, so users should only access it by clicking the button on the Ghost modal, not through regular internet browsing.

MiroslavProchazka:
> Please prefix the constant name with `TREZOR_SUPPORT_` and move it lower in the file among the other support URLs.
> 
> We already have [HELP_CENTER_FIRMWARE_REVISION_CHECK](https://trezor.io/learn/a/trezor-firmware-revision-check), which is used in the banner. Shouldn't it also be replaced with this one?

edfbd1374506e705d6f015de7b5e1b675a73a6ad

komret (approved):

PR #13858, opened 2024-08-16
feat(suite-native): for coin enabling show first empty account

### LINKED ISSUE #13844: Coin Enabling - show at least one account for each enabled coin
When coin (and coin enabling in general) is enabled, and no account with txns exist, show at least one default account (empty). 

If coin enabling is disabled only discover non empty accounts

vytick:
## Description

If coin enabling is enabled discover at least first normal account even if it is empty

## Related Issue

Resolve #13844

## Screenshots:

https://github.com/user-attachments/assets/560c2deb-0ca7-4883-b7bf-1202829f7545

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -25,6 +25,7 @@ import { AccountType, Network, NetworkSymbol, getNetworkType } from '@suite-comm
 import { DiscoveryStatus } from '@suite-common/wallet-constants';
 import { requestDeviceAccess } from '@suite-native/device-mutex';
 import { analytics, EventType } from '@suite-native/analytics';
+import { FeatureFlag, selectIsFeatureFlagEnabled } from '@suite-native/feature-flags';
 
 import { selectDiscoveryInfo, setDiscoveryInfo } from './discoveryConfigSlice';
 import {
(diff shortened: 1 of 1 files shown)


suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -293,13 +294,25 @@ const discoverAccountsByDescriptorThunk = createThunk(
...
+                    getState(),
+                    FeatureFlag.IsCoinEnablingActive,
+                );
+
+                const isVisible =
+                    // If the account is not empty, it should be visible.
+                    !accountInfo.empty ||
+                    // If the feature flag is enabled, first normal account should be visible.
+                    (isCoinEnablingActive &&
+                        bundleItem.accountType === 'normal' &&
```
Nodonisko, line 307: Why only normal type and why only 0 index? Shoudn't it be last account instead of first?

vytick: @Nodonisko 
First empty normal account because there might be more invisible accounts and we want to show user the first possible and not to skip any. If user had empty account on index0 and visible on index 1, he could have received there funds and this would break discovery causing index 1 not to be discovered since we stop on first empty account.

Normal account because we dont want to show empty account for for example ledger or legacy types only this default(normal) is visible if empty for enabled coins.

This logic is similar to what desktop does.

Nodonisko (approved):

PR #13869, opened 2024-08-19
fix(suite): do not override BigNumber config as it is set to 1e9 anyway

tomasklim:
## Description

- in case user used number input and he staked big amount of ETH, it broke his app 
- it is not required to change the config anymore as we do it globaly to pretty big number
- as it goes to release, I wanted to keep it as safe as possible

**How to test this?**
- extend `ETH staking` e2e test to first go into send form, fill really long number into amount form (need to be valid), has to be more than 21 numbers and then navigate to staking dashboard with the mocked data in the existing test (the account has 1234 ETH, so you should be able to fill that long amount)

## Related Issue

Resolve sentry issue
Screenshot 2024-08-19 at 11 42 13

### DIFF
--- a/packages/suite/src/components/suite/NumberInput.tsx
+++ b/packages/suite/src/components/suite/NumberInput.tsx
@@ -59,14 +59,7 @@ const cleanValueString = (value: string, locale: Locale) => {
         cleanedValue = cleanedValue.slice(0, cleanedValue.length - 1);
     }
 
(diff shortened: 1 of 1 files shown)


martykan (approved):

STew790:
**QA OK**
After changing the test as instructed, the test went OK.

image

PR #13875, opened 2024-08-19
fix(connect-popup): theme provider on logs page

martykan:
## Description

Fix theme injected in ThemeProvider on logs page in popup. 
There were some changes related to the Button component which broke it and required updating the theme.


mroz22 (approved):
nice thanks for proactively fixing it!

PR #13879, opened 2024-08-19
fix(suite-native): currency selector improvements

### LINKED ISSUE #7197: FIAT currency improvements
1. Add FIAT shortcut (e.g. Euro - EUR) in front of currency when user picks the FIAT currency. It is visible when I pick it but not in the toaster selection. See printscreens
2. Implement latest version of currency settings from connect

EDIT Jul 7, 2024: Align the labels with the desktop, but do not add BTC/ETH for now as it is not yet on desktop and do not rename the tab.

When user picks:
image

When user go through the selection:
image

yanascz:
Labels unified with desktop Suite, fiat currency symbol also shown in the drop-down menu.
The `fiatCurrencies` constant has been restructured to better match semantic of individual properties.

## Related Issue

Resolve [7197](https://github.com/trezor/trezor-suite/issues/7197)

## Screenshots:

image
image
image


suite-native/module-settings/src/components/CryptoUnitsSelector.tsx
```diff
@@ -25,7 +25,7 @@ export const CryptoUnitsSelector = () => {
 
     return (
         <Select<PROTO.AmountUnit>
-            selectLabel="Bitcoin Amount Units"
+            selectLabel="Bitcoin units"
```
vytick: I know this is not part of the task, but since you are touching this, what do you think about adding intl (change selectLabel to ReactNode and use ) here, so we continue gradually adding translations everywhere.

suite-native/module-settings/src/components/CurrencySelector.tsx
```diff
@@ -27,9 +27,8 @@ export const CurrencySelector = () => {
     return (
         <Select<FiatCurrencyCode>
             items={fiatCurrencyItems}
-            selectLabel="Currency"
-            selectValue={selectedFiatCurrency.label}
-            valueLabel={selectedFiatCurrency.label.toUpperCase()}
+            selectLabel="Fiat currency"
```
vytick: also intl here?

suite-native/module-settings/src/components/CryptoUnitsSelector.tsx
yanascz: Yep, I thought about it as well but decided not to change it now. I'm not sure ReactNode is a way to go, would prefer translation ID instead. And that's probably something that might deserve a team discussion/consensus first.

vytick: 💪  I am fine with translationID as well, since this is only used twice, it wont make any difference.

Nodonisko: We already use ReactNode is most of places so I would use that also here. Passing only ID is problematic when you have translation with variables because you need to add also props for these variables and then typing become issue etc etc.

suite-native/module-settings/src/components/CurrencySelector.tsx
yanascz: Done.

suite-native/module-settings/src/components/CryptoUnitsSelector.tsx
yanascz: `<Translation ... />` used.

Nodonisko (approved):

vytick (approved):
Tested, works nice. Good job and welcome in the team! :) 🚀

PR #13883, opened 2024-08-19
feat(transport): add DebugLink support to NodeUsbTransport and UdpTransport

szymonlesisz:
…

## Description

Add support for `DebugLink` messages.
Using debug link interface would not break current workflow (like SignTx) and make it possible to send `DebugLinkDecisions`

useful for testing on physical device (debug link build FW) and emulator without `trezor-user-env`

## TODO and to discuss:
- [ ] add tests
- [ ] add support to `WebUsbTransport` ?
- [ ] add support to `transport-bridge/TrezordNode` ? (add /debug endpoints like in old bridge) ?
- [ ]  should it be possible to run a Transport with and without `debugLink` simultaneously? (then `debugLink` param should be passed to each api method: open, close, read, write)

### DIFF
--- a/packages/transport/src/api/udp.ts
+++ b/packages/transport/src/api/udp.ts
@@ -18,9 +18,11 @@ export class UdpApi extends AbstractApi {
 
     private enumerationTimeout: ReturnType<typeof setTimeout> | undefined;
     private enumerateAbortController = new AbortController();
+    private debugLink?: boolean;
 
-    constructor({ logger }: AbstractApiConstructorParams) {
+    constructor({ logger, debugLink }: AbstractApiConstructorParams & { debugLink?: boolean }) {
         super({ logger });
+        this.debugLink = debugLink;
     }
 
     listen() {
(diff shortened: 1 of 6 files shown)


mroz22:
I just got to this PR after some time. As it is, it looks alright but as you pointed out - it needs some tests to make it actually useful

> add tests

Yes, it would make sense maybe as part of USB API test that can be run with real device that we have here https://github.com/trezor/trezor-suite/blob/develop/packages/transport-test/e2e/api/api.test.ts#L46 

> add support to WebUsbTransport ?
> add support to transport-bridge/TrezordNode ? (add /debug endpoints like in old bridge) ?

My first impression is that it should be enough to test it at the lowest level possible (USB API).

> should it be possible to run a Transport with and without debugLink simultaneously? (then debugLink param should be passed to each api method: open, close, read, write)

Hmm do we need it actually? 🤔 On the other hand, because of THP you do need to use this on the transport layer and not only on the API layer I guess?

mroz22 (approved):

PR #13890, opened 2024-08-20
fix(components): scroll container gradient color in NewModal component

seibei-iguchi:


jvaclavik (approved):
👍

nit: we might delete elevation from useShadow and use it only this way if it's not working correctly

PR #13901, opened 2024-08-20
fix(connect): skip tests for unsupported coins on new devices

martykan:
## Description

Found this while working on #13865, decided it would probably be a better idea to make a separate PR for this. 

The nightly Connect tests currently run even for coins that are no longer supported on the Safe series - specifically NEM, EOS, Bgold, Dash, Decred

This PR skips the NEM, EOS suites altogether for these models.
Since the other coins are Bitcoin-based, I added a new rule format to the legacyTest rules, which can be used to exclude device models - eg. `'!T2B1'`


mroz22 (approved):
makes sense

PR #13903, opened 2024-08-20
fix(suite): select overlapping with modal

### LINKED ISSUE #13867: Select & Modal overlap
## Describe the bug
When the Select and Modal components are active simultaneously, the Select menu overlaps the Modal.

## Info
The CSS property `z-index` is the same for both components. 

## How to reproduce
1. Create a Select component using `menuIsOpen` property.
2. Open a Modal when the Select options are opened. 

## Expected behavior
The Modal should appear in front of the Select option menu.

## Screenshots
obrazek

seibei-iguchi:
Selects had z-index set to the same value as modals because of the case when they were rendered inside of one. This could cause an overlap when a modal was opened while a select outside the modal was active. 

### Before:
Screenshot 2024-08-20 at 20 46 41

### After:
Screenshot 2024-08-20 at 20 45 06

### DIFF
--- a/packages/components/src/components/form/Select/Select.tsx
+++ b/packages/components/src/components/form/Select/Select.tsx
@@ -34,10 +34,11 @@ const reactSelectClassNamePrefix = 'react-select';
 const createSelectStyle = (
     theme: DefaultTheme,
     elevation: Elevation,
+    isRenderedInModal: boolean,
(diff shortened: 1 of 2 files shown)


jvaclavik (approved):
Nice 👏 

Just thinking if this problem is just with the modal or also with something else. But let's merge it and see

PR #13904, opened 2024-08-21
chore(suite): add data-testid for Staking tab

komret:


komret:
Will be added elsewhere.

PR #13914, opened 2024-08-21
Feat(suite): Enhance HiddenPlaceholder

### LINKED ISSUE #10567: Improve the "Hide balances" feature

Lemonexe:
Enhance it to actually replace the characters with `####`, making them completely illegible.

## Description

TODO. WIP

- originally I tried to replace the number with random characters 
  - then I tried replacing it with constant four characters `####`
  - it does not work for short strings - when the `####` is wider than 0, it leads to horrible flickering if you hover at certain place
  - instead it replaces with as many `#` as per original length, but max 4
  - still not ideal, at some places it can still flicker, TBD

## Related Issue

Resolve #10567

## Screenshots:

Uncovered numbers for reference:
uncovered
Original, only blurred, but you can kinda discern the numbers:
only blurred
Blured and redacted:
blurred and redacted
Blured and redacted onhover:
blurred and redacted onhover

### DIFF
--- a/packages/react-utils/jest.config.js
+++ b/packages/react-utils/jest.config.js
@@ -1,0 +1,8 @@
+const baseConfig = require('../../jest.config.base');
+
+module.exports = {
+    ...baseConfig,
+    collectCoverage: true,
+    collectCoverageFrom: ['src/**/*.ts'],
+    testEnvironment: '../../JestCustomEnv.js',
+};
--- a/packages/react-utils/src/rewriteReactNodeRecursively.ts
+++ b/packages/react-utils/src/rewriteReactNodeRecursively.ts
@@ -1,0 +1,37 @@
+import { ReactNode, isValidElement, cloneElement, Key } from 'react';
+
+type RewriteReactNodeRecursivelyCallback = (stringSubNode: string) => string;
+
+/**
+ * Recursively crawls ReactNode and applies callback on all string subnodes
+ * @param node ReactNode to be recursively crawled & modified
+ * @param callback function to be applied on all string subnodes
+ * @param reactKey optional React key applied on the root node, if it is a ReactElement. Used internally.
+ * @returns modified ReactNode
+ */
+export const rewriteReactNodeRecursively = (
+    node: ReactNode,
+    callback: RewriteReactNodeRecursivelyCallback,
+    reactKey?: Key,
+): ReactNode => {
+    if (typeof node === 'string') return callback(node);
+    if (typeof node === 'number') return callback(node.toString());
+
+    if (Array.isArray(node)) {
+        return node.map((child, index) => {
+            const newReactKey: Key | undefined = isValidElement(child)
+                ? child.key ?? index
+                : undefined;
+
+            return rewriteReactNodeRecursively(child, callback, newReactKey);
+        });
+    }
+
+    if (isValidElement(node)) {
+        const rewrittenChildren = rewriteReactNodeRecursively(node.props.children, callback);
+
+        return cloneElement(node, { ...node.props, key: reactKey, children: rewrittenChildren });
+    }
+
+    return node;
+};
--- a/packages/suite/src/components/suite/HiddenPlaceholder.tsx
+++ b/packages/suite/src/components/suite/HiddenPlaceholder.tsx
@@ -3,7 +3,7 @@ import styled, { css } from 'styled-components';
 import { useSelector } from 'src/hooks/suite';
 import { selectIsDiscreteModeActive } from 'src/reducers/wallet/settingsReducer';
 import { rewriteReactNodeRecursively } from '@trezor/react-utils';
-import { redactNumericalSubstring } from '@suite-common/wallet-utils';
+import { redactNumericalSubstring, RedactNumbersContext } from '@suite-common/wallet-utils';
 
 interface WrapperProps {
     $intensity: number;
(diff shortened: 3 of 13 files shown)


packages/react-utils/jest.config.js
```diff
@@ -0,0 +1,8 @@
+const baseConfig = require('../../jest.config.base');
+
+module.exports = {
+    ...baseConfig,
+    collectCoverage: true,
+    collectCoverageFrom: ['src/**/*.ts'],
+    testEnvironment: '../../JestCustomEnv.js',
+};
```
Lemonexe, line 8: `react-utils` did not have unit tests at all. I set it up for the first test. Code coverage is not great :smile:

packages/react-utils/src/rewriteReactNodeRecursively.ts
```diff
@@ -0,0 +1,37 @@
...
+                : undefined;
+
+            return rewriteReactNodeRecursively(child, callback, newReactKey);
+        });
+    }
+
+    if (isValidElement(node)) {
+        const rewrittenChildren = rewriteReactNodeRecursively(node.props.children, callback);
+
+        return cloneElement(node, { ...node.props, key: reactKey, children: rewrittenChildren });
```
Lemonexe, line 33: **TL;DR: Should I scrap this?** :thinking: 

My idea was to have `HiddenPlaceholder.tsx` alone handle redacting numbers, so you would just wrap any content with it and it'd be completely automatic.
As I wrote in the comment there, this only works for simple cases of components wrapping children, such as:
`<Wrapper>Label: <SomeStyle>12345 CZK</SomeStyle></Wrapper>`.

But this procedure simply stops when it encounters more complex components, which calculate their content instead of just wrapping children, like  `<Formatter value={1234} />`. They're returned "as is".

So those components have to do the redacting themselves: see `prepareFiatAmountFormatter.ts`, `prepareCryptoAmountFormatter.ts`.

I'm not happy with this mixed approach, where _sometimes_ it does things automagically under the hood, but _sometimes_ you just have to do it yourself. I fear this might lead to confusion and mistakes later on.
So should we drop this magic, and redact numbers at all specific places where we display them?
Reminder: this won't make `HiddenPlaceholder.tsx` useless, it still does CSS blur, and will provide the context for `shouldRedactNumbers`.

packages/suite/src/components/suite/HiddenPlaceholder.tsx
```diff
@@ -34,11 +36,13 @@ export const HiddenPlaceholder = ({
     children,
     enforceIntensity,
     className,
-    ...rest
+    'data-testid': dataTestId,
```
Lemonexe, line 39: incidental cleanup, `rest` did not contain anything else

packages/react-utils/src/rewriteReactNodeRecursively.ts
```diff
@@ -0,0 +1,37 @@
...
+    callback: RewriteReactNodeRecursivelyCallback,
+    reactKey?: Key,
+): ReactNode => {
+    if (typeof node === 'string') return callback(node);
+    if (typeof node === 'number') return callback(node.toString());
+
+    if (Array.isArray(node)) {
+        return node.map((child, index) => {
+            const newReactKey: Key | undefined = isValidElement(child)
+                ? child.key ?? index
```
Lemonexe, line 23: When rewriting the `ReactNode`, I **do** have to provide key even when the original element did not.
This is not an edge case, it's actually very common:
`<span>label<i>123</i></span>`
the `span` has children = array `["label", <i>123</i>]`. Now the `<i>` has to have a `key`, idk why :facepalm:

suite-common/formatters/src/makeFormatter.tsx
```diff
@@ -40,11 +44,11 @@ export const makeFormatter = <TInput, TOutput, TDataContext extends DataContext
...
-    formatter.displayName = displayName;
+    const FormatterComponent: Formatter<TInput, TOutput, TDataContext> = props =>
+        <>{format(props.value, props, useShouldRedactNumbers())}</> ?? null;
+    FormatterComponent.displayName = displayName;
 
-    formatter.format = (value, dataContext = {}) => format(value, dataContext);
+    FormatterComponent.format = (value, dataContext = {}) => format(value, dataContext);
 
-    return formatter;
+    return FormatterComponent;
```
Lemonexe, line 53: It is actually a React component, so its name should be in CamelCase.
Otherwise I get an unjustified error from eslint rule of hooks because of calling `useShouldRedactNumbers`.

suite-common/wallet-utils/src/discreetModeUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+type RedactNumbersContextData = { shouldRedactNumbers: boolean };
+export const RedactNumbersContext = createContext<RedactNumbersContextData>({
+    shouldRedactNumbers: false,
+});
+/**
+ * Determine whether we are under a component which currently requests to redact the numbers for discreet mode (see HiddenPlaceholder).
+ * @returns shouldRedactNumbers whether numbers should be redacted in displayed output
+ */
+export const useShouldRedactNumbers = () =>
+    useContext(RedactNumbersContext)?.shouldRedactNumbers ?? false;
```
Lemonexe, line 32: Not sure if all of this belongs to `wallet-utils`.  Seemed suitable to me. Where would you place it?

Lemonexe:
Closing in favor of https://github.com/trezor/trezor-suite/pull/14528

PR #13916, opened 2024-08-21
chore(suite-desktop-core): TrezordNode in ThreadProxy

martykan:
## Description

Utilize ThreadProxy (which is also used for coinjoin) to run TrezordNode in a separate thread. 

Needs some more testing & maybe polishing.

### DIFF
--- a/packages/suite-desktop-core/src/modules/bridge.ts
+++ b/packages/suite-desktop-core/src/modules/bridge.ts
@@ -8,8 +8,7 @@ import { validateIpcMessage } from '@trezor/ipc-proxy';
 import { app, ipcMain } from '../typed-electron';
 import { BridgeProcess } from '../libs/processes/BridgeProcess';
 import { b2t } from '../libs/utils';
-import { Logger } from '../libs/logger';
-import { convertILoggerToLog } from '../utils/IloggerToLog';
+import { ThreadProxy } from '../libs/thread-proxy';
 
 import type { Module, Dependencies } from './index';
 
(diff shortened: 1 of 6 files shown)


packages/suite-desktop-core/src/modules/bridge.ts
```diff
@@ -71,35 +97,24 @@ const getBridgeInstance = (store: Dependencies['store']) => {
         legacyBridgeReasonRollout ||
         !allowPrerelease
     ) {
-        return new BridgeProcess();
+        //return new BridgeProcess();
```
martykan: remove before merging

marekrjpolak:
I've made some changes, hopefully useful:

- https://github.com/trezor/trezor-suite/pull/13916/commits/6e64887c59441d9d0784dca013c2df727b6967bf - `TrezordNode.start` method is now correctly rejected in case of failure (e.g. 21325 occupied by trezord); it seems that unhandled errors in utility process kill it immediately
- https://github.com/trezor/trezor-suite/pull/13916/commits/08a76ae3a794ba8dfa65e3d6b53cad040064ab52 - some thread proxy improvements
  - proxied object doesn't have to be `EventEmitter` anymore
  - `ThreadProxy` also not an event emitter, instead it has two internal emitters: one for proxied events from the utility process (`subscribe`/`unsubscribe` methods) and one for lifecycle events of process itself (`watch`/`unwatch` methods)
  - apart from already existing `disposed` event, `started` and `exited` were added as well
- https://github.com/trezor/trezor-suite/pull/13916/commits/a358b35d09040c17e6dc661f1d2a8b03430d89bc - CJ proxy adjusted to changes from previous commit
- https://github.com/trezor/trezor-suite/pull/13916/commits/e56d5eeae5c85e6d3f152fc3df247942de377f82 - Proxified `TrezordNode` improvements
  - `TrezordNodeThread` removed and bare `TrezordNode` used instead of it in utility process (it existed just to extend `EventEmitter` which was required at that time, am I right?)
  - `TrezordNodeProcess` now watches `started` event and calls start method on it (without that, respawned process is doing nothing)
  - `stop` method now dispose the process so it's not idling in the background (again in case of occupied 21325, where the unused process even respawns for nothing)

Does it make sense? There are definitely some discussion-worthy issues, as that dying utility process is logged as error even in usual cases so it may spam Sentry a bit, dunno.

mroz22 (approved):
works for me 🥳

PR #13919, opened 2024-08-21
fix: do not fetch fiat rates for uknown tokens

Nodonisko:
## Description

I noticed that mobile app is trying to fetch quite often tokens that doesn't have definitions resulting in wasted failed call.

## Related Issue

Resolve


AdamSchinzel (approved):
LGTM

PR #13894, opened 2024-08-20
fix(suite): fix missing fee rate for ada

MiroslavProchazka:
Fixing missing fee rate string in ADA send confirmation 
Resolves https://github.com/trezor/trezor-suite/issues/13882
image

PR #13920, opened 2024-08-21
chore(suite): Exhaustive typing for TransactionReviewSummary fee description

Lemonexe:
## Description

Following https://github.com/trezor/trezor-suite/pull/13894, clean up the boilerplate using a mapper which:
- gets the translationId per networkType
- does so in a type-exhaustive way!

## Screenshots:

Example with BTC TEST (still has "Fee rate"):
image

### DIFF
--- a/packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewSummary.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewSummary.tsx
@@ -10,6 +10,7 @@ import { Account, Network } from 'src/types/wallet';
 import { GeneralPrecomposedTransactionFinal, StakeType } from '@suite-common/wallet-types';
 import { useSelector } from 'src/hooks/suite/useSelector';
 import { selectLabelingDataForSelectedAccount } from 'src/reducers/suite/metadataReducer';
+import { NetworkType } from '@suite-common/wallet-config';
 
 const Wrapper = styled.div`
     padding: 20px 15px 12px;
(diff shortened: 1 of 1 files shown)


packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewSummary.tsx
```diff
@@ -190,6 +191,24 @@ const ReviewLeftDetailsLineRight = styled.div<{ $color: string; $uppercase?: boo
...
+        case 'ripple':
+            return 'TR_TX_FEE';
+        case 'solana':
+            return 'TR_TX_FEE';
+        case 'cardano':
+            return 'TR_TX_FEE';
+    }
+
+    // @ts-expect-error (must be unreachable code to make sure the switch is exhaustive)
+    return 'TR_FEE_RATE';
```
Lemonexe: Relying on unreachable code error is not the only way to do exhaustive typing. Absurd type declaration works as well, we have that already in [sendFormThunks](https://github.com/trezor/trezor-suite/blob/d4fd4e0af0d378db3fbb2f52d1f60d968f44d97a/suite-common/wallet-core/src/send/sendFormThunks.ts#L150-L177).
Which pattern do you like better?

komret: I'd prefer to use one of the approaches from our [code style guide](https://www.notion.so/satoshilabs/Good-Practices-Defensive-Programming-424acc122e3c429384141ab50b0aff5c). The example from `sendFormThunks` is unnecessarily lengthy in my opinion. The approach proposed here seems better to me, but the `@ts-expect-error` part is actually unnecessary, the compiler will warn you if you miss a case thanks to the return type. But the error message is not so clear in the case, which is why I prefer [type mapping](https://www.notion.so/satoshilabs/Good-Practices-Defensive-Programming-424acc122e3c429384141ab50b0aff5c?pvs=4#8ed213c9a555457ba8849a3f663ed3cd).

Lemonexe: Yeah, I could have realized I don't need a function for that, just a map will be enough.
I have refactored it :slightly_smiling_face:

komret (approved):

PR #13925, opened 2024-08-21
fix(connect): Device cancelable prompts (pin, word, passphrase)

### LINKED ISSUE #14507: cancel from host - meta issue
## Cancel from host
_lets finally fix it everywhere_

There are a couple of button requests that may be cancelled from the host: 
- common button request (eg get address), basically whatever "confirm on device"
- passphrase
- pin (t1)

image

All of these should have a cancel button available somewhere in UI.

As @szymonlesisz found, it didn't always work correctly and prepared a fix in #13925

please let's test the following matrix: 

|   	| Webusb  	| bridge 2.0.27 * 	| bridge 2.0.33  	|  bridge 3+  	|
|---	|---	|---	|---	|---	|
| get address  	|   	|   	|   	|   	|
| pin  T1	|   	|   	|   	|   	|
| passphrase  	|   	|   	|   	|   	|

* here you will likely not see cancel buttons in Suite.

Also please check these issues - they may be affected by the fix as well: 
- #14214
- #13777
- #12210

szymonlesisz:
## Description

Fix unreported (?) behavior: calling `Cancel` on device when promptPin, promptPassphrase, promptWord is rejected or `TrezorConnect.cancel` was invoked.

Changes:
- unify all mentioned prompt functions and moved to standalone module.
- each prompt calls Cancel when rejected
- each prompt assigns cancelableRequest to the `Device`
- when prompt is resolved/rejected cancelableRequest is cleared from the `Device`
- when TrezorConnect.cancel is called cancelableRequest is invoked (call Cancel then reject prompt promise)

Lets test and discuss :)

resolve https://github.com/trezor/trezor-suite/issues/14507 (maybe it won't resolve everything, I just want to trigger QA :P)

### DIFF
--- a/packages/connect/e2e/tests/device/cancel.test.ts
+++ b/packages/connect/e2e/tests/device/cancel.test.ts
@@ -1,0 +1,232 @@
+import { StartEmu, SetupEmu } from '@trezor/trezor-user-env-link';
+import { getController, initTrezorConnect } from '../../common.setup';
+import TrezorConnect from '../../../src';
+
+const getAddress = (showOnTrezor: boolean, coin: string = 'regtest') => {
+    return TrezorConnect.getAddress({
+        path: "m/84'/1'/0'/0/0",
+        coin,
+        showOnTrezor,
+    });
+};
+
+const passphraseHandler = (value: string) => () => {
+    TrezorConnect.uiResponse({
+        type: 'ui-receive_passphrase',
+        payload: {
+            passphraseOnDevice: false,
+            value,
+        },
+    });
+    TrezorConnect.removeAllListeners('ui-request_passphrase');
+};
+
+const addressHandler = () => () => {
+    TrezorConnect.uiResponse({
+        type: 'ui-receive_confirmation',
+        payload: true,
+    });
+    TrezorConnect.removeAllListeners('ui-request_confirmation');
+};
+
+const assertGetAddressWorks = async () => {
+    // validate that further communication is possible without any glitch
+    TrezorConnect.on('ui-request_passphrase', passphraseHandler(''));
+    TrezorConnect.on('ui-request_confirmation', addressHandler());
+    const getAddressResponse = await getAddress(false, 'testnet');
+    expect(getAddressResponse).toMatchObject({
+        success: true,
+        payload: { address: 'tb1qnspxpr2xj9s2jt6qlhuvdnxw6q55jvygcf89r2' },
+    });
+};
+
+describe('TrezorConnect.cancel', () => {
+    const controller = getController();
+
+    const setupTest = async ({
+        setupParams,
+        bridgeVersion,
+    }: {
+        setupParams: SetupEmu & StartEmu;
+        bridgeVersion: string;
+    }) => {
+        await controller.stopBridge();
+        await controller.stopEmu();
+        await controller.startEmu({ wipe: true, ...setupParams });
+        await controller.setupEmu(setupParams);
+        await controller.startBridge(bridgeVersion);
+
+        await initTrezorConnect(controller, { debug: true });
+    };
+
+    beforeAll(async () => {
+        await controller.connect();
+    });
+
+    afterAll(async () => {
+        await TrezorConnect.dispose();
+        controller.dispose();
+    });
+
+    [
+        // '2.0.27', // todo: 2.0.27 is removed from trezor-user-env
+        '2.0.33',
+        // '3.0.0' // todo: add support in trezor-user-env
+    ].forEach(bridgeVersion => {
+        describe(`Bridge ${bridgeVersion}`, () => {
+            afterEach(async () => {
+                await TrezorConnect.dispose();
+                await controller.stopEmu();
+                await controller.stopBridge();
+            });
+
+            // the goal is to run this test couple of times to uncover possible race conditions/flakiness
+            it(`GetAddress - ButtonRequest_Address - Cancel `, async () => {
+                await setupTest({
+                    setupParams: {},
+                    bridgeVersion,
+                });
+
+                TrezorConnect.removeAllListeners();
+                const getAddressCall = getAddress(true);
+                await new Promise<void>(resolve => {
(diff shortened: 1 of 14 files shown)


packages/connect/e2e/tests/device/cancel.test.ts
```diff
@@ -0,0 +1,134 @@
...
+        await initTrezorConnect(controller, { debug: false });
+    };
+
+    afterAll(async () => {
+        controller.dispose();
+        await TrezorConnect.dispose();
+    });
+
+    // the goal is to run this test couple of times to uncover possible race conditions/flakiness
+    [0, 1, 10, 100, 300].forEach(delay => {
```
mroz22: it looks like that higher delays are usually ok but the lower ones tend to be flaky

packages/connect/e2e/tests/device/cancel.test.ts
```diff
@@ -0,0 +1,134 @@
...
+            setupParams: {
+                mnemonic: 'mnemonic_all',
+                passphrase_protection: true,
+            },
+        });
+
+        const getAddressCall = getAddress(true);
+        await new Promise(resolve => {
+            TrezorConnect.on('UI_EVENT', event => {
+                if (event.type === 'ui-request_passphrase') {
```
mroz22: + add ui-response with nullable value (effectivelly cancels)

packages/connect/e2e/tests/device/cancel.test.ts
```diff
@@ -0,0 +1,134 @@
...
+            //   payload: {
+            //     error: 'Initialize failed: Unexpected message, code: Failure_UnexpectedMessage',
+            //     code: 'Device_InitializeFailed'
+            //   }
+            await assertGetAddressWorks();
+        });
+    });
+
+    test('Synchronous Cancel', async () => {
+        await setupTest({
```
mroz22: maybe automatic button rensposes

packages/connect/src/device/DeviceCommands.ts
```diff
@@ -633,53 +551,39 @@ export class DeviceCommands {
...
-         */
+    /**
+     * Bridge version =< 2.0.28 has a bug that doesn't permit it to cancel
+     * user interactions in progress, so we have to do it manually.
+     */
+    async cancelWithFallback() {
         const { name, version } = this.transport;
         if (name === 'BridgeTransport' && !versionUtils.isNewer(version, '2.0.28')) {
             try {
+                // not sure if this even works
```
szymonlesisz: this

packages/connect/src/device/DeviceCommands.ts
```diff
@@ -633,53 +551,39 @@ export class DeviceCommands {
...
+        }
+    }
 
-            if (this.callPromise) {
+    async cancel() {
+        if (this.disposed) {
+            return;
+        }
+        this.dispose();
+        // not sure that this is correct
```
szymonlesisz: this

packages/connect/src/device/Device.ts
```diff
@@ -829,15 +851,14 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
-                }
+                await this.cancelableAction?.();
+                await this.commands?.cancel();
 
                 return this.transport.release({
                     session: this.transportSession,
                     path: this.originalDescriptor.path,
                     onClose: true,
-                });
+                }).promise;
```
szymonlesisz: this

packages/connect/e2e/tests/device/cancel.test.ts
szymonlesisz: this is a deeper rabbit hole, briefly there are 2 problems:
- core `methodSynchronize` is async (loading method) and it is resolved in the next tick, before method is assigned there is nothing to cancel 
- even if here wouldnt be race condition mentioned above there is no possibility to stop current "acquire" process from the outside, therefore it continues, device gets acquired and workflow continues

since it's not related to my current changes i'm leavaing this for later with todo's and comments

mroz22 (approved):
Looks good and should be fixing stuff. Also it doesn't seem to be conflicting with @marekrjpolak's views of life (or at least I hope it doesn't).

I created #14507 to let QA know what should be tested. I'd say we may merge it now so that it receives proper testing long enough before the freeze that is scheduled in a week or so.

mroz22:
It seems to fail for some [popup tests](https://github.com/trezor/trezor-suite/actions/runs/11015742702/job/30593169621#step:5:952). I will investigate it tomorrow

marekrjpolak:
@mroz22 some popup test was failing even before in one of my PR. We've briefly discussed it with @martykan and came to conclusion that the test is probably somehow faulty, dunno.

mroz22:
There is indeed one difference in behaviour caught by[ tests](https://github.com/trezor/trezor-suite/blob/develop/packages/connect-popup/e2e/tests/popup-pages.test.ts#L41)

develop

image

image

this branch 

it doesn't kill the method upon disconnection of the device, it lets the permissions screen active 
image

and when I approve it, only after that I would see this error screen

image

image

packages/connect/src/device/Device.ts
```diff
@@ -774,9 +772,9 @@ export class Device extends TypedEmitter<DeviceEvents> {
         // TODO: cleanup everything
         _log.debug('Disconnect cleanup');
 
-        this.transportSession = null; // set to null to prevent transport.release
-        this.interruptionFromUser(ERRORS.TypedError('Device_Disconnected'));
-        delete this.runPromise;
+        this.transportSession = null; // set to null to prevent transport.release and cancelableAction
+
+        return this.interruptionFromUser(ERRORS.TypedError('Device_Disconnected'));
```
szymonlesisz, line 777: this was the reason of [failing tests](https://github.com/trezor/trezor-suite/pull/13925#issuecomment-2373109255)

`interruptionFromUser` was not awaited and in the same time `runPromise` was deleted therefore if couldn't be rejected by `interruptionFromUser`

mroz22 (approved):

PR #13929, opened 2024-08-22
fix(suite): tokens tab rerendering

AdamSchinzel:
## Description

Tabs inside tokens section were randomly rerendering almost like every second. This should fix it.

### DIFF
--- a/packages/suite/src/views/wallet/tokens/common/TokensLayout/TokensLayoutNavigation.tsx
+++ b/packages/suite/src/views/wallet/tokens/common/TokensLayout/TokensLayoutNavigation.tsx
@@ -71,6 +71,30 @@ interface TokensLayoutNavigationProps {
     setSearchQuery: Dispatch<SetStateAction<string>>;
 }
 
+const Item = ({
+    route,
+    title,
+    icon,
+    currentRoute,
+    count,
+}: {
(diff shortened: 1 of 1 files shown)


packages/suite/src/views/wallet/tokens/common/TokensLayout/TokensLayoutNavigation.tsx
```diff
@@ -71,6 +71,31 @@ interface TokensLayoutNavigationProps {
...
+    count,
+}: {
+    route: Route['name'];
+    title: TranslationKey;
+    icon: IconName;
+    currentRoute?: Route['name'];
+    count: number;
+}) => (
+    <NavListItem
+        data-testid={`@coinmarket/menu/${route}`}
```
jvaclavik: Are you sure about this testid? You haven't used this before

jvaclavik (approved):
One question but approving. Good job

packages/suite/src/views/wallet/tokens/common/TokensLayout/TokensLayoutNavigation.tsx
AdamSchinzel: Sorry leftover, removed

PR #13931, opened 2024-08-22
fix(staking): max with low amount

### LINKED ISSUE #13316: Staking "Max" when amount is lower than 0.13 ETH + current fee
**Description**
Right now `10%`, `20%`, `50%` and `Max` buttons are disabled when balance is from `0.1 ETH` to `0.1299999 ETH`. 
Max becomes enabled from `0.13 ETH` (included)

**Current solution**
- Unable to use `Max` when `<0.13 ETH`
Image

- `Max` button sets less than `0.1 ETH` when balance is `0.13 ETH` (because of fees)
Image

 
**Changes:**
Please enable max button when user has `>=0.1 ETH` (minimum amount)
- in case balance is less than `0.13 ETH + current fee to stake`, set `0.1 ETH` (minimum amount) into input.
- warning `It's recommended to leave 0.03 THOL so you can pay for withdrawal fees.` should explain the rest
Image

dev-pvl:
## Description

Updated the logic to allow the use of the Max button even when the balance is less than 0.13 ETH. Also, the max button is now enabled when the user amount  is >=0.1 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13316

## Screenshots:
image

### DIFF
--- a/packages/suite/src/actions/wallet/stake/stakeFormEthereumActions.ts
+++ b/packages/suite/src/actions/wallet/stake/stakeFormEthereumActions.ts
@@ -31,7 +31,11 @@ import {
     prepareStakeEthTx,
     prepareUnstakeEthTx,
 } from 'src/utils/suite/stake';
-import { MIN_ETH_FOR_WITHDRAWALS } from 'src/constants/suite/ethStaking';
+import {
+    MIN_ETH_AMOUNT_FOR_STAKING,
+    MIN_ETH_BALANCE_FOR_STAKING,
+    MIN_ETH_FOR_WITHDRAWALS,
+} from 'src/constants/suite/ethStaking';
 import { NetworkSymbol } from '@suite-common/wallet-config';
 import { ComposeActionContext } from '@suite-common/wallet-core';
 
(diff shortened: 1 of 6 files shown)


tomasklim (changes requested):
I've clicked `Max`. Amount + fee is not 0.1 so I get error. That's ok. 
However, I am getting the info banner that 0.03 ETH was left as reserve. @dev-pvl 
Screenshot 2024-10-08 at 18 22 28

tomasklim (changes requested):
I have less than `0.13`.
In normal circumstances we leave `0.03` for unstake and claim fees.
However, in this case we can't leave `0.03` but we leave maximum we can.

We need to modify the message in the banner. @jirih-stsh

Screenshot 2024-10-08 at 18 24 19

Any ideas? 
It is an edge-case so maybe just calculate what we left? (but thats maybe too complicated) 
So another option instead of amount to put there `some reserve`?

tomasklim:
btw there are type-check errors

jirih-stsh:
> I have less than `0.13`. In normal circumstances we leave `0.03` for unstake and claim fees. However, in this case we can't leave `0.03` but we leave maximum we can.
> 
> We need to modify the message in the banner. @jirih-stsh
> 
> Screenshot 2024-10-08 at 18 24 19
> 
> Any ideas? It is an edge-case so maybe just calculate what we left? (but thats maybe too complicated) So another option instead of amount to put there `some reserve`?


@tomasklim  since its an edge-case I would just go with the simpler option. 
So write something like: "We've left a small amount of ETH out so you can pay for withdrawal fees."

dev-pvl:
@tomasklim I hid the info banner when the error appeared. Another info banner is added when the amount is less than 0.13. Also, fixed type-check errors. https://github.com/trezor/trezor-suite/pull/13931/commits/ca67abd5c697d1e039fa6ba75fbfa01237b077dd

tomasklim (approved):

PR #13933, opened 2024-08-22
fix(suite): Update bg colors for Switch and tertiary button

jvaclavik:
## Related Issue

Resolve 

## Screenshots:
before
Screenshot 2024-08-22 at 13 29 18

after
image

fixing also bg color of buy button:

image


packages/theme/src/colors.ts
```diff
@@ -75,10 +76,10 @@ const light = {
     backgroundSurfaceElevation2: palette.lightGray100,
     backgroundSurfaceElevation3: palette.lightWhiteAlpha1000,
     backgroundSurfaceElevationNegative: palette.lightGray200,
-    backgroundTertiaryDefaultOnElevation0: palette.lightGray200,
+    backgroundTertiaryDefaultOnElevation0: hexToRgba(palette.lightGray1000, 0.12),
```
seibei-iguchi, line 79: What's the reason behind this? Why is there alpha channel only on some of elevations? Some of the colors in pallete.ts are already using alpha channel in hex format (like lightWhiteAlpha300 or lightSecondaryEmeraldBase10opacity). Is it a good idea to introduce a concept like this in what should IMO be a simple mapping? I'd be very confused seeing this file for the first time:))

PeKne:
Careful. By editing `colors.ts` you are also editing color tokens for `suite-native` which IMO should not happen yet.

packages/theme/src/colors.ts
peter-sanderson: I agree, I am confused when to use `rbg()` and when `#...` hex colors. Both supports alpha. So why mix them?

jvaclavik:
Closing for now. We are waiting for complete new palette from designers.

PR #13936, opened 2024-08-22
feat(transport-bridge): it should work with suite web onion

mroz22:
https://trezor.io/learn/a/tor-in-trezor-suite

### DIFF
--- a/packages/node-utils/src/http.ts
+++ b/packages/node-utils/src/http.ts
@@ -364,8 +364,13 @@ const checkOrigin = ({
 
     if (origin) {
         isOriginAllowed = origins.some(o => {
-            // match from the end to allow subdomains
-            return new URL(origin).hostname.endsWith(new URL(o).hostname);
+            try {
+                // match from the end to allow subdomains
+                return new URL(origin).hostname.endsWith(new URL(o).hostname);
+            } catch (error) {
+                // If parsing URL fails we don't want it to crash but silently logs error.
+                logger.error(`Failed parsing URL: ${error}`);
+            }
         });
     }
     if (!isOriginAllowed) {
--- a/packages/transport-bridge/src/http.ts
+++ b/packages/transport-bridge/src/http.ts
@@ -182,12 +182,16 @@ export class TrezordNode {
                     ) {
                         next(req, res);
                     } else {
-                        allowOrigins(['https://sldev.cz', 'https://trezor.io', 'http://localhost'])(
-                            req,
-                            res,
-                            next,
-                            context,
-                        );
+                        allowOrigins([
+                            'https://sldev.cz',
+                            'https://trezor.io',
+                            'http://localhost',
+                            // When using Tor it will send string "null" as default, and it will not allow calling to localhost.
+                            // To allow it to be sent, you can go to about:config and set the attributes below:
+                            // "network.http.referer.hideOnionSource - false"
+                            // "network.proxy.allow_hijacking_localhost - false"
+                            'http://suite.trezoriovpjcahpzkrewelclulmszwbqpzmzgub37gbcjlvluxtruqad.onion',
+                        ])(req, res, next, context);
                     }
                 },
             ]);


mroz22:
regardless of this PR, does suite-web in tor browser work for you guys? for me no. so I can't check whether it works with the new bridge.

cc @trezor/qa @MiroslavProchazka

bosomt:
Works on my computer

image

**Info:**
 - Suite version: web 24.8.3 (054258687669f8ac2d1bcbd002e96bf5cba9a403)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/115.0
 - OS: MacIntel
 - Screen: 1400x700
 - Device: Trezor T2B1 2.7.2 regular (revision da75d8f4b67410b40a9cfd2954d183d81dd6e8e8)

mroz22:
ok, it is something with my setup and I don't know what. could you @karliatto please test this/finish it? basically suite-web over onion URL should work with the new bridge

packages/transport-bridge/src/http.ts
```diff
@@ -182,12 +182,12 @@ export class TrezordNode {
...
-                            req,
-                            res,
-                            next,
-                            context,
-                        );
+                        allowOrigins([
+                            'https://sldev.cz',
+                            'https://trezor.io',
+                            'http://localhost',
+                            'http://suite.trezoriovpjcahpzkrewelclulmszwbqpzmzgub37gbcjlvluxtruqad.onion/web',
```
karliatto: Tor browser is not sending the real origin since `.onion` addresses are considered to be secret so it sends string "null" instead. https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/32865

It is still following standards like see  in: https://datatracker.ietf.org/doc/html/rfc6454#section-7.3

```
 Whenever a user agent issues an HTTP request from a "privacy-
   sensitive" context, the user agent MUST send the value "null" in the
   Origin header field.
```

karliatto (commented):
I think this was not working with legacy bridge.
But now it is working for me with the changes I added with fixup 39a9e7cf877818f89d09593e45e7c8c55bb679f3, but it is required to add 2 flags to Tor browser as I added comment in the code 


It requires to go to about:config and set the attributes below:
*  "network.http.referer.hideOnionSource - false"
*  "network.proxy.allow_hijacking_localhost - false"
And after that it should work.

packages/transport-bridge/src/http.ts
karliatto: That's is why this flag is required in Tor `"network.http.referer.hideOnionSource - false"`

karliatto: Maybe we could create a Knowledge based documentation on how to use suite from Tor browser. @MiroslavProchazka

MiroslavProchazka: We already have it - see https://trezor.io/learn/a/tor-in-trezor-suite

karliatto: Wow :open_mouth: ! That's great! And I did not realize it is in the description of this PR. :cry:

packages/node-utils/src/http.ts
```diff
@@ -364,8 +364,13 @@ const checkOrigin = ({
 
     if (origin) {
         isOriginAllowed = origins.some(o => {
-            // match from the end to allow subdomains
-            return new URL(origin).hostname.endsWith(new URL(o).hostname);
+            try {
```
karliatto, line 367: I am adding this try catch to avoid the node server to crash when origin is defined but it is not a parseable URL like in the case of default Tor when origin is "null" as a string.

mroz22:
approved 👍 thanks

karliatto (approved):

bosomt:
**QA OK**

Tested with latest dev version in onion browser and with latest dev version app as source of node bridge
https://www.notion.so/satoshilabs/node-bridge-support-onion-suite-web-fffdc526060680388f8cf634e3bbb373?pvs=4

PR #12261, opened 2024-05-03
chore(utxo-lib): uprgade tiny-secp256k1

peter-sanderson:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/7612

PR #13938, opened 2024-08-22
Chore/update mission deps

komret:
Update all Mission-related dependencies except Electron (updated recently by https://github.com/trezor/trezor-suite/pull/13795) and `tiny-secp256k1` (WIP in https://github.com/trezor/trezor-suite/pull/12261). Also, not realy sure what to do with `bcrypto`.


komret:
I don't know why this breaks a mobile XRP test...

vytick:
Looks like upgrade of jssha might have broken something related to XRP.
Screenshot 2024-08-23 at 16 33 38

komret:
No idea how to fix that :( I reverted it back to the previous version. This dependency will be handled by @trezor/suite-connect team.

packages/utxo-lib/src/bip32.ts
```diff
@@ -6,7 +6,7 @@
 // - `fromBase58` and `toBase58` methods are using additional "network" param in bs58check.encode/decode (Decred support).
 
 import ecc from 'tiny-secp256k1';
-import wif from 'wif';
+import * as wif from 'wif';
```
Lemonexe, line 9: :thought_balloon: Do we prefer to import everything from ES exports?
Why not `import { encode } from 'wif'`
or `import { encode as wifEncode } from 'wif'` to make the usage clearer?

Same with `bip66`..

Lemonexe (approved):
LGTM, CI is fine :heavy_check_mark:

PR #13947, opened 2024-08-23
chore(scripts): add list-outdated

komret:
Move package lists from [Notion](https://www.notion.so/satoshilabs/3rd-party-package-team-responsibilities-210e017395ae40349c40006f851244ba) to the repo and add a script to check outdated dependencies based on a domain, e.g. `yarn list-outdated mission` lists all outdated dependencies managed by the Mission team.

TODO separately:
- check for dependencies missing from the lists
- check for duplicates among the lists

I don't know what to do with the failing shellcheck, adding quotes breaks the script. So I would just ignore it.

Screenshot 2024-08-23 at 14 38 28


jvaclavik (approved):
Wow! Good job! ❤️

One though – we will need some script to check new dependencies that are not in the list so far

scripts/list-outdated-dependencies/common-dependencies.txt
```diff
@@ -0,0 +1,110 @@
+
```
jvaclavik: maybe this empty line is not necessary

scripts/list-outdated-dependencies/usability-dependencies.txt
```diff
@@ -0,0 +1,55 @@
+
```
jvaclavik: Same here

scripts/list-outdated-dependencies/common-dependencies.txt
komret: ✅

scripts/list-outdated-dependencies/usability-dependencies.txt
komret: ✅

komret:
> we will need some script to check new dependencies that are not in the list so far

Yes, I mentioned it above, but unfortunately not enough time today to add it.

PR #13950, opened 2024-08-23
feat(suite): Enhance HiddenPlaceholder using special font

### LINKED ISSUE #10567: Improve the "Hide balances" feature

Lemonexe:
Enhance it to actually replace every number with `$`, making them completely illegible.

:construction: **PROOF OF CONCEPT** :construction:

## Description

- JS approach has been tried in [#13914](https://github.com/trezor/trezor-suite/pull/13914) _(complex, messy code, unreliable)_,
- this time try it with only CSS using `font-family`.
- special font was prepared using the free online tool [Glyphr Studio](https://www.glyphrstudio.com/app/)
- the font simply has ever digit replaced with `$`

**Disadvantage:** we still show number of digits..

:information_source:  I chose `$` because it's narrower than `#` or `&`, but still it's not an ideal pick, you can see that the line height is higher, so all content expands down :see_no_evil:
Maybe I messed up something in Glyphr Studio, but idk, this is just PoC! I can later perfect it using `*` maybe?

## Related Issue

Resolve #10567

## Screenshots:

Video shows behavior with hover:

[Video shows behavior with hover](https://github.com/user-attachments/assets/0b41f5db-3549-4403-a498-c343764c062a)


Lemonexe:
Closing, because creating the special font would be too much work if we want to look consistent _(this was just a demo, kinda broken)_.
The font was created by an external partner, and we'd have to contract them again for it.

Will focus on JS solution.

PR #13951, opened 2024-08-23
fix(suite-native): align styles of the Input component with design

### LINKED ISSUE #12672: Input field refactor
As a part of the passphrase project, we've encountered a mismatch in the design of inputfields. 

Design: 
the height of the "input" element is 56 px

Prod: 
the height of the "input" element is 64 px

when discussed with @juriczech, it shouldn't bee too much of work on dev and no work on design. 

As the component is global, the change should then easily propagate all around. What we should check prior to releasing it, is if it changes/breaks the layout structure anywhere in the app. 

As a part of this, let's also add border for both default & focus states of the inputfield. This requires adding 2 design tokens.

yanascz:
Input height and default/focused border aligned with design.

## Related Issue

Resolve #12672

## Screenshots:

image
image
image

Known iOS issue with artificial bottom padding that we couldn't fix:
image


Nodonisko (approved):
looks nice

packages/theme/src/colors.ts
```diff
@@ -92,6 +92,8 @@ const light = {
     borderElevationNegative: palette.lightGray400,
     borderSecondary: palette.lightSecondaryEmerald800,
     borderSubtleInverted: palette.lightWhiteAlpha400,
+    borderInputDefault: palette.lightGray400,
```
PeKne, line 95: Where does these two new colors come from? I can not find them anywhere in the design files.  Did @shenkys want to add these colors?

Otherwise looks good 👌.

yanascz: Yep, @shenkys asked for this change. I guess he forgot to update the ticket as well.

suite-native/atoms/src/Input/Input.tsx
```diff
@@ -67,25 +71,32 @@ type InputLabelStyleProps = {
...
-    (utils, { hasError, hasWarning, isFocused, elevation }) => ({
-        backgroundColor: utils.colors.backgroundNeutralSubtleOnElevation0,
-        borderColor: utils.colors.backgroundNeutralSubtleOnElevation1,
+    (utils, { hasError, hasWarning, isDisabled, isFocused, elevation }) => ({
+        backgroundColor: isDisabled
+            ? utils.colors.backgroundNeutralSubtleOnElevation1
+            : utils.colors.backgroundNeutralSubtleOnElevation0,
+        borderColor: isDisabled
+            ? utils.colors.backgroundNeutralSubtleOnElevation1
+            : utils.colors.borderInputDefault,
```
jvaclavik, line 88: It doesn't respect elevation semantics btw

jvaclavik (approved):

suite-native/atoms/src/Input/Input.tsx
yanascz: This was discussed with @shenkys, the border colour needs to match the background colour when disabled. There will be some UI revision with @PeKne AFAIK. If you believe it should be done differently, let's discuss in person. 🙏

PR #13953, opened 2024-08-23
refactor(suite): Delete secondary buttons

jvaclavik:
## Related Issue

Resolve

### DIFF
--- a/packages/suite/src/views/wallet/details/index.tsx
+++ b/packages/suite/src/views/wallet/details/index.tsx
@@ -139,7 +139,7 @@ const Details = () => {
                                 />
                                 <ActionColumn>
                                     <StyledActionButton
-                                        variant="secondary"
+                                        variant="tertiary"
                                         data-testid="@wallets/details/show-xpub-button"
                                         onClick={handleXpubClick}
                                         isDisabled={disabled}
--- a/packages/components/src/components/Warning/Warning.stories.tsx
+++ b/packages/components/src/components/Warning/Warning.stories.tsx
@@ -19,9 +19,6 @@ const meta: Meta = {
             <WarningComponent {...rest} variant="primary">
                 {children}
             </WarningComponent>
-            <WarningComponent {...rest} variant="secondary">
-                {children}
-            </WarningComponent>
             <WarningComponent {...rest} variant="tertiary">
                 {children}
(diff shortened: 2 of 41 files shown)


packages/components/src/components/Warning/WarningButton.tsx
```diff
@@ -6,9 +6,18 @@ export const WarningButton = ({ children, ...rest }: ButtonProps) => {
     const { variant } = useWarningContextContext();
     const value = { variant: DEFAULT_VARIANT };
 
+    const getVariant = () => {
+        switch (variant) {
+            case 'secondary':
+                return 'primary';
```
peter-sanderson: wut?

packages/suite/src/views/wallet/details/index.tsx
```diff
@@ -107,6 +107,7 @@ const Details = () => {
                                 description={<Translation id={accountTypeDesc} />}
                                 buttonLink={accountTypeUrl}
                             />
+<<<<<<< HEAD
```
peter-sanderson: !!!!

seibei-iguchi:
Why not update the uses of tertiary to secondary and delete tertiary?

packages/components/src/components/Warning/WarningIconButton.tsx
```diff
@@ -6,12 +6,21 @@ export const WarningIconButton = ({ icon, ...rest }: IconButtonProps) => {
     const { variant } = useWarningContextContext();
     const value = { variant: DEFAULT_VARIANT };
 
+    const getVariant = () => {
+        switch (variant) {
+            case 'secondary':
+                return 'primary';
```
peter-sanderson: wat?

jvaclavik:
> Why not update the uses of tertiary to secondary and delete tertiary?

It's going to happen, after this step, but I wanted to make sure there are no secondary ones.

packages/components/src/components/Warning/WarningIconButton.tsx
jvaclavik: We don't have anything to pair with. Do you have a better idea?

peter-sanderson (approved):
image

seibei-iguchi (approved):

PR #13955, opened 2024-08-23
fix(suite-native): if cardano is on, discovery ends too early

### LINKED ISSUE #13954: Mobile - bug in discovery
**Describe the bug**

If Cardano is enabled (default state for everyone) for device with unsupported derivation, this one is not included into the count for started discovery, but `createDescriptorPreloadedDiscoveryThunk` is being run for it. This can cause count being off by 1 in `finishNetworkTypeDiscoveryThunk`, meaning that the discovery will end before the last network is being discovered.

**How to reproduce**

1. enable Coin Enabling in dev settings and connect model T
2. enable Cardano
3. go to dashboard and wait for the discovery to finish
4. go to settings and enable Dash 
5. Dash is never discovered

https://github.com/user-attachments/assets/0d55c810-e6e3-4885-938e-cf22f9cca92e

**Expected behavior**

All networks are discovered

vytick:
Discovery can end before all the networks are discovered if Cardano is discovered and adding one more network.

## Description

If Cardano is enabled (default state for everyone) for device with unsupported derivation, this one is not included into the count for started discovery, but `createDescriptorPreloadedDiscoveryThunk` is being run for it. This can cause count being off by 1 in `finishNetworkTypeDiscoveryThunk` meaning the discovery will end before the last network is being discovered.

Before, the check for this was in the `createDescriptorPreloadedDiscoveryThunk` and this PR moves the check to `startDescriptorPreloadedDiscoveryThunk` and uses correct list of networks to discover for both creating the discovery in the redux and running the discovery for each network

## Related Issue

Resolve #13954

https://github.com/user-attachments/assets/6c78b804-a1ea-4d07-8be4-524be083e855

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -511,13 +511,15 @@ const discoverNetworkBatchThunk = createThunk(
 
 export const createDescriptorPreloadedDiscoveryThunk = createThunk(
     `${DISCOVERY_MODULE_PREFIX}/createDescriptorPreloadedDiscoveryThunk`,
-    async (
+    (
         {
             deviceState,
             networks,
+            availableCardanoDerivations,
         }: {
             deviceState: string;
             networks: readonly Network[];
+            availableCardanoDerivations: ('normal' | 'legacy' | 'ledger')[] | undefined;
         },
         { dispatch, getState },
     ) => {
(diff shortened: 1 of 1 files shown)


suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -629,10 +613,28 @@ export const startDescriptorPreloadedDiscoveryThunk = createThunk(
             }),
         );
 
+        // Some cardano derivation are not supported by the device. We need to filter them out.
+        let availableCardanoDerivations: ('normal' | 'legacy' | 'ledger')[] | undefined;
```
Nodonisko: maybe you can set this to `= []` so you don't need to do `?? []` later.

Nodonisko (approved):

suite-native/discovery/src/discoveryThunks.ts
vytick: eddc551dae

Nodonisko (approved):

PR #13956, opened 2024-08-24
feat(mobile): Refactor AccountList + TokenSelectModal

Nodonisko:
## Related Issue

Resolve 

## Screenshots:
Simulator Screenshot - iPhone 14 Pro - 2024-08-24 at 15 05 46
Simulator Screenshot - iPhone 14 Pro - 2024-08-24 at 15 05 42
Simulator Screenshot - iPhone 14 Pro - 2024-08-24 at 15 05 36


suite-native/accounts/src/components/AccountListItem.tsx
```diff
@@ -7,40 +8,28 @@ import {
...
-}));
+    onPress?: TouchableOpacityProps['onPress'];
+    disabled?: boolean;
+};
 
-export const AccountListItem = ({ account, hideTokens = false }: AccountListItemProps) => {
-    const { applyStyle } = useNativeStyles();
+export const AccountListItem = ({
+    account,
+    hideTokens = false,
```
PeKne: nit: put the default value at the end

suite-native/accounts/src/components/AccountListItemBase.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { TouchableOpacity, TouchableOpacityProps } from 'react-native';
+
+import { Box, HStack, Text } from '@suite-native/atoms';
+import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
+
+export type AccountListItemBaseProps = {
+    icon: React.ReactNode;
+    title: React.ReactNode;
+    badges?: React.ReactNode;
```
PeKne: optionals should be at the end

suite-native/atoms/src/Card/Card.tsx
```diff
@@ -10,16 +10,17 @@ import { AlertBox, AlertBoxProps } from '../AlertBox';
 export type CardProps = {
     children: ReactNode;
     style?: NativeStyleObject;
+    noPadding?: boolean;
```
PeKne, line 13: what about
```suggestion
    isZeroPadding?: boolean;
```

suite-native/accounts/src/components/AccountsList.tsx
```diff
@@ -29,15 +33,36 @@ export const AccountsList = ({
                 DeviceRootState,
         ) => selectFilteredDeviceAccountsGroupedByNetworkAccountType(state, filterValue),
     );
+    const bottomSheetAccountAtom = useMemo(() => atom<Account | null>(null), []);
```
PeKne, line 36: Why is the atom defined inside of the component and then passed as a prop? What is the benefit of it? Could not be `useState` used in this case?

suite-native/assets/src/components/Assets.tsx
```diff
@@ -63,8 +63,8 @@ export const Assets = () => {
 
     return (
         <>
-            <Card>
-                <VStack spacing={19}>
+            <Card noPadding>
+                <VStack spacing={0}>
```
PeKne: `VStack` with `spacing={0}` has no meaning. Replace this with a `Box`.

PeKne (approved):
Mostly just nits 👍

suite-native/atoms/src/Card/Card.tsx
Nodonisko: It's feels strange even if it's correct according to our code style guide. IDK

suite-native/accounts/src/components/AccountsList.tsx
Nodonisko: It's actually very handy because it doesn't rerender this component when modal is opened because we use only setter here. Also another benefit is that you can use setter from child component.

Also `useMemo` is used because this component could have multiple instances across the app but we don't want to share this atom with other instances so now it basically acts as little bit smarter `useState`. It's mentioned somewhere in jotai docs that use can use it like this.

suite-native/assets/src/components/Assets.tsx
Nodonisko: Removed completely, `Card` is enough

suite-native/accounts/src/components/AccountListItemBase.tsx
Nodonisko: fixed... we should take a look if these things coudn't be easily automated in ESlint

suite-native/accounts/src/components/AccountListItem.tsx
Nodonisko: fixed

PR #13957, opened 2024-08-24
chore(repo): add Nx config for depcheck

Nodonisko:
## Description

Depcheck should be much faster now because:
1. It's cached
2. It's run only for modified packages (previously for modified packages and their dependants)

## Related Issue

Resolve


mroz22 (approved):
lets try!

PeKne (approved):

PR #13961, opened 2024-08-26
feat(components): table

AdamSchinzel:
## Description

Created a basic table components ([inspiration from chakra-ui](https://v2.chakra-ui.com/docs/components/table/usage)). Just wanted to keep the naming more self-describing than just `THead`, `Tr`, ..., ...

Futuru usage will be for dashboard assets, tokens and collectibles. Maybe I will find there some places where I can improve the general components to better suite the usage.

## Screenshots:
Screenshot 2024-08-26 at 8 49 11 AM

### DIFF
--- a/packages/components/src/components/Table/Table.stories.tsx
+++ b/packages/components/src/components/Table/Table.stories.tsx
@@ -1,0 +1,40 @@
+import { Meta, StoryObj } from '@storybook/react';
+
+import { Table as TableComponent, allowedTableFrameProps } from './Table';
+import { getFramePropsStory } from '../../utils/frameProps';
+
+const EXAMPLE_TOKENS = [
+    { name: 'USDT', balance: '10', price: '$1,00' },
+    { name: 'USDC', balance: '100', price: '$1,00' },
+    { name: 'MANA', balance: '20', price: '$2,15' },
+];
+
+const meta: Meta = {
+    title: 'Table',
+} as Meta;
+export default meta;
+
+export const Table: StoryObj = {
+    render: props => (
+        <TableComponent {...props}>
+            <TableComponent.HeaderRow>
+                <TableComponent.HeaderCell>Token</TableComponent.HeaderCell>
+                <TableComponent.HeaderCell $alignRight>Balance</TableComponent.HeaderCell>
+                <TableComponent.HeaderCell>Price</TableComponent.HeaderCell>
+            </TableComponent.HeaderRow>
+            {EXAMPLE_TOKENS.map((token, i) => (
+                <TableComponent.Row key={i}>
+                    <TableComponent.Cell>{token.name}</TableComponent.Cell>
+                    <TableComponent.Cell $alignRight>{token.balance}</TableComponent.Cell>
+                    <TableComponent.Cell>{token.price}</TableComponent.Cell>
+                </TableComponent.Row>
+            ))}
+        </TableComponent>
+    ),
+    args: {
+        ...getFramePropsStory(allowedTableFrameProps).args,
+    },
+    argTypes: {
+        ...getFramePropsStory(allowedTableFrameProps).argTypes,
+    },
+};
--- a/packages/components/src/components/Table/Table.tsx
+++ b/packages/components/src/components/Table/Table.tsx
@@ -1,0 +1,46 @@
+import { ReactNode } from 'react';
+import styled from 'styled-components';
+
+import { borders, Elevation, mapElevationToBackground } from '@trezor/theme';
+
+import { TableHeader } from './TableHeader';
+import { TableCell, TableCellProps } from './TableCell';
+import { TableRow } from './TableRow';
+import { useElevation } from '../ElevationContext/ElevationContext';
+import { FrameProps, FramePropsKeys, withFrameProps } from '../../utils/frameProps';
+import { TransientProps } from '../../utils/transientProps';
+
+export const allowedTableFrameProps: FramePropsKeys[] = ['margin'];
+type AllowedFrameProps = Pick<FrameProps, (typeof allowedTableFrameProps)[number]>;
+
+const Container = styled.table<{ $elevation: Elevation } & TransientProps<AllowedFrameProps>>`
+    display: flex;
+    width: 100%;
+    flex-direction: column;
+    background: ${mapElevationToBackground};
+    border-radius: ${borders.radii.md};
+    box-shadow: ${({ theme, $elevation }) => $elevation === 1 && theme.boxShadowBase};
+
+    ${withFrameProps}
+`;
+
+interface TableProps {
+    children: ReactNode;
+}
+
+export const Table = ({ children }: TableProps) => {
+    const { elevation } = useElevation();
+
+    return <Container $elevation={elevation}>{children}</Container>;
+};
+
+Table.Row = TableRow;
+Table.Cell = TableCell;
+Table.HeaderRow = TableHeader;
+Table.HeaderCell = (props: TableCellProps) => (
+    <TableCell $isHeader {...props}>
+        {props.children}
+    </TableCell>
+);
+
+export default Table;
--- a/packages/components/src/components/Table/TableCell.tsx
+++ b/packages/components/src/components/Table/TableCell.tsx
@@ -1,0 +1,26 @@
+import { spacingsPx, typography } from '@trezor/theme';
+import { ReactNode } from 'react';
+import styled, { css } from 'styled-components';
+
+export interface TableCellProps {
+    children: ReactNode;
+    $alignRight?: boolean;
+    $isHeader?: boolean;
+}
+
+export const TableCell = styled(({ $isHeader, $alignRight, ...props }: TableCellProps) =>
+    $isHeader ? <th {...props} /> : <td {...props} />,
+)<TableCellProps>`
+    ${({ $isHeader }) => ($isHeader ? typography.hint : typography.body)}
+    color: ${({ theme, $isHeader }) => ($isHeader ? theme.textSubdued : theme.textDefault)};
+    display: flex;
+    align-items: center;
+    justify-self: left;
+
+    ${({ $alignRight }) =>
+        $alignRight &&
+        css`
+            justify-self: right;
+            padding-right: ${spacingsPx.xxxl};
+        `}
+`;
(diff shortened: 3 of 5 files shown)


packages/components/src/components/Table/TableHeader.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+import { spacingsPx } from '@trezor/theme';
+
+export const TableHeader = styled.div`
+    display: grid;
+    grid-auto-flow: column;
+    grid-auto-columns: 1fr;
+    align-items: center;
+    padding: 0 ${spacingsPx.xl};
+    border-bottom: 1px solid ${({ theme }) => theme.borderElevation2};
```
jvaclavik: please use 

```tsx
border-bottom: 1px solid ${mapElevationToBorder};
```

Check how it's implemented in other places

packages/components/src/components/Table/TableHeaderCell.tsx
```diff
@@ -0,0 +1,16 @@
+import { spacingsPx, typography } from '@trezor/theme';
+import styled, { css } from 'styled-components';
+
+export const TableHeaderCell = styled.div<{ $alignRight?: boolean }>`
```
seibei-iguchi: This and TableRowCell are very similar. I'd make them one component and change the few details with a prop (a variant perhaps?)

packages/components/src/components/Table/TableContainer.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const Container = styled.div`
+    flex-direction: column;
+`;
+
+interface TableContainerProps {
+    children: ReactNode;
+}
+
+export const TableContainer = ({ children }: TableContainerProps) => (
```
seibei-iguchi: Why not just Table?

packages/components/src/components/Table/TableContainer.tsx
```diff
@@ -0,0 +1,18 @@
+import { ReactNode } from 'react';
+import styled from 'styled-components';
+
+import { Card } from '../Card/Card';
+
+export const Container = styled.div`
```
seibei-iguchi: I know it's a bit of a pain, but I'd really love if we started making the Suite more accessible and use proper semantic elements. Could we use table, thead, tr, th, td and so on, please? Should be fine, if you set the display (flex, block and so on).

packages/components/src/components/Table/TableContainer.tsx
```diff
@@ -0,0 +1,18 @@
...
+    flex-direction: column;
```
jvaclavik: Is it necessary? You don't have flex here.

packages/components/src/components/Table/TableContainer.tsx
seibei-iguchi: I'm thinking the other components could be exported as part of the Table component, i.e. Table.Row = TableRow and so on. Then you only have to import the main Table component. What do you think?

seibei-iguchi (changes requested):

packages/components/src/components/Table/TableContainer.tsx
```diff
@@ -0,0 +1,18 @@
...
+`;
+
+interface TableContainerProps {
+    children: ReactNode;
+}
+
+export const TableContainer = ({ children }: TableContainerProps) => (
+    <Card paddingType="none">
```
seibei-iguchi: Perhaps this doesn't have to be part of the component? It'd make the Table component more flexible.

packages/components/src/components/Table/TableContainer.tsx
jvaclavik: I agree and I suggest following naming:

```
Table 
Table.Row 
Table.Cell 
Table.HeaderRow 
Table.HeaderCell

File: packages/components/src/components/Table/TableRow.tsx [hunk 7]
```diff
@@ -0,0 +1,16 @@
+import { spacingsPx } from '@trezor/theme';
+import styled from 'styled-components';
+
+export const TableRow = styled.div`
+    display: grid;
+    grid-auto-flow: column;
+    grid-auto-columns: 1fr;
+    align-items: center;
+    margin: 0 ${spacingsPx.xl};
+    padding: ${spacingsPx.md} 0;
```
seibei-iguchi: Why do both the row and the cell have vertical paddings?

packages/components/src/components/Table/TableContainer.tsx
```diff
@@ -0,0 +1,18 @@
...
+        <Container>{children}</Container>
+    </Card>
```
jvaclavik: I believe that this component shouldn't be part of DS component but it should occur in its usage.

packages/components/src/components/Table/TableHeaderCell.tsx
AdamSchinzel: I was think about the same thing that I can pass props to it but I like it more when it's different in name so it's more explicit.

packages/components/src/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+    { name: 'USDC', balance: '100', price: '$1.00' },
+    { name: 'MANA', balance: '20', price: '$2.15' },
+];
+
+const meta: Meta = {
+    title: 'Table',
+} as Meta;
+export default meta;
+
+export const Table: StoryObj = {
```
jvaclavik, line 17: Could you please add also `margin` from `frameProps`? 🙏 


Plese copy implementation from here

```
packages/components/src/components/Divider/Divider.tsx
```

and storybook's args/argTypes from here:
```
packages/components/src/components/Divider/Divider.stories.tsx

File: packages/components/src/components/Table/TableContainer.tsx (hunk 3 above)
Comment (AdamSchinzel) [reply packages/components/src/components/Table/TableContainer.tsx]: To me the table is `TableRow`, `TableHeader` and other components together, this is just a container. But don't have strong opinion on this.

File: packages/components/src/components/Table/TableContainer.tsx (hunk 4 above)
Comment (AdamSchinzel) [reply packages/components/src/components/Table/TableContainer.tsx]: Good point

File: packages/components/src/components/Table/TableHeaderCell.tsx (hunk 2 above)
Comment (seibei-iguchi) [reply packages/components/src/components/Table/TableHeaderCell.tsx]: I understand, but they are almost the same:)

File: packages/components/src/components/Table/TableContainer.tsx (hunk 3 above)
Comment (jvaclavik) [reply packages/components/src/components/Table/TableContainer.tsx]: Then please rename it to Table 🙏  we might change it in the future if it will not fit.

File: packages/components/src/components/Table/TableHeaderCell.tsx (hunk 2 above)
Comment (seibei-iguchi) [reply packages/components/src/components/Table/TableHeaderCell.tsx]: If you want to, you can make a TableCell with a prop and then the two other components which would use the TableCell component. Just to make it more DRY.

File: packages/components/src/components/Table/Table.tsx [hunk 10]
```diff
@@ -0,0 +1,46 @@
...
+import { TableCell, TableCellProps } from './TableCell';
+import { TableRow } from './TableRow';
+import { useElevation } from '../ElevationContext/ElevationContext';
+import { FrameProps, FramePropsKeys } from '../../utils/frameProps';
+import { TransientProps } from '../../utils/transientProps';
+
+export const allowedTableFrameProps: FramePropsKeys[] = ['margin'];
+type AllowedFrameProps = Pick<FrameProps, (typeof allowedTableFrameProps)[number]>;
+
+export const Container = styled.table<
```
seibei-iguchi: Why export?

packages/components/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,46 @@
...
+    { $elevation: Elevation } & TransientProps<AllowedFrameProps>
```
jvaclavik: You also need to add 
```
${withFrameProps}
```

and pass it here

AdamSchinzel: cafdd5d13d635df5e31f83b5fec3589494e5e8f8

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface TableCellProps {
+    children: ReactNode;
+    $alignRight?: boolean;
+    $isHeader?: boolean;
+}
+
+export const TableCell = styled(({ $isHeader, $alignRight, ...props }: TableCellProps) =>
+    $isHeader ? <th {...props} /> : <td {...props} />,
+)<TableCellProps>`
+    ${({ $isHeader }) => ($isHeader ? typography.hint : typography.body)}
```
seibei-iguchi, line 14: Aren't you missing display: flex here?

packages/components/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,46 @@
...
+>`
+    display: flex;
+    width: 100%;
+    flex-direction: column;
+    background: ${mapElevationToBackground};
+    border-radius: ${borders.radii.md};
+    box-shadow: ${({ theme, $elevation }) => $elevation === 1 && theme.boxShadowBase};
```
jvaclavik, line 22: btw I think this might lead to `box-shadow: false` if it's not elevation 1. Also it should be `boxShadowElevated` for elevation 1 and higher and `boxShadowBase` for elevation 0

packages/components/src/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+    render: props => (
+        <TableComponent {...props}>
```
jvaclavik, line 19: nit: you can use

```tsx
const { Cel,HeaderCell,HeaderRow } = TableComponent

packages/components/src/components/Table/Table.tsx
AdamSchinzel: Hmm this is copy pasted from `Card` component. Should it be different then?

packages/components/src/components/Table/Table.stories.tsx
AdamSchinzel: If you don't mind I like to use the . Cell etc. I like it more 😁

packages/components/src/components/Table/Table.tsx
AdamSchinzel: Leftover, deleted it

seibei-iguchi (approved):

PR #13976, opened 2024-08-26
feat(suite-native): enable coin via add new coin account flow

### LINKED ISSUE #13260: Coin Enabling - enable via Add Coin
- [x] Show all supported networks
- [x] If selected network is not enabled, enable it, run discovery for the coin
- [x] if discovery finds some account, show them 
- [x] user can go back while discovery is running and this closes Add coin flow
- [x] allow to add new coin even if discovery has found account(s)
- [x] if no account is found in discovery, add new (normal) account and go into its detail like in current version

Screenshot 2024-07-09 at 22 40 14
Screenshot 2024-07-09 at 22 40 18
Screenshot 2024-07-09 at 22 40 24

[Figma](https://www.figma.com/design/oQtgXoPIgGSlh01TLHnmeE/Coin-enabling?node-id=391-27727&t=bZoil3v12NSuv9No-0)

vytick:
User can enable new coin via add new coin flow and also discover all relevant accounts

## Description

- [x] Show all supported networks
- [x] If selected network is not enabled, enable it, run discovery for the coin
- [x] if discovery finds some account, show them 
- [x] user can go back while discovery is running and this closes Add coin flow
- [x] allow to add new coin even if discovery has found account(s)
- [x] if no account is found in discovery, add new (normal) account and go into its detail like in current version

## Related Issue

Resolve #13260

## Screenshots:

https://github.com/user-attachments/assets/790a0b7c-c2d6-4bbc-a0c9-2e310cb5c860


suite-native/intl/src/en.ts
```diff
@@ -175,6 +175,18 @@ export const en = {
                 confirm: 'Continue with <type></type>',
             },
         },
+        coinDiscoveryRunningScreen: {
+            title: 'Checking {coin} for balances & transactions.',
+            subtitle: 'This should take just a moment.',
+        },
+        coinDiscoveryFinishedScreen: {
+            title: {
+                singular: 'We’ve found <count>1</count> <coin>Bitcoin</coin> account',
```
PeKne: nit: 
Does these "default values have any purpose here? What about omitting them? I have no strong opinion about this, rather thinking out loud.

```suggestion
                singular: 'We’ve found <count></count> <coin></coin> account',
```

suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
```diff
@@ -223,6 +230,28 @@ export const useAddCoinAccount = () => {
...
+        flowType: AddCoinFlowType,
+    ) => {
+        clearNetworkWithTypeToBeAdded();
+        navigation.replace(AddCoinAccountStackRoutes.AddCoinDiscoveryRunning, {
+            networkSymbol,
+            flowType,
+        });
+    };
+
+    const navigateToAddCoinDiscoveryFinishedScreen = (
```
PeKne: IMO this hook is getting quite bulky. What about not moving this navigation helper from here to `AddCoinDiscoveryRunningScreen` definition? Personally I would preffer to move also other navigation helpers out of here. This hooks has already more than 400 lines and has responsibility for too many things (I has 11 exports!). But this is also maybe just a personal prefference. I did not do the reviews of PRs of this flow, so maybe there is a good reason for having them here?

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+    networkSymbol: NetworkSymbol;
+};
+
+export const AddCoinDiscoveryRunningScreen = ({ route }) => {
+    const { networkSymbol, flowType } = route.params;
+    const dispatch = useDispatch();
+    const accounts = useSelector((state: AccountsRootState & DeviceRootState) =>
+        selectDeviceAccountsByNetworkSymbol(state, networkSymbol),
+    );
+    const normalAccounts = accounts.filter(a => a.accountType === 'normal');
```
PeKne: these two values are needed only in context of `handleFinish` function and does not need to be calculated on every occasion. Move them inside of the function please.

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const nonEmptyAccounts = accounts.filter(a => !a.empty);
+
+    const discovery = useSelector(selectDeviceDiscovery);
+    const { navigateToAddCoinDiscoveryFinishedScreen, navigateToSuccessorScreen } =
+        useAddCoinAccount();
+    const { enabledNetworkSymbols, applyDiscoveryChanges } = useCoinEnabling();
+    const [loadingResult, setLoadingResult] = useState<SpinnerLoadingState>('idle');
+
+    const goToAccountDetail = ({ account }: { account: Account }) => {
+        navigateToSuccessorScreen({
```
PeKne, line 39: what is the benefit of calling this navigation helper here rather calling naigation.replace("AccountDetail", ...)? Does not the helper function just add an extra step. IMO it decreases readability, because you need to go to the `navigateToSuccessorScreen` and check its definition. IF I'm correct, it should always resolve to navigation to an account detail, right?

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+            networkSymbol &&
+            !enabledNetworkSymbols.includes(networkSymbol) &&
+            accounts.length === 0 &&
+            !discovery
+        ) {
+            dispatch(toggleEnabledDiscoveryNetworkSymbol(networkSymbol));
+            applyDiscoveryChanges();
+        }
+
+        if (accounts.length > 0 && !discovery && loadingResult !== 'success') {
```
PeKne:
```suggestion
        if (accounts.length > 0 && !discovery) {
```

IMO the last part of the check is not needed. WDYT?

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const accounts = useSelector((state: AccountsRootState & DeviceRootState) =>
+        selectDeviceAccountsByNetworkSymbol(state, networkSymbol),
+    );
+    const normalAccounts = accounts.filter(a => a.accountType === 'normal');
+    const nonEmptyAccounts = accounts.filter(a => !a.empty);
+
+    const discovery = useSelector(selectDeviceDiscovery);
+    const { navigateToAddCoinDiscoveryFinishedScreen, navigateToSuccessorScreen } =
+        useAddCoinAccount();
+    const { enabledNetworkSymbols, applyDiscoveryChanges } = useCoinEnabling();
```
PeKne:
```suggestion
    const { applyDiscoveryChanges } = useCoinEnabling();
    const enabledNetworkSymbols = useSelector(selectEnabledDiscoveryNetworkSymbols);

```
WHy is it needed to access the symbols from a custom hook if it is possible to access it directly from the state? I think this adds an extra complexity again.

suite-native/intl/src/en.ts
```diff
@@ -175,6 +175,18 @@ export const en = {
...
+                plural: 'We’ve found <count>2</count> <coin>Bitcoin</coin> accounts',
```
PeKne: Format.js has rules for linting plural strings. What about trying to use that? Disclaimer: I never used that yet.

https://formatjs.io/docs/core-concepts/icu-syntax#plural-format

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryFinishedScreen.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    <Translation
+                        id={titleKey}
+                        values={{
+                            count: _ => accounts.length.toString(),
+                            coin: _ => networks[networkSymbol].name,
+                        }}
+                    />
+                </Text>
+            </Box>
+            <Card style={{ paddingHorizontal: 0, paddingTop: 0 }}>
```
PeKne: no inline styles please.

PeKne (commented):
I have some concers but in general It looks good.

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
vytick: It should either go to receive or account detail (different UI), depending on from where the flow started: `flowType`

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
vytick: For me it actually helps with readability and ensures that correct selectors are used everytime we want to work with coin enabling. 

Lets discuss + and -  of this on Monday 🙏

suite-native/intl/src/en.ts
vytick: I have tried to use it, but didnt succeed.

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryFinishedScreen.tsx
vytick: f471517693

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
vytick: f471517693

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
vytick: f471517693

suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
vytick: I did some cleaning there
f471517693

suite-native/intl/src/en.ts
vytick: f471517693

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
PeKne: the `flowType` is accessible is also accessible in this screen no?

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
PeKne: IMO the selecetor should be named in the way you always know which one you want to use.  Then this is just extra abstraction. But I'm not gonna push it. Up to you.

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
vytick: Hmm. Good point. I'll thing about it once more. Thanks

vytick: there is a last  use of navigateToSuccessorScreen and that one is in the hook. I can replace it with actual content of that function, but to me it felt it makes the code less readable, because it adds 20 lines of code. The other navigation helpers were already removed

suite-native/module-add-accounts/src/screens/AddCoinDiscoveryRunningScreen.tsx
vytick: I'll refactor this as we spoke

vytick: 6467982f20

PeKne (approved):
thanks for considering the comments 👌

PR #13979, opened 2024-08-26
feat(suite): refresh on connection return

### LINKED ISSUE #13913: Change the "All backends are down" error message to "Connection lost"
Please replace the "All backends are down" error message by "Connection lost". (instead of "Bitcoin: All backends are down" it should read "Bitcoin: Connection lost" etc. Please make sure it's translated to all Suite languages (it's not in Crowdin).

In most cases there are other reasons for not being able to reach backends than that they would be "down". User internet connection issues, rate limits etc.

Also please, if possible, hit automatically [Retry] once the connection has been re-established. As seen below, we can detect it, we remove the red banner when reconnected. Together with that, we should ideally Retry discovery instead of waiting for the user to do it manually.
Image
Image

enjojoy:
## Description

Changed message to the translated string "Connection lost". Set up suite redux middleware to refresh on offline -> online.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13913

## Screenshots:

image

### DIFF
--- a/packages/suite/src/middlewares/suite/suiteMiddleware.ts
+++ b/packages/suite/src/middlewares/suite/suiteMiddleware.ts
@@ -10,6 +10,7 @@ import {
     handleDeviceConnect,
     handleDeviceDisconnect,
     observeSelectedDevice,
+    restartDiscoveryThunk,
     selectDeviceThunk,
 } from '@suite-common/wallet-core';
 import { notificationsActions } from '@suite-common/toast-notifications';
@@ -104,6 +105,12 @@ const suite =
                 break;
             case SUITE.REQUEST_AUTH_CONFIRM:
                 api.dispatch(authConfirm());
+                break;
+            case SUITE.ONLINE_STATUS:
+                // Restart discovery to reconnect to backends when user goes offline -> online.
+                if (action.payload === true) {
+                    api.dispatch(restartDiscoveryThunk());
+                }
                 break;
             default:
                 break;
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -1651,6 +1651,10 @@ export default defineMessages({
         defaultMessage: 'Connect',
         id: 'TR_CONNECT',
     },
+    TR_CONNECTION_LOST: {
+        defaultMessage: 'Connection lost',
+        id: 'TR_CONNECTION_LOST',
+    },
     TR_DISCONNECT: {
         defaultMessage: 'Disconnect',
         id: 'TR_DISCONNECT',
--- a/packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
+++ b/packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
@@ -107,10 +107,19 @@ interface ExceptionProps {
     discovery?: Discovery;
 }
 
+const getAccountError = (accountError: string) => {
+    if (accountError === 'All backends are down') {
+        return <Translation id="TR_CONNECTION_LOST" />;
+    }
+
+    return accountError;
+};
+
 const discoveryFailedMessage = (discovery?: Discovery) => {
     if (!discovery) return '';
     if (discovery.error) return <div>{discovery.error}</div>;
-    // group all failed networks into array of errors
+
+    // Group all failed networks into array of errors.
     const networkError: string[] = [];
     const details = discovery.failed.reduce((value, account) => {
         const n = accountUtils.getNetwork(account.symbol)!;
@@ -119,7 +128,7 @@ const discoveryFailedMessage = (discovery?: Discovery) => {
 
         return value.concat(
             <div key={account.symbol}>
-                {n.name}: {account.error}
+                {n.name}: {getAccountError(account.error)}
             </div>,
         );
     }, [] as JSX.Element[]);
--- a/packages/blockchain-link/src/workers/solana/index.ts
+++ b/packages/blockchain-link/src/workers/solana/index.ts
@@ -513,8 +513,9 @@ class SolanaWorker extends BaseWorker<SolanaAPI> {
 
     private lazyTokens = createLazy(() => solanaUtils.getTokenMetadata());
 
-    tryConnect(url: string): Promise<SolanaAPI> {
+    async tryConnect(url: string): Promise<SolanaAPI> {
         const api = new Connection(url, { wsEndpoint: url.replace('https', 'wss') });
+        await api.getLatestBlockhash('finalized');
         this.post({ id: -1, type: RESPONSES.CONNECTED });
 
         return Promise.resolve(api);


packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
```diff
@@ -91,6 +102,7 @@ const Container = ({ title, description, cta, dataTestBase }: ContainerProps) =>
                         isLoading={isLocked()}
                         onClick={a.action}
                         data-testid={`@exception/${dataTestBase}/${a.variant || 'primary'}-button`}
+                        ref={retryButtonRef}
```
tomasklim: I would expect that using a ref in a map can lead to unintended consequences.

I think you need something like
```
ref={a.label === 'TR_RETRY' ? retryButtonRef : null}  // assign ref conditionally
```

but tbh is it even needed? you assign the action you need in this component, you can "click" on it using this action `onClick={a.action}`

This is probably not a good approach either but what you wanted to do can be simply done by this
```
    useEffect(() => {
        if (isOnline) {
            const retryAction = actions.find(a => a.label === 'TR_RETRY');
            if (retryAction) {
                retryAction.action();
            }
        }
    }, [isOnline, actions]);
```

Have you considered using redux middleware? modifying this component like this can again lead to unintended consequences.

tomasklim (changes requested):

packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
```diff
@@ -66,6 +67,16 @@ interface ContainerProps {
...
+    const isOnline = useSelector(state => state.suite.online);
+    const retryButtonRef = useRef<HTMLButtonElement | null>(null);
+
+    useEffect(() => {
+        //TODO: pass exception type and check for specific exception, where we want to do the forced refresh
+        console.log('IS ONLINE: ', isOnline);
+        if (isOnline && retryButtonRef && retryButtonRef.current !== null) {
+            retryButtonRef.current.click();
+        }
+    }, [isOnline]);
```
mroz22: this sounds anti-patternish to me. shouldn't retry be handled somewhere on action/middleware level?

enjojoy:
Upon agreement with @mroz22 :
- The message "All backends are down" is coming from blockchain-link, I won't rewrite it there because it's a public API and it could cause problems for someone using it. 

- [x]  I will rewrite it in suite and use a Translation component. 

- In blockchain-link, I will rewrite only Solana message, because it's not consistent with the rest of the networks. 

image

packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
enjojoy: You're right, thank you.

marekrjpolak:
In https://github.com/trezor/trezor-suite/pull/13979/commits/6cd3e56ab5d27ce823b3c253f3da4b6ee2974475 I've added an arbitary API call into Solana worker's `tryConnect` method so in case of connectivity issues it should throw as any other backend type, and return standard "All backends are down" message when no backend can be connected.

enjojoy:
> In [6cd3e56](https://github.com/trezor/trezor-suite/commit/6cd3e56ab5d27ce823b3c253f3da4b6ee2974475) I've added an arbitary API call into Solana worker's `tryConnect` method so in case of connectivity issues it should throw as any other backend type, and return standard "All backends are down" message when no backend can be connected.

thank you ✨

packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
tomasklim: > Have you considered using redux middleware? modifying this component like this can again lead to unintended consequences

?

packages/suite/src/support/messages.ts
```diff
@@ -1651,6 +1651,10 @@ export default defineMessages({
         defaultMessage: 'Connect',
         id: 'TR_CONNECT',
     },
+    TR_CONNECTION_LOST: {
+        defaultMessage: 'Conenction lost',
```
tomasklim:
```suggestion
        defaultMessage: 'Connection lost',
```

packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
enjojoy: Sorry, I haven't considered using it, I just did a little research, and just to clarify: I'd create a redux middleware function that would watch a connection status and refresh anytime the device goes offline -> online, right? So it wouldn't even come to the moment when the user would be able to refresh the page manually and the button would be obsolete?

enjojoy:
@tomasklim , I figured it out and used redux middleware to refresh the page

packages/suite/src/middlewares/suite/suiteMiddleware.ts
```diff
@@ -114,6 +115,11 @@ const suite =
             api.dispatch(observeSelectedDevice());
         }
 
+        // Restart discovery to reconnect to backends when user goes offline -> online.
```
tomasklim: You miss check going from offline to online 🧐

packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
```diff
@@ -117,7 +117,12 @@ const discoveryFailedMessage = (discovery?: Discovery) => {
 
         return value.concat(
             <div key={account.symbol}>
-                {n.name}: {account.error}
+                {n.name}:{' '}
+                {account.error === 'All backends are down' ? (
```
tomasklim: Create a method here 

“getAccountError” and handle this logic there

packages/suite/src/middlewares/suite/suiteMiddleware.ts
enjojoy: This is how this action works based on this example & testing.
image

enjojoy: Do I have to check it additionally?

AdamSchinzel: I think you have it correct. The online status is either `true` or `false` so it works both ways.

```js
export const updateOnlineStatus = (payload: boolean): SuiteAction => ({
    type: SUITE.ONLINE_STATUS,
    payload,
});
```

enjojoy: Yes. It doesn't refresh on online -> offline though.

tomasklim: Why would you restart it when going from online to offline? I don't get it 

Moreover, move it please to the switch above instead of introducing if statement

packages/suite/src/views/dashboard/components/PortfolioCard/components/Exception.tsx
```diff
@@ -117,7 +130,7 @@ const discoveryFailedMessage = (discovery?: Discovery) => {
 
         return value.concat(
             <div key={account.symbol}>
-                {n.name}: {account.error}
+                {n.name}: {getAccountError({ accountError: account.error })}
```
tomasklim: nit: I think it can just accept param
```suggestion
                {n.name}: {getAccountError(account.error )}
```

```
const getAccountError = (accountError: string) => {
```

tomasklim (changes requested):

packages/suite/src/middlewares/suite/suiteMiddleware.ts
enjojoy: It was not restarting due to how restartDiscoveryThunk worked, but I added the check anyway, so now it will restart only on going offline - online for sure

tomasklim (approved):
LGTM, not tested

enjojoy:
> LGTM, not tested

Thanks, I tested It

PR #13981, opened 2024-08-27
connect-explorer: update codemirror libs

mroz22:
quickly checked. https://dev.suite.sldev.cz/connect/udpate-codemirror/methods/bitcoin/getAddress/ it seems to work after the update 
image


martykan (approved):
LGTM

PR #13983, opened 2024-08-27
fix(suite): useCallback in Icon.web to control rerenders

### LINKED ISSUE #13970: Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside ...
Sentry Issue: [TREZOR-SUITE-10NG](https://satoshilabs.sentry.io/issues/5747219007/?referrer=github_integration)

```
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
  at mh (@trezor/suite-build/../../node_modules/react-dom/cjs/react-dom.production.min.js:266:249)
  at nh.enqueueSetState (@trezor/suite-build/../../node_modules/react-dom/cjs/react-dom.production.min.js:145:225)
  at E.prototype.setState (@trezor/suite-build/../../node_modules/react/cjs/react.production.min.js:12:237)
  at _proto.componentDidUpdate (@trezor/suite-build/../../node_modules/react-svg/dist/react-svg.esm.js:171:12)
  at lk (@trezor/suite-build/../../node_modules/react-dom/cjs/react-dom.production.min.js:261:316)
...
(8 additional frame(s) were not displayed)
```

Lemonexe:
Fix React crashing ( :scream: ) at Coinmarket Sell page caused by excessive rerenders (maximum updates reached).

## Description

- [Sentry error](https://satoshilabs.sentry.io/issues/5747219007/?project=5193825&referrer=github_integration)
  - Web only, not desktop!
  - URL `/accounts/coinmarket/sell#/btc/0/normal`
- It might not be apparent at first glance how the cause is connected to the consequence
  - See [the error in console](https://github.com/user-attachments/files/16757791/react.error.txt), which hints at `ReactSVG` component used in `Icon.web`
- The error description hint is misleading
  - it is not caused an infinite loop caused by incorrectly used hooks
  - It's really just ordinary rerenders, but too much of them..
- Missing `useCallback` reduced _some_ unnecessary rerenders
- :thinking:  This is just the tip of iceberg. Our codebase often lacks memoization, so it will probably pop up again

## Related Issue

Resolve #13970

## Screenshots:

react crash

### DIFF
--- a/suite-common/icons/src/webComponents/Icon.web.tsx
+++ b/suite-common/icons/src/webComponents/Icon.web.tsx
@@ -1,5 +1,5 @@
 import { ReactSVG } from 'react-svg';
-import { MouseEvent } from 'react';
+import { MouseEvent, useCallback } from 'react';
 
 import styled, { css, useTheme } from 'styled-components';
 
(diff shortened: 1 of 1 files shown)


suite-common/icons/src/webComponents/Icon.web.tsx
```diff
@@ -58,34 +58,43 @@ export const Icon = ({
 
     const iconSize = getIconSize(size);
 
-    const handleOnKeyDown = (e: React.KeyboardEvent) => {
-        if (e.key === 'Enter' || e.key === ' ') {
-            onClick?.();
-        }
-    };
+    const handleOnKeyDown = useCallback(
```
peter-sanderson, line 61: lets test/figure out why parent of this component is triggering re-renders, this seems as curing the symptoms, not the cause

Lemonexe:
Closing, this is not the proper way to fix it. The problem is somewhere in `CoinmarketSellFormComponent` component, which rerenders excessively. It is an Invity domain.

PR #13991, opened 2024-08-27
feat(wallet-core): disableAccountsThunk can exclude portfolio accounts

### LINKED ISSUE #13986: Coin Enabling - when network is not enabled, it deletes portfolio tracker accounts
**Describe the bug**
When network is disabled, it wrongly removes accounts of that network also for portfolio tracker. Those accounts should not be touched by Coin enabling setup.

**How to reproduce**

1. Go to portfolio tracker, add ethereum
2. Enable coin enabling
3. go to settings - coin enablling
4. have disabled ethereum
5. go back to dashboard and eth accounts are gone (actually accounts for all not enabled networks)

**Expected behavior**
Portfolio tracker should not be influenced at all by coin enabling

vytick:
Previously disabling coin removed Portfolio tracker accounts, it should stay there now.

## Related Issue

Resolve #13986

## Screenshots:

https://github.com/user-attachments/assets/0c24bc49-dd58-4ed4-9763-272fbb8b733e

### DIFF
--- a/suite-common/wallet-core/src/accounts/accountsThunks.ts
+++ b/suite-common/wallet-core/src/accounts/accountsThunks.ts
@@ -37,7 +37,9 @@ export const disableAccountsThunk = createThunk(
             .filter(n => !enabledNetworks.includes(n.symbol) || n.isHidden)
             .map(n => n.symbol);
         // find accounts for disabled networks
-        const accountsToRemove = accounts.filter(a => disabledNetworks.includes(a.symbol));
+        const accountsToRemove = accounts.filter(
+            a => disabledNetworks.includes(a.symbol) && !a.imported,
+        );
 
         if (accountsToRemove.length) {
             dispatch(accountsActions.removeAccount(accountsToRemove));


suite-common/wallet-core/src/accounts/accountsThunks.ts
```diff
@@ -20,24 +20,38 @@ import { notificationsActions } from '@suite-common/toast-notifications';
...
+            : [];
+
         // find disabled networks
         const disabledNetworks = networksCompatibility
             .filter(n => !enabledNetworks.includes(n.symbol) || n.isHidden)
             .map(n => n.symbol);
         // find accounts for disabled networks
-        const accountsToRemove = accounts.filter(a => disabledNetworks.includes(a.symbol));
+        const accountsToRemove = accounts.filter(
+            a => disabledNetworks.includes(a.symbol) && !portfolioAccountKeys.includes(a.key),
```
matejkriz: It can be simplified by using `imported` flag that is true for portfolio tracker accounts.

matejkriz (changes requested):
I think that portfolio accounts can be always excluded from disableAccountsThunk so the extra parametr is not needed.

suite-common/wallet-core/src/accounts/accountsThunks.ts
```diff
@@ -20,24 +20,38 @@ import { notificationsActions } from '@suite-common/toast-notifications';
...
+    selectPortfolioTrackerAccounts,
+} from './accountsReducer';
 import { ACCOUNTS_MODULE_PREFIX } from './accountsConstants';
 import { selectBlockchainHeightBySymbol } from '../blockchain/blockchainReducer';
 
 export const disableAccountsThunk = createThunk(
     `${ACCOUNTS_MODULE_PREFIX}/disableAccountsThunk`,
-    (_, { dispatch, extra, getState }) => {
+    (
+        { excludePortfolioTrackerAccounts = false }: { excludePortfolioTrackerAccounts?: boolean },
```
Nodonisko: `{ excludePortfolioTrackerAccounts = false }: { excludePortfolioTrackerAccounts?: boolean } | undefined = {},`

little akward but then you shoudn't need to put `{}` to every call

suite-common/wallet-core/src/accounts/accountsThunks.ts
vytick: great idea!

vytick: makes it pretty simple indeed
4d181fd4de

matejkriz (approved):
👍 works as expected even on desktop.

PR #14004, opened 2024-08-28
chore(mobile): upgrade deps to RN 0.75

Nodonisko:
## Description

I will need ugpraded RN because of new CPU profiling tool in 0.75

## Related Issue

Resolve


Nodonisko:
Chcked all mentioned packages by socket.dev and they seems fine 

@SocketSecurity ignore-all

Nodonisko:
@SocketSecurity ignore-all

matejkriz (approved):
Except `Invalid hermes-engine.podspec file: undefined method visionos for #<Pod::Specification name="hermes-engine/Pre-built">.` error during `yarn run native:prebuild:clean` it works fine for me. I hope it's only local issue.

PeKne (approved):
Tested both for Android and iOS without an issue. I was not able to reproduce @matejkriz's error.

PR #14006, opened 2024-08-28
chore(suite): set node-bridge rollout to 100% EAP users

MiroslavProchazka:
Bumping the Node bridge to 100% of EAP users and slightly adjusting the copy, so we can test it more in the upcoming 24.9.1 release.


mroz22:
lets give it a try

mroz22 (approved):

Issue #13821, opened 2024-08-15
Performance issues of Graph on mobile


Nodonisko:
First loading could take a long time (depends on how many transactions user has) - we should consider some kind of progress bar

App become unresponsive during graph loading (Sometime it's impossible to navigate to Settings while graph is still loading)

matejkriz:
6. App become unresponsive during graph loading (I can't navigate to Settings while graph is still loading)

matejkriz:
> 5. ETC doesn't work correctly `0x500910C4c42F1F2eeAF7E36d0c6f41e0F8858d5f` not locally nor from blockbook. It's showing negative values - very low prio right now 

moved to https://github.com/trezor/trezor-suite/issues/13987

matejkriz:
> 1. It can go to negative balance (could be reproduced using all seed) - discovered by @vytick 

moved to https://github.com/trezor/trezor-suite/issues/13988

matejkriz:
related https://github.com/trezor/trezor-suite/issues/13019

matejkriz:
> 2. Pending transactions are added to graph balance incorrectly, because they will be added for whole graph line not for just last point or not at all.

moved to https://github.com/trezor/trezor-suite/issues/14003

matejkriz:
> 3. Staking is not reflected in graph

Reflect staking in mobile app has it's own issue https://github.com/trezor/trezor-suite/issues/12187

matejkriz:
Closing this now as it seems to be mostly resolved. Moving to Approved directly, doesn't make sense to test separately.

PR #14008, opened 2024-08-28
chore(suite-native): memoize expensive selector to save CPU during graph loading

matejkriz:
## Description

Very small step to optimize app lagging during graph loading of account with lot of transactions. Just saving CPU by memoizing non-trivial selector.

## Related Issue

related to https://github.com/trezor/trezor-suite/issues/13821

## Screenshots:
without memoization:
image

with memoization:
image


vytick (approved):
good find! 💪

PR #13790, opened 2024-08-13
test(transport-test): implement `other call in progress` check for reaceive/send/call endpoints

mroz22:
concurrent receive/send/call old  vs new bridge behaves differently

a couple of not strictly related commits (f139b86, 1441e59, 2e920b5, 444f939),  cherry-picked into #13907. low hanging-ness guaranteed.

cc @szymonlesisz

PR #14009, opened 2024-08-28
fix(transport): usb api in node and abort

mroz22:
try out 

```
yarn workspace @trezor/transport-test test:e2e:api:node:hw
```

on develop and this branch


mroz22:
works for me both in web and node

mroz22:
part of #13790

PR #14010, opened 2024-08-28
fix(suite-native): ignore first evm ledger derivation path in discovery

### LINKED ISSUE #13996: Doubled ETH account normal vs ledger derivation path
**Describe the bug**
I see two Ethereum #1 accounts, both are the same, just different derivation path.

Only the #1 is doubled, other accounts appear only once.

**Info:**

-   Mobile Suite Version: 24.8
-   OS: android

**How to reproduce**

1. Load wallet with Ethereum account (it can be allall seed)
1. See error

**Expected behavior**
I should see no doubled accounts

**Screenshots**
image

**Additional context**
related? https://github.com/trezor/trezor-suite/pull/13734

vytick:
- ignore first ledger evm account, because the derivation path is the same as trezor 1st path
- add `isAccountAlreadyDiscovered` check before adding account to state in case something similar happens in the future. First check in `discoverNetworkBatchThunk` is not sufficient because the account does not have to be in the state already

## Related Issue

Resolve #13996

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -290,7 +290,13 @@ const discoverAccountsByDescriptorThunk = createThunk(
                 ...getAccountInfoDetailsLevel(bundleItem.coin),
             });
 
-            if (success) {
+            const isAccountAlreadyDiscovered = selectIsAccountAlreadyDiscovered(getState(), {
+                deviceState,
+                networkSymbol: bundleItem.coin,
+                path: bundleItem.path,
+            });
+
+            if (success && !isAccountAlreadyDiscovered) {
                 if (accountInfo.empty) {
                     isFinalRound = true;
                 }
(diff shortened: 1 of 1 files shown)


PeKne (approved):

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -440,7 +446,12 @@ const discoverNetworkBatchThunk = createThunk(
...
-            const index = lastDiscoveredAccountIndex + batchIndex;
+            const isEvmLedgerDerivationPath =
+                network.networkType === 'ethereum' && network.accountType === 'ledger';
+            // first Ledger derivation path for EVM networks is the same as trezor, so we need to skip it (consistent with desktop)
+            const index =
+                lastDiscoveredAccountIndex + batchIndex + (isEvmLedgerDerivationPath ? 1 : 0);
+
             const accountPath = network.bip43Path.replace('i', index.toString());
 
             const isAccountAlreadyDiscovered = selectIsAccountAlreadyDiscovered(getState(), {
```
matejkriz, line 457: Have you tried to remove this check? Is it really useful here? Or do we leave it on the basis of "don't touch what isn't broken" (which is fine too)?

vytick: it speeds up discovery of this network in case it was interrupted and not finished, but device is in view only. In such case, already discovered accounts are not fetched again.

Issue #13152, opened 2024-07-02
EVM Tokens / Generic table structure


tomasklim:
Create generic asset table which should be reusable on `Dashboard` and in `Tokens` tab in `Account`.
Please use css `grid`. The table should look the same as before, it should use newly created component `Table`

- [x] Create table https://github.com/trezor/trezor-suite/pull/13961
- [x] Use it in `Tokens` tab https://github.com/trezor/trezor-suite/pull/14013
- [x] Use it on `Dashboard`

Issue created based on feedback in [notion](https://www.notion.so/satoshilabs/Tokens-tab-improvements-b1a0a31ef3dd43fcb1079401c84c0bc6?pvs=4#9f54463c0a3c4a64b6f379195155ff97)

Tokens ✅ :
![Screenshot 2024-09-30 at 9 19 54](https://github.com/user-attachments/assets/475c14bf-9ca9-409d-8cbb-54988f2a5a31)

Dashboard 🟡 :
![Screenshot 2024-09-30 at 9 20 07](https://github.com/user-attachments/assets/6a4a1009-ff52-4224-96ac-8bcce38cfe79)

enjojoy:
The table should be there instead of Cards on the mobile view as well?

bosomt:
**QA OK**

**Info:**
 - Suite version: web 24.11.0 (a682ea653d7d39851dc7129991b6a11eae493bdc)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.4 regular (revision ab96ce7954c7db384b26bc863d8752150027adbf)
 - Transport: BridgeTransport 2.0.33

PR #14013, opened 2024-08-29
chore(suite): new token table

### LINKED ISSUE #12822: EVM tokens/ Icons
**Description:**

Add token icon to tokens list to the left of token name. Including all EVMs, Solana and Cardano
- do not show ring, just icon of token
- 3 letters are better than 1 in case of tokens without icon
Screenshot 2024-08-02 at 21 06 07

Tokens available in token definitions should have an icon, no need to fetch icons for the ones without token definitions. They are not going to be available.

**Prerequisite issue:**

#13630

AdamSchinzel:
## Description

Rewrite token table to new table component, add icons and alight price and trend columns.

## Related Issue

Resolve #12822
Related #13152

### Before
Screenshot 2024-09-11 at 16 22 38
Screenshot 2024-09-11 at 16 22 25

### After
Screenshot 2024-09-11 at 16 21 03
Screenshot 2024-09-12 at 16 30 11

https://github.com/user-attachments/assets/21b0c00a-5679-4c58-a88d-cb4ee9ba6038


packages/components/src/components/AssetLogo/AssetLogo.tsx
```diff
@@ -15,8 +20,8 @@ type AllowedFrameProps = Pick<FrameProps, (typeof allowedAssetLogoFrameProps)[nu
 
 export type AssetLogoProps = AllowedFrameProps & {
     size: AssetLogoSize;
-    coingeckoId: string;
-    contractAddress?: string;
+    symbol: NetworkSymbol;
+    contractAddress?: TokenAddress;
```
jvaclavik: Please keep origin API of the component. invity can use also symbols and tokens that we don't know.

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -1,10 +1,12 @@
...
 import { ReactNode } from 'react';
 import styled, { css } from 'styled-components';
 
+import { spacingsPx, typography } from '@trezor/theme';
+
 export interface TableCellProps {
-    children: ReactNode;
+    children?: ReactNode;
     $alignRight?: boolean;
+    $withRightMargin?: boolean;
```
jvaclavik: 1) Btw since it's in the DS it should have predictable API so we shouldn't use $ which means we need to create a wrapper above styled component and use `makeTransientProps`

2) Also thinking that maybe using `margin` from `frameProps` would be more universal than `withRightMargin`

3) please could you update API to `align = 'left' | 'right'`? 🙏

packages/components/src/index.ts
```diff
@@ -30,6 +31,7 @@ export * from './components/RotateDeviceImage/RotateDeviceImage';
 export * from './components/DataAnalytics';
 export * from './components/Divider/Divider';
 export * from './components/Dropdown/Dropdown';
+export * from './components/Dropdown/Menu';
```
jvaclavik: why do you need this?

packages/components/src/index.ts
```diff
@@ -72,6 +74,7 @@ export * from './components/typography/Paragraph/Paragraph';
 export * from './components/typography/Text/Text';
 export * from './components/typography/TruncateWithTooltip/TruncateWithTooltip';
 export * from './components/Warning/Warning';
+export * from './components/Table/Table';
```
jvaclavik: btw we prefer to be specific here and not export everything

jvaclavik (changes requested):

MiroslavProchazka:
Is it also implemented in Dashboard and My Assets, as stated in the original issue?

AdamSchinzel:
> Is it also implemented in Dashboard and My Assets, as stated in the original issue?

What issue do you mean? Icons and alignment are only for tokens table.

Rewriting assets table to table component will be in a another PR in the end so we can merge this today.

packages/components/src/index.ts
AdamSchinzel: Okey but why there is `export * from` almost everywhere 😄?

packages/components/src/index.ts
AdamSchinzel: I'm using GroupedMenuItems and I used to import it like this:

` import { GroupedMenuItems } from '@trezor/components/src/components/Dropdown/Menu';`

Think it's better to import it from `@trezor/components`

packages/components/src/components/Table/TableCell.tsx
AdamSchinzel: Yes make sense 👍

packages/components/src/index.ts
AdamSchinzel: I will be then specific and change it to:

```js
export type { GroupedMenuItems } from './components/Dropdown/Menu';
```

packages/components/src/components/AssetLogo/AssetLogo.tsx
AdamSchinzel: I think it's better to derive coingeckoId in the component and just as for symbol and contractAddress.

For example in the tokens table I would use the getCoingeckoId function anyway so it makes more sense to me to do it one step before.

@adderpositive @FreeWall What do you think guys?

packages/components/src/index.ts
jvaclavik: Historic reasons :) but we will update it in the future.

packages/components/src/components/AssetLogo/AssetLogo.tsx
AdamSchinzel: With Invity team we agreed on what is inside this fixup - 65fe3bdd114421663f7917eb79b51f80828074cb.

We will keep `coingeckoId` so they can show icons for networks we don't support but use symbol for correctly transforming contract address inside `getContractAddressForNetwork`.

packages/components/package.json
```diff
@@ -21,7 +21,10 @@
         "@suite-common/icons": "workspace:*",
         "@suite-common/suite-constants": "workspace:*",
         "@suite-common/suite-utils": "workspace:*",
+        "@suite-common/token-definitions": "workspace:*",
         "@suite-common/validators": "workspace:*",
+        "@suite-common/wallet-config": "workspace:*",
+        "@suite-common/wallet-types": "workspace:*",
```
peter-sanderson: UI components shall NOT depend on any bitcoin/crypto related libs and packages IMO

packages/components/src/components/AssetLogo/AssetLogo.tsx
```diff
@@ -53,7 +59,9 @@ export const AssetLogo = ({
 }: AssetLogoProps) => {
     const [isLoading, setIsLoading] = useState(true);
     const [isPlaceholder, setIsPlaceholder] = useState(false);
-    const fileName = contractAddress ? `${coingeckoId}--${contractAddress}` : coingeckoId;
+    const fileName = contractAddress
+        ? `${coingeckoId}--${getContractAddressForNetwork(symbol, contractAddress)}`
```
peter-sanderson: This is purely UI component. The `getContractAddressForNetwork(symbol, contractAddress)` shall be resolved outside of component package

packages/components/src/components/AssetLogo/AssetLogo.tsx
peter-sanderson, line 1: @jvaclavik I would even argue this AssetLogo component shall not be part of the Componets package

jvaclavik: Well, I'm the author of `AssetLogo` 🫣 it should probably be in product-components, but it's for different PR.

packages/components/src/components/AssetLogo/AssetLogo.tsx
```diff
@@ -39,11 +44,12 @@ const Logo = styled.img<{ $size: number; $isVisible: boolean }>(
...
 );
 
-const getAssetLogoUrl = (fileName: string, quality?: '@2x') =>
+const getAssetLogoUrl = (fileName?: string, quality?: '@2x') =>
     `${ICONS_URL_BASE}${fileName}${quality === undefined ? '' : quality}.webp`;
 
 export const AssetLogo = ({
     size,
+    symbol,
     coingeckoId,
```
peter-sanderson: this is ignored in case   `contractAddress` is present

in cases like this lets type it properly so it is clear from types what is goin on.

For icon (`packages/components/src/components/Icon/Icon.tsx`)

we use this pattern so it can be used from there:
```ts
export type ExclusiveColorOrVariant =
    | { variant?: IconVariant; color?: undefined }
    | {
          variant?: undefined;
          /** @deprecated Use only is case of absolute desperation. Prefer using `variant`. */
          color?: string;
      };
      ```

Comment (AdamSchinzel) [reply packages/components/src/components/AssetLogo/AssetLogo.tsx]: I removed it so the API for the component is the same

File: packages/components/src/components/AssetLogo/AssetLogo.tsx (hunk 6 above)
Comment (AdamSchinzel) [reply packages/components/src/components/AssetLogo/AssetLogo.tsx]: Removed, it will be handled outside of the component. Primarily because of the imports to the component package mentioned above.

File: packages/components/src/components/AssetLogo/AssetLogo.tsx [hunk 8]
```diff
@@ -39,7 +41,7 @@ const Logo = styled.img<{ $size: number; $isVisible: boolean }>(
     `,
 );
 
-const getAssetLogoUrl = (fileName: string, quality?: '@2x') =>
+const getAssetLogoUrl = (fileName?: string, quality?: '@2x') =>
```
tomasklim:
```suggestion
const getAssetLogoUrl = (fileName?: string, quality?: '@2x') =>
```
This is going to build wrong url in case file name is undefined

packages/components/package.json
```diff
@@ -21,6 +21,7 @@
         "@suite-common/icons": "workspace:*",
         "@suite-common/suite-constants": "workspace:*",
         "@suite-common/suite-utils": "workspace:*",
+        "@suite-common/token-definitions": "workspace:*",
```
tomasklim: This is going to include redux, jws and other non-ui libs into components package

packages/components/tsconfig.json
```diff
@@ -20,9 +20,18 @@
...
+        },
         {
             "path": "../../suite-common/validators"
         },
+        {
+            "path": "../../suite-common/wallet-config"
+        },
+        {
+            "path": "../../suite-common/wallet-types"
+        },
```
tomasklim: `yarn refs`

packages/components/src/components/AssetLogo/AssetLogo.stories.tsx
```diff
@@ -17,8 +20,8 @@ export default meta;
 export const AssetLogo: StoryObj<AssetLogoProps> = {
     args: {
         size: 24,
-        coingeckoId: 'binance-smart-chain',
-        contractAddress: '0x0203d275d2a65030889af45ed91d472be3948b92',
+        symbol: 'eth',
+        contractAddress: '0xdac17f958d2ee523a2206206994597c13d831ec7' as TokenAddress,
```
tomasklim: no longer valid

tomasklim (changes requested):

tomasklim (changes requested):
~Not implemented~

> Tokens available in token definitions should have an icon, no need to fetch icons for the ones without token definitions. They are not going to be available.

Screenshot 2024-09-01 at 8 53 23

tomasklim (changes requested):
~In the design, there is just a small gap between icon and name of the token~

Implementation of width of columns is missing

Screenshot 2024-09-01 at 8 59 17

Screenshot 2024-09-01 at 8 58 27

~not sure if related but table on dashboard is broken~

Screenshot 2024-09-01 at 9 12 57

AdamSchinzel:
> In the design, there is just a small gap between icon and name of the token
> 
> Implementation of width of columns is missing
> 
> Screenshot 2024-09-01 at 8 59 17 Screenshot 2024-09-01 at 8 58 27
> not sure if related but table on dashboard is broken
> 
> Screenshot 2024-09-01 at 9 12 57

Yes good point it could be smaller.

The table is broken in develop so not related by this PR.

packages/components/src/components/AssetLogo/AssetLogo.tsx
```diff
@@ -19,6 +21,7 @@ export type AssetLogoProps = AllowedFrameProps & {
     contractAddress?: string;
     shouldTryToFetch?: boolean;
     placeholder: string;
+    isKnownToken?: boolean;
```
jvaclavik: Why do you need it? You don't use it anywhere

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -1,26 +1,55 @@
...
+import { makePropsTransient, TransientProps } from '../../utils/transientProps';
+
+export const allowedTableCellFrameProps: FramePropsKeys[] = ['margin'];
+type AllowedFrameProps = Pick<FrameProps, (typeof allowedTableCellFrameProps)[number]>;
+
+export type TableCellProps = AllowedFrameProps & {
+    children?: ReactNode;
+    align?: 'left' | 'right';
+    isHeader?: boolean;
+    isSmaller?: boolean;
```
tomasklim: Using this solution, the icon column has different size in legit and hidden tokens (different number of columns). It does not look nice.

And instead of `isSmaller`, I would like something more generic. I would like to be able to set `grid-template-columns` to the table or `grid-column` to cells. I don't know what is best solution, but this is not generic.

Please help @jvaclavik @adamhavel

tomasklim (changes requested):
~There are errors in console~
Screenshot 2024-09-08 at 10 41 25

Now:
~the columns collide~
Screenshot 2024-09-03 at 1 51 56

Before:
Screenshot 2024-09-03 at 1 55 34

packages/components/src/index.ts
```diff
@@ -9,6 +9,7 @@ export { Checkbox, type CheckboxProps } from './components/form/Checkbox/Checkbo
 export * from './components/animations/DeviceAnimation';
 export * from './components/animations/LottieAnimation';
 export * from './components/AssetShareIndicator/AssetShareIndicator';
+export * from './components/AssetLogo/AssetLogo';
```
jvaclavik: Please export just what you need 🙏

packages/components/src/components/Table/Table.stories.tsx
```diff
@@ -19,13 +19,13 @@ export const Table: StoryObj = {
         <TableComponent {...props}>
```
jvaclavik, line 19: btw, I tried add margin (from frameProps) and it's not working here

packages/components/src/components/Table/TableCell.tsx
jvaclavik: I quite agree with @tomasklim. Setting `grid-template-columns` or `grid-column` is much more universal than this. 

let's wait for @adamhavel 's opinion, he is the grid guru 🙂

packages/components/src/index.ts
```diff
@@ -72,6 +74,7 @@ export * from './components/typography/Paragraph/Paragraph';
 export * from './components/typography/Text/Text';
 export * from './components/typography/TruncateWithTooltip/TruncateWithTooltip';
 export * from './components/Warning/Warning';
+export { Table } from './components/Table/Table';
```
jvaclavik: I think we should be able to use `Table` also without `Container`. And I think `Container` could be just `Card` component. It's basically the same, right? 

```tsx
<Card paddingType="none">
  <Table ... />
</Card>
```

jvaclavik: And one more thing - I think `TableRow` should have `hasBorder: boolean` prop

packages/suite/src/views/wallet/tokens/common/TokensTable/TokensTable.tsx
```diff
@@ -43,26 +27,25 @@ const ChevronIcon = styled(Icon)<{ $isActive: boolean }>`
...
     cursor: pointer;
     align-items: center;
     ${typography.label}
     color: ${({ theme }) => theme.textSubdued};
-    ${({ $isActive }) =>
+    ${({ $isActive, $elevation, theme }) =>
         $isActive &&
         css`
-            border-bottom: 1px solid ${({ theme }) => theme.borderElevation2};
+            border-bottom: 1px solid ${mapElevationToBorder({ $elevation, theme })};
```
jvaclavik: you can use just 

```tsx
border-bottom: 1px solid ${mapElevationToBorder};

File: packages/suite/src/views/wallet/tokens/common/TokensTable/TokensTable.tsx [hunk 18]
```diff
@@ -82,31 +65,34 @@ export const TokenList = ({
...
-                        <ColName>
+                        </Table.HeaderCell>
+                        <Table.HeaderCell>
                             <Translation id="TR_7D_CHANGE" />
-                        </ColName>
+                        </Table.HeaderCell>
                     </>
                 )}
-            </Columns>
+                <Table.HeaderCell /> {/* Actions */}
```
jvaclavik: Can you delete this comment please? 🙏

jvaclavik (changes requested):

packages/components/src/components/Table/TableCell.tsx
seibei-iguchi: Well I suggest to keep it as simple as possible. I'm thinking props like columns={n} on the Table and span={n} on the TableCell. But if we wanna set grid on the Table, then we'd have to use subgrid on the rows. It's doable but maybe we could just use native table rendering and not override it (i.e. why recreate it with grid when it can be done without css with table). Then we could use attributes like colspan. By the way, I'm creating a simple Grid implementation in the design system right now, but I'd like to keep it separate from the Table (semantics and so on)

packages/suite/src/views/wallet/tokens/common/TokensTable/TokensTable.tsx
AdamSchinzel: Screenshot 2024-09-03 at 9 55 13 AM

I had it like this before but TS was complaining.

tomasklim:
~Zero token balance part is broken~
Screenshot 2024-09-08 at 10 38 03

~There are duplicated keys in `Hidden`~
Screenshot 2024-09-08 at 10 38 43

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -1,26 +1,45 @@
...
-    ${({ $alignRight }) =>
-        $alignRight &&
-        css`
-            justify-self: right;
-            padding-right: ${spacingsPx.xxxl};
-        `}
+    justify-content: ${({ $align }) => mapAlignmentToJustifyContent($align)};
 `;
+
+export const TableCell = ({ children, span = 1, align = 'left' }: TableCellProps) => {
```
jvaclavik: Not sure about `span` prop name. Why not `colSpan`?

jvaclavik (approved):
Like it a lot! 💪

tomasklim (approved):
Looks good! 👍 

Responsiveness looks a bit like an issue but I can try to solve it myself when I am back. Everything looks good with nice numbers, but tokens can have really small fiat values, so users can have millions of them and it can have e.g. 18 decimals too.

https://github.com/user-attachments/assets/e4e54fd0-2a05-4b97-b867-b7de99938e9d

And I do not like jumping when showing zero balance tokens

seibei-iguchi:
> Looks good! 👍 
> 
> 
> 
> Responsiveness looks a bit like an issue but I can try to solve it myself when I am back. Everything looks good with nice numbers, but tokens can have really small values, so users can have millions of them and it can have e.g. 18 decimals too.
> 
> 
> 
> https://github.com/user-attachments/assets/e4e54fd0-2a05-4b97-b867-b7de99938e9d
> 
> 
> 
> And I do not like jumping when showing zero balance tokens 

Thanks! I'll look into responsiveness and edge cases today. Also will try to tackle the jumping.

Hermez-cz:
Hi, it's looking good as far I can see from the provided screenshots, thank you!

Only this might be better aligned. 
image

seibei-iguchi:
> Hi, it's looking good as far I can see from the provided screenshots, thank you!
> 
> Only this might be better aligned. image

I can try do something but these are separate tables with different number of columns so their layout is calculated dynamically. But I might fix this one by setting the width of the first column to a fixed value.

packages/components/src/components/Table/Table.tsx
```diff
@@ -1,56 +1,63 @@
...
+export const Table = ({ children, margin, colWidths }: TableProps) => {
+    const { scrollElementRef, onScroll, ShadowContainer, ShadowLeft, ShadowRight } =
+        useScrollShadow();
 
     return (
-        <Container $elevation={elevation} {...frameProps}>
-            {children}
-        </Container>
+        <ShadowContainer>
+            <ShadowLeft backgroundColor="backgroundSurfaceElevation1" />
```
adamhavel, line 40: I'm not sure how to handle this. Either I set this via prop on Table component or Table will always be in a Card component. What do you think?

tomasklim (commented):
Really great changes. LGTM 🚀 

Three small things to discuss
Screenshot 2024-09-13 at 6 31 16

1. I would wrap symbol next to balance to the next line in case it is too long (done in screenshot). Symbols in Suite can have up to 10 chars
2. We can decrease padding of cells when the window is small
3. We can hide `7 day change` when window is small and put send/receive into `...` (I think there will also be Trade button in the future)

However, if 1,2,3 does not make sense, I am okay with scrolling, looks great -> would be nice if first column would be sticky, so that user knows the context of the right scrolled part.

packages/components/src/components/Table/Table.tsx
adamhavel: It seems to me using Card internally in Table is a lesser evil, what do you think?

seibei-iguchi:
> Really great changes. LGTM 🚀
> 
> Three small things to discuss Screenshot 2024-09-13 at 6 31 16
> 
> 1. I would wrap symbol next to balance to the next line in case it is too long (done in screenshot). Symbols in Suite can have up to 10 chars
> 2. We can decrease padding of cells when the window is small
> 3. We can hide `7 day change` when window is small and put send/receive into `...` (I think there will also be Trade button in the future)
> 
> However, if 1,2,3 does not make sense, I am okay with scrolling, looks great -> would be nice if first column would be sticky, so that user knows the context of the right scrolled part.

Well, I've tried the sticky first column. Let me know what you think:)

packages/components/src/components/Table/TableRow.tsx
```diff
@@ -1,30 +1,36 @@
...
+    border-top: 1px solid ${mapElevationToBorder};
+
+    &:first-child {
+        border-top: 0;
     }
+
+    ${({ $isCollapsed }) =>
+        $isCollapsed &&
+        css`
+            visibility: collapse;
```
tomasklim, line 18: AssetLogo's visibility collides with this and makes icons visible from `Zero-balance tokens` section even though it is not open
Screenshot 2024-09-14 at 7 31 33

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -1,26 +1,60 @@
...
+
+const Cell = styled.td<{ $isHeader: boolean; $elevation: Elevation }>`
     ${({ $isHeader }) => ($isHeader ? typography.hint : typography.body)}
     color: ${({ theme, $isHeader }) => ($isHeader ? theme.textSubdued : theme.textDefault)};
+    text-align: left;
+    padding: ${spacingsPx.sm} ${spacingsPx.lg};
+    max-width: 300px;
+
+    &:first-child {
+        position: sticky;
```
tomasklim, line 35: It does not work for me. I think `overflow: hidden;` has to be removed from `table`tag

packages/suite/src/views/wallet/tokens/common/TokensTable/TokensTable.tsx
```diff
@@ -0,0 +1,130 @@
...
+                                network={network}
+                                tokenStatusType={tokenStatusType}
+                                isUnverifiedTable={isUnverifiedTable}
+                                hideRates={hideRates}
+                            />
+                        ))}
+                        {tokensWithoutBalance.length !== 0 && (
+                            <>
+                                <Table.Row>
+                                    <Table.Cell colspan={hideRates ? 3 : 5}>
```
tomasklim: I think to make this sticky, it has to have colspan="1" but the text has to overflow over columns

tomasklim (commented):
When I implement suggested changes, it looks great 

Screenshot 2024-09-14 at 7 58 42

packages/components/src/components/Table/TableCell.tsx
seibei-iguchi: It works for me, weird. You have chromium based browser?

packages/components/src/components/Table/TableRow.tsx
seibei-iguchi: Weird! Doesn't happen on my side. Will investigate.

packages/suite/src/views/wallet/tokens/common/TokensTable/TokensTable.tsx
seibei-iguchi: Yes you're right, I've experimented with this and it also works with colspan 2 without overflowing

packages/components/src/components/Card/Card.tsx
```diff
@@ -77,6 +77,7 @@ const CardContainer = styled.div<
     padding: ${mapPaddingTypeToPadding};
     background: ${mapElevationToBackground};
     border-radius: ${borders.radii.md};
+    overflow: hidden;
```
jvaclavik: I'm not sure if this doesn't break anything. Thinking about adding it as a frame props. WDYT?

packages/components/src/components/Table/TableCell.tsx
seibei-iguchi: It worked in Firefox but not Blink-based browsers, fixed.

packages/components/src/components/Table/TableRow.tsx
seibei-iguchi: Fixed.

packages/suite/src/views/wallet/tokens/common/TokensTable/TokensTable.tsx
seibei-iguchi: Fixed.

packages/components/src/components/Table/Table.tsx
```diff
@@ -1,56 +1,63 @@
...
-interface TableProps {
+const ScrollContainer = styled.div`
+    overflow: auto hidden;
+    -webkit-overflow-scrolling: touch;
+`;
+
+export type TableProps = AllowedFrameProps & {
     children: ReactNode;
-}
+    colWidths?: string[];
```
jvaclavik, line 32: btw shouldn't we rename `colspan` to `colSpan` to have it in the same style?

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -1,26 +1,61 @@
...
+
+    return map[align];
+};
+
+const Cell = styled.td<{ $isHeader: boolean; $elevation: Elevation }>`
     ${({ $isHeader }) => ($isHeader ? typography.hint : typography.body)}
     color: ${({ theme, $isHeader }) => ($isHeader ? theme.textSubdued : theme.textDefault)};
+    text-align: left;
+    padding: ${spacingsPx.sm} ${spacingsPx.lg};
+    max-width: 300px;
```
jvaclavik, line 31: Why do we need this?

packages/components/src/components/typography/Text/Text.tsx
```diff
@@ -20,9 +19,16 @@ type AllowedTextTextProps = Pick<TextPropsCommon, (typeof allowedTextTextProps)[
 export const allowedTextFrameProps = ['margin'] as const satisfies FramePropsKeys[];
 type AllowedFrameProps = Pick<FrameProps, (typeof allowedTextFrameProps)[number]>;
 
-export type TextVariant =
-    | Extract<UIVariant, 'primary' | 'secondary' | 'tertiary' | 'info' | 'warning' | 'destructive'>
-    | 'purple';
+export const textVariants = [
```
jvaclavik: We should use UIVariants (discussed in person)

jvaclavik (approved):

packages/components/src/components/Card/Card.tsx
seibei-iguchi: Agree, makes sense.

PR #14017, opened 2024-08-29
Ci/add preminor releases

karliatto:
## Description
Modify the bump versions scripts for connect so it is possible to do preminor releases, right now the process will be:

   - minor: Increase minor version (if version 9.5.0-beta.3 --> 9.5.0)
   - patch: Increase patch version (if version 9.4.2-beta.3 --> 9.4.2)
   - preminor: Increase preminor version, pre-release (if version 9.4.1 --> 9.5.0-beta.1)
   - prepatch: Increase prepatch version, pre-release (if version 9.4.1 --> 9.4.2-beta.1)
   - prerelease: Increase prerelease version (if version 9.4.1-beta.1 --> 9.4.1-beta.2)


Hannsek:
> minor: Increase minor version (if version 9.4.1 --> 9.5.0)
> patch: Increase patch version (if version 9.4.1 --> 9.4.2)

between these there will always be 9.5.0-beta.X or 9.4.2-beta.X.

karliatto:
> between these there will always be 9.5.0-beta.X or 9.4.2-beta.X.

Yes, you are right:

* if you are in version 9.5.0-beta.x and run it with `minor` it will create version `9.5.0`
* if you are in version 9.5.1-beta.x and run it with `patch` it will create version `9.5.1`

Since that is the way we are going to use it, better use that as example :+1:

vdovhanych (approved):

PR #14019, opened 2024-08-29
fix(transport): remove BridgeTransport.receive timeout

szymonlesisz:
## Description

`/read` should never timeout (scheduleAction default timeout is set to 10sec)

### DIFF
--- a/packages/transport/src/transports/bridge.ts
+++ b/packages/transport/src/transports/bridge.ts
@@ -294,24 +294,27 @@ export class BridgeTransport extends AbstractTransport {
         session,
         protocol: customProtocol,
     }: AbstractTransportMethodParams<'receive'>) {
-        return this.scheduleAction(async signal => {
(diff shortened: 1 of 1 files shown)


packages/transport/src/transports/bridge.ts
```diff
@@ -269,25 +269,28 @@ export class BridgeTransport extends AbstractTransport {
...
+                });
+                if (!response.success) {
+                    return response;
+                }
 
-            return this.success(undefined);
-        });
+                return this.success(undefined);
+            },
+            { timeout: undefined },
```
mroz22, line 316: but this is post, not read/received as stated in PR description?

szymonlesisz: lol, sorry my bad! fixed, forcepushed

mroz22 (approved):

Issue #13839, opened 2024-08-16
Refactor and type network config


tomasklim:
There is file `networksConfig.ts`

In this file, there is `networks` variable declaring all networks available in Suite and there is also `networksCompatibility` which is a transformation of `networks` to format used in the past.

**Issues:**
- `networks` variable is not typed, the type is generated from the structure of file
  - it would be best to type it so that it is easier to work with it
- `networksCompatibility ` is dynamic and it is almost a blackbox how to work with it. It would be great to finally get rid of it in behalf of `networks`

- together with those changes we can maybe change (as a bonus, not really needed but would be nice)
  - `explorer` can have `base` url and then just `path` for `tx, account, address`,...
  - `default` account type which is now on root of each network could be moved to `accountTypes` under `default` type
  - `customBackends` can be called just `backendTypes` or `backends`
  - I am not sure if we really need `support`, it is also defined in `coins.json`
    - cardano: old fw version of T is deprecated anyway so we can just bump supported version in there maybe
    - solana is 1:1
    - probably we need it but extended for seed type, see Optimisation #13738

**Why:**
- cleaner codebase
- easier work
- easier to implement - different `networkSymbol` from `networkCurrency` required for L2
  - `ln`: `BTC`
  - `op`: `ETH`
  - `base`: `ETH`
  - ...

Lemonexe:
All of the PRs above :arrow_up: solve only the first two bullet points, not the 3rd (_together with those changes we can..._)

The bonus work is **not done** yet. I'd suggest moving it to another issue, because wow, this was already much more work than I anticipated at first :open_mouth:

Also, there are no unit tests for [wallet-config utils](https://github.com/trezor/trezor-suite/blob/develop/suite-common/wallet-config/src/utils.ts) :see_no_evil:

Lemonexe:
Created
- #14444 unit tests
- #14445 moved the bonus bullet points except the 2nd one - I don't think it's necessary, there is a helper for this purpose introduced in [this commit of #14281](https://github.com/trezor/trezor-suite/pull/14281/commits/8f0270267d714f747542facdee446861ef20ed69)

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 24.10.0 (53303523db1a941907766902bcf79b3f320b4dd2)
[...]

PR #14021, opened 2024-08-29
chore(wallet-utils): Typecheck networksConfig declaration

Lemonexe:
1st of many PRs to refactor & cleanup `networksConfig`.

- Implement a generic type check for declaration of `networks`
- Rename old Network type to NetworkCompatibility
  - lots of changes across the rpo, but only TS, so no need to QA

## Description

Currently `networks` have type inferred from its value, and readonly. IMHO we want to keep that.

But currently it lacks any typechecking when declaring it, so any mistake will just silently carry on to the inferred type:

```
export const networks = {
    btc: {
        name: 'Bitcoin',
        BAD_PROPERTY: "lol",
        networkType: 'bitcoin',
```

This PR adds a kind of generic template type which is not exported, we will still rely on exact type inferrence from value for convenient DX, but serves to type check the declaration.

## Related Issue

Part of #13839

### DIFF
--- a/suite-common/wallet-config/src/networksConfig.ts
+++ b/suite-common/wallet-config/src/networksConfig.ts
@@ -653,7 +653,10 @@ export const networks = {
 type InferredNetworks = typeof networks;
 type InferredNetwork = InferredNetworks[NetworkSymbol];
 
-export type Network = Without<InferredNetwork, 'accountTypes'> & {
+/**
+ * @deprecated Old network object interface for backwards copatibility.
+ */
+export type NetworkCompatible = Without<InferredNetwork, 'accountTypes'> & {
     symbol: NetworkSymbol;
     accountType?: AccountType;
     backendType?: BackendType;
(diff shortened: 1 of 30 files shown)


suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -1,5 +1,101 @@
 import { DeviceModelInternal } from '@trezor/connect';
-import type { Keys, Without } from '@trezor/type-utils';
+import type { Without } from '@trezor/type-utils';
+
```
Lemonexe, line 3: I intend to migrate the types to separate file later. Now it would create a circular dependency..

suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -1,5 +1,101 @@
...
+};
+
+export type Account = {
+    bip43Path: string;
+    backendType?: BackendType;
+    features?: NetworkFeature[];
+    isDebugOnlyAccountType?: boolean;
+};
+
+// template types serve only to check if `networks` satisfies it. Exact type is inferred below
```
peter-sanderson, line 74: if they are used only for `satisfies` check, than we shall not export them I guess

peter-sanderson: I would limit exports here for now and allow exports only once we are sure what we really need elswhere

peter-sanderson (approved):

suite-common/wallet-config/src/networksConfig.ts
Lemonexe: But there are many parts of code which work with "a network in general", not a specific network, and those currently import `NetworkCompatible`. When I refactor it, they shall import `Network` instead, so it'll be needed.

Also I intend to split this file into several, so it's a little preparation for that.

PR #14027, opened 2024-08-29
fix(suite-native): display fiat value skeleton while fiat rates are loading

### LINKED ISSUE #13299: Loading fiat value should show skeleton instead of 0 value
**Describe the bug**
- fiat values of portfolio are not cached
- when you start app when offline you see all fiat values with zero value
- before fiat values are fetched from servers after you see start app all values are zero too for quite a long time
- same results for latest dev version, 24.7.1 version and two weeks old dev version

**Steps to reproduce:**
1. Import multiple accounts
2. Kill app
3. Start app with phone in flight mode or observe app before values are fetched from servers

**Info:**
 - Suite version: Release notes for 24.6.0 (350)
Last commit hash: 8a20d9cf4567a8077dd95d88212f9bd2eb66f89e

**Screenshots:**

telegram-cloud-photo-size-4-5801191960428528094-y

https://github.com/trezor/trezor-suite/assets/31506317/f965e330-d720-4c4c-add4-57c88a00244a

yanascz:
Make sure we handle fiat values as `null` until fiat rates are loaded.

## Related Issue

Resolve #13299

## Screenshots:

https://github.com/user-attachments/assets/c3f89c0f-47ff-4a42-9b79-1d34884bae0f

### DIFF
--- a/suite-native/assets/src/assetsSelectors.ts
+++ b/suite-native/assets/src/assetsSelectors.ts
@@ -17,15 +17,12 @@ import { BigNumber } from '@trezor/utils/src/bigNumber';
 export interface AssetType {
     symbol: NetworkSymbol;
     assetBalance: string;
-    fiatBalance: string;
+    fiatBalance: string | null;
 }
 
 type AssetsRootState = AccountsRootState & FiatRatesRootState & SettingsSliceRootState;
 
-const sumCryptoBalance = (balances: string[]): BigNumber =>
-    balances.reduce((prev, balance) => prev.plus(balance), new BigNumber(0));
-
-/* 
+/*
 We do not memoize any of following selectors because they are using only with `useSelectorDeepComparison` hook which is faster than memoization in proxy-memoize.
 */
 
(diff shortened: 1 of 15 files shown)


suite-native/assets/src/assetsSelectors.ts
```diff
@@ -80,22 +77,19 @@ export const selectDeviceAssetsWithBalances = (state: AssetsRootState & DeviceRo
...
-                return prev;
-            }, 0)
-            .toFixed();
-
-        const cryptoValue = sumCryptoBalance(networkAccounts.map(account => account.cryptoValue));
+        const assetBalance = networkAccounts.reduce(
+            (sum, { cryptoValue }) => sum.plus(cryptoValue),
+            new BigNumber(0),
+        );
+        const fiatBalance = networkAccounts.reduce(
```
Nodonisko, line 84: I am quite concern about performance here, this selector is already on my list for being slow and BigNumber is one of biggest perf bottlenecks.

suite-native/formatters/src/components/FiatAmountFormatter.tsx
```diff
@@ -21,9 +21,12 @@ export const FiatAmountFormatter = ({
 }: FiatAmountFormatterProps) => {
     const { FiatAmountFormatter: formatter } = useFormatters();
 
-    const isTestnetValue = !!network && isTestnet(network);
```
Nodonisko, line 24: Can you please keep it as variable?

suite-native/assets/src/assetsSelectors.ts
yanascz: `BigNumber` was already used to add crypto amounts and is also used elsewhere when dealing with fiat amounts. I believe it makes sense considering the rounding errors with `number`.

What other solution did you have in mind?

suite-native/formatters/src/components/FiatAmountFormatter.tsx
yanascz: What added value does it have?

suite-native/assets/src/assetsSelectors.ts
yanascz: Changed back to `Number` based on our discussion in person.

Nodonisko (approved):

suite-native/formatters/src/components/FiatAmountFormatter.tsx
Nodonisko: It's easier to understand what `if` condition is for. Here is relatively simple so it's probably fine dropping it but in many cases it helps a lot.

suite-native/module-accounts-management/src/components/AccountDetailGraphHeader.tsx
```diff
@@ -66,26 +47,17 @@ const CryptoBalance = ({
...
         selectAccountByKey(state, accountKey),
     );
-
     const tokenSymbol = useSelector((state: AccountsRootState) =>
         selectEthereumAccountTokenSymbol(state, accountKey, tokenAddress),
     );
+    const setPoint = useSetAtom(selectedPointAtom);
 
-    const firstGraphPoint = useAtomValue(referencePointAtom);
+    // Reset selected point on unmount so that it doesn't display on device change
```
matejkriz, line 59: 👏😍

PR #14028, opened 2024-08-29
fix(transport): UdpApi trigger transport-interface-change event

szymonlesisz:
fix missing device disconnect or late device connect events

## Description
Turns out `UdpApi` never emits `transport-interface-change` therefore `device-disconnect` or late `device-connect` was never reported

### DIFF
--- a/packages/transport/src/api/udp.ts
+++ b/packages/transport/src/api/udp.ts
@@ -1,5 +1,5 @@
 import UDP from 'dgram';
-import { isNotUndefined } from '@trezor/utils';
+import { isNotUndefined, arrayPartition } from '@trezor/utils';
 
 import {
     AbstractApi,
@@ -7,12 +7,13 @@ import {
     AbstractApiConstructorParams,
     DEVICE_TYPE,
 } from './abstract';
-
+import { DescriptorApiLevel } from '../types';
 import * as ERRORS from '../errors';
 
 export class UdpApi extends AbstractApi {
     chunkSize = 64;
 
+    protected devices: DescriptorApiLevel[] = [];
     protected interface = UDP.createSocket('udp4');
     protected communicating = false;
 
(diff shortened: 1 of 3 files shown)


packages/transport/src/api/udp.ts
```diff
@@ -33,13 +34,31 @@ export class UdpApi extends AbstractApi {
...
         enumerateRecursive();
     }
 
+    private handleEnumerate(result: Awaited<ReturnType<typeof this.enumerate>>) {
+        if (result.success) {
+            const [known, unknown] = arrayPartition(result.payload, device =>
+                this.devices.includes(device),
+            );
+
+            if (known.length !== this.devices.length || unknown.length > 0) {
```
mroz22: isn't uknown.length > 0 enough here?

packages/transport/src/api/udp.ts
```diff
@@ -33,13 +34,31 @@ export class UdpApi extends AbstractApi {
...
+                this.enumerate(this.enumerateAbortController.signal)
+                    .then(r => this.handleEnumerate(r))
+                    .finally(enumerateRecursive);
             }, 500);
         };
 
         enumerateRecursive();
     }
 
+    private handleEnumerate(result: Awaited<ReturnType<typeof this.enumerate>>) {
```
mroz22: enumerate in usb assings it's result to to the this.devices variable. https://github.com/trezor/trezor-suite/blob/develop/packages/transport/src/api/usb.ts#L176
 
maybe this fn should be called also somewhere on line 194?

packages/transport/src/api/udp.ts
szymonlesisz: nope, it will not work for disconnect event, example:
i've got 1 device stored in `this.devices` now after disconnection i have 0 devices in known and 0 in unknown

packages/transport/src/api/udp.ts
szymonlesisz: not sure if this is correct behavior, usb emits `transport-interface-change` from the `listen` function relying on the fact that `enumerate` was called before that. but what if it wasnt? what if i called listen without calling enumerate first?

btw, usb also assigns `this.devices` on line 59 and 82

szymonlesisz: on second thought, if i go with expected way (first enumerate then listen) then `transport-interface-change` will be emitted regardless, so i agree that i should prevent that, but i cant assign `this.devices` (inside enumerate fn) before i check for changes (after enumerate)

i will think about it

szymonlesisz: i've moved handleEnumerate (renamed to handleDevicesChange) to be called after eumerate. 
`transport-interface-change` is emitted only if api is listening

mroz22 (approved):

PR #14045, opened 2024-08-30
chore(wallet-utils): Decouple utility functions from networkCompatibility

Lemonexe:
3rd of many PRs to refactor & cleanup `networksConfig`: prepare new utility functions based on the `networks` object

## Description

:eyes: CR recommended commit by commit:
1. updates the `networks` object with `symbol`, which **automatically** makes it usable at some places
2. prepare new utility functions which are based on `networks` instead of `networksCompatible`
  - in some cases, the new functions work out of the box :tada: 
  - in others, backward compatible utility functions are created
4. :broom:  extract all of the utility functions to a separate file 

## Related Issue

Part of #13839


packages/suite/src/views/wallet/coinmarket/exchange/components/ExchangeForm/Inputs/SendCryptoInput/SendCryptoSelect/index.tsx
```diff
@@ -1,7 +1,7 @@
 import { Select, CoinLogo } from '@trezor/components';
 import { Controller } from 'react-hook-form';
 import styled from 'styled-components';
-import { NetworkSymbol, getEthereumTypeNetworkSymbols } from '@suite-common/wallet-config';
+import { NetworkSymbol, ethereumTypeNetworkSymbols } from '@suite-common/wallet-config';
```
Lemonexe, line 4: see [comment](https://github.com/trezor/trezor-suite/pull/14045#discussion_r1738769103)

suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -86,8 +87,8 @@ export type Network = {
     accountTypes: NetworkAccountTypes;
     isHidden?: boolean; // not used here, but supported elsewhere
     chainId?: number;
-    features?: NetworkFeature[];
-    customBackends?: BackendType[];
+    features: NetworkFeature[];
+    customBackends: BackendType[];
```
Lemonexe, line 91: Forgotten in #14021 - these properties are actually present on all networks.

suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -77,6 +77,7 @@ export type NetworkAccountTypes = Partial<Record<AccountType, Account>>;
 export type NetworkDeviceSupport = Partial<Record<DeviceModelInternal, string>>;
 
 export type Network = {
+    symbol: NetworkSymbol;
```
Lemonexe, line 80: This was also desired (discussed with @tomasklim) , originally I wanted to do it later, but it's suitable to do it now, as it makes a lot of usecases for the util functions work out of the box.

suite-common/wallet-config/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+    Networks,
+    Network,
+    NetworkFeature,
+    networks,
+    networksCompatibility,
+    NetworkSymbol,
+} from './networksConfig';
+
+const networksCollection: Network[] = Object.values(networks);
+
```
Lemonexe, line 12: in some cases we can use `getMainnets` right away - there are some differences between `networks` and `networksCompatibility`, but a lot of it is common.
When the usecase does not rely on those properties which are different, we may use `getMainnets` right away. Those usecases which rely on network.accountType - the `getMainnetsCompatible` and `getTestnetsCompatible` allow to do it later, and have the PRs smaller.

suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -725,15 +725,26 @@ export const getMainnets = (debug = false, bnb = false) =>
...
 
-export const getTestnets = (debug = false) =>
+export const getTestnetsCompatible = (debug = false) =>
     networksCompatibility.filter(
         n => !n.accountType && n.testnet === true && (!n.isDebugOnlyNetwork || debug),
     );
 
-export const getAllNetworkSymbols = () => networksCompatibility.map(n => n.symbol);
+const networksCollection: Network[] = Object.values(networks);
+
```
Lemonexe: `getMainnets` and `getTestnets` are new, but I have manually checked that the **output is the same** as the old versions, except the networks now contain `accountTypes: [...]`. So it's only extended, not limited.

Still - I had to create the `Compatible` functions because of interoperability with `networksCompatibility` - they are often used in conjunction, and TS would scream, even though in this case, there is nothing missing, because both mainnets and testnets have `accountType: undefined`, but TS can't know that :see_no_evil:

suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -725,15 +725,26 @@ export const getMainnets = (debug = false, bnb = false) =>
...
+export const getMainnets = (debug = false, bnb = false) =>
+    networksCollection.filter(
+        n => !n.testnet && (!n.isDebugOnlyNetwork || debug) && (bnb || n.symbol !== 'bnb'),
+    );
+
+export const getTestnets = (debug = false) =>
+    networksCollection.filter(n => n.testnet === true && (!n.isDebugOnlyNetwork || debug));
+
+export const networkSymbols = networksCollection.map(n => n.symbol);
```
Lemonexe: The original `getAllNetworkSymbols` was **unused**.
- It does not have to be a function, as its output was always constant
- Also it did not work properly, it returned array of non-unique symbols, now they're unique :heavy_check_mark:

suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -725,15 +725,26 @@ export const getMainnets = (debug = false, bnb = false) =>
...
 
-export const getEthereumTypeNetworkSymbols = () =>
-    networksCompatibility.filter(n => n.networkType === 'ethereum').map(n => n.symbol);
+export const ethereumTypeNetworkSymbols = networksCollection
+    .filter(n => n.networkType === 'ethereum')
+    .map(n => n.symbol);
```
Lemonexe: - It does not have to be a function, as its output was always constant
- Also it did not work properly, it returned array of non-unique symbols, now they're unique
  - The function was used, but making it unique won't break anything: 
  - All usages are only to query for `ethereumTypeNetworkSymbols.includes`
  - and for that, unique or non-unique doesn't matter :heavy_check_mark:

suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -746,13 +757,11 @@ export const isDebugOnlyAccountType = (
...
 
-    const network = networks?.[symbol];
+    const network = (networks as Networks)?.[symbol];
 
     if (!network) return false;
 
-    const accountTypeInfo = (network.accountTypes as Record<AccountType, NetworkCompatible>)[
-        accountType
-    ];
+    const accountTypeInfo = network.accountTypes[accountType];
```
Lemonexe: Incidental TS enhancement.

suite-common/wallet-types/src/backend.ts
```diff
@@ -10,10 +10,8 @@ export type BlockbookUrl = {
     tor?: boolean; // Added by TOR
 };
 
-export type BackendType = 'blockbook' | 'electrum' | 'ripple' | 'blockfrost' | 'solana';
-
 export type CustomBackend = {
-    coin: (typeof networksCompatibility)[number]['symbol'];
```
Lemonexe, line 16: The changes in this file are because of TS error which was not clear to me, but this way it is cleaner anyway.

suite-common/wallet-config/src/networksConfig.ts
komret: Let's remove it if it is not used.

komret (approved):

suite-common/wallet-config/src/networksConfig.ts
Lemonexe: Done :heavy_check_mark:

PR #13997, opened 2024-08-27
Remove legacy Icons, use only new icons

peter-sanderson:

PR #14046, opened 2024-08-30
fix(suite-common): unify icon SVG path fill to #000

yanascz:
Some icons are not visible in the mobile app (nor IDE) because there's fill missing. 

## Related Issue

Caused by #13997


jvaclavik (approved):
nice 👌

PR #14047, opened 2024-08-30
fix(suite): Display rotation

jvaclavik:
## Related Issue

Resolve 

## Screenshots:

before
image

after
image

### DIFF
--- a/packages/suite/src/views/settings/SettingsDevice/DisplayRotation.tsx
+++ b/packages/suite/src/views/settings/SettingsDevice/DisplayRotation.tsx
@@ -3,17 +3,20 @@ import { analytics, EventType } from '@trezor/suite-analytics';
 
 import { SettingsSectionItem } from 'src/components/settings';
 import { ActionColumn, TextColumn, Translation } from 'src/components/suite';
-import { Button, ButtonGroup } from '@trezor/components';
+import { SelectBar } from '@trezor/components';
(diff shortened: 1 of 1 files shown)


packages/suite/src/views/settings/SettingsDevice/DisplayRotation.tsx
```diff
@@ -3,17 +3,20 @@ import { analytics, EventType } from '@trezor/suite-analytics';
...
 import { useDevice, useDispatch } from 'src/hooks/suite';
 import { applySettings } from 'src/actions/settings/deviceSettingsActions';
 import { SettingsAnchor } from 'src/constants/suite/anchors';
 
-const DISPLAY_ROTATIONS = [
+type RotationValue = 0 | 90 | 180 | 270;
+type Rotation = { label: JSX.Element; value: RotationValue };
+
+const DISPLAY_ROTATIONS: Array<Rotation> = [
     { label: <Translation id="TR_NORTH" />, value: 0 },
```
seibei-iguchi, line 15: Isn't something like up, left, etc. more simple to understand?

seibei-iguchi (approved):

packages/suite/src/views/settings/SettingsDevice/DisplayRotation.tsx
jvaclavik: We are going to use arrow icons in the future

PR #14048, opened 2024-08-30
Enable user interaction after a revision check check fails

komret:
## Description
This should allow users with a scammy firmware to access Suite and attempt to send funds away. Generating new addresses is disabled.

@MiroslavProchazka please check copy. Note that the instructions in the initial modal don't explicitly guide you to send funds away from it (I am not saying it has to be there).

## Related Issue

Resolve https://github.com/trezor/trezor-suite-private/issues/110

## Screenshots:
Screenshot 2024-08-30 at 17 09 43
Screenshot 2024-08-30 at 17 09 55
Screenshot 2024-08-30 at 17 10 08
Screenshot 2024-08-30 at 17 10 16

### DIFF
--- a/packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
+++ b/packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
@@ -15,6 +15,7 @@ export const DeviceCompromised = () => {
 
     const revision = device?.features?.revision;
     const version = getFirmwareVersion(device);
+    const vendor = device?.features?.fw_vendor;
     const authenticityError =
         device?.features && device.authenticityChecks?.firmwareRevision?.success === false
             ? device.authenticityChecks.firmwareRevision?.error
@@ -28,7 +29,7 @@ export const DeviceCompromised = () => {
     };
 
     useEffect(() => {
-        const contextData = { revision, version, authenticityError };
+        const contextData = { revision, version, vendor, authenticityError };
 
         withSentryScope(scope => {
             scope.setLevel('error');
@@ -38,7 +39,7 @@ export const DeviceCompromised = () => {
                 scope,
             );
         });
-    }, [revision, version, authenticityError]);
+    }, [authenticityError, revision, vendor, version]);
 
     return (
         <WelcomeLayout>
--- a/packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
+++ b/packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
@@ -1,12 +1,16 @@
-import { WelcomeLayout } from 'src/components/suite';
-import { SecurityCheckFail } from '../SecurityCheck/SecurityCheckFail';
-import { Card } from '@trezor/components';
 import { useEffect } from 'react';
-import { captureSentryMessage, withSentryScope } from '../../../utils/suite/sentry';
-import { useDevice } from '../../../hooks/suite';
+
+import { Card } from '@trezor/components';
 import { getFirmwareVersion } from '@trezor/device-utils';
 
+import { WelcomeLayout } from 'src/components/suite';
+import { useDevice, useDispatch } from 'src/hooks/suite';
+import { captureSentryMessage, withSentryScope } from 'src/utils/suite/sentry';
+import { SecurityCheckFail } from '../SecurityCheck/SecurityCheckFail';
+import { deviceActions } from '@suite-common/wallet-core';
+
 export const DeviceCompromised = () => {
+    const dispatch = useDispatch();
     const { device } = useDevice();
 
     const revision = device?.features?.revision;
@@ -15,6 +19,13 @@ export const DeviceCompromised = () => {
         device?.features && device.authenticityChecks?.firmwareRevision?.success === false
             ? device.authenticityChecks.firmwareRevision?.error
             : undefined;
+
+    const goToSuite = () => {
+        // Condition to satisfy TypeScript, device.id is always defined at this point.
+        if (device?.id) {
+            dispatch(deviceActions.dismissFirmwareRevisionCheck(device.id));
+        }
+    };
 
     useEffect(() => {
         const contextData = { revision, version, authenticityError };
@@ -32,7 +43,7 @@ export const DeviceCompromised = () => {
     return (
         <WelcomeLayout>
             <Card data-testid="@device-compromised">
-                <SecurityCheckFail />
+                <SecurityCheckFail goBack={goToSuite} />
             </Card>
         </WelcomeLayout>
     );
--- a/packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
+++ b/packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
@@ -78,12 +78,17 @@ const supportChatUrl = `${TREZOR_SUPPORT_URL}#open-chat`;
 
 interface SecurityCheckFailProps {
     goBack?: () => void;
+    useSoftMessaging?: boolean;
 }
 
-export const SecurityCheckFail = ({ goBack }: SecurityCheckFailProps) => {
-    const heading = goBack ? 'TR_DEVICE_COMPROMISED_HEADING_SOFT' : 'TR_DEVICE_COMPROMISED_HEADING';
-    const text = goBack ? 'TR_DEVICE_COMPROMISED_TEXT_SOFT' : 'TR_DEVICE_COMPROMISED_TEXT';
-    const items = goBack ? softChecklistItems : checklistItems;
+export const SecurityCheckFail = ({ goBack, useSoftMessaging }: SecurityCheckFailProps) => {
+    const heading = useSoftMessaging
+        ? 'TR_DEVICE_COMPROMISED_HEADING_SOFT'
+        : 'TR_DEVICE_COMPROMISED_HEADING';
+    const text = useSoftMessaging
+        ? 'TR_DEVICE_COMPROMISED_TEXT_SOFT'
+        : 'TR_DEVICE_COMPROMISED_TEXT';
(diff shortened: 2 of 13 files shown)


MiroslavProchazka (commented):
Please use "TR_DEVICE_COMPROMISED_HEADING" together with "TR_DEVICE_COMPROMISED_TEXT" instead of "TR_PLAY_IT_SAFE" and "TR_DEVICE_COMPROMISED_TEXT_SOFT" as was originally in the first implementation

MiroslavProchazka (commented):
Please add tracking for `fw_vendor` property to Sentry, so we can eliminate false-positive cases

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
```diff
@@ -16,6 +20,12 @@ export const DeviceCompromised = () => {
             ? device.authenticityChecks.firmwareRevision?.error
             : undefined;
 
+    const goToSuite = () => {
+        if (device?.id) {
```
Lemonexe, line 26: :thinking: It seems impossible to display `DeviceCompromised` if `device` is undefined (looking at `Preloader.tsx`). I would just put a comment about it, otherwise, I'd wonder if `goToSuite` works correctly for the case of `device?.id = undefined`, but that's not a releavant case at all.

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -772,6 +783,27 @@ export const selectSelectedDeviceAuthenticity = (state: DeviceRootState) => {
     return device?.id ? deviceAuthenticity?.[device.id] : undefined;
 };
 
+export const selectFailedSecurityChecks = (state: DeviceRootState) => {
+    const device = selectDevice(state);
+
+    return device?.features && device.authenticityChecks !== undefined
```
Lemonexe: Why is `device?.features` relevant?

Lemonexe (approved):
LGTM, tested and it works as expected.
Just two small questions.

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -31,16 +31,15 @@ export type State = {
     devices: TrezorDevice[];
     selectedDevice?: TrezorDevice;
     deviceAuthenticity?: Record<string, StoredAuthenticateDeviceResult>;
+    dismissedSecurityChecks?: {
+        firmwareRevision?: string[];
```
peter-sanderson, line 35: optional/nit: adding a nominal type (branded type) for this would be great

peter-sanderson (approved):

suite-common/wallet-core/src/device/deviceReducer.ts
peter-sanderson: @Lemonexe this check will trigger type narrowing in Typescript and the device is then resolved to `KnownDevice` which has `authenticityChecks` property. Other types unioned in `device` does not have it.

Lemonexe: Ah, thanks for clarification. Maybe introducing a typeguard function for it might make it clearer & reusable, but that's a tiny nit :slightly_smiling_face:

peter-sanderson: > typeguard function

good point! something `isKnownDevice(..)` would be great!

komret:
> Please use "TR_DEVICE_COMPROMISED_HEADING" together with "TR_DEVICE_COMPROMISED_TEXT" instead of "TR_PLAY_IT_SAFE" and "TR_DEVICE_COMPROMISED_TEXT_SOFT" as was originally in the first implementation

501391365d5a09b8f5f1d115b98411533252d4a5

> Please add tracking for `fw_vendor` property to Sentry, so we can eliminate false-positive cases

6b7e9787d40b9765b10eb8bb0a59a74d8ee7a70d

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
komret: 6dbac509081c221c50c51f1fb900e8f1565a16e7

suite-common/wallet-core/src/device/deviceReducer.ts
komret: I didn't add a type guard, but hopefully made it more readable in 6dbac509081c221c50c51f1fb900e8f1565a16e7.

suite-common/wallet-core/src/device/deviceReducer.ts
komret: Good point, but `device.id` type goes deep so I wouldn't do it here.

suite-common/wallet-core/src/device/deviceReducer.ts
peter-sanderson: I talked to @mroz22 some time ago and he pointed me to check features to assert its `KnownDevice`. I think its a common pattern in the codebase. 

I am even for keeping it here and then refactor it into typeguead at once rather then introducing other "style" of checking for it.

komret:
Tiny design changes in 560796fbab8aaa544f9bc64002eb6f9a3063603c:

Screenshot 2024-09-02 at 12 27 31

suite-common/wallet-core/src/device/deviceReducer.ts
peter-sanderson: its not device.id or is it? I am talking about the `Revision` and I think it has not that many usages

peter-sanderson: but its a nit, also I would prefer to do it in different PR if doing it

komret: It is actually an array of device IDs. I can see it is not quite apparent. The idea is that there would be a similar array for failed hash checks etc. in future.

packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
```diff
@@ -59,44 +40,35 @@ const checklistItems = [
...
     goBack?: () => void;
+    useSoftMessaging?: boolean;
 }
 
-export const SecurityCheckFail = ({ goBack }: SecurityCheckFailProps) => {
-    const heading = goBack ? 'TR_DEVICE_COMPROMISED_HEADING_SOFT' : 'TR_DEVICE_COMPROMISED_HEADING';
-    const text = goBack ? 'TR_DEVICE_COMPROMISED_TEXT_SOFT' : 'TR_DEVICE_COMPROMISED_TEXT';
-    const items = goBack ? softChecklistItems : checklistItems;
+export const SecurityCheckFail = ({ goBack, useSoftMessaging }: SecurityCheckFailProps) => {
+    const heading = useSoftMessaging
```
peter-sanderson, line 51: I see as a  good practice to type variables like this. 

1. It will enable auto-competition for IDE
2. Type error will be triggered here, at the source of the issue and not down below in the component usage

packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
```diff
@@ -59,44 +40,35 @@ const checklistItems = [
...
+        ? 'TR_DEVICE_COMPROMISED_HEADING_SOFT'
+        : 'TR_DEVICE_COMPROMISED_HEADING';
+    const text = useSoftMessaging
```
peter-sanderson, line 54: same here

packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
peter-sanderson, line 51:
```suggestion
    const heading: TranslatioKey = useSoftMessaging
```

peter-sanderson (approved):

packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
komret: I don't know about this, it feels too pedantic to me.

Ad 1, it is questionable if this saves you time because you have to type more.
Ad 2, I agree.
3) This breaks the inference as `heading` becomes a general `TranslationKey` instead of a literal union.

It could be a topic for Code Improvements though.

Lemonexe: :thinking:  :thought_balloon:  **If** inference isn't enough, and we want explicit declaration for stricter typesafety _(imho we don't)_, then how about `satisfies`?
It'd tell us if `useSoftMessaging` returns something else than we expect, but at the same time, keep type inferrence, which is more specific than `TranslationKey`

komret: Good idea, I like that a bit more, although it doesn't enable autocompletion as per point 1

peter-sanderson: > Ad 1, it is questionable if this saves you time because you have to type more.

Its not (just) for you, but for anybody who needs to change it in the future.

> 3) This breaks the inference as heading becomes a general TranslationKey instead of a literal union.

Has this any implications? I don't see any.

bosomt:
@komret sent this issue, hint is rendered even when using legit device https://github.com/trezor/trezor-suite/issues/14084

packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
komret: > Has this any implications? I don't see any.

Not much TBH. It shows you a narrower type on hover. It would be convenient in case you wanted to manipulate the strings later in code.

PR #14049, opened 2024-08-30
chore(suite-native): coin enabling analytics

### LINKED ISSUE #13035: Coin Enabling - analytics
1. Coins Active on Startup: Total Number of Coins - Overall
- Which coins are enabled by user when the selection is displayed to them
- Which coins are initially active on Start up (after user did the selection previously)
2. Coins Active on Startup - Overall
3. Average of active coins
4. Median of active coins
5. Change Enabled coins in settings
- When user change the enabled coins in settings

vytick:
Discussed with Tomas, and based on his opinion added enabled network symbols to appReady event, and new event for initial setup and also a new event for changing if network is enabled in settings.

referenced in https://www.notion.so/satoshilabs/da354c80a23a44a2814748968f701c5a?v=6a0a9e88bd04494db46c139336122d37

## Related Issue

Resolve #13035


suite-native/coin-enabling/src/components/NetworkSymbolSwitchItem.tsx
```diff
@@ -95,6 +98,16 @@ export const NetworkSymbolSwitchItem = ({
             });
         }
         dispatch(toggleEnabledDiscoveryNetworkSymbol(networkSymbol));
+
+        if (allowChangeAnalytics) {
```
PeKne, line 101: Just an idea... Should not this be dispatched only if the user leaves the screen? Like this he might turn it on and off multiple times but the coin is not enabled/disabled until he leaves the screen. WDYT?

Otherwise looks good.

vytick: confirming once more the best behaviour (not only tech-wise) with @mnuky

vytick: It does not feel to be a common way of user interaction and (imo) increased complexity does not bring much value. Wdyt?

PeKne: IDK, this should be decided by the data team. Or they should be at least informed about this fact, so they know that the measured data could be skewed.

PeKne:
Please add the changes also to the analytics changelog: https://www.notion.so/satoshilabs/Changelog-Mobile-app-946307159a294bcdb90b94e90eefd292

PeKne (approved):

suite-native/coin-enabling/src/components/NetworkSymbolSwitchItem.tsx
vytick: Checked with Tomas, as you suggested, and he said, it is fine like this.

PR #14050, opened 2024-09-01
chore(suite): Add error rule for not overriding components and product-components

jvaclavik:
## Description

- overriding components is now throwing error, not just warning
- all existing occurrences have comment for disabling this rule as exception (it's going to be refactored later)
- add rule also for product-components

This allows not seeing 300+ warnings during checking every PR

## Related Issue

Resolve

### DIFF
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/CoinmarketTermsModal.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/CoinmarketTermsModal.tsx
@@ -123,7 +123,7 @@ export const CoinmarketTermsModal = ({
                 </Row>
                 <Row alignItems="center">
                     <IconWrapper $backgroundColor="#e8f3fa">
-                        <Icon size={24} name={`checkActive`} color="#1d88c5" />
+                        <Icon size={24} name="checkActive" color="#1d88c5" />
                     </IconWrapper>
 
                     <ContentWrapper>
@@ -140,7 +140,7 @@ export const CoinmarketTermsModal = ({
                 </Row>
                 <Row alignItems="center">
                     <IconWrapper $backgroundColor="#f9f4e6">
-                        <Icon size={24} name={`pencil`} color="#c19009" />
+                        <Icon size={24} name="pencil" color="#c19009" />
                     </IconWrapper>
                     <ContentWrapper>
                         <Text typographyStyle="highlight">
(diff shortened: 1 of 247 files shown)


packages/components/src/index.ts
```diff
@@ -21,6 +21,7 @@ export * from './components/Icon/Icon';
 export * from './components/buttons/PinButton/PinButton';
 export * from './components/buttons/TextButton/TextButton';
 export * from './components/buttons/TooltipButton/TooltipButton';
+export * from './components/Table/Table';
```
peter-sanderson: you wanted to not to do `*` exports maybe lets not adding a new one

packages/transport-bridge/src/ui/components/Card.tsx
```diff
@@ -1,7 +1,8 @@
 import styled from 'styled-components';
 
-import { Card as ComponentsCard } from '@trezor/components/src/components/Card/Card';
+import { Card as ComponentsCard } from '@trezor/components';
```
peter-sanderson, line 3: :clap:

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/CoinmarketTermsModal.tsx
```diff
@@ -123,7 +123,7 @@ export const CoinmarketTermsModal = ({
                 </Row>
                 <Row alignItems="center">
                     <IconWrapper $backgroundColor="#e8f3fa">
-                        <Icon size={24} name={`checkActive`} color="#1d88c5" />
```
peter-sanderson, line 126: It is interesting that this is not covered by current eslint rules. Can it be?

peter-sanderson (approved):
In general I really like the idea of ignoring all the current occurrences and fixing them later without bugging devs with ton of warnings.

But it shall be done rather sooner then later, as this `// ignore` comment can be easily taken as an inspiration for new overrides :smile:

packages/components/src/index.ts
jvaclavik: This shouldn't be in this PR 🫣 deleting

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/CoinmarketTermsModal.tsx
jvaclavik: It should be definitely possible to set it in prettier and auto correct it.

peter-sanderson: I think its already there, but probably not working / not configured for backticks ``

PR #14053, opened 2024-09-02
fix(suite): change misleading copy mentioning browser in desktop debu…

mroz22:
based on discussion here https://satoshilabs.slack.com/archives/CKZHV2G13/p1724747991665409

the screen you will see when for whatever reason bridge dies (or does not start) while using desktop application.

image


MiroslavProchazka (approved):

MiroslavProchazka:
I wonder whether simple Suite restart/refresh can solve this in some cases. If yes, we may add it above the "Try restarting your computer"

mroz22:
As of now, it can't probably.

PR #14058, opened 2024-09-02
fix(suite-desktop-core): await bridge module load

martykan:
## Description

Bridge module loading was previously not awaited, according to a comment - to mimic previous behavior. 

However this can sometimes cause the "No transport" screen to appear for a short time while the app is starting up. 
Awaiting the module solves this.

### DIFF
--- a/packages/suite-desktop-core/src/modules/bridge.ts
+++ b/packages/suite-desktop-core/src/modules/bridge.ts
@@ -4,6 +4,7 @@
 import { TrezordNode } from '@trezor/transport-bridge';
 import { isDevEnv } from '@suite-common/suite-utils';
 import { validateIpcMessage } from '@trezor/ipc-proxy';
+import { scheduleAction } from '@trezor/utils';
 
 import { app, ipcMain } from '../typed-electron';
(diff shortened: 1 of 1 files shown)


mroz22:
two notes to discuss:

1] the "No transport" screen" could also be handled by application logic. It might be related to some of the recent refactorings of DeviceList
2] maybe awaited modules should have some timeout in case they are not critical? Or maybe non critical modules shoudln't be awaited at all? 🤔 

lets summon @marekrjpolak 😄

martykan:
The application logic is currently responding to the `TRANSPORT.ERROR` event, so I'm not sure if we can do anything sensible from that side. Awaiting the module seems like the best solution in my opinion, but we definitely need to make sure the edge cases are covered

marekrjpolak (approved):
We've agreed that
- awaiting the Bridge module is the cleanest solution,
- it should be capped (e.g. to 3 seconds) after which obtaining `TRANSPORT.ERROR` is inevitable, and
- `if (loaded) return` will be kept (because it's also possible to restart Bridge on every soft app reload (Ctrl+R), but it probably doesn't make much sense).

:heavy_check_mark:

PR #14060, opened 2024-09-02
chore(wallet-config): FIle reorganisation (extract types)

Lemonexe:
4th of many PRs to refactor & cleanup `networksConfig`: 

## Description

- move types from `networkConfig.ts` into `types.ts`
- move type guard functions from `types.ts` to `utils.ts`
  - it'd create circular dependency, but more importantly they're not really types, so they don't belong there!
  - remove unused `isEthereumBasedNetwork`

## Related Issue

Part of #13839

### DIFF
--- a/suite-common/wallet-config/src/types.ts
+++ b/suite-common/wallet-config/src/types.ts
@@ -1,9 +1,98 @@
-import { networks, NetworkSymbol } from './networksConfig';
+import { DeviceModelInternal } from '@trezor/connect';
 
-export const isNetworkSymbol = (symbol: NetworkSymbol | string): symbol is NetworkSymbol =>
-    symbol in networks;
+export type NetworkSymbol =
+    | 'btc'
+    | 'ltc'
+    | 'eth'
+    | 'etc'
+    | 'xrp'
+    | 'bch'
+    | 'btg'
+    | 'dash'
+    | 'dgb'
+    | 'doge'
+    | 'nmc'
+    | 'vtc'
+    | 'zec'
+    | 'ada'
+    | 'sol'
+    | 'matic'
+    | 'bnb'
+    | 'test'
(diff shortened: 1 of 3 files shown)


suite-common/wallet-config/src/types.ts
```diff
@@ -1,9 +1,98 @@
...
+    | 'tsep'
+    | 'thol'
+    | 'txrp'
+    | 'tada'
+    | 'dsol';
 
-export const isEthereumBasedNetwork = (
-    network: (typeof networks)[NetworkSymbol],
-): network is (typeof networks)[NetworkSymbol] & { chainId: number } =>
-    network.networkType === 'ethereum';
```
Lemonexe, line 9: `isEthereumBasedNetwork` is unused, I deleted it

suite-common/wallet-config/src/utils.ts
```diff
@@ -69,3 +62,6 @@ export const getNetworkFeatures = (symbol: NetworkSymbol) =>
     networks[symbol]?.features as unknown as NetworkFeature;
 
 export const getCoingeckoId = (symbol: NetworkSymbol) => networks[symbol]?.coingeckoId;
+
+export const isNetworkSymbol = (symbol: NetworkSymbol | string): symbol is NetworkSymbol =>
+    networks.hasOwnProperty(symbol);
```
Lemonexe, line 67: I changed this from `symbol in networks` to `hasOwnProperty`, because for example `'toString'` shouldn't be evaluated as a valid network symbol :innocent: 

:thought_balloon: IMHO `in` is fine if we are checking a specific key which surely isn't inherited from prototype, like `'btc' in networks`, but a function that checks any string in general should use `hasOwnProperty` to exclude inherited properties.

komret (approved):

PR #14066, opened 2024-09-02
chore: add isDeviceAcquired type guard

komret:
## Description

Add `isDeviceAcquired` type guard as discussed [here](https://github.com/trezor/trezor-suite/pull/14048#discussion_r1740590685).


Lemonexe (approved):
LGTM, thanks :+1:

PR #14068, opened 2024-09-02
feat(suite-native): transaction detail screen facelift

### LINKED ISSUE #13764: Send coins - sign transaction improvements
The transaction detail page has to be redesigned according to the attached Figma link. After are these changes take place, the successfull send flow redirects to the transaction detail page, where user can see all the transaction data and metadata.

## Acceptance criteria 
- [ ] When user sends the transaction, transaction is proceeded to mempool to be mined
- [ ] User is redirected to detail of the transaction
- [ ] When transaction is sent in the mobile app, user sees the status of the transaction Pending
- [ ] When transaction is confirmed, it is marked as Confirmed

## Resources
- link to [DESIGN](https://www.figma.com/design/sqKFZAIvk3zERxiWgthK8K/Send?node-id=1724-35592&t=KTigu9fek47mphXN-4)

PeKne:
## Description
- [x] When user sends the transaction, transaction is proceeded to mempool to be mined
- [x] User is redirected to detail of the transaction
- [x] When transaction is sent in the mobile app, user sees the status of the transaction Pending
- [x] When transaction is confirmed, it is marked as Confirmed

## Related Issue

Resolve #13764

## Screenshots:
https://github.com/user-attachments/assets/853b8afb-467e-493c-af10-4915d8e564c6


suite-native/transactions/src/components/TransactionDetail/TransactionDetailRow.tsx
```diff
@@ -10,11 +10,11 @@ type TransactionDetailRowProps = {
     children: ReactNode;
 };
 
-const rowStyle = prepareNativeStyle(utils => ({
+const rowStyle = prepareNativeStyle(() => ({
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
-    marginVertical: utils.spacings.small,
+    height: 42,
```
matejkriz: Fixed value for height doesn't feel right here, is it really necessary?

suite-native/transactions/src/components/TransactionDetail/TransactionDetailSheets.tsx
```diff
@@ -27,11 +28,14 @@ const sheetToAnalyticsEventMap: Record<SheetType, TransactionSheetAnalyticsEvent
     inputs: EventType.TransactionDetailInputOutput,
 };
 
+const cardStyle = prepareNativeStyle(() => ({ paddingVertical: 12, paddingRight: 12 }));
```
matejkriz, line 31: Better to use spacing util instead of hardcoded values? Also overriding styles for Card atom component is not nice. Maybe we need multiple Card component variants?

suite-native/module-send/src/components/OutputsReviewFooter.tsx
```diff
@@ -61,17 +65,14 @@ export const OutputsReviewFooter = ({ accountKey }: { accountKey: AccountKey })
     if (!signedTransaction || !account) return <ConfirmOnTrezorImage />;
 
     const handleSendTransaction = async () => {
-        const sendResponse = await dispatch(
-            sendTransactionAndCleanupSendFormThunk({ account, signedTransaction }),
-        );
+        const sendResponse = await dispatch(sendTransactionAndCleanupSendFormThunk({ account }));
```
matejkriz, line 68: I was "stucked" on the Send transaction screen (Review transaction), when I click on the Send button. Maybe it's because of this await? If so, we should add some loading state on that page too.

suite-native/transactions/src/components/TransactionDetail/NetworkTransactionDetailSummary.tsx
```diff
@@ -46,14 +63,14 @@ export const NetworkTransactionDetailSummary = ({
...
             )}
-            <TransactionDetailStatusSection txid={txid} accountKey={accountKey} />
             {A.isNotEmpty(transactionOutputAddresses) && (
                 <TransactionDetailAddressesSection
                     addressesType="outputs"
                     addresses={transactionOutputAddresses}
                     onShowMore={onShowMore}
                 />
             )}
+            <VerticalSeparator isMultiAddress={transactionInputAddresses.length > 2} />
```
matejkriz: maybe?
```suggestion
            <VerticalSeparator isMultiAddress={transactionOutputAddresses.length > 2} />
```

matejkriz: I don't get this, one thing is if we should check length of input or output addresses, other thing is why to check for more than 2 and not just more than 1?

matejkriz: Would be awesome to solve it without this param completely, but not sure if there is an easy solution for dynamic height here.

suite-native/transactions/src/components/TransactionDetail/NetworkTransactionDetailSummary.tsx
```diff
@@ -7,11 +7,28 @@ import {
...
 import { TransactionDetailAddressesSection } from './TransactionDetailAddressesSection';
-import { TransactionDetailStatusSection } from './TransactionDetailStatusSection';
+
+const separatorStyle = prepareNativeStyle<{ isMultiAddress: boolean }>(
+    (utils, { isMultiAddress }) => ({
+        position: 'absolute',
+        left: 7.5,
+        top: 37,
+        width: 1,
+        height: isMultiAddress ? 90 : 34,
```
matejkriz: Is the static height really needed? And why 90 for "multi address"? There can be more outputs...

suite-native/transactions/src/components/TransactionDetail/TransactionDetailSheets.tsx
matejkriz: I see, we don't have a constant for 12 🙈

suite-native/transactions/src/utils.ts
```diff
@@ -42,3 +44,17 @@ export const sortTargetAddressesToBeginning = (
         A.intersection(addresses, targetAddresses),
         A.difference(addresses, targetAddresses),
     ) as VinVoutAddress[];
+
+export const getTransactionValueSign = (transactionType: TransactionType) => {
+    const signValueMap = {
```
matejkriz: nitpick, this constant can be define outside the function to prevent redefinition on every function call...

matejkriz (commented):
I see mostly nitpicks, the screen looks really nice now 🎉

Only real thing is if it is true that there is some waiting once you click on the Send button that has no loading indicator.

suite-native/transactions/src/components/TransactionDetail/TransactionDetailRow.tsx
PeKne: It was redundant there, removed in 59369acb4d

suite-native/transactions/src/components/TransactionDetail/TransactionDetailSheets.tsx
PeKne: Yeah, there is no 12 spacing const. But IMO we should add it. Because most of the new designs are based on values 8, 12 and 16. So we are hardcoding the value 12 too often. Let's discuss with other team devs on Slack.

suite-native/module-send/src/components/OutputsReviewFooter.tsx
PeKne: What do you mean by "stuck"? Did it just take some time to redirect or did the navigation not happen at all?

suite-native/transactions/src/components/TransactionDetail/NetworkTransactionDetailSummary.tsx
PeKne: For the separator is important how many inputs transaction has, because it is position on the left side of the inputs.

But you are right, the check of `> 1` is better. 59369acb4d

suite-native/transactions/src/components/TransactionDetail/NetworkTransactionDetailSummary.tsx
PeKne: There are only two options defined by design:

A) Single input/output:
Screenshot 2024-09-04 at 10 40 38

B) Multiple inputs/outputs:
Screenshot 2024-09-04 at 10 34 23


For this reason I believe that no dynamic height checking is needed. I at least improved the naming and put the static values to `const`s.  59369acb4d

suite-native/transactions/src/utils.ts
PeKne: fixed: 59369acb4d

suite-native/module-send/src/components/OutputsReviewFooter.tsx
matejkriz: It just take some time so it feels like it is stucked.

matejkriz: Lets create a separate issue for it

suite-native/transactions/src/components/TransactionDetail/NetworkTransactionDetailSummary.tsx
matejkriz: just check == 2, please

suite-native/transactions/src/components/TransactionDetail/NetworkTransactionDetailSummary.tsx
matejkriz: I get it now

suite-native/transactions/src/components/TransactionDetail/NetworkTransactionDetailSummary.tsx
PeKne: fixed by d57a5ac5f3

matejkriz: inputsCount is also nicer name ;-)

matejkriz (approved):
Nice improvement 👍

suite-native/intl/src/en.ts
```diff
@@ -688,6 +689,18 @@ export const en = {
             subtitle: 'Get started by receiving coins',
             button: 'Receive',
         },
+        detail: {
+            header: '<transactionType></transactionType> transaction',
```
vytick, line 693: fyi: Recently i realised, that you can also use {}
```suggestion
            header: '{transactionType} transaction',
```

and later in code:

``` values={{ transactionType: transaction.type }}```

PR #14072, opened 2024-09-02
feat(suite-native): staking balance

### LINKED ISSUE #12187: Reflect balances & staking in the app
Since the ETH staking has been introduced in Trezor Suite, we need to reflect it in TSL, in order not to scare users that they've lost their ETH but indicate that it's been staked.
We should indicated that:
- ETH funds are staked
- Amount of staked ETH
- Rewards

Total balance:
- Staked ETH is still reflected in total amount (= ETH + staked ETH + rewards)

Inform user that in case they want to Unstake they have to do it Trezor Suite web or desktop

[Figma design](https://www.figma.com/design/ylfHUP23P94yZPcVTiPTZ6/ETH-Staking?node-id=255-25893&t=puFlKGLmJswhbHwv-11)

dev-pvl:
## Description
Added the staking balance display to the Account modal. Updated UI to include the new staking balance information.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12187
## Screenshots:

Simulator Screenshot - iPhone 15 Pro Max - 2024-09-10 at 16 28 18
Simulator Screenshot - iPhone 15 Pro Max - 2024-09-10 at 16 31 03
Simulator Screenshot - iPhone 15 Pro Max - 2024-09-10 at 16 34 11
Simulator Screenshot - iPhone 15 Pro Max - 2024-09-10 at 16 37 00


suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -604,6 +604,18 @@ export const getStakingFiatBalance = (account: Account, rate: number | undefined
     return toFiatCurrency(balanceInEther, rate, 2);
 };
 
+export const getAccountCryptoBalanceWithStakingFormatted = (account: Account) => {
```
Nodonisko: Please use `calculateBalance` instead of these functions and <CryptoAmountFormatter /> to format it.

suite-native/assets/src/components/AssetItem.tsx
```diff
@@ -72,6 +75,17 @@ export const AssetItem = memo(
             ) => selectNumberOfUniqueTokensForCoinPerDevice(state, cryptoCurrencySymbol),
         );
 
+        const accountsWithStaking = useSelector(
```
Nodonisko: This will break perf of this component and will rerender with every change to any account which is happening all the time. Please memoize this selector or better just return that boolean we need here directly from selector.

suite-native/assets/src/components/AssetItem.tsx
```diff
@@ -72,6 +75,17 @@ export const AssetItem = memo(
...
+            (
+                state: AccountsRootState &
+                    DeviceRootState &
+                    FiatRatesRootState &
+                    SettingsSliceRootState,
+            ) => selectVisibleDeviceAccountsWithStakingByNetworkSymbol(state, cryptoCurrencySymbol),
+        );
+
+        const isStaking = accountsWithStaking.length > 0;
```
Nodonisko: Name should be something like `hasAnyAccountWithStaking`

suite-native/intl/src/en.ts
```diff
@@ -92,6 +92,8 @@ export const en = {
     accountList: {
         numberOfTokens: '+{numberOfTokens} tokens',
         tokens: 'Tokens',
+        staking: 'Staking',
+        hasStaking: '+Staking',
```
Nodonisko: nit: I prefer something like `stakingBadge`

suite-native/assets/src/assetsSelectors.ts
```diff
@@ -69,10 +73,12 @@ export const selectDeviceAssetsWithBalances = (state: AssetsRootState & DeviceRo
             shouldIncludeStaking: true,
         });
 
+        const cryptoBalanceWithStaking = getAccountCryptoBalanceWithStakingFormatted(account);
```
Nodonisko: `getAccountFiatBalance` few lines before should be capable including staking balance and I think it's probably happening by accident because there is `shouldIncludeStaking: true`

suite-native/accounts/src/components/StakingItem.tsx
```diff
@@ -0,0 +1,61 @@
...
+    alignItems: 'center',
+    paddingVertical: 12,
+    paddingHorizontal: utils.spacings.medium,
+}));
+
+type StakingItemProps = {
+    account: Account;
+};
+
+export const StakingItem = ({ account }: StakingItemProps) => {
```
Nodonisko: This component should use `AccountListItemBase` or this component shoudn't exist at all and `AccountListItemBase` could be used directly

suite-common/icons/assets/icons/spinnerGap.svg
dev-pvl, line 1: Should this icon be animated? And if so, can I use any existing component from the project for this purpose?

suite-native/intl/src/en.ts
dev-pvl: already removed. There is a badge with no text instead.

suite-native/accounts/src/components/StakingItem.tsx
dev-pvl: done. This component has been deleted. https://github.com/trezor/trezor-suite/pull/14072/commits/565eadf4d973e9eded8dc5caa0e6da4d7dc37733

suite-native/assets/src/components/AssetItem.tsx
dev-pvl: added a separate selector for this. https://github.com/trezor/trezor-suite/pull/14072/commits/565eadf4d973e9eded8dc5caa0e6da4d7dc37733

suite-native/assets/src/components/AssetItem.tsx
dev-pvl: https://github.com/trezor/trezor-suite/pull/14072/commits/565eadf4d973e9eded8dc5caa0e6da4d7dc37733

suite-common/wallet-utils/src/accountUtils.ts
dev-pvl: https://github.com/trezor/trezor-suite/pull/14072/commits/565eadf4d973e9eded8dc5caa0e6da4d7dc37733

suite-native/assets/src/assetsSelectors.ts
dev-pvl: I probably got it wrong, but getAccountFiatBalance calculated correctly. The issue occurs for the asset's crypto balance (account.formattedBalance). It's calculated without staking, as I see.

Nodonisko:
Thanks for PR, but we had to do some bigger refactor in the end so we implemented it ourselves. Some code of this PR was moved to #14561

PR #14073, opened 2024-09-02
Chore/add list missing to ci

komret:
## Description

Some improvements to the `list-missing-dependencies` script and the domain lists.
- The check is now run automatically in CI as part of the "Other checks" job.
- It now uses `yarn info` instead of `yarn outdated` so it checks all direct dependencies, not just outdated ones.
- It also checks for package names that are listed redundantly under the domains (in case the dependency was removed, but forgotten in a list).
- It allows for dependency categorisation inside the domain lists because lines starting with `#` are skipped.
- Also some cleanup in the lists resulting from the improved checks.

## Screenshots:
Screenshot 2024-09-18 at 12 30 58


scripts/ci/list-missing-dependencies.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+# List dependencies missing from the domain lists.
+
+domains=("common" "connect" "mission" "mobile" "qa" "trends" "usability")
+
+dependencies_in_lists=()
```
Lemonexe, line 6: Nit: maybe I'd name it `dependencies_in_all_domain_lists`, to make it really clear. But it's not necessary.

Lemonexe (approved):
:ok_hand: looks perfect!

It's great that the check now works both ways. I tested it locally.
Looks like you solved the issue with string replacements.

scripts/ci/list-missing-dependencies.sh
```diff
@@ -0,0 +1,84 @@
...
+    fi
+
+    if [ ${#dependencies_listed_but_unused[@]} -ne 0 ]; then
+        echo -e "\nThere are dependencies in the domain lists that are not installed:"
+        printf '%s\n' "${dependencies_listed_but_unused[@]}"
+        echo "Remove them from the respective .txt files in $domain_lists_directory."
+    fi
+
+    exit 1  # Exit with a non-zero status code to indicate failure
+fi
```
vdovhanych, line 84: super nit-picky but there is no newline at the end

scripts/ci/list-missing-dependencies.sh
```diff
@@ -0,0 +1,84 @@
...
+    if [[ "$found" == false ]]; then
+        dependencies_listed_but_unused+=("$listed_package")
+    fi
+done
+
+# Output the results
+if [ ${#dependencies_missing_from_lists[@]} -eq 0 ] && [ ${#dependencies_listed_but_unused[@]} -eq 0 ]; then
+    echo "All dependencies are correctly listed under their domains."
+else
+    domain_lists_directory="scripts/list-outdated-dependencies"
```
vdovhanych, line 70: How are we updating this file? Will it be run manually, and will you create a PR with updated lists?

vdovhanych (approved):

komret:
Sorry I merged without noticing the comments :(

scripts/ci/list-missing-dependencies.sh
komret: Yes, this has to be done manually. The files are up to date now and they will be updated whenever a dependency is added/removed - in the same PR.

PR #14076, opened 2024-09-02
fix: storage middleware fiat rates error

Nodonisko:
## Description

Simple fix but it's concerning that TS didn't complain and it seems that it didn't know return type of `updateTxsFiatRatesThunk.fulfilled` sadly I don't have time to investigate it now.

Also it seems that this code may be not needed anyway because of recent fiat rates refactor, maybe @AdamSchinzel should investigate it.

## Related Issue

Resolve


marekrjpolak (approved):
Not sure whether `updateTxsFiatRatesThunk` should ever return undefined account on fulfill, but as a quick fix good enough.

PR #14080, opened 2024-09-02
feat: improve token definitions perf

Nodonisko:
## Description

This PR greatly improves performance of checking token definitions. There is something about ~100x speedup. For big accounts with few hundreds token this result to about `70ms -> 1.5ms`, for really huge account with 2k tokens it was about `8000ms -> 70ms`

For nerds: `Array.includes` is `O(n)`, but `Set.has` is `O(1)`

## Related Issue

Resolve

### DIFF
--- a/suite-common/token-definitions/src/tokenDefinitionsUtils.ts
+++ b/suite-common/token-definitions/src/tokenDefinitionsUtils.ts
@@ -14,6 +14,9 @@ export const getContractAddressForNetwork = (
     contractAddress: string,
 ) => {
     switch (networkSymbol) {
+        case 'eth':
+            // Specyfing most common network as first case improves performance little bit
+            return contractAddress.toLowerCase();
         case 'sol':
         case 'dsol':
             return contractAddress;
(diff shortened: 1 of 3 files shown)


suite-common/token-definitions/src/tokenDefinitionsUtils.ts
```diff
@@ -14,6 +14,9 @@ export const getContractAddressForNetwork = (
     contractAddress: string,
 ) => {
     switch (networkSymbol) {
+        case 'eth':
```
tomasklim, line 17: note: would be nice to change this method to network type instead of network symbol, that would speed it even more

tomasklim (approved):
LGTM,  not tested

suite-common/token-definitions/src/tokenDefinitionsUtils.ts
Nodonisko: FYI Impact of this my change is pretty small, maybe like 3% and it was hardly measurable (it more my obsession than anything)

matejkriz (approved):

PR #14081, opened 2024-09-02
feat(mobile): improve Account import flow performance

Nodonisko:
## Description
This needs to be merged first #14080 

This PR greatly improves perf Account import flow on mobile. Tested it with ETH account that has few hundreds tokens and it works mostly smoothly compared to few seconds freeze or forever freeze on low end devices.

## Related Issue

Resolve


matejkriz (approved):
It works fine for me 👍

PR #14089, opened 2024-09-03
feat(components): add new set of icons

seibei-iguchi:


packages/components/src/components/Icon/Icon.stories.tsx
```diff
@@ -21,7 +34,7 @@ export const Icon: StoryObj<IconProps> = {
     },
     argTypes: {
         name: {
-            options: iconNames,
+            options: [...allIcons],
```
jvaclavik, line 37: I think `options: allIcons` should be enough

packages/components/src/components/Icon/icons.stories.tsx
```diff
@@ -54,6 +58,13 @@ const Render = (props: IconProps) => {
     const [search, setSearch] = useState('');
     const [copied, setCopied] = useState<string | null>(null);
     const theme = useTheme();
+
+    const iconNames = Object.keys(icons) as IconName[];
+    const iconsDeprecatedNames = Object.keys(iconsDeprecated) as IconNameDeprecated[];
```
jvaclavik: Could we differentiate label for deprecated icons? Like add suffix `(legacy)` or similar. And maybe we could also add radio button to controls with "show all icons / show deprecated icons / show all". WDYT?

packages/components/src/components/Icon/Icon.tsx
```diff
@@ -28,6 +27,20 @@ type ExclusiveColorOrVariant =
 export const allowedIconFrameProps: FramePropsKeys[] = ['margin', 'pointerEvents'];
 type AllowedFrameProps = Pick<FrameProps, (typeof allowedIconFrameProps)[number]>;
 
+export const iconSizes = {
```
jvaclavik, line 38: don't we want to load this from shared package? (I'm not sure)

packages/components/src/components/Icon/icons.stories.tsx
```diff
@@ -71,18 +82,18 @@ const Render = (props: IconProps) => {
                 showClearButton="always"
             />
             <Wrapper>
-                {Object.keys(icons)
+                {[...allIcons]
```
jvaclavik: why not just `allIcons.filter(...)`?

suite-common/icons/generateIcons.js
```diff
@@ -53,6 +43,8 @@ const optimizeSvgAssets = assetsDirname => {
     const assetsDir = path.join(__dirname, assetsDirname);
     const assetFileNames = fs.readdirSync(assetsDir);
 
+    console.log(assetFileNames);
```
jvaclavik: do you need this?

packages/components/src/components/Icon/icons.stories.tsx
seibei-iguchi: Sets don't have filter:)

packages/components/src/components/Icon/Icon.tsx
seibei-iguchi: I don't think so. The shared package contains only the files. I think the sizes should reside with the component, similar to other components.

packages/components/src/components/Icon/icons.stories.tsx
seibei-iguchi: If we add a suffix, we would have to add it to all the currently used icons and we would lose the benefit that some of the names are same and those switched to the new icons automatically.

packages/components/src/components/Icon/icons.stories.tsx
jvaclavik: I see!

jvaclavik (approved):
Suite desktop: LGTM!

Good job, let's wait for approval from Suite mobile.

vdovhanych (approved):
Ok for the gh actions part.

PR #14095, opened 2024-09-03
fix(suite): Button minWidth

Lemonexe:
## Related Issue

Resolve 

## Screenshots:

These `Buttons` in `AccountEmpty` should have 160 min width:
image

React error currently thrown:

image

### DIFF
--- a/packages/components/src/components/buttons/Button/Button.tsx
+++ b/packages/components/src/components/buttons/Button/Button.tsx
@@ -16,7 +16,7 @@ import { makePropsTransient, TransientProps } from '../../../utils/transientProp
 import { FrameProps, FramePropsKeys, withFrameProps } from '../../../utils/frameProps';
 import { Icon, IconName } from '../../Icon/Icon';
 
-export const allowedButtonFrameProps: FramePropsKeys[] = ['margin'];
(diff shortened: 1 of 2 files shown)


packages/components/src/components/buttons/Button/Button.tsx
```diff
@@ -16,7 +16,7 @@ import { makePropsTransient, TransientProps } from '../../../utils/transientProp
 import { FrameProps, FramePropsKeys, withFrameProps } from '../../../utils/frameProps';
 import { Icon, IconName } from '../../Icon/Icon';
 
-export const allowedButtonFrameProps: FramePropsKeys[] = ['margin'];
+export const allowedButtonFrameProps = ['margin', 'minWidth'] as const satisfies FramePropsKeys[];
```
Lemonexe, line 19: This needs to be inferred type, not declarative `FramePropsKeys[]`.
Otherwise, `(typeof allowedButtonFrameProps)[number] = FramePropsKeys` 
and  `Pick<FrameProps, (typeof allowedButtonFrameProps)[number]> = FrameProps`

But we can use `satisfies FramePropsKeys[]` to type check that it doesn't contain nonsense.

PR #14100, opened 2024-09-03
fix(transport): improve types

mroz22:
couple of improvements in types and error handling in @trezor/transport package. should be discussed with @szymonlesisz 

For example getting rid of `{ success: false, error: string }` 
image

Discussion points:
- top-level members of transport API (call, send, receive) should always catch potential errors and return { success: false, error: 'unexpected-error'}. for transport classes (webusb/bridge/etc) it is defined here https://github.com/trezor/trezor-suite/blob/develop/packages/transport/src/transports/abstract.ts#L508
- node-bridge: [final catch should be this one](https://github.com/trezor/trezor-suite/blob/develop/packages/transport/src/api/abstract.ts#L190) and calls down the call stack such as readUtil and writeUtil now don't need to catch errors themselves. Also couple of throws were replaced with { success: false } for better type clarity.

### DIFF
--- a/packages/transport/src/transports/abstract.ts
+++ b/packages/transport/src/transports/abstract.ts
@@ -74,6 +74,21 @@ export const isTransportInstance = (transport?: AbstractTransport) => {
 
 const getKey = ({ path, product }: Descriptor) => `${path}${product}`;
 
+type ReadWriteError =
+    | typeof ERRORS.HTTP_ERROR
+    | typeof ERRORS.WRONG_RESULT_TYPE
+    | typeof ERRORS.OTHER_CALL_IN_PROGRESS
+    | typeof PROTOCOL_MALFORMED
+    | typeof ERRORS.DEVICE_DISCONNECTED_DURING_ACTION
+    | typeof ERRORS.UNEXPECTED_ERROR
+    | typeof ERRORS.SESSION_NOT_FOUND
+    | typeof ERRORS.ABORTED_BY_TIMEOUT
+    | typeof ERRORS.ABORTED_BY_SIGNAL
+    | typeof ERRORS.WRONG_ENVIRONMENT
+    | typeof ERRORS.DEVICE_NOT_FOUND
+    | typeof ERRORS.INTERFACE_UNABLE_TO_OPEN_DEVICE
+    | typeof ERRORS.INTERFACE_DATA_TRANSFER;
+
 export abstract class AbstractTransport extends TypedEmitter<{
     [TRANSPORT.UPDATE]: DeviceDescriptorDiff;
     [TRANSPORT.ERROR]:
@@ -268,21 +283,7 @@ export abstract class AbstractTransport extends TypedEmitter<{
         name: string;
         data: Record<string, unknown>;
         protocol?: TransportProtocol;
-    }): AbortableCall<
-        undefined,
-        | typeof ERRORS.DEVICE_DISCONNECTED_DURING_ACTION
-        // bridge
-        | typeof ERRORS.HTTP_ERROR
-        | typeof ERRORS.WRONG_RESULT_TYPE
-        | typeof ERRORS.OTHER_CALL_IN_PROGRESS
-        // webusb + bridge
-        | typeof PROTOCOL_MALFORMED
-        | typeof ERRORS.UNEXPECTED_ERROR
-        | typeof ERRORS.SESSION_NOT_FOUND
-        | typeof ERRORS.ABORTED_BY_TIMEOUT
-        | typeof ERRORS.ABORTED_BY_SIGNAL
-        | typeof ERRORS.WRONG_ENVIRONMENT
-    >;
+    }): AbortableCall<undefined, ReadWriteError>;
 
     /**
      * Only read from transport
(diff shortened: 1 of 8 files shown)


packages/protocol/src/protocol-v1/encode.ts
```diff
@@ -11,6 +11,7 @@ export const getChunkHeader = (_data: Buffer) => {
 export const encode: TransportProtocolEncode = (data, options) => {
     const { messageType } = options;
     if (typeof messageType === 'string') {
+        // todo: this may throw, should we start returning { success: bool } too?
```
mroz22: todo: discuss

packages/transport/src/transports/abstractApi.ts
```diff
@@ -174,20 +174,26 @@ export abstract class AbstractApiTransport extends AbstractTransport {
...
-                    session,
-                });
-                if (!getPathBySessionResponse.success) {
-                    // session not found means that device was disconnected
-                    if (getPathBySessionResponse.error === 'session not found') {
-                        return this.error({ error: ERRORS.DEVICE_DISCONNECTED_DURING_ACTION });
+                const handleError = (error: string) => {
+                    // if user revokes usb permissions in browser we need a way how propagate that the device was technically disconnected,
+                    if (error === ERRORS.DEVICE_DISCONNECTED_DURING_ACTION) {
+                        this.enumerate();
```
mroz22, line 180: note: this is not new, just moved. but it should be moved to a higher layer - connect

packages/transport/src/utils/receive.ts
```diff
@@ -3,29 +3,45 @@ import { Root } from 'protobufjs/light';
...
-    receiver: () => Promise<Buffer>,
+    receiver: T,
     offset: number,
     expectedLength: number,
     chunkHeader: Buffer,
-): Promise<void> {
+) {
     if (offset >= expectedLength) {
-        return;
+        return error({ error: ERRORS.UNEXPECTED_ERROR, message: 'offset >= expected length' });
```
mroz22: this is somewhere in between unexpected and expected error. it really shouldn't not happen, its just a sanity check. so I am not sure whether it should be strongly typed or not.

packages/transport/src/transports/abstract.ts
```diff
@@ -70,6 +70,23 @@ export const isTransportInstance = (transport?: AbstractTransport) => {
     return false;
 };
 
+type ReadCallSendError =
+    | typeof ERRORS.HTTP_ERROR
+    | typeof ERRORS.WRONG_RESULT_TYPE
+    | typeof ERRORS.OTHER_CALL_IN_PROGRESS
+    // webusb + bridge
```
mroz22: rm

packages/transport/src/transports/abstract.ts
```diff
@@ -70,6 +70,23 @@ export const isTransportInstance = (transport?: AbstractTransport) => {
     return false;
 };
 
+type ReadCallSendError =
```
mroz22: ReadWriteError

packages/transport/src/utils/receive.ts
mroz22: unknown

mroz22: or maybe let it throw

packages/transport/src/transports/abstractApi.ts
```diff
@@ -187,51 +193,40 @@ export abstract class AbstractApiTransport extends AbstractTransport {
...
-                    });
-                    const chunks = createChunks(
-                        bytes,
-                        protocol.getChunkHeader(bytes),
-                        this.api.chunkSize,
-                    );
-                    const apiWrite = (chunk: Buffer) => this.api.write(path, chunk, signal);
-                    const sendResult = await sendChunks(chunks, apiWrite);
-                    if (!sendResult.success) {
-                        throw new Error(sendResult.error);
```
mroz22, line 206: this is the main difference, I already had typed error, but used throw statement here, which leads to loss of type information which is again band-aided on line 228.

packages/transport/src/transports/abstractApi.ts
```diff
@@ -187,51 +193,40 @@ export abstract class AbstractApiTransport extends AbstractTransport {
                 }
                 const { path } = getPathBySessionResponse.payload;
 
-                try {
```
mroz22, line 190: try catch is not needed here, the function body is wrapped into custom schedule action which has its own catch handler.

mroz22:
rebased, and added a fixup. found even more room for simplification [2540b95](https://github.com/trezor/trezor-suite/pull/14100/commits/2540b95475fa8a529a542edda670a2a1a05ca6f3)

packages/transport/src/transports/abstract.ts
```diff
@@ -74,6 +74,21 @@ export const isTransportInstance = (transport?: AbstractTransport) => {
 
 const getKey = ({ path, product }: Descriptor) => `${path}${product}`;
 
+type ReadWritError =
```
marekrjpolak: nit: typo

packages/transport/src/utils/receive.ts
```diff
@@ -3,29 +3,44 @@ import { Root } from 'protobufjs/light';
...
     chunkHeader: Buffer,
-): Promise<void> {
+) {
     if (offset >= expectedLength) {
         return;
     }
-    const data = await receiver();
+    const readResult = await receiver();
+    if (!readResult.success) {
+        return readResult;
```
marekrjpolak: If I understand correctly, returning unsuccessful result here is irrelevant, because `receiveRest` is used as a procedure, not a function. Therefore, even if one `receiveRest` iteration fails (without throwing an error unlike before), `receive` will still return success with half-empty buffer.

Also, I'd like to convert the recursion into a cycle in the future, but no reason to do it right now.

marekrjpolak (commented):
Seems logical, apart from one nit and one comment.

packages/transport/src/utils/receive.ts
```diff
@@ -42,15 +16,25 @@ export async function receive<T extends () => ReturnType<AbstractApi['read']>>(
...
-    const result = Buffer.alloc(length);
+    let result = Buffer.alloc(length);
     const chunkHeader = protocol.getChunkHeader(Buffer.from(data));
 
-    if (length) {
-        payload.copy(result);
-    }
+    let offset = 0;
+    payload.copy(result);
+    offset = payload.length;
```
mroz22: yes I know

marekrjpolak (approved):
Looks good, provided that I haven't missed something from the newer fixups.
:seal:

PR #14105, opened 2024-09-03
feat(suite): Add anchor to Button if it's needed

jvaclavik:
## Description

This allows us to prevent components where `button` is inside of `a` like this one:
image

## Related Issue

Resolve

### DIFF
--- a/packages/components/src/components/buttons/Button/Button.tsx
+++ b/packages/components/src/components/buttons/Button/Button.tsx
@@ -26,6 +26,7 @@ type ButtonContainerProps = TransientProps<AllowedFrameProps> & {
     $hasIcon?: boolean;
     $isFullWidth?: boolean;
     $isSubtle: boolean;
+    as?: 'a' | 'button';
     $borderRadius?: typeof borders.radii.sm | typeof borders.radii.full; // Do not allow all, we want consistency
 };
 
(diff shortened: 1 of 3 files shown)


packages/components/src/components/buttons/Button/Button.tsx
```diff
@@ -26,6 +27,7 @@ type ButtonContainerProps = TransientProps<AllowedFrameProps> & {
     $hasIcon?: boolean;
     $isFullWidth?: boolean;
     $isSubtle: boolean;
+    as?: 'a' | 'button';
```
seibei-iguchi, line 29: What's the reason for this? Isn't it already decided by the presence of a href prop?

jvaclavik: We need to change tag to `a` when we want to use `href`. it's not working on `button` AFAIK

seibei-iguchi: Yeah but I don't see you use it anywhere? Shouldn't the styled ButtonContainer change to something else than button?

seibei-iguchi (approved):

packages/components/src/components/buttons/Button/Button.tsx
jvaclavik: Solved in DM. it's native feature of styled components

Issue #13819, opened 2024-08-15
Coin Enabling - Flow polishing


vytick:
- [x] Trigger init screen from any location
- [x] prevent dismissing init screen by hw back while no network is selected
- [x] enable coin enabling for develop and debug
- [x] Make sure showing the UI works correctly with biometric, pin, connecting screen, etc. 
- [x] preselect view only networks when user enters settings Coin enabling before initial setup takes place

STew790:
**QA OK**
All mentioned points are OK.

**Info**
24.9.2 7a4ddc9

PR #14106, opened 2024-09-03
fix(suite-native): move coin enabling init check to navigator

vytick:
- [x] Trigger init screen from any location
- [x] prevent dismissing init screen by hw back while no network is selected
- [x] enable coin enabling for develop and debug
- [x] improve checks for showing init UI 

## Related Issue

Part of #13819


matejkriz (changes requested):
Not working for me, it redirects out of the initial coin enabling screen after selection of something without me clicking on Confirm button.

matejkriz (approved):
Now it works smoothly! 👏

PR #14109, opened 2024-09-04
Chore(suite): Use new Networks in AddAccountModal

Lemonexe:
Part 1 of refactoring usages of deprecated `networksCompatibility` to `networks`,
including corresponding types & utility functions such as `getMainnets`, `getTestnets`.

## Description

The usages are very interconnected, so refactoring one usage creates incompatibility and the refactoring propagates, unless I make temporary compatibility "fix" at the interfaces.

:eyes:  For easier orientation in CR, this was the "path" I took (not listing the simple cases, only where I had to refactor logic):

- `AddAccountModal`
  - `useEnabledNetworks` and all its usages, all simple
  - `AccountTypeSelect`
  - `AddAccountButton`
    - `AddCoinjoinAccountButton`

## Related Issue

Part of #13839

## Screenshots

All of the steps of Add Account modal work as before:

activate new network
existing network selection
account type select
add account

### DIFF
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
@@ -7,11 +7,7 @@ import {
     selectDeviceModel,
 } from '@suite-common/wallet-core';
 import { arrayPartition } from '@trezor/utils';
-import {
-    networksCompatibility,
-    NetworkCompatible,
-    NetworkSymbol,
-} from '@suite-common/wallet-config';
+import { networks, Network, NetworkSymbol, NetworkAccount } from '@suite-common/wallet-config';
 import { CollapsibleBox, NewModal } from '@trezor/components';
 import { FirmwareType } from '@trezor/connect';
 import { spacings, spacingsPx } from '@trezor/theme';
@@ -53,8 +49,8 @@ export const AddAccountModal = ({ device, onCancel, symbol, noRedirect }: AddAcc
 
     const { mainnets, testnets, enabledNetworks: enabledNetworkSymbols } = useEnabledNetworks();
 
-    const getNetworks = (networks: NetworkCompatible[], getUnsupported = false) =>
-        networks.filter(
+    const getNetworks = (networksToFilterFrom: Network[], getUnsupported = false) =>
+        networksToFilterFrom.filter(
             ({ symbol }) => getUnsupported !== supportedNetworkSymbols.includes(symbol),
         );
 
@@ -70,17 +66,17 @@ export const AddAccountModal = ({ device, onCancel, symbol, noRedirect }: AddAcc
     const networkPinned = !!symbol;
     const preselectedNetwork = symbol && supportedNetworks.find(n => n.symbol === symbol);
     // or in case of only btc is enabled on bitcoin-only firmware
-    const bitcoinNetwork = networksCompatibility.find(n => n.symbol === 'btc');
     const bitcoinOnlyDefaultNetworkSelection =
         device.firmwareType === FirmwareType.BitcoinOnly &&
         supportedMainnets.length === 1 &&
         allTestnetNetworksDisabled
-            ? bitcoinNetwork
+            ? networks.btc
             : undefined;
 
-    const [selectedNetwork, setSelectedNetwork] = useState<NetworkCompatible | undefined>(
+    const [selectedNetwork, setSelectedNetwork] = useState<Network | undefined>(
         preselectedNetwork || bitcoinOnlyDefaultNetworkSelection,
     );
+    const [selectedAccount, setSelectedAccount] = useState<NetworkAccount | undefined>(undefined);
 
     const isSelectedNetworkEnabled =
         !!selectedNetwork && enabledNetworkSymbols.includes(selectedNetwork.symbol);
@@ -93,7 +89,7 @@ export const AddAccountModal = ({ device, onCancel, symbol, noRedirect }: AddAcc
     );
 
     // Collect all empty accounts related to selected device and selected accountType
-    const currentType = selectedNetwork?.accountType ?? 'normal';
+    const currentType = selectedAccount?.accountType ?? 'normal';
     const emptyAccounts = selectedNetwork
         ? accounts.filter(
               a =>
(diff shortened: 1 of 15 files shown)


suite-common/wallet-config/src/types.ts
```diff
@@ -97,3 +99,5 @@ export type Network = {
 export type Networks = {
     [key in NetworkSymbol]: Network;
 };
+
+export type NetworkAccountCollection = Array<Account & { accountType: AccountType }>;
```
Lemonexe: There are many usecases where I need to iterate through accounts, but they are a dict object, which is sooo cumbersome. Array is much more practical, so I've added a function in `utils.ts` to convert it.

Here I've purposefully only added the array type, and not a type with the modified Account - IMHO it'd be quite confusing to have two interfaces for Account - how to name properly the "Account that is used in collection"?
This way it is `NetworkAccountCollection[number]` which is more words, but clear. But I still don't like it :disappointed: 

:point_right:  **Alternatively**, I could just always specify `accountType`, so it would duplicate the key:

```
segwit: {
  accountType: 'segwit',
  bip43Path: "m/49'/0'/i'",
},
```
**And** use a generic type, which would enforce that the `key` is the same as `accountType`, so the duplication would be safe. What do you think? :thinking:

suite-common/wallet-config/src/types.ts
```diff
@@ -54,6 +54,8 @@ export type NetworkFeature =
     | 'coin-definitions'
     | 'nft-definitions';
 
+export type Bip43Path = string; // TODO define a more specific type
```
Lemonexe, line 57: Placeholder for a possible future refactoring - compose a template literal type (not a taxative string union).
When we relied on inferred types, `NetworkCompatible['bip43Path']` was used, this replaces it so that it can be later refactored en-masse (would be difficult if I placed `string` everywhere).

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
```diff
@@ -108,25 +113,29 @@ export const AddAccountModal = ({ device, onCancel, symbol, noRedirect }: AddAcc
...
-                      isDebugOnlyAccountType === undefined || (isDebugOnlyAccountType && isDebug),
-              )
+              A.filter(({ backendType }) => backendType !== 'coinjoin' || isCoinjoinVisible),
+              A.filter(({ isDebugOnlyAccountType }) => !isDebugOnlyAccountType || isDebug),
+              // the default account is expected to be the first one
+              A.prepend({
+                  bip43Path: selectedNetwork.bip43Path,
+                  accountType: 'normal',
+              } as NetworkAccountCollection[number]),
+          ) as NetworkAccountCollection)
```
Lemonexe: This is a good usecase for functional programming, but unfortunately `mobily A` is poorly typed. It converts all return types to `readonly`, so I have to use `as` to make it mutable again, otherwise  types are incompatible :facepalm: 

I'm kind of disappointed with `mobily` compared to `ramda` or `fp-ts`, and maybe I should just use vanillaJS here. What do you think? :thinking:

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
```diff
@@ -70,17 +73,19 @@ export const AddAccountModal = ({ device, onCancel, symbol, noRedirect }: AddAcc
...
+            ? networks.btc
             : undefined;
 
-    const [selectedNetwork, setSelectedNetwork] = useState<NetworkCompatible | undefined>(
+    const [selectedNetwork, setSelectedNetwork] = useState<Network | undefined>(
         preselectedNetwork || bitcoinOnlyDefaultNetworkSelection,
     );
+    const [selectedAccount, setSelectedAccount] = useState<
+        NetworkAccountCollection[number] | undefined
+    >(undefined);
```
Lemonexe: What is going on here, you may ask?
The UI lets you first choose network, then its account.
Previously, the `selectedNetworks` state served for both selections, IMHO quite obscure codewise.
(in coin selection, you select a network, but only from the mainnet/testnet ones, then in account selection, you select a _different network object_ from the same set) 

Now, each UI selection has its own state, which is better readable I think.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
```diff
@@ -108,25 +113,29 @@ export const AddAccountModal = ({ device, onCancel, symbol, noRedirect }: AddAcc
...
+              accountTypesToArray,
               /**
                * Filter out coinjoin account type if it is not visible.
                * Visibility of coinjoin account type depends on coinjoin feature config in message system.
                * By default it is visible publicly, but it can be remotely hidden under debug menu.
                */
-              .filter(({ backendType }) => backendType !== 'coinjoin' || isCoinjoinVisible)
-              .filter(
-                  ({ isDebugOnlyAccountType }) =>
-                      isDebugOnlyAccountType === undefined || (isDebugOnlyAccountType && isDebug),
```
Lemonexe, line 121: This logic was redundant, I simplified it.

Lemonexe:
Note: `[Test] suite-native Android E2E / run_android_e2e_tests` is currently degraded, see #14117

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
```diff
@@ -70,17 +73,19 @@ export const AddAccountModal = ({ device, onCancel, symbol, noRedirect }: AddAcc
     const networkPinned = !!symbol;
     const preselectedNetwork = symbol && supportedNetworks.find(n => n.symbol === symbol);
     // or in case of only btc is enabled on bitcoin-only firmware
-    const bitcoinNetwork = networksCompatibility.find(n => n.symbol === 'btc');
     const bitcoinOnlyDefaultNetworkSelection =
         device.firmwareType === FirmwareType.BitcoinOnly &&
         supportedMainnets.length === 1 &&
         allTestnetNetworksDisabled
-            ? bitcoinNetwork
+            ? networks.btc
```
peter-sanderson, line 73: This is awesome! Finally direct access for networks!

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
peter-sanderson: I am not that much fun those libraries, I think in most cases vanilla JS is just fine.

peter-sanderson: Not sure if this is applicable here but: 

In case you need to filer array of some types into a subtype, I suggest to simply use `reduce` instead of filter. It will return reduced type as well.

```ts
const arr: A|B = ["a","b"]


const: A = arr.reduce((acc, it) => {
    if (it === "a") {
        acc.push(it)
    }

    return acc
}, [] as A[])

```

suite-common/wallet-config/src/types.ts
peter-sanderson: Lets duplicate the `key` (`accountType`) inside original dictionary, this would pollute just one place and not leak into whole codebase

packages/suite/src/hooks/settings/useEnabledNetworks.ts
```diff
@@ -1,21 +1,56 @@
...
+// TODO refactor all usages of this hook to use the new one
+type EnabledNetworksCompatible = {
     mainnets: NetworkCompatible[];
     testnets: NetworkCompatible[];
     enabledNetworks: NetworkCompatible['symbol'][];
     setEnabled: (symbol: NetworkCompatible['symbol'], enabled: boolean) => void;
 };
 
-export const useEnabledNetworks = (): EnabledNetworks => {
+export const useEnabledNetworksCompatible = (): EnabledNetworksCompatible => {
```
peter-sanderson: it may be useful to add 

```ts
/**
 * @deprecated This is deprecated in favor of SomethingElse, dont use it!
  */
const deprecatedStuff = ...
```

So it will be ~striked~ in IDEs

Lemonexe: I will get rid of it completely. At first glance I thought that some usages of `useEnabledNetworks` are more complex to refactor, but they are in fact easy :relaxed:

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
Lemonexe: I get what you're saying - that `reduce` written this way restricts the type, which could be useful. Though in this case, restricting type does not help; the code that consumes `accountTypes` does not look at those properties so it'd make no difference.

But it's a nice pattern. Though if we needed this, the same could be done with `filter` (imho more intuitive to to read):
```
const arr2: A[] = arr.filter((item): item is A => item === "a")
```

packages/suite/src/hooks/settings/useEnabledNetworks.ts
Lemonexe: https://github.com/trezor/trezor-suite/pull/14109/commits/2b501f9ea5eb60e15f1a9492e433987c295ab632

suite-common/wallet-config/src/types.ts
Lemonexe: :magic_wand: :sparkles:  https://github.com/trezor/trezor-suite/pull/14109/commits/b2967eece225048bae32bad56a94e2b7108a2a59

I've done the same thing for network symbol. Previously, this would be allowed for Networks, but now it gives nice error:
ts magic Networks

And with this change, also for Accounts:
ts magic Accounts

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddAccountModal/AddAccountModal.tsx
Lemonexe: https://github.com/trezor/trezor-suite/pull/14109/commits/a41e14f60bf9aebe8b2c0dc83b019d9a3a2fbc0a

peter-sanderson (approved):

suite-common/wallet-config/src/types.ts
peter-sanderson: :clap:

PR #14113, opened 2024-09-04
fix(suite-desktop-core): segfault in bridge autostart

martykan:
## Description

After some changes, probably with node bridge in ThreadProxy, the process in daemon mode starts segfaulting. 

I found out moving waiting for `app.whenReady()` first solves the issue


mroz22 (approved):
cc @MiroslavProchazka 
this is 100% safe to cherrypick. no change in standard app flow.

PR #14115, opened 2024-09-04
Change default bitcoin like network fee unit to sat/vB

matejkriz:
## Description

We were still using `sat/B` as a fee unit. After SegWit it should be changed for `sat/vB` for Bitcoin.

 On mempool.space, there is `sat/vB`, on Blockbook we use `sat/vByte`. You can find someone using `sats/vByte`, but that is basically the same.
 In firmware we also show `sat/vB`.

> This will change the displayed unit in Suite for all bitcoin-like coins including Litecoin (where lit/vB is more common) or those without SegWit. Separation per network would need more refactoring.

## Screenshots:
image
image
image

### DIFF
--- a/packages/suite-web/e2e/tests/coinmarket/exchange.test.ts
+++ b/packages/suite-web/e2e/tests/coinmarket/exchange.test.ts
@@ -175,7 +175,7 @@ describe.skip('Coinmarket exchange', () => {
                     .first()
                     .should('contain.text', 'REGTEST');
             });
-        cy.getTestElement('@modal').within(() => cy.contains('1 sat/B'));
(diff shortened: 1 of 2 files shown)


komret (approved):

bosomt:
**QA OK**


**Info:**
 - Suite version: web 24.10.0 (d48f1d337c96e29fad684a19fbb55bd842b9df74)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:130.0) Gecko/20100101 Firefox/130.0
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: BridgeTransport 3.0.0-bundled.24.10.0

Issue #14117, opened 2024-09-04
Spike: Check the flakiness of our E2E tests


matejkriz:
Android E2E tests are not stable, failing from time to time for no obvious reason. 

Nightly: https://github.com/trezor/trezor-suite/pull/12630

komret:
Some of my recent fails:

https://github.com/trezor/trezor-suite/actions/runs/10665429295/attempts/1
https://github.com/trezor/trezor-suite/actions/runs/10632558852
https://github.com/trezor/trezor-suite/actions/runs/10616016089
https://github.com/trezor/trezor-suite/actions/runs/10594575772/attempts/1
https://github.com/trezor/trezor-suite/actions/runs/10525596559

jvaclavik:
maybe this one https://github.com/trezor/trezor-suite/actions/runs/10698542969/job/29659273802?pr=14089

Lemonexe:
Two times I've got the same as @jvaclavik, on third rerun I've got a different one
https://github.com/trezor/trezor-suite/actions/runs/10697608810/job/29663701700

matejkriz:
Let's close this one now, it's more stable. We can create another spike once we get to it.

PR #14120, opened 2024-09-04
chore(suite): use new Networks in coinjoin

Lemonexe:
Refactoring deprecated `networksCompatibility` to `networks` in `createCoinjoinAccount`.

## Description

Follow-up to #14109, where Add Account Modal was refactored.
In order to split the PRs, `createCoinjoinAccount` was refactored here (with a temporary fix placed in-between, now removed).

## Related Issue

Part of #13839

### DIFF
--- a/packages/suite/src/actions/wallet/__tests__/coinjoinAccountActions.test.ts
+++ b/packages/suite/src/actions/wallet/__tests__/coinjoinAccountActions.test.ts
@@ -70,7 +70,12 @@ describe('coinjoinAccountActions', () => {
             testMocks.setTrezorConnectFixtures(f.connect);
             jest.spyOn(console, 'log').mockImplementation(() => {});
 
-            await store.dispatch(coinjoinAccountActions.createCoinjoinAccount(f.params as any)); // params are incomplete
+            await store.dispatch(
+                coinjoinAccountActions.createCoinjoinAccount(
+                    f.params.network as any,
+                    f.params.account as any,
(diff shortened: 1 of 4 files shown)


packages/suite/src/actions/wallet/__tests__/coinjoinAccountActions.test.ts
```diff
@@ -70,7 +70,12 @@ describe('coinjoinAccountActions', () => {
             testMocks.setTrezorConnectFixtures(f.connect);
             jest.spyOn(console, 'log').mockImplementation(() => {});
 
-            await store.dispatch(coinjoinAccountActions.createCoinjoinAccount(f.params as any)); // params are incomplete
+            await store.dispatch(
+                coinjoinAccountActions.createCoinjoinAccount(
+                    f.params.network as any,
+                    f.params.account as any,
+                ),
```
Lemonexe, line 77: Note: `@total-typescript/shoehorn` is a magic :magic_wand: fix that can help us get rid of `any`.

But not so easily in this case, because in the fixture, there are a lot of strings (inferred as `string`) that would have to be individually cast as `NetworkType` or `AccountType` and it would be quite noisy :disappointed:

Lemonexe:
Note: `[Test] suite-native Android E2E / run_android_e2e_tests` is currently degraded, see #14117

marekrjpolak (approved):

PR #14122, opened 2024-09-04
fix(suite-native): tweak fiat amount skeletons to prevent layout shift

### LINKED ISSUE #13299: Loading fiat value should show skeleton instead of 0 value

yanascz:
Also one big skeleton for fiat balance was split into two smaller ones.

## Related Issue

Resolve #13299

## Screenshots:
graph
accounts

### DIFF
--- a/suite-native/formatters/src/components/EmptyAmountSkeleton.tsx
+++ b/suite-native/formatters/src/components/EmptyAmountSkeleton.tsx
@@ -1,0 +1,13 @@
+import { BoxSkeleton, HStack } from '@suite-native/atoms';
+
+import { EmptyAmountText } from './EmptyAmountText';
+
+export const EmptyAmountSkeleton = () => {
+    // Usage of EmptyAmountText ensures the correct line height.
+    return (
+        <HStack alignItems="center">
+            <EmptyAmountText />
+            <BoxSkeleton width={48} height={20} />
+        </HStack>
+    );
+};
--- a/suite-native/graph/src/components/GraphFiatBalance.tsx
+++ b/suite-native/graph/src/components/GraphFiatBalance.tsx
@@ -1,8 +1,9 @@
 import { Atom, useAtomValue } from 'jotai';
 
 import { FiatGraphPoint } from '@suite-common/graph';
-import { BoxSkeleton, DiscreetTextTrigger, HStack } from '@suite-native/atoms';
+import { Box, BoxSkeleton, DiscreetTextTrigger, HStack, VStack } from '@suite-native/atoms';
 import { FiatBalanceFormatter } from '@suite-native/formatters';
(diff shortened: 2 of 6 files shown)


suite-native/graph/src/components/GraphFiatBalance.tsx
```diff
@@ -37,13 +46,13 @@ export const GraphFiatBalance = ({
     const firstGraphPoint = useAtomValue(referencePointAtom);
 
     if (!firstGraphPoint) {
-        return <BoxSkeleton width={120} height={73} />;
+        return <Skeleton />;
     }
 
     return (
-        <>
+        <Box alignItems="center" style={{ height: 72 }}>
```
matejkriz: Avoid inline styles, please. 

And also - is the fixed height here really necessary? It's kind of ok for the skeletons, but here it's container for real values and if user have bigger font stale, the value can't fit here, I guess

suite-native/formatters/src/components/EmptyAmountSkeleton.tsx
```diff
@@ -0,0 +1,13 @@
+import { BoxSkeleton, HStack } from '@suite-native/atoms';
+
+import { EmptyAmountText } from './EmptyAmountText';
+
+export const EmptyAmountSkeleton = () => {
+    // Usage of EmptyAmountText ensures the correct line height.
```
matejkriz, line 6: Good idea!

suite-native/graph/src/components/GraphFiatBalance.tsx
yanascz: I don't like it here either but I don't know how to ensure that the height doesn't change after the loading is done. That's what causes the layout to shift. 😕 

I'm open to other ideas if you have some.

matejkriz: Never mind, it looks OK even on the biggest scale, so let's not spend more time on it now
image

matejkriz (approved):
Very nice beautiful 😍

PR #13984, opened 2024-08-27
test(suite-desktop): add running bridge version assertion

mroz22:
just a tiny update of spawn-bridge test in desktop.  the only notable addition is that it now asserts expected bridge version.

PR #14123, opened 2024-09-04
test(suite-desktop-core): bridge daemon e2e

martykan:
## Description

Add a basic E2E test for bridge daemon. It starts the background process, checks the bridge status, then starts the UI and checks its status. It then exits the UI and checks the bridge is still running.

### DIFF
--- a/packages/suite-desktop-core/src/modules/bridge.ts
+++ b/packages/suite-desktop-core/src/modules/bridge.ts
@@ -105,8 +105,9 @@ const shouldUseLegacyBridge = (store: Dependencies['store']) => {
     const newBridgeRollout = store.getBridgeSettings().newBridgeRollout || 0;
     // note that this variable is duplicated with UI
     const NEW_BRIDGE_ROLLOUT_THRESHOLD = 1;
-    const legacyBridgeReasonRollout =
-        !isDevEnv && !skipNewBridgeRollout && newBridgeRollout >= NEW_BRIDGE_ROLLOUT_THRESHOLD;
+    const legacyBridgeReasonRollout = !isDevEnv && newBridgeRollout >= NEW_BRIDGE_ROLLOUT_THRESHOLD;
+
+    if (skipNewBridgeRollout && !legacyRequestedByArg) return false;
 
     return (
         legacyRequestedBySettings ||
--- a/.github/workflows/test-suite-desktop-e2e.yml
+++ b/.github/workflows/test-suite-desktop-e2e.yml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - TEST_FILE: "spawn-bridge suite-guide wallet-discovery"
+          - TEST_FILE: "spawn-bridge spawn-bridge-daemon suite-guide wallet-discovery"
             CONTAINERS: "trezor-user-env-unix"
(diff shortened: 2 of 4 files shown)


mroz22:
https://github.com/trezor/trezor-suite/blob/develop/.github/workflows/test-suite-desktop-e2e.yml#L28

mroz22 (commented):
@karliatto how comes that this test is passing in the CI but the other one doing basically the same does not? 🤔  #13984

martykan:
@mroz22 I didn't notice the startBridge command in launchSuite, this test might be wrong

packages/suite-desktop-core/src/modules/bridge.ts
```diff
@@ -105,8 +105,9 @@ const shouldUseLegacyBridge = (store: Dependencies['store']) => {
     const newBridgeRollout = store.getBridgeSettings().newBridgeRollout || 0;
     // note that this variable is duplicated with UI
     const NEW_BRIDGE_ROLLOUT_THRESHOLD = 1;
-    const legacyBridgeReasonRollout =
-        !isDevEnv && !skipNewBridgeRollout && newBridgeRollout >= NEW_BRIDGE_ROLLOUT_THRESHOLD;
+    const legacyBridgeReasonRollout = !isDevEnv && newBridgeRollout >= NEW_BRIDGE_ROLLOUT_THRESHOLD;
+
+    if (skipNewBridgeRollout && !legacyRequestedByArg) return false;
```
martykan, line 110: To bypass EAP condition

karliatto:
> @karliatto how comes that this test is passing in the CI but the other one doing basically the same does not? 🤔 #13984

I am not sure, but I was doing 2 things in that PR:
1. making your changes in the spawn-bridge test pass
2. getting rid of the old image from gitlab "registry.gitlab.com/satoshilabs/trezor/trezor-suite/base:latest" 

I separated point 2 into new PR https://github.com/trezor/trezor-suite/pull/14163

packages/suite-desktop-core/src/modules/bridge.ts
karliatto: That change looks like quite important, to make the node bridge tests.  :+1:

karliatto:
I think this PR is adding some useful changes for testing node bridge in CI, so if this PR is OK after rebasing I think we should merge this one first and then I will update https://github.com/trezor/trezor-suite/pull/13984.

packages/suite-desktop-core/src/modules/bridge.ts
mroz22: hmm we now have rollout 100%. so we might want to stop using rollout for users in EAP (allowPrerelease) and start using it for normal users. 

would you maybe rework it now in that way that:
- if you are in EAP -> node-bridge always
- if you are not in EAP -> apply rollout
- set rollout to 0 (for now)

martykan: But that's probably for a separate PR, right?

mroz22: yeah, could be, I just thought it might simplify conditions here a bit. But yes, not needed as part of this.

mroz22 (approved):

PR #14127, opened 2024-09-04
test(connect): remove ADDRESS_N utility from tests

szymonlesisz:
connect accepts `address_n` in string format

follow up for https://github.com/trezor/trezor-suite/commit/82910e0766eb23ed7731f6fa1519e449a3ccfdf5


szymonlesisz:
~~turns out it doesn't in web 🤔~~

edit: `global.TestUtils` needs to be defined in both jest.setup (used by jest) and common.setup (used by karma and jest)

mroz22 (approved):
# 🧹

Issue #13597, opened 2024-08-01
Mobile offline state


vytick:
Show banner, that the app is offline

___
### [Design here](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Bugfixes?node-id=259-2068&t=gZI3UoFGeSkVq5WN-11)

STew790:
**QA OK**
A yellow banner showing user that they are offline is shown

**Info:**

24.8.0.  5cd19fd

vytick:
@STew790 Please retest after latest update

vytick:
actually wait a bit. I'll do one more check

vytick:
No need to test now. I'll have to make a change. Sorry for ping

vytick:
@STew790 please check now

PR #14131, opened 2024-09-04
feat(suite-native): use our custom URL to check network reachability to prevent pinging google

matejkriz:
## Description

No visible changes, only it should now using our CDN for network check.

## Related Issue

Related #13597

### DIFF
--- a/suite-native/app/src/App.tsx
+++ b/suite-native/app/src/App.tsx
@@ -10,7 +10,7 @@ import { selectIsAppReady, selectIsConnectInitialized, StoreProvider } from '@su
 import { FormatterProvider } from '@suite-common/formatters';
 import { NavigationContainerWithAnalytics } from '@suite-native/navigation';
 import { FeatureMessageScreen, MessageSystemBannerRenderer } from '@suite-native/message-system';
-import { OfflineBanner } from '@suite-native/connection-status';
+import { configureNetInfo, OfflineBanner } from '@suite-native/connection-status';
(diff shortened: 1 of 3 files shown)


vytick (approved):
Great that you push this! ❤️ 

LGTM

suite-native/app/src/App.tsx
```diff
@@ -34,6 +34,11 @@ const APP_STARTED_TIMESTAMP = Date.now();
 // Keep the splash screen visible while we fetch resources
 SplashScreen.preventAutoHideAsync();
 
+// Global configuration of NetInfo for network status monitoring.
```
Nodonisko, line 37: Wouldn't be better place for this component in file with function definition?

matejkriz: You mean for the comment? I was also thinking about that, but I believe it's better to see here why it is being called and what is it about. I don't expect us to expand this in any major way in the future. But both places seem fine to me. @vytick , we need one more vote 😉

matejkriz: I'll merge as it is, anyone feel free to open PR to enhance the comment 🙏

STew790:
**QA OK**
Works as expected. Correctly shows offline/online status and reloading of graphs or etc works also good.

**Info:**
24.9.1 https://github.com/trezor/trezor-suite/commit/bf893a5e0f152087cb4e47a4921a64c3c4c1624f

PR #14138, opened 2024-09-04
feat(suite-native) improve coin enabling flow

### LINKED ISSUE #13819: Coin Enabling - Flow polishing

vytick:
- [x] Make sure showing the UI works correctly with biometric prompt
- [X] preselect networks for view only devices when user enters Coin enabling in settings before initial setup takes place and dont go through initial setup after any network is enabled this way.

## Related Issue

Resolve #13819


matejkriz (approved):
👌 works fine for me

PR #14149, opened 2024-09-05
feat(suite-native): QR code scanner bottom sheet

### LINKED ISSUE #10866: Send coins - address QR scanner
## Description
It should be possible to fill in the address value by scanning a QR code by phone camera or by importing a photo from the gallery.  The functionality of XPUB scan screen will be enhanced and this same component will be reused for the send flow.

## Acceptance criteria
- [ ] No new component is created, the existing XPUB scanning component is reused and improved
- User can insert the recipient address
    - [ ] Via scanning the QR code
    - [ ] from photo library
    
## Resources
- Link to [DESIGN 1](https://www.figma.com/design/sqKFZAIvk3zERxiWgthK8K/Send?node-id=2135-10290&t=KTigu9fek47mphXN-4)
- Link to [DESIGN 2](https://www.figma.com/design/sqKFZAIvk3zERxiWgthK8K/Send?node-id=2135-11081&t=KTigu9fek47mphXN-4)

PeKne:
## Description
- `expo-barcode-scanner` lib was removed because it was deprecated by `expo-camera`
- pick an image from the gallery implemented (`expo-image-picker`)
- scanner is now rendered inside of the bottom sheet instead of a separated modal screen

> 💡NOTE: there is a delay between when the user clicks the pick image button and when the native gallery picker opens. There will probably be a loader added in the follow-up PR. This needs to be discussed with the designer first.

## Related Issue

Resolve #10866

## Screenshots:

https://github.com/user-attachments/assets/60d2e36a-2001-422b-b28b-e56d467b6730


suite-common/icons-deprecated/assets/icons/lightbulb.svg
matejkriz, line 1: beware, I think this icon is used on desktop too...

PeKne: Now I'm confused. Are they still using the `icons-deprecated` package? Even if so, I'm not changing the layout of the icon, it was just optimized by running `yarn generate-icons` from @suite-common/icons-deprecated.

yanascz: FYI, I unified the fill color of all icons to `#000` a while ago. Are you maybe using old `develop` branch as base?

Would be great to keep the color black.

PeKne: I can do that, but does the color matter? The desired color is applied inside of the Icon component anyway.

matejkriz: >  Are they still using the icons-deprecated package?

yes, desktop will migrate to new icons gradually.

I made the comment because of the color. I don't know if it affects desktop or not, would be better to test it. I belive, that on desktop if there is no fill, they will add it automatically, but if there is some fill, it stays.

suite-native/app/app.config.ts
```diff
@@ -75,11 +75,18 @@ const getPlugins = (): ExpoPlugins => {
             },
         ],
         [
-            'expo-barcode-scanner',
+            'expo-camera',
```
matejkriz, line 78: Maybe we should also add `"recordAudioAndroid": false` because default value seems to be `true`? https://docs.expo.dev/versions/latest/sdk/camera/#configurable-properties

PeKne: good point 👍

suite-common/icons-deprecated/assets/icons/lightbulb.svg
yanascz: No, the color doesn't matter, that's why we unified the color to `#000`. And it's more obvious that custom coloring is broken. Not a big deal, of course.

suite-native/module-accounts-import/src/navigation/AccountsImportStackNavigator.tsx
```diff
@@ -34,11 +33,5 @@ export const AccountsImportStackNavigator = () => (
                 component={AccountImportSummaryScreen}
             />
         </AccountsImportStack.Group>
-        <AccountsImportStack.Group screenOptions={{ presentation: 'modal' }}>
-            <AccountsImportStack.Screen
-                name={AccountsImportStackRoutes.XpubScanModal}
```
matejkriz, line 39: It should be removed from navigators and routes too, right?

suite-native/app/app.config.ts
vytick: Maybe, can you, please, remove the old one and add the new libraries to this list?
 https://github.com/trezor/trezor-suite/blob/a35125abeee5c78d9e7f284b0f9bd80ca6ff7765/scripts/list-outdated-dependencies/mobile-dependencies.txt#L35


And I see, there is also `netInfo` missing from the list, can you please add this as well? 🙏

suite-native/module-send/src/components/QrCodeBottomSheetIcon.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const [isVisible, setIsVisible] = useState(false);
+
+    const toggleBottomSheet = () => {
+        setIsVisible(prevState => !prevState);
+    };
+
+    return (
+        <>
+            <TouchableOpacity onPress={toggleBottomSheet}>
+                <Icon name="qrCode" size="large" />
```
vytick, line 24: maybe it would be nice to have bigger touchable opacity around the icon, so it is easier to hit. wdyt?

suite-native/qr-code/src/components/QRCodeScanner.tsx
```diff
@@ -38,33 +49,52 @@ export const QRCodeScanner = ({ onCodeScanned }: QRCodeScannerProps) => {
...
             <Box alignItems="center" justifyContent="center" flex={1}>
-                <Loader title="Loading camera..." />
+                <Loader />
             </Box>
         );
     }
 
     switch (cameraPermissionStatus) {
-        // If the status is "UNDETERMINED" the expo-barcode-scanner library shows a native permission dialog itself.
+        // If the status is "UNDETERMINED" the expo-camera library shows a native permission dialog itself.
```
vytick, line 64: Tested on iOS: App asked for the permission, received it (confirmed in the device settings) and worked once. But second time, it returns `PermissionStatus.UNDETERMINED` meaning nothing is rendered.

vytick: putting app to background with this state of UI and opening right away fixes it 🤷‍♂️. Even repeated tries make it the same: nothing showed, app to bg, open app, scanning showed.

suite-native/qr-code/src/components/PickQRFromGalleryButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+        const pickedImage = await ImagePicker.launchImageLibraryAsync({});
+        const imageUri = pickedImage?.assets?.[0].uri;
+
+        try {
+            const scannedResults = await scanFromURLAsync(imageUri!, ['qr']);
+            const QrData = scannedResults[0];
+
+            onImagePicked(QrData);
+        } catch (error) {
+            showToast({
```
vytick, line 31: sidenote: toast is being rendered below the overlay making it barely visible (I thought app did nothing at the moment). Not sure if fixable in this pr 🤔

suite-native/module-send/src/components/QrCodeBottomSheetIcon.tsx
PeKne: Did you have problem with hitting the button? Extra hitSlop added in [22fe1f4](https://github.com/trezor/trezor-suite/pull/14149/commits/22fe1f4715253e96af6a467f7e4a0a71ea8fd79e).

suite-native/module-accounts-import/src/navigation/AccountsImportStackNavigator.tsx
PeKne: That is right, removed in [22fe1f4](https://github.com/trezor/trezor-suite/pull/14149/commits/22fe1f4715253e96af6a467f7e4a0a71ea8fd79e)

suite-native/app/app.config.ts
PeKne: @matejkriz permissions fixed in 22fe1f4715.

@vytick the QR code-related libraries were already added there with the original commit. Since netinfo is unrelated to this issue, I created a separate PR for it: https://github.com/trezor/trezor-suite/pull/14222

suite-native/qr-code/src/components/PickQRFromGalleryButton.tsx
PeKne: Oh, you are right. It is a known issue, but it definitely does not look nice here. I changed it so the bottom sheet is hidden in case of error, so the toast is visible. [22fe1f4](https://github.com/trezor/trezor-suite/pull/14149/commits/22fe1f4715253e96af6a467f7e4a0a71ea8fd79e)

suite-native/qr-code/src/components/QRCodeScanner.tsx
PeKne: Should work now as expected: [22fe1f4](https://github.com/trezor/trezor-suite/pull/14149/commits/22fe1f4715253e96af6a467f7e4a0a71ea8fd79e)

PeKne:
@matejkriz The performance problems discovered by you were caused by infinite useEffect triggered from `useCameraPermission`. An event listener was listening to an AppState change event and triggering the `requestCameraPermision` system dialog. This dialog was changing the AppState as well, so the listener was triggered in an infinite loop. Fixed by [22fe1f4](https://github.com/trezor/trezor-suite/pull/14149/commits/22fe1f4715253e96af6a467f7e4a0a71ea8fd79e)

suite-common/icons-deprecated/assets/icons/lightbulb.svg
PeKne: @yanascz fixed in ede5e8459b. I did not know you unified all of them. Don't want to ruin your efforts ;).

suite-native/app/app.config.ts
vytick: Sorry, I must be blind 🙈 . And thanks!  🙏

suite-native/qr-code/src/components/QRCodeScanner.tsx
vytick: Works nicely! 👏 👌

vytick (approved):
Tested with Portfolio tracker on iOS. Works nicely, Great job! 

And thanks for the netinfo PR 🙇

PR #14150, opened 2024-09-05
Chore(wallet-utils): use new Network in wallet-utils

Lemonexe:
Refactoring deprecated `networksCompatibility` to `networks` mostly in `wallet-utils`, and related files.

## Description

-  refactor util functions (low hanging fruit)
- refactor related types

## Related Issue

Part of #13839


suite-common/wallet-constants/src/account.ts
```diff
@@ -1,12 +0,0 @@
...
-    ETHEREUM: 'ethereum',
-    RIPPLE: 'ripple',
-};
-
-export const ACCOUNT_TYPE = {
-    NORMAL: 'normal',
-    LEGACY: 'legacy',
-    SEGWIT: 'segwit',
-    TAPROOT: 'taproot',
-};
```
Lemonexe, line 12: Not used anymore

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -1010,17 +1005,16 @@ export const getPendingAccount = ({
     };
 };
 
-export const getNetworkFeatures = ({
-    networkType,
+export const getNetworkAccountFeatures = ({
```
Lemonexe, line 1017: - `networkType` is absolutely redundant, it's both ways dependent on `symbol`
- renamed to reflect that it gets features either from `account`, or `network` (for `'normal'`)

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -1030,11 +1024,7 @@ export const hasNetworkFeatures = (
         return false;
     }
 
-    const networkFeatures = getNetworkFeatures(account);
-
-    if (!networkFeatures) {
-        return false;
-    }
```
Lemonexe, line 1037: can't be falsy anymore (at least `[]` is returned), see above

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -418,6 +420,9 @@ export const getAllAccounts = (deviceState: string | typeof undefined, accounts:
     return accounts.filter(a => a.deviceState === deviceState && a.visible);
 };
 
+/**
+ * @deprecated use directly networks[symbol], you can import { networks } from '@suite-common/wallet-config'
+ */
```
Lemonexe, line 431: Shall be done in following PR (has slightly larger splash zone)

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -273,11 +275,7 @@ export const getAccountTypeUrl = (path: string) => {
     }
 };
 
-export const getAccountDecimals = (symbol: NetworkSymbol) => {
-    const network = networksCompatibility.find(n => n.symbol === symbol);
-
-    return network?.decimals;
-};
+export const getAccountDecimals = (symbol: NetworkSymbol) => networks[symbol]?.decimals;
```
Lemonexe, line 278: the `?.` is not necessary TS-wise, but there'd be runtime errors, because at some places, `getAccountDecimals` gets called with with string cast as `NetworkSymbol` :disappointed:

suite-common/wallet-types/src/account.ts
```diff
@@ -87,7 +91,7 @@ export type Account = {
     unlockPath?: PROTO.UnlockPath; // parameter used to unlock SLIP-25/coinjoin keychain
     descriptor: string;
     descriptorChecksum?: string;
-    accountType: NonNullable<NetworkCompatible['accountType']>;
+    accountType: NonNullable<NetworkAccountType>;
```
komret: This s weird. Why is it `NonNullable`? It doesn't seem to do anything. Looks like `accountType` can just be `AccountType` and we can remove the reexport from this file.

suite-common/wallet-utils/src/accountUtils.ts
komret: This would still create some unexpected behaviour, then. It might be better to just let the app crash in those cases so that QA or Sentry can catch it.

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -365,22 +364,26 @@ export const formatTokenAmount = (tokenTransfer: TokenTransfer) => {
...
-            const accountType = n.accountType || ACCOUNT_TYPE.NORMAL;
+export const sortByCoin = (accounts: Account[]) => {
+    const orderedNetworkSymbols = Object.keys(networks) as NetworkSymbol[];
 
-            return accountType === a.accountType && n.symbol === a.symbol;
-        });
-        const bIndex = networksCompatibility.findIndex(n => {
-            const accountType = n.accountType || ACCOUNT_TYPE.NORMAL;
+    return accounts.sort((a, b) => {
+        // sort by order of networks key (symbols)
```
komret: Typo: network keys

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -365,22 +364,26 @@ export const formatTokenAmount = (tokenTransfer: TokenTransfer) => {
...
+        if (aSymbolIndex !== bSymbolIndex) return aSymbolIndex - bSymbolIndex;
 
-            return accountType === b.accountType && n.symbol === b.symbol;
-        });
-        if (aIndex === bIndex) return a.index - b.index;
+        // same network, so sort by accountType keys
+        const network = networks[a.symbol];
+        const orderedAccountTypes = Object.keys(network.accountTypes) as AccountType[];
+        const aAccountTypeIndex = orderedAccountTypes.indexOf(a.accountType);
+        const bAccountTypeIndex = orderedAccountTypes.indexOf(b.accountType);
```
komret: This is new, isn't it? Is going to be used somewhere? If not, I wouldn't do it.

komret (changes requested):

suite-common/wallet-utils/src/accountUtils.ts
Lemonexe: Yes, it is necessary to keep the behavior same. Let me illustrate:

**Before, with `networksCompatibility`**:
```
networksCompatibility: 
- btc, normal
- btc, taproot
- eth, normal
- eth, ledger
```
We need to sort by network as primary criterion, and by account as secondary (that's what the old fn did).
It was a simpler implementation, because `networksCompatibility` is a flat structure already sorted exactly the way we as we want, so we just look up the indices.

**Now with `networks`** the nested structure, I really have to do two steps. First sort by network (btc vs. eth, but normal vs. taproot are not yet sorted), and then when we have them grouped by network, sort them by account (btc normal vs. btc taproot).

Still, I tried to remove the four lines to see what happens, and test coverage fails :x:

suite-common/wallet-types/src/account.ts
Lemonexe: Thanks for the catch! `NonNullable` is really useless.

But this isn't just reexport of modified interface, it is a completely new interface:

`NetworkAccount` in `networksConfig.ts` is just a universal _definition_ of the account type,
but this `Account` is for the actual "instances" of the account type, with specific addresses, balance, history etc.

suite-common/wallet-utils/src/accountUtils.ts
Lemonexe: I have written a more meaningful commentary, [as per comment below](https://github.com/trezor/trezor-suite/pull/14150#discussion_r1745648881)

suite-common/wallet-types/src/account.ts
komret: Sorry I meant the reexport of `AccountType`, not `Account`.

Lemonexe: Yeah, I missed that. Removed in e08f68d0c880f23d58c79108e7705d0a970a0a01

suite-common/wallet-utils/src/accountUtils.ts
Lemonexe: As discussed in PM, we shall postpone it (usages of `getAccountDecimals` rely on this behavior)

komret (approved):

komret:
There is an unused dependency in `@suite-native/config`, see [validation](https://github.com/trezor/trezor-suite/actions/runs/10723441391/job/29737482793?pr=14150).

Lemonexe:
> There is an unused dependency in `@suite-native/config`, see [validation](https://github.com/trezor/trezor-suite/actions/runs/10723441391/job/29737482793?pr=14150).

Cool, didn't know we had that. Updated in  https://github.com/trezor/trezor-suite/pull/14150/commits/bab94d6f5dde8fe3632a213a2fb3be46a5ef69fa

EDIT: oh I also have to do `yarn update-project-references` afterwards..

Lemonexe:
\rebase

PR #14153, opened 2024-09-05
Chore/add versioning to transport bridge

karliatto:
## Description
Add version to tranport-bridge containing suite version as `<transport-bridge>-bundled.<suite-version>`.

image

### DIFF
--- a/packages/suite-desktop-core/src/threads/bridge.ts
+++ b/packages/suite-desktop-core/src/threads/bridge.ts
@@ -29,6 +29,7 @@ const init = (settings: TrezordNodeSettings) =>
             }),
             { serviceName: 'trezord-node' },
         ),
+        bundledVersion: process.env.VERSION,
     });
 
 createThread(init);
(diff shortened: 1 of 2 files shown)


mroz22 (changes requested):
This is another manual step we are going to forget about. And since we are not going to release a standalone version anytime soon, I'd say we shouldn't do version bumps like this. Instead I propose something like:

- where node bridge is initiated, pass suite-destkop version as a param
- from node-bridge endpoints returning version, return something like `3.0.0-bundled.24.9.0`

karliatto:
> This is another manual step we are going to forget about. And since we are not going to release a standalone version anytime soon, I'd say we shouldn't do version bumps like this. Instead I propose something like:
> 
>     * where node bridge is initiated, pass suite-destkop version as a param
> 
>     * from node-bridge endpoints returning version, return something like `3.0.0-bundled.24.9.0`

Ok, let's do it like that then, I am experimenting with that way.. So when we run trerzordnode as standalone it should only display `3.0.0` ?

mroz22 (approved):
looks ok 👍

PR #14160, opened 2024-09-05
refactor(ds): Add divider to Flex

jvaclavik:
## Related Issue

Resolve 

## Screenshots:
image

### DIFF
--- a/packages/components/src/components/Flex/Flex.tsx
+++ b/packages/components/src/components/Flex/Flex.tsx
@@ -1,7 +1,9 @@
-import { SpacingValues } from '@trezor/theme';
-import styled from 'styled-components';
+import { Elevation, mapElevationToBorder, SpacingValues } from '@trezor/theme';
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Flex/Flex.tsx
```diff
@@ -82,6 +130,9 @@ export type FlexProps = AllowedFrameProps & {
     flex?: Flex;
     flexWrap?: FlexWrap;
     isReversed?: boolean;
+    hasDivider?: boolean;
+    /** @deprecated Use only is case of absolute desperation. Prefer keep it according to elevation. */
```
seibei-iguchi, line 134: Why adding a prop if we immediately say it's deprecated?

seibei-iguchi (approved):

packages/components/src/components/Flex/Flex.tsx
jvaclavik: Good question, we might delete it if it's not necessary

PR #14162, opened 2024-09-05
Remove `SecurityCheckButton`

komret:
## Description

The `SecurityCheckButton` had a set `height` so that it does not change height when switching between `SecurityCheck` and `SecurityCheckFail` (when you click the secondary button). This broke during Suite redesign when `Button` changed its default height.

I think we actually don't need it to keep a constant height between the views so I deleted it to keep things simple.

I also changed the button layout to resemble that of Suite modals. However, it doesn't look great, does it? Maybe we should consider adding `flex` or `flexGrow` etc. to `frameProps`, @jvaclavik?

## Related Issue

Resolve 

## Screenshots:
Before:
Screenshot 2024-09-05 at 16 45 27
After:
Screenshot 2024-09-05 at 16 46 19

### DIFF
--- a/packages/components/src/components/buttons/Button/Button.tsx
+++ b/packages/components/src/components/buttons/Button/Button.tsx
@@ -160,6 +160,7 @@ export const Button = ({
     return (
         <ButtonContainer
             as={isLink ? 'a' : 'button'}
+            target={isLink ? target || '_blank' : undefined}
             href={href}
             $variant={variant}
             $size={size}
--- a/packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
+++ b/packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
@@ -1,12 +1,11 @@
 import styled from 'styled-components';
 
-import { H2, Row, Text, useElevation } from '@trezor/components';
+import { Button, H2, Row, Text, useElevation } from '@trezor/components';
 import { Elevation, mapElevationToBorder, spacings, spacingsPx } from '@trezor/theme';
 import { TREZOR_SUPPORT_FW_CHECK_FAILED } from '@trezor/urls';
 
 import { Translation, TrezorLink } from 'src/components/suite';
 import { SecurityChecklist } from '../../../views/onboarding/steps/SecurityCheck/SecurityChecklist';
-import { SecurityCheckButton } from '../../../views/onboarding/steps/SecurityCheck/SecurityCheckButton';
 import { SecurityCheckLayout } from './SecurityCheckLayout';
 
 const TopSection = styled.div<{ $elevation: Elevation }>`
@@ -19,10 +18,6 @@ const TopSection = styled.div<{ $elevation: Elevation }>`
 const StyledTrezorLink = styled(TrezorLink)`
     /* flex-grow has no effect on a link, display is set to contents so that it can be read from the child */
     display: contents;
-`;
-
-const StyledSecurityCheckButton = styled(SecurityCheckButton)`
-    flex-grow: 1;
 `;
 
 const checklistItems = [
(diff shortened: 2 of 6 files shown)


packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
```diff
@@ -231,27 +220,29 @@ const SecurityCheckContent = ({
...
-                        </TimeEstimateWrapper>
-                    </SecurityCheckButtonWithSecondLine>
+                        <Column>
+                            <Translation id={primaryButtonTopText} />
+                            <TimeEstimateWrapper>
+                                <Icon size={12} name="clock" color={theme.iconOnPrimary} />
+                                <Translation id="TR_TAKES_N_MINUTES" />
+                            </TimeEstimateWrapper>
+                        </Column>
+                    </Button>
```
jvaclavik: How about using `isFullWidth` prop for all buttons here?

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
```diff
@@ -69,18 +68,8 @@ const Buttons = styled.div<{ $elevation: Elevation }>`
     display: flex;
     flex-wrap: wrap;
```
jvaclavik, line 71: You probably need to change also flex wrapping here. It should be in media query where`flex-wrap: wrap` is set only for small resolutions.

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
komret: It works differently, though.

1. Doesn't work well with `flex-wrap`
2. When only one of the buttons `isFullWidth`, the other one shrinks:
flex-grow:
Screenshot 2024-09-05 at 17 44 37
width:
Screenshot 2024-09-05 at 17 37 16
3. When each of the `isFullWidth`, they grow to the same width instead of growing proportionally:
flex-grow:
Screenshot 2024-09-05 at 17 46 01
width:
Screenshot 2024-09-05 at 17 45 21

jvaclavik:
PR was updated. 

Current design:

Screenshot 2024-09-10 at 12 06 03

Screenshot 2024-09-10 at 12 06 12

Screenshot 2024-09-10 at 12 06 15

image

packages/components/src/components/buttons/Button/Button.tsx
```diff
@@ -160,6 +160,7 @@ export const Button = ({
     return (
         <ButtonContainer
             as={isLink ? 'a' : 'button'}
+            target={isLink ? target || '_blank' : undefined}
```
jvaclavik, line 163: This was a bug in button with `href` (`target` didn't work)

packages/urls/src/urls.ts
```diff
@@ -93,7 +93,7 @@ export const HELP_CENTER_EVM_ADDRESS_CHECKSUM =
 export const HELP_CENTER_FIRMWARE_REVISION_CHECK =
     'https://trezor.io/learn/a/trezor-firmware-revision-check';
 export const TREZOR_SUPPORT_FW_CHECK_FAILED =
```
jvaclavik, line 95: I don't understand why it worked before

packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
```diff
@@ -1,23 +1,21 @@
...
     margin-top: ${spacingsPx.xs};
-    padding-bottom: ${spacingsPx.xl};
+    margin-bottom: ${spacingsPx.xl};
     width: 100%;
 `;
 
-const StyledTrezorLink = styled(TrezorLink)`
-    /* flex-grow has no effect on a link, display is set to contents so that it can be read from the child */
-    display: contents;
+const Flex = styled.div`
```
jvaclavik, line 17: Not a perfect solution, but it's OK for now until we come with something better.

packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
```diff
@@ -50,30 +48,29 @@ export const SecurityCheckFail = ({ goBack, useSoftMessaging }: SecurityCheckFai
         ? 'TR_DEVICE_COMPROMISED_TEXT_SOFT'
         : 'TR_DEVICE_COMPROMISED_TEXT';
 
-    const { elevation } = useElevation();
-
     return (
```
jvaclavik, line 51: I prioritized it in our design backlog that we are going to polish these screens in near future.

jvaclavik (approved):

PR #14163, opened 2024-09-05
tests: get rid of gitlab old trezor-suite image in suite desktop tests

karliatto:
## Description

This PR stop using docker image ` registry.gitlab.com/satoshilabs/trezor/trezor-suite/base:latest` that was required in gitlab but we do not need it anymore, so it is possible to simplify and delete some code.

Remove duplicated suite-desktop-nightly tests .github/workflows/test-suite-desktop-nightly.yml it is not necessary to have exactly same code we can just make same workflow to run in multiple situations.

## Related Issue

Related https://github.com/trezor/trezor-suite/pull/13984/files


mroz22 (approved):
Perfect. Now it should be also faster.

PR #14165, opened 2024-09-05
fix(suite-native): add missing bottom margin to discovery assets loader

### LINKED ISSUE #14090: Wrong position of Retrieving balances title when loading
Reported by @shenkys:
 
~~- [ ] Last network has wrong balance on dashboard (zero)~~ moved to separate issue https://github.com/trezor/trezor-suite/issues/14177
- [ ] While discovery is running, text Retrieving balances is too close to bottom edge (was not previously)

**Screenshots**

yanascz:
## Related Issue

Partially resolves #14090

## Screenshots:
image

### DIFF
--- a/suite-native/assets/src/components/DiscoveryAssetsLoader.tsx
+++ b/suite-native/assets/src/components/DiscoveryAssetsLoader.tsx
@@ -32,7 +32,7 @@ export const DiscoveryAssetsLoader = ({ numberOfAssets }: { numberOfAssets: numb
                 <ListItemSkeleton key={i} />
             ))}
 
-            <HStack justifyContent="center">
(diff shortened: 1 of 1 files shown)


matejkriz (approved):
nitpick: it also miss align-items center so the text is not vertically aligned perfectly

yanascz:
> nitpick: it also miss align-items center so the text is not vertically aligned perfectly

There's no visual change if I add it. Tried both platforms.

matejkriz:
> > nitpick: it also miss align-items center so the text is not vertically aligned perfectly
> 
> There's no visual change if I add it. Tried both platforms.

It is, just barely visible :-D That is why I marked it as a nitpick, since it is visible only on different font scales

PR #14091, opened 2024-09-03
fix(suite): fix condition to display tooltip over used address button

komret:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14084

PR #14166, opened 2024-09-05
Enable displaying receive addresses when revision check is failed but disabled

### LINKED ISSUE #14084: Incorrect hint in receive section
**Describe the bug**
`Your device may have been compromised. Do not send funds to it.` message as hint when revealing receive address with legit device

**Steps to reproduce:**
1. Navigate to receive section
2. Reveal address
3. Observe hint

**Info:**
 - Suite version: web 24.9.0 (983e42f2cc2f6fb51463a26c7f6206c4f56a3d14)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:129.0) Gecko/20100101 Firefox/129.0
 - OS: MacIntel
 - Screen: 2560x1440

**Screenshots:**

https://github.com/user-attachments/assets/60c65325-98a7-4fa2-86a3-1c5caa08f566

komret:
## Description

The button to display address is disabled when firmware revision check has failed. However, the original implementation didn't take into account the option to disabled the check. Now, if the check is disabled (but it still fails under the hood), you should be able to generate a new address and display used addresses.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14084 (see https://github.com/trezor/trezor-suite/issues/14084#issuecomment-2328000550)

### DIFF
--- a/packages/suite/src/components/suite/Preloader/Preloader.tsx
+++ b/packages/suite/src/components/suite/Preloader/Preloader.tsx
@@ -13,20 +13,16 @@ import {
     selectPrerequisite,
     selectIsLoggedOut,
     selectSuiteFlags,
+    selectIsFirmwareRevisionCheckEnabledAndFailed,
 } from 'src/reducers/suite/suiteReducer';
 import { SuiteStart } from 'src/views/start/SuiteStart';
 import { PrerequisitesGuide } from '../PrerequisitesGuide/PrerequisitesGuide';
 import { LoggedOutLayout } from '../layouts/LoggedOutLayout';
 import { WelcomeLayout } from '../layouts/WelcomeLayout/WelcomeLayout';
 import { ViewOnlyPromo } from 'src/views/view-only/ViewOnlyPromo';
-import {
-    selectDevice,
-    selectFailedSecurityChecks,
-    selectIsFirmwareRevisionCheckDismissed,
-} from '@suite-common/wallet-core';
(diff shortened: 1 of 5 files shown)


packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -92,12 +84,8 @@ export const Preloader = ({ children }: PreloaderProps) => {
     if (
         (router.route?.app === undefined ||
             !ROUTES_TO_SKIP_REVISION_CHECK.includes(router.route?.app)) &&
-        selectedDevice?.features &&
-        selectedDevice.connected === true &&
```
komret, line 96: I think this should fail for remembered wallets, too, so I removed this condition.

packages/suite/src/reducers/suite/suiteReducer.ts
komret, line 1: I moved the selector here because I could not access `suite.settings` in `deviceReducer`. This is not used by the mobile app yet, so it doesn't need to be in `suite-common`.

Lemonexe (approved):
LGTM.

The new suiteReducer is easier to reuse than the previous device reducer :+1:  

I did not think of this case before when reviewing #14091, but now I have tested:
- legit device, incl. remembered wallet :heavy_check_mark:
- compromised with only the Back button clicked, incl. remembered wallet :heavy_check_mark:
- compromised with firmware check disabled , incl. remembered wallet :heavy_check_mark:

PR #14178, opened 2024-09-06
ci(docker): get rid of trezor-suite gitlab image

karliatto:
## Description
When running CI in GitLab we were using docker image "registry.gitlab.com/satoshilabs/trezor/trezor-suite/base:latest" but now it is not needed anymore. 
So with this PR I am getting rid of that image and simplifying a bit the process.

### DIFF
--- a/docker/docker-connect-test.sh
+++ b/docker/docker-connect-test.sh
@@ -48,8 +48,6 @@ FIRMWARE_URL=""
 INCLUDED_METHODS=""
 EXCLUDED_METHODS=""
 DOCKER=true
-# TODO: DOCKER_PATH appears unused. Remove or export?
-DOCKER_PATH="docker"
 USE_TX_CACHE=true
 USE_WS_CACHE=true
 PATTERN=""
@@ -58,13 +56,10 @@ RANDOMIZE=false
 
 # user options
 OPTIND=2
-while getopts ":p:i:e:f:u:m:D:hdcr" opt; do
+while getopts ":p:i:e:f:u:m:hdcr" opt; do
   case $opt in
   d)
     DOCKER=false
-    ;;
-  D)
-    DOCKER_PATH="$OPTARG"
     ;;
   c)
     USE_TX_CACHE=false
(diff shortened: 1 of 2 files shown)


vdovhanych (commented):
Please change this to all of the testing flows not only connect. Its being used for everything and i would like to change it in one go.

docker/docker-connect-test.sh
```diff
@@ -47,9 +44,6 @@ FIRMWARE=""
 FIRMWARE_URL=""
 INCLUDED_METHODS=""
 EXCLUDED_METHODS=""
-DOCKER=true
-# TODO: DOCKER_PATH appears unused. Remove or export?
-DOCKER_PATH="docker"
```
mroz22: these are used, you can run tests from command line in one script. and it either starts trezor-user-env or not.

karliatto:
> Please change this to all of the testing flows not only connect. Its being used for everything and i would like to change it in one go.

@vdovhanych I cannot find the image "registry.gitlab.com/satoshilabs/trezor/trezor-suite/base:latest" anywhere else in the repo but maybe I am missing somethig.

docker/docker-connect-test.sh
karliatto: Ok, I did not think about use case where you want to run this script locally with trezor-user-env locally. I will keep it.

vdovhanych:
> > Please change this to all of the testing flows not only connect. Its being used for everything and i would like to change it in one go.
> 
> @vdovhanych I cannot find the image "registry.gitlab.com/satoshilabs/trezor/trezor-suite/base:latest" anywhere else in the repo but maybe I am missing somethig.

Hmm, I thought it's used for e2e testing also, but it's actually using a cypress image, right? Well, if it's not anywhere else, then it's good, I guess :).

docker/docker-connect-test.sh
karliatto: Done, by those fixups 
* https://github.com/trezor/trezor-suite/pull/14178/commits/90aef7972ae1681c8caa66420dcbcce749d357ab
* https://github.com/trezor/trezor-suite/pull/14178/commits/69ec61cd7459065dc65b84e3a7562c2e1d30361d

Not that functionality of the script is back.

karliatto:
It looks like all tests are now passing here.

vdovhanych (approved):

PR #14180, opened 2024-09-06
fix: sessions backround sharedworker truly shared

### LINKED ISSUE #12002: connect: webusb: sync broken - not shared sharedworker
If you acquire a device in either suite or connect, you won't be able to work with it in the other application. 

image

image

Actions steps: 
1. display a meaningful error 
2. fix synchronization between the two. The main problem is that suite has local connect which means that connect there and normal connect have different origins which leads to having 2 sharedworkers active at the same time which of course completely stops any synchronization.

mroz22:
## Description

Connect-iframe has been bundled into suite-web since the very beginning. This, however, caused a problem: the shared workers used for synchronizing device access between tabs were not really shared (between suite-web and connect) because there were two of them at the same time.

The tldr change by this PR is that from now on the single source of the sharedworker is @trezor/connect-web deployment and suite-web may be merely pointed to it. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12002

## Screenshots:

note: screenshots are outdated. I changed the naming later. 

Before: 
image

After:
image

## Test env

- sldev suite-web https://dev.suite.sldev.cz/suite-web/shared-sharedworker/web/
- sldev connect-exp https://dev.suite.sldev.cz/connect/shared-sharedworker/methods/bitcoin/getAddress/

### DIFF
--- a/packages/transport-native/src/nativeUsb.ts
+++ b/packages/transport-native/src/nativeUsb.ts
@@ -2,7 +2,6 @@ import { WebUSB } from '@trezor/react-native-usb';
 import {
     Transport as AbstractTransport,
     AbstractApiTransport,
-    SessionsClient,
     SessionsBackground,
     UsbApi,
 } from '@trezor/transport';
@@ -11,20 +10,10 @@ export class NativeUsbTransport extends AbstractApiTransport {
     // TODO: Not sure how to solve this type correctly.
     public name = 'NativeUsbTransport' as any;
 
-    private readonly sessionsBackground;
+    private readonly sessionsBackground = new SessionsBackground();
 
     constructor(params: ConstructorParameters<typeof AbstractTransport>[0]) {
         const { messages, logger } = params;
-        const sessionsBackground = new SessionsBackground();
-
-        const sessionsClient = new SessionsClient({
-            requestFn: args => sessionsBackground.handleMessage(args),
-            registerBackgroundCallbacks: () => {},
-        });
-
-        sessionsBackground.on('descriptors', descriptors => {
-            sessionsClient.emit('descriptors', descriptors);
-        });
 
         super({
             messages,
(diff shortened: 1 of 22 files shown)


suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -108,11 +108,18 @@ export const connectInitThunk = createThunk(
 
         cardanoConnectPatch(getEnabledNetworks);
 
+        const sessionsBackground =
+            window.location.origin +
+            `${process.env.ASSET_PREFIX || ''}/static/connect/workers/sessions-background-sharedworker.js`.replaceAll(
+                /\/+/g,
+                '/',
+            );
```
mroz22: by using a `_` prefixed variable I am trying to express that although this is part of public API, you are not supposed to use it. 
 
I was also thinking about other options how to achieve the same without introducing a new connect init param, such as setting the URL somewhere deeper using env variable but it felt even clumsier than this.

packages/transport/src/transports/abstract.ts
```diff
@@ -163,6 +165,7 @@ export abstract class AbstractTransport extends TypedEmitter<{
         signal.addEventListener('abort', abort);
 
         this.logger = logger;
+        this.signal = signal;
```
mroz22: because of https://github.com/trezor/trezor-suite/pull/14180/commits/c3d4d2fbff5c4009ba377b99fe1b904fd9c5650f#diff-3a72a34ba3cff1d3318d6902cfe59925e7c3b6e9fc7270295e294d6f9677f18bR32

packages/transport-native/src/nativeUsb.ts
```diff
@@ -1,45 +1,24 @@
...
                 logger,
             }),
-            sessionsClient,
         });
-        this.sessionsBackground = sessionsBackground;
     }
 
     public listen() {
         this.api.listen();
-        this.sessionsBackground.dispose();
```
mroz22, line 42: hmmm why was this needed?

packages/transport-native/src/nativeUsb.ts
```diff
@@ -1,45 +1,24 @@
...
-        const sessionsBackground = new SessionsBackground();
-
-        const sessionsClient = new SessionsClient({
-            requestFn: args => sessionsBackground.handleMessage(args),
-            registerBackgroundCallbacks: () => {},
-        });
-
-        sessionsBackground.on('descriptors', descriptors => {
-            sessionsClient.emit('descriptors', descriptors);
-        });
```
mroz22, line 27: this was moved to parent - init method

packages/transport-native/src/nativeUsb.ts
marekrjpolak: Oof. It's because when I removed abort signal from session background, I wanted to keep it 1:1. But `sessionsBackground.dispose()` should've been part of `stop()`, not `listen()`, of course. So it's my mistake.

marekrjpolak (approved):

PR #14192, opened 2024-09-06
fix(suite-native): use default letter spacing

### LINKED ISSUE #14177: Wrong line break of fiat values on dashboard
- Spotted so far only on Android
- Already in production
- With smallest system font scale it's "fixed"
- Only some values are broken, 0.00, but also for example $5.35

image
image
image

matejkriz:
## Description

POC solution to see how default letter spacing looks like.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14177

## Screenshots:
image


matejkriz:
the current state of custom letter spacing is what we want

PR #14194, opened 2024-09-06
chore(suite): update mission dependencies

Lemonexe:
## Description

Update most Mission-related dependencies:

- `electron-builder`
- `simple-git`
  - moved to usability dependencies, as it pertains Suite Guide
- `electron-updater`
- `varuint-bitcoin`

These were **not updated**:
- `electron` 31
  - it is [supported by nixos](https://search.nixos.org/packages?channel=24.05&show=electron&from=0&size=50&sort=relevance&type=packages&query=electron), but not properly, could break build unexpectedly in future, [see comment](https://github.com/trezor/trezor-suite/pull/14194#discussion_r1750271053) 
-  `tiny-secp256k1` WIP in [#12261](https://github.com/trezor/trezor-suite/pull/12261) 
- `electron-store` requires refactoring our electron main process to ES
  - that is possible since [electron 28](https://github.com/electron/electron/blob/main/docs/tutorial/esm.md), which we have since [#13795](https://github.com/trezor/trezor-suite/pull/13795), but is out of scope

Besides CI checks :heavy_check_mark: , I have tested: local suite:dev, suite:dev:desktop, linux build :heavy_check_mark:

:information_source: For reference, last bump mission deps PR was #13938

### DIFF
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -47,5 +47,6 @@ module.exports = {
         '^bcrypto/lib/(.*)$': 'bcrypto/lib/$1-browser',
         // Enforce usage of CommonJS version of uuid because ESM version is not working in Jest
         '^uuid$': require.resolve('uuid'), // https://stackoverflow.com/questions/73203367/jest-syntaxerror-unexpected-token-export-with-uuid-library
+        '^uint8array-tools$': require.resolve('uint8array-tools'), // same case as with uuid
     },
 };
--- a/jest.config.native.js
+++ b/jest.config.native.js
@@ -44,5 +44,4 @@ module.exports = {
         '<rootDir>/../../suite-native/test-utils/src/expoMock.js',
         '<rootDir>/../../suite-native/test-utils/src/walletSdkMock.js',
     ],
-    moduleNameMapper: { '^uuid$': require.resolve('uuid') },
 };
--- a/packages/suite/jest.config.js
+++ b/packages/suite/jest.config.js
@@ -1,3 +1,5 @@
+const baseConfig = require('../../jest.config.base');
+
 // all tests have same UTC timezone
 process.env.TZ = 'UTC';
 process.env.LANG = 'en-US';
@@ -31,9 +33,7 @@ module.exports = {
         '^src/(.+)': '<rootDir>/src/$1',
         '\\.(mp4)$': '<rootDir>/__mocks__/import-mp4.js',
         '\\.(svg)$': '<rootDir>/__mocks__/import-svg.js',
-        uuid: require.resolve('uuid'), // https://stackoverflow.com/questions/73203367/jest-syntaxerror-unexpected-token-export-with-uuid-library
-        // Enforce usage of JS version of bcrypto in tests because on CI we don't build native modules because it's slowing yarn install
-        '^bcrypto/lib/(.*)$': 'bcrypto/lib/$1-browser',
+        ...baseConfig.moduleNameMapper,
     },
     moduleFileExtensions: ['js', 'ts', 'tsx'],
     coverageDirectory: './coverage',
--- a/packages/utxo-lib/src/bufferutils.ts
+++ b/packages/utxo-lib/src/bufferutils.ts
@@ -7,15 +7,17 @@
 
 import BN from 'bn.js';
 import pushdata from 'pushdata-bitcoin';
-import varuint from 'varuint-bitcoin';
+import * as varuint from 'varuint-bitcoin';
 import { Int64LE } from 'int64-buffer';
 import { bufferUtils } from '@trezor/utils';
 import * as types from './types';
 
+const OUT_OF_RANGE_ERROR = 'value out of range';
+
 export function verifuint(value: number, max: number) {
     if (typeof value !== 'number') throw new Error('cannot write a non-number as a number');
     if (value < 0) throw new Error('specified a negative value for writing an unsigned value');
-    if (value > max) throw new Error('value out of range');
+    if (value > max) throw new Error(OUT_OF_RANGE_ERROR);
     if (Math.floor(value) !== value) throw new Error('value has a fractional component');
 }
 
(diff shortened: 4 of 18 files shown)


packages/utxo-lib/src/bufferutils.ts
```diff
@@ -82,18 +84,20 @@ export function writeInt64LE(buffer: Buffer, value: number, offset: number) {
 }
 
 export function readVarInt(buffer: Buffer, offset: number) {
-    const result = varuint.decode(buffer, offset);
+    const { numberValue, bytes } = varuint.decode(buffer, offset);
```
Lemonexe, line 87: The old interface returned just the number, and stored `bytes` directly on the `readVarInt`. Basically, the imported function (shared instance) had state containing a result of each function call :see_no_evil: 
Now it's a pure function, much better  :+1: 

See lib code: [old](https://github.com/bitcoinjs/varuint-bitcoin/blob/a05cb2cd32be43f6cab477704d60574066e3bb48/index.js#L21), [new](https://github.com/bitcoinjs/varuint-bitcoin/blob/d0b5f695480177cb25df7090215989a258da8413/src/cjs/index.cjs#L109)

packages/utxo-lib/src/bufferutils.ts
```diff
@@ -82,18 +84,20 @@ export function writeInt64LE(buffer: Buffer, value: number, offset: number) {
...
+
+    if (numberValue === null) throw new Error(OUT_OF_RANGE_ERROR);
```
Lemonexe, line 89: They also added `bigint` support, so the lib fn no longer throws error when out of range, but instead returns null number (and non-null bigint number).
In this PR I wanted to **keep the same functionality**, not refactor our `utxo-lib` to support bigints (I don't even know if it's needed), so the error is thrown.

jest.config.base.js
```diff
@@ -47,5 +47,6 @@ module.exports = {
         '^bcrypto/lib/(.*)$': 'bcrypto/lib/$1-browser',
         // Enforce usage of CommonJS version of uuid because ESM version is not working in Jest
         '^uuid$': require.resolve('uuid'), // https://stackoverflow.com/questions/73203367/jest-syntaxerror-unexpected-token-export-with-uuid-library
+        '^uint8array-tools$': require.resolve('uint8array-tools'), // same case as in uuid
```
Lemonexe: This lib has been added as transitive dependency of `varuint-bitcoin`.
I checked [the code](https://github.com/bitcoinjs/uint8array-tools) that it contains nothing sus.
But like `uuid`, it has misconfigured CJS vs MJS exports. This has to be done so that jest can resolve the import :see_no_evil: 

:information_source: :bulb: FIY, this is only necessary for test suites using `jsdom` env. The `utxo-lib` test suite has `node` env, and it can handle the import fine, so its test passes even without this fix, but it would fail in _other packages_, where `utxo-lib` package is indirectly used

jest.config.native.js
```diff
@@ -44,5 +44,4 @@ module.exports = {
         '<rootDir>/../../suite-native/test-utils/src/expoMock.js',
         '<rootDir>/../../suite-native/test-utils/src/walletSdkMock.js',
     ],
-    moduleNameMapper: { '^uuid$': require.resolve('uuid') },
```
Lemonexe, line 47: See above in this file. We are importing `moduleNameMapper` from `baseConfig`, but overwriting it.
The only difference is, that here it is missing `bcrypto`. I tried WIP

packages/suite/jest.config.js
```diff
@@ -31,9 +33,7 @@ module.exports = {
         '^src/(.+)': '<rootDir>/src/$1',
         '\\.(mp4)$': '<rootDir>/__mocks__/import-mp4.js',
         '\\.(svg)$': '<rootDir>/__mocks__/import-svg.js',
-        uuid: require.resolve('uuid'), // https://stackoverflow.com/questions/73203367/jest-syntaxerror-unexpected-token-export-with-uuid-library
-        // Enforce usage of JS version of bcrypto in tests because on CI we don't build native modules because it's slowing yarn install
-        '^bcrypto/lib/(.*)$': 'bcrypto/lib/$1-browser',
```
Lemonexe, line 36: This repeats what is in `baseConfig`, and now I needed to also add the `uint8array-tools` everywhere. So let's keep it DRY.

Lemonexe:
Concerning socket-security:

- `read-binary-file-arch@1.0.6`, a new a transitive dep of electron-builder, [code seems legit](https://github.com/samuelmaddock/read-binary-file-arch/blob/37156c22f07966dd132be84797fb7c9eb0a4cebe/index.js#L66)
- `@electron/rebuild@3.6.0` is a new transitive dep directly from `@electron`, so I'd trust that?
- `builder-util@25.0.3` is an update of existing dep from `electron-builder`
  - owned by the [same authors](https://www.npmjs.com/package/builder-util) as [electron-builder](https://www.npmjs.com/package/electron-builder), so I'd trust that?

Lemonexe:
@SocketSecurity ignore npm/read-binary-file-arch@1.0.6 npm/@electron/rebuild@3.6.0 npm/builder-util@25.0.3

Discussed with @komret , see comment above with explanation

shell.nix
```diff
@@ -8,7 +8,7 @@ with import
 
 let
   # unstable packages
-  electron = electron_30; # use the same version as defined in packages/suite-desktop/package.json
+  electron = electron; # use the same version as defined in packages/suite-desktop/package.json. Note that the 31 version is named just "electron"
```
Lemonexe: Not entirely sure if we want to use it, if it doesn't have specific version... Could it break in future?

komret (approved):

shell.nix
szymonlesisz: i think it breaks desktop builds on nixos, but 31.4 [should be there](https://github.com/NixOS/nixpkgs/commit/98636237874295db2a5b8b46d784e74a783d74c5) already

szymonlesisz (commented):
update of `varuint-bitcoin` looks ok :+1:

shell.nix
Lemonexe: Well, there is `electron-chromedriver_31` listed, but not `electron_31`. [They did add it](https://search.nixos.org/packages?channel=24.05&show=electron&from=0&size=50&sort=relevance&type=packages&query=electron), but only named as `electron`. At first I thought it's no problem, but you're right, it will break desktop builds when they bump to 32 at nixos repository.

So I'm gonna drop the electron bump from this PR and finish that later. I'll write an email to the maintainers to ask about it :slightly_smiling_face:

PR #14195, opened 2024-09-06
feat(mobile): update Metro - 3x faster JS build

Nodonisko:
## Description

This Metro version introduces new much faster resolver. Our build now takes around ~20s on my Mac compared to ~60s before.

## Related Issue

Resolve


vytick (approved):
LGTM 🚀 

I have not measured improvements, but it builds and works as expected

PR #14202, opened 2024-09-09
chore: update socks-proxy-agent from 6.1.1 to 8.0.4

### LINKED ISSUE #14179: update socks-proxy-agent
we are two major versions behind 
```
 6.1.1 -> 8.0.4
```

if you could @karliatto please update it when you find some spare time 🙏

karliatto:
## Description
update socks-proxy-agent from 6.1.1 to 8.0.4

* https://github.com/trezor/trezor-suite/pull/14202/commits/69ed9d5ade051c666e17f2a731df9f5b5b398108#diff-49f49f01ad60f5c7fc2414cb881d318ffe90f54838287ee5c847e6981fa12162L28 - Define type `SocksProxyAgentOptions` in order to keep the same API that we had before and not introduce breaking changes.
* https://github.com/trezor/trezor-suite/pull/14202/commits/933fab2e08a9af3ca66a5003df66b884fe5b6399#diff-fe77c9995deb7c750724b7d0de5b926f9942a35be2b8d3ae54eb2255a7aee63bR144 - transforming the previous API to the new one to work with new version o `SocksProxyAgent`.
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14179

### DIFF
--- a/packages/blockchain-link-types/package.json
+++ b/packages/blockchain-link-types/package.json
@@ -22,8 +22,7 @@
     "dependencies": {
         "@solana/web3.js": "^1.95.0",
         "@trezor/type-utils": "workspace:*",
-        "@trezor/utxo-lib": "workspace:*",
-        "socks-proxy-agent": "8.0.4"
+        "@trezor/utxo-lib": "workspace:*"
     },
     "devDependencies": {
         "tsx": "^4.16.3"
--- a/packages/blockchain-link-types/package.json
+++ b/packages/blockchain-link-types/package.json
@@ -23,7 +23,7 @@
         "@solana/web3.js": "^1.95.0",
         "@trezor/type-utils": "workspace:*",
         "@trezor/utxo-lib": "workspace:*",
-        "socks-proxy-agent": "6.1.1"
+        "socks-proxy-agent": "8.0.4"
     },
     "devDependencies": {
         "tsx": "^4.16.3"
--- a/packages/blockchain-link-types/src/common.ts
+++ b/packages/blockchain-link-types/src/common.ts
@@ -1,4 +1,6 @@
-import type { SocksProxyAgentOptions } from 'socks-proxy-agent';
+import type tls from 'tls';
+import type { Url } from 'url';
+import type { SocksProxy } from 'socks';
 
 import type { Transaction as BlockbookTransaction, VinVout } from './blockbook';
 import type {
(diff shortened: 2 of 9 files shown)


karliatto:
@marekrjpolak It would be nice if you have a look at this PR. I have made changes keeping in mind not to introduce breaking changes in connect API. Let me know what you think.

mroz22 (approved):
looks ok, did not test

marekrjpolak (changes requested):
I repeatedly wasn't able to do a basic all seed discovery with Tor turned on. On current develop it was ok.

Also, do you know why is it necessary to pass the uri to the new SocksProxyAgent constructor as a string instead of object as noted [here](https://github.com/TooTallNate/proxy-agents/blob/main/packages/socks-proxy-agent/CHANGELOG.md#upgrading-from-5x-to-6x)? I've tried it and it required more fields than we have and than is shown in the example, but I have no idea why.

karliatto:
@marekrjpolak I probably broke it by trying to satisfy the types but I should work now with this latest change https://github.com/trezor/trezor-suite/pull/14202/commits/eb2eeee6dc5a6359b40410b36707575bde7b6f2c

As you mentioned using the other way it requires some fields that we do not use.  But they way I am using is the one they use in their tests https://github.com/TooTallNate/proxy-agents/blob/main/packages/socks-proxy-agent/test/test.ts#L34 and it should work now.

marekrjpolak (approved):
Ok, I've checked basic Ehtereum (multi-identity) discovery, Electrum backend on onion address and CJ discovery and all seems to be working. :+1: Still, I'm a bit paranoid whether the identities are correctly separated which is kinda hard to verify. But it's probably fine.

karliatto:
>  Still, I'm a bit paranoid whether the identities are correctly separated which is kinda hard to verify. But it's probably fine.


It is hard to verify in the complex context of suite, but it is quite straight forward using the simple node script below while Tor from suite running:

```javascript
import https from 'https';
import { SocksProxyAgent } from 'socks-proxy-agent';

const socksServerUrl = new URL(`socks://127.0.0.1:36133`);
socksServerUrl.username = 'user';
socksServerUrl.password = 'password';


const agent = new SocksProxyAgent(socksServerUrl);

https.get('https://ipinfo.io', { agent }, (res) => {
	console.log(res.headers);
	res.pipe(process.stdout);
});
```

With that script if you change user or password get different IP in response.

Issue #13780, opened 2024-08-13
Node bridge third party wallets


bosomt:
**Describe the bug**

Issue for tracking of support of new Node bridge 3.0 by third party standalone wallets
Developers needs to be contacted or we can implement it on our side to make new bridge wallet for all wallets that support old Trezor Bridge 2.0

https://www.notion.so/satoshilabs/Test-NodeBridge-with-Third-party-wallets-53cc2e97be974ed6b7356bc663e0b053?pvs=4

| Wallet | Does it work | Download Link |
| ------------- | ------------- | ------------- |
| Electrum | ❌ | https://electrum.org/#download |
| Sparrow | ❌ | https://sparrowwallet.com/ |
| Frame | ❌ | https://frame.sh/ |
| Monero GUI | ❌ | https://www.getmonero.org/it/downloads/ |
| Feather Wallet | ❌ | https://featherwallet.org/download/ |
| Specter	| ❌ | https://specter.solutions/desktop/index.html |
| Wasabi | ❌ |	https://wasabiwallet.io/ |
| Exodus | ❌ | https://www.exodus.com/ |

mroz22:
It should work without contacting 3rd parties.

Hannsek:
I'd much rather use [this](https://www.notion.so/satoshilabs/5f35444cbf544204a083a2310d539ce6?v=b3dbfe70a21c44a0baacec775244e253) dedicated table if needed.

mroz22:
<img width="727" alt="image" src="https://github.com/user-attachments/assets/ddab8920-3dcd-4cbd-842a-c5868bdb2dd2">

electrum works just fine for me. monerogui no worse then with the old bridge. could you provide any more information regarding to what doesn't work?

bosomt:
@mroz22 device was not detected at all in mentioned wallets, i will retest it again later with latest build of node bridge

bosomt:
- retested with latest node bridge from yesterday dev app build
- ✅  Elecrum wallet detect device
- ❌ Exodus does not detect device
- ✅  Monero GUI wallet detects device 
- ❌  Wasabi wallet does not detect device

mroz22:
This is interesting, could you be more specific what exactly does not work? maybe some logs would be helpful. Also I would be grateful if you could follow the same steps that did not work for you using the new bridge also using the old one. I am quite convinced that interfaces should be pretty much same now and that it should either work or fail for both of them. I just tested exodus and some basic pairing worked.

[...]

PR #14211, opened 2024-09-09
transport masked path

mroz22:
follow-up after #14126 and #14143 in the branch of quests that should result in having 0 differences between old and new bridge API. 

This time it is about masking the device serial number with a plain `${number}` placeholder, originally probably as a countermeasure against sniffing/tracking for Trezors by malicious software installed in users' computers.

In some very wild scenarios, this could also affect #13780
I wanted to test whether it still works with multiple devices connected but I apparently broke it recently see #14221

before: 
image

after: 
image

### DIFF
--- a/packages/transport-bridge/src/http.ts
+++ b/packages/transport-bridge/src/http.ts
@@ -12,7 +12,7 @@ import {
     RequestWithParams,
     Response,
 } from '@trezor/node-utils';
-import { Descriptor, Session } from '@trezor/transport/src/types';
+import { Descriptor, PathPublic, Session } from '@trezor/transport/src/types';
 import { validateProtocolMessage } from '@trezor/transport/src/utils/bridgeProtocolMessage';
 import { Log, arrayPartition, Throttler } from '@trezor/utils';
 import { AbstractApi } from '@trezor/transport/src/api/abstract';
@@ -36,11 +36,12 @@ const validateDescriptorsJSON: RequestHandler<JSON, Descriptor[]> = (request, re
 };
 
 const validateAcquireParams: ParamsValidatorHandler<{
-    path: string;
+    path: PathPublic;
     previous: Session | 'null';
 }> = (request, response, next) => {
     if (
         typeof request.params.path === 'string' &&
+        /^[1-9][0-9]*$/.test(request.params.path) &&
         typeof request.params.previous === 'string' &&
         /^\d+$|^null$/.test(request.params.previous)
     ) {
--- a/packages/transport/src/sessions/background.ts
+++ b/packages/transport/src/sessions/background.ts
@@ -22,11 +22,11 @@ import type {
     HandleMessageParams,
     HandleMessageResponse,
 } from './types';
-import type { Descriptor, Success } from '../types';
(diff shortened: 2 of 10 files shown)


packages/transport-bridge/src/http.ts
```diff
@@ -36,11 +36,12 @@ const validateDescriptorsJSON: RequestHandler<JSON, Descriptor[]> = (request, re
 };
 
 const validateAcquireParams: ParamsValidatorHandler<{
-    path: string;
+    path: PathPublic;
     previous: Session | 'null';
 }> = (request, response, next) => {
     if (
         typeof request.params.path === 'string' &&
+        Number.parseInt(request.params.path, 10) &&
```
marekrjpolak: Probably not a big deal, but this allows negative numbers or even hex numbers and disallows zero. The `previous` check below is much more strict (`\d+`).

packages/transport/src/sessions/background.ts
```diff
@@ -294,6 +313,12 @@ export class SessionsBackground extends TypedEmitter<{
         };
     }
 
+    private getInternal(pathPublic: PathPublic): PathInternal | undefined {
+        return Object.entries(this.pathInternalPathPublicMap).find(
+            ([_internal, pub]) => pub === pathPublic,
+        )?.[0];
```
marekrjpolak: Maybe a bit better readable?
```suggestion
        return Object.keys(this.pathInternalPathPublicMap).find(
            internal => this.pathInternalPathPublicMap[internal] === pathPublic,
        );
```

packages/transport/src/sessions/background.ts
```diff
@@ -219,8 +241,9 @@ export class SessionsBackground extends TypedEmitter<{
         return Promise.resolve(this.success({ descriptors: Object.values(this.descriptors) }));
     }
 
+    // todo: this is wrong, there may be 2 same sessions for different paths
     private getPathBySession({ session }: GetPathBySessionRequest) {
-        let path: string | undefined;
+        let path: PathInternal | undefined;
```
marekrjpolak: Better readable? But this selects the first descriptor with the given session, not the last one as before. :man_shrugging: 
```suggestion
        const path = Object.keys(this.descriptors).find(
            pathKey => this.descriptors[pathKey]?.session === session,
        );
```

packages/transport/src/sessions/background.ts
```diff
@@ -118,18 +122,27 @@ export class SessionsBackground extends TypedEmitter<{
...
         disconnectedDevices.forEach(d => {
-            delete this.descriptors[d.path];
+            delete this.descriptors[d];
+            delete this.pathInternalPathPublicMap[d];
         });
 
         payload.descriptors.forEach(d => {
+            if (!this.pathInternalPathPublicMap[d.path]) {
+                this.pathInternalPathPublicMap[d.path] = `${
+                    Object.keys(this.pathInternalPathPublicMap).length + 1
```
marekrjpolak: Incremental numbering could be safer, as discussed personally.

marekrjpolak (approved):

PR #14212, opened 2024-09-09
feat(repo): TS 5.5

Nodonisko:
## Related Issue

Resolve

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
         ]
     },
     "resolutions": {
-        "typescript": "5.3.3",
+        "typescript": "5.5.4",
         "react-native": "0.75.2",
         "prettier": "3.2.5",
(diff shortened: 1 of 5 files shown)


mroz22 (commented):
uf, these are type-box related. maybe @martykan could take a look. also I just noted that there is yet another version of type-box so maybe updating might help (blind shot here)

martykan:
Yeah unfortunately this is the part where I was doing some black magic to infer a type for "keyof enum". 
I will try to see if it's fixable

martykan:
I think we're good to go, should I rebase?

Nodonisko:
@martykan Thanks for fixing it!

Yes, rebase please.

martykan (approved):

PR #14215, opened 2024-09-09
Passphrase flow checks device for empty and not for visible accounts

### LINKED ISSUE #14208: Adding an empty passphrase don't prompt me to check my passphrase
**Describe the bug**
When I add a new/empty passphrase, I should be prompted/shown that the passphrase is empty and if I want to continue and I should re-enter it but I am not on latest develop

**Info:**

-   Mobile Suite Version: 24.9.1 b415aa2
-   OS: Android

**How to reproduce**

1. Connect a device
2. Go to passphrase
3. Enter an empty passphrase
4. It goes straight to discovery (without confirming it is empty)

**Expected behavior**
I should be prompted that the passphrase is empty and I should re-enter it.

vytick:
- Passphrase flow previously checked accounts visibility, now it checks if account is empty instead.
- I also renamed `selectIsEmptyDevice` to `selectIsDiscoveredAccountlessDevice` because that is what it actually returns.

Resolve #14208

## Screenshots:
https://github.com/user-attachments/assets/300c2a20-f02b-4d33-b966-a8c13cbfa900

### DIFF
--- a/suite-common/wallet-core/src/accounts/accountsReducer.ts
+++ b/suite-common/wallet-core/src/accounts/accountsReducer.ts
@@ -12,6 +12,7 @@ import { formattedAccountTypeMap } from './accountsConstants';
 import {
     DeviceRootState,
     selectDevice,
+    selectDeviceState,
     selectHasOnlyPortfolioDevice,
 } from '../device/deviceReducer';
 import { deviceActions } from '../device/deviceActions';
@@ -270,6 +271,9 @@ export const selectVisibleNonEmptyDeviceAccountsByNetworkSymbol = (
         account => !account.empty || account.visible,
     );
 
+export const selectNonEmptyDeviceAccounts = (state: AccountsRootState & DeviceRootState) =>
+    selectDeviceAccounts(state).filter(account => !account.empty);
+
 export const selectAccountsByNetworkAndDeviceState = memoizeWithArgs(
     (state: AccountsRootState, deviceState: string, networkSymbol: NetworkSymbol) => {
         const accounts = selectAccounts(state);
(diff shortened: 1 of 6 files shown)


yanascz (approved):
Seems to work fine!

suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -397,9 +401,23 @@ export const selectIsEmptyDevice = (
...
         const hasOnlyPortfolioDevice = selectHasOnlyPortfolioDevice(state);
 
-        return isEmptyDevice && hasOnlyPortfolioDevice;
+        return isDiscoveredAccountlessDevice && hasOnlyPortfolioDevice;
+    },
+);
+
+// If we know that device is not empty, we can return true right away.
+// Otherwise we return null and wait until discovery finishes
+export const selectDeviceIsNotEmpty = memoize(
```
matejkriz: It returns `boolean` or `null`, right? I see no point of memoizing this. And I would change the name:
```suggestion
export const selectIsDeviceNotEmpty = (
```

nitpick: I find `if` conditions without brackets unnecessary harder to read, especially if the condition is longer and it's not just `if(isLoading) return null;` or something like that... But feel free to disagree :-)

suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -386,7 +390,7 @@ export const selectIsDeviceAccountless = (state: AccountsRootState & DeviceRootS
     pipe(selectVisibleDeviceAccounts(state), A.isEmpty);
 
 // Selected device has no accounts and no active discovery. It can be empty portfolio device.
-export const selectIsEmptyDevice = (
+export const selectIsDiscoveredAccountlessDevice = (
```
matejkriz: nitpick, feel free to ignore
```suggestion
export const selectIsDiscoveredDeviceAccountless = (
```

suite-common/wallet-core/src/accounts/accountsReducer.ts
vytick: I actually do feel the same and dont know, why I havent used those. I'll update and remove memoization

suite-common/wallet-core/src/accounts/accountsReducer.ts
vytick: e6791b5ea8

suite-common/wallet-core/src/accounts/accountsReducer.ts
vytick: e6791b5ea8

matejkriz (approved):

PR #14200, opened 2024-09-07
fix formatting of review txs value

ericvalcik:
## Description

Fix wrapping of Coin/Token symbol in the `TransactionReviewSummary.tsx`.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13603

## Screenshots:

before:
image

after changes:
image

PR #14216, opened 2024-09-09
fix(suite): fix symbol formatting of review txs value

### LINKED ISSUE #13603: Symbol wrapping
**Describe the bug**
Coin/Token symbol should not be wrapped in the middle of word

**Expected behavior**

Screenshot 2024-08-02 at 11 00 40

**Screenshots now**
This is an example from tx review modal but I think that it can occur everywhere, so it would be fine to fix it in some global component

Screenshot 2024-08-02 at 10 19 09
Screenshot 2024-08-02 at 10 18 26
Screenshot 2024-08-02 at 10 16 17
Screenshot 2024-08-02 at 9 55 06

ericvalcik:
## Description

Fix wrapping of Coin/Token symbol when reviewing a transaction. This is a follow up PR on #14200.

## Related Issue

Resolve #13603.

## Screenshots:

before:
image

after changes:
image

### DIFF
--- a/packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewSummary.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewSummary.tsx
@@ -70,7 +70,6 @@ const Headline = styled.div`
     font-size: 16px;
     font-weight: 600;
     margin-top: 20px;
-    word-break: break-all;
 `;
 
 const AccountWrapper = styled.div`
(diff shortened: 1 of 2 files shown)


tomasklim (dismissed):

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewSummary.tsx
```diff
@@ -69,7 +69,7 @@ const Headline = styled.div`
     font-size: 16px;
     font-weight: 600;
     margin-top: 20px;
-    word-break: break-all;
+    word-break: break-word;
```
tomasklim: I think it works same with `normal` as I proposed, or not?

tomasklim: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break `break-word` is deprecated

ericvalcik: Ahh, I see, well `work-break: normal` is the default, so I think removing the `word-break` entry entirely is the best approach? I pushed the changes

tomasklim (approved):
Thank you for your contribution. Please squash the commits into 1

PR #12122, opened 2024-04-24
feat(unstaking):  Instant unstake forecasting

EvgeniiVoznyuk:
## Description
Instant unstake forecasting is added to the eth unstaking modal

## Related Issue

Resolve [trezor#11402](https://github.com/trezor/trezor-suite/issues/11402)

## Screenshots:
image
image

PR #14220, opened 2024-09-09
Display warning if the address entered in Send Form is a contract address. Ethereum  mainnet only.

### LINKED ISSUE #7244: Warn about sending funds to a contract address
**Design:** 
https://www.figma.com/file/hZ22ckyTRlCYRLTNUbTdWO/Exploring-Accounts?node-id=4390%3A39787&t=VEeGefz3JzzqMRnu-0

**Spec:** 
As a user I want to be warned when I paste a token contract address to the recipient address field when attempting to send ether or other tokens out of my Ethereum account.

**Acceptance criteria**

- When I paste a token contract address to the “Recipient address” field, warning box will be displayed right under the “Recipients address” field
    - Each pasted string has to be verified against Blockbook which knows all existing contracts, each contract has the following information
api-blockbook
- Warning box contains the following information:
    - Red triangle warning icon on the left
    - “You are about to send to a contract address which could result in a loss of funds.” text next to the warning icon
    - The warning text is followed by “Lear more” link, which redirect to the knowledge base article
        - Article URL: **https://trezor.io/learn/a/sending-funds-to-a-contract-address**
    - Button “I understand” is located on the right side
        - When I click the button to acknowledge the warning, the whole warning box closes

enjojoy:
## Description

Implemented only on Ethereum mainnet accounts. 
If the user enters a contract address, he can proceed with sending only after pressing the "I understand the risk" button

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/7244

## Screenshots:

Screenshot 2024-12-16 at 15 01 13

### DIFF
--- a/packages/suite/src/components/wallet/InputError.tsx
+++ b/packages/suite/src/components/wallet/InputError.tsx
@@ -7,23 +7,23 @@ import { spacings } from '@trezor/theme';
 import { LearnMoreButton } from '../suite/LearnMoreButton';
 
 type ButtonProps = { onClick: MouseEventHandler<HTMLButtonElement>; text: string };
-type LinkProps = { url: Url };
 
 export type InputErrorProps = {
-    button?: ButtonProps | LinkProps;
+    buttonProps?: ButtonProps;
+    learnMoreUrl?: Url;
     message?: string;
 };
 
-export const InputError = ({ button, message }: InputErrorProps) => (
+export const InputError = ({ buttonProps, learnMoreUrl, message }: InputErrorProps) => (
     <Row gap={spacings.xs} justifyContent="space-between" flex="1">
-        <Paragraph>{message}</Paragraph>
-        {button &&
-            ('url' in button ? (
-                <LearnMoreButton url={button.url} />
-            ) : (
-                <Button size="tiny" variant="tertiary" onClick={button.onClick} textWrap={false}>
-                    {button.text}
-                </Button>
-            ))}
+        <Row gap={spacings.xs}>
+            <Paragraph>{message}</Paragraph>
+            {learnMoreUrl && <LearnMoreButton url={learnMoreUrl} />}
+        </Row>
+        {buttonProps?.text && (
+            <Button size="tiny" variant="tertiary" onClick={buttonProps.onClick}>
+                {buttonProps.text}
+            </Button>
+        )}
     </Row>
 );
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -2036,6 +2036,14 @@ export default defineMessages({
         defaultMessage: 'Unable to verify address history. Check that the address is correct.',
         id: 'TR_ETH_ADDRESS_CANT_VERIFY_HISTORY',
     },
+    TR_EVM_ADDRESS_IS_CONTRACT: {
+        defaultMessage: 'You are sending funds to a contract address.',
+        id: 'TR_EVM_ADDRESS_IS_CONTRACT',
+    },
+    TR_I_UNDERSTAND_THE_RISK: {
+        defaultMessage: 'I understand',
+        id: 'TR_I_UNDERSTAND_THE_RISK',
+    },
     TR_NEEDS_ATTENTION_BOOTLOADER: {
         defaultMessage: 'Trezor is in Bootloader mode.',
         id: 'TR_NEEDS_ATTENTION_BOOTLOADER',
--- a/packages/suite/src/views/wallet/send/Outputs/Address.tsx
+++ b/packages/suite/src/views/wallet/send/Outputs/Address.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 
 import { checkAddressCheckSum, toChecksumAddress } from 'web3-utils';
 import styled from 'styled-components';
@@ -16,11 +16,15 @@ import {
     isTaprootAddress,
     isBech32AddressUppercase,
     getInputState,
+    checkIsAddressNotUsedNotChecksummed,
 } from '@suite-common/wallet-utils';
 import { getNetworkSymbolForProtocol } from '@suite-common/suite-utils';
-import { HELP_CENTER_EVM_ADDRESS_CHECKSUM } from '@trezor/urls';
 import { spacings } from '@trezor/theme';
 import { CoinLogo } from '@trezor/product-components';
+import {
+    HELP_CENTER_EVM_ADDRESS_CHECKSUM,
+    HELP_CENTER_EVM_SEND_TO_CONTRACT_URL,
+} from '@trezor/urls';
 
 import { scanOrRequestSendFormThunk } from 'src/actions/wallet/send/sendFormThunks';
 import { useSendFormContext } from 'src/hooks/wallet';
@@ -59,6 +63,7 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
     const [addressDeprecatedUrl, setAddressDeprecatedUrl] =
         useState<ReturnType<typeof isAddressDeprecated>>(undefined);
     const [hasAddressChecksummed, setHasAddressChecksummed] = useState<boolean | undefined>();
+    const contractAddressWarningDismissed = useRef(false);
     const dispatch = useDispatch();
     const { device } = useDevice();
     const {
@@ -72,6 +77,7 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
         metadataEnabled,
         watch,
         setDraftSaveRequest,
+        trigger,
     } = useSendFormContext();
     const { translationString } = useTranslation();
     const { descriptor, networkType, symbol } = account;
@@ -89,9 +95,14 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
     const options = getDefaultValue('options', []);
     const broadcastEnabled = options.includes('broadcast');
     const isOnline = useSelector(state => state.suite.online);
+
+    useEffect(() => {
+        contractAddressWarningDismissed.current = false;
+    }, [address]);
+
     const getInputErrorState = () => {
-        if (hasAddressChecksummed) {
-            return 'primary';
+        if (hasAddressChecksummed && !addressError) {
+            return 'default';
         }
         if (addressError) {
             return getInputState(addressError);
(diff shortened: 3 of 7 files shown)


MiroslavProchazka:
Nit - copy adjustment: 

"“This is a contract address. Are you sure you want to proceed with sending to this address?”"

wdyt @Hermez-cz ?

enjojoy:
> Nit - copy adjustment:
> 
> "“This is a contract address. Are you sure you want to proceed with sending to this address?”"
> 
> wdyt @Hermez-cz ?

I discussed it with Ben and he also proposed the "This". Also I will add a Learn More button with a link to this article https://trezor.io/support/a/where-is-my-ethereum

enjojoy:
@MiroslavProchazka  I updated the screenshot

tomasklim:
Please consider smart contract wallets (Account Abstraction). e.g. https://safe.global/wallet
More and more often people will intentionally want to send it there. Probably just inform Ben about it (maybe the warning should not be that strong) and update KB https://www.halborn.com/blog/post/what-is-a-smart-contract-wallet

packages/suite/src/components/wallet/InputError.tsx
```diff
@@ -28,21 +29,39 @@ type LinkProps = { url: Url };
...
+                        type="label"
+                    >
+                        {translationString('TR_LEARN_MORE')}
+                    </Link>
+                )}
+            </ContentWrapper>
+            {button &&
+                ('url' in button ? (
+                    <LearnMoreButton url={button.url} />
+                ) : (
```
tomasklim: This is used just in one edge-case in the whole app. Please unify it that all occurences of learn more "button" are text url.

tomasklim: Could you please refactor this component that in case
- `button` prop is available, render button. I would also like better name than `button`. It is not button, it is just wrapper of `text` and `onClick`. `buttonProps` would be better but maybe there is even better naming for that.
- `url` prop is available, render this learn more button

this `'url' in button` is mega hack and not a clean code, please fix it 🙏

packages/suite/src/support/messages.ts
```diff
@@ -2131,6 +2131,15 @@ export default defineMessages({
         defaultMessage: 'Unable to verify address history. Check that the address is correct.',
         id: 'TR_ETH_ADDRESS_CANT_VERIFY_HISTORY',
     },
+    TR_ETH_ADDRESS_IS_CONTRACT: {
```
tomasklim: `EVM` instead of `ETH`

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -185,6 +190,14 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                         }),
                     text: translationString('TR_CONVERT_TO_LOWERCASE'),
                 };
+            case 'contract':
+                return {
+                    onClick: () => {
+                        contractAddressWarningDismissed = true;
```
tomasklim: so this should work just once per send form session?

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -185,6 +190,14 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                        trigger(inputName);
```
tomasklim: what does this do?

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -260,6 +273,24 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
                 }
             },
+            contract: async (address: string) => {
+                if (networkType === 'ethereum' && !contractAddressWarningDismissed) {
+                    if (isOnline) {
+                        const params = {
+                            descriptor: address,
+                            coin: symbol,
+                        };
+                        const result = await TrezorConnect.getAccountInfo(params);
```
tomasklim: For evm accounts we now do 2 same requests. How about merging contract and checksum into one validation? `evmChecks`, `validateEvmAddress`, `evmAddressValidator`,.. checksum being first

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -303,7 +334,15 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
         if (addressError) {
             return (
-                <InputError message={addressError.message} button={getValidationButtonProps()} />
+                <InputError
+                    message={addressError.message}
+                    learnMoreAsTextUrl={
+                        addressError.type === 'contract'
+                            ? HELP_CENTER_EVM_SEND_TO_CONTRACT_URL
+                            : undefined
+                    }
```
tomasklim: this is not nice.

instead please improve the `InputError` component and also `button={getValidationButtonProps()}`
can be used somehow like this
```
{...getInputErrorProps()}
```

tomasklim (dismissed):

packages/suite/src/components/wallet/InputError.tsx
enjojoy: I agree, will refactor

packages/suite/src/views/wallet/send/Outputs/Address.tsx
enjojoy: No, considering the fact Address component is re-rendered every time the address input value changes, it works every time a new address is entered.


https://github.com/user-attachments/assets/9b0f5949-3bf1-4454-8d2c-7c9f023d1a17

packages/suite/src/views/wallet/send/Outputs/Address.tsx
enjojoy: trigger starts re-validation. same as setValue, but without changing the value. 

https://react-hook-form.com/docs/useform/trigger

enjojoy: it's needed to remove the contract warning

packages/suite/src/views/wallet/send/Outputs/Address.tsx
enjojoy: I also thought about it but I was worried that the validation rule would be bigger and worse readable. 

But ok I agree and I will refactor it

packages/suite/src/views/wallet/send/Outputs/Address.tsx
tomasklim: Hm, interesting, but that's not clean code. Could you please come up with a solution which will also replace
```
    const [hasAddressChecksummed, setHasAddressChecksummed] = useState<boolean | undefined>();
```
there can be one state for both `evmChecks` (object)

packages/suite/src/views/wallet/send/Outputs/Address.tsx
tomasklim: camel case please

packages/suite/src/views/wallet/send/Outputs/Address.tsx
enjojoy: I think that due to the nature of the errors it would be better to later group them into objects based on their logic

These 2 errors require different closure, checksum is dependent on the actual state of the address and can't be affected by the user's behavior, also when the address is changed I want to display bottom text and it works well with useState variable. 

Contract on the other side requires an action from the user and can be dismissed and I don't want it to be shown only based on the state of the address (the address can be the same but state of the error can be different)

So I don't want to put these errors together in an object because it requires picking either state based handling or normal variable, and the other error logic breaks 

This is why I would group some other errors later based on their logic

If I've missed something or misunderstood your request please do let me know, but this is my conclusion after approx. a day playing with the logic of these errors

packages/suite/src/views/wallet/send/Outputs/Address.tsx
enjojoy: Did this in the fixup commit

packages/suite/src/components/wallet/InputError.tsx
```diff
@@ -24,25 +23,34 @@ const ContentWrapper = styled.div`
...
-            ('url' in button ? (
-                <LearnMoreButton url={button.url} />
-            ) : (
+export const InputError = ({ buttonProps, message, learnMoreUrl }: InputErrorProps) => {
+    const { translationString } = useTranslation();
+
+    return (
+        <Wrapper>
+            <ContentWrapper>
+                {message}{' '}
```
tomasklim: `{' '}` should be there only in case there is `Link`

comment below is not really needed

packages/suite/src/views/wallet/send/Outputs/Address.tsx
tomasklim: Okay, makes sense. However, I still don't like "let" variable here and be dependant on rerender

enjojoy: @tomasklim I found a way to achieve the same behavior by using useRef, it's in the fixup

enjojoy:
I updated the screenshot with a new version of the shorter and less implicit version after discussing it with Ben and reminded him about the KB article update @tomasklim

jvaclavik:
Btw I think "I understand the risk" shouldn't be a button. Because button evokes you might be redirected somewhere else (different page, modal or something else). 
This is pretty clear use case for checkbox and it shouldn't be possible to send form without ticking it.

@Hermez-cz WDYT?

Hermez-cz:
> Btw I think "I understand the risk" shouldn't be a button. Because button evokes you might be redirected somewhere else (different page, modal or something else). This is pretty clear use case for checkbox and it shouldn't be possible to send form without ticking it.
> 
> @Hermez-cz WDYT?

Hi, the design linked in https://github.com/trezor/trezor-suite/issues/7244 is already quite old, but the button would make sense in there as part of the warning banner to dismiss it. In the current implementation of an error message, the button really feels odd. Checkbox makes better logical sense, but might make the interface fuzzy and the flow not so clear. Will have to check with the designer.

tomasklim:
@enjojoy 🚨🚨📢 does it work for Polygon and BNB?

enjojoy:
> @enjojoy 🚨🚨📢 does it work for Polygon and BNB?

It works on Polygon but not on BNB, because BNB smart contracts don't provide any contract info for some reason, it's the only way I can know if it's a smart contract address

enjojoy:
❌ Blocked - no contractInfo on BNB and partially on POL (missing internal txs)

https://satoshilabs.slack.com/archives/C03SRP8PSS2/p1727182740567229?thread_ts=1725683114.300189&cid=C03SRP8PSS2
https://satoshilabs.slack.com/archives/C03SRP8PSS2/p1727697251440999

tomasklim:
~Maybe we can use rpc calls to node using logic added here https://github.com/trezor/trezor-suite/pull/12122/commits~ we cant, we have `eth_call` not `eth_getAccount`

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -158,39 +168,56 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+        buttonProps?: InputErrorProps['buttonProps'];
+    } => {
         switch (addressError?.type) {
             case 'deprecated':
-                if (addressDeprecatedUrl) {
+                return {
+                    learnMoreUrl: addressDeprecatedUrl ? URLS[addressDeprecatedUrl] : undefined,
+                };
+            case 'evmchecks':
+                if (!checkAddressCheckSum(address)) {
```
martykan, line 182: I get a deprecation warning here, maybe it's better to fix it now rather than later:

> Will be removed in next release. Please use `web3-validator` package instead.

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -230,38 +257,52 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+            evmchecks: async (address: string) => {
+                if (networkType === 'ethereum') {
                     if (isOnline) {
                         const params = {
                             descriptor: address,
                             coin: symbol,
                         };
-                        // 1. If the address is used but unchecksummed, then Suite will automatically
-                        // convert the address to the correct checksummed form and inform the user as described in the OP.
                         const result = await TrezorConnect.getAccountInfo(params);
```
martykan: What if `result.success` is false? Before it would fall through to `TR_ETH_ADDRESS_CANT_VERIFY_HISTORY`, now nothing happens

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -230,38 +257,52 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                                }
                             }
-
-                            // 2. If the address is not checksummed at all and not found in blockbook
-                            // offer to checksum it with a button.
-                            if (!hasHistory && address === address.toLowerCase()) {
-                                return translationString('TR_ETH_ADDRESS_NOT_USED_NOT_CHECKSUMMED');
+                        }
+                        //2. Check if address is a contract address
+                        if (!contractAddressWarningDismissed.current && symbol === 'eth') {
```
martykan: `symbol === 'eth'` limits this to Ethereum mainnet, is this intended or do we want to support other EVMs as well?

packages/suite/src/views/wallet/send/Outputs/Address.tsx
martykan: Note - doesn't have to be in this PR

packages/suite/src/views/wallet/send/Outputs/Address.tsx
enjojoy: Yes, it's intended, now we agreed to do it only on mainnet.

packages/suite/src/views/wallet/send/Outputs/Address.tsx
enjojoy: I refactored this part in this fixup https://github.com/trezor/trezor-suite/pull/14220/commits/bd7d4fe21e2e92580bc47f4c746bb2394ee76537 , what do you think ?

packages/suite/src/views/wallet/send/Outputs/Address.tsx
enjojoy: Created an issue for this https://github.com/trezor/trezor-suite/issues/15989

martykan (approved):
LGTM

enjojoy:
@qa To test it find any contract address on etherscan (or somewhere else), and copy paste it into send form. This feature is implemented only on Ethereum mainnet.

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -230,38 +258,46 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                                setHasAddressChecksummed,
+                            );
+                        if (checksumAndUsageValidationResult) {
+                            return translationString('TR_ETH_ADDRESS_NOT_USED_NOT_CHECKSUMMED');
                         }
                     }
 
-                    return translationString('TR_ETH_ADDRESS_CANT_VERIFY_HISTORY');
+                    //2. Check if address is a contract address
+                    if (!contractAddressWarningDismissed.current && symbol === 'eth') {
```
tomasklim, line 295: @enjojoy can you please add also `thol` and `tsep`? Please take this array ["eth", "tsep", "thol"].includes(symbol) to some variable and put there comment. Thank you. 

`TODO: only for synced Blockbooks, wait for syncing internal tx to add others`

enjojoy: here https://github.com/trezor/trezor-suite/pull/16130

PR #14228, opened 2024-09-09
fix(suite-native): after coin enabling is enabled, show relevant normal accounts

### LINKED ISSUE #14225: View only wallet with no accounts not showing empty accounts after turning ON Coin enabling
**Describe the bug**
If I enable `View only` for an empty wallet and then turn ON the `Coin enable` feature flag, the discovery will not run for the enabled coins even though it should show the first empty accounts.

**Info:**

Mobile Suite Version: 24.9.1 https://github.com/trezor/trezor-suite/commit/b415aa2c5f4d8c5c767055211d4e22c85865c05c
OS: Android

**How to reproduce**

1. Connect an empty wallet
2. Enable view only for that wallet
3. Turn on the `Coin enable` flag and enable some coins
4. Kill the app
5. Start the app again
6. The wallet should show empty accounts for the enabled coins but it shows just "Empty wallet"

**Expected behavior**
Discovery should be done for enabled coins even when they are empty

vytick:
When checking for discovery updates, make sure that first normal account is visible for enabled networks when coin enabling is active.

This might be needed in case user has View only device from before coin enabling was active in such case the first normal account can be invisible. We need to make it visible.

## Related Issue

Resolve #14225

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -17,6 +17,7 @@ import {
     selectDeviceAccountsForNetworkSymbolAndAccountType,
     selectDeviceState,
     disableAccountsThunk,
+    selectFirstNormalAccountForNetworkSymbol,
 } from '@suite-common/wallet-core';
 import { selectIsAccountAlreadyDiscovered } from '@suite-native/accounts';
 import TrezorConnect from '@trezor/connect';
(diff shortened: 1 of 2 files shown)


matejkriz (commented):
Have you considered migration instead? It can happen only once and only for those that had some view only devices on older version.

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -694,7 +698,28 @@ export const discoveryCheckThunk = createThunk(
...
+        // This might be needed in case user has View only device from before coin enabling was active
+        // in such case the first normal account can be invisible. We need to make it visible.
+        if (isCoinEnablingActive) {
+            const enabledDiscoveryNetworkSymbols = selectEnabledDiscoveryNetworkSymbols(getState());
+            enabledDiscoveryNetworkSymbols.forEach(networkSymbol => {
+                const firstNormalAccount = selectDeviceAccountsForNetworkSymbolAndAccountType(
+                    getState(),
+                    networkSymbol,
+                    NORMAL_ACCOUNT_TYPE,
+                ).filter(account => account.index === 0 && !account.visible);
```
matejkriz: what about making a memoized selector from it?

vytick: Good idea 5e9aa79893

matejkriz (approved):

matejkriz:
For a reference, the migration is not used because it has to happen when device is connected, not on app start.

Issue #13019, opened 2024-06-21
App is unusuable(or very slow) with ETH account with large number of tokens


Nodonisko:
**Describe the bug**
Import this account into app: `0xdac17f958d2ee523a2206206994597c13d831ec7` and app will be freezing all the time.

**Info:**

-   Mobile Suite Version: version/commit
-   OS: name, version

**How to reproduce**

1. Click on '....'
1. Scroll down to '....'
1. See error

**Expected behavior**

**Screenshots**

**Additional context**

Nodonisko:
There are still two major issues left:

1. When ETH asset or account is clicked and modal is opened, if there is more than few tokens it will freeze app for few seconds. It's because everything is rendered at once. This component should use FlashList to show tokens so only visible will be rendered. It will require some structural changes to components so it could be used in FlashList.
2. Homescreen Assets component is rerendered very often during discovery/app start. That's probably because loading of fiat rates for tokens and every updated fiat rate will trigger rerender. Solution here is to separate fiat values from main assets selector a create atomic component just for fiat rate that will be use direct selector for example something like `selectAssetFiatValue(state, 'eth')` so it will update minimal amount of components (other solutions possible but this seems easier to me).

Both of these issue should be relative quick to do.

STew790:
**QA OK**
App is usable even with large amount of tokens and values.

**Info**
24.10.1 d87ce74

**Additional notes**
Just additional notes for @Nodonisko 
When there is only the account you shared with me (and no device connected) the app is a little faster then with the device connected.
With the device connected it is like this:
https://github.com/user-attachments/assets/d4804c81-9016-4e23-971e-ae79044e2d3d

And this is with the really big account shared above:

https://github.com/user-attachments/assets/2fdc1139-be05-48fc-adcb-6c6421506c90

PR #14234, opened 2024-09-09
fix(mobile): fix app perf for big number of tokens

Nodonisko:
## Description

1. First major improvement is batching of fiat rates updates => this is big issue for apps with bigger amount of tokens/accounts because every dispatched action has quite big overhead. Now most of updates happen in one call which speeds up app responsiveness by several seconds on start, on periodic updates etc. Also it speeds up account import even more.
2. Fix issue with tokens definitions check and memoization + use WeakMap to prevent memory leaks
3. Refactor whole logic of assets selectors so minimum amount of components is rerendered
4. Use `tokensFilter` of graph to remove all unknown tokens from graph calculations => some noticeable speedup for graph with big accounts
5. Make token selectors work correctly so no more `+1 token` everywhere

## Related Issue

Resolve second point of #13019


suite-native/assets/src/assetsSelectors.ts
```diff
@@ -73,25 +82,66 @@ export const selectDeviceAssetsWithBalances = (state: AssetsRootState & DeviceRo
...
+    let totalFiatBalance = 0;
+
+    const assets = deviceNetworksWithAssets.map((networkSymbol: NetworkSymbol) => {
         const networkAccounts = accountsWithFiatBalance.filter(
             account => account.symbol === networkSymbol,
         );
         const assetBalance = networkAccounts.reduce(
-            (sum, { cryptoValue }) => sum.plus(cryptoValue),
-            new BigNumber(0),
+            (sum, { cryptoValue }) => sum + Number(cryptoValue),
```
matejkriz, line 92: For fiat value it's ok to loose the precision, but not sure about crypto value 🤔

suite-native/assets/src/assetsSelectors.ts
```diff
@@ -73,25 +82,66 @@ export const selectDeviceAssetsWithBalances = (state: AssetsRootState & DeviceRo
...
             null as number | null,
         );
 
+        totalFiatBalance += fiatBalance ?? 0;
+
         const asset: AssetType = {
             symbol: networkSymbol,
-            assetBalance: assetBalance.toFixed(),
+            // For assets we should always only 8 decimals to save space
+            assetBalance: assetBalance.toFixed(8),
```
matejkriz, line 105: What space to save? Ethereum has 18 decimals, solana 9...

suite-native/assets/src/components/AssetItem.tsx
```diff
@@ -1,13 +1,15 @@
-import { memo } from 'react';
+import React, { memo } from 'react';
```
matejkriz: nitpick: mixing `React.memo` and `memo` in this file

suite-native/assets/src/assetsSelectors.ts
matejkriz: Now I see that it's already rounded in AssetItem to 8 decimals and it's used only on the dashboard. It might be nice to define some constant for it and use it on both places so the connection is clear, but it's a nitpick.

suite-native/assets/src/assetsSelectors.ts
matejkriz: If it's rounded to 8 decimals anyway, I guess it's fine.

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraph.tsx
```diff
@@ -55,7 +49,7 @@ export const PortfolioGraph = forwardRef<PortfolioGraphRef>((_props, ref) => {
 
     return (
         <VStack spacing="large" testID="@home/portfolio/graph">
-            {isAnyMainnetAccountPresent ? <PortfolioGraphHeader /> : null}
+            <PortfolioGraphHeader />
```
matejkriz: Is removal of `isAnyMainnetAccountPresent` intentional?

matejkriz (approved):
👏 Really nice refactor of the assets and batching fiat rates 👍  Great idea with removing unknown tokens from graph calculations! Use of getUntracked is black magic ⛧ 

Just double check the `isAnyMainnetAccountPresent` removal, please.

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraph.tsx
Nodonisko: definitely not, probably bad rebase

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraph.tsx
```diff
@@ -55,7 +56,7 @@ export const PortfolioGraph = forwardRef<PortfolioGraphRef>((_props, ref) => {
 
     return (
         <VStack spacing="large" testID="@home/portfolio/graph">
-            {isAnyMainnetAccountPresent ? <PortfolioGraphHeader /> : null}
+            {isAnyMainnetAccountPresent ? <PortfolioGraphHeader /> : null}{' '}
```
matejkriz: `{' '}` was intentional?
```suggestion
            {isAnyMainnetAccountPresent ? <PortfolioGraphHeader /> : null}
```

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraph.tsx
matejkriz: 👌 check the extra `{' '}`

suite-common/token-definitions/src/tokenDefinitionsSelectors.ts
```diff
@@ -23,10 +23,13 @@ export const selectNftDefinitions = (
 ) => state.tokenDefinitions?.[networkSymbol]?.nft;
 
 export const selectCoinDefinition = (
-    state: TokenDefinitionsRootState,
+    stateTracked: TokenDefinitionsRootState,
     networkSymbol: NetworkSymbol,
     contractAddress: TokenAddress,
 ) => {
+    // We need to use getUntracked here otherwise this could interfere with memoization in isTokenDefinitionKnown
+    const state = getUntracked(stateTracked) ?? stateTracked;
```
vytick, line 31: In docs they say that: ```[Notes] This function is for debugging purpose. It’s not supposed to be used in production and it’s subject to change.```

might that cause any problem?

suite-native/accounts/src/components/AccountListItem.tsx
```diff
@@ -24,6 +25,20 @@ export type AccountListItemProps = {
...
+const TokenBadge = React.memo(({ accountKey }: { accountKey: AccountKey }) => {
+    const numberOfTokens = useSelector((state: TokensRootState) =>
+        selectNumberOfAccountTokensWithFiatRates(state, accountKey),
+    );
+
+    return (
+        <Badge
+            elevation="1"
+            size="small"
+            label={<Translation id="accountList.numberOfTokens" values={{ numberOfTokens }} />}
```
vytick, line 37: Should this return `+1 tokens` instead of `+1 token` ? for one token?

suite-native/assets/src/assetsSelectors.ts
```diff
@@ -39,10 +42,10 @@ export const selectVisibleDeviceAccountsKeysByNetworkSymbol = (
     return accounts.map(account => account.key);
 };
 
-export const selectDeviceAssetsWithBalances = (state: AssetsRootState & DeviceRootState) => {
+export const selectDeviceNetworksWithAssets = (state: AssetsRootState) => {
```
vytick, line 45: memoization might help here, maybe?

suite-common/token-definitions/src/tokenDefinitionsSelectors.ts
Nodonisko: I don't think so, since we are not upgrading that lib anyway. Also this sadly don't have any other reasonable solution

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraph.tsx
Nodonisko: Idk, maybe @yanascz knows?

yanascz: The empty text value should not be necessary, if I remember correctly. The idea is to completely remove the header if there are only testnet accounts available.

suite-native/accounts/src/components/AccountListItem.tsx
Nodonisko: Good point but I don't want to do that in this PR

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraph.tsx
matejkriz: @Nodonisko It is added in this PR, not by @yanascz. And it also triggers `Error: Text strings must be rendered within a <Text> component.` So I'm pretty sure it was not intentional...

Nodonisko: Sorry I was confused I thought it was there and I removed it.

matejkriz (changes requested):
I spotted one issue on this branch. All fiat values except BTC stay as skeleton on dashboard, even though I can already see whole portfolio balance and fiat value in account detail.
image

matejkriz:
> I spotted one issue on this branch. All fiat values except BTC stay as skeleton on dashboard, even though I can already see whole portfolio balance and fiat value in account detail. 

I can't reproduce it anymore after wiping the data. I tried also switching between develop and this branch. Hopefully it was my local issue. Unblocking this PR if I'm the only one who seen that...

matejkriz (approved):

PR #14236, opened 2024-09-10
test(suite-desktop-core): check bridge version running in app

karliatto:
## Description

This changes check that the version of Bridge running when Trezor Suite app is running is the right one.

## Related Issue

Related to PR https://github.com/trezor/trezor-suite/pull/13984


mroz22 (approved):
looks like it works

PR #14241, opened 2024-09-10
test(suite-desktop-core): bridge runs when apps reloads

karliatto:
## Description

Add tests cases to test bridge runs when apps reloads.

I keep the todo `    // todo: add test case when INTERNAL bridge is killed and auto-restarted` as a follow-up.

## Related Issue

Resolve  https://github.com/trezor/trezor-suite/pull/13984/

### DIFF
--- a/packages/suite-desktop-core/e2e/tests/spawn-bridge.test.ts
+++ b/packages/suite-desktop-core/e2e/tests/spawn-bridge.test.ts
@@ -41,6 +41,13 @@ testPlaywright.describe.serial('Bridge', () => {
         const { version } = json;
         expectPlaywright(version).toEqual(expectedBridgeVersion);
 
+        // bridge is running after renderer window is refreshed
+        await suite.window.reload();
+        await suite.window.title();
+        // bridge is running
+        const bridgeRes2 = await request.get('http://127.0.0.1:21325/status/');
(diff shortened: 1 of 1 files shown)


packages/suite-desktop-core/e2e/tests/spawn-bridge.test.ts
```diff
@@ -51,4 +59,23 @@ testPlaywright.describe.serial('Bridge', () => {
             // ok
         }
     });
+
+    testPlaywright(
+        'App acquired device, EXTERNAL bridge is restarted, app reconnects',
+        async () => {
+            await TrezorUserEnvLink.startEmu({ wipe: true, version: '2-latest', model: 'T2T1' });
+            await TrezorUserEnvLink.setupEmu({});
```
mroz22, line 67: shoudln't it start bridge here? emu is probably not important

karliatto: Right, the test was passing since bridge was running from the previous test, so I changed stopBridge in beforeAll to beforeEach ot start each test the same way and then start it in this test.

The emu is required to "passThroughInitialRun" so we make sure device is connected before stopping bridge.

mroz22 (approved):
gracias 🙏

PR #14233, opened 2024-09-09
feat(message-system):  add native testing message for enabled vtc coin

vytick:
Agreed with @STew790 this adds message for mobil with enabled VTC to test #13470

⚠️ Should be released for `DEV` and removed after testing.

[In notion](https://www.notion.so/satoshilabs/Testing-native-message-for-enabled-coins-f660f921403e4d888c1d1b548e54448b)

PR #14242, opened 2024-09-10
chore(message-system): allow to release message system on DEVELOP environment from any branch

matejkriz:
## Description

- Allow to release message system on DEVELOP environment from any branch.
- Keep production release restricted to develop.

Just an idea for simplified testing, I see no point in restricting releases to develop environment from develop branch. If we can release it from any branch, it could prevent unintentional production release of testing message like https://github.com/trezor/trezor-suite/pull/14233

### DIFF
--- a/.github/workflows/release-suite-config.yml
+++ b/.github/workflows/release-suite-config.yml
@@ -36,13 +36,13 @@ jobs:
           role-to-assume: ${{ github.event.inputs.environment == 'develop-message' && 'arn:aws:iam::538326561891:role/gh_actions_suite_develop_message' || 'arn:aws:iam::538326561891:role/gh_actions_suite_production_message' }}
           aws-region: ${{ env.AWS_REGION }}
 
-      - name: Build and sign ${{ github.event.inputs.environment }} config
-        if: ${{ github.event.inputs.environment == 'develop-message' && github.ref == 'refs/heads/develop' }}
+      - name: Build and sign develop message-system config file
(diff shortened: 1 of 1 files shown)


MiroslavProchazka (approved):
Great, I was thinking that something like this would definitely make testing easier. It doesn’t make sense to merge new messages into the develop branch just to test them and then revert it. So thanks, good job!

vytick:
I would love to have something like this, butmaybe it would be better to have a special branch for those releases so we can easily track what is actually released in dev messages and how it changed over the time. wdyt?

vdovhanych:
We should not release test messages from any branch; we would not have a source of truth about what is actually deployed there. The messages are for the develop version of the app environment, and it makes perfect sense to release them from develop branch. I see why you would want to release it from any branch but i feel it would create some messy situations in the future. 

What I would do, though, is to have production messages be released from protected release branches, not from develop.

vdovhanych:
> I would love to have something like this, butmaybe it would be better to have a special branch for those releases so we can easily track what is actually released in dev messages and how it changed over the time. wdyt?

That is what develop should be. Why have some other branch for it?

vytick:
👍 Sure. I just believe, that having two separate sources for two separate environments is a good thing. Introducing release branch for prod messages would do the trick as well 👌.

matejkriz:
next try: https://github.com/trezor/trezor-suite/pull/14336

PR #14252, opened 2024-09-10
fix(transport): node bridge version parsing issue

martykan:
## Description

Fix version parsing for Node Bridge. 
Since #14153 the Node Bridge version is reported in format such as "3.0.0-bundled.24.10.0".

However in `BridgeTransport` there is some logic to check if the version is outdated and it fails to parse it.
So with this PR we throw away the part behind the dash.


marekrjpolak (approved):

karliatto (approved):
Thanks!

PR #14254, opened 2024-09-10
fix(suite-native): check btc firmware instead of btc only device for coin enabling

### LINKED ISSUE #14255: Coin Enabling works wrong for BTC only device
**Describe the bug**
For BTC only device without testnets allowed it shows initial coin enabling settings and in settings it shows networks to use with only BTC. 

**How to reproduce**
Have BTC only device, connect to not set app with fresh app with coin enabling on. It shows initial dialog and in settings there is BTC instead of BTC only UI.

**Expected behavior**
Dont show initial dialog for BTC only firmware.
In settings show BTC only graphic

**Screenshots**
This should not be seen

**Additional context**
This should be in settings instead

vytick:
Fix for correct behaviour of coin enabling when user has BTC only firmware.

- Using `hasBitcoinOnlyFirmware` that check firmware, instead of previos `selectIsBitcoinOnlyDevice` that checks device. 
- Checking symbols count instead of networks to show correct UI in settings

Resolve #14255

### DIFF
--- a/suite-common/wallet-core/src/device/deviceReducer.ts
+++ b/suite-common/wallet-core/src/device/deviceReducer.ts
@@ -7,6 +7,7 @@ import { Device, Features, UI } from '@trezor/connect';
 import {
     getFirmwareVersion,
     getFirmwareVersionArray,
+    hasBitcoinOnlyFirmware,
     isBitcoinOnlyDevice,
 } from '@trezor/device-utils';
 import { NetworkDeviceSupport, NetworkSymbol, networks } from '@suite-common/wallet-config';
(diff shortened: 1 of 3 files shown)


suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -959,3 +960,6 @@ export const selectInstacelessUnselectedDevices = memoize((state: DeviceRootStat
 
 export const selectIsBitcoinOnlyDevice = (state: DeviceRootState) =>
     isBitcoinOnlyDevice(selectDevice(state));
+
+export const selectHasBitcoinOnlyFirmware = (state: DeviceRootState) =>
```
Nodonisko, line 966: What's the difference between these two selectors?

vytick: The first one checks device, the second checks firmware

vytick: but yes, the name of the selectIsBitcoinOnlyDevice confused me previously as well

matejkriz: can be renamed to BitcoinOnlyTrezor... or comment added

vytick: ddaeae5da7

matejkriz: `// BTC only device is a device that has BTC only firmware`
That is not true or at least confusing. 

suggestions:
`// Bitcoin only Trezor with bitcoin-only firmware.`

Because `selectIsBitcoinOnlyDevice` returns `false` for regular Trezor with btc-only FW...

vytick: You are right, that is better, Device is confusing whether it is hw device or our suite device. Trezor is much better

fa0dcac981

matejkriz (approved):

PR #8700, opened 2023-06-14
Fix/memoization

matejkriz:
## Description

Removing lot of unnecessary memoization from selectors. 

- It makes no sense to memoize selectors returning primitive values and not doing any costly calculations.
- Even if selector returns object, it doesn't mean that memoization is necessary. Performance should be measured in that case to verify if unnecessary rerenders costs more than the memoization. Check https://github.com/dai-shi/proxy-memoize/issues/81 and https://github.com/dai-shi/proxy-memoize/issues/80.

 In suite-desktop/web we use shallow equality check to prevent unnecessary rerenders. In suite-native there is just strict equality check used in useSelector.

I also verified, that using memoized selector inside other memoized selector is fine (I suspected if changes in the affected properties of the nested selector correctly invalidate the cache and yes.)

<details>
  <summary>How did I debug it?</summary>
  
  
  `settingsReducer.ts`:
  ```js
let counterSomething = 1;
const selectSomething = memoizeWithArgs((state: AppState) => {
    console.log('selectSomethingSettings called', ++counterSomething, state.resize?.screenWidth);
    return (state.resize?.screenWidth || 0) > 1000;
});

let counter = 1;
export const selectEnabledNetworksMemoized = memoizeWithArgs((state: AppState) => {
    console.log('selectEnabledNetworks called', ++counter);
    const something = selectSomething(state);
    return {
        enabledNetworks: state.wallet.settings.enabledNetworks,
        something,
    };
});
  ```
  
  `SettingsGeneral.tsx`:
  ```js
  console.time('selectEnabledNetworks');
  const { enabledNetworks, something } = useSelector(selectEnabledNetworksMemoized);
  console.timeEnd('selectEnabledNetworks');
  console.log('SettingsGeneral', something);
  
  
  {something && <h1>something</h1>}
  ```
</details>

PR #8659, opened 2023-06-08
Downgrade `proxy-memoize` to 2.0.2

marekrjpolak:
## Description

Upgrade of `proxy-memoize` to 2.0.4 (in https://github.com/trezor/trezor-suite/commit/eb588a8c4d2237f67a3d0bc49232639e2139208c) caused some issues with selector memoization. Therefore it was downgraded back to 2.0.2 until we fix the memoized selectors themselves.

## Related Issue

Could resolve #8658, could resolve #8660 and other various issues.

PR #14264, opened 2024-09-10
feat(mobile): minor perf stuff

Nodonisko:
## Description

This improved performance during discovery  for about ~10 FPS on iPhone. Android not measured but could be even better.

Surprisingly main issues was that memoized selector (I start to really dislike proxy-memoize because of too much magic and unpredictable performance)

## Related Issue

Resolve

### DIFF
--- a/suite-common/wallet-core/src/accounts/accountsReducer.ts
+++ b/suite-common/wallet-core/src/accounts/accountsReducer.ts
@@ -144,15 +144,14 @@ export const prepareAccountsReducer = createReducerWithExtraDeps(
 
 export const selectAccounts = (state: AccountsRootState) => state.wallet.accounts;
 
-export const selectAccountsByDeviceState = memoizeWithArgs(
-    (state: AccountsRootState, deviceState: string): Account[] =>
-        pipe(
-            selectAccounts(state),
-            A.filter(account => account.deviceState === deviceState),
-        ) as Account[],
-    // cache up to 3 devices to make sure it works correctly
-    { size: 3 },
-);
+export const selectAccountsByDeviceState = (
+    state: AccountsRootState,
+    deviceState: string,
+): Account[] =>
+    pipe(
+        selectAccounts(state),
+        A.filter(account => account.deviceState === deviceState),
+    ) as Account[];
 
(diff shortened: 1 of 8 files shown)


suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -144,15 +144,14 @@ export const prepareAccountsReducer = createReducerWithExtraDeps(
...
-export const selectAccountsByDeviceState = memoizeWithArgs(
-    (state: AccountsRootState, deviceState: string): Account[] =>
-        pipe(
-            selectAccounts(state),
-            A.filter(account => account.deviceState === deviceState),
-        ) as Account[],
-    // cache up to 3 devices to make sure it works correctly
-    { size: 3 },
-);
+export const selectAccountsByDeviceState = (
```
vytick, line 147: Maybe it is missing memoization, but it feels that switching between home, assets and receive tab takes really long time. In my ios simulator it took around 5 seconds on fully discovered allseed.

matejkriz:
I really dislike proxy-memoize for a long time 🙉 It was causing some really unpredictable bugs on desktop. (See https://github.com/trezor/trezor-suite/pull/8659, https://github.com/trezor/trezor-suite/pull/8700). The idea sounds really cool and smart, but in reality it's too much black magic.

suite-common/wallet-core/src/accounts/accountsReducer.ts
vytick: Hmm. Now i tried the same on different branch and it also took a long time 🤔. So maybe not related to this change, but still feels very slow.

Nodonisko: Good to know I will check what's happening.

Nodonisko: I think it's because there is quite a lot of accounts on all seed and there is no FlashList for Accounts/Receive screen.

Nodonisko: not related to any of my changes

matejkriz (approved):

PR #14267, opened 2024-09-11
fix(suite-native): check discovery existence instead of whether it is active

vytick:
Replace `selectIsDeviceDiscoveryActive` with `selectHasDeviceDiscovery` to check whether there is a discovery, because if Cardano was enabled, the discovery could have been changed to `STOPPED` by cardano derivation thunk and this caused our UI not to behave correctly while waiting for discovery to finish.

On mobile, we dont rely on discovery being active, because we delete the object once it finishes so we dont need to check the state and existence check is sufficient.

### DIFF
--- a/suite-native/accounts/src/components/AddAccountsButton.tsx
+++ b/suite-native/accounts/src/components/AddAccountsButton.tsx
@@ -12,7 +12,7 @@ import {
 } from '@suite-native/navigation';
 import { IconButton } from '@suite-native/atoms';
 import {
-    selectDeviceDiscovery,
+    selectHasDeviceDiscovery,
     selectIsDeviceInViewOnlyMode,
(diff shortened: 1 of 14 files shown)


matejkriz (commented):
It doesn't look like proper fix to me. `selectDeviceDiscovery` returns the object of discovery, right? I think we need new selector for suite-native, something like `selectHasDeviceDiscovery`. Or find a way how not to stop discovery by Cardano on mobile...

vytick:
You are right, I will update this PR once #14273 is merged and use `selectHasDeviceDiscovery` on all relevant places

vytick:
Updated in 013ec1156e @matejkriz

matejkriz (approved):
It seems to be working fine 👍

PR #14269, opened 2024-09-11
fix(components): reintroduce className to Heading component

### LINKED ISSUE #14261: Components ignoring className
While it is discouraged to overload `@trezor/components` with styles, there is still some code that does this. However, `className` prop is ignored by some components (e.g. `H2`), which leads to broken layouts.

Add `className` prop to all components where it once was, unless you are sure no code relies on it.

For example, all this code is ignored:
Screenshot 2024-09-10 at 16 49 52

seibei-iguchi:


jvaclavik (approved):
sad but true :D

PR #14273, opened 2024-09-11
fix(suite-native): let passphrase flow finish sooner if some account is found

vytick:
Before, it waited for discovery to end. Now it takes you to dashboard as soon as first account is discovered of i no is discovered you wait for discovery to finish. 

https://github.com/user-attachments/assets/afdc5121-df01-415c-aebd-2ab7fd60d5e2

### DIFF
--- a/suite-common/wallet-core/src/accounts/accountsReducer.ts
+++ b/suite-common/wallet-core/src/accounts/accountsReducer.ts
@@ -16,7 +16,11 @@ import {
     selectHasOnlyPortfolioDevice,
 } from '../device/deviceReducer';
 import { deviceActions } from '../device/deviceActions';
-import { DiscoveryRootState, selectIsDeviceDiscoveryActive } from '../discovery/discoveryReducer';
+import {
+    DiscoveryRootState,
+    selectHasDeviceDiscovery,
+    selectIsDeviceDiscoveryActive,
+} from '../discovery/discoveryReducer';
 
 export type AccountsState = Account[];
 
@@ -424,11 +428,17 @@ export const selectIsDeviceNotEmpty = (
 ) => {
     const deviceState = selectDeviceState(state);
     const nonEmptyAccounts = selectNonEmptyDeviceAccounts(state);
-    const isDeviceDiscoveryActive = selectIsDeviceDiscoveryActive(state);
+    const hasDiscovery = selectHasDeviceDiscovery(state);
 
-    if ((isDeviceDiscoveryActive || !deviceState) && A.isEmpty(nonEmptyAccounts)) {
+    const isNotEmpty = A.isNotEmpty(nonEmptyAccounts);
+
+    if (isNotEmpty) {
+        return true;
+    }
+
+    if (hasDiscovery || !deviceState) {
         return null;
     }
 
-    return A.isNotEmpty(nonEmptyAccounts);
+    return isNotEmpty;
 };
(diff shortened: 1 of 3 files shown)


suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -425,9 +429,9 @@ export const selectIsDeviceNotEmpty = (
 ) => {
     const deviceState = selectDeviceState(state);
     const nonEmptyAccounts = selectNonEmptyDeviceAccounts(state);
-    const isDeviceDiscoveryActive = selectIsDeviceDiscoveryActive(state);
+    const hasDiscovery = selectDeviceDiscovery(state);
 
-    if ((isDeviceDiscoveryActive || !deviceState) && A.isEmpty(nonEmptyAccounts)) {
+    if ((hasDiscovery || !deviceState) && A.isEmpty(nonEmptyAccounts)) {
```
matejkriz: You can have discovery but already know that the device is not empty. The method comment also claimed that. 

It would makes sense to add some short circuit evaluation `if(isNotEmpty) { return true; }` before the hasDiscovery condition.

suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -425,9 +429,9 @@ export const selectIsDeviceNotEmpty = (
 ) => {
     const deviceState = selectDeviceState(state);
     const nonEmptyAccounts = selectNonEmptyDeviceAccounts(state);
-    const isDeviceDiscoveryActive = selectIsDeviceDiscoveryActive(state);
+    const hasDiscovery = selectDeviceDiscovery(state);
```
PeKne: the `hasDiscovery` variable name suggests that the value is boolean but it is actually a discovery object. Fix that please. Either by renaming the variable or creating new selector that returns boolean.

suite-common/wallet-core/src/accounts/accountsReducer.ts
vytick: cc10595a14

suite-common/wallet-core/src/accounts/accountsReducer.ts
vytick: cc10595a14

vytick: Good idea, I will change it here and once merged expand this to all relevant places in #14267 👍

suite-common/wallet-core/src/accounts/accountsReducer.ts
matejkriz: I was wrong, now I see that it was already working the same, just the condition wasn't so clear. Now it's easier to read but works the same, right? 👍

suite-common/wallet-core/src/accounts/accountsReducer.ts
matejkriz: The new `selectHasDeviceDiscovery` returns false or discovery object? I would expect boolean.

vytick: My bad:
55a8159217

suite-common/wallet-core/src/accounts/accountsReducer.ts
vytick: It should, but you are right, this is more readable

matejkriz (approved):
Works nice, I've tested both empty and used passphrases.

STew790:
**QA OK**
I am taken to discovery after first acc is found.

**Info**
24.9.1 5524f8d

PR #14281, opened 2024-09-11
chore(suite-native): use new Networks structure

Lemonexe:
Refactoring deprecated `networksCompatibility` to `networks` + related utils in Suite Native

## Description

- 8f0270267d714f747542facdee446861ef20ed69 add a new helper to make working with account types easier & less confusing
- 35616d6c0270ad8ad24976d0f204aa801e49f56e
  - refactor account discovery
  - update adding coinjoin account
  - keep backwards compatibility for suite, until it is refactored there too 

## Related Issue

Part of #13839

## Screenshots

Discovered accounts at Android - the same screenshot on develop & this branch: :heavy_check_mark: 
Screenshot from 2024-09-19 09-39-05

To be really sure discovery works the same under the hood, I console logged [discoveryNetworksTotalCount (TOTAL)](https://github.com/trezor/trezor-suite/blob/b92c24e0f3a100b1c17fa3ff3856b5b2c9240860/suite-native/discovery/src/discoveryThunks.ts#L569), [finishedNetworksCount (LOADED)](https://github.com/trezor/trezor-suite/blob/b92c24e0f3a100b1c17fa3ff3856b5b2c9240860/suite-native/discovery/src/discoveryThunks.ts#L125) and [network+accountType? when dispatching new discovery](https://github.com/trezor/trezor-suite/blob/b92c24e0f3a100b1c17fa3ff3856b5b2c9240860/suite-native/discovery/src/discoveryThunks.ts#L698).
These are the results - I verifed they are the same at develop & this branch: :heavy_check_mark: 
[accountDiscovery.txt](https://github.com/user-attachments/files/17055704/accountDiscovery.txt)

### DIFF
--- a/suite-common/wallet-config/src/utils.ts
+++ b/suite-common/wallet-config/src/utils.ts
@@ -1,5 +1,14 @@
 import { networks, networksCompatibility } from './networksConfig';
-import { AccountType, Network, NetworkFeature, Networks, NetworkSymbol } from './types';
+import {
+    AccountType,
+    Network,
+    NetworkFeature,
+    Networks,
+    NetworkSymbol,
+    NormalizedNetworkAccount,
+} from './types';
+
+export const NORMAL_ACCOUNT_TYPE = 'normal' satisfies AccountType;
 
 /**
  * array from `networks` as a `Network[]` type instead of inferred type
@@ -35,6 +44,23 @@ export const getTestnets = (debug = false) =>
     networksCollection.filter(n => n.testnet === true && (!n.isDebugOnlyNetwork || debug));
 
 export const getTestnetSymbols = () => getTestnets().map(n => n.symbol);
+
+/**
+ * For a given network, return a collection of accounts incl. 'normal', and with missing properties backfilled from 'normal'
+ */
+export const normalizeNetworkAccounts = (network: Network): NormalizedNetworkAccount[] => {
+    const normalAccount: NormalizedNetworkAccount = {
+        accountType: NORMAL_ACCOUNT_TYPE,
+        bip43Path: network.bip43Path,
+        features: network.features,
+    };
+    const alternativeAccounts = Object.values(network.accountTypes).map(account => ({
+        ...normalAccount,
+        ...account,
+    }));
+
+    return [normalAccount, ...alternativeAccounts];
+};
 
 export const isBlockbookBasedNetwork = (symbol: NetworkSymbol) =>
     networks[symbol]?.customBackends.some(backend => backend === 'blockbook');
--- a/suite-common/wallet-config/src/utils.ts
+++ b/suite-common/wallet-config/src/utils.ts
@@ -1,4 +1,4 @@
-import { networks, networksCompatibility } from './networksConfig';
+import { networks } from './networksConfig';
 import {
     AccountType,
     Network,
@@ -14,26 +14,6 @@ export const NORMAL_ACCOUNT_TYPE = 'normal' satisfies AccountType;
  * array from `networks` as a `Network[]` type instead of inferred type
  */
 export const networksCollection: Network[] = Object.values(networks);
-
-/**
- * @deprecated See `networksCompatibility`
- */
-export const getMainnetsCompatible = (debug = false, bnb = false) =>
-    networksCompatibility.filter(
-        n =>
-            !n.accountType &&
-            !n.testnet &&
-            (!n.isDebugOnlyNetwork || debug) &&
-            (bnb || n.symbol !== 'bnb'),
-    );
-
-/**
- * @deprecated See `networksCompatibility`
- */
-export const getTestnetsCompatible = (debug = false) =>
-    networksCompatibility.filter(
-        n => !n.accountType && n.testnet === true && (!n.isDebugOnlyNetwork || debug),
-    );
 
 export const getMainnets = (debug = false, bnb = false) =>
     networksCollection.filter(
--- a/suite-common/wallet-core/src/discovery/discoveryThunks.ts
+++ b/suite-common/wallet-core/src/discovery/discoveryThunks.ts
@@ -10,7 +10,14 @@ import {
 } from '@suite-common/wallet-utils';
 import { Discovery, DiscoveryItem, PartialDiscovery } from '@suite-common/wallet-types';
 import { getTxsPerPage } from '@suite-common/suite-utils';
-import { networksCompatibility, NetworkSymbol } from '@suite-common/wallet-config';
+import {
+    networksCompatibility,
+    NetworkSymbol,
+    Network,
+    networksCollection,
+    normalizeNetworkAccounts,
+    NormalizedNetworkAccount,
+} from '@suite-common/wallet-config';
 import { getFirmwareVersion } from '@trezor/device-utils';
 import { versionUtils } from '@trezor/utils';
 
@@ -36,7 +43,49 @@ type ProgressEvent = BundleProgress<AccountInfo | null>['payload'];
 
 export const LIMIT = 10;
 
+/**
+ * Filter collection of activated networks to only include those supported by device & suite
+ */
 export const filterUnavailableNetworks = (
+    enabledNetworks: NetworkSymbol[],
+    device?: TrezorDevice,
+): Network[] =>
+    networksCollection.filter(n => {
+        const firmwareVersion = getFirmwareVersion(device);
+        const internalModel = device?.features?.internal_model;
+
+        const isSupportedInSuite =
+            !n.support || // support is not defined => is supported
+            !internalModel || // typescript. device undefined. => supported
+            (n.support[internalModel] && // support is defined for current device
+                versionUtils.isNewerOrEqual(firmwareVersion, n.support[internalModel] as string)); // device version is newer or equal to support field in networks => supported
+
+        return (
+            enabledNetworks.includes(n.symbol) &&
+            !n.isHidden &&
+            !device?.unavailableCapabilities?.[n.symbol] && // exclude by network symbol (ex: xrp on T1B1)
+            isSupportedInSuite
+        );
+    });
+
+/**
+ * For a given network, return a collection of normalized accounts (incl. 'normal'), excluding types unsupported by device or suite
+ */
+export const filterUnavailableAccountTypes = (
+    network: Network,
+    device?: TrezorDevice,
+): NormalizedNetworkAccount[] =>
+    normalizeNetworkAccounts(network).filter(
+        networkAccount =>
+            isTrezorConnectBackendType(networkAccount.backendType) && // exclude accounts with unsupported backend type, such as coinjoin
+            !device?.unavailableCapabilities?.[networkAccount.accountType!], // exclude by account types (ex: taproot)
+    );
+
+/**
+ * @deprecated Old implementation, temporarily duplicated here for suite, while the new one is used in suite-native
+ * Fully replaced by combined use of filterUnavailableNetworks and filterUnavailableAccountTypes
+ */
+export const filterUnavailableNetworksCompatible = (
     enabledNetworks: NetworkSymbol[],
     device?: TrezorDevice,
 ) =>
@@ -230,8 +279,8 @@ export const getBundleThunk = createThunk(
             return bundle;
         }
 
-        const networks = filterUnavailableNetworks(discovery.networks, device);
-        networks.forEach(configNetwork => {
+        const filteredNetworks = filterUnavailableNetworksCompatible(discovery.networks, device);
+        filteredNetworks.forEach(configNetwork => {
             // find all existed accounts
             const accountType = configNetwork.accountType || 'normal';
             const prevAccounts = accountsByDeviceState
@@ -660,7 +709,9 @@ export const createDiscoveryThunk = createThunk(
             selectors: { selectEnabledNetworks },
         } = extra;
         const enabledNetworks = selectEnabledNetworks(getState());
-        const networks = filterUnavailableNetworks(enabledNetworks, device).map(n => n.symbol);
+        const networksSymbols = filterUnavailableNetworksCompatible(enabledNetworks, device).map(
+            n => n.symbol,
+        );
 
         dispatch(
             createDiscovery({
@@ -668,11 +719,11 @@ export const createDiscoveryThunk = createThunk(
                 authConfirm: !device.useEmptyPassphrase,
                 index: 0,
                 status: DiscoveryStatus.IDLE,
-                total: LIMIT * networks.length,
+                total: LIMIT * networksSymbols.length,
                 bundleSize: 0,
                 loaded: 0,
                 failed: [],
-                networks,
+                networks: networksSymbols,
             }),
         );
     },
@@ -690,19 +741,22 @@ export const updateNetworkSettingsThunk = createThunk(
         discovery.forEach(d => {
             const devices = selectDevices(getState());
             const device = devices.find(dev => dev.state === d.deviceState);
-            const networks = filterUnavailableNetworks(enabledNetworks, device).map(n => n.symbol);
+            const networksSymbols = filterUnavailableNetworksCompatible(
+                enabledNetworks,
+                device,
+            ).map(n => n.symbol);
 
             const progress = dispatch(
                 calculateProgress({
                     ...d,
-                    networks,
+                    networks: networksSymbols,
                     failed: [],
                 }),
             );
             dispatch(
                 updateDiscovery({
                     ...progress,
-                    networks,
+                    networks: networksSymbols,
                     failed: [],
                 }),
             );
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -19,6 +19,7 @@ import {
     disableAccountsThunk,
     selectFirstNormalAccountForNetworkSymbol,
     selectHasDeviceDiscovery,
+    filterUnavailableAccountTypes,
 } from '@suite-common/wallet-core';
 import { selectIsAccountAlreadyDiscovered } from '@suite-native/accounts';
 import TrezorConnect, { StaticSessionId } from '@trezor/connect';
@@ -26,9 +27,12 @@ import { Account, DiscoveryItem } from '@suite-common/wallet-types';
 import { getDerivationType, tryGetAccountIdentity } from '@suite-common/wallet-utils';
 import {
     AccountType,
-    NetworkCompatible,
+    Network,
     NetworkSymbol,
+    getNetwork,
     getNetworkType,
+    NetworkAccount,
+    normalizeNetworkAccounts,
 } from '@suite-common/wallet-config';
 import { DiscoveryStatus } from '@suite-common/wallet-constants';
 import { requestDeviceAccess } from '@suite-native/device-mutex';
@@ -45,11 +49,8 @@ import {
     selectNetworksWithUnfinishedDiscovery,
     selectShouldRunDiscoveryForDevice,
 } from './discoverySelectors';
-import { getNetworkSymbols } from './utils';
 
 const DISCOVERY_DEFAULT_BATCH_SIZE = 2;
-
-export const NORMAL_ACCOUNT_TYPE: AccountType = 'normal';
 
 const DISCOVERY_BATCH_SIZE_PER_COIN: Partial<Record<NetworkSymbol, number>> = {
     bch: 1,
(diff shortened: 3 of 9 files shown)


suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -668,8 +666,18 @@ export const startDescriptorPreloadedDiscoveryThunk = createThunk(
...
+                .filter(
+                    ({ accountType }: NetworkAccount) =>
+                        network.networkType !== 'cardano' ||
+                        (availableCardanoDerivations as string[]).includes(
+                            accountType ?? NORMAL_ACCOUNT_TYPE,
+                        ),
+                )
+                .forEach(({ accountType }: NetworkAccount) => {
+                    dispatch(discoverNetworkBatchThunk({ deviceState, network, accountType }));
+                });
```
Lemonexe, line 696: Change of logic here: previously `network` was duplicated for each `accountType`, so `dispatch` was sent simply for each network, now it's nested.

suite-native/discovery/src/utils.ts
```diff
@@ -29,10 +20,5 @@ export const getNetworksWithUnfinishedDiscovery = (
...
 
-export const getNetworkSymbols = (networks: readonly NetworkCompatible[]): NetworkSymbol[] =>
-    F.toMutable(
-        A.uniqBy(
-            networks.map(n => n.symbol),
-            F.identity,
-        ),
-    );
+export const getNetworkSymbols = (networks: readonly Network[]): NetworkSymbol[] =>
+    F.toMutable(networks.map(n => n.symbol));
```
Lemonexe, line 24: the new `networks` are guaranteed to be unique, so no need to make them unique

suite-native/discovery/src/utils.ts
```diff
@@ -1,24 +1,15 @@
...
+    enabledNetworks: readonly Network[],
     accounts: Account[],
     accountsLimit: number,
 ) =>
     enabledNetworks.filter(network => {
-        const networkAccountsOfType = accounts.filter(
-            account =>
-                account.symbol === network.symbol &&
-                account.accountType === normalizedAccountType(network.accountType),
-        );
```
Lemonexe, line 21: I am not sure if I'm right about that, but it seems `enabledNetworks` are only networks with normal accounts. So no need to filter by `accountType` too.
`enabledNetworks` originally come from [selectDeviceSupportedNetworks](https://github.com/trezor/trezor-suite/blob/a1ba1fd58d9e17059fe0e5af8ea7f5534c50e523/suite-common/wallet-core/src/device/deviceReducer.ts#L734-L769) where networks are reduced to symbols, so without additional information on account types, it should be just normal accounts.

EDIT: OUTDATED

suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
```diff
@@ -122,23 +122,8 @@ export const useAddCoinAccount = () => {
...
-        return supportedNetworks.filter(
-            network =>
-                network.symbol === networkSymbol &&
-                (type === NORMAL_ACCOUNT_TYPE
-                    ? network.accountType === undefined
-                    : network.accountType === type),
-        )[0];
-    };
+    const getNetworkToAdd = ({ networkSymbol }: { networkSymbol: NetworkSymbol }) =>
+        supportedNetworks.filter(network => network.symbol === networkSymbol)[0];
```
Lemonexe: This gets simplified from "get network specific for the accountType"
to just getting the main network (for normal accountType),
and `accountType` is now passed alongside.

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -645,18 +650,11 @@ export const startDescriptorPreloadedDiscoveryThunk = createThunk(
...
-                (availableCardanoDerivations as string[]).includes(
-                    (network.accountType ?? NORMAL_ACCOUNT_TYPE) as string,
-                ),
-        );
 
         await dispatch(
             createDescriptorPreloadedDiscoveryThunk({
                 deviceState,
-                networks: networksFiltered,
+                networks: networksWithUnfinishedDiscovery,
```
Lemonexe: Doesn't matter what accounts there are, because in `createDescriptorPreloadedDiscoveryThunk`, the `networks` are mapped to symbols, so that information is dropped.

So I believe that even in previous implementation, we could have passed `networksWithUnfinishedDiscovery`

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -424,10 +427,12 @@ const discoverNetworkBatchThunk = createThunk(
             deviceState,
             round = 1,
             network,
+            accountType,
         }: {
             deviceState: string;
             round?: number;
-            network: NetworkCompatible;
+            network: Network;
+            accountType: AccountType;
```
Lemonexe, line 434: this function relied on getting a specific `network` for given `accountType`, now it's normal `network` and `accountType` passed alongside.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -39,6 +44,31 @@ export const LIMIT = 10;
...
+                versionUtils.isNewerOrEqual(firmwareVersion, n.support[internalModel] as string)); // device version is newer or equal to support field in networks => supported
+
+        return (
+            enabledNetworks.includes(n.symbol) &&
+            !n.isHidden &&
+            !device?.unavailableCapabilities?.[n.symbol] && // exclude by network symbol (ex: xrp on T1B1)
+            isSupportedInSuite
+        );
+    });
+
```
Lemonexe, line 70: Because this fn is common for discovery both in `suite` and `suite-native`, and in this PR I want to refactor only native, I copied the old fn for `suite` as `filterUnavailableNetworksCompatible`, while the two new fns are used in native

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -438,27 +445,13 @@ const discoverNetworkBatchThunk = createThunk(
...
-            network.networkType === 'cardano' &&
-            (network.accountType === 'ledger' || network.accountType === 'legacy') &&
-            !discovery.availableCardanoDerivations?.includes(network.accountType);
-
-        if (isIncompatibleCardanoType) {
-            dispatch(finishNetworkTypeDiscoveryThunk());
-
-            return;
-        }
-
```
Lemonexe, line 455: Already handled in `startDescriptorPreloadedDiscoveryThunk`, which is the only place which calls `discoverNetworkBatchThunk` (this function)

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -560,14 +555,27 @@ export const createDescriptorPreloadedDiscoveryThunk = createThunk(
...
 
-        const networksSymbols = networks.map(network => network.symbol);
+        // Some cardano derivation may not be supported by the device. We need to exclude them from total count.
+        const isAvailableAccount = (networkType: NetworkType, accountType: AccountType) =>
+            networkType !== 'cardano' ||
+            (availableCardanoDerivations as string[]).includes(accountType);
 
-        const discoveryNetworksTotalCount = networksSymbols.length;
+        const discoveryNetworksTotalCount = networks.reduce((count, network) => {
+            count += isAvailableAccount(network.networkType, NORMAL_ACCOUNT_TYPE) ? 1 : 0;
```
Lemonexe: I could have written `count += Number(isAvailableAccount..);` instead of ternary, but this is clearer :smile:

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -645,11 +653,15 @@ export const startDescriptorPreloadedDiscoveryThunk = createThunk(
...
+        // filter out cardano-type networks, if none of their derivations are available
         const networksFiltered = networksWithUnfinishedDiscovery.filter(
             network =>
                 network.networkType !== 'cardano' ||
-                (availableCardanoDerivations as string[]).includes(
-                    (network.accountType ?? NORMAL_ACCOUNT_TYPE) as string,
+                // first try normal derivation, if it's not available, try other derivations
+                availableCardanoDerivations.includes(NORMAL_ACCOUNT_TYPE) ||
+                filterUnavailableAccountTypes(network).some(({ accountType }: NetworkAccount) =>
+                    (availableCardanoDerivations as string[]).includes(accountType),
```
Lemonexe, line 667: If `filterUnavailableAccountTypes` leaves empty array, `.some()` will be false, which is expected - network will be filtered out :heavy_check_mark: 
But that probably won't happen, cause `filterUnavailableAccountTypes` filters out non-standard accounts...

suite-native/discovery/src/utils.ts
```diff
@@ -1,38 +1,37 @@
...
 import { Account } from '@suite-common/wallet-types';
 
-const NORMAL_ACCOUNT_TYPE = 'normal';
-
-// network uses undefined for normal type, but account uses 'normal'
-const normalizedAccountType = (accountType?: AccountType) => accountType ?? NORMAL_ACCOUNT_TYPE;
-
 export const getNetworksWithUnfinishedDiscovery = (
-    enabledNetworks: readonly NetworkCompatible[],
+    enabledNetworks: readonly Network[],
```
Lemonexe: This whole function used to filter out `NetworkCompatible`, now it filters `Network` with regard to its accounts (if there are some). So the collection length is different (smaller) - will it break something?
I think no, this is where it is used:
- `selectNetworksWithUnfinishedDiscovery`, and that is used:
  - `selectShouldRunDiscoveryForDevice` which [only looks if length > 0](https://github.com/trezor/trezor-suite/blob/5b29207bb5024babd48fd5ed504c3ca3b6ee51d7/suite-native/discovery/src/discoverySelectors.ts#L136-L138) :heavy_check_mark: 
  - `startDescriptorPreloadedDiscoveryThunk` which handles the foreach account :heavy_check_mark:

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -659,19 +701,21 @@ export const createDiscoveryThunk = createThunk(
...
+        );
 
         dispatch(
             createDiscovery({
                 deviceState,
                 authConfirm: !device.useEmptyPassphrase,
                 index: 0,
                 status: DiscoveryStatus.IDLE,
-                total: LIMIT * networks.length,
+                total: LIMIT * networksSymbols.length,
```
vytick, line 722: originally the number was  LIMIT * (networkSymbols.lenght * accountType.lenght), will it work correctly now?

Because previously networks were combination of networkSymbols and supported accountTypes

suite-native/discovery/src/utils.ts
```diff
@@ -1,38 +1,37 @@
...
 
-export const getNetworkSymbols = (networks: readonly NetworkCompatible[]): NetworkSymbol[] =>
-    F.toMutable(
-        A.uniqBy(
-            networks.map(n => n.symbol),
-            F.identity,
-        ),
-    );
+export const getNetworkSymbols = (networks: readonly Network[]): NetworkSymbol[] =>
+    F.toMutable(networks.map(n => n.symbol));
```
vytick:
```suggestion
networks.map(n => n.symbol);
```

suite-native/discovery/src/utils.ts
```diff
@@ -1,38 +1,37 @@
...
+            return (
+                A.isEmpty(networkAccountsOfType) ||
+                (networkAccountsOfType.length < accountsLimit &&
+                    networkAccountsOfType.every(account => account.visible))
+            );
+        };
+
+        if (isUndiscoveredAccountType('normal')) return true;
+
+        Object.values(network.accountTypes).forEach(({ accountType }) => {
```
vytick: we dont want to discover or check 'coinjoin' accounts

suite-native/discovery/src/utils.ts
```diff
@@ -1,38 +1,37 @@
...
-
         // if there is no account for this network -> we should have at least one account even if added via Add Coin
         // or if there is at least one visible account of this type, we should have at least one hidden account so adding funds outside of this app is detected and account shown
-        return (
-            A.isEmpty(networkAccountsOfType) ||
-            (networkAccountsOfType.length < accountsLimit &&
-                networkAccountsOfType.every(account => account.visible))
-        );
+        // the whole network is considered undiscovered if this is found for either 'normal' account, or any of the other accountTypes
+        const isUndiscoveredAccountType = (accountType: AccountType) => {
```
vytick: isUndiscoveredAccountType is created for each network over and over again, I would put it outside of filter

suite-native/discovery/src/utils.ts
```diff
@@ -1,38 +1,37 @@
...
+            );
+
+            return (
+                A.isEmpty(networkAccountsOfType) ||
+                (networkAccountsOfType.length < accountsLimit &&
+                    networkAccountsOfType.every(account => account.visible))
+            );
+        };
+
+        if (isUndiscoveredAccountType('normal')) return true;
```
vytick: overall can we simplify the code so all account types including 'normal' are checked in one go? This two step feels harder to read. 

ot: Is there a strong reason why normal type is part of the network and other account types/derivations are not? If all the account types were treated the same this wouldnt have to be there

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -439,27 +446,13 @@ const discoverNetworkBatchThunk = createThunk(
...
-
         const chunkBundle: Array<DiscoveryItem> = [];
 
         A.makeWithIndex(batchSize, batchIndex => {
             const isEvmLedgerDerivationPath =
-                network.networkType === 'ethereum' && network.accountType === 'ledger';
+                network.networkType === 'ethereum' && accountType === 'ledger';
             // first Ledger derivation path for EVM networks is the same as trezor, so we need to skip it (consistent with desktop)
             const index =
                 lastDiscoveredAccountIndex + batchIndex + (isEvmLedgerDerivationPath ? 1 : 0);
```
vytick, line 458: from now, using  `network.bip43Path` will always use normal account's derivation path. It should use derivation for each accounttype

suite-common/wallet-core/src/discovery/discoveryThunks.ts
Lemonexe: ~I agree, this is a mistake, thanks for catching it!~

EDIT: it works because `filterUnavailableNetworksCompatible`

suite-native/discovery/src/utils.ts
Lemonexe: Yeah, I placed it there so it has only one argument, I assumed creating a function won't create a performance problem, but it is messy. I can put it even to top scope, why not :slightly_smiling_face:

suite-native/discovery/src/utils.ts
Lemonexe: I can create a helper for it - a function `Network => NetworkAccount[]`, but with 1) `'normal'` account constructed from Network properties 2) missing properties for alternative accounts will be backfilled from `Network`.

So I will never again have to do stuff like `account.features ?? network.features` etc.

It will be helpful also for `suite` :slightly_smiling_face:

vytick: pretty, please :) 🙏

suite-native/discovery/src/utils.ts
Lemonexe: Right, I didn't realize.. then let's get rid of it, it's not very helpful now.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
Lemonexe: Oh this is actually only called in `suite` desktop. To make sure, I console logged the function call, it's not called at all in `suite-native`

suite-native/discovery/src/discoveryThunks.ts
Lemonexe: That's what L662 was for. But I used `normalizeNetworkAccounts` anyway in https://github.com/trezor/trezor-suite/pull/14281/commits/b92c24e0f3a100b1c17fa3ff3856b5b2c9240860, and it's much cleaner now I think :slightly_smiling_face:

suite-native/discovery/src/utils.ts
Lemonexe: Deleted in fixup https://github.com/trezor/trezor-suite/pull/14281/commits/8213b3178a8abe0f1e5c72685fbbfc303f2eefe0

suite-native/discovery/src/utils.ts
Lemonexe: Done in https://github.com/trezor/trezor-suite/pull/14281/commits/b92c24e0f3a100b1c17fa3ff3856b5b2c9240860, now it's using `normalizeNetworkAccounts`, which does this under-the-hood - it uses `filterUnavailableAccountTypes`, and it filters `isTrezorConnectBackendType('coinjoin')=false` :heavy_check_mark:

suite-native/discovery/src/utils.ts
Lemonexe: no longer necessary. Now using `normalizeNetworkAccounts` in https://github.com/trezor/trezor-suite/commit/b92c24e0f3a100b1c17fa3ff3856b5b2c9240860, this function was no longer called from two places, but only from one, so I inlined it again, as it was before this PR.

suite-native/discovery/src/utils.ts
Lemonexe: Created in https://github.com/trezor/trezor-suite/pull/14281/commits/8f0270267d714f747542facdee446861ef20ed69, used in https://github.com/trezor/trezor-suite/commit/b92c24e0f3a100b1c17fa3ff3856b5b2c9240860

suite-common/wallet-core/src/discovery/discoveryThunks.ts
vytick: I know. If it does not break desktop I am okay with that, but I have not tested that one.

suite-native/discovery/src/discoveryThunks.ts
Lemonexe: Fixed in https://github.com/trezor/trezor-suite/pull/14281/commits/b92c24e0f3a100b1c17fa3ff3856b5b2c9240860

suite-common/wallet-config/src/utils.ts
```diff
@@ -1,31 +1,20 @@
...
+import {
+    AccountType,
+    Network,
+    NetworkFeature,
+    Networks,
+    NetworkSymbol,
+    NormalizedNetworkAccount,
+} from './types';
+
+export const NORMAL_ACCOUNT_TYPE = 'normal' satisfies AccountType;
```
Lemonexe, line 11: `satisfies` is imho better than `NORMAL_ACCOUNT_TYPE: AccountType = 'normal'`, because then TS would think that it _could be_ `segwit` or `ledger` or whatever. This way it's only `'normal'`

suite-common/wallet-core/src/discovery/discoveryThunks.ts
Lemonexe: It does not, because I'm using `filterUnavailableNetworksCompatible`, which returns duplicated symbols. This is refactored in subsequent PR #14398 .

I realized that when writing the code originally, then forgot about it when I answered your comment :man_facepalming: Yes, this was working from the start, sry about confusion :heavy_check_mark:

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -675,9 +682,18 @@ export const startDescriptorPreloadedDiscoveryThunk = createThunk(
             return;
         }
 
-        // Start discovery for every network account type.
+        // Start discovery for every network - normal accounts as well as alternative accounts
         networksFiltered.forEach(network => {
```
vytick, line 686: `createDescriptorPreloadedDiscoveryThunk` and `discoverNetworkBatchThunk` are still dispatched for all account types for example in case BTC or ADA is selected, because there might be unsupported account types: coinjoin (btc) or legacy (ada).

This is because now we consider all network undiscovered if those two are not. This makes discovery to be started even when not needed. This causes some unwanted consequences (possible passphrase requests, discovery and fiat rates loading, etc.)

Lets discuss this later on call or in DMs.

suite-native/discovery/src/utils.ts
```diff
@@ -1,38 +1,26 @@
...
         // if there is no account for this network -> we should have at least one account even if added via Add Coin
         // or if there is at least one visible account of this type, we should have at least one hidden account so adding funds outside of this app is detected and account shown
-        return (
-            A.isEmpty(networkAccountsOfType) ||
-            (networkAccountsOfType.length < accountsLimit &&
-                networkAccountsOfType.every(account => account.visible))
-        );
-    });
+        // the whole network is considered undiscovered if this is found for either 'normal' account, or any of the other accountTypes
+        normalizeNetworkAccounts(network).some(({ accountType }) => {
```
vytick, line 21: normalizeNetworkAccounts makes it much more readable 👌 <3

vytick (approved):
Great job 🚀 

Thanks for all the heavy lifting 🏋️‍♂️ and welcome aboard to mobile development 😄 !

vytick:
@STew790  Please, test, that mobile discovery works the same as before, after this PR is merged 🙏. I tested this, but it might have broken something in coin enabling, adding coin, etc.. So more tests the merrier.

Thanks 🙇‍♂️

Lemonexe:
All suite-web E2E tests recurrently degraded. In this PR, it was last working at [this commit](https://github.com/trezor/trezor-suite/commit/d5c12b9a9babc78bf477dd48dd80ead0bfea44d4) last week, and since then, no changes have been done to suite desktop or web, only native..

STew790:
**QA OK**
I already mentioned this on the slack but the discovery together with the coin enabling now feels really smooth and fast. Nonetheless, I will definitely monitor this further, just to be super safe.

**Info**
24.9.2 https://github.com/trezor/trezor-suite/commit/b718af4a2b5d9fde28e1832276405f36a1c2399b

PR #14283, opened 2024-09-11
Support semver labels in version utils

marekrjpolak:
## Description

Even with #14252, Suite is sometimes still failing to properly compare new `3.0.0-bundled.24.10` Bridge version format. Therefore I extended `versionUtils` to be able to parse (and ignore) these semver labels, and, while I was in it, refactored the utils a bit.

The last commit removes the changes from #14252, which is optional. It's perfectly ok to leave it there.


marekrjpolak:
Ok, so I broke it. Previously it was working because `[]` translates to `undefined.undefined.undefined` which was handled as a valid semver version :angry:

Fix is on the way, but we'll have to be more careful about what we're using as a version.

mroz22 (approved):

PR #14286, opened 2024-09-11
fix(suite): Fix tertiary button with rgba

jvaclavik:
## Related Issue

Resolve


jvaclavik:
Fixed by elevation, with alpha colors, we are waiting for new palette

PR #14292, opened 2024-09-11
feat(suite-native): allow sending bitcoin transactions denominated in sats

### LINKED ISSUE #13823: Send coins - satoshi unit for BTC
## Acceptance criteria
- [ ] If user selected satoshi unit in settings, he also sees amount in satoshi in send form

yanascz:
Switching bitcoin units while a form draft is stored is not yet implemented. Will be handled separately.

## Related Issue

Resolve #13823

## Screenshots:
Simulator Screenshot - iPhone 15 Pro Max - 2024-09-11 at 16 38 06
Simulator Screenshot - iPhone 15 Pro Max - 2024-09-11 at 16 38 14
Simulator Screenshot - iPhone 15 Pro Max - 2024-09-11 at 16 38 49


suite-native/settings/src/settingsSlice.ts
```diff
@@ -53,11 +54,20 @@ export const appSettingsSlice = createSlice({
...
     state.appSettings.fiatCurrencyCode;
 export const selectBitcoinUnits = (state: SettingsSliceRootState) => state.appSettings.bitcoinUnits;
+export const selectAreSatsAmountUnit = (state: SettingsSliceRootState) =>
+    selectBitcoinUnits(state) === PROTO.AmountUnit.SATOSHI;
 export const selectIsOnboardingFinished = (state: SettingsSliceRootState) =>
     state.appSettings.isOnboardingFinished;
 export const selectViewOnlyCancelationTimestamp = (state: SettingsSliceRootState) =>
     state.appSettings.viewOnlyCancelationTimestamp;
 
+export const selectIsValueInSats = (
```
yanascz: Any idea where to place this one? Or do you think we need `useBitcoinAmountUnit` (partially) migrated to `suite-common` and used instead of this one?

suite-native/settings/src/settingsSlice.ts
```diff
@@ -53,11 +54,20 @@ export const appSettingsSlice = createSlice({
...
+    state: SettingsSliceRootState,
+    networkSymbol: NetworkSymbol | null | undefined,
```
vytick, line 66: Why to allow null and undefined for networkSymbol?

yanascz: Just to make using the selector easier. Most places where it's used have potentially `null` or `undefined` symbol. 😕 How would you suggest to change this?

suite-native/settings/src/settingsSlice.ts
vytick: To me, this sounds to be a good place

suite-native/settings/src/settingsSlice.ts
```diff
@@ -53,11 +54,20 @@ export const appSettingsSlice = createSlice({
...
+): boolean => {
+    return networkSymbol === 'btc' && selectAreSatsAmountUnit(state);
```
vytick: nit: extra `{ }` and return
```suggestion
    => networkSymbol === 'btc' && selectAreSatsAmountUnit(state);
```

suite-native/settings/src/settingsSlice.ts
vytick: I see. I have not checked all usages before. In this case it makes sense 👍 , sorry for the comment! :)

suite-native/settings/src/settingsSlice.ts
yanascz: Thanks, I like!

suite-native/settings/src/settingsSlice.ts
yanascz: It turns out the app doesn't support devices that do not allow sending in sats. Thus no need to use/migrate the `useBitcoinAmountUnit` hook.

suite-native/module-send/src/hooks.ts
```diff
@@ -0,0 +1,27 @@
+import { useSelector } from 'react-redux';
```
PeKne: Please change name of this file to `useSendAmountTransformers`. Every file should should contain only one hook. You can create parent `hooks/` folder if you expect that there will be more hooks in the future.

suite-native/state/src/extraDependencies.ts
```diff
@@ -26,7 +30,8 @@ const transports = transportsPerDeviceType[deviceType];
 export const extraDependencies: ExtraDependencies = mergeDeepObject(extraDependenciesMock, {
     selectors: {
         selectEnabledNetworks: selectEnabledDiscoveryNetworkSymbols,
-        selectBitcoinAmountUnit: () => PROTO.AmountUnit.BITCOIN,
+        selectBitcoinAmountUnit: selectBitcoinUnits,
```
PeKne, line 35: what are these two extra dependencies selectors used for now? Updating `selectBitcoinAmountUnit` make sense to me, because it at least mimics behaviour of desktop (even though I believe it is not used anywhere now). On the other hand `selectAreSatsAmountUnit` is not implemented in desktop so IMO it does not have place here.

PeKne (changes requested):
Looks good. Tested and works smoothly. Please see these two minor comments:

suite-native/module-send/src/hooks.ts
yanascz: Done.

suite-native/state/src/extraDependencies.ts
yanascz: The `selectAreSatsAmountUnit` selector is actually used in `composeBitcoinTransactionFeeLevelsThunk` and thus we need it to get the fees calculated correctly.

PeKne: You got me. It is correct ✅

PeKne (approved):

suite-native/module-send/src/sendFormThunks.ts
```diff
@@ -87,6 +89,21 @@ export const cleanupSendFormThunk = createThunk(
...
+export const cleanupBitcoinSendFormsThunk = createThunk(
+    `${SEND_MODULE_PREFIX}/cleanupBitcoinSendFormsThunk`,
+    (_, { dispatch, getState }) => {
+        const sendFormDrafts = selectSendFormDrafts(getState());
+        const accountKeys = Object.keys(sendFormDrafts);
+
+        accountKeys.forEach(accountKey => {
+            const account = selectAccountByKey(getState(), accountKey);
+            if (account && hasNetworkFeatures(account, 'amount-unit')) {
+                dispatch(cleanupSendFormThunk({ accountKey }));
```
PeKne:
```suggestion
                dispatch(sendFormActions.removeDraft({ accountKey }));
```

should be enough. Your implementation does not do any harm now, but might later if is the `cleanupSendFormThunk` modified or the send flow navigation between screens changed.

suite-native/module-send/src/sendFormThunks.ts
```diff
@@ -87,6 +89,21 @@ export const cleanupSendFormThunk = createThunk(
     },
 );
 
+export const cleanupBitcoinSendFormsThunk = createThunk(
```
PeKne: I would prefer a better name. This suggests it cleans up either:
1. only BTC send forms
2. send forms of all BTC-like coins

None of the above applies. I suggest something like cleanupSatoshisSuportingSendFormsThunk etc.

suite-native/module-send/src/sendFormThunks.ts
```diff
@@ -87,6 +89,21 @@ export const cleanupSendFormThunk = createThunk(
...
+    `${SEND_MODULE_PREFIX}/cleanupBitcoinSendFormsThunk`,
+    (_, { dispatch, getState }) => {
+        const sendFormDrafts = selectSendFormDrafts(getState());
```
PeKne: just an idea, decide yourself if you like it:

Create a selector that returns only the accounts that support satoshis and have any send form draft. Then you would not have to filter the array inside of this thunk.

suite-native/module-send/src/sendFormThunks.ts
yanascz: OK, makes sense.

suite-native/module-send/src/sendFormThunks.ts
yanascz: True, `removeSendFormDraftsSupportingAmountUnitThunk` it is.

suite-native/module-send/src/sendFormThunks.ts
yanascz: I don't want to create anything tailored for this since we don't need it elsewhere.

vytick: this selector can be placed in this package and not exported outside

PeKne (approved):

suite-native/module-send/src/sendFormThunks.ts
yanascz: What will be the added value of it then? I'm using the same selectors as in `convertSendFormDraftsBtcAmountUnitsThunk` and I don't think it'll ever be used elsewhere. If yes, we can always extract it.

Anyway, if I'm missing something, I'm all ears.

PR #14293, opened 2024-09-11
fix(suite-native): fix extra line breaks on some Android phones

### LINKED ISSUE #14177: Wrong line break of fiat values on dashboard

matejkriz:
## Description

On some Android devices (reported from Xiaomi) there were extra line breaks on some places i.e. portfolio balance, dashboard asset fiat value.

Only some specific values had that bug and only on some font scales.

Bug reported https://github.com/facebook/react-native/issues/46436

~As a workaround automatic font scaling is disabled and handled manually.~ (Didn't work well, just line breaking on other places.)

As a workaround custom letter spacing is disabled for crypto and fiat values on Android.

## Related Issue

Resolve [#14177](https://github.com/trezor/trezor-suite/issues/14177)

## Notes for QA

- Test different system font scales.

### DIFF
--- a/suite-native/formatters/package.json
+++ b/suite-native/formatters/package.json
@@ -21,6 +21,7 @@
         "@suite-common/wallet-utils": "workspace:*",
         "@suite-native/atoms": "workspace:*",
         "@suite-native/settings": "workspace:*",
+        "@trezor/env-utils": "workspace:*",
         "@trezor/styles": "workspace:*",
         "@trezor/utils": "workspace:*",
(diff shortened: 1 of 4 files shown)


Nodonisko (approved):
Approving but not sure if it won't be issue without that prop

Nodonisko (commented):
Lgtm

PR #14297, opened 2024-09-11
Fix/native/issues with unsupported coins

### LINKED ISSUE #14289: Coin disappeared from view only wallet after connecting device not supporting said coin
**Describe the bug**
When I connect first TT and activate for example BTC and VTC and I enable view only for it then I connect my TS5 that doesn't actually have VTC it correctly shows only BTC on dashboard and in the settings. BUT when I select my view only TT I can only see on the dashboard my BTC but in the settings I can see that VTC is enabled too so I should be able to see it on dashboard as well.
Additional info: When I connect TT back, the VTC is discovered again.
Another info: When I enable only VTC when I first connect my TT and then I connect TS5, the screen will get stuck on the "Connecting" screen.

**Info:**

-   Mobile Suite Version: 24.9.1. 692d09c
-   OS: Android

**How to reproduce**

1. Connect TT
2. Select for example BTC and VTC in the coin enable settings
3. Enable view only for the device
4. Connect TS5 
5. The discovery will only discover BTC (which is correct)
6. In the coin enable settings I can see correct coins for TS5
7. When I switch back to TT I can see only BTC in the dashboard
8. When I check the coin enable settings I can see that BTC and VTC is enabled

**Video:**
https://youtu.be/wpoP8STN5t8

vytick:
- dont use only enabled coins supported by current device for deleting unselected networks
 - hide device loading for devices with no enabled supported coin

## Related Issue

Resolve #14289

## Screenshots:

https://github.com/user-attachments/assets/bc0668c2-e6d7-4c88-b300-dd5e43557aab

### DIFF
--- a/suite-native/state/src/extraDependencies.ts
+++ b/suite-native/state/src/extraDependencies.ts
@@ -25,6 +25,8 @@ const transports = transportsPerDeviceType[deviceType];
 
 export const extraDependencies: ExtraDependencies = mergeDeepObject(extraDependenciesMock, {
     selectors: {
+        // using all enabled networks even those current device does not support,
+        // otherwise disableAccountsThunk might erase accounts not supported by current device
         selectEnabledNetworks: selectEnabledDiscoveryNetworkSymbols,
         selectBitcoinAmountUnit: () => PROTO.AmountUnit.BITCOIN,
         selectLocalCurrency: selectFiatCurrencyCode,
--- a/suite-native/coin-enabling/src/components/DiscoveryCoinsFilter.tsx
+++ b/suite-native/coin-enabling/src/components/DiscoveryCoinsFilter.tsx
@@ -10,7 +10,7 @@ import { Icon } from '@suite-common/icons-deprecated';
 import {
     applyDiscoveryChangesThunk,
     selectDiscoverySupportedNetworks,
-    selectEnabledDiscoveryNetworkSymbols,
+    selectDeviceEnabledDiscoveryNetworkSymbols,
 } from '@suite-native/discovery';
 
 import { NetworkSymbolSwitchItem } from './NetworkSymbolSwitchItem';
@@ -25,7 +25,7 @@ export const DiscoveryCoinsFilter = ({
     allowChangeAnalytics = true,
 }: DiscoveryCoinsFilterProps) => {
     const dispatch = useDispatch();
-    const enabledNetworkSymbols = useSelector(selectEnabledDiscoveryNetworkSymbols);
+    const enabledNetworkSymbols = useSelector(selectDeviceEnabledDiscoveryNetworkSymbols);
     const availableNetworks = useSelector(selectDiscoverySupportedNetworks);
 
     useFocusEffect(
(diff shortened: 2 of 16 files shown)


vytick:
I'll have to resolve circular dependency in device first

suite-native/device/src/hooks/useDelayedNavigation.ts
```diff
@@ -41,8 +49,14 @@ export const useDelayedNavigation = () => {
...
 
     // If Device is authorized and loading accounts, redirect to the Home screen.
+    // If connected device has no supported networks and coin enabling init was finished,
+    // also redirect to the Home screen otherwise user would be blocked forever.
+    // This can happen if user enabled only COIN on device A and then connects device B which does not support COIN.
     useEffect(() => {
-        if (isDeviceReadyToUseAndAuthorized && isTimeoutFinished) {
+        if (
+            (isDeviceReadyToUseAndAuthorized && isTimeoutFinished) ||
+            (deviceEnabledDiscoveryNetworkSymbols.length === 0 && isCoinEnablingInitFinished)
```
PeKne, line 57: should not be coin enabling screen showed instead if user has no coins enabled?

suite-native/state/src/extraDependencies.ts
```diff
@@ -25,6 +25,8 @@ const transports = transportsPerDeviceType[deviceType];
 
 export const extraDependencies: ExtraDependencies = mergeDeepObject(extraDependenciesMock, {
     selectors: {
+        // using all enabled networks even those current device does not support,
```
PeKne, line 28: why is this comment here? Should not it be rather at the place where is this extra dep used?

PeKne (commented):
The code looks good.  Not tested (Will do tomorrow morning if no one else does it today).

suite-native/state/src/extraDependencies.ts
vytick: It feels obsolete to have it in disableAccountsThunk or other places (we don use desktop discovery), because that is the place where you actually expect this behaviour. I've added the info here to make sure that we know why this selector was used over the device specific one.

suite-native/device/src/hooks/useDelayedNavigation.ts
vytick: Because this will show empty dashboard where is prompt to add new account and this will also enable the network.

vytick: But this part causes circular dependency and I have to think how to get rid of it.

vytick: I've moved this hook in 75a57dc15c to the package to where it is used to solve circular dependency

PeKne (approved):
One minor problem. Could you please rename the `useDelayedNavigation` so it better describes its purpose? It is a very general name, and without inspecting the implementation, it is tough to guess what it does.

vytick:
> One minor problem. Could you please rename the useDelayedNavigation so it better describes its purpose? It is a very general name, and without inspecting the implementation, it is tough to guess what it does.

👍  Good idea, renaming it to `useOnDeviceReadyNavigation`

PR #14298, opened 2024-09-11
chore: remove prettier-eslint dependency, upgrade @typescript-eslint

peter-sanderson:
- `prettier-eslint` seems to be unused so I removed it
- `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` upgraded to `8.5.0`


packages/suite-web/e2e/tests/analytics/events.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
```
peter-sanderson, line 1: This is because stuff like:

`expect(suiteReadyEvent.osName).to.exist.and.not.to.be.empty;` is used in Cypress

Nodonisko (approved):
lgtm

PR #14304, opened 2024-09-12
test(suite-native): e2e test improvements

PeKne:
## Description:
- POC of mocking of the `TrezorConnect.getAccountInfo` method so the E2E test does not have to rely on flaky network backends.
- If any test fails, the Jest retry mechanism is used instead of the Detox retry. That means only the individual test is retried, not the whole test suite.
- `waitFor` timeouts exceed at some places to avoid test flakiness.
- Accounts import optimized so the app is not restarted between the tests. Leads to time optimization.


suite-native/app/metro.config.js
```diff
@@ -54,6 +56,13 @@ const config = {
                 };
             }
 
+            if (process.env.IS_DETOX_BUILD && moduleName === '@trezor/connect') {
```
matejkriz, line 57: This looks a bit overcomplicated to me and it is global for all tests. Have you tried `jest.mock`?

matejkriz: Aha, I see it is probably not working in Detox, right? https://wix.github.io/Detox/docs/guide/mocking/

PeKne: Yes, it does not :/

PeKne:
@matejkriz this PR was rebased and marked as ready for review. Please look at it and if there is no concern, let's merge it.

suite-native/app/metro.config.js
```diff
@@ -4,6 +4,8 @@ const nodejs = require('node-libs-browser');
 const { getSentryExpoConfig } = require('@sentry/react-native/metro');
 const { mergeConfig } = require('@react-native/metro-config');
 
+console.log(process.env.IS_DETOX_BUILD);
```
matejkriz: If we really want the log here, there should be some explanation message prepended...

PeKne: removed

.github/workflows/test-suite-native-e2e-android.yml
```diff
@@ -158,11 +158,11 @@ jobs:
...
           emulator-options: -no-window -gpu swiftshader_indirect -no-snapshot -noaudio -no-boot-anim -grpc 8554
-          script: yarn test:e2e android.emu.release --headless --take-screenshots failing --record-videos failing --retries 2
+          script: yarn test:e2e android.emu.release --headless --take-screenshots failing --record-videos failing
 
       - name: "Store failed test screenshot artifacts"
         if: ${{failure()}}
         uses: actions/upload-artifact@v4
         with:
-          name: failed-android-tests-screenshots
+          name: failed-android-tests
```
matejkriz, line 167: 👏

matejkriz (approved):
Looks good to me, tests passing, let's merge it!

PR #14104, opened 2024-09-03
feat(coinmarket): rename exchange to swap

adderpositive:
## Description
To follow industry standards where most self-custody wallets call coin-to-coin exchange a “swap”. Additionally, in testing, we saw a significant uplift in confidence and easier-to-find trade features when the navigation buttons were separate for fiat-crypto and coin-to-coin operations.
Add also an analytics event on Swap - `wallet-coinmarket-exchange` (`accounts/actions`)

## Solution
- Account detail - header: Split Trade button into “Buy & Sell” (it will lead to the Buy tab) and “Swap” (it will lead to the former Exchange tab)
- Account detail - trade box: Rename Exchange for Swap
- Trade screen tabs: Rename Exchange for Swap
- Rename input and transaction labels
- Update modal content for CEX and DEX

### Split buttons
image
### Replaced button title from `Exchange` to `Swap`
image
### Replaced label from `Exchange amount` to `Swap amount`
image
### Replaced tab title from `Exchange` to `Swap` and updated icon
image
### Replaced `Exchange` button title with `Swap`
image
### Replaced `exchange` strings with `swap` (CEX)
image
### Replaced `exchange` strings with `swap` (DEX)
image
### Replaced `Exchange` label with `Swap` (DEX)
image

## Related Issue
Resolve https://github.com/trezor/trezor-suite/issues/14103

PR #14307, opened 2024-09-12
Revert "feat(coinmarket): rename exchange to swap"

MiroslavProchazka:
Reverts trezor/trezor-suite#14104

Reverting this PR for the time being until the following issues are resolved:

### Responsiveness
- [ ] When decreasing the screen width, the Send + Receive buttons should always remain visible. The Buy, Sell, and Swap buttons should disappear first into a dots. The Send + Receive buttons must be the last to wrap.
- [ ] Buttons should not wrap vertically like this.
image
image

### Icons
- [ ] Icons need to be different (the coin icon is used elsewhere, and the refresh icon should not be used for Swap).”

	
image


jvaclavik (approved):

Hermez-cz:
The 'Buy & Sell' button could be shortened to 'Buy/Sell.'

Text within buttons should not wrap but remain on one line.

As the window shrinks, buttons should never squish or overlap. Instead, they should move into the [...] menu in this order: [Buy/Sell] first, [Swap] second. The [Send | Receive] button group should never be split, so it should disappear last, and all at once. (Before that, it should be minimized to arrow icons, though this change isn't necessarily part of this PR.)

PR #14320, opened 2024-09-12
feat(coinmarket): possibility to proceed with unverified address

### LINKED ISSUE #14310: Coinmarket - Allow buy without connected Trezor device
## Issue

When a user gets to the second step in the Buy flow where they select a specific address to receive crypto and the Trezor is disconnected, Suite only allows to Show an unverified address and copy it so the user is stuck.
obrazek

## Solution
Instead of “Show unverified address” in this dialog show “Proceed with unverified address” button, close the dialog, and allow a user to proceed to the partner’s checkout.

obrazek

adderpositive:
## Issue

When a user gets to the second step in the Buy flow where they select a specific address to receive crypto and the Trezor is disconnected, Suite only allows to Show an unverified address and copy it so the user is stuck.

### Design before
obrazek

## Solution
Instead of “Show unverified address” in this dialog show “Proceed with unverified address” button, close the dialog, and allow a user to proceed to the partner’s checkout.

### Design after
obrazek
obrazek

## Related
Resolve https://github.com/trezor/trezor-suite/issues/14310

### DIFF
--- a/packages/suite-data/files/translations/en.json
+++ b/packages/suite-data/files/translations/en.json
@@ -1700,6 +1700,7 @@
   "TR_SHOW_MORE_ADDRESSES": "Show more ({count})",
   "TR_SHOW_UNVERIFIED_ADDRESS": "Show unverified address",
   "TR_SHOW_UNVERIFIED_XPUB": "Show unverified public key",
+  "TR_PROCEED_UNVERIFIED_ADDRESS": "Proceed with unverified address",
   "TR_SIDEBAR_ADD_COIN": "Add a coin",
   "TR_SIGN": "Sign",
   "TR_SIGNATURE": "Signature",
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConfirmUnverifiedModal.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConfirmUnverifiedModal.tsx
@@ -1,53 +1,35 @@
-import { useEffect } from 'react';
-import styled from 'styled-components';
-
-import { applySettings } from 'src/actions/settings/deviceSettingsActions';
-import { Translation, Modal } from 'src/components/suite';
+import { Translation } from 'src/components/suite';
 import { TranslationKey } from 'src/components/suite/Translation';
 import { useDevice, useDispatch, useSelector } from 'src/hooks/suite';
-import { ThunkAction } from 'src/types/suite';
-import { Button } from '@trezor/components';
+import { Dispatch, GetState } from 'src/types/suite';
+import { Button, H3, NewModal, Paragraph } from '@trezor/components';
 import { onCancel } from 'src/actions/suite/modalActions';
 import { selectDeviceLabelOrName } from '@suite-common/wallet-core';
-
-const StyledModal = styled(Modal)`
-    width: 520px;
-`;
-
-// eslint-disable-next-line local-rules/no-override-ds-component
-const StyledButton = styled(Button)`
-    flex-grow: 1;
-`;
+import { applySettings } from 'src/actions/settings/deviceSettingsActions';
+import { useEffect } from 'react';
 
 interface ConfirmUnverifiedModalProps {
-    showUnverifiedButtonText: TranslationKey;
-    showUnverified: () => ThunkAction;
-    verify: () => ThunkAction;
+    action: {
+        event: () => (dispatch: Dispatch) => void;
+        title: TranslationKey;
+        closeAfterEventTriggered?: boolean;
+    };
+    verifyProcess?: () => (dispatch: Dispatch, getState: GetState) => Promise<void>;
     warningText: TranslationKey;
 }
 
 export const ConfirmUnverifiedModal = ({
-    showUnverifiedButtonText,
-    showUnverified,
-    verify,
+    action,
     warningText,
+    verifyProcess,
 }: ConfirmUnverifiedModalProps) => {
     const deviceLabel = useSelector(selectDeviceLabelOrName);
+    const { device } = useDevice();
     const dispatch = useDispatch();
-    const { device, isLocked } = useDevice();
-
-    // Device connected while the modal is open -> switch to verification modal.
-    useEffect(() => {
-        if (device?.connected) {
-            dispatch(verify());
-        }
-    }, [device?.connected, dispatch, verify]);
-
-    // just to make TS happy
-    if (!device) return null;
+    const { isLocked } = useDevice();
 
     const isDeviceLocked = isLocked();
-    const isPassphraseRequired = device.connected && !device.available;
+    const isPassphraseRequired = device?.connected && !device.available;
     const deviceStatus = isPassphraseRequired
         ? 'TR_DEVICE_LABEL_IS_UNAVAILABLE'
         : 'TR_DEVICE_LABEL_IS_NOT_CONNECTED';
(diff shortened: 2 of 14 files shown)


packages/suite-data/files/translations/en.json
```diff
@@ -572,6 +572,7 @@
   "TR_COINMARKET_KYC_YES_REFUND": "KYC requested in exceptional cases. KYC not required for refunds. 🤝",
   "TR_COINMARKET_KYC_NO_KYC": "KYC never required. Exceptional cases automatically refunded. 👍",
   "TR_COINMARKET_KYC_DEX": "KYC never required. DEX swaps either succeed or fail. 👍",
+  "TR_COINMARKET_CONFIRM_ADDRESS": "Confirm address",
```
komret: This string is generic enough that it could be used elsewhere in the app, let's just call it `TR_CONFIRM_ADDRESS`.

BTW you don't have to edit these files, adding a new string to `messages.ts` is enough.

packages/suite/src/views/wallet/coinmarket/common/CoinmarketSelectedOffer/CoinmarketVerify/CoinmarketVerify.tsx
```diff
@@ -238,7 +253,13 @@ export const CoinmarketVerify = ({ coinmarketVerifyAccount, currency }: Coinmark
                                     }
                                 }}
                             >
-                                <Translation id="TR_BUY_CONFIRM_ON_TREZOR" />
+                                <Translation
+                                    id={
+                                        !device?.connected
+                                            ? 'TR_COINMARKET_CONFIRM_ADDRESS'
+                                            : 'TR_CONFIRM_ON_TREZOR'
```
komret: nit: might be easier to read without the negation:
```
device?.connected
    ? 'TR_CONFIRM_ON_TREZOR'
    : 'TR_COINMARKET_CONFIRM_ADDRESS'
```

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ProceedUnverifiedAddressModal.tsx
komret, line 1: A new component might be unnecessary. Could you use `ConfirmUnverifiedModal` instead? It could be updated to use the `NewModal` while you're at it.

komret (changes requested):

Hermez-cz:
Change the modal component to the warning one please: https://www.figma.com/design/dYIRTk1qVcKsrJDqOzZ7SG/Modals?node-id=4037-2006&t=6JHZmmhTkl2jrwnV-1

Agreed with the UX designer that the yellow warning state should use only together with the icon as seen in the design.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ProceedUnverifiedAddressModal.tsx
adderpositive: I'm not sure about that. I think the `ConfirmUnverifiedModal` would become messier if more logic were added. What do you think?

komret: I just checked, it wouldn't really change much. Minor change to props. Updating that modal to new design would be a nice bonus.

packages/suite/src/views/wallet/coinmarket/common/CoinmarketSelectedOffer/CoinmarketVerify/CoinmarketVerify.tsx
```diff
@@ -127,6 +132,15 @@ export const CoinmarketVerify = ({ coinmarketVerifyAccount, currency }: Coinmark
...
 
+    // close modals and reset addressVerified on device connection change
+    useEffect(() => {
+        dispatch({
+            type: COINMARKET_BUY.VERIFY_ADDRESS,
+            addressVerified: undefined,
+        });
+        dispatch(modalActions.onCancel());
+    }, [device?.connected, dispatch]);
+
```
komret, line 142: Perhaps this should be moved inside the modal. It is only related to `ProceedUnverifiedAddressModal`, isn't it?

If you use `ConfirmUnverifiedModal`, the `useEffect` is already there.

adderpositive: Yes, it is related. But it is not possible to clear the  `addressVerified`, if the modal is not opened.

komret: Ok I thought `addressVerified` was only set inside the modal.

adderpositive: It is only set in the modal. But It is also necessary to set the value as `undefined` if the modal is not active and the connection of Trezor will change.

adderpositive:
I have created a new component `ConfirmUnverifiedModalContainer` where I moved all logic of _unverified_ modals. It handles **receive**, **xpub**, and **proceed** modals. What do you think about that?

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/UserContextModal.tsx
```diff
@@ -41,12 +39,12 @@ import {
     UnhideTokenModal,
 } from 'src/components/suite/modals';
 import type { AcquiredDevice } from 'src/types/suite';
-import { openXpubModal, showXpub } from 'src/actions/wallet/publicKeyActions';
 import type { ReduxModalProps } from '../ReduxModal';
 import { CryptoId } from 'invity-api';
 import { EverstakeModal } from './UnstakeModal/EverstakeModal';
 import { PassphraseMismatchModal } from './PassphraseMismatchModal';
 import { FirmwareRevisionOptOutModal } from './FirmwareRevisionOptOutModal';
+import { ConfirmUnverifiedModalContainer } from 'src/components/suite/modals/ReduxModal/UserContextModal/ConfirmUnverifiedModalContainer';
```
tomasklim: nit: Import from `'src/components/suite/modals'`

tomasklim (changes requested):

tomasklim:
Remembered wallet, `Confirm address` is spinning

https://github.com/user-attachments/assets/cfdce460-8b03-408f-af89-ca12909aa939

adderpositive:
> Remembered wallet, `Confirm address` is spinning
> Screen.Recording.2024-09-24.at.15.06.10.mov

it is happening when you are not connected

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/UserContextModal.tsx
adderpositive: done

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConfirmUnverifiedModal.tsx
```diff
@@ -53,47 +34,50 @@ export const ConfirmUnverifiedModal = ({
...
+                    <Button onClick={handleClose} variant="tertiary">
                         <Translation id="TR_BACK" />
                     </Button>
                 </>
             }
-        />
+        >
+            <H3>
+                <Translation id={deviceStatus} values={{ deviceLabel: device.label }} />
+            </H3>
```
komret: Could you use the `heading` prop?

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConfirmUnverifiedModalContainer.tsx
komret, line 1: Certainly an improvement, although I'd prefer to see the shared logic in `ConfirmUnverifiedModal` and have two or three wrappers with specifics, kind of like `ConfirmUnverifiedAddress` worked. This looks like three components in one with little code overlap between them. The effect and `enablePassphraseAndContinue` could be in `ConfirmUnverifiedModal `, unless I am missing something.

komret (changes requested):

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConfirmUnverifiedModal.tsx
adderpositive: The prop `heading` will add the title to the modal's header. And it doesn't look good in order header -> icon -> paragraph.

komret: What does @adamhavel think?

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConfirmUnverifiedModalContainer.tsx
adderpositive: updated

komret (approved):

tomasklim (approved):

PR #14323, opened 2024-09-12
fix(suite-native): do not display fiat amount for testnet coins

yanascz:
## Screenshots:
image
image
image

### DIFF
--- a/suite-native/formatters/src/components/CryptoToFiatAmountFormatter.tsx
+++ b/suite-native/formatters/src/components/CryptoToFiatAmountFormatter.tsx
@@ -1,17 +1,15 @@
 import { TextProps } from '@suite-native/atoms';
 import { NetworkSymbol } from '@suite-common/wallet-config';
-import { useFormatters } from '@suite-common/formatters';
 
 import { FormatterProps } from '../types';
-import { AmountText } from './AmountText';
 import { useFiatFromCryptoValue } from '../hooks/useFiatFromCryptoValue';
+import { FiatAmountFormatter } from './FiatAmountFormatter';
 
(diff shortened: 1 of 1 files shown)


suite-native/formatters/src/components/CryptoToFiatAmountFormatter.tsx
```diff
@@ -34,7 +29,5 @@ export const CryptoToFiatAmountFormatter = ({
         cryptoValue: value,
     });
 
-    const formattedFiatValue = FiatAmountFormatter.format(fiatValue ?? '0');
-
-    return <AmountText value={formattedFiatValue} isDiscreetText={isDiscreetText} {...textProps} />;
+    return <FiatAmountFormatter network={network} value={fiatValue} {...textProps} />;
```
vytick, line 32: why to exclude `isDiscreetText`?

yanascz: It's not used anywhere with this component and the default is the same as in `FiatAmountFormatter`. It's practically dead code. ☠️

PeKne (approved):

suite-native/formatters/src/components/CryptoToFiatAmountFormatter.tsx
vytick: makes sense 👍

STew790:
**QA OK**
I don't see fiat value anywhere.

**Info:**
24.9.1 https://github.com/trezor/trezor-suite/commit/306d448b5e568d81359228294e847c80cea6c25e

PR #14327, opened 2024-09-13
chore: NX output style

martykan:
## Description

Trying to change NX output style to work like before, to include output of individual tasks


Lemonexe (approved):
Seems to work fine, I made `type-check` [fail with this change](https://github.com/trezor/trezor-suite/actions/runs/10878898368/job/30182695466?pr=14349), and there is now the error description.

PR #14328, opened 2024-09-13
fix(coinmarket): responsive menu and icons 

### LINKED ISSUE #14331: Coinmarket - incorrect menu/tab icons & responsive behavior
## Responsiveness
- When decreasing the screen width, the Send + Receive buttons should always remain visible. The Buy, Sell, and Swap buttons should disappear first into a dots. The Send + Receive buttons must be the last to wrap.
- Buttons should not wrap vertically like this.
image
image

## Icons
- Icons need to be different (the coin icon is used elsewhere, and the refresh icon should not be used for Swap).”

	
image

adderpositive:
### Responsive
- When decreasing the screen width, the Send + Receive buttons should always remain visible. The Buy, Sell, and Swap buttons should disappear first into a dots. The Send + Receive buttons must be the last to wrap.
- Buttons should not wrap vertically like this.
image
image

### Icons
- Icons need to be different (the coin icon is used elsewhere, and the refresh icon should not be used for Swap).”
image

## After fix
obrazek
obrazek
obrazek
- hidden `Buy & Sell` under 1200px of device window width, it's in the submenu (`...`) 

## Related issues
Resolve https://github.com/trezor/trezor-suite/issues/14331

### DIFF
--- a/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/HeaderActions.tsx
+++ b/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/HeaderActions.tsx
@@ -124,7 +124,7 @@ export const HeaderActions = () => {
             {isCoinmarketAvailable && (
                 <AppNavigationTooltip>
                     <ButtonComponent
-                        icon="coinVertical"
+                        icon="currencyCircleDollar"
                         onClick={() => {
                             goToWithAnalytics('wallet-coinmarket-buy', { preserveParams: true });
                         }}
@@ -136,7 +136,7 @@ export const HeaderActions = () => {
                         <Translation id="TR_COINMARKET_BUY_AND_SELL" />
                     </ButtonComponent>
                     <ButtonComponent
-                        icon="arrowsClockwise"
+                        icon="arrowsLeftRight"
                         onClick={() => {
                             goToWithAnalytics('wallet-coinmarket-exchange', {
                                 preserveParams: true,
--- a/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/HeaderActions.tsx
+++ b/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/HeaderActions.tsx
@@ -9,6 +9,7 @@ import {
     IconButton,
     IconButtonProps,
     IconName,
+    variables,
 } from '@trezor/components';
 import { spacingsPx } from '@trezor/theme';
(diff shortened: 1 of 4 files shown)


Hermez-cz:
Hi, thanks! How does [Swap] behive when resizing please?

Lemonexe (commented):
Looks ok, but since you're doing this, won't it make sense to do the same for Swap? :thinking:
It has the same problem; on some widths swap is visible, and rest is cut off:
Screenshot from 2024-09-17 08-55-34

:thought_balloon: Though it's a completely different question why this header cuts off content instead of wrapping. But that does not concern this PR, it's preexisting :upside_down_face:   Tagging @jvaclavik

tomasklim:
@adderpositive could you please check conflicts 🙏

adderpositive:
@tomasklim solved

packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/HeaderActions.tsx
```diff
@@ -18,13 +18,21 @@ import { useDevice, useDispatch, useSelector } from 'src/hooks/suite';
 import { goto } from 'src/actions/suite/routerActions';
 import { AppNavigationTooltip } from 'src/components/suite/AppNavigation/AppNavigationTooltip';
 import { selectSelectedAccount } from 'src/reducers/wallet/selectedAccountReducer';
+import { SCREEN_QUERY } from '@trezor/components/src/config/variables';
```
jvaclavik: Please use:
```
import { variables } from '@trezor/components';
```

and use it this way:
```
variables.SCREEN_QUERY
```

jvaclavik:
btw is this just low-quality screenshot or border-top is missing here?

Screenshot 2024-09-24 at 16 10 27

adderpositive:
@jvaclavik low quality it is

packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/HeaderActions.tsx
adderpositive: Nice

adderpositive: updated

tomasklim (approved):
LGTM. Receive and Send button sometimes dissapear but thats not an issue of this PR

Issue #14285, opened 2024-09-11
Not able to communicate with two devices at once when using node bridge


bosomt:
**Describe the bug**
Not able to communicate with two devices at once when using node bridge.
You can break communication easily when using two devices at once.
Even reconnect of second device does not help.
Sometimes it works sometimes it dont, so im not able to replicate it everytime

**Steps to reproduce:**
1. Connect two devices /TT and TS5 were used/
2. Start suite with EAP /node bridge enabled/ 
3. Discover both wallets
4. Reload Suite

**Info:**
 - Suite version: desktop 24.9.1 (82c6d9b468f417d99f99d90b3c77d0f02650616b)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.9.1 Chrome/124.0.6367.243 Electron/30.3.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.2 regular (revision 37006592dbc3f2d817def0c75ed6a62513895ec5)
 - Transport: BridgeTransport 3.0.0

**Screenshots:**
<img width="876" alt="image" src="https://github.com/user-attachments/assets/27d434d7-0480-4369-a73c-c12e13766eb6">

![image](https://github.com/user-attachments/assets/1c141288-81fe-4cde-9e62-b37400a909d2)

![image](https://github.com/user-attachments/assets/44d2216a-f2df-4c7d-ab20-470b47e69af8)

mroz22:
Somewhat simplified steps:

1. 2 devices, connected unlocked
2. start suite-desktop
3. observe that only one device is emitted to suite

so far I was able to observe that this enumeration here
https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/device/DeviceList.ts#L336
correctly returns 2 descriptors
but things start to be weird roughly here 
https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/device/DeviceList.ts#L215 
where `_createAndSaveDevice` is called only once. 

maybe @marekrjpolak could take a quick look please? it could be related to your recent refactorings

marekrjpolak:
![image](https://github.com/user-attachments/assets/e81bbf69-1919-45e5-9ff0-a824ef6fff27)

bosomt:
Managed to replicate it in Brave browser when webUSB is used as transport, 

[...]

PR #14335, opened 2024-09-13
transport-test: add case for multi devices

mroz22:
- [1cb3186](https://github.com/trezor/trezor-suite/pull/14335/commits/1cb31869b363ed4b93dca09632c5aae69e6930e4)
  - add support for running a test with multiple connected hw devices at the same time 
- [1cb3186](https://github.com/trezor/trezor-suite/pull/14335/commits/1cb31869b363ed4b93dca09632c5aae69e6930e4)
  - while I haven't been able to reproduce the behavior reported in https://github.com/trezor/trezor-suite/issues/14285 and this tests seems to be passing quite nicely (local only) I'd say that it already gives us some defensive capabilities regarding future refactorings and fixes, such as  #14339

### DIFF
--- a/packages/transport-test/e2e/bridge/controller.ts
+++ b/packages/transport-test/e2e/bridge/controller.ts
@@ -26,6 +26,8 @@ class Controller extends TrezorUserEnvLinkClass {
     private logger: Console;
     private nodeBridge: TrezordNode | undefined = undefined;
 
+    private expectedNumOfDevices = 0;
+
     private originalApi: {
         connect: typeof TrezorUserEnvLinkClass.prototype.connect;
         startBridge: typeof TrezorUserEnvLinkClass.prototype.startBridge;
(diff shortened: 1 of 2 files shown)


packages/transport-test/e2e/bridge/controller.ts
```diff
@@ -89,22 +91,31 @@ class Controller extends TrezorUserEnvLinkClass {
...
+        const expected = this.expectedNumOfDevices;
         this.logger.log(
             `${env.USE_HW ? '[MANUAL ACTION REQUIRED] ' : ''} waiting for ${expected} device to be connected`,
         );
 
         return scheduleAction(
             async () => {
                 const devices = (await webusb.getDevices()).filter(d => d.productName === 'TREZOR');
-                if (devices.length === expected) {
+                if (devices.length >= expected) {
```
mroz22, line 118: check on this

packages/transport-test/e2e/bridge/controller.ts
```diff
@@ -26,6 +26,8 @@ class Controller extends TrezorUserEnvLinkClass {
     private logger: Console;
     private nodeBridge: TrezordNode | undefined = undefined;
 
+    private expectedNumOfDevices = 0;
```
martykan, line 29: Maybe this should be reset somewhere, when `stopEmu` is not called it will be incorrect

PR #14336, opened 2024-09-13
Release message system config from branches by environment

matejkriz:
## Description

Allow to release message system config only from predefined branches so it's easier to track in git what has been released where.

- `release-message-system-develop`: unrestricted branch, anyone (with write access) can push there anything and release on develop environment
- `release-message-system-production`: restricted, requires PR

Dispatching the action in github is still manual, but doesn't need environment specified because it's now derived from the branch.

### DIFF
--- a/.github/workflows/release-suite-message-system-config.yml
+++ b/.github/workflows/release-suite-message-system-config.yml
@@ -1,4 +1,4 @@
-name: "[Release] suite config"
+name: "[Release] suite message system config"
 
 permissions:
   id-token: write # for fetching the OIDC token
@@ -6,6 +6,7 @@ permissions:
 
 on:
   workflow_dispatch:
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
--- a/.github/workflows/release-suite-config.yml
+++ b/.github/workflows/release-suite-config.yml
@@ -6,26 +6,20 @@ permissions:
 
 on:
   workflow_dispatch:
-    inputs:
-      environment:
-        description: "Release environment"
-        type: choice
-        options:
-          - develop-message
-          - production-message
-        required: true
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
   release-config:
-    if: github.repository == 'trezor/trezor-suite'
-    environment: ${{ github.event.inputs.environment }}
+    if: ${{ github.repository == 'trezor/trezor-suite' && (github.ref == 'refs/heads/release-message-system-production' || github.ref == 'refs/heads/release-message-system-develop') }}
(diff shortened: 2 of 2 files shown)


MiroslavProchazka (approved):
Great, thank you!

komret (approved):

.github/workflows/release-suite-message-system-config.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+  workflow_dispatch:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  release-config:
+    if: ${{ github.repository == 'trezor/trezor-suite' && (github.ref == 'refs/heads/release-message-system-production' || github.ref == 'refs/heads/release-message-system-develop') }}
```
vdovhanych, line 16: Why having a separate branch for develop? I dont really see the advantage in this.

.github/workflows/release-suite-message-system-config.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ env.ROLE_TO_ASSUME  }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Build and sign ${{ env.RELEASE_ENV }} message-system config file
```
vdovhanych, line 34: I dont like it like this, this way we would have to have secret for develop and produtction enviroment present and configured i think. Or maybe the string would be just empty sinc the secret would not be in the env secrets.

.github/workflows/release-suite-message-system-config.yml
```diff
@@ -0,0 +1,44 @@
+name: "[Release] suite message system config"
+
+permissions:
+  id-token: write # for fetching the OIDC token
+  contents: read # for actions/checkout
+
+on:
+  workflow_dispatch:
```
vdovhanych, line 8: add empty line after this

vdovhanych (approved):
I think its ok but im not sure about the special branch for develop messages.

.github/workflows/release-suite-message-system-config.yml
matejkriz: This is the reason: https://github.com/trezor/trezor-suite/pull/14233 
From time to time we need to test something that should never go to production and better than reverting it is to not merge it to develop at all.

.github/workflows/release-suite-message-system-config.yml
matejkriz: Empty string would be ok...

.github/workflows/release-suite-message-system-config.yml
matejkriz: 857865d4ad93a8a7df1123901a12cab8301eb9b6

matejkriz:
Protection rules updated, I'm about to merge this cc @MiroslavProchazka @komret

PR #14341, opened 2024-09-13
Coin enabling settings disabled during discovery

### LINKED ISSUE #14275: Coin enabling - handle coin enabling during discovery
Do not allow to change coin settings during discovery. Block opening settings and show spinner/loader when discovery is running

[figma](https://www.figma.com/design/oQtgXoPIgGSlh01TLHnmeE/Coin-enabling?node-id=0-1&node-type=canvas&t=QZfImcKwcqfIj4Yw-0)

Screenshot 2024-09-12 at 15 48 55

matejkriz:
## Description

Check by commits. First I refactored and simplified ListItem a bit, It wasn't universal and it was used in Settings only and have unused props.

Second commit is adding the loading indicator.

There is an unhandled edge case when user have view-only wallet, change enabled coins, then goes out of the Coin Enabling screen to apply the change, goes back to Coin Enabling screen and then connect the device on that screen 🤪 I believe it's too much of an edge case to handle.

What I don't like on this solution is that if you enable some more coins and go back from that screen, you always see the loading and you can't go immediately back. But again, not worth handling now IMHO. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14275

## Screenshots:
imageimage


suite-native/module-dev-utils/src/screens/DevUtilsScreen.tsx
```diff
@@ -27,12 +27,10 @@ export const DevUtilsScreen = ({
...
                 <Card>
                     <VStack spacing="medium">
-                        {!isDebugEnv() && (
-                            <ListItem
-                                subtitle={`${getEnv()}-${getSuiteVersion()}, commit ${getCommitHash()}`}
-                                title="Build version"
-                            />
-                        )}
+                        <TitleHeader
+                            title="Build version"
```
vytick, line 31: Please, can you use intl here, while touching the code 🙏 ?

suite-native/module-settings/src/components/SettingsSectionItem.tsx
```diff
@@ -1,11 +1,53 @@
...
+    isLoading,
+}: SettingsSectionItemProps) => {
+    const { applyStyle, utils } = useNativeStyles();
+
+    const isDisabled = !onPress || isLoading;
+
+    return (
+        <TouchableOpacity disabled={isDisabled} onPress={onPress} testID={testID}>
+            <Box flexDirection="row">
+                {iconName && <SettingsSectionItemIcon iconName={iconName} />}
```
vytick: nit: 
It seems, that iconName is always set, so this can maybe be required param?

vytick (approved):
Works nicely 🚀. Love that you refactored the ListItem 👍 

I have one thing to discuss, but up to you whether to resolve it or no.

suite-native/module-dev-utils/src/screens/DevUtilsScreen.tsx
vytick: this is dev utils. no need to do that

suite-native/module-settings/src/components/SettingsSectionItem.tsx
```diff
@@ -1,11 +1,53 @@
...
+    title,
+    subtitle,
+    iconName,
+    onPress,
+    testID,
+    isLoading,
+}: SettingsSectionItemProps) => {
+    const { applyStyle, utils } = useNativeStyles();
+
+    const isDisabled = !onPress || isLoading;
```
vytick, line 36: no need to have here onPress, since default value for this is `() => void` making it always false

suite-native/module-settings/src/components/SettingsSectionItemText.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Box, Text } from '@suite-native/atoms';
+
+export type SettingsSectionItemTextProps = {
+    title: ReactNode;
+    subtitle?: ReactNode;
+};
+
+export const SettingsSectionItemText = ({ title, subtitle }: SettingsSectionItemTextProps) => {
+    return (
+        <Box justifyContent="center" flexDirection="column" flex={2}>
```
PeKne: Nit: No need for specific flex values as `2`. The other siblings in the parent flexBox have no flex value set. `flex 1` should be sufficient.
```suggestion
        <Box justifyContent="center" flexDirection="column" flex={1}>
```

Apart for this, LGTM 👏

matejkriz: https://github.com/trezor/trezor-suite/pull/14341/commits/86f647ce514f4d47df7c534495510902e9b80f66

suite-native/module-settings/src/components/SettingsSectionItem.tsx
matejkriz: https://github.com/trezor/trezor-suite/pull/14341/commits/86f647ce514f4d47df7c534495510902e9b80f66

PR #14345, opened 2024-09-15
fix(suite-native): make cardano discovery work correct

vytick:
Cardano discovery can fail on Connect methods to get derivations being called with `useCardanoDerivation === false` making discovery to set to stopped state and graph behave odd.

BY calling getFeatures, cardanoConnectPatch is applied and connect starts using useCardanoDerivation set to true. 

The other change makes sure that even if derivation fetch fails it does not set discovery state to stopped not making graph on mobile behave strange

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -201,8 +201,22 @@ const getCardanoSupportedAccountTypesThunk = createThunk(
             return undefined;
         }
         const availableCardanoDerivationsResponse = await requestDeviceAccess({
-            deviceCallback: () =>
-                dispatch(getAvailableCardanoDerivationsThunk({ deviceState, device })).unwrap(),
+            deviceCallback: async () => {
+                // calling method like getFeatures with keepSession: false (default) will make connect initialize again with useCardanoDerivation: true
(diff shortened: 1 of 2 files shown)


suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -201,8 +201,22 @@ const getCardanoSupportedAccountTypesThunk = createThunk(
             return undefined;
         }
         const availableCardanoDerivationsResponse = await requestDeviceAccess({
-            deviceCallback: () =>
-                dispatch(getAvailableCardanoDerivationsThunk({ deviceState, device })).unwrap(),
+            deviceCallback: () => {
+                // calling method like getFeatures will make connect useCardanoDerivation set to true with use other calls
+                // using cardanoConnectPatch otherwise getting Cardano derivation might fail (if useCardanoDerivation is false)
+                TrezorConnect.getFeatures({
```
Nodonisko: why there is no await?

vytick: good point. I’ll add it but it works anyway

vytick: 2e9e127b4e

Nodonisko (approved):

PR #14348, opened 2024-09-16
feat(suite-native): send form normal fee level validation

### LINKED ISSUE #13782: Send coins - fee level unsuficient balance
WIP 🚧: If the user fills out the send form and then selects a fee level that he does not have enough funds for, we should inform him about it and prevent him from continuing to the on-device review step.

PeKne:
## Description
- normal fee level is now part of send form validation schema
- If user selects an send amount X, where `(X + normal fee) > available balance` form shows error and user can not continue until he lowers the amount

## Related Issue

Resolve #13782

### DIFF
--- a/suite-native/module-send/src/screens/SendOutputsScreen.tsx
+++ b/suite-native/module-send/src/screens/SendOutputsScreen.tsx
@@ -1,5 +1,5 @@
 import { useDispatch, useSelector } from 'react-redux';
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useWatch } from 'react-hook-form';
 import Animated, { SlideInDown, SlideOutDown } from 'react-native-reanimated';
 
@@ -12,7 +12,6 @@ import {
     SendRootState,
     composeSendFormTransactionFeeLevelsThunk,
     selectAccountByKey,
-    selectDevice,
     selectNetworkFeeInfo,
     selectSendFormDraftByAccountKey,
     sendFormActions,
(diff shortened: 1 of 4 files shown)


suite-native/module-send/src/sendOutputsFormSchema.ts
```diff
@@ -42,7 +43,16 @@ const isAmountHigherThanBalance = (amount: string, context?: SendFormFormContext
...
-    const { networkSymbol, networkFeeInfo, availableAccountBalance, isValueInSats } = context;
+    const {
+        networkSymbol,
+        networkFeeInfo,
+        availableAccountBalance,
+        isValueInSats,
+        precomposedFeeLevels,
+    } = context;
+
+    // transaction amount + normal fee has to be lower than account balance
```
yanascz: How is this comment relevant to the following line? 🤔 It's not obvious at all.

suite-native/module-send/src/screens/SendOutputsScreen.tsx
```diff
@@ -44,6 +43,14 @@ const buttonWrapperStyle = prepareNativeStyle(utils => ({
     padding: utils.spacings.medium,
 }));
 
+const DEFAULT_VALUES = [
+    {
+        amount: '',
+        address: '',
+        fiat: '',
+    },
+] as const satisfies SendOutputsFormValues['outputs'];
```
yanascz, line 53: ❤️

suite-native/module-send/src/sendOutputsFormSchema.ts
PeKne: better comment added in 5189ad143f

suite-native/module-send/src/screens/SendOutputsScreen.tsx
```diff
@@ -117,32 +118,27 @@ export const SendOutputsScreen = ({
...
-            }),
-        );
-    }, [accountKey, dispatch, getValues]);
+    const precomposeTransactionFeeLevels = useCallback(async () => {
+        const formState = constructFormDraft(getValues());
 
-    useEffect(() => {
-        if (watchedFormValues && isValid) debounce(storeFormDraftIfValid);
-    }, [isValid, storeFormDraftIfValid, watchedFormValues, debounce]);
+        if (!account) return;
```
yanascz: Would it make sense to move this at the very beginning of the callback block and change like this?
```suggestion
        if (!account || !networkFeeInfo) return;
```

PeKne:
@yanascz In the end, the solution utilizes the calculation of send max with the normal fee level, as you suggested. So the normal fee send max amount is calculated right away when the user enters the send form screen and recalculated if the address is changed. Like this, we have a numeric value to compare with the amount, and we don't have to recalculate fees on every change. The solution of calculating send max was inspired by desktop implementation. Hit me up in Slack if you need to clarify anything. 

The original Commits were squashed and force-pushed because the new solution conflicted with the original one. Please do the full review again 🙏.

matejkriz (approved):
👏 It was hard for me to understand, but after a huddle consultation it makes sense to me now.

PR #14353, opened 2024-09-16
chore(suite-native): show testnet fiat amounts as plain zero

yanascz:
The blank space looks as a bug when testing some layouts.

## Screenshots:
image
image
image

image


matejkriz (approved):
👍

PeKne:
IMO, the zero without unit looks as UI bug to me :D. But as I said during today SU, no user will care about this. So feel free to merge 🙏.

yanascz:
Turns out this solves some inconsistencies but leads to different ones so the benefit is questionable. It’s even misleading in some cases.

PR #14357, opened 2024-09-16
fix(mobile): different total fiat across the app

Nodonisko:
## Description

I am not sure about different balance in device switcher but this is my best guess.

## Related Issue

Resolve


matejkriz (commented):
shouldIncludeStaking makes sense, but how can filtering by account visibility help?

Nodonisko:
I though that if some accounts are already discovered, after disabling it will only hide them but according to @vytick they are deleted so I will remove this part probably.

matejkriz (approved):

PR #14359, opened 2024-09-16
Chore(suite): use new Networks in suite utils

Lemonexe:
## Description

Refactoring deprecated `networksCompatibility` to `networks` in these low-hanging-fruit files:
- https://github.com/trezor/trezor-suite/pull/14359/commits/3e4116897cd3d7999300e89ab59761617899945f `useBitcoinAmountUnit`
- https://github.com/trezor/trezor-suite/pull/14359/commits/9a883b85ce93b4a3387e5b3a9c02759dff1c33f9 `router`
- https://github.com/trezor/trezor-suite/pull/14359/commits/00aa985e2fbe158e4edbde349f9da20f96e8488b `prepareNetworkSymbolFormatter`

:information_source: I think the 1st and 3rd don't need testing, but I did test the `router` on suite web to be sure: these routes load the specific accounts as they should:
- http://localhost:8000/accounts#/btc
- http://localhost:8000/accounts#/btc/0
- http://localhost:8000/accounts#/eth/0/normal
- http://localhost:8000/accounts#/btc/1/taproot
- http://localhost:8000/accounts#/asdf/3 (fallback to first account, same as http://localhost:8000/accounts)

## Related Issue

Part of #13839

### DIFF
--- a/suite-common/formatters/src/formatters/prepareNetworkSymbolFormatter.ts
+++ b/suite-common/formatters/src/formatters/prepareNetworkSymbolFormatter.ts
@@ -1,5 +1,5 @@
 import { UNIT_ABBREVIATIONS } from '@suite-common/suite-constants';
-import { networksCompatibility, NetworkSymbol } from '@suite-common/wallet-config';
+import { getNetwork, NetworkSymbol } from '@suite-common/wallet-config';
 
 import { FormatterConfig } from '../types';
 import { makeFormatter } from '../makeFormatter';
@@ -12,8 +12,7 @@ export const prepareNetworkSymbolFormatter = (config: FormatterConfig) =>
             const { bitcoinAmountUnit } = config;
             const { areAmountUnitsEnabled = true } = dataContext;
 
-            const { features: networkFeatures, testnet: isTestnet } =
-                networksCompatibility.find(network => network.symbol === symbol) ?? {};
+            const { features: networkFeatures, testnet: isTestnet } = getNetwork(symbol);
             const areAmountUnitsSupported = !!networkFeatures?.includes('amount-unit');
             let formattedSymbol = symbol.toUpperCase();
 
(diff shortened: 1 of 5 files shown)


suite-common/formatters/src/formatters/prepareNetworkSymbolFormatter.ts
```diff
@@ -12,8 +12,8 @@ export const prepareNetworkSymbolFormatter = (config: FormatterConfig) =>
             const { bitcoinAmountUnit } = config;
             const { areAmountUnitsEnabled = true } = dataContext;
 
-            const { features: networkFeatures, testnet: isTestnet } =
-                networksCompatibility.find(network => network.symbol === symbol) ?? {};
+            // symbol: NetworkSymbol, so TS says getNetwork always returns Network, but fallback just in case (we also have unit tests for this)
```
Lemonexe: the question is, if we need [those unit tests](https://github.com/trezor/trezor-suite/blob/74d30844a27d67274395ad23e5102c4c3fefe1c1/suite-common/formatters/src/formatters/tests/prepareCryptoAmountFormatter.test.ts#L78-L103).
But I'd rather be conservative, maybe they are there for a reason _(e.g. someone somewhere overrides TS? )_

peter-sanderson: In general I am against tests that tests situation where TS would raise error. 

I think they are duplicate and if there is some place where we for example parse JSON and types can be wrong we shall have type-safe validation (yup, ...)

peter-sanderson (approved):

suite-common/formatters/src/formatters/tests/prepareCryptoAmountFormatter.test.ts
```diff
@@ -74,32 +73,6 @@ describe('CryptoAmountFormatter', () => {
                 }),
             ).toBe('0.00014898… ETH');
         });
-
```
Lemonexe, line 77: useless tests, they go against TS.

`CryptoAmountFormatter.format` does not let you call it with `symbol` other than `NetworkSymbol`, so unless someone overrides TS with `as` or `any`, this should never happen. And if someone does override TS, let's find out? :smiling_imp:  We don't want that in our codebase when we have utils like `isNetworkSymbol`.

matejkriz (approved):
Works fine on mobile 👌

Lemonexe:
\rebase

PR #14374, opened 2024-09-17
chore(suite): refactor usages of deprecated getNetworkCompatible

Lemonexe:
Refactoring deprecated [getNetworkCompatible](https://github.com/trezor/trezor-suite/blob/722f5c99d35a53488f00b2ff46c7b500d813b31f/suite-common/wallet-utils/src/accountUtils.ts#L490-L494) to either [getNetwork, or getNetworkOptional](https://github.com/trezor/trezor-suite/blob/722f5c99d35a53488f00b2ff46c7b500d813b31f/suite-common/wallet-config/src/utils.ts#L73-L76)

## Description

`getNetworkCompatible`, which returns deprecated `NetworkCompatible`, is swapped for either:
- `getNetwork` if symbol is guaranteed to be `NetworkSymbol` type, then return `Network` and remove need for nullable handling :slightly_smiling_face: 
- `getNetworkOptional` if we can't be sure, then return `Network | undefined` _(as before)_

:information_source: Most occurrences were refactored - only the simple cases, where it doesn't change anything (the subsequent code consumes only those properties of `Network` which are common with `NetworkCompatible`).
Only several usages are waiting for a future refactoring.

## Related Issue

Part of #13839

## Screenshots:

It's quite straightforward refactoring, but still I selected a few easy-to-test places:

`AccountNavigation` still has correct tabs as per network's features 
AccountNavigation

`PortfolioCardException` reproducible with Tor + localhost REGTEST :upside_down_face: 
PortfolioCardException regtest + tor

`NoTransaction` not reproducible, had to comment code to get it, but the blockchain link still works:  
NoTransaction

### DIFF
--- a/suite-common/formatters/src/formatters/NetworkNameFormatter.ts
+++ b/suite-common/formatters/src/formatters/NetworkNameFormatter.ts
@@ -1,9 +1,8 @@
-import { NetworkSymbol } from '@suite-common/wallet-config';
-import { getNetworkCompatible } from '@suite-common/wallet-utils';
+import { getNetwork, NetworkSymbol } from '@suite-common/wallet-config';
 
 import { makeFormatter } from '../makeFormatter';
 
 export const NetworkNameFormatter = makeFormatter<NetworkSymbol, string>(
-    value => getNetworkCompatible(value)?.name || value,
+    value => getNetwork(value).name,
     'NetworkNameFormatter',
 );
--- a/packages/suite/src/actions/wallet/cardanoStakingActions.ts
+++ b/packages/suite/src/actions/wallet/cardanoStakingActions.ts
@@ -5,13 +5,10 @@ import { PendingStakeTx, PoolsResponse, CardanoNetwork } from 'src/types/wallet/
 import { Account, WalletAccountTransaction } from 'src/types/wallet';
 import { Dispatch, GetState } from 'src/types/suite';
 import { getUnixTime } from 'date-fns';
-import {
-    isPending,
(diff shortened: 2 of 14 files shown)


suite-common/formatters/src/formatters/NetworkNameFormatter.ts
```diff
@@ -1,9 +1,8 @@
-import { NetworkSymbol } from '@suite-common/wallet-config';
-import { getNetworkCompatible } from '@suite-common/wallet-utils';
+import { getNetwork, NetworkSymbol } from '@suite-common/wallet-config';
 
 import { makeFormatter } from '../makeFormatter';
 
 export const NetworkNameFormatter = makeFormatter<NetworkSymbol, string>(
-    value => getNetworkCompatible(value)?.name || value,
+    value => getNetwork(value).name,
```
Lemonexe, line 6: only used in native [here](https://github.com/trezor/trezor-suite/blob/5b29207bb5024babd48fd5ed504c3ca3b6ee51d7/suite-native/accounts/src/components/AccountListItem.tsx#L71) & [here](https://github.com/trezor/trezor-suite/blob/89d864e8c7aff4a5b6b46b49c4cbc6d12d3ca07d/suite-native/assets/src/components/AssetItem.tsx#L116), there it is called with `NetworkSymbol` type, so it's guaranteed to get a `Network`

komret (approved):

matejkriz (approved):
mobile part looks good 👌

PR #14377, opened 2024-09-17
feat(connect): return device info with method response

martykan:
## Description

Return additional info about device along with response payload. 
This is especially useful for core-in-popup or mobile, due to the lack of device events.

Eg.

```
{
  "id": 2,
  "success": true,
  "payload": {
    ...
  },
  "device": {
    "path": "E97D0BCE6A73CC77DDBA324D",
    "state": {
      "sessionId": "...",
      "staticSessionId": "abc@xyz:0"
    },
    "instance": 0
  }
}
```

### DIFF
--- a/packages/connect/src/events/call.ts
+++ b/packages/connect/src/events/call.ts
@@ -1,7 +1,8 @@
 import { serializeError } from '../constants/errors';
 import type { IFRAME } from './iframe';
 import type { TrezorConnect } from '../types/api';
-import type { CommonParams } from '../types/params';
+import type { CommonParams, DeviceIdentity } from '../types/params';
+import { Device } from '../device/Device';
 
 // conditionally unwrap TrezorConnect api method Success<T> response
 type UnwrappedResponse<T> =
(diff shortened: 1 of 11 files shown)


packages/connect/src/events/call.ts
```diff
@@ -58,17 +59,24 @@ export interface MethodResponseMessage {
...
     event: RESPONSE_EVENT,
     type: RESPONSE_EVENT,
     id,
     success,
 
     payload: success ? payload : serializeError(payload),
+    device: {
+        path: device?.originalDescriptor.path,
+        state: device?.getState(),
+        instance: device?.instance,
```
mroz22: maybe device should be nullable as the whole object if all of its properties are undefined 🤔

mroz22 (approved):
imho ok. 

@Hannsek, this deserves a new blog record. 

Could [connect-docs-blog-guide](https://connect.trezor.io/9/guides/) be something Ondra Zik would be willing to cooperate on with us? cc @MiroslavProchazka

PR #14383, opened 2024-09-17
fix(connect): keepSession with changing useCardanoDerivation

### LINKED ISSUE #14369: Cardano related TrezorConnect calls might fail even with useCardanoDerivation: true
It seems, that if `TrezorConnect` call had `keepSession: true` followed by enabling Cardano and asking for cardano derivations, it fails. 

According to our discussion with @mroz22 this might probably happen, because the Device has connect already initialized and not disposed (`this.commands?.disposed === false`), meaning, that [this condition](https://github.com/trezor/trezor-suite/blob/b1c111c8f21bda5221367cbbf3a204663b3804b2/packages/connect/src/device/Device.ts#L333) does not pass and the call is run without change in `useCardanoDerivation` even if it was passed in the call options. It probably also works the same the other way around, when cardano is disabled. Even patching in `connectInitThunk` with `cardanoConnectPatch` does not fix it in this case.

It seems that, workaround to this is to send another trezor connect request with `keepSession: false` which probably sets `this.commands?.disposed` to `true` and correctly applies `useCardanoDerivation`. 

We are using this workaround in #14345

martykan:
## Description

Make changing `useCardanoDerivation` work with `keepSession`.
Previously the issue was that initialize was not called again with the existing session. 
Now we keep track of the current `useCardanoDerivation` setting and we call initialize again if it's changed.

## Related Issue

Resolve #14369

### DIFF
--- a/packages/connect/e2e/tests/device/keepSession.test.ts
+++ b/packages/connect/e2e/tests/device/keepSession.test.ts
@@ -1,0 +1,42 @@
+import TrezorConnect from '../../../src';
+
+import { getController, setup, conditionalTest, initTrezorConnect } from '../../common.setup';
+
+const controller = getController();
+
+describe('keepSession common param', () => {
+    beforeAll(async () => {
+        await TrezorConnect.dispose();
+        await setup(controller, {
+            mnemonic: 'mnemonic_all',
+        });
+        await initTrezorConnect(controller);
+    });
+
+    afterAll(async () => {
+        controller.dispose();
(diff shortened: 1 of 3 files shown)


mroz22:
heads up @vytick, could you please check if it solves your troubles? are you able to get rid of that workaround?

vytick:
@mroz22 @martykan It works now without workaround👌

mroz22 (approved):

PR #14388, opened 2024-09-17
Fix/tokens table

tomasklim:
## Description

- nfts were included in tokens, filter them out
- there were errors in console when tokens had same name -> key should be contract, not symbol
- asset logo component did not use `shouldTryToFetch`

### DIFF
--- a/packages/components/src/components/AssetLogo/AssetLogo.tsx
+++ b/packages/components/src/components/AssetLogo/AssetLogo.tsx
@@ -58,7 +58,7 @@ export const AssetLogo = ({
     ...rest
 }: AssetLogoProps) => {
     const [isLoading, setIsLoading] = useState(true);
-    const [isPlaceholder, setIsPlaceholder] = useState(false);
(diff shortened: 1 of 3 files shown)


seibei-iguchi (approved):

packages/components/src/components/AssetLogo/AssetLogo.tsx
```diff
@@ -71,10 +71,6 @@ export const AssetLogo = ({
         setIsPlaceholder(true);
     };
 
-    useEffect(() => {
-        setIsPlaceholder(!shouldTryToFetch);
-    }, [shouldTryToFetch]);
```
jvaclavik: Is it correct? It's not going to update placeholder when you programatically update `shouldTryToFetch `

tomasklim: reverted

tomasklim:
Merging, even though tests are red. Those changes could not break them

PR #14393, opened 2024-09-17
chore(address-validator): update base-x 3.0.7 -> 5.0.0

mroz22:
## Related Issue

Resolve


martykan (approved):
LGTM

Unfortunately we still have 3 different versions of this through transitive dependencies, but that's another problem...

PR #14395, opened 2024-09-17
feat(suite): disable tx bumpfee except lowestnonc

### LINKED ISSUE #13529: Disable “Bump Fee” for txs with higher nonce than the lowest pending
Users are currently experiencing difficulty determining which transaction to bump to ensure it gets confirmed. The current system can lead to confusion, as bumping the oldest transaction with a small fee may still not result in confirmation due the a fact that oldest tx (time) may not have lowest nonce (was already bumped once)

Solution to address this issue:

1.	Disable “Bump Fee” for txs with higher nonce: Only the bump button for the transaction with the lowest nonce should remain enabled. This will guide users to prioritize the oldest transaction first.
2.	Add Tooltips to Disabled Buttons: Implement tooltips on disabled buttons indicating that the user must first bump the fee for the transaction with the lowest nonce. This tooltip will provide clarity on the required action.

Possible downsides:
1. In case user has all txs underprices, he has to wait for the oldest one to get confirmed to bump the rest. If that is problematic, we can introduce modal or add nonce to pending tx 

Screenshot 2024-07-28 at 20 53 13

Please take in mind also this change which can make it even more confusing https://github.com/trezor/trezor-suite/pull/12975

enjojoy:
## Description

Disable all the bump fees buttons except the transaction with the lowest nonce (the oldest transaction). Tooltip is displayed only on the disabled buttons.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13529

## Screenshots:

image

### DIFF
--- a/packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
+++ b/packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
@@ -2,7 +2,7 @@ import { memo, useMemo, useState } from 'react';
 import styled, { css } from 'styled-components';
 import { AnimatePresence } from 'framer-motion';
 import { selectIsPhishingTransaction } from '@suite-common/wallet-core';
-import { variables, Button, Card } from '@trezor/components';
+import { variables, Button, Card, Link } from '@trezor/components';
 import { Translation } from 'src/components/suite';
 import { useDispatch, useSelector } from 'src/hooks/suite';
 import { openModal } from 'src/actions/suite/modalActions';
@@ -35,6 +35,8 @@ import { BlurWrapper } from './TransactionItemBlurWrapper';
 import { selectSelectedAccount } from 'src/reducers/wallet/selectedAccountReducer';
 import { getInstantStakeType } from 'src/utils/suite/stake';
 import { isStakeTypeTx } from '@suite-common/suite-utils';
+import { Tooltip } from '@trezor/components';
+import { HELP_CENTER_REPLACE_BY_FEE } from '@trezor/urls';
 
 // eslint-disable-next-line local-rules/no-override-ds-component
 const Wrapper = styled(Card)<{
@@ -88,6 +90,7 @@ interface TransactionItemProps {
     network: Network;
     accountType: AccountType;
     className?: string;
+    disableBumpFee?: boolean;
     index: number;
 }
 
@@ -101,6 +104,7 @@ export const TransactionItem = memo(
         network,
         accountType,
         className,
+        disableBumpFee,
         index,
     }: TransactionItemProps) => {
         const [limit, setLimit] = useState(0);
@@ -182,6 +186,42 @@ export const TransactionItem = memo(
         const dataTestBase = `@transaction-item/${index}${
             transaction.deadline ? '/prepending' : ''
         }`;
+
+        const BumpFeeButton = ({ isDisabled }: { isDisabled: boolean }) => (
+            <Button
+                variant="tertiary"
+                onClick={() => openTxDetailsModal(true)}
+                isDisabled={isDisabled}
+            >
+                <Translation id="TR_BUMP_FEE" />
+            </Button>
+        );
+
+        const DisabledBumpFeeButtonWithTooltip = () => (
+            <Tooltip
+                content={
+                    <div>
+                        <Translation
+                            id="TR_BUMP_FEE_DISABLED_TOOLTIP"
+                            values={{
+                                a: chunks => (
+                                    <Link
+                                        href={HELP_CENTER_REPLACE_BY_FEE}
+                                        variant="nostyle"
+                                        icon="externalLink"
+                                        type="hint"
+                                    >
+                                        {chunks}
+                                    </Link>
+                                ),
+                            }}
+                        />
+                    </div>
+                }
+            >
+                <BumpFeeButton isDisabled={true} />
+            </Tooltip>
+        );
 
         // we are using slightly different layout for 1 targets txs to better match the design
         // the only difference is that crypto amount is in the same row as tx heading/description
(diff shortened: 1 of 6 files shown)


enjojoy:
Learn more link - https://trezor.io/learn/a/replace-by-fee-rbf-ethereum

MiroslavProchazka:
What about using "oldest pending transaction", instead of first? Does that make sense?

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
```diff
@@ -418,12 +422,46 @@ export const TransactionItem = memo(
...
+                                            }
+                                        >
+                                            <Button
+                                                variant="tertiary"
+                                                onClick={() => openTxDetailsModal(true)}
+                                                isDisabled={true}
+                                            >
+                                                <Translation id="TR_BUMP_FEE" />
+                                            </Button>
+                                        </Tooltip>
```
Lemonexe: I'd extract the tooltipped button to a new component (just here in this file), as the indentation is IMHO getting quite wild (gotta love prettier :smile:  ).

:thought_balloon:  Since the `Button` inside the `Tooltip` is almost the same, only changes are 1) being wrapped by `Tooltip` and 2) `isDisabled`, maybe there could be a wrapper component, which would either render
`<Tooltip ...>{children}</Tooltip>` or just `children`. But it's overkill to do it just to prevent duplication of four lines, so I'll leave that at your discretion :slightly_smiling_face:

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -1034,3 +1034,23 @@ export const groupTokensTransactionsByContractAddress = (
...
     return groupedTokensTxs;
 };
+
+export const getTransactionWithLowestNonce = (
+    transactionGroups: GroupedTransactionsByDate,
+): WalletAccountTransaction | null => {
+    let lowestNonceTransaction: WalletAccountTransaction | null = null;
+
+    Object.keys(transactionGroups).forEach(dateKey => {
+        transactionGroups[dateKey].forEach(transaction => {
```
Lemonexe: This code is fine, but it could be simplified :slightly_smiling_face: 

1. You can grab `Object.values` instead of `keys`
2. You can flatten the array, and avoid the nested `forEach`

```suggestion
    Object.values(transactionGroups)
        .flat()
        .forEach(transaction => {
```

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -1034,3 +1034,23 @@ export const groupTokensTransactionsByContractAddress = (
...
+                lowestNonceTransaction === null ||
+                currentNonce < Number(lowestNonceTransaction.ethereumSpecific?.nonce)
+            ) {
+                lowestNonceTransaction = transaction;
+            }
+        });
+    });
+
+    return lowestNonceTransaction;
+};
```
Lemonexe: Actually now that I played around with it, you could "simplify" it further using `.reduce`, though it's questionable if it makes it better readable :smile:  
This code should do the same. And that being said, **this function could do with tests** (you could use values from the `groupTransactionsByDate` test)

```
export const getTransactionWithLowestNonce = (
    transactionGroups: GroupedTransactionsByDate,
): WalletAccountTransaction | null =>
    Object.values(transactionGroups)
        .flat()
        .reduce<WalletAccountTransaction | null>((lowestNonceTransaction, transaction) => {
            const currentNonce = Number(transaction.ethereumSpecific?.nonce);
            if (
                lowestNonceTransaction === null ||
                currentNonce < Number(lowestNonceTransaction.ethereumSpecific?.nonce)
            ) {
                lowestNonceTransaction = transaction;
            }

            return lowestNonceTransaction;
        }, null);
```

Lemonexe (commented):
Codewise looks good, just some comments to make it neat.
And please, create the unit test if possible.

I didn't yet get to test it, then I'll finish the CR..

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -1034,3 +1034,23 @@ export const groupTokensTransactionsByContractAddress = (
...
 };
+
+export const getTransactionWithLowestNonce = (
+    transactionGroups: GroupedTransactionsByDate,
+): WalletAccountTransaction | null => {
+    let lowestNonceTransaction: WalletAccountTransaction | null = null;
+
+    Object.keys(transactionGroups).forEach(dateKey => {
+        transactionGroups[dateKey].forEach(transaction => {
+            const currentNonce = Number(transaction.ethereumSpecific?.nonce);
```
Lemonexe, line 1045: :warning:  I just realized a problem here - for non-ethereum, this function returns the first transaction. I think it should return `null`, because it's an ethereum-only feature.
So I would add before this line: `if(!transaction.ethereumSpecific) return`

Or you could do it in `TransactionGroupedList` - don't call this function if `network.networkType !== 'ethereum'`

Lemonexe (changes requested):
I tested it and it works correctly for ethereum-type networks :heavy_check_mark:
But it also disables button & displays tooltip for **non-ethereum** network txs (bitcoin testnet) :x:
I believe this is undesired, right?

enjojoy:
> I tested it and it works correctly for ethereum-type networks ✔️ But it also disables button & displays tooltip for **non-ethereum** network txs (bitcoin testnet) ❌ I believe this is undesired, right?

Yes, thank you, will fix that

tomasklim:
> What about using "oldest pending transaction", instead of first? Does that make sense?

Oldest by nonce, not by time

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
enjojoy: I just took it out in a variable so the indentation is not as crazy and It's better readable. If we'd have more occasions of buttons with tooltips, i think it'd be worth it creating a component for that.

suite-common/wallet-utils/src/transactionUtils.ts
enjojoy: Thank you, that's big brain 🧠 Maybe it's a little less readable than using for each, but I like it

suite-common/wallet-utils/src/transactionUtils.ts
enjojoy: I would do both for safety

Lemonexe (approved):
Thanks, it's working as expected :heavy_check_mark: 
Codewise looking great :rocket:

~`getTransactionWithLowestNonce` could do with unit tests~

enjojoy:
> Thanks, it's working as expected ✔️ Codewise looking great 🚀
> 
> `getTransactionWithLowestNonce` could do with unit tests, but oh well, I'm guilty of untested code too 😅 Can be done later 🤷

I pushed the tests 😼

Lemonexe (approved):
Good job adding the unit tests, now the code is a bit safer :+1:

PR #14398, opened 2024-09-17
Chore: finish networks refactoring

### LINKED ISSUE #13839: Refactor and type network config

Lemonexe:
Finish refactoring of old `networksCompatible` to `networks` :tada:

## Description

- 342638d3792077b5fb5031a195442433ad320c9e the heavy lifting :muscle: These changes had to be done together, as they are tightly interlinked:
  - refactor `selectedAccount` state
  - refactor discovery (a lot of logic changed!) 
  - all the easy stuff related to either discovery || selected account
- 3332ef75f45f96c05155d19e033170e3c0acd4c5 nuke legacy code :broom: :bucket: :radioactive:
- d0a25d92d563e08cf2c462d27c8705bffcd9851c incidental fix of pre-existing discovery bug :bug: with `tada` :tada: , discovery would seemingly never finish

## Related Issue

Resolve #13839 , finally :relaxed: :tada:

## Screenshots:

I tested in app that the discovery works as before: all accounts are discovered & unfinished/in progress is indicated correctly. But to be **really** sure, I console logged [`loaded` and `total` at `calculateProgress#L150`](https://github.com/trezor/trezor-suite/blob/d0a25d92d563e08cf2c462d27c8705bffcd9851c/suite-common/wallet-core/src/discovery/discoveryThunks.ts#L150)
- I activated every coin except `tada` on T2T1 w/ academic seed, and this is how the discovery progressed on this branch vs. develop: [discovery.test.txt](https://github.com/user-attachments/files/17060772/discovery.test.txt)
 (it's the same, both converged on `65 / 65` :heavy_check_mark: )
- I verified the same with another wallet with even more accounts, both converged on `92 / 92`
- before :tada: :bug: was solved, it'd converge on for example `16 / 36`, indicating unfinished discovery forever:
tada bug

### DIFF
--- a/suite-common/test-utils/src/mocks.ts
+++ b/suite-common/test-utils/src/mocks.ts
@@ -22,7 +22,7 @@ import {
     WalletAccountTransaction,
     BlockchainNetworks,
 } from '@suite-common/wallet-types';
-import { networksCompatibility } from '@suite-common/wallet-config';
+import { networksCollection } from '@suite-common/wallet-config';
 
 // in-memory implementation of indexedDB
 import 'fake-indexeddb/auto';
@@ -558,8 +558,7 @@ const intlMock = {
     formatMessage: (s: any) => s.defaultMessage,
 };
 
-const mockedBlockchainNetworks = networksCompatibility.reduce((result, network) => {
-    if (network.accountType) return result;
+const mockedBlockchainNetworks = networksCollection.reduce((result, network) => {
     result[network.symbol] = {
         connected: false,
         explorer: network.explorer,
--- a/suite-common/wallet-config/src/networksConfig.ts
+++ b/suite-common/wallet-config/src/networksConfig.ts
@@ -1,14 +1,6 @@
 import { DeviceModelInternal } from '@trezor/connect';
-import type { Without } from '@trezor/type-utils';
 
-import {
-    AccountType,
-    BackendType,
-    Explorer,
-    NetworkFeature,
-    Networks,
-    NetworkSymbol,
-} from './types';
+import { Networks } from './types';
 
 export const networks = {
     btc: {
@@ -641,44 +633,3 @@ export const networks = {
         coingeckoNativeId: undefined,
     },
 } as const satisfies Networks;
-
-type InferredNetworks = typeof networks;
-type InferredNetwork = InferredNetworks[NetworkSymbol];
-
-/**
- * @deprecated Old network object interface for backwards copatibility.
- */
-type NetworkCompatible = Without<InferredNetwork, 'accountTypes'> & {
-    symbol: NetworkSymbol;
-    accountType?: AccountType;
-    backendType?: BackendType;
-    testnet?: boolean;
-    isHidden?: boolean;
-    chainId?: number;
-    features?: NetworkFeature[];
-    support?: {
-        [key in DeviceModelInternal]: string;
-    };
-    isDebugOnlyNetwork?: boolean;
-    isDebugOnlyAccountType?: boolean;
-    explorer: Explorer;
-    coingeckoId?: string;
-};
-
-/**
- * @deprecated Please use `networks` instead.
- * Transforms the network object into the previously used format so we don't have to change
- * every occurence. We could gradually start to use the network object directly and in the end
- * this could be removed.
- */
-export const networksCompatibility: NetworkCompatible[] = Object.values(networks).flatMap(
-    ({ symbol, accountTypes, ...network }) => [
-        { symbol, ...network },
-        ...Object.entries(accountTypes).map(([accountType, networkOverride]) => ({
-            symbol,
-            accountType,
-            ...network,
-            ...networkOverride,
-        })),
-    ],
-);
--- a/suite-common/wallet-config/src/networksConfig.ts
+++ b/suite-common/wallet-config/src/networksConfig.ts
@@ -648,7 +648,7 @@ type InferredNetwork = InferredNetworks[NetworkSymbol];
 /**
  * @deprecated Old network object interface for backwards copatibility.
  */
-export type NetworkCompatible = Without<InferredNetwork, 'accountTypes'> & {
+type NetworkCompatible = Without<InferredNetwork, 'accountTypes'> & {
     symbol: NetworkSymbol;
     accountType?: AccountType;
     backendType?: BackendType;
--- a/suite-common/wallet-core/src/accounts/accountsThunks.ts
+++ b/suite-common/wallet-core/src/accounts/accountsThunks.ts
@@ -1,6 +1,6 @@
 import TrezorConnect, { AccountInfo, TokenInfo } from '@trezor/connect';
 import { Account, AccountKey } from '@suite-common/wallet-types';
-import { networksCompatibility } from '@suite-common/wallet-config';
+import { networksCollection } from '@suite-common/wallet-config';
 import {
     analyzeTransactions,
     findAccountDevice,
@@ -33,7 +33,7 @@ export const disableAccountsThunk = createThunk(
         const accounts = selectAccounts(getState());
         const enabledNetworks = selectEnabledNetworks(getState());
         // find disabled networks
-        const disabledNetworks = networksCompatibility
+        const disabledNetworks = networksCollection
             .filter(n => !enabledNetworks.includes(n.symbol) || n.isHidden)
             .map(n => n.symbol);
         // find accounts for disabled networks
--- a/suite-common/wallet-core/src/blockchain/blockchainThunks.ts
+++ b/suite-common/wallet-core/src/blockchain/blockchainThunks.ts
@@ -2,7 +2,7 @@ import { createThunk } from '@suite-common/redux-utils';
 import {
     getNetworkOptional,
     isNetworkSymbol,
-    networksCompatibility,
+    networksCollection,
     NetworkSymbol,
 } from '@suite-common/wallet-config';
 import {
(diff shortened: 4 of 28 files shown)


suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -74,44 +73,32 @@ export const filterUnavailableAccountTypes = (
...
+        ).length;
+
+        // numbers of accounts of givent types
+        const { numberOfNonCardano, numberOfCardano } = networksToCount.reduce(
+            (acc, network) => {
+                const { symbol } = network;
+
+                // increment the appropriate counter based on the network symbol
+                const incrementAccumulator = () => {
+                    if (symbol === 'ada' || symbol === 'tada') acc.numberOfCardano += 1;
```
Lemonexe, line 102: Note: at first I thought just including `'tada'` here would solve the :tada: :bug:,
but I also had to do the "1 or 2 thing" with `numberOfCardanoCoins` instead of always 1.

Otherwise, with `ada` & `tada` at the same time, discovery was wrongly indicated as finished, with debug showing `16 / 6` as loaded / total :scream:
I would have to activate some more coins to "manually" start discovery of `ada` or `tada`..

suite-common/test-utils/src/mocks.ts
```diff
@@ -560,8 +560,7 @@ const intlMock = {
     formatMessage: (s: any) => s.defaultMessage,
 };
 
-const mockedBlockchainNetworks = networksCompatibility.reduce((result, network) => {
-    if (network.accountType) return result;
```
Lemonexe, line 562: this line means that when using `networksCompatibility`, only `'normal'` accounts were processed, which makes the refactoring really easy :relaxed: Analogical diff in several other files below.

suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -641,44 +633,3 @@ export const networks = {
...
-    ({ symbol, accountTypes, ...network }) => [
-        { symbol, ...network },
-        ...Object.entries(accountTypes).map(([accountType, networkOverride]) => ({
-            symbol,
-            accountType,
-            ...network,
-            ...networkOverride,
-        })),
-    ],
-);
```
Lemonexe, line 684: farewell :wave:

suite-common/wallet-core/src/accounts/accountsThunks.ts
```diff
@@ -33,7 +33,7 @@ export const disableAccountsThunk = createThunk(
         const accounts = selectAccounts(getState());
         const enabledNetworks = selectEnabledNetworks(getState());
         // find disabled networks
-        const disabledNetworks = networksCompatibility
+        const disabledNetworks = networksCollection
             .filter(n => !enabledNetworks.includes(n.symbol) || n.isHidden)
             .map(n => n.symbol);
```
Lemonexe, line 38: :information_source: this is different! With `networksCompatibility`, the array of symbols is not unique, so e.g. `'btc'` is duplicated foreach account. Now with `networksCollection` it is unique.
BUt in this case it's ok, see L41, `disabledNetworks` is only used to test `.includes`

suite-common/wallet-core/src/blockchain/blockchainThunks.ts
```diff
@@ -61,7 +61,7 @@ export const preloadFeeInfoThunk = createThunk(
     `${BLOCKCHAIN_MODULE_PREFIX}/preloadFeeInfoThunk`,
     async (_, { dispatch }) => {
         // Fetch default fee levels
-        const networks = networksCompatibility.filter(n => !n.isHidden && !n.accountType);
```
Lemonexe, line 65: filters only `'normal'` accounts, so it's the same :heavy_check_mark:

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -74,44 +73,32 @@ export const filterUnavailableAccountTypes = (
...
     (discovery: Discovery) =>
     (_dispatch: any, getState: any): PartialDiscovery => {
-        const { numberOfNonCardano, numberOfCardano } = discovery.networks.reduce(
-            (acc, symbol) => {
-                if (symbol === 'ada') {
-                    acc.numberOfCardano += 1;
-                } else {
-                    acc.numberOfNonCardano += 1;
-                }
+        const device = selectDevice(getState());
```
Lemonexe, line 87: what's going on here? See [L741, L748](https://github.com/trezor/trezor-suite/blob/ced3f2a8a54c2636010c61e1f11be1bce03c692c/suite-common/wallet-core/src/discovery/discoveryThunks.ts#L741-L748) where it's first called.

**Previously**, `networksSymbols` from `filterUnavailableNetworksCompatible` was non-unique, [like in comment above](https://github.com/trezor/trezor-suite/pull/14398/files/d8e95808ac13344515846ca1d08255ad06b957a4..ced3f2a8a54c2636010c61e1f11be1bce03c692c#r1764541695), except here **it actually mattered**, and imho the previous implementation was kinda occult.
In `calculateProgress`, we need the _number_ of loaded accounts, so the number of duplicated `discovery.networks` symbols was counted.

**Now**, `discovery.networks` is unique, and it is responsibility of `calculateProgress` to iterate through accounts.

:bulb:  Note how `filterUnavailableNetworksCompatible` is split between `filterUnavailableNetworks` and `filterUnavailableAccountTypes`.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -279,58 +265,64 @@ export const getBundleThunk = createThunk(
...
-                    networkType: configNetwork.networkType,
+                    networkType,
                     derivationType: getDerivationType(accountType),
                     suppressBackupWarning: true,
                 });
             }
-        });
+        };
+
+        // foreach network, start discovery foreach normalized account (normal as well as all alternative accounts)
```
Lemonexe, line 320: "normalized account" is a very helpful helper introduced in [#14281 here](https://github.com/trezor/trezor-suite/commit/8f0270267d714f747542facdee446861ef20ed69), and used in `filterUnavailableAccountTypes`.

**TL;DR:** It's kinda like `networksCompatibility` but **only** for accounts, which actually does make sense!
It gets all accounts of network, but 1) synthesizes a `'normal'` account from the network properties, and 2) for the alternative accs it backfills missing properties from normal.

:arrow_right: never again u have to do stuff like `account.bip43Path ?? network.bip43Path`, because it's very easy to do a mistake there :hear_no_evil:
You just iterate over all accounts, and the correct properties are guaranteed to be there :ok_hand:

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -709,17 +701,23 @@ export const createDiscoveryThunk = createThunk(
...
+        const maxTotalAccounts = LIMIT * availableConfigAccounts;
 
         dispatch(
             createDiscovery({
                 deviceState,
                 authConfirm: !device.useEmptyPassphrase,
                 index: 0,
                 status: DiscoveryStatus.IDLE,
-                total: LIMIT * networksSymbols.length,
+                total: maxTotalAccounts,
```
Lemonexe, line 720: I suspect this is not used anywhere? But I'm reaaally not sure. I updated it anyway.

suite-native/module-send/src/screens/SendOutputsScreen.tsx
```diff
@@ -163,9 +163,9 @@ export const SendOutputsScreen = ({
         // Keyboard has to be dismissed here before navigating, so it's animation is not interfering with the animations on the FeesScreen.
         Keyboard.dismiss();
 
-        const network = getNetworkCompatible(account?.symbol);
+        if (!account) return;
 
-        if (!network) return;
+        const network = networks[account.symbol];
```
Lemonexe: when `!!account`, `network` is now guaranteed :heavy_check_mark:

suite-common/wallet-core/src/discovery/discoveryThunks.ts
tomasklim: Thank you! and for tests too!

suite-common/wallet-core/src/discovery/discoveryThunks.ts
tomasklim: Thank you! (Maybe I am just blind but it would be worth it to add tests for that?)

tomasklim (approved):
🎉🎊🥳🎈🎶🍾🍻🍹🍸🍷🥂🍺🎂🎇🎆💃🕺🎧🎤🎵🎺🎸🎻🥁🎡🎠🎢🚀🪅🍕🍔🌭🍿🍰🧁🍨🍩🍭🍬🍫🪩🎯🕹️🎲💥🔥🌟✨💫🌈👯‍♂️👯‍♀️👑👒🥳

suite-common/wallet-core/src/discovery/discoveryThunks.ts
Lemonexe: Good idea. I'd put it into followup work, so that we can merge this, and not accumulate conflicts - after all, there were no tests previously.
EDIT: added to the [appropriate issue](https://github.com/trezor/trezor-suite/issues/14444)

Issue #14399, opened 2024-09-17
Token definitions fetching improvements


Nodonisko:
We should come up with better solution for fetching token definitions. We now have hardcoded `eth`, `pol`, `bnb` & `sol` which is suboptimal. ~Also it will fetch that definitions during startup which is also not good because parsing it is kind of slow.~

- [x] Only definitions for enabled networks and networks of imported accounts should be fetched. (DONE in #15263)
- ~[ ] Don't fetch token definitions right on app start, not to make it slow.~

possible followups:
#15252 #15253

matejkriz:
Notes for QA: check that token definitions are not missing when needed (tokens have fiat rates, are visible in transaction list)

STew790:
**QA OK**
Since this https://github.com/trezor/trezor-suite/issues/15332 was fixed, I haven't found any other issue with tokens :tada: 

**Info**
24.11.1    35433e7d5213f924061bc36bb74edf017bd39db0

PR #14400, opened 2024-09-17
fix(mobile): token defintions for portfolio tracker

### LINKED ISSUE #7646: ERC20 n.4.1 - Display tokens on Dashboard and reflect them in assets values
As a user I would like to see the ERC-20 tokens reflected in total balance on dashboard in order not to be confused that some assets are included and some not. 

To Do design: 
Display badge "+ token(s) next to ETH
Put badge to all tokens with coin they are related to

Acceptance criteria:
ERC-20 tokens balances are reflected on dashboard Total balances
ERC-20 tokens balances are reflected in assets on dashboard
[EDIT NEW]: ERC-20 tokens balances are reflected in Wallet switcher
When user has some token and clicks on ETH on dashboard, he gets the bottom sheet with selection of assets.

[Figma](https://www.figma.com/file/xSWRW8Csq7zYJ3TiqwoIxB/Bugfixes?type=design&node-id=251-409&mode=design&t=oYqjjCX9bZBGJo5t-11)

Nodonisko:
## Description

It's just hotfix until we come up with proper solution. Follow up #14399

## Related Issue

Resolve #7646

### DIFF
--- a/suite-common/test-utils/src/extraDependenciesMock.ts
+++ b/suite-common/test-utils/src/extraDependenciesMock.ts
@@ -75,6 +75,10 @@ export const extraDependenciesMock: ExtraDependencies = {
         selectBitcoinAmountUnit: mockSelector('selectBitcoinAmountUnit', PROTO.AmountUnit.BITCOIN),
         selectAreSatsAmountUnit: mockSelector('selectAreSatsAmountUnit', false),
         selectEnabledNetworks: mockSelector('selectEnabledNetworks', ['btc', 'test']),
+        selectTokenDefinitionsEnabledNetworks: mockSelector(
+            'selectTokenDefinitonsEnabledNetworks',
+            ['eth'],
+        ),
         selectLocalCurrency: mockSelector('selectLocalCurrency', 'usd'),
         selectIsPendingTransportEvent: mockSelector('selectIsPendingTransportEvent', false),
         selectDebugSettings: mockSelector('selectDebugSettings', {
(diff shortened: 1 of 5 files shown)


suite-common/test-utils/src/extraDependenciesMock.ts
```diff
@@ -75,6 +75,10 @@ export const extraDependenciesMock: ExtraDependencies = {
         selectBitcoinAmountUnit: mockSelector('selectBitcoinAmountUnit', PROTO.AmountUnit.BITCOIN),
         selectAreSatsAmountUnit: mockSelector('selectAreSatsAmountUnit', false),
         selectEnabledNetworks: mockSelector('selectEnabledNetworks', ['btc', 'test']),
+        selectTokenDefinitionsEnabledNetworks: mockSelector(
+            'selectTokenDefinitonsEnabledNetworks',
+            ['eth'],
```
vytick, line 80: Maybe you can make selector, that fetches all accounts, take those imported, get their symbols and check with `isCoinWithTokens` and return all token enabled network symbols. 

This way even before we change how token definitions are fetched, we would have only one place where to set which definitions are enabled. We would just update `isCoinWithTokens` with polygon or whatever

matejkriz (approved):
Works fine as a workaround 👌 It means token definitions are fetched on mobile on every start even if user don't use ETH at all, so let's optimize later...

suite-common/test-utils/src/extraDependenciesMock.ts
```diff
@@ -75,6 +75,10 @@ export const extraDependenciesMock: ExtraDependencies = {
         selectBitcoinAmountUnit: mockSelector('selectBitcoinAmountUnit', PROTO.AmountUnit.BITCOIN),
         selectAreSatsAmountUnit: mockSelector('selectAreSatsAmountUnit', false),
         selectEnabledNetworks: mockSelector('selectEnabledNetworks', ['btc', 'test']),
+        selectTokenDefinitionsEnabledNetworks: mockSelector(
+            'selectTokenDefinitonsEnabledNetworks',
```
tomasklim, line 79: There is typo definitons -> definitions

PR #14402, opened 2024-09-17
fix(suite-native): hide coin enabling when disabled device connect ff

vytick:
Dont show Coin Enabling settings on ios (when device connect is disabled)


matejkriz (approved):

STew790:
**QA OK**
No coin enabling for iOS.
**Info:**
24.9.1 01653e6

PR #14404, opened 2024-09-18
fix(suite-native): Senf flow UI/UX review fixes vol. 1

### LINKED ISSUE #14356: Send coins - fixes of UI/UX review
Fix all the problems discovered during the [UI/UX review](https://www.figma.com/design/sqKFZAIvk3zERxiWgthK8K/Send?node-id=3241-9623&t=rZPWs70FeVAIyzjH-4) that are marked as critical for the first send release.

PeKne:
## Description
- fixes all the critical UI/UX issues mentioned in the Figma UI/UX review file (see issue attached)

## Issue 
Resolve #14356


suite-native/atoms/src/Hint.tsx
```diff
@@ -15,6 +15,9 @@ type HintProps = {
     children?: ReactNode;
 };
 
+const ICON_SIZE = 14;
+const SPACE_SIZE = 6;
```
vytick, line 19: is there a reason to have these numbers in form of variables, when they are used only one place?

suite-native/module-send/src/components/AddressOriginHelpButton.tsx
```diff
@@ -7,30 +7,48 @@ import { AddressReviewSheetSection } from './AddressReviewSheetSection';
...
+            }
+        ></AddressReviewHelpSheet>
     );
 };
+
+// title={
+//     <VStack>
+//         <Text variant="titleSmall">
+//             <Translation id="moduleSend.review.address.originBottomSheet.title" />
+//         </Text>
```
vytick: please, if there is no need to have this commented, remove it

suite-native/module-send/src/components/SendScreen.tsx
```diff
@@ -19,7 +19,6 @@ export const SendScreen = ({ children, footer, screenHeader, subheader }: SendSc
             screenHeader={screenHeader}
             subheader={subheader}
             footer={footer}
-            keyboardShouldPersistTaps="never"
```
vytick, line 22: this probably means, that keyboardShouldPersistTaps is never set anywhere, maybe to remove the prop?

vytick (commented):
I havent tested this yet, here are few comments before I get to it. Overall it looks nice :)

suite-native/alerts/package.json
```diff
@@ -14,6 +14,7 @@
         "@suite-common/icons-deprecated": "workspace:*",
```
vytick, line 14: It seems that send btn always visible in selecting amount to send is not resolved. Not sure if it should be included, just mentioning this.

suite-native/module-send/src/components/FiatAmountInput.tsx
```diff
@@ -43,9 +43,15 @@ export const FiatAmountInput = ({
         valueTransformer: fiatAmountTransformer,
     });
 
+    // Validation is assigned to the crypto field, so we need to check if that field has an error.
+    const { hasError } = useField({
+        name: cryptoFieldName,
+    });
+
     const handleChangeValue = (newValue: string) => {
```
vytick, line 51: it seems that it does not support use of decimal comma `,`


https://github.com/user-attachments/assets/b0b2d8c3-ff26-44bc-a103-25a71bef254e

suite-native/atoms/src/Hint.tsx
PeKne: I prefer to use descriptive consts for these numbers, so it is easier to review and reuse them. In this case, both of them are used only once and it is easy to read what the numbers mean, so it is not necessary. But I would prefer to keep it this way if you don't mind.

suite-native/module-send/src/components/AddressOriginHelpButton.tsx
PeKne: Completely obsolete. Removed in d6fb6a99c8.

suite-native/module-send/src/components/SendScreen.tsx
PeKne: removed d6fb6a99c8

suite-native/module-send/src/components/FiatAmountInput.tsx
PeKne: good catch, fixed in d6fb6a99c8

suite-native/alerts/package.json
PeKne: Sorry for the misleading PR description. This will be solved by a separate PR. [New issue here](https://github.com/trezor/trezor-suite/issues/14431)

suite-native/atoms/src/Hint.tsx
vytick: I understand the approach, but reading what number was used is more difficult with const than direct use. 

And in case it is used on one place only it (IMO) feels harder to understand the code. But if you prefer it this way, I dont have a strong opinion on this. Feel free to resolve

suite-native/module-send/src/components/FiatAmountInput.tsx
vytick: Works now nicely 💪

vytick (approved):
Looks nice, good job 👏 🚀

suite-native/atoms/src/Hint.tsx
PeKne: Just to clarify my point of view... I see the numbers inside of the JSX same as an inline style. IMO all these stylings and configurations should be at the beginning of the file.

PR #14412, opened 2024-09-18
fix(connect): fw releases value for custom device model

szymonlesisz:
## Description

Addition to https://github.com/trezor/trezor-suite/commit/0915d889312b572b2f2bfdbe70f9dcc847096e3a

Custom models doesn't have defined releases, this leads to unexpected runtime errors like [here](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/device/Device.ts#L569) or [here](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/api/checkFirmwareAuthenticity.ts#L35).

- added empty array fallback for `getReleases` fn
- fixed types of `releases` (could be undefined)


mroz22 (approved):

mroz22:
test unit result 
image

PR #14417, opened 2024-09-18
feat(connect): push tx permission

martykan:
## Description

This PR introduces a separate permission in Connect methods that are pushing (broadcasting) transactions to the network. 
This is especially useful with the upcoming deeplinks, since it could be a security risk to push transactions from untrusted sources.

### DIFF
--- a/packages/connect-popup/src/view/permissions.ts
+++ b/packages/connect-popup/src/view/permissions.ts
@@ -8,14 +8,12 @@ const getPermissionText = (permissionType: string, _deviceName: string) => {
     switch (permissionType) {
         case 'read':
             return 'Read public keys from Trezor device';
-        case 'read-meta':
-            return 'Read metadata from Trezor device';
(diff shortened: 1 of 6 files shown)


packages/connect-popup/src/view/permissions.ts
```diff
@@ -8,14 +8,12 @@ const getPermissionText = (permissionType: string, _deviceName: string) => {
...
             return 'Read public keys from Trezor device';
-        case 'read-meta':
-            return 'Read metadata from Trezor device';
         case 'write':
             return 'Prepare Trezor device for transaction and data signing';
-        case 'write-meta':
-            return 'Write metadata to Trezor device';
         case 'management':
             return 'Modify device settings';
+        case 'push_tx':
```
mroz22, line 15: maybe strongly typed? MethodPermission

mroz22 (approved):

PR #14439, opened 2024-09-19
chore(suite-native): capture console errors to sentry on mobile too

matejkriz:
## Description

Desktop Suite reports console.error to Sentry, but mobile not yet. But console.error is used also on some places in shared common code instead of throwing an error like for example parsing message system config, so we miss those errors.

Example test error: https://satoshilabs.sentry.io/issues/5873989872/?environment=debug&project=4504214699245568&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=0

image


package.json
```diff
@@ -107,7 +107,6 @@
         "electron": "30.3.1",
         "@types/node": "20.12.7",
         "@types/react": "18.2.55",
-        "@sentry/types": "7.92.0",
```
matejkriz, line 110: I just removed this resolution and typescript wasn't complaining, so I guess it's not needed anymore.

Nodonisko (approved):

PR #14440, opened 2024-09-19
transport branded types and fix release device

mroz22:
## Related Issue

Resolve


marekrjpolak (approved):
Branded types are good, the second commit probably as well.

PR #14447, opened 2024-09-19
chore(message-system): add banner for deprecated coins

MiroslavProchazka:
This PR  implements a message system banner that will inform users about the deprecation of Dash, Bitcoin Gold, Namecoin, Vertcoin, and DigiByte.

For the first time, this banner will also be included in the mobile app. I’ve set the minimum version as 24.9.1 @matejkriz, please confirm if that’s correct.

Some additional details:
	•	“priority”: 93
	•	“dismissible”: true
	•	“variant”: “warning”
	•	“category”: “banner”

Translations were proofread by ChatGPT o1, so they should be reliable. The “Learn more” link will direct users to KB article sitting on "https://trezor.io/learn/a/deprecated-coins", which is currently in preparation.

I've also lowered priority on POL banner as it was clashing between each other.

Tracked in Notion in [here](https://www.notion.so/satoshilabs/Suite-coins-deprecation-10cdc5260606807e92b2c3867ef9cf3d?pvs=4)

Tested on **`release-message-system-develop`** by QA, slack thread [here](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1727352279546489)

Suite 
image

Mobile
image

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
-    "timestamp": "2024-08-028T00:00:00+00:00",
-    "sequence": 62,
+    "timestamp": "2024-09-26T00:00:00+00:00",
+    "sequence": 63,
     "actions": [
         {
             "conditions": [
@@ -107,7 +107,7 @@
             ],
             "message": {
                 "id": "c8ca8958-7868-467f-9278-f072edc234ac",
-                "priority": 93,
+                "priority": 92,
                 "dismissible": true,
                 "variant": "info",
                 "category": "banner",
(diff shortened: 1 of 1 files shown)


tomasklim:
How about creating one message mentioning all deprecating coins? This means that if user has all coins enabled, he will see 5 almost same messages

MiroslavProchazka:
> How about creating one message mentioning all deprecating coins? This means that if user has all coins enabled, he will see 5 almost same messages

In that case we would need to display  all affected coins in the message, which will definitely prolong it. 

Also I'm not sure whether this would work (to have several coins in the settings, but I will try).

vytick:
@MiroslavProchazka fyi: there is no intention to deploy that message to production. It was testing message

tomasklim:
@MiroslavProchazka it should work
and you can use network symbols instead of names 

I would rather see `BTG` instead of `Bitcoin Gold`

vytick (approved):
lgtm :)

PR #14448, opened 2024-09-19
feat(mobile): fancy QR codes

Nodonisko:
## Description

This is just experiment and we may or may not merge it. I did not followed any desing etc. just played around little bit.

## Related Issue

Resolve 

## Screenshots:
Simulator Screenshot - iPhone 15 Pro - 2024-09-19 at 22 29 01
Simulator Screenshot - iPhone 15 Pro - 2024-09-19 at 22 29 21


matejkriz:
It looks cool, but relying on QR code error correction capatibilities so we don't want it.

PR #14455, opened 2024-09-20
unit test for filtering accounts

### LINKED ISSUE #13978: Create unit tests to test cover the complex condition for accounts filtering
Follow up https://github.com/trezor/trezor-suite/pull/13723

enjojoy:
## Description

Test a complex condition in account filtering

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13978

### DIFF
--- a/packages/suite/src/hooks/wallet/coinmarket/form/useCoinmarketVerifyAccount.tsx
+++ b/packages/suite/src/hooks/wallet/coinmarket/form/useCoinmarketVerifyAccount.tsx
@@ -1,4 +1,4 @@
-import { isDebugOnlyAccountType, Network, networksCollection } from '@suite-common/wallet-config';
+import { Network, networksCollection } from '@suite-common/wallet-config';
 import { selectDevice } from '@suite-common/wallet-core';
 import { useEffect, useMemo, useState } from 'react';
 import { useForm } from 'react-hook-form';
@@ -22,6 +22,7 @@ import {
 } from 'src/types/coinmarket/coinmarketVerify';
 import { useAccountAddressDictionary } from 'src/hooks/wallet/useAccounts';
 import { TrezorDevice } from '@suite-common/suite-types';
+import { filterReceiveAccounts } from '@suite-common/wallet-utils';
 
 const getSelectAccountOptions = (
     suiteReceiveAccounts: Account[] | undefined,
@@ -79,33 +80,13 @@ const getSuiteReceiveAccounts = ({
                 ((n.isDebugOnlyNetwork && isDebug) || !n.isDebugOnlyNetwork),
         );
 
-        const isSameDevice = (account: Account) => account.deviceState === device?.state;
-        const isSameNetwork = (account: Account) => account.symbol === receiveNetwork;
-        const isDebugAndIsAccountDebugOnly = (account: Account) =>
-            isDebugOnlyAccountType(account.accountType, account.symbol) && isDebug;
-        const isNotDebugOnlyAccount = (account: Account) =>
-            !isDebugOnlyAccountType(account.accountType, account.symbol);
-        // Check if the account is not empty
-        const isNotEmptyAccount = (account: Account) => !account.empty;
-        // Check if the account is marked as visible
-        const isVisibleAccount = (account: Account) => account.visible;
-        const isFirstNormalAccount = (account: Account) =>
-            account.accountType === 'normal' && account.index === 0;
-        const isCoinjoinAccount = (account: Account) => account.accountType === 'coinjoin';
-
-        if (receiveNetworks.length > 0) {
-            // Get accounts of the current symbol belonging to the current device.
-            return accounts.filter(
-                account =>
-                    isSameDevice(account) &&
-                    isSameNetwork(account) &&
-                    !isCoinjoinAccount(account) &&
-                    (isDebugAndIsAccountDebugOnly(account) || isNotDebugOnlyAccount(account)) &&
-                    (isNotEmptyAccount(account) ||
-                        isVisibleAccount(account) ||
-                        isFirstNormalAccount(account)),
-            );
-        }
+        return filterReceiveAccounts({
+            accounts,
+            deviceState: device?.state,
+            receiveNetwork,
+            isDebug,
+            receiveNetworks,
+        });
     }
 
     return undefined;
@@ -142,6 +123,7 @@ const useCoinmarketVerifyAccount = ({
             }),
         [accounts, currency, device, isDebug, receiveNetwork],
     );
+
     const selectAccountOptions = useMemo(
         () => getSelectAccountOptions(suiteReceiveAccounts, device),
         [device, suiteReceiveAccounts],
--- a/suite-common/wallet-utils/src/__tests__/filterReceiveAccounts.test.ts
+++ b/suite-common/wallet-utils/src/__tests__/filterReceiveAccounts.test.ts
@@ -1,0 +1,128 @@
+import { testMocks } from '@suite-common/test-utils';
+import { Network, networksCollection } from '@suite-common/wallet-config';
+import { Account } from '@suite-common/wallet-types';
+
+import { isDebugOnlyAccountType, filterReceiveAccounts } from '../filterReceiveAccounts';
+
+const { getSuiteDevice, getWalletAccount } = testMocks;
+
+const accountsList: Account[] = [
+    getWalletAccount({ symbol: 'eth', accountType: 'legacy' }),
+    getWalletAccount({ symbol: 'eth', accountType: 'normal' }),
+    getWalletAccount({ symbol: 'eth', accountType: 'ledger' }),
+    getWalletAccount({ symbol: 'btc', accountType: 'coinjoin' }),
+    getWalletAccount({ symbol: 'btc', accountType: 'taproot' }),
+    getWalletAccount({ symbol: 'btc', accountType: 'legacy' }),
+    getWalletAccount({ symbol: 'btc', accountType: 'segwit' }),
+    getWalletAccount({ symbol: 'btc', accountType: 'ledger' }),
+    getWalletAccount({ symbol: 'pol', accountType: 'legacy' }),
+    getWalletAccount({ symbol: 'pol', accountType: 'normal' }),
+    getWalletAccount({ symbol: 'pol', accountType: 'ledger' }),
+    getWalletAccount({ symbol: 'sol', accountType: 'normal', empty: true, visible: false }),
+    getWalletAccount({ symbol: 'sol', accountType: 'ledger' }),
+    getWalletAccount({
+        symbol: 'sol',
+        accountType: 'ledger',
+        empty: true,
+        visible: false,
+    }),
+];
+
+type RunFilterReceiveAccountsTestParams = {
+    isDebug?: boolean;
+    receiveNetwork?: string;
+    deviceState?: `${string}@${string}:${number}`;
+    accounts?: Account[];
+};
+
+const runFilterReceiveAccouns = ({
+    isDebug = true,
+    receiveNetwork = 'eth',
+    deviceState = '1stTestnetAddress@device_id:0',
+    accounts = accountsList,
+}: RunFilterReceiveAccountsTestParams) => {
+    const device = getSuiteDevice({
+        unavailableCapabilities: {
+            dash: 'no-support',
+        },
+        state: deviceState,
+    });
+    const unavailableCapabilities = device?.unavailableCapabilities ?? {};
+
+    const receiveNetworks = networksCollection.filter(
+        (n: Network) =>
+            n.symbol === receiveNetwork &&
+            !unavailableCapabilities[n.symbol] &&
+            ((n.isDebugOnlyNetwork && isDebug) || !n.isDebugOnlyNetwork),
+    );
+
+    return filterReceiveAccounts({
+        accounts,
+        deviceState: device.state,
+        receiveNetwork,
+        isDebug,
+        receiveNetworks,
+    });
+};
+
+describe('filter receive accounts', () => {
+    it('checks if account is debug only type', () => {
+        expect(isDebugOnlyAccountType('legacy', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('segwit', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('coinjoin', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('taproot', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('ledger', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('legacy', 'eth')).toBe(true);
+        expect(isDebugOnlyAccountType('ledger', 'eth')).toBe(true);
+        expect(isDebugOnlyAccountType('normal', 'regtest')).toBe(false);
+    });
+
+    it('returns no results when given an empty accounts array', () => {
+        expect(runFilterReceiveAccouns({ accounts: [] })).toEqual([]);
+    });
+
+    it('returns no results when given a non-existing network in acccounts list', () => {
+        expect(runFilterReceiveAccouns({ receiveNetwork: 'bnb' })).toEqual([]);
+    });
+
+    it('returns all accounts when debug mode is on', () => {
+        const filteredAccounts = [
+            getWalletAccount({ symbol: 'eth', accountType: 'legacy' }),
+            getWalletAccount({ symbol: 'eth', accountType: 'normal' }),
+            getWalletAccount({ symbol: 'eth', accountType: 'ledger' }),
+        ];
+        expect(runFilterReceiveAccouns({})).toEqual(filteredAccounts);
+    });
+
+    it('returns only non-debug accounts when debug mode is off', () => {
+        const filteredAccounts = [getWalletAccount({ symbol: 'eth', accountType: 'normal' })];
+
+        expect(runFilterReceiveAccouns({ isDebug: false })).toEqual(filteredAccounts);
+    });
+
+    it('returns no results when device is not the same', () => {
+        expect(runFilterReceiveAccouns({ deviceState: '2ndTestnetAddress@device_id:0' })).toEqual(
+            [],
+        );
+    });
+
+    it('excludes coinjoin accounts for BTC network (also tests isAnotherNetwork and isCoinjoinAccount methods)', () => {
+        const filteredAccounts = [
+            getWalletAccount({ symbol: 'btc', accountType: 'taproot' }),
+            getWalletAccount({ symbol: 'btc', accountType: 'legacy' }),
+            getWalletAccount({ symbol: 'btc', accountType: 'segwit' }),
+            getWalletAccount({ symbol: 'btc', accountType: 'ledger' }),
+        ];
+
+        expect(runFilterReceiveAccouns({ receiveNetwork: 'btc' })).toEqual(filteredAccounts);
+    });
+
+    it('returns account when when its either first normal account (no matter is empty or not visible) or it is not empty and visible', () => {
+        const filteredAccounts = [
+            getWalletAccount({ symbol: 'sol', accountType: 'normal', empty: true, visible: false }),
+            getWalletAccount({ symbol: 'sol', accountType: 'ledger' }),
+        ];
+
+        expect(runFilterReceiveAccouns({ receiveNetwork: 'sol' })).toEqual(filteredAccounts);
+    });
+});
--- a/suite-common/wallet-config/src/utils.ts
+++ b/suite-common/wallet-config/src/utils.ts
@@ -3,7 +3,6 @@ import {
     AccountType,
     Network,
     NetworkFeature,
-    Networks,
     NetworkSymbol,
     NormalizedNetworkAccount,
 } from './types';
@@ -44,21 +43,6 @@ export const normalizeNetworkAccounts = (network: Network): NormalizedNetworkAcc
 
 export const isBlockbookBasedNetwork = (symbol: NetworkSymbol) =>
     networks[symbol]?.customBackends.some(backend => backend === 'blockbook');
-
-export const isDebugOnlyAccountType = (
-    accountType: AccountType,
-    symbol?: NetworkSymbol,
-): boolean => {
-    if (!symbol) return false;
-
-    const network = (networks as Networks)?.[symbol];
-
-    if (!network) return false;
-
-    const accountTypeInfo = network.accountTypes[accountType];
-
-    return !!accountTypeInfo?.isDebugOnlyAccountType;
-};
 
 export const getNetworkType = (symbol: NetworkSymbol) => networks[symbol]?.networkType;
 
--- a/suite-common/wallet-utils/src/filterReceiveAccounts.ts
+++ b/suite-common/wallet-utils/src/filterReceiveAccounts.ts
@@ -1,0 +1,51 @@
+import { AccountType, getNetwork, Network, NetworkSymbol } from '@suite-common/wallet-config';
+import { Account } from '@suite-common/wallet-types';
+
+export const isDebugOnlyAccountType = (
+    accountType: AccountType,
+    symbol?: NetworkSymbol,
+): boolean => {
+    if (!symbol) return false;
+
+    const network = getNetwork(symbol);
+
+    const accountTypeInfo = network.accountTypes[accountType];
+
+    return !!accountTypeInfo?.isDebugOnlyAccountType;
+};
+
+type FilterReceiveAccountsProps = {
+    accounts: Account[];
+    deviceState: string | undefined;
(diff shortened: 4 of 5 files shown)


suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -1153,3 +1155,59 @@ export const isTokenMatchesSearch = (token: TokenInfo, search: string) => {
...
+    const isDebugAndIsAccountDebugOnly = (account: Account) =>
+        isDebugOnlyAccountType(account.accountType, account.symbol) && isDebug;
+    const isNotDebugOnlyAccount = (account: Account) =>
+        !isDebugOnlyAccountType(account.accountType, account.symbol);
+    const isNotEmptyAccount = (account: Account) => !account.empty;
+    const isVisibleAccount = (account: Account) => account.visible;
+    const isFirstNormalAccount = (account: Account) =>
+        account.accountType === 'normal' && account.index === 0;
+
+    if (receiveNetworks.length > 0) {
```
tomasklim: I think you do not need this condition. Filter of `[]` will return `[]`

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -1153,3 +1155,59 @@ export const isTokenMatchesSearch = (token: TokenInfo, search: string) => {
...
+
+type FilterRecieveAccountsProps = {
+    accounts: Account[];
+    deviceState: string | undefined;
+    receiveNetwork?: string;
+    isDebug: boolean;
+    receiveNetworks: Network[];
+};
+
+export const filterRecieveAccounts = ({
```
tomasklim:
```suggestion
export const filterReceiveAccounts = ({
```

tomasklim (changes requested):

suite-common/wallet-utils/src/accountUtils.ts
enjojoy: This check was there initially for another reason I think, but not sure why, it was there before I refactored the checks

image

enjojoy: I think it's a check wether the network is supported by suite at all and if not, it would be impossible to buy. So I would let it be there to not break anything that was written before me

packages/suite/src/hooks/wallet/coinmarket/form/useCoinmarketVerifyAccount.tsx
```diff
@@ -79,31 +80,15 @@ const getSuiteReceiveAccounts = ({
...
-        }
+        const filteredAccounts = filterReceiveAccounts({
+            accounts,
+            deviceState: device?.state,
+            receiveNetwork,
+            isDebug,
+            receiveNetworks,
+        });
+
+        return filteredAccounts;
```
peter-sanderson: nit: `return filterReceiveAccounts({...` direclty

packages/suite/src/hooks/wallet/coinmarket/form/useCoinmarketVerifyAccount.tsx
```diff
@@ -140,6 +125,8 @@ const useCoinmarketVerifyAccount = ({
             }),
         [accounts, currency, device, isDebug, receiveNetwork],
     );
+
+    console.log('SUITE RECEIVE ACCOUNTS: ', suiteReceiveAccounts);
```
peter-sanderson: probably forgotten log

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,119 @@ describe('account utils', () => {
         account.addresses = enhanceAddresses(accountInfo, account);
         expect(account.addresses.change).toEqual([]);
     });
+
+    it('isDebugOnlyAccountType', () => {
```
peter-sanderson: I think that the test description shall be informative and shall explain what is expected and why:

something like => `it('asserts that only ledger and legacy eth accounts are considered as debug, as we don't ..... <real reason why here>' ...`

I even think we have something like this in some code style guide.

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,119 @@ describe('account utils', () => {
...
+                accounts,
+                deviceState: device.state,
+                receiveNetwork,
+                isDebug,
+                receiveNetworks,
+            }),
+        ).toEqual(expectedAccounts);
+    };
+
+    it('filterReceiveAccounts: isDebug', () => {
```
peter-sanderson: `it('filters out the debug account', ...`

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: sorry, re-requested review the same moment you posted this comment. will do

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
peter-sanderson: ... and similarly for other cases

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,119 @@ describe('account utils', () => {
...
+        const unavailableCapabilities = device?.unavailableCapabilities ?? {};
+
+        const receiveNetworks = networksCollection.filter(
+            (n: Network) =>
+                n.symbol === receiveNetwork &&
+                !unavailableCapabilities[n.symbol] &&
+                ((n.isDebugOnlyNetwork && isDebug) || !n.isDebugOnlyNetwork),
+        );
+
+        const accounts = [
```
peter-sanderson: I think this can be extracted as constant outside of the `runFilterReceiveAccountsTest` function

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,119 @@ describe('account utils', () => {
...
+    });
+
+    type RunFilterReceiveAccountsTestParams = {
+        isDebug?: boolean;
+        expectedAccounts?: Account[];
+        receiveNetwork?: string;
+        deviceState?: `${string}@${string}:${number}`;
+    };
+
+    const runFilterReceiveAccountsTest = ({
```
peter-sanderson: I would separate the "execution" of the test and assertion. I prefer to see assertions in the tests and not hidden somewhere.

This is kinda subjective, but I think this function only passes args down the stack and dont add anything regarding to assert

peter-sanderson: So I would just return `filterReceiveAccounts` the filtered account from this, and asserted them separately in every `it` test-case

peter-sanderson: So the sole responsibility of the `runFilterReceiveAccountsTest` would be to provide dependecies (`device` and `receiveNetworks`)

and responsibility for assertion is separated in the test-case

peter-sanderson: The reason why I like it is that then in the `it` block it is clearly visible what test does and it follows the 

`given` -> `when` -> `then` pattern (in this case: `input data` -> `run tested code`  -> `assert result`)

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -1153,3 +1155,59 @@ export const isTokenMatchesSearch = (token: TokenInfo, search: string) => {
...
+
+export const isDebugOnlyAccountType = (
+    accountType: AccountType,
+    symbol?: NetworkSymbol,
+): boolean => {
+    if (!symbol) return false;
+
+    const network = (networks as Networks)?.[symbol];
+
+    if (!network) return false;
```
Lemonexe: :bulb: This can be simplified without `as`, and without the `if`.
Btw I added `wallet-config/getNetwork` only recently, before it had to be like this, but now it can be nicer :blush: 
```suggestion
    const network = getNetwork(symbol);
```

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -1153,3 +1155,59 @@ export const isTokenMatchesSearch = (token: TokenInfo, search: string) => {
...
+    receiveNetwork,
+    isDebug,
+    receiveNetworks,
+}: FilterRecieveAccountsProps): Account[] => {
+    const isSameDevice = (account: Account) => account.deviceState === deviceState;
+    const isSameNetwork = (account: Account) => account.symbol === receiveNetwork;
+    const isDebugAndIsAccountDebugOnly = (account: Account) =>
+        isDebugOnlyAccountType(account.accountType, account.symbol) && isDebug;
+    const isNotDebugOnlyAccount = (account: Account) =>
+        !isDebugOnlyAccountType(account.accountType, account.symbol);
```
Lemonexe: This is quite confusing to read, it all sounds the same :sweat_smile:
I know you only moved the code around, but it's an opportunity to simplify: both of these fns can be replaced with one, which will do:
`isDebug ||  !isDebugOnlyAccountType(account.accountType, account.symbol);`

Edit: perhaps the new fn could be named `shouldDisplayDebugOnly` :thought_balloon:

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
Lemonexe: to be fair, other its in this file are labeled similarly, for example `enhanceAddresses`.
I don't mean to to defend it, I think this file could do with a cleanup, I'd like to see one `describe` per one tested function instead of  `describe('account utils'`, and `it` should describe tested behavior.

But this PR is consistent with current state :shrug:

peter-sanderson: I see. I am not really fan of multiple (exported) functions in one file. I would like to see separate files for functions and separate test-files for them respectively.

And this way the `describe` would not be needed at all.

suite-common/wallet-utils/src/accountUtils.ts
enjojoy: Thank you, it's a nice note

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: I think I will leave the descriptions how they are for the consistency with the file, but it's a good point and we can do some refactoring later

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: I see what you mean and I agree

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: done

suite-common/wallet-utils/src/accountUtils.ts
enjojoy: Thanks, good to know

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,114 @@ describe('account utils', () => {
...
+        );
+    });
+
+    it('filterReceiveAccounts: isAnotherNetwork', () => {
+        const filteredAccounts = [
+            getWalletAccount({ symbol: 'btc', accountType: 'coinjoin' }),
+            getWalletAccount({ symbol: 'btc', accountType: 'taproot' }),
+            getWalletAccount({ symbol: 'btc', accountType: 'legacy' }),
+            getWalletAccount({ symbol: 'btc', accountType: 'segwit' }),
+            getWalletAccount({ symbol: 'btc', accountType: 'ledger' }),
```
Lemonexe: Could add a non btc account here, so you can test that it has been filtered out.
Analogically to case of `filterReceiveAccounts: isDebug`

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,114 @@ describe('account utils', () => {
...
+            receiveNetworks,
+        });
+    };
+
+    it('filterReceiveAccounts: isDebug', () => {
+        const filteredAccounts = [
+            getWalletAccount({ symbol: 'eth', accountType: 'legacy' }),
+            getWalletAccount({ symbol: 'eth', accountType: 'normal' }),
+            getWalletAccount({ symbol: 'eth', accountType: 'ledger' }),
+        ];
```
Lemonexe: Could add an actually debug account here, so you can test that it has been _filtered out_.

:thought_balloon: tip to make your life easier: you can rename this `filteredAccounts` to `normalAccounts`, then
`const filteredAccounts = [...normalAccounts, getWalletAccount(regtest or something)]`,
and test against `normalAccounts`

Lemonexe (approved):
As far as I'm concerned, LGTM :rocket: but please address other reviewers first :slightly_smiling_face: 
Also it'd be ideal to just slightly expand the tests, see my new comments.

suite-common/wallet-utils/src/accountUtils.ts
tomasklim: Because before the method returned `undefined`. Now you return `[]`

tomasklim: imho if it should return `[]` then the condition really does not need to be there.

enjojoy: Oh right, I see now

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: I will leave the description consistent with the file

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: Eth legacy and ledger accounts are debug accounts

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: But I'm filtering them from a list of all the accounts I have declared above

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
tomasklim: regtest is debug network
eth legacy is debug account type

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,113 @@ describe('account utils', () => {
...
+
+        expect(runFilterReceiveAccouns({})).toEqual(filteredAccounts);
+    });
+
+    it('filterReceiveAccounts: isNotDebug', () => {
```
tomasklim: Please wrap those `filterReceiveAccounts` `it`s into new `describe` to make it cleaner

`describe`s can be nested so

```
describe('account-utils', () => {
    ....

    describe('filterReceiveAccounts', () => {
        it('blablabla', () => {
            ....
        });

        it('blabla', () => {
           ...
        });
    });
```
should be fine

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -1153,3 +1154,52 @@ export const isTokenMatchesSearch = (token: TokenInfo, search: string) => {
         token.policyId?.toLowerCase().includes(search)
     );
 };
+
+export const isDebugOnlyAccountType = (
```
tomasklim: tbh this file is huge, I would put those new methods into a new file and same with tests. It is going to be better also because the fact that there are helper methods and types for the tests

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
tomasklim: Let's ignore this in case you move it to a new file

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,113 @@ describe('account utils', () => {
...
+        }),
+    ];
+
+    type RunFilterReceiveAccountsTestParams = {
+        isDebug?: boolean;
+        receiveNetwork?: string;
+        deviceState?: `${string}@${string}:${number}`;
+    };
+
+    const runFilterReceiveAccouns = ({
```
tomasklim: is it safe to create such complex helper method for tests? what if there is bug too @Lemonexe @peter-sanderson

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,113 @@ describe('account utils', () => {
...
+    });
+
+    it('filterReceiveAccounts: isEmptyAccount and isNotVisibleAccount + isEmptyAccount and isNotVisibleAccount but isFirstNormalAccount', () => {
+        const filteredAccounts = [
+            getWalletAccount({ symbol: 'sol', accountType: 'normal', empty: true, visible: false }),
+            getWalletAccount({ symbol: 'sol', accountType: 'ledger' }),
+        ];
+
+        expect(runFilterReceiveAccouns({ receiveNetwork: 'sol' })).toEqual(filteredAccounts);
+    });
```
tomasklim: I think we are missing test for:
1. empty accounts array
2. non-existent network in network collection
3. mixed account types - debug and non-debug

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -337,4 +340,113 @@ describe('account utils', () => {
         account.addresses = enhanceAddresses(accountInfo, account);
         expect(account.addresses.change).toEqual([]);
     });
+
+    it('isDebugOnlyAccountType', () => {
+        expect(isDebugOnlyAccountType('legacy', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('segwit', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('coinjoin', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('taproot', 'btc')).toBe(false);
```
tomasklim: please add regtest here

tomasklim (changes requested):

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -1153,3 +1154,52 @@ export const isTokenMatchesSearch = (token: TokenInfo, search: string) => {
...
+    if (!symbol) return false;
+
+    const network = getNetwork(symbol);
+
+    const accountTypeInfo = network.accountTypes[accountType];
+
+    return !!accountTypeInfo?.isDebugOnlyAccountType;
+};
+
+type FilterRecieveAccountsProps = {
```
tomasklim:
```suggestion
type FilterReceiveAccountsProps = {
```

tomasklim (dismissed):

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
peter-sanderson: I don't really agree with the argument of consistency in general, as it can be used to stop any improvement. 


However, if you separate this `isDebugOnlyAccountType`  (and `filterReceiveAccounts`) into its own files as Tom asks [here](https://github.com/trezor/trezor-suite/pull/14455/files#r1776429181) the consistency concern won't be the case anyway

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
peter-sanderson: I see this as necessary helper to orchestrate the test inputs. basically creating mocks and passing them into the `filterReceiveAccounts` so it probably necessary somehow

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: Not sure what you mean by mixed account types, I have a test for when Debug is on - then we get all the account types - debug and non-debug, when debug is off we get only non-debug, I have a test for that too.

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: Okay yes, I extracted it to another file and will reword the descs

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: I made a new file

suite-common/wallet-utils/src/accountUtils.ts
enjojoy: I extracted it into a new file

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: I think it's necessary, but I didn't add anything new to it, just copied the behavior I've seen in useCoinmarketVerifyAccount

image

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
enjojoy: added, tho it seems like it's not debug-only

suite-common/wallet-utils/src/accountUtils.ts
tomasklim: Where?

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
tomasklim: because it is debug only network, not debug account type

suite-common/wallet-utils/src/accountUtils.ts
enjojoy: Oh sorry, I only extracted tests, will extract this as well🤦🏻‍♀️

suite-common/wallet-utils/src/__tests__/filterReceiveAccounts.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        accounts,
+        deviceState: device.state,
+        receiveNetwork,
+        isDebug,
+        receiveNetworks,
+    });
+};
+
+describe('filter receive accounts', () => {
+    it('check if account is debug only type', () => {
```
peter-sanderson: absolute nitpick:
```suggestion
    it('checks if account is debug only type', () => {
```

suite-common/wallet-utils/src/__tests__/filterReceiveAccounts.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        expect(isDebugOnlyAccountType('segwit', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('coinjoin', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('taproot', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('ledger', 'btc')).toBe(false);
+        expect(isDebugOnlyAccountType('legacy', 'eth')).toBe(true);
+        expect(isDebugOnlyAccountType('ledger', 'eth')).toBe(true);
+        expect(isDebugOnlyAccountType('normal', 'regtest')).toBe(false);
+    });
+
+    it('empty accounts array', () => {
```
peter-sanderson:
```suggestion
    it('returns no results, when no accounts are given', () => {
```

suite-common/wallet-utils/src/__tests__/filterReceiveAccounts.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        expect(runFilterReceiveAccouns({ accounts: [] })).toEqual([]);
+    });
+
+    it('when given a non-existing network in acccounts list', () => {
```
peter-sanderson:
```suggestion
    it('returns no results for non-existing network in acccounts list', () => {
```

suite-common/wallet-utils/src/__tests__/filterReceiveAccounts.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        expect(runFilterReceiveAccouns({ receiveNetwork: 'bnb' })).toEqual([]);
+    });
+
+    it('when debug mode is on', () => {
```
peter-sanderson:
```suggestion
    it('filters only debug accounts, when debug mode is on', () => {
```


.. and similarly for rest of the `it`s

peter-sanderson (approved):

peter-sanderson (approved):
Good job adding new test :clap:

enjojoy:
> Good job adding new test 👏

Thank you, I reworded the its

PR #14468, opened 2024-09-23
ci: improving update device authenticity script

karliatto:
## Description

 Before creating the new PR with new keys for device authenticity data we check if there was already previous PR open, if so we add a comment and close the previous one before creating new one, since we always should merge the latest one because it contains all the changes since last merged PR.

### DIFF
--- a/scripts/ci/check-connect-data.ts
+++ b/scripts/ci/check-connect-data.ts
@@ -2,7 +2,7 @@ import fetch from 'cross-fetch';
 import fs from 'fs-extra';
 import path from 'path';
 import crypto from 'crypto';
-import { commit } from './helpers';
+import { comment, commit } from './helpers';
 
 const { exec } = require('./helpers');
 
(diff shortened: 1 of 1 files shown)


scripts/ci/check-connect-data.ts
```diff
@@ -81,6 +81,49 @@ const updateConfigFromJSON = async () => {
...
+                '--state=open',
+            ]);
+
+            console.log('prList', prList);
+
+            if (prList.stdout !== '') {
+                const prNumbers = prList.stdout.match(/(?<=\t)(\d+)(?=\t)/g);
+
+                console.log(`Found open pull requests ${prNumbers}. Closing...`);
+                for (const prNumber of prNumbers) {
```
mroz22, line 102: I guess after this is merged we can always (probably) expect only one record in prNumbers, right? but it is always better to expect the worst case scenario

mroz22 (approved):
looks useful

scripts/ci/check-connect-data.ts
karliatto: Yes, you are right. If everything works well we should never see more than one of those PRs open.

PR #14472, opened 2024-09-23
fixup! chore(suite-native): use new Networks structure

vytick:
Idea on how to achieve previous behaviour for coinjoin and available derivations discovery

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -635,9 +635,22 @@ export const startDescriptorPreloadedDiscoveryThunk = createThunk(
             return;
         }
 
+        const enabledNetworks = selectDeviceEnabledDiscoveryNetworkSymbols(getState());
+
+        let availableCardanoDerivations: ('normal' | 'legacy' | 'ledger')[] = [];
+        if (enabledNetworks.some(network => network.networkType === 'cardano')) {
+            availableCardanoDerivations =
+                (await dispatch(
+                    getCardanoSupportedAccountTypesThunk({
+                        deviceState,
+                    }),
+                ).unwrap()) ?? [];
+        }
+
         const networksWithUnfinishedDiscovery = selectNetworksWithUnfinishedDiscovery(
             getState(),
             forcedAreTestnetsEnabled,
+            Array.from(availableCardanoDerivations),
         );
 
         // Start tracking duration and networks for analytics purposes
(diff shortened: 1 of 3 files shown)


suite-native/discovery/src/utils.ts
```diff
@@ -1,18 +1,34 @@
...
     enabledNetworks.filter(network =>
         // if there is no account for this network -> we should have at least one account even if added via Add Coin
         // or if there is at least one visible account of this type, we should have at least one hidden account so adding funds outside of this app is detected and account shown
-        // the whole network is considered undiscovered if this is found for either 'normal' account, or any of the other accountTypes
+        // the whole network is considered undiscovered if this is found for either 'normal' account, or any of supported accountTypes
         normalizeNetworkAccounts(network).some(({ accountType }) => {
+            // coinjoin and some cardano derivation might not be supported (cardano derivations are provided)
+            if (
+                'coinjoin' === accountType ||
+                (network.symbol === 'ada' &&
```
Lemonexe, line 25: Maybe rather `networkType` for consistency? Such pattern is used in `discoveryThunks.ts`.
Also, this could maybe break `tada`, which is the only other cardano-type network.

```suggestion
                (network.networkType !== 'cardano' &&
```

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -635,9 +635,22 @@ export const startDescriptorPreloadedDiscoveryThunk = createThunk(
             return;
         }
 
+        const enabledNetworks = selectDeviceEnabledDiscoveryNetworkSymbols(getState());
+
+        let availableCardanoDerivations: ('normal' | 'legacy' | 'ledger')[] = [];
+        if (enabledNetworks.some(network => network.networkType === 'cardano')) {
```
Lemonexe, line 641: TS broken. This does what is intended :slightly_smiling_face: 

```suggestion
        if (enabledNetworks.some(networkSymbol => getNetwork(networkSymbol).networkType === 'cardano')) {
```

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -635,9 +635,22 @@ export const startDescriptorPreloadedDiscoveryThunk = createThunk(
...
+                    getCardanoSupportedAccountTypesThunk({
+                        deviceState,
+                    }),
+                ).unwrap()) ?? [];
+        }
+
         const networksWithUnfinishedDiscovery = selectNetworksWithUnfinishedDiscovery(
             getState(),
             forcedAreTestnetsEnabled,
+            Array.from(availableCardanoDerivations),
```
Lemonexe, line 653: what is the reason for `Array.from`?

Lemonexe (approved):
LGTM, just with two improvement ideas and one fix.

But I can do that in the parent PR as well..

suite-native/discovery/src/discoveryThunks.ts
vytick: good point no need

Lemonexe:
I will implement my comments in the parent PR.

PR #14479, opened 2024-09-23
feat(suite-native): send decimals validation

### LINKED ISSUE #14370: Display error when more than max allowed decimal digits are used
**How to reproduce**
Go to the send form screen and type an amount with more than 8 decimal digits.
image

**Expected behavior**
The same error message as in the desktop Suite shall be displayed.
image

PeKne:
## Related Issue

Resolve #14370

### DIFF
--- a/suite-native/module-send/src/screens/SendOutputsScreen.tsx
+++ b/suite-native/module-send/src/screens/SendOutputsScreen.tsx
@@ -98,6 +98,8 @@ export const SendOutputsScreen = ({
         selectSendFormDraftByAccountKey(state, accountKey),
     );
 
+    const network = account ? getNetworkCompatible(account.symbol) : null;
+
     const form = useForm<SendOutputsFormValues>({
         validation: sendOutputsFormValidationSchema,
         context: {
@@ -106,6 +108,7 @@ export const SendOutputsScreen = ({
             availableAccountBalance: account?.availableBalance,
             isValueInSats: isAmountInSats,
             normalFeeMaxAmount,
+            decimals: network?.decimals,
         },
         defaultValues: {
             outputs: sendFormDraft?.outputs ?? DEFAULT_VALUES,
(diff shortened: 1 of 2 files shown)


suite-native/module-send/src/screens/SendOutputsScreen.tsx
```diff
@@ -98,6 +98,8 @@ export const SendOutputsScreen = ({
         selectSendFormDraftByAccountKey(state, accountKey),
     );
 
+    const network = account ? getNetworkCompatible(account.symbol) : null;
```
vytick, line 101: please, change not to use deprecated method:

```
/**
 * @deprecated use `networks[symbol]` (exact type) or `getNetwork` (generic type) from '@suite-common/wallet-config'
 */
```

vytick (changes requested):
Looks good 👍, but please change how to fetch network

suite-native/module-send/src/screens/SendOutputsScreen.tsx
PeKne: Could we please leave it as it is? 🙏 @Lemonexe is already changing this file in his follow-up refactor PR: https://github.com/trezor/trezor-suite/pull/14398#discussion_r1771440838. Then the typing will behave correctly.

Lemonexe: Yes, the `composeContext` still works with `networkCompatible`, and that is refactored in the last PR.

I left it as the last thing, because it's heavily interconnected with `selectedAccountState`, and with suite discovery, so it'be a nightmare to refactor one without the other.

vytick: okay, then 👍

vytick (approved):
As discussed the way to get network data will be changed in following PR

PR #14483, opened 2024-09-23
Feat/empty device label

### LINKED ISSUE #12881: Do not rename device default name to "My Trezor"
Firmware currently names devices by their model name as default, like "Trezor Safe 3". Please do not rename it to "My Trezor" during initialization.

Image

komret:
## Description
- Do not set device label automatically, it can only be set by user action (previously it was set when creating a new wallet to "My Trezor".
- If device label is not set, display device name instead (e.g. Trezor Safe 5).

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12881

### Before:
Screenshot 2024-09-23 at 13 25 54
Screenshot 2024-09-23 at 13 58 41
Screenshot 2024-09-23 at 13 58 57

### After:
Screenshot 2024-09-23 at 15 51 20
Screenshot 2024-09-23 at 13 58 28
This is not value but placeholder:
Screenshot 2024-09-23 at 16 25 00

### DIFF
--- a/packages/suite/src/components/suite/notifications/NotificationRenderer/NotificationRenderer.tsx
+++ b/packages/suite/src/components/suite/notifications/NotificationRenderer/NotificationRenderer.tsx
@@ -1,6 +1,8 @@
 import type { ComponentType } from 'react';
+import { useSelector } from 'react-redux';
 
 import { AUTH_DEVICE, type NotificationEntry } from '@suite-common/toast-notifications';
+import { selectDeviceLabelOrName } from '@suite-common/wallet-core';
 import { DEVICE } from '@trezor/connect';
 
 import { NotificationViewProps } from 'src/components/suite';
@@ -62,6 +64,8 @@ export const NotificationRenderer = ({
     notification,
     render,
 }: NotificationRendererProps): JSX.Element => {
+    const deviceLabel = useSelector(selectDeviceLabelOrName);
+
     switch (notification.type) {
         case 'acquire-error':
             return error(render, notification, 'TOAST_ACQUIRE_ERROR');
@@ -223,7 +227,7 @@ export const NotificationRenderer = ({
                     variant="info"
                     message="EVENT_DEVICE_CONNECT"
                     messageValues={{
-                        label: notification.device.label,
+                        label: deviceLabel,
                     }}
                 />
             );
--- a/suite-common/wallet-core/src/device/deviceReducer.ts
+++ b/suite-common/wallet-core/src/device/deviceReducer.ts
@@ -796,12 +796,6 @@ export const selectIsPortfolioTrackerDevice = (state: DeviceRootState) => {
     return device?.id === PORTFOLIO_TRACKER_DEVICE_ID;
 };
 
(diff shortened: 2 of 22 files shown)


suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -811,10 +805,28 @@ export const selectDeviceNameById = (
     return device?.name ?? null;
 };
 
-export const selectDeviceLabel = (state: DeviceRootState) => {
-    const selectedDevice = selectDevice(state);
+export const selectDeviceLabel = (state: DeviceRootState): string => {
+    const device = selectDevice(state);
+
+    return device?.features?.label ?? '';
```
Nodonisko: Please don't use fallback to empty string or don't use `if("empty string")` later, it's confusing and too magical that empty string is false in JS

komret: ecc4143c4b862e3cc2ee9ca90503f5b881d476f9

Lemonexe: Exact `=== ''` comparison is less confusing, but still, `null` would be more semantic than `''`.

Could `selectDeviceLabel` return `string | null`? I see it is only used in two places: `ChangeDeviceLabel, DeviceInfoModalScreen`, everywhere else it's `selectDeviceLabelOrName` which would still return only `string`.
But maybe it's not worth it, we can discuss IRL :slightly_smiling_face:

packages/suite/src/components/suite/notifications/NotificationRenderer/NotificationRenderer.tsx
```diff
@@ -223,7 +223,7 @@ export const NotificationRenderer = ({
                     variant="info"
                     message="EVENT_DEVICE_CONNECT"
                     messageValues={{
-                        label: notification.device.label,
+                        label: notification.device.features?.label || notification.device.name,
```
Lemonexe: I can see that `useSelector(selectDeviceLabelOrName)` because it'd have to be defined at top of component, above `switch`. Why not then extract this as a component, so it can call the hook internally?

But it's a nit, I guess this is fine.

Lemonexe (approved):
Tested with physical trezor and onboarding works fine :heavy_check_mark:
I tested some of the usages `selectDeviceLabelOrName` and they work fine, so I'd trust the others :heavy_check_mark: 
Codewise LGTM, though [with comment above](https://github.com/trezor/trezor-suite/pull/14483#discussion_r1772691257) :speech_balloon:

suite-common/wallet-core/src/device/deviceReducer.ts
komret: 324f51237dc1ecd5bb34b65caf3e8e5b1a712206

packages/suite/src/components/suite/notifications/NotificationRenderer/NotificationRenderer.tsx
komret: 🤔 I think there's no perfect solution. e6ecafe58388ed7a484741dbf0258774b65dae10

Lemonexe (approved):

PR #14484, opened 2024-09-23
feat(connect-deeplink): deeplinkUrl setting (for app linking)

martykan:
## Description

Add `deeplinkUrl` setting to allow changing the URL to Suite. 
Add intent filters for opening the app from https://dev.suite.sldev.cz/connect/.../deeplink/
Update example app URL for testing
Add [fallback page](https://dev.suite.sldev.cz/connect/feat/native-connect-popup-deeplink-url/deeplink/) for deeplinks (telling user to download app)

### DIFF
--- a/packages/connect-deeplink/src/index.ts
+++ b/packages/connect-deeplink/src/index.ts
@@ -19,7 +19,7 @@ export class TrezorConnectDeeplink implements ConnectFactoryDependencies {
     private _settings: ConnectSettings;
     private messagePromises: Record<number, Deferred<any>> = {};
     private messageID = 0;
-    private schema = 'trezorsuitelite';
+    private defaultDeeplinkUrl = 'trezorsuitelite://connect';
 
     public constructor() {
         this._settings = {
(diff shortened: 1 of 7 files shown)


packages/connect-deeplink/src/index.ts
```diff
@@ -174,7 +175,7 @@ export class TrezorConnectDeeplink implements ConnectFactoryDependencies {
     }
 
     private buildUrl(method: string, params: any, callback: string) {
-        return `${this.schema}://connect?method=${method}&params=${encodeURIComponent(
+        return `${this._settings.deeplinkUrl}?method=${method}&params=${encodeURIComponent(
```
karliatto, line 170: I am having here some changes related to this https://github.com/trezor/trezor-suite/pull/14490/files#diff-161cfff841ef61661da9b9f9b576e2f9d4d96db4dbc697b6202506404d5acdcdR4

But we can adjust once we are ready to merge it.

suite-native/app/app.config.ts
```diff
@@ -169,6 +169,29 @@ export default ({ config }: ConfigContext): ExpoConfig => {
                 monochromeImage: './assets/appIcon_android.png',
                 ...appIconAndroid,
             },
+            intentFilters:
```
karliatto, line 171: If I understand well this configuration is only for Android, right?

martykan: Yes this PR only supports Android, I would leave iOS for later, after the testing phase

karliatto (approved):
Looks good.

Issue #13470, opened 2024-07-23
Coin Enabling - message system compliance


vytick:
Make sure that message system can correctly detect which coins are enabled to correctly show banner based on this.

STew790:
**QA NOK**
https://github.com/trezor/trezor-suite/pull/14233 this is still not merged :thinking:

matejkriz:
> **QA NOK** #14233 this is still not merged 🤔

Now it's released on develop, you can test it.

vytick:
@STew790 can be tested now with a bit different testsing message (for nmc, dash, btg, vtc)

STew790:
**QA OK**
Yes, this is now working correctly for me :tada:

PR #14486, opened 2024-09-23
Simplify mobile ENV variables

matejkriz:
## Description

- There was a bug using `.env.production` even on develop build of Android in createBundleReleaseJsAndAssets step of gradle.
- ENV variables from `eas.json` are used for EAS builds.
- `.env.development` and `.env.test` files are used for local builds and CI E2E builds.
- Local ENV vars can be overriden by `.env.development.local` and `.env.test.local` files.
- We don't use special decorators any more, so we can switch to "default" babel-preset-expo.

## Related Issue

I hope it can unblock testing of https://github.com/trezor/trezor-suite/issues/13470


vdovhanych (approved):

suite-native/app/.env.debug
```diff
@@ -1,3 +0,0 @@
-EXPO_PUBLIC_ENVIRONMENT=debug
-EXPO_PUBLIC_CODESIGN_BUILD=false
-EXPO_PUBLIC_IS_ANALYTICS_LOGGER_ENABLED=false
```
vytick, line 3: `EXPO_PUBLIC_IS_ANALYTICS_LOGGER_ENABLED`  is now intended to use only in my local .env right?

suite-native/config/README.md
```diff
@@ -1,11 +1,3 @@
 # @suite-native/config
 
 This package contains helper functions based on configuration variables for different environments of our React Native mobile application.
-
-## Config variables
```
vytick, line 5: It might be good to add some comment on how to set/change ENV variables somewhere, wdyt?

Nodonisko:
What about e2e where we are triggering JS bundling manually and not in EAS?

vytick (approved):
Looks good, tested locally in emulator and on device with assembleRelease debug build

suite-native/app/.env.debug
matejkriz: can be used in .env.development.local, mentioned in README

suite-native/config/README.md
matejkriz: Check the README now, please

PR #14487, opened 2024-09-23
feat(mobile): use flashlist for tokens bottom sheet

### LINKED ISSUE #13019: App is unusuable(or very slow) with ETH account with large number of tokens

Nodonisko:
## Description

Token select bottom now uses FlashList so when there is big number of tokens it's much faster. Also simplified overall structure of Account List components.

## Related Issue

Resolve #13019


suite-native/accounts/src/selectors.ts
```diff
@@ -93,3 +88,51 @@ export const selectAccountFiatBalance = (
...
+
+export const selectAccountListSections = memoizeWithArgs(
+    (state: NativeAccountsRootState, accountKey?: AccountKey | null) => {
+        if (!accountKey) return EMPTY_ARRAY;
+        const account = selectAccountByKey(state, accountKey);
+        if (!account) return EMPTY_ARRAY;
+
+        const sections: AccountSelectBottomSheetSection[] = [];
+
+        const couldHasTokens = isCoinWithTokens(account.symbol);
```
vytick:
```suggestion
        const canHaveTokens = isCoinWithTokens(account.symbol);
```

suite-native/accounts/src/selectors.ts
```diff
@@ -93,3 +88,51 @@ export const selectAccountFiatBalance = (
...
+        const tokens = selectFilterKnownTokens(state, account.symbol, account.tokens ?? []);
+        const hasAnyKnownTokens = couldHasTokens && !!tokens.length;
```
vytick:
```suggestion
        const hasKnownTokens = couldHasTokens && !!tokens.length;
```

suite-native/atoms/src/Sheet/BottomSheet.tsx
```diff
@@ -112,22 +114,28 @@ export const BottomSheet = ({
...
+                                    keyboardShouldPersistTaps="handled"
+                                >
+                                    <Animated.View>
+                                        <Box paddingHorizontal="medium" {...boxProps}>
+                                            {children}
+                                        </Box>
+                                    </Animated.View>
+                                </ScrollView>
+                            ) : (
+                                <Box {...boxProps} style={{ height: '100%' }}>
```
vytick, line 135: this has different style than the box inside scrollview and there is no Animated.View. Is that intended?

suite-native/accounts/src/components/TokenSelectBottomSheet.tsx
```diff
@@ -17,27 +20,30 @@ export const TokenSelectBottomSheet = ({
...
-                setSelectedAccount(null);
-            }}
-        >
-            <AccountListItemInteractive
-                key={selectedAccount.key}
-                account={selectedAccount}
-                onSelectAccount={handleSelectAccount}
-            />
-        </BottomSheet>
+        <AccountSelectBottomSheet
```
vytick, line 43: it due tohow it behaves for `isScrollable === false` it can now look like this. I am not sure whether this is expected:

matejkriz (commented):
The performance is amazing! 

I just faced the problem of scrolling up. It's possible, but the scrollview has to be moving, so you have to start scrolling down a bit and then scroll up, otherwise bottomsheet is closed instead 🙈 But I guess it's separate issue and minor.
https://github.com/user-attachments/assets/b07c41da-856d-48d0-ae09-411d97769458

suite-native/atoms/src/Sheet/BottomSheet.tsx
Nodonisko: I think `Animated.View` might be even redudant inside that scrollview, it seems it doesn't have any obvious purpose. Maybe @PeKne may know why it's there?

Nodonisko:
@matejkriz I noticed that randomly but I was not able to reproduce it so thanks for finding out. I think it will require some extensive changes to BottomSheet, not sure if I want to make this part of this PR, I will probably create issue for that.

suite-native/accounts/src/components/TokenSelectBottomSheet.tsx
Nodonisko: Probably yes, I decided that I will tweak that in another PR because it will require some extensive BottomSheet refactorings

suite-native/accounts/src/selectors.ts
Nodonisko: Renamed

suite-native/accounts/src/selectors.ts
Nodonisko: This one doesn't feel right for me :D

suite-native/atoms/src/Sheet/BottomSheet.tsx
Nodonisko: Anyway I created follow up #14503 which could be used to refine this.

suite-native/accounts/src/components/TokenSelectBottomSheet.tsx
vytick: Okay. Some refactoring there might be useful. it does odd things with scrolling for no obvious reason (detecting pan instead of scroll).

vytick (approved):
nice improvement 👍  🚀

PR #14489, opened 2024-09-23
Remove `device.label`

komret:
## Description
Follow-up to https://github.com/trezor/trezor-suite/pull/14483.
Let's not set `device.label`. Instead, fall back to `device.name` if needed. The label can be accessed via `device.features.label`.
This is a breaking change.


mroz22:
todo: 
breaking, include in connect v 10 branch

PR #14490, opened 2024-09-24
Feat: native popup improvements

karliatto:
## Description
* Add version to deeplink URL
* Use buildURL function from `@trezor/connect-deeplink' package so it is not duplicated.


karliatto:
I am moving the versioning discussion to new PR https://github.com/trezor/trezor-suite/pull/14621

PR #14495, opened 2024-09-24
Feat/desktop user agent

tomasklim:
## Description

- add Suite version to user agent of requests going from desktop app 
- version is correctly added to the user agent

If needed, we can change the hardcoded `Trezor Suite` name to a dynamic value in the future.

## Screenshots:
Cloudflare:
Screenshot 2024-09-24 at 10 18 20

### DIFF
--- a/packages/blockchain-link/src/workers/blockbook/websocket.ts
+++ b/packages/blockchain-link/src/workers/blockbook/websocket.ts
@@ -21,6 +21,7 @@ import type {
 } from '@trezor/blockchain-link-types/src/params';
 
 import { BaseWebsocket } from '../baseWebsocket';
+import { getSuiteVersion } from '@trezor/env-utils';
 
 interface BlockbookEvents {
     block: BlockNotification;
@@ -52,7 +53,7 @@ export class BlockbookAPI extends BaseWebsocket<BlockbookEvents> {
             agent: this.options.agent,
             headers: {
                 Origin: 'https://node.trezor.io',
-                'User-Agent': 'Trezor Suite',
+                'User-Agent': `Trezor Suite ${getSuiteVersion()}`,
                 ...this.options.headers,
             },
         });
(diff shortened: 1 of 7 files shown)


packages/blockchain-link/src/workers/blockbook/websocket.ts
```diff
@@ -52,7 +53,7 @@ export class BlockbookAPI extends BaseWebsocket<BlockbookEvents> {
             agent: this.options.agent,
             headers: {
                 Origin: 'https://node.trezor.io',
-                'User-Agent': 'Trezor Suite',
+                'User-Agent': `Trezor Suite ${getSuiteVersion()}`,
```
marekrjpolak, line 56: While I believe this shouldn't break anything and therefore can be merged, I see two possible downsides:

1) As the comment stated, web builds ignore the options (second param) passed to `WebSocket`, therefore this won't work in web Suite. Which is probably expected, right?

2) `getSuiteVersion` util just returns anything what's in `process.env.VERSION`, so if a 3rd party uses Connect in Node environment, it's possible to have something like `Trezor Suite 9.x.x` there, because of connect version. Again, not a big deal?

tomasklim: 1. yes
2. I am fine with this

marekrjpolak (approved):

bosomt:
**QA OK**

image

**Info:**
 - Suite version: desktop 24.10.1 (3017e5864a176715b1023945397ef49e8aa24d5e)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.10.1 Chrome/126.0.6478.234 Electron/31.6.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: BridgeTransport 3.0.0-bundled.24.10.1

tomasklim:
@bosomt it was like that even before. This change is for desktop and it is probably not possible to test

STew790:
I just checked it with Tali and it is OK from him as well :+1: .

PR #14496, opened 2024-09-24
chore(tenv): always set the bridge version when calling tenv, 2.0.33 if not set

grdddj:
Should minimize future issues with changing the bridges in `tenv`

Potentially, the `2.0.33` could be an exportable constant, and tests can use it as an expected version (couple tests currently hardcode `2.0.33`)

### DIFF
--- a/packages/trezor-user-env-link/src/api.ts
+++ b/packages/trezor-user-env-link/src/api.ts
@@ -61,6 +61,8 @@ export const MNEMONICS = {
         'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about',
 };
 
+export const DEFAULT_BRIDGE_VERSION = '2.0.33';
+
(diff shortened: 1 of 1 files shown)


vdovhanych (approved):

packages/trezor-user-env-link/src/api.ts
```diff
@@ -158,6 +158,7 @@ export class TrezorUserEnvLinkClass extends TypedEmitter<WebsocketClientEvents>
         return null;
     }
     async startBridge(version?: string) {
```
mroz22: good enough, just one nitpick -  javascript way of writing this is:

```
    async startBridge(version = '2.0.33') {
```

grdddj: Right, thanks, corrected with a constant set

PR #14505, opened 2024-09-24
fix(suite): test id in account menu

seibei-iguchi:


Lemonexe (approved):
If this fixes the tests, then LGTM, thanks :+1:

PR #14506, opened 2024-09-24
fix(suite-common): sanitise fetching graph history rates for empty account

### LINKED ISSUE #14502: Error in graph on empty account with range "ALL"
**Describe the bug**
When I sync or connect and empty account and select "ALL" on the graph range, the graph shows an error.

**Info:**

-   OS: name, version

**How to reproduce**

1. Connect an empty seed device or sync empty account
2. On the graph select the "ALL" range
3. Observer the error in the graph

**Screenshots**
image

vytick:
When no start timestamp was selected for wallet with only empty accounts, it threw range error. This was because `getTimestampsInTimeFrame` would receive equal start date and end date. This can happen only if user has no account with txn history at all.  

So in such case I check and change the start date to 1y ago, to work the same as when such option(1y) is  selected.

## Related Issue

Resolve #14502

### DIFF
--- a/suite-common/graph/src/graphDataFetching.ts
+++ b/suite-common/graph/src/graphDataFetching.ts
@@ -409,6 +409,19 @@ export const getMultipleAccountBalanceHistoryWithFiat = async ({
             findOldestBalanceMovementTimestamp,
             fromUnixTime,
         );
+
+        // if there were no balance movements at all,
+        // findOldestBalanceMovementTimestamp resulted with start date being the same as end date
+        // use 1 year into past and return zeroes for the range
+        if (startOfTimeFrameDate.getTime() === endOfTimeFrameDate.getTime()) {
+            let startDate = startOfTimeFrameDate;
+            startDate.setDate(startDate.getHours() - 8760);
+
+            return [
+                ...getTimestampsInTimeFrame(startDate, endOfTimeFrameDate, numberOfPoints - 1),
+                getUnixTime(endOfTimeFrameDate),
+            ].map(t => ({ date: fromUnixTime(t), value: 0 }));
+        }
     }
 
     // Last timestamp must be endOfTimeFrameDate because blockchainGetAccountBalanceHistory it's not always reliable for coins like ETH.


suite-common/graph/src/graphDataFetching.ts
```diff
@@ -409,6 +409,13 @@ export const getMultipleAccountBalanceHistoryWithFiat = async ({
             findOldestBalanceMovementTimestamp,
             fromUnixTime,
         );
+
+        // if there were no balance movements at all,
+        // findOldestBalanceMovementTimestamp resulted with start date being the same as end date
+        // use 1 year into past instead, otherwise getTimestampsInTimeFrame will throw range error
+        if (startOfTimeFrameDate.getTime() === endOfTimeFrameDate.getTime()) {
```
Nodonisko, line 416: I am not sure if this si best fix. I think it may be better to just return early here since if there is no movement we know that graph will just zero all the time. I think `return []` or return `[{ date: 1, value: 0 }, {date: 2, value: 0}]` should work and it will be faster because otherwise we will be fetching timestamps for that whole year.

vytick: So in such case it should probably check for every time if there are no movements of any coin to return early for each option 1d/1w/1m/...

I was thinking about this, but it felt suboptimal for such edge case. And this is the same behaviour there is for 1y option.

vytick: Ok. I'll modify the behaviour for empty history.

matejkriz: Another thing I'd be a little worried about is mutating `startOfTimeFrameDate` which is passed as parameter to this function. I'd rather define a new constant for it locally... But maybe it's ok, it's just harder to think about possible consequences. (But the mutation was there even before this PR, I know...)

vytick: I've added check for empty account in 2091d876ab and kept the timeframe update to better UX when touching the graph.

Nodonisko: Sorry but it seems unecessary complicated now, I need to take a proper look and think about it.

Nodonisko: I still think that best solution is return early if  `startOfTimeFrameDate === endOfTimeFrameDate` because that's easy and will fix the issue. Don't care about the other timeframes it's not worth added complexity.

vytick: simplified [49b17fd](https://github.com/trezor/trezor-suite/pull/14506/commits/49b17fd0b5ef5065cce409f9e1a7d69e19f0f1d9)

Nodonisko (approved):

PR #14508, opened 2024-09-24
fix(suite): header alignment in accounts menu

seibei-iguchi:
### Before
Screenshot 2024-09-24 at 14 02 29

### After
Screenshot 2024-09-24 at 14 01 29


jvaclavik (approved):
Took me a while to see what was changed  :D

PR #14517, opened 2024-09-25
Finetune popup tests

mroz22:


packages/connect-popup/e2e/tests/popup-close.test.ts
```diff
@@ -434,6 +434,11 @@ test('popup should behave properly with subsequent calls', async ({ page, contex
     popupClosedPromise = new Promise(resolve => {
         popup.on('close', () => resolve(undefined));
     });
+
+    log(
+        'aha, this is probably expected to close automatically. but it does not. the reason is that this one already opens getPubKey popup, not getFeatures. getFetures seems to be handled without popup',
+    );
```
mroz22: UX-wise, it looks good. Can't it be that I already have permissions saved from a previous test so I don't need to accept permissions? 
I can't recall what recent change might have caused this.

martykan: I don't exactly know what change affected this. I was looking at this yesterday with Marek and on local the behavior was the same as the test - eg. the popup did show up for getFeatures, but it closes itself quickly since it's not actually needed. But in CI the popup doesn't even open.

packages/connect-popup/e2e/support/helpers.ts
```diff
@@ -149,7 +149,7 @@ export const setConnectSettings = async (
     { trustedHost = false, connectSrc }: { trustedHost?: boolean; connectSrc?: string },
     _isWebExtension?: boolean,
 ) => {
-    await explorerPage.goto(formatUrl(explorerUrl, `settings/index.html`));
+    await explorerPage.goto(formatUrl(explorerUrl, `settings`));
```
martykan: It's necessary to specify `index.html` for webextension

martykan:
Maybe `formatUrl` could intelligently add `index.html` based on the environment 🤔

martykan (approved):

PR #14518, opened 2024-09-25
Finetune install connect test

mroz22:
there is a [test](https://github.com/trezor/trezor-suite/actions/runs/11024014375/job/30616416523) that connect is installable from npm, types work, and we didn't screw up. 

the problem is that it is now falling in CI so we might have screwed up actually. 

to get this information earlier, I suggest extending it also to beta channel. 

the problem is that I don't understand what is going on there right now. Maybe @martykan got some clue 

image

### DIFF
--- a/packages/connect/e2e/test-yarn-install.sh
+++ b/packages/connect/e2e/test-yarn-install.sh
@@ -21,5 +21,5 @@ yarn add @trezor/connect@"$1"
 echo import TrezorConnect from \"@trezor/connect\" >index.ts
 
 # compile with typescript
-yarn add typescript@5.3.2
-yarn tsc ./index.ts --types node --esModuleInterop
+yarn add typescript@5.5.4
+yarn tsc ./index.ts --types node --types w3c-web-usb --esModuleInterop
--- a/packages/connect/e2e/test-yarn-install.sh
+++ b/packages/connect/e2e/test-yarn-install.sh
@@ -16,9 +16,9 @@ npm init -y
 touch yarn.lock
 
 # install connect package
-yarn add @trezor/connect
+yarn add @trezor/connect@"$1"
 # prepare minimal typescript implementation
-echo import TrezorConnect from \"@trezor/connect\" > index.ts
+echo import TrezorConnect from \"@trezor/connect\" >index.ts
 
 # compile with typescript
 yarn add typescript@5.3.2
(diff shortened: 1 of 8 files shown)


mroz22:
it looks like it can't resolve imports from sinclair. It reminded me of this script that is replacing imports from monorepo with actual ones https://github.com/trezor/trezor-suite/blob/develop/scripts/replace-imports.sh#L19 but it shouldn't be the case here 

also .mjs is suspicious. could be that I just need some additional configuration

martykan:
Yeah looks like we have a problem. 
Replacing `@sinclair/typebox/build/esm/index.mjs` -> `@sinclair/typebox` in the build does solve the issue.
In the old version it was correctly importing as `@sinclair/typebox`. Not sure what changed.

There is just one more regarding USB
Screenshot 2024-09-25 at 09 41 05

mroz22:
usb lib depends on w3c-web-usb types. so we probably need to make sure that it is always installed 

image

this should help I think 74c7bca553dc4e92709e9d8547b267db934179c9

packages/connect/e2e/test-yarn-install.sh
```diff
@@ -16,10 +16,10 @@ npm init -y
...
-yarn add @trezor/connect
+yarn add @trezor/connect@$1
 # prepare minimal typescript implementation
 echo import TrezorConnect from \"@trezor/connect\" > index.ts
 
 # compile with typescript
-yarn add typescript@5.3.2
-yarn tsc ./index.ts --types node --esModuleInterop
+yarn add typescript@5.5.4
+yarn tsc ./index.ts --types node --types w3-web-usb --esModuleInterop
```
martykan:
```suggestion
yarn tsc ./index.ts --types node --types w3c-web-usb --esModuleInterop
```

packages/connect/package.json
```diff
@@ -91,6 +91,7 @@
     "devDependencies": {
         "@trezor/trezor-user-env-link": "workspace:*",
         "@types/karma": "^6.3.8",
+        "@types/w3c-web-usb": "^1.0.10",
```
martykan: Shows up as an unused dependency, which makes sense since it's used in `@packages/transport` not `@packages/connect`

mroz22: and it shouldn't be needed since it is listed in `usb` package 

image

martykan (approved):

PR #14528, opened 2024-09-25
Feat(suite): Enhance discreet mode

### LINKED ISSUE #10567: Improve the "Hide balances" feature

Lemonexe:
Enhance it to actually replace the characters with `###`, making them completely illegible.

## Description

**Most of stuff done in 1035518754db3258f5e38c272b97a46f17151e5f:**

- HiddenPlaceholder now creates a context to inform components downstream that they should redact numbers
  - unfortunately it has to be responsibility of downstream components to change the content to `###`.
  - I attempted a centralized solution in [#13914](https://github.com/trezor/trezor-suite/pull/13914), and it was a dead-end :disappointed: 
- create helpers `RedactNumericalValue` and `useShouldRedactNumbers', utilize them in several such components
  - update it in `formatters`, `FiatHeader`, `FormattedCryptoAmounts`, `SummaryCards`
  - some of those components are broadly used, and may not be used within `HiddenPlaceholder` at all
  - but _if they are_ under downstream of `HiddenPlaceholder`, they will now redact numbers

**Tweaks:**
- 654893c82f4ed513ce3a4328b4bea5d65e96c47c Slightly refactor `FiatHeader` so it displays `###` only once, not twice
- 130d6c94e673211c3ed920d05a80d670c93d72e3 Allow `disableKeepingWidth`, because in order for it to work, it needs   `display:inline-block`, but that breaks text overflow in `TxDetailModal/IOAddress.tsx` (also it's unnecessary there)

**Occurrences:**

I verified all of the usages of `HiddenPlaceholder`, and _on main parts_ of the app they now do the `###`, not just blur :heavy_check_mark:

Still not done in some less mission-critical components, can be done subsequently: :construction:
[CardanoRewards](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/views/wallet/staking/components/CardanoRewards.tsx), [CoinmarketBalance](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/views/wallet/coinmarket/common/CoinmarketBalance.tsx), [CoinmarketFormInputAccountOption](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/views/wallet/coinmarket/common/CoinmarketForm/CoinmarketFormInput/CoinmarketFormInputAccountOption.tsx), [SellTransaction](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayout/CoinmarketAccountTransactions/SellTransaction.tsx), [BuyTransaction](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/views/wallet/coinmarket/common/CoinmarketLayout/CoinmarketAccountTransactions/BuyTransaction.tsx)

## Related Issue

Resolve #10567

## Screenshots:

Video showing it in action with hover behavior.
This is without cad5c662dabdf3c7eb7ca14271fcefe400e1e220, so blur is toned down, and bg is applied so you can see the widths..
EDIT: sry the video is outdated 🤦‍♂️ the "7 transactions" are now "### transactions"

[discreet.webm](https://github.com/user-attachments/assets/7e6f08c1-f124-455a-b2ee-78572c5ac940)

### DIFF
--- a/packages/suite/src/components/suite/HiddenPlaceholder.tsx
+++ b/packages/suite/src/components/suite/HiddenPlaceholder.tsx
@@ -1,15 +1,18 @@
-import { ReactNode, useEffect, useRef, useState } from 'react';
+import { ReactNode, useLayoutEffect, useRef, useState } from 'react';
 import styled, { css } from 'styled-components';
 import { useSelector } from 'src/hooks/suite';
 import { selectIsDiscreteModeActive } from 'src/reducers/wallet/settingsReducer';
+import { RedactNumbersContext } from '@suite-common/wallet-utils';
 
-interface WrapperProps {
+type WrapperProps = {
     $intensity: number;
     $discreetMode: boolean;
-}
+    $minWidth?: number;
+};
 
 const Wrapper = styled.span<WrapperProps>`
     font-variant-numeric: tabular-nums;
+    display: inline-block;
 
     ${({ $intensity, $discreetMode }: WrapperProps) =>
         $discreetMode &&
@@ -21,29 +24,49 @@ const Wrapper = styled.span<WrapperProps>`
                 filter: none;
             }
         `}
+
+    ${({ $minWidth }: WrapperProps) =>
+        $minWidth &&
+        css`
+            min-width: ${$minWidth}px;
+        `}
 `;
 
-export interface HiddenPlaceholderProps {
+export type HiddenPlaceholderProps = {
     enforceIntensity?: number;
     children: ReactNode;
     className?: string;
     ['data-testid']?: string;
-}
+};
 
+/**
+ * Wrapper for sensitive information, which should be hidden in discreet mode.
+ * ATTENTION: this is only half of the functionality, see also RedactNumericalValue
+ *
+ * HiddenPlaceholder does following:
+ * 1) applies CSS blur, which is removed on hover
+ * 2) through RedactNumbersContext it passes information downstream, so the content itself can be redacted
+ * 3) to prevent flickering on hover, it upholds min-width of redacted content when it is uncovered
+ *
+ * Note: content cannot be changed centrally here, it's responsibility of downstream components,
+ * they may consume the context using useShouldRedactNumbers(), or use RedactNumericalValue helper
+ */
 export const HiddenPlaceholder = ({
     children,
     enforceIntensity,
     className,
-    ...rest
+    'data-testid': dataTestId,
 }: HiddenPlaceholderProps) => {
     const ref = useRef<HTMLSpanElement>(null);
     const [automaticIntensity, setAutomaticIntensity] = useState(10);
-    const discreetMode = useSelector(selectIsDiscreteModeActive);
+    const [wrapperMinWidth, setWrapperMinWidth] = useState<undefined | number>(undefined);
+    const [isHovered, setIsHovered] = useState(false);
 
-    useEffect(() => {
-        if (ref.current === null) {
-            return;
-        }
+    const discreetMode = useSelector(selectIsDiscreteModeActive);
+    const shouldRedactNumbers = discreetMode && !isHovered;
+
+    useLayoutEffect(() => {
+        if (ref.current === null) return;
 
         const fontSize = Number(
             window
@@ -54,15 +77,33 @@ export const HiddenPlaceholder = ({
         setAutomaticIntensity(fontSize / 5);
     }, []);
 
+    // we only need to handle the case when revealed content is smaller than redacted, not vice versa.
+    // in such case, onMouseEnter shrinks content, and it may immediately onMouseLeave even when cursor is still.
+    const onMouseEnter = () => {
+        setIsHovered(true);
+        if (ref?.current) {
+            setWrapperMinWidth(ref.current.getBoundingClientRect().width);
+        }
+    };
+    const onMouseLeave = () => {
+        setIsHovered(false);
+        setWrapperMinWidth(undefined);
+    };
+
     return (
         <Wrapper
+            onMouseEnter={onMouseEnter}
+            onMouseLeave={onMouseLeave}
             $discreetMode={discreetMode}
             $intensity={enforceIntensity !== undefined ? enforceIntensity : automaticIntensity}
+            $minWidth={wrapperMinWidth}
             className={className}
             ref={ref}
-            data-testid={rest['data-testid']}
+            data-testid={dataTestId}
         >
-            {children}
+            <RedactNumbersContext.Provider value={{ shouldRedactNumbers }}>
+                {children}
+            </RedactNumbersContext.Provider>
         </Wrapper>
     );
 };
--- a/packages/suite/src/components/suite/HiddenPlaceholder.tsx
+++ b/packages/suite/src/components/suite/HiddenPlaceholder.tsx
@@ -8,11 +8,12 @@ type WrapperProps = {
     $intensity: number;
     $discreetMode: boolean;
     $minWidth?: number;
+    $disableKeepingWidth?: boolean;
 };
 
 const Wrapper = styled.span<WrapperProps>`
     font-variant-numeric: tabular-nums;
-    display: inline-block;
+    display: inline;
 
     ${({ $intensity, $discreetMode }: WrapperProps) =>
         $discreetMode &&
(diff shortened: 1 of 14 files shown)


packages/suite/src/components/suite/HiddenPlaceholder.tsx
```diff
@@ -21,29 +25,53 @@ const Wrapper = styled.span<WrapperProps>`
...
+ * 3) to prevent flickering on hover, it upholds min-width of redacted content when it is uncovered
+ *
+ * Note: content cannot be changed centrally here, it's responsibility of downstream components,
+ * they may consume the context using useShouldRedactNumbers(), or use RedactNumericalValue helper
+ */
 export const HiddenPlaceholder = ({
     children,
     enforceIntensity,
     className,
-    ...rest
```
Lemonexe, line 37: incidental cleanup, `rest` did not contain anything else but `'data-testid'`

suite-common/formatters/src/makeFormatter.tsx
```diff
@@ -40,11 +44,11 @@ export const makeFormatter = <TInput, TOutput, TDataContext extends DataContext
...
-    formatter.displayName = displayName;
+    const FormatterComponent: Formatter<TInput, TOutput, TDataContext> = props =>
+        <>{format(props.value, props, useShouldRedactNumbers())}</> ?? null;
+    FormatterComponent.displayName = displayName;
 
-    formatter.format = (value, dataContext = {}) => format(value, dataContext);
+    FormatterComponent.format = (value, dataContext = {}) => format(value, dataContext);
 
-    return formatter;
+    return FormatterComponent;
```
Lemonexe, line 53: It is actually a React component, so its name should be in CamelCase.
Otherwise I get an _unjustified_ error from eslint rule of hooks because of calling `useShouldRedactNumbers`.
Such injustice, smh.

suite-common/wallet-utils/src/discreetModeUtils.ts
```diff
@@ -0,0 +1,30 @@
+import { createContext, useContext } from 'react';
```
Lemonexe, line 1: Not sure if _all of this_ belongs to `wallet-utils`. Seemed suitable to me. Where would you place it?

Note: there are already some react-related functions in this package, so it's not strictly business logic.

packages/suite/src/views/wallet/transactions/components/SummaryCards.tsx
```diff
@@ -53,6 +57,13 @@ const DateWrapper = styled.span`
     white-space: nowrap;
 `;
 
+const NumberOfTransactions = ({ value }: { value: number }) => (
+    <Translation
+        id="TR_N_TRANSACTIONS"
+        values={{ value: useShouldRedactNumbers() ? DISCREET_PLACEHOLDER : value }}
+    />
+);
```
Lemonexe, line 62: Just FIY: this is a kinda catch with using `NumberOfTransactions` - you have to extract this as component, so the `useShouldRedactNumbers` is called downstream of `HiddenPlaceholder`; you can't place this code in `SummaryCards`. I'm just wondering if this might confuse people in future. Someone might think that it should work, because the hook is called inside `HiddenPlaceholder`, but it's still called in `SummaryCards` = still upstream. 

IMHO it's clear basic React, but then again, not everyone will study the `HiddenPlaceholder`, so they might not realize this catch.

packages/suite/src/components/suite/HiddenPlaceholder.tsx
```diff
@@ -21,29 +25,53 @@ const Wrapper = styled.span<WrapperProps>`
...
+    const [isHovered, setIsHovered] = useState(false);
+
     const discreetMode = useSelector(selectIsDiscreteModeActive);
+    const shouldRedactNumbers = discreetMode && !isHovered;
 
-    useEffect(() => {
-        if (ref.current === null) {
-            return;
-        }
+    useLayoutEffect(() => {
```
Lemonexe, line 73: not just because of this PR - I believe this should have always been `useLayoutEffect`

tomasklim:
1. sometimes I can see `###` for a while

https://github.com/user-attachments/assets/8781ba89-f402-4e96-ad8e-4aee87016b9d

2. would it be possible to set width of placeholder to the width of original text? 

https://github.com/user-attachments/assets/6f4378d8-aeb3-4489-b06a-004c0eb6eada

3. would it be possible to increase a `z-index` of those blurred texts so that the shadow is not "cut"?
Screenshot 2024-09-26 at 7 44 16

suite-common/formatters/src/formatters/prepareFiatAmountFormatter.ts
```diff
@@ -14,24 +15,29 @@ export const prepareFiatAmountFormatter = (config: FormatterConfig) =>
...
         const { intl, fiatCurrency } = config;
         const { style, currency, minimumFractionDigits, maximumFractionDigits } = dataContext;
         const fiatValue = new BigNumber(value);
+        const currencyForDisplay = currency ?? fiatCurrency;
 
         if (fiatValue.isNaN()) {
             return null;
         }
 
+        let formattedValue: string = '';
```
peter-sanderson: extract into function so we get rid of `let`

suite-common/formatters/src/formatters/prepareCryptoAmountFormatter.ts
```diff
@@ -84,10 +89,10 @@ export const prepareCryptoAmountFormatter = (config: FormatterConfig) =>
             if (withSymbol) {
                 const NetworkSymbolFormatter = prepareNetworkSymbolFormatter(config);
 
-                return `${formattedValue} ${NetworkSymbolFormatter.format(symbol!)}`;
+                formattedValue = `${formattedValue} ${NetworkSymbolFormatter.format(symbol!)}`;
```
peter-sanderson: refactor to get rid of `let` :pray:

packages/suite/src/views/wallet/transactions/components/SummaryCards.tsx
peter-sanderson: we have same issue with ElevationProvider

packages/suite/src/components/suite/HiddenPlaceholder.tsx
```diff
@@ -54,15 +82,32 @@ export const HiddenPlaceholder = ({
         setAutomaticIntensity(fontSize / 5);
     }, []);
 
+    // we only need to handle the case when revealed content is smaller than redacted, not vice versa.
+    // in such case, onMouseEnter shrinks content, and it may immediately onMouseLeave even when cursor is still.
+    const onMouseEnter = () => {
+        setIsHovered(true);
+        if (ref?.current) setWrapperMinWidth(ref.current.getBoundingClientRect().width);
```
peter-sanderson: pls, pls, pls `{}` :pray:

peter-sanderson (approved):

packages/suite/src/components/suite/HiddenPlaceholder.tsx
Lemonexe: aa3a8a060863ec493641ce84336c0e7ea679dc23

suite-common/formatters/src/formatters/prepareCryptoAmountFormatter.ts
Lemonexe: 34332c75ffd936149b0dbd00adca39ab3b7e7403
This one was quite something :sweat_smile:

peter-sanderson (approved):

Lemonexe:
@tomasklim 

> 1. sometimes I can see `###` for a while

Yes, this is expected, seems to me it's still worth it. Though I will check with product. If this is a big issue, I'll work on it.

EDIT: as per @sime it's minor issue, I'll do it in subsequent PR.

> 2. would it be possible to set width of placeholder to the width of original text?

Hmm, what you are showing is a minor visual glitch :thinking: I am only setting width of uncovered to that of redacted, which is possible, because it happens during lifecycle of component, so I can measure the DOM width as I do. Doing it vice versa cannot be reliable, because the component may not have rendered uncovered text during its lifecycle.
I'll discuss with product if this is an issue, before merging. 

EDIT: this specific case solved easily by [setting specific width of the column](https://github.com/trezor/trezor-suite/pull/14528/files#diff-eceb919d8b20092b921e549fd4bed2c299874374e41126cf2ba48953fec79439) (which was already there for the first col). Though it is a tricky thing, and it might bite us again at another time another place.

> 3. would it be possible to increase a `z-index` of those blurred texts so that the shadow is not "cut"?

This issue is preexisting, I haven't touched it :smile: but really, I am afraid that z-index could cause much more issues than this. I'd have to investigate, if we proplery use isolation: isolate throughout the app, to prevent z-index wars. Anyway I'll postpone it, as this is not caused by this PR.

Issue #14467, opened 2024-09-23
Build app from PR


matejkriz:
- Add possibility to run the mobile app from PR easily for quick review and testing
- Rebuild native part only if necessary to optimize EAS costs

matejkriz:
Note for QA: Please make sure you are able to use it and check this doc: https://www.notion.so/satoshilabs/Trezor-Suite-Lite-Preview-12adc526060680628b84d92dc8bf411a?pvs=4

STew790:
**QA OK**
Works like a charm :tada: I was able to check the branch via the QR code.

PR #14530, opened 2024-09-25
Add mobile preview builds

matejkriz:
## Description

This is a first step, adding preview builds from PRs and develop branch. 
Next step will be to stop creating develop builds to actually save some build time, but I would like to see that this is working fine first before breaking the old process. 

## Related Issue

Resolve partially https://github.com/trezor/trezor-suite/issues/14467


matejkriz:
@SocketSecurity ignore npm/expo-updates@0.25.24

suite-native/app/.fingerprintignore
```diff
@@ -0,0 +1,3 @@
+ios/**/*
```
Nodonisko, line 1: Not sure about this

Nodonisko (approved):

suite-native/app/.fingerprintignore
matejkriz: neither do I, but I hope that any change in ios folder should be caused by some other not-ignored change like, dependency or config update

Nodonisko: Yea, all source of truth for iOS folder should be dependencies and expo config file so it should be fine.

.github/workflows/release-suite-native-preview-update.yml
```diff
@@ -0,0 +1,49 @@
...
+    paths:
+      - "suite-native/**"
+      - "suite-common/**"
+  workflow_dispatch:
+
+  jobs:
+    update:
+      if: github.repository == 'trezor/trezor-suite' || github.repository == 'trezor/trezor-suite-private'
+    name: EAS Update
+    environment: develop-suite-native
```
matejkriz: I guess this is not correct @vdovhanych?

.github/workflows/release-suite-native-preview-update.yml
```diff
@@ -0,0 +1,49 @@
+name: "[Release] suite-native preview update"
+
+on:
+  push:
```
PeKne: Do we want to trigger this workflow on push to develop? It seems that changes from https://github.com/trezor/trezor-suite/pull/14531 are already taking care of doing develop preview builds. I would remove it to make sure that there is no extra build created from here.

.github/workflows/release-suite-native-preview-update.yml
```diff
@@ -0,0 +1,49 @@
...
+      - "suite-common/**"
+  workflow_dispatch:
+
+  jobs:
+    update:
+      if: github.repository == 'trezor/trezor-suite' || github.repository == 'trezor/trezor-suite-private'
+    name: EAS Update
+    environment: develop-suite-native
+    runs-on: ubuntu-latest
+    concurrency: fingerprint-${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
PeKne: What about adding a `cancel-in-progress` argument to concurrency runs? So we can save some resources? Just an idea. Quite often someone pushes to the PR a few times in a short period. This parameter would ensure that the job won't run for every push.

PeKne (approved):
Awesome, can't wait to test it 🤩

.github/workflows/release-suite-native-preview-update.yml
vdovhanych: The environment is only available from protected branches. I can create an environment that is accessible from all branches, but for that, let's do a different org in the expo so we do not expose the API key to all of the branches.

.github/workflows/release-suite-native-preview-update.yml
vdovhanych: agree

.github/workflows/release-suite-native-preview-update.yml
```diff
@@ -0,0 +1,49 @@
...
+    env:
+      SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
```
vdovhanych: Are we sure we want to have sentry for the preview app builds?

.github/workflows/release-suite-native-preview-update.yml
```diff
@@ -0,0 +1,49 @@
+name: "[Release] suite-native preview update"
```
vdovhanych: I'm not sure I like the name of this, mainly release, as it is more like a build test app rather than release.

matejkriz: It is the same as it was with [Release] suite-native develop, but let's change it to build and keep release just for the public release.

.github/workflows/release-suite-native-preview-update.yml
matejkriz: The next PR #14531 is building different profile (develop vs preview) so it's different build

.github/workflows/release-suite-native-preview-update.yml
matejkriz: yes, there will be not much events from preview builds so it will not drain our quota, but it's usefull to see the errors early and be able to test changes related to sentry quicly

.github/workflows/release-suite-native-preview-update.yml
matejkriz: I would be afraid that if we skip some run, we can unintentionally skip the native build. In case the job run already updated the fingerprint in the database before cancelation. But at the same time the native build should happen only on that change, so it's not that much resources wasted. We will see, we can change it later, but for now I would prefer this safer option.

.github/workflows/release-suite-native-preview-update.yml
PeKne: I understand. But why we want to do a `preview` build if there is a push to develop? Then should be the `develop` build enough, right?

matejkriz: No, it's distributed different way and now it's even different app (see latest fixups)

.github/workflows/release-suite-native-preview-update.yml
matejkriz: https://github.com/trezor/trezor-suite/pull/14530/commits/087a775cc2295874b616cd4d2ae8d79ead1f542a

Nodonisko (approved):
lgtm

vdovhanych (approved):

.github/workflows/build-suite-native-preview.yml
```diff
@@ -0,0 +1,59 @@
...
+  push:
+    # push develop(default) branch is necessary for this action to update its fingerprint database
+    branches: [develop]
+  pull_request:
+    paths:
+      - "suite-native/**"
+      - "packages/react-native-usb/**"
+      - "packages/transport-native/**"
+      - "yarn.lock"
+      # list of paths is not complete, but it's always possible to dispatch manually
```
Nodonisko, line 13: I would suggest add suite-common/** so people from other teams could easily test changes that can affect mobile app

.github/workflows/build-suite-native-preview.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Install libs
+        run: yarn workspaces focus @suite-native/app
+
+      - name: Create preview builds if fingerprint changed
+        uses: expo/expo-github-action/preview-build@main
+        with:
+          command: eas build --profile preview --platform all
+          working-directory: suite-native/app
+
+      - name: Install deps and build libs
```
Nodonisko: Why are you calling `yarn install` again here? `yarn workspaces focus @suite-native/app` above should allready have installed all dependendecies needed

suite-native/app/app.config.ts
```diff
@@ -67,16 +67,6 @@ const projectIds = {
 } as const satisfies Record<BuildType, string>;
 
 const buildType = (process.env.EXPO_PUBLIC_ENVIRONMENT as BuildType) ?? 'debug';
-const isCI = process.env.CI == 'true' || process.env.CI == '1';
-
```
Nodonisko, line 71: what's wrong with this? it already saved out ass once or twice I think :D

Nodonisko (commented):
few nits, otherwise looks good

suite-native/app/app.config.ts
matejkriz: I've tried to explain it in commit message - Detox build was broken because expo-updates is evaluating app.config.ts even though it's not necessary for Dextox build. I belive that if those ENV variables are missing, we will find that soon anyway? https://github.com/trezor/trezor-suite/actions/runs/11483547009/job/31959232440?pr=14530#step:10:3143

.github/workflows/build-suite-native-preview.yml
matejkriz: Good point. I'll added, but maybe in next PR. I found out that currently it does the build on every new push even if the runtimeVersion is not changed, so I have to fix that first.

.github/workflows/build-suite-native-preview.yml
matejkriz: 👍 I wasn't sure all the dependencies for message system config are there, but it works fine, so removed, thanks for noticing!

PR #14531, opened 2024-09-25
chore(suite-native): build develop only on native change

### LINKED ISSUE #14467: Build app from PR

matejkriz:
## Description

Do develop native build only if necessary to save builds on EAS. Keep it distributed via Firebase/Testflight so we don't need to register every ios device.

## Related Issue

Followup of https://github.com/trezor/trezor-suite/pull/14530
Resolve https://github.com/trezor/trezor-suite/issues/14467

### DIFF
--- a/.github/workflows/release-suite-native-develop.yml
+++ b/.github/workflows/release-suite-native-develop.yml
@@ -9,16 +9,13 @@ on:
       - "suite-common/**"
   workflow_dispatch:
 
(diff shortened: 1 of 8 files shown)


matejkriz:
no longer relevant

PR #14536, opened 2024-09-26
fix(suite): Fix product-components dep

jvaclavik:
## Related Issue

Resolve


tomasklim (approved):
I think you miss change of `yarn.lock`
Screenshot 2024-09-26 at 7 35 17

PR #14539, opened 2024-09-26
fix(suite-desktop): fix bridge controls in debug menu

### LINKED ISSUE #13587: suite-desktop debug: bridge running switch
1. start suite-desktop with bundled bridge
1. turn off bundled bridge in debug settings
1. try to turn on bundled bridge in debug settings - checkbox does not react
2. transport is however started and TRANSPORT_START event emitted 

image
image

mroz22:
image

fixing bridge controls in debug menu. this feature might be useful for some old T ones edgecase handling - turning on legacy bridge instead of the new bridge in the runtime of application. cc @MiroslavProchazka that we are able to achieve that, eg we detected unreadable device, we give you a button "use old bridge". 

resolve #13587

### DIFF
--- a/packages/suite-desktop-core/src/app.ts
+++ b/packages/suite-desktop-core/src/app.ts
@@ -18,7 +18,7 @@ import { clearAppCache, initUserData } from './libs/user-data';
 import { initSentry } from './libs/sentry';
 import { initModules, mainThreadEmitter } from './modules';
 import { init as initTorModule } from './modules/tor';
-import { init as initBridgeModule } from './modules/bridge';
+import { init as initBridgeModule, initUi as initBridgeUi } from './modules/bridge';
 import { createInterceptor } from './libs/request-interceptor';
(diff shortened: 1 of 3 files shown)


martykan (approved):
Works for me. I think it's okay to split the init code just for bridge, since it's handled in a special way due for autostart.

mroz22:
My main concern is that I might be doing some non-systematic changes in code that you guys, mainly @marekrjpolak won't like. On the other hand if it get its job done, lets move on and talk about improvements later

marekrjpolak:
> mainly @marekrjpolak won't like

I won't pretend that I'm not gonna cry because of this. But let's just stop calling Bridge (and maybe even Tor) modules as they're handled on their own, and I'll be quite happy.

In the future, we may for example separate the Bridge & Tor process **logic** (which will be handled purely in the main) from **control** which can be handled as before (=> it will become module again?). Anyway, this is good for now.

PR #14553, opened 2024-09-26
feat(suite): normal accounts list collapsible of multiple accounts av…

seibei-iguchi:


jvaclavik (approved):

Lemonexe:
Concerning the failing tests: I see, previously the account groups were expanded, now they are folded.
I also see that the arrow elements have `data-testid="@account-menu/arrow"`.
I played with it directly in local cypress, and it seems you can simply get the collection of all these arrows, elem.click(), them, and the test then continues normally, or is there another problem? :thinking:  

EDIT: sorry, I didn't read that properly, `clickAllAccountArrows()` already does that

seibei-iguchi:
> Concerning the failing tests: I see, previously the account groups were expanded, now they are folded. I also see that the arrow elements have `data-testid="@account-menu/arrow"`. I played with it directly in local cypress, and it seems you can simply get the collection of all these arrows, elem.click(), them, and the test then continues normally, or is there another problem? 🤔
> 
> image image

The accounts were always folded, that's why there's the clickAllAccountArrows function call. The issue is the default account was not collapsible before, now it is, but is also expanded by default - that's why it's closed while all the other accounts are opened.

tomasklim (approved):
I love it @adamhavel, please get it into next release 🙏 

Screenshot 2025-01-01 at 13 12 55

bosomt:
**QA OK**

**Info:**
 - Suite version: web 25.1.0 (06bb732f9ea451f6e9dfdf88a7220b8adc64d398)
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.7 regular (revision e3183e58f35f6bd6b98fc9262434abb773f6ae9a)
 - Transport: WebUsbTransport

PR #14564, opened 2024-09-26
refactor(suite): deduplicate common translation formatters and placeholders

ghost:
## Description

Move  common formatters and recurring placeholder resolvers from props to Translation component itself.


MiroslavProchazka:
Hello @mempoolboy ,

Thank you for your contribution!
I've asked my colleagues to do a code review. Once this will be done, I will create a separate PR from your source code in our repository to run the pipelines. Then we can merge this.

Thanks

komret:
Good idea, saves some lines, but it is a bit magicky, not sure if worth it. I'll let @jvaclavik decide.

tomasklim:
I like it but it needs documentation

PR #14568, opened 2024-09-27
Run tests with node bridge

mroz22:
node-bridge should be available in trezor-user-env since https://github.com/trezor/trezor-user-env/pull/261/files
but it still doesn't work, something is missing and I am not sure yet what it is. It can't see device yet. 
image

image

one of the problems might be missing ui files for the status page. 
image

The end game goal is to start using it in our tests instead of trezord-go.

### DIFF
--- a/packages/connect/e2e/common.setup.ts
+++ b/packages/connect/e2e/common.setup.ts
@@ -78,7 +78,7 @@ export const setup = async (
     TrezorUserEnvLink.state = options;
 
     // after all is done, start bridge again
-    await TrezorUserEnvLink.startBridge();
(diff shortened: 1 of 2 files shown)


grdddj:
Could be solved by https://github.com/trezor/trezor-user-env/pull/265 - I will build a testing image and share it here, to be tested

The testing image: ghcr.io/trezor/trezor-user-env:9f7d122fdf4f4c2bfc298bac8e2cc349b5448cd7@sha256:442827ad49462b1e2ae0cac900f9d5f2911dcb823edebfadb14d3f1a4745878a

PR #14577, opened 2024-09-27
fix(connect): replace import for @sinclair/typebox due to TS issue

martykan:
## Description

There is a prevailing issue with the imports for `@sinclair/typebox` in TypeScript declaration files that are published in the build library output, which is causing issues with using the library by 3rd parties.  

It's importing `@sinclair/typebox/build/esm/index.mjs`, however this doesn't work properly since our published package is CommonJS, not ESM. 

This PR adds a patch via the `replace-imports.sh` script, which replaces `@sinclair/typebox/build/esm/index.mjs` -> `@sinclair/typebox`.

### DIFF
--- a/scripts/replace-imports.sh
+++ b/scripts/replace-imports.sh
@@ -24,3 +24,14 @@ if grep -Rl "$SEARCH_PATTERN" "$1"; then
 else
     echo "All occurrences of '@trezor/*/src' have been successfully replaced."
 fi
+
+# Patch for Typebox import issue, where TS uses an ESM import path, but our package is CommonJS
+# @sinclair/typebox/build/esm/index.mjs -> @sinclair/typebox
+
+REGEX="s/@sinclair\/typebox\/build\/esm\/index.mjs/@sinclair\/typebox/g"
+
+if [[ "$OS" == "Darwin" ]]; then
+    find "$1" -type f -exec sed -i '' -E "$REGEX" {} +
+else
+    find "$1" -type f -exec sed -i -E "$REGEX" {} +
+fi


Nodonisko (approved):
I think this is not right solution but if you guys are fine with it I have no issue with that :D

martykan:
I agree it would be best to find the root cause, but right now we have a broken release and need a reliable fix ☹️

mroz22:
# :shipit:

zxpectre:
Hi! there, I'm the dev behind GameChanger Wallet which is a web app and has Trezor support since years, there is a hard fork taking place soon on Cardano and I've been trying to import the new mandatory version since months now without full success. 

Last working version was `@trezor/connect-web@9.0.11`

Dep tree:
```
└─┬ @trezor/connect-web@9.4.7
  └─┬ @trezor/connect@9.4.7
    └─┬ @trezor/schema-utils@1.2.3
      └── @sinclair/typebox@0.33.22 
```

Time is running up for us :/

How a loader with `customize-cra ` should be implemented to support the new releases?

A loader like this
```js
const typeboxFixPlugin =config => {
  config.module.rules.push({
    include: /node_modules/,
    test: /\.mjs$/,
    type: 'javascript/auto'
  });
  return config;
}
```

fails with 

```

./node_modules/@sinclair/typebox/build/cjs/type/mapped/mapped.js 107:663
Module parse failed: Unexpected token (107:663)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|     (0, kind_1.IsOptional)(T) ? (0, index_12.Optional)(FromSchemaType(K, (0, index_2.Discard)(T, [index_1.OptionalKind]))) : (0, kind_1.IsReadonly)(T) ? (0, index_14.Readonly)(FromSchemaType(K, (0, index_2.Discard)(T, [index_1.ReadonlyKind]))) : // unevaluated mapped types
|     (0, kind_1.IsMappedResult)(T) ? FromMappedResult(K, T.properties) : (0, kind_1.IsMappedKey)(T) ? FromMappedKey(K, T.keys) : // unevaluated types
>     (0, kind_1.IsConstructor)(T) ? (0, index_5.Constructor)(FromRest(K, T.parameters), FromSchemaType(K, T.returns), options) : (0, kind_1.IsFunction)(T) ? (0, index_6.Function)(FromRest(K, T.parameters), FromSchemaType(K, T.returns), options) : (0, kind_1.IsAsyncIterator)(T) ? (0, index_4.AsyncIterator)(FromSchemaType(K, T.items), options) : (0, kind_1.IsIterator)(T) ? (0, index_9.Iterator)(FromSchemaType(K, T.items), options) : (0, kind_1.IsIntersect)(T) ? (0, index_8.Intersect)(FromRest(K, T.allOf), options) : (0, kind_1.IsUnion)(T) ? (0, index_16.Union)(FromRest(K, T.anyOf), options) : (0, kind_1.IsTuple)(T) ? (0, index_15.Tuple)(FromRest(K, T.items ?? []), options) : (0, kind_1.IsObject)(T) ? (0, index_11.Object)(FromProperties(K, T.properties), options) : (0, kind_1.IsArray)(T) ? (0, index_3.Array)(FromSchemaType(K, T.items), options) : (0, kind_1.IsPromise)(T) ? (0, index_13.Promise)(FromSchemaType(K, T.item), options) : T
|   );
| } // prettier-ignore

```

and transpilng with `babel-loader`  do not crash the build but the app crashes with terrible recursive loops on the transpiled typebox code.

---

### UPDATE:

I've manage to spot the exact version gap that introduced these issues:

@trezor/connect-web@9.4.0 ---> @trezor/connect-web@9.4.1 

fail us with

```
Failed to compile.

./node_modules/@sinclair/typebox/build/esm/type/indexed/indexed-from-mapped-key.mjs
Can't import the named export 'Index' from non EcmaScript module (only default export is available)
```

I think the webpack updates you applied on those commits introduced the problem somehow.

zxpectre:
I'm trying to adapt our code to use https://connect.trezor.io/9.4.0/trezor-connect.js but issue is for Cardano there are so many types and constants definitions outside that bundle that I'm forced to load types and constants using @trezor/connect-web and app breaks either at build time or runtime (`Uncaught TypeError: Cannot convert a BigInt value to a number`).

Can you at least publish @types/@trezor/connect-web so I can try to import that instead of the conflictive packages?

martykan:
@zxpectre I looked at your repo https://github.com/zxpectre/react-csl as a base to replicate this issue. 
I installed and used `@trezor/connect-web`, without additional loaders.

I did see an issue with trying to import ESM, but only when running `npm start`. 
With `npm run build` everything was working fine. Is that the case for you as well? 
Which leads me to believe it's some issue with the tooling.
I tried updating to `react-scripts": "^5.0.1"` and both started working, there were just some warnings about missing source maps.

zxpectre:
Hi @martykan , thanks, I'm currently spending the day trying to update the tooling while correcting all the breaking changes on the project, which is not good on this short time frame. So far I'm introducing more problems with this approach.

Any other ideas regarding tweaking the loader?

As reported, on some versions npm run build works, but app fails on runtime, on other versions npm run build fails.

martykan:
I looked at `babel-loader` and the issue with `Module parse failed: Unexpected token` seems to be caused by the null-coalescing (`??`) operator, so [babel-plugin-transform-nullish-coalescing-operator](https://babeljs.io/docs/babel-plugin-transform-nullish-coalescing-operator) might be a place to start, but I wasn't able to get it fully working. 
I would say it's best to keep things simple, so I would not use the loader workarounds and focus on updating the tooling.

PR #14593, opened 2024-09-27
fix: Elevation context wrapping

peter-sanderson:
Fixes the incorrect use of React Context so the elevation value is passed down correctly.

This also fixes the base elevation of modal, as elevation context sets the elevation of the plane the `NewModal` is onto.

image


packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -157,74 +157,78 @@ const NewModalBase = ({
...
+                </ScrollContainer>
+                <ShadowBottom />
+            </ShadowContainer>
+            {bottomContent && <Footer>{bottomContent}</Footer>}
+        </Container>
     );
 };
 
+const NewModalBase = (props: NewModalProps) => (
+    <ElevationContext baseElevation={prevElevation[MODAL_ELEVATION]}>
```
peter-sanderson, line 227: Here the `prevElevation` must be used, as `ElevationContext` sets the level of the plane, on which NewModal sits on.

If we want `NewModal` to be `0` it must sit on plane of elevation `-1`

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -157,74 +159,78 @@ const NewModalBase = ({
...
-                                    variant="tertiary"
-                                    icon="close"
-                                    data-testid="@modal/close-button"
-                                    onClick={onCancel}
-                                    size="small"
-                                />
-                            )}
-                        </Header>
+        <Container
+            $elevation={elevation}
```
peter-sanderson, line 163: `MODAL_ELEVATION` no longer needs to be hard-coded as we can use value from `useElevation` now, because we are in nested component and the Context (ElevationContext) of this component is correctly set by React.

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -131,7 +131,7 @@ type NewModalProps = AllowedFrameProps & {
     'data-testid'?: string;
 };
 
-const NewModalBase = ({
+const _NewModalBase = ({
```
seibei-iguchi, line 134: I really don't like this naming:D There must be something better

seibei-iguchi (approved):

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -157,74 +159,78 @@ const NewModalBase = ({
...
-                                            !!onCancel && !heading && !description && !onBackClick
-                                        }
-                                    >
-                                        <Icon name={icon} size={ICON_SIZE} variant={variant} />
-                                    </IconWrapper>
-                                )}
-                                {children}
-                            </Body>
-                        </ScrollContainer>
-                        <ShadowBottom backgroundColor={modalBackgroundColor} />
```
peter-sanderson, line 219: Problem here is, that `ShadowBottom` component is placed in JSX in the context of `NewModalBase` parent component. 

The fact its wrapped in `<ElevationContext ...` does not set its context to it, but keeps the context of the `NewModalBase`.

We have to create new component function, so the new context is created and the correct `elevation` value is retrieved from it (in this case the value `-1 + 1 = 0` <-- correct.


The reason why it "worked" (kinda), is because `useElevation` has a default behavior and fallbacks to `0` (which resulted in this case in `0 + 1 = 1` <-- incorrect.

In the light of this issue, I kinda think we should **remove** default behavior and throw error if `useElevation` is used without context being set. 

Defaults are evil.

cc @jvaclavik :point_up: 

image

packages/components/src/components/NewModal/NewModal.tsx
peter-sanderson: Feel free to come up with better name / rename it later ;)

bosomt:
**QA OK**


**Info:**
 - Suite version: desktop 24.10.0 (2187dc0a75fc5ffb03e51b661ff61952c5159ebf)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.10.0 Chrome/126.0.6478.234 Electron/31.6.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: BridgeTransport 3.0.0-bundled.24.10.0

PR #14600, opened 2024-09-27
Fix/accounts tab search does not work correctly with special chars

### LINKED ISSUE #11383: Accounts tab - search does not work correctly with special chars
**Describe the bug:**
Search in accounts tab does not correctly filter accounts with special chars `#`

**Info:**

- Suite Version: 24.8.1

**How to reproduce**

- try to filter accounts using `#` or other special characters (you have to enable labeling to be able to add other chars)

**Expected behavior**
It would find my account whatever name it has

**Screenshots**
Screenshot 2024-09-08 at 9 57 14
Screenshot 2024-09-08 at 9 57 04
Screenshot 2024-09-08 at 9 56 48
Screenshot 2024-09-08 at 9 56 42

https://github.com/trezor/trezor-suite/assets/858321/ecb6c2fc-6cfb-49bf-b773-d65834826477

ghost:
Unlabeled accounts are represented by default labels created as translation strings. As these are derived from account data based on user settings and require an intl provider, they are not saved in the store and therefore are not included in the search).

## Description

Include default labels in subject of account search, consolidate and refactor related code.

## Related Issue

Resolve [#11383](https://github.com/trezor/trezor-suite/issues/11383)

## Screenshots:

image
image
image
image

### DIFF
--- a/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/PageNames/AccountName/AccountDetails.tsx
+++ b/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/PageNames/AccountName/AccountDetails.tsx
@@ -11,8 +11,7 @@ import {
     FiatValue,
     AmountUnitSwitchWrapper,
 } from 'src/components/suite';
-import { useAccountLabel } from 'src/components/suite/AccountLabel';
-import { useSelector } from 'src/hooks/suite';
+import { useDefaultAccountLabel, useSelector } from 'src/hooks/suite';
 import { selectLabelingDataForSelectedAccount } from 'src/reducers/suite/metadataReducer';
 import { CoinLogo } from '@trezor/product-components';
 
@@ -83,7 +82,7 @@ export const AccountDetails = ({ selectedAccount, isBalanceShown }: AccountDetai
     const [shouldAnimate, setShouldAnimate] = useState(false);
     const [hasMounted, setHasMounted] = useState(false);
     const selectedAccountLabels = useSelector(selectLabelingDataForSelectedAccount);
-    const { defaultAccountLabelString } = useAccountLabel();
+    const { getDefaultAccountLabel } = useDefaultAccountLabel();
     const { symbol, key, path, index, accountType, formattedBalance } = selectedAccount;
 
     useEffect(() => {
@@ -121,7 +120,7 @@ export const AccountDetails = ({ selectedAccount, isBalanceShown }: AccountDetai
                         defaultValue: path,
                         value: selectedAccountLabels.accountLabel,
                     }}
-                    defaultEditableValue={defaultAccountLabelString({ accountType, symbol, index })}
+                    defaultEditableValue={getDefaultAccountLabel({ accountType, symbol, index })}
                     updateFlag={isBalanceShown}
                 />
             </AccountHeading>
--- a/packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountsList.tsx
+++ b/packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountsList.tsx
@@ -41,8 +41,9 @@ export const AccountsList = ({ onItemClick }: AccountListProps) => {
         searchString || coinFilter
             ? list.filter(account => {
                   const { key, accountType, symbol, index } = account;
-                  const accountLabel =
-                      accountLabels[key] || getDefaultAccountLabel({ accountType, symbol, index });
+                  const accountLabel = accountLabels.hasOwnProperty(key)
+                      ? accountLabels[key]
+                      : getDefaultAccountLabel({ accountType, symbol, index });
 
                   return accountSearchFn(account, searchString, coinFilter, accountLabel);
               })
--- a/packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountsList.tsx
+++ b/packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountsList.tsx
@@ -1,7 +1,7 @@
 import { sortByCoin, getFailedAccounts, accountSearchFn } from '@suite-common/wallet-utils';
 import { Account } from '@suite-common/wallet-types';
 import { useAccountSearch, useDiscovery, useSelector } from 'src/hooks/suite';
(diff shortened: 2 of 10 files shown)


MiroslavProchazka:
Hello @mempoolboy ,

Thank you for your contribution! 
I've asked my colleagues to do a code review. Once this will be done, I will create a separate PR from your source code in our repository to run the pipelines. Then we can merge this.

Thanks

packages/suite/src/hooks/suite/useDefaultAccountLabel.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const getDefaultAccountLabel = useCallback(
+        ({ accountType, symbol, index = 0 }: GetDefaultAccountLabel): string => {
+            if (accountType === 'coinjoin') {
+                return translationString(getTitleForCoinjoinAccount(symbol));
+            }
+
+            return translationString('LABELING_ACCOUNT', {
+                networkName: translationString(getTitleForNetwork(symbol)), // Bitcoin, Ethereum, ...
+                index: index + 1, // This is the number which shows after hash, e.g. Ethereum #3
```
Lemonexe: Nit: I know this line was copied as is from the old function, but we could make it neater :slightly_smiling_face:
I'd extract `index+1` as a variable, could be named for example `displayedAccountNumber` or something, then comment is not needed. Code that explains itself is preferred over commentary, but it's a tiny thing :slightly_smiling_face:

packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/PageNames/AccountName/AccountDetails.tsx
```diff
@@ -82,7 +81,7 @@ export const AccountDetails = ({ selectedAccount, isBalanceShown }: AccountDetai
     const [shouldAnimate, setShouldAnimate] = useState(false);
     const [hasMounted, setHasMounted] = useState(false);
     const selectedAccountLabels = useSelector(selectLabelingDataForSelectedAccount);
-    const { defaultAccountLabelString } = useAccountLabel();
+    const { getDefaultAccountLabel } = useDefaultAccountLabel();
```
Lemonexe, line 85: thanks for renaming the returned callback, the old name didn't make much sense :+1:

packages/suite/src/hooks/suite/useDefaultAccountLabel.ts
```diff
@@ -0,0 +1,32 @@
+import { getTitleForNetwork, getTitleForCoinjoinAccount } from '@suite-common/wallet-utils';
+import { useCallback } from 'react';
+import { useTranslation } from './useTranslation';
+import { AccountType, NetworkSymbol } from '@suite-common/wallet-config';
+
+export interface GetDefaultAccountLabel {
```
Lemonexe: Nit: I think `GetDefaultAccountLabelParams` would be more suitable, because this isn't type of the function, just its params.

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountsList.tsx
```diff
@@ -37,7 +39,13 @@ export const AccountsList = ({ onItemClick }: AccountListProps) => {
     const list = sortByCoin(accounts.filter(a => a.deviceState === device.state).concat(failed));
     const filteredAccounts =
         searchString || coinFilter
-            ? list.filter(a => accountSearchFn(a, searchString, coinFilter, accountLabels[a.key]))
+            ? list.filter(account => {
+                  const { key, accountType, symbol, index } = account;
+                  const accountLabel =
+                      accountLabels[key] || getDefaultAccountLabel({ accountType, symbol, index });
```
Lemonexe: Nit: this is imho more semantic - although it's longer, it makes it clear we're not checking for truthiness, but if the value exists on the record at all. Though this code is also fine :ok_hand: 
```suggestion
                      accountLabels.hasOwnProperty(key) ? accountLabels[key] : getDefaultAccountLabel({ accountType, symbol, index });
```

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountsList.tsx
```diff
@@ -37,7 +39,13 @@ export const AccountsList = ({ onItemClick }: AccountListProps) => {
...
+
+                  return accountSearchFn(account, searchString, coinFilter, accountLabel);
```
Lemonexe, line 48: :thought_balloon:  I'd add a test case for this at [accountUtils.test.ts#L217](https://github.com/trezor/trezor-suite/blob/bd936caf71aaee85cbe4f6579a30069381401021/suite-common/wallet-utils/src/__tests__/accountUtils.test.ts#L217).
Although the `accountSearchFn` function is itself unchanged, its usage has slightly changed. There is now no test case to match substring with _supplied_ label, only whole string.
This will test it, and at the same time it'll help describe what is the fn typically used for :slightly_smiling_face: 
```
expect(accountSearchFn(btcAcc, '#1', undefined, 'Bitcoin #1')).toBe(true);
```

Lemonexe (approved):
I love this improvement, we were missing it, thank you! :tada: 
Codewise looking good, I have only small codestyle nits :slightly_smiling_face: 
Tested locally rebased on develop, and works as expected :heavy_check_mark: 

Note: there is conflict in one file, but it's trivial, so no need to re-review after solving it.

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountsList.tsx
ghost: I really like this solution. 🤩

ghost:
> I love this improvement, we were missing it, thank you! 🎉 Codewise looking good, I have only small codestyle nits 🙂 Tested locally rebased on develop, and works as expected ✔️
> 
> Note: there is conflict in one file, but it's trivial, so no need to re-review after solving it.

Thanks for the thorough CR, it's much needed and appreciated. 😃 

I've rebased the branch, resolved the conflict, addressed the issues and added the test scenario (well, **I**'ve rebased the branch and resolved the conflict 😄).

Lemonexe:
Revived as #15035

PR #14608, opened 2024-09-29
T3W1 support in Suite

MiroslavProchazka:
Added support for T3W1

- device authenticity check does not work, you have to turn it off in settings
- added basic support for new device
- added placeholder images of TS7
- I've used some nasty values to fields where we do not have any data at this moment (`FW version` "2.99.99", `pubkeys` "you shall not pass", etc.)


packages/protobuf/src/messages-schema.ts
```diff
@@ -6,6 +6,7 @@ export enum DeviceModelInternal {
     T2B1 = 'T2B1',
     T3B1 = 'T3B1',
```
Lemonexe, line 7: I ran `yarn update-protobuf` locally to see if it is in sync.
- The addition of T3W1 is in sync (no diffs against _this PR_) :heavy_check_mark: 
- But it removed T3B1 :warning: 

:point_right: it's not related to this PR but someone should fix T3B1 in firmware repo

mroz22 (approved):
Idon't see any problem here

packages/protobuf/src/messages-schema.ts
komret: Not happening anymore, at least not on my machine :)

komret:
@mroz, please recheck the connect part and protobuf.
@matejkriz, please check the mobile part.
@Lemonexe, please check 9fe230cbf61bd59a7f27431b9e67038bda234d96

I lowered minimum firmware version from concrete values to 1.0.0 or 2.0.0 at some places. My assumption is that while this should have no impact on functionality, it simplifies adding new devices in future, because we usually don't know the exact firmware version that will be released at the time of adding support to Suite.

I added a `// TODO T3W1` comment wherever code is temporary - correct values should be added later when they are known or when they should be made public. All images and icons are temporary, but not marked in code.

packages/connect/src/utils/__tests__/deviceFeaturesUtils.test.ts
```diff
@@ -118,7 +118,7 @@ describe('utils/deviceFeaturesUtils', () => {
         const featT2B1 = {
             major_version: 2,
             minor_version: 6,
-            patch_version: 1,
+            patch_version: 2,
```
mroz22, line 121: 🤔 hmm why is that?

mroz22 (approved):

Lemonexe (approved):
Suite part LGTM :shipit: :rocket:

suite-native/receive/src/components/DevicePaginationButton.tsx
```diff
@@ -15,7 +15,6 @@ type DeviceScreenPaginationProps = {
     onPress: () => void;
 };
 
-const T2B1_BUTTON_COLOR = '#2B2D2B';
```
Nodonisko, line 18: please do not remove this const, it's not clear what's is that value now

Nodonisko (approved):
just one nit for native part

suite-native/receive/src/components/DevicePaginationButton.tsx
komret: I put it back in 7552647f58de52aa1bcd03aa3f568118b034bb8a, but the naming is quite weird, it says T2B1, but is not used for T2B1 🤷

PR #14609, opened 2024-09-30
feat(suite-native): device connection/lock events during send flow

### LINKED ISSUE #14405: Mobile send - Trezor device connection states
Handle  these connection events in context of the send flow:
- connection/disconnection of Trezor device during the send flow
- pin locking

[DESIGN](https://www.figma.com/design/sqKFZAIvk3zERxiWgthK8K/Send?node-id=3202-15202&t=rZPWs70FeVAIyzjH-4).

PeKne:
## Description
- these device states are handled by this PR:
  - disconnected view-only device
  - device disconnection during the on-device review
  - PIN requested before and during the on-device review
- all these mentioned events should also work in combination with passphrase 

## Related Issue

Resolve #14405

### Going to on-device review of a disconnected view-only device

https://github.com/user-attachments/assets/9c7bece9-d7ba-43b6-a2e4-a3c0ee5a9919

### View-only device disconnected during the on device review

https://github.com/user-attachments/assets/54b80baa-249a-4a04-820c-920156c50721

### Not remembered device disconnected during the review

https://github.com/user-attachments/assets/9de79a65-50da-4508-9db3-977f9835dd93

### Going to on device review with PIN locked device

https://github.com/user-attachments/assets/8055f3f1-5725-4c50-a4d5-3fa4cf0ef6fb


suite-common/wallet-core/src/send/sendFormTypes.ts
```diff
@@ -53,6 +53,7 @@ export type ComposeFeeLevelsError = {
 
 export type SignTransactionError = {
     error: 'sign-transaction-failed';
+    connectErrorCode?: string;
```
matejkriz: nitpick: I would prefer just `errorCode`

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -83,9 +90,11 @@ export const useHandleDeviceConnection = () => {
         if (isNoPhysicalDeviceConnected && isOnboardingFinished) {
             // This accidentally gets triggered by finishing onboarding with no device connected,
             // so this prevents from redirect being duplicated.
```
matejkriz: nitpick: This comment is related to isPreviousRouteOnboarding only, what about moving the comment on the line just before isPreviousRouteOnboarding const definition?

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -43,6 +45,10 @@ export const useHandleDeviceConnection = () => {
     const navigation = useNavigation<NavigationProp>();
     const dispatch = useDispatch();
 
+    const isSendStackFocused =
```
matejkriz, line 48: Looks fragile in case of some nested routes, but let's just test it thoroughly (note to myself)

suite-native/navigation/src/types.tsx
```diff
@@ -64,3 +68,7 @@ export type TabsOptions = {
         params?: Record<string, unknown>;
     };
 };
+
+export type NavigateParameters<TParamList extends ParamListBase = ParamListBase> = Parameters<
+    NavigationHelpers<TParamList>['navigate']
+>;
```
matejkriz: It's probably not the most elegant way, I don't want to spend too much time on it but if you use the type of first item in the array, you don't have to wrap it in an array and then destructuring (...onCancelNavigationTarget)...
```suggestion
export type NavigateParameters<TParamList extends ParamListBase = ParamListBase> = Parameters<
    NavigationHelpers<TParamList>['navigate']
>[0];
```

suite-native/module-send/src/components/AddressReviewStepList.tsx
```diff
@@ -36,13 +46,16 @@ type NavigationProps = StackToStackCompositeNavigationProps<
     RootStackParamList
 >;
 
+const CONNECT_DEVICE_DISCONNECTED_ERROR = 'device disconnected during action';
```
matejkriz: You don't want to import @trezor/transport? I guess TRANSPORT_ERROR.DEVICE_DISCONNECTED_DURING_ACTION could be used here, or would be nice to mention DEVICE_DISCONNECTED_DURING_ACTION in comment here  so that when someone changes it, they can find this place too.

suite-native/module-send/src/components/AddressReviewStepList.tsx
```diff
@@ -81,6 +97,48 @@ export const AddressReviewStepList = () => {
             );
 
             if (isRejected(response)) {
+                const connectErrorCode = response.payload?.connectErrorCode;
+                const message = response.payload?.message;
+                // In case that the signing review is interrupted, restart the flow so user can try again.
+                if (
+                    connectErrorCode === 'Failure_PinCancelled' || // User cancelled the pin entry on device
+                    connectErrorCode === 'Method_Cancel' || // User canceled the pin entry in the app UI.
```
matejkriz: Testing with T1, view-only enabled. If I cancel confirmation of receive address on T1, Device locked toast appear and I'm navigated back to review screen. I guess that it can be **Method_Cancel** in both cases - cancel pin entry or cancel the review flow.

matejkriz: Same behaviour on TT and TS3

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -35,6 +36,7 @@ export const useHandleDeviceConnection = () => {
     const isNoPhysicalDeviceConnected = useSelector(selectIsNoPhysicalDeviceConnected);
     const isPortfolioTrackerDevice = useSelector(selectIsPortfolioTrackerDevice);
     const isOnboardingFinished = useSelector(selectIsOnboardingFinished);
+    const isDeviceRemembered = useSelector(selectIsDeviceRemembered);
```
matejkriz, line 39: If I have some other device remembered, and disconnect non-remembered device (imagine standard wallet remembered, passphrase not), it doesn't work as expected. Instead of redirecting to dashboard with warning about disconnected device, I'm redirected to account detail and then after couple of seconds to the dashboard without any warning. My suspicion is, that the condition shouldBlockSendReviewRedirect is not working in that case because isDeviceRemembered returns true if the non-remembered device is disconnected and remembered device is selected instead. But I haven't checked yet.

matejkriz (commented):
Great job, code looks good, only few nitpicks, feel free to not resolve them. I spotted two bugs during testing mentioned in comments (and one unrelated bug on T1).

matejkriz:
RESOLVED ✅

If I go to Send flow with disconnected passhprase device, I'm asked to connect the device before review and once it's connected and unlocked (TT), I see review address screen with Confirm on Trezor bottom badge already (see screenshot)
image
that is not so big deal, but it confused me, because I had to click Next button, device was not waiting for any confirmation at that point.


EDIT: It happened to me twice, but on 3rd attempt I don't see the Confirm on trezor badge at that point 🤷

matejkriz:
RESOLVED ✅

Again passphrase view-only and disconnected device, if I enter wrong passhprase, it redirects me to the account detail without any explanation, warning. In receive flow we have it handled by the "Tha passphrase you've entered is incorrect." warning.
image

matejkriz:
Wouldn't it be better to ask for the device connection at the last step of receive address confirmation? So only after clicking the Next button to the end? Or even better - add extra box informing that connecting and unlocking the device will be needed if we know that device is not ready at that point?

suite-common/wallet-core/src/send/sendFormTypes.ts
PeKne: fixed in faf9766feb

Also Connect typing improved in [a631499](https://github.com/trezor/trezor-suite/pull/14609/commits/a63149905359c4094e0488aefc48a103ae318095). @mroz22 Could you please look at it? 🙏

suite-native/device/src/hooks/useHandleDeviceConnection.ts
PeKne: https://github.com/trezor/trezor-suite/commit/faf9766feb14f843bc6760e496a14cedb8586b25

suite-native/device/src/hooks/useHandleDeviceConnection.ts
PeKne: This variable says if the SendStack is selected in context of RootStack. Do you have an idea how to improve this?

suite-native/navigation/src/types.tsx
PeKne: https://github.com/trezor/trezor-suite/commit/faf9766feb14f843bc6760e496a14cedb8586b25

suite-native/module-send/src/components/AddressReviewStepList.tsx
PeKne: https://github.com/trezor/trezor-suite/commit/faf9766feb14f843bc6760e496a14cedb8586b25 TRANSPORT_ERROR imported

suite-native/module-send/src/components/AddressReviewStepList.tsx
PeKne: https://github.com/trezor/trezor-suite/commit/faf9766feb14f843bc6760e496a14cedb8586b25 Fixed by removing the toast. There is now way how to distinguish between device being locked and on device review cancellation. Both return `{code: 'Failure_ActionCancelled', error: 'Cancelled}`.

The toast was not in the design file anyway, so I believe it is not a big deal.

suite-native/device/src/hooks/useHandleDeviceConnection.ts
PeKne: I was unable to reproduce that :/

PeKne:
@matejkriz Since the handling of review errors became quite extensive, I decided to move its logic to separate hook `useHandleSendReviewFailure`. All the mentioned passphrase bugs should be resolved as well.  [faf9766](https://github.com/trezor/trezor-suite/pull/14609/commits/faf9766feb14f843bc6760e496a14cedb8586b25)

suite-native/device/src/hooks/useHandleDeviceConnection.ts
matejkriz: No idea, and it seems to work fine, I just wanted to focus on it during testing. Resolving...

suite-native/device/src/hooks/useHandleDeviceConnection.ts
matejkriz: Check the video, I can reproduce it even without any remembered device, but I guess my description wasn't clear at all...
https://github.com/user-attachments/assets/cce8a55d-fed3-4a98-b869-f5972998a36b

matejkriz: I'm disconnecting (unplugging the cable) after address confirmations.

suite-native/module-send/src/components/AddressReviewStepList.tsx
matejkriz: Looks much better, thanks

suite-native/device/src/hooks/useHandleDeviceConnection.ts
PeKne: Now I got it! The device disconnection error is possibly also TRANSPORT_ERROR.UNEXPECTED_ERROR. Handling of this error type added.

Also, the warning alert is sometimes displayed before the navigation takes place. So when the navigation happens later, the alert is hidden. An extra timeout was added to display the alert.

Both are fixed by 45d6143b98.

matejkriz: Not a fan of timeouts but we can live with that here 👌

matejkriz:
I've seen one bug twice, but can't reproduce now, @STew790 please double check during testing.

- Open app with remembered wallet, Trezor disconnected
- Go to Send, fill the form, go to review
- I'm asked to connect the Trezor
- Connect and unlock trezor (tested with TT)
It was supposted to continue in the review flow, and usually it does, but I was **redirected back to the dashboard** twice.

matejkriz (approved):
Great job, no more bugs that I can reliably reproduce 😉 And the happy path is really nice.

PR #14621, opened 2024-09-30
feat: connect deeplink versioning

karliatto:
## Description

Replacing the connect setting `deeplinkUrl` with `connectSrc` + semantic version. This would allow us to make breaking changes in the URL that we can version because it will get always updated with different connect versions.

### DIFF
--- a/packages/connect-examples/deeplink-expo/App.tsx
+++ b/packages/connect-examples/deeplink-expo/App.tsx
@@ -29,13 +29,13 @@ export const App = () => {
                 email: 'developer@xyz.com',
                 appUrl: 'http://your.application.com',
             },
+            connectSrc: 'https://dev.suite.sldev.cz/connect/develop/',
             deeplinkOpen: url => {
                 // eslint-disable-next-line no-console
                 console.log('deeplinkOpen', url);
                 Linking.openURL(url);
             },
             deeplinkCallbackUrl: Linking.createURL('/connect'),
-            // deeplinkUrl: 'https://dev.suite.sldev.cz/connect/develop/deeplink/',
         });
     };
 
--- a/suite-native/module-connect-popup/src/hooks/useConnectParseParams.ts
+++ b/suite-native/module-connect-popup/src/hooks/useConnectParseParams.ts
@@ -3,10 +3,27 @@ import { useMemo } from 'react';
 import { ParsedURL } from 'expo-linking';
 
 import TrezorConnect from '@trezor/connect';
+import { DEEPLINK_VERSION } from '@trezor/connect/src/data/version';
+import { TypedError, TrezorError } from '@trezor/connect/src/constants/errors';
 
-export const useConnectParseParams = (url: ParsedURL) => {
-    const popupOptions = useMemo(() => {
-        const { queryParams } = url;
+type UseConnectParseParamsType =
+    | {
+          popupOptions: {
+              method: string;
+              params: any;
+              callback: string;
+          };
+          parseParamsError: undefined;
+      }
+    | {
+          popupOptions: undefined;
+          parseParamsError: TrezorError;
+      };
+
+export const useConnectParseParams = (url: ParsedURL): UseConnectParseParamsType => {
+    return useMemo(() => {
+        const { queryParams, path } = url;
+        const version = path && path.split('/').slice(-2, -1)[0];
         if (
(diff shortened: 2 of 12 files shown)


packages/connect-examples/deeplink-expo/App.tsx
```diff
@@ -35,6 +35,8 @@ export const App = () => {
                 Linking.openURL(url);
             },
             deeplinkCallbackUrl: Linking.createURL('/connect'),
+            // TODO: what is the use case of 3rd party implementor setting the deep link URL?
+            // Shouldn't it be something just set in connect-deeplink package?
             // deeplinkUrl: 'https://dev.suite.sldev.cz/connect/develop/deeplink/',
```
karliatto: We need a way to tell connect-deeplink to use `https://dev.suite.sldev.cz` instead of `https://connect.trezor.io/` so it uses the development app when we are developing. 

But I think we could set it in `packages/connect-deeplink/src/data/version.ts`, and in the Connect API add something like `_use_DevDeeplinkUrl`. 

@martykan WDYT?

martykan: What if we just used `${connectSrc}/deeplink/`
I think that would be good for consistency

karliatto: Let's do that https://github.com/trezor/trezor-suite/pull/14621/files#diff-eb7311883974581223f67cd18ddccac5c7a0efd61c687773bf6d106fa4c60ea5R167

martykan:
I switched to a separate version counter just for the deeplink protocol, instead of just using the Connect version. 
This way the app can actually differentiate if there was a breaking change to the protocol or it's just a newer version of Connect with unrelated changes. In case of receiving a newer protocol than supported it will display an error. Older protocols can be handled in a backward compatible way or eventually deprecated.

We also want to expose a fallback HTML page, and deploying that for each Connect version would be impractical. This will expose the page for the current version and also the past ones: 
https://dev.suite.sldev.cz/connect/feat/versioning-deeplink/deeplink/1/

suite-native/module-connect-popup/src/hooks/useConnectParseParams.ts
```diff
@@ -3,10 +3,31 @@ import { useMemo } from 'react';
 import { ParsedURL } from 'expo-linking';
 
 import TrezorConnect from '@trezor/connect';
+import { DEEPLINK_VERSION } from '@trezor/connect/src/data/version';
 
-export const useConnectParseParams = (url: ParsedURL) => {
-    const popupOptions = useMemo(() => {
-        const { queryParams } = url;
+export enum ParseParamsErrors {
+    INVALID_PARAMS = 'invalid_params',
```
karliatto: We probably could user here 

```typescript
throw ERRORS.TypedError( 'Method_InvalidParameter');
```

suite-native/module-connect-popup/src/hooks/useConnectParseParams.ts
```diff
@@ -3,10 +3,31 @@ import { useMemo } from 'react';
...
+    VERSION_UNSUPPORTED = 'version_unsupported',
```
karliatto: I don't see any error related to version in connect errors, but maybe we could add it there https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/constants/errors.ts#L0-L1 ?

suite-native/module-connect-popup/src/hooks/useConnectParseParams.ts
martykan: True but I can't throw here in `useMemo`

karliatto: Sure, but maybe just using `ERRORS.TypedError( 'Method_InvalidParameter')` ?

martykan: Ok like this? 58e1fb6bd845f552f57fbf4296ddf58974e4ec02

karliatto: Yes, that looks awesome.

karliatto (commented):
Looks good to me. :+1:

martykan (approved):

PR #14622, opened 2024-09-30
Build node bridge in CI

grdddj:
Tries to build an exportable artifact with bin.js file containing node-bridge code

To be used by `tenv` - taking the latest artefact from `develop`

### DIFF
--- a/.github/workflows/build-node-bridge.yml
+++ b/.github/workflows/build-node-bridge.yml
@@ -1,0 +1,66 @@
+name: "[Build] suite-node-bridge-bin-js"
+
+permissions:
+  id-token: write # for fetching the OIDC token
+  contents: read # for actions/checkout
+
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    types: [labeled]
+  workflow_dispatch:
+
+jobs:
+  node-bridge-bin-js:
+    name: Build node-bridge bin.js file
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          lfs: true
+          submodules: true
+
+      - name: Install node and yarn
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: ".nvmrc"
+
+      - name: Install deps and build libs
+        run: |
+          yarn install --immutable
+
+      - name: Build bin.js file
+        run: |
+          yarn workspace @trezor/transport-bridge build:js
+
+      - name: Upload node-bridge artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: node_bridge_bin_js
+          path: |
+            packages/transport-bridge/dist/bin.js
+          retention-days: 3
+
+      - name: Configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::538326561891:role/gh_actions_trezor_suite_dev_deploy
+          aws-region: eu-central-1
+
+      - name: Extract branch name
+        id: extract_branch
+        run: |
+          echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+
+      - name: Upload dist to dev server
+        shell: bash
+        env:
+          DEPLOY_PATH: s3://dev.suite.sldev.cz/transport-bridge/${{ steps.extract_branch.outputs.branch }}/dist
+        run: |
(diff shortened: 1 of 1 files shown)


grdddj:
Seems to work fine already - see https://github.com/trezor/trezor-suite/actions/runs/11105572536

Consideration - instead (in addition) to saving this as GH-Actions artifacts, it might be beneficial to save it on `dev.trezor.io` as well ... one benefit is that we would not need ani GH API keys to get it nightly into `tenv`, and it would be just one non-changing URL
- there are already some CI jobs in `Suite` uploading stuff there, I guess we would just need to agree on (and create?) a specific folder (URL) for it
- in that case, we would change it to run only on `develop` branch probably, so it is not being rewritten by other branches

.github/workflows/build-node-bridge.yml
```diff
@@ -0,0 +1,49 @@
+name: "[Build] suite-node-bridge-bin-js"
```
mroz22, line 1: if you search for 

```
yarn workspace @trezor/transport-bridge build:lib
```
there are a couple of occurrences. and this very script actually calls 
```
yarn workspace @trezor/transport-bridge build:js
```

not sure how good idea this is but it could be feasible just to reuse artifacts from another job and spare us one yml config file. for example https://github.com/trezor/trezor-suite/blob/develop/.github/workflows/test-suite-desktop-e2e.yml#L1 could work.

mroz22:
I am not familiar with `dev.trezor.io` but we are using `dev.suite.sldev.cz` heavily. For example, recently we added a webusb testere utility here https://dev.suite.sldev.cz/transport-test/develop/.

I am totally ok for using it for this artifact. It should be uploaded probably only for develop branch so that it doesn't bloat too much.

.github/workflows/build-node-bridge.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::538326561891:role/gh_actions_trezor_suite_dev_deploy
+          aws-region: eu-central-1
+
+      - name: Upload dist to dev server
+        shell: bash
+        env:
+          DEPLOY_PATH: s3://dev.suite.sldev.cz/transport-bridge/dist
```
mroz22: I am ok with it but maybe, for consistency purposes - shouldn't it contain branch name here? consider other jobs

.github/workflows/build-node-bridge.yml
```diff
@@ -0,0 +1,62 @@
...
+          submodules: true
+
+      - name: Install node and yarn
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: ".nvmrc"
+
+      - name: Install deps and build libs
+        run: |
+          yarn install --immutable
```
mroz22, line 33: is this full yarn install needed? wouldn't `yarn focus` to transport-bridge folder suffice?

.github/workflows/build-node-bridge.yml
grdddj: I have deleted the branch name on purpose, as it will only run in develop branch ... but will include it again

.github/workflows/build-node-bridge.yml
grdddj: No idea at all here, but will try it

mroz22: this way it installs all the dependencies. In frequently run jobs, it makes sense to optimize it using yarn focus. here it is less relevant but still a good practice to follow if possible

mroz22 (approved):

.github/workflows/build-node-bridge.yml
```diff
@@ -0,0 +1,66 @@
...
+permissions:
+  id-token: write # for fetching the OIDC token
+  contents: read # for actions/checkout
+
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    types: [labeled]
```
vdovhanych, line 12: This makes no sense; it will build a node bridge for any PR that you add a label to. Why??

grdddj: Because I do not know workflow syntax well and I copied it like this from some other .yml file. Agreed that it is redundant, I will remove it.

Do you have some other possible improvements, when I will be already making another PR with changes?

vdovhanych: This was the only thing that i noticed

grdddj: Thanks for that. I created a PR for it - https://github.com/trezor/trezor-suite/pull/14742

PR #14625, opened 2024-09-30
feat(suite): change assets table

### LINKED ISSUE #13152: EVM Tokens / Generic table structure

enjojoy:
## Description

Changed the cards grid to the new Table component

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13152

## Screenshots:

image


packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetTableHeader.tsx
```diff
@@ -1,43 +1,25 @@
...
-            <Translation id="TR_EXCHANGE_RATE" />
-        </HeaderRight>
-        <Header>
-            <Translation id="TR_7D_CHANGE" />
-        </Header>
-        <Header /> {/* Buy Button */}
-    </AssetTableRowGrid>
+    <Table.Header>
+        <Table.Row>
+            <Table.Cell>{/* Logo */}</Table.Cell>
```
seibei-iguchi: I'd put the asset logo in the same cell like the name, like in the tokens table.

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetTableHeader.tsx
```diff
@@ -1,43 +1,25 @@
...
+            <Table.Cell>
+                <Translation id="TR_VALUES" />
+            </Table.Cell>
+            <Table.Cell>
+                <Translation id="TR_EXCHANGE_RATE" />
+            </Table.Cell>
+            <Table.Cell>
+                <Translation id="TR_7D_CHANGE" />
+            </Table.Cell>
+            <Table.Cell>{' ' /* Buy Button */}</Table.Cell>
```
seibei-iguchi: I'd delete these and just add colspan={3} to the previous cell.

tomasklim (changes requested):
The table should behave the same as before 😢 

Before:

https://github.com/user-attachments/assets/a0beea26-e189-4580-8e52-555cbd835a4f

Now:

https://github.com/user-attachments/assets/c354e255-bfce-4691-939e-3f3495458a22

+
scrolling is broken

https://github.com/user-attachments/assets/758aad51-4f9c-4147-8f6e-74756fc19aca

seibei-iguchi:
> The table should behave the same as before 😢
> 
> Before:
> 
>  Screen.Recording.2024-09-30.at.19.48.41.mov 
> 
> Now:
> 
>  Screen.Recording.2024-09-30.at.19.49.23.mov 
> 
> 
> scrolling is broken
> 
>  Screen.Recording.2024-09-30.at.19.45.55.mov

The change in color on row hover is something we would have to add to the Table component. Not sure if by default, or activated by a prop?

tomasklim:
> > The table should behave the same as before 😢
> > Before:
> > Screen.Recording.2024-09-30.at.19.48.41.mov
> > Now:
> > Screen.Recording.2024-09-30.at.19.49.23.mov
> > scrolling is broken
> > Screen.Recording.2024-09-30.at.19.45.55.mov
> 
> The change in color on row hover is something we would have to add to the Table component. Not sure if by default, or activated by a prop?

Maybe activate by prop? But I like it, I would add it also for tokens table

enjojoy:
> > > The table should behave the same as before 😢
> > > Before:
> > > Screen.Recording.2024-09-30.at.19.48.41.mov
> > > Now:
> > > Screen.Recording.2024-09-30.at.19.49.23.mov
> > > scrolling is broken
> > > Screen.Recording.2024-09-30.at.19.45.55.mov
> > 
> > 
> > The change in color on row hover is something we would have to add to the Table component. Not sure if by default, or activated by a prop?
> 
> Maybe activate by prop? But I like it, I would add it also for tokens table

Should I then add it as additional styling for now or should I add the prop functionality/wait for it to be added?

seibei-iguchi:
> > > > The table should behave the same as before 😢
> 
> > > > Before:
> 
> > > > Screen.Recording.2024-09-30.at.19.48.41.mov
> 
> > > > Now:
> 
> > > > Screen.Recording.2024-09-30.at.19.49.23.mov
> 
> > > > scrolling is broken
> 
> > > > Screen.Recording.2024-09-30.at.19.45.55.mov
> 
> > > 
> 
> > > 
> 
> > > The change in color on row hover is something we would have to add to the Table component. Not sure if by default, or activated by a prop?
> 
> > 
> 
> > Maybe activate by prop? But I like it, I would add it also for tokens table
> 
> 
> 
> Should I then add it as additional styling for now or should I add the prop functionality/wait for it to be added?

Feel free to add it yourself. Something like highlightRows? As it will be a Table prop, you'd have to create a context so that the Row can reach it, though.

enjojoy:
> > > > > The table should behave the same as before 😢
> > 
> > 
> > > > > Before:
> > 
> > 
> > > > > Screen.Recording.2024-09-30.at.19.48.41.mov
> > 
> > 
> > > > > Now:
> > 
> > 
> > > > > Screen.Recording.2024-09-30.at.19.49.23.mov
> > 
> > 
> > > > > scrolling is broken
> > 
> > 
> > > > > Screen.Recording.2024-09-30.at.19.45.55.mov
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > The change in color on row hover is something we would have to add to the Table component. Not sure if by default, or activated by a prop?
> > 
> > 
> > > 
> > 
> > 
> > > Maybe activate by prop? But I like it, I would add it also for tokens table
> > 
> > 
> > Should I then add it as additional styling for now or should I add the prop functionality/wait for it to be added?
> 
> Feel free to add it yourself. Something like highlightRows? As it will be a Table prop, you'd have to create a context so that the Row can reach it, though.

I will add it myself, yes, by the way on develop hover didn't work on light theme, I will try to fix that as well

packages/components/src/components/Table/HighlightContext.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { createContext, useContext } from 'react';
```
seibei-iguchi: I think it's not necessary to have a separate file. I'd do it the same way it's done in TableHeader for such a simple context.

packages/components/src/components/Table/Table.tsx
```diff
@@ -30,30 +31,36 @@ const ScrollContainer = styled.div`
...
-                        </colgroup>
-                    )}
-                    {children}
-                </Container>
-            </ScrollContainer>
-            <ShadowRight
-                backgroundColor={mapElevationToBackgroundToken({ $elevation: parentElevation })}
-            />
-        </ShadowContainer>
+        <HighlightProvider highlightRowOnHover={hightlightRowOnHover}>
```
seibei-iguchi: I'd just name it TableContext.Provider and useTable

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetTableHeader.tsx
```diff
@@ -1,43 +1,20 @@
...
-    padding-right: ${spacingsPx.xxxl};
-`;
+import { Table } from '@trezor/components';
 
 export const AssetTableHeader = () => (
-    <AssetTableRowGrid>
-        <Header /> {/* Logo */}
-        <Header>
+    <Table.Header>
+        <Table.Cell colSpan={3}>
```
seibei-iguchi: This is missing Table.Row. You probably deleted it because you don't want it to be highlighted? You can use the existing context hook from TableHeader to get info if the row is in header or not.

enjojoy: Yes, I didn't want it to be affected by the styling, okay, thank you, I will use the hook

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -37,6 +38,17 @@ const Cell = styled.td<{ $isHeader: boolean; $elevation: Elevation }>`
...
+
+    ${({ $highlight, $isHeader }) =>
+        $highlight &&
+        !$isHeader &&
+        css`
+            &:first-child:hover {
+                .hovered {
+                    background-color: red;
+                }
+            }
```
tomasklim: what should this do?

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetTableHeader.tsx
```diff
@@ -1,43 +1,24 @@
...
+                    <Translation id="TR_VALUES" />
+                </Table.Cell>
+                <Table.Cell>
+                    <Translation id="TR_EXCHANGE_RATE" />
+                </Table.Cell>
+                <Table.Cell colSpan={3}>
+                    <Translation id="TR_7D_CHANGE" />
+                </Table.Cell>
+            </Table.Row>
+        </Table.Header>
```
tomasklim: Just put this component in `AssetTable` not create a new one for header.

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetRow.tsx
```diff
@@ -27,42 +25,10 @@ const Coin = styled.div`
     overflow: hidden;
 `;
 
-const StyledCol = styled.div<{ $isLastRow?: boolean }>`
+const StyledCol = styled.div`
```
tomasklim: Can't we get rid of those `CoinBalanceContainer`, `CryptoBalanceWrapper`, `ExchangeRateWrapper`,...? I think most of the styled components here are not really needed

Btw `variables.SCREEN_QUERY.MOBILE` is user here but this table is never visible on mobile

tomasklim (changes requested):
1.
I think button and arrow should change color on row hover because on hover of buttons itself, the background is not visible

Screenshot 2024-10-02 at 17 56 05

Same in tokens table
Screenshot 2024-10-02 at 18 02 11


2. 
On click of row the same action is done as on on click of arrow button, maybe it should become active?
Screenshot 2024-10-02 at 17 55 59

3. 

There is cursor pointer on tokens table but you cannot click it so the cursor pointer should not be there
So you should probably set it only if `onClick` is available
Screenshot 2024-10-02 at 18 00 21

4. 
Zero balance row does not occupy whole width so it does not change color of whole row. I recommend adding empty `<td></td>`s
Screenshot 2024-10-02 at 18 01 08

packages/components/src/components/Table/TableCell.tsx
enjojoy: forgotten debug hack 🙂‍↔️

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetTableHeader.tsx
enjojoy: Oki, it was like that before, for better readability I guess, but it's not that complex so makes sense to put it back into the main table file

tomasklim: Before the component was huge because of styles. Now it is not true anymore

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetRow.tsx
enjojoy: Right, I removed all of them and made sure the styles stayed the same by using generic components

enjojoy:
> Zero balance row does not occupy whole width so it does not change color of whole row. I recommend adding empty `<td></td>`s

This is fixed, zero balance row occupies the whole row, also the hover and cursor pointer is applied only to the rows that have an onClick property.

enjojoy:
> There is cursor pointer on tokens table but you cannot click it so the cursor pointer should not be there
> So you should probably set it only if `onClick` is available

Done✅

enjojoy:
> On click of row the same action is done as on on click of arrow button, maybe it should become active?

AFAIK buttons have only 2 states: normal one and the change of the appearance on hover, but not on click, do you mean that the button will behave the same as the row, which means it will be hovered when the row is hovered? 

Or do you mean that it should have an entirely new state when clicked, the "active" state. 

If you mean that it should react to the onClick, I would approach it generically and make all the buttons reactive to the onClick, then we could use it here as well instead of making just one button behave differently. 

As discussed before, this button has a purely "hinting" purpose, for the user to understand that it's possible to click on the row.

enjojoy:
> I think button and arrow should change color on row hover because on hover of buttons itself, the background is not visible

As discussed before, I'm creating an issue for this. This problem occurs because elevation colors are repeating themselves.

enjojoy:
> > I think button and arrow should change color on row hover because on hover of buttons itself, the background is not visible
> 
> As discussed before, I'm creating an issue for this. This problem occurs because elevation colors are repeating themselves.

https://github.com/trezor/trezor-suite/issues/14711

packages/components/src/components/Table/TableRow.tsx
```diff
@@ -1,17 +1,55 @@
...
+        css`
+            &:hover {
+                cursor: pointer;
+                background-color: ${mapElevationToBackground({
+                    theme,
+                    $elevation: nextElevation[$elevation],
+                })};
+
+                & > td:first-child {
+                    background: linear-gradient(
```
seibei-iguchi, line 38: Perhaps make a const or func out of this so you can use it here and in the Cell? I'll try to figure out if there's a better way to do this, but I think it's good enough for now:)

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetRow.tsx
```diff
@@ -221,28 +117,28 @@ export const AssetRowSkeleton = (props: { animate?: boolean }) => {
     const animate = props.animate ?? shouldAnimate;
```
tomasklim: Please move `AssetRowSkeleton` into a new file. It does not use styled components anymore, so no need to have it here

Please also improve `(props: { animate?: boolean })` so that it is unified with other components

packages/components/src/components/Table/Table.tsx
```diff
@@ -30,30 +38,36 @@ const ScrollContainer = styled.div`
 export type TableProps = AllowedFrameProps & {
     children: ReactNode;
     colWidths?: string[];
+    highlightRowOnHover?: boolean;
 };
 
-export const Table = ({ children, margin, colWidths }: TableProps) => {
+export const Table = ({ children, margin, colWidths, highlightRowOnHover = false }: TableProps) => {
```
tomasklim, line 47: Please add `highlightRowOnHover` to storybook

tomasklim (changes requested):
Last small nits and lets merge it

tomasklim (approved):
You have type check error

btw you can unify usage of `isAnimating`, `animate`, `isAnimated`,.. to something like `shouldAnimate`

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetRow.tsx
enjojoy: Done

packages/components/src/components/Table/Table.tsx
enjojoy: Done

enjojoy:
> You have type check error
> 
> btw you can unify usage of `isAnimating`, `animate`, `isAnimated`,.. to something like `shouldAnimate`

shouldAnimate is a hookFunction so I can't use that again, but I agree that it's odd with these similar namings, not sure how to tackle it here though

PR #14626, opened 2024-09-30
Improve error handling in firmware revision check

komret:
## Description

Previously, whenever there was an error while downloading and reading the remote `releases.json` file, the user would be instructed to go online. This message is misleading when there is some other cause for an error such as incorrect URL or incorrectly formatted JSON file.

It would be nice to log these types of errors to Sentry, but then we'd have to store the stack trace somewhere temporarily before firing the event... maybe next step.

Copy could be improved and the [KB article](https://trezor.io/learn/a/trezor-firmware-revision-check) updated.

**QA:** This cannot be tested without some code modification. But you can test that I did not break the original flow.

## Related Issue

Resolve 

## Screenshots:
Screenshot 2024-09-30 at 14 35 12

### DIFF
--- a/packages/connect/src/device/checkFirmwareRevision.ts
+++ b/packages/connect/src/device/checkFirmwareRevision.ts
@@ -2,7 +2,7 @@ import { isEqual } from '@trezor/utils/src/versionUtils';
 import { PROTO } from '../constants';
 import { downloadReleasesMetadata } from '../data/downloadReleasesMetadata';
 import { FirmwareRelease, VersionArray } from '../types';
-import { FirmwareRevisionCheckResult } from '../types/device';
+import { FirmwareRevisionCheckError, FirmwareRevisionCheckResult } from '../types/device';
 import { calculateRevisionForDevice } from './calculateRevisionForDevice';
 
 type GetOnlineReleaseMetadataParams = {
@@ -20,8 +20,8 @@ const getOnlineReleaseMetadata = async ({
 };
 
 const failFirmwareRevisionCheck = (
-    error: 'revision-mismatch' | 'firmware-version-unknown' | 'firmware-version-unknown',
-): FirmwareRevisionCheckResult => ({ success: false, error });
+    error: FirmwareRevisionCheckError,
+): Extract<FirmwareRevisionCheckResult, { success: false }> => ({ success: false, error });
 
 export type CheckFirmwareRevisionParams = {
     firmwareVersion: VersionArray;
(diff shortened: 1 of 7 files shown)


packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -59,6 +59,7 @@ export const checkFirmwareRevision = async ({
     deviceRevision,
     expectedRevision,
 }: CheckFirmwareRevisionParams): Promise<FirmwareRevisionCheckResult> => {
+    console.log(expectedRevision);
```
peter-sanderson: forgotten log?

komret: fixed

packages/suite/src/components/suite/banners/SuiteBanners/FirmwareRevisionCheckBanner.tsx
```diff
@@ -1,20 +1,36 @@
...
 import { useSelector } from 'src/hooks/suite';
 
+const getMessage = (error: FirmwareRevisionCheckError) => {
+    switch (error) {
+        case 'cannot-perform-check-offline':
+            return 'TR_DEVICE_FIRMWARE_REVISION_CHECK_UNABLE_TO_PERFORM';
+        case 'other-error':
+            return 'TR_FIRMWARE_REVISION_CHECK_OTHER_ERROR';
+        default:
+            return 'TR_FIRMWARE_REVISION_CHECK_FAILED';
```
peter-sanderson: instead of default behavior, I would suggest to use map, this way whenever somebody adds new `FirmwareRevisionCheckError` it will tell him (via TS error) to handle the case

komret:
[Here](https://github.com/trezor/data/blob/master/firmware/t3b1/releases.json)'s an example of a poorly formatted JSON which cannot be parsed and causes an error:. I'll fix it in another PR.

peter-sanderson (approved):
However I would like to see what errors are causing this and somehow fix them. Json file error or incorrect URl should not happen

packages/suite/src/components/suite/banners/SuiteBanners/FirmwareRevisionCheckBanner.tsx
komret: 😢 b81ec4f5a25442a1596c902cdbd587d8ffc7597e

mroz22:
commits organized nicely 👏

PR #14629, opened 2024-09-30
Fix/native/send form keyboard events

### LINKED ISSUE #14431: Send coins - form submit button scroll position
- [ ] When the user fills in the amount to send in the send form and the submit button appears, the content of the screen should scroll so that both crypto and fiat inputs are visible above the button. [More here](https://www.figma.com/design/sqKFZAIvk3zERxiWgthK8K/Send?node-id=3256-9757&node-type=instance&t=vOcgCS1B9dwftZr4-0)
- [ ] If a user has opened the software keyboard, he  should be able to click fiat <==> crypto input without the need for an extra tap to hide the keyboard

PeKne:
## Description
- send form scrolls so both amount inputs are visible
- user can click amount switch button with opened keyboard
- if users enters the form screen with existing draft, the form is validated right away

## Related Issue

Resolve #14431

## Screenshots:

https://github.com/user-attachments/assets/628c7ae1-222c-46f5-b530-ade371c1f3e6

### DIFF
--- a/suite-native/module-send/src/components/AmountInputs.tsx
+++ b/suite-native/module-send/src/components/AmountInputs.tsx
@@ -1,6 +1,6 @@
 import { useSharedValue, withTiming } from 'react-native-reanimated';
 import { useRef, useState } from 'react';
-import { TextInput } from 'react-native';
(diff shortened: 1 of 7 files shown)


matejkriz (approved):
Looks much better 🎉

Issue #14196, opened 2024-09-06
Polygon - base logic


vytick:
- [x] change `isCoinWithTokens` to work with more token enabled networks
- [x] make txns work with token enabled networks
- [x] change ethereum token selectors to token selectors and make it network agnostic

MiroslavProchazka:
We will be doing the migration from MATIC to Polygon in the upcoming weeks, so please be aware. of it. @tomasklim is preparing all needed in [here](https://github.com/trezor/trezor-suite/pull/14093)

vytick:
@MiroslavProchazka Thanks! I
am aware and waiting for this to be merged. but mostly this is can be done even before.

vytick:
I have split this issue to multiple ones and this one was resolved by #14633 and already tested there

PR #14633, opened 2024-09-30
feat(suite-native): make token selectors network agnostic

vytick:
This PR does not touch any UI so far, only contains changes for working with tokens.

 - [x] change isCoinWithTokens to work with more token enabled networks
 - [x] make txns work with token enabled networks
 - [x]  change ethereum token selectors to token selectors and make it network agnostic

## Related Issue

Part of #14196

## Screenshots:

https://github.com/user-attachments/assets/b552aef6-7d41-42f7-ab52-b1b24e41788c


suite-native/tokens/src/tokensSelectors.ts
```diff
@@ -1,70 +1,249 @@
...
+                transaction =>
+                    isNotEmptyTransaction(transaction) ||
+                    (areTokenOnlyTransactionsIncluded &&
+                        isTransactionWithTokenTransfers(transaction)),
+            ),
+        ) as WalletAccountTransaction[],
+    { size: 500 },
+);
+
+export const selectAccountOrTokenAccountTransactions = (
```
Nodonisko: Name is strange, it returns only transactions. Something like `selectAccountTokenTransactions` is more clear

suite-native/tokens/src/tokensSelectors.ts
```diff
@@ -1,70 +1,249 @@
...
+                ) as TypedTokenTransfer[],
+            })),
+            A.filter(
+                transaction =>
+                    isNotEmptyTransaction(transaction) ||
+                    (areTokenOnlyTransactionsIncluded &&
+                        isTransactionWithTokenTransfers(transaction)),
+            ),
+        ) as WalletAccountTransaction[],
+    { size: 500 },
```
Nodonisko: I don't think there will be 500 accounts

suite-native/tokens/src/tokensSelectors.ts
```diff
@@ -1,70 +1,249 @@
...
+                })),
+            })),
+            A.filter(transaction =>
+                A.some(
+                    transaction.tokens,
+                    tokenTransfer => tokenTransfer.contract === tokenAddress,
+                ),
+            ),
+        ) as WalletAccountTransaction[],
+    { size: 500 },
```
Nodonisko: I don't think we need such a big number, something like 50 will be also fine there because there won't more than one mounted instance of component that uses this selector

Nodonisko (approved):
just few nits

suite-native/tokens/src/tokensSelectors.ts
vytick: 👍  Will change. I kept the number from renamed and moved ethereum selector.

Nodonisko:
Also E2E seems to fail

suite-native/tokens/src/tokensSelectors.ts
vytick: It returns either all txns for whole account (eg. eth + tokens)or only for some token (usdc on eth)

vytick: e8dde2170e

suite-native/tokens/src/tokensSelectors.ts
vytick: e8dde2170e

suite-native/tokens/src/tokensSelectors.ts
vytick: e8dde2170e

vytick:
@STew790 Please test, that everything works as before:
- with disabled poygon and disabled bnb only eth has tokens and nothing else
- eth has correctly token txns toggle and nothing else (unless enabled pol/bnb) has it
- nothing breaks with tokens and transactions

STew790:
**QA OK**
- Only ETH has tokens (I don't see them anywhere else) with BNB and Polygon disabled
- ETH has toggle in transactions for tokens
- tokens and other transactions LGTM

**Info**
24.9.2 https://github.com/trezor/trezor-suite/commit/b718af4a2b5d9fde28e1832276405f36a1c2399b

PR #14637, opened 2024-10-01
chore(suite): unreadable device tips improvements

### LINKED ISSUE #14284: Confusing instructions on landing page for old devices
When you connect an old device not supported by WebUsb while using web Suite, you get the following instructions, which are confusing and not helpful. I tried this with bootloader 1.3.3 on devkit.

Web:
Screenshot 2024-09-11 at 13 52 06
1. The top message "Device detected in incorrect state" is inaccurate. There is a problem with the app, not the device.
1. It then tells you to download Bridge, but does not tell you how to get it.
1. It leads you to Bridge status page, but you probably don't have Bridge at all. (Copy could be better.)
1. It tells you to use desktop app, which won't work either because it doesn't use Bridge anymore.
1. The button to disable WebUSB and use Bridge instead won't help you because you probably don't have Bridge.

mroz22:
addressing #14284

todos: 
- do we have a knowledge base article addressing old hid devices fw update issues?
- do we wan't to include button to switch to using old bridge (desktop only)

image

cc @MiroslavProchazka 

resolve #14284

### DIFF
--- a/packages/suite/src/components/suite/PrerequisitesGuide/Transport.tsx
+++ b/packages/suite/src/components/suite/PrerequisitesGuide/Transport.tsx
@@ -1,23 +1,28 @@
 import { Translation, TroubleshootingTips } from 'src/components/suite';
+
 import {
     TROUBLESHOOTING_TIP_SUITE_DESKTOP,
     TROUBLESHOOTING_TIP_RESTART_COMPUTER,
     TROUBLESHOOTING_TIP_WEBUSB_ENVIRONMENT,
 } from 'src/components/suite/troubleshooting/tips';
 
-export const Transport = () => (
-    // No transport layer (bridge/webUSB) is available
-    // On web it makes sense to
-    // - offer downloading Trezor Suite desktop, or
-    // - use a browser that supports WebUSB
-    // Desktop app should have Bridge transport layer available as it is built-in, if it is not available we fucked up something.
-    <TroubleshootingTips
-        label={<Translation id="TR_TROUBLESHOOTING_DEVICE_NOT_DETECTED" />}
-        items={[
-            TROUBLESHOOTING_TIP_WEBUSB_ENVIRONMENT,
-            TROUBLESHOOTING_TIP_SUITE_DESKTOP,
-            TROUBLESHOOTING_TIP_RESTART_COMPUTER,
-        ]}
-        data-testid="@connect-device-prompt/bridge-not-running"
-    />
-);
+export const Transport = () => {
+    const items = [
+        TROUBLESHOOTING_TIP_WEBUSB_ENVIRONMENT,
+        TROUBLESHOOTING_TIP_SUITE_DESKTOP,
+        TROUBLESHOOTING_TIP_RESTART_COMPUTER,
+    ];
+
+    return (
+        // No transport layer (bridge/webUSB) is available
+        // On web it makes sense to
+        // - offer downloading Trezor Suite desktop, or
+        // - use a browser that supports WebUSB
+        // Desktop app should have Bridge transport layer available as it is built-in, if it is not available we fucked up something.
+        <TroubleshootingTips
+            label={<Translation id="TR_TROUBLESHOOTING_DEVICE_NOT_DETECTED" />}
+            items={items}
+            data-testid="@connect-device-prompt/bridge-not-running"
+        />
+    );
+};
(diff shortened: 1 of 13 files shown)


MiroslavProchazka:
I would not lead them to KB article as there is a significant chance they will brick the device. I would just lead them directly to write tech support

MiroslavProchazka:
Regarding the button switch, @Hannsek and @Hermez-cz shall decide. I would be personakls up for adding it there.

Hannsek:
If we can target the specific firmwares then yes, let's do it. I had it in the design suggestion somewhere.

But if it is just ~ 10 users, we shouldn't spend time on this.

Hannsek:
We have this on KB:

image

mroz22:
I would do something like this (for desktop)

image

and this for web 

image

explanation why I believe it makes sense as suggested here https://github.com/trezor/trezor-suite/pull/14637/files#diff-4f4ba83e95f4c0800b23ced0ac2643abf14ea387d3de6e91d0ac38b47113e2bfR123

cc @Hannsek please check on it and please give me the KB link. 

Also connect-device troubleshooting probably should be revisited as well.

Hannsek:
It seems fine to me. The KB article is [here](https://trezor.io/support/a/trezor-suite-doesn-t-see-my-device).

@Hermez-cz please check it as well.

Hermez-cz:
Hi, I didn't catch if we can recognize affected devices. Is that possible please?

The bridge toggle might be puzzling minds of all users with connection issues and there's no way they'll understand what it's about. They have no idea what bridge is, what might be an alternative implementation, what does the 3.0.0-bundled.24.10.0 mean etc. All feels very non-transparent and anyone might just feel forced to give it a blind shot.

Also old Trezor model vs. firmware version sounds fuzzy like this. Should we refer only to firmware / device software perhaps? Because it doesn't mean a thing if my Model One would be very old if I'd keep it up to date.

The URLs are the most prominent thing in the whole page, more prominent than those bullet point titles, is that intentional?

Hannsek:
Please do some adjustments:

- [ ] display this 4 points error only for HID devices
- [ ] other devices should see only first and fourth point (is there another troubleshooting step that may help core devices?)
- [ ] adjust second point so that after clicking on the link, a bridge modal would appear where the toggle to switch to old bridge will be present

packages/suite/src/components/suite/PrerequisitesGuide/Transport.tsx
```diff
@@ -1,23 +1,28 @@
...
-    <TroubleshootingTips
-        label={<Translation id="TR_TROUBLESHOOTING_DEVICE_NOT_DETECTED" />}
-        items={[
-            TROUBLESHOOTING_TIP_WEBUSB_ENVIRONMENT,
-            TROUBLESHOOTING_TIP_SUITE_DESKTOP,
-            TROUBLESHOOTING_TIP_RESTART_COMPUTER,
-        ]}
-        data-testid="@connect-device-prompt/bridge-not-running"
-    />
-);
```
mroz22, line 23: no changes here. sorry for diff

martykan (approved):
Looks good 👍 

Only thing I noticed when testing is when toggling bridge implementations the error changes into "App can't communicate with device" momentarily. But that's probably not important

Hannsek:
@evgenysl Once you'll be testing that, please post here screenshots of how does it look now. 🙏🏻

bosomt:
This is what i see on 24.11.2 when i connect `1.3.3` device

Image

**Info:**
 - Suite version: web 24.11.2 (b44a924ad9e7cb96c8eaadfdca45e4183314bd66)
 - OS: MacIntel
 - Screen: 1512x982
 - Device: 
 - Transport: WebUsbTransport

bosomt:
This is Node bridge and app with same device

image


**Info:**
 - Suite version: desktop 24.11.2 (b44a924ad9e7cb96c8eaadfdca45e4183314bd66)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.11.2 Chrome/128.0.6613.162 Electron/32.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: 
 - Transport: BridgeTransport 3.0.0-bundled.24.11.2

mroz22:
Hmm I think it should have displayed something different. Could you please also post old bridge status page with that device?

bosomt:
Same device with old bridge running

image
Screenshot 2024-11-22 at 10 45 10

PR #14640, opened 2024-10-01
fix(suite): fix incorrectly listed bridge version in /bridge page

### LINKED ISSUE #14639: webusb/bridge detection
bridge detection might still be in trouble. see:

image

also exodus suggests installing bridge even though it is already running 

image

could it be connected with the new bride format version we are now serving from the `/` endpoint?

mroz22:
fix https://github.com/trezor/trezor-suite/issues/14639#issuecomment-2384978680

### DIFF
--- a/packages/suite/src/views/suite/bridge/index.tsx
+++ b/packages/suite/src/views/suite/bridge/index.tsx
@@ -50,12 +50,9 @@ export const BridgeUnavailable = () => {
     // if bridge is running, user will never be directed to this page, but since this page is accessible directly over /bridge url
     // it makes sense to show some meaningful information here
     if (transport?.type) {
-        // not 100%, this is true only for latest suite-desktop release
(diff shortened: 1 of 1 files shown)


packages/suite/src/views/suite/bridge/index.tsx
```diff
@@ -50,12 +50,9 @@ export const BridgeUnavailable = () => {
     // if bridge is running, user will never be directed to this page, but since this page is accessible directly over /bridge url
     // it makes sense to show some meaningful information here
     if (transport?.type) {
-        // not 100%, this is true only for latest suite-desktop release
-        const isBundled = transport?.bridge?.version?.[0] === 3;
```
mroz22, line 54: `bundled` is now part of version string

marekrjpolak (approved):

PR #14645, opened 2024-10-01
chore(suite): bump node bridge to 1%

MiroslavProchazka:
Bumping node bridge to 1% for users apart from EAP


mroz22:
Ok, se we decided to merge this AFTER the freeze. So the upcoming release will be 100% EAP users, 0% regular users.

mroz22 (approved):

PR #14648, opened 2024-10-01
Change the EVM legacy discovery error on old firmware

### LINKED ISSUE #14638: Change "Forbidden key path" for EVM legacy accounts
EVM legacy paths are not available on old fw versions (e.g. 2.5.3TT). Instead of skipping it for old fw, let's change the text to something like:

`Ethereum Legacy: Update firmware to discover this account`

Screenshot 2024-09-30 at 14 24 17

enjojoy:
## Description

Changed the error message for non-normal EVM account types, when discovery error occurs. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14638

## Screenshots:

image

### DIFF
--- a/packages/suite/src/views/dashboard/PortfolioCard/PortfolioCardException.tsx
+++ b/packages/suite/src/views/dashboard/PortfolioCard/PortfolioCardException.tsx
@@ -7,7 +7,7 @@ import {
     authorizeDeviceThunk,
     restartDiscoveryThunk as restartDiscovery,
 } from '@suite-common/wallet-core';
-import { getNetwork } from '@suite-common/wallet-config';
+import { getNetwork, NetworkType } from '@suite-common/wallet-config';
 import { variables, Button, H3, Image, IconName } from '@trezor/components';
 import { Discovery } from '@suite-common/wallet-types';
 
@@ -102,9 +102,13 @@ const Container = ({ title, description, cta, dataTestBase }: ContainerProps) =>
     );
 };
 
-const getAccountError = (accountError: string) => {
+const getAccountError = (accountError: string, networkType: NetworkType) => {
     if (accountError === 'All backends are down') {
         return <Translation id="TR_CONNECTION_LOST" />;
+    }
+
+    if (networkType === 'ethereum' && accountError === 'Forbidden key path') {
+        return <Translation id="TR_UPGRADE_FIRMWARE_TO_DISCOVER_ACCOUNT_ERROR" />;
     }
 
     return accountError;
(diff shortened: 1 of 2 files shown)


packages/suite/src/views/dashboard/PortfolioCard/PortfolioCardException.tsx
```diff
@@ -121,9 +125,12 @@ const discoveryFailedMessage = (discovery?: Discovery) => {
         if (networkError.includes(account.symbol)) return value;
         networkError.push(account.symbol);
 
+        const accountTypeDisplay = account.accountType !== 'normal' ? account.accountType : '';
+
         return value.concat(
             <div key={account.symbol}>
-                {network.name}: {getAccountError(account.error)}
+                {network.name} {accountTypeDisplay}:{' '}
```
tomasklim: This means, that if it is normal, there is ` `

enjojoy: Right, let me remove the " "

enjojoy: Removed, now the space will be only when the type is not normal

tomasklim (changes requested):
https://github.com/trezor/trezor-suite/blob/develop/COMMITS.md

enjojoy:
> https://github.com/trezor/trezor-suite/blob/develop/COMMITS.md

sorry, no idea how I could forget this. reworded.

tomasklim (approved):

Hermez-cz:
The copywriting is clear and concise, directly explaining the issue. However, the [Retry] CTA feels out of place. From the user's perspective, a CTA to update the firmware would be more appropriate, as retrying will lead to the same result without the update. Since we can't change the flow at this point, how about adding '(see the blue banner above)' at the end of the body text?

enjojoy:
> The copywriting is clear and concise, directly explaining the issue. However, the [Retry] CTA feels out of place. From the user's perspective, a CTA to update the firmware would be more appropriate, as retrying will lead to the same result without the update. Since we can't change the flow at this point, how about adding '(see the blue banner above)' at the end of the body text?

Sure, I will add this note

evgenysl:
**QA OK**

- FW 2.6.0 and newer - no issues with Legacy account discovery


- FW 2.5.3  

Screenshot 2024-10-01 at 15 45 03


Screenshot 2024-10-01 at 15 46 15


Message is small and is not present in Legacy account itself (second screenshot), but this is an edge case for firmware from November 2022 and older.

PR #14649, opened 2024-10-01
Fix/device label selector

komret:
## Description

Use correct selector in `DeviceStatus` so that a correct device label is shown when multiple devices are connected.

## Screenshots:
before:
Screenshot 2024-10-01 at 12 03 35
after:
Screenshot 2024-10-01 at 12 00 43


Lemonexe (approved):
Looks good :rocket:

:thought_balloon:  Maybe the selectors such as `selectDeviceName` could be named `selectCurrentDeviceName` and etc., but ofc that does not relate to this PR. Just thinking out loud that because of the unclear name, it is very easy to make such a mistake.

STew790:
**QA OK**
image

**Info:**
 - Suite version: web 24.10.0 (bfe42f2d115a933baeb1c44901ebcae176e1e1e8)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080

PR #14662, opened 2024-10-01
refactor(connect): connect error codes typing

PeKne:
## Description

- TrezorConnect methods now return error code of union type instead of generic string for better type safety.


mroz22 (approved):
looks useful. thanks.

PR #14667, opened 2024-10-01
test: deeplink connect native

karliatto:
## Description

Add new native e2e android test `suite-native/app/e2e/tests/deeplinkPopup.test.ts` where we call a `@trezor/connect-mobile` and get response in callback of a test server to check that we are getting the right response.

## Related Issue
Related https://github.com/trezor/trezor-suite/issues/14744

### DIFF
--- a/packages/trezor-user-env-link/src/api.ts
+++ b/packages/trezor-user-env-link/src/api.ts
@@ -59,6 +59,7 @@ export const MNEMONICS = {
     mnemonic_12: 'alcohol woman abuse must during monitor noble actual mixed trade anger aisle',
     mnemonic_abandon:
         'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about',
+    mnemonic_immune: 'immune enlist rule measure fan swarm mandate track point menu security fan',
 };
 
 export const DEFAULT_BRIDGE_VERSION = '2.0.33';
--- a/suite-native/app/e2e/tests/deeplinkPopup.test.ts
+++ b/suite-native/app/e2e/tests/deeplinkPopup.test.ts
@@ -1,0 +1,142 @@
+import http from 'http';
+import { exec } from 'child_process';
+
+import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
+import TrezorConnect from '@trezor/connect-mobile';
+
+import {
+    appIsFullyLoaded,
+    disconnectTrezorUserEnv,
+    openApp,
+    prepareTrezorEmulator,
+    restartApp,
+} from '../utils';
+import { onOnboarding } from '../pageObjects/onboardingActions';
+import { onCoinEnablingInit } from '../pageObjects/coinEnablingActions';
+import { onHome } from '../pageObjects/homeActions';
+
+const SERVER_PORT = 8080;
+const SERVER_URL = `http://localhost:${SERVER_PORT}`;
+
+let server: http.Server | undefined;
+
+const openUriScheme = (url: string, platformToOpen: 'android') => {
+    const command = `npx uri-scheme open '${url.replace(/'/g, '')}' --${platformToOpen}`;
+
+    exec(command, (err, stdout, stderr) => {
+        if (err) {
+            console.error(err);
+
+            return;
+        }
+        // eslint-disable-next-line no-console
+        console.info(stdout);
+        console.error(stderr);
+    });
+};
+
+describe('Deeplink connect popup.', () => {
+    beforeAll(async () => {
+        await new Promise(resolve => {
+            server = http.createServer((req, res) => {
+                if (req.url) {
+                    const url = new URL(req.url, SERVER_URL);
+                    TrezorConnect.handleDeeplink(url.href);
+                    res.statusCode = 200;
+                    res.setHeader('Content-Type', 'text/plain');
+                    res.end('Callback URL received successfully!\n');
+                }
+            });
+
+            server.listen(SERVER_PORT, 'localhost', () => {
+                // eslint-disable-next-line no-console
+                console.info(`Server running at ${SERVER_URL}`);
+                resolve(null);
+            });
+        });
+
+        await prepareTrezorEmulator();
+        await openApp({ newInstance: true });
+        await onOnboarding.finishOnboarding();
+
+        await onCoinEnablingInit.waitForScreen();
+        await onCoinEnablingInit.enableNetwork('regtest');
+        await onCoinEnablingInit.clickOnConfirmButton();
+
+        await waitFor(element(by.id('skip-view-only-mode')))
+            .toBeVisible()
+            .withTimeout(10000); // communication between connected Trezor and app takes some time.
+
+        await element(by.id('skip-view-only-mode')).tap();
+
+        // This `TrezorConnect` instance here is pretending to be the integrator or @trezor/connect-mobile
+        await TrezorConnect.init({
+            manifest: {
+                email: 'developer@xyz.com',
+                appUrl: 'http://your.application.com',
+            },
+            deeplinkOpen: url => {
+                openUriScheme(url, 'android');
+            },
+            deeplinkCallbackUrl: `${SERVER_URL}/connect/`,
+            connectSrc: 'https://dev.suite.sldev.cz/connect/develop/',
+        });
+    });
+
+    beforeEach(async () => {
+        await restartApp();
+
+        await device.reverseTcpPort(SERVER_PORT);
+
+        await appIsFullyLoaded();
+
+        await waitFor(element(by.id('@screen/ConnectingDevice')))
+            .toBeVisible()
+            .withTimeout(10000);
+
+        await onHome.waitForScreen();
+    });
+
+    afterAll(async () => {
+        disconnectTrezorUserEnv();
+        await new Promise(resolve => {
+            if (server) {
+                server.close(() => {
+                    resolve(null);
+                });
+            }
+        });
+        await device.terminateApp();
+    });
+
+    it('Handle deeplink', async () => {
+        const promise = TrezorConnect.getAddress({
+            path: "m/49'/0'/0'/0/0",
+            coin: 'btc',
+        });
+
+        await element(by.id('@popup/deeplink-info'));
+
+        await waitFor(element(by.id('@popup/call-device')))
+            .toBeVisible()
+            .withTimeout(10 * 1000);
+        await element(by.id('@popup/call-device')).tap();
+
+        await TrezorUserEnvLink.pressYes();
+
+        const response = await promise;
+        const expectedResponse = {
+            id: 1,
+            payload: {
+                path: [49, 0, 0, 0, 0],
+                serializedPath: 'm/49/0/0/0/0',
+                address: '3J7UQSLAaFh1nkUomVdm8ArifPEvYfNr1o',
+            },
+            success: true,
+        };
+
+        if (JSON.stringify(response) !== JSON.stringify(expectedResponse)) {
+            throw new Error('Result does not match expected.');
+        }
+    });
+});
--- a/suite-native/app/e2e/tests/deeplinkPopup.test.ts
+++ b/suite-native/app/e2e/tests/deeplinkPopup.test.ts
@@ -3,6 +3,7 @@ import { exec } from 'child_process';
 
 import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
 import TrezorConnect from '@trezor/connect-mobile';
+import { conditionalDescribe } from '@suite-common/test-utils';
 
 import {
     appIsFullyLoaded,
@@ -35,7 +36,7 @@ const openUriScheme = (url: string, platformToOpen: 'android') => {
     });
 };
 
-describe('Deeplink connect popup.', () => {
+conditionalDescribe(device.getPlatform() !== 'android', 'Deeplink connect popup.', () => {
     beforeAll(async () => {
         await new Promise(resolve => {
             server = http.createServer((req, res) => {
--- a/suite-common/test-utils/src/conditionalDescribe.ts
+++ b/suite-common/test-utils/src/conditionalDescribe.ts
@@ -1,0 +1,11 @@
+export const conditionalDescribe = (
+    skipCondition: boolean,
+    title: string,
+    fn: jest.EmptyFunction,
+) => {
+    if (skipCondition) {
+        describe.skip(title, fn);
+    } else {
+        describe(title, fn);
+    }
+};
--- a/suite-common/test-utils/src/index.ts
+++ b/suite-common/test-utils/src/index.ts
@@ -1,3 +1,4 @@
 export * from './mocks';
 export * from './configureMockStore';
 export * from './extraDependenciesMock';
+export * from './conditionalDescribe';
--- a/suite-native/app/e2e/utils.ts
+++ b/suite-native/app/e2e/utils.ts
@@ -1,7 +1,7 @@
 import { resolveConfig } from 'detox/internals';
 import { expect as detoxExpect } from 'detox';
 
-import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
+import { MNEMONICS, TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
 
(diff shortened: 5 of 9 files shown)


suite-native/app/e2e/tests/deeplinkPopup.test.ts
```diff
@@ -0,0 +1,115 @@
...
+import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
+import TrezorConnect from '@trezor/connect-deeplink';
+
+import { openApp } from '../utils';
+import { onOnboarding } from '../pageObjects/onboardingActions';
+import { onCoinEnablingInit } from '../pageObjects/coinEnablingActions';
+
+const TREZOR_DEVICE_LABEL = 'Trezor T - Tester';
+// Contains only one BTC account with a single transaction to make the discovery as fast as possible.
+const SIMPLE_SEED = 'immune enlist rule measure fan swarm mandate track point menu security fan';
```
mroz22: just recently, we started gathering all the mnemonics in one place here https://github.com/trezor/trezor-suite/blob/develop/packages/trezor-user-env-link/src/api.ts#L57 
not a required change, just a note to be considered

karliatto: Makes sense, I added it there https://github.com/trezor/trezor-suite/pull/14667/commits/17bb3c99871bb0ed3d0c7c9252459a9bf43bfdb0

karliatto:
I think this is ready.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    disconnectTrezorUserEnv,
+    openApp,
+    prepareTrezorEmulator,
+    restartApp,
+} from '../utils';
+import { onOnboarding } from '../pageObjects/onboardingActions';
+import { onCoinEnablingInit } from '../pageObjects/coinEnablingActions';
+import { onHome } from '../pageObjects/homeActions';
+
+const SERVER_PORT = 8080;
```
mroz22, line 19: Is there any chance that I would be running this locally while having connect-explrorer dev server active? I think that we are using this port somewhere already. clash is probably unlikely however.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
```diff
@@ -0,0 +1,162 @@
...
+                res.statusCode = 200;
+                res.setHeader('Content-Type', 'text/plain');
+                res.end('Callback URL received successfully!\n');
+            }
+        });
+
+        server.listen(SERVER_PORT, 'localhost', () => {
+            // eslint-disable-next-line no-console
+            console.info(`Server running at ${SERVER_URL}`);
+        });
```
mroz22: nitpick: execution doesn't wait for server to be already listening. it could lead to a racecondition that is very unlikely to happen

mroz22 (commented):
looks good to me. just left 2 useless comments, feel free to ignore.

maybe just ask mobile guys if they want to have a look too? Or is @HajekOndrej interested here?

suite-native/app/e2e/tests/deeplinkPopup.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        await TrezorConnect.init({
+            manifest: {
+                email: 'developer@xyz.com',
+                appUrl: 'http://your.application.com',
+            },
+            deeplinkOpen: url => {
+                openUriScheme(url, 'android');
+            },
+            deeplinkCallbackUrl: `${SERVER_URL}/connect/`,
+            connectSrc: 'https://dev.suite.sldev.cz/connect/9.1.2/',
```
mroz22: btw this link is alive?

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: As far as I know connect-explorer is running on port 8088.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: That is a very good point. This should love that https://github.com/trezor/trezor-suite/pull/14667/commits/9b9116f4048b692faf897efdda225a05ab936f31

suite-native/app/e2e/tests/deeplinkPopup.test.ts
mroz22: right, I saw it here https://github.com/trezor/trezor-suite/blob/develop/docs/features/coinjoin.md#L15

.github/workflows/test-suite-native-e2e-android.yml
```diff
@@ -115,6 +115,14 @@ jobs:
           path: node_modules
           key: node_modules/${{ github.ref }}/${{github.run_id}}
 
+      - name: Install adb (Android SDK Platform Tools)
```
PeKne: This should not be needed. All the Android tooling should already be included in the `reactivecircus/android-emulator-runner@v2` workflow used below.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: No, it actually works because it satisfy the intentFilters https://github.com/trezor/trezor-suite/blob/develop/suite-native/app/app.config.ts#L200

Not critical for tests, but yes, better to have real link, changed:

https://github.com/trezor/trezor-suite/pull/14667/commits/dd885269503998d6a820532bddf455f69ff34c54

So it works also with the nice fallback in web popup.

packages/trezor-user-env-link/src/api.ts
```diff
@@ -59,6 +59,8 @@ export const MNEMONICS = {
     mnemonic_12: 'alcohol woman abuse must during monitor noble actual mixed trade anger aisle',
     mnemonic_abandon:
         'abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about',
+    // Contains only one BTC account with a single transaction to make the discovery as fast as possible.
+    mnemonic_immune: 'immune enlist rule measure fan swarm mandate track point menu security fan',
```
PeKne, line 62: Would it not make sense to name the seeds by their attributes rather than phrase words? The attribute of this seed is that it contains only a single BTC account and nothing else. Wdyt? Just thinking out loud.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+            return;
+        }
+        // eslint-disable-next-line no-console
+        console.info(stdout);
+        console.error(stderr);
+    });
+};
+
+const reverseServerPort = (port: number) => {
```
PeKne: There is a built-in functionality for reversing ports in the detox framework. Use [this instead](https://wix.github.io/Detox/docs/api/device/#devicereversetcpport-android-only).

suite-native/app/e2e/tests/deeplinkPopup.test.ts
```diff
@@ -0,0 +1,165 @@
...
+import { onOnboarding } from '../pageObjects/onboardingActions';
+import { onCoinEnablingInit } from '../pageObjects/coinEnablingActions';
+import { onHome } from '../pageObjects/homeActions';
+
+const SERVER_PORT = 8080;
+const SERVER_URL = `http://localhost:${SERVER_PORT}`;
+
+let server: http.Server | undefined;
+
+const openUriScheme = (url: string, platformToOpen: 'android') => {
```
PeKne, line 24: Use the [detox device API](https://wix.github.io/Detox/docs/guide/mocking-open-with-url) instead, please.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
```diff
@@ -0,0 +1,165 @@
...
+            deeplinkOpen: url => {
+                openUriScheme(url, 'android');
+            },
+            deeplinkCallbackUrl: `${SERVER_URL}/connect/`,
+            connectSrc: 'https://dev.suite.sldev.cz/connect/develop/',
+        });
+    });
+
+    beforeEach(async () => {
+        await prepareTrezorEmulator();
```
PeKne: is not Trezor emulator already "prepared" from the  `beforeAll` step? Is it needed here again?

suite-native/app/e2e/tests/deeplinkPopup.test.ts
```diff
@@ -0,0 +1,165 @@
...
+            id: 1,
+            payload: {
+                path: [49, 0, 0, 0, 0],
+                serializedPath: 'm/49/0/0/0/0',
+                address: '3J7UQSLAaFh1nkUomVdm8ArifPEvYfNr1o',
+            },
+            success: true,
+        };
+
+        await new Promise((resolve, reject) => {
```
PeKne: why is the Promise object needed here? What about simplifying it to something like:
```suggestion
        if (JSON.stringify(response) !== JSON.stringify(expectedResponse)) {
                new Error('Result does not match expected.')
            }
```

PeKne (changes requested):
Great idea to write an automated test for this. This will save us a lot of time debugging later if someone breaks the connect popup integration. 👏

packages/trezor-user-env-link/src/api.ts
karliatto: Well, it contains only a single BTC account for now, it does not necessarily has to be the same tomorrow, and it is public seed. 
But anyway in the tests now we are using regtest so  it actually does not make any sense. I am thinking now that we should even delete the comment.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: I got that "inspiration" from https://github.com/trezor/trezor-suite/blob/develop/suite-native/app/e2e/tests/send.test.ts#L72 but if it not necessary let's get rid of it.

packages/trezor-user-env-link/src/api.ts
PeKne: You are probably right. The mainnet discovery is still happening in the [onboardAndConnect.test.ts](https://github.com/trezor/trezor-suite/blob/1150abeef6bdb9026be5acf7c9587914da4ea45e/suite-native/app/e2e/tests/onboardAndConnect.test.ts) but we are not doing discovery of all networks (since we have now coin enabling on mobile as well).  So feel free to delete the comment.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
PeKne: Try to remove it. IMO, it should not be needed here. In case it does not work, feel free to put it back.

.github/workflows/test-suite-native-e2e-android.yml
karliatto: Sure, I removed it since `device.reverseTcpPort()` works nicely.

https://github.com/trezor/trezor-suite/pull/14667/commits/d4003508eaa78fb168e359f9eacec13dc32e80e0

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: It sounds like good idea but I have been testing it and does not really works as this test need, so I would keep this approach.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: Done simpler https://github.com/trezor/trezor-suite/pull/14667/commits/72c023f4fce281731b15d7681bb6b6e99a6a9912#diff-0cba7b27a14d30c9b8c7c5b3ac0c073b9e94a549ac4c8aeb74012a0676faba91R123

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: Yes, you are right, it was not required. Removed.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: Done, it works nicely https://github.com/trezor/trezor-suite/pull/14667/commits/72c023f4fce281731b15d7681bb6b6e99a6a9912#diff-0cba7b27a14d30c9b8c7c5b3ac0c073b9e94a549ac4c8aeb74012a0676faba91R74

karliatto:
@PeKne I responded to all your request and tests are still passing. Let me know!

suite-native/app/e2e/tests/deeplinkPopup.test.ts
PeKne: This, unfortunately, does not work I tried to put a dummy value to the `expectedResponse` e.g. `{success: false}`, and the test still passed. It is because if the values do not match, an error object is created, but it is not thrown.

To ensure the values are compared correctly, you must use the jest assert function: `expect (response).toEqual(expectedResponse)`. I am aware that I recommended you to use the Error object in the previous comment, but I was rather focusing on removing the `Promise` object.

PeKne:
Also one more thing. Now the test is working only for Android platform, but our tests are written for both Android and iOS. So it will always fail on Apple devices. For this reason, some platform condition should be introduced. ChatGPT offered me this solution (I tested it and it works):

```ts
...

// Define a function to conditionally skip or run the test suite
function conditionalDescribe(skipCondition: boolean, title: string, fn: jest.EmptyFunction) {
    if (skipCondition) {
        describe.skip(title, fn);
    } else {
        describe(title, fn);
    }
}

conditionalDescribe(device.getPlatform() !== 'android', 'Deeplink connect popup.', () => {
...
```
IMO it would make sense to put the `conditionalDescribe` to test utils in suite-common. It might be helpful later to someone else.

mroz22:
> IMO it would make sense to put the conditionalDescribe to test utils in suite-common. It might be helpful later to someone else.

this should be solved even more globally. we already use a custom conditionalTest in connect for example. cc @HajekOndrej

martykan:
I believe we can test iOS as well, it should work on both with the emulator.

martykan:
So I attempted to adjust the test for iOS, however there are some issues during onboarding and coin enabling, which block the progress of the test:
```
The app is busy with the following tasks:
• There are 1 work items pending on the dispatch queue: "Main Queue (<OS_dispatch_queue_main: com.apple.main-thread>)".
• 1 enqueued native timers:
- Timer #1:
    + Fire date: none.
    + Time until fire: 0.000.
    + Repeat interval: 0.
    + Is recurring: YES.
• Run loop "Main Run Loop" is awake. 
```
So overall the iOS E2E tests first need changes to support device connection.

PeKne:
@martykan We do not want to make iOS tests behave like Android devices. At the moment, the flows and UI are different between iOS and Android. So, it is not intended to make the tests do the same thing on both platforms. Thus, device connection should not be allowed for iOS e2e tests now until we introduce a Trezor device supporting Bluetooth.

martykan:
Ok, understandable, but currently all the iOS tests don't match the real behavior of the app. 
When I build and run `ios.sim.release` tests, it's actually using the Debug variant, which supports connecting device and the flow is the same as Android. 
Or maybe my setup is wrong?

PeKne:
@martykan Good point. We have not used iOS tests in CI until now, so it is highly possible that the iOS test does not behave as it should. However, soon, there will also be an iOS test run in CI, so it will need to be fixed.

suite-native/app/e2e/tests/deeplinkPopup.test.ts
karliatto: Sure, let's throw https://github.com/trezor/trezor-suite/pull/14667/commits/407c5d0bb0220465be8e30e83a4515590214958d

`expect` does not work as expected, so I would keep the current approach but throwing the error, I tested with success and fail and it worked.

PeKne: Strange, the expect worked for me locally. I like the `expect` keyword more because it prints out the difference between the two compared values in case of failure. Never mind, this works as well :).

PeKne (approved):
Good job 👏

Issue #13990, opened 2024-08-27
Silent Update - UI


peter-sanderson:
Figma: https://www.figma.com/design/d8Ekwm865R3YR6PtH4gdyc/Auto-update?node-id=4083-20734&t=D2t4VMq29llhHKn3-1

- [x] Legacy Icons refactoring
- [x] Icons (QuickActions)
- [x] Icon with corner Icon component
- [x] Update Icons (to display FW and Suite update state)
- [x] Tooltip Update Only
- [x] Update Tooltip on WEB (what to do on Web? hide the Suite part?)
- [x] Update Icon is clickable 
- [x] Tooltips
- [x] Implement update mini-popup-banner (there are many states see figma)
![image](https://github.com/user-attachments/assets/e76daec7-bb43-4d67-ae0d-199ddb62e49d)
![image](https://github.com/user-attachments/assets/af557c76-a1af-4429-8b76-a4e69a7b97eb)
- [x] Click on mini-popup-banner "Just updated" will lead to changelog
- [x] Implement new Update Modals (should be probably new Issue, its a big stuff)
- [x] App (Suite) "Whats new"  click works to show changelog
- [x] remove it from experimental features

bosomt:
QA task here https://www.notion.so/satoshilabs/Silent-update-11fdc526060680a0b7c2f11c5d999c98#11fdc526060681bb9a60d1fc55b6bcec

bosomt:
**QA OK**

sanity tests for multiple platforms ok, if anything raises we will create separate issue
 - Suite version: desktop 24.11.1 (9e9dadff47c5243bf4a10f539a898fe48da46506)

PR #14679, opened 2024-10-02
feat: NewModal can have color (to override the variant), also CSSColo…

peter-sanderson:
- Allows branding of the modal to custom color (while preserving the button variants)
- Fixes the string in IconColor and enforces stronly typed `CSSColor` everywhere

Partially related to https://github.com/trezor/trezor-suite/issues/13990

image

Example usage: 
image

### DIFF
--- a/packages/components/src/components/Badge/Badge.tsx
+++ b/packages/components/src/components/Badge/Badge.tsx
@@ -18,7 +18,7 @@ export type BadgeSize = Extract<UISize, (typeof badgeSizes)[number]>;
 export const allowedBadgeFrameProps = ['margin'] as const satisfies FramePropsKeys[];
 type AllowedFrameProps = Pick<FrameProps, (typeof allowedBadgeFrameProps)[number]>;
 
-type BadgeVariant = Extract<UIVariant, 'primary' | 'tertiary' | 'destructive'>;
+export type BadgeVariant = Extract<UIVariant, 'primary' | 'tertiary' | 'destructive'>;
 
 export type BadgeProps = AllowedFrameProps & {
     size?: BadgeSize;
@@ -150,7 +150,7 @@ export const Badge = ({
                     name={icon}
                     color={
                         isDisabled
-                            ? 'iconDisabled'
+                            ? theme.iconDisabled
                             : mapVariantToIconColor({ $variant: variant, theme })
                     }
                 />
--- a/packages/components/src/components/NewModal/NewModal.stories.tsx
+++ b/packages/components/src/components/NewModal/NewModal.stories.tsx
@@ -46,6 +46,8 @@ export default meta;
 export const NewModal: StoryObj<NewModalProps> = {
     args: {
         variant: 'primary',
+        iconName: undefined,
+        iconComponent: undefined,
         heading: 'Modal heading',
         description: 'Modal description',
         children:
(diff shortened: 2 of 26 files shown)


packages/components/src/components/Badge/Badge.tsx
```diff
@@ -149,9 +149,11 @@ export const Badge = ({
                 <Icon
                     name={icon}
                     color={
-                        isDisabled
-                            ? 'iconDisabled'
-                            : mapVariantToIconColor({ $variant: variant, theme })
+                        theme[
```
peter-sanderson: I believe this was a bug. I think this could not have been working.

packages/components/src/components/NewModal/NewModal.stories.tsx
```diff
@@ -62,13 +67,30 @@ export const NewModal: StoryObj<NewModalProps> = {
             control: {
                 type: 'radio',
             },
-            options: ['primary', 'warning', 'destructive'],
+            options: [
+                'primary',
+                'warning',
+                'destructive',
+                undefined,
+            ] satisfies NewModalProps['variant'][],
```
peter-sanderson, line 72: We shall start using this pattern to match this list with type

packages/theme/src/types.ts
```diff
@@ -2,11 +2,5 @@ export type CSSColor =
     | `#${string}`
     | `rgb(${number}, ${number}, ${number})`
     | `rgba(${number}, ${number}, ${number}, ${number})`
-    | 'transparent';
-
-export const isCSSColor = (color: string): color is CSSColor => {
```
peter-sanderson, line 7: Not used anywhere

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -306,7 +306,7 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
 
     const getBottomTextIconComponent = () => {
         if (hasAddressChecksummed) {
-            return <Icon name="check" size="medium" color="iconDisabled" />;
```
peter-sanderson, line 309: This must have been bug.

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -158,6 +162,12 @@ const _NewModalBase = ({
         }
     });
 
+    const wrapperColorOrVariant: TransientProps<NewModalExclusiveColorOrVariant> =
+        color === undefined ? { $variant: variant ?? 'primary' } : { $color: color };
```
peter-sanderson: This if-magic is necessary to satisfy Typescript

seibei-iguchi:
I don't think the variant and color mixes - they are not really exclusive. If they were, the buttons should also be purple - but we do not want that. I'd rather see a prop like iconColor and so on, and not make it unnecessarily complicated.

peter-sanderson:
@adamhavel good point, I reworked it

packages/theme/src/types.ts
```diff
@@ -2,11 +2,5 @@ export type CSSColor =
...
-    const cssColorRegex =
-        /^#([0-9a-fA-F]{3,8})$|^rgba?\((\d{1,3}),\s*(\d{1,3}),\s*(\d{1,3}),?\s*(\d{1,3})\)$/;
-
-    return cssColorRegex.test(color);
-};
+    | 'transparent'
+    | 'currentColor';
```
jvaclavik, line 6: what is `currentColor`? Is it this one? https://www.w3schools.com/colors/colors_currentcolor.asp

jvaclavik (approved):

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -105,41 +101,33 @@ const Footer = styled.footer`
...
     alignment?: NewModalAlignment;
     size?: NewModalSize;
     'data-testid'?: string;
-};
+} & ExclusiveIconNameOrComponent;
 
 const _NewModalBase = ({
     children,
-    variant = 'primary',
+    variant,
```
seibei-iguchi, line 124: Why remove the default variant? How does this work?:)

PR #14681, opened 2024-10-02
feat(mobile): bottom sheet flashlist

### LINKED ISSUE #14503: BottomSheet refactor to better support other ScrollViews
1. From @matejkriz 

> I just faced the problem of scrolling up. It's possible, but the scrollview has to be moving, so you have to start scrolling down a bit and then scroll up, otherwise bottomsheet is closed instead 🙈 But I guess it's separate issue and minor.
[user-attachments/assets/b07c41da-856d-48d0-ae09-411d97769458](https://github.com/user-attachments/assets/b07c41da-856d-48d0-ae09-411d97769458)

2. Another issue is that when used with `isScrollable=false` it will fill nearly entire screen even if there is only few items - bottom sheet height is adapting to number of items in sheet, for example if there are only two items it will fill about 30% of screen.
3. Toast messages are hidden under BottomSheet - for example when you click Staking in Receive flow

We can either refactor BottomSheet or create some special bottom sheet for FlashList or maybe both with some shared primitives.

Nodonisko:
## Description

1. Created special version of BottomSheet for FlashList
2. Fixed some bugs related to scrolling
3. Staking toast messages are now visible when BottomSheetFlashlist is opened 
4. Some other bugfixes etc.

## Related Issue

Resolve #14503

### DIFF
--- a/suite-native/accounts/src/components/AccountSelectBottomSheet.tsx
+++ b/suite-native/accounts/src/components/AccountSelectBottomSheet.tsx
@@ -1,10 +1,8 @@
 import React, { useCallback } from 'react';
 
-import { FlashList } from '@shopify/flash-list';
-
-import { BottomSheet } from '@suite-native/atoms';
-import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
-import { useToast } from '@suite-native/toasts';
+import { BottomSheetFlashList } from '@suite-native/atoms';
+import { ToastRenderer, useToast } from '@suite-native/toasts';
+import { Translation } from '@suite-native/intl';
 
 import { AccountSelectBottomSheetSection, OnSelectAccount } from '../types';
 import { AccountsListItem } from './AccountsList/AccountsListItem';
@@ -19,9 +17,7 @@ type AccountSelectBottomSheetProps = {
     onClose: () => void;
 };
 
-const contentContainerStyle = prepareNativeStyle(utils => ({
-    paddingHorizontal: utils.spacings.sp16,
-}));
+const ESTIMATED_ITEM_SIZE = 76;
 
 export const AccountSelectBottomSheet = React.memo(
     ({
@@ -30,7 +26,6 @@ export const AccountSelectBottomSheet = React.memo(
         isStakingPressable = false,
         onClose,
     }: AccountSelectBottomSheetProps) => {
-        const { applyStyle } = useNativeStyles();
         const { showToast } = useToast();
 
         const renderItem = useCallback(
@@ -63,7 +58,9 @@ export const AccountSelectBottomSheet = React.memo(
                                     } else {
                                         showToast({
                                             variant: 'warning',
-                                            message: 'Staking is not available in this context.',
+                                            message: (
+                                                <Translation id="accountList.stakingDisabled" />
+                                            ),
                                         });
                                     }
                                 }}
@@ -93,14 +90,16 @@ export const AccountSelectBottomSheet = React.memo(
         );
 
         return (
-            <BottomSheet isVisible isCloseDisplayed={false} onClose={onClose} isScrollable={false}>
-                <FlashList
-                    data={data}
-                    renderItem={renderItem}
-                    contentContainerStyle={applyStyle(contentContainerStyle)}
-                    estimatedItemSize={76}
-                />
-            </BottomSheet>
+            <BottomSheetFlashList<AccountSelectBottomSheetSection>
+                isVisible
+                isCloseDisplayed={false}
+                onClose={onClose}
+                data={data}
+                renderItem={renderItem}
+                estimatedItemSize={ESTIMATED_ITEM_SIZE}
+                estimatedListHeight={ESTIMATED_ITEM_SIZE * data.length}
+                ExtraProvider={ToastRenderer}
+            />
         );
     },
 );
--- a/suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
+++ b/suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
@@ -1,38 +1,42 @@
-import { useEffect, useRef, useState, ReactNode } from 'react';
+import { useEffect, useRef, ReactNode } from 'react';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import Animated from 'react-native-reanimated';
-import { ScrollView, PanGestureHandler } from 'react-native-gesture-handler';
-import { GestureResponderEvent, Pressable } from 'react-native';
+import { PanGestureHandler } from 'react-native-gesture-handler';
+import { GestureResponderEvent, Pressable, Dimensions } from 'react-native';
+
+import { FlashList, FlashListProps } from '@shopify/flash-list';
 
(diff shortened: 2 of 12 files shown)


suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
```diff
@@ -0,0 +1,146 @@
...
+    ExtraProvider?: React.ComponentType;
+    estimatedListHeight?: number;
+} & FlashListProps<TItem>;
+
+const DEFAULT_INSET_BOTTOM = 50;
+
+const sheetWrapperStyle = prepareNativeStyle(utils => ({
+    backgroundColor: utils.colors.backgroundSurfaceElevation0,
+    borderTopLeftRadius: utils.borders.radii.large,
+    borderTopRightRadius: utils.borders.radii.large,
```
yanascz: This is obsolete, should use `r20` after rebasing.
```suggestion
    borderTopLeftRadius: utils.borders.radii.r20,
    borderTopRightRadius: utils.borders.radii.r20,
```

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { BottomSheetHeader } from './BottomSheetHeader';
+
+export type BottomSheetFlashListProps<TItem> = {
+    isVisible: boolean;
+    isCloseDisplayed?: boolean;
+    onClose: (isVisible: boolean) => void;
+    title?: ReactNode;
+    subtitle?: ReactNode;
+    ExtraProvider?: React.ComponentType;
+    estimatedListHeight?: number;
```
yanascz, line 23: Would it make sense to swap these two lines?
```suggestion
    estimatedListHeight?: number;
    ExtraProvider?: React.ComponentType;
```

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const handlePressOutside = (event: GestureResponderEvent) => {
+        if (event.target === event.currentTarget) closeSheetAnimated();
+    };
+
+    const insetBottom = Math.max(insets.bottom, DEFAULT_INSET_BOTTOM);
+
+    return (
+        <BottomSheetContainer
+            isVisible={isVisible}
+            onClose={handleCloseSheet}
```
yanascz: What about using `closeSheetAnimated` here directly? 🤔 
```suggestion
            onClose={closeSheetAnimated}
```

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+import { Box } from '../Box';
+import { BottomSheetContainer } from './BottomSheetContainer';
+import { useBottomSheetAnimation } from './useBottomSheetAnimation';
+import { BottomSheetHeader } from './BottomSheetHeader';
+
+export type BottomSheetFlashListProps<TItem> = {
+    isVisible: boolean;
+    isCloseDisplayed?: boolean;
+    onClose: (isVisible: boolean) => void;
```
yanascz, line 19: Do we wanna keep mandatory properties together at the beginning?

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                title={title}
+                                subtitle={subtitle}
+                                isCloseDisplayed={isCloseDisplayed}
+                                onCloseSheet={closeSheetAnimated}
+                            />
+
+                            <Box
+                                style={{
+                                    // only estimated height is used, so we need to add some buffer to prevent unnecessary scrolling, when list is shorter than estimated
+                                    height: estimatedListHeight * 1.1,
```
yanascz: I'd prefer the multiplier to be first. Not a big deal.
```suggestion
                                    height: 1.1 * estimatedListHeight,
```

suite-native/accounts/src/components/AccountSelectBottomSheet.tsx
```diff
@@ -93,14 +90,16 @@ export const AccountSelectBottomSheet = React.memo(
...
-                />
-            </BottomSheet>
+            <BottomSheetFlashList<AccountSelectBottomSheetSection>
+                isVisible
+                isCloseDisplayed={false}
+                onClose={onClose}
+                data={data}
+                renderItem={renderItem}
+                estimatedItemSize={ESTIMATED_ITEM_SIZE}
+                estimatedListHeight={ESTIMATED_ITEM_SIZE * data.length}
```
yanascz, line 100: This order makes more sense to me. Not a big deal.
```suggestion
                estimatedListHeight={data.length * ESTIMATED_ITEM_SIZE}
```

yanascz:
Is the double spacing between items supposed to be fixed in this PR?
Also the _Select Account_ title is now missing.

image

suite-native/accounts/src/components/AccountSelectBottomSheet.tsx
Nodonisko: It doesn't make sense for me :D

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
Nodonisko: I think overall consensus is that we want keep them in begging but I think it doesn't matter. In case of these code "stylystic" issue it should either have autofixable eslint rule or shoudn't be enforced.

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
Nodonisko: I don't see any benefit doing that manually, if you like it in some specific order there are ESLint plugins like this one https://github.com/azat-io/eslint-plugin-perfectionist where you can setup everything in this matter. I think it would be nice to have something like this implemented, feel free to prepare PR if you are unhappy with current state.

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
Nodonisko: Make sense

Nodonisko:
Fix for bad groups is not part of this PR, will do follow-up. `Select Account` title should it's intentionally removed according to Sheny.

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
yanascz: Why do you think it's enforced? I'm just trying to figure out what's the team consensus.

suite-native/atoms/src/Sheet/BottomSheetFlashList.tsx
yanascz: I don't think such a plugin would make it better. From my point of view, semantic sorting is the best approach and I don't think it can be done automagically. At least not yet.

Anyway, I'm not unhappy, I'm just asking what you think.

yanascz (approved):

PR #14683, opened 2024-10-02
fix(suite-native): send review old flow outputs

### LINKED ISSUE #14678: Send coins - old flow outputs (Trezor M1)
The Trezor M1 supports only FW of version <= 1.x.x. That means that the set of on-device send review outputs is constructed according to the old send flow ([see this](https://github.com/trezor/trezor-suite/blob/18eb018eba350e02573f531f5ae51b3c3b952f3f/suite-common/wallet-utils/src/reviewTransactionUtils.ts#L337)). As a result, there might be `regular_legacy` and `fee` outputs present in the review.

## Acceptance criteria
Are outputs of old flow correctly handled, and the user can do a device review using Trezor model 1.

PeKne:
## Related Issue

Resolve #14678 

## Screenshots:

https://github.com/user-attachments/assets/0bdd91cf-97f4-436e-89bb-a978b3761373

### DIFF
--- a/suite-native/module-send/src/components/ReviewOutputItem.tsx
+++ b/suite-native/module-send/src/components/ReviewOutputItem.tsx
@@ -13,28 +13,33 @@ type ReviewOutputItemProps = {
     reviewOutput: StatefulReviewOutput;
     onLayout: (event: LayoutChangeEvent) => void;
 };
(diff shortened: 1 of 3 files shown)


yanascz (approved):
Works fine! 🥳

PR #14685, opened 2024-10-02
fix(suite-native): ignore unknown URL schemes by connect popup

yanascz:
Make sure the app starts without error when running it locally.


matejkriz (approved):
makes sense to me, haven't tested it

PR #14690, opened 2024-10-02
connect - core in suite-desktop

mroz22:
So far only some preliminary research and exploration of replacing connect-popup with suite-desktop. My general idea is to merge this feature behind debug. The motivation is that there are similar UX problems to be solved for both core-in-suite-native and core-in-suite-desktop so it makes sense to me to have some playground available for both.

suite-desktop:
- in dev or when `--expose-connect-ws` flag is on, suite-desktop opens websocket on 8090 and accepts connect calls here.
- the 1st commit has support for inter-module inducted window-focus and window-blur. it probably needs some more polishing.

connect-web:
- added new imp `core-in-suite-desktop`. what is a better name? `core-in-ws-server`?
- this impl should later be available also from base `connect` package. And also from `connect-webextension`. I need to discuss this with @karliatto 
-

image

### DIFF
--- a/packages/connect-web/src/impl/core-in-suite-desktop.ts
+++ b/packages/connect-web/src/impl/core-in-suite-desktop.ts
@@ -23,8 +23,7 @@ import { createDeferred, createDeferredManager, DeferredManager } from '@trezor/
 import { parseConnectSettings } from '../connectSettings';
 
 /**
- * Base class for CoreInPopup methods for TrezorConnect factory.
- * This implementation is directly used here in connect-web, but it is also extended in connect-webextension.
+ * CoreInSuiteDesktop implementation for TrezorConnect factory.
  */
 export class CoreInSuiteDesktop implements ConnectFactoryDependencies<ConnectSettingsWeb> {
     public eventEmitter = new EventEmitter();
--- a/packages/connect-web/src/impl/core-in-suite-desktop.ts
+++ b/packages/connect-web/src/impl/core-in-suite-desktop.ts
@@ -1,0 +1,192 @@
+import EventEmitter from 'events';
+
+// NOTE: @trezor/connect part is intentionally not imported from the index so we do include the whole library.
+import {
+    IFRAME,
+    UiResponseEvent,
+    CallMethodPayload,
+    CallMethodAnyResponse,
+    POPUP,
+} from '@trezor/connect/src/events';
+import * as ERRORS from '@trezor/connect/src/constants/errors';
+import type {
+    ConnectSettings,
+    ConnectSettingsPublic,
+    ConnectSettingsWeb,
+    Manifest,
+    Response,
+} from '@trezor/connect/src/types';
+import { ConnectFactoryDependencies, factory } from '@trezor/connect/src/factory';
+import { Login } from '@trezor/connect/src/types/api/requestLogin';
+import { createDeferred, createDeferredManager, DeferredManager } from '@trezor/utils';
+
+import { parseConnectSettings } from '../connectSettings';
+
+/**
+ * Base class for CoreInPopup methods for TrezorConnect factory.
+ * This implementation is directly used here in connect-web, but it is also extended in connect-webextension.
+ */
+export class CoreInSuiteDesktop implements ConnectFactoryDependencies<ConnectSettingsWeb> {
+    public eventEmitter = new EventEmitter();
+    protected _settings: ConnectSettings;
+    private ws?: WebSocket;
+    private readonly messages: DeferredManager<CallMethodAnyResponse>;
+
+    public constructor() {
+        this._settings = parseConnectSettings();
+        this.messages = createDeferredManager();
+    }
+
+    public manifest(data: Manifest) {
+        this._settings = parseConnectSettings({
+            ...this._settings,
+            manifest: data,
+        });
+    }
+
+    public dispose() {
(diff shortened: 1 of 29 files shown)


packages/suite-desktop-core/src/app.ts
```diff
@@ -195,6 +195,25 @@ const initUi = async ({
...
+        app.dock?.show();
+        if (isMacOs()) app.show();
+        if (!mainWindow.isVisible()) mainWindow.show();
+        if (mainWindow.isMinimized()) mainWindow.restore();
+        mainWindow.focus();
+    });
+
+    mainThreadEmitter.on('blur-window', () => {
+        app.hide();
+    });
```
mroz22: one preliminary question - is it the correct way to achieve inter-module focus-window and blur-window? if yes, then of course code duplication needs to be solved. 

cc @marekrjpolak, @martykan

martykan: We already have handlers in `window-controls` that respond to ipcMain `app/hide` and `app/focus`

martykan: However it can only focus an existing window, not recreate it

packages/connect-web/src/impl/core-in-suite-desktop.ts
```diff
@@ -0,0 +1,172 @@
...
+     * 2. sends request to popup where the request is handled by core
+     * 3. returns response
+     */
+    public async call(params: CallMethodPayload): Promise<CallMethodAnyResponse> {
+        try {
+            await this.handshake();
+
+            this.ws?.send(
+                JSON.stringify({
+                    type: IFRAME.CALL,
```
karliatto, line 139: Maybe using ids for messages as we use in `iframe` and new core-in-module is a good idea https://github.com/trezor/trezor-suite/blob/develop/packages/connect-web/src/impl/core-in-iframe.ts#L258

mroz22: yes. I think we already have couple of places where we use the `send and listen for event` communication mechanism. It would be nice to find some common abstraction for it

martykan:
Current state:
- Suite runs Websocket server at http://localhost:2335/connect-ws
- New Core mode in Connect (`suite-desktop`) that communicates with said WS
- Modal that shows info about the method being called, needs to be confirmed every time
- Security info about the process that created the connection + web origin header
Screenshot 2024-11-18 at 14 10 38

Missing features:
- Support better device modals (refactor modal state)
    - ConfirmAddressModal
    - ConfirmXpubModal
    - TransactionReviewModal
- Special methods with UI requests
	- getAccountInfo
    	- SELECT_ACCOUNT
	- composeTransaction
	    - SELECT_FEE, UPDATE_CUSTOM_FEE, INSUFFICIENT_FUNDS
- Permissions handling
  - Keep using internal Connect logic - but need different settings → Second instance of connect? Different API
  - Or skip internal Connect logic and use wrapper?

packages/connect-web/src/impl/core-in-suite-desktop.ts
```diff
@@ -0,0 +1,192 @@
...
+    Response,
+} from '@trezor/connect/src/types';
+import { ConnectFactoryDependencies, factory } from '@trezor/connect/src/factory';
+import { Login } from '@trezor/connect/src/types/api/requestLogin';
+import { createDeferred, createDeferredManager, DeferredManager } from '@trezor/utils';
+
+import { parseConnectSettings } from '../connectSettings';
+
+/**
+ * Base class for CoreInPopup methods for TrezorConnect factory.
```
mroz22: comment mentioning CoreInPopup. this might be outdated

packages/connect-web/src/impl/core-in-suite-desktop.ts
```diff
@@ -0,0 +1,192 @@
...
+                type: POPUP.HANDSHAKE,
+            }),
+        );
+        try {
+            await promise;
+        } catch (err) {
+            console.error(err);
+            throw new Error('Handshake timed out');
+        }
+    }
```
mroz22, line 70: is this handshake actually waiting for some response?

packages/connect/src/constants/errors.ts
```diff
@@ -78,6 +78,9 @@ export const serializeError = (payload: any) => {
     if (payload && payload.error instanceof Error) {
         return { error: payload.error.message, code: payload.error.code };
     }
+    if (payload instanceof TrezorError) {
+        return { error: payload.message, code: payload.code };
+    }
```
mroz22, line 83: hmm why is this needed? 🤔

martykan: It helps here
https://github.com/trezor/trezor-suite/pull/14690/files/b1630f7dfba7a66977e205e605b1c41372282c97#diff-2e9d057f0bfe2cc92fe50d4ce28838622d9e79fcca010ab8847a0fa288da13fdR245
For some reason the original method serializes payload.error, not the actual error

packages/connect-web/src/impl/core-in-suite-desktop.ts
martykan: Yeah, the response is received by `this.ws.addEventListener('message', ...)` like all other messages

mroz22 (commented):
# 🦭 ✅

martykan (approved):

PR #14701, opened 2024-10-03
fix(suite): HiddenPlaceholder onHover when not discreet mode

Lemonexe:
## Description

Fix oversight from #14528 
- when **not** in discreet mode, turn off the "keeping width behavior" altogether.
- refactor Table so that `colWidths` pro _enforces_ the width rather than just _preferring_ it

These two may seem unrelated, but are related to `TokensTable`

### DIFF
--- a/packages/suite/src/components/suite/HiddenPlaceholder.tsx
+++ b/packages/suite/src/components/suite/HiddenPlaceholder.tsx
@@ -8,7 +8,6 @@ type WrapperProps = {
     $intensity: number;
     $discreetMode: boolean;
     $minWidth?: number;
-    $disableKeepingWidth?: boolean;
 };
 
 const Wrapper = styled.span<WrapperProps>`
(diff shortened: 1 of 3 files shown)


packages/suite/src/components/suite/HiddenPlaceholder.tsx
```diff
@@ -95,14 +93,15 @@ export const HiddenPlaceholder = ({
         setWrapperMinWidth(undefined);
     };
 
+    const shouldEnforceMinWidth = discreetMode && !disableKeepingWidth;
```
Lemonexe, line 96: Rather than making the condition in Wrapper more complex, I made it slightly dumber, and instead moved _decisions_ here, and Wrapper is now concerned with only properties that directly translate to CSS.

seibei-iguchi (approved):

jvaclavik (approved):

PR #14717, opened 2024-10-03
feat(connect): dispatch call in progress event - needed for immediate…

mroz22:
1. introduce a new connect `"call in progress"` event
2. use this event to set `suite.locks[device lock]` in suite and get rid of the whitelist of wrapped methods. From now on, the lock is set whenever the connect method that is called works with a device and released when this method posts response.

cherry-picked from #14690


suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -45,7 +46,9 @@ export const connectInitThunk = createThunk(
         });
 
         TrezorConnect.on(UI_EVENT, ({ event: _, ...action }) => {
-            // dispatch event as action
+            if (action.type === 'ui-call_in_progress') {
```
mroz22, line 46: Hmmm maybe it should be `DEVICE_EVENT` 🤔

mroz22:
not finished, to be discussed, I'd like to nominate @marekrjpolak 🎖️

suite-common/connect-init/src/connectInitThunks.ts
marekrjpolak: I agree that `DEVICE_EVENT` sounds a bit better.

packages/connect/src/core/index.ts
```diff
@@ -518,7 +517,13 @@ const onCall = async (context: CoreContext, message: IFrameCallMessage) => {
...
     }
 
-    return await onCallDevice(context, message, method);
+    sendCoreMessage(createUiMessage(UI.CALL_IN_PROGRESS, { value: true }));
+
+    try {
+        return await onCallDevice(context, message, method);
+    } finally {
+        sendCoreMessage(createUiMessage(UI.CALL_IN_PROGRESS, { value: false }));
+    }
```
marekrjpolak: I believe that the lock should be removed at all cases, even when the method throws or returns prematurely. So I proposed a fixup.

suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -61,50 +64,26 @@ export const connectInitThunk = createThunk(
...
             // typescript complains about params and return type, need to be "any"
             const original: any = TrezorConnect[key];
             if (!original) return;
             (TrezorConnect[key] as any) = async (params: any) => {
-                dispatch(lockDevice(true));
-                const result = await synchronize(() => original(params));
-                dispatch(lockDevice(false));
+                if (selectIsDeviceLocked(getState())) {
+                    return await synchronize(() => original(params));
+                }
```
marekrjpolak: I'm afraid this won't work as expected. When the first locking call is executed, there's no device lock yet so it would be out of `synchronize()` and the next call in `synchronize()` will run in parallel.

There should be something like `when locked is true, wait until locked is false and go`? But when you accidentally start two methods at once, you're still fucked.

Maybe, if we want the lock to be unpenetrable, it has to be defined at the same place as consumed? Meaning that if you want to await the lock from Connect in Suite, it will always be prone to race condition?

suite-common/connect-init/src/connectInitThunks.ts
marekrjpolak: Also, I thought it's more obvious not to edit Connect events here in `connectInitThunks` but pass them directly as redux actions whenever possible. So I tried to catch this event in Suite middleware and transform it to `SUITE.LOCK_DEVICE` there, but then it probably wouldn't work in native, because they don't use the same middlewares? So I'm not sure about that now.

suite-common/connect-init/src/connectInitThunks.ts
mroz22: Sounds right. Maybe I can sort of go back to the previous behaviour, but still more robust simply by defining a new utility method on connect, something like 
```
TrezorConnect.getMethods
```
Which would return method details, in this case I would be interested in all methods that have `method.useDevice=true` and create a wrapper for them.

suite-common/connect-init/src/connectInitThunks.ts
mroz22: Sure, middleware sounds like a place to put it. I just don't like to use middleware if there is only a single use of something. In that case, it feels more readable without it. 

🤔 but there are also `ui-*` device-related events

```
'ui-device_needs_backup'
'ui-request_passphrase_on_device'
'ui-request_pin'
```

marekrjpolak: > I just don't like to use middleware if there is only a single use of something

Yes, I agree, that's why my first idea was to directly replace `SUITE.LOCK_DEVICE` with `UI.CALL_IN_PROGRESS` everywhere. But that would be a bigger change.

suite-common/connect-init/src/connectInitThunks.ts
marekrjpolak: Yeah, that would be an improvement.

And why do we even need these locks on Suite side? If it is impossible to do two calls in parallel because of Connect, shouldn't Connect handle it? Maybe we can add a new general method param, something like `allowQueue`, which will cause synchronization on Connect's side instead of throwing call in progress?

Issue #11921, opened 2024-04-08
Design refactoring: Padding, Spacing, Radius


shenkys:
As of now, we've got some mismatch between code & design. 

So far discovered 

Radius Small = 12px in design, 8px in code
Radius Large = 20px in design, 24px in code 

Radius used on (almost) all cards:
Design = radius-lg (20px)
Code = radius-md (16px)

There's definitely more to it so....

....we should dedicate a day or two with dev to: 

Check for mismatching values in already existing tokens 
- Check for any possible breaking points (e.g. radius value used for different padding or some other )

Check for mismatching usage of these tokens (e.g. the cards) & and align on it long term  
- Card radii & padding
- Bottomsheet radii & padding
- etc.

This will help us with: 
- Design handoff (won't need to anotate everything as we'll have a reliable source of truth)
- Programming (won't have to use hard-coded values) 
- Visual language (we'll do one more step for unification & any future changes can be done from 1 place)

vytick:
I agree, that unification of design tokens and values used in the code would be beneficial in order to speak the same language. Not only in case of radius, but also other properties (spacings, paddings, ...)

Also the design and app current state is not in real sync now partially because of this.

yanascz:
QA, would be great to check that nothing is fundamentally broken.

STew790:
**QA OK**
Haven't found anything broken but will definitely monitor this before release.

**Info**
24.11.1 https://github.com/trezor/trezor-suite/commit/ff16b788c576e89bbbfc17cbe783154223440336

PR #14720, opened 2024-10-04
fix(suite-native): use 20px border radius instead of 24px

yanascz:
20px border radius is the biggest one available according to design.

## Related Issue

#11921

### DIFF
--- a/suite-native/module-authorize-device/src/components/passphrase/PassphraseForm.tsx
+++ b/suite-native/module-authorize-device/src/components/passphrase/PassphraseForm.tsx
@@ -42,12 +42,6 @@ type PassphraseFormProps = {
     noPassphraseEnabled?: boolean;
 };
 
-const formStyle = prepareNativeStyle(utils => ({
-    backgroundColor: utils.colors.backgroundSurfaceElevation1,
-    borderRadius: utils.borders.radii.r24,
-    gap: utils.spacings.sp16,
-}));
-
 const cardStyle = prepareNativeStyle(_ => ({
     padding: FORM_CARD_PADDING,
 }));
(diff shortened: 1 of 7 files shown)


suite-native/module-authorize-device/src/components/passphrase/PassphraseForm.tsx
```diff
@@ -44,7 +44,6 @@ type PassphraseFormProps = {
 
 const formStyle = prepareNativeStyle(utils => ({
     backgroundColor: utils.colors.backgroundSurfaceElevation1,
-    borderRadius: utils.borders.radii.r24,
```
yanascz, line 47: There doesn't seem to be any visible border so it doesn't make sense here. Correct me if I'm wrong. 🙏

Nodonisko: You don't need border to make something rounded, background is enough

yanascz: I understand that, I'm saying there's no visible effect of the radius change. I believe even the background colour is redundant here.

PeKne: seems like that

yanascz: This is the area we're talking about:
image

yanascz: Removed the form style altogether. 1412384

PeKne (approved):

Issue #14480, opened 2024-09-23
Reset storage button in Settings/Application does not do anything


bosomt:
**Describe the bug**
Reset storage button in Settings/Application does not do anything.
You only get `Storage cleared` notification but nothing is actually reset.

**Steps to reproduce:**
1. Navigate to `Settings/Application`
2. Press `Reset app`
3. Observe that nothing is reset

**Info:**
 - Suite version: desktop 24.10.0 (0c147a1d37dc615e286a114d90bc8b637bbb929f)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.10.0 Chrome/124.0.6367.243 Electron/30.3.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.3 regular (revision 7f373ae71eca855dd41b4a0fdcc7cadaa13a8281)
 - Transport: BridgeTransport 2.0.33

**Screenshots:**
<img width="1421" alt="Screenshot 2024-09-23 at 14 32 13" src="https://github.com/user-attachments/assets/66cdd167-25dc-404a-b4f9-c1f774edf4b3">

https://github.com/user-attachments/assets/51aa2f03-ed01-4a0f-b93f-47f87fce29be

evgenysl:
**Arm Mac** I have the same issue:
1. Reset app button
2. Trezor Suite minimizes into Dock
3. Wallets are not removed

On **Windows** Reset app works as expected. 
Trezor Suite fully restarts and wallets are removed.

bosomt:
same results for `Settings/Debug/Wipe data` 
im not able to quit app after i try to reset/wipe it 

looks like every reset spawns one more binary in memory

<img width="825" alt="image" src="https://github.com/user-attachments/assets/50e6f986-539e-4a8c-8ccc-676ea19dfc29">

komret:
Broken by https://github.com/trezor/trezor-suite/commit/f3a11183d459a1d8974d11d0d891aec1fb6a1c76 👀 @martykan

mroz22:
On my machine, it works nicely provided I don't have the autostart feature enabled. To find out whether you have it enabled or not, please consult this issue too #14721

[...]

PR #14722, opened 2024-10-04
fix(suite): autostart control (debug): do not store in redux, use ele…

### LINKED ISSUE #14721: autostart changed in OS setting not reflected
1. go to debug 
2. turn on autostart
3. go somewhere else
4. turn off autostart in system settings 
5. go back to debug menu - autostart still appears to be on, should be off

image

image

mroz22:
A side quest taken while looking into #14480. 

fixes #14721


martykan:
Good change, but needs an exception for Linux, since Electron doesn't handle autostart there.
We can do the check based on the existence of the autostart file.

mroz22:
something like this? ea1aaf1bda7092b426bd8c9c89e5a23a21f4453a

martykan (approved):

martykan (approved):

PR #14723, opened 2024-10-04
feat: Refactor Update Flow into NotificationBanener and add support for Silent Update

peter-sanderson:
Adds a popover with little modal-like notification  to inform user about update status.

Part of https://github.com/trezor/trezor-suite/issues/13990

imageimage


komret (approved):

packages/suite/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@trezor/suite",
-    "suiteVersion": "24.11.0",
+    "suiteVersion": "0.0.0",
```
peter-sanderson: !!!!

komret (approved):

PR #14726, opened 2024-10-04
Multitransport prerequisites

### LINKED ISSUE #15062: multitransport part 1
implement multitransport prerequisites 

## QA 🙏 

this is touching, again, everything related to all transports
- multi devices
- session stealing
- multi tabs/apps
- bridge/node-bride/webusb

marekrjpolak:
## Description

Refactoring (mostly) of abstract transport class (in `@trezor/transport`) and DeviceList (in `@trezor/connect`) so they're hopefully a bit more resilient and better prepared for multitransport support.

## Commits

- https://github.com/trezor/trezor-suite/pull/14726/commits/8d6270bc3fa2d766d1590317bc55cbf0ff49f15c - split `TRANSPORT.UPDATE` event per separate descriptors
- https://github.com/trezor/trezor-suite/pull/14726/commits/a78a9e3c20f7f9bfd2371fda46e3073c2f8edd14 - extract device creation locking into a separate util (but it will be removed later anyway)
- https://github.com/trezor/trezor-suite/pull/14726/commits/b705e9d2391adecf3b7be2b3a6f34a908d0b7920 - device-related methods on `DeviceList` improved so they better suit their real usage
- https://github.com/trezor/trezor-suite/pull/14726/commits/9f9ab9d4e24eac77a791fc485a92f34626bdcfce - stop using device changed event in favor of acquired/released events (will be changed later anyway)
- https://github.com/trezor/trezor-suite/pull/14726/commits/7e30624780251f366d480c46b446c771c1787fc1 - all the Device fields accessibility restricted as much as possible (potentially split into getter/setter), so it's easier to handle them
- https://github.com/trezor/trezor-suite/pull/14726/commits/7007e81aa4368a84fbcd1fcc3db4753897118b61 - **BIG ONE** - devices are now identified by randomly-generated unique ids (`DeviceUniquePath` branded type) instead of real transport paths, which will be necessary once there's a device accessible by multiple transports
- https://github.com/trezor/trezor-suite/pull/14726/commits/d5fee829c1585b603d8ba5ea4440389f4639986b - **BIGGEST ONE** - complete rework of device handling on transport/connect boundary
  - a bond between `acquire`/`release`/`listen` transport calls (e.g. postponing listen responses incoming during acquire after the acquire etc.) removed from abstract transport; it's handled in Connect's Device object instead where it's much easier to implement proper locks/promises
  - `Device` object is created in `DeviceList` as soon as first seen so it can be used for handling asynchronous cases without temporarily stored info (e.g. `creatingDevicesDescriptors`); that's why the constructor was simplified (there is only one variant now)
  - complex device-related logic (as the one above) moved into `Device` class where it belongs -> device lifecycle events emitted from there, acquiring, retrying to acquire and failing to acquire handled there, waiting for incoming descriptors while acquiring/releasing handled there...
  - added new `releaseRequest` event into sessions handling -> it was needed so the sessions are **not** changed during acquire intent while it's still possible to signal that some device should be released by others
  - `TRANSPORT.UPDATE` event divided even more into connected/disconnected/changed events
  - device creation locking extensively simplified so there's only one very simple `handshakeLock` now
- https://github.com/trezor/trezor-suite/pull/14726/commits/437f04a9e2d7cfe9b468c6beecf12f291ad902b0 - all the types of tests adjusted to the new changes

I apologize for not being able to make the commits clearer as I usually strive for, but the changes were too big and chaotic. Also, taxation is theft, which may seem unrelated, but it's pretty important anyway.

resolve #15062

### DIFF
--- a/packages/transport/tests/abstractUsb.test.ts
+++ b/packages/transport/tests/abstractUsb.test.ts
@@ -113,7 +113,12 @@ describe('Usb', () => {
         it('handleDescriptorsChange', async () => {
             const { transport } = await initTest();
             const spy = jest.fn();
-            transport.on('transport-update', spy);
+            transport.on('transport-device_connected', descriptor =>
+                spy({ type: 'connected', descriptor }),
+            );
+            transport.on('transport-device_disconnected', descriptor =>
+                spy({ type: 'disconnected', descriptor }),
+            );
 
             transport.handleDescriptorsChange([{ path: PathPublic('1'), session: null, type: 1 }]);
 
@@ -156,7 +161,9 @@ describe('Usb', () => {
             const { transport } = await initTest();
             jest.useFakeTimers();
             const spy = jest.fn();
-            transport.on('transport-update', spy);
+            transport.on('transport-device_connected', spy);
+            transport.on('transport-device_disconnected', spy);
+            transport.on('transport-device_session_changed', spy);
 
             await transport.enumerate();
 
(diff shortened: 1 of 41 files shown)


packages/transport/tests/abstractUsb.test.ts
```diff
@@ -171,61 +180,6 @@ describe('Usb', () => {
             expect(spy).toHaveBeenCalledTimes(0);
         });
 
-        it('acquire. transport listening. missing descriptor', async () => {
```
martykan, line 174: Are these test cases no longer relevant?

marekrjpolak: Well, not on this level, as transport now considers `acquire`/`release` calls and incoming `listen` responses completely independent things, so e.g. acquire is successful even if session in the next listen response differs.

mroz22 (approved):

PR #14732, opened 2024-10-04
fix(suite-native): enable to confirm receive address outside our app

### LINKED ISSUE #14716: Confirming recipient address outside Trezor app freezes Trezor
When sending BTC, switching to another application to review the presented recipient address and pressing confirm whilst outside of Trezor Suite freezes the Trezor

1. Send BTC
2. See recipient address
3. Switch to another application
4. Press confirm

https://github.com/user-attachments/assets/89669e83-6415-49a3-b5a4-1c85a58cbd33

matejkriz:
## Description

- We are closing device connection on going to background, so we have to reopen it once needed. We can open the device even on background. But Suite has to be still running, not killed.
- Closing all the devices OnActivityEntersBackground is a good thing because we want to allow other apps (like browser) to work with the Trezor too and it is a good practice to frees up system resources.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14716

### DIFF
--- a/packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
+++ b/packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
@@ -150,7 +150,15 @@ class ReactNativeUsbModule : Module() {
             for (device in devicesList) {
                 if (usbManager.hasPermission(device)) {
                     Log.d(LOG_TAG, "Has permission, send event onDeviceConnected: $device")
-                    val webUsbDevice = openDevice(device.deviceName)
+                    
+                    val webUsbDevice = if (hasOpenedConnection(device.deviceName)) {
+                        Log.d(LOG_TAG, "Device already opened: $device")
+                        getWebUSBDevice(device)
+                    } else {
+                        Log.d(LOG_TAG, "Opening device: $device")
+                        openDevice(device.deviceName)
+                    }
+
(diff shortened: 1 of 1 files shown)


packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
```diff
@@ -150,7 +150,8 @@ class ReactNativeUsbModule : Module() {
             for (device in devicesList) {
                 if (usbManager.hasPermission(device)) {
                     Log.d(LOG_TAG, "Has permission, send event onDeviceConnected: $device")
-                    val webUsbDevice = openDevice(device.deviceName)
+                    openDevice(device.deviceName)
+                    val webUsbDevice = getWebUSBDevice(device)
```
Nodonisko: coudn't these two lines be replaced with `getOpenedConnection` ?

matejkriz: We need `WebUSBDevice` here, and `getOpenedConnection` returns `UsbDeviceConnection`, so I don't see it as a replacement.

matejkriz:
image


🤖 Open this link on your Android devices (or scan the QR code) to install the app:
https://expo.dev/accounts/trezorcompany-develop/projects/trezor-suite-preview/builds/e4689da6-ee10-4f83-953c-ea8576b11ef2

Nodonisko (approved):

PR #14740, opened 2024-10-07
wip: core-in-whatever

mroz22:
extremely unfinished draft

prereq of #14690

we are a little step from having single nice and clean standardized logic for creating client-facing implementations of connect core (that have additional powers such  as impl switching) in pretty much any environment. once this is finished it could help us with: 
- deeplink
- connect in suite desktop (#14690)
- connect-web as a module (getting rid of iframe from suite-web if we want to)

### DIFF
--- a/packages/connect/src/utils/assetUtils.ts
+++ b/packages/connect/src/utils/assetUtils.ts
@@ -12,7 +12,7 @@ const firmwareAssets: Record<DeviceModelInternal, NodeRequire> = {
     [DeviceModelInternal.T3T1]: require('@trezor/connect-common/files/firmware/t3t1/releases.json'),
 };
 
-export const getAssetByUrl = (url: string) => {
+export const tryLocalAssetRequire = (url: string) => {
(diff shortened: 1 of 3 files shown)


packages/connect/src/impl/core-in-module.ts
```diff
@@ -0,0 +1,272 @@
+import EventEmitter from 'events';
```
mroz22: this new impl should be implemented in 

`connect/src/index`

and also it can be used in connect-web package that is implemented in suite-web to get rid of iframe.

packages/connect/src/impl/dynamic.ts
```diff
@@ -0,0 +1,153 @@
+import EventEmitter from 'events';
```
mroz22: just moved from connect-web

karliatto:
I have moved part of this PR into https://github.com/trezor/trezor-suite/pull/14740

So we simplify with the dynamic connect. And later on there can be a follow-up for core-in-module for suite-web.

PR #13542, opened 2024-07-29
Feat: TrezorHostProtocol

szymonlesisz:
heavy draft

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/thp&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/thp&lookback=7)

PR #14752, opened 2024-10-07
Feat(protobuf): build messages.json using javascript

szymonlesisz:
## Description

Rework protbuf json build and use typescript instead of shell.

- removed unused external imports `google/protobuf/descriptor.proto`
  protobuf package does [not rely ](https://github.com/protobufjs/protobuf.js/blob/master/README.md#compatibility) on it and so do we. not needed in runtime.
  
- removed unused options: `deprecated`, `(has_bitcoin_only_values)`, `(experimental_message)`, `(experimental_field)`
  current (develop) shell script removes options `packed` from several places  and leaves others (type uint32, uint64) 🤔 this option is **required** in runtime at least for uint32/uint64
  
Build script can be used directly as standalone script or imported as module
useful for building different sets like [thp](https://github.com/trezor/trezor-suite/pull/13542)
```
tsx ./packages/protobuf/scripts/protobuf-definitions.ts /path/to/proto-files --only=bitcoin
tsx ./packages/protobuf/scripts/protobuf-definitions.ts /path/to/proto-files --skip=monero
```

we can extend it in the future to build different sets for different coins/networks and load them dynamically when needed:
['bitcoin-only', 'ethereum-only' ...etc]


packages/protobuf/scripts/protobuf-definitions.ts
```diff
@@ -0,0 +1,156 @@
...
+    // remove unused "options"
+    const ignoreOptions = ['(experimental_field)', '(experimental_message)', '(has_bitcoin_only_values)', 'deprecated'];
+
+    if (def.options) {
+        const options = Object.keys(def.options);
+        if (def.type && options.includes('packed')) {
+            const obj = root.lookup(def.type);
+            if (obj) {
+                try {
+                    // remove "packed" option from custom types
```
szymonlesisz: cc @mroz22 - compatibility with previous json, otherwise it will add option.packed = false

mroz22 (approved):
looks good, 
yarn workspace @trezor/protobuf update:protobuf works on my machine

szymonlesisz:
i did some extra work and 've removed unused options definitions which extends `google.protobuf.*` + valuesOptions pointing to them

mroz22 (approved):

PR #14766, opened 2024-10-08
feat: FW hash check automatically

### LINKED ISSUE #8489: Firmware hash check every time the device is plugged in
Currently we have in debug menu hidden possibility to check FW authenticity but it requires downloading the binary, that is why it's not enabled by default.

https://satoshilabs.slack.com/archives/C03D2GGJ8G0/p1684221988914369 
An idea of how to do the firmware hash check every time the device is plugged in, without having to keep a downloaded version of the firmware:

- Every day our backend will generate a random challenge and compute the firmware hash for each firmware version released after May 2022 (older ones don't support the feature).
- The challenge and the list of firmware versions and hashes will be placed in something similar to current secure messaging system.
- Suite will use the challenge from there to execute a firmware hash every time the device is plugged in and it will compare the result with the reference result.

Both the challenge and the responses can be truncated to small values. For example as little as 4 bytes for the random challenge and the first 2 bytes for the hash result seems acceptable. (We can double these numbers just to be on the safe side.)

Lemonexe:
## Description

:eyes:  CR commit by commit recommended

- c0e7783c3f1024a0b517cf7c7aeeeb62601726d5 Remove previous experimental implmentation of automatic FW hash check
  - that was turned off by default, turned on secretly in debug settings, it displayed only a notification
  - and it was done through suite instead of connect, which is not ideal
- 0d66d0180f1de93896a2350bfae485e59904b927 Automatically do FW hash check in connect, when device is initialized, similarly to existing revision check
  - only do this, if a new optional setting `binFilesBaseUrl` is passed by suite to connect
- [3rd + 4th](https://github.com/trezor/trezor-suite/pull/14766/files/0d66d0180f1de93896a2350bfae485e59904b927..9ea46715e0b0eb0794af04535cfff5b26514ef50) Pick up the result in Suite, and incorporate the new hash check & existing revision into single UI
  - any of those errors will display compromised device modal (can be dismissed)
  - then a banner displaying specific error message
  - received buttons are disabled for either error
  - existing settings button now turns **both off and on** simultaneously (via modal)
  - and hash check can be granularly disabled universally via message system, same as firmware check
    - if you disable one of them, the normal settings toggle will indicate it's off, inviting you to turn it all on
- 568f37bd65607d1bf82cfb9a16317bfadcec2aa2 fix E2E tests which use T1B1 by disabling the check entirely (tenv cannot provide correct hash check)
- 20d337ff30f1f7bcbb640f314d7df295396caac3 add granular controls to Debug settings to turn off either hash or firmware

:arrow_right: TODO in subsequent PRs:
- suite: handle the case if suite firmware is newer than suite (not urgent, as it's not relevant till next-next-release)
- suite: find a smarter way for `connectInitThunks.ts` (I started it, and it affects cca 10 files so I'd rather not bloat this)  
- connect: abstract the logic that calculates hash, challenges device and compares it, to deduplicate code
  - connect calls which receive `baseUrl` should use the settings baseUrl by default, and have the params optional (so as not to break API)
- discuss [new messages](https://github.com/trezor/trezor-suite/pull/14766/files#diff-988baecac61429698981f7b86b979d1627f33f484918f65402c4b345b5c57cba) with product, and have official text added into Crowdin

## Related Issue

Resolve #8489

Resolve https://github.com/trezor/trezor-suite-private/issues/100

## Screenshots:

Demo of granular debug toggle vs. main toggle.
[granular debug settings.webm](https://github.com/user-attachments/assets/0e551b53-8f55-4fec-b5ed-a61c679d0db1)
 

FW hash check failures with T1B1 & tenv (which can't pass hash check, so it simulates hacked device):
- 1.10.5 :warning: doesn't support hash check, so doesn't fail, but you get warned to update 
- 1.12.1 :x: hash check mismatch, but revision check is fine :heavy_check_mark: 
- 1-main :x: version unrecognized (no binary can be loaded), and also fails fw revision check :x:

[T1B1 various failures.webm](https://github.com/user-attachments/assets/3f0897e3-1cad-4a05-a45d-a06ad450bc7a)

devices other than T1B1 should pass (I tested both physical & tenv), because the check is not done :heavy_check_mark: 

physical T1B1 should pass (I tested) :heavy_check_mark:


packages/connect/src/device/Device.ts
```diff
@@ -567,6 +574,10 @@ export class Device extends TypedEmitter<DeviceEvents> {
         const { message } = await this.getCommands().typedCall('GetFeatures', 'Features', {});
         this._updateFeatures(message);
 
+        if (this.authenticityChecks.firmwareHash === null) {
+            this.authenticityChecks.firmwareHash = await this.checkFirmwareHash();
+        }
+
```
mroz22, line 580: not a good place, it would get too many invocations. not sure what is the best place but for example `initialize` is surely better

packages/connect/src/device/Device.ts
```diff
@@ -595,6 +606,47 @@ export class Device extends TypedEmitter<DeviceEvents> {
         }
     }
 
+    async checkFirmwareHash(): Promise<FirmwareHashCheckResult> {
+        const createFailResult = (error: FirmwareHashCheckError) => ({ success: false, error });
+
+        // optional setting, see types/settings.ts
+        const baseUrl = DataManager.getSettings('binFilesBaseUrl');
```
mroz22, line 618: dependency on external module. maybe we want to get rid of those slowly? not sure.

packages/connect/src/device/Device.ts
```diff
@@ -595,6 +606,47 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
+        if (!baseUrl) return createFailResult('binary-unavailable');
+
+        const checkSupported = this.atLeast(FW_HASH_SUPPORTED_VERSIONS);
```
mroz22: if `FW_HASH_SUPPORTED_VERSIONS` is used only here, I would move definition to this place.

packages/connect/src/types/settings.ts
```diff
@@ -34,6 +34,8 @@ export interface ConnectSettingsPublic {
     deeplinkOpen?: (url: string) => void;
     deeplinkCallbackUrl?: string;
     deeplinkUrl?: string;
+    // URL for binary files such as firmware, may be local or remote. If not provided, firmware hash check won't be done.
+    binFilesBaseUrl?: string;
```
mroz22, line 38: this is basically `baseUrl` which is part of some calls right? do these files need to be bin?

packages/connect/src/device/Device.ts
Lemonexe: What do you mean by external? `DataManager` is part of `connect` :thinking: 
And I'm not yet very familiar with `connect`, so I can't really say what's desired :smile:

packages/connect/src/device/Device.ts
Lemonexe: It's also used in `onCallFirmwareUpdate`, where I extracted it from, so I can't place it in this file, because it'd create circular dependency. Placing it in `onCallFirmwareUpdate` would work, but does not make much sense semantically.
:bulb:  But I noticed there is `src/constants/` folder, so I created `src/constants/firmware.ts`. Looks good?

packages/connect/src/types/settings.ts
Lemonexe: yes. But because the fw hash check is to be done before discovery, I placed it in `getFeatures()`, and so I need to supply it as `connect` setting.
This way it _also_ fulfills another purpose - if **not** provided, fw hash check will be skipped, which, AFAIK, is essential for 3rd party users of `connect`. We want it only for suite desktop & web.
- on desktop (dev) it points to `./packages/suite-desktop/build/static/bin/`
- on web it points to `/static/connect/data`

packages/suite/src/components/suite/SecurityCheck/DeviceCompromisedHashCheck.tsx
```diff
@@ -0,0 +1,58 @@
...
+                `Firmware hash check failed! ${JSON.stringify(contextData)}`,
+                scope,
+            );
+        });
+    }, [authenticityError, revision, vendor, version]);
+
+    return (
+        <WelcomeLayout>
+            <Card data-testid="@device-compromised/hash-check">
+                {/* TODO #14766 REMOVE ME ! */}
```
Lemonexe, line 49: :warning: :construction:

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/FirmwareHashOptOutModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    <NewModal.Button variant="tertiary" onClick={onCancel}>
+                        <Translation id="TR_CANCEL" />
+                    </NewModal.Button>
+                </>
+            }
+            variant="warning"
+        >
+            <H3>
+                <Translation id="TR_DEVICE_FIRMWARE_REVISION_CHECK_TITLE" />
+                {/* TODO #14766 REMOVE ME ! */}
```
Lemonexe, line 45: :warning: :construction:

packages/suite/src/components/suite/banners/SuiteBanners/FirmwareHashCheckBanner.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { HELP_CENTER_FIRMWARE_REVISION_CHECK } from '@trezor/urls';
+
+import { Translation, TrezorLink } from 'src/components/suite';
+import { useSelector } from 'src/hooks/suite';
+
+const messages: Record<
+    Exclude<FirmwareHashCheckError, 'check-skipped'>,
+    TranslationKey | string
+> = {
+    // TODO #14766 actual translation strings
```
Lemonexe, line 15: :warning: :construction:

packages/connect/src/device/Device.ts
```diff
@@ -567,6 +573,10 @@ export class Device extends TypedEmitter<DeviceEvents> {
         const { message } = await this.getCommands().typedCall('GetFeatures', 'Features', {});
         this._updateFeatures(message);
 
+        if (this.authenticityChecks.firmwareHash === null) {
+            this.authenticityChecks.firmwareHash = await this.checkFirmwareHash();
```
Lemonexe, line 577: now it's boilerplate - **if** we decide this is the way to go, I'll refactor the common parts of this function in subsequent PR

packages/suite/src/components/suite/SecurityCheck/DeviceCompromisedHashCheck.tsx
Lemonexe, line 1: copied from `DeviceCompromisedRevisionCheck.tsx`

packages/suite/src/components/suite/banners/SuiteBanners/FirmwareHashCheckBanner.tsx
Lemonexe, line 1: copied from `FirmwareRevisionCheckBanner.tsx`

packages/suite/src/components/suite/banners/SuiteBanners/FirmwareHashCheckBanner.tsx
```diff
@@ -0,0 +1,50 @@
+import { TranslationKey } from '@suite-common/intl-types';
```
Lemonexe, line 1: copied from `FirmwareRevisionCheckBanner.tsx`

packages/suite/src/components/suite/SecurityCheck/DeviceCompromisedHashCheck.tsx
```diff
@@ -0,0 +1,58 @@
+import { useEffect } from 'react';
```
Lemonexe, line 1: copied from `DeviceCompromisedRevisionCheck.tsx`

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/FirmwareHashOptOutModal.tsx
```diff
@@ -0,0 +1,69 @@
+import { useState } from 'react';
```
Lemonexe, line 1: copied from `FirmwareRevisionOptOutModal.tsx`

packages/suite/src/views/settings/SettingsDevice/FirmwareHashCheck.tsx
```diff
@@ -0,0 +1,68 @@
+import {
```
Lemonexe, line 1: copied from `FirmwareRevisionCheck.tsx` (the button)

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -593,10 +594,16 @@ export const prepareDeviceReducer = createReducerWithExtraDeps(initialState, (bu
             if (!state.dismissedSecurityChecks.firmwareRevision) {
                 state.dismissedSecurityChecks.firmwareRevision = [];
             }
-            state.dismissedSecurityChecks.firmwareRevision = [
-                payload,
-                ...state.dismissedSecurityChecks.firmwareRevision,
-            ];
+            state.dismissedSecurityChecks.firmwareRevision.unshift(payload);
```
Lemonexe, line 597: as it is wrapped in redux toolkit `createReducer`, it uses "draft" state so we can safely mutate it :heavy_check_mark:

packages/connect/src/device/Device.ts
mroz22: we are trying to decouple logic and make it flexible by passing data in params instead of relying on global scope. But in this module (Device.ts) DataMangare is already used this way, so its ok.

packages/connect/src/types/settings.ts
mroz22: it makes sense. Maybe it should be used also in other calls? at least those that accept `baseUrl` param? And if yes, what should take precedence?

packages/suite-web/e2e/support/utils/shortcuts.ts
```diff
@@ -84,7 +86,21 @@ export const passThroughSetPin = () => {
...
+        window.store.dispatch({
+            type: SuiteActions.SET_DEBUG_MODE,
+            payload: { showDebugMenu: true },
+        });
+    });
+};
+
+export const disableFirmwareHashCheck = () => {
+    // window.store may not be ready at this point, we need to wait for app to load
+    cy.getTestElement('@welcome-layout/body');
```
Lemonexe, line 98: Idk if waiting on `@welcome-layout/body` is the best way. What I tried:
- waiting on `@device-compromised/hash-check`, but that's not possible because it shows up only for T1B1. I'd need to _optionally_ wait for it, could do that with `MutationObserver` that checks each new node if it might be the error modal, but idk, it's awkward
- `cy.get('[data-testid]')` wait for _any_ test element as proof that React app has rendered. But it's not enough, the[ useCypress()](https://github.com/trezor/trezor-suite/blob/develop/packages/suite-web/src/support/useCypress.ts) has not fired during the _first such render cycle_. It works with `cy.wait(500)` but that's hacky

packages/suite-web/e2e/tests/onboarding/t1b1-create-wallet.test.ts
```diff
@@ -3,15 +3,16 @@
 
 describe('Onboarding - create wallet', () => {
     beforeEach(() => {
+        cy.task('startEmu', { model: 'T1B1', version: '1-latest', wipe: true });
         cy.task('startBridge');
         cy.viewport(1440, 2560).resetDb();
         cy.prefixedVisit('/');
+        cy.disableFirmwareHashCheck();
```
Lemonexe, line 10: the test is skipped, I tried it locally and it's still broken :x:
Fixing it is out of the scope, but I at least verified that `disableFirmwareHashCheck` is necessary, and that it works (proceed to next parts where the test fails). So when someone wants to fix the test in future, they'll have one less thing to fix :slightly_smiling_face: 

Also moved `startEmu` to `beforeEach`, only for code consistency

packages/suite-web/e2e/tests/onboarding/t1b1-recovery-basic.test.ts
```diff
@@ -3,13 +3,13 @@
...
 describe.skip('Onboarding - recover wallet T1B1', () => {
-    before(() => {
+    beforeEach(() => {
         cy.task('startEmu', { model: 'T1B1', version: '1-latest', wipe: true });
         cy.task('startBridge');
 
         cy.viewport(1440, 2560).resetDb();
         cy.prefixedVisit('/');
-        cy.task('startEmu', { wipe: true });
+        cy.disableFirmwareHashCheck();
```
Lemonexe, line 12: [see comment at t1b1-create-wallet.test.ts](https://github.com/trezor/trezor-suite/pull/14766/files/287e17d3481a6284f5fdcc4fbc558408341653c7#diff-2622e5905cf96c51bdd66580564de3f1e49c09a18fc7c9c2c2294a2b2622ae61)

Though I fixed two more problems:
- idk why, but `before` does not work - it fails to initialize emu, while `beforeEach` works :heavy_check_mark: 
- the 2nd call of `startEmu` was killing the T1B1 which this test wants, and loading T2T1 instead :x:

packages/connect/src/types/settings.ts
Lemonexe: It'd make sense to me that those connect calls which currently receive `baseUrl` param, should use the settings baseUrl by default, and have the param optional which would override - this approach would not break API.

I added it to TODOs in PR description.

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -586,17 +586,14 @@ export const prepareDeviceReducer = createReducerWithExtraDeps(initialState, (bu
...
-            if (!state.dismissedSecurityChecks.firmwareRevision) {
-                state.dismissedSecurityChecks.firmwareRevision = [];
+            if (!state.dismissedSecurityChecks.firmwareAuthenticity) {
+                state.dismissedSecurityChecks.firmwareAuthenticity = [];
             }
-            state.dismissedSecurityChecks.firmwareRevision = [
-                payload,
-                ...state.dismissedSecurityChecks.firmwareRevision,
-            ];
+            state.dismissedSecurityChecks.firmwareAuthenticity.unshift(payload);
```
Lemonexe, line 596: as it is wrapped in redux toolkit `createReducer`, the state here is a "draft" state. so we can safely mutate it :heavy_check_mark:

suite-common/wallet-core/src/firmware/getBinFilesBaseUrlThunk.ts
```diff
@@ -1,6 +1,7 @@
 import { isDesktop } from '@trezor/env-utils';
 import { resolveStaticPath } from '@suite-common/suite-utils';
 import { createThunk } from '@suite-common/redux-utils';
+
```
Lemonexe, line 4: forgotten lint warning from #14800 :see_no_evil:

packages/suite/src/support/messages.ts
```diff
@@ -6976,6 +6976,24 @@ export default defineMessages({
...
+    },
+    TR_DEVICE_FIRMWARE_HASH_CHECK_UNKNOWN_RELEASE: {
+        id: 'TR_DEVICE_FIRMWARE_HASH_CHECK_UNKNOWN_RELEASE',
+        defaultMessage: 'Firmware unrecognized. Your Trezor may be counterfeit.',
+    },
+    TR_DEVICE_FIRMWARE_HASH_CHECK_OTHER_ERROR: {
+        id: 'TR_DEVICE_FIRMWARE_HASH_CHECK_OTHER_ERROR',
+        defaultMessage:
+            "Firmware hash check couldn't be performed. Your Trezor may be counterfeit.",
+    },
```
Lemonexe, line 7042: TODO dev placeholder :arrow_right: discuss with product, and have official text added into Crowdin

packages/suite/src/actions/suite/suiteActions.ts
```diff
@@ -309,6 +310,10 @@ export const checkFirmwareRevision =
             type: SUITE.DEVICE_FIRMWARE_REVISION_CHECK,
             payload: { isDisabled },
         });
+        dispatch({
+            type: SUITE.DEVICE_FIRMWARE_HASH_CHECK,
+            payload: { isDisabled },
+        });
```
Lemonexe, line 316: this thunk now dispatches two more granular actions.
I find that preferable to having debug-only granular settings + master switch in normal setting, it could lead to  confusion - forgetting to use either of those switches when writing new code.

:thought_balloon: and it doesn't violate with redux philosophy - an action (or thunk) essentially describes an event = "thing that user did", not a "command to change a thing in state". So an action or thunk can do multiple things in state. 
Yeah I just read the tutorial recently :sweat_smile: :muscle:

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -42,22 +45,20 @@ const getFullscreenApp = (route: AppState['router']['route']): FC | undefined =>
...
 };
 
-interface PreloaderProps {
-    children: React.ReactNode;
-}
-
 // Preloader is a top level wrapper used in _app.tsx.
 // Decides which content should be displayed basing on route and prerequisites.
-export const Preloader = ({ children }: PreloaderProps) => {
+export const Preloader = ({ children }: PropsWithChildren) => {
```
Lemonexe, line 50: shortcut :slightly_smiling_face:

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
```diff
@@ -11,37 +11,47 @@ import { useDevice, useDispatch } from 'src/hooks/suite';
...
+const reportCheckFail = (checkType: 'revision' | 'hash', contextData: any) => {
+    withSentryScope(scope => {
+        scope.setLevel('error');
+        scope.setTag('deviceAuthenticityError', `firmware ${checkType} check failed`);
+        captureSentryMessage(
+            `Firmware ${checkType} check failed! ${JSON.stringify(contextData)}`,
+            scope,
+        );
+    });
+};
```
Lemonexe, line 25: the tag for firmware _revision_ stay the same (continuity in sentry)

packages/suite/src/middlewares/suite/__tests__/buttonRequestMiddleware.test.ts
```diff
@@ -63,6 +63,8 @@ describe('buttonRequest middleware', () => {
         store.getActions().pop();
         expect(store.getActions()).toMatchObject([
             { type: connectInitThunk.pending.type, payload: undefined },
+            { type: getBinFilesBaseUrlThunk.pending.type },
+            { type: getBinFilesBaseUrlThunk.fulfilled.type },
```
Lemonexe, line 67: The order isn't random, so I believe this will not be flaky.
First connectInit gets initiated, then getBinFiles but gets resolved first (`createThunk` is always async, but in this case there is nothing async going on and it gets resolved immediately)

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -432,30 +435,64 @@ export const selectHasExperimentalFeature =
...
+
+/**
+ * Determine hard failure of firmware revision check - specific error types which are severe.
+ * If Suite is offline and cannot perform check or there is some unexpected error, a banner is shown but device is accessible.
+ */
+const selectIsFirmwareRevisionCheckEnabledAndFailed = (state: StateForFirmwareChecks): boolean => {
+    const error = selectFirmwareRevisionCheckError(state);
 
+    return error ? !['cannot-perform-check-offline', 'other-error'].includes(error) : false;
+};
```
Lemonexe, line 465: Took me a long while of iterating and making a mess until I came up with the best way:
split `selectFirmwareRevisionCheckError` to get the raw error or null, instead of having it all in `selectIsFirmwareRevisionCheckEnabledAndFailed`.

See usages of `selectFirmwareRevisionCheckError`, and hash respectively. In all those cases I'd have a lot of boilerplate, potentially making a mistake... and such mistake is currently in develop, see [comment](https://github.com/trezor/trezor-suite/pull/14766/files#r1795749059)

packages/suite/src/components/suite/banners/SuiteBanners/SuiteBanners.tsx
```diff
@@ -58,15 +63,15 @@ export const SuiteBanners = () => {
...
+    // this handles firmware hash being invalid after a firmware update, not the regular firmware hash check
     if (device?.id && firmwareHashInvalid.includes(device.id)) {
         banner = <FirmwareHashMismatch />;
         priority = 92;
-    } else if (
-        !isFirmwareRevisionCheckDisabled &&
-        isDeviceAcquired(device) &&
-        device.authenticityChecks !== undefined &&
-        device.authenticityChecks.firmwareRevision !== null && // check was performed
-        device.authenticityChecks.firmwareRevision.success === false
```
Lemonexe, line 69: Using `selectFirmwareRevisionCheckError` not only deduplicates logic, but it also fixes a bug :bug: 
This is missing an if for messageSystem `selectIsFeatureDisabled(state, Feature.firmwareRevisionCheck)`

_Currently bug does not present, but if we decided to turn it off globally, this would still be on_ :scream:

MiroslavProchazka:
CC @Hannsek - Let’s discuss how much we should encourage users to update their firmware, especially on T1B1.

packages/connect/src/device/Device.ts
```diff
@@ -595,6 +605,52 @@ export class Device extends TypedEmitter<DeviceEvents> {
         }
     }
 
+    async checkFirmwareHash(): Promise<FirmwareHashCheckResult> {
+        const createFailResult = (error: FirmwareHashCheckError) => ({ success: false, error });
+
+        // optional setting for `connect`, see types/settings.ts
+        const baseUrl = DataManager.getSettings('binFilesBaseUrl');
+        // Initially rolled out only for Model One; in fugure we may remove that condition and do it for all models
```
peter-sanderson:
```suggestion
        // Initially rolled out only for Model One; in the future we may remove that condition and do it for all models
```

packages/connect/src/device/Device.ts
```diff
@@ -595,6 +605,52 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
+        const checkSupported = this.atLeast(FIRMWARE.FW_HASH_SUPPORTED_VERSIONS);
+        if (!checkSupported) return createFailResult('check-unsupported');
+
+        try {
+            const currentVersion = this.getVersion()!; // guaranteed because this.atLeast
+            const release = getReleases(this.features.internal_model).find(r =>
+                versionUtils.isEqual(r.version, currentVersion),
+            );
+            // if version is expected to support hash check, but the release is unknown, then firmware is considered unofficial
+            if (!release) return createFailResult('unknown-release');
```
peter-sanderson: nit:

```suggestion
            if (release === undefined) return createFailResult('unknown-release');
```

peter-sanderson: (and all similar situations)

packages/connect/src/device/Device.ts
```diff
@@ -595,6 +605,52 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
+            const deviceResponse = await this.getCommands().typedCall(
+                'GetFirmwareHash',
+                'FirmwareHash',
+                { challenge },
+            );
+
+            if (!deviceResponse.message.hash) return createFailResult('other-error');
+
+            if (deviceResponse.message.hash !== expectedHash)
+                return createFailResult('hash-mismatch');
```
peter-sanderson: :pray: `{}` when multiline

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -84,8 +85,8 @@ export const Preloader = ({ children }: PreloaderProps) => {
     if (
         (router.route?.app === undefined ||
             !ROUTES_TO_SKIP_REVISION_CHECK.includes(router.route?.app)) &&
```
peter-sanderson:
```suggestion
            !ROUTES_TO_SKIP_FIRMWARE_CHECK.includes(router.route?.app)) &&
```

Lemonexe: 2aaa6b0d230c96135fee5d8b53e8a85b9bde285a

packages/connect/src/device/Device.ts
Lemonexe: 5006dfa6f5c32d6b1d8ccdba34d6348be384fc81
572747189f21c54d0273469dc81258f34ddb0046

packages/connect/src/device/Device.ts
Lemonexe: bfb933969284397744a157c96ffe08c95fb23843

packages/connect/src/device/Device.ts
Lemonexe: bfb933969284397744a157c96ffe08c95fb23843

packages/connect/src/device/Device.ts
```diff
@@ -595,6 +605,59 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
+        const release = getReleases(this.features.internal_model).find(r =>
+            versionUtils.isEqual(r.version, firmwareVersion),
+        );
+        // if version is expected to support hash check, but the release is unknown, then firmware is considered unofficial
+        if (release === undefined) return createFailResult('unknown-release');
+
+        const btcOnly = this.firmwareType === FirmwareType.BitcoinOnly;
+        let binary: ArrayBuffer;
+        try {
+            binary = await getBinary({ baseUrl, release, btcOnly });
```
peter-sanderson: Ideally this shall be in a function that would return `ArrayBuffer | null` in case no binary was found (-> `null` as evidence of absence).

Then you check for `=== null` and return `'check-unsupported'`. 

This way you get rid of the `let`

Lemonexe: [6eb21696](https://github.com/trezor/trezor-suite/pull/14766/commits/6eb21696d07aa93adfb798ae640f13caaa137d36)

peter-sanderson (approved):

PR #14772, opened 2024-10-08
feat: remove suport for BIP-69, keep the default order of the inputs and outputs

### LINKED ISSUE #10765: Abandon the use of BIP-69
**Introduction**

**Specification**

peter-sanderson:
This PR removes the support for BIP-69. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/10765


packages/utxo-lib/tests/__fixtures__/compose.ts
```diff
@@ -40,7 +40,7 @@ export default [
                     customField: 'prove that payment output is generic',
                 },
             ],
-            outputsPermutation: [0],
+            outputsPermutation: undefined,
```
peter-sanderson, line 43: It no longer makes sense to test permutations by fixture, they are random now. But the permutation is tested by using it to de-randomazize array to assert it.

packages/utxo-lib/tests/__fixtures__/compose.ts
```diff
@@ -228,13 +228,13 @@ export default [
             inputs: [UTXO],
             outputs: [
                 {
-                    address: '1LetUsDestroyBitcoinTogether398Nrg',
-                    amount: '20000',
+                    address: '1BitcoinEaterAddressDontSendf59kuE',
```
peter-sanderson, line 231: After re-randomization of outputs, they now follow the order in which they were placed by the user.

peter-sanderson:
=> https://github.com/trezor/trezor-suite/pull/14827

PR #14774, opened 2024-10-09
tiny connect chores

mroz22:
again some by the way chores:

- db9b9ed184620264159ad4f668b55618943daefa -  comments in this commit are not valid anymore. we are now releasing all internal dependencies to npm
- 6f95fcd8b7be76ed01d515795b653ee6442fd9ba - tiny cleanup in Device.ts and types tightening.

PR #14776, opened 2024-10-09
chore(connect): types tightening - device.path typed as PathPublic

mroz22:
# 🧹 

a followup after #14774

maybe little controversial

todos: 

- [x] check that saving device works

### DIFF
--- a/packages/connect/src/types/device.ts
+++ b/packages/connect/src/types/device.ts
@@ -1,3 +1,4 @@
+import { Descriptor } from '@trezor/transport';
 import type { PROTO } from '../constants';
 import type { ReleaseInfo } from './firmware';
 
@@ -66,7 +67,7 @@ export type FirmwareHashCheckResult =
     | { success: false; error: FirmwareHashCheckError };
 
 type BaseDevice = {
-    path: string;
+    path: Descriptor['path'];
     name: string;
 };
 
--- a/packages/suite/src/storage/definitions.ts
+++ b/packages/suite/src/storage/definitions.ts
@@ -3,11 +3,10 @@ import { FieldValues } from 'react-hook-form';
 
 import type { SuiteState } from 'src/reducers/suite/suiteReducer';
 import type { FormState, RatesByTimestamps } from '@suite-common/wallet-types';
-import type { AcquiredDevice } from 'src/types/suite';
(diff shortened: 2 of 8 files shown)


packages/connect/src/types/device.ts
```diff
@@ -56,7 +57,7 @@ export type FirmwareRevisionCheckResult =
       };
 
 type BaseDevice = {
-    path: string;
```
mroz22, line 69: this is the change - type tightening of device.path. Till now, it was only internally defined as PathPublic (branded type). From now on, this branded type is propagated to higher layers.

packages/suite/src/storage/definitions.ts
```diff
@@ -60,7 +60,7 @@ export interface SuiteDBSchema extends DBSchema {
     };
     devices: {
         key: string;
-        value: AcquiredDevice;
+        value: Omit<AcquiredDevice, 'path'>;
```
mroz22: it looks like that device saving logic always enforced device with `path = ''`.
I am not sure whether I should stick to this or I may safely assume that we don't want to save path at all. Maybe @szymonlesisz knows?

suite-common/test-utils/src/mocks.ts
```diff
@@ -196,7 +201,10 @@ const getConnectDevice = (dev?: Partial<Device>, feat?: Partial<Features>): Devi
  * @param {Partial<Features>} [feat]
  * @returns {TrezorDevice}
  */
-const getSuiteDevice = (dev?: Partial<TrezorDevice>, feat?: Partial<Features>): TrezorDevice => {
+const getSuiteDevice = (
+    dev?: Partial<Omit<TrezorDevice, 'path'> & { path: string }>,
```
mroz22: accepting path as string in mock so that we don't need to initiate branded type in fixtures

packages/suite/src/storage/definitions.ts
mroz22: ok, it plays some role, at least here https://github.com/trezor/trezor-suite/blob/develop/suite-common/wallet-core/src/device/deviceReducer.ts#L133 so I am going to put it back

marekrjpolak (approved):

PR #14780, opened 2024-10-09
Send coins - switch amount inputs when the inactive one is tapped

yanascz:
Using `onPress` on the `Input` component works fine on iOS but doesn't work on Android for non-editable inputs.
Wrapping the component into `Pressable` and removing ☝️ works on Android but not on iOS. 🤯

👉 I had to combine the two approaches. Please show me that I'm missing something.

## Screenshots:

https://github.com/user-attachments/assets/606b0c89-831d-4bb9-a896-4b8ba09b2306

### DIFF
--- a/suite-native/module-send/src/components/AmountInputs.tsx
+++ b/suite-native/module-send/src/components/AmountInputs.tsx
@@ -118,6 +118,7 @@ export const AmountInputs = ({ index, accountKey }: AmountInputProps) => {
                     inputRef={cryptoRef}
                     isDisabled={!isCryptoSelected}
                     networkSymbol={networkSymbol}
+                    onPress={!isCryptoSelected ? handleSwitchInputs : undefined}
                     onFocus={handleInputFocus}
                 />
                 {isFiatDisplayed && (
@@ -130,6 +131,7 @@ export const AmountInputs = ({ index, accountKey }: AmountInputProps) => {
                             inputRef={fiatRef}
                             isDisabled={isCryptoSelected}
                             networkSymbol={networkSymbol}
+                            onPress={isCryptoSelected ? handleSwitchInputs : undefined}
                             onFocus={handleInputFocus}
                         />
                     </>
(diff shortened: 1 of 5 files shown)


suite-native/module-send/src/components/CryptoAmountInput.tsx
```diff
@@ -26,13 +27,19 @@ export const sendAmountInputWrapperStyle = prepareNativeStyle<{ isDisabled: bool
...
     isDisabled: boolean;
+    onPress?: TextInputProps['onPress'];
 }) => {
     const textColor: Color = isDisabled ? 'textSubdued' : 'textDefault';
 
-    return <Text color={textColor}>{children}</Text>;
+    return (
+        <Pressable onPress={onPress}>
+            <Text color={textColor}>{children}</Text>
+        </Pressable>
```
yanascz, line 41: Not sure this is worth doing but without it tapping on the label doesn't lead to a switch.

suite-native/module-send/src/components/AmountInputs.tsx
```diff
@@ -118,6 +118,7 @@ export const AmountInputs = ({ index, accountKey }: AmountInputProps) => {
                     inputRef={cryptoRef}
                     isDisabled={!isCryptoSelected}
                     networkSymbol={networkSymbol}
+                    onPress={!isCryptoSelected ? handleSwitchInputs : undefined}
```
matejkriz, line 121: Otherwise it would be broken for testnet coins, I guess.
```suggestion
                    onPress={isFiatDisplayed && !isCryptoSelected ? handleSwitchInputs : undefined}
```

matejkriz: Ok, it's not, because `!isFiatDisplayed => isCryptoSelected`

matejkriz (approved):

PeKne (approved):

STew790:
**QA OK**
Works correctly :tada: 
**Info**
24.10.1 https://github.com/trezor/trezor-suite/commit/6f4c0b1102e3df30853adce30c53a0de8453a48a

PR #14787, opened 2024-10-09
chore(suite): deduplicate code with selectBinFilesBaseUrl

Lemonexe:
## Description

Create one selector to get path to static bin files, currently you have to do two different approaches for desktop & web.

### DIFF
--- a/packages/suite/src/reducers/desktop/index.ts
+++ b/packages/suite/src/reducers/desktop/index.ts
@@ -18,5 +18,3 @@ export const desktopReducer = (
             return state;
     }
 };
-
-export const selectDesktopBinDir = (state: DesktopState) => state?.paths?.binDir;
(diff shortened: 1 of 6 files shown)


packages/suite/src/reducers/desktop/index.ts
```diff
@@ -18,5 +18,3 @@ export const desktopReducer = (
             return state;
     }
 };
-
-export const selectDesktopBinDir = (state: DesktopState) => state?.paths?.binDir;
```
Lemonexe, line 22: this was unused. Only the same named `selectDesktopBinDir` from `extraDependencies` was used.

peter-sanderson (approved):

Lemonexe:
Closing in favor of #14800

PR #14791, opened 2024-10-10
Making dynamic connect factory the default

karliatto:
## Description

With this PR we are moving to have Dynamic Trezor Connect as default, so in each environment (web, webextension, mobile) we can use different Trezor Connect implementation and dynamically change from one to the other.

## Related Issue
Related to https://github.com/trezor/trezor-suite/pull/14740

### DIFF
--- a/packages/connect-web/src/index.ts
+++ b/packages/connect-web/src/index.ts
@@ -1,107 +1,44 @@
-import { ConnectFactoryDependencies, factory } from '@trezor/connect/src/factory';
+import { factory } from '@trezor/connect/src/factory';
+import { TrezorConnectDynamic } from '@trezor/connect/src/impl/dynamic';
 import { CoreInIframe } from './impl/core-in-iframe';
 import { CoreInPopup } from './impl/core-in-popup';
-import ProxyEventEmitter from './utils/proxy-event-emitter';
-import type { ConnectSettings, ConnectSettingsPublic, Manifest } from '@trezor/connect/src/types';
-import EventEmitter from 'events';
-import { CallMethodPayload } from '@trezor/connect/src/events';
+import type { ConnectSettingsPublic } from '@trezor/connect';
 import { getEnv } from './connectSettings';
-
-type TrezorConnectType = 'core-in-popup' | 'iframe';
 
 const IFRAME_ERRORS = ['Init_IframeBlocked', 'Init_IframeTimeout', 'Transport_Missing'];
 
-/**
- * Implementation of TrezorConnect that can dynamically switch between iframe and core-in-popup implementations
- */
-export class TrezorConnectDynamicImpl implements ConnectFactoryDependencies {
-    public eventEmitter: EventEmitter;
-
-    private currentTarget: TrezorConnectType = 'iframe';
-    private coreInIframeImpl: CoreInIframe;
-    private coreInPopupImpl: CoreInPopup;
-
-    private lastSettings?: Partial<ConnectSettings>;
-
-    public constructor() {
-        this.coreInIframeImpl = new CoreInIframe();
-        this.coreInPopupImpl = new CoreInPopup();
-        this.eventEmitter = new ProxyEventEmitter([
-            this.coreInIframeImpl.eventEmitter,
-            this.coreInPopupImpl.eventEmitter,
-        ]);
-    }
-
-    private getTarget() {
-        return this.currentTarget === 'iframe' ? this.coreInIframeImpl : this.coreInPopupImpl;
-    }
-
-    private async switchTarget(target: TrezorConnectType) {
-        if (this.currentTarget === target) {
-            return;
-        }
-
-        await this.getTarget().dispose();
-        this.currentTarget = target;
-        await this.getTarget().init(this.lastSettings);
-    }
-
-    public manifest(manifest: Manifest) {
-        this.lastSettings = {
-            ...this.lastSettings,
-            manifest,
-        };
-
-        this.getTarget().manifest(manifest);
-    }
-
-    public async init(settings: Partial<ConnectSettingsPublic> = {}) {
+const impl = new TrezorConnectDynamic<'core-in-popup' | 'iframe'>({
+    implementations: [
+        {
+            type: 'iframe',
+            impl: new CoreInIframe(),
+        },
+        {
+            type: 'core-in-popup',
+            impl: new CoreInPopup(),
+        },
+    ],
+    getInitTarget: (settings: Partial<ConnectSettingsPublic>) => {
         const env = getEnv();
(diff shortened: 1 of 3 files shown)


martykan:
I think it's a good idea, but to make it truly generic for use by all the packages, the `TrezorConnectDynamic` class would need some more refactoring, currently the logic is still tied to the implementations used in the `connect-web` package, where for other packages there could be different implementations and logic

karliatto:
> I think it's a good idea, but to make it truly generic for use by all the packages, the `TrezorConnectDynamic` class would need some more refactoring, currently the logic is still tied to the implementations used in the `connect-web` package, where for other packages there could be different implementations and logic

Sure, what about moving that `handleErrorFallback` to instance initiation like https://github.com/trezor/trezor-suite/pull/14791/files#diff-e53c2a11939014264b9f02396492ab90e4b77ded32292f31886abf13fac5a7c6R10 ?

martykan:
Yeah that's a good solution, maybe `TrezorConnectType` could also be turned into a generic.

karliatto:
> Yeah that's a good solution, maybe `TrezorConnectType` could also be turned into a generic.

Even better :+1: https://github.com/trezor/trezor-suite/pull/14791/commits/34e1db2d18db3de026bb7a7ad1a4a6beea98ad5b

martykan:
Perfect, but something's breaking in CI

packages/connect/src/impl/dynamic.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    public async switchTarget(target: TrezorConnectImplType) {
+        if (this.currentTarget === target) {
+            return;
+        }
+
+        if (!this.lastSettings) {
+            throw new Error('Connect dynamic not initialized.');
+        }
+        this.currentTarget = target;
```
martykan, line 60: Previously there was also dispose before switching
```
await this.getTarget().dispose();
```

packages/connect-web/src/index.ts
```diff
@@ -1,107 +1,45 @@
...
+    getInitTarget: (settings: Partial<ConnectSettingsPublic>) => {
         const env = getEnv();
         if (settings.coreMode === 'iframe' || settings.popup === false || env === 'webextension') {
-            this.currentTarget = 'iframe';
+            return 'iframe';
         } else if (settings.coreMode === 'popup') {
-            this.currentTarget = 'core-in-popup';
+            return 'core-in-popup';
         } else {
+            // TODO: it does not look like a good idea to modify settings like this, create a public method in dynamic for that?
```
martykan: Yeah, the current code probably can't work

packages/connect/src/impl/dynamic.ts
karliatto: Right, that was the breaking CI thing.

packages/connect-web/src/index.ts
karliatto: I think it actually works because JS objects are copied by reference. But it is probably not a good idea.
I am thinking is it is actually needed? if "auto" means it will use "iframe" as first option.

Shouldn't the code below works the same way?


```typescript
getInitTarget: (settings: Partial<ConnectSettingsPublic>) => {
        if (settings.coreMode === 'popup') {
            return 'core-in-popup';
        } 

        return 'iframe';
}
```
Do you know the reason why `settings.coreMode = 'auto';` is required here?

karliatto: If you know why it is required we can keep doing it by adding a public method to TrezorConnectDynamic like:

```typescript
    public updateSettings(settingToUpdate: Partial<ConnectSettingsPublic>) {}
```

martykan: It's just written like this so coreMode is explicitly set instead of undefined. 
But we could modify other code to consider coreMode undefined as auto (eg. condition in `handleErrorFallback`)

packages/connect-web/src/index.ts
```diff
@@ -1,107 +1,40 @@
...
-            }
-
-            throw error;
-        }
-    }
+        return 'iframe';
+    },
+    handleErrorFallback: async (errorCode: string) => {
+        const isCoreModeAuto =
+            impl.lastSettings?.coreMode === 'auto' || !impl.lastSettings?.coreMode === undefined;
```
martykan:
```suggestion
            impl.lastSettings?.coreMode === 'auto' || impl.lastSettings?.coreMode === undefined;
```
Extra `!` right?

packages/connect-web/src/index.ts
```diff
@@ -1,107 +1,40 @@
...
-            ...this.lastSettings,
-            manifest,
-        };
-
-        this.getTarget().manifest(manifest);
-    }
-
-    public async init(settings: Partial<ConnectSettingsPublic> = {}) {
-        const env = getEnv();
-        if (settings.coreMode === 'iframe' || settings.popup === false || env === 'webextension') {
```
martykan, line 60: This condition now disappears but maybe it's not a problem. 
Before this meant that for `settings.popup === false || env === 'webextension'` it would not be in auto mode by default, it would strictly use iframe

packages/connect-web/src/index.ts
karliatto: https://github.com/trezor/trezor-suite/pull/14791/commits/5a921d4f1745fe384a321aacb6eb0975170cfd09

packages/connect-web/src/index.ts
karliatto: What happens if you have settings.coreMode === 'popup' in env === 'webextension'  ?

Is it possible to user coreMode === 'popup' in a webextension using connect-web?

martykan: Undefined behavior... it could probably work, but I haven't tested it and I don't think there is any reason to use it.

karliatto: Right, and even webextension using connect-web is kind of legacy since it is not supported in MV3.
Ok, let's keep the previous logic c806de2c446d09781dee2bbd19dfe656e09459d5

martykan: Well that condition should probably be put into `isCoreModeAuto`, because right now it doesn't affect anything in `getInitTarget`

karliatto: Do you mean in case when there is an error it should still keep 'iframe' in case that `env === 'webextenstion`?

Like:

```js
        const isCoreModeAuto =
            env !== 'webextension' ||
            impl.lastSettings?.coreMode === 'auto' ||
            impl.lastSettings?.coreMode === undefined;
```

martykan: I just pushed a fixup that should be equivalent to how it worked previously

karliatto: :+1:  looks good to me, rebasing the PR...

martykan (approved):

PR #14794, opened 2024-10-10
feat(mobile-connect-popup): handle all URLs, add feature flag for popup

martykan:
## Description

Fix deeplink handling for develop and prod HTTPS URLs, eg.:

- https://connect.trezor.io/9/deeplink/
- https://dev.suite.sldev.cz/connect/develop/deeplink/

Note: Prod URL only opens on prod app, and vice versa for dev URL

Add a feature flag to enable/disable Connect Poup. By default it's enabled on dev builds.

### DIFF
--- a/suite-native/app/app.config.ts
+++ b/suite-native/app/app.config.ts
@@ -168,14 +168,20 @@ export default ({ config }: ConfigContext): ExpoConfig => {
                 monochromeImage: './assets/appIcon_android.png',
                 ...appIconAndroid,
             },
-            intentFilters:
-                buildType === 'production'
-                    ? []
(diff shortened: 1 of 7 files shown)


karliatto:
In order to have it for production we need to have the https://connect.trezor.io/.well-known/assetlinks.json available, right?

martykan:
Yes it's not there yet, I will prepare it and send it to Tali. Also the equivalent for iOS.

karliatto (approved):
As agreed let's move with this even there is not https://connect.trezor.io/.well-known/assetlinks.json  yet.

:+1:

PR #14797, opened 2024-10-10
fix(protobuf): repo path in build script

martykan:
## Description

There is an issue with running `protobuf-build.sh` in CI due to the REPO_PATH not being able set correctly if the directory does not exist yet. 

With this change the REPO_PATH will always be set to `trezor-firmware-probuf-update` within the parent directory which holds the trezor-suite repo. 

Before: https://github.com/trezor/trezor-suite/actions/runs/11265164884/job/31326531995

After: https://github.com/trezor/trezor-suite/actions/runs/11272831120/job/31348572302


mroz22 (approved):

mroz22:
maybe @szymonlesisz wants to test whether it still works locally for him?

PR #14799, opened 2024-10-10
feat(suite): send form token select

### LINKED ISSUE #13749: EVM Tokens / Contract address in send form
## Description
Improve token select feature in send form. It should replace the currency (token) mini select menu in the send form.

Image

[Figma design](https://www.figma.com/design/o6eb4wUMjSu5xMb34423vh/Tokens?node-id=4518-22627&t=wM3XYIq1PipOxf5S-1)

### Token select element
- a new element which enable users to choose a token in send form
- the element is placed at the top to highlight and clarify what is being sent. The user can change it at any time; it includes the mainnet coin and tokens available on the selected account
- an 'i' icon will appear next to it, and hovering over the icon will show the full contract address in a tooltip.

Screenshot 2024-09-30 at 10 52 52
- in case of Cardano, please show both `assetId` and `fingerprint`

### Token select modal _TBD_
- Reuse choose token modal #14551 
- _Design TBD_ (it should not be dropdown but modal from 🔝 
- Searchable by `name`, `symbol`, `contract address`
- Choose `Shown` / `Hidden`
- Display contract address

Screenshot 2024-09-30 at 11 13 23
Screenshot 2024-09-30 at 11 12 49
 Image

enjojoy:
## Description

- Improved UI of token selection in send form. 
- Copy token contract address in the tooltip. 
- Implemented new modal instead of a dropdown
- Removed EVM warning 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13749

### Before: 
image

### After: 
Screenshot 2024-11-22 at 18 41 37

Screenshot 2024-11-22 at 17 22 28

- No tokens available on account with supported tokens
Screenshot 2024-11-22 at 18 42 54

- if the network doesn't support tokens
Screenshot 2024-11-22 at 18 43 36

- On hover on address it can be copied or user can go to scanner web:
Screenshot 2024-11-22 at 18 42 26


adderpositive (approved):

adderpositive:
I have found only this strange design. Not sure it is a related issue.
obrazek

enjojoy:
> I have found only this strange design. Not sure it is a related issue. obrazek

AFAIK @adamhavel fixed this, probably not on develop yet

seibei-iguchi:
> > I have found only this strange design. Not sure it is a related issue. obrazek
> 
> AFAIK @adamhavel fixed this, probably not on develop yet

I might have broken something else, will check:)

enjojoy:
> > > I have found only this strange design. Not sure it is a related issue. obrazek
> > 
> > 
> > AFAIK @adamhavel fixed this, probably not on develop yet
> 
> I might have broken something else, will check:)

This branch wasn't rebased on develop for a few days, so maybe it's just not there yet

seibei-iguchi:
> > > > I have found only this strange design. Not sure it is a related issue. obrazek
> > > 
> > > 
> > > AFAIK @adamhavel fixed this, probably not on develop yet
> > 
> > 
> > I might have broken something else, will check:)
> 
> This branch wasn't rebased on develop for a few days, so maybe it's just not there yet

Will you try rebasing it?

enjojoy:
> > > > > I have found only this strange design. Not sure it is a related issue. obrazek
> > > > 
> > > > 
> > > > AFAIK @adamhavel fixed this, probably not on develop yet
> > > 
> > > 
> > > I might have broken something else, will check:)
> > 
> > 
> > This branch wasn't rebased on develop for a few days, so maybe it's just not there yet
> 
> Will you try rebasing it?

Yes but not right now, I will let you know in case the issue persists after rebasing.

enjojoy:
Seems to be fixed when rebased on develop @adamhavel @adderpositive 

image

enjojoy:
image

packages/suite/src/views/wallet/send/Outputs/Outputs.tsx
```diff
@@ -35,9 +36,10 @@ const StyledEvmExplanation = styled.div`
...
 interface OutputsProps {
     disableAnim?: boolean; // used in tests, with animations enabled react-testing-library can't find output fields
+    areTokensSupported: boolean;
 }
 
-export const Outputs = ({ disableAnim }: OutputsProps) => {
+export const Outputs = ({ disableAnim, areTokensSupported }: OutputsProps) => {
     const [height, setHeight] = useState(0);
     const [hasRenderedOutputs, setHasRenderedOutputs] = useState(false);
     const size = useLayoutSize();
```
tomasklim: can you resolve `areTokensSupported` in this component?

packages/suite/src/components/suite/CoinBalance.tsx
```diff
@@ -3,7 +3,7 @@ import { FormattedCryptoAmount } from 'src/components/suite';
 
 interface CoinBalanceProps {
     value: string;
-    symbol: Account['symbol'];
+    symbol: Account['symbol'] | (string & {});
```
tomasklim, line 6: ?

packages/components/src/components/AssetLogo/AssetLogo.tsx
```diff
@@ -73,8 +72,10 @@ export const AssetLogo = ({
         setIsPlaceholder(true);
     };
     useEffect(() => {
-        setIsPlaceholder(!shouldTryToFetch);
-    }, [shouldTryToFetch]);
+        if (shouldTryToFetch) {
```
tomasklim: why?

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -125,14 +129,15 @@ export const SelectAssetModal = ({
             ? item.coingeckoId === searchCategory.coingeckoId ||
               item.coingeckoId === searchCategory.coingeckoNativeId
             : true;
+        const networkContractAddress = `${item.networkSymbol}${item.contractAddress ? `--${item.contractAddress}` : ''}`;
```
tomasklim: I dont understand this

enjojoy:
TODO:

- [x]  Remove network warning banner 
- [x]  Remove tooltip from address
- [x]  TX model inputs-outputs - component for address, (first 6 and last 4 digits), on hover buttons to copy and show in blockbook.
- [x]  On copy it has a similar modal as when copy in tokens. 

- [x]  In modal select between My Tokens and Hidden & Unverified too
- [x]  In modal no search by network. 
- [x]  Modal is too long - can be shorter

- [x]  In send form: button Clear All is missing, 
- [x] No currency sign in send field when native token is picked. 
- [x]   Full coin amount and fiat
- [x] remove shadow
- [x] add unrecognized badge
- [x] icons to token category picker
- [x]  No balance under Amount
- [x]  Icon size 24 
- [x] check cardano, there is fingerprint and asset id. No contract address, but those two
- [x]  Check with figma design

suite-common/token-definitions/src/tokenDefinitionsUtils.ts
```diff
@@ -63,3 +64,12 @@ export const buildTokenDefinitionsFromStorage = (
 
     return tokenDefinitions;
 };
+
+export const getTokenExplorerUrl = (network: Network, token: EnhancedTokenInfo) => {
```
tomasklim: I think it is more `wallet-utils`

enjojoy:
@tomasklim  I checked for cardano and fixed the logo for it here https://github.com/trezor/trezor-suite/pull/14799/commits/a07e774133cf0198667103857f9c2bfd7e62a4c8 


If chatgpt is not lying, it should be policy ID, so I will also change the text on the TokenSelect and on the modal for copying 
image

UPD: I will use policy ID because on the explorer where the link leads, policy ID is displayed https://github.com/trezor/trezor-suite/pull/14799/commits/75d9b794ca24653cd7e2c9e49ede2b5f80cef12b

tomasklim:
Check tokens section, we show policy id and fingerprint

enjojoy:
> Check tokens section, we show policy id and fingerprint

Yes, in tokens section we show all 3 things

image

Should we also display all three things in the token select? I feel like policyID is enough, but we can display all 3 and a link to explorer separately after the 3 identifiers

enjojoy:
@tomasklim This commit was needed because with the addition of cardano tokens fiat rates request was failing, because when request is done with PolicyID it was always not found. https://github.com/trezor/trezor-suite/pull/14799/commits/ac5f41e126b210c7d426a5c6af4d36c9b0831497

Is there any case you know of where this can be useful or it was there just as a precaution and never really tested?

tomasklim:
> it was always not found

Always is a strong word. On what tokens is it based on?

> Is there any case you know of where this can be useful or it was there just as a precaution and never really tested?

I always test things...

enjojoy:
> > it was always not found
> 
> Always is a strong word. On what tokens is it based on?
> 
> > Is there any case you know of where this can be useful or it was there just as a precaution and never really tested?
> 
> I always test things...

Maybe the API changed since then? In my case it was based on these coins, when I removed that part the error disappeared and it stopped lagging in send form, too.

image

P.S. I didn't mean to question your testing consistency, asked just in case because it was behaving oddly

enjojoy:
> > Check tokens section, we show policy id and fingerprint
> 
> Yes, in tokens section we show all 3 things
> 
> image
> Should we also display all three things in the token select? I feel like policyID is enough, but we can display all 3 and a link to explorer separately after the 3 identifiers

Asked this question in Cardano group, let's stick to policyID: 
image

tomasklim:
https://www.coingecko.com/en/coins/snek

enjojoy:
> https://www.coingecko.com/en/coins/snek

I see, thanks. So there are different types of tokens, I'll handle this case then. 

upd: will fix it in a later PR, there's an issue for that

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenAddress.tsx
```diff
@@ -0,0 +1,134 @@
+import { HiddenPlaceholder } from 'src/components/suite/HiddenPlaceholder';
+import { Icon, Link, Text, TextProps } from '@trezor/components';
+import { useMemo, useState } from 'react';
+import styled, { css, useTheme } from 'styled-components';
+
+const IconWrapper = styled.div`
+    display: none;
+    padding: 1px;
```
seibei-iguchi: Please use predefined theme variables for all these numbers.

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenAddress.tsx
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            &:hover,
+            &:focus {
+                ${onHoverTextOverflowContainerHover}
+            }
+        `}
+`;
+
+const SpanTextStart = styled.span`
```
seibei-iguchi: This can be all done with Text component.

seibei-iguchi (dismissed):

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenAddress.tsx
enjojoy: Thanks, checking the whole PR for it.

enjojoy:
@adderpositive  I added some changes to your part, please check it out, on the first glance it was working well.

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -167,19 +210,30 @@ export const SelectAssetModal = ({
...
+                        />
+                    )}
 
                 {filteredData.length === 0 ? (
                     <AssetItemNotFound
                         listHeight={LIST_HEIGHT}
                         listMinHeight={LIST_MIN_HEIGHT}
                         searchCategory={searchCategory}
-                        networkCategories={networkCategories}
+                        networkCategories={filterCategories.categories as NetworkFilterCategory[]}
```
adderpositive: Perhaps we could find a way to pass only `searchNetworkName` instead of passing `searchCategory` and `networkCategories` and do all logic in this component. We will also avoid forcing type. What do you think?

adderpositive (dismissed):
Fine by me, just one nit.

adderpositive:
I forgot that when changing currency in Send, the form jumps slightly.

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
enjojoy: What do you think about this? https://github.com/trezor/trezor-suite/pull/14799/commits/a7f3d42ace60630dfd61ceca683642ba62c77186

I think it's nice because we'll be able to add a token send form specific not found message if needed.

enjojoy:
> I forgot that when changing currency in Send, the form jumps slightly.

I don't see this in send, which network are you using? Can you record a video please?

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
adderpositive: Fine!

adderpositive:
> > I forgot that when changing currency in Send, the form jumps slightly.
> 
> I don't see this in send, which network are you using? Can you record a video please?

I'm not able to replicate it now, it is fixed I think. :)

packages/suite/src/views/wallet/send/index.tsx
```diff
@@ -76,6 +76,8 @@ const SendLoaded = ({ children, selectedAccount }: SendLoadedProps) => {
 
     const { symbol } = selectedAccount.account;
 
+    const areTokensSupported = selectedAccount.network?.features?.includes('tokens') ?? false;
```
tomasklim: `getNetworkFeatures(networkSymbol).includes('tokens')`

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,301 @@
...
+                                            </Row>
+                                            <Row justifyContent="flex-start">
+                                                {option.contractAddress && option.cryptoName && (
+                                                    <Text variant="tertiary" typographyStyle="hint">
+                                                        <Row gap={spacings.xxs}>
+                                                            {account.symbol === 'ada' ? (
+                                                                <Translation id="TR_POLICY_ID_ADDRESS" />
+                                                            ) : (
+                                                                <Translation id="TR_CONTRACT_ADDRESS" />
+                                                            )}{' '}
```
tomasklim: I think instead of {' '} it is better to increase flex gap

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,301 @@
...
+                                                        amount={option.balance ?? '0'}
+                                                        symbol={option.networkSymbol}
+                                                    />
+                                                </Text>
+                                            </Row>
+                                            <Row justifyContent="flex-start">
+                                                {option.contractAddress && option.cryptoName && (
+                                                    <Text variant="tertiary" typographyStyle="hint">
+                                                        <Row gap={spacings.xxs}>
+                                                            {account.symbol === 'ada' ? (
```
tomasklim: `network === cardano` so that it is also for cardano testnet

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,301 @@
...
+                                                                )}
+                                                                shouldAllowCopy={true}
+                                                                typographyStyle="hint"
+                                                                variant="tertiary"
+                                                                tokenExplorerUrl={
+                                                                    option.tokenExplorerUrl
+                                                                }
+                                                                onCopy={() =>
+                                                                    onCopyAddress(
+                                                                        option.contractAddress ||
```
tomasklim: `getContractAddressForNetwork(` also here

packages/suite/src/views/wallet/send/Outputs/TokenSelect/tokenSelectUtils.ts
```diff
@@ -0,0 +1,137 @@
...
+            box-shadow: ${theme.boxShadowBase};
+        `}
+    cursor: ${({ $isDisabled }) => ($isDisabled ? 'not-allowed' : 'pointer')};
+`;
+
+export const onCopyAddress = (
+    address: string,
+    addressType: AddressType,
+    shouldShowCopyAddressModal: boolean,
+    dispatch: Dispatch,
```
tomasklim: utils should not accept `dispatch`. it should be action/thunk or just do

`dispatch(onCopyAddress)` and this util will `return openModal` and `return(notificationActions` but i think thunk/action would be better for this

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,301 @@
...
+                                                                            '',
+                                                                        'contract',
+                                                                        shouldShowCopyAddressModal,
+                                                                        dispatch,
```
tomasklim: do not send `dispatch` as param

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,301 @@
...
+                                        <Column alignItems="flex-start">
+                                            <Row justifyContent="flex-start">
+                                                <Text variant="default" typographyStyle="body">
+                                                    {option.cryptoName}
+                                                </Text>
+                                            </Row>
+                                            <Row>
+                                                <Text variant="tertiary" typographyStyle="hint">
+                                                    {option.balance ?? '0'}{' '}
+                                                    {option.symbol.toUpperCase()} ≈{' '}
```
tomasklim: show `≈` only if rate is available (not a testnet and also rate is available for token)

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,301 @@
...
+                        $elevation={elevation}
+                        fillType="none"
+                        margin={{ bottom: spacings.sm }}
+                        paddingType="large"
+                        onClick={
+                            options.length > 1
+                                ? () => {
+                                      setIsTokensModalActive(true);
+                                  }
+                                : () => {}
```
tomasklim: undefined

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenAddress.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const [isClicked, setIsClicked] = useState(false);
+    const theme = useTheme();
+
+    const copy = (e: React.MouseEvent) => {
+        e.stopPropagation();
+        onCopy();
+    };
+
+    const shortenAddress = useMemo(() => {
+        return `${tokenContractAddress.slice(0, 6)}...${tokenContractAddress.slice(-4)}`;
```
tomasklim: maybe we should expand it on hover?

tomasklim (changes requested):

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenAddress.tsx
enjojoy: I think it's a good idea but let's include it in further contributions after alignment with PM, I wouldn't introduce new features just before the freeze (unless if this PR wont be included in todays freeze for some reason)

packages/suite/src/components/suite/CoinBalance.tsx
adderpositive: This provides broader support for account symbols. By default, you'll see Account['symbol'] as the value passed, but you can use any string. This applies specifically to the trade section, where additional symbols are supported. We have created with @enjojoy after discussion.

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
adderpositive: I assume this refers to support for `cryptoId`.

suite-common/token-definitions/src/tokenDefinitionsUtils.ts
enjojoy: Sure

packages/suite/src/views/wallet/send/Outputs/Outputs.tsx
enjojoy: yes https://github.com/trezor/trezor-suite/pull/14799/commits/330da4255450b23d2cfbaad51f213baac0ae55c2

packages/suite/src/components/suite/CoinBalance.tsx
enjojoy: Yes, I think it's because in coinmarket the types are not consistent with the other part of suite

packages/components/src/components/AssetLogo/AssetLogo.tsx
enjojoy: not sure why I did it, returning it back in fixup

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
enjojoy: Yes, this is there because before it was searching by cryptoId (cryptoId is specific to coinmarket and has this structure of network--contract), now it's no longer there, because in my case I want to disable search by network, so I recreated it this way in case we search by network will be enabled

packages/suite/src/views/wallet/tokens/common/TokensTable/TokenRow.tsx
```diff
@@ -158,7 +148,7 @@ export const TokenRow = ({
                     <AssetLogo
                         coingeckoId={coingeckoId || ''}
                         placeholder={token.name || token.symbol || 'token'}
-                        contractAddress={networkContractAddress as TokenAddress}
+                        contractAddress={token.contract.toLowerCase() as TokenAddress}
```
tomasklim: This change is breaking Solana icons

tomasklim:
Screenshot 2024-11-05 at 15 47 15

hidden placeholder is missing

💚 done

packages/product-components/src/components/SelectAssetModal/mockData.ts
```diff
@@ -7,224 +8,255 @@ export const selectAssetModalOptions: SelectAssetOptionProps[] = [
...
+        symbol: 'BCH',
         cryptoName: 'Bitcoin Cash',
         coingeckoId: 'bitcoin-cash',
+        contractAddress: null,
     },
     {
         type: 'currency',
-        value: 'binancecoin',
-        label: 'BNB',
+        networkSymbol: 'binancecoin',
```
tomasklim: this is `coingeckoNativeId`, definitely not `networkSymbol`

tomasklim (changes requested):

packages/suite/src/types/coinmarket/coinmarket.ts
```diff
@@ -178,11 +177,19 @@ export interface CoinmarketCoinLogoProps {
...
 export interface CoinmarketCryptoSelectItemProps
-    extends Omit<SelectAssetOptionCurrencyProps, 'value'> {
+    extends Omit<
+        SelectAssetOptionCurrencyProps,
+        'symbol' | 'networkSymbol' | 'contractAddress' | 'cryptoName'
+    > {
     value: CryptoId;
+    label: string;
+    contractAddress?: string;
+    cryptoName: string | undefined;
```
tomasklim: so it is first omitted to be added again?

packages/suite/src/views/wallet/send/Outputs/TokenSelect/tokenSelectUtils.ts
```diff
@@ -0,0 +1,118 @@
...
+            label: translationString('TR_TOKENS'),
+        },
+        {
+            type: 'hiddenWithBalance',
+            label: translationString('TR_HIDDEN'),
+        },
+    ],
+});
+
+export const copyAddressAction =
```
tomasklim: move this to another file as it is now reused from two places and it is action, not util

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenAddress.tsx
```diff
@@ -0,0 +1,132 @@
...
+    onCopy: () => void;
+}
+
+// This is needed because icon interferes with pointer events of Select
+// eslint-disable-next-line local-rules/no-override-ds-component
+const IconWithNoPointer = styled(Icon)`
+    pointer-events: none;
+`;
+
+export const TokenAddress = ({
```
tomasklim: Are we okay that we have 2 components doing almost the same? `IOAddress` and this one `TokenAddress`.

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,324 @@
...
+import { SUITE } from 'src/actions/suite/constants';
+import {
+    CoinLogo,
+    SelectAssetModal,
+    SelectAssetOptionCurrencyProps,
+} from '@trezor/product-components';
+import { getContractAddressForNetwork, isTestnet } from '@suite-common/wallet-utils';
+import { borders, Elevation, mapElevationToBackground, spacings } from '@trezor/theme';
+import { FiatValue, HiddenPlaceholder, Translation } from 'src/components/suite';
+import { TokenAddress } from './TokenAddress';
```
Lemonexe: Duplicate identifier; compilation error

tomasklim (dismissed):

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,420 @@
...
+                                                <Text variant="tertiary" typographyStyle="hint">
+                                                    <HiddenPlaceholder>
+                                                        {option.balance ?? '0'}
+                                                    </HiddenPlaceholder>{' '}
+                                                    {option.symbol.toUpperCase()}
+                                                    {!isTestnet(
+                                                        option.networkSymbol as NetworkSymbol,
+                                                    ) &&
+                                                        option.balance &&
+                                                        ' ≈ '}
```
Lemonexe: nit: one ternary would be nicer than the two `&&` short circuits.
:thought_balloon:  Though ideally, the `FiatValue` would also be in the condition. This code relies on the behavior that it returns nothing, which is not so clear. But that's fine
```suggestion
                                                    (!isTestnet(option.networkSymbol as NetworkSymbol) && option.balance) ? ' ≈ ' : null
```

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,420 @@
...
+                                                                ? (option.contractAddress as TokenAddress)
+                                                                : undefined
+                                                        }
+                                                        amount={option.balance ?? '0'}
+                                                        symbol={option.networkSymbol}
+                                                    />
+                                                </Text>
+                                            </Row>
+                                            <Row justifyContent="flex-start">
+                                                {option.contractAddress && option.cryptoName && (
```
Lemonexe: Having the condition within `Row` renders empty Row for currencies (useless I think?)

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,420 @@
...
+                                                    <Text variant="tertiary" typographyStyle="hint">
+                                                        <Row gap={spacings.xxs}>
+                                                            {option.networkSymbol === 'cardano' ? (
```
Lemonexe: This still renderes "Contract address" for Cardano (expected "Policy ID")
Probably should be `=== 'ada'`, but idk, it's typed as `NetworkSymbol | string` :see_no_evil:

packages/suite/src/actions/suite/copyAddressActions.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { copyToClipboard } from '@trezor/dom-utils';
+import { notificationsActions } from '@suite-common/toast-notifications';
+import { openModal } from 'src/actions/suite/modalActions';
+import { Dispatch } from 'redux';
+
+export const copyAddressAction =
+    (address: string, addressType: AddressType, shouldShowCopyAddressModal: boolean) =>
+    (dispatch: Dispatch) => {
+        if (shouldShowCopyAddressModal) {
```
Lemonexe: Personally, I'd split this to two thunks, one for if & one for else.
The caller can decide which one to call, instead of passing `shouldShowCopyAddressModal` to split the logic.
There is no common logic for the if & else, it's just that..

:thought_balloon:  _btw it is technically synchronous thunk isn't it?_

packages/suite/src/actions/suite/copyAddressActions.ts
```diff
@@ -0,0 +1,25 @@
...
+            dispatch(
+                openModal({
+                    type: 'copy-address',
+                    addressType,
+                    address,
+                }),
+            );
+        } else {
+            const result = copyToClipboard(address);
+            if (typeof result !== 'string') {
```
Lemonexe: the condition is unclear at first glance, but a named variable makes it self-explanatory:
```suggestion
            const isError = typeof result === 'string'
            if (!isError) {
```

Lemonexe:
EDIT: resolved, thank you :+1: 

:question:  concerning behavior when I have no tokens, just currency - it feels weird to me that it displays the :no_entry_sign: cursor on the whole card.
As if the UI is telling me that Polygon itself is blocked from sending.
Maybe :no_entry_sign: cursor should be only on the arrow, then it's telling me I can't select (as there isn't anything to select)
blocked cursor

suite-common/wallet-utils/src/tokenUtils.ts
```diff
@@ -22,3 +23,12 @@ export const getContractAddressForNetwork = (
             return contractAddress.toLowerCase();
     }
 };
+
+export const getTokenExplorerUrl = (network: Network, token: TokenInfo) => {
+    const explorerUrl =
+        network.networkType === 'cardano' ? network.explorer.token : network.explorer.account;
+    const contractAddress = network.networkType === 'cardano' ? token.fingerprint : token.contract;
+    const queryString = network.explorer.queryString ?? '';
```
Lemonexe, line 34: Is empty string ok? I honestly don't know, just asking if you're counting with it.

packages/suite/src/components/suite/copy/TokenAddressRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+            <HiddenPlaceholder disableKeepingWidth>
+                <TextOverflowContainer
+                    onMouseLeave={() => setIsClicked(false)}
+                    data-testid="@tx-detail/txid-value"
+                    id={tokenContractAddress}
+                    $shouldAllowCopy={shouldAllowCopy}
+                >
+                    <Text textWrap="nowrap">{shortenAddress}</Text>
+                    {shouldAllowCopy ? (
+                        <IconWrapper onClick={e => copy(e)}>
```
Lemonexe:
```suggestion
                        <IconWrapper onClick={copy}>
```

packages/suite/src/components/suite/copy/TokenAddressRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+    tokenExplorerUrl,
+    shouldAllowCopy = true,
+    typographyStyle = 'label',
+    variant = 'default',
+    onCopy,
+}: TokenAddressRowProps) => {
+    const [isClicked, setIsClicked] = useState(false);
+    const theme = useTheme();
+
+    const copy = (e: React.MouseEvent) => {
```
Lemonexe: Nit: UMD global React is used here; better would be to import it explicitly, but I don't even know how it affects compilation :shrug: I'm just thinking "explicit better than implicit"

packages/suite/src/components/suite/copy/TokenAddressRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+        e.stopPropagation();
+        onCopy();
+    };
```
Lemonexe, line 90: Currently, the check icon is never displayed!
```suggestion
        setIsClicked(true);
    };
```

packages/suite/src/components/suite/copy/TokenAddressRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    const shortenAddress = useMemo(() => {
```
Lemonexe: useless memo, it's not used in any dependency. It neither prevent any rerenders, nor saves expensive computation..

Also typo I think, it's string not a fn, so `shortenedAddress` would be better name I think

Lemonexe (commented):
Unloading some minor codestyle comments, but I am not finished reviewing yet :hourglass:

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -106,33 +135,49 @@ const getNetworkCount = (options: SelectAssetOptionProps[]) => {
...
-    const [searchCategory, setSearchCategory] = useState<SelectAssetSearchCategoryType>(null); // coingeckoNativeId as fallback for ex. polygon
+    const [searchCategory, setSearchCategory] = useState<SelectAssetSearchCategory>(null); // coingeckoNativeId as fallback for ex. polygon
     const [end, setEnd] = useState(options.length);
     const data = useMemo(() => getData(options), [options]);
     const { scrollElementRef, onScroll, ShadowTop, ShadowBottom, ShadowContainer } =
         useScrollShadow();
     const networkCount = getNetworkCount(options);
 
+    const searchPlaceholder = searchPlaceholderText
+        ? searchPlaceholderText
```
Lemonexe: Can be simplified with `??`

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -106,33 +135,49 @@ const getNetworkCount = (options: SelectAssetOptionProps[]) => {
...
         return (
             (searchFor(item.name) ||
-                searchFor(item.cryptoId) ||
                 searchFor(item.badge) ||
-                searchFor(item.symbol)) &&
+                (!disableSearchByNetwork && searchFor(item.symbol)) ||
+                (disableSearchByNetwork
+                    ? searchFor(contractAddress)
+                    : searchFor(networkContractAddress))) &&
             categoryFilter
```
Lemonexe: I know this is original code, but here is an opportunity to enhance:
`categoryFilter` is defined at beginning of callback, and here it is in one big `&& categoryFilter`.
:arrow_right: useless, you can early return false if `!categoryFilter` at the beginning.

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -167,19 +212,30 @@ export const SelectAssetModal = ({
...
+                )}
+                {filterCategories?.categoriesType === 'tokens' &&
+                    setPickedSendTokenCategory &&
+                    pickedSendTokenCategory && (
+                        <SendTokenTabs
+                            sendTokenCategories={filterCategories.categories}
+                            pickedCategory={pickedSendTokenCategory}
+                            setPickedCategory={setPickedSendTokenCategory}
+                        />
+                    )}
```
Lemonexe: Ad in person discussion with @tomasklim : this is what I was talking about - this should be rendered above (in coinmarket or send form respectively), and passed to `SelectAssetModal` as JSX.
Like `children`, but in this case I'd name the prop `categoryTabs: ReactNode` or something, because `children` would be misleading name (it's just specific part of the content, not whole content).
:arrow_right: it will make this component dumber, and that's how I like it.

I think it's possible here; both `NetworkTabs` and `SendTokenTabs` use only props from above, or values directly calculated from them. But if I'm mistaken, disregard.

packages/product-components/src/components/SelectAssetModal/SendTokenTabs.tsx
```diff
@@ -0,0 +1,57 @@
...
+import styled from 'styled-components';
+import { CheckableTag } from './CheckableTag';
+
+interface NetworkTabsWrapperProps {
+    $elevation: Elevation;
+}
+
+const NetworkTabsWrapper = styled.div<NetworkTabsWrapperProps>`
+    margin-left: -${spacingsPx.md};
+    width: calc(100% + ${spacings.md * 2}px);
```
Lemonexe: Seems more elegant to me than `calc`.
And works the same - the fine divider line spans the whole width :heavy_check_mark: 
```suggestion
    margin: 0 -${spacingsPx.md};
```

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
Lemonexe: Oh no, the `searchCategory` comes from here. That's a mess :see_no_evil:

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
tomasklim: true, thank you! it should be `account.networkType`

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
tomasklim: Yea, now it renders `≈` even though fiat rate is not available

tomasklim: I've added formattedCryptoAmount for the balance above 

```
                                    <HiddenPlaceholder>
                                        <FormattedCryptoAmount
                                            value={selectedOption.balance ?? '0'}
                                            symbol={selectedOption.symbol}
                                        />
                                    </HiddenPlaceholder>
```

tomasklim: I can try something like

```ts
const renderFiatValueComponent = () => {
    const fiatValue = (
        <FiatValue
            tokenAddress={option.contractAddress ? (option.contractAddress as TokenAddress) : undefined}
            amount={option.balance ?? '0'}
            symbol={option.networkSymbol}
        />
    );
    
    return fiatValue;
};
```

and use it like this
```
                    {renderFiatValueComponent() && (
                        <> ≈ {renderFiatValueComponent()} </>
                    )}
```

but I am not sure if it would work

Lemonexe: Oh I get it now, it's more complex than I thought at first, because `FiatValue` can render null if fiat rate is not available, which is unknown here :confused:

Lemonexe: it'd be ideal to extract the `FiatValue` inner logic to hook, so we can access the logic without rendering the component and checking its return for `null`. But if we have to, I'd render it to a constant:
```
const fiatValue = <FiatValue ...props>;
```
and then use it like
```
{fiatValue !== null ? <> ≈ {fiatValue} </> : null}
```
Which is the same thing as you have written but simpler.

But please add a todo to extract the `FiatValue` logic, because this is still not very nice.

packages/suite/src/components/suite/copy/TokenAddressRow.tsx
tomasklim: 🚀

suite-common/wallet-utils/src/tokenUtils.ts
tomasklim: yes, it is available only for solana

packages/suite/src/actions/suite/copyAddressActions.ts
tomasklim: oh absolutely true

tomasklim:
> ❓ concerning behavior when I have no tokens, just currency - it feels weird to me that it displays the 🚫 cursor on the whole card. As if the UI is telling me that Polygon itself is blocked from sending. Maybe 🚫 cursor should be only on the arrow, then it's telling me I can't select (as there isn't anything to select) blocked cursor

I removed the Select component, it was useless in this case

packages/suite/src/actions/suite/copyAddressActions.ts
tomasklim: yes it is

tomasklim: it is now longer but who cares

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
tomasklim: I totally agree

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
tomasklim: I've done it with the logic you mentioned in slack, i will create and issue to revisit this, this is not the only occurence

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
tomasklim: no user would ever search by cryptoId

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
tomasklim:
```
   const filteredData = data.filter(item => {
        if (
            searchCategory &&
            item.coingeckoId !== searchCategory.coingeckoId &&
            item.coingeckoId !== searchCategory.coingeckoNativeId
        ) {
            return false;
        }

        const contractAddress = item.contractAddress || undefined;
        const searchFor = (property: string | undefined) =>
            property?.toLocaleLowerCase().includes(search.toLocaleLowerCase());

        return (
            searchFor(item.name) ||
            searchFor(item.badge) ||
            (!disableSearchByNetwork && searchFor(item.symbol)) ||
            (disableSearchByNetwork ? searchFor(contractAddress) : searchFor(item.networkSymbol))
        );
    });
```
I believe this is it

Lemonexe: Yeah that's what I meant.

tomasklim:
Screenshot 2024-11-07 at 13 48 44

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
tomasklim: done

packages/product-components/src/components/SelectAssetModal/TokenTabs.tsx
```diff
@@ -0,0 +1,54 @@
+import { ReactNode } from 'react';
+
+import styled from 'styled-components';
+
+import { Icon, Row, useElevation } from '@trezor/components';
+import { borders, Elevation, mapElevationToBorder, spacings, spacingsPx } from '@trezor/theme';
+
+import { CheckableTag } from './CheckableTag';
+
+const NetworkTabsWrapper = styled.div<{ $elevation: Elevation }>`
```
Lemonexe: nit: rename to `TokenTabsWrapper`

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,412 @@
...
+                                            />
+                                        </Row>
+                                    </Text>
+                                </Row>
+                            )}
+                        </Column>
+                    </Row>
+
+                    <Tooltip
+                        content={<Translation id="TR_TOKEN_NOT_FOUND" />}
```
Lemonexe: Currently this message is shown always, even when you **have** tokens!
I was curious so I took a look at `Tooltip` component, and apparently the `disabled` prop only affects cursor, it doesn't disable the tooltip.
image
```suggestion
                        content={isDisabled ? <Translation id="TR_TOKEN_NOT_FOUND" /> : null}
```

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,412 @@
...
+                        disabled={isDisabled}
+                    >
+                        <CursorWrapper $isDisabled={isDisabled}>
```
Lemonexe: Followup to comment above: please take a look at `Tooltip` `cursor` prop. Maybe it can be used instead of `CursorWrapper`. But maybe not, idk :slightly_smiling_face:

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,412 @@
...
+    enhanceTokensWithRates,
+    getTokens,
+    sortTokensWithRates,
+} from 'src/utils/wallet/tokenUtils';
+import { useDispatch, useSelector, useTranslation } from 'src/hooks/suite';
+import { useSendFormContext } from 'src/hooks/wallet';
+import { TokenAddressRow } from 'src/components/suite/copy/TokenAddressRow';
+import { copyAddressToClipboard, showCopyAddressModal } from 'src/actions/suite/copyAddressActions';
+
+// eslint-disable-next-line local-rules/no-override-ds-component
```
Lemonexe: I don't understand this component.
If you need to override design system component, there should be an explaantion why.
But also, why the check for  `$elevation === 1`? Do you expect it to vary?

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,412 @@
...
+    height: ITEM_HEIGHT,
+});
+
+const buildTokenOptions = (
+    accountTokens: Account['tokens'],
+    symbol: Account['symbol'],
+    coinDefinitions: TokenDefinitions['coin'],
+    activeTokenTab: TokenTab['tab'],
+): AssetProps[] => {
+    const result = [];
```
Lemonexe: this is implicitly typed as `any`
```suggestion
    const result: AssetProps[] = [];
```

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,412 @@
...
+        clearErrors(amountInputName);
+        // remove Amount if isSetMaxActive or ETH data options are enabled
+        if (isSetMaxActive || dataEnabled) setAmount(outputId, '');
+        // remove ETH data option
+        if (dataEnabled) toggleOption('ethereumData');
+        // compose (could be prevented because of Amount error from re-validation above)
+        composeTransaction(amountInputName);
+    };
+
+    const isDisabled = !account.tokens?.length;
```
Lemonexe: I'd rather name it `hasNoTokens`. I think naming "what it is" is clearer than "what is it used for".
For example the `const onOpenSelectAssetModal` two lines below will make more sense :slightly_smiling_face:

packages/suite/src/actions/suite/copyAddressActions.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const copyAddressToClipboard = (address: string) => (dispatch: Dispatch) => {
+    const result = copyToClipboard(address);
+
+    const isSuccess = result === true;
+
+    if (isSuccess) {
+        dispatch(notificationsActions.addToast({ type: 'copy-to-clipboard' }));
+    }
+};
```
Lemonexe, line 28: :+1:

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -34,110 +28,43 @@ export interface SelectAssetOptionGroupProps {
...
-    onFavoriteClick?: (isFavorite: boolean) => void;
-    onClose: () => void;
-}
-
-interface AssetProps
-    extends Pick<SelectAssetOptionCurrencyProps, 'coingeckoId' | 'contractAddress'> {
-    cryptoId: string;
-    symbol: string;
-    name: string;
+export interface AssetProps
```
Lemonexe: I can see that only `AssetProps` is actually used here, while `SelectAssetOptionProps` (and its both constituents) are used in Coinmarket.
And `AssetProps` is defined as subset of `SelectAssetOptionCurrencyProps`..

:bulb: it'd make more sense to me to turn it around - define `SelectAssetOptionCurrencyProps` as extension of `AssetProps` with extra properties. Then you can move it to coinmarket, and have only `AssetProps` here.

Also, you could define `Omit<AssetProps, 'height'>` so you don't have to repeat it so much..

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -34,110 +28,43 @@ export interface SelectAssetOptionGroupProps {
...
-    return networkNetworkGroups.length + networkCurrencies.length;
-};
+export interface SelectAssetModalProps {
+    options: AssetProps[];
+    searchPlaceholderText?: string;
+    onSelectAsset: (selectedAsset: Omit<AssetProps, 'height'>) => Promise<void> | void;
+    onClose: () => void;
+    searchInput?: ReactNode;
+    filterTabs?: ReactNode;
+    noItemsAvailablePlaceholder: { heading: string | ReactNode; body?: string | ReactNode };
```
Lemonexe: `string` is assignable to `ReactNode`
```suggestion
    noItemsAvailablePlaceholder: { heading: ReactNode; body?: ReactNode };
```

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -34,110 +28,43 @@ export interface SelectAssetOptionGroupProps {
...
-                    group.coingeckoId === item.coingeckoId ||
-                    group.coingeckoNativeId === item.coingeckoId,
-            ),
-    );
 
-    return networkNetworkGroups.length + networkCurrencies.length;
-};
+export interface SelectAssetModalProps {
+    options: AssetProps[];
+    searchPlaceholderText?: string;
```
Lemonexe: prop is unused anywhere

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -34,110 +28,43 @@ export interface SelectAssetOptionGroupProps {
...
     onClose,
+    filterTabs,
+    searchInput,
+    noItemsAvailablePlaceholder,
 }: SelectAssetModalProps) => {
     const intl = useIntl();
-    const [search, setSearch] = useState('');
-    const [searchCategory, setSearchCategory] = useState<SelectAssetSearchCategoryType>(null); // coingeckoNativeId as fallback for ex. polygon
+
     const [end, setEnd] = useState(options.length);
```
Lemonexe, line 48: This seems useless. You only assign to it, never read `end`. Or is it there just to trigger rerender?

packages/product-components/src/components/SelectAssetModal/AssetItemNotFound.tsx
```diff
@@ -1,61 +1,27 @@
...
 
 import { Column, Paragraph, Text } from '@trezor/components';
 import { spacings } from '@trezor/theme';
 
-import { SelectAssetNetworkProps, SelectAssetSearchCategoryType } from './SelectAssetModal';
-
 interface AssetItemNotFoundProps {
-    searchCategory: SelectAssetSearchCategoryType;
-    networkCategories: SelectAssetNetworkProps[];
+    noItemsAvailablePlaceholder: { heading: string | ReactNode; body?: string | ReactNode };
```
Lemonexe: `string` is assignable to `ReactNode`
```suggestion
    noItemsAvailablePlaceholder: { heading: ReactNode; body?: ReactNode };
```

packages/product-components/src/components/SelectAssetModal/AssetItemNotFound.tsx
```diff
@@ -1,61 +1,27 @@
...
-            </Text>
+}: AssetItemNotFoundProps) => (
+    <Column
+        alignItems="center"
+        justifyContent="center"
+        height={listHeight}
+        minHeight={listMinHeight}
+    >
+        <Text typographyStyle="body">{noItemsAvailablePlaceholder.heading}</Text>
+        {noItemsAvailablePlaceholder.body && (
```
Lemonexe, line 24: :heart:

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,412 @@
...
+
+const buildTokenOptions = (
+    accountTokens: Account['tokens'],
+    symbol: Account['symbol'],
+    coinDefinitions: TokenDefinitions['coin'],
+    activeTokenTab: TokenTab['tab'],
+): AssetProps[] => {
+    const result = [];
+
+    if (activeTokenTab === 'tokens') {
```
Lemonexe, line 80: I'd add comment `// this represents native currency`
Maybe it's just me being slow, but it took me a while to figure it out from the properties `symbol`, `networkSymbol`, `cryptoName`. Those are clear when building a token, but here it can be kinda confusing what means what, so a comment can save time to future readers :smile:

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
```diff
@@ -0,0 +1,412 @@
...
+            box-shadow: ${theme.boxShadowBase};
+        `}
+    cursor: ${({ $isDisabled }) => ($isDisabled ? 'default' : 'pointer')};
+`;
+
+export const CursorWrapper = styled.div<{ $isDisabled: boolean }>`
+    cursor: ${({ $isDisabled }) => ($isDisabled ? 'not-allowed' : 'pointer')};
+`;
+
+const getToken = (token: TokenInfo, symbol: Account['symbol'], shouldTryToFetch: boolean) => ({
```
Lemonexe: I'd rename it to `createTokenOption`, or something like that. Otherwise, `getToken` looks very familiar to imported `getTokens`, but they do something very different.

packages/suite-data/files/translations/en.json
```diff
@@ -1872,7 +1872,7 @@
   "TR_TOKENS_EMPTY": "No tokens... yet.",
   "TR_TOKENS_EMPTY_CHECK_HIDDEN": "No tokens. They may be hidden.",
   "TR_TOKENS_SEARCH_TOOLTIP": "Search by token, symbol, or contract address.",
-  "TR_TOKEN_NOT_FOUND": "Token not found",
+  "TR_TOKEN_NOT_FOUND": "No token found",
```
Lemonexe: This will be regenerated by Crowdin, right? But I guess no big deal :shrug:

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
```diff
@@ -1,29 +1,23 @@
...
     type: 'currency';
-    value: string; // CryptoId (networkId + contractAddress)
-    label: string; // token shortcut
-    cryptoName: string | undefined; // full name
-    coingeckoId: string; // CryptoId (networkId)
-    contractAddress?: string; // CryptoId (contractAddress)
+    symbol: string;
+    label?: string;
+    balance?: string;
+    badge?: string | ReactNode;
```
Lemonexe: `string` is assignable to `ReactNode`
```suggestion
    badge?: ReactNode;
```

Lemonexe (approved):
:eyes: CR round 2 after complete rework:

:information_source: I reviewed only the `product-components` and Send Form side, **not** the coinmarket side (relying on CR by Invity).

LGTM, works as supposed to, with [just one minor glitch](https://github.com/trezor/trezor-suite/pull/14799#discussion_r1836569153) :heavy_check_mark:

:memo: Codewise: quite a few new comments, but mostly nits, nothing major. Really, overall this looks much much better than the previous version :heart: :+1: :tada: 
Thank you for taking the time to untangle the `SelectAssetModal`, because it was really so much logic at wrong place.
The abstraction with `searchInput`, `noItemsAvailablePlaceholder`, `filterTabs` is great! This is the React way :sunglasses: 

After resolving comments I think it's good to go :rocket:

adderpositive (approved):

adderpositive (approved):

packages/product-components/src/components/TokenIconSet/TokenIconSet.tsx
```diff
@@ -60,6 +60,7 @@ export const TokenIconSet = ({ network, tokens }: TokenIconSetProps) => {
                     contractAddress={getContractAddressForNetwork(network, token.contract)}
                     placeholder={token.symbol?.toUpperCase() ?? ''}
                     placeholderWithTooltip={false}
+                    // TODO: shouldFetch?
```
adderpositive, line 63: I would do it right now. 😃

adderpositive (approved):

packages/suite-data/files/translations/en.json
tomasklim: yes

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
tomasklim: 🤷‍♂️ will check

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
tomasklim: I think it is not there. I dont see it in the code 😀

packages/product-components/src/components/SelectAssetModal/SelectAssetModal.tsx
tomasklim: great idea, done

packages/suite/src/views/wallet/send/Outputs/TokenSelect/TokenSelect.tsx
tomasklim: it does not make sense. I think it was some hack for react-select component

tomasklim: I was able to completely remove this component

PR #14800, opened 2024-10-10
chore(wallet-core): deduplicate code with getBinFilesBaseUrlThunk

Lemonexe:
## Description

Create one thunk to get path to static bin files, currently you have to do two different approaches for desktop & web. 

## Screenshots

I console logged the `const binFilesBaseUrl` in `checkFirmwareHashThunk.ts ` and `firmwareThunks.ts`:
Screenshot from 2024-10-10 14-54-58
Screenshot from 2024-10-10 14-58-17

Firmware update still works:
Screenshot from 2024-10-10 14-54-43

### DIFF
--- a/packages/suite/src/reducers/desktop/index.ts
+++ b/packages/suite/src/reducers/desktop/index.ts
@@ -18,5 +18,3 @@ export const desktopReducer = (
             return state;
     }
 };
-
-export const selectDesktopBinDir = (state: DesktopState) => state?.paths?.binDir;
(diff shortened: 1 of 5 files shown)


packages/suite/src/reducers/desktop/index.ts
```diff
@@ -18,5 +18,3 @@ export const desktopReducer = (
             return state;
     }
 };
-
-export const selectDesktopBinDir = (state: DesktopState) => state?.paths?.binDir;
```
Lemonexe, line 22: unused

peter-sanderson (approved):

PR #14803, opened 2024-10-11
chore(suite-native): send feature flag used for btc like coins

### LINKED ISSUE #14759: Send coins - adjust feature flag
- [ ] adjust feature flag so the send flow is always available in production build for ordinary users.
- [ ] Bitcoin based coins are possible to send after enabling feature flag
- [ ] It is still impossible to send any other coins.
- [ ] make sure that iOS user can not enter send flow

PeKne:
## Description
- BTC (+ TEST and REGTEST) send is available to all users of all builds
- Bitcoin based coins send can be enabled via feature flag
- send of other networks is still always disabled
## Related Issue

Resolve #14759 

## Screenshots:
https://github.com/user-attachments/assets/8f42161e-9386-4d06-8dcf-40cd73737184

### DIFF
--- a/suite-native/feature-flags/src/featureFlagsSlice.ts
+++ b/suite-native/feature-flags/src/featureFlagsSlice.ts
@@ -5,7 +5,7 @@ import { isDebugEnv, isDetoxTestBuild, isDevelopOrDebugEnv } from '@suite-native
 
 export const FeatureFlag = {
     IsDeviceConnectEnabled: 'isDeviceConnectEnabled',
-    IsSendEnabled: 'isSendEnabled',
+    IsBitcoinLikeSendEnabled: 'isBitcoinLikeSendEnabled',
     IsRegtestEnabled: 'isRegtestEnabled',
     IsPolygonEnabled: 'IsPolygonEnabled',
     IsBscEnabled: 'IsBscEnabled',
@@ -20,7 +20,7 @@ export type FeatureFlagsRootState = {
 
 export const featureFlagsInitialState: FeatureFlagsState = {
     [FeatureFlag.IsDeviceConnectEnabled]: isAndroid() || isDebugEnv(),
-    [FeatureFlag.IsSendEnabled]: isAndroid() && isDevelopOrDebugEnv(),
+    [FeatureFlag.IsBitcoinLikeSendEnabled]: isAndroid() && isDevelopOrDebugEnv(),
     [FeatureFlag.IsRegtestEnabled]: isDebugEnv() || isDetoxTestBuild(),
     [FeatureFlag.IsPolygonEnabled]: false,
     [FeatureFlag.IsBscEnabled]: false,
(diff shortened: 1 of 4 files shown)


suite-native/module-accounts-management/src/components/TransactionListHeader.tsx
```diff
@@ -99,7 +101,7 @@ export const TransactionListHeader = memo(
         tokenContract,
     }: AccountDetailHeaderProps) => {
         const navigation = useNavigation<AccountsNavigationProps>();
-        const [isSendEnabled] = useFeatureFlag(FeatureFlag.IsSendEnabled);
+        const [IsDeviceConnectEnabled] = useFeatureFlag(FeatureFlag.IsDeviceConnectEnabled);
```
matejkriz:
```suggestion
        const [isDeviceConnectEnabled] = useFeatureFlag(FeatureFlag.IsDeviceConnectEnabled);
```

suite-native/feature-flags/src/featureFlagsSlice.ts
```diff
@@ -20,15 +21,14 @@ export type FeatureFlagsRootState = {
...
-    [FeatureFlag.IsSendEnabled]: isAndroid() && isDevelopOrDebugEnv(),
+    [FeatureFlag.IsBitcoinLikeSendEnabled]: isAndroid() && isDevelopOrDebugEnv(),
     [FeatureFlag.IsRegtestEnabled]: isDebugEnv() || isDetoxTestBuild(),
     [FeatureFlag.IsPolygonEnabled]: false,
     [FeatureFlag.IsBscEnabled]: false,
 };
 
 export const featureFlagsPersistedKeys: Array<keyof FeatureFlagsState> = [
     FeatureFlag.IsDeviceConnectEnabled,
-    FeatureFlag.IsSendEnabled,
```
matejkriz, line 31: Is it intentional, that `IsBitcoinLikeSendEnabled` is not persisted? It doesn't actually matter if it's by default true on develop, but it could be confusing...

suite-native/feature-flags/src/featureFlagsSlice.ts
```diff
@@ -47,5 +47,26 @@ export const featureFlagsSlice = createSlice({
 export const selectIsFeatureFlagEnabled = (state: FeatureFlagsRootState, key: FeatureFlag) =>
     state.featureFlags[key];
 
+const SEND_COINS_WHITELIST: Readonly<NetworkSymbol[]> = ['btc', 'test', 'regtest'];
+
+export const selectIsNetworkSendFlowEnabled = (
```
matejkriz: nitpick: It seems to me it belongs more in the module-send or module-accounts-management, but it can stay here. It is temporary anyway...

matejkriz (changes requested):

suite-native/feature-flags/src/featureFlagsSlice.ts
PeKne: 879afffe7b

suite-native/feature-flags/src/featureFlagsSlice.ts
PeKne: You are right. It should not be here. 879afffe7b

suite-native/module-accounts-management/src/components/TransactionListHeader.tsx
PeKne: 879afffe7b

matejkriz (approved):

PR #14807, opened 2024-10-11
feat(suite): add coin selection prompt to empty passphrase wallet

### LINKED ISSUE #14229: Passphrase empty: maybe enable some coins?
**Problem**

User enters passphrase expecting money
Discovery doesn’t find anything
They perhaps have something, but the coin is not enabled

**Opportunity**

Explain user to perhaps enable the coin manually

**Design and spec**
When the user enters a passphrase and confirms it on device (or enters it on it right away), we'll check if there's any balance or transaction history on all currently enabled coins.
If nothing is found, we will show a "This passphrase is empty" state of the wallet switcher.

Please add a new small element to the "Expecting passphrase wallet with funds?" part of it as per [design](https://www.figma.com/design/wGshbwWuR7shAA1i1UBqR8/View-only?m=auto&node-id=3614-43910&t=C7CKBVPhnQcDIlyX-1).

Image

It shows all currently enabled coins and an [+ Add] button
(button not show if all coins are enabled; in case of Bitcoin only FW, don't show this new element at all as nothing else can be checked)

At this MVP solution, the [+Add] button will simply lead the user to Settings > Coins to set up the coins, they will have to manually get back to the Passphrase entry afterwards.
_______________________________________________________________________________
In case the user will continue to open an empty Passphrase wallet instead, we'll show them a new Passphrase wallet empty state including a similar new small element showing coins checked as per [design](https://www.figma.com/design/wGshbwWuR7shAA1i1UBqR8/View-only?m=auto&node-id=3614-43717&t=C7CKBVPhnQcDIlyX-1).
Similarly as in the above described wallet switcher, it should not be shown for the Bitcoin only FW.
Also here, the [+Add] button will simply lead the user to Settings > Coins to set up the coins.

Image

seibei-iguchi:
- added coin selection prompt to empty passphrase wallet
- refactored DS components overrides
- removed last instances of custom renderer for device selector modals

### Before
Screenshot 2024-10-14 at 10 48 09
Screenshot 2024-10-14 at 10 48 27

### After
Screenshot 2024-10-14 at 10 46 38
Screenshot 2024-10-14 at 10 45 56


tomasklim (changes requested):
Networks overflow in case there is a lot of them activated
Screenshot 2024-10-14 at 11 04 39
Screenshot 2024-10-14 at 11 02 12

Do we need different string for `Checked coins` and `Checked balances on:`?

Hermez-cz:
> Networks overflow in case there is a lot of them activated Screenshot 2024-10-14 at 11 04 39 Screenshot 2024-10-14 at 11 02 12
> 
> Do we need different string for `Checked coins` and `Checked balances on:`?

Let's have `Checked balances on:` for both case please.

packages/components/src/components/Flex/Flex.tsx
```diff
@@ -147,6 +147,7 @@ export type FlexProps = AllowedFrameProps & {
     className?: string;
     onClick?: () => void;
     'data-testid'?: string;
+    as?: string;
```
enjojoy, line 150: What is this needed for?

packages/components/src/components/Icon/Icon.tsx
enjojoy, line 1:
```js
const SVG = styled(ReactSVG)`
    display: flex;
    align-items: center;
    justify-content: center;

    div {
        display: flex;
        align-items: center;
        justify-content: center;
    }

    path {
        transition:
            stroke 0.15s,
            fill 0.15s;
    }
` as typeof ReactSVG; 
```

Is it possible to partially replace it with Row or Column component with flex props?

packages/suite/src/views/dashboard/PortfolioCard/PortfolioCard.tsx
```diff
@@ -19,6 +19,7 @@ import { selectLocalCurrency } from 'src/reducers/wallet/settingsReducer';
 import { hasBitcoinOnlyFirmware } from '@trezor/device-utils';
 
 const Body = styled.div`
+    display: flex;
     align-items: center;
```
enjojoy: Can you replace it with Flex component?

enjojoy (approved):
You have 2 linter warnings, and I added some nit comments. Other than that tested, previous comments seem to be implemented well.

LGTM🚀

tomasklim (approved):

packages/components/src/components/Flex/Flex.tsx
seibei-iguchi: For semantic reasons, where I want the Row be for example a section or header HTML element.

packages/components/src/components/Icon/Icon.tsx
seibei-iguchi: I think it wouldn't work in this case.

PR #14820, opened 2024-10-14
chore(suite): tweak suite fw hash selectors

Lemonexe:
## Description

Codestyle cleanup after PR #14766

### DIFF
--- a/packages/suite/src/reducers/suite/suiteReducer.ts
+++ b/packages/suite/src/reducers/suite/suiteReducer.ts
@@ -10,7 +10,13 @@ import { isDeviceAcquired } from '@suite-common/suite-utils';
 import { discoveryActions, DeviceRootState, selectDevice } from '@suite-common/wallet-core';
 import { versionUtils } from '@trezor/utils';
 import { isWeb } from '@trezor/env-utils';
-import { TRANSPORT, TransportInfo, ConnectSettings } from '@trezor/connect';
+import {
+    TRANSPORT,
+    TransportInfo,
+    ConnectSettings,
+    FirmwareHashCheckError,
+    FirmwareRevisionCheckError,
+} from '@trezor/connect';
 
 import { getIsTorEnabled, getIsTorLoading } from 'src/utils/suite/tor';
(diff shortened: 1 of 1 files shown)


packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -460,16 +466,20 @@ export const selectFirmwareRevisionCheckError = (state: StateForFirmwareChecks)
...
-    return error ? !['cannot-perform-check-offline', 'other-error'].includes(error) : false;
+    return error !== null ? !softErrors.includes(error) : false;
 };
 
 /**
  * Get firmware hash check error, or null if check was successful / skipped.
  */
 export const selectFirmwareHashCheckError = (state: StateForFirmwareChecks) => {
     const device = selectDevice(state);
-    if (!isDeviceAcquired(device) || !device.authenticityChecks) return false;
```
Lemonexe, line 472: This should be `null`, see `selectFirmwareRevisionCheckError`.

I didn't catch it because I didn't do `error !== null` as I should have

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -460,16 +466,20 @@ export const selectFirmwareRevisionCheckError = (state: StateForFirmwareChecks)
  */
 const selectIsFirmwareRevisionCheckEnabledAndFailed = (state: StateForFirmwareChecks): boolean => {
     const error = selectFirmwareRevisionCheckError(state);
+    const softErrors: FirmwareRevisionCheckError[] = [
+        'cannot-perform-check-offline',
+        'other-error',
+    ];
 
-    return error ? !['cannot-perform-check-offline', 'other-error'].includes(error) : false;
```
Lemonexe, line 464: I didn't realize, but this was typed as `(string []).includes(FirmwareRevisionCheckError)` so I could write `'other-errooooor'` and TS would be fine, hence `softErrors` declaration.

komret (approved):

PR #14825, opened 2024-10-14
feat(suite-native): correct network name for tokens in account detail

### LINKED ISSUE #14191: Polygon - Make account detail work with all evms not only ethereum
- [x] textual warning mentioning directly ethereum while included tokens, should mention actual network

vytick:
Reflect actual network while showing include token txns in account detail

## Related Issue

Resolve #14191


PeKne (approved):
🚀

PR #14827, opened 2024-10-14
feat: remove suport for BIP-69, randomize inputs and outputs 

### LINKED ISSUE #10765: Abandon the use of BIP-69

peter-sanderson:
This PR removes the support for BIP-69. 

TODO:
- [ ] use https://github.com/trezor/trezor-suite/pull/14899 and drop code from this PR
- [ ] add sorting strategies: `'bip69' | 'random' | 'disabled'` (or similar naming) so we dont break 3rd parties
- [ ] do not touch original tests, add new for random sorting
- [ ] for random sorting test use mock of randomness to have stable fixtures

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/10765


packages/utxo-lib/tests/compose.test.ts
```diff
@@ -14,10 +14,37 @@ describe('composeTx', () => {
...
         const network = getNetwork(f.request.network);
         const request = { ...f.request, network };
-        const result: any = { ...f.result };
+        const expected = { ...f.result };
+
         it(f.description, () => {
             const tx = composeTx(request as any);
-            expect(tx).toEqual(result);
+
+            const deRandomize = () => {
```
peter-sanderson, line 22: I am not particularly happy about this (quite complicated test), but I tried to utilize existing test-fixtures and test-cases to verify I haven't broke anything.

packages/utxo-lib/src/getRandomInt.ts
```diff
@@ -0,0 +1,25 @@
...
+        );
+    }
+
+    if (min >= max) {
+        throw new RangeError(
+            `The value of "max" is out of range. It must be greater than the value of "min" (${min}). Received ${max}`,
+        );
+    }
+
+    const randomValue = parseInt(randomBytes(4).toString('hex'), 16);
```
peter-sanderson: is `4` enough?

packages/utxo-lib/tests/compose.test.ts
peter-sanderson: TODO: mock random, make it stable

peter-sanderson:
=> https://github.com/trezor/trezor-suite/pull/14917

PR #14828, opened 2024-10-14
Improve random int

peter-sanderson:
- renames `getRandomNumberInRange` to `getWeakRandomNumberInRange`
- add `randomInt` that uses `crypto` lib (because `randomInt` from crypto is not supported by `crypto-browserify` => https://github.com/browserify/crypto-browserify/issues/224)
- `arrayShuffle`  now uses  `randomInt` and therefore `crypto` lib


packages/suite-build/configs/base.webpack.config.ts
```diff
@@ -44,7 +44,7 @@ const config: webpack.Configuration = {
             src: path.resolve(__dirname, '../../suite/src/'),
         },
         fallback: {
-            // Polyfills crypto API for NodeJS libraries in the browser. 'crypto' does not run without 'stream'
+            // Polyfills crypto API for Node.js libraries in the browser. 'crypto' does not run without 'stream'
```
peter-sanderson, line 47: I received a notice, that this is the official™ spelling is `Node.js`. :shrug:

packages/transport-test/package.json
```diff
@@ -14,7 +14,7 @@
         "test:e2e:new-bridge:hw": "USE_HW=true USE_NODE_BRIDGE=true yarn test:e2e:bridge",
         "test:e2e:new-bridge:emu": "USE_HW=false USE_NODE_BRIDGE=true yarn test:e2e:bridge",
         "build:e2e:api:node": "yarn esbuild ./e2e/api/api.test.ts --bundle --outfile=./e2e/dist/api.test.node.js --platform=node --target=node18 --external:usb",
-        "build:e2e:api:browser": "yarn esbuild ./e2e/api/api.test.ts --bundle --outfile=./e2e/dist/api.test.browser.js --platform=browser --external:usb  && cp e2e/ui/api.test.html e2e/dist/index.html",
+        "build:e2e:api:browser": "yarn esbuild ./e2e/api/api.test.ts --bundle --alias:crypto=crypto-browserify --alias:stream=stream-browserify --outfile=./e2e/dist/api.test.browser.js --platform=browser --external:usb  && cp e2e/ui/api.test.html e2e/dist/index.html",
```
peter-sanderson, line 17: Took me a while to figure this out: https://esbuild.github.io/api/#alias

packages/connect-web/package.json
```diff
@@ -45,7 +45,8 @@
     "dependencies": {
         "@trezor/connect": "workspace:*",
         "@trezor/connect-common": "workspace:*",
-        "@trezor/utils": "workspace:*"
+        "@trezor/utils": "workspace:*",
+        "crypto-browserify": "^3.12.0"
```
mroz22, line 49: dependency or devDependency ?  🤔

packages/connect-popup/package.json
```diff
@@ -25,6 +25,7 @@
         "@trezor/device-utils": "workspace:*",
         "@trezor/transport": "workspace:*",
         "@trezor/urls": "workspace:*",
+        "crypto-browserify": "^3.12.0",
```
mroz22, line 28: and what about `stream-browserify`. this is transient?

peter-sanderson: yes, I used the same approach as was in `suite` package

packages/connect-web/package.json
peter-sanderson: It depends... :rofl:

mroz22: it should, imho, be only `devDependency`. because we only need it to produce `lib` using `build:lib`. it shouldn't be needed in the runtime of this package at all I believe.

mroz22: connect-web itself only imports scheduleAction, createDeferred and createDeferredManager

peter-sanderson:
=> https://github.com/trezor/trezor-suite/pull/14878

PR #14830, opened 2024-10-14
feat(suite): hide fw hash check banner for old versions

Lemonexe:
## Description

- Remove the SuiteBanner for FW hash case `check-unsupported` (old firmware)
- new TS util

## Related Issue

Followup to https://github.com/trezor/trezor-suite-private/issues/100

## Screenshots

if
if

else
else

### DIFF
--- a/packages/type-utils/src/utils.ts
+++ b/packages/type-utils/src/utils.ts
@@ -58,3 +58,17 @@ export type PartialRecord<K extends keyof any, T> = {
 
 // distributive conditional types to the rescue! This way we can infer union literal type from ReturnType but exclude undefined
 export type DefinedUnionMember<T> = T extends string ? T : never;
+
+/**
+ * Type-guard if a value is a subset of an array.
+ * Useful to narrow down a union type `value` to a subset of the union (typeof `arr`).
+ *
+ * Example:
+ * type Variant = 'a' | 'b' | 'c' | 'd';
+ * const skippedVariants = ['a', 'b'] satisfies Variant[];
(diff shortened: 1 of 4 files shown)


packages/type-utils/src/utils.ts
```diff
@@ -58,3 +58,17 @@ export type PartialRecord<K extends keyof any, T> = {
...
+ *
+ * Example:
+ * type Variant = 'a' | 'b' | 'c' | 'd';
+ * const skippedVariants = ['a', 'b'] satisfies Variant[];
+ * if(isArrayMember(variant, skippedVariants)) // variant is 'a' | 'b', else 'c' | 'd'
+ */
+export const isArrayMember = <Value extends string, Subset extends Value>(
+    value: Value,
+    arr: Subset[],
+): value is Subset => arr.some(v => v === value);
```
Lemonexe, line 74: :bulb: `arr.some` is better than `arr.includes`.

Unfortunately, `Array.includes` is declared such that it only allows `value` to be `(typeof arr)[number]`,
so you have to either typecast `(arr as string[]).includes`, or `.includes(value as Subset)` :face_with_diagonal_mouth: 
This tiny detail has really bugged me about TS, see the many `.includes` occurrences in our repo :sweat_smile: :facepalm:  :see_no_evil: 
[Example..](https://github.com/trezor/trezor-suite/blob/39c0c639f5dbf486b21f556e0aa6408287126f84/suite-common/wallet-core/src/stake/stakeTypes.ts#L30)
But this solves it! :slightly_smiling_face:

peter-sanderson (approved):

PR #14833, opened 2024-10-14
Show confirmation pill for PIN input during FW update

komret:
When user is prompted to enter PIN during FW update, Suite did not direct them towards the device.

To test this, lock your Trezor by holding a button/screen and then start FW update. You will be prompted  to enter your PIN twice, Suite should tell you to "Confirm on Trezor" both times.

Video showing the bug:

https://github.com/user-attachments/assets/848ae136-5362-428b-97cb-eff7720b2f1d

#### Fixed:

Before installation, PIN entry on Trezor screen:
Screenshot 2024-10-14 at 13 57 08

After installation, PIN entry on Trezor screen. The "Select Trezor" button does not do anything in this case, but there seems to be no way to differentiate this from another situation when the button is necessary :/ The button only appears on web.
Screenshot 2024-10-14 at 13 59 21

### DIFF
--- a/packages/suite/src/hooks/suite/useFirmware.ts
+++ b/packages/suite/src/hooks/suite/useFirmware.ts
@@ -77,10 +77,10 @@ export const useFirmware = () => {
                 firmware.uiEvent.payload.code,
             )) ||
         // When a PIN-protected device reconnects to normal mode after installation, PIN is requested.
-        // There is a false positive in case such device is wiped (including PIN) during installation,
-        // but there is no straightforward way of detecting that. Some logic could be added to Connect, though.
+        // There is a false positive in case such device is wiped (including PIN) during custom installation.s
         (firmware.uiEvent?.type === UI.FIRMWARE_RECONNECT &&
-            originalDevice?.features?.pin_protection);
+            originalDevice?.features?.pin_protection &&
+            !deviceWillBeWiped);
 
     const showConfirmationPill =
         !showReconnectPrompt &&
--- a/packages/suite/src/hooks/suite/useFirmware.ts
+++ b/packages/suite/src/hooks/suite/useFirmware.ts
@@ -68,19 +68,23 @@ export const useFirmware = () => {
     const deviceWillBeWiped = willDeviceBeWiped();
 
     const confirmOnDevice =
-        // Show the confirmation pill at the start of installation using the "wait" or "manual" method,
-        // after ReconnectDevicePrompt is closed.
+        // Show the confirmation pill before starting the installation using the "wait" or "manual" method,
+        // after ReconnectDevicePrompt is closed and user selects the option to install firmware while in bootloader.
         // Also in case the device is PIN-locked at the start of the process.
         (firmware.uiEvent?.type === DEVICE.BUTTON &&
             firmware.uiEvent.payload.code !== undefined &&
             ['ButtonRequest_FirmwareUpdate', 'ButtonRequest_PinEntry'].includes(
                 firmware.uiEvent.payload.code,
             )) ||
-        // When a PIN-protected device reconnects to normal mode after installation, PIN is requested.
-        // There is a false positive in case such device is wiped (including PIN) during custom installation.s
+        // Show the confirmation pill right after ReconnectDevicePrompt is closed while using the "wait" or "manual" method,
+        // before user selects the option to install firmware while in bootloader
+        // When a PIN-protected device reconnects to normal mode after installation, PIN is requested and the pill is shown.
+        // There is a false positive in case such device is wiped (including PIN) during custom installation.
         (firmware.uiEvent?.type === UI.FIRMWARE_RECONNECT &&
-            originalDevice?.features?.pin_protection &&
-            !deviceWillBeWiped);
+            (firmware.uiEvent.payload.target === 'bootloader' ||
+                (firmware.uiEvent.payload.target === 'normal' &&
+                    originalDevice?.features?.pin_protection &&
+                    !deviceWillBeWiped)));
 
     const showConfirmationPill =
         !showReconnectPrompt &&
--- a/packages/suite/src/hooks/suite/useFirmware.ts
+++ b/packages/suite/src/hooks/suite/useFirmware.ts
@@ -2,12 +2,13 @@ import { useDispatch } from 'react-redux';
 
 import { FirmwareStatus } from '@suite-common/suite-types';
 import { firmwareUpdate, selectFirmware, firmwareActions } from '@suite-common/wallet-core';
-import { DEVICE, FirmwareType, UI } from '@trezor/connect';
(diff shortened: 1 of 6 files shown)


packages/suite/src/hooks/suite/useFirmware.ts
```diff
@@ -48,13 +52,39 @@ export const useFirmware = () => {
         modal.context === MODAL.CONTEXT_DEVICE &&
         modal.windowType === 'ButtonRequest_FirmwareCheck';
 
-    const isCurrentlyBitcoinOnly = hasBitcoinOnlyFirmware(originalDevice);
+    // Device may be wiped during firmware type switch because Universal and Bitcoin-only firmware have different vendor headers,
+    // except T1B1 and T2T1. There may be some false negatives here during custom installation.
+    // TODO: Determine this in Connect.
+    const willDeviceBeWiped = () => {
```
Lemonexe: is there a reason to have this in function?
It adds no semantic value when we have the variable, IMHO this is clear enough:
```
const deviceWillBeWiped = (
            shouldSwitchFirmwareType &&
            deviceModelInternal !== undefined &&
            ![DeviceModelInternal.T1B1, DeviceModelInternal.T2T1].includes(deviceModelInternal)
        );
```

packages/suite/src/hooks/suite/useFirmware.ts
```diff
@@ -48,13 +52,39 @@ export const useFirmware = () => {
...
+            ['ButtonRequest_FirmwareUpdate', 'ButtonRequest_PinEntry'].includes(
+                firmware.uiEvent.payload.code,
+            )) ||
+        // Show the confirmation pill right after ReconnectDevicePrompt is closed while using the "wait" or "manual" method,
+        // before user selects the option to install firmware while in bootloader
+        // When a PIN-protected device reconnects to normal mode after installation, PIN is requested and the pill is shown.
+        // There is a false positive in case such device is wiped (including PIN) during custom installation.
+        (firmware.uiEvent?.type === UI.FIRMWARE_RECONNECT &&
+            (firmware.uiEvent.payload.target === 'bootloader' ||
+                (firmware.uiEvent.payload.target === 'normal' &&
```
Lemonexe, line 83: I don't know if the previous `firmware.uiEvent.payload.method !== 'wait'` is equivalent to `firmware.uiEvent.payload.target === 'bootloader' || firmware.uiEvent.payload.target === 'normal'`
If you are sure about it, then fine, asking just in case, so nothing is missed.

packages/suite/src/hooks/suite/useFirmware.ts
```diff
@@ -22,12 +23,15 @@ export const useFirmware = () => {
...
     const transport = useSelector(state => state.suite.transport);
     const modal = useSelector(state => state.modal);
+    const routeName = useSelector(selectRouteName);
     const { device } = useDevice();
 
     // Device in its state before installation is cached when installation begins.
     // Until then, access device as normal.
     const originalDevice = firmware.cachedDevice || device;
 
+    const shouldSwitchFirmwareType = routeName === 'firmware-type';
```
Lemonexe: I understand the benefit of not having to pass the prop anymore. But this sounds like a future problem. If someone moves the functionality to other page, or adds the functionality at other place, this won't work, because it's hardcoded to one specific route..

I'm fine with it, but at least, I think the `'firmware-type'` should be extracted as constant, which would also be used in `ForegroundAppModal.tsx`, and possibly other places. Idk.

Lemonexe (approved):
LGTM, tested and it works as expected, both switching to btc-only, and updating version.

Just some minor codestyle comments :slightly_smiling_face:

packages/suite/src/hooks/suite/useFirmware.ts
komret: I think the previous condition was either wrong or it worked coincidentally. The new condition is more descriptive and I tried to help it with the comment. But it is still quite messy :(

packages/suite/src/hooks/suite/useFirmware.ts
komret: 7782d9366e8e0553c7c56ed37314d20fa7409996

packages/suite/src/hooks/suite/useFirmware.ts
komret: You're right, this is not nice. I refactored it in 7782d9366e8e0553c7c56ed37314d20fa7409996, but maybe I made it worse by having to pass it to the hook where it could be forgotten (yet I don't want to make it required because some components don't care about it). I could pass it to the functions as previously, but then I'd have to turn even `confirmOnDevice` to a function. I don't like either solution :) What do you think?

packages/suite/src/views/firmware/FirmwareUpdate.tsx
```diff
@@ -1,44 +1,29 @@
...
 
     const heading = shouldSwitchFirmwareType ? 'TR_SWITCH_FIRMWARE' : 'TR_INSTALL_FIRMWARE';
 
     return (
-        <FirmwareModal
-            deviceWillBeWiped={deviceWillBeWiped}
-            heading={heading}
-            install={installTargetFirmware}
-        >
+        <FirmwareModal heading={heading} install={installTargetFirmware}>
```
Lemonexe: Isn't `shouldSwitchFirmwareType` prop missing here? :thinking: 
If it is so, yeah I know, it's a drawback of this solution, it's easy to miss.. :disappointed:
EDIT: yeah I think it should be here. Previously, FirmwareModal relied on `deviceWillBeWiped`. Now it needs `shouldSwitchFirmwareType` instead, because that is used to determine `deviceWillBeWiped` in `useFirmware`

komret: f163fa3a9f0abc8f45d12fa67da2488010c8e650

Lemonexe (approved):

bosomt:
Noticed that its not possible to cancel PIN dialogue from host, no X in pill or in modal.
When you cancel it on device PIN dialogue is displayed twice and you have to cancel it twice to get 
`Installation failed: Error: PIN entry cancelled` error

Is it worht follow up issue @komret  ? Otherwise **QA OK**

image


**Info:**
 - Suite version: desktop 24.11.0 (a337b23e886188d52160cc2c00dfa4a8000801d0)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/24.11.0 Chrome/128.0.6613.162 Electron/32.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: BridgeTransport 3.0.0-bundled.24.11.0

komret:
@bosomt Hmmm yes please, I'll try to fix it.

komret:
The double PIN cancellation seems to be caused by https://github.com/trezor/trezor-suite/pull/14910, not this.

Issue #14852, opened 2024-10-14
Trading - long amount or symbols does not fit to UI nicely


tomasklim:
**Describe the bug**

Version: 24.10.1

![Screenshot 2024-10-14 at 19 46 16](https://github.com/user-attachments/assets/b8732246-cca4-4b70-9afa-25997b24e51d)
![Screenshot 2024-10-14 at 19 46 43](https://github.com/user-attachments/assets/f5d9756d-cf0f-435c-98eb-961d697367d3)
![Screenshot 2024-10-14 at 19 44 10](https://github.com/user-attachments/assets/3c555ffc-1b78-4076-91d5-1ec48d78e65d)
![Screenshot 2024-10-14 at 19 44 59](https://github.com/user-attachments/assets/937c1475-79c5-4207-b7e1-f854718f6568)

bosomt:
**QA OK**

![Image](https://github.com/user-attachments/assets/3f9856df-f357-4dc8-a4de-a647c30aca16)
![Image](https://github.com/user-attachments/assets/b94788d8-c017-46c4-98d3-4cd810871b6f)

**Info:**
 - Suite version: web 24.11.0 (ccc4abd7435e4efabeac8fa846d6e4f7a4e23432)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 5120x1440
 - Device: Trezor T2T1 2.8.1 regular (revision 632b9561559b7ab6824bb7eeac072874e07b7b82)
 - Transport: WebUsbTransport

PR #14862, opened 2024-10-15
14798 send form total value overflows to second row moving the incl fee row for fiat

### LINKED ISSUE #14798: Send form "Total" value overflows to second row moving the "Incl. fee" row for fiat
**Describe the bug**
In Send form, the "Total" that is showing value in crypto can overflow to the second row which is for "Incl. fee" for fiat and that also moves the "Review & Send" button.

**Steps to reproduce:**
1. Go to Ethereum send form
2. Enter for example 1$ as an amount
3. See the result on the right with "Review & Send" button

**Info:**
 - Suite version: desktop 24.10.1 (3cbe786b5c8effd03b331faec20a41ef01d6671b)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.10.1 Chrome/126.0.6478.234 Electron/31.6.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Transport: BridgeTransport 3.0.0-bundled.24.10.1

**Expected result:**
Total for crypto and fiat stays on one line.

**Actual result:**
Total value for crypto overflows to second row for "Incl. fee" value

**Screenshots:**
[2024-10-10 13-16-16.webm](https://github.com/user-attachments/assets/605bd8a4-b462-432f-aba0-9f35fe91c104)

seibei-iguchi:
- fixed fees wrapping in send form
- fixed elevation handling in Card and CollapsibleBox components

### Before:
Screenshot 2024-10-17 at 14 16 36

### After:
Screenshot 2024-10-17 at 14 18 16


packages/components/src/components/CollapsibleBox/CollapsibleBox.tsx
```diff
@@ -174,6 +174,51 @@ export const CollapsibleBox = ({
         setIsOpen(!isOpen);
     };
 
+    const headerContent = (
```
tomasklim, line 177: How about moving this out of this component `CollapsibleBoxHeader`

packages/components/src/components/Card/utils.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+    return paddingMap[$paddingType];
+};
+
+export const mapPaddingTypeToPadding = ({ $paddingType }: MapArgs): number | string => {
+    const paddingMap: Record<PaddingType, number | string> = {
+        none: 0,
+        small: spacingsPx.sm,
+        normal: spacingsPx.lg,
+        large: spacingsPx.xl,
```
tomasklim, line 37: normal -> lg padding
large -> xl padding

How about 
having normal, large and xlarge with proper spacings

seibei-iguchi: Sorry, I don't understand the point:)

tomasklim: we call paddingType `normal` but we use `lg` spacing.

seibei-iguchi: As consumers of the component, we don't care about the right side of the mapping function, that's just internal design system token naming:) For this component, normal means such and such padding. Hope it makes sense:)

enjojoy (approved):
LGTM 🚀 Tested the change, it works. Fixed card elevation looks good.

packages/components/src/components/Card/Card.tsx
```diff
@@ -103,33 +83,38 @@ const CardContainer = styled.div<
     /* when theme changes from light to dark */
     transition: background 0.3s, box-shadow 0.2s;
 
+    ${({ $fillType, $elevation, theme }) =>
+        $fillType === 'none' &&
```
peter-sanderson: [optional] Not sure if its worth it here, but adding a new fillType wont produce type-error here. So programmer does not know, they need to adjust CSS

```ts
const map: Record<FillType, SomeCSSStyledComponentType> = {
    'none': css`<here is the css for none>`, 
    'default': css`<here is the css for default>`,
}
```

a then do:

```ts
${map[$fillType]}  // <--- here you get type-error if you add new FillType into union
```

packages/components/src/components/Card/Card.tsx
```diff
@@ -103,33 +83,38 @@ const CardContainer = styled.div<
...
+        css`
+            background: none;
```
peter-sanderson: Also this is resetting something from above. I would rather see a explicit styling for every `FillType`  separate.

peter-sanderson: So my logical flow when I read the code is straightforward

packages/suite/src/components/backup/BackupSeedCard.tsx
```diff
@@ -105,7 +105,7 @@ export const BackupSeedCard = ({
     };
 
     return (
-        <Container forceElevation={2} $checked={isChecked} data-testid={dataTest}>
```
peter-sanderson, line 108: :clap:

packages/suite/src/views/suite/SwitchDevice/CardWithDevice.tsx
```diff
@@ -64,7 +64,7 @@ export const CardWithDevice = ({
     const { elevation } = useElevation();
 
     return (
-        <Card paddingType="small" forceElevation={0}>
```
peter-sanderson, line 67: :clap: :clap:

peter-sanderson (approved):
Just that one theoretical/philosophical point about type mapping. We can discuss on the call if you want, but its not a big deal. 

Good job removing that hacky `forceElevation` :tada: 

:shipit:

seibei-iguchi:
> Just that one theoretical/philosophical point about type mapping. We can discuss on the call if you want, but its not a big deal. 
> 
> 
> 
> Good job removing that hacky `forceElevation` :tada: 
> 
> 
> 
> :shipit: 

The force elevation wasn't even doing anything:) If you look at the Card component, it wasn't properly passed.

tomasklim:
Adam when you are into this, could you also check it in trade section? (In case it is really the same) https://github.com/trezor/trezor-suite/issues/14852

peter-sanderson (approved):

PR #14872, opened 2024-10-15
fix(suite-native): send amount inputs scrolling

### LINKED ISSUE #14871: Send coins - inputs not visible on focus
### Current behavior
- if you focus amount input on a device with short screen, the defocused amount input might be hidden behind the footer button

https://github.com/user-attachments/assets/6f17fb22-dd58-45db-be65-065c6460a82e

### Expected behavior
If user focuses any of the amount inputs or change its value, the both inputs are always visible

PeKne:
## Description
- Both amount inputs are scrolled to be visible on every focus or change so they always stay visible.

## Related Issue

Resolve #14871 

## Screenshots:

https://github.com/user-attachments/assets/b1953a5b-007b-41a3-94a7-531a1da955a1


matejkriz (approved):
I don't like the double-phased scroll, but I haven't found any better solution now, let's use this...

PR #14877, opened 2024-10-15
chore: rename connect-deeplink to connect-mobile

martykan:
## Description

Rename `@trezor/connect-deeplink` to `@trezor/connect-mobile`


Nodonisko:
How we will call mobile sdk?

martykan:
@mroz22's vision is it will be both in the same package

mroz22:
yes, the idea is to:
1. have packages called based on the env they are supposed to be used in (web, webextension, mobile)
2. have plugin-in based architecture for "core implementation" and one of the implementation could be accessing core over deeplings and the other one could be accessing core as a local module (provided as some plugin)

mroz22 (approved):

PR #14884, opened 2024-10-15
feat(suite-common): add pol and bnb to graph

### LINKED ISSUE #14851: Polygon - make sure POL and BNB works with graph
POL and BNB networks should be counted in portfolio graph (same rules as eth)

vytick:
POL and BNB networks should be counted in portfolio graph. 

- Adding `BEP20` as a contract type for BSC tokens
- fetching POL and BNB history movement events

## Related Issue

Resolve #14851

### DIFF
--- a/suite-common/graph/src/balanceHistoryUtils.ts
+++ b/suite-common/graph/src/balanceHistoryUtils.ts
@@ -134,6 +134,7 @@ const getAccountHistoryMovementItemRipple = ({
     return { main: Array.from(summaryMap.values()).sort((a, b) => a.time - b.time), tokens: {} };
 };
 
+// this can be also used for networks of Ethereum type (like ETH, POL or BNB)
 export const getAccountHistoryMovementItemETH = ({
     transactions,
     from,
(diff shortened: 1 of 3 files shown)


suite-common/graph/src/balanceHistoryUtils.ts
```diff
@@ -230,8 +230,9 @@ export const getAccountHistoryMovementItemETH = ({
         }
 
         tx.tokens.forEach(token => {
-            // skip empty amounts and non-ERC20 tokens
-            if (token.amount === '' || token.standard !== 'ERC20') return;
+            // skip empty amounts and non-ERC20 non-BEP20 tokens
```
Nodonisko, line 234: please add comment that BEP is related to polygon

Nodonisko (approved):

suite-common/graph/src/balanceHistoryUtils.ts
vytick: b09a27fe8f

suite-common/graph/src/balanceHistoryUtils.ts
```diff
@@ -320,6 +322,8 @@ export const getAccountHistoryMovementFromTransactions = ({
         case 'xrp':
             return getAccountHistoryMovementItemRipple({ transactions, from, to });
         case 'eth':
+        case 'pol':
+        case 'bnb':
             return getAccountHistoryMovementItemETH({ transactions, from, to });
```
matejkriz, line 328: Maybe add a comment to the `getAccountHistoryMovementItemETH` method that it's not only for Ethereum, but for all ethereum-type networks?

BTW, @Nodonisko, can we use `getAccountHistoryMovementItemBTC` for all btc-like coins? Does it make sense or it's better to keep it for BTC only?

vytick: updated

vytick: updated

Nodonisko: It's actually unused code right now, I kept it there because I thinked we will download all transactions anyway, but since we are not doing that is much better and faster to use blockbook to get account movements

PR #14899, opened 2024-10-16
feat: implement strong randomInt that works in browser without need for polyfill lib

peter-sanderson:
We need cryptographically unbiased random function.

This function
- Does NOT need any polyfill for webpack. It uses `window.crypto` for the Browser and `crypto` for Node.js
- Uses *Rejection Sampling* technique to generate random integer in given range [without modulo bias](https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/). 

:exclamation: **Disclaimer:** Please review it VERY carefully. I am not a cryptographer and this was quite hard (but very enlightening) for me to do.

There is similar code in Firmware: https://github.com/trezor/trezor-firmware/blob/main/crypto/rand.c#L77

### DIFF
--- a/packages/utils/src/getRandomInt.ts
+++ b/packages/utils/src/getRandomInt.ts
@@ -1,0 +1,58 @@
+import { getRandomValues as cryptoGetRandomValues } from 'crypto';
+
+/**
+ * Before changing anything here, see the Modulo Bias problem!
+ * @see https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/
+ *
+ * @param min Inclusive
+ * @param max Exclusive (to match the crypto.randomInt() function API
+ */
+export const getRandomInt = (min: number, max: number) => {
+    if (!Number.isSafeInteger(min)) {
+        throw new RangeError(
+            `The "min" argument must be a safe integer. Received type ${typeof min} (${min})`,
+        );
+    }
+
+    if (!Number.isSafeInteger(max)) {
+        throw new RangeError(
+            `The "max" argument must be a safe integer. Received type ${typeof max} (${max})`,
+        );
+    }
+
+    if (min >= max) {
+        throw new RangeError(
+            `The value of "max" is out of range. It must be greater than the value of "min" (${min}). Received ${max}`,
+        );
+    }
+
+    const MAX_RANGE_32_BITS = 0xffffffff + 1;
+    const range = max - min;
+
+    if (range > MAX_RANGE_32_BITS) {
+        throw new RangeError(
+            `This function only provide 32 bits of entropy, therefore range cannot be more then 2^32.`,
+        );
+    }
+
+    const getRandomValues =
+        typeof window !== 'undefined'
+            ? (array: Uint32Array) => window.crypto.getRandomValues(array)
+            : (array: Uint32Array) => cryptoGetRandomValues(array);
+
+    const array = new Uint32Array(1); // This provides 32 bits of entropy.
+
+    // It is crucial to avoid modulo bias.
+    // See: https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/
+
+    // We calculate the maximum possible value that can be evenly distributed across the desired range.
+    const maxRange = MAX_RANGE_32_BITS - (MAX_RANGE_32_BITS % range);
+
+    let randomValue: number;
+    do {
+        getRandomValues(array);
+        randomValue = array[0];
+    } while (randomValue >= maxRange);
+
+    return min + (randomValue % range);
+};
--- a/packages/utils/tests/getRandomInt.test.ts
+++ b/packages/utils/tests/getRandomInt.test.ts
@@ -1,0 +1,80 @@
+import { randomInt } from 'crypto';
+
+import { getRandomInt } from '../src/getRandomInt';
+
+describe(getRandomInt.name, () => {
+    it('raises same error as randomInt from crypto when max <= min', () => {
+        const EXPECTED_ERROR = new RangeError(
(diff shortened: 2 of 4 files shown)


packages/utils/src/getRandomInt.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * Before changing anything here, see the Modulo Bias problem!
+ * @see https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/
+ *
+ * @param min Inclusive
+ * @param max Exclusive (to match the crypto.randomInt() function API
```
peter-sanderson, line 8: I had a quite a debate with @szymonlesisz if we should have this exclusive or not. 

My argument is, that I like to have it aligned with  `crypto.randomInt()` API. (Also `Math.random()` does `0-1` where `1` is exclusive.

packages/utils/tests/getRandomInt.test.ts
```diff
@@ -0,0 +1,75 @@
...
+            'The "max" argument must be a safe integer. Received type number (9007199254740992)',
+        );
+
+        expect(() => randomInt(0, UNSAFE_INTEGER)).toThrowError(EXPECTED_ERROR_MAX);
+        expect(() => getRandomInt(0, UNSAFE_INTEGER)).toThrowError(EXPECTED_ERROR_MAX);
+    });
+
+    // This test takes 100+seconds to run. It is very needed for development and debugging,
+    // but due to the time it takes, it is skipped in CI.
+    it.skip('return value in given range and in uniform distribution', () => {
```
mroz22, line 52: skiiiip

Lemonexe: The commentary explains it. I ran it locally, it passes and took 2 min, so it makes sense to skip. Also it's a random test, while the its above are deterministic, this one could fail randomly, and the chance of it failing depends on your tolerance. With 10% tolerance, the chance is probably negligible :laughing:  

Though maybe it's no harm to enable it - after all, github runners run parallel, and unit tests are not by far the slowest job, so slowing it down would not slow down CI as a whole.. But it could add up as people add more tests. FWIW I'm fine with it either way :shrug:

packages/utils/tests/getRandomInt.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        const EXPECTED_ERROR_MIN = new RangeError(
+            'The "min" argument must be a safe integer. Received type number (9007199254740992)',
+        );
+
+        expect(() => randomInt(UNSAFE_INTEGER, UNSAFE_INTEGER + 1)).toThrowError(
+            EXPECTED_ERROR_MIN,
+        );
+        expect(() => getRandomInt(UNSAFE_INTEGER, UNSAFE_INTEGER + 1)).toThrowError(
+            EXPECTED_ERROR_MIN,
+        );
```
Lemonexe, line 35: What about negative values? `(-UNSAFE_INTEGER-1, -UNSAFE_INTEGER)`

packages/utils/tests/getRandomInt.test.ts
Lemonexe: Alternatively: lower the number of samples, if it would still have statistical significance. And developers can increase it locally as they wish..

Lemonexe (approved):
Codewise looks good, but I'm not a cryptographer, so a review from someone more knowledgable will be needed.

packages/utils/tests/getRandomInt.test.ts
peter-sanderson: When I did less iterations, it failed from time to time. I assume it would be super-annoying to have test that fails randomly.

peter-sanderson: I would have to also increase the tolerance. I fear that it would render the test useless.

Lemonexe: Yes, then you would have to increase tolerance. I tried playing around with samples vs. tolerance locally. But even when you have high number of samples, _or_ high tolerance, it still _can_ randomly fail, although with very tiny probability (negligible).
BUt still, it just feels to me methodically incorrect to run an indeterministic test in CI, even when statistics are overwhelmingly on our side. So I'm fine with skipping it, but that's just my opinion :smile:

packages/utils/tests/getRandomInt.test.ts
peter-sanderson: fixed (added test cases)

packages/utils/tests/getRandomInt.test.ts
peter-sanderson: Just sidenote we already have similar test here: `packages/utils/tests/arrayShuffle.test.ts`

packages/utils/src/getRandomInt.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    const MAX_32_BIT_INT = 0xffffffff;
+    const range = max - min;
+
+    if (range > MAX_32_BIT_INT) {
+        throw new RangeError(
+            `This function only provide 32 bits of entropy, therefore range cannot be more then 2^32.`,
+        );
+    }
```
onvej-sl: If you call `getRandomInt(0, 2^32)`, you'll get the exception: `This function only provide 32 bits of entropy, therefore range cannot be more then 2^32.` First, I think this is too restrictive, since we can easily handle this case as well. Second, the message in the exception is not true since the range is `2^32` which is not more than `2^32`.

If you change `MAX_32_BIT_INT` to `2^32` (and perhaps rename the variable), it should resolve both the problems. At the same time, it should not cause any issue elsewhere: It will hold that `0 <= maxRange <= 2^32` and the loop will generate evenly distributed  `randomValue` such that `0 <= randomValue < maxRange`.

PR #14905, opened 2024-10-16
fix(suite-desktop-core): restart app when using autostart on linux

martykan:
## Description

We see some issues with restarting after auto update on Ubuntu with Suite 24.9.2. 
The app will quit, but not re-open itself. 

One problem could be due to the `--no-sandbox` flag, however even after solving that it seems there is an issue. 

Based on a bug from Sentry (#14904), it could be that there is a race condition where the old app is not fully closed when the new instance is started. 
This means that the `second-instance` handler is triggered, which doesn't work since the app is already closing. 

This fix removes the `second-instance` and `before-quit` handlers (only on Linux), allowing the app to quit immediately.

### DIFF
--- a/packages/suite-desktop-core/src/modules/auto-updater.ts
+++ b/packages/suite-desktop-core/src/modules/auto-updater.ts
@@ -214,6 +214,12 @@ export const init: Module = ({ mainWindowProxy, store }) => {
         setImmediate(() => {
             // Removing listeners & closing window (https://github.com/electron-userland/electron-builder/issues/1604)
             app.removeAllListeners('window-all-closed');
+            if (process.platform === 'linux') {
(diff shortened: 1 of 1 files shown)


mroz22:
@karliatto if you could please review and check it on your machine 🙏

karliatto:
If I understand well the issue that this is trying to solve is when the app is updated to new version and it restarts. It was crashing sometimes if autostart was enabled in Linux and this should fix it. Right?

Is there a quick way how to test app update?

martykan:
I don't really know a quick way. 
When I was testing the update process before, I served the `build-electron/` directory via a webserver on localhost and set it as a feed for Electron updater. It should have the necessary YML files. 
Then when you bump the version and build it, it should offer you the update from the previous version.

karliatto (approved):
LGTM.
I fully tested locally with update and restart of the app having the background from autostart running and it was working OK.

PR #14917, opened 2024-10-17
Refactoring of the transaction composition so it allows for multiple sorting strategies

peter-sanderson:


packages/utxo-lib/src/types/compose.ts
```diff
@@ -68,11 +68,34 @@ export interface ComposeChangeAddress {
...
 }
 
-export interface ComposeRequest<
+export type TransactionInputOutputSortingStrategy =
+    // BIP69 sorting
+    | 'bip69'
+
+    // Inputs are randomized, outputs are kept as they were provided in the request,
+    // and change is randomly placed somewhere between outputs
+    // | 'random' // Todo: implement
```
Lemonexe: Nit: reference issue, if it has been created, or create one.

packages/utxo-lib/src/types/compose.ts
```diff
@@ -68,11 +68,34 @@ export interface ComposeChangeAddress {
...
+
+    // It keeps the inputs and outputs as they were provided in the request.
+    // This is useful for RBF transactions where the order of inputs and outputs must be preserved.
+    | 'none';
+
+type SortingStrategyPropsWithCompatibility =
+    | {
+          /** @deprecated use `sortingStrategy=none` instead if you want to keep order of inputs and outputs */
```
Lemonexe, line 85: I don't think the comment is working. I'd expect that when supplying boolean would, IDE would mark it as deprecated.
At least Idea webstorm doesn't mark [the usage](https://github.com/trezor/trezor-suite/blob/d18ba9a879d8a6e6aa1967533a7bda042824729a/packages/connect/src/api/bitcoin/TransactionComposer.ts#L167) as deprecated, although there it's strictly `boolean`.
Maybe it's not possible, as the deprecation can't "propagate" through the chain of functions, then nevermind :shrug: :+1: 

image

packages/utxo-lib/src/compose/transaction.ts
```diff
@@ -5,76 +5,30 @@ import {
...
-
-        return convertOutput(output, { type: 'change', ...request.changeAddress });
-    });
+    const sortingStrategy: TransactionInputOutputSortingStrategy =
+        // eslint-disable-next-line no-nested-ternary
+        request.sortingStrategy === undefined
+            ? request.skipPermutation === true
+                ? 'none'
+                : 'bip69'
+            : request.sortingStrategy;
```
Lemonexe: Nit: what do you think about this instead of nested ternary? 
```suggestion
    const sortingStrategy: TransactionInputOutputSortingStrategy =
        request.sortingStrategy ?? (request.skipPermutation === true ? 'none' : 'bip69');
```

packages/utxo-lib/src/compose/sorting/bip69SortingStrategy.ts
```diff
@@ -0,0 +1,39 @@
...
+export const bip69SortingStrategy: SortingStrategy = ({ result, request, convertedInputs }) => {
+    const defaultPermutation: number[] = [];
+    const convertedOutputs = result.outputs.map((output, index) => {
+        defaultPermutation.push(index);
+        if (request.outputs[index]) {
+            return convertOutput(output, request.outputs[index]);
+        }
+
+        return convertOutput(output, { type: 'change', ...request.changeAddress });
+    });
```
Lemonexe, line 27: This code is duplicated with `noneSortingStrategy`, where the only difference is absence of `defaultPermutation`.
:arrow_right: I would suggest extracting a fn that would do this:
`(result, request) => ({ convertedOutputs, defaultPermutation })` 
and in `bip69` and `none` you could call it & destructure just what is needed.

Lemonexe (commented):
Mostly LGTM
- TS :heart: 
- I carefully reviewed `noneSortingStrategy` and `bip69SortingStrategy` to check it implements the same logic from old `createTransaction` :heavy_check_mark: 
- 4 minor codestyle comments :speech_balloon: 
- `createTransaction` could maybe do with its own code coverage?
  - but not necessary, as we have tests for `composeTx`, and it covers bip69 in most cases, and it also [has one case for none](https://github.com/trezor/trezor-suite/blob/ecb87bccdc8619e5ec604cfecf62c9611d055003/packages/utxo-lib/tests/__fixtures__/compose.ts#L1017), though with old API
  - :point_right: IMHO at least `composeTx` test should be expanded to cover both cases with new API (just duplicate selected two fixtures & remake them to new API?)

packages/utxo-lib/src/compose/transaction.ts
Lemonexe: EDIT: outdated, feel free to ignore and resolve, the new function is good :ok_hand:

packages/utxo-lib/src/types/compose.ts
peter-sanderson: This will be implemented in future PR. I added it here so to demonstrate the motivation why we are doing such refactoring.

packages/utxo-lib/src/compose/transaction.ts
peter-sanderson: :+1: resolving

packages/utxo-lib/src/types/compose.ts
Lemonexe: Yes, I just meant referencing a github issue in the code comment, so the TODO isn't forgotten. But it's not necessary if it's already planned in near future,

packages/utxo-lib/src/types/compose.ts
peter-sanderson: fixed: https://github.com/trezor/trezor-suite/pull/14917/commits/d1d19cd921bb20ca4ec8a592b607aa3533b9c490

packages/utxo-lib/tests/coinselect/coinselect.test.ts
```diff
@@ -1,10 +1,10 @@
 import { coinselect } from '../../src/coinselect';
-import fixtures from '../__fixtures__/coinselect/coinselect-index';
-import fixturesErrors from '../__fixtures__/coinselect/coinselect-errors';
+import { coinselectIndexFixture } from '../__fixtures__/coinselect/coinselect-index';
+import { coinselectFixturesErrors } from '../__fixtures__/coinselect/coinselect-errors';
```
Lemonexe, line 3: :+1:

Lemonexe (approved):

PR #14918, opened 2024-10-17
fix(suite): clear snowflakeBinaryPath when disabling from external feat

### LINKED ISSUE #14731: Tor cannot be connected
Trying with Snowflake and pure. How to view Tor debug logs?

karliatto:
## Description

Issue where `snowflakeBinaryPath` is not cleared if snowflake is disable from experimental features or if experimental features are switch off all at once. So even it was disable in this case it was trying to use it. As described in issue https://github.com/trezor/trezor-suite/issues/14731

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14731

### DIFF
--- a/packages/suite/src/constants/suite/experimental.ts
+++ b/packages/suite/src/constants/suite/experimental.ts
@@ -1,4 +1,6 @@
 import { TranslationKey } from '@suite-common/intl-types';
+import { desktopApi } from '@trezor/suite-desktop-api';
+
 import { EXPERIMENTAL_PASSWORD_MANAGER_KB_URL, TOR_SNOWFLAKE_PROJECT_URL, Url } from '@trezor/urls';
 
 import { Dispatch } from '../../types/suite';
@@ -27,5 +29,16 @@ export const EXPERIMENTAL_FEATURES: Record<ExperimentalFeature, ExperimentalFeat
         title: 'TR_EXPERIMENTAL_TOR_SNOWFLAKE',
         description: 'TR_EXPERIMENTAL_TOR_SNOWFLAKE_DESCRIPTION',
         knowledgeBaseUrl: TOR_SNOWFLAKE_PROJECT_URL,
+        onToggle: async ({ newValue }) => {
+            if (!newValue) {
+                const result = await desktopApi.getTorSettings();
+                if (result.success && result.payload.snowflakeBinaryPath !== '') {
+                    await desktopApi.changeTorSettings({
+                        ...result.payload,
+                        snowflakeBinaryPath: '',
+                    });
+                }
+            }
+        },
     },
 };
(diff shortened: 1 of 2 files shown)


komret (changes requested):
I think it would be better to use the [onToggle](https://github.com/trezor/trezor-suite/blob/d7c8594311840661c5cc75645af34d2064f75325/packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx#L40) property for this.

Maybe also only clear the `snowflakeBinaryPath` in case it was set.

karliatto:
> I think it would be better to use the [onToggle](https://github.com/trezor/trezor-suite/blob/d7c8594311840661c5cc75645af34d2064f75325/packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx#L40) property for this.
> 

It looks to me that that onToggle is currently not doing anything https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/constants/suite/experimental.ts#L16

But sure, I could extend the EXPERIMENTAL_FEATURES.'tor-snowflake' to have it. 

Other thing I want to achieve is to delete `snowflakeBinaryPath` when "Experimental Features" are disable, even if the user did not disable 'tor-snowflake' explicitly, if all the "Experimental Features" are disable it should be cleared.

karliatto:
@komret what about https://github.com/trezor/trezor-suite/pull/14918/commits/6f76794dd3f12b47e9a16399e6047c1e3fc65dd6 ?? I think that works.

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -28,7 +28,7 @@ import { Action, Lock, TorBootstrap, TorStatus } from 'src/types/suite';
 import { getExcludedPrerequisites, getPrerequisiteName } from 'src/utils/suite/prerequisites';
 import { RouterRootState, selectRouter } from './routerReducer';
 import { NetworkSymbol } from '@suite-common/wallet-config';
-import { SuiteThemeVariant } from '@trezor/suite-desktop-api';
+import { SuiteThemeVariant, desktopApi } from '@trezor/suite-desktop-api';
```
komret: forgotten import

komret:
> It looks to me that that onToggle is currently not doing anything https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/constants/suite/experimental.ts#L16

It was originally implemented for Suite automatic update when it was in experimental features, currently it si not used.

> Other thing I want to achieve is to delete `snowflakeBinaryPath` when "Experimental Features" are disable, even if the user did not disable 'tor-snowflake' explicitly, if all the "Experimental Features" are disable it should be cleared.

It should work due to logic [here](https://github.com/trezor/trezor-suite/blob/7ce9c0af34e6d6a7891634cb58a511d540e31ff5/packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx#L90). Although this could be improved to only trigger `onToggle` when switching experimental features off, now it is triggered when they are turned on, too.

komret (approved):

PR #14921, opened 2024-10-17
chore: rimraf only on root package.json

karliatto:
## Description
Dependency `rimraf` is declared in root package.json, and there is not need to have it in other packages.


karliatto:
Just a test.

PR #14931, opened 2024-10-17
feat(suite): Fix & enhance onUpdate FW hash check error handling 

### LINKED ISSUE #5868: Improve GetFirmwareHash error handling
**Describe the bug**

When `GetFirmwareHash` returns an error such as "Unknown message", the dialog below is displayed, but there is no red banner, nothing saying that the user should contact Trezor support, see https://github.com/trezor/trezor-suite/issues/5377.

image

**Expected behavior**

If the reported firmware version supports `GetFirmwareHash` and the device returns an error response to the `GetFirmwareHash` message, then Suite should behave the same way as if an incorrect hash was returned. If this seems too harsh, then Suite can retry the operation twice, but I think that's not even necessary.

Lemonexe:
FW hash check is done automatically after FW update, but it has been broken for some time!
Also we want to handle a generic error, not just hash mismatch, because a counterfeit FW could just fail the hash check on purpose, so that should be handled more strictly than before. 

## Description

:eyes: CR commit by commit recommended

- 7c5eb3d3c9b778bbdd38ade2b1968cab628cb9bb Fixing an inconspicuous bug that suppresses all error handling
  - that took me a while to spot. TS doesn't give any error.
- f501f934fd73422ab8402a37c8b26b029aeb783b When hash check fails to perform, treat it just as severely as hash mismatch.
  - :thinking:  This creates a problem though, because while hash `mismatch` guarantees counterfeit FW, an `other-error` could be transient.
  - But there is no way to exonerate a marked device, except by wiping it, or reseting the suite
  - Can't even retry FW update, the modal is blocked by error
- 3643159a85c713c60a535c363ec3fa22b9c757c6 :arrow_right:  so now, when a device is marked "failed last fw update hash check", you can just start the fw update process again. If it passes (error was transient), it will be exonerated.
  - btw if you truly installed bad binary, you'll ofc also get ghosts on reload because [#14766](https://github.com/trezor/trezor-suite/pull/14766)

I've tested using a ~[debug branch](https://github.com/trezor/trezor-suite/tree/fix/fw-update-hash-check-error-handling-DEBUG)~ (gone), and also by really doing a FW update, but with modified [onCallFirmwareUpdate.ts](https://github.com/trezor/trezor-suite/blob/f501f934fd73422ab8402a37c8b26b029aeb783b/packages/connect/src/core/onCallFirmwareUpdate.ts)

## Related Issue

Resolve #5868

## Screenshots:

**Before**
even with failing FW hash check, the FW update procedure appears successful :scream: 
image

**After first commit**
Now, at the very least, we're at the situation described by issue  #5868 :sweat_smile: 
soft error

**Final state**
Video showing all flows using mocked results [using this branch](https://github.com/trezor/trezor-suite/tree/fix/fw-update-hash-check-error-handling-DEBUG)
[mocked update flows.webm](https://github.com/user-attachments/assets/a79bfc8f-4b6f-4f9f-805c-9cb5093c95f0)


packages/suite/src/views/firmware/FirmwareModal.tsx
```diff
@@ -77,19 +82,13 @@ export const FirmwareModal = ({
...
-            );
-        }
-
         switch (status) {
             case 'error':
+                // Special and hopefully very rare case. This appears when somebody tried to fool user into using a hacked firmware. This check is skipped when installing custom FW.
+                if (device?.id && firmwareHashInvalid.includes(device.id)) {
+                    return <FirmwareUpdateHashCheckError error={error} />;
+                }
+
```
Lemonexe, line 91: Moving the code here is what enables you to reopen FW update modal, so you can try reinstalling it.

When the condition was above the switch, it meant that once a device was marked as failed, reopening the FW update modal would immediately display the error as if it just finished in error state, and never let you reinstall. Even after reloading app or reconnecting device, because the devices marked as failed are persistent.
I think that doesn't make sense, so this lets you try again (and together with `clearInvalidHash`, it can exonerate the device)

suite-common/wallet-core/src/firmware/firmwareActions.ts
```diff
@@ -17,9 +17,19 @@ const setHashInvalid = createAction(
     }),
 );
 
-const setError = createAction(`${FIRMWARE_MODULE_PREFIX}/set-error`, (payload?: string) => ({
```
Lemonexe, line 20: I renamed `setError` to a more descriptive `setFirmwareUpdateError`, I verified it is not used for anything else that FW hash check on FW update

peter-sanderson (approved):

Lemonexe:
Updated codestyle as per CR

Issue #12710, opened 2024-06-03
EPIC: connect 10 - tracker


mroz22:
Lets start with:
- marking places in code that require refactoring but it comes at the cost of introducing breaking changes with a comment 
```
// todo: connect10
```
- at the same time, if breaking change should be introduced in connect 10, lets put a [deprecation warning](https://github.com/trezor/trezor-suite/issues/12676) where it makes sense.

Lets start listing future breaking changes here:

```[tasklist]
### Future breaking changes
- [ ] remove `connectSettings.webusb`
- [ ] remove workarounds in poup.html (cc @martykan, please edit)
- [ ] device.state string to object - https://github.com/trezor/trezor-suite/pull/12887
- [ ] support native bigint. remove bignumber.js etc where possible
- [ ] remove `TrezorConnect.manifest()`
- [ ] Turn `requestLogin` into a normal method by removing `asyncChallenge`
- [ ] remove `skipPermutation` from `composeTransaction` call (replaced by `sortingStrategy`) https://github.com/trezor/trezor-suite/issues/15070
```

Connect 10 API wish list
- strongly typed errors
- exposed validations? could connect somehow tell its consumers whether the call is allowed and whether it will succeed?

szymonlesisz:
related issue: https://github.com/trezor/trezor-suite/issues/4517

related notes: https://www.notion.so/satoshilabs/Version-10-api-breaking-changes-212426f265074e1ab6651f8ee5096655

martykan:
Closing as we are managing Connect 10 issues separately

PR #14932, opened 2024-10-17
feat(connect): try to reuse initialized session

szymonlesisz:
## Description

`DeviceState` contains all the information to call `Initialize` message and potentially reuse already initialized session on the device.

This should fix the issue with app reload and entering the passphrase.

## Follow up
- i didnt change connect storage loadState() cc @mroz22 

## Related Issue

Related to [Epic connect 10 ](https://github.com/trezor/trezor-suite/issues/12710): Device.state as string > DeviceState

### DIFF
--- a/packages/connect/src/core/AbstractMethod.ts
+++ b/packages/connect/src/core/AbstractMethod.ts
@@ -53,6 +53,30 @@ function validateStaticSessionId(input: unknown): StaticSessionId {
         'DeviceState: invalid staticSessionId: ' + input,
     );
 }
+// validate expected state from method parameter.
+// it could be undefined
+function validateDeviceState(input: unknown): DeviceState | undefined {
+    if (typeof input === 'string') {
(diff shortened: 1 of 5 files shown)


packages/connect/src/core/AbstractMethod.ts
```diff
@@ -53,6 +53,30 @@ function validateStaticSessionId(input: unknown): StaticSessionId {
         'DeviceState: invalid staticSessionId: ' + input,
     );
 }
+// validate expected state from method parameter.
+// it could be undefined
+function validateDeviceState(input: unknown): DeviceState | undefined {
+    if (input === 'string') {
```
martykan:
```suggestion
    if (typeof input === 'string') {
```

martykan: That seems to be causing the tests to fail, other than that looks good

martykan (approved):

PR #14947, opened 2024-10-18
feat(connect): uncouple enableFirmwareHashCheck setting from binFilesBaseUrl

Lemonexe:
## Description

Code cleanup after #14766. On second thoughts, I think it is quite obscure that supplying `binFilesBaseUrl` also incidentally turns on automatic firmware hash check, and explicitly supplying the flag is cleaner.

### DIFF
--- a/packages/connect/src/data/connectSettings.ts
+++ b/packages/connect/src/data/connectSettings.ts
@@ -154,5 +154,9 @@ export const parseConnectSettings = (input: Partial<ConnectSettings> = {}) => {
         settings.binFilesBaseUrl = input.binFilesBaseUrl;
     }
 
+    if (typeof input.enableFirmwareHashCheck === 'boolean') {
+        settings.enableFirmwareHashCheck = Boolean(input.enableFirmwareHashCheck);
+    }
+
     return settings;
 };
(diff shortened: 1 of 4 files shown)


packages/connect/src/data/connectSettings.ts
```diff
@@ -154,5 +154,7 @@ export const parseConnectSettings = (input: Partial<ConnectSettings> = {}) => {
         settings.binFilesBaseUrl = input.binFilesBaseUrl;
     }
 
+    settings.enableFirmwareHashCheck = Boolean(input.enableFirmwareHashCheck);
```
mroz22: don't we want to keep the same structure as with other fields?

Lemonexe: No problem. My idea was, that if we have an optional param of `boolean` type specifically ,
we don't need `true | false | undefined`, and can instead default `undefined` to `false`, as they have the same meaning.
It was a suggestion, but for the sake of consistency I will use the old pattern.

:thought_balloon:  though if you agree with the proposed pattern, I can do that in a separate PR for _all_ optional booleans :slightly_smiling_face:

mroz22: arrrgh. if you want, go for it. but is it such a big deal? 😄 we are old and tired people 👵

Lemonexe: Yeah you're right. We all have better things to do than refactoring interfaces :smiley:

mroz22 (approved):

PR #14950, opened 2024-10-18
chore(connect): split types of different Connect implementations

martykan:
## Description

Previously, the settings and methods were the same for all Connect packages (eg. connect, connect-web, connect-webextension, connect-mobile).
This is not ideal, because some methods and settings are only available in some of them. 

The goal is to make it possible for each package to have it's own type - specifically for the settings and for additional methods besides the common ones.

### DIFF
--- a/packages/connect-web/src/impl/core-in-iframe.ts
+++ b/packages/connect-web/src/impl/core-in-iframe.ts
@@ -19,7 +19,12 @@ import {
     CoreEventMessage,
     CallMethodPayload,
 } from '@trezor/connect/src/events';
-import type { ConnectSettings, DeviceIdentity, Manifest } from '@trezor/connect/src/types';
+import type {
+    ConnectSettings,
+    ConnectSettingsWeb,
+    DeviceIdentity,
+    Manifest,
+} from '@trezor/connect/src/types';
 import { ConnectFactoryDependencies, factory } from '@trezor/connect/src/factory';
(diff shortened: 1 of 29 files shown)


packages/connect-web/src/impl/core-in-iframe.ts
```diff
@@ -221,6 +227,7 @@ export class CoreInIframe implements ConnectFactoryDependencies {
 
             // auto init with default settings
             try {
+                // @ts-expect-error manifest could have been changed in the meantime, hard to check this
```
martykan, line 230: previously this worked due to weaker typing

packages/connect/src/types/settings.ts
```diff
@@ -13,27 +13,20 @@ export interface ConnectSettingsPublic {
     manifest?: Manifest;
     connectSrc?: string;
     debug?: boolean;
-    hostLabel?: string;
-    hostIcon?: string;
     popup?: boolean;
     transportReconnect?: boolean;
-    webusb?: boolean; // deprecated
+    /** @deprecated */
+    webusb?: boolean;
```
mroz22: are we ready to remove it from types? (backwards comp stuff)?

mroz22 (approved):
I think it makes sense

packages/connect/src/types/settings.ts
martykan: We can remove it, it looks like it doesn't even work anymore

PR #14957, opened 2024-10-18
fix(suite): routes for tokens

### LINKED ISSUE #13142: EVM Tokens / Change route logic to avoid closing search and loading
Try to use https://v5.reactrouter.com/web/api/Switch

Instead of having token routes defined in the main routes file, add also smaller router to main token component.

This will fix 
- loading between Coins / Collectibles / Hidden 
- closing search
- resetting filter

```
const TokenComponent = () => (

<div>
  <TokensSelectorAndSearch />
  <Switch>
     <Route id="coins" component={Coins} />
     <Route id="nfts"...
     <Route id="hidden"...
)
```

https://github.com/trezor/trezor-suite/assets/33235762/0c008f7d-cb44-438a-bd73-8d025d13659f

### LINKED ISSUE #14985: Rename Coins/Hidden
We need to rename Coins/Hidden to something like Visible/Hidden or Legit/Hidden is the tokens section. Soon NFT section will be added so it should be consistent. 

image

enjojoy:
## Description

- Introduced nested routing in Tokens section. Now the search is not being re-rendered when changing tabs. Tho in the future it will be better to update routing utils functions to support nested routing by design.
- Refactored some legacy styling. 
- Renamed Coins to Tokens

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13142 
Resolve https://github.com/trezor/trezor-suite/issues/14985

## Screenshots:

https://github.com/user-attachments/assets/ccd568c1-1a7e-469c-b82c-3436064864ac


packages/suite-desktop-ui/src/support/Router.tsx
```diff
@@ -43,8 +41,6 @@ const components: { [key: string]: ComponentType<any> } = {
     'wallet-index': Transactions,
     'wallet-receive': Receive,
     'wallet-details': WalletDetails,
-    'wallet-tokens-coins': Coins,
-    'wallet-tokens-hidden': HiddenTokens,
```
tomasklim, line 47: this means that there are no tokens in desktop app at all

packages/suite-web/src/support/Router.tsx
```diff
@@ -29,11 +29,12 @@ const components: Record<PageName, LazyExoticComponent<ComponentType<any>>> = {
...
     ),
+    'wallet-tokens': lazy(() => import(/* webpackChunkName: "wallet" */ 'src/views/wallet/tokens')),
     'wallet-tokens-coins': lazy(
-        () => import(/* webpackChunkName: "wallet" */ 'src/views/wallet/tokens/coins'),
+        () => import(/* webpackChunkName: "wallet" */ 'src/views/wallet/tokens'),
     ),
     'wallet-tokens-hidden': lazy(
-        () => import(/* webpackChunkName: "wallet" */ 'src/views/wallet/tokens/hidden-tokens'),
+        () => import(/* webpackChunkName: "wallet" */ 'src/views/wallet/tokens'),
     ),
```
tomasklim: I think it does not need to be here. you can delete it

packages/suite/src/components/wallet/WalletLayout/AccountTopPanel/AccountNavigation.tsx
```diff
@@ -44,14 +38,13 @@ export const AccountNavigation = () => {
...
         {
-            id: 'wallet-tokens-coins',
+            id: 'wallet-tokens',
             callback: () => {
-                goToWithAnalytics('wallet-tokens-coins', { preserveParams: true });
+                goToWithAnalytics('wallet-tokens', { preserveParams: true });
             },
             title: <Translation id="TR_NAV_TOKENS" />,
             isHidden: !['cardano', 'ethereum', 'solana'].includes(networkType),
-            activeRoutes: ['wallet-tokens-coins', 'wallet-tokens-hidden'],
```
tomasklim, line 53: but now the `Tokens` tab is not active

tomasklim, line 53: btw page header changes when navigating to shown/hidden
Screenshot 2024-10-18 at 22 53 11

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountItemsGroup.tsx
```diff
@@ -55,7 +55,7 @@ export const AccountItemsGroup = ({
 
     const tokensFiatBalance = getTokensFiatBalance(account, localCurrency, rates, tokens);
 
-    const tokensRoutes = ['wallet-tokens-coins', 'wallet-tokens-hidden'];
+    const tokensRoutes = ['wallet-tokens'];
```
tomasklim: I think all three should now be there

packages/suite/src/views/wallet/tokens/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <Switch>
+                <Route path="/accounts/tokens/coins">
+                    <CoinsTable selectedAccount={selectedAccount} searchQuery={searchQuery} />
+                </Route>
+                <Route path="/accounts/tokens/hidden">
+                    <HiddenTokensTable
+                        selectedAccount={selectedAccount}
+                        searchQuery={searchQuery}
+                    />
+                </Route>
```
tomasklim:
```suggestion
                <Route path="/accounts/tokens/hidden">
                    <HiddenTokensTable
                        selectedAccount={selectedAccount}
                        searchQuery={searchQuery}
                    />
                </Route>
                <Route path="*">
                    <CoinsTable selectedAccount={selectedAccount} searchQuery={searchQuery} />
                </Route>
```

You want to always show the `Shown` tokens unless there is hidden so thats why put it second and put there `*`

tomasklim (changes requested):
This will not fix it completely but at least some help

tomasklim:
Please also get rid of `accounts/tokens/coins#...` path. This path should be on `accounts/tokens#...`
By that we will get rid of that nonsense coins in tokens 😀 

So just     `const tokensRoutes = ['wallet-tokens', 'wallet-tokens-hidden'];`

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountItemsGroup.tsx
enjojoy: Wdym by all three? The nft route too? I was thinking to add it in another PR anyway, not to mix things together. Or did you mean something else

packages/suite/src/components/wallet/WalletLayout/AccountTopPanel/AccountNavigation.tsx
enjojoy: fixed

packages/suite-desktop-ui/src/support/Router.tsx
enjojoy: Yes, I've added the Tokens component back

packages/suite/src/components/wallet/WalletLayout/AccountTopPanel/AccountNavigation.tsx
enjojoy: now it is

enjojoy:
@trezor/qa  Please test both web and desktop app

enjojoy:
@adderpositive  Martin, I think that this routing method and NavigationItem styling can be useful in coinmarket, as you also have a similar design in Trading section, please consider implementing

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
```diff
@@ -52,6 +61,18 @@ const Container = styled(NavigationItemBase)<{
                       }
                   }
               `}
+    ${({ $isRounded }) =>
+        $isRounded &&
+        css`
+            border-radius: ${borders.radii.full};
```
seibei-iguchi, line 67: Use 50% instead:)

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
```diff
@@ -52,6 +61,18 @@ const Container = styled(NavigationItemBase)<{
...
+            padding: ${spacingsPx.xs} ${spacingsPx.md};
+        `}
+    ${({ $typographyStyle }) =>
```
seibei-iguchi: I think this can be removed and the Translation element wrapped in Paragraph. And the Count element can also be removed, and replaced with Text/Paragraph with variant="tertiary".

packages/suite/src/views/wallet/tokens/TokensNavigation.tsx
```diff
@@ -1,104 +1,31 @@
...
 import { EventType, analytics } from '@trezor/suite-analytics';
 
 import { useDispatch, useSelector } from 'src/hooks/suite';
-import { NavigationItem } from '../../../../../components/suite/layouts/SuiteLayout/Sidebar/NavigationItem';
+import { NavigationItem } from 'src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem';
 import { getTokens } from 'src/utils/wallet/tokenUtils';
 import { selectIsDebugModeActive } from 'src/reducers/suite/suiteReducer';
 import { SearchAction } from 'src/components/wallet/SearchAction';
 import { openModal } from 'src/actions/suite/modalActions';
+import { Divider } from '@trezor/components';
```
seibei-iguchi: Add these imports to the import above

packages/suite/src/views/wallet/tokens/TokensNavigation.tsx
```diff
@@ -133,25 +60,33 @@ export const TokensLayoutNavigation = ({
     }, [account.symbol, account.index, account.accountType, setSearchQuery]);
 
     return (
-        <Wrapper>
-            <List>
-                <Item
-                    route="wallet-tokens-coins"
-                    title="TR_COINS"
+        <Row alignItems="center" justifyContent="space-between" margin={{ bottom: spacings.md }}>
```
seibei-iguchi, line 63: Why use two Rows? I think they can be combined.

seibei-iguchi (changes requested):

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
enjojoy: Why do we have it as a variable then? :D

enjojoy: Also, 50% in question 😳
image

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
enjojoy: Idk about removing typographyStyle because I still need to have a different gap, otherwise it looks weird with smaller text

packages/suite/src/views/wallet/tokens/TokensNavigation.tsx
enjojoy: Yes one of the Rows could be removed

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
seibei-iguchi: The Paragraph has margin prop, you can set the values conditionally.

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
seibei-iguchi: Ok that's weird, the full variable is 100px, right? What should it look like?:D

enjojoy: It should look like on the video attached to the desc of this PR

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
enjojoy: Done

adderpositive:
Small visual bug - search in the middle.
image

packages/suite/src/views/wallet/tokens/TokensNavigation.tsx
```diff
@@ -0,0 +1,101 @@
...
+export const TokensNavigation = ({
+    selectedAccount,
+    searchQuery,
+    setSearchQuery,
+}: TokensNavigationProps) => {
+    const { account } = selectedAccount;
+
+    const [isExpanded, setExpanded] = useState(false);
+
+    const routeName = useSelector(state => state.router.route?.name);
```
adderpositive: Nit: you can use `selectRouteName` instead of `state => state.router.route?.name`

packages/suite/src/views/wallet/tokens/hidden-tokens/HiddenTokensTable.tsx
```diff
@@ -3,8 +3,8 @@ import { TokenManagementAction, selectCoinDefinitions } from '@suite-common/toke
 
 import { getTokens } from 'src/utils/wallet/tokenUtils';
 import { useSelector } from 'src/hooks/suite';
-import { NoTokens } from '../../common/NoTokens';
-import { TokensTable } from '../../common/TokensTable/TokensTable';
+import { NoTokens } from '../common/NoTokens';
```
adderpositive, line 6: Nit: consider using absolute paths

adderpositive (approved):

adderpositive:
image
Can you please make a clickable whole circle? It is more user-friendly.

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
```diff
@@ -109,10 +124,24 @@ export const NavigationItem = ({
...
             <Icon name={icon} size={iconSize} color={theme.iconSubdued} pointerEvents="none" />
-            <Translation id={nameId} values={values} />
-            {itemsCount && <Count>{itemsCount}</Count>}
+            <Paragraph
+                variant="tertiary"
+                typographyStyle={typographyStyle}
+                margin={{
+                    right: spacing,
+                    left: spacing,
+                }}
```
adderpositive: Please consider using `gap` instead. This `margin` breaks a navigation in the trade section.

Big white spaces:
image

adderpositive:
> @adderpositive Martin, I think that this routing method and NavigationItem styling can be useful in coinmarket, as you also have a similar design in Trading section, please consider implementing

In the trade section, NavigationItem is used with some small overrides for a responsive design. Routing method should be consider. I like it.

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
adderpositive: This is the main problem for me. )

adderpositive (changes requested):

enjojoy:
> Small visual bug - search in the middle. image

Thank you, didn't test it with debug on

packages/suite/src/views/wallet/tokens/TokensNavigation.tsx
enjojoy: No, in the end both rows are needed

enjojoy:
> image Can you please make a clickable whole circle? It is more user-friendly.

yes, good point

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
enjojoy: Yes, understandable, will bring the gap back

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
seibei-iguchi: Ok, I get how it works now, though it's slightly hacky. Please use the full variable then:)

enjojoy:
> @adderpositive Martin, I think that this routing method and NavigationItem styling can be useful in coinmarket, as you also have a similar design in Trading section, please consider implementing

@adderpositive primarily talking of routing, it's more smooth without unnecessary page reloading

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
```diff
@@ -109,10 +124,18 @@ export const NavigationItem = ({
             className={className}
             tabIndex={0}
             $elevation={elevation}
+            $isRounded={isRounded}
+            $typographyStyle={typographyStyle}
         >
             <Icon name={icon} size={iconSize} color={theme.iconSubdued} pointerEvents="none" />
-            <Translation id={nameId} values={values} />
-            {itemsCount && <Count>{itemsCount}</Count>}
+            <Paragraph variant="tertiary" typographyStyle={typographyStyle}>
```
seibei-iguchi: This wasn't grey before, right? I think the tertiary only applies to the count.

packages/suite/src/components/wallet/SearchAction.tsx
```diff
@@ -102,13 +109,9 @@ export const SearchAction = ({
...
                 innerAddon={
-                    <Icon
-                        name="search"
-                        cursorPointer
-                        size={16}
-                        color={theme.iconDefault}
-                        onClick={!isExpanded ? openAndSelect : undefined}
-                    />
+                    <SearchIconContainer onClick={!isExpanded ? openAndSelect : undefined}>
+                        <Icon name="search" cursorPointer size={16} color={theme.iconDefault} />
```
seibei-iguchi: Use size="medium"

seibei-iguchi: Could you please add variant="default" to the Icon component with this color so you can use variant instead? See the Text component, it works the same.

seibei-iguchi (changes requested):

packages/suite/src/components/wallet/SearchAction.tsx
```diff
@@ -42,6 +37,18 @@ const StyledInput = styled(Input)<{ $isExpanded: boolean }>`
     }
 `;
 
+const SearchIconContainer = styled.div`
+    cursor: pointer;
+    height: 38px;
```
seibei-iguchi: Use the input width const

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
enjojoy: True, good catch

packages/suite/src/components/wallet/SearchAction.tsx
```diff
@@ -102,13 +109,9 @@ export const SearchAction = ({
...
                 innerRef={inputRef}
                 innerAddon={
-                    <Icon
-                        name="search"
-                        cursorPointer
-                        size={16}
-                        color={theme.iconDefault}
-                        onClick={!isExpanded ? openAndSelect : undefined}
-                    />
+                    <SearchIconContainer onClick={!isExpanded ? openAndSelect : undefined}>
```
seibei-iguchi, line 114: Perhaps you don't need this at all if you add the onClick handler on the StyledInput itself.

seibei-iguchi (changes requested):

packages/suite/src/components/wallet/SearchAction.tsx
enjojoy: Not really because then it doesn't click on the icon :D

packages/suite/src/views/wallet/tokens/TokensNavigation.tsx
enjojoy: I combined the imports where it was possible, did you mean something else?

adderpositive (approved):

seibei-iguchi (approved):

suite-common/suite-types/src/route.ts
```diff
@@ -6,8 +6,9 @@ type RouteKeys =
     | 'params'
     | 'exact'
     | 'isForegroundApp'
+    | 'isNestedRoute'
```
tomasklim, line 9: is this used somehow? or does it have just informational meaning?

packages/suite/src/views/wallet/tokens/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    return (
+        <WalletLayout title="TR_TOKENS" account={selectedAccount}>
+            <TokensNavigation
+                selectedAccount={selectedAccount}
+                searchQuery={searchQuery}
+                setSearchQuery={setSearchQuery}
+            />
+            <Switch>
+                <Route path="/accounts/tokens/hidden">
```
tomasklim: use `resolveStaticPath` otherwise it will not work

packages/suite/src/views/wallet/tokens/index.tsx
```diff
@@ -1,16 +1,22 @@
...
 import { HiddenTokensTable } from './hidden-tokens/HiddenTokensTable';
+import { goto } from 'src/actions/suite/routerActions';
 
 export const Tokens = () => {
     const [searchQuery, setSearchQuery] = useState('');
 
     const { selectedAccount } = useSelector(state => state.wallet);
+    const dispatch = useDispatch();
+
+    if (!selectedAccount.network?.features.includes('tokens')) {
```
tomasklim: as the `network` can be undefined (you used `?`), it redirects user even when he refreshes in ETH tokens tab

You should use useEffect and check if account is properly loaded and then do decision 
~Another option is to get symbol from route hash and get features for this network and do it based on that. But I dont know if we do that and use it~

tomasklim (changes requested):

suite-common/suite-types/src/route.ts
enjojoy: It's necessary for the type where PageName is extracted from Route type. 

```js
export type PageName = Exclude<Route, { isForegroundApp: true } | { isNestedRoute: true }>['name'];
```

Nested Route is excluded because it doesn't have it's own page. Otherwise I'd have to assign some component to the route, and it's exactly what we want to avoid with nested components, because they get a component assigned inside it's parent route. 

For more context you can also check Router.tsx: 

If i remove that exclusion of isNestedRoute, typescript complains

image

tomasklim: ok makes sense, thanks

packages/suite/src/views/wallet/tokens/index.tsx
enjojoy: resolveStaticPath seems to be used only for static files. I've added an asset prefix to the path and it works on the staging web.

enjojoy: Okay, I added the prefix and now it works on the staging web

packages/suite/src/views/wallet/tokens/index.tsx
enjojoy: Interesting, I couldn't replicate the error myself but I added the suggested solution, it will be safer like that anyway https://github.com/trezor/trezor-suite/pull/14957/commits/97011477a5d86c0081f286b09ed1f67c68654ed8

tomasklim (approved):

PR #14959, opened 2024-10-18
iframeless suite - use connect core as a module in suite-web

karliatto:
## Description

This PR makes changes in order to allow Trezor Suite to use connect as a module instead of using iframe.

* [feat(connect): core in module](https://github.com/trezor/trezor-suite/pull/14959/commits/76470c2db765556687941b05ffc8ea33ed4312a3) - new core-in-module implementation of Trezor Connect. - It required changes in DataManager so json files like `releases.json` are included in the conenct bundle instead of fetched. It makes the bundle a bit bigger but it requires less fetches.
* [feat(connect-iframe): build core in module for suite](https://github.com/trezor/trezor-suite/pull/14959/commits/f3053de37660b9701cd909f00c91127bb8ebbd68) - refactoring webpack in iframe so it allows to be used for iframe and core-in-module as well as core-in-popup. Eventually iframe could be ditched.
* [feat(connect-web): core in module for web](https://github.com/trezor/trezor-suite/pull/14959/commits/1678edfe47f35f438f2ba07053f5eb02eadc48f6) - implementation of core-in-module for connect-web - it uses `TrezorConnectDynamic`  and use the web code like WebUSB here so it in the environment where it is used.
* [feat(suite-common): integrate connect as module](https://github.com/trezor/trezor-suite/pull/14959/commits/8e638499a60935192a031c5dbb0243dd86cd2c99) - changes in suite-common so it fetches properly the new connect data using `resolveConnectPath`.
* [feat(suite-build): use core in module for suite](https://github.com/trezor/trezor-suite/pull/14959/commits/d14610c837fc5472b1ad7b591449ec9beddf0c11) - suite-builds now are able to build the whole connect. This could be a middle step to the final one where this webpack will build the whole connect instead of delegating it to the one in iframe.
* [feat(suite): update connectSrc in suite](https://github.com/trezor/trezor-suite/pull/14959/commits/3fa3ff612bfc6cc086b12badd716676539765bf9) - changing to new connect path in suite.
* [test(suite-web): update test for suite with core in module](https://github.com/trezor/trezor-suite/pull/14959/commits/d412eda4b7ea8f9f9ae4ab2d7efc1cfe36a70b90) - updating tests so they work with new core-in-module in suite.
* [chore(connect-popup): update core-module build process](https://github.com/trezor/trezor-suite/pull/14959/commits/a7c624cc7505d18fdf36c3e2be7d46c8d47350ce) - update popup to use core module from the new updated webpacks in iframe instead of copying.
* [ci(github): add asset path to popup build](https://github.com/trezor/trezor-suite/pull/14959/commits/bc817b9e4d25ac9de7e9925e128329ead4440320) - update release connect action to use the assets path.
* [ci(github): remove unnecessary build steps in suite-web](https://github.com/trezor/trezor-suite/pull/14959/commits/db158fe2db4fa15468de710cb71c3d14dc67c354)
* [feat(utils): cloneCyclicObject new util](https://github.com/trezor/trezor-suite/pull/14959/commits/0d31f2614e10864684993ee0445bb8123f15c3c7) - new utility that allows to clone object that contain cyclic objects inside

Now you should be able to run `yarn suite:dev` and have all the suite web built including the connect part.

### DIFF
--- a/.github/actions/release-connect/action.yml
+++ b/.github/actions/release-connect/action.yml
@@ -82,6 +82,7 @@ runs:
       shell: bash
       env:
         NODE_ENV: ${{ inputs.nodeEnv }}
+        ASSET_PREFIX: /${{ inputs.serverPath }}
       run: |
         yarn workspace @trezor/connect-popup build
 
--- a/packages/connect-iframe/package.json
+++ b/packages/connect-iframe/package.json
@@ -3,9 +3,10 @@
     "version": "9.0.0",
     "private": true,
     "scripts": {
-        "build:iframe": "TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/prod.webpack.config.ts --stats-children",
+        "build:iframe": "TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/iframe.webpack.config.ts --stats-children",
+        "build:sessions": "TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/sessions.webpack.config.ts --stats-children",
         "build:core-module": "TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/core.webpack.config.ts --stats-children",
-        "build": "yarn rimraf build && yarn build:iframe && yarn build:core-module",
+        "build": "yarn rimraf build && yarn build:iframe && yarn build:core-module && yarn build:sessions",
         "___NOTE__": "iframe build is one of the prerequisites of suite-web. build:lib script provides it together with other libraries",
         "build:lib": "yarn build",
         "type-check": "yarn g:tsc --build tsconfig.json",
--- a/packages/connect-iframe/webpack/core.webpack.config.ts
+++ b/packages/connect-iframe/webpack/core.webpack.config.ts
@@ -3,8 +3,16 @@ import webpack from 'webpack';
 import merge from 'webpack-merge';
 
 import baseConfig from './base.webpack.config';
+import { getDistPathForProject } from './utils';
 
-const DIST = path.resolve(__dirname, '../build');
+const project = process.env.PROJECT || 'iframe';
+
+if (project !== 'iframe' && project !== 'suite-web' && project !== 'popup') {
+    throw new Error(`Unsupported project: ${project}`);
+}
+const DIST = getDistPathForProject(project);
+
+const CORE_PUBLIC_PATH = process.env.ASSET_PREFIX ? `${process.env.ASSET_PREFIX}/` : '/';
 
 export const config: webpack.Configuration = {
     target: 'web',
@@ -15,7 +23,7 @@ export const config: webpack.Configuration = {
     output: {
         filename: 'js/[name].js',
         path: DIST,
-        publicPath: './',
+        publicPath: CORE_PUBLIC_PATH,
         library: {
             type: 'module',
         },
@@ -23,7 +31,6 @@ export const config: webpack.Configuration = {
     experiments: {
         outputModule: true,
     },
-    // todo: nx implicit dependencies
 };
 
 export default merge([config, baseConfig]);
--- a/packages/connect-iframe/webpack/sessions.webpack.config.ts
+++ b/packages/connect-iframe/webpack/sessions.webpack.config.ts
@@ -1,0 +1,29 @@
+import path from 'path';
+import webpack from 'webpack';
+import merge from 'webpack-merge';
+
+import { config as baseConfig } from './base.webpack.config';
+
+const basePath = path.join(__dirname, '..');
+
+const DIST = path.join(basePath, 'build');
+
+const config: webpack.Configuration = {
+    // common instructions that are able to build correctly imports from @trezor/connect (reusing this in popup)
+    entry: {
+        ['sessions-background-sharedworker']: {
+            filename: 'workers/[name].js',
+            import: path.resolve(
+                __dirname,
+                '../../transport/src/sessions/background-sharedworker.ts',
+            ),
+        },
+    },
+    output: {
+        filename: 'js/[name].[contenthash].js',
+        path: DIST,
+        publicPath: './',
+    },
+};
+
+export default merge([config, baseConfig]);
--- a/packages/connect-iframe/webpack/utils.ts
+++ b/packages/connect-iframe/webpack/utils.ts
@@ -1,0 +1,17 @@
+import path from 'path';
+
+type Project = 'iframe' | 'popup' | 'suite-web';
+
+export const getDistPathForProject = (project: Project = 'iframe') => {
+    const basePath = path.join(__dirname, '..', '..');
+    switch (project) {
+        case 'iframe':
+            return path.join(basePath, 'connect-iframe', 'build');
+        case 'suite-web':
+            return path.join(basePath, 'suite-web', 'build');
+        case 'popup':
+            return path.join(basePath, 'connect-popup', 'build');
+        default:
+            throw new Error('Missing project.');
+    }
+};
--- a/packages/connect-web/src/module/index.ts
+++ b/packages/connect-web/src/module/index.ts
@@ -1,0 +1,61 @@
+import { factory } from '@trezor/connect/src/factory';
+import { config } from '@trezor/connect/src/data/config';
+import { TrezorConnectDynamic } from '@trezor/connect/src/impl/dynamic';
+import { CoreInModule } from '@trezor/connect/src/impl/core-in-module';
+import type { ConnectSettingsPublic } from '@trezor/connect/src/types';
+import type { ConnectFactoryDependencies } from '@trezor/connect/src/factory';
+import { CoreRequestMessage, ERRORS, TRANSPORT } from '@trezor/connect/src/exports';
+
+interface ConnectWebDynamicImplementation
+    extends ConnectFactoryDependencies<ConnectSettingsPublic> {
+    handleCoreMessage: (message: CoreRequestMessage) => void;
+}
+
+const impl = new TrezorConnectDynamic<
+    'core-in-module',
+    ConnectSettingsPublic,
+    ConnectWebDynamicImplementation
+>({
+    implementations: [
+        {
+            type: 'core-in-module',
+            impl: new CoreInModule(),
+        },
+    ],
+    getInitTarget: () => 'core-in-module',
+    handleErrorFallback: () => new Promise(resolve => resolve(false)),
+});
+
+const disableWebUSB = () => {
+    if (!impl.lastSettings) {
+        throw ERRORS.TypedError('Init_NotInitialized');
+    }
+
+    impl.getTarget().handleCoreMessage({ type: TRANSPORT.DISABLE_WEBUSB });
+};
+
+const requestWebUSBDevice = async () => {
+    await window.navigator.usb.requestDevice({ filters: config.webusb });
+
+    impl.getTarget().handleCoreMessage({ type: TRANSPORT.REQUEST_DEVICE });
+};
+
+const TrezorConnect = factory(
+    {
+        eventEmitter: impl.eventEmitter,
+        init: impl.init.bind(impl),
+        call: impl.call.bind(impl),
+        manifest: impl.manifest.bind(impl),
+        requestLogin: impl.requestLogin.bind(impl),
+        uiResponse: impl.uiResponse.bind(impl),
+        cancel: impl.cancel.bind(impl),
+        dispose: impl.dispose.bind(impl),
+    },
+    {
+        disableWebUSB: disableWebUSB.bind(impl),
+        requestWebUSBDevice: requestWebUSBDevice.bind(impl),
+    },
+);
+
+export default TrezorConnect;
+export * from '@trezor/connect/src/exports';
--- a/packages/connect/src/data/DataManager.ts
+++ b/packages/connect/src/data/DataManager.ts
@@ -1,71 +1,41 @@
 // origin: https://github.com/trezor/connect/blob/develop/src/js/data/DataManager.js
 
-import { httpRequest } from '../utils/assets';
+import coins from '@trezor/connect-common/files/coins.json';
+import coinsEth from '@trezor/connect-common/files/coins-eth.json';
+import bridge from '@trezor/connect-common/files/bridge/releases.json';
+import messages from '@trezor/protobuf/messages.json';
+
 import { parseCoinsJson } from './coinInfo';
 import { parseFirmware } from './firmwareInfo';
 import { parseBridgeJSON } from './transportInfo';
 import { ConnectSettings, DeviceModelInternal } from '../types';
+import { firmwareAssets } from '../utils/assetUtils'; // Adjust the path as necessary
 
 type AssetCollection = { [key: string]: Record<string, any> };
-
-const assets = [
-    {
-        name: 'coins',
-        url: './data/coins.json',
-    },
-    {
-        name: 'coinsEth',
-        url: './data/coins-eth.json',
-    },
-    {
-        name: 'bridge',
-        url: './data/bridge/releases.json',
-    },
-    {
-        name: 'firmware-t1b1',
-        url: './data/firmware/t1b1/releases.json',
-    },
-    {
-        name: 'firmware-t2t1',
-        url: './data/firmware/t2t1/releases.json',
-    },
-    {
-        name: 'firmware-t2b1',
-        url: './data/firmware/t2b1/releases.json',
-    },
-    {
-        name: 'firmware-t3b1',
-        url: './data/firmware/t3b1/releases.json',
-    },
-    {
-        name: 'firmware-t3t1',
-        url: './data/firmware/t3t1/releases.json',
-    },
-    {
-        name: 'firmware-t3tw1',
-        url: './data/firmware/t3w1/releases.json',
-    },
-];
 
 export class DataManager {
     static assets: AssetCollection = {};
 
     private static settings: ConnectSettings;
-    private static messages: Record<string, any>;
+    private static messages: Record<string, any> = messages;
 
-    static async load(settings: ConnectSettings, withAssets = true) {
-        const ts = settings.env === 'web' ? `?r=${settings.timestamp}` : '';
+    static load(settings: ConnectSettings, withAssets = true) {
         this.settings = settings;
 
         if (!withAssets) return;
 
-        const assetPromises = assets.map(async asset => {
-            const json = await httpRequest(`${asset.url}${ts}`, 'json');
-            this.assets[asset.name] = json;
-        });
-        await Promise.all(assetPromises);
-
-        this.messages = await httpRequest('./data/messages/messages.json', 'json');
+        const assetsMap = {
+            coins,
+            coinsEth,
+            bridge,
+            ...Object.fromEntries(
+                Object.entries(firmwareAssets).map(([key, value]) => [
+                    `firmware-${key.toLowerCase()}`,
+                    value,
+                ]),
+            ),
+        };
+        Object.assign(this.assets, assetsMap);
 
         // parse bridge JSON
         parseBridgeJSON(this.assets.bridge);
--- a/packages/connect/src/impl/core-in-module.ts
+++ b/packages/connect/src/impl/core-in-module.ts
@@ -1,6 +1,6 @@
 import EventEmitter from 'events';
 
-import { DeferredManager, createDeferredManager, cloneObject } from '@trezor/utils';
+import { DeferredManager, createDeferredManager, cloneObjectCyclic } from '@trezor/utils';
 
 import * as ERRORS from '../constants/errors';
 import {
@@ -207,7 +207,7 @@ export class CoreInModule implements ConnectFactoryDependencies<ConnectSettingsP
     public async call(params: CallMethodPayload) {
         try {
             const { promiseId, promise } = this._messagePromises.create();
-            const payload = cloneObject(params);
+            const payload = cloneObjectCyclic<any>(params);
             this.handleCoreMessage({
                 type: IFRAME.CALL,
                 id: promiseId,
--- a/packages/connect/src/impl/core-in-module.ts
+++ b/packages/connect/src/impl/core-in-module.ts
@@ -1,0 +1,286 @@
+import EventEmitter from 'events';
+
+import { DeferredManager, createDeferredManager, cloneObject } from '@trezor/utils';
+
+import * as ERRORS from '../constants/errors';
+import {
+    POPUP,
+    IFRAME,
+    UI,
+    UI_EVENT,
+    DEVICE_EVENT,
+    RESPONSE_EVENT,
+    TRANSPORT_EVENT,
+    BLOCKCHAIN_EVENT,
+    createErrorMessage,
+    UiResponseEvent,
+    CoreEventMessage,
+    CallMethodPayload,
+    CORE_EVENT,
+    CoreRequestMessage,
+} from '../events';
+import type { ConnectSettings, ConnectSettingsPublic, DeviceIdentity, Manifest } from '../types';
+import { ConnectFactoryDependencies, factory } from '../factory';
+import { Log, initLog } from '../utils/debug';
+import { parseConnectSettings } from '../data/connectSettings';
+
+export class CoreInModule implements ConnectFactoryDependencies<ConnectSettingsPublic> {
+    public eventEmitter = new EventEmitter();
+    public _settings: ConnectSettings;
+
+    private _coreManager?: any;
+    private _log: Log;
+    private _messagePromises: DeferredManager<{
+        id: number;
+        success: boolean;
+        payload: any;
+        device?: DeviceIdentity;
+    }>;
+
+    private readonly boundOnCoreEvent = this.onCoreEvent.bind(this);
+
+    public constructor() {
+        this._settings = parseConnectSettings();
+        this._log = initLog('@trezor/connect-web');
+        this._messagePromises = createDeferredManager({ initialId: 1 });
+    }
+
+    private async initCoreManager() {
+        const { connectSrc } = this._settings;
+        const { initCoreState, initTransport } = await import(
+            /* webpackIgnore: true */ `${connectSrc}js/core.js`
+        ).catch(_err => {
+            this._log.error('_err', _err);
+        });
+
+        if (!initCoreState) return;
+
+        if (initTransport) {
+            this._log.debug('initiating transport with settings: ', this._settings);
+            await initTransport(this._settings);
+        }
+
+        this._coreManager = initCoreState();
+
+        return this._coreManager;
+    }
+
+    public manifest(data: Manifest) {
+        this._settings = parseConnectSettings({
+            ...this._settings,
+            manifest: data,
+        });
+    }
+
+    public dispose() {
+        this.eventEmitter.removeAllListeners();
+        this._settings = parseConnectSettings();
+        if (this._coreManager) {
+            this._coreManager.dispose();
+        }
+
+        return Promise.resolve(undefined);
+    }
+
+    public cancel(error?: string) {
+        if (this._coreManager) {
+            const core = this._coreManager.get();
+            if (!core) {
+                throw ERRORS.TypedError('Runtime', 'postMessage: _core not found');
+            }
+
+            this.handleCoreMessage({
+                type: POPUP.CLOSED,
+                payload: error ? { error } : null,
+            });
+        }
+    }
+
+    // handle messages to core
+    public handleCoreMessage(message: CoreRequestMessage) {
+        const core = this._coreManager.get();
+        if (!core) {
+            throw ERRORS.TypedError('Runtime', 'postMessage: _core not found');
+        }
+        core.handleMessage(message);
+    }
+
+    // handle message received from Core
+    private onCoreEvent(message: CoreEventMessage) {
+        const { event, type, payload } = message;
+
+        if (type === UI.REQUEST_UI_WINDOW) {
+            this._coreManager.get()?.handleMessage({ type: POPUP.HANDSHAKE });
+
+            return;
+        }
+
+        if (type === POPUP.CANCEL_POPUP_REQUEST) return;
+
+        switch (event) {
+            case RESPONSE_EVENT: {
+                const { id = 0, success, device } = message;
+                const resolved = this._messagePromises.resolve(id, {
+                    id,
+                    success,
+                    payload,
+                    device,
+                });
+                if (!resolved) this._log.warn(`Unknown message id ${id}`);
+                break;
+            }
+            case DEVICE_EVENT:
+                // pass DEVICE event up to html
+                this.eventEmitter.emit(event, message);
+                this.eventEmitter.emit(type, payload); // DEVICE_EVENT also emit single events (connect/disconnect...)
+                break;
+
+            case TRANSPORT_EVENT:
+                this.eventEmitter.emit(event, message);
+                this.eventEmitter.emit(type, payload);
+                break;
+
+            case BLOCKCHAIN_EVENT:
+                this.eventEmitter.emit(event, message);
+                this.eventEmitter.emit(type, payload);
+                break;
+
+            case UI_EVENT:
+                // pass UI event up
+                this.eventEmitter.emit(event, message);
+                this.eventEmitter.emit(type, payload);
+                break;
+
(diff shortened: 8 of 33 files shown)


mroz22:
nice, I can run 
`yarn suite:dev` without building iframe, connect loads, there is no iframe injected
image

but:
1] I am getting some error, might be unrelated, rebase might be needed. Or maybe the reason is missing static files? If yes, it probably should be a sidequest to handle this error case better
2] Codewise it feels like we should simplify it a bit, there is lot of code duplication with https://github.com/trezor/trezor-suite/blob/feat/use-core-in-suite-web/packages/connect/src/index.ts#L0-L1. My original idea was to defined connect/src/impl/core-in-module, and then use it A] in connect, B] in connect-web/src/impl/core-in-module that would define needed overrides for web. is that possible?

karliatto:
> nice, I can run
> `yarn suite:dev` without building iframe, connect loads, there is no iframe injected

Right, if you build `iframe` the whole discovery is it going to work, but not because it is using `iframe` but becuase currently I made some changes to `iframe` build so it works this way. So my goal now it go change it so it can be built separately but without much code duplication.

> My original idea was to defined connect/src/impl/core-in-module

Here https://github.com/trezor/trezor-suite/pull/14959/commits/88cd2c92b7e1d81c48d4805464be4a10762a73c1 I made the changes to move `core-in-module` to connect/src/impl`. But as you said probably big part of it could be reused from `connect/src/index.ts`.

packages/connect-iframe/package.json
```diff
@@ -4,7 +4,8 @@
     "private": true,
     "scripts": {
         "lint:js": "yarn g:eslint '**/*.{ts,tsx,js}'",
-        "build:iframe": "TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/prod.webpack.config.ts --stats-children",
+        "build:iframe": "TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/iframe.webpack.config.ts --stats-children",
+        "build:sessions": "TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/sessions.webpack.config.ts --stats-children",
```
mroz22, line 7: maybe `build:workers`? in theory, coud this webpack config build any worker?

packages/connect-iframe/webpack/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+type Project = 'iframe' | 'suite-web';
+
+export const getDistPathForProject = (project: Project = 'iframe') => {
+    const basePath = path.join(__dirname, '..', '..');
+    switch (project) {
+        case 'iframe':
+            return path.join(basePath, 'connect-iframe', 'build');
+        case 'suite-web':
+            return path.join(basePath, 'suite-web', 'build', 'static', 'connect');
```
mroz22: so this is to keep the same connect-related folder structure for web, right? the only change is that iframe is going away but the code that was previously in the iframe is still expecting some statics at the same paths.

packages/connect-web/src/module/index.ts
mroz22, line 1: this looks nice, core-in-module for web, reusing coreInModule from node only with some overrides 👍

packages/connect/src/data/DataManager.ts
mroz22, line 1: I am not sure I understand what is going on here. if iframe then load locally? (I didn't check internals of `httpRequest` yet.  anyway, this is something I will be thinking how to avoid bringing this kind of decision making logic into DataManager. At the first glance it doesn't feel right.

packages/connect/src/impl/core-in-module.ts
```diff
@@ -0,0 +1,298 @@
...
+        this._log = initLog('@trezor/connect-web');
+        this._messagePromises = createDeferredManager({ initialId: 1 });
+    }
+
+    private async initCoreManager() {
+        const connectSrc = this._settings.connectSrc;
+        const { initCoreState, initTransport } = await import(
+            /* webpackIgnore: true */ `${connectSrc}js/core.js`
+        ).catch(_err => {
+            this._log.error('_err', _err);
```
mroz22, line 53: what is the reason for dynamic load here?

mroz22 (commented):
I gave it a quick walkthrough and noted couple of questions.

packages/connect/src/impl/core-in-module.ts
karliatto: Making it flexible, so in the context of `suite-web` connectSrc is going to be different as if we use it somewhere else. 
How would you do it?

packages/connect/src/data/DataManager.ts
karliatto: Yes, that feel wrong to me as well, there is other way to do it like https://github.com/trezor/trezor-suite/pull/14959/commits/7696f94f1648aacf9031f4a9d5cacff4f02d1967#diff-54a3dc9015ab9933751016bc35e7987f834b59b24a8d0cc05ef51d1ddabdd1c0 and it works same.

packages/connect-iframe/webpack/utils.ts
karliatto: Yes, this is for using the same webpack configs for both projects.
So it is built directly from `suite-build` instead of copying the build from `connect-iframe` as was done before.

packages/connect-iframe/package.json
karliatto: Do you mean to call it "build:workers" instead of "build:sessions" ?

packages/connect/src/data/DataManager.ts
```diff
@@ -54,18 +26,35 @@ export class DataManager {
...
 
-        const assetPromises = assets.map(async asset => {
-            const json = await httpRequest(`${asset.url}${ts}`, 'json');
-            this.assets[asset.name] = json;
-        });
-        await Promise.all(assetPromises);
+        const coins = require('@trezor/connect-common/files/coins.json');
+        const coinsEth = require('@trezor/connect-common/files/coins-eth.json');
+        const bridge = require('@trezor/connect-common/files/bridge/releases.json');
+        const t1b1 = require('@trezor/connect-common/files/firmware/t1b1/releases.json');
```
mroz22: I am still seeing a fetch for releases.json, its probably because of some firmware security check. something to discuss
image

image

packages/connect/src/data/DataManager.ts
```diff
@@ -1,51 +1,23 @@
...
+require('@trezor/connect-common/files/coins.json');
+require('@trezor/connect-common/files/coins-eth.json');
+require('@trezor/connect-common/files/bridge/releases.json');
+require('@trezor/connect-common/files/firmware/t1b1/releases.json');
+require('@trezor/connect-common/files/firmware/t2t1/releases.json');
+require('@trezor/connect-common/files/firmware/t2b1/releases.json');
+require('@trezor/connect-common/files/firmware/t3b1/releases.json');
+require('@trezor/connect-common/files/firmware/t3t1/releases.json');
+require('@trezor/connect-common/files/firmware/t3w1/releases.json');
+require('@trezor/protobuf/messages.json');
```
mroz22: this block is weird. could you please expand the comment? I am not sure I follow

packages/connect-web/src/module/index.ts
mroz22, line 1: btw why did you chose to implement connect-web/src/module? My first thought was that it could be connect-web/impl/core-in-module which would basically only import the connect/src/core-in-module and only add or override some web specificities. It wouldn't work?

packages/suite-web/e2e/cypress.config.ts
```diff
@@ -69,7 +69,7 @@ export default defineConfig({
         video: true,
         trashAssetsBeforeRuns: false,
         chromeWebSecurity: false,
-        experimentalFetchPolyfill: true,
+        // experimentalFetchPolyfill: true, // This makes Cardano webassembly loading fail.
```
mroz22: all the changes in this commit were required because of this going away?

mroz22 (commented):
I can confirm that suite-web now works on my machine without any iframe. Leaving the first batch of comments and questions. I still haven't read it fully though

packages/connect-iframe/package.json
```diff
@@ -3,7 +3,9 @@
     "version": "9.0.0",
     "private": true,
     "scripts": {
-        "build:iframe": "TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/prod.webpack.config.ts --stats-children",
+        "lint:js": "yarn g:eslint '**/*.{ts,tsx,js}'",
```
mroz22: is this line relevant? maybe it went away as the result of refactorings by @peter-sanderson ?

packages/connect-iframe/webpack/core.webpack.config.ts
```diff
@@ -3,8 +3,16 @@ import webpack from 'webpack';
...
 
-const DIST = path.resolve(__dirname, '../build');
+const project = process.env.PROJECT || 'iframe';
+
+if (project !== 'iframe' && project !== 'suite-web' && project !== 'popup') {
+    throw new Error(`Unsupported project: ${project}`);
+}
+const DIST = getDistPathForProject(project);
+
+const CORE_PUBLIC_PATH = process.env.ASSET_PREFIX ? `${process.env.ASSET_PREFIX}/` : '/';
```
mroz22, line 15: note, this is little susceptible to not defining ASSET_PREFIX with a trailing '/'. I am just brainstorming whether it would make sense to validate it, ideally using some standard API 🤔. Somewhere I did string.replaceAll('//'. '/'), just in case. but it is not very nice

packages/connect-iframe/webpack/sessions.webpack.config.ts
```diff
@@ -0,0 +1,33 @@
+import path from 'path';
+import webpack from 'webpack';
+import merge from 'webpack-merge';
+
+import { config as baseConfig } from './base.webpack.config';
+import { getSharedworkerDistPathForProject } from './utils';
+
+const project = process.env.PROJECT || 'iframe';
```
mroz22: do we need to specify project anymore? I mean - this should be only deployed with connect now so my first intuition is that accessing env.PROJECT might not be needed.

packages/connect-iframe/webpack/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+        case 'suite-web':
+            return path.join(basePath, 'suite-web', 'build');
+        case 'popup':
+            return path.join(basePath, 'connect-popup', 'build');
+        default:
+            throw new Error('Missing project.');
+    }
+};
+
+// TODO: maybe sharedworker is not needed anymore.
```
mroz22: yes, maybe not. it should only be build by connect and suite is using it via this clumsy mechanism here https://github.com/trezor/trezor-suite/blob/feat/use-core-in-suite-web/suite-common/connect-init/src/connectInitThunks.ts#L111

packages/connect/src/data/DataManager.ts
karliatto: Yes, that is probably an optimization that can be done. Interesting.

packages/suite-web/e2e/cypress.config.ts
karliatto: Nop, all the changes in this commit were fixing other issues before this one. No changes in the tests were required after removing this.
I am thinking that it will be better to delete the line `// experimentalFetchPolyfill:` instead of commenting.
It is also a deprecated experimental feature by cypress https://docs.cypress.io/app/references/experiments

```
Note: experimentalFetchPolyfill has been deprecated in Cypress 6.0.0 and will be removed in a future release. Consider using [cy.intercept()](https://docs.cypress.io/api/commands/intercept) to intercept fetch requests instead.
```

packages/connect/src/data/DataManager.ts
mroz22: oh, but it points to `data.trezor.io`. Do you actually need that @Lemonexe ? don't you have releases.json always available "locally"? I am starting to dislike that we sometimes use data.trezor.io and sometimes local copy 🤔

packages/suite-web/e2e/tests/metadata/interval-fetching.test.ts
```diff
@@ -30,8 +30,15 @@ describe('Metadata - suite is watching cloud provider and syncs periodically', (
             cy.task('startEmu', { wipe: true });
             cy.task('setupEmu', { mnemonic: 'mnemonic_all' });
             cy.task('startBridge');
+            cy.prefixedVisit('/', {
+                onBeforeLoad: win => {
+                    cy.stub(win, 'open').callsFake(stubOpen(win));
+                    cy.stub(win, 'fetch').callsFake(rerouteMetadataToMockProvider);
+                },
+            });
+            cy.getTestElement('@welcome-layout/body', { timeout: 60_000 });
```
mroz22: why exactly was this change needed?

packages/suite-web/e2e/cypress.config.ts
mroz22: I am totally in favor of removing it. But also - won't it require bigger refactoring? They say you cant use cy.stub and need to use cy.intercept. We use the former many times in the code base at the moment 🤔

packages/connect/src/data/DataManager.ts
Lemonexe: Yes and no.
What we really need is only to have _up to date_ version of `releases.json`.
- On suite **desktop** that does mean we need to fetch `data.trezor.io`, because suite might be outdated
- But suite **web** is always up to date, so yes, `releases.json` can be included in the bundle :heavy_check_mark: 
  - theoretically you could have the same browser tab open for months, but that is a negligible edge case IMO

packages/connect/src/data/DataManager.ts
karliatto: Sure, I added extended more clear comment and changed the code a bit with same functionality https://github.com/trezor/trezor-suite/pull/14959/commits/54aaa8af443d6c93e6b906f32b9a8bc294db293b

packages/connect-web/src/module/index.ts
karliatto: The reason is that this `/module/index.ts` is more like `/index.ts` at the root of `connect-web` https://github.com/trezor/trezor-suite/blob/feat/use-core-in-suite-web/packages/connect-web/src/index.ts#L0-L1

Where in `connect-web/impl/` we have the implementations in `connect-web/src/index.ts` and in `connect-web/src/module/index.ts` we have the `TrezorConnectDynamic` since the implementation of `CoreInModule` is in  `connect/src/impl/core-in-module.ts` https://github.com/trezor/trezor-suite/blob/feat/use-core-in-suite-web/packages/connect/src/impl/core-in-module.ts

karliatto: I think it makes sense, but please feel free to challenge my opinion. :)

packages/connect-iframe/package.json
karliatto: Ups :upside_down_face:  that was definitely not intended result of some rebase. Fixed it here https://github.com/trezor/trezor-suite/pull/14959/commits/5d30b2460f81f244b678f840a9841e577f9a1e7f

packages/suite-web/e2e/tests/metadata/interval-fetching.test.ts
karliatto: For some reason `cy.clock()` was interfering with the discovery using core in module. So moving the discovery to happen before `cy.clock()` is initiated was fixing it. The time out should be `30_ 000`, I only changed it  to `60_000` during development, by I don't think it is needed so I changed it here https://github.com/trezor/trezor-suite/pull/14959/commits/99e500089ebd82c24a0f75f20dd5c59397a9e413

packages/connect-iframe/webpack/core.webpack.config.ts
karliatto: I was trying to use something we already have like `ASSET_PREFIX` but I am thinking it is going to be better to be more explicit and make it `CONNECT_ASSETS_PREFIX`  so we when building we can declare it with the proper `/` at the end.

packages/suite-web/e2e/cypress.config.ts
karliatto: Exactly but we use `cy.intercept` already. It is possible that config was never needed just added as a boilerplate. And tests are working properly without it.

karliatto: Removed 4010bf9b1820bba91585722501d7132979b10b27

packages/connect-iframe/webpack/utils.ts
karliatto: If I understand well with the latest changes it is still needed for developer/testing environments  in:
* connect-explorer - https://github.com/trezor/trezor-suite/blob/develop/suite-common/connect-init/src/connectInitThunks.ts#L136
* suite-web - https://github.com/trezor/trezor-suite/blob/develop/suite-common/connect-init/src/connectInitThunks.ts#L136


Right :question:

mroz22: yes, but there is only one "source of build". previously suite-web and connect had their own builds which was wrong. Now only connect builds and uses its instance of sharedworker as a default. 

This logic in connectInitThunks is there because suite has its own copy of connect but we do not want this copy to by using also a custom copy of sharedworker - it should be looking for the "master" one.

packages/connect/src/data/DataManager.ts
mroz22: hmm I am sure it must be solvable in a better way, this kind of requires shouldn't be needed 🤔

packages/suite-web/e2e/tests/metadata/interval-fetching.test.ts
mroz22: Its probably because connect in iframe used to be isolated context, so cy.clock did not affect timers within it.

karliatto: Yes, something like that sounds like the case.

packages/connect/src/data/DataManager.ts
karliatto: We already do something similar in https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/utils/assetUtils.ts#L8

karliatto: But I can try different way to tell Webpack that it only has to include those files.

packages/connect-web/src/module/index.ts
mroz22: Hmm I will need to have this explained in person 😸 I am looking into it. It might make some sense, yes. Since suite-web is a very special consumer of @trezor/connect-web (it has its own copy) it probably makes some sense that it has very special handling - for example it absolutely doesn't need the `handleFallbackError` because there is simply no fallback possible.

packages/connect-iframe/webpack/utils.ts
mroz22: on https://dev.suite.sldev.cz/suite-web/feat/use-core-in-suite-web/web/

I get 
image

It looks like that sharedworker is not deployed properly to the expected path on https://dev.suite.sldev.cz/connect/feat/use-core-in-suite-web/

karliatto: How did you get to that error? 
I can see it is here `https://dev.suite.sldev.cz/suite-web/feat/use-core-in-suite-web/web/workers/sessions-background-sharedworker.js`  I can see the worker is there but the request you displayed in the image is missing `web` in `https://dev.suite.sldev.cz/suite-web/feat/use-core-in-suite-web/web/` so I am curious how it happened.

Edit: Ok, I am able to reproduce it now. Investigating.

karliatto: Fixed by https://github.com/trezor/trezor-suite/pull/14959/commits/bb3eb4d186f45c783eeb7f932f232214e19864a5

mroz22:
since connect won't be isolated inside the iframe anymore we are going to receive many more sentry reports (which is good). but I am afraid of this one - connect first checks if bridge is running and if not, it starts using webusb. this will probably get logged into sentry eating our quota heavily. is there a way how to work around this? 
image

edit: it can't be done probably...

mroz22:
I am still getting this 

image

packages/connect-iframe/webpack/utils.ts
mroz22: I don't think it works. I am still getting the same error on connect-explorer build on sldev. did it build the latest commit?

martykan:
Regarding Sentry, I don't think it's an actual uncaught exception, it's just that Chrome shows failed network requests in the log. I don't see any requests going to Sentry about this.

karliatto:
> since connect won't be isolated inside the iframe anymore we are going to receive many more sentry reports (which is good). but I am afraid of this one - connect first checks if bridge is running and if not, it starts using webusb. this will probably get logged into sentry eating our quota heavily. is there a way how to work around this? 
> edit: it can't be done probably...

It should be possible to filter out form sentry at the sentry level a specific type of error, we have been doing it for couple of errors for quite long time.

karliatto:
> I am still getting this
> image


@mroz22 good catch! It was missing this build instruction that is used for connect-explorer https://github.com/trezor/trezor-suite/pull/14959/commits/7404ffa1a591e2b403ce5b4477cb61ccbb6f86b1

Now it works.

image

packages/connect-iframe/webpack/utils.ts
karliatto: It was missing for connect-explorer build. This https://github.com/trezor/trezor-suite/pull/14959/commits/7404ffa1a591e2b403ce5b4477cb61ccbb6f86b1fixes it.

mroz22 (changes requested):
now it registers two sharedworkers again 
using 
- https://dev.suite.sldev.cz/suite-web/feat/use-core-in-suite-web/web/ and 
- https://dev.suite.sldev.cz/connect/feat/use-core-in-suite-web/

image

karliatto:
> now it registers two sharedworkers again using
> 
>     * https://dev.suite.sldev.cz/suite-web/feat/use-core-in-suite-web/web/ and
> 
>     * https://dev.suite.sldev.cz/connect/feat/use-core-in-suite-web/
> 
> 
> image


I see this issue should happen only on develop since we are using different sources of the worker, in production we use always same source https://connect.trezor.io/9/workers/sessions-background-sharedworker.js

karliatto:
> now it registers two sharedworkers again using
> 
>     * https://dev.suite.sldev.cz/suite-web/feat/use-core-in-suite-web/web/ and
> 
>     * https://dev.suite.sldev.cz/connect/feat/use-core-in-suite-web/
> 
> 
> image

It should be fixed now.

.github/actions/release-connect/action.yml
```diff
@@ -82,6 +82,7 @@ runs:
       shell: bash
       env:
         NODE_ENV: ${{ inputs.nodeEnv }}
+        ASSET_PREFIX: /${{ inputs.serverPath }}
```
mroz22, line 85: this is going to affect production releases, so lets double check

packages/connect/src/utils/pathUtils.ts
```diff
@@ -207,16 +209,18 @@ export const fixPath = <
...
-    // make sure that address_n is an array
-    if (utxo.address_n && typeof utxo.address_n === 'string') {
-        utxo.address_n = getHDPath(utxo.address_n);
+    // Make sure that address_n is an array
+    if (newUtxo.address_n && typeof newUtxo.address_n === 'string') {
+        newUtxo.address_n = getHDPath(newUtxo.address_n);
     }
 
-    return utxo as ProtoWithAddressN<T>;
+    return newUtxo as ProtoWithAddressN<T>;
```
mroz22: wow this commit looks unrelated. but I get the point - missing serialization. but shouldn't it be handled somewhere centrally?

packages/suite-build/configs/base.webpack.config.ts
```diff
@@ -132,7 +132,7 @@ const config: webpack.Configuration = {
                     },
                 ],
             },
-            // Workers
+            // Workers TODO(karliatto): is this worker-loader really loading anything here?
```
mroz22: does it work without it?

packages/connect-iframe/webpack/sessions.webpack.config.ts
mroz22: @karliatto? what do you think? is it needed?

packages/suite-build/configs/base.webpack.config.ts
mroz22: I think that if you remove it, it would break discovery on web

packages/connect/src/utils/pathUtils.ts
karliatto: My thinking here was to have a pure function, since modifying the argument passed specially when it is an object in JS does not look like a good idea. If you modify the provided object in the function the side effects can be unpredictable. So it is safer to create a new object and return the new object modified.

packages/suite-build/configs/base.webpack.config.ts
karliatto: Ok, so it removing actually works in web but it breaks discovery suite-desktop. That loader is the base used for suite-web and suite-desktop. So I updated the comment to https://github.com/trezor/trezor-suite/pull/14959/commits/4b3fd0276aea0b28a72f5fa3f72332cec35a736a

.github/actions/release-connect/action.yml
karliatto: Yes.  That is a good idea, I am double checking potential places where it could fail.

packages/connect/src/utils/pathUtils.ts
mroz22: 🤔 
btw: where it was failing? some CI test? Because I assume that it started failing because of missing serialization but in suite-desktop, connect is run inside node and I think there is no serialization either. So this should be happening there as well.

.github/actions/release-connect/action.yml
karliatto: Yes, I double checked the actions and it should work. We use the same action for dev environment and production so it is very close. In production the serverPath will be:

* The version relase => `serverPath: 9.4.4-beta.1`
* The static conenect v9 release => `serverPath: 9`

packages/suite-build/configs/base.webpack.config.ts
```diff
@@ -132,7 +132,7 @@ const config: webpack.Configuration = {
                     },
                 ],
             },
-            // Workers
+            // This worker loader is used for suite-desktop
```
mroz22, line 135: I am still not sure, I just commented out this part 
```
  // {
                    //     loader: 'worker-loader',
                    //     options: {
                    //         filename: 'static/worker.[contenthash].js',
                    //     },
                    // },
```
and discovery works for me using `yarn suite:dev:desktop`

karliatto: Really? I was broken for me, but let me try again.

packages/connect/src/utils/pathUtils.ts
karliatto: It was failing in suite-web send test if I remember well. Basically it was modifying an object that was freezed, probably because it was an object from the reducer that is not allowed to be directly modified.

mroz22: Aha, I know.
Carlos, please lets do it like this:

- create an issue that there might be some places causing side/effects object mutations
- link this issue with a todo to this file (packages/connect/src/utils/pathUtils.ts)
- suite-destkop actually has serialization because data passed to connect calls first go through renderer-main interaface. 
- so lets mimic the same behaviour for connect-in-module for now. I would clone the input param of connect here https://github.com/trezor/trezor-suite/blob/feat/use-core-in-suite-web/packages/connect/src/impl/core-in-module.ts#L208

packages/suite-build/configs/base.webpack.config.ts
karliatto: It definitely does not work for me if I comment that in desktop, I get error:

image

mroz22: Oh, this is Graph worker. It has nothing to do with connect. So ok, I tested it with a seed with no balance so it could be it.

packages/utils/src/index.ts
```diff
@@ -49,3 +49,4 @@ export * from './throttler';
 export * from './extractUrlsFromText';
 export * from './isFullPath';
 export * from './asciiUtils';
+export * from './cloneCyclicObject';
```
mroz22: nit: I would call it `cloneObjectCyclic` so that it is nicely sorted alphabetically just after `cloneObject`. totally feel free to dismiss

packages/connect/src/utils/pathUtils.ts
```diff
@@ -1,5 +1,7 @@
 // origin: https://github.com/trezor/connect/blob/develop/src/js/utils/pathUtils.js
 
+import { cloneObject } from '@trezor/utils';
```
mroz22: so it can go away now after https://github.com/trezor/trezor-suite/pull/14959/files#diff-8c98ba84fc78374812ebc5a974004cbaf72f6245466b70f82c6e0a53f8eadcd3R210 was added?

suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -124,19 +124,11 @@ export const connectInitThunk = createThunk(
...
-                const assetPrefixArr = (process.env.ASSET_PREFIX || '').split('/').filter(Boolean);
-                const relevantSegments = assetPrefixArr
-                    .map((segment, index) => {
-                        const first = index === 0;
-                        const last = index === assetPrefixArr.length - 1;
-                        if (segment === 'suite-web' && first) return 'connect';
-                        if (segment === 'web' && last) return null;
-
-                        return segment;
-                    })
```
mroz22: are you sure that we don't need this segment? I already had this file marked as "viewed" and now this change appeared here again I think

mroz22: I am getting this again. feels related
image

packages/utils/src/index.ts
karliatto: Sure, `cloneObjectCyclic` looks better. Thanks! :heavy_check_mark: Done

suite-common/connect-init/src/connectInitThunks.ts
karliatto: Yes, you are right, that came back as a result of the rebase, I missed it, but it is back now it is working as expected.

packages/connect/src/utils/pathUtils.ts
karliatto: We probably could, but why do you think it will be better?
For me, an utility function that modifies the passed attribute, sounds like a bad practice.

packages/connect/src/utils/pathUtils.ts
karliatto: Ok, I created issue https://github.com/trezor/trezor-suite/issues/15559 and liked it to that file.

packages/utils/src/cloneObjectCyclic.ts
```diff
@@ -0,0 +1,37 @@
+// Makes a deep copy of an object, handling cyclical references.
```
mroz22, line 1: cc @marekrjpolak for review

packages/connect/src/utils/pathUtils.ts
mroz22: please remove it 🙏 I would like to validate that the new global utility really does the job

packages/utils/src/cloneObjectCyclic.ts
marekrjpolak: I like the util. :+1: My only comment is whether we want to replace old `cloneObject` with it or not as this one seems safer. So it narrows down to performance matter probably?

karliatto: Yes, I did not want to replace it in case it cause some unexpected issues, since this PR is already doing quite a big changes. Maybe we can do it in future release.

mroz22: yes, please lets do this in a followup 👍

packages/connect/src/utils/pathUtils.ts
karliatto: https://github.com/trezor/trezor-suite/pull/14959/commits/a30a0a676b248ac173a0b525d07ad8704efb916c

karliatto: ok, so after removing it tests are green https://github.com/trezor/trezor-suite/actions/runs/12031049869/job/33541253694?pr=14959

:relaxed:

mroz22:
perfect. I will do some final testing and we can merge it

mroz22 (approved):

PR #14963, opened 2024-10-20
feat(mobile): new icons

### LINKED ISSUE #11596: Icons don't follow theme right after theme change
**Describe the bug**
Quite a weird bug but when I switch theme and go back to settings screen the icons have the old theme styles (attached recording)

**Info:**
-   Mobile Suite 24.3.1
-   OS: iOS

**How to reproduce**

1. Switch themes
1. See go back icon
1. See icons on settings screen

**Expected behavior**
Should have styles of the currently selected theme.

**Screenshots**
https://github.com/trezor/trezor-suite/assets/66002635/c59a7fc1-e713-4c98-8a67-037abdd1e84a

### LINKED ISSUE #12433: Iconset replacement
We've finally received the unified iconset where we're no longer using strokes which caused that we had to keep 5 different variants of each icon. Now we're using scalable stroke so no matter at what size you export it, it'll look consistently throughout. 

What we’ll need to do (likely) is to replace all the icons in the repo with new ones. I’d just need someone to check: 
- If the naming convention is the same 
- If we’re missing any icons
- If some icons aren’t matching
- Replace it in repo

[Outline.zip](https://github.com/user-attachments/files/16837224/Outline.zip)
source: https://www.figma.com/design/BdbvEYgjDdntL3haFcXRgV/%5BAssets%5D-Icons?node-id=1746-54&node-type=SECTION&t=u59DUGSGadXCIoTQ-11

### LINKED ISSUE #14311: Sporadic missing icons in dark mode
Missing icons appear sporadically, so far, only noticed in dark mode. Artifact is resolved when navigating to another screen and navigating back.

image
image

Nodonisko:
I recommend check this commit by commit, first one is creating new mobile icons package and some refactoring. Second one is just replacing deprecated package with new one.

## Description

This new icons package is created with main focus on performance and minimal performance overhead so it should be super fast. I also did some refactoring and simplification of some `Button` components which also helped with performance little bit.

## Related Issue

Resolve #12433 
Fixes https://github.com/trezor/trezor-suite/issues/11596
Fixes #14311


Nodonisko:
@SocketSecurity ignore npm/ttf2woff2@5.0.0

Nodonisko:
`ttf2woff2` package looks safe and it's used only locally for generating font files so it should be safe

suite-native/icons/src/Icon.tsx
```diff
@@ -0,0 +1,127 @@
...
+    ios: RNText,
+});
+
+export type IconProps = {
+    name: IconName;
+    size?: IconSize | number;
+    color?: IconColor;
+};
+
+export const Icon = ({ name, size = 'large', color = 'iconDefault' }: IconProps) => {
```
matejkriz: Not sure what exact change caused Device info and Open passphrase buttons to be cut off, but looks ok on develop.

image

matejkriz:
I guess this can fix also 
https://github.com/trezor/trezor-suite/issues/14311
and 
https://github.com/trezor/trezor-suite/issues/11596
right, @Nodonisko ?

Nodonisko:
Yes it should fix all icon related issue we have now.

suite-native/module-dev-utils/src/screens/DemoScreen.tsx
```diff
@@ -24,13 +24,13 @@ import {
     TextButton,
     NumPadButton,
     TextButtonVariant,
-    Card,
-    ListItemSkeleton,
+    // Card,
```
matejkriz, line 27: leftover...

matejkriz (approved):
I didn't go through the individual icon changes, I mostly check only the first commit and it works nice, including switching between dark and light mode 🎉 

Only issue I found out is the cut off buttons in the Device switcher.

It would be great to update README of @suite-common/icons too, but it can be a followup.

suite-common/icons/generateIconFont.ts
```diff
@@ -0,0 +1,160 @@
...
+    'pencilSimpleLine',
+    'piggyBank',
+    'piggyBankFilled',
+    'plus',
+    'plusCircle',
+    'qrCode',
+    'question',
+    'question',
+    'question',
+    'question',
```
matejkriz: duplicated

suite-common/icons/generateIconFont.ts
```diff
@@ -0,0 +1,160 @@
...
+    'stack',
+    'swap',
+    'trashSimple',
+    'treeStructure',
+    'trezorLogo',
+    'trezorModelOne',
+    'trezorModelOneFilled',
+    'trezorModelT',
+    'trezorModelTFilled',
+    'trezorSafe3',
```
matejkriz, line 92: Not as part of this PR, but it would be great to use internal names for file names for Trezor models (T2B1 instead of modelOne and so on). It's then easier to search for them, easier mapping and so on.

matejkriz, line 92: Icon for Safe 3 is somehow broken 
image

correct:
image

matejkriz (changes requested):
It looks like the icon for Safe 3 is broken only in the font generated from correct SVG, so maybe there are more icons broken, only not so obvious?

seibei-iguchi (approved):

suite-common/icons/generateIconFont.ts
Nodonisko: Figma is source of truth here, I don't want to clash with it because it would require manual work everytime when updating icon from Figma in future. For us it's abstracted away in DeviceIcon.tsx anyway which should use correct model ids.

Nodonisko: There could be sometimes problem when svg uses `fill-rule` attribute but I verified other icons we use that have this property and they look correct.

I fixed this icon manually and added guide how to fix it to README in case we will need it.

suite-native/icons/src/Icon.tsx
Nodonisko: I am not able to reproduce this one, can you please try again? Maybe it was some random temp issue.

suite-common/icons/generateIconFont.ts
matejkriz: Agree, internal names should be ideally used directly in Figma.

matejkriz: Looks good now 👍 
image
image

matejkriz (approved):
👏

matejkriz (approved):
👏

PR #14966, opened 2024-10-21
fix(protobuf): skip benchmark messages

martykan:
## Description

`messages-benchmark.proto` seems to be causing some issue in build and they probably shouldn't be included anyway, add them to the list of skipped defs.


mroz22 (approved):
we are not using it imho. let skip that

PR #14982, opened 2024-10-21
feat(suite-native): send max

### LINKED ISSUE #13791: Send coins - send max
## Acceptance criteria
- [ ] there is a Send max button next to the `amount to be sent` input
- [ ] When user selects send max, balance is calculated as `balance - normal_fee`
- [ ] When user changes fee level on the next screen, the totalAmount stays unchanged and the outputs are adjusted so it is always possible to send the transaction with any fee level

PeKne:
## Description
- send max button is displayed in the send form
- if user processed to fee selection with send max enabled, the total amount is always equal to the account balance

## Related Issue

Resolve #13791 

## Screenshots:

https://github.com/user-attachments/assets/8a573f1d-393b-4798-b9f9-bcd4df7bb57d

### DIFF
--- a/suite-native/module-send/src/components/FiatAmountInput.tsx
+++ b/suite-native/module-send/src/components/FiatAmountInput.tsx
@@ -13,6 +13,7 @@ import { SendAmountCurrencyLabelWrapper, sendAmountInputWrapperStyle } from './C
 import { SendAmountInputProps } from '../types';
 import { useSendAmountTransformers } from '../hooks/useSendAmountTransformers';
 import { getOutputFieldName } from '../utils';
+import { SendOutputsFormValues } from '../sendOutputsFormSchema';
 
 export const FiatAmountInput = ({
     recipientIndex,
@@ -25,7 +26,7 @@ export const FiatAmountInput = ({
     isDisabled = false,
 }: SendAmountInputProps) => {
     const { applyStyle } = useNativeStyles();
-    const { setValue } = useFormContext();
+    const { setValue } = useFormContext<SendOutputsFormValues>();
     const fiatCurrencyCode = useSelector(selectFiatCurrencyCode);
     const { fiatAmountTransformer } = useSendAmountTransformers(networkSymbol);
     const converters = useCryptoFiatConverters({ networkSymbol });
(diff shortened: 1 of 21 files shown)


suite-native/module-send/src/components/FiatAmountInput.tsx
```diff
@@ -57,6 +57,7 @@ export const FiatAmountInput = ({
         setValue(cryptoFieldName, converters?.convertFiatToCrypto?.(transformedValue), {
             shouldValidate: true,
         });
+        setValue('setMaxOutputId', undefined);
```
yanascz, line 62: The name is also used in `CryptoAmountInput`, would it make sense to create a constant for that?

suite-native/module-send/src/components/SendMaxButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+    useEffect(() => {
+        if (isSendMaxAvailable) calculateFeeLevelsMaxAmount();
+        else setMaxAmountValue(undefined);
+    }, [isSendMaxAvailable, calculateFeeLevelsMaxAmount]);
+
+    const setOutputSendMax = () => {
+        if (!maxAmountValue) return;
+        Keyboard.dismiss();
+
+        setValue('setMaxOutputId', outputIndex);
```
yanascz, line 72: The same field name used for the third time... 🤔

PeKne:
@yanascz The problem with inconsistent `fee per unit` and `fee` values should be solved by: [cca0c93](https://github.com/trezor/trezor-suite/pull/14982/commits/cca0c93481177c9d3a23066593b34f7ed7b3aa69)

suite-native/module-send/src/components/FiatAmountInput.tsx
PeKne: That is the name of the field from [SendOutputsFormValues](https://github.com/trezor/trezor-suite/blob/acb46c296763d940e73bb7e7e33ad917c9ca0ad9/suite-native/module-send/src/sendOutputsFormSchema.ts#L139) and Typescript should tell you what the possible values are.

Unfortunately, now I realize that the typing of our custom form hook was wrong 😑. So I fixed the typescript type check in [acb46c2](https://github.com/trezor/trezor-suite/pull/14982/commits/acb46c296763d940e73bb7e7e33ad917c9ca0ad9). Since now TypeScript correctly tells you what you should fill in, the constant is not needed in my opinion.

suite-native/module-send/src/components/SendMaxButton.tsx
PeKne: The comment from above applies here as well.

suite-native/module-send/src/components/FiatAmountInput.tsx
yanascz: OK, then it makes sense like this. 👍

yanascz (approved):
Works fine, I'm unable to break it. 🥳

PR #14993, opened 2024-10-21
feat: implement random strategy for utxo sorting

### LINKED ISSUE #10765: Abandon the use of BIP-69

peter-sanderson:
This PR introduces a replacement of BIP69 with random sorting. See the linked issue for explanation.

Resolves https://github.com/trezor/trezor-suite/issues/10765

### DIFF
--- a/packages/connect/src/constants/utxo.ts
+++ b/packages/connect/src/constants/utxo.ts
@@ -1,3 +1,3 @@
 import type { TransactionInputOutputSortingStrategy } from '@trezor/utxo-lib';
 
-export const DEFAULT_SORTING_STRATEGY: TransactionInputOutputSortingStrategy = 'bip69';
+export const DEFAULT_SORTING_STRATEGY: TransactionInputOutputSortingStrategy = 'random';
--- a/packages/utxo-lib/src/compose/sorting/randomSortingStrategy.ts
+++ b/packages/utxo-lib/src/compose/sorting/randomSortingStrategy.ts
@@ -1,0 +1,36 @@
+import { SortingStrategy } from './sortingStrategy';
+import { convertOutput } from './convertOutput';
+import { arrayShuffle, getRandomInt } from '@trezor/utils';
+
+export const randomSortingStrategy: SortingStrategy = ({ result, request, convertedInputs }) => {
+    const nonChangeOutputPermutation: number[] = [];
+    const changeOutputPermutation: number[] = [];
+
+    const convertedOutputs = result.outputs.map((output, index) => {
+        if (request.outputs[index]) {
+            nonChangeOutputPermutation.push(index);
+
+            return convertOutput(output, request.outputs[index]);
+        }
+
+        changeOutputPermutation.push(index);
+
+        return convertOutput(output, { type: 'change', ...request.changeAddress });
+    });
+
+    /**
+     * The goal here is to randomly insert change outputs into the outputs array.,
+     * so you cannot tell what is the change just by the order of the transaction.
+     */
+    const permutation = [...nonChangeOutputPermutation];
+    const newPositionOfChange = getRandomInt(0, permutation.length + 1);
+    permutation.splice(newPositionOfChange, 0, ...changeOutputPermutation);
+    const sortedOutputs = permutation.map(index => convertedOutputs[index]);
+
+    return {
+        /** Randomly shuffle inputs to make it harder to fingerprint the Trezor Suite. */
+        inputs: arrayShuffle(convertedInputs, { randomInt: getRandomInt }),
+        outputs: sortedOutputs,
+        outputsPermutation: permutation,
+    };
+};
--- a/packages/utxo-lib/src/compose/sorting/randomSortingStrategy.ts
+++ b/packages/utxo-lib/src/compose/sorting/randomSortingStrategy.ts
@@ -24,6 +24,7 @@ export const randomSortingStrategy: SortingStrategy = ({ result, request, conver
      */
     const permutation = [...nonChangeOutputPermutation];
     const newPositionOfChange = getRandomInt(0, permutation.length + 1);
+
     permutation.splice(newPositionOfChange, 0, ...changeOutputPermutation);
     const sortedOutputs = permutation.map(index => convertedOutputs[index]);
 
--- a/packages/utxo-lib/tests/__fixtures__/compose.ts
+++ b/packages/utxo-lib/tests/__fixtures__/compose.ts
@@ -30,6 +30,7 @@ type Fixture = {
         changeAddress: AnyComposeRequest['changeAddress'] & { path?: number[] | string };
     };
     result: AnyComposeResult;
+    randomIntSequence?: number[];
 };
 
 export const composeTxFixture: Fixture[] = [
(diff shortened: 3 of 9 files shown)


packages/utxo-lib/tests/compose.test.ts
```diff
@@ -53,7 +80,7 @@ describe('composeTx addresses cross-check', () => {
 
             addrKeys.slice(0, offset).forEach(addressType => {
                 const key = `${txType}-${addressType}` as keyof typeof f.result;
-                it(`${String(key)} ${f.description}`, () => {
```
peter-sanderson, line 56: @Lemonexe I have no idea why, but Idea no longer bitches about this. :shrug:

packages/connect/src/constants/utxo.ts
```diff
@@ -1,3 +1,3 @@
 import type { TransactionInputOutputSortingStrategy } from '@trezor/utxo-lib/libDev/src';
```
Lemonexe: Bad import.
I'm always on the lookout for this, because IDEA does it all the time, and I have not yet figured out how to configure it :see_no_evil:

packages/utxo-lib/src/compose/sorting/randomSortingStrategy.ts
```diff
@@ -0,0 +1,39 @@
...
+        changeOutputPermutation.push(index);
+
+        return convertOutput(output, { type: 'change', ...request.changeAddress });
+    });
+
+    /**
+     * The goal here is to randomly insert change outputs into the outputs array.,
+     * so you cannot tell what is the change just by the order of the transaction.
+     */
+    const permutation = [...nonChangeOutputPermutation];
```
Lemonexe, line 25: Why introduce this variable?
Is it to make the code look similar with `bip69sortingStrategy`? Imho needless, but I'm fine with that :slightly_smiling_face: 
But in any case, you can simplify `permutation = nonChangeOutputPermutation`, as there appears to be no reason to keep the latter immutable.

packages/utxo-lib/src/compose/sorting/randomSortingStrategy.ts
```diff
@@ -0,0 +1,39 @@
...
+    // Min (0) is inclusive, max (permutation.length + 1) is exclusive
+    // Example: for array [0, 1, 2] the result can be: 0, 1, 2, 3
```
Lemonexe: I don't think the comment is necessary. One click on `getRandomInt` will tell you the same thing.

packages/utxo-lib/tests/compose.test.ts
```diff
@@ -5,6 +5,29 @@ import { verifyTxBytes } from './compose.utils';
...
 import { fixturesCrossCheck } from './__fixtures__/compose.crosscheck';
 
+jest.mock('@trezor/utils', () => {
+    const actual = jest.requireActual('@trezor/utils');
+
+    let fakeRandomIndex = 0;
+    const fakeRandom = [
+        1, // Called from `sortingStrategy=random` test, for outputs (so change output is inserted at index 1 after the first output)
+        1, // Called from `sortingStrategy=random` test, when shuffling inputs
+    ];
```
Lemonexe: Why can't we simply have `getRandomInt: () => 1`

I assume you'd like to future-proof this, so in another test you can work with `getRandomInt` returning something else than `1`, is that right?

Then this doesn't seem very maintainable to me - you have maintain the link that `fakeRandom[0]` is for _this test_, while `fakeRandom[1]` is for _that test_, but just rearranging the tests will break this link, if I'm not mistaken?
Currently they're both 1, but at any rate, the comments would no longer make sense.

If I understood that correctly, I'd suggest this:
- here: `getRandomInt: () => 1`
- in future test you redefine the mock so it returns `123` or whatever, for that test file
  - that seems to me as the normal approach, to redefine mocks as needed for specific tests

packages/connect/src/constants/utxo.ts
peter-sanderson: Me neither, I hate it. If anyone figures the configuration out it would be great

packages/utxo-lib/tests/__fixtures__/compose.ts
```diff
@@ -524,6 +524,59 @@ export const composeTxFixture: Fixture[] = [
...
+                    amount: '150000',
+                    type: 'payment',
+                },
+                {
+                    address: '1CrwjoKxvdbAnPcGzYjpvZ4no4S71neKXT',
+                    amount: '50262',
+                    type: 'change',
+                },
+            ],
+            outputsPermutation: [0, 1],
```
Lemonexe: :thinking: This test case technically does cover the `'random'`, but if I'm not mistaken, this is the same result that would be obtained from `'none'`.
- it only tests that `randomSortingStrategy` works at all for the simplest case 
- but you could test a multi-output transaction, so there's really something to shuffle
  - I assume that change would have default index 2, but the mocked random would produce `[0, 2, 1]`, is that right? That would be interesting test case :slightly_smiling_face:

Lemonexe (approved):
LGTM :rocket: :game_die:
But it'd be great if someone more knowledgable took a look at the `randomSortingStrategy` itself - the implementation makes sense to me, but it's not trivial :slightly_smiling_face: 

Mostly minor comments on codestyle & tests.
But I'd make the tests more thorough, [see comment](https://github.com/trezor/trezor-suite/pull/14993#discussion_r1810518868).

packages/utxo-lib/src/compose/sorting/randomSortingStrategy.ts
peter-sanderson: I didnt want to do a mutation on `nonChangeOutputPermutation` as it comes from code above and there is logical boundary

Lemonexe: Resolving; we clarified in person, `permutation` contains both change and nonChange, so it can't be called nonChange

packages/utxo-lib/tests/compose.test.ts
```diff
@@ -5,13 +5,31 @@ import { verifyTxBytes } from './compose.utils';
 import { composeTxFixture } from './__fixtures__/compose';
 import { fixturesCrossCheck } from './__fixtures__/compose.crosscheck';
 
+import { getRandomInt, arrayShuffle } from '@trezor/utils';
+
+jest.mock('@trezor/utils');
+
+(arrayShuffle as jest.Mock).mockImplementation(jest.requireActual('@trezor/utils').arrayShuffle);
```
Lemonexe: Nit: this would be more conservative approach. A combination of your new code & your previous code - instead of mocking whole import and restoring only `arrayShuffle`, this restores whole import, but only `getRandomInt` is mocked. Writing it this way makes it clear that we _only need to_ mock the getRandomInt.

Pick whatever you like better :slightly_smiling_face: 

```suggestion
jest.mock('@trezor/utils', () => ({
    ...jest.requireActual('@trezor/utils'),
    getRandomInt: jest.fn(),
}));
```

packages/utxo-lib/tests/__fixtures__/compose.ts
```diff
@@ -524,6 +525,79 @@ export const composeTxFixture: Fixture[] = [
...
     },
+    {
+        description:
+            'sorts the inputs randomly and puts change at random place between user-defined outputs' +
+            'when sortingStrategy=random',
+        randomIntSequence: [
+            1, // For outputs (so change output is inserted at index 1 after the first output)
+            1, // Shuffling inputs (Fisher-Yates swap last, the 3rd with second)
+            0, // Shuffling inputs (Fisher-Yates swap second with the first)
+        ],
```
Lemonexe, line 536: :heart:

Lemonexe (approved):

packages/utxo-lib/tests/compose.test.ts
peter-sanderson: Good point, improved!

PR #14994, opened 2024-10-21
chore: improve fixture for compose.test (add strict types)

peter-sanderson:
Part of https://github.com/trezor/trezor-suite/issues/10765

### DIFF
--- a/packages/utxo-lib/tests/__fixtures__/compose.crosscheck.ts
+++ b/packages/utxo-lib/tests/__fixtures__/compose.crosscheck.ts
@@ -1,6 +1,23 @@
 import { UTXO } from './compose';
+import {
+    ComposeChangeAddress,
+    ComposeInput,
+    ComposeOutput,
+    ComposeRequest,
+} from '../../src/types/compose';
+import { CoinSelectPaymentType } from '../../src';
 
-export const fixturesCrossCheck = [
+type AnyComposeRequest = ComposeRequest<ComposeInput, ComposeOutput, ComposeChangeAddress>;
+
+type Fixture = {
+    description: string;
+    request: Omit<AnyComposeRequest, 'network'> & {
+        network?: AnyComposeRequest['network'];
+    };
+    result: Partial<Record<`${CoinSelectPaymentType}-${CoinSelectPaymentType}`, { bytes: number }>>;
+};
+
+export const fixturesCrossCheck: Fixture[] = [
     {
         description: '1 input, 1 output, no change',
         request: {
@@ -210,7 +227,6 @@ export const fixturesCrossCheck = [
                 {
                     type: 'opreturn',
                     dataHex: 'deadbeef',
-                    amount: '0',
                 },
             ],
         },
--- a/packages/utxo-lib/tests/__fixtures__/compose.ts
+++ b/packages/utxo-lib/tests/__fixtures__/compose.ts
@@ -1,4 +1,14 @@
-export const UTXO = {
+import {
+    ComposeChangeAddress,
+    ComposeInput,
+    ComposeOutput,
+    ComposeRequest,
+    ComposeResult,
+    ComposeResultFinal,
+} from '../../src/types/compose';
+import { bitcoincash, doge } from '../../src/networks';
+
+export const UTXO: ComposeInput & { path: number[] } = {
     path: [44, 1, 0, 0, 0], // NOTE: this field is not required by the ComposeInput interface, yet it is accepted in and received out
     coinbase: false,
     own: true,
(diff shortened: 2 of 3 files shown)


packages/utxo-lib/tests/compose.test.ts
```diff
@@ -1,27 +1,24 @@
-import { composeTx } from '../src';
+import { composeTx } from '../src/compose';
```
Lemonexe, line 1: Nit: seems like unnecessary change?

packages/utxo-lib/tests/compose.test.ts
```diff
@@ -1,27 +1,24 @@
...
 
 import { verifyTxBytes } from './compose.utils';
 import { composeTxFixture } from './__fixtures__/compose';
 import { fixturesCrossCheck } from './__fixtures__/compose.crosscheck';
 
-// keyof typeof NETWORKS;
-const getNetwork = (name?: string) =>
-    // @ts-expect-error expression of type string can't be used to index type
-    typeof name === 'string' && NETWORKS[name] ? NETWORKS[name] : NETWORKS.bitcoin;
-
```
Lemonexe, line 12: Good riddance :smile: :wave:

packages/utxo-lib/tests/compose.test.ts
```diff
@@ -1,27 +1,24 @@
...
+        const result = { ...f.result };
 
         it(f.description, () => {
-            const tx = composeTx(request as any);
+            const tx = composeTx(request);
             expect(tx).toEqual(result);
 
-            if (tx.type === 'final') {
-                verifyTxBytes(tx, f.request.txType as any, network);
+            expect(f.request.txType).not.toEqual('p2wsh');
```
Lemonexe, line 18: This could imho do with a short comment: why we are explicitly testing that the _**fixtures**_ don't contain a `p2wsh` type transaction? 
I can see that `verifyTxBytes` does not accept a `p2wsh` type transaction, _whatever that is_, but it's not apparent until you play around with the code.

packages/utxo-lib/tests/compose.test.ts
```diff
@@ -36,24 +33,27 @@ describe('composeTx addresses cross-check', () => {
...
             // skip test for each addressType if there is nothing to replace (example: 7 inputs test)
-            const offset = f.request.outputs.find(o => o.address === 'replace-me')
+            const offset = f.request.outputs.find(o => 'address' in o && o.address === 'replace-me')
                 ? addrKeys.length
                 : 1;
 
             addrKeys.slice(0, offset).forEach(addressType => {
                 const key = `${txType}-${addressType}` as keyof typeof f.result;
-                it(`${key} ${f.description}`, () => {
+                it(`${String(key)} ${f.description}`, () => {
```
Lemonexe, line 56: this seems useless?

packages/utxo-lib/tests/__fixtures__/compose.crosscheck.ts
```diff
@@ -1,6 +1,23 @@
...
+import { CoinSelectPaymentType } from '../../src';
 
-export const fixturesCrossCheck = [
+type AnyComposeRequest = ComposeRequest<ComposeInput, ComposeOutput, ComposeChangeAddress>;
+
+type Fixture = {
+    description: string;
+    request: Omit<AnyComposeRequest, 'network'> & {
+        network?: AnyComposeRequest['network'];
+    };
```
Lemonexe, line 16: IMHO we should give some thought to explore https://www.npmjs.com/package/@total-typescript/shoehorn 

Not the scope of this PR ofc, this is ok :slightly_smiling_face: 

Also, don't we have some type generic util for Nullable fields on object? Can't find it, so maybe I'm wrong.

packages/utxo-lib/tests/__fixtures__/compose.ts
```diff
@@ -958,7 +983,7 @@ export const composeTxFixture = [
         description: 'builds a simple tx without change (cashaddr)',
         request: {
             changeAddress: { address: 'bitcoincash:qzppkat2v7xu9fr3yeuqdnggjqqltrs7pcg8swvhl0' },
-            network: 'bitcoincash',
+            network: bitcoincash,
```
Lemonexe, line 986: This was literally wrong, thx :+1: :+1:

packages/utxo-lib/tests/__fixtures__/compose.ts
```diff
@@ -39,7 +63,7 @@ export const composeTxFixture = [
                     amount: '100000',
                     type: 'payment',
                     customField: 'prove that payment output is generic',
-                },
+                } as AnyComposeFinalResult['outputs'][number], // hack for `customField`
```
Lemonexe, line 66: Why don't you declare the same thing for `Fixture.result` as for `Fixture.request` if you want to support `customField`?
Boilerplate is quicker to read than inconsistent approaches.

Lemonexe (approved):
Thanks a lot! It's a great code improvement, makes the tests less hacky and magic.

I do have some comments for codestyle or TS.

packages/utxo-lib/tests/compose.test.ts
peter-sanderson: I think that we shall not import in test from barrel index files, but rather from the file we are directly testing.


Barrel files just define the outer API of the package, and tests are internal stuff of that package.

packages/utxo-lib/tests/compose.test.ts
peter-sanderson: TBH I just tweak it to pass the checks. This is mystery to me and I dont understand why this is the case.

It was out-of-scope for me.

Lemonexe: Oh that's fine :ok_hand:   I was just wondering if you have more insight :slightly_smiling_face:

packages/utxo-lib/tests/compose.test.ts
peter-sanderson: I got an error (from Idea I guess) that this is a proper way how to convert `symbol` to string. That just doing 
```
`${...}`
```

may fail. 

Obviously not here, I wanted to get rid of the error.

packages/utxo-lib/tests/__fixtures__/compose.ts
peter-sanderson: That doesnt work, you cannot `Omit` a fields as ComposeResult is union type, where that field is not present in all components. :(

Lemonexe: Got it! Did not realise that.

PR #14996, opened 2024-10-21
Display tokens and staking information at dashboard

### LINKED ISSUE #14456: EVM Tokens / Add tokens and staking info to dashboard assets overview
**Description:**
Please add `Tokens` and `Staking` info to the dashboard "My Assets" section as per design below and as is already available in sidebar

[Design Figma](https://www.figma.com/design/0Kz3w6WDlvQm62RjOm87na/Staking?node-id=6130-28049&t=bVgXOf4lz3Nz6qDE-1)

**Current state:**
Screenshot 2024-09-30 at 9 54 52
Screenshot 2024-09-30 at 10 14 34

________
**Future state:**

- prerequisite [EVM Tokens / Token icons bundle #14611](https://github.com/trezor/trezor-suite/issues/14611)

**Table view:**
- prerequisite [EVM Tokens / Generic table structure #13152](https://github.com/trezor/trezor-suite/issues/13152)

image

**Card view:**
image

enjojoy:
## Description

- Display tokens and staking information for each asset.
- Some minor refactors

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14456

### Before:

- Piggy has a bellybutton
image

- Dashboard table
image

- Dashboard grid
image

### After:

- Piggy doesn't have a bellybutton anymore
image

- Dashboard table
image
image

- Dashboard grid
image
image


tomasklim (changes requested):
Please validate it against designs provided in Figma

e.g.

Missing "tokens" and "staking" text next to pig and token icons
Screenshot 2024-10-28 at 10 57 04

Figma:
Screenshot 2024-10-28 at 10 57 34

enjojoy:
Thank you, I also fixed the overall table alignment so it's more like in figma. It's in fixup https://github.com/trezor/trezor-suite/pull/14996/commits/b4628e27cc66918ccfacaafee490abb7e52e97b7
image

@tomasklim Should I also change Balance to Amount like in the design?

packages/suite/src/views/dashboard/AssetsView/AssetCoinLogo.tsx
```diff
@@ -35,7 +26,7 @@ export const AssetCoinLogo = ({ symbol, assetsFiatBalances, index }: AssetCoinLo
         : undefined;
 
     return (
-        <LogoWrapper>
+        <Row alignItems="center" justifyContent="center">
```
seibei-iguchi: alignItems="center" is the default

packages/suite/src/views/dashboard/AssetsView/AssetCard/AssetCard.tsx
```diff
@@ -21,11 +21,25 @@ import { FiatHeader } from 'src/components/wallet/FiatHeader';
...
+import { selectAssetAccountsThatStaked } from '@suite-common/wallet-core';
+import { BigNumber } from '@trezor/utils';
+import { selectCoinDefinitions } from '@suite-common/token-definitions';
+import { getTokens } from 'src/utils/wallet/tokenUtils';
+import { enhanceTokensWithRates } from 'src/utils/wallet/tokenUtils';
+import { sortTokensWithRates } from 'src/utils/wallet/tokenUtils';
+import { FiatCurrencyCode } from '@suite-common/suite-config';
 
 // eslint-disable-next-line local-rules/no-override-ds-component
 const StyledCard = styled(Card)`
```
seibei-iguchi, line 43: Please remove this override and the custom padding, use paddingType instead.

packages/suite/src/views/dashboard/AssetsView/AssetCard/AssetCard.tsx
```diff
@@ -34,7 +48,12 @@ const StyledCard = styled(Card)`
 `;
 
 const Content = styled.div`
```
seibei-iguchi: Can be done with Column. The padding is not needed, I think.

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetRow.tsx
```diff
@@ -49,64 +95,113 @@ export const AssetRow = memo(
...
-                </Table.Cell>
-            </Table.Row>
+                    </Table.Cell>
+                    <Table.Cell align="right">
+                        {!isTestnet(symbol) && <PriceTicker symbol={symbol} />}
+                    </Table.Cell>
+
+                    <Table.Cell>{!isTestnet(symbol) && <TrendTicker symbol={symbol} />}</Table.Cell>
+                    <Table.Cell align="right" colSpan={2}>
+                        <Row gap={16}>
```
seibei-iguchi: Use spacings

seibei-iguchi (changes requested):

seibei-iguchi:
I'm not sure this is a good approach, it seems too fragile and magical, using the padding and positioning. I'd rather not use separate rows for this use case and do it inside one row, I think.

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetTable.tsx
```diff
@@ -1,60 +1,63 @@
...
-    </Table>
-);
+    localCurrency,
+    currentFiatRates,
+}: AssetTableProps) => {
+    return (
+        <Table isRowHighlightedOnHover margin={{ top: spacings.md, bottom: spacings.md }}>
+            <Table.Header>
+                <Table.Row>
+                    <Table.Cell colSpan={1} padding={{ right: spacingsPx.xxs }}></Table.Cell>
```
seibei-iguchi: Why the empty cell?

packages/suite/src/views/dashboard/AssetsView/AssetCoinLogo.tsx
enjojoy: https://github.com/trezor/trezor-suite/pull/14996/commits/d1b9c30e115df36aba8ad13e537786aa33226844

packages/suite/src/views/dashboard/AssetsView/AssetCard/AssetCard.tsx
enjojoy: https://github.com/trezor/trezor-suite/pull/14996/commits/e57c6a41d63a7760b80706c1b73811aeca389776

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetRow.tsx
enjojoy: https://github.com/trezor/trezor-suite/pull/14996/commits/4458ddcc3a6e480a9f577c96610327de1345259a

enjojoy:
> I'm not sure this is a good approach, it seems too fragile and magical, using the padding and positioning. I'd rather not use separate rows for this use case and do it inside one row, I think.

What I like about this approach is that it comes with all the paddings and hovers that a row has and I don't have to introduce too much custom stuff. 

image

Regarding custom padding, I'm not sure there's a better way to make the first and second column be as close to each other as in the design. If I'd put them in one column, I think it would be hard to align the name of the coin with "Token" and "Staking" labels, so it looks like they're in the same column, because the token icon is different size than staking and tokens icons. 

That being said, imo it would be ok to remake the rows and put them in one row, even though I don't really see how it's fragile, but I wouldn't change how the columns are done. 

What do you think?

seibei-iguchi:
> > I'm not sure this is a good approach, it seems too fragile and magical, using the padding and positioning. I'd rather not use separate rows for this use case and do it inside one row, I think.
> 
> What I like about this approach is that it comes with all the paddings and hovers that a row has and I don't have to introduce too much custom stuff.
> 
> image
> 
> Regarding custom padding, I'm not sure there's a better way to make the first and second column be as close to each other as in the design. If I'd put them in one column, I think it would be hard to align the name of the coin with "Token" and "Staking" labels, so it looks like they're in the same column, because the token icon is different size than staking and tokens icons.
> 
> That being said, imo it would be ok to remake the rows and put them in one row, even though I don't really see how it's fragile, but I wouldn't change how the columns are done.
> 
> What do you think?

I'll try to quickly come up with something in code and do a commit, we can then discuss over that:)

seibei-iguchi:
@enjojoy So I've made some changes. I've kept the separate rows, but I've cleaned it up a little and made it a little bit more robust and less magical. I'd suggest you abstract the Section component into a separate file, because the only variable part are the top and bottom properties. You can control them via a prop with values like top|middle|bottom so that it renders differently depending on where it is used (tokens, staking etc). Let me know what you think!

packages/components/src/components/Table/TableRow.tsx
```diff
@@ -79,6 +85,7 @@ export const TableRow = ({
     onClick,
     onHover,
     isHighlightedOnHover,
+    isBorderTop = true,
```
seibei-iguchi: perhaps change to hasBorderTop?

seibei-iguchi:
By the way, wouldn't it be nice if the tokens and stakings row would also link to the respective sections?

enjojoy:
> @enjojoy So I've made some changes. I've kept the separate rows, but I've cleaned it up a little and made it a little bit more robust and less magical. I'd suggest you abstract the Section component into a separate file, because the only variable part are the top and bottom properties. You can control them via a prop with values like top|middle|bottom so that it renders differently depending on where it is used (tokens, staking etc). Let me know what you think!

So when there's only one additional row, it's okay, althogh I fixed it a bit (top: calc(50% - 1px);) so the spacings between the end and the beginning of the dotted lines are consistent and it looks like one straight line. 

But when there's two rows, it looks bad either way :( 
Too little (your solution)
image


Too much (added 1px difference )
image

I think it looked good before, but you didn't like the hardcoded values right?

enjojoy:
> By the way, wouldn't it be nice if the tokens and stakings row would also link to the respective sections?

I also thought about it and I think it's a good idea. @tomasklim  What do you think?

packages/components/src/components/Table/TableRow.tsx
enjojoy: https://github.com/trezor/trezor-suite/pull/14996/commits/9f9ea451a2e3b1eadccbb362382b4508d8632267

enjojoy:
> > @enjojoy So I've made some changes. I've kept the separate rows, but I've cleaned it up a little and made it a little bit more robust and less magical. I'd suggest you abstract the Section component into a separate file, because the only variable part are the top and bottom properties. You can control them via a prop with values like top|middle|bottom so that it renders differently depending on where it is used (tokens, staking etc). Let me know what you think!
> 
> So when there's only one additional row, it's okay, althogh I fixed it a bit (top: calc(50% - 1px);) so the spacings between the end and the beginning of the dotted lines are consistent and it looks like one straight line.
> 
> But when there's two rows, it looks bad either way :( Too little (your solution) image
> 
> Too much (added 1px difference ) image
> 
> I think it looked good before, but you didn't like the hardcoded values right?

Nevermind, before it was like this too. I think the only way to fix is is to make a line with absolute position or something
image

enjojoy:
> > By the way, wouldn't it be nice if the tokens and stakings row would also link to the respective sections?
> 
> I also thought about it and I think it's a good idea. @tomasklim What do you think?

I think it would be a little tricky because when I click on the asset, it takes me to the first asset, but if staking can be on another asset, so we'd have to iterate over it and the same with tokens. So maybe open an issue for it to do it later. Because now these rows don't get the account that has the staking or the tokens

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetStakingRow.tsx
```diff
@@ -0,0 +1,64 @@
+import { isTestnet } from '@suite-common/wallet-utils';
+import { Column, Icon, Table } from '@trezor/components';
+import { spacings, borders } from '@trezor/theme';
+import { CoinBalance, HiddenPlaceholder, Translation } from 'src/components/suite';
+import styled, { css } from 'styled-components';
+import { NetworkSymbol } from '@suite-common/wallet-config';
+import { FiatValue } from 'src/components/suite';
+import { Text } from '@trezor/components';
+
+const Section = styled.div<{ $renderBothRows: boolean }>`
```
seibei-iguchi: Can you please abstract this into a separate file and use it in all three types of rows?

seibei-iguchi (changes requested):

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetStakingRow.tsx
enjojoy: https://github.com/trezor/trezor-suite/pull/14996/commits/2ebf646952b19aed134aa8613c51dd5d924c5fb1 
Thanks for a good idea

seibei-iguchi (approved):

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetRow.tsx
```diff
@@ -10,25 +10,56 @@ import {
...
+        assetTokens,
+        stakingAccounts,
+        assetsFiatBalances,
+        localCurrency,
+        currentFiatRates,
+        accounts,
+    }: AssetTableRowProps & {
+        localCurrency: FiatCurrencyCode;
+        currentFiatRates?: RatesByKey;
+    }) => {
```
tomasklim: I dont like how you reuse `AssetTableRowProps` :( it makes the code really confusing

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetTokenRow.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { AssetTableExtraRowsSection as Section } from './AssetTableExtraRowsSection';
+
+export const AssetTokenRow = ({
+    tokenIconSetWrapper,
+    network,
+    tokensDisplayFiatBalance,
+}: {
+    tokenIconSetWrapper: React.ReactNode;
+    network: AssetTableRowProps['network'];
+    tokensDisplayFiatBalance: string;
```
tomasklim, line 14: missing interface

suite-common/wallet-core/src/transactions/transactionsReducer.ts
```diff
@@ -409,6 +409,15 @@ export const selectAccountHasStaked = (state: TransactionsRootState, account: Ac
...
 };
 
+export const selectAssetAccountsThatStaked = (
+    state: TransactionsRootState,
+    accounts: Account[],
+) => {
+    const accountsThatStaked = accounts.filter(account => selectAccountHasStaked(state, account));
+
+    return accountsThatStaked;
+};
```
tomasklim, line 419:
```suggestion
    return accounts.filter(account => selectAccountHasStaked(state, account));
};
```

packages/suite/src/views/dashboard/AssetsView/AssetCard/AssetCard.tsx
```diff
@@ -122,21 +132,52 @@ export const AssetCard = ({
...
+        localCurrency,
+        symbol,
+        currentFiatRates,
+    );
+    const sortedTokens = tokensWithRates.sort(sortTokensWithRates);
+    const tokensFiatBalance = sortedTokens.reduce((total, token) => {
+        return total.plus(token?.fiatValue ?? 0);
+    }, new BigNumber(0));
+    const shouldRenderStakingRow = accountsThatStaked.length > 0 && assetStakingBalance.gt(0);
+    const shouldRenderTokenRow = tokens.shownWithBalance?.length > 0 && tokensFiatBalance.gt(0);
```
tomasklim: this code looks like a duplicate of code in `AssetRow.tsx`

Can `AssetCard` and `AssetRow` have a parent which would resolve this logic? and it will just accept some param like `displayType: 'card' | 'row'`

tomasklim: If yes, please do it in separate commit

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -669,6 +669,26 @@ export const getTokensFiatBalance = (
...
 };
 
+export const getAssetTokensFiatBalance = (
+    accounts: Account[],
+    localCurrency: FiatCurrencyCode,
+    rates?: RatesByKey,
+) => {
+    let totalBalance = new BigNumber(0);
+
+    accounts.forEach(account => {
```
tomasklim: would be nicer to use reduce

```
    accounts.reduce((total, account) => {
        const tokensBalance = getAccountTokensFiatBalance(
            account,
            localCurrency,
            rates,
            account.tokens,
        );
        return total.plus(tokensBalance ?? 0);
    }, new BigNumber(0)).toFixed();
```

if it works

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetStakingRow.tsx
```diff
@@ -0,0 +1,46 @@
+import { Column, Icon, Table, Text } from '@trezor/components';
+import { spacings } from '@trezor/theme';
+import { CoinBalance, HiddenPlaceholder, Translation, FiatValue } from 'src/components/suite';
+import { NetworkSymbol } from '@suite-common/wallet-config';
+import { AssetTableExtraRowsSection as Section } from './AssetTableExtraRowsSection';
+
+interface AssetStakingRowProps {
+    stakingTotalBalance: string;
+    symbol: NetworkSymbol;
+    renderBothRows: boolean;
```
tomasklim: `shouldRenderBothRows`

tomasklim: or maybe `shouldRenderTokenRow`? as you know that you should render staking at that place

packages/suite/src/components/wallet/TokenIconSetWrapper.tsx
```diff
@@ -3,30 +3,57 @@ import {
...
+
+    const aggregatedTokens = Object.values(
+        tokens.reduce((acc: Record<string, TokensWithRates>, token) => {
+            const { contract, balance, fiatValue } = token;
+
+            if (!acc[contract]) {
+                acc[contract] = {
+                    ...token,
+                    balance: balance ?? '0',
+                    fiatValue: fiatValue ?? { s: 0, e: 0, c: [0] },
```
tomasklim: what is s, e, c?

packages/product-components/src/components/TokenIconSet/TokenIconSet.tsx
```diff
@@ -49,9 +52,9 @@ export const TokenIconSet = ({ network, tokens }: TokenIconSetProps) => {
     return (
         <IconContainer $length={length}>
             {length > 3 && <TokenIconPlaceholder $elevation={elevation} />}
-            {visibleTokens.map(token => (
+            {visibleTokens.map((token, i) => (
```
tomasklim: not needed as duplicated cannot be there right?

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetRow.tsx
enjojoy: Understandable, I got rid of this
https://github.com/trezor/trezor-suite/pull/14996/commits/db24b70b62881dce04ca81e408a9e95506a2cc3a

packages/suite/src/views/dashboard/AssetsView/AssetCard/AssetCard.tsx
enjojoy: I feel like it doesn't make sense to create a parent component because card is an independent component and row is a part of a table, but I agree that it makes sense to take out the data handlings because they essentially do the same thing, so I extracted them in a separate function. 
https://github.com/trezor/trezor-suite/pull/14996/commits/59a5790a04f98d8d432a98d107ee7ce1b62fa2d5

suite-common/wallet-utils/src/accountUtils.ts
enjojoy: Yes with a slight change, here in a fixup
https://github.com/trezor/trezor-suite/pull/14996/commits/b2b5096161bd3f365cb8e613cc7f173d8a16d47a

packages/suite/src/views/dashboard/AssetsView/AssetTable/AssetStakingRow.tsx
enjojoy: Yes, good catch https://github.com/trezor/trezor-suite/pull/14996/commits/5a47ab41b461e17bd4f7eeafdfd91ebfc2884ae7

packages/suite/src/components/wallet/TokenIconSetWrapper.tsx
enjojoy: It is a BigNumber as an object, but probably forgot to change it to a normal BigNumber, here it is fixed https://github.com/trezor/trezor-suite/pull/14996/commits/1a6ffd52c4948eb5c17e79078aaab4088ffadd44

packages/product-components/src/components/TokenIconSet/TokenIconSet.tsx
enjojoy: Yes, not anymore https://github.com/trezor/trezor-suite/pull/14996/commits/78aaf22d419fdf3f33b44af2ded70d95baf30f2d

tomasklim (approved):

PR #14998, opened 2024-10-21
chore(libs): update git-url-parse from 13.1.0 to 15.0.0

mroz22:
used only for this right? 
image

it seems to work the same as before


martykan (approved):
We could probably get rid of this sometime

PR #14999, opened 2024-10-21
feat(suite-native): enable Solana via feature flag

### LINKED ISSUE #14895: Solana - create feature flag
BY enabling feature flag, user should be able to:

- enable Solana and discover accounts
- import Solana to Portfolio Tracker
- See SPL tokens and txns for the account

vytick:
Add Solana to mobile via feature flag and make it discoverable

## Related Issue

Resolve #14895

### DIFF
--- a/suite-common/graph/src/balanceHistoryUtils.ts
+++ b/suite-common/graph/src/balanceHistoryUtils.ts
@@ -326,6 +326,11 @@ export const getAccountHistoryMovementFromTransactions = ({
         case 'pol':
         case 'bnb':
             return getAccountHistoryMovementItemETH({ transactions, from, to });
+        case 'sol':
+            return {
+                main: [] as AccountHistoryMovementItem[],
+                tokens: {},
+            } as AccountHistoryMovement;
         default:
             coin satisfies never;
             throw new Error(`getAccountHistoryMovementItem: Unsupported network ${coin}`);
--- a/suite-common/graph/src/constants.ts
+++ b/suite-common/graph/src/constants.ts
@@ -9,6 +9,7 @@ export const LOCAL_BALANCE_HISTORY_COINS = [
     'pol',
     'bnb',
     'xrp',
+    'sol',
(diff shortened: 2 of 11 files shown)


suite-native/app/index.js
```diff
@@ -1,3 +1,4 @@
+import 'react-native-get-random-values';
```
PeKne: Could not we use AI for this? 😅

Just joking. But why is this needed? You are not importing anything specific from this package. Is it some polyfill that Solana uses in the background? If so, add some comment explaining that please

vytick: It has react native implementation of `crypto.getRandomValues()`. Otherwise Solana api wont work

PeKne: Maybe it should be placed in the metro.config.ts, then. We are already polyfilling the `crypto` module there. Cc @Nodonisko, what is your opinion about it?

Nodonisko: I don't like it, we should either use polyfill that we already have (it's only JS so it will be slow) or use `react-native-quick-crypto` which is estabilished lib a will want to use it in future anyway.

Nodonisko: If we need only that one function we already using expo-crypto in app which has getRandomValues too https://docs.expo.dev/versions/latest/sdk/crypto/#cryptogetrandomvaluesttypedarray

vytick: I just need get polyfill for crypto.getRandomValues and this one does just this one thing, has bunch of installs (900k) and is suggested by solana mobile for expo: https://docs.solanamobile.com/react-native/expo

vytick: updated in c92279d039

vytick: Fyi: I have a followup branch to get SOL data for `getAccountHistoryMovementFromTransactions`, but I dont want to make this PR too complex

PeKne (approved):

PR #15001, opened 2024-10-22
chore(suite): isArrayMember TS improvement

Lemonexe:
## Description
A new typescript util isArrayMember was added in #14830, and it solves a common issue with `Array.includes`.

:arrow_right: implement the util in suite-common & suite

:information_source: PRs for connect & suite-native will follow.


packages/type-utils/src/utils.ts
```diff
@@ -70,5 +70,5 @@ export type DefinedUnionMember<T> = T extends string ? T : never;
  */
 export const isArrayMember = <Value extends string, Subset extends Value>(
     value: Value,
-    arr: Subset[],
+    arr: readonly Subset[],
```
Lemonexe, line 73: - Problem was, that `readonly string[]` was not assignable to `string[]`.
  - for example [packages/coinjoin/tests/tools/discovery.ts](https://github.com/trezor/trezor-suite/pull/15001/files#diff-9df02e40e1f0264d59a761a011db49b9085cc301b485d86f4799cc28a6fc0f3f) (because `as const`)
- The function _should_ accept both.
- I believe this is the best way to place it
  - it does not _require_ `arr` to be `readonly`
  - it merely widens the acceptable input type
  - and the inferred generic type `Subset` is the same, because it's primitive value, so no such thing as readonly, so it doesn't change the type-narrowing behavior of `isArrayMember` in any way

suite-common/graph/src/graphDataFetching.ts
```diff
@@ -226,9 +226,7 @@ const getAccountBalanceHistory = async ({
                     accountMovementHistory.tokens?.[t.contract as TokenAddress] ?? [],
                 ] as const,
         ),
-        A.filter(
-            ([contractId]) => !tokensFilter || tokensFilter.includes(contractId as TokenAddress),
-        ),
```
Lemonexe, line 231: incidental easy fix, `isArrayMember` is not needed here :slightly_smiling_face:

suite-common/wallet-core/src/fiat-rates/fiatRatesThunks.ts
```diff
@@ -53,7 +53,7 @@ export const updateTxsFiatRatesThunk = createThunk(
         const groupedTokensTxs = groupTokensTransactionsByContractAddress(txs);
 
         for (const token in groupedTokensTxs) {
-            const hasCoinDefinitions = getNetworkFeatures(account.symbol as NetworkSymbol).includes(
+            const hasCoinDefinitions = getNetworkFeatures(account.symbol).includes(
```
Lemonexe, line 56: incidental easy fix, not related to `isArrayMember` at all, but I found it with my regex that searched for ` as ` in vicinity of `includes` :smile:

peter-sanderson (approved):
:clap:

PR #15003, opened 2024-10-22
fix(connect-popup): analytics for core in popup

### LINKED ISSUE #14983: Connect analytics calls in popup mode
Currently the AppReady event is only being sent from `connect-iframe`, not `connect-popup`. 
This means that data in popup coreMode is not reported correctly.

martykan:
## Description

Add AppReady event handling to core in popup flow.

## Related Issue

Resolve #14983

### DIFF
--- a/packages/connect-popup/src/index.tsx
+++ b/packages/connect-popup/src/index.tsx
@@ -320,6 +320,22 @@ const handleMessageInCoreMode = (
                 info: method.info,
             });
             reactEventBus.dispatch({ type: 'state-update', payload: getState() });
+
+            const { settings } = getState();
+            const transport = core.getTransportInfo();
+            analytics.report({
+                type: EventType.AppReady,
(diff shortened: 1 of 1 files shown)


packages/connect-popup/src/index.tsx
```diff
@@ -320,6 +320,22 @@ const handleMessageInCoreMode = (
                 info: method.info,
             });
             reactEventBus.dispatch({ type: 'state-update', payload: getState() });
+
+            const { settings } = getState();
+            const transport = core.getTransportInfo();
+            analytics.report({
+                type: EventType.AppReady,
```
mroz22, line 327: IFRAME.CALL sounds weird but it makes sense, there is, right now, one iframe.call per popup open, so probably good enough 🤷‍♂️

mroz22 (approved):

PR #15004, opened 2024-10-22
chore(suite-native): isArrayMember TS improvement

Lemonexe:
## Description
A new typescript util [isArrayMember](https://github.com/trezor/trezor-suite/blob/e10d590dabb68d4654ef4b91b5bf39ea80e575cc/packages/type-utils/src/utils.ts#L62-L74) was added in #14830, and it solves [a common issue](https://github.com/trezor/trezor-suite/pull/15004#discussion_r1810377227) with `Array.includes`

:arrow_right: implement the util in `suite-native` to remove some `as` occurences (no change in behavior)

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -37,6 +37,7 @@ import {
 import { DiscoveryStatus } from '@suite-common/wallet-constants';
 import { requestDeviceAccess } from '@suite-native/device-mutex';
 import { analytics, EventType } from '@suite-native/analytics';
+import { isArrayMember } from '@trezor/utils';
 
 import {
     selectDiscoveryInfo,
@@ -561,7 +562,8 @@ export const createDescriptorPreloadedDiscoveryThunk = createThunk(
                 ({ accountType }) =>
                     // Some cardano derivation may not be supported by the device. We need to exclude them from total count.
                     network.networkType !== 'cardano' ||
-                    (availableCardanoDerivations as string[]).includes(accountType),
+                    (availableCardanoDerivations !== undefined &&
+                        isArrayMember(accountType, availableCardanoDerivations)),
             );
 
             return count + availableAccountTypes.length;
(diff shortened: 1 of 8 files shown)


suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -561,7 +562,8 @@ export const createDescriptorPreloadedDiscoveryThunk = createThunk(
                 ({ accountType }) =>
                     // Some cardano derivation may not be supported by the device. We need to exclude them from total count.
                     network.networkType !== 'cardano' ||
-                    (availableCardanoDerivations as string[]).includes(accountType),
+                    (availableCardanoDerivations !== undefined &&
+                        isArrayMember(accountType, availableCardanoDerivations)),
```
Lemonexe, line 566: This was actually a potential runtime error caused by the `as`, because `availableCardanoDerivations.includes` would throw if `undefined`.
At least as far as TS is concerned, it can be undefined. Unclear if that can really happen :man_shrugging:

suite-native/tokens/src/utils.ts
```diff
@@ -11,7 +12,5 @@ export const getTokenName = (tokenName?: string) => {
 export const NETWORK_SYMBOLS_WITH_TOKENS = ['eth', 'pol', 'bnb'] satisfies Array<NetworkSymbol>;
 export type NetworkSymbolWithTokens = (typeof NETWORK_SYMBOLS_WITH_TOKENS)[number];
 
-export const isCoinWithTokens = (symbol: NetworkSymbol): symbol is NetworkSymbolWithTokens => {
-    // We typecast because TS does complain that other coins like btc are not included in NETWORKS_WITH_TOKENS which is whole point of this function to check it
```
Lemonexe, line 15: indeed, that was my motivation to create `isArrayMember` :sunglasses:

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -37,6 +37,7 @@ import {
 import { DiscoveryStatus } from '@suite-common/wallet-constants';
 import { requestDeviceAccess } from '@suite-native/device-mutex';
 import { analytics, EventType } from '@suite-native/analytics';
+import { isArrayMember } from '@trezor/type-utils';
```
Nodonisko: `type-utils` package is suposed to contain types only without runtime code. Use `@trezor/utils` instead

Lemonexe: Done in #15019

Nodonisko (approved):
lgtm

PR #15011, opened 2024-10-22
feat(suite): enable pin entry cancellation during fw update

komret:
## Description

Enable PIN entry cancellation in Suite during FW update.

To reproduce, lock your device by holding screen/button before starting the update flow.

Note that the PIN entry is cancellable only before the installation begins, not the second time PIN is requested after installation is completed. This is because the device is not acquired at that point, thus cannot be interacted with.

Resolves https://github.com/trezor/trezor-suite/pull/14833#issuecomment-2428686169

## Screenshots:
https://github.com/user-attachments/assets/4fb52538-f8df-4bd4-80ee-b7f90684c128

### DIFF
--- a/packages/connect/src/device/DeviceCommands.ts
+++ b/packages/connect/src/device/DeviceCommands.ts
@@ -373,10 +373,10 @@ export class DeviceCommands {
     }
 
     async _commonCall(type: MessageKey, msg?: DefaultPayloadMessage['message']) {
-        const resp = await this.call(type, msg);
         if (this.disposed) {
             throw ERRORS.TypedError('Runtime', 'typedCall: DeviceCommands already disposed');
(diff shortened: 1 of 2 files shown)


mroz22:
@marekrjpolak  please investigate 🙏

marekrjpolak (approved):
@komret I've added the smallest & safest fix possible, so there should be displayed `Error: Cancelled` as discussed.

mroz22 (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 24.12.0 (19aeb1a0f7d306fcbe7cc61ed330a2481a3d0e4e)
 - OS: MacIntel
 - Screen: 1512x982

PR #15012, opened 2024-10-22
fix(suite): tighten condition to set uiEvent for firmware update

komret:
## Description

See the comment in code for explanation.

How to reproduce the bug:
1. Use TS5 with PIN.
2. Click _Change PIN_.
3. Cancel on device while entering PIN.
4. Open update modal - see unexpected confirmation pill on top.

## Screenshots:
Before:

https://github.com/user-attachments/assets/4caf9d38-297d-4396-8519-f3a1f167bd72

### DIFF
--- a/suite-common/wallet-core/src/firmware/firmwareReducer.ts
+++ b/suite-common/wallet-core/src/firmware/firmwareReducer.ts
@@ -96,7 +96,10 @@ export const prepareFirmwareReducer = createReducerWithExtraDeps(initialState, (
                 action.type === UI.FIRMWARE_PROGRESS ||
                 action.type === DEVICE.BUTTON,
             (state, action) => {
-                state.uiEvent = action;
+                // DEVICE.BUTTON can be dispatched outside of the firmware update flow and that should not change the uiEvent,
+                // otherwise it could result in confirmation pill being displayed unintentionally.
+                if (!(action.type === DEVICE.BUTTON && state.status === 'initial'))
+                    state.uiEvent = action;
             },
         );
 });


suite-common/wallet-core/src/firmware/firmwareReducer.ts
```diff
@@ -96,7 +96,10 @@ export const prepareFirmwareReducer = createReducerWithExtraDeps(initialState, (
                 action.type === UI.FIRMWARE_PROGRESS ||
                 action.type === DEVICE.BUTTON,
             (state, action) => {
-                state.uiEvent = action;
+                // DEVICE.BUTTON can be dispatched outside of the firmware update flow and that should not change the uiEvent,
+                // otherwise it could result in confirmation pill being displayed unintentionally.
+                if (!(action.type === DEVICE.BUTTON && state.status === 'initial'))
```
Lemonexe, line 101: I am no expert on redux toolkit, but I think the condition belongs to the matcher function.

So L97 would be `(action.type === DEVICE.BUTTON && state.status !== 'initial')`

Lemonexe (commented):
1. I couldn't reproduce the bug. Please confirm that it really happens, and is fixed by this change, but if you are, I'll approve.
2. But if it works, then another question is how it works? It seems to me more appropriate to fix the condition in `useFirmware`, which determines `confirmOnDevice`. Just IMHO it's easier to understand the code, than if it is here. But idk, it's just at first glance

suite-common/wallet-core/src/firmware/firmwareReducer.ts
komret: That would indeed make more sense semantically, but I don't think I can access `state` in the matcher function.

komret:
> 1. I couldn't reproduce the bug. Please confirm that it really happens, and is fixed by this change, but if you are, I'll approve.

Sorry, I edited the reproduction steps. In step 3, it has to be cancelled during PIN entry, not before.

> 2. But if it works, then another question is how it works? It seems to me more appropriate to fix the condition in `useFirmware`, which determines `confirmOnDevice`. Just IMHO it's easier to understand the code, than if it is here. But idk, it's just at first glance

I think it is better fixed here, because this is the source: the state is changing while it should not, firmware reducer is affected by actions unrelated to firmware.

suite-common/wallet-core/src/firmware/firmwareReducer.ts
Lemonexe: Right, that makes sense!

Lemonexe (approved):
LGTM, works well

> I think it is better fixed here, because this is the source: the state is changing while it should not, firmware reducer is affected by actions unrelated to firmware.

Sounds good, it was only a suggestion. It wasn't clear to me before, but now I see that `state.firmware.uiEvent` is something that's related _only to fw update_, and no other fw interaction (as I assumed before, and was worried about side effects)

bosomt:
**QA OK**

**Info:**
 - Suite version: web 24.11.0 (05d128a974d81cb841314a86ed3c97e11cdb6ec4)
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.4 regular (revision 31d228a325b5968a2f3565b803818aaf7fa29bbb)

PR #15016, opened 2024-10-22
fix(suite-native): prevent device actions cutoff by removing explicit height

matejkriz:
## Description

This issue was not present in production; it was introduced in the development branch at https://github.com/trezor/trezor-suite/pull/14963. The issue was visible only with non-standard system font scaling.

With ACCESSIBILITY_FONTSIZE_MULTIPLIER == 1, everything worked fine, but with a lower value, the text was cut off.

The fixed height doesn't seems to be necessary. Tested with different font scales on both android and ios...

## Screenshots:

BEFORE:
image

AFTER:
image

### DIFF
--- a/suite-native/device-manager/src/components/DeviceAction.tsx
+++ b/suite-native/device-manager/src/components/DeviceAction.tsx
@@ -1,7 +1,7 @@
 import { ReactNode } from 'react';
 import { Pressable } from 'react-native';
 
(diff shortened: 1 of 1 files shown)


vytick (approved):
tested in emulator, looks good 🚀

STew790:
**QA OK**
LGTM on Android and iOS

**Info**
24.11.1 https://github.com/trezor/trezor-suite/commit/c214aa306ebf33a0255973753fb55b487dcdc6db

PR #15019, opened 2024-10-22
chore(utils): move isArrayMember to utils

Lemonexe:
## Description

I created a new utility in `type-utils` in #14830, but I was [told afterwards](https://github.com/trezor/trezor-suite/pull/15004#discussion_r1810489291) that it should rather be in `utils`

- so move `isArrayMember` to utils
- write test, which I omitted before :see_no_evil: 

Partial revert of #15001, where `type-utils` were newly introduced into some packages.


marekrjpolak (approved):
:seal: :heavy_check_mark:

PR #15022, opened 2024-10-23
Add IconCircle component

seibei-iguchi:
- added IconCircle component
- added negative spacings
- refactored Suite to use IconCircle component and removed style overrides

### Before
Screenshot 2024-10-23 at 10 13 43
Screenshot 2024-10-23 at 10 13 54
Screenshot 2024-10-23 at 10 14 14

### After
Screenshot 2024-10-23 at 10 10 34
Screenshot 2024-10-23 at 10 10 44
Screenshot 2024-10-23 at 10 11 04


packages/theme/src/spacings.ts
```diff
@@ -12,21 +12,35 @@ export const spacings = {
...
-export const spacingsPx = (Object.keys(spacings) as Array<SpacingSize>).reduce(
+export const spacingsPx = (Object.keys(spacings) as Array<Spacing>).reduce(
     (result, key) => {
-        (result as Record<SpacingSize, string>)[key] = `${spacings[key]}px`;
+        (result as Record<Spacing, string>)[key] = `${spacings[key]}px`;
 
         return result;
     },
-    {} as { [K in SpacingSize]: `${(typeof spacings)[K]}px` },
+    {} as { [K in Spacing]: `${(typeof spacings)[K]}px` },
```
peter-sanderson: For simplicity I would suggest:

1. export the ```{ [K in Spacing]: `${(typeof spacings)[K]}px` }``` and name it
2. the type can be simplified to: ```{ [K in Spacing]: `${Spacings[K]}px` }```

```ts
type SpacingPx = { [K in Spacing]: `${Spacings[K]}px` };
```

packages/components/src/components/IconCircle/IconCircle.tsx
```diff
@@ -0,0 +1,73 @@
...
+            $hasBorder={hasBorder}
+            {...wrapperColorOrVariant}
+            {...frameProps}
+        >
+            <Icon name={name} size={size} {...iconColorOrVariant} />
+        </IconCircleWrapper>
+    );
+};
+
+export type { IconCircleVariant, IconCircleColors };
```
peter-sanderson, line 74: why this? can we just export it directly?

packages/components/src/components/IconCircle/IconCircle.tsx
```diff
@@ -0,0 +1,73 @@
...
+    FramePropsKeys,
+    pickAndPrepareFrameProps,
+    withFrameProps,
+} from '../../utils/frameProps';
+
+export const allowedIconCircleFrameProps = ['margin'] as const satisfies FramePropsKeys[];
+type AllowedFrameProps = Pick<FrameProps, (typeof allowedIconCircleFrameProps)[number]>;
+
+type IconCircleWrapperProps = TransientProps<IconCircleExclusiveColorOrVariant> &
+    TransientProps<AllowedFrameProps> & {
```
peter-sanderson: Can be simplified to:

```ts
type IconCircleWrapperProps = TransientProps<
    IconCircleExclusiveColorOrVariant & AllowedFrameProps
> & {
    $size: number;
    $hasBorder: boolean;
};
```

packages/components/src/components/IconCircle/IconCircle.tsx
seibei-iguchi: To simplify the exports in index.ts, I guess

packages/components/src/components/IconCircle/utils.tsx
```diff
@@ -0,0 +1,55 @@
...
+        1: {
+            primary: 'backgroundPrimarySubtleOnElevation2',
+            warning: 'backgroundAlertYellowSubtleOnElevation2',
+            destructive: 'backgroundAlertRedSubtleOnElevation2',
+            info: 'backgroundAlertBlueSubtleOnElevation2',
+            tertiary: 'backgroundTertiaryDefaultOnElevation1',
+        },
+    };
+
+    return theme[colorMap[Number($hasBorder)][$variant]];
```
peter-sanderson: With `Number($hasBorder)` the benefit of having a map is absent. Here I would just do this:

constant outside of the function scope
```ts

const noBorderColorMap: Record<IconCircleVariant, Color>> = {....}

const borderColorMap: Record<IconCircleVariant, Color>> = {....}
```

and then 

```ts
const map = $hasBorder ? borderColorMap : noBorderColorMap;

return theme[map][$variant]];
```

peter-sanderson: If you want to preserve the 2D map, it would need to convert `$hasBorder,` into `$hasBorder: "border" | "no-border"` so you don't need to cast type (with  `Number()`)

peter-sanderson: Maybe also 

```ts
return theme[colorMap[$hasBorder ? "border" : "no-border"][$variant]];
```

would be an option

packages/components/src/components/NewModal/NewModal.tsx
```diff
@@ -6,29 +6,36 @@ import {
...
+    IconButton,
+    IconCircle,
+    IconName,
+    H3,
+    Text,
+    useScrollShadow,
+    useElevation,
+    ElevationContext,
+    ElevationUp,
+} from '../../index';
```
peter-sanderson: I am not sure about this. I feel that inside of the package we should not import from index files. 

I see those as API of the package for the outside use.  Internally, package should be working without it.

I have no strong argument for it, it just feels right. Is there any technical benefit of one or the other?

packages/components/src/components/IconCircle/IconCircle.tsx
peter-sanderson: Oh I see, its a re-export.

packages/components/src/components/NewModal/NewModal.tsx
seibei-iguchi: I don't have a strong opinion, it just felt cleaner:)

peter-sanderson: I also have a feeling back in my mind, that this can cause a circular dependecny problem, but I don't know exact conditions.

packages/suite/src/components/wallet/AccountExceptionLayout.tsx
```diff
@@ -1,86 +1,51 @@
...
-`;
-
-const Actions = styled.div`
-    display: flex;
-    justify-content: center;
-    width: 100%;
-    margin-bottom: 20px;
-`;
-
-// eslint-disable-next-line local-rules/no-override-ds-component
```
peter-sanderson, line 43: :love_you_gesture:

adderpositive (approved):
LGTM. Is it possible to delete this empty element? It leaves a lot of empty white space.

obrazek

peter-sanderson (approved):

PR #15027, opened 2024-10-23
fix(mobile): icons fixes

Nodonisko:
## Description

Token icons have be generated manually as before.

## Related Issue

Resolve


yanascz (approved):
It seems to work fine at first glance. 👍

PR #15028, opened 2024-10-23
chore(connect): isArrayMember TS improvement

Lemonexe:
## Description

In #15019, `isArrayMember` was created in `@trezor/utils`, and implemented in suite.

:arrow_right: Implement it in `connect` to remove some `as` occurences (no change in behavior)


marekrjpolak (approved):
:shipit:

PR #15031, opened 2024-10-23
feat(suite): expand FW hash check on all models (v2)

Lemonexe:
Followup to #14766 where FW hash check was implemented for T1B1.

## Description

- a8ba1e8d50c58808fb002a38ccd93ba43a597ddf Reapply [#15018](https://github.com/trezor/trezor-suite/pull/15018): Expand the FW hash check on all models
- be5b44056c1a1f851830c0db9fcf4c72e78f4bff Fix Playwright tests
  - since playwright is concerned with desktop-specific app interface, and not so much the UI of react app itself, I think it is all right to take the easier path:
  - just dismiss the FW hash check modal via "Back" button
  - This will leave the error banner, but that doesn't interfere with scope of playwright tests
- 4182592e5203ea1066505a85d3e529d8fd8fdf69 Fix Cypress tests by using the already existing `cy.disableFirmwareHashCheck()`
  - most tests don't need it, as they use [passThroughInitialRun](https://github.com/trezor/trezor-suite/blob/77cd90627c60241c55fe16b42d46898dc37834de/packages/suite-web/e2e/support/utils/shortcuts.ts#L12-L13) that takes care of it
  - but some tests do not use passThroughInitialRun, since they do something customized instead
  - so it's a bit of boilerplate, but IMHO it's analogous to having to confirm analytics consent in all those tests :shrug: 
  - BTW unlike playwright, in Cypress there are some tests that would be broken by the warning banner, if the modal is just dismissed: such as testing for other SuiteBanners, or receiving addresses etc.

## Related Issue

resolves https://github.com/trezor/trezor-suite-private/issues/114

### DIFF
--- a/packages/suite-desktop-core/e2e/support/pageActions/dashboardActions.ts
+++ b/packages/suite-desktop-core/e2e/support/pageActions/dashboardActions.ts
@@ -5,8 +5,16 @@ import { NetworkSymbol } from '@suite-common/wallet-config';
 import { waitForDataTestSelector } from '../common';
 
 class DashboardActions {
+    optionallyDismissFwHashCheckError(window: Page) {
+        // dismiss the error modal only if it appears (handle it async in parallel, not necessary to block the rest of the flow)
+        window
+            .$('[data-testid="@device-compromised/back-button"]')
+            .then(dismissFwHashCheckButton => dismissFwHashCheckButton?.click());
+    }
+
     async passThroughInitialRun(window: Page) {
         await waitForDataTestSelector(window, '@welcome/title');
+        this.optionallyDismissFwHashCheckError(window);
         await window.getByTestId('@analytics/continue-button').click();
         await window.getByTestId('@onboarding/exit-app-button').click();
 
--- a/packages/suite-web/e2e/tests/analytics/events.test.ts
+++ b/packages/suite-web/e2e/tests/analytics/events.test.ts
@@ -20,6 +20,7 @@ describe('Analytics Events', () => {
 
     it('reports transport-type, suite-ready and device-connect/device-disconnect events when analytics is initialized and enabled', () => {
         cy.prefixedVisit('/');
+        cy.disableFirmwareHashCheck();
 
         // go to settings and enable analytics (makes analytics enabled and initialized)
         onNavBar.openSettings();
@@ -96,6 +97,10 @@ describe('Analytics Events', () => {
         cy.interceptDataTrezorIo(requests);
 
         cy.prefixedVisit('/', forceLightMode);
+        cy.disableFirmwareHashCheck();
+        // The next step is to press Settings Icon, and that can be done from DeviceCompromised modal as well as from a normal scren.
+        // But it can happen that Cypress tries to click it immediately, but fails, because the app is rerendering right now (modal is disappearing)
+        cy.wait(500);
 
         // change few settings to see if it is different from default values in suite-ready
         onNavBar.openSettings();
--- a/packages/suite-web/e2e/tests/analytics/toggle.test.ts
+++ b/packages/suite-web/e2e/tests/analytics/toggle.test.ts
@@ -18,13 +18,13 @@ describe('Analytics Toggle - Enablement and Disablement', () => {
         cy.viewport(1440, 2560).resetDb();
 
         requests = [];
-    });
-
-    it('should respect disabled analytics in onboarding with following enabling in settings', () => {
         cy.interceptDataTrezorIo(requests).as('data-fetch');
 
         cy.prefixedVisit('/');
+        cy.disableFirmwareHashCheck();
+    });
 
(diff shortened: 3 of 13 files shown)


packages/suite-desktop-core/e2e/support/pageActions/dashboardActions.ts
```diff
@@ -5,8 +5,16 @@ import { NetworkSymbol } from '@suite-common/wallet-config';
 import { waitForDataTestSelector } from '../common';
 
 class DashboardActions {
+    optionallyDismissFwHashCheckError(window: Page) {
+        // dismiss the error modal only if it appears (handle it async in parallel, not necessary to block the rest of the flow)
+        window
+            .$('[data-testid="@device-compromised/back-button"]')
```
Lemonexe, line 11: `window.getByTestId` throws if not found, while `window.$` just returns `null`

mroz22 (approved):
acked connect part, cc @HajekOndrej for tests part

packages/suite-desktop-core/e2e/support/pageActions/dashboardActions.ts
HajekOndrej: Is this proper handling? I am no JS expert, so I don't know, but it looks, that you may sometimes call .then() on a null.

packages/suite-web/e2e/tests/analytics/events.test.ts
```diff
@@ -134,6 +135,7 @@ describe('Analytics Events', () => {
         // nothing should be reported until analytics is initialized and enabled
         cy.wrap(requests).should('have.length', 0);
 
+        cy.disableFirmwareHashCheck();
```
HajekOndrej: Maybe it would be better to just disable it at the start rather than in the middle of the test? Unless it's necessary for the test, I think it kinda breaks the flow of arrange/act/assert (or given/when/then). 

If that is possible, then I think we can move the prefixedVisit and this straight to the beforeEach.

packages/suite-web/e2e/tests/analytics/toggle.test.ts
```diff
@@ -24,6 +24,7 @@ describe('Analytics Toggle - Enablement and Disablement', () => {
         cy.interceptDataTrezorIo(requests).as('data-fetch');
 
         cy.prefixedVisit('/');
+        cy.disableFirmwareHashCheck();
```
HajekOndrej, line 24: Similarly to the previous file, we could refactor 
cy.interceptDataTrezorIo(requests).as('data-fetch');
cy.prefixedVisit('/');
cy.disableFirmwareHashCheck();

to beforeEach if I am not mistaken

packages/suite-web/e2e/tests/analytics/events.test.ts
Lemonexe: Oh yes, good call! At first I placed it at beginning, but the test was hanging because of this, and moving the procedure fixed it, but I'm really puzzled why :confused: 
[This is the run](https://github.com/trezor/trezor-suite/actions/runs/11479515554/job/31948847103), and [here is artifact (screenshot)](https://github.com/user-attachments/assets/2ac2f7ac-8e62-4797-9127-b00f15f6fa1f).
EDIT: I think I figured it out, I'll try something.

packages/suite-desktop-core/e2e/support/pageActions/dashboardActions.ts
Lemonexe: that's why I added `?.click()` :slightly_smiling_face: 
It will either call it if truthy, or ignore if null

HajekOndrej: Yeah, that's not what I meant, I thought that the .$ would return null, but it returns a promise whose value can be null. Everything okay! :)

packages/suite-web/e2e/tests/analytics/toggle.test.ts
Lemonexe: fixed in 9b8bd8bdd2d014b3d78dc61d1ff0d0e79e5dec86

packages/suite-web/e2e/tests/analytics/events.test.ts
Lemonexe: Oh yeah, see comment in 9b8bd8bdd2d014b3d78dc61d1ff0d0e79e5dec86
I am not a fan of using `cy.wait` to randomly fix things, but in this case I think I proved the reason why exactly it's necessary.

EDIT: As for the `beforeEach`, the second part of test starts bridge, while the first part is running without bridge. I guess to test that some settings can be modified even without it? IDK. But I think the `beforeEach` can't be practically extracted because of the difference.

HajekOndrej: I see. It makes more sense like this anyway, thanks.

HajekOndrej (approved):

PR #15034, opened 2024-10-23
feat(connect): display descriptive info when firmware cant serve current method

### LINKED ISSUE #15032: Unknown message error with Ethereum and Bitcoin-only firmware
**Describe the bug**

When using Ethereum with Bitcoin-only firmware, I'm getting "Unknown message" error with all Ethereum methods.

Screenshot 2024-10-23 at 11 44 46

**Info:**
- Transport: both webUSB and Bridge 3.0.0
- Firmware: 1.12.1
- Connect: 9.3.0 - 9.4.2

**Expected behaviour:**

Change message to notify user that he is using Bitcoin-only firmware and needs to switch to Universal firmware.

cc. @mroz22

mroz22:
very dirty draft, discussion is going on in https://satoshilabs.slack.com/archives/C02V2PSDNA2/p1729692326462089?thread_ts=1729676861.081469&cid=C02V2PSDNA2
resolve #15032 
image

### DIFF
--- a/packages/connect/src/core/AbstractMethod.ts
+++ b/packages/connect/src/core/AbstractMethod.ts
@@ -16,6 +16,7 @@ import {
 import { getHost } from '../utils/urlUtils';
 import type { Device } from '../device/Device';
 import type { FirmwareRange, DeviceState, StaticSessionId, DeviceUniquePath } from '../types';
+import { Capability } from '@trezor/protobuf/src/messages';
 
 export type Payload<M> = Extract<CallMethodPayload, { method: M }> & { override?: boolean };
 export type MethodReturnType<M extends CallMethodPayload['method']> = CallMethodResponse<M>;
@@ -132,6 +133,8 @@ export abstract class AbstractMethod<Name extends CallMethodPayload['method'], P
     allowDeviceMode: DeviceMode[]; // used in device management (like ResetDevice allow !UI.INITIALIZED)
 
     requireDeviceMode: DeviceMode[];
+
+    requiredDeviceCapabilities: Capability[] = [];
 
     network: NETWORK.NetworkType;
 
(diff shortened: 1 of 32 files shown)


Hannsek:
What about these alternatives:
image

martykan:
Regarding the error UX this is what I have for now, which I feel is good enough. 
Learn more leads you to https://trezor.io/learn/a/bitcoin-only-firmware-on-trezor

Screenshot 2024-11-01 at 11 45 31

To implement a direct link to FW install in Suite would have some caveats, such as if we want to open desktop/web Suite, if desktop we need to add a deeplink.

Hannsek:
Yes, this is good enough. Thanks!

packages/connect/src/core/AbstractMethod.ts
```diff
@@ -333,6 +336,24 @@ export abstract class AbstractMethod<Name extends CallMethodPayload['method'], P
...
 
+    checkDeviceCapability() {
+        const deviceHasAllRequiredCapabilities = (this.requiredDeviceCapabilities || []).every(
+            capability => this.device.features.capabilities.includes(capability),
+        );
+        if (!deviceHasAllRequiredCapabilities) {
+            if (this.device.firmwareType === 'bitcoin-only') {
+                throw ERRORS.TypedError(
+                    'Device_MissingCapabilityBtcOnly',
+                    `Trezor has Bitcoin-only firmware installed, which does not support the ${this.info} method. Please install Universal firmware through Trezor Suite.`,
```
mroz22: this.info may sometimes yield quite variable results 🤔 just thinking whether it is a good idea to intertwine it together here

packages/connect/src/core/index.ts
```diff
@@ -428,6 +428,8 @@ const inner = async (context: CoreContext, method: AbstractMethod<any>, device:
         sendCoreMessage(createPopupMessage(POPUP.CANCEL_POPUP_REQUEST));
     }
 
+    method.checkDeviceCapability();
```
mroz22: I am just thinking what is the line to place this check. There are other similar checks, for example this one https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/core/index.ts#L214
maybe @marekrjpolak has a stronger opinion?

mroz22 (commented):
looks good to me, left two comments, don't have the right to aprove

packages/connect/src/core/AbstractMethod.ts
martykan: Yeah we could just write "does not support this method" or maybe "does not support this operation", since method may sound strange for less technical people

martykan (approved):

mroz22:
maybe put it ahead of permissions check? c4f0ef90da8896b94a3caea5c943535415f58d4c

martykan:
Sure, looks like it belongs with the earlier checks

PR #15041, opened 2024-10-23
improve connect test matrix

mroz22:
-  update legacy results for cardano.
- `connect-test-matrix-generator.js` reworked now it:
  -   creates an array of all possible combinations of test modifiers (`'model', 'firmware', 'env', 'groups', 'cache_tx', 'transport'`). 
  -  accepts args in the format `--someStrArg=meow` or `--someArrOfStrArg=wuff,meow` and uses it to filter the array produced in the previous step. This way we only receive combinations we want to use to create the desired jobs matrix. 
  - special keyword `all` might be passed to allow all of the possibilities, for example: `--env=all` means the same as `--env=node,web`
  - some of  the inputs are not actively used yet. I created this mainly because of @szymonlesisz request to be able to extend test parametrization by specifying `transport`. this is now possible but the newly added param is not consumed anywhere yet.
- connect tests jobs slightly reorganized.
  - As we add more tests, we need to find a way to run them more efficiently and target them more precisely. So from now on in regular PRs, only the `api` section of connect is tested (so no `methods`). Methods (eth, nem, ada..) are the most time-consuming and at the same time, they don't break that often so they are now run only in nightly jobs. There will be a follow-up that will create new packages for all the shitcoin method families that will allow us to run them in regular PR pipelines again.
  - anyway, you can check yourself how the [full testing strategy including nightly looks like](https://github.com/trezor/trezor-suite/actions/runs/11549791189)

just a few screenshots to make the PR more visually appealing

image

image

what I don't understand is why github sometimes wraps two jobs under one dropdown and sometimes not, see this screenshot 
image

### DIFF
--- a/scripts/ci/connect-test-matrix-generator.js
+++ b/scripts/ci/connect-test-matrix-generator.js
@@ -1,140 +1,227 @@
 const process = require('process');
 
+const DEBUG = false;
+
+const log = (...args) => {
+    if (DEBUG) {
+        return console.log(...args);
+    }
+};
+
 const groups = {
     api: {
         name: 'api',
         pattern:
-            'init authorizeCoinjoin cancelCoinjoinAuthorization passphrase unlockPath setBusy override checkFirmwareAuthenticity keepSession cancel.test info.test',
-        methods: '',
+            'init authorizeCoinjoin cancelCoinjoinAuthorization passphrase unlockPath setBusy checkFirmwareAuthenticity keepSession cancel.test info.test',
+        includeFilter: '',
+    },
+    // temporarily created group for flaky test - to spend less time on reruns and to make test result in CI more readable without investigating long logs
+    apiFlaky: {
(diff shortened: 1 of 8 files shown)


martykan (approved):
LGTM

scripts/ci/connect-test-matrix-generator.js
```diff
@@ -1,140 +1,227 @@
...
-const legacyFirmware = {
-    firmwares: ['2.3.0'],
-    tests: daily.tests
-        // Cardano supports >=2.6.0
-        .filter(test => test.name !== 'cardano'),
-};
+const inputs = [
+    {
+        key: 'model',
+        value: ['T1B1', 'T2T1', 'T2B1', 'T3B1', 'T3T1'],
```
mroz22, line 90: add here to run model 7 cc @komret

Issue #14445, opened 2024-09-19
networksConfig refactoring afterthoughts


Lemonexe:
Followup to https://github.com/trezor/trezor-suite/issues/13839

Further refactoring of `networksConfig` to reduce information, and make property names clearer.

```[tasklist]
### Tasks
- [ ] https://github.com/trezor/trezor-suite/issues/16035
- [ ] https://github.com/trezor/trezor-suite/issues/16034
- [x] https://github.com/trezor/trezor-suite/issues/16271
```

PR #15043, opened 2024-10-24
Chore: bip43Path template type

### LINKED ISSUE #16271: define `Bip43Path` as a string template literal type rather than just `string`

Lemonexe:
## Description

- Define `Bip43Path` as a string template literal type rather than just `string`, to ensure that we don't accidentally produce a non-conforming path.
  - lot of TS errors had to be fixed, as we often had `string` directly, and _so far_ that was interchangeable with `Bip43Path`, but now it's not, it has to be used everywhere
  - but still there are interfaces with `connect` were the path is coming in as `string`
    - `suite-common` imports are not allowed in connect
    - :arrow_right: future TODO: move the path types to another package, for example `crypto-utils`..?  
- Create a helper `substituteBip43Path` to centralize substitution of index for `'i'` placeholder
  - it requires `as`, because `String.replace` returns `string` :disappointed: 
  - cover with unit test

## Related Issue

Part of #14445 
Resolves #16271

### DIFF
--- a/suite-common/wallet-config/src/types.ts
+++ b/suite-common/wallet-config/src/types.ts
@@ -57,10 +57,14 @@ export type NetworkFeature =
     | 'nft-definitions';
 
 type Level = `/${number}'`;
-type LevelOrIndex = `/${number | 'i'}'`;
 type MaybeApostrophe = `'` | '';
+type MaybeLevel = `/${number}${MaybeApostrophe}` | '';
+type LevelOrIndex = `/${number | 'i'}'`;
 type MaybeLevelOrIndex = `/${number | 'i'}${MaybeApostrophe}` | '';
-export type Bip43Path = `m${Level}${Level}${LevelOrIndex}${MaybeLevelOrIndex}${MaybeLevelOrIndex}`;
+// template with i in place of account index, which shall be substituted with a number
+export type Bip43PathTemplate =
+    `m${Level}${Level}${LevelOrIndex}${MaybeLevelOrIndex}${MaybeLevelOrIndex}`;
+export type Bip43Path = `m${Level}${Level}${Level}${MaybeLevel}${MaybeLevel}`;
 
 export type Explorer = {
     tx: string;
@@ -73,7 +77,7 @@ export type Explorer = {
 
 type NetworkAccountWithSpecificKey<TKey extends AccountType> = {
     accountType: TKey;
-    bip43Path: Bip43Path;
+    bip43Path: Bip43PathTemplate;
     backendType?: BackendType;
     features?: NetworkFeature[];
     isDebugOnlyAccountType?: boolean;
@@ -91,7 +95,7 @@ type NetworkWithSpecificKey<TKey extends NetworkSymbol> = {
     symbol: TKey;
     name: string;
     networkType: NetworkType;
-    bip43Path: Bip43Path;
+    bip43Path: Bip43PathTemplate;
     decimals: number;
     testnet: boolean;
     explorer: Explorer;
--- a/suite-common/wallet-config/src/types.ts
+++ b/suite-common/wallet-config/src/types.ts
@@ -56,7 +56,11 @@ export type NetworkFeature =
     | 'coin-definitions'
     | 'nft-definitions';
 
-export type Bip43Path = string; // TODO define a more specific type
+type Level = `/${number}'`;
+type LevelOrIndex = `/${number | 'i'}'`;
+type MaybeApostrophe = `'` | '';
+type MaybeLevelOrIndex = `/${number | 'i'}${MaybeApostrophe}` | '';
+export type Bip43Path = `m${Level}${Level}${LevelOrIndex}${MaybeLevelOrIndex}${MaybeLevelOrIndex}`;
 
 export type Explorer = {
     tx: string;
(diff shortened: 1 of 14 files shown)


suite-common/wallet-config/src/types.ts
```diff
@@ -56,7 +56,11 @@ export type NetworkFeature =
     | 'coin-definitions'
     | 'nft-definitions';
 
-export type Bip43Path = string; // TODO define a more specific type
+type Level = `/${number}'`;
+type LevelOrIndex = `/${number | 'i'}'`;
+type MaybeApostrophe = `'` | '';
+type MaybeLevelOrIndex = `/${number | 'i'}${MaybeApostrophe}` | '';
+export type Bip43Path = `m${Level}${Level}${LevelOrIndex}${MaybeLevelOrIndex}${MaybeLevelOrIndex}`;
```
Lemonexe: Fun fact: [BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) prescribes that first three fields are always hardened (have the `'`), and only the 4th & 5th have it optional. But at first I tried making it optional everywhere:
```
type MaybeApostrophe = `'` | '';
type Level = `/${number}${MaybeApostrophe}`;
type LevelOrIndex = `/${number | 'i'}${MaybeApostrophe}`;
type MaybeLevelOrIndex = LevelOrIndex | '';
```
But it's too many combinations :skull: TS itself still works, but IDE integration does not take it well when error message has cca 120 kB :upside_down_face: 
Restricting the `'` reduces combinations, so TS error message has only 2 kB :relieved: 
This is probably the first time in my life I _optimized TypeScript performance_ :rofl: But it's also correct as per BIP44..


:thought_balloon:  also btw we call it Bip43Path, despite the fact that [BIP43](https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki) defines only the purpose field,
while the path as a whole is defined by BIP44, but I guess we'll leave it at that :sweat_smile: :see_no_evil:

suite-native/module-accounts-import/src/accountsImportThunks.ts
```diff
@@ -59,7 +59,7 @@ export const importAccountThunk = createThunk(
                     deviceState,
                     discoveryItem: {
                         index: deviceNetworkAccounts.length, // indexed from 0
-                        path: accountInfo?.path ?? '',
+                        path: (accountInfo?.path ?? '') as Bip43Path,
```
Lemonexe, line 62: unclear if `''` can happen, and how this `path` is consumed, but hey, the code was already like that; let's not fix what is probably not broken :shrug:

suite-common/wallet-core/src/accounts/tests/accountsReducer.test.ts
```diff
@@ -30,6 +30,8 @@ const getAccount = (a?: Partial<Account>) => ({
     ...a,
 });
 
+const testBip43Path = "m/84'/0'/0'";
```
peter-sanderson:
```suggestion
const testBip43Path: Bip43Path = "m/84'/0'/0'";
```

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
```diff
@@ -88,6 +89,14 @@ describe('account utils', () => {
         });
     });
 
+    describe(substituteBip43Path.name, () => {
+        it("substitutes 'i' placeholder in path with index", () => {
+            expect(substituteBip43Path("m/84'/0'/i'", 7)).toEqual("m/84'/0'/7'");
+            expect(substituteBip43Path("m/44'/0'/i'/0", '4')).toEqual("m/44'/0'/4'/0");
+            expect(substituteBip43Path("m/10025'/1'/i'/1'")).toEqual("m/10025'/1'/0'/1'");
```
peter-sanderson: nit: this is great case to use dataProviders/fixtures to reduce boilerplate

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -223,6 +223,9 @@ export const getBip43Type = (path: string) => {
     }
 };
 
+export const substituteBip43Path = (bip43Path: Bip43Path, accountIndex: string | number = '0') =>
```
peter-sanderson: maybe we hall separate `Bip43PathTemplate` (with `i`) from  `Bip43Path` 

This function would be then type-mapping from Bip43PathTemplate -> Bip43Path

suite-common/wallet-core/src/accounts/tests/accountsReducer.test.ts
Lemonexe: ba3e0dfbee9483849d071aec3a4406618818b6e3

suite-common/wallet-utils/src/__tests__/accountUtils.test.ts
Lemonexe: ba3e0dfbee9483849d071aec3a4406618818b6e3

suite-common/wallet-utils/src/accountUtils.ts
Lemonexe: 5b7d207d46ef6cbf6e29ceb179ef451acbf653b3 that was a little bit more work, as I had to dissect the current usages of `Bip43Path`, to determine at each place if we are dealing with template, final path, or both (in case of `getAccountTypeName`)

peter-sanderson (approved):

PR #15044, opened 2024-10-24
test(suite-web): remove temporary fixes for fw revision check

Lemonexe:
## Description

Finish or remove temporary workarounds for FW _revision_ check in suite-web E2E tests (cypress)

### DIFF
--- a/packages/suite-web/e2e/tests/onboarding/t2t1-create-wallet.test.ts
+++ b/packages/suite-web/e2e/tests/onboarding/t2t1-create-wallet.test.ts
@@ -6,8 +6,6 @@ describe('Onboarding - create wallet', () => {
         cy.task('startBridge');
         cy.viewport(1920, 1080).resetDb();
         cy.prefixedVisit('/');
-
-        // TODO: add workaround that switches off the firmware revision check in settings/device
     });
 
     it('Success (Shamir backup)', () => {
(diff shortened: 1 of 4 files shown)


packages/suite-web/e2e/tests/onboarding/t2t1-create-wallet.test.ts
```diff
@@ -16,8 +14,8 @@ describe('Onboarding - create wallet', () => {
         // in production, and we locally bumped emulator version, we would get into a screen saying "update your firmware" and the test would fail.
         cy.task('startEmu', { wipe: true, model: 'T2T1', version: '2-main' });
 
-        // TODO: compromised device workaround, refactor into more stable solution
-        cy.contains('Back').click();
+        // firmware revision check will fail with 2-main version, but we just need to get through onboarding, so it's fine to just dismiss it
+        cy.getTestElement('@device-compromised/back-button').click();
```
Lemonexe, line 18: This is the only case where clicking on Back button remains necessary, as it purposefully uses `2-main`
But it's better to use `data-testid` which was introduced in #15031

packages/suite-web/e2e/tests/settings/t2t1-device-settings.test.ts
```diff
@@ -82,14 +82,11 @@ describe('T2T1 - Device settings', () => {
     });
 
     it('able to change homescreen in firmware >= 2.5.4', () => {
-        cy.task('startEmu', { wipe: true, model: 'T2T1', version: '2-main' });
+        cy.task('startEmu', { wipe: true, model: 'T2T1', version: '2-latest' });
```
Lemonexe, line 85: I am not sure why this test required `2-main`, I suspect it was an oversight.
`2-latest` IMHO makes sense for this test. :thinking:

HajekOndrej (approved):

PR #15055, opened 2024-10-24
fix: cursor in search-action when collaped

### LINKED ISSUE #11312: Search button/field doesn't work properly
**Describe the bug**
Search button/field doesn't work properly.
image

1. when mouse over the button, there should be the "hand" and not cursor.
image

2. when you type something into the field, the "x" button doesn't work 
image

**Info:**

-   Remembered Wallet: yes
-   Suite Version: 24.2.2

**How to reproduce**

1. hover over the button
2. type some text into the text field and try to delete with the "x" button

**Expected behavior**
1. See a hand on hover
2. be able to delete the whole text with the "x" button

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite/issues/11312

image

### DIFF
--- a/packages/suite/src/components/wallet/SearchAction.tsx
+++ b/packages/suite/src/components/wallet/SearchAction.tsx
@@ -36,9 +36,12 @@ const StyledInput = styled(Input)<{ $isExpanded: boolean }>`
     overflow: hidden;
     border-radius: ${borders.radii.full};
 
+    cursor: ${({ $isExpanded }) => (!$isExpanded ? 'pointer' : undefined)};
+
     input {
         height: ${INPUT_WIDTH};
         border: none;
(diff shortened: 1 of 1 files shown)


Lemonexe (approved):
LGTM, tested :heavy_check_mark:

packages/suite/src/components/wallet/SearchAction.tsx
```diff
@@ -104,14 +107,14 @@ export const SearchAction = ({
                 innerAddon={
                     <Icon
                         name="search"
-                        cursorPointer
```
Lemonexe: now I noticed that the icon is clickable when the input is expanded (it folds it), so it should have 
`cursorPointer`

peter-sanderson: Wow, nice catch. :clap:

PR #15059, opened 2024-10-24
feat(suite): permit unrecognized FW with outdated suite

Lemonexe:
## Description

Currently, FW hash check will display ghost modal if FW version is not recognized locally.
But if Suite 24.12 offers newer firmware, user installs it and connects device to Suite 24.11, it would falsely flag the devices as non-existent version.

In that case, skip ghost modal, don't block receive addresses, but display urgent warning to update Suite.

:information_source: btw, FW _revision_ check tries to download `releases.json`, so it doesn't have the "uncertain state" unless offline. That could be done for hash check, but the hash check itself would still be skipped (binary not present). Urging people to update suite, because fw is too new, is better than urging them to update fw because too old, AFAIK?

## Related Issue

Resolve https://github.com/trezor/trezor-suite-private/issues/115

## Screenshots:

new error banner

## QA instructions

This time, we do not need to verify that FW hash check works correctly for legit vs. fake devices. This is just _an exemption_ from FW hash check error, so it is sufficient to test with `trezor-user-env` (which simulates fake device).

There are 8 cases to verify:
- this branch vs. [testing branch](https://github.com/trezor/trezor-suite/tree/feat/hash-check-on-newer-fw-than-suite-TESTING) (it simulates outdated suite with available update)
- latest released version vs. unreleased version (for example T1B1 `1.12.1` vs. `1-main`)
- desktop vs. web

**Expected results**: all for both desktop & web, and starting with both fw checks **on**:
- latest released version, both branches:
Ghost modal & can't receive & banner "Firmware hash check failed. Your Trezor may be counterfeit."
- unreleased version, this branch:
Ghost modal & can't receive & banner "Firmware revision check failed. Your Trezor may be counterfeit."
  - Turn off fw **revision** check and reload:
  Ghost modal & can't receive & banner "Firmware unrecognized. Your Trezor may be counterfeit."
  - Turn fw revision check back on
- unreleased version, testing branch:
Ghost modal & can't receive & banner "Firmware revision check failed. Your Trezor may be counterfeit."
  - Turn off fw revision check and reload:
  **no** Ghost & **can** receive & banner "Firmware unrecognized. Please update your Trezor Suite to verify your Trezor."

### DIFF
--- a/packages/suite/src/reducers/suite/suiteReducer.ts
+++ b/packages/suite/src/reducers/suite/suiteReducer.ts
@@ -1,11 +1,7 @@
 import produce from 'immer';
 
 import type { InvityServerEnvironment } from '@suite-common/invity';
-import {
-    Feature,
-    MessageSystemRootState,
-    selectIsFeatureDisabled,
-} from '@suite-common/message-system';
+import { Feature, selectIsFeatureDisabled } from '@suite-common/message-system';
 import { isDeviceAcquired } from '@suite-common/suite-utils';
 import { discoveryActions, DeviceRootState, selectDevice } from '@suite-common/wallet-core';
 import { versionUtils } from '@trezor/utils';
@@ -24,13 +20,14 @@ import { ensureLocale } from 'src/utils/suite/l10n';
 import type { Locale } from 'src/config/suite/languages';
 import { SUITE, STORAGE } from 'src/actions/suite/constants';
 import { ExperimentalFeature } from 'src/constants/suite/experimental';
-import { Action, Lock, TorBootstrap, TorStatus } from 'src/types/suite';
+import { Action, AppState, Lock, TorBootstrap, TorStatus } from 'src/types/suite';
 import { getExcludedPrerequisites, getPrerequisiteName } from 'src/utils/suite/prerequisites';
 import { RouterRootState, selectRouter } from './routerReducer';
 import { NetworkSymbol } from '@suite-common/wallet-config';
 import { SuiteThemeVariant } from '@trezor/suite-desktop-api';
 import { AddressDisplayOptions, WalletType } from '@suite-common/wallet-types';
 import { SIDEBAR_WIDTH_NUMERIC } from 'src/constants/suite/layout';
+import { UpdateState } from './desktopUpdateReducer';
 
 export interface SuiteRootState {
     suite: SuiteState;
@@ -443,12 +440,10 @@ export const selectHasExperimentalFeature =
     (feature: ExperimentalFeature) => (state: SuiteRootState) =>
         state.suite.settings.experimental?.includes(feature) ?? false;
 
-type StateForFirmwareChecks = SuiteRootState & DeviceRootState & MessageSystemRootState;
-
 /**
  * Get firmware revision check error, or null if check was successful / skipped.
  */
-export const selectFirmwareRevisionCheckError = (state: StateForFirmwareChecks) => {
+export const selectFirmwareRevisionCheckError = (state: AppState) => {
     const device = selectDevice(state);
     if (!isDeviceAcquired(device) || !device.authenticityChecks) return null;
 
@@ -464,7 +459,7 @@ export const selectFirmwareRevisionCheckError = (state: StateForFirmwareChecks) 
  * Determine hard failure of firmware revision check - specific error types which are severe.
  * If Suite is offline and cannot perform check or there is some unexpected error, a banner is shown but device is accessible.
  */
-const selectIsFirmwareRevisionCheckEnabledAndFailed = (state: StateForFirmwareChecks): boolean => {
+const selectIsFirmwareRevisionCheckEnabledAndFailed = (state: AppState): boolean => {
     const error = selectFirmwareRevisionCheckError(state);
     const softErrors: FirmwareRevisionCheckError[] = [
         'cannot-perform-check-offline',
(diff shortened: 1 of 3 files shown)


packages/suite/src/support/messages.ts
```diff
@@ -7047,6 +7047,11 @@ export default defineMessages({
         defaultMessage:
             "Firmware hash check couldn't be performed. Your Trezor may be counterfeit.",
     },
+    TR_DEVICE_FIRMWARE_UNRECOGNIZED_OUTDATED_SUITE: {
+        id: 'TR_DEVICE_FIRMWARE_UNRECOGNIZED_OUTDATED_SUITE',
+        defaultMessage:
+            'Firmware unrecognized. Please update your Trezor Suite to verify your Trezor.',
```
Lemonexe, line 7053: dev placeholder, shall be changed in crowdin

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -443,12 +440,10 @@ export const selectHasExperimentalFeature =
     (feature: ExperimentalFeature) => (state: SuiteRootState) =>
         state.suite.settings.experimental?.includes(feature) ?? false;
 
-type StateForFirmwareChecks = SuiteRootState & DeviceRootState & MessageSystemRootState;
-
 /**
  * Get firmware revision check error, or null if check was successful / skipped.
  */
-export const selectFirmwareRevisionCheckError = (state: StateForFirmwareChecks) => {
+export const selectFirmwareRevisionCheckError = (state: AppState) => {
```
Lemonexe, line 446: `AppState` describes more accurately what is expected than `StateForFirmwareChecks`

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -489,12 +484,26 @@ export const selectFirmwareHashCheckError = (state: StateForFirmwareChecks) => {
...
+ */
+export const selectIsUnrecognizedFirmwareWithOutdatedSuite = (state: AppState): boolean => {
+    const device = selectDevice(state);
+    if (!isDeviceAcquired(device) || !device.authenticityChecks?.firmwareHash) return false;
+    const isUpdateAvailable = state.desktopUpdate.state === UpdateState.Available;
+    const checkResult = device.authenticityChecks.firmwareHash;
+
+    return !checkResult.success && checkResult.error === 'unknown-release' && isUpdateAvailable;
+};
+
```
Lemonexe, line 495: I hope it's clear from description - this select isn't meant to return "is ok" | "not ok",
but rather "is this a specific error case" | "any other case, be it success or error"

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -489,12 +484,26 @@ export const selectFirmwareHashCheckError = (state: StateForFirmwareChecks) => {
     return isCheckEnabled && checkResult?.success === false ? checkResult.error : null;
 };
 
+/**
+ * Determine if there is a FW hash check error that could be caused by outdated Suite version (a subset of firmware hash error)
+ */
+export const selectIsUnrecognizedFirmwareWithOutdatedSuite = (state: AppState): boolean => {
+    const device = selectDevice(state);
+    if (!isDeviceAcquired(device) || !device.authenticityChecks?.firmwareHash) return false;
+    const isUpdateAvailable = state.desktopUpdate.state === UpdateState.Available;
```
Lemonexe, line 490: I verified that `desktopReducer` is in the basic set of reducers common for all apps, so it is guaranteed to be on state.
If I understand correctly, it gets updated by actions which are **only** triggered on desktop (`suite-desktop-ui`)

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -489,12 +484,26 @@ export const selectFirmwareHashCheckError = (state: StateForFirmwareChecks) => {
     return isCheckEnabled && checkResult?.success === false ? checkResult.error : null;
 };
 
+/**
+ * Determine if there is a FW hash check error that could be caused by outdated Suite version (a subset of firmware hash error)
```
peter-sanderson: To me, it seams that the comment says same think as the name of the method. (I think it is not necessary)

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -489,12 +484,26 @@ export const selectFirmwareHashCheckError = (state: StateForFirmwareChecks) => {
...
+ */
+export const selectIsUnrecognizedFirmwareWithOutdatedSuite = (state: AppState): boolean => {
+    const device = selectDevice(state);
+    if (!isDeviceAcquired(device) || !device.authenticityChecks?.firmwareHash) return false;
+    const isUpdateAvailable = state.desktopUpdate.state === UpdateState.Available;
+    const checkResult = device.authenticityChecks.firmwareHash;
+
+    return !checkResult.success && checkResult.error === 'unknown-release' && isUpdateAvailable;
```
peter-sanderson, line 493: Just note that it may take some time from start of the suite to get this `isUpdateAvailable` to be true. It is doing network stuff and on slow connection / when our servers are slow this may take a while.

peter-sanderson (approved):

packages/suite/src/reducers/suite/suiteReducer.ts
Lemonexe: I am aware of that. Alas, I see no other way that wouldn't compromise security of the check :disappointed: 
What do you think?

packages/suite/src/reducers/suite/suiteReducer.ts
Lemonexe: Removed. I was afraid it wouldn't be descriptive enough without it.

PR #15064, opened 2024-10-24
chore(suite): device state to new object format

### LINKED ISSUE #14972: suite: start using device._state for connect calls
now `device.state` is used to identify specific device on TrezorConnect level. 

`device.state` is deprecated and will be replaced with `device._state` which we should start using now. 

image

this should also cause passphrase to persist suite reloads if implemented correctly

now: 
1. entre passphrase
2. remember device 
3. reload app
4. get address -> get prompted 
image

later: 
number 4 shouldn't happen

## QA notes
- everything with saved devices can break
- everything that is using device state as a reference to something can break (saved transactions, labels)
- needs migration

martykan:
## Description

To goal of this PR is to enable Suite to persist the full state object.
Suite is already saving both `state` and `_state`, so the persistence logic for the state is already handled, we just needed to start sending the `state` object instead of the string.

The approach in this PR is forward-compatible with Connect 10, where the state as string will be removed and replaced by the `DeviceState` object.

So this is implemented by updating Suite's `TrezorDevice` type to have `state: DeviceState` ahead of the change in Connect and by adapting the rest of the code. This includes a migration of the old persisted values.

## Related Issue

Resolve #14972


mroz22 (changes requested):
it looks like it makes wallet labeling data disappear

image

image

we have a test that would have caught that - hadn't it been skipped! :( https://github.com/trezor/trezor-suite/blob/feat/suite-remember-device-state/packages/suite-web/e2e/tests/metadata/wallet-metadata.test.ts#L15 but maybe it will get fixed soon, we just discussed that with @HajekOndrej couple of days ago

martykan:
Labeling should be working now. But yeah this whole PR needs thorough testing.

mroz22 (approved):

PR #15076, opened 2024-10-25
fix: remove sudo auto-install of the udevrules, simplify it

### LINKED ISSUE #15074: Automatic trezor UDEV rules instalaltion loads indefinetly
When you try to use the automatic udev rules installation on Linux, it loads indefinitely and never progresses.

**Info:**
Suite Version: app 24.10.1
OS: Ubuntu 24.04
Bridge: 2.0.33

image

Since the udev rules have now been merged into systemd, this function might not be needed anymore? https://github.com/systemd/systemd/commit/a78a52465298e8f5a927da9c9fc56c41837018aa

peter-sanderson:
Packaging formats are moving towards sand-boxing, we no longer can do a root-code-execution. 
 - https://docs.flatpak.org/en/latest/sandbox-permissions.html
 - https://snapcraft.io/docs/security-policies
 - https://snapcraft.io/docs/interfaces

Also udev rules [are now in the systemd](https://github.com/systemd/systemd/pull/33829/commits/5888f032ea75fb7e4df0e0b7532e1b3fdc597df3) (https://github.com/systemd/systemd/pull/33813) we can expect that it will bubble up through releases of the distribution over few years.  

Resolves: https://github.com/trezor/trezor-suite/issues/15074

image


packages/suite/src/components/suite/PrerequisitesGuide/DeviceUnreadable.tsx
```diff
@@ -114,14 +27,14 @@ interface DeviceUnreadableProps {
 export const DeviceUnreadable = ({ device }: DeviceUnreadableProps) => {
     const selectedDevice = useSelector(selectSelectedDevice);
 
+    // generic troubleshooting tips
+    const items = [];
+
     // this error is dispatched by trezord when udev rules are missing
     if (isLinux() && device?.error === 'LIBUSB_ERROR_ACCESS') {
-        return <> {isDesktop() ? <UdevDesktop /> : <UdevWeb />}</>;
+        items.push(TROUBLESHOOTING_TIP_UDEV);
```
peter-sanderson, line 35: Instead of download flow, we just shot tip to install udev rules. Its same tip as on Web where user selects from deb vs. rpm.

packages/suite/src/components/suite/UdevDownload.tsx
```diff
@@ -5,10 +5,10 @@ import styled from 'styled-components';
 import { DATA_URL, HELP_CENTER_UDEV_URL } from '@trezor/urls';
 import { Button, Select, Spinner } from '@trezor/components';
 import { spacings, typography } from '@trezor/theme';
+import { suggestUdevInstaller } from '@trezor/connect';
+import { getInstallerPackage } from '@trezor/connect-common'; // Todo: this import is not dependency, and shall not be
```
peter-sanderson: I need help from somebody from connect team to help me figure out the best way how to incorporate this into suite. :pray:

packages/suite/src/components/suite/UdevDownload.tsx
```diff
@@ -51,16 +51,12 @@ interface Installer {
...
-    const transport = useSelector(selectTransport);
+    const installers: Installer[] = suggestUdevInstaller(getInstallerPackage()).packages.map(p => ({
+        label: p.name,
+        value: DATA_URL + p.url.substring(1),
+        preferred: p.preferred,
+    }));
 
-    const installers: Installer[] =
-        transport && transport.udev
-            ? transport.udev.packages.map(p => ({
```
peter-sanderson: - [ ] This is only available on the web. It was fixed [here](https://github.com/trezor/trezor-suite/pull/15771) I think we shall do something similar for desktop as well.

packages/connect/src/exports.ts
```diff
@@ -3,3 +3,4 @@ export * from './events';
 export * from './types';
 
 export { parseConnectSettings } from './data/connectSettings';
+export { suggestUdevInstaller } from './data/udevInfo';
```
peter-sanderson: Probably shall not be done and it shall be send to state via emitting event the same way it is done on the web

packages/suite/src/components/suite/PrerequisitesGuide/DeviceUnreadable.tsx
```diff
@@ -1,106 +1,19 @@
...
-        );
-    }
-
-    return (
-        <TroubleshootingTips
-            opened={response === 0}
-            label={<Translation id="TR_TROUBLESHOOTING_UNREADABLE_UDEV" />}
-            cta={
-                <Button onClick={handleCtaClick}>
-                    <Translation id="TR_TROUBLESHOOTING_UDEV_INSTALL_TITLE" />
```
mroz22, line 84: some records in messages.ts should probably be cleared as well (`TR_TROUBLESHOOTING_UDEV_INSTALL_TITLE`...)

packages/suite/src/components/suite/UdevDownload.tsx
mroz22: hmm the only place where suggestUdevInstaller is now used in connect is this 
https://github.com/trezor/trezor-suite/blob/udev-rules/packages/connect-web/src/module/index.ts#L35
and this is also only used by suite now so we are free to decide where we want it. maybe we can keep  this logic in suite only now? any ideas? @marekrjpolak

packages/suite/src/components/suite/PrerequisitesGuide/DeviceUnreadable.tsx
```diff
@@ -1,106 +1,19 @@
...
+    TROUBLESHOOTING_TIP_UDEV,
 } from 'src/components/suite/troubleshooting/tips';
-import { useSelector, useDispatch } from 'src/hooks/suite';
+import { useSelector } from 'src/hooks/suite';
 import type { TrezorDevice } from 'src/types/suite';
 
-// linux web
-const UdevWeb = () => (
-    <TroubleshootingTips
-        label={<Translation id="TR_TROUBLESHOOTING_UNREADABLE_UDEV" />}
```
peter-sanderson, line 24: translation deleted

packages/suite/src/components/suite/PrerequisitesGuide/DeviceUnreadable.tsx
```diff
@@ -1,106 +1,19 @@
...
-        items={[
-            {
-                key: 'udev-about',
-                noBullet: true,
-                description: <Translation id="TR_UDEV_DOWNLOAD_DESC" />,
```
peter-sanderson, line 29: still used in `UdevRules` compoent

packages/suite/src/components/suite/PrerequisitesGuide/DeviceUnreadable.tsx
```diff
@@ -1,106 +1,19 @@
...
-
-            setResponse(0);
-        }
-    };
-
-    if (response === 1) {
-        return (
-            <TroubleshootingTips
-                opened={false}
-                label={<Translation id="TR_RECONNECT_IN_NORMAL" />}
```
peter-sanderson, line 71: still used in `ReconnectDevicePrompt`

packages/suite/src/components/suite/PrerequisitesGuide/DeviceUnreadable.tsx
```diff
@@ -1,106 +1,19 @@
...
-                items={[]}
-                data-testid="@connect-device-prompt/unreadable-udev"
-            />
-        );
-    }
-
-    return (
-        <TroubleshootingTips
-            opened={response === 0}
-            label={<Translation id="TR_TROUBLESHOOTING_UNREADABLE_UDEV" />}
```
peter-sanderson, line 81: deleted

packages/suite/src/components/suite/PrerequisitesGuide/DeviceUnreadable.tsx
peter-sanderson, line 84: deleted

peter-sanderson: Good point, deleted

packages/suite/src/components/suite/PrerequisitesGuide/DeviceUnreadable.tsx
```diff
@@ -1,106 +1,19 @@
...
-            cta={
-                <Button onClick={handleCtaClick}>
-                    <Translation id="TR_TROUBLESHOOTING_UDEV_INSTALL_TITLE" />
-                </Button>
-            }
-            items={[
-                {
-                    key: 'udev-about',
-                    description: <Translation id="TR_UDEV_DOWNLOAD_DESC" />,
```
peter-sanderson, line 90: used in `UdevRules`

mroz22 (approved):

peter-sanderson:
## For QA:

Linux only :point_down: 

1. desktop suite 
2. uninstall udev-rules 
3. reconnect device
4. you see the info with hints
5. if you click through udev-rules -> you shall see the modal and you shall be able to download deb or rmp package

bosomt:
**QA OK**

- tested on mint and ubuntu
- tested on latest dev build
- rpm is preselected and probably not detected
- more common package is deb, maybe we can consider changing this behavior ?

PR #15083, opened 2024-10-25
feat(suite-native): solana local graph history movements

### LINKED ISSUE #14969: Solana - decode txn to get event history
To add Solana to Graph we need to decode all wallet transactions to get amount changes. It is already done for EVMs, but not for Solana.

- Count history movements and show graph with events for Solana, similar to EVMs

vytick:
Adds historical movements for Solana. If there is a lot of txns it can take longer time to fetch all the transaction data necessary from the chain.

## Related Issue

Resolve #14969

### DIFF
--- a/suite-common/graph/src/balanceHistoryUtils.ts
+++ b/suite-common/graph/src/balanceHistoryUtils.ts
@@ -305,6 +305,152 @@ export const getAccountHistoryMovementItemETH = ({
     };
 };
 
+export const getAccountHistoryMovementItemSOL = ({
+    transactions,
+    from,
+    to,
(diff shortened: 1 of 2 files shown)


Nodonisko (approved):

Nodonisko:
We may consider bigger page size for fetching transactions. There could be even some heuristic for that based on total txs of account and timeframe because we know this in advance.

vytick:
> We may consider bigger page size for fetching transactions. There could be even some heuristic for that based on total txs of account and timeframe because we know this in advance.

For sure, but I would like to test this in the wild, so we know how it behaves, and do optimisations afterwards.

PR #15088, opened 2024-10-27
feat(suite#4941): warn about non-ASCII characters in passcode

### LINKED ISSUE #4941: Warn about non-ASCII characters in passphrase
## User story

**[Instructions + Design](https://github.com/trezor/trezor-suite/issues/4941#issuecomment-2338726186)**

As a Trezor Suite user
I want to protect my privacy
So that my funds are safe when my seed is compromised

As a Trezor Suite user
I want to create a hidden wallet passphrase on suite
And I want to confirm it on my device
So that I can trust I have create a hidden wallet securely

As a Trezor Suite user
I want to access my hidden wallet
And I want to enter my passphrase on a device I trust (Model T)
So that I can access my funds

## Acceptance criteria
Given I am on 'Select wallet type'
When I type '€'
Then I should see a 'Unable to create a wallet with this passphrase'
And the 'Access Hidden wallet' button should be disabled
And the 'Enter passphrase on Trezor' link should be disabled

**Caveat**: Backwards compatibility issue, potentially blocking users with UTF-8 characters in passphrases.

### Passphrase 'Š'
IMG_20220214_155118

### Passphrase '€'
IMG_20220214_160314

TomasSlama:
resolves #4941 

This pull request includes several changes to improve the `PassphraseTypeCard` component, add new features, and enhance the user experience. The most important changes include adding support for non-ASCII character detection, updating the `PassphraseTypeCard` stories, and introducing new tests for the added functionality.

### Enhancements to `PassphraseTypeCard`:

* Added support for detecting non-ASCII characters and displaying appropriate warnings. This includes the `useNonAsciiChars` hook and the `getSubmitLabel` utility function. 
* Updated `PassphraseTypeCardContent` to include new props and display a banner for non-ASCII character warnings. 

### Storybook Updates:

* Renamed the default export in `PassphraseTypeCard.stories.tsx` and added a new story for the hidden passphrase type. (`packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCard.stories.tsx`)

### Translation Updates:

* Added new translation keys for non-ASCII character warnings. 

### Video

https://github.com/user-attachments/assets/1b5b99fc-1861-4430-93b0-11f8e66334ea

### Screenshot

Screenshot 2024-10-31 at 18 19 51


packages/product-components/package.json
```diff
@@ -24,6 +24,7 @@
         "@trezor/connect": "workspace:*",
         "@trezor/dom-utils": "workspace:*",
         "@trezor/env-utils": "workspace:*",
+        "@trezor/protobuf": "workspace:*",
```
TomasSlama: I couldn't figure out a better way how to include `BackupType`

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCardContent.tsx
```diff
@@ -175,6 +184,11 @@ export const PassphraseTypeCardContent = ({
                         {value && !isPassphraseTooLong && (
                             <PasswordStrengthIndicator password={value} />
                         )}
+                        {showAsciiBanner && (
+                            <Banner variant={asciiBannerVariant}>
+                                TBD
```
TomasSlama: I was not able to find already done components that would be able to render such design, as I don't have access to the Figma file, I wanted to create new components `Code` and `Prev` in `packages/components`, but I noticed that there is no `mono` font used so far, so to finish this part I would need more details about the design (font, spacing).

packages/product-components/src/components/PassphraseTypeCard/getSubmitLabel.test.tsx
TomasSlama, line 1: as mentioned in the PR description this test is only representative as there is not Jest setup for this package

packages/product-components/src/components/PassphraseTypeCard/getSubmitLabel.tsx
```diff
@@ -0,0 +1,28 @@
+import type { ReactNode } from 'react';
+import React from 'react';
+import { Translation } from '../../../../suite/src/components/suite/Translation';
```
TomasSlama: I don't like this solution, as there is not a single usage of `Translation` in `/packages/product-components` so far, so it seems I'm breaking some kind of rule here. 

Also, I know that I'm creating circular dependency with this. It seems that the ideal solution for me would be to extract Translation from `suite` so it could be used even here 🤔

The other option that I considered was to pass the translations from the above level, but then more refactoring would be needed as we don't have access to the `value` on the above level. Also it does not make sense to pass it from the level above as I believe these Translations will never change and should be part of the component 🤔 I'm interested in your thoughts on this topic 🙏

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.test.ts
TomasSlama, line 1: the same as the test above, as jest is not setup in the package it is only representative

packages/suite-data/files/translations/en.json
TomasSlama, line 1: I didn't add other languages as they were not provided, I could have used chatGPT to translate it for me, but I'm not sure what is the proper process here.

tomasklim: It is not required to add anything to `en.json`. It is controlled by crowdin. `messages.ts` is enough

packages/product-components/package.json
tomasklim:
```
import { Features } from '@trezor/connect';

backupType: Features["backup_type"]
```

This could be the way

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useState } from 'react';
+
+// Regular expression to match non-ASCII characters
+const nonAsciiPattern = /[^\x00-\x7F]/g;
```
tomasklim: The `nonAsciiPattern` (`/[^\x00-\x7F]/g`) together with `match` could make a nice utility in `@trezor/utils` package. Then, you’d only need to test that utility, avoiding the need to test the hook itself.

By the way, is the hook even necessary? It seems like it might be overkill.

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.test.ts
tomasklim: Is the test just generated by some AI tool? 
The reason I’m asking is that the design shows certain characters. Simplified Chinese is nice, but it would also be good to test some other "real" non-ASCII characters in our context, as we’re based in Europe, where Chinese isn’t typically the first we think of 😀

packages/product-components/src/components/PassphraseTypeCard/getSubmitLabel.tsx
tomasklim: We use 
`<FormattedMessage id="ID_HERE" />`

packages/product-components/src/components/PassphraseTypeCard/getSubmitLabel.test.tsx
tomasklim: no need to test this, we usually do not do that

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCardContent.tsx
tomasklim: I think we use `mono` in `DeviceDisplayText` 
Figma props here:
Screenshot 2024-10-27 at 15 57 07

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.test.ts
TomasSlama: yeah, I usually let chat gpt cover my functions with unit tests, when the function is simple enough the quality of the test is usually good. Good point to replace mandarine with czech letters for example. Thanks for the feedback.

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.ts
TomasSlama: I agree about moving the regexp to the util function, but I believe I would still need to use useEffect, to be able to show the banner when the user would write first invalid character and keep it open after that 🤔 

The test does not cover only the basic logic of regexp, but also the requirements, especially the one which does not have to be obvious right away: when the banner is shown it will stay shown even after the user removes invalid characters, so I like having tests like these as when you read them you can explicitly see what are the requirements.

packages/product-components/package.json
TomasSlama: it works, thank you 🙇

packages/components/src/components/typography/Code/Code.tsx
```diff
@@ -0,0 +1,20 @@
...
+    font-family: RobotoMono, PixelOperatorMono8, monospace;
+    display: inline;
+    font-size: inherit;
+    line-height: 1.5;
+    font-weight: 400;
+    letter-spacing: -0.4px;
+    padding: 0px ${() => spacingsPx.xxxs};
+    background-color: ${({ theme }) => theme.backgroundNeutralSubtleOnElevation1};
+    box-shadow: inset 0 0 0 1px ${({ theme}) => theme.borderElevation0 };
+    border-radius: ${() => borders.radii.xxs};
```
TomasSlama: this is not pixel perfect, I would still need more details from the figma, from the image it is very hard to tell. I believe it is close enough for our purpose and could be slightly refactored by someone who has access.

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
```diff
@@ -0,0 +1,79 @@
...
+                <SpecialChars>
+                    {'! " # $ % & \\ \' ( ) * +  - . / : ; < = > ? @ [  ] ^ _ ` { | } ~'}
+                </SpecialChars>
+                <SpecialDescription>
+                    <FormattedMessage
+                        defaultMessage="Using non-listed special characters risk future compatibility"
+                        id="TR_PASSPHRASE_NON_ASCII_CHARS_WARNING"
+                    />
+                    <ButtonWrapper>
+                        {/* TODO: better would be to reuse LearnMoreButton */}
```
TomasSlama, line 64: I noticed in suite, there is `LearnMoreButton`, but I didn't want to introduce circular dependency between packages, so I duplicated the code for now.

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.test.ts
TomasSlama: I removed the test for now as it was not running anyway.

packages/product-components/src/components/PassphraseTypeCard/getSubmitLabel.tsx
```diff
@@ -0,0 +1,36 @@
+import type { ReactNode } from 'react';
```
tomasklim, line 1: In `messages.ts` you can add `label`
```ts
    TR_NON_ASCII_CHAR: {
        defaultMessage: '{label} (with non-recommended \"{char}\")',
        id: 'TR_NON_ASCII_CHAR',
    },
    TR_NON_ASCII_CHARS: {
        defaultMessage: '{label} (with non-recommended characters)',
        id: 'TR_NON_ASCII_CHARS',
    },
```

So then you can refactor it like this:
```ts
import React, { ReactNode } from 'react';
import { FormattedMessage } from 'react-intl';

type SubmitLabelProps = {
    nonAsciiChars: string[] | null;
    label: ReactNode;
};

export const getSubmitLabel = ({ nonAsciiChars, label }: SubmitLabelProps) => {
    if (!nonAsciiChars || !nonAsciiChars.length) {
	return label;
    }

    const isSingleChar = nonAsciiChars.length === 1;
    const messageId = isSingleChar ? "TR_NON_ASCII_CHAR" : "TR_NON_ASCII_CHARS";
    const defaultMessage = isSingleChar
        ? '(with non-recommended "{char}")'
        : "(with non-recommended characters)";

    return (
          <FormattedMessage
              id={messageId}
              defaultMessage={defaultMessage}
              values={{ label, char: nonAsciiChars[0] }}
          />
    );
};
```

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const SpecialDescription = styled.div`
+    background-color: ${({ theme }) => theme.backgroundNeutralDisabled};
+    padding: ${spacingsPx.xxs};
+    display: flex;
+    gap: ${spacingsPx.sm};
+    ${typography.label}
+`;
+
+type Props = {
```
tomasklim, line 40: nit:
```suggestion
type NonAsciiBannerProps = {
```

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCardContent.tsx
```diff
@@ -43,12 +44,18 @@ const Description = styled.div`
...
+    submitVariant?: 'primary' | 'warning';
     type: WalletType;
     singleColModal?: boolean;
     displayValue: string;
     isPassphraseTooLong: boolean;
     value: string;
     setValue: (value: string) => void;
+    /**
+     * If true, the banner with non-ASCI chars warning will be displayed
+     */
```
tomasklim: nit: we try to omit comments if not necessary

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCard.tsx
```diff
@@ -59,16 +62,20 @@ export type PassphraseTypeCardProps = {
...
     singleColModal?: boolean;
     deviceModel?: DeviceModelInternal;
+    deviceBackup?: Features['backup_type'] | null;
     onSubmit: (value: string, passphraseOnDevice?: boolean) => void;
     learnMoreTooltipOnClick?: TooltipProps['addon'];
     learnMoreTooltipAppendTo?: TooltipProps['appendTo'];
 };
 
+
+
```
tomasklim:
```suggestion
```

packages/utils/src/getNonAsciiChars.ts
```diff
@@ -0,0 +1,6 @@
+// Regular expression to match non-ASCII characters
```
tomasklim: there is already `isAscii` util. I am thinking about having `asciiUtil` file, moving both utils to 1 file and sharing the pattern

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCardContent.tsx
```diff
@@ -43,12 +44,18 @@ const Description = styled.div`
 
 type PassphraseTypeCardContentProps = {
     submitLabel: ReactNode;
+    submitVariant?: 'primary' | 'warning';
```
tomasklim, line 47: @adamhavel is it better to allow strictly just those two variants here or should we rather use:
```suggestion
    submitVariant?: ButtonVariant;
```

same for banner `BannerVariant`

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react';
```
tomasklim: I think it can be simplified to something like

```
export const useNonAsciiChars = (value: string) => {
    const [showAsciiBanner, setShowAsciiBanner] = useState(false);

    const nonAsciiChars = useMemo(() => getNonAsciiChars(value), [value]);

    useEffect(() => {
        if (Array.isArray(nonAsciiChars)) {
            setShowAsciiBanner(true);
        }
    }, [nonAsciiChars]);

    return { nonAsciiChars, showAsciiBanner };
};
```

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.ts
```diff
@@ -0,0 +1,23 @@
...
+import { getNonAsciiChars } from '@trezor/utils';
+
+export const useNonAsciiChars = (value: string) => {
+    const [nonAsciiChars, setNonAsciiChars] = useState<string[] | null>(null);
+    const [hasNonAsciiChars, setHasNonAsciiChars] = useState(false);
+    const [showAsciiBanner, setShowAsciiBanner] = useState(false);
+
+    useEffect(() => {
+        const nonAsciiChars = getNonAsciiChars(value);
```
tomasklim: if we modify this util to return `[]` instead of null when no matches chars are found. We can use `nonAsciiChars.length` and avoid unnecessary state of `hasNonAsciiChars`

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
```diff
@@ -0,0 +1,79 @@
...
+                </SpecialChars>
+                <SpecialDescription>
+                    <FormattedMessage
+                        defaultMessage="Using non-listed special characters risk future compatibility"
+                        id="TR_PASSPHRASE_NON_ASCII_CHARS_WARNING"
+                    />
+                    <ButtonWrapper>
+                        {/* TODO: better would be to reuse LearnMoreButton */}
+                        <Button
+                            href="https://trezor.io/learn/a/passphrases-and-hidden-wallets#Security"
```
tomasklim: we have `@trezor/urls` package for those

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
tomasklim: Yes, agree, we plan to reorganise this whole `PassphraseTypeCard` better, the composition is not optimal.

cc @adamhavel

tomasklim (changes requested):
LGTM, few more improvement proposals here

and CI jobs are failing
```
> nx run @trezor/suite:"lint:styles"

@trezor/product-components: src/components/PassphraseTypeCard/NonAsciiBanner.tsx
@trezor/product-components:  [25](https://github.com/trezor/trezor-suite/actions/runs/11579874962/job/32384206177?pr=15088#step:7:26):30  ✖  Expected quotes around "PixelOperatorMono8"  font-family-name-quotes
@trezor/product-components: 1 problem (1 error, 0 warnings)
@trezor/components: src/components/typography/Code/Code.tsx
@trezor/components:   6:30  ✖  Expected quotes around "PixelOperatorMono8"  font-family-name-quotes
@trezor/components:  12:15  ✖  Unexpected unit                              length-zero-no-unit
@trezor/components: 2 problems (2 errors, 0 warnings)
```

Check files for correct formatting
```
Run yarn nx format:check
packages/components/src/components/typography/Code/Code.stories.tsx
packages/components/src/components/typography/Code/Code.tsx
packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCard.tsx
packages/suite/src/support/messages.ts
```

packages/product-components/src/components/PassphraseTypeCard/getSubmitLabel.tsx
TomasSlama: much better solution, thank you!

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
TomasSlama: and did I choose the proper URL? It was not in the issue.

image

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
TomasSlama: ✅

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
TomasSlama: ✅

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCard.tsx
TomasSlama: ✅

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCardContent.tsx
TomasSlama: ✅

packages/product-components/src/components/PassphraseTypeCard/getSubmitLabel.tsx
TomasSlama: ✅

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
tomasklim: yes, looks good

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.ts
TomasSlama: ✅

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.ts
TomasSlama: I agree, I kept the `null` as it will never be an empty array, we can just use `nonAsciiChars` in condition

packages/utils/src/getNonAsciiChars.ts
TomasSlama: ✅

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.ts
tomasklim: you have console log there

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
tomasklim: I don't see it changed

tomasklim:
@TomasSlama after resolving the issues
Screenshot 2024-11-03 at 14 26 22

Please squash fixups and rebase to current develop 🙏

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCardContent.tsx
Lemonexe: As for `BannerVariant`, there is now an inconsistency between `PassphraseTypeCardContent` and `NonAsciiBanner`, so I'll use `BannerVariant`  both places.

Lemonexe:
FIY I will finish this PR (implement CR, do some minor cleanup) and merge it.
I will do it in another branch (I can't push to this fork).
:arrow_right: #15238

packages/product-components/src/components/PassphraseTypeCard/PassphraseTypeCardContent.tsx
Lemonexe: https://github.com/trezor/trezor-suite/pull/15238/commits/02ebd9a7bcb4cf72732beba84150775e945f5cc8

packages/product-components/src/components/PassphraseTypeCard/NonAsciiBanner.tsx
Lemonexe: Resolved in https://github.com/trezor/trezor-suite/pull/15238

packages/product-components/src/components/PassphraseTypeCard/useNonAsciiChars.ts
Lemonexe: Resolved in https://github.com/trezor/trezor-suite/pull/15238

tomasklim:
@TomasSlama Thank you for your contribution! Your improvement is going to be part of the new app release in 2 weeks. See you soon!

Lemonexe:
https://github.com/trezor/trezor-suite/pull/15238 is merged :heavy_check_mark: :rocket:

PR #15096, opened 2024-10-29
chore(connect-explorer): remove redundant code

mroz22:
did spreading state into copy spread copy of state make any sense?


martykan (approved):
That's indeed some strange code

PR #15099, opened 2024-10-29
feat(suite-native): ethereum send MVP

### LINKED ISSUE #15060: Send Ethereum - mainnet and tokens MVP
We already have send form working for BTC. Make the Bitcoin version of the send form work for ETH and its tokens as well.

## Acceptance criteria
- It is possible to dispatch an ETH or ERC20 transaction from Trezor Suite Lite
- the UI and UX are not adjusted to ETH it still looks the same as BTC flow at this stage

PeKne:
## Description
Now, sending ETH and its tokens via the mobile send form is possible.

The PR is divided into two commits. First of the enabled ETH mainnet send flow. The second one focuses on tokens.
## Related Issue

Resolve #15060 

## Screenshots:

https://github.com/user-attachments/assets/ba6631bd-7e36-4bad-b875-e5e915720947


suite-native/module-send/src/components/RecipientsSummary.tsx
```diff
@@ -78,20 +84,24 @@ export const RecipientsSummary = ({ accountKey, selectedFeeLevel }: FeesRecipien
...
+                            <CoinAmountFormatter
                                 variant="hint"
                                 color="textSubdued"
                                 value={output.amount}
-                                network={networkSymbol}
                                 isBalance={isBalance}
+                                accountKey={accountKey}
+                                tokenContract={tokenContract}
+                                adjustsFontSizeToFit
+                                numberOfLines={1}
```
yanascz, line 104: Is this something we'll ever want to set to value other than 1? 🤔 Would consider removing from the component's interface.

suite-native/module-send/src/components/SendScreenSubHeader.tsx
```diff
@@ -33,34 +42,50 @@ export const AccountBalanceScreenHeader = ({ accountKey }: AccountBalanceScreenH
...
+
+    const tokenBalance = useSelector((state: AccountsRootState) =>
+        selectAccountTokenBalance(state, accountKey, tokenContract),
+    );
+
     if (!networkSymbol) {
         return;
     }
 
+    const accountTitle = tokenSymbol ? `${accountLabel} - ${tokenSymbol}` : accountLabel;
```
yanascz: Can we use en dash (–) instead of hyphen (-) here?

suite-native/module-send/src/components/RecipientsSummary.tsx
PeKne: These formatters are only reusing the [Text component](https://github.com/trezor/trezor-suite/blob/1996ba9d97bb1d604a0673680ebec0f6fc5cd106/suite-native/atoms/src/Text.tsx) API. If we would like to always set this prop equal to 1 if the `adjustFontSizeToFit` is true, we should refactor every usage of it in the repo. I'm not saying I'm against it, but not sure if it is worth it. Feel free to create an issue for that, and we can discuss it during the next standup meeting.

suite-native/module-send/src/components/SendScreenSubHeader.tsx
PeKne: fixed [3107f98](https://github.com/trezor/trezor-suite/pull/15099/commits/3107f982ed5688ece658c67db8ed8a4767c62eee)

suite-native/module-send/src/components/RecipientsSummary.tsx
Nodonisko: This components are wrapped in our own, if you set it as default in `CoinAmountFormatter` and `CoinToFiatAmountFormatter` it shoudn't be issue and it shoudn't require any fefactorings/

PeKne: You are right that I can do it for these two formatters created in this PR. But IMO, it would make sense to have the API unified everywhere (also other `...Formatter` components, maybe even `Text`?). Because it might be kind of counterintuitive that some components work with `adjustsFontSizeToFit` only if you specify `numberOfLines` and some do not.

What do you think?

yanascz: Some refactoring might make sense but as a follow-up. I was asking in the context of this change.

yanascz (approved):
Seems to work fine after the fixes. 👍

Issue #14616, opened 2024-09-30
Mobile device management - change PIN


Nodonisko:
[Figma](https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Manage-device?node-id=367-2922&t=um4cRUycAW4gJnBL-1)

**User can change PIN of the device in Device settings**

- If user has PIN disabled (not set) he can only Enable
- When user clicks on Enable he is prompted to continue on Trezor and confirm it there
- T1
- if user has PIN protection enabled and wants to change on T1, he clicks on Change and is prompted to confirm on Trezor. Then he is prompted to enter current PIN on mobile screen, then enters the new PIN on mobile screen.
- If user wants to disable PIN protection on T1, he clicks on Disable and is prompted to confirm it on Trezor device, then enters current PIN on mobile screen.

**Other models**

- If user has PIN protection enabled and wants to change, he clicks on Change and is prompted to confirm on Trezor. Then he is prompted to enter current PIN on Trezor device, then enters new PIN on Trezor Device
- If user wants to disable PIN protection, he clicks on Disable and is prompted to confirm it on Trezor device, then enters current PIN on Trezor device

STew790:
**QA OK**
- tried the PIN on T1 and TS5
- I was able to enable/disable/change the PIN
- cancel at any point (either on device or on host) worked just fine and I always returned to the settings page

**Info**
24.10.2 https://github.com/trezor/trezor-suite/commit/17d794f5dee37026247df980bdccae1f7e489cae

Issue #14615, opened 2024-09-30
Mobile device management - enable/disable PIN


Nodonisko:
[Figma](https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Manage-device?node-id=367-2923&t=um4cRUycAW4gJnBL-1)

- Use can enable or disable PIN for device

STew790:
**QA OK**
Already approved here https://github.com/trezor/trezor-suite/issues/14616

PR #15103, opened 2024-10-29
feat(suite-native): add enable/change/disable PIN protection e2e test

yanascz:
## Related Issues

#14615 #14616

### DIFF
--- a/suite-native/app/e2e/tests/deviceSettings.test.ts
+++ b/suite-native/app/e2e/tests/deviceSettings.test.ts
@@ -1,0 +1,80 @@
+import { conditionalDescribe } from '@suite-common/test-utils';
+import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
+
+import { onAlertSheet } from '../pageObjects/alertSheetActions';
+import { onBottomSheet } from '../pageObjects/bottomSheetActions';
+import { onCoinEnablingInit } from '../pageObjects/coinEnablingActions';
+import { onConnectingDevice } from '../pageObjects/connectingDevice';
+import { onDeviceManager } from '../pageObjects/deviceManagerActions';
+import { onDeviceSettings } from '../pageObjects/deviceSettingsActions';
+import { onOnboarding } from '../pageObjects/onboardingActions';
+import {
+    appIsFullyLoaded,
+    disconnectTrezorUserEnv,
+    openApp,
+    prepareTrezorEmulator,
+    restartApp,
+} from '../utils';
+
+conditionalDescribe(device.getPlatform() !== 'android', 'Device settings', () => {
+    beforeAll(async () => {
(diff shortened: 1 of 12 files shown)


suite-native/app/e2e/pageObjects/homeActions.ts
```diff
@@ -7,6 +7,19 @@ class HomeActions {
             .withTimeout(10000);
     }
 
+    async waitForConnectingDeviceScreen() {
```
PeKne: I'm not sure if this should be placed in this file. I believe that this file should contain only actions that are related specifically to the `HomeScreen`. `@screen/ConnectingDevice` is a different screen.

suite-native/app/e2e/pageObjects/homeActions.ts
```diff
@@ -7,6 +7,19 @@ class HomeActions {
...
+        await waitFor(element(by.id('@screen/ConnectingDevice')))
+            .toBeVisible()
+            .withTimeout(10000);
+    }
+
+    async tapDeviceManagerSwitch() {
```
PeKne: What about placing this to `deviceManagerActions`?

suite-native/app/e2e/tests/deviceSettings.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+        await onDeviceSettings.tapEnablePinProtectionButton();
+
+        await TrezorUserEnvLink.pressNo();
+        await onAlertSheet.tapPrimaryButton();
+
+        await TrezorUserEnvLink.pressYes();
+        await TrezorUserEnvLink.inputEmu('42');
+        await TrezorUserEnvLink.inputEmu('42');
+        await TrezorUserEnvLink.pressYes();
```
PeKne, line 57: There is not check if the action was successful. What about asserting if the success toast was displayed after each of these three pin actions?

PeKne: Or at least checking that redirect back to the device settings screen was successful after the last step of disabling the PIN again.

PeKne (approved):

suite-native/app/e2e/pageObjects/homeActions.ts
yanascz: Moved to a new file.

suite-native/app/e2e/pageObjects/homeActions.ts
yanascz: Makes sense.

suite-native/app/e2e/tests/deviceSettings.test.ts
yanascz: Added `await onDeviceSettings.waitForScreen();`.

PR #15108, opened 2024-10-29
chore(suite): unify staking files

### LINKED ISSUE #13754: Unify staking files
Right now utils and types are all over suite in many files.

Please unify as much as possible.

There is `stake.ts` utils file in `packages/suite`
There is `stake.ts` utils file in `@suite-common/suite-utils`
Then there is `stakingUtils.ts` in `@suite-common/wallet-utils`

There is `stake.ts` utils file in `@suite-common/wallet-types`
There is `stakeTypes.ts` utils file in `@suite-common/wallet-core`
- there are also values which are definitely not types in this file. Maybe should be in file `ethStaking.ts`

The file with constants `ethStaking.ts` is also hard to find, please come up with better naming

Please do some kind of cleaning and organising so that it is easy to find what we are looking for.

dev-pvl:
## Description

Combined all stake utils into a single file. 
Moved constants into wallet-constants and pulled non-type items out of type files.
Renamed a few files to make things easier to find

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13754


tomasklim:
@dev-pvl please resovle conflicts

dev-pvl:
@tomasklim done

suite-common/wallet-utils/src/ethereumStakingUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+            acc.push(networkSymbol);
+        }
+
+        return acc;
+    }, [] as SupportedNetworkSymbol[]);
+
+// Define signature constants
+const STAKE_SIGNATURE = '0x3a29dbae';
+const UNSTAKE_SIGNATURE = '0x76ec871c';
+const CLAIM_SIGNATURE = '0x33986ffa';
```
tomasklim, line 91: Can those be added to your lib? and we just import it from @everstake/sdk?

suite-common/wallet-utils/src/ethereumStakingUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+export const getSignatureByEthereumDataHex = (dataHex: string) => `0x${dataHex.slice(0, 8)}`;
+
+export const getTxStakeNameByDataHex = (dataHex: string | undefined): StakeType | null => {
+    if (!dataHex) return null;
+    const signature = getSignatureByEthereumDataHex(dataHex);
+
+    return isStakeTypeTx(signature) ? signatureToStakeNameMap[signature] : null;
+};
+
+export const getUnstakeAmountByEthereumDataHex = (dataHex?: string) => {
```
tomasklim, line 120: I think some of those utils would be interesting for your sdk or not?

tomasklim (approved):

suite-common/wallet-utils/src/ethereumStakingUtils.ts
dev-pvl: Agree, it could be added. But let's do this as a part of the separate PR. I remember we also had some hardcoded values in the blockbook that didn't belong to it.

suite-common/wallet-utils/src/ethereumStakingUtils.ts
dev-pvl: yeah, you're right. But let's do it separately

tomasklim:
You have unused dependencies error @dev-pvl

PR #15110, opened 2024-10-29
chore(connect): move storage to `@trezor/connect`

martykan:
## Description

Simply move Storage from `@trezor/connect-common` to `@trezor/connect`. 
This lets us type things properly and makes more sense for the future. 

I also added a benign test for systemInfo in `@trezor/connect-common`, so there are some tests in the package and Jest doesn't complain


mroz22:
hmm how about with modularization? Aren't we going to move pretty much everything to @trezor/connect-common again?

martykan:
That's true, for the current situation it makes sense, but not considering the modularization

PR #15120, opened 2024-10-30
feat(suite-native): send token drafts

### LINKED ISSUE #15078: Send Ethereum - token send form drafts
The current implementation can store only one send form draft per account. Since we have a separate flow for sending mainnet and token, we need to change the implementation so that a token draft and mainnet draft can be stored without overriding each other.

## Acceptance criteria
- If the user fills the Ethereum mainnet send form, leaves it, and goes to a token send form, the values filled in the mainnet flow are not visible in the token form. 
- User can have a draft of multiple tokens without erasing each other.

PeKne:
## Description
There is a send form draft stored for each of the account tokens so the mainnet and token accounts do not override each others draft.

## Related Issue

Resolve #15078 

## Screenshots:

https://github.com/user-attachments/assets/6dbcc93f-1fc5-4827-aea7-ce4977035d82

### DIFF
--- a/suite-common/wallet-core/src/send/sendFormReducer.ts
+++ b/suite-common/wallet-core/src/send/sendFormReducer.ts
@@ -5,12 +5,15 @@ import {
     FormState,
     GeneralPrecomposedTransactionFinal,
     Output,
+    SendFormDraftKey,
+    TokenAddress,
 } from '@suite-common/wallet-types';
 import { cloneObject } from '@trezor/utils';
 import { createReducerWithExtraDeps } from '@suite-common/redux-utils';
 import { BlockbookTransaction } from '@trezor/blockchain-link-types';
 import { NetworkSymbol, networks } from '@suite-common/wallet-config';
 import { DeviceModelInternal } from '@trezor/connect';
+import { getSendFormDraftKey } from '@suite-common/wallet-utils';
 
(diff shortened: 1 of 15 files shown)


suite-common/wallet-core/src/send/sendFormReducer.ts
```diff
@@ -96,22 +104,24 @@ export const selectSendSerializedTx = (state: SendRootState) => state.wallet.sen
 export const selectSendSignedTx = (state: SendRootState) => state.wallet.send.signedTx;
 export const selectSendFormDrafts = (state: SendRootState) => state.wallet.send.drafts;
 
-export const selectSendFormDraftByAccountKey = (
+export const selectSendFormDraftByKey = (
```
vytick, line 107: nit: 
```suggestion
export const selectSendFormDraftByDraftKey = (
```

vytick (approved):
lgtm 🚀

suite-common/wallet-core/src/send/sendFormReducer.ts
PeKne: Feels redundant to me. It makes the name unnecessarily long.

martykan:
Screenshot 2024-11-01 at 15 07 18

I'm getting some TS trouble from here

PR #15123, opened 2024-10-30
Session stealing fix

marekrjpolak:
## Description

Fixes bug when `usedElsewhere` did nothing if device wasn't currently occupied here, so it could e.g. be shown as `connected` in two separate tabs at once.


mroz22 (approved):
lgbtm 🌈

PR #15125, opened 2024-10-30
fix(suite): ethereum/cardano button request hack conflicted with labeling

### LINKED ISSUE #15111: No Confirm on device modal when enabling labeling on Ethereum / Cardano account pages
**Describe the bug**
No Confirm on device modal when enabling labeling on account with remembered wallet

**Steps to reproduce:**
1. Don't have labeling enabled
2. Reload Suite
3. Go to an Ethereum / Cardano account
4. Enable labeling on this coin
5. Suite stops responding 

**Info:**
 - Suite version: web 24.11.0 (3400c549ef8efaa2f1dccc6cf8641aa3ec04c8d3)
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.4 regular (revision da7ddd5c8c0b2040eef38247d4fd73d4576e6708)
 - Transport: WebUsbTransport

**Screenshots:**

https://github.com/user-attachments/assets/2160fa5e-5db9-4a66-bb17-94cab466d5eb

martykan:
## Description

There is an ugly hack for button requests in Suite that rewrites `ButtonRequest_Other` to `ButtonRequest_SignTx` on Ethereum/Cardano account screens. 
This clashes with other calls to enable labeling, which may also happen on these screens. 

This PR modifies Connect, adding a `method` field to the button request payload, that identifies the message type to which this event is responding.

This field is then used in Suite to exclude labeling enabling (`CipherKeyValue` calls) from this hack.
 
It could also be used to remove the conditions for the hack that are based on the router to be entirely based on the method field, but I didn't want to change any  behavior unintentionally as part of this PR. 

## Related Issue

Resolve #15111

### DIFF
--- a/packages/connect/src/api/firmware/uploadFirmware.ts
+++ b/packages/connect/src/api/firmware/uploadFirmware.ts
@@ -10,7 +10,10 @@ import type { TypedCall } from '../../device/DeviceCommands';
 const postConfirmationMessage = (device: Device) => {
     // only if firmware is already installed. fresh device does not require button confirmation
     if (device.features.firmware_present) {
-        device.emit(DEVICE.BUTTON, device, { code: 'ButtonRequest_FirmwareUpdate' });
+        device.emit(DEVICE.BUTTON, device, {
+            code: 'ButtonRequest_FirmwareUpdate',
+            method: 'FirmwareUpload',
+        });
     }
 };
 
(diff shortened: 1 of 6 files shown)


martykan:
For reference - we are looking for a better solution that would involve fixing things in FW properly and adding a more elegant patch in Connect for backwards compatibility

PR #15134, opened 2024-10-31
feat(suite): soften FW hash check for other-error

Lemonexe:
## Description

Create a softer version of `DeviceCompromised` modal specifically for FW hash check `'other-error'` to handle a false positive edge case (see issue)

:information_source: Designed by myslef, shall be reworded or tweaked with the product team.

## Related Issue

Resolve https://github.com/trezor/trezor-suite-private/issues/119

## Screenshots:

**Before:** Normal ghost modal
image

**After (only for `'other-error'`):** Softer ghost modal
image

### DIFF
--- a/packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
+++ b/packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
@@ -11,7 +11,8 @@ import {
     selectFirmwareHashCheckError,
     selectFirmwareRevisionCheckError,
 } from 'src/reducers/suite/suiteReducer';
-import { SecurityCheckFail } from 'src/components/suite/SecurityCheck/SecurityCheckFail';
+import { SecurityCheckFail, SecurityCheckFailProps } from './SecurityCheckFail';
+import { softFailureChecklistItems } from './checklistItems';
 
 const reportCheckFail = (checkType: 'revision' | 'hash', contextData: any) => {
     withSentryScope(scope => {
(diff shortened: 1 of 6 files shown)


packages/suite/src/components/suite/SecurityCheck/SecurityCheckFail.tsx
```diff
@@ -19,33 +21,20 @@ const Flex = styled.div`
...
-    },
-    {
-        icon: 'hand',
-        content: <Translation id="TR_AVOID_USING_DEVICE" />,
-    },
-    {
-        icon: 'chat',
-        content: <Translation id="TR_USE_CHAT" values={{ b: chunks => <b>{chunks}</b> }} />,
-    },
-] as const;
```
Lemonexe, line 35: Moved to a new file for brevity

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
```diff
@@ -48,12 +49,22 @@ export const DeviceCompromised = () => {
         if (hashCheckError) reportCheckFail('hash', { ...commonData, hashCheckError });
     }, [revisionCheckError, hashCheckError, revision, vendor, version]);
 
+    const isSoftFailure = revisionCheckError === null && hashCheckError === 'other-error';
```
Lemonexe, line 52: A countefeit device could deliberately crash hash check plus fail revision check, so if revision check did not pass, it has precedence (:arrow_right: display normal ghost modal)

packages/suite/src/components/suite/SecurityCheck/checklistItems.tsx
```diff
@@ -0,0 +1,31 @@
+import { Translation } from '../Translation';
+
+export const hardFailureChecklistItems = [
```
peter-sanderson: Maybe declarative typing instead of `as const`?

peter-sanderson: `: SecurityChecklistItem[]`?

peter-sanderson (approved):

packages/suite/src/components/suite/SecurityCheck/checklistItems.tsx
Lemonexe: oh yeah, fixed. As we discussed yesterday the use-cases for `as const`, this really isn't one, I mindlessly copied old code..

PR #15153, opened 2024-10-31
fix(suite-native): device state migration

martykan:
## Description

Add migration for device.state in Suite native's persist reducer and adjust the type in a few remaining places. 
 
## Related Issue

Follow-up to #15064

### DIFF
--- a/suite-native/storage/src/tests/migrations/deviceV2.test.ts
+++ b/suite-native/storage/src/tests/migrations/deviceV2.test.ts
@@ -1,0 +1,63 @@
+import { migrateDeviceState } from '../../migrations/device/v2';
+
+describe('migrateDeviceState', () => {
+    it('should migrate old devices without state to new format', () => {
+        const oldDevices = [
+            {
+                state: undefined,
+                otherKey: 'otherValue',
+            },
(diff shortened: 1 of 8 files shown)


matejkriz (approved):
Looks good and works as expected. I also tried to check if staticSessionId is used everywhere it should and it's a bit tricky with the account.deviceState, but haven't found any missed places 👏

suite-native/storage/src/tests/migrations/deviceV2.test.ts
```diff
@@ -0,0 +1,63 @@
...
+describe('migrateDeviceState', () => {
+    it('should migrate old devices without state to new format', () => {
+        const oldDevices = [
+            {
+                state: undefined,
+                otherKey: 'otherValue',
+            },
+        ];
+
+        const migratedAccounts = migrateDeviceState(oldDevices);
```
vytick, line 12: why the name migratedAccounts in the whole file?

martykan: Copied from another test and forgot to change it 😄

PR #15156, opened 2024-10-31
Enable 'prefer-const'

peter-sanderson:
I think we don't need to disable this rule. In whole code-base, we have only one situation where we need to do it (`setTimer` with `clear`).


komret:
I wish this could be done without the disabling comment but could not come up with anything.

There are some more line which needs to be fixed, see validation.

Nodonisko (approved):

komret (approved):

PR #15158, opened 2024-10-31
chore: add recommanded checks from eslint-plugin-jest

peter-sanderson:
We are using `eslint-plugin-jest` but we are not using any of the rules from it. So I tried to add `plugin:jest/recommended` and this is the result. 

Mostly:
- do not export from test files
- use canonical `toThrow` instead of `toThrowError`
- safer usage of promises
- duplicates of the description text
- duplicate tests

There were some issues that I just silenced by `// eslint-disable` as they were in some complicated/old code. The rule is there for new tests and the old one, we can fix someday or never.

Namely it was:
- usage of Jasmine in Karma
- skipped and commented tests
- some `expects` out of the `it` scope

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,6 +25,7 @@ module.exports = {
         'plugin:react/recommended',
         'plugin:import/typescript',
         'plugin:react-hooks/recommended',
+        'plugin:jest/recommended',
     ],
     settings: {
         react: {
@@ -303,6 +304,16 @@ module.exports = {
             'error',
             { packageName: '@trezor/product-components' },
         ],
+
+        // Jest plugin config
+        'jest/valid-title': 'off', // This rule does not use Typescript and produces false positives
+        'jest/no-disabled-tests': 'off', // we still have a lot of skipped tests (WIP)
+        'jest/no-conditional-expect': 'off', // Todo: we shall solve this, this is bad practice
+        'jest/expect-expect': 'off', // Todo: we have test with no assertions, this may be legit but it needs to be checked
+        'jest/no-standalone-expect': [
+            'error',
+            { additionalTestBlockFunctions: ['conditionalTest'] },
+        ],
     },
     overrides: [
         {
@@ -367,6 +378,18 @@ module.exports = {
                 'import/no-extraneous-dependencies': 'off',
                 'import/no-unresolved': 'off',
                 'import/no-default-export': 'off',
+            },
+        },
+        {
+            files: ['packages/suite-web/e2e/**/*'],
+            rules: {
+                'jest/valid-expect': 'off', // Cypress
+            },
+        },
+        {
+            files: ['packages/connect/e2e/**/*'],
+            rules: {
+                'jest/no-jasmine-globals': 'off', // Kamma tests
             },
         },
     ],
--- a/packages/blockchain-link/tests/unit/connection.test.ts
+++ b/packages/blockchain-link/tests/unit/connection.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-jasmine-globals */
 import { BackendWebsocketServerMock } from '@trezor/e2e-utils';
 import workers from './worker';
 import BlockchainLink from '../../src';
@@ -127,27 +128,29 @@ workers.forEach(instance => {
             expect(server.getFixtures()!.length).toEqual(2);
         });
 
-        it('Handle connect event', done => {
-            blockchain.on('connected', done);
-            blockchain.connect().then(result => {
-                expect(result).toEqual(true);
-            });
-        });
+        it('Handle connect event', () =>
+            new Promise<void>(done => {
+                blockchain.on('connected', done);
+                blockchain.connect().then(result => {
+                    expect(result).toEqual(true);
+                });
+            }));
 
-        it('Handle disconnect event', done => {
-            blockchain.on('disconnected', done);
-            blockchain.connect().then(() => {
-                // TODO: ripple-lib throws error when disconnect is called immediately
-                // investigate more, use setTimeout as a workaround
-                // Error [ERR_UNHANDLED_ERROR]: Unhandled error. (websocket)
-                // at Connection.RippleAPI.connection.on (../../node_modules/ripple-lib/src/api.ts:133:14)
-                if (instance.name === 'ripple') {
-                    setTimeout(() => blockchain.disconnect(), 100);
-                } else {
-                    blockchain.disconnect();
-                }
-            });
-        });
+        it('Handle disconnect event', () =>
+            new Promise<void>(done => {
+                blockchain.on('disconnected', done);
+                blockchain.connect().then(() => {
+                    // TODO: ripple-lib throws error when disconnect is called immediately
+                    // investigate more, use setTimeout as a workaround
+                    // Error [ERR_UNHANDLED_ERROR]: Unhandled error. (websocket)
+                    // at Connection.RippleAPI.connection.on (../../node_modules/ripple-lib/src/api.ts:133:14)
+                    if (instance.name === 'ripple') {
+                        setTimeout(() => blockchain.disconnect(), 100);
+                    } else {
+                        blockchain.disconnect();
+                    }
+                });
+            }));
 
         it('Connect (only one endpoint is valid)', async () => {
             // blockfrost has only one valid endpoint
--- a/packages/blockchain-link/tests/unit/workers.test.ts
+++ b/packages/blockchain-link/tests/unit/workers.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-jasmine-globals */
 import TinyWorker from 'tiny-worker';
 import BlockchainLink from '../../src';
 
--- a/packages/coinjoin/tests/client/inputRegistration.test.ts
+++ b/packages/coinjoin/tests/client/inputRegistration.test.ts
@@ -342,15 +342,16 @@ describe('inputRegistration', () => {
         expect(spy.mock.calls.length).toBeGreaterThan(0);
 
         // 2. wait for confirmation interval to resolve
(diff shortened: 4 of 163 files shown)


packages/connect/e2e/karma.setup.js
```diff
@@ -1,4 +1,4 @@
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000;
+jest.setTimeout(30000);
```
peter-sanderson: :eyes: ?

.eslintrc.js
```diff
@@ -369,5 +382,17 @@ module.exports = {
...
             },
         },
+        {
+            files: ['packages/suite-web/e2e/**/*'],
+            rules: {
+                'jest/valid-expect': 'off', // Cypress
+            },
+        },
+        {
+            files: ['packages/connect/e2e/**/*'],
```
peter-sanderson, line 390: When upgrading to ESLint 9, I will move this to the config to the `packages/connect/e2e` local config. This will be dropped anyway.

packages/blockchain-link/tests/unit/connection.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-jasmine-globals */
```
peter-sanderson, line 1: This is due to usage of `fail`. It is out-of-scope of this PR. We can solve it separately.

packages/blockchain-link/tests/unit/workers.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-jasmine-globals */
```
peter-sanderson, line 1: This is due to usage of `fail`. It is out-of-scope of this PR. We can solve it separately.

packages/connect/src/api/bitcoin/__tests__/Fees.test.ts
```diff
@@ -110,6 +110,7 @@ describe('api/bitcoin/Fees', () => {
 
     // const e2eNetworks = ['BTC', 'TEST', 'BCH', 'BTG', 'DASH', 'DGB', 'DOGE', 'LTC', 'NMC', 'VTC'];
     // e2eNetworks.forEach(network => {
+    // eslint-disable-next-line jest/no-commented-out-tests
```
peter-sanderson, line 113: We can discuss if we just shall delete them all separately. Currently I assume there may be some important information in there and I leave them here.

packages/transport-test/e2e/bridge/bridge.test.ts
```diff
@@ -23,6 +23,7 @@ describe('bridge', () => {
 
         const enumerateResult = await bridge.enumerate();
         assertSuccess(enumerateResult);
+        // eslint-disable-next-line jest/no-standalone-expect
```
peter-sanderson, line 26: `expect` in beforeAll ... I am not gonna fixing that now... :exploding_head:

packages/suite/src/reducers/wallet/__tests__/cardanoStakingReducer.test.ts
```diff
@@ -162,28 +162,6 @@ describe('cardanoStakingReducer reducer', () => {
         });
     });
 
-    it('CARDANO_STAKING.SET_FETCH_ERROR mainnet', () => {
```
peter-sanderson, line 165: duplicated test 1:1 with one above it

packages/coinjoin/tests/client/inputRegistration.test.ts
```diff
@@ -342,6 +342,7 @@ describe('inputRegistration', () => {
         expect(spy.mock.calls.length).toBeGreaterThan(0);
 
         // 2. wait for confirmation interval to resolve
+        // eslint-disable-next-line jest/valid-expect-in-promise
```
peter-sanderson: @marekrjpolak `await` here crashes the test on timeout, but I think it shall be here? Will you please look into it?

packages/utils/tests/createLazy.test.ts
```diff
@@ -62,8 +62,8 @@ describe('createLazy', () => {
 
         jest.advanceTimersByTime(400);
 
-        expect(initPromise).rejects.toThrow('Disposed during initialization');
         expect(lazy.getPending()).toEqual(undefined);
         expect(lazy.get()).toEqual(undefined);
+        await expect(initPromise).rejects.toThrow('Disposed during initialization');
```
peter-sanderson, line 67: @marekrjpolak :pray: review this

packages/coinjoin/tests/client/inputRegistration.test.ts
marekrjpolak: Not my code, but it seems that `input.clearConfirmationInterval()` must be called right after `input.getConfirmationInterval()` in order to resolve it, therefore it can't be awaited. How 'bout something like this? Will it satisfy the eslint rule?

```typescript
        // 2. wait for confirmation interval to resolve
        const promises = Promise.all(
            response.inputs.map(input => input.getConfirmationInterval()?.promise),
        );

        // 1. abort confirmation interval
        response.inputs.forEach(input => input.clearConfirmationInterval());

        const res = await promises;
        res.forEach(input => {
            expect(input?.getConfirmationInterval()).toBeUndefined();
        });
```

packages/utils/tests/createLazy.test.ts
marekrjpolak: This is correct, I believe.

.eslintrc.js
```diff
@@ -303,6 +304,18 @@ module.exports = {
             'error',
             { packageName: '@trezor/product-components' },
         ],
+
+        // Jest plugin config
+        'jest/valid-title': 'off', // This rule does not use Typescript and produces false positives
+        'jest/valid-describe-callback': 'off', // This rule does not use Typescript and produces false positives
+        'jest/no-disabled-tests': 'off', // Well, what can I say... ¯\_(ツ)_/¯ We skip tests sometimes.
```
Lemonexe: nit: it's public repo, so I'd write a less "personal" comment `// we still have a lot of skipped tests (WIP)`

.eslintrc.js
```diff
@@ -303,6 +304,18 @@ module.exports = {
...
+        'jest/no-focused-tests': 'off', // Same as above, but // Todo: shall be easy to get rid of this
```
Lemonexe: Please, could you do it in this PR, it has only [one occurence](https://github.com/trezor/trezor-suite/blob/f9047a6258eba6694776548211a714534c5c5fd8/suite-common/suite-utils/src/__tests__/build.test.ts#L14)?
It is clearly forgotten, because its unneeded, test passees without `.only`
Clear demonstration that this rule is useful :+1:

packages/connect/src/api/bitcoin/__tests__/Fees.test.ts
Lemonexe: Sad but true. Though ideally we should be consistent - either `//` or `.skip`.
It's just a question of lesser evil, but if we have to deal with it, I am definitely fan of `.skip`, commented code doesn't belong in repo.
Though no need to do it in this PR imho :shrug:

packages/utils/tests/getWeakRandomInt.test.ts
```diff
@@ -6,12 +6,7 @@ describe(getWeakRandomInt.name, () => {
...
-        expect(() => getWeakRandomInt(0, -1)).toThrowError(EXPECTED_ERROR);
-    });
-
-    it('returns same value when range is trivial', () => {
-        expect(getWeakRandomInt(0, 1)).toEqual(0);
-        expect(getWeakRandomInt(100, 101)).toEqual(100);
+        expect(() => getWeakRandomInt(0, -1)).toThrow(EXPECTED_ERROR);
     });
 
     it('returns same value when range is trivial', () => {
```
Lemonexe, line 12: Nit: this label should be changed, it fits the `it` you have deleted. The `it` below is not trivial range.

Lemonexe (approved):
Looks like a great improvement, thank you :rocket: 
There is a lot of room for more, but the PR makes sense with its current scope :+1:

.eslintrc.js
```diff
@@ -369,5 +380,17 @@ module.exports = {
...
+        {
+            files: ['packages/suite-web/e2e/**/*'],
+            rules: {
+                'jest/valid-expect': 'off', // Cypress
+            },
+        },
+        {
+            files: ['packages/connect/e2e/**/*'],
+            rules: {
+                'jest/no-jasmine-globals': 'off', // Kamma tests
```
komret, line 392:
```suggestion
                'jest/no-jasmine-globals': 'off', // Karma tests
```

marekrjpolak (approved):

suite-common/test-utils/src/conditionalDescribe.ts
```diff
@@ -4,8 +4,10 @@ export const conditionalDescribe = (
     fn: jest.EmptyFunction,
 ) => {
     if (skipCondition) {
+        // eslint-disable-next-line jest/valid-describe-callback
```
komret, line 7: Seems to be the same case as `jest/valid-title`, I'd disable this rule globally.

komret (changes requested):

packages/utils/tests/getWeakRandomInt.test.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/15273

.eslintrc.js
peter-sanderson: https://github.com/trezor/trezor-suite/pull/15273

suite-common/test-utils/src/conditionalDescribe.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/15273

PR #15160, opened 2024-10-31
fix(suite-native): update gorhom bottom sheet to get rid of warning

vytick:
By updating `@gorhom/bottom-sheet` we get rid of unwanted warning

So you should no longer see:
```
If you don't want to see this message, you can disable the `strict` mode. Refer to:
https://docs.swmansion.com/react-native-reanimated/docs/debugging/logger-configuration for more details.
 WARN  [Reanimated] Reading from `value` during component render. Please ensure that you do not access the `value` property or use `get` method of a shared value while React is rendering a component.
```


yanascz (approved):
Haven’t tried.

PR #15166, opened 2024-11-01
fix(suite-native): send minimal amount validation

PeKne:
## Description

While fetching a minimal amount of each fee level, we were filtering out values of `type: 'non-final'` that are returned if the recipient address is still not known. For this reason, the validation did not work as expected in the form without the filled-in receive address.


yanascz (approved):
Works fine! 👍

PeKne:
QA notes: Please test that it is possible to fill in a send amount while the address field is blank. No error should be displayed if the amount is lower than an account balance.

STew790:
**QA OK**

image

**Info**
24.11.1 https://github.com/trezor/trezor-suite/commit/ff16b788c576e89bbbfc17cbe783154223440336

PR #15170, opened 2024-11-01
feat(suite-native): enhance layout of the account import summary screen

### LINKED ISSUE #14065: Make the button "Confirm" in syncing coins sticky to the bottom of the screen
When user wants to sync coin with tokens, button with Confirming it is hidden at the bottom and user must scroll. 
Please change the behaviour so that it is more accessible for users. 

Acceptance criteria:
When user syncs the coin with and is on the page where the coin must be confirmed. Button "Confirm" is sticked to the bottom of the screen so that it's always accessible for the user.

yanascz:
The Confirm button fixed to the bottom of the screen.

## Related Issue

Resolve #14065


PeKne:
I tested it on the iPhone 13 mini, and it works nicely!

One thing I don't like is that if you start filling in the account label input, the submit button is hidden behind the keyboard :/. Sometimes, even the input is hidden behind it. Can we change that? 

https://github.com/user-attachments/assets/b8b3f648-5690-4d60-8273-67c0b81c9ebe

But it was probably like that, even with the old version, right? So maybe let's just create an issue and fix it separately.

yanascz:
> One thing I don't like is that if you start filling in the account label input, the submit button is hidden behind the keyboard :/. Sometimes, even the input is hidden behind it. Can we change that?

I used `KeyboardAvoidingView` for the summary footer as in `XpubHint`. It's not ideal but at least it works the same.

PeKne (approved):

PeKne:
Tested on both platforms 👍

PR #15171, opened 2024-11-01
fix(suite-native): CryptoIcon in send

vytick:
after two different PR merged, CryptoIcon props in send need to be fixed


yanascz (approved):

PeKne:
Thanks for the fix, @vytick 🙏🫡

PR #15172, opened 2024-11-01
chore(ci): check existing EAS builds before creating a new one

### LINKED ISSUE #14467: Build app from PR

matejkriz:
## Description

- runtimeVersion should ensure the compatibility, we should not create a new builds with the same runtimeVersion.
- Listing two latest successful builds is fragile, it can list i.e. two android builds if ios had some issue, but should be good enough for now.

## Related Issue

Resolve [#14467](https://github.com/trezor/trezor-suite/issues/14467)

### DIFF
--- a/.github/workflows/build-suite-native-preview.yml
+++ b/.github/workflows/build-suite-native-preview.yml
@@ -8,9 +8,11 @@ on:
     types: [opened, synchronize, labeled]
     paths:
       - "suite-native/**"
+      - "suite-common/**"
       - "packages/react-native-usb/**"
       - "packages/transport-native/**"
       - "yarn.lock"
+      - ".github/workflows/build-suite-native-preview.yml"
       # list of paths is not complete, but it's always possible to dispatch manually using 'build-mobile' label
   workflow_dispatch:
     # manual dispatch will not add any comment to PR, use label 'build-mobile' if PR exists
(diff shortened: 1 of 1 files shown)


matejkriz:
I just realized that build-mobile label is not working as expected, because it is still limited to PRs touching whitelisted paths (the path condition also applies for the labeled event) 🤦

.github/workflows/build-suite-native-preview.yml
```diff
@@ -44,7 +46,26 @@ jobs:
...
         run: yarn workspaces focus @suite-native/app
 
+      - name: Check runtimeVersion builds
+        run: |
+          cd suite-native/app
+          # Read runtimeVersion from dynamic app.json
+          RUNTIME_VERSION=$(npx expo config --json | jq -r '.runtimeVersion')
+          echo "Current runtimeVersion is $RUNTIME_VERSION"
+          # Check if there is a build with the same runtimeVersion on EAS already
+          EXISTING_BUILD=$(eas build:list --status=finished  --build-profile=preview --channel=preview --limit=1  --non-interactive --runtime-version=$RUNTIME_VERSION)
```
Nodonisko, line 56: Why not list like 20 of them and simply `grep` in it?

Nodonisko (approved):

.github/workflows/build-suite-native-preview.yml
matejkriz: What would be the benefit? There is a filter by --runtime-version, so I believe that 1 existing build is sufficient.

Nodonisko: Oh I didn't noticed that filter then it's fine

PR #15174, opened 2024-11-01
feat(suite-native): for only passphrase remembered add also standard wallet

### LINKED ISSUE #12722: Passphrase: When Standard wallet not remembered
When only passphrase wallet is remembered and device is connected, we keep the user in the initially visible passphrase wallet, instead of switching him to standard (which might be empty). 

As a workaround to allow users to  actually open it, we will display the "Standard wallet" when connected. It would show no balance in the switcher at first and discovery would happen only after the user would open it with it. 

After that, it'd be normally loaded, balance would appear in wallet switcher and user could effortlessly switch back & forth. 

[Design here](https://www.figma.com/design/mmDnYtctfjnreWhTjLfeQH/Passphrase?node-id=2978-47942&t=zSlSsyD2w6IXuXa3-11)

vytick:
Passphrase: When Standard wallet not remembered add it when trezor is connected.

## Related Issue

Resolve #12722

### DIFF
--- a/suite-native/device-manager/src/components/WalletItem.tsx
+++ b/suite-native/device-manager/src/components/WalletItem.tsx
@@ -6,9 +6,17 @@ import { Translation } from '@suite-native/intl';
 import { Icon } from '@suite-native/icons';
 import { TrezorDevice } from '@suite-common/suite-types';
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
-import { selectDevice, selectDeviceByState } from '@suite-common/wallet-core';
+import {
+    AccountsRootState,
+    DeviceRootState,
+    FiatRatesRootState,
+    selectDevice,
+    selectDeviceByState,
+    selectDeviceHasAccountsByDeviceState,
+} from '@suite-common/wallet-core';
 import { FiatAmountFormatter } from '@suite-native/formatters';
 import { selectDeviceTotalFiatBalanceNative } from '@suite-native/device';
+import { SettingsSliceRootState } from '@suite-native/settings';
 
 type WalletItemProps = {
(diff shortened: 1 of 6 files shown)


yanascz (approved):
Would be great for the standard wallet to be placed before the passphrase wallets. Otherwise it works fine! 👍

matejkriz (changes requested):
The zero balance should be hidden before it's selected and so discovered.

>  It would show no balance in the switcher at first and discovery would happen only after the user would open it with it.

vytick:
> The zero balance should be hidden before it's selected and so discovered.
> 

6c2047ad66

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -54,6 +58,9 @@ const labelStyle = prepareNativeStyle(() => ({
 export const WalletItem = ({ deviceState, onPress, isSelectable = true }: WalletItemProps) => {
     const { applyStyle } = useNativeStyles();
     const device = useSelector((state: any) => selectDeviceByState(state, deviceState));
+    const hasAccounts = useSelector((state: any) =>
+        device?.state ? selectDeviceHasAccountsByDeviceState(state, device.state) : false,
```
matejkriz: Could you please put the check for device.state existence inside the selector itself? 

And I don't like the `any` type usage in this file. I know it doesn't add a much type safety to use root states in useSelector, but it's easy to do (you can import all the root state types besides the selectors) and it's always good to avoid `any` if possible.

vytick: 8eed4d1661

matejkriz (approved):

PR #15175, opened 2024-11-03
fix: sessions sharedworker correctly shared for all deployments

mroz22:


packages/transport/src/transports/webusb.browser.ts
```diff
@@ -4,11 +4,10 @@ import { UsbApi } from '../api/usb';
 
 import { BrowserSessionsBackground } from '../sessions/background-browser';
 
+console.log('process.env.ASSET_PREFIX', process.env.ASSET_PREFIX);
```
martykan: forgotten console log?

martykan (approved):
Seems to work

PR #15179, opened 2024-11-04
feat(suite-native): send address checksum

### LINKED ISSUE #14933: Send Ethereum - address checksum
## Acceptance criteria 
- When the user inserts the address in the non-checksum format:
    1. When a user inserts an already used address that is not checksummed at all, conversion to checksum is done automatically, and the user is informed about that action.
    2. When the user inserts an address that has not been used yet and is not checksummed, he is informed that the address is not in valid format (is not checksummed) and that he should convert it to checksum. The user can’t send the transaction without a checksummed address.
    3. When the user inserts an invalid address where the checksum is not possible (no matter whether it is used or not). The user is informed that the address is invalid

### Design
- [Figma](https://www.figma.com/design/sqKFZAIvk3zERxiWgthK8K/Send?node-id=2208-7212&t=4jNMD8UvChwkIvDF-4)

### Dev note
You can get inspired by the [desktop implementation](https://github.com/trezor/trezor-suite/blob/07fff7930b73d02eb3e0bf68cda3439973e468c6/packages/suite/src/views/wallet/send/Outputs/Address.tsx).

PeKne:
## Description
- address checksum functionality added for Ethereum-based accounts
- if the address has not been used yet, an explanation alert is displayed (see the issue description for more details).
## Related Issue

Resolve #14933 

## Screenshots:

https://github.com/user-attachments/assets/014ab9fa-a181-4b6a-a409-520d28fec205

### DIFF
--- a/suite-native/module-send/src/hooks/useAddressValidationAlerts.tsx
+++ b/suite-native/module-send/src/hooks/useAddressValidationAlerts.tsx
@@ -1,0 +1,138 @@
+import { useCallback, useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
+
+import { useRoute, RouteProp } from '@react-navigation/native';
+import { checkAddressCheckSum, toChecksumAddress } from 'web3-utils';
+import { G } from '@mobily/ts-belt';
+
+import { SendStackParamList, SendStackRoutes } from '@suite-native/navigation';
+import { Translation } from '@suite-native/intl';
+import { selectAccountTokenSymbol, TokensRootState } from '@suite-native/tokens';
+import { useAlert } from '@suite-native/alerts';
+import { useFormContext } from '@suite-native/forms';
+import { isAddressValid } from '@suite-common/wallet-utils';
+import { AccountsRootState, selectAccountNetworkSymbol } from '@suite-common/wallet-core';
+import TrezorConnect from '@trezor/connect';
+import { Link } from '@suite-native/link';
+
+import { getOutputFieldName } from '../utils';
+import { TokenOfNetworkAlertBody } from '../components/TokenOfNetworkAlertContent';
+
+type UseAddressValidationAlertsArgs = {
+    inputIndex: number;
+};
+
+const CHECKSUM_LINK_URL = 'https://trezor.io/learn/a/evm-address-checksum-in-trezor-suite';
+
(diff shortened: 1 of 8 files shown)


suite-native/module-send/src/hooks/useAddressValidationAlerts.tsx
```diff
@@ -0,0 +1,138 @@
...
+                // Already used addresses are checksumed without displaying the alert.
+                const isUsedAddress = addressInfo.payload.history.total !== 0;
+                if (isUsedAddress) {
+                    convertAddressToChecksum();
+
+                    return;
+                }
+            }
+
+            showAlert({
```
vytick, line 78: if call to `getAccountInfo` fails, it shows the alert. We might not want this, wdyt?

PeKne: Good thinking. But this is a design requirement. If we are unable to get information if the address was user or not, the bottom sheet should be displayed. It works the same in desktop implementation. The only difference is that they do not use a bottom sheet but a banner instead.

suite-native/module-send/src/hooks/useAddressValidationAlerts.tsx
```diff
@@ -0,0 +1,138 @@
+import { useCallback, useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
+
+import { useRoute, RouteProp } from '@react-navigation/native';
+import { checkAddressCheckSum, toChecksumAddress } from 'web3-utils';
```
vytick, line 5: fyi: `checkAddressCheckSum` is marked deprecated

vytick (approved):
Works nicely 👌. Just one concern about deprecated method

suite-native/module-send/src/hooks/useAddressValidationAlerts.tsx
PeKne: I will let the desktop trends team know. [The web3-utils is their dependency](https://github.com/trezor/trezor-suite/blob/1b02b0cd14a4d2ece15bbd5b6019cc04b031cb8f/scripts/list-outdated-dependencies/trends-dependencies.txt#L7). So I will leave this up to them how to resolve it.

PR #15185, opened 2024-11-04
feat(suite): autostart in bridge requested modal

martykan:
## Description

Add a CTA for the autostart setting into the bridge requested modal. 
Also fix an error in the bridge deeplink handling.

## Screenshots:
Screenshot 2024-11-04 at 13 04 58

### DIFF
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -9232,6 +9232,11 @@ export default defineMessages({
         defaultMessage:
             "Are you sure? Your device can only be used by one app at a time. If you're currently using another app with your Trezor device, finish that session first.",
     },
+    TR_BRIDGE_TIP_AUTOSTART: {
+        id: 'TR_BRIDGE_TIP_AUTOSTART',
+        defaultMessage:
+            'Tip: You can enable the auto-start feature and have the Bridge always running in the background.',
+    },
     TR_BRIDGE_NEEDED_DESCRIPTION: {
         id: 'TR_BRIDGE_NEEDED_DESCRIPTION',
         defaultMessage:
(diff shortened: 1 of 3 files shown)


packages/suite/src/support/messages.ts
```diff
@@ -9232,6 +9232,11 @@ export default defineMessages({
         defaultMessage:
             "Are you sure? Your device can only be used by one app at a time. If you're currently using another app with your Trezor device, finish that session first.",
     },
+    TR_BRIDGE_TIP_AUTOSTART: {
+        id: 'TR_BRIDGE_TIP_AUTOSTART',
+        defaultMessage:
+            'Tip: You can enable the auto-start feature and have the Bridge always running in the background.',
```
mroz22, line 9238: this is for @Hannsek I am not sure we want to be mentioning `Bridge` anywhere.

packages/suite/src/views/suite/bridge-requested/index.tsx
```diff
@@ -93,6 +97,21 @@ export const BridgeRequested = () => {
         >
             <Metadata title="Bridge | Trezor Suite" />
             <StyledImage image="CONNECT_DEVICE" width="360" />
+
+            {isDesktop() && isDevEnv && (
```
mroz22: `(isDesktop() &&  isDevEnv) || isDebugEnabled` maybe?

mroz22 (approved):

packages/suite/src/support/messages.ts
Hannsek: That is a good quesiton. I don't have the answer for that right now.

PR #15194, opened 2024-11-04
fix(suite): WelcomeLayout gap on top

martykan:
## Description

My personal pet peeve

There is a gap due to a wrapper for message system banner, even if there is no message.

## Screenshots:
Before
Screenshot 2024-11-04 at 15 33 50
After
Screenshot 2024-11-04 at 15 34 06


mroz22 (approved):
ok from my POV. isn't there somebody else with an opinion?

martykan:
Maybe @jvaclavik but he's off 😄

PR #15201, opened 2024-11-04
fix(connect): skip fw hash check when binary too tiny (DX)

Lemonexe:
## Description

When FW binary is only a git LFS pointer (which is the case on all web dev deployments), FW hash is mismatched so you get Ghost modal even with legit physical Trezor - [example](https://dev.suite.sldev.cz/suite-web/chore/bump-mission-deps/web/).

:arrow_right: ignore the FW hash check for too small binary files

:eyes: You can verify [on this dev deployment](https://dev.suite.sldev.cz/suite-web/fix/fw-hash-check-with-lfs-pointer/web/)

:information_source: This is DX only thing, won't happen in production

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -773,8 +773,14 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
         const btcOnly = this.firmwareType === FirmwareType.BitcoinOnly;
         const binary = await getBinaryOptional({ baseUrl, btcOnly, release });
+        // release was found, but not its binary - happens on desktop, where only local files are searched
         if (binary === null) {
-            // release was found, but not its binary - happens on desktop, where only local files are searched
+            return createFailResult('check-unsupported');
+        }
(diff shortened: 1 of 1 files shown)


packages/connect/src/device/Device.ts
```diff
@@ -773,8 +773,9 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
         const btcOnly = this.firmwareType === FirmwareType.BitcoinOnly;
         const binary = await getBinaryOptional({ baseUrl, btcOnly, release });
-        if (binary === null) {
+        if (binary === null || binary.byteLength < 200) {
```
mroz22: I would personally be more explicit, something like 

```
if (binary.byteLenght < 200) {
  log('skipping too small);
...
}
```

otherwise 👍

mroz22 (approved):

packages/connect/src/device/Device.ts
Lemonexe: Done :heavy_check_mark:

PR #15144, opened 2024-10-31
Upgrade ESLint and all plugis & revalidate the config and ALL of the rules

peter-sanderson:
I have done:
- Removed all ESLint we had
- Reintroduced the new verison
- Manually initialized ESLint for whole repository 
- Manually **one-by-one** added/off all the rules that we had different
- I also tried to write comments to them and explained most of the configuration

------------

- Useful tool to migrate old config to new configs: `npx @eslint/migrate-config <path-to>/.eslintrc.js`

**To solve:**
- [x] ESLint 9 does not support sub-config files problem https://github.com/eslint/eslint/discussions/18574 -> solved by `-flag unstable_config_lookup_from_file`

**To check we have all rules that we explicitly ON**
- [x] `@typescript-eslint/prefer-ts-expect-error` deprecated for https://typescript-eslint.io/rules/ban-ts-comment and part of recommended
- [x] `func-style` (re-added into `jestConfig.js`)
- [x] `no-case-declarations` Part of recommended 
- [x] `react/jsx-filename-extensio` - added
- [x] `import/order` - added
- [x] `import/no-extraneous-dependencies` - added
- [x] `no-nested-ternary` - added
- [x] `import/no-anonymous-default-export'` - added
- [x] `react/jsx-curly-brace-presence` - added
- [x] `no-console`
- [x] `react-hooks/rules-of-hooks` - added
- [x] `react-hooks/exhaustive-deps` - added
- [x] `@typescript-eslint/no-useless-constructor` - added (From TS Strict)
- [x] `@typescript-eslint/no-use-before-define` - added
- [x] `require-await` - added
- [x] `react/jsx-key` - in recomended
- [x] `prefer-destructuring` - added

Node:
- [x] `handle-callback-err`
- [x] `no-mixed-requires`
- [x] `no-new-require`
- [x] `no-path-concat`
- [x] `no-process-exit`
- [x] `no-restricted-modules`
- [x] `eol-last`

...

- [x] `import/no-default-export` - added
- [x] `no-label-var` - added
- [ ] `@typescript-eslint/no-shadow` - :exclamation: :exclamation: <-- TODO: this is an issue
- [x] `no-shadow-restricted-names` - already in recommended 
- [x] `no-undef-init` - added
- [x] `no-unused-expressions` - added (and from `0` -> `error`) no ida what `0` means
- [ ] `chai-friendly/no-unused-expressions` <-- :exclamation: TODO :exclamation: 
[...]

PR #15202, opened 2024-11-04
chore: add no-unsafe-optional-chaining as it became part of recommanded settings

peter-sanderson:
https://eslint.org/docs/latest/rules/no-unsafe-optional-chaining became part of recommended settings

Prep, for https://github.com/trezor/trezor-suite/pull/15144

### DIFF
--- a/packages/components/src/components/Dropdown/Dropdown.tsx
+++ b/packages/components/src/components/Dropdown/Dropdown.tsx
@@ -124,7 +124,7 @@ export const Dropdown = forwardRef(
                 return;
             }
 
-            const { width, height } = menuRef.current?.getBoundingClientRect();
+            const { width, height } = menuRef.current.getBoundingClientRect();
 
             const adjustedCoords = getAdjustedCoords({
                 coords: coordsToUse,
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -124,7 +124,7 @@ export const Dropdown = forwardRef(
                 return;
             }
 
-            const { width, height } = menuRef.current?.getBoundingClientRect();
```
peter-sanderson, line 127: Its checked above, `!menuRef.current` => early return, the `?` not needed

packages/components/src/components/typography/TruncateWithTooltip/TruncateWithTooltip.tsx
```diff
@@ -29,8 +29,13 @@ export const TruncateWithTooltip = ({
     useEffect(() => {
         if (!containerRef.current || !scrollWidth || !scrollHeight) return;
         const resizeObserver = new ResizeObserver(entries => {
-            const { inlineSize: elementWidth, blockSize: elementHeight } =
-                entries[0].borderBoxSize?.[0];
+            const borderBoxSize = entries[0].borderBoxSize?.[0];
+            if (!borderBoxSize) {
```
peter-sanderson, line 33: This probably never happens, but I wasn't sure

Lemonexe: Can't tell from documentation :/ But in any case, it's not worse than previous code, which throws error if `!borderBoxSize`

Lemonexe (approved):
LGTM

PR #15203, opened 2024-11-04
feat(suite-desktop-core): tray icon

martykan:
## Description

Add a tray icon showing status of the bridge while running in background. 
It can be enabled/disabled in Debug settings. 
Currently it displays status of the bridge, but in the future it should show connected devices as well, but for that to work properly we need to move `TrezorConnect.init` to the desktop core process.

## Screenshots

Screenshot 2024-11-05 at 09 42 16


packages/suite-desktop-core/src/modules/tray.ts
```diff
@@ -0,0 +1,146 @@
...
+            },*/
+            {
+                type: 'separator',
+            },
+            {
+                label: 'Launch Trezor Suite',
+                click: () => mainThreadEmitter.emit('app/show'),
+            },
+            {
+                label: 'Stop Trezor Bridge and Quit',
```
mroz22, line 79: Again, I am not sure if we want to use the term "Trezor Bridge" at all. but this one is on @Hannsek

mroz22 (approved):
# :shipit:

PR #15220, opened 2024-11-05
fix: missing devDependency

peter-sanderson:
`import * as messages from '@trezor/protobuf/messages.json';`

is used in that package

image


mroz22 (approved):

mroz22:
how comes it passed the checks?

peter-sanderson:
@mroz22 I haven investigated, but in branch with ESLint9 it triggered the check.

PR #15225, opened 2024-11-05
fix: local-rules/no-override-ds-component were not riggered correcly due to incorrect configuration

peter-sanderson:


seibei-iguchi (approved):

eslint-local-rules/rules.ts
```diff
@@ -19,31 +19,36 @@ export default {
...
-                        packageName: {
-                            type: 'string',
+                        packageNames: {
+                            type: 'array',
+                            items: { type: 'string' },
+                            minItems: 1,
                         },
                     },
-                    additionalProperties: false,
+                    // additionalProperties: false,
```
komret: Should be deleted?

PR #15235, opened 2024-11-05
fix(e2e): Fixed rest of viewports in currently enabled tests

### LINKED ISSUE #15148: Fix viewport in E2E web tests
Viewport used in E2E tests for web makes no sense. It is used only because proper handling of scrolling to elements is not done. We need to be testing on some standard viewport. This is wrong from testing POV as well as usability of the Cypress app POV.

HajekOndrej:
## Description

Fixed more viewports. 

## Related Issue

Resolve #15148


mroz22 (changes requested):
please squash 🙏

mroz22 (approved):

PR #15240, opened 2024-11-05
fix(suite-desktop-core): trezor connect module reload

martykan:
## Description

Fix an issue with disposing TrezorConnect when reloading the window with Cmd+R


packages/suite-desktop-core/src/modules/trezor-connect.ts
```diff
@@ -64,3 +61,12 @@ export const initBackground: ModuleInitBackground = ({ store }: Pick<Dependencie
 
     return { onLoad, onQuit };
 };
+
+export const init: ModuleInit = () => {
```
Lemonexe, line 65: :warning: `ModuleInit` is not defined

Lemonexe (approved):
I admit I do not fully understand the cause, but I can see
- the bug is resolved :heavy_check_mark: 
- `init` method was present before yesterday

After resolving typescript error, LGTM..

PR #15241, opened 2024-11-05
chore: enable no-constant-binary-expression as it becames recommanded default in new ESLint

peter-sanderson:
https://eslint.org/docs/latest/rules/no-constant-binary-expression will became default in recommanded config

### DIFF
--- a/packages/suite/src/components/wallet/Pagination.tsx
+++ b/packages/suite/src/components/wallet/Pagination.tsx
@@ -102,7 +102,7 @@ export const Pagination = ({
                     <PageItem
                         key={i}
                         data-testid={`@wallet/accounts/pagination/${i}`}
-                        data-test-activated={i === currentPage ?? 'true'}
+                        data-test-activated={i === currentPage}
                         onClick={() => onPageSelected(i)}
                         $isActive={i === currentPage}
                     >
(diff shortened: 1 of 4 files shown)


packages/suite/src/components/wallet/Pagination.tsx
```diff
@@ -99,7 +99,7 @@ export const Pagination = ({
                     <PageItem
                         key={i}
                         data-testid={`@wallet/accounts/pagination/${i}`}
-                        data-test-activated={i === currentPage ?? 'true'}
```
peter-sanderson, line 105: I think this is always `true|false` and never result in `'true'` (string).

The `??` works like this:
image

packages/suite/src/views/dashboard/PortfolioCard/PortfolioCard.tsx
```diff
@@ -77,8 +77,8 @@ export const PortfolioCard = memo(() => {
     }
 
     const isWalletEmpty = !discoveryStatus && isDeviceEmpty;
-    const isWalletLoading = discoveryStatus?.status === 'loading' ?? false;
```
peter-sanderson, line 82: same here, even if `discoveryStatus?.status` is undefined, the `=== 'loading` turns it into `false` and it never fallback to the `??`

packages/suite/src/views/wallet/coinmarket/common/CoinmarketBalance.tsx
```diff
@@ -48,7 +48,7 @@ export const CoinmarketBalance = ({
     });
 
     if (showOnlyAmount) {
-        if (!balance ?? isNaN(Number(balance))) return null;
+        if (balance === undefined || isNaN(Number(balance))) return null;
```
peter-sanderson: Discussed with @adderpositive

Lemonexe (approved):

PR #15243, opened 2024-11-05
Connect data manager cleanup

mroz22:
# 🧹 

removing some no-longer-useful indirections


martykan (approved):
Looks good, are we aiming to get rid of it everywhere?

mroz22:
not sure yet. I am now researching how we would create coin-specific connect modules and these were the first problems I met - that every file imports from everywhere so splitting code into modules is quite difficult now.

martykan:
I guess to solve that we will have to start moving things into `connect-common` and `connect` will just be a facade/wrapper.

PR #15258, opened 2024-11-06
chore(connect): remove rebootToBootloader api method

mroz22:
# 🧹 

we don't need TrezorConnect.rebootToBootloader anymore. We don't use it and it was never part of public API.

edit:
or maybe it could be useful for something? reboot to bootloader from suite to allow factory reset, or something like this? cc @Hannsek


Hannsek:
Yes, factory reset, also it is useful for installing custom firmware.

How big trouble is it for us to have this in our code.

mroz22:
But we don't use it anywhere at the moment - on TrezorConnect API level. We only use it internally which stays. Are we going to use it as a single, standalone connect call?

Hannsek:
Sorry, I don't follow. Are you saying that it would stay in Suite, but it won't be possible for 3rd party apps to call this method?

marekrjpolak (approved):
No added line -> big green checkmark from me. Provided that nobody has a problem with removing it ofc.

PR #15261, opened 2024-11-06
chore: add no-async-promise-executor as it becames recommanded

peter-sanderson:
https://eslint.org/docs/latest/rules/no-async-promise-executor

I just ignored 2 occurrences we have so far. This way we can fix them later and we are warn to not create another.

image

### DIFF
--- a/packages/suite-desktop-core/e2e/support/regtest.ts
+++ b/packages/suite-desktop-core/e2e/support/regtest.ts
@@ -16,33 +16,30 @@ export const generateBlock = () =>
         method: 'GET',
     });
 
-export const waitForCoinjoinBackend = () =>
-    // Todo: refactor to not use async-promise-executor
-    // eslint-disable-next-line no-async-promise-executor
-    new Promise<void>(async (resolve, reject) => {
-        const limit = 60;
-        const error = '';
+export const waitForCoinjoinBackend = async () => {
+    const limit = 60;
+    const error = '';
 
-        console.log('waiting for coinjoin backend');
+    console.log('waiting for coinjoin backend');
 
-        for (let i = 0; i < limit; i++) {
-            if (i === limit - 1) {
-                console.log(`waiting for coinjoin backend: ${error}\n`);
-            }
-
-            await new Promise(resolve => setTimeout(() => resolve(undefined), 1000));
-
-            try {
-                const res = await fetch('http://localhost:19121/');
-                if (res.status === 200) {
-                    console.log('coinjoin backend is online');
-
-                    return resolve();
-                }
-            } catch {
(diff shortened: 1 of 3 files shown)


packages/suite-desktop-core/e2e/support/regtest.ts
```diff
@@ -17,6 +17,8 @@ export const generateBlock = () =>
     });
 
 export const waitForCoinjoinBackend = () =>
+    // Todo: refactor to not use async-promise-executor
+    // eslint-disable-next-line no-async-promise-executor
```
Lemonexe: Is it that much work to rewrite these two functions?
If I am not mistaken, this is low hanging fruit, not huge effort to do

```
export const waitForCoinjoinBackend = async  () => {
       const limit = 60;
        const error = '';
        ...
        return; instead of resolve()
        throw; instead of reject()
}
```

packages/trezor-user-env-link/src/websocket-client.ts
```diff
@@ -268,7 +268,7 @@ export class WebsocketClient extends TypedEmitter<WebsocketClientEvents> {
     }
 
     waitForTrezorUserEnv() {
-        return new Promise<void>(async (resolve, reject) => {
+        return async () => {
```
Lemonexe: Thank you, `regtest.ts` looks perfect, but here you accidentally changed it to HOF.
I meant just this, without the `return async () => {`:

```
    async waitForTrezorUserEnv() {
        // unfortunately, it can take incredibly long for trezor-user-env to start, we should
        // do something about it
        const limit = 300;
        let error = '';

        console.log('waiting for trezor-user-env');

        for (let i = 0; i < limit; i++) {
            if (i === limit - 1) {
                console.log(`cant connect to trezor-user-env: ${error}\n`);
            }
            await delay(1000);

            try {
                const res = await fetch(USER_ENV_URL.DASHBOARD);
                if (res.status === 200) {
                    console.log('trezor-user-env is online');

                    return;
                }
            } catch (err) {
                error = err.message;
                // using process.stdout.write instead of console.log since the latter always prints also newline
                // but in karma, this code runs in browser and process is not available.
                if (typeof process !== 'undefined') {
                    process.stdout.write('.');
                } else {
                    console.log('.');
                }
            }
        }

        throw error;
    }
```

Lemonexe (changes requested):
Currently broken..

packages/trezor-user-env-link/src/websocket-client.ts
peter-sanderson: ojoj, shit, 

thank you for catching it

Lemonexe (approved):

PR #15262, opened 2024-11-06
chore(suite): fix aborted discovery process in wallet switcher

### LINKED ISSUE #13396: Opening device switcher immediately on startup ("Click to discover wallet")
**Describe the bug**
When opening the device switcher immediately after starting Suite, before discovery starts, it goes into a state where it doesn't load the accounts and only shows "Click to discover wallet", which actually doesn't do anything. 

**Steps to reproduce:**
1. Have a Trezor connected and set up, but don't enable view-only
2. Re-start Suite
3. Immediately click on the device switcher menu

**Info:**
 - Suite version: desktop 24.7.2 (3282fe8f6f491e024c0cc9cae99f3085c229c96c)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.7.2 Chrome/118.0.5993.159 Electron/27.3.8 Safari/537.36
 - OS: MacIntel
 - Screen: 1800x1169
 - Device: Trezor T2B1 2.8.0 regular (revision dd4671a5104952ef505d28d1f9e94d1484b4607a)

**Expected result:**
I click on "Click to discover wallet" and it restarts discovery.

**Actual result:**
Nothing happens upon clicking. 

**Screenshots:**
Screenshot 2024-07-17 at 14 37 17

seibei-iguchi:


packages/components/src/components/Card/Card.tsx
```diff
@@ -60,6 +62,23 @@ const CardContainer = styled.div<
         border-color 0.2s;
     cursor: ${({ $isClickable }) => ($isClickable ? 'pointer' : 'default')};
 
+    ${({ theme, $isHiglighted, $paddingType }) =>
+        $isHiglighted &&
+        css`
+            overflow: hidden;
+            padding-left: calc(${spacingsPx.xxs} + ${mapPaddingTypeToPadding({ $paddingType })});
+
+            &::before {
```
peter-sanderson, line 71: This looks a bit :magic_wand: magic, but if its the only way...

seibei-iguchi: No magic, this is the (css) way:D

packages/components/src/components/Card/utils.tsx
```diff
@@ -30,9 +31,9 @@ export const mapPaddingTypeToLabelPadding = ({ $paddingType }: PaddingMapArgs):
     return paddingMap[$paddingType];
 };
 
-export const mapPaddingTypeToPadding = ({ $paddingType }: PaddingMapArgs): number | string => {
-    const paddingMap: Record<PaddingType, number | string> = {
-        none: 0,
+export const mapPaddingTypeToPadding = ({ $paddingType }: PaddingMapArgs): SpacingPxValues => {
+    const paddingMap: Record<PaddingType, SpacingPxValues> = {
+        none: '0px',
```
peter-sanderson, line 36: I would rather have just one way of doing stuff. So I think we have `SpacingValues` everywhere now.

packages/suite/src/views/suite/SwitchDevice/DeviceItem/WalletInstance.tsx
```diff
@@ -133,107 +79,121 @@ export const WalletInstance = ({
...
+    const handleClick = discoveryProcess
+        ? () => {
+              if (!editing) {
+                  dispatch(selectDeviceThunk({ device: instance }));
+                  dispatch(redirectAfterWalletSelectedThunk());
+                  onCancel(false);
+              }
+          }
+        : () => {
+              device &&
```
peter-sanderson: please lets not use && as a call-chain operator, I think it should be one with good-old if

peter-sanderson (approved):
Just some small things, also discussed in person

tomasklim (approved):

PR #15263, opened 2024-11-06
feat(suite-native): only download token definitions for relevant networks

vytick:
- Only definitions for enabled networks and networks of imported accounts are fetched.
 
## Related Issue

Part of #14399


suite-native/discovery/src/discoveryConfigSlice.ts
```diff
@@ -193,6 +197,15 @@ export const selectDeviceEnabledDiscoveryNetworkSymbols = memoizeWithArgs(
     { size: 2 },
 );
 
+export const selectTokenDefinitionsEnabledNetworks = memoize(
```
yanascz, line 200: Looking at the selector name, it's not obvious what is being selected... Token definitions or enabled networks? 🤔

vytick: I just kept the name used elsewhere

yanascz: It's confusing but if it's used elsewhere... 😏

yanascz (approved):
Seems to work fine! 💪

PR #15271, opened 2024-11-06
chore: no-control-regex as it becames recommanded

peter-sanderson:
https://eslint.org/docs/latest/rules/no-control-regex

image


packages/utils/src/asciiUtils.ts
```diff
@@ -1,4 +1,5 @@
 // Regular expression to match non-ASCII characters
+// eslint-disable-next-line no-control-regex
 const nonAsciiPattern = /[^\x00-\x7F]/g;
```
komret: Let's avoid the control characters instead:

```suggestion
const nonAsciiPattern = /[^\x20-\x7E]/g;
```

komret (changes requested):

komret (approved):

PR #15276, opened 2024-11-06
fix(connect): emit changed event with state change in keepSession

### LINKED ISSUE #15226: Suite needs passphrase input for every action after reconnecting to view only wallet
**Describe the bug**
When I reconnect my device to Suite and I have my wallet in view-only mode, Suite wants passphrase input for "every" action there is and not only once per session.

**Steps to reproduce:**
1. Connect device with passphrase
2. Enable view only for the wallet
3. Disconnect device
4. Reconnect device
5. Go to any account and select receive address
6. Suite asks for passphrase
7. Now go to any other receive address to show it
8. Suite asks for passphrase again

**Info:**
 - Suite version: web 24.11.0 (b7afbd05152736013d619dfd9dc9f10143fbd216)
 - OS: Linux x86_64
 - Screen: 1920x1080

**Expected result:**
Suite wants passphrase input only once per session.

**Actual result:**
Suite wants passphrase input for "every" action there is and not only once per session.

martykan:
## Description

We update `sessionId` in saved device state in Suite from receiving a `DEVICE.CHANGED` event.
However this event wasn't being sent for calls with `keepSession: true`, such as during account discovery, this meant that if the passphrase was entered during those calls, it wasn't saved.

## Related Issue

Resolve #15226

### DIFF
--- a/packages/connect/src/core/index.ts
+++ b/packages/connect/src/core/index.ts
@@ -696,6 +696,15 @@ const onCallDevice = async (
         }
         // Work done
 
+        if (
+            method.keepSession &&
+            method.deviceState &&
+            method.deviceState.sessionId !== device.getState()?.sessionId
+        ) {
+            // if session was changed from the one that was sent, send a device changed event
+            sendCoreMessage(createDeviceMessage(DEVICE.CHANGED, device.toMessageObject()));
+        }
+
         // TODO: This requires a massive refactoring https://github.com/trezor/trezor-suite/issues/5323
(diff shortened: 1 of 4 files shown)


mroz22:
btw what about `skipFinalReload`? Does it affect this case too?

martykan:
Cardano doesn't work so I assume there is an issue with `skipFinalReload`

martykan:
Nevermind, it's a different issue with Cardano, due to the device state being current on subsequent calls

mroz22 (approved):
looks legit

PR #15281, opened 2024-11-06
feat(message-system): add ab testing message

### LINKED ISSUE #15067: AB Testing capability
We’d like to have an ability to run experiments in Trezor Suite, using Feature Flags.

### Experiment Configuration
Experiments would use Message System to config Feature Flags (i.e. there will be no UI to config experiments, deploy of the messaging system will be required - independed from Suite release though).

https://github.com/trezor/trezor-suite/blob/develop/docs/features/message-system.md 

This needs to be extended to allow configuration:

- Experiment ID (e.g. rename-trade-section)
- Multiple Variation IDs (e.g. control, buy-swap) - this may be unlimited, but in reality we won’t use more than 4 for one experiment
- Each would have distribution percentage set between 0 - 100%

### Traffic distribution
Distribution would be based on mathematical function applied to instanceID.

- number will be generated from instanceID
- we need to be able to edit test configuration and change the % - ideally this should only extend / reduce existing pool of test segment accounts, not randomize it completely

### Tracking, etc.
Anything else - tracking, logic, will be implemented in the Suite as part of each experiment/variation. The reporting will use IDs from the Message System config.

### LINKED ISSUE #15279: Extend message system to include AB testing variant
## Description 
- Experiments would use Message System to configure Feature Flags (i.e. there will be no UI to config experiments, deploy of the messaging system will be required - independent from Suite release though).
- Experiment ID (e.g. rename-trade-section) 
   - Multiple Variation IDs (e.g. control, buy-swap) - this may be unlimited, but in reality, we won’t use more than 4 for one experiment
   - Each would have a distribution percentage set between 0 - 100%

adderpositive:
## Description
- Experiments would use a Message System to configure Feature Flags (i.e. there will be no UI to config experiments, deployment of the messaging system will be required - independent from Suite release though).
- Experiment ID (e.g. rename-trade-section) 
   - Multiple Variation IDs (e.g. control, buy-swap) - this may be unlimited, but in reality, we won’t use more than 4 for one experiment
   - Each would have an experiment distribution percentage between 0 - 100%

## Example of experiments
`experiments` property is placed in the top level of _config.v[version].json_

```json
  "experiments": [
      {
          "conditions": [
              {
                  "environment": {
                      "desktop": ">=24.5.1",
                      "mobile": "!",
                      "web": ">=24.5.1"
                  }
              }
          ],
          "experiment": {
              "id": "3bed56a4-ecd8-4e0f-2910-014b484c2aff",
              "groups": [
                  {
                      "variant": "A",
                      "percentage": 30
                  },
                  {
                      "variant": "B",
                      "percentage": 70
                  }
              ]
          }
      }
  ]
```
- `conditions` are the same property as for the action - https://github.com/trezor/trezor-suite/blob/develop/docs/features/message-system.md
- `experiment` has required parameters `id` and `groups`
  - `id` must be unique, but can also be used for nice naming (e.g. "button-testing")
  - `groups` as an array of testing groups - minimum length of items is two
    - `percentage` is a number between 0 and 100, a sum of all percentages has to be 100
    - `variant` can be any string - name of the group
- updated documentation - https://github.com/trezor/trezor-suite/blob/feat/ab-testing-on-message-system/docs/features/message-system.md

## TODO
- [ ] update _config.v1.json_ on the develop and stable - add `experiments` property

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15279
Resolve https://github.com/trezor/trezor-suite/issues/15067

## Testing

This PR was tested separately. More information you can find here - https://github.com/trezor/trezor-suite/pull/16145 and https://www.notion.so/satoshilabs/A-B-testing-170dc526060680ab9d5af4e5c9d87876

### DIFF
--- a/docs/features/message-system.md
+++ b/docs/features/message-system.md
@@ -8,7 +8,9 @@ Message system was implemented to allow sending emergency messages to Trezor Sui
 
 [Notion for production deployment](https://www.notion.so/satoshilabs/Message-system-production-release-c0ac4275461f4755bc9050ff2de57425)
 
-## Types of in-app messages
+## Features of message system
+
+### Messages
 
 There are multiple ways of displaying message to a user:
 
@@ -18,8 +20,14 @@ There are multiple ways of displaying message to a user:
     -   _TODO: missing implementation_
 -   context
     -   messages on specific places in app (e.g. settings page, banner in account page)
--   feature
-    -   disabling some feature with an explanation message
+
+### Feature
+
+Disabling some feature with an explanation message.
+
+### A/B testing
+
+The ability to create A/B testing.
 
 ## Implementation
 
@@ -80,7 +88,7 @@ If changes made to the message system are incompatible with the previous version
 
 Structure of config, types and optionality of specific keys can be found in the schema or in generated types. Example config is commented below.
 
-```javascript
+```json
 {
     // Version of message system implementation. Bump if new version is not backward compatible.
     "version": 1,
@@ -118,11 +126,7 @@ Structure of config, types and optionality of specific keys can be found in the 
                     - Options: gte, lt, ranges, tildes, carets,... are supported, see semver lib for more info.
                     */
                     "os": {
-                        "macos": [
-                            "10.14",
-                            "10.18",
-                            "11"
-                        ],
+                        "macos": ["10.14", "10.18", "11"],
                         "linux": "*",
                         "windows": "!",
                         "android": "*",
@@ -137,18 +141,12 @@ Structure of config, types and optionality of specific keys can be found in the 
                         "revision": "7281ac61483e38d974625c2505bfe5efd519aacb"
                     },
                     "browser": {
-                        "firefox": [
-                            "82",
-                            "83"
-                        ],
+                        "firefox": ["82", "83"],
                         "chrome": "*",
                         "chromium": "!"
                     },
                     "transport": {
-                        "bridge": [
-                            "2.0.30",
-                            "2.0.27"
-                        ],
+                        "bridge": ["2.0.30", "2.0.27"],
                         "webusbplugin": "*"
                     },
                     /*
@@ -255,16 +253,48 @@ Structure of config, types and optionality of specific keys can be found in the 
                 },
                 // Used only for context.
                 "context": {
-                    "domain": [
-                        "coins.receive",
-                        "coins.btc"
-                  ]
-                }
-                 // Used only for feature
+                    "domain": ["coins.receive", "coins.btc"]
+                },
+                // Used only for feature
                 "feature": [
                     {
                         "domain": "coinjoin",
                         "flag": false
+                    }
+                ]
+            }
+        }
+    ],
+    // Used for AB testing
+    "experiments": [
+        {
+            // Same as for actions, check conditions above
+            "conditions": [
+                {
+                    "environment": {
+                        "desktop": ">=24.5.1",
+                        "mobile": "!",
+                        "web": ">=24.5.1"
+                    }
+                }
+            ],
+            // Detail of an experiment
+            "experiment": {
+                // Used as a selector for the test
+                // For `id` use UUID and then use mapping in `suite-common/message-system/src/experiment/experiments.ts`
+                "id": "e2e8d05f-1469-4e47-9ab0-53544e5cad07",
+                // Array of testing groups - minimum length of items is two
+                // The sum of group percentages must equal 100 in total
+                "groups": [
+                    {
+                        // Name of variant - any string
+                        "variant": "A",
+                        // The percentage range a user can be assigned to (0-100)
+                        "percentage": 30
+                    },
+                    {
+                        "variant": "B",
+                        "percentage": 70
                     }
                 ]
             }
@@ -287,6 +317,14 @@ Current priorities of existing banners can be found [here](https://github.com/tr
 #### Targeting Linux version
 
 Unfortunately, it is not possible to target specific distributions and versions of Linux. It is possible to only target all Linux users using `*` or exclude all Linux users using `!`.
+
+#### Experiment implementation
+
+The experiment is based on the `instanceId` used by analytics. It will remain active regardless of user consent. However, no data will be sent to analytics without user consent.
+
+##### Component experiment
+
+To create an experiment for components, use ExperimentWrapper.tsx. It requires the `id` of the test and the `components` to be tested as properties. Check the implementation details for more information [here](https://github.com/trezor/trezor-suite/blob/feat/ab-testing-on-message-system/packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx).
 
 ### Application steps
 
--- a/packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
+++ b/packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
@@ -1,0 +1,40 @@
+import { ReactElement } from 'react';
+
+import { ExperimentNameType } from '@suite-common/message-system';
+
+import { useExperiment } from 'src/hooks/experiment/useExperiment';
+
+interface ExperimentWrapperProps {
+    id: ExperimentNameType;
+    components: Array<{
+        variant: string;
+        element: ReactElement;
+    }>;
+}
+
+/**
+ * @param components first item in components is default
+ */
+export const ExperimentWrapper = ({
+    id,
+    components,
+}: ExperimentWrapperProps): ReactElement | null => {
+    const { experiment, activeExperimentVariant } = useExperiment(id);
+    const areComponentsEmpty = !components.length;
+
+    if (areComponentsEmpty) return null;
+
+    const defaultComponent = components[0];
+    const experimentOrVariantNotFound = !experiment || !activeExperimentVariant;
+    const experimentAndComponentsMismatch = experiment?.groups.length !== components.length;
+
+    if (experimentOrVariantNotFound || experimentAndComponentsMismatch) {
+        return defaultComponent.element;
+    }
+
+    const activeComponent = components.find(
+        component => component.variant === activeExperimentVariant.variant,
+    );
+
+    return activeComponent?.element ?? defaultComponent.element;
+};
--- a/packages/suite/src/hooks/experiment/useExperiment.ts
+++ b/packages/suite/src/hooks/experiment/useExperiment.ts
@@ -1,0 +1,26 @@
+import { useMemo } from 'react';
+
+import { selectAnalyticsInstanceId } from '@suite-common/analytics';
+import {
+    ExperimentNameType,
+    experiments,
+    selectActiveExperimentGroup,
+    selectExperimentById,
+} from '@suite-common/message-system';
+
+import { useSelector } from 'src/hooks/suite';
+
+export const useExperiment = (id: ExperimentNameType) => {
+    const experimentUuid = experiments[id];
+    const instanceId = useSelector(selectAnalyticsInstanceId);
+    const experiment = useSelector(selectExperimentById(experimentUuid));
+    const activeExperimentVariant = useMemo(
+        () => selectActiveExperimentGroup({ instanceId, experiment }),
+        [instanceId, experiment],
+    );
+
+    return {
+        experiment,
+        activeExperimentVariant,
+    };
+};
--- a/packages/suite/src/middlewares/suite/messageSystemMiddleware.ts
+++ b/packages/suite/src/middlewares/suite/messageSystemMiddleware.ts
@@ -6,6 +6,7 @@ import {
     messageSystemActions,
     categorizeMessages,
     getValidMessages,
+    getValidExperiments,
 } from '@suite-common/message-system';
 
 import { SUITE } from 'src/actions/suite/constants';
@@ -43,10 +44,19 @@ const messageSystemMiddleware =
                     enabledNetworks,
                 },
             });
-
             const categorizedValidMessages = categorizeMessages(validMessages);
 
+            const validExperiments = getValidExperiments(config, {
+                device,
+                transport,
+                settings: {
+                    tor: getIsTorEnabled(torStatus),
+                    enabledNetworks,
+                },
+            }).map(item => item.id);
+
             api.dispatch(messageSystemActions.updateValidMessages(categorizedValidMessages));
+            api.dispatch(messageSystemActions.updateValidExperiments(validExperiments));
         }
 
         return action;
--- a/suite-common/message-system/schema/config.schema.v1.json
+++ b/suite-common/message-system/schema/config.schema.v1.json
@@ -30,186 +30,7 @@
                 "required": ["conditions", "message"],
                 "properties": {
                     "conditions": {
-                        "type": "array",
-                        "items": {
-                            "title": "Condition",
-                            "type": "object",
-                            "properties": {
-                                "duration": {
-                                    "title": "Duration",
-                                    "type": "object",
-                                    "required": ["from", "to"],
-                                    "properties": {
-                                        "from": {
-                                            "$ref": "#/definitions/date-time"
-                                        },
-                                        "to": {
-                                            "$ref": "#/definitions/date-time"
-                                        }
-                                    },
-                                    "additionalProperties": false
-                                },
-                                "os": {
-                                    "title": "Operating System",
-                                    "type": "object",
-                                    "required": [
-                                        "macos",
-                                        "linux",
-                                        "windows",
-                                        "android",
-                                        "ios",
-                                        "chromeos"
-                                    ],
-                                    "properties": {
-                                        "macos": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "linux": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "windows": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "android": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "ios": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "chromeos": {
-                                            "$ref": "#/definitions/version"
-                                        }
-                                    },
-                                    "additionalProperties": {
-                                        "$ref": "#/definitions/version"
-                                    }
-                                },
-                                "environment": {
-                                    "title": "Environment",
-                                    "type": "object",
-                                    "required": ["desktop", "mobile", "web"],
-                                    "properties": {
-                                        "desktop": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "mobile": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "web": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "revision": {
-                                            "type": "string"
-                                        }
-                                    }
-                                },
-                                "browser": {
-                                    "title": "Browser",
-                                    "type": "object",
-                                    "required": ["firefox", "chrome", "chromium"],
-                                    "properties": {
-                                        "firefox": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "chrome": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "chromium": {
-                                            "$ref": "#/definitions/version"
-                                        }
-                                    },
-                                    "additionalProperties": {
-                                        "$ref": "#/definitions/version"
-                                    }
-                                },
-                                "transport": {
-                                    "title": "Transport",
-                                    "type": "object",
-                                    "required": ["bridge", "webusbplugin"],
-                                    "properties": {
-                                        "bridge": {
-                                            "$ref": "#/definitions/version"
-                                        },
-                                        "webusbplugin": {
-                                            "$ref": "#/definitions/version"
-                                        }
-                                    },
-                                    "additionalProperties": {
-                                        "$ref": "#/definitions/version"
-                                    }
-                                },
-                                "settings": {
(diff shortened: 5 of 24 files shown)


packages/suite/src/utils/suite/distribution.ts
```diff
@@ -0,0 +1,64 @@
...
+
+interface DistributionGetGroupByInclusion {
+    groups: DistributionMessageProps['groups'];
+    inclusion: number;
+}
+
+/**
+ * @returns number between 0 and 99 generated from instanceId
+ */
+export const getInclusionFromInstanceId = (instanceId: string) => {
```
matejkriz: The numbers 0 to 95 are slightly more likely to occur than numbers 96 to 99. 🤓 

https://chatgpt.com/share/672e3087-1564-800b-8479-39ce0b483749

But practically, for our use, it should be completely negligible 👌

MiroslavProchazka:
Hello @adderpositive , please let's wait for @tomasklim and let's discuss this in more broader context. Personally I do not like the idea to mix it with message system, so lets discuss.

adderpositive:
> Hello @adderpositive , please let's wait for @tomasklim and let's discuss this in more broader context. Personally I do not like the idea to mix it with message system, so lets discuss.

Yes, of course. We have been already discussing the extension of the system. I will go ahead and briefly describe it here. 
1. The first solution was implemented directly into the message-system messages. After brainstorming, I have refactored it to the solution in the PR.
2. The next solution is to rename the system - _message-system_ can be confusing if it manages more functionality than messages only. Examples of some names _toggle-system_, _push-system_, ...
3. But @matejkriz had an excellent point of view the message system should provide a solution how to safely and unforgettably deliver a message to the Suite at runtime - the message can be distinguished as _banner_, _feature_ or _experiment_, etc. It is different than it should display only some messages. 

I hope it is understandable a little bit. :)

adderpositive:
~~Tests are failing because the property `experiments` doesn't exist in the `config.v1.json`. It should also be updated on the _develop_ and _stable_.~~

Fixed

tomasklim:
> Tests are failing because the property experiments doesn't exist in the config.v1.json. It should also be updated on the develop and stable.

Please make it optional. So that the app does not crash when `experiments` are not available at all. We will maybe separate AB testing, feature flags, and message system into 3 systems in the future

docs/features/message-system.md
```diff
@@ -21,6 +23,10 @@ There are multiple ways of displaying message to a user:
 -   feature
     -   disabling some feature with an explanation message
```
tomasklim: this is not `messages`

introduce `## Feature Flags`

docs/features/message-system.md
```diff
@@ -255,16 +251,40 @@ Structure of config, types and optionality of specific keys can be found in the
...
                 // Used only for context.
                 "context": {
-                    "domain": [
-                        "coins.receive",
-                        "coins.btc"
-                  ]
+                    "domain": ["coins.receive", "coins.btc"]
                 }
-                 // Used only for feature
-                "feature": [
```
tomasklim, line 264: is feature completely missing from the example now?

docs/features/message-system.md
```diff
@@ -8,7 +8,9 @@ Message system was implemented to allow sending emergency messages to Trezor Sui
 
 [Notion for production deployment](https://www.notion.so/satoshilabs/Message-system-production-release-c0ac4275461f4755bc9050ff2de57425)
 
-## Types of in-app messages
+## Possibility of message system
```
tomasklim:
```suggestion
## Features of message system
```

docs/features/message-system.md
```diff
@@ -8,7 +8,9 @@ Message system was implemented to allow sending emergency messages to Trezor Sui
...
+
+### Use actions (messages)
```
tomasklim:
```suggestion
### Messages
```

docs/features/message-system.md
```diff
@@ -21,6 +23,10 @@ There are multiple ways of displaying message to a user:
 -   feature
     -   disabling some feature with an explanation message
 
+### Use (Experiments)
```
tomasklim:
```suggestion
### A/B testing
```

docs/features/message-system.md
```diff
@@ -279,7 +299,7 @@ When updating message system config, sequence number must always be higher than
 
 Updated config is automatically uploaded by CI job to the corresponding S3 bucket based on the current branch.
 
-#### Priorities of messages
+#### Priorities of actions (messages)
```
tomasklim: we use priorities only for messages right? so it does not need change

packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @param components last item in components is default
+ */
+export const ExperimentWrapper = ({
+    id,
+    components,
+}: ExperimentWrapperProps): ReactElement | null => {
+    const { experiment, activeExperimentVariant } = useExperiment(id);
+    const areComponentEmpty = !components.length;
```
tomasklim:
```suggestion
    const areComponentsEmpty = !components.length;
```

tomasklim: or just put `!components.length` into condition

packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+    if (areComponentEmpty) return null;
+
+    const defaultComponent = components[components.length - 1];
```
tomasklim: why the last one is default?

packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const experimentOrVariantNotFound = !experiment || !activeExperimentVariant;
+    const experimentAndComponentsMismatch = experiment?.groups.length !== components.length;
+
+    if (experimentOrVariantNotFound || experimentAndComponentsMismatch)
+        return defaultComponent.element;
```
tomasklim:
```suggestion
    if (experimentOrVariantNotFound || experimentAndComponentsMismatch){
        return defaultComponent.element;
        }
```

packages/suite/src/hooks/experiment/useExperiment.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { selectAnalyticsInstanceId } from '@suite-common/analytics';
+import { selectExperimentById } from '@suite-common/message-system';
+
+import { useSelector } from 'src/hooks/suite';
+import { selectActiveExperimentGroup } from 'src/utils/suite/experiment';
+
+export const useExperiment = (id: string) => {
+    const state = useSelector(state => state);
+    const instanceId = selectAnalyticsInstanceId(state);
```
tomasklim:
```suggestion
    const instanceId = useSelector(selectAnalyticsInstanceId);
```

suite-common/message-system/config/config.v1.json
```diff
@@ -1125,5 +1125,6 @@
                 }
             }
         }
-    ]
+    ],
+    "experiments": []
```
tomasklim: app should be running even though `experiments` are not there

suite-common/message-system/schema/config.schema.v1.json
```diff
@@ -346,6 +167,55 @@
...
+                        "description": "Used for AB testing",
+                        "type": "object",
+                        "required": ["id", "groups"],
+                        "properties": {
+                            "id": {
+                                "type": "string"
+                            },
+                            "groups": {
+                                "type": "array",
+                                "minItems": 2,
```
tomasklim: I think not needed

tomasklim (changes requested):

packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
adderpositive: I assumed if any experiment is active, the last test will be always default one.

suite-common/message-system/schema/config.schema.v1.json
adderpositive: It doesn't make sense to have an experiment with zero or one item. What do you think?

packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
adderpositive: Another options is to use default property in the config.

adderpositive:
> > Tests are failing because the property experiments doesn't exist in the config.v1.json. It should also be updated on the develop and stable.
> 
> Please make it optional. So that the app does not crash when `experiments` are not available at all. We will maybe separate AB testing, feature flags, and message system into 3 systems in the future

Ok. In my opinion, it’s a good approach to separate messages, features, experiments, etc., but only on the application layer. On the data layer (_config.vx.jsw_), splitting the config into more parts would make maintenance difficult and confusing.

suite-common/message-system/schema/config.schema.v1.json
```diff
@@ -346,6 +167,54 @@
...
+                                            "description": "The name of the variant, e.g., 'A' or 'B'"
+                                        },
+                                        "percentage": {
+                                            "type": "number",
+                                            "minimum": 0,
+                                            "maximum": 100,
+                                            "description": "Percentage of users for this variant (0-100)"
+                                        }
+                                    },
+                                    "additionalProperties": false
```
matejkriz: I would definitely put this to `true` to allow arbitrary properties for future experiments. Suite should ignore them if it doesn't know them.

docs/features/message-system.md
```diff
@@ -269,6 +264,40 @@ Structure of config, types and optionality of specific keys can be found in the
...
+                    "environment": {
+                        "desktop": ">=24.5.1",
+                        "mobile": "!",
+                        "web": ">=24.5.1"
+                    }
+                }
+            ],
+            // Detail of an experiment
+            "experiment": {
+                // Used as a selector for the test
```
matejkriz, line 283: Please remind me, have we considered some human-readable name for the experiments and use it as a reference in the code instead of the id? Or do we want to put human-readable ids here? Or do we want to use the id as random uuid and then define some map to human-readable keys in the Suite code? I guess we should document the decision here in the docs.

packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
matejkriz: This is only fallback logic for this component, that should basically never be user, right? We should always pass components for every variant. In that case I think it's ok to have it like this or change it for the first component (I would prefer that).

packages/suite/src/utils/suite/experiment.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        currentPercentage += group.percentage;
+
+        return result;
+    });
+
+    return extendedExperiment.find(
+        group => group.range[0] <= inclusion && group.range[1] >= inclusion,
+    )?.group;
+};
```
matejkriz: This way, same user will always have the same value (same number between 0 and 99) for all the experiments, right? Because instanceId is persistent. 

I’m not sure, but I can imagine, it might affect the validity of the experiments if they are not completely independent. This is because the same users will always see the first variant in all experiments.

matejkriz: The other option is to mix the experiment id somehow with the instanceId, so we still have constant value for one experiment, but it's different across other experiments.

docs/features/message-system.md
adderpositive: I prefer to use random uuid to make it unique without map of human-readable keys in the Suite. Map of keys will add more complexity, with the need to wait for releases.

adderpositive: There was a request to make it human-readable, but it will require more control from the person who will add it.

packages/suite/src/utils/suite/experiment.ts
adderpositive: Very good point.

suite-common/message-system/schema/config.schema.v1.json
matejkriz: And how we will handle the experiments that have the clear winner? I guess we would need to keep that experiment for older Suite versions with only one group and 100%. We can't just remove the experiment if the winner is not the default option.

docs/features/message-system.md
matejkriz: For the map I mean simple constant somewhere in the codebase that can be used instead of the uuid like 

```javascript
const experiments = {
 orangeSendButton: 'fb0eb1bc-8ec3-44d4-98eb-53301d73d981',
 somethingElse: 'def30808-589b-4aab-92a2-334b2104e6c4',
} as const;
```

I think that is a no-brainer and we should translate those random ids to something readable and not put the strings everywhere. We can even keep the old experiments there so we will never reuse the same id.

packages/suite/src/utils/suite/experiment.ts
matejkriz, line 1: I would expect this file to be in suite-common as it is reusable for the mobile.

docs/features/message-system.md
adderpositive: I see. Yes, that makes sense.

suite-common/message-system/src/messageSystemUtils.ts
```diff
@@ -234,76 +232,94 @@ export const getValidMessages = (config: MessageSystem | null, options: Options)
...
+        return [];
+    }
+
+    return config.experiments
+        .filter(
+            experiment =>
+                !experiment.conditions.length ||
+                experiment.conditions.some(condition => validateConditions(condition, options)),
+        )
+        .filter(experiment => validateExperiment(experiment.experiment))
```
matejkriz: Is this filter needed here? This should be ensured by the scheme itself. Here I think we want to be as benevolent as possible.

adderpositive: Of course `experiment.groups.length >= 2 &&` is ambiguous, but the schema does not handle control of the sum. Or is there a way how to do it in the schema?

suite-common/message-system/schema/config.schema.v1.json
adderpositive: Group A - 100%, Group B - 0%

packages/suite/src/utils/suite/experiment.ts
adderpositive: Moved to the message-system in [9f28739](https://github.com/trezor/trezor-suite/pull/15281/commits/9f2873926aaa12d0a251229555800cb3db34d33d).

docs/features/message-system.md
adderpositive: I have added the mapping here https://github.com/trezor/trezor-suite/pull/15281/commits/de0fb36bae7f1e391cbbe7f7afa413f21b130b3f and also created relatively strong typing.

suite-common/message-system/schema/config.schema.v1.json
tomasklim: makes sense

docs/features/message-system.md
matejkriz: It's not super consistent with Context (ContextDomain) and Feature (FeatureDomain) naming, but it can work fine 👌

suite-common/message-system/src/messageSystemUtils.ts
matejkriz: I believe that this filter really should go away, it's worse to check it here than no check at all.

The check for sum can be added as a custom validation to the scheme, but not sure if it's necessary. I would rather define some fallback solution in case the sum is not 100.

adderpositive: In this step, I would rather consider the experiment invalid than rely on a less unreadable fallback, such as group percentage calculations. Additionally, the percentages will not match those in the _config.json_ file if they are recalculated in a fallback.

suite-common/message-system/schema/config.schema.v1.json
matejkriz: It really doesn't makes sense to me to require at least 2 items. What is the benefit? The downside is, that if we will have multiple closed experiments, there will be useless groups with 0%.

packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { useExperiment } from 'src/hooks/experiment/useExperiment';
+
+interface ExperimentWrapperProps {
+    id: ExperimentNameType;
+    components: Array<{
+        variant: string;
+        element: ReactElement;
+    }>;
+}
```
tomasklim, line 13:
```suggestion
    experiments: Array<{
        variant: string;
        component: ReactElement;
    }>;
}
```

suite-common/message-system/schema/config.schema.v1.json
adderpositive: It can be deleted but then I would leave control of the sum there as it is. 🙃

packages/suite/src/components/suite/Experiment/ExperimentWrapper.tsx
adderpositive: I would use rather to use 
```
    components: Array<{  
        variant: string;  
        component: ReactElement;  
    }>;  
```
with regard to on using `experiment` variable in `ExperimentWrapper` with different type could lead to misunderstandings.

adderpositive:
@matejkriz After our discussion, I removed strict validation. Additionally, you can see a description below if the sum does not equal 100. I hope this helps with your review. 🙃 Changes here https://github.com/trezor/trezor-suite/pull/15281/commits/c83ddc52bb24ef4302d8a7be81ff3317092d675c

If an array of groups is empty the `defaultComponent` (first in components) will be displayed.
If the sum of the items in the group is more than 100, the percentage created from `instanceId` will assigned to the matching group.
If the sum of the items in the group is less than 100 and the percentage created from `instanceId` does not match any group (e.g. A - 30%, B - 30%, user percentage - 69), it will return `defaultComponent` (first in components).

suite-common/message-system/schema/config.schema.v1.json
adderpositive: Solved: https://github.com/trezor/trezor-suite/commit/c83ddc52bb24ef4302d8a7be81ff3317092d675c

suite-common/message-system/src/messageSystemUtils.ts
adderpositive: Solved: https://github.com/trezor/trezor-suite/commit/c83ddc52bb24ef4302d8a7be81ff3317092d675c

tomasklim (approved):

packages/suite/src/middlewares/suite/messageSystemMiddleware.ts
```diff
@@ -43,10 +44,19 @@ const messageSystemMiddleware =
                     enabledNetworks,
                 },
             });
-
             const categorizedValidMessages = categorizeMessages(validMessages);
 
+            const validExperiments = getValidExperiments(config, {
```
matejkriz, line 49: It's a nitpick and too late, I guess, so feel free to ignore, but since the validation params should be consistent for getValidMessages few lines above and here, it would be nice to deduplicate it (put it in some constant and pass that for example). And the mapping to ids (.map(item => item.id)) can happen inside getValidExperiments.

packages/suite/src/middlewares/suite/messageSystemMiddleware.ts
```diff
@@ -43,10 +44,19 @@ const messageSystemMiddleware =
...
+                device,
+                transport,
+                settings: {
+                    tor: getIsTorEnabled(torStatus),
+                    enabledNetworks,
+                },
+            }).map(item => item.id);
+
             api.dispatch(messageSystemActions.updateValidMessages(categorizedValidMessages));
+            api.dispatch(messageSystemActions.updateValidExperiments(validExperiments));
```
matejkriz, line 59: Sorry for noticing that so late, but now I'm trying to test it on mobile and just realized, that this is only in desktop middleware and it's missing in mobile middleware so it does not support mobile even though it looks like only this part is missing (I believe we don't need ExperimentWrapper for the initial implementation, we can add it together with some real experiments targeting mobile).

packages/suite/src/middlewares/suite/messageSystemMiddleware.ts
tomasklim: Can be resolved after freeze right? So I will merge it and we can change it later

matejkriz: There was no reason to merge it before freeze 🙉 But of course we can do is as a followup...

suite-common/message-system/src/experiment/index.ts
matejkriz, line 1: Using index file is deprecated in our repo with the only exception for one index file on root of the package for exports. I'll prepare a followup PR.

packages/suite/src/middlewares/suite/messageSystemMiddleware.ts
matejkriz: followup here https://github.com/trezor/trezor-suite/pull/16243

suite-common/message-system/src/experiment/index.ts
matejkriz: check https://github.com/trezor/trezor-suite/pull/16243

packages/suite/src/middlewares/suite/messageSystemMiddleware.ts
matejkriz: https://github.com/trezor/trezor-suite/pull/16243

PR #15289, opened 2024-11-06
fix(suite): order of network filter

### LINKED ISSUE #9165: Accounts tab - order of network filter is different than accounts order
**Describe the bug**
The order of networks in the sidebar network filter is now based on the order in which the networks were enabled.

**Info:**

-   Suite Version: 24.8.1

**How to reproduce**

1. Disable all networks in `Settings` in `Coins` section
2. Enable `Litecoin`, `Polygon`, `Dogecoin`, `Bitcoin`
3. Check sidebar order of networks in filter vs order of accounts 
4. `Litecoin`, `Polygon`, `Dogecoin`, `Bitcoin` vs `Bitcoin`, `Polygon`, `Litecoin`, `Dogecoin`

**Expected behavior**
Networks filter should have same order as networks in Settings / accounts tab

**Screenshots**
Screenshot 2024-10-21 at 11 59 35
Screenshot 2024-10-21 at 12 00 07

izmy:
## Description

Fixed order of network filter. The order is now the same as the account order.

## Related Issue

Resolve #9165

### DIFF
--- a/suite-common/wallet-config/src/utils.ts
+++ b/suite-common/wallet-config/src/utils.ts
@@ -14,6 +14,11 @@ export const NORMAL_ACCOUNT_TYPE = 'normal' satisfies AccountType;
  */
 export const networksCollection: Network[] = Object.values(networks);
 
+/**
+ * array of network symbols
+ */
+export const networkSymbolCollection = networksCollection.map(n => n.symbol);
+
 export const getMainnets = (debug = false) =>
     networksCollection.filter(n => !n.testnet && (!n.isDebugOnlyNetwork || debug));
 
--- a/packages/suite/src/actions/settings/__fixtures__/walletSettings.ts
+++ b/packages/suite/src/actions/settings/__fixtures__/walletSettings.ts
@@ -46,7 +46,7 @@ export default [
         initialState: { enabledNetworks: [] },
         action: () => walletSettingsActions.changeNetworks(['ltc', 'eth']),
         result: {
-            enabledNetworks: ['ltc', 'eth'],
+            enabledNetworks: ['eth', 'ltc'],
         },
     },
     {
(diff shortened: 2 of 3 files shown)


suite-common/wallet-config/src/utils.ts
```diff
@@ -14,6 +14,11 @@ export const NORMAL_ACCOUNT_TYPE = 'normal' satisfies AccountType;
  */
 export const networksCollection: Network[] = Object.values(networks);
 
+/**
+ * array of network symbols
+ */
+export const networkSymbolList = networksCollection.map(n => n.symbol);
```
tomasklim: nit:
`networkSymbolCollection` so that it is unified with `networksCollection`

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/CoinsFilter.tsx
```diff
@@ -46,7 +47,9 @@ export const CoinsFilter = () => {
     const device = useSelector(selectDevice);
 
     const unavailableCapabilities = device?.unavailableCapabilities ?? {};
-    const supportedNetworks = enabledNetworks.filter(symbol => !unavailableCapabilities[symbol]);
+    const supportedNetworks = enabledNetworks
```
tomasklim: This solution would fix it just for this one particular case. `enabledNetworks` is used on more places in codebase, how about storing it in redux directly in the right order?

A migration for current users would be nice.

tomasklim (changes requested):
Thanks for your contribution, few things can be improved. Check the proposed changes ⬇️

tomasklim (approved):
LGTM. I've created a PR with your commit to run CI pipelines here #15335
I also added a [migration](https://github.com/trezor/trezor-suite/pull/15289#discussion_r1832931348) so that all old users have correct order.

tomasklim (changes requested):
Unit tests are unfortunately failing https://github.com/trezor/trezor-suite/actions/runs/11775212413/job/32795344436?pr=15335

tomasklim (approved):
Thank you! Approving and closing in favor of https://github.com/trezor/trezor-suite/pull/15335 to have CI checks

PR #15300, opened 2024-11-07
feat(connect): make transportReconnect implicitly true

martykan:
## Description

Make transportReconnect implicitly true, so 3rd parties can reconnect when bridge was not running.

Existing tests affected by this are modified to use `transportReconnect: false`

### DIFF
--- a/packages/connect/e2e/common.setup.ts
+++ b/packages/connect/e2e/common.setup.ts
@@ -127,6 +127,7 @@ export const initTrezorConnect = async (
         debug: false,
         popup: false,
         pendingTransportEvent: true,
+        transportReconnect: false,
         connectSrc: process.env.TREZOR_CONNECT_SRC, // custom source for karma tests
         ...options,
     });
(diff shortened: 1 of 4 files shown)


mroz22:
woudln't it be easier to add 

`transportReconnect: true` default here? 
https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/data/connectSettings.ts#L31

martykan:
Yeah, that's simpler, I changed it. I originally chose the approach of changing it in core, so it doesn't need updating on the 3rd party side, but that shouldn't be necessary anyway since iframe will parse them again and it wasn't previously defaulting to false.

martykan:
I just need to figure out what to do about the tests. With transportReconnect this is called on dispose/cleanup:
https://github.com/trezor/trezor-suite/blob/a4e5832d3c9cfae9506bc58f4c5d931b538767c3/packages/connect/src/device/DeviceList.ts#L398
Which results in an unhandled promise rejection (since no one is listening at that point anymore) and Jest considers that a failure.

martykan:
Ok I think it should be in good shape now, the dispose issue was solved by some changes in DeviceList

mroz22 (approved):
looks ok

PR #15302, opened 2024-11-07
fix(suite-desktop-core): is autostart enabled condition on windows

### LINKED ISSUE #15272: "Start Trezor Suite automatically" button is not clickable on Windows
"Start Trezor Suite automatically" button is not clickable on Windows 10 and 11. 

Observed on latest develop Suite 24.11.0 while testing https://github.com/trezor/trezor-suite/pull/15256

martykan:
## Description

Windows needs a special case for checking if autostart is enabled.

## Related Issue

Resolve #15272


mroz22 (approved):
I can't test it, but it looks meaningful

PR #15304, opened 2024-11-07
[discuss] chore(connect): coinAbstractMethod

mroz22:
I am exploring modularization possibilities. And it of course brings about many ideas for refactoring and improvements.
While thinking (#15291) about how to reduce the amount of data in config.ts - where specifically I don't like the information about [feature support](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/data/config.ts#L109) - IMHO it should be defined per coin-module. 

Would it make sense to define common stuff and stuff like this in a coin specific abstract method? cc @marekrjpolak @karliatto @martykan

### DIFF
--- a/packages/connect/src/api/binance/api/binanceAbstractMethods.ts
+++ b/packages/connect/src/api/binance/api/binanceAbstractMethods.ts
@@ -1,0 +1,21 @@
+import { Capability } from '@trezor/protobuf/src/messages-schema';
+import { AbstractMethod, Payload } from '../../../core/AbstractMethod';
+import { CallMethodPayload } from '../../../events';
+import { getFirmwareRange } from '../../common/paramsValidator';
+import { getMiscNetwork } from '../../../data/coinInfo';
+
+export abstract class BinanceAbstractMethod<
+    Name extends CallMethodPayload['method'],
+    Params = undefined,
+> extends AbstractMethod<Name, Params> {
+    requiredDeviceCapabilities = [Capability.Binance];
+
+    constructor(message: { id?: number; payload: Payload<Name> }) {
+        super(message);
+        this.firmwareRange = getFirmwareRange(this.name, getMiscNetwork('BNB'), this.firmwareRange);
+    }
+
+    // firmwareRangeRequirements could be defined here instead of config.ts
+    // loading coin-specific protobuf could be done here
+    // anything else?
+}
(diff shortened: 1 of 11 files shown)


packages/connect/src/api/binance/api/binanceGetAddress.ts
```diff
@@ -2,28 +2,28 @@
...
 
 type Params = PROTO.BinanceGetAddress & {
     address?: string;
 };
 
-export default class BinanceGetAddress extends AbstractMethod<'binanceGetAddress', Params[]> {
+export default class BinanceGetAddress extends BinanceAbstractMethod<
+    'binanceGetAddress',
+    Params[]
+> {
```
mroz22, line 20: if I give it a shorter name + remove comments, I am optimistic that this might be a line saving solution :D

martykan:
Yeah seems good. Also when I look at GetAddress, GetPublicKey init, it seems like it's mostly the same between all methods

packages/connect/src/api/binance/api/binanceAbstractMethods.ts
```diff
@@ -0,0 +1,21 @@
...
+    Params = undefined,
+> extends AbstractMethod<Name, Params> {
+    requiredDeviceCapabilities = [Capability.Binance];
+
+    constructor(message: { id?: number; payload: Payload<Name> }) {
+        super(message);
+        this.firmwareRange = getFirmwareRange(this.name, getMiscNetwork('BNB'), this.firmwareRange);
+    }
+
+    // firmwareRangeRequirements could be defined here instead of config.ts
```
mroz22: specifically firmwareRangeRequirement for coin. basically we have now 3 levels of fw requirements 
- coin
- method
- method feature

coin could be defined in this abstract method

mroz22:
hmm, ti doesn't work very well, there could be `<coinComposeTranaction>` which is different (has firmware range, doesn't have unavailable capability because it has useDevice: false)

packages/connect/src/core/AbstractMethod.ts
```diff
@@ -341,6 +341,9 @@ export abstract class AbstractMethod<Name extends CallMethodPayload['method'], P
     }
 
     checkDeviceCapability() {
+        if (!this.useDevice) {
+            return true;
+        }
```
mroz22, line 346: if cardanoComposeTransaction, or any method that has useDevice: false should also inherit from a coin specific abstract method, we need to handle cases like this. I am starting to dislike it

PR #15305, opened 2024-11-07
Revert "chore(suite-desktop-core): bump openpgp"

Lemonexe:
This reverts commit a66ece5a64dd1fabe2918634c8e2f4956b7a1605 (downgrades OpenPGP back to 5.11.2)

[See slack convo](https://satoshilabs.slack.com/archives/G019WLX2P7B/p1730986359058219), it broke windows build :scream:
This fixes the windows build :heavy_check_mark: 

Reason not yet known, but I guess that when they rewritten the library to `mjs`, it's not exported as it should, or maybe has hacky code that hardcodes CWD at build time, IDK, I'll have to check.

### DIFF
--- a/packages/suite-desktop-core/package.json
+++ b/packages/suite-desktop-core/package.json
@@ -36,7 +36,7 @@
         "electron-localshortcut": "^3.2.1",
         "electron-store": "8.2.0",
         "electron-updater": "6.3.9",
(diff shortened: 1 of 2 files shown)


martykan (approved):
Confirming this seems to fix it

PR #15308, opened 2024-11-07
Improved transactions for mobile

### LINKED ISSUE #15301: Mobile align transactions list with desktop to show all txns
- [x] show all transactions (we hide some now)
- [x] show native transactions and token txns without filtering (now we have switch in account detail to include token txns)
- [x] extract names for transactions  (used in `TransactionHeader`)

vytick:
- show all transactions, no hiding
- show native transactions and token txns without need to tap the switch
- extract names for transactions (EVMs)
 
## Related Issue

Resolve #15301

## Screenshots:

https://github.com/user-attachments/assets/84e565aa-9cf0-405c-8bd7-0985d95dedd6


vytick:
@shenkys has just told me not to remove the switch filtering normal and token txns until they go through that with product team. I'll put this to draft until then.

vytick:
This task was approved to finalise

suite-native/intl/src/en.ts
```diff
@@ -822,6 +822,22 @@ export const en = {
...
+            pending: 'Pending',
+            contract: 'Contract',
+            self: 'Self',
+            joint: 'Joined',
+            failed: 'Failed',
+            unknown: 'Unknown',
+            withdrawal: 'Rewards withdrawal',
+            stake_delegation: 'Stake delegation',
+            stake_registration: 'Registration of a stake address',
+            stake_deregistration: 'Deregistration of a stake address',
```
yanascz: Why snake case here? I'd go with camelCase as everywhere else in this file.

suite-native/transactions/src/components/TransactionName.tsx
```diff
@@ -0,0 +1,64 @@
...
+            return 'transactions.name.failed';
+        case 'joint':
+            return 'transactions.name.joint';
+        case 'contract':
+            return 'transactions.name.contract';
+        case 'self':
+            return getSelfTransactionMessageByType({
+                type: transaction.cardanoSpecific?.subtype,
+            });
+        case 'unknown':
```
yanascz: Why list `'unknown'` when it's immediately followed by `default`? 🤔

suite-native/transactions/src/components/TransactionName.tsx
```diff
@@ -0,0 +1,64 @@
...
+        default:
+            return isPending ? 'transactions.name.pending' : 'transactions.name.unknown';
+    }
+};
+
+export const TransactionName = ({ transaction, isPending }: TransactionNameProps) => {
+    const ethName = transaction.ethereumSpecific?.parsedData?.name;
+
+    // use name of eth txns, but not for Transfer
+    if (ethName && !(ethName === 'Transfer' && ['recv', 'sent'].includes(transaction.type))) {
```
yanascz: Are you sure about the `&&` operator after `'Transfer'`? 🤔 

Anyway, I'd propagate the negation deeper. Since the original condition is off from my PoV, notice I kept the `&&` operators.
```suggestion
    if (ethName && (ethName !== 'Transfer' && transaction.type !== 'recv' && transaction.type !== 'sent')) {
```

suite-native/tokens/src/tokensSelectors.ts
```diff
@@ -146,85 +145,24 @@ export const selectAnyOfTokensIsKnown = (
...
-        (_state, _accountKey: AccountKey, areTokenOnlyTransactionsIncluded: boolean) =>
-            areTokenOnlyTransactionsIncluded,
-    ],
-    (
-        transactions,
-        tokenDefinitions,
-        areTokenOnlyTransactionsIncluded,
-    ): WalletAccountTransaction[] => {
-        return pipe(
+export const selectAccountTransactionsWithTokensWithFiatRates = createMemoizedSelector(
```
Nodonisko: Can we drop fiat rates from name? It is not valid anymore right?

Nodonisko (approved):
Few nits but looks good

suite-native/intl/src/en.ts
vytick: 32c38f50ee

suite-native/transactions/src/components/TransactionName.tsx
vytick: 32c38f50ee

suite-native/tokens/src/tokensSelectors.ts
vytick: 32c38f50ee

suite-native/transactions/src/components/TransactionName.tsx
vytick: 32c38f50ee

vytick: ah dc0ee8c39f

vytick: It should apply ethName only for txnst that are not Transfer AND (sent || recv)

yanascz (approved):

PR #15324, opened 2024-11-08
Feat/multitransport

### LINKED ISSUE #16322: implement multi-transport support
connect and suite should support communication over 2 different mediums (USB, Bluetooth, udp) at the same time.

marekrjpolak:
## Description

This PR adds support for multiple active transports in parallel, meaning that e.g. it's possible to have both emulator and physical device connected to the same instance of Suite, or to have both usb and bluetooth (when implemented) active at the same time.

Many changes must've been done on Transport, Connect and Suite level, so anything could break. Connect API stood mostly unchanged (mostly transport events) or extended (`setTransports` method) so there should be no issues regarding backwards compatibility.

Only one transport of each `apiType` (= transport medium) can be active at the same time, but all the current transports are of `usb` type except `UdpTransport` (which is `udp`).

 ## Features & fixes (approximately corresponds to commits)

- devices in `DeviceList` are now stored and identified not only by their transport path but also the transport itself (so it would hypothetically be possible to have two devices with colliding path but on different transport medium)
- all the asynchronous transport-related tasks in `DeviceList` (initializing/reconnecting/stopping) are now protected against overlapping (so the latter will abort the former and wait gracefully for it to finish before executing itself)
- running transport instances are now seamlessly reused when `DeviceList` is reinitialized (if this transport is still allowed of course)
- when waiting for yet unidentified devices on transport start, disconnecting the device resolves the promise in the same way as acquiring it (until now, it timeouts)
- newly implemented `TrezorConnect.setTransports` method allows to change the set of enabled transports without needing to reinitialize the whole Connect (= generalized `disableWebUSB` method, but it was preserved as well for the sake of backwards compatibility)
- multitransport implementation (as stated in the description section), and the task overlapping is handled per `apiType` (`usb` transport may be reconnecting at the same time as the `udp` one is stopping)
- Suite (mostly reducers and selectors) must've been adjusted to all of this
- message system was adjusted to allow targetting users based on any of their multiple transports instead of a single one (and also fixed as it was broken until now)

## TODOs

- handle how multiple transports are sent into analytics (https://github.com/trezor/trezor-suite/pull/15324/commits/bc4f10e24686eddc3a8e3cbfa1dba19b9fdacd38)
- some e2e tests would be nice
- ???

## How to test

In debug settings, transport clients section, it's now possible to turn on/off all possible transports for given platform and it changes immediately. So the best way is probably wildly switching them there, manually starting/stopping bridge, connecting/disconnecting physical devices and starting/stopping emulator.

In web Suite there's no udp support so only one transport can be active there, but it should still be possible to switch between bridge and webusb like this.

Everything should work like a charm (which it won't, of course). Please report any inconvenience.

resolve #16322


packages/connect/src/device/DeviceList.ts
```diff
@@ -185,105 +275,144 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> implements IDevic
...
     }
 
-    private async selectTransport([transport, ...rest]: Transport[]): Promise<Transport> {
-        const result = await transport.init();
+    private async selectTransport(
+        [transport, ...rest]: Transport[],
+        signal: AbortSignal,
+    ): Promise<Transport> {
+        if (signal.aborted) throw new Error(signal.reason);
+        if (transport === this.transport) return transport;
```
szymonlesisz: `transport: Transport` will never satisfy `this.transport: Partial<Record<TransportApiType, Transport>>`

packages/connect/src/impl/dynamic.ts
```diff
@@ -114,6 +115,12 @@ export class TrezorConnectDynamic<
         }
     }
 
+    public setTransports({ transports }: SetTransports) {
+        // @ts-expect-error hell knows why this is not working
```
szymonlesisz: because `this.lastSettings` is typed as possible undefined and in that case `manifest` will be missing, the result will be just `this.lastSettings = { transport };`, 

works if you use `{ ...this.lastSettings! }`

szymonlesisz (approved):
im using this commits in [bluetooth](https://github.com/trezor/trezor-suite/tree/feat/rust-bluetooth) branch for a while now and i didn't find any issue 🥇 🔥🔥🔥

packages/connect/src/device/DeviceList.ts
marekrjpolak: This was indeed a bug, thank you! https://github.com/trezor/trezor-suite/pull/15324/commits/54cd8e6aace469fa5951b62499e01c85d8505f70

packages/connect/src/impl/dynamic.ts
marekrjpolak: Right, I thought it's the same case as in `manifest` method, but it wasn't.
https://github.com/trezor/trezor-suite/pull/15324/commits/1e377da0b338d9b5b5cc6c3eed3e4371559e3258

szymonlesisz (approved):

packages/connect-web/src/impl/core-in-iframe.ts
```diff
@@ -212,6 +212,11 @@ export class CoreInIframe implements ConnectFactoryDependencies<ConnectSettingsW
         }
     }
 
+    public setTransports() {
+        // TODO: implement
+        throw new Error('Unsupported right now');
+    }
```
mroz22, line 218: it is probably not necessary. at least for now. and in the future when suite-desktop becomes the connect provider for 3rd parties, it won't be needed either.

packages/connect-web/src/impl/core-in-suite-desktop.ts
```diff
@@ -119,6 +119,11 @@ export class CoreInSuiteDesktop implements ConnectFactoryDependencies<ConnectSet
         return await this.handshake();
     }
 
+    public setTransports() {
+        // TODO: implement
+        throw new Error('Unsupported right now');
```
mroz22, line 124: this is now exclusively 3rd parties interface and I don't think we want them to touch this API 👍

packages/connect-analytics/src/types/events.ts
```diff
@@ -28,8 +28,8 @@ export type ConnectAnalyticsEvent =
               referrerEmail?: string;
               method?: string;
               payload?: string[];
-              transportType?: string;
-              transportVersion?: string;
+              transportTypes?: string[];
+              bridgeVersion?: string;
```
mroz22, line 32: there are breaking changes to analytics now, am I right? if yes, maybe need to log it somewhere 🤔

mroz22 (commented):
I tested popup and found no regression.

this is unrelated since it happens also on develop but I found it here. using "core in popup" mode with bridge active and only one device connected - I am getting the device selection screen which is IMHO wrong. Could you please check on it @evgenysl I can explain details if needed.

image

image

mroz22 (commented):
one more note before I forget - @evgenysl, please don't forget to test "core-in-popup" fallback logic

 https://github.com/trezor/trezor-suite/blob/develop/packages/connect-web/src/index.ts#L60

mroz22 (approved):
high level it makes sense and looks ok. I havent spent much time checking details yet. 

### notes for QA: 
- no coin specific logic needs to be tested. only device and transport specific. lets focus also on edgecases
- popup under various conditions: webusb un/supported
- popup when using @trezor/connect-web and @trezor/connect-webextension. Let's revisit our webextension testing process. 
- popup using different modes:
  - core-in-poup
  - core-in-iframe
- bridge outdated screen in popup (I think you would need to have bridge less than 27) which is not very frequent nowadays
- automatic fallback from core-in-iframe to core-in-popup https://github.com/trezor/trezor-suite/blob/develop/packages/connect-web/src/index.ts#L60
- killing bridge in the course of "something"

PR #15325, opened 2024-11-08
chore(suite-analytics): remove suite-analytics from common code

matejkriz:
## Description

During optimization of suite-native build, I found out that `@suite-common/wallet-core` depends on `@trezor/suite-analytics`, so mobile app also depends on web/desktop analytics package. This PR fixes that by reporting the same events from middleware in suite package.

Please test that it works the same as before 🙏


jvaclavik (commented):
From code perspective it seems fine, But I'm going to also test the app.

jvaclavik (approved):
looks good!

PR #15327, opened 2024-11-08
revert adding standard wallet for only passphrase wallet remembered

vytick:
This reverts commit 9e4882a35af4fe5acd765ea7e0026d7ed0e91aef.

This solution brought some problems with passphrase. The goal is to revert and add just fake standard wallet to device manager and add after user clicks it.

## Related Issue

Opens #12722 again


matejkriz (approved):
Why didn’t you use git revert? :-)

vytick:
I didnt like it :)

PR #15330, opened 2024-11-08
feat(suite-native): show placeholder when standard wallet is not present

### LINKED ISSUE #12722: Passphrase: When Standard wallet not remembered

vytick:
Add placeholder to device switcher, when standard device is not present and only passphrase wallet is. On select of placeholder, create the standard wallet and select it. 

## Related Issue

Resolve #12722

## Screenshots:

https://github.com/user-attachments/assets/913aa614-aafb-41dd-8c9d-99d29914f2e0

### DIFF
--- a/suite-native/device-manager/src/components/WalletList.tsx
+++ b/suite-native/device-manager/src/components/WalletList.tsx
@@ -1,20 +1,60 @@
-import { useSelector } from 'react-redux';
+import { useSelector, useDispatch } from 'react-redux';
 
-import { selectDeviceInstances } from '@suite-common/wallet-core';
-import { VStack } from '@suite-native/atoms';
+import { A } from '@mobily/ts-belt';
+
 import { TrezorDevice } from '@suite-common/suite-types';
+import {
+    createDeviceInstanceThunk,
+    selectDevice,
+    selectDeviceInstances,
+    selectIsPortfolioTrackerDevice,
+} from '@suite-common/wallet-core';
+import { VStack } from '@suite-native/atoms';
+import { selectHasNoDeviceWithEmptyPassphrase } from '@suite-native/device';
 
 import { WalletItem } from './WalletItem';
+import { WalletItemBase } from './WalletItemBase';
 
 type WalletListProps = {
     onSelectDevice: (device: TrezorDevice) => void;
 };
 
 export const WalletList = ({ onSelectDevice }: WalletListProps) => {
+    const dispatch = useDispatch();
     const devices = useSelector(selectDeviceInstances);
+    const selectedDevice = useSelector(selectDevice);
+    const hasNoDeviceWithEmptyPassphrase = useSelector(selectHasNoDeviceWithEmptyPassphrase);
+    const isPortfolioTrackerDevice = useSelector(selectIsPortfolioTrackerDevice);
+    const isSelectable = devices.length > 1 || hasNoDeviceWithEmptyPassphrase;
+
+    // we want to show placeholder in case there are only passphrase wallets without standard and not portfolio
+    const showPlaceholder =
+        hasNoDeviceWithEmptyPassphrase && A.isNotEmpty(devices) && !isPortfolioTrackerDevice;
+
+    // on tap of placeholder we actually create device with empty passphrase and select it
+    const handlePlaceholderPress = async () => {
+        if (selectedDevice) {
+            await dispatch(
+                createDeviceInstanceThunk({
+                    device: selectedDevice,
+                    useEmptyPassphrase: true,
+                }),
+            )
+                .unwrap()
+                .then(result => onSelectDevice(result.device));
+        }
+    };
 
     return (
         <VStack spacing="sp12" paddingHorizontal="sp16">
+            {showPlaceholder && (
+                <WalletItemBase
+                    variant="standard"
+                    onPress={handlePlaceholderPress}
+                    isSelectable
+                    isSelected={false}
+                />
+            )}
             {devices.map(device => {
                 if (!device.state) {
                     return null;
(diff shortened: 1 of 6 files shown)


suite-native/device/src/selectors.ts
```diff
@@ -99,3 +100,7 @@ export const selectViewOnlyDevicesAccountsNetworkSymbols = memoize(
         );
     },
 );
+
+export const selectHasNonPassphraseDevice = memoize((state: DeviceRootState) =>
```
PeKne: nit: 
```suggestion
export const selectHasNoPassphraseDevice = memoize((state: DeviceRootState) =>
```

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -1,20 +1,59 @@
...
 
 export const WalletList = ({ onSelectDevice }: WalletListProps) => {
+    const dispatch = useDispatch();
     const devices = useSelector(selectDeviceInstances);
+    const selectedDevice = useSelector(selectDevice);
+    const hasNonPassphraseDevice = useSelector(selectHasNonPassphraseDevice);
+    const deviceState = useSelector(selectDeviceState);
+    const isPortfolioTrackerDevice = useSelector(selectIsPortfolioTrackerDevice);
+
+    const isSelectable = devices.length > 1 || !hasNonPassphraseDevice;
```
PeKne: Nit: You are negating already negated boolean value `!hasNonPassphraseDevice` and you are using it the same way few lines below. What about changing it to `hasPassphraseDevice`? The readability is better. You would have to change the selector, though.

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -1,20 +1,59 @@
...
+    const handleOnPlaceholderPress = async () => {
+        if (selectedDevice) {
+            await dispatch(
+                createDeviceInstanceThunk({
+                    device: selectedDevice,
+                    useEmptyPassphrase: true,
+                }),
+            )
+                .unwrap()
+                .then(result => onSelectDevice(result.device));
```
PeKne, line 44: What if the `createDeviceInstanceThunk` fails?

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -1,20 +1,59 @@
...
+    const deviceState = useSelector(selectDeviceState);
+    const isPortfolioTrackerDevice = useSelector(selectIsPortfolioTrackerDevice);
+
+    const isSelectable = devices.length > 1 || !hasNonPassphraseDevice;
+
+    // we want to show placeholder in case there is only passphrase wallet without standard and not portfolio
+    const showPlaceholder = !hasNonPassphraseDevice && deviceState && !isPortfolioTrackerDevice;
+
+    // on tap of placeholder we actually create device with empty passphrase and select it
+    const handleOnPlaceholderPress = async () => {
```
PeKne: nit: There is an unwritten rule that the component prop functions starts with prefix `on...` and the functions that handle their behaviour (are passed there) are named `handle...` [More here.](https://medium.com/@master_43681/understanding-and-proper-usage-of-on-and-handle-prefixes-in-react-function-naming-119f0beea400)
```suggestion
    const handlePlaceholderPress = async () => {
```

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -1,20 +1,59 @@
...
+                .unwrap()
+                .then(result => onSelectDevice(result.device));
+        }
+    };
 
     return (
         <VStack spacing="sp12" paddingHorizontal="sp16">
+            {showPlaceholder && (
+                <WalletItem
+                    deviceState={deviceState}
```
PeKne: Do I understand correctly that the `deviceState` of the currently selected passphrase device is passed to the placeholder? Why? What is it there for? Could not you just display a dummy placeholder that does not need any selectors inside instead?

PeKne (commented):
I have two questions. Other comments marked as `nit` are not important, if you do not agree with those, feel free to resolve them.

suite-native/device-manager/src/components/WalletList.tsx
vytick: hmm, good thinking. I need a device to make an instance of, but it can be current device. I'll rework this. thanks!

suite-native/device-manager/src/components/WalletList.tsx
vytick: Not having standard wallet view only, but passphrase yes and wanting to get to standard one is already edge case, so imo we dont have to handle even more edge case scenario. Wdyt?

suite-native/device-manager/src/components/WalletList.tsx
vytick: bc094e834e

suite-native/device-manager/src/components/WalletList.tsx
vytick: bc094e834e

suite-native/device-manager/src/components/WalletList.tsx
vytick: bc094e834e

suite-native/device/src/selectors.ts
vytick: bc094e834e

suite-native/device-manager/src/components/WalletList.tsx
PeKne: Fair enough

PeKne (approved):

PR #15331, opened 2024-11-08
feat(suite-native): add authenticity check to device settings

### LINKED ISSUE #14617: Mobile device management - check device authenticity
[Figma](https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Manage-device?node-id=367-2933&t=um4cRUycAW4gJnBL-1)

Check device authenticity (=DAC)

- User can trigger DAC from device settings
- Authenticity check is available only for TS family devices (TS3, TS5)
- When user clicks on Check device authenticity he is informed about the purpose of DAC
- When he continues he is prompted to continue and confirm on Trezor Device
- If the DAC is successful, user is informed that the device is genuine and he can continue using it
- If the DAC is not successful, user is informed that device might be compromised and he should contact our customer service

Implementation notes:
- POC in this PR #10247 
- Latest version of react-native-quick-crypto should support subtle crypto features we need to make this work, would be also nice to use this library to replace whole crypto-browserify because it's much faster (when I tried last time there was some issue with messaging system, but we should be able to make it work)

yanascz:
Added possibility to check device authenticity if a connected device supports that.

## Related Issue

Resolve #14617

## Screenshots:
Simulator Screenshot - iPhone 16 Pro - 2024-11-08 at 19 21 57
Simulator Screenshot - iPhone 16 Pro - 2024-11-08 at 19 22 00
Simulator Screenshot - iPhone 16 Pro - 2024-11-08 at 19 22 14
Simulator Screenshot - iPhone 16 Pro - 2024-11-08 at 19 22 24


suite-native/app/globalPolyfills.js
```diff
@@ -1,49 +1,9 @@
-import * as Crypto from 'expo-crypto';
-
-if (typeof global.crypto !== 'object') {
-    global.crypto = {};
-}
-// crypto.getRandomValues polyfill is needed for Solana
-global.crypto.getRandomValues = Crypto.getRandomValues;
-
-global.Buffer = require('buffer').Buffer;
```
vytick, line 9: Removing this buffer related stuff is safe? What does Dan say?

suite-native/module-device-settings/src/components/DeviceSettingsCardLayout.tsx
```diff
@@ -16,7 +16,7 @@ type DeviceSettingsCardProps = {
     alertBoxProps?: Omit<AlertBoxProps, 'borderRadius'>;
 };
 
-export const DeviceSettingsCard = ({
```
vytick, line 19: Why the name change?

suite-native/app/package.json
```diff
@@ -112,6 +111,7 @@
         "react-native-gesture-handler": "2.18.1",
         "react-native-keyboard-aware-scroll-view": "0.9.5",
         "react-native-mmkv": "2.12.2",
+        "react-native-quick-crypto": "0.7.6",
```
vytick, line 113: you probably want to change runtimeVersion in app.config.js because of preview build

suite-common/suite-utils/src/__tests__/device.test.ts
```diff
@@ -1,10 +1,21 @@
...
+import { isDeviceAuthenticityCheckSupported, isDeviceWithButtons } from '../device';
 
 describe('device utils', () => {
     test('isDeviceWithButtons', () => {
         expect(isDeviceWithButtons(DeviceModelInternal.T3B1)).toBe(true);
         expect(isDeviceWithButtons(DeviceModelInternal.T3T1)).toBe(false);
     });
+
+    test.each([
+        [DeviceModelInternal.T1B1, false],
```
Nodonisko: Tbh I don't see any benefit in testing that super simple code

Nodonisko (approved):
Did not tested yet but code looks good

suite-native/app/globalPolyfills.js
Nodonisko: I think it should use native buffer from different library so it should be safe. Please test it account imports still work just in case.

yanascz: The buffer stuff is handled in the `install` function.

suite-common/suite-utils/src/__tests__/device.test.ts
yanascz: I basically transformed the original constant into a test. We can drop it a thus reduce number of changes.

suite-native/module-device-settings/src/components/DeviceSettingsCardLayout.tsx
yanascz: Because the name is misleading, I myself got confused. IMO the new name better reflects the purpose of the class.

Can revert if you don't think it's an improvement.

suite-common/suite-utils/src/__tests__/device.test.ts
Nodonisko: I think we should drop it

Nodonisko:
Some e2e seems to fail

suite-native/app/package.json
Nodonisko: Yea, expo dev client crash now

suite-common/suite-utils/src/__tests__/device.test.ts
yanascz: OK, removed.

suite-native/app/package.json
yanascz: Incremented.

suite-native/module-device-settings/src/components/DeviceSettingsCardLayout.tsx
vytick: It feels the same or maybe a bit worse to understand, but leave it be :)

PR #15333, opened 2024-11-10
chore: replace proxy-memoize with reselect

### LINKED ISSUE #14785: Skeleton remains in the last coins fiat value on the dashboard after discovery
**Describe the bug**
When a discovery is finished (being finished) there's a chance that there is a skeleton that remains in the fiat value and it's always with the last coin (enabled) on the dashboard. This can be "fixed" by going into, for example, settings and back, the value suddenly appears.

**Info:**

-   Mobile Suite Version: [version/commit](https://github.com/trezor/trezor-suite/pull/14281)
-   OS: Android

**How to reproduce**

1. Connect a device with some enabled coins
2. Wait for the discovery of the coins and specifically for the last enabled one
3. There remains a skeleton where the fiat value should be even after quite some time
4. This can be "fixed" by going for example into the settings and back

**Screenshots**

Nodonisko:
## Description

I migrated from `proxy-memoize` because it was too magical and caused issues that some selectors did not updated correctly randomly and we were also stucked on old version because on new it was even worse and it throwed some errors on mobile too.

Reselect is more standart library and it works, even that DX is somewhat worse especially when you have multiple nested parametrized selectors.

## Related Issue

Resolve #14785


suite-common/redux-utils/src/index.ts
```diff
@@ -7,3 +7,5 @@ export * from './createReducerWithExtraDeps';
 export * from './createActionWithExtraDeps';
 export * from './createSingleInstanceThunk';
 export * from './hooks/useSelectorDeepComparison';
+export * from './selectors';
+export * from 'reselect';
```
PeKne: Is this intentional?

suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -147,303 +150,300 @@ export const prepareAccountsReducer = createReducerWithExtraDeps(
     },
 );
 
-export const selectAccounts = (state: AccountsRootState) => state.wallet.accounts;
+const createAppSelector = createWeakMapSelector.withTypes<
```
PeKne: why is this has `...App...` in the name?

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -27,6 +27,10 @@ import {
 import { ConnectDeviceSettings, deviceActions } from './deviceActions';
 import { PORTFOLIO_TRACKER_DEVICE_ID } from './deviceConstants';
 
+
+// For selectors with parameters, use WeakMap memoization
+const createAppSelector = createWeakMapSelector.withTypes<DeviceRootState>();
```
PeKne: Why is this comment in this file but in other reducer files not? Maybe move it to `redux-utils/selectors.ts`?

suite-common/wallet-core/src/accounts/accountsReducer.ts
PeKne: What about changing it to `createMemoizedSelector`?

suite-common/wallet-core/src/transactions/transactionsReducer.ts
```diff
@@ -66,7 +65,7 @@ export type TransactionsRootState = {
             transactions: { [key: AccountKey]: (WalletAccountTransaction | null | undefined)[] };
         };
     };
-};
+} & AccountsRootState;
```
PeKne, line 72: Does this mean that it will be sufficient to send a only one RootState type to the `useSelector()` call? Love it :clap:

suite-native/app/app.config.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable no-console */
+
```
PeKne: This should be probably returned

suite-native/tokens/src/tokensSelectors.ts
```diff
@@ -47,63 +54,52 @@ export const selectAccountTokenInfo = memoizeWithArgs(
...
-export const selectAccountTokenTransactions = memoizeWithArgs(
-    (
-        state: TransactionsRootState,
-        accountKey: AccountKey,
-        tokenAddress: TokenAddress,
-    ): WalletAccountTransaction[] =>
+export const selectAccountTokenTransactions = createAppSelector(
+    [
+        selectAccountTransactions,
+        (_state, _accountKey: AccountKey, tokenAddress: TokenAddress) => tokenAddress,
```
PeKne, line 98: Damn, this is so ugly that you have to pass a function instead of a simple argument value 🥴. But it is how it is

PeKne (commented):
Tested and works good. I did not encounter any performance drop or crash 👏

suite-common/message-system/src/messageSystemSelectors.ts
```diff
@@ -16,53 +19,57 @@ export const selectMessageSystemCurrentSequence = (state: MessageSystemRootState
...
 );
 
-export const selectContextMessageContent = memoizeWithArgs(
-    (state: MessageSystemRootState, domain: ContextDomain, language: string) => {
-        const activeContextMessages = selectActiveContextMessages(state);
+export const selectContextMessageContent = createAppSelector(
+    [
+        selectActiveContextMessages,
+        (_state, domain: ContextDomain) => domain,
+        (_state, _domain, language: string) => language,
```
PeKne, line 66: nit: 
```suggestion
        (_state, domain: ContextDomain, language: string) => {domain, language},
```

Nodonisko: I am not sure, but I think this will break memoization, but IDK

suite-native/tokens/src/tokensSelectors.ts
Nodonisko: It's even worse if you have multiple nested selectors with different params :D

suite-native/app/app.config.ts
Nodonisko: It doesn't throw anywhere so I guess it's fine

suite-common/wallet-core/src/transactions/transactionsReducer.ts
Nodonisko: Yea, I don't see reason why not to do this

suite-common/wallet-core/src/accounts/accountsReducer.ts
Nodonisko: It's taken from their docs, but `createMemoizedSelector` looks good

suite-common/redux-utils/src/index.ts
Nodonisko: It's correct because initially I wanted to use more than only one function from reselect package, but it seems we are fine with just one so I will remove it.

suite-common/wallet-core/src/device/deviceReducer.ts
Nodonisko: Yea good idea

suite-common/redux-utils/src/selectorsUtils.ts
```diff
@@ -0,0 +1,13 @@
+import { createSelectorCreator, weakMapMemoize } from 'reselect';
+
+const EMPTY_STABLE_ARRAY: unknown[] = [];
+
+export const returnStableArrayIfEmpty = <T>(array?: readonly T[] | T[]): T[] => {
```
matejkriz, line 9: What about some comment on when/why to use it?

matejkriz:
Could you please provide some at least basic PR description why this change is happening?

suite-common/redux-utils/src/selectorsUtils.ts
Nodonisko: added

Nodonisko:
@matejkriz done

bosomt:
**QA OK**

tested on all and academic seed
- discovery
- adding account
- send tx


**Info:**
 - Suite version: web 24.12.0 (c4aeab255612457b91be531a3b20b4ce72e9f010)
 - OS: MacIntel
 - Screen: 1512x982

matejkriz (approved):
From my testing it looks ok and I'm really glad we finally get rid off the magic proxy-memoize that I spent lot of hours debugging 🎉

PR #15335, opened 2024-11-11
CLONE: fix(suite): order of network filter #15289

### LINKED ISSUE #9165: Accounts tab - order of network filter is different than accounts order

tomasklim:
## Description

- clone of #15289 with added migration

Fixed order of network filter. The order is now the same as the account order.

## Related Issue

Resolve #9165

### DIFF
--- a/packages/suite/src/reducers/wallet/settingsReducer.ts
+++ b/packages/suite/src/reducers/wallet/settingsReducer.ts
@@ -1,12 +1,13 @@
 import produce from 'immer';
 
-import { WalletSettings } from '@suite-common/wallet-types';
+import type { WalletSettings } from '@suite-common/wallet-types';
 import { PROTO } from '@trezor/connect';
+import { networkSymbolCollection } from '@suite-common/wallet-config';
 
-import { STORAGE } from 'src/actions/suite/constants';
 import { WALLET_SETTINGS } from 'src/actions/settings/constants';
 import * as walletSettingsActions from 'src/actions/settings/walletSettingsActions';
-import { Action, AppState } from 'src/types/suite';
+import { STORAGE } from 'src/actions/suite/constants';
+import type { Action, AppState } from 'src/types/suite';
 
 export type State = WalletSettings;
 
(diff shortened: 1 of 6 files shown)


packages/suite/src/reducers/wallet/settingsReducer.ts
```diff
@@ -36,7 +37,10 @@ const settingsReducer = (state: State = initialState, action: Action): State =>
 
             case walletSettingsActions.changeNetworks.type: {
                 if (walletSettingsActions.changeNetworks.match(action)) {
-                    draft.enabledNetworks = action.payload;
+                    draft.enabledNetworks = [...action.payload].sort(
```
Lemonexe, line 40: Nit: `toSorted` is exactly for this purpose, [see here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) :slightly_smiling_face: 
```suggestion
                    draft.enabledNetworks = action.payload.toSorted(
```

packages/suite/src/storage/index.ts
```diff
@@ -5,7 +5,7 @@ import { reloadApp } from 'src/utils/suite/reload';
 import { migrate } from './migrations';
 import type { SuiteDBSchema } from './definitions';
 
-const VERSION = 48; // don't forget to add migration and CHANGELOG when changing versions!
+const VERSION = 49; // don't forget to add migration and CHANGELOG when changing versions!
```
Lemonexe, line 8: [changelog](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/storage/CHANGELOG.md) is missing?

Lemonexe (approved):
LGTM!

PR #15338, opened 2024-11-11
fix node-bridge segfault

mroz22:
1st commit - just refactoring I felt makes sense

### DIFF
--- a/packages/transport/src/api/usb.ts
+++ b/packages/transport/src/api/usb.ts
@@ -354,7 +354,7 @@ export class UsbApi extends AbstractApi {
     }
 
     public async closeDevice(path: string) {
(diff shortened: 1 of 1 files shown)


packages/transport/src/api/usb.ts
```diff
@@ -354,21 +354,34 @@ export class UsbApi extends AbstractApi {
     }
 
     public async closeDevice(path: string) {
-        const device = this.findDevice(path);
+        let device = await this.findDevice(path, true);
```
mroz22: this is the important change that makes it work on my machine. adding `true` here. Other places could maybe force enumeration too but I wanted to make changes minimal.

marekrjpolak (approved):
LGTM

PR #15342, opened 2024-11-11
fix(suite-native): redirect to token send on review cancel

### LINKED ISSUE #15341: Canceling sending of ERC2O token on device and continuing again selects ETH instead of the token in transaction
**Describe the bug**
When I want to send a ERC20 token transaction and I cancel the flow on the device, it doesn't return me back to the send flow of the ERC20 token but rather on just ETH.

**Info:**

-   Mobile Suite Version: 24.11.1 https://github.com/trezor/trezor-suite/commit/ff16b788c576e89bbbfc17cbe783154223440336
-   OS: Android

**How to reproduce**

1. Connect device
2. Navigate to ETH account
3. Select any token
4. Select "Send"
5. Continue with the "Send" flow until you are prompted on device
6. Cancel the flow on the device
7. Try to continue with the flow again on host
8. You can see that ETH is selected instead of your desired token

PeKne:
## Description
The token contract is passed to the navigation action on the device review cancelation. So the user is redirected back to the TOKEN send flow again instead of MAINNET is at was now.

## Related Issue

Resolve #15341

## Screenshots:

https://github.com/user-attachments/assets/c915db55-dfc9-4b09-9f23-6abed3573271

### DIFF
--- a/suite-native/module-send/src/components/AddressReviewStepList.tsx
+++ b/suite-native/module-send/src/components/AddressReviewStepList.tsx
@@ -49,7 +49,11 @@ export const AddressReviewStepList = () => {
 
     const [childHeights, setChildHeights] = useState<number[]>([]);
     const [stepIndex, setStepIndex] = useState(0);
-    const handleSendReviewFailure = useHandleSendReviewFailure({ accountKey, transaction });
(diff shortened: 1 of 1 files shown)


vytick (approved):
lgtm

PR #15344, opened 2024-11-11
feat(suite-native): fetch token definitions while importing account

### LINKED ISSUE #15332: When importing first ETH accounts, tokens are not displayed on summary screen
In portfolio tracker when importing first token for coin that supports tokens, tokens are not displayed on import summary because token definitions are not fetched. It will work for second account because it will fetch token definitions for first account.

vytick:
While fetching account data for imported account, fetch also token definitions for this network, if needed.

## Related Issue

Resolve #15332

## Screenshots:

https://github.com/user-attachments/assets/71759c78-e00f-4cdc-847b-608e02e5d84c

### DIFF
--- a/suite-native/module-accounts-import/src/accountsImportThunks.ts
+++ b/suite-native/module-accounts-import/src/accountsImportThunks.ts
@@ -12,8 +12,11 @@ import { getAccountIdentity, shouldUseIdentities } from '@suite-common/wallet-ut
 import { Timestamp, TokenAddress } from '@suite-common/wallet-types';
 import { FiatCurrencyCode } from '@suite-common/suite-config';
 import {
+    DefinitionType,
+    getTokenDefinitionThunk,
     periodicCheckTokenDefinitionsThunk,
     selectFilterKnownTokens,
+    selectNetworkTokenDefinitions,
 } from '@suite-common/token-definitions';
 
 import { paymentTypeToAccountType } from './constants';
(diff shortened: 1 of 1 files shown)


suite-native/module-accounts-import/src/accountsImportThunks.ts
```diff
@@ -116,6 +119,21 @@ export const getAccountInfoThunk = createThunk<
...
 
             if (fetchedAccountInfo?.success) {
+                const tokenDefinitions = selectNetworkTokenDefinitions(getState(), networkSymbol);
+
+                // fetch token definitions for this network in case they are needed
+                if (!tokenDefinitions) {
+                    const definitionTypes = getSupportedDefinitionTypes(networkSymbol);
+
+                    definitionTypes.forEach(async type => {
+                        await dispatch(
```
Nodonisko: What about doing this sooner? Like on `XpubScanScreen` we already know network there will be more time for request to finish and also it will be happening when app is "less busy". We can also dispatch this call only for that specific network instead of all.

PeKne (approved):
Awesome 👌

suite-native/module-accounts-import/src/accountsImportThunks.ts
```diff
@@ -116,6 +119,21 @@ export const getAccountInfoThunk = createThunk<
             ]);
 
             if (fetchedAccountInfo?.success) {
+                const tokenDefinitions = selectNetworkTokenDefinitions(getState(), networkSymbol);
+
+                // fetch token definitions for this network in case they are needed
+                if (!tokenDefinitions) {
+                    const definitionTypes = getSupportedDefinitionTypes(networkSymbol);
```
Nodonisko: Can we just hardcode tokens here? We don't have any nfts or something in app

vytick: good idea: f7d2969817

suite-native/module-accounts-import/src/accountsImportThunks.ts
vytick: This happens when there is a loading screen present and user is already waiting. It feels to be the best moment in the flow. 

Also only token definitions for imported network are being fetched, no other networks

Nodonisko: Still think it will be better to start sooner, it could run in background and have more time to finish. Also you are not awaiting it here so there is no guarantee that it will be fetched b when next screen is displayed

Nodonisko: I don't have strong onion, so you may keep it here but at fix that await

vytick: It is already awaited in f7d2969

Nodonisko (approved):

PR #15346, opened 2024-11-11
fix(suite): fix wrong key for translation

MiroslavProchazka:
image

PR #15347, opened 2024-11-11
Add type check to messages.ts and fix translations

komret:
## Description

Translation key and `id` did not match in some cases which led to some strings not being translated in the UI. There was no TS check 🤯 Example of a bug caused by this: https://github.com/trezor/trezor-suite/pull/15346

I added the TS check and fixed all broken translations. I usually kept the key that is already in translated in Crowdin. In two cases, I decided to change the key instead: `TR_ADDRESSES` to `TR_ADDRESS` because the former did not make sense and `TR_CHECK_RECOVERY_SEED_DESC_T2B1` => `TR_CHECK_RECOVERY_SEED_DESC_T3B1` because it is dynamic.

The idea for a type-checking function comes from [here](https://www.reddit.com/r/typescript/comments/o4k08k/is_there_any_way_to_enforce_a_key_to_be_equal_to/).

## Screenshots:
Screenshot 2024-11-11 at 20 01 24


MiroslavProchazka (approved):

packages/suite/src/support/messages.ts
```diff
@@ -1,6 +1,24 @@
...
+    description?: string;
+    // Must be set to true for programmatically constructed keys. Otherwise, the keys will be deleted by the list-duplicates script.
+    dynamic?: boolean;
+};
+
+/**
+Checks whether id corresponds to the property name. Otherwise, text is not translated. Returns the same value that was passed in.
+ */
+const defineMessagesWithTypeCheck = <Keys extends string>(messages: {
+    [K in Keys]: MessageDescriptor<K>;
```
Lemonexe: Nit: a string union should IMHO be named with singular rather than plural, which is for Array type.
I think it's a convention, but I don't have any source to back it up :smile: 
```suggestion
const defineMessagesWithTypeCheck = <TranslationKey extends string>(messages: {
    [K in TranslationKey]: MessageDescriptor<K>;
```

Lemonexe (approved):
:ok_hand: perfect! :rocket: 

:thought_balloon: Just FIY: I did something similar a while ago in [networks config](https://github.com/trezor/trezor-suite/blob/a33bf1d19feacec314c3debbbbfbf52f8426e674/suite-common/wallet-config/src/types.ts#L114-L116), but with static type. I wondered if this can be also written with  type, without the need for an actual function `messages => messages`.
But no. While `NetworkSymbol` is string union statically declared beforehand, here we need to infer `TranslationKey`, so the fn is the only way :slightly_smiling_face:

packages/suite/src/support/messages.ts
komret: I changed `Keys` => `Key`.

PR #15351, opened 2024-11-12
fix(mobile): infite loading of fiat rates during discovery

### LINKED ISSUE #14785: Skeleton remains in the last coins fiat value on the dashboard after discovery

Nodonisko:
## Description

Main account was not included in fiat rates request but it worked because it was probably triggered by something else but randomly.

## Related Issue

Resolve #14785


PeKne (approved):

PeKne (approved):
👌

PR #15352, opened 2024-11-12
Doge smart fees

mroz22:
this PR makes fee estimation and tx composition using values returned by doge blockbooks. Txs should now have fee high enough to get mined. On the other hand, we might be overpaying slightly now.

### DIFF
--- a/packages/blockchain-link/webpack/dev.js
+++ b/packages/blockchain-link/webpack/dev.js
@@ -23,7 +23,6 @@ module.exports = {
             directory: `${SRC}ui`,
         },
         hot: false,
-        https: false,
         port: PORT,
     },
     module: {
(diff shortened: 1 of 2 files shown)


szymonlesisz (approved):
shipit :)

STew790:
**QA OK**
LGTM, transaction is already confirmed. 
Here is a tx I sent from develop an hour ago:
https://doge5.trezor.io/tx/e8f4b3ae822afa7f6e75241cf351c98028a4e51084dffdca18b6e6c2a696208f
and here is the tx from this branch which is already confirmed:
https://doge5.trezor.io/tx/29ee01c7795f933c01688303952b809503f0f3a30d3b867d5aeaa946d7622054

PR #15364, opened 2024-11-13
feat(suite-native): Send Custom Fee

### LINKED ISSUE #15180: Send Coins - Custom fee
## Acceptance criteria
User is allowed to specify custom fee rate if he does not like any of the predefined fee levels provided.

PeKne:
## Description
- custom fee implemented for all the enabled network (both Bitcoin-like and Ethereum-like + tokens)
- working with a send max enabled as well (the total amount is always adjusted so all the balance is send from the account)
## Related Issue

Resolve  #15180

### BTC 

https://github.com/user-attachments/assets/0738cc27-cbcd-4698-8765-bd81366346f2

### BTC - Send Max

https://github.com/user-attachments/assets/efde781e-dce4-43ac-bf08-44ff4e464961

### ETH 

https://github.com/user-attachments/assets/60ba68da-c4be-4bf2-8e92-f7ab96ac4e01

### DIFF
--- a/suite-native/module-send/src/components/CustomFeeBottomSheet.tsx
+++ b/suite-native/module-send/src/components/CustomFeeBottomSheet.tsx
@@ -1,0 +1,114 @@
+import { useSelector } from 'react-redux';
+import Animated, {
+    FadeInDown,
+    FadeOutDown,
+    SlideInDown,
+    SlideOutDown,
+    useAnimatedStyle,
+    withTiming,
+} from 'react-native-reanimated';
+
+import { useRoute } from '@react-navigation/native';
+
+import { AlertBox, BottomSheet, Button, HStack, Text, VStack } from '@suite-native/atoms';
+import { Translation } from '@suite-native/intl';
+import { useFormContext } from '@suite-native/forms';
+import { CryptoAmountFormatter, CryptoToFiatAmountFormatter } from '@suite-native/formatters';
+import { AccountsRootState, selectAccountNetworkSymbol } from '@suite-common/wallet-core';
+import { SendStackParamList, SendStackRoutes, StackProps } from '@suite-native/navigation';
+
+import { SendFeesFormValues } from '../sendFeesFormSchema';
+import { CustomFeeInputs } from './CustomFeeInputs';
+import { useCustomFee } from '../hooks/useCustomFee';
+
+type CustomFeeBottomSheetProps = {
+    isVisible: boolean;
+    onClose: () => void;
+};
+
+type RouteProps = StackProps<SendStackParamList, SendStackRoutes.SendAddressReview>['route'];
+
+export const CustomFeeBottomSheet = ({ isVisible, onClose }: CustomFeeBottomSheetProps) => {
+    const route = useRoute<RouteProps>();
+    const { accountKey, tokenContract } = route.params;
+
+    const { feeValue, isFeeLoading, isSubmittable, isErrorBoxVisible } = useCustomFee({
+        accountKey,
+        tokenContract,
+    });
+
+    const networkSymbol = useSelector((state: AccountsRootState) =>
+        selectAccountNetworkSymbol(state, accountKey),
+    );
+
+    const { setValue, handleSubmit } = useFormContext<SendFeesFormValues>();
+
+    const handleSetCustomFee = handleSubmit(() => {
+        setValue('feeLevel', 'custom');
+        onClose();
+    });
+
+    const animatedButtonContainerStyle = useAnimatedStyle(
+        () => ({
+            height: withTiming(isSubmittable && isVisible ? 50 : 0),
+        }),
+        [isSubmittable, isVisible],
+    );
+
+    if (!networkSymbol) return null;
+
+    return (
+        <BottomSheet
+            isVisible={isVisible}
+            onClose={onClose}
+            title={<Translation id="moduleSend.fees.custom.bottomSheet.title" />}
+        >
+            <VStack spacing="sp24" justifyContent="space-between" flex={1}>
+                <CustomFeeInputs networkSymbol={networkSymbol} />
+                <HStack
+                    flex={1}
+                    justifyContent="space-between"
+                    alignItems="center"
+                    paddingHorizontal="sp1"
+                >
+                    <Text variant="highlight">
+                        <Translation id="moduleSend.fees.custom.bottomSheet.total" />
+                    </Text>
+                    <VStack alignItems="flex-end">
+                        <CryptoToFiatAmountFormatter
+                            value={feeValue}
+                            isLoading={isFeeLoading}
+                            network={networkSymbol}
+                        />
+                        <CryptoAmountFormatter
+                            value={feeValue}
+                            network={networkSymbol}
+                            variant="body"
+                            isLoading={isFeeLoading}
+                            isBalance={false}
+                        />
+                    </VStack>
+                </HStack>
+                {isErrorBoxVisible && (
+                    <Animated.View entering={FadeInDown} exiting={FadeOutDown}>
+                        <AlertBox
+                            variant="error"
+                            title={<Translation id="moduleSend.fees.error" />}
+                            contentColor="textDefault"
+                        />
+                    </Animated.View>
+                )}
+
+                <Animated.View style={animatedButtonContainerStyle}>
+                    {isSubmittable && (
+                        <Animated.View entering={SlideInDown} exiting={SlideOutDown}>
+                            <Button onPress={handleSetCustomFee}>
+                                <Translation id="moduleSend.fees.custom.bottomSheet.confirmButton" />
+                            </Button>
+                        </Animated.View>
+                    )}
+                </Animated.View>
+            </VStack>
+        </BottomSheet>
+    );
+};
--- a/suite-native/module-send/src/components/CustomFeeCard.tsx
+++ b/suite-native/module-send/src/components/CustomFeeCard.tsx
@@ -1,0 +1,123 @@
+import { useSelector } from 'react-redux';
+import Animated, { FadeInLeft, FadeOutLeft } from 'react-native-reanimated';
+
+import { useRoute } from '@react-navigation/native';
+
+import { Card, HStack, VStack, Text, Box, Button } from '@suite-native/atoms';
+import { useFormContext } from '@suite-native/forms';
+import { Translation } from '@suite-native/intl';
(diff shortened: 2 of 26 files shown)


suite-native/module-send/src/components/CustomFeeBottomSheet.tsx
```diff
@@ -0,0 +1,114 @@
...
+    );
+
+    const { setValue, handleSubmit } = useFormContext<SendFeesFormValues>();
+
+    const handleSetCustomFee = handleSubmit(() => {
+        setValue('feeLevel', 'custom');
+        onClose();
+    });
+
+    const animatedButtonContainerStyle = useAnimatedStyle(
```
Nodonisko, line 51: I think you might want to use this https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/layout-transitions#linear-transition

It should work better and it should also add some nice animation when that error box is diplayed

suite-native/module-send/src/components/CustomFeeCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+                                    isBalance={false}
+                                    variant="hint"
+                                    numberOfLines={1}
+                                    adjustsFontSizeToFit
+                                />
+                            </VStack>
+                        </HStack>
+                    </VStack>
+                    <HStack flex={1} justifyContent="space-between">
+                        <Box flex={1 / 3}>
```
Nodonisko: I think if you use just 1 and later 2 it will work same from what I know about flexbox

suite-native/module-send/src/components/SendFeesForm.tsx
```diff
@@ -97,43 +122,56 @@ export const SendFeesForm = ({ accountKey, tokenContract, feeLevels }: SendFormP
     const transactionBytes = normalFee.bytes as number;
 
     // If trezor-connect was not able to compose the fee level, we have calculate total amount locally.
-    const mockedFee = transactionBytes * Number(feePerUnit);
-    const mockedTotalAmount = mockedFee + (Number(normalFee.totalSpent) - Number(normalFee.fee));
-    const isSubmittable = selectedFeeLevelTransaction.type === 'final';
+    const mockedFee = BigNumber(transactionBytes)
```
Nodonisko: Can we memo these BigNumber operations? They are usually quite slow

suite-native/module-send/src/hooks/useCustomFee.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }, [accountKey, dispatch, networkType, setError, tokenContract, translate, trigger, getValues]);
+
+    useEffect(() => {
+        setIsFeeLoading(true);
+        debounce(handleValuesChange);
+    }, [watchedFeePerUnit, watchedFeeLimit, handleValuesChange, debounce]);
+
+    // TODO: estimate do not work for error state EtH
+    // If the trezor-connect is unable to compose the transaction, we display rough estimate of the fee instead.
+    const feeEstimate = watchedFeePerUnit
```
Nodonisko: can we memo this?

suite-native/module-send/src/sendFeesFormSchema.ts
```diff
@@ -1,11 +1,99 @@
...
+        .test(
+            'fee-too-low',
+            'Fee is too low.',
+            (value, { options: { context } }: yup.TestContext<SendFeesFormContext>) => {
+                if (!value) return true;
+                const { networkFeeInfo } = context!;
+
+                if (!networkFeeInfo) return false;
+
+                const feeBig = new BigNumber(value);
```
Nodonisko: It is necessary to have BN here? How does numbers looks like?

I guess this validation runs with every change of form so it should be super quick.

suite-native/module-send/src/sendFormSlice.ts
```diff
@@ -59,3 +75,31 @@ export const sendFormSlice = createSliceWithExtraDeps({
             });
     },
 });
+
+export const selectFeeLevels = (state: NativeSendRootState) => state.wallet.send.feeLevels;
+export const selectCustomFeeLevel = (
+    state: NativeSendRootState,
+): GeneralPrecomposedTransaction | undefined => state.wallet.send.feeLevels.custom;
+
+export const selectFeeLevelTransactionBytes = (
```
Nodonisko: use memoize, especially if you use BN

Nodonisko (commented):
Checked, not tested on real device also I don't know much about how fees works

suite-native/module-send/src/components/CustomFeeCard.tsx
PeKne: 61136b9777

suite-native/module-send/src/components/SendFeesForm.tsx
PeKne: 61136b9777

suite-native/module-send/src/hooks/useCustomFee.tsx
PeKne: 61136b9777

suite-native/module-send/src/sendFeesFormSchema.ts
PeKne: you are right, no need for BN here 61136b9777

suite-native/module-send/src/sendFormSlice.ts
PeKne: 61136b9777

suite-native/module-send/src/components/CustomFeeBottomSheet.tsx
PeKne: The layout linear transition works good for calculating animation of the component wrappen in the Animated.View. But I decided to use this kind of height transformation, because it also nicely resizes the whole bottomSheet (wrapper of the content) without doing any nasty layout jumps. If you have some better solution to address this issue, I will be happy to fix that.

Nodonisko: cool didnt know that, thanks for explanation

suite-native/module-send/src/components/SendFeesForm.tsx
```diff
@@ -47,27 +51,60 @@ type SendFeesNavigationProps = StackToStackCompositeNavigationProps<
...
+    const feeLevels = useSelector(selectFeeLevels);
+
+    const networkFeeInfo = useSelector((state: FeesRootState) =>
+        selectNetworkFeeInfo(state, account?.symbol),
+    );
+
+    const normalFeeLevelInfo = networkFeeInfo?.levels.find(l => l.label === 'normal');
+    const networkType = account?.symbol ? getNetworkType(account.symbol) : undefined;
+
+    const minimalFeeLimit =
```
Nodonisko, line 69: can you do only  `feeLevels?.normal?.estimatedFeeLimit` ?

suite-native/module-send/src/components/SendFeesForm.tsx
```diff
@@ -86,54 +123,60 @@ export const SendFeesForm = ({ accountKey, tokenContract, feeLevels }: SendFormP
...
     });
 
-    const normalFee = feeLevels.normal as PrecomposedTransactionFinal; // user is not allowed to enter this screen if normal fee is not final
-    const transactionBytes = normalFee.bytes as number;
-
-    // If trezor-connect was not able to compose the fee level, we have calculate total amount locally.
-    const mockedFee = transactionBytes * Number(feePerUnit);
-    const mockedTotalAmount = mockedFee + (Number(normalFee.totalSpent) - Number(normalFee.fee));
-    const isSubmittable = selectedFeeLevelTransaction.type === 'final';
+    const mockedTotalAmount = BigNumber(normalFee.totalSpent)
```
Nodonisko: useMemo

suite-native/module-send/src/hooks/useCustomFee.tsx
```diff
@@ -0,0 +1,121 @@
...
+        trigger,
+        getValues,
+        watch,
+    } = useFormContext<SendFeesFormValues>();
+
+    const feePerUnitFieldName = 'customFeePerUnit';
+    const feeLimitFieldName = 'customFeeLimit';
+
+    const watchedFeePerUnit = watch(feePerUnitFieldName, '0');
+    const watchedFeeLimit = watch(feeLimitFieldName, '1') as string;
```
Nodonisko, line 56: why it's necessary typecast here?

suite-native/module-send/src/hooks/useCustomFee.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    const normalLevelTransactionBytes = useSelector((state: NativeSendRootState) =>
+        selectFeeLevelTransactionBytes(state, 'normal'),
+    );
+
+    const handleValuesChange = useCallback(async () => {
+        const { customFeePerUnit, customFeeLimit } = getValues();
+
+        trigger();
```
Nodonisko, line 65: what does trigger do? shoudn't we trigger submit after following condition?

suite-native/module-send/src/components/SendFeesForm.tsx
PeKne: The `feeLevels.normal` might be of a type that does not include `estimatedFeeLimit`. So the Typescript shouts that we are trying to access a prop that might not be present.

suite-native/module-send/src/components/SendFeesForm.tsx
PeKne: c6f6b03e3d

suite-native/module-send/src/hooks/useCustomFee.tsx
PeKne: Because the feeLimit is optional so, its type is equal to `string | undefined`. For this reason, there is a default value `1` passed to the watch function, so the result of the function should always be a `string`. `react-hook-form` TypeScript implementation is not the best, so it returns the type of `string | undefined` anyway. The typecast fixes that.

suite-native/module-send/src/hooks/useCustomFee.tsx
PeKne: The `trigger` function validates the form fields. If an error is found, it is displayed to the user. Ideally, we would not have to call `trigger` empirically but rather use the `mode: 'onChange'` validation mode of the react-hook-form, so the library itself handles the validation. But we want to validate on `debounce()` so it is not validated while the user is still typing. For this reason, the trigger is called in the debounce onChange function.

Nodonisko (approved):

Issue #15311, opened 2024-11-07
(Do not) Reintroduce `@typescript-eslint/no-shadow` 


peter-sanderson:
I think it brings more hard then benefit. It supper annoying and I am not sure if it really can help with something.

With Typescript, its very hard to introduce an error by shadowing anyway,

Even with most lenient configuration
```
'@typescript-eslint/no-shadow': [
    'error',
    {
        builtinGlobals: false,
        hoist: 'never', // most lenient variant
        allow: ['_', 'error', 'resolve', 'reject', 'fetch'],
    },
],
```

It is 192 errors

Nodonisko:
We agreed to turn this on not that long ago and I prefer to have this enabled. 

You may disable this for this desktop if you want but please leave this on for mobile.

PR #15366, opened 2024-11-13
chore: reintroduce no-shadow in minimal form

peter-sanderson:
Solves: https://github.com/trezor/trezor-suite/issues/15311

### DIFF
--- a/packages/eslint/src/typescriptConfig.mjs
+++ b/packages/eslint/src/typescriptConfig.mjs
@@ -5,22 +5,6 @@ export const typescriptConfig = [
     {
         rules: {
             // Additional rules
-            '@typescript-eslint/no-unused-vars': [
-                'error',
-                {
-                    vars: 'all',
-                    args: 'none',
-                    ignoreRestSiblings: true,
-                    varsIgnorePattern: '^_',
-                },
-            ],
-            '@typescript-eslint/no-shadow': [
-                'off',
-                {
-                    builtinGlobals: false,
-                    allow: ['_', 'error', 'resolve', 'reject', 'fetch'],
-                },
-            ],
             '@typescript-eslint/no-use-before-define': ['error'],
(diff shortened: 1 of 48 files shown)


Lemonexe (approved):

packages/eslint/src/typescriptConfig.mjs
```diff
@@ -5,22 +5,6 @@ export const typescriptConfig = [
...
-            '@typescript-eslint/no-unused-vars': [
-                'error',
-                {
-                    vars: 'all',
-                    args: 'none',
-                    ignoreRestSiblings: true,
-                    varsIgnorePattern: '^_',
-                },
-            ],
-            '@typescript-eslint/no-shadow': [
```
Nodonisko, line 17: It was not off, it was on only for new packages from moment I added it. Please bring it back at least for suite-native

Nodonisko (changes requested):

packages/eslint/src/typescriptConfig.mjs
peter-sanderson: I had some hard time with this rule: I created a follow-up issue here: https://github.com/trezor/trezor-suite/issues/15311

peter-sanderson: Oh, I missed you comment there. I try for native.

peter-sanderson: done

Nodonisko (approved):

Lemonexe (approved):
Looks.. Acceptable to me :neutral_face: 

Renaming `variable` to `variable2` is not pretty, but if the goal is to annoy us developers so that we rename it appropriately when working on these files in future, then I guess it's better than `disable-next-line`, as it will prevent bugs, and certainly give enough annoyance :laughing: 

:rocket:

PR #15375, opened 2024-11-13
fix(connect): remove unwanted acquire in case when deriveCardano chan…

mroz22:


matejkriz (approved):
Looks good and fixes our issue 🎉 🎖️

PR #15376, opened 2024-11-13
chore: add eslint-plugin-cypress

### LINKED ISSUE #15297: Introduce eslint-plugin-cypress and use recommaned setings
https://www.npmjs.com/package/eslint-plugin-cypress

HajekOndrej:
## Related Issue

Resolve #15297

### DIFF
--- a/packages/eslint/src/index.mjs
+++ b/packages/eslint/src/index.mjs
@@ -47,6 +47,7 @@ export const eslint = [
     ...jestConfig,
     ...localRulesConfig,
     ...chaiFriendlyConfig,
+
     jsxA11y.flatConfigs.recommended,
 
     // Tests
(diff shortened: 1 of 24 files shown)


packages/eslint/src/index.mjs
```diff
@@ -47,7 +48,9 @@ export const eslint = [
     ...jestConfig,
     ...localRulesConfig,
     ...chaiFriendlyConfig,
+
     jsxA11y.flatConfigs.recommended,
+    pluginCypress.configs.recommended,
```
peter-sanderson: I am thinking, can we have this required not on global level, but rather on the local level where we have cypress used. 

So the config is loaded only in case cypress is used there.

I am just thinking laud here. Maybe we have cypress scattered across many packages. Also in case we add it to the new package and we forget about importing this, no cypress ESlint will be checked.

peter-sanderson (approved):
Thanks! Good job!

packages/eslint/src/index.mjs
HajekOndrej: I have changed it. There is no other cypress afaik and I want to get rid of it completely anyway at some point.

PR #15381, opened 2024-11-13
fix(e2e): Switched firmware to 2-latest in onboarding tests

HajekOndrej:
## Description

Hotfix for failing tests. With 2-main firmware the tenv gets stuck.

## Related Issue

Resolve

### DIFF
--- a/packages/suite-web/e2e/tests/onboarding/t2t1/t2t1-create-wallet.test.ts
+++ b/packages/suite-web/e2e/tests/onboarding/t2t1/t2t1-create-wallet.test.ts
@@ -1,23 +1,19 @@
 // @group_device-management
 // @retry=2
 
-import { onDeviceCompromisedModal } from '../../../support/pageObjects/deviceCompromisedObject';
-
 describe('Onboarding - create wallet', () => {
     beforeEach(() => {
         cy.task('startBridge');
         cy.viewport('macbook-13').resetDb();
         cy.prefixedVisit('/');
+        cy.disableFirmwareHashCheck();
     });
 
     it('Success (Shamir backup)', () => {
         // note: this is an example of test that can not be parametrized to be both integration (isolated) test and e2e test.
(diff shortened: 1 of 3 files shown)


packages/suite-web/e2e/tests/onboarding/t2t1/t2t1-create-wallet.test.ts
```diff
@@ -1,23 +1,19 @@
...
 
     it('Success (Shamir backup)', () => {
         // note: this is an example of test that can not be parametrized to be both integration (isolated) test and e2e test.
         // the problem is that it always needs to run the newest possible emulator. If this was pinned to use emulator which is currently
         // in production, and we locally bumped emulator version, we would get into a screen saying "update your firmware" and the test would fail.
-        cy.task('startEmu', { wipe: true, model: 'T2T1', version: '2-main' });
-
-        // firmware revision check will fail with 2-main version, but we just need to get through onboarding, so it's fine to just dismiss it
-        onDeviceCompromisedModal.ignoreDeviceCompromisedWarning();
+        cy.task('startEmu', { wipe: true, model: 'T2T1', version: '2-latest' });
```
Lemonexe, line 16: I came across this line before, and my interpretation of the comment above was that I should leave `2-latest` here, because this test specifically requires it. Same for `t2t1-recovery-fail.test.ts`.

But it seems there is no problem with it, or we'll see later I guess . If it fixes PR tests, that's most important 👍

Lemonexe (approved):
Thanks! Tests are passing, so LGTM..

packages/suite-web/e2e/tests/onboarding/t2t1/t2t1-create-wallet.test.ts
HajekOndrej: You mean `2-main` should be left there, right?  I will try to find out, but for now at least it passes and to me it doesn't seem like a bad test with the latest firmware.

PR #15382, opened 2024-11-13
chore(message-system): warn about pos backend issues

MiroslavProchazka:
Adding a new sequence to the messaging system to warn users on Suite >= 24.9.2 to update the app to resolve ongoing issues with backends.

Also lowering priority on the older Polygon message so they do not collide.

Demanded in [here](https://satoshilabs.slack.com/archives/C02F11Q9UCR/p1731526025085689) 

Tracked in Notion [here](https://www.notion.so/satoshilabs/Warn-users-about-onboing-isues-with-Polygon-BE-13ddc526060680f7bf69e4081cf8dde2?pvs=4)

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
-    "timestamp": "2024-09-26T00:00:00+00:00",
-    "sequence": 63,
+    "timestamp": "2024-11-14T00:00:00+00:00",
+    "sequence": 64,
     "actions": [
         {
             "conditions": [
@@ -107,7 +107,7 @@
             ],
             "message": {
                 "id": "c8ca8958-7868-467f-9278-f072edc234ac",
-                "priority": 92,
+                "priority": 91,
                 "dismissible": true,
                 "variant": "info",
                 "category": "banner",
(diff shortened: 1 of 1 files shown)


evgenysl:
There's no banner at 24.9.2 and 24.8.3

vdovhanych (approved):

suite-common/message-system/config/config.v1.json
```diff
@@ -147,6 +147,64 @@
...
+            "conditions": [
+                {
+                    "settings": [
+                        {
+                            "matic": true
+                        }
+                    ],
+                    "environment": {
+                        "desktop": "<=24.9.2",
+                        "mobile": "!",
```
PeKne, line 160: we will be releasing POL for mobile (24.11.1) next week. I guess that this message should be displayed there as well.

MiroslavProchazka: But you will be already using the correct symbol (POL) and new backends, right? This message is aimed for users on older Suites that are currently connected to old BE with MATIC as a symbol (so before migration).

suite-common/message-system/config/config.v1.json
```diff
@@ -147,6 +147,64 @@
...
+        {
+            "conditions": [
+                {
+                    "settings": [
+                        {
+                            "matic": true
+                        }
+                    ],
+                    "environment": {
+                        "desktop": "<=24.9.2",
```
tomasklim, line 159: i think this can be `*` as web, it should be filtered out by `settings` `matic: true` which does not exist in new versions

PR #15383, opened 2024-11-13
chore(mobile) : expo upgrade sdk 52 🚀

Nodonisko:
## Related Issue

Resolve


matejkriz (approved):
App icon and splashscreen are somewhat broken, but let's fix that as a followup to save more conflicts.

PR #15386, opened 2024-11-14
test(transport-test): use 2-latest emu (main has some breaking change)

mroz22:
2-main should only run in nightly jobs,


HajekOndrej (approved):

karliatto (approved):
Ok, but wouldn't be useful to have also the same tests to run nightly with `2-main`?

PR #15388, opened 2024-11-14
chore: connect examples node maintenance

karliatto:
## Description
Some maintenance for connect-examples in this case the `node` using it in typescript so it plays well in the monorepo.


martykan (approved):
Works 👍

PR #15389, opened 2024-11-14
test(connect): specify transport

mroz22:
- add testing of different bridge implementations to the nightly pipeline. 
- there is an [issue](https://github.com/trezor/trezor-suite/actions/runs/11854311553/job/33038450097) with node-bridge variant, so this is going to be a followup

### DIFF
--- a/.github/workflows/test-connect.yml
+++ b/.github/workflows/test-connect.yml
@@ -70,21 +70,26 @@ jobs:
       dailyMatrix: ${{ steps.set-matrix-daily.outputs.dailyMatrix }}
       otherDevicesMatrix: ${{ steps.set-matrix-other-devices.outputs.otherDevicesMatrix }}
       allFwsMatrix: ${{ steps.set-matrix-all-firmwares.outputs.allFwsMatrix }}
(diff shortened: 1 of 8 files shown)


docker/docker-connect-test.sh
```diff
@@ -53,10 +54,12 @@ USE_WS_CACHE=true
 PATTERN=""
 FIRMWARE_MODEL=""
 RANDOMIZE=false
+TRANSPORT="2.0.32"
```
mroz22: 2.0.33

.github/workflows/test-connect.yml
```diff
@@ -158,3 +163,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix: ${{ fromJson(needs.set-matrix.outputs.otherDevicesMatrix) }}
+
+  all-transports:
+    needs: [build, set-matrix]
+    name: all-transports-api ${{ matrix.key }} {{ matrix.transport }}
```
martykan:
```suggestion
    name: all-transports-api ${{ matrix.key }}
```
It's missing $ so it doesn't show correctly, but it can be removed altogether since it's shown in testDescription

marekrjpolak (approved):

PR #15413, opened 2024-11-15
feat(suite-common): remove solana from graph

### LINKED ISSUE #15247: Exclude some networks from graph and adjust UI
For various reasons we want to exclude some networks from graph and not show the graph in the account detail as well. The main example is Solana (desktop also does not show Ripple), where fetching txns takes a long time and costs a lot of resources.

We need to adjust UI for this:

- [x] Hide graph on dashboard if user has only graph-disabled networks
- [x] Hide graph in disabled networks account details
- [x] Tell user that some networks (SOL) are not part of the graph
- [x] remove infinite scrolling and introduce button to load more txns

**[FIGMA](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Bugfixes?node-id=0-1&node-type=canvas&t=SxZFvZrePzBclO0J-0)**

vytick:
- Hide graph on dashboard if user has only graph-disabled networks
- Hide graph in disabled networks account details
- Tell user that some networks (SOL) are not part of the graph
- remove infinite scrolling and introduce button to load more txns

## Related Issue

Resolve #15247

## Screenshots:

https://github.com/user-attachments/assets/3788b571-e7a2-45ec-81f0-e0752fdc56cf


Nodonisko:
Why do we hide entire graph? Why not show only that banner? And why do you show that banner when graph is hidden? It doesn't make much sense IMHO

suite-common/graph/src/constants.ts
```diff
@@ -9,10 +9,14 @@ export const LOCAL_BALANCE_HISTORY_COINS = [
...
 export type LocalBalanceHistoryCoin = (typeof LOCAL_BALANCE_HISTORY_COINS)[number];
 
 export const isLocalBalanceHistoryCoin = (
     symbol: NetworkSymbol,
 ): symbol is LocalBalanceHistoryCoin => isArrayMember(symbol, LOCAL_BALANCE_HISTORY_COINS);
+
+export const IGNORED_BALANCE_HISTORY_COINS = ['sol'] satisfies Array<NetworkSymbol>;
+export const isIgnoredBalanceHistoryCoin = (
+    symbol: NetworkSymbol,
+): symbol is LocalBalanceHistoryCoin => isArrayMember(symbol, IGNORED_BALANCE_HISTORY_COINS);
```
PeKne: The ignored coin probably should not be typd as a `LocalBalanceHistoryCoin`.
```suggestion
): symbol => isArrayMember(symbol, IGNORED_BALANCE_HISTORY_COINS);
```

suite-native/intl/src/en.ts
```diff
@@ -35,6 +35,9 @@ export const en = {
         },
     },
     moduleHome: {
+        ignoredNetworks: {
```
PeKne: nit: 
```suggestion
        graphIgnoredNetworks: {
```

suite-native/module-accounts-management/src/components/AccountDetailGraph.tsx
```diff
@@ -47,16 +52,23 @@ export const AccountDetailGraph = ({ accountKey, tokenContract }: AccountDetailG
         <VStack spacing="sp24">
             <AccountDetailGraphHeader accountKey={accountKey} tokenAddress={tokenContract} />
```
PeKne: Could you please rename this component please? Now it is kind of confusing that `GraphHeader` is rendered for accounts that do not support graph. Maybe change it to something like `AccountDetailHeader`?

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraphHeader.tsx
```diff
@@ -10,10 +10,12 @@ import {
     referencePointAtom,
     selectedPointAtom,
 } from '../portfolioGraphAtoms';
+import { selectHasDeviceHistoryEnabledAccounts } from '@suite-native/graph/src/selectors';
 
 export const PortfolioGraphHeader = () => {
```
PeKne: If you will be renanimg the AccountDetailGraphHeader, please rename this component too.

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
```diff
@@ -65,6 +67,38 @@ type TransactionListItem =
     | (TypedTokenTransferWithTx | MonthKey)
     | (WalletAccountTransaction | MonthKey);
 
+type TransactionsListFooterProps = {
+    isLoading: boolean;
+    canShowButton: boolean;
+    onButtonPress: () => void;
+};
+
+const TransactionsListFooter = ({
```
PeKne: This file is already too long. Could you please move this component to a separate file?

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
```diff
@@ -152,7 +186,12 @@ export const TransactionList = ({
             areTokensIncluded,
         ),
     );
-    const isLoaderVisible = isLoadingTransactions && transactions.length === 0;
+
+    const areAllTxnsFetched = useSelector((state: TransactionsRootState) =>
```
PeKne: Call of this selector can be moved to the `TransactionsListFooter` so it does not have to be passed as a prop there. It is not used anywhere else in this file.

suite-native/transactions/src/components/TransactionsList/TransactionListItem.tsx
```diff
@@ -71,7 +71,7 @@ export const TransactionListItemValues = ({
                         <SignValueFormatter value={getTransactionValueSign(transaction.type)} />
                     )}
                     <CryptoToFiatAmountFormatter
-                        value={transaction.amount}
+                        value={Math.abs(Number(transaction.amount))}
```
PeKne: Why is the `Math.abs` needed here and also in `TokenToFiatAmountFormatter` now? Is it because of some specific feature of Solana?

What about modyfing the `...Formatter` components itself, so the `.abs()` is applied there? It should prevent us from needing the `abs()` anywhere else in the future.

suite-common/graph/src/graphDataFetching.ts
```diff
@@ -153,6 +157,12 @@ const getAccountBalanceHistory = async ({
     }
 
     const getBalanceHistory = async () => {
+        if (isIgnoredBalanceHistoryCoin(coin)) {
```
PeKne, line 160: Is this needed? I believe that the networkSymbol of a graph-disabled coin should not me passed to the graph calculation at all? What do you think about filtering the coins before passing them to the graph calculation? Then this change nor the change in `graphBalanceEvents` would not be needed.

PeKne (changes requested):
I have some questions. But overall it looks good.

suite-native/transactions/src/components/TransactionsList/TransactionListItem.tsx
vytick: for solana `-` sign is part of the amount which causes visual inconsistencies (double -). This is similar mitigation that desktop does.

suite-common/graph/src/graphDataFetching.ts
vytick: If it goes through for solana, `TrezorConnect.blockchainGetAccountBalanceHistory` is called, but there is no method for this and it throws error which is rendered in the graph afterwards

vytick:
> Why do we hide entire graph? Why not show only that banner? And why do you show that banner when graph is hidden? It doesn't make much sense IMHO

We hide the graph for portfolio, when there is only solana in the portfolio. If there is any other network, the graph is visible. The banner is shown, whenever Solana is part of portfolio even in case the graph is not, so you know the reason for this.

suite-native/transactions/src/components/TransactionsList/TransactionListItem.tsx
vytick: 🤔  but it would be needed on more places and it would be better to handle separately. I'll remove this and create issue

suite-native/module-accounts-management/src/components/AccountDetailGraph.tsx
```diff
@@ -12,6 +12,8 @@ import { TokenAddress } from '@suite-common/wallet-types';
 
 import { AccountDetailGraphHeader } from './AccountDetailGraphHeader';
 import { referencePointAtom, selectedPointAtom } from '../accountDetailGraphAtoms';
+import { selectIsHistoryEnabledAccountByAccountKey } from '@suite-native/graph/src/selectors';
```
Nodonisko, line 12: please import from package directly

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraphHeader.tsx
```diff
@@ -10,10 +10,12 @@ import {
     referencePointAtom,
     selectedPointAtom,
 } from '../portfolioGraphAtoms';
+import { selectHasDeviceHistoryEnabledAccounts } from '@suite-native/graph/src/selectors';
```
Nodonisko: import directly from package

suite-native/transactions/src/components/TransactionsList/TokenTransferListItem.tsx
```diff
@@ -63,7 +63,7 @@ export const TokenTransferListItemValues = ({
         <>
             <TokenToFiatAmountFormatter
                 networkSymbol={transaction.symbol}
-                value={tokenTransfer.amount}
+                value={Math.abs(Number(tokenTransfer.amount))}
```
Nodonisko: can you just use `decimals={0}` ?

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
```diff
@@ -152,7 +186,12 @@ export const TransactionList = ({
...
+        selectAreAllAccountTransactionsLoaded(state, accountKey),
+    );
+
+    const isLoaderVisible = isLoadingTransactions;
```
Nodonisko: not sure if this has any added value

suite-common/graph/src/constants.ts
```diff
@@ -9,10 +9,14 @@ export const LOCAL_BALANCE_HISTORY_COINS = [
...
     'xrp',
-    'sol',
 ] satisfies Array<NetworkSymbol>;
 export type LocalBalanceHistoryCoin = (typeof LOCAL_BALANCE_HISTORY_COINS)[number];
 
 export const isLocalBalanceHistoryCoin = (
     symbol: NetworkSymbol,
 ): symbol is LocalBalanceHistoryCoin => isArrayMember(symbol, LOCAL_BALANCE_HISTORY_COINS);
+
+export const IGNORED_BALANCE_HISTORY_COINS = ['sol'] satisfies Array<NetworkSymbol>;
```
Nodonisko, line 21: maybe add some comments why we need to ignore some coins

Nodonisko (commented):
few nits but it looks fine

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraphHeader.tsx
vytick: it seems Cursor is not that smart

suite-native/transactions/src/components/TransactionsList/TransactionListItem.tsx
vytick: https://github.com/trezor/trezor-suite/issues/15424

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
vytick: 10127190bd

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
vytick: 10127190bd

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraphHeader.tsx
vytick: 10127190bd

suite-native/module-accounts-management/src/components/AccountDetailGraph.tsx
vytick: 10127190bd

suite-native/intl/src/en.ts
vytick: 10127190bd

suite-common/graph/src/constants.ts
vytick: 10127190bd

PeKne (approved):
:shipit:

suite-native/transactions/src/components/TransactionsList/TokenTransferListItem.tsx
vytick: removed

suite-common/graph/src/constants.ts
vytick: added in 14995a1a5b

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
vytick: good point, removed in 14995a1a5b

suite-native/module-home/src/screens/HomeScreen/components/PortfolioGraphHeader.tsx
vytick: ef74c791c9

suite-native/module-accounts-management/src/components/AccountDetailGraph.tsx
vytick: ef74c791c9

Issue #7406, opened 2023-01-17
BIP-329 style labeling import/export


prusnak:
There is a BIP-329 standard for labeling now. It should be a low-hanging fruit to implement import/export of such labeling format. 

More info: https://github.com/bitcoin/bips/blob/master/bip-0329.mediawiki

xavierfiechter:
@prusnak any progress on BIP-329 export/import for address/txn/outputs?

prusnak:
> @prusnak any progress on BIP-329 export/import for address/txn/outputs?

No, would you like to help?

xavierfiechter:
@prusnak I am building https://labelbase.space, curating https://bip329.org and focusing on other projects related to BIP-329. So, yes, I'd love to help. I can consult on labeling and UX topics, but I am not familiar with TypeScript.

Lemonexe:
This is already done in Desktop, Web, and Mobile ✔️

PR #15421, opened 2024-11-18
metadata export

mroz22:
image
 

- [x] export raw files to make them transferable between google drive and dropbox https://satoshilabs.slack.com/archives/CKN5C51CJ/p1731686820096629
- #7406 - possibly reformat data to match standard export/import format

### DIFF
--- a/packages/suite/src/views/settings/SettingsDebug/Metadata.tsx
+++ b/packages/suite/src/views/settings/SettingsDebug/Metadata.tsx
@@ -1,0 +1,33 @@
+import { useState } from 'react';
+
+import { Button } from '@trezor/components';
+
+import { ActionColumn, SectionItem, TextColumn } from 'src/components/suite';
+import { useDispatch } from 'src/hooks/suite';
+import { exportMetadataToLocalFile } from 'src/actions/suite/metadataActions';
+
+export const Metadata = () => {
+    const dispatch = useDispatch();
+    const [exporting, setExporting] = useState(false);
+
+    const onClick = () => {
+        setExporting(true);
+        dispatch(exportMetadataToLocalFile()).finally(() => {
+            setExporting(false);
+        });
+    };
+
+    return (
+        <SectionItem data-testid="@settings/debug/metadata">
+            <TextColumn
+                title="Export"
+                description="Export labeling files to your computer. You may use this to transfer your labeling files from your Google drive account to your Dropbox account."
+            />
+            <ActionColumn>
+                <Button onClick={onClick} isDisabled={exporting} isLoading={exporting}>
+                    Export
+                </Button>
+            </ActionColumn>
+        </SectionItem>
+    );
+};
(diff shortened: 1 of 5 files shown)


bosomt:
**QA OK**

- tested on 2 independent wallets and devices 
- exported from dropbox by mistake - lost of 429 errors from Dropbox, expected as export from dropbox is not needed/supported
- exported from Google, archive file was generated
- imported into Dropbox into Trezor directory
- After i loaded wallet account/wallet/transactions labels were present 

**Info:**
 - Suite version: desktop 24.12.0 (fe586851abda63dde5fa47366758545b9c73cd08)
 - OS: MacIntel
 - Screen: 1512x982

mroz22:
Thanks @bosomt. It was a blind shot by me so I am happy to see it works. The fact that it works makes it low-hanging fruits now. You can only add some toasts to handle the error branches and that's it. 

It can maybe already be useful for support? Also I was thinking that we maybe could add hidden routes so that support doesn't need to ask users to enable debug.

karliatto:
@mroz22 I have been testing and it works, I just needed rebasing it before otherwise I was getting some errors not sure why.

I am having a look if some error handling in the UI is required or something.

But from observation and some testing I think this will open at least 2 follow-up tasks:

* We need to allow importing.
* I think in order to be compatible with https://github.com/bitcoin/bips/blob/master/bip-0329.mediawiki it requires to be in unencrypted .json so it can be import by other wallets as well.

mroz22:
> I am having a look if some error handling in the UI is required or something.

I think I already forced pushed some notifications toasts to handle potential errors. So for purposes of debug mode, it should be ok. 

> We need to allow importing.

yes, next step.

> I think in order to be compatible with https://github.com/bitcoin/bips/blob/master/bip-0329.mediawiki it requires to be in unencrypted .json so it can be import by other wallets as well.

yes, next step.

packages/suite/src/views/settings/SettingsDebug/Metadata.tsx
```diff
@@ -0,0 +1,33 @@
...
+        dispatch(exportMetadataToLocalFile()).finally(() => {
+            setExporting(false);
+        });
+    };
+
+    return (
+        <SectionItem data-testid="@settings/debug/metadata">
+            <TextColumn
+                title="Export"
+                description="Export labeling files to your computer. You may use this to transfer your labeling files from your Google drive account to your Dropbox account."
```
karliatto, line 24: Nit: This is not translated but I guess for it is fine for debug menu

karliatto (approved):
Looks good to me!
I really like the idea of extending Suite export/import functionality. It could be very useful for users.

packages/suite/src/views/settings/SettingsDebug/Metadata.tsx
mroz22: yeah that should be good for now.

PR #15436, opened 2024-11-18
fix(mobile): missing date 📅 for transactions

### LINKED ISSUE #15433: Transaction history not showing date and time of tx
**Describe the bug**
Transaction history of all accounts is not showing date and time of transaction.

**Info:**

-   Mobile Suite Version: 24.11.1  4cc822b
-   OS: name, version

**How to reproduce**

1. Connect a device
2. Go to any account to see transaction history

**Screenshots**
image

Nodonisko:
## Description

Fix dates for transactions and input/output sheet details.

There were switched params in selectors because there are both just string TS was not aware of it => We should use branded types as much as possible.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15433

### DIFF
--- a/suite-native/notifications/src/components/TransactionNotification.tsx
+++ b/suite-native/notifications/src/components/TransactionNotification.tsx
@@ -88,7 +88,7 @@ export const TransactionNotification = ({
     );
 
     const transactionTargetAddress = useSelector((state: TransactionsRootState) =>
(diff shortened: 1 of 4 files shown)


matejkriz (approved):
I'm not sure if the emoji in commit message is needed 🙉 😆  But good spot, that mistake was very easy to make 🤔

Nodonisko:
Commit messages are boring, let's make them more fun 😅

PR #15445, opened 2024-11-19
Radical Solana refresh optimizations

### LINKED ISSUE #15641: User cannot send Solana because we set low fees
**Describe the bug**
Users are sometimes not able to send Solana transactions and have to try repeatedly. 
This is most most likely happens when fees on the network are high, because we are setting a very low hardcoded minimum fee

User report: https://www.reddit.com/r/TREZOR/comments/1h138m4/comment/lzazrln/

> Sometimes when i try to send my USDC to another wallet i encounter this
> error:
> TransactionExpiredBlockheightExceededError: Signature
> sometimes I have to wait like 20 minutes and try to send again later in hopes it will work or spam it a lot until it goes through. I also read it could be because of network congestion and trezor putting the lowest fees possible
> 

**How to reproduce**

1. Go to Solana account, when the average network fees are high
2. Try to send some tokens
3. The transaction might fail because of fees and you get: `TransactionExpiredBlockheightExceededError: Signature`

**Expected behavior**
The transaction does not fail, is approved, and user is able to send tokens on first try

martykan:
## Description

Reduce the amount of periodic calls for Solana

1. Check if the account was updated only based on txids. This means that it uses only 1 request per account + 1 per token. Previously this was more than 5 requests per account + 1 per token. 
2. Remove Solana block subscription from Suite's side and move it to send form only, at a reduced interval of 50 seconds.
3. Fixed baseFee calculation, increase priority fee unit price x3, added simulations for precise calculation of compute limit, resulting in overall lower fees but higher priority. 
4. Fixed Solana account subscriptions. Added solution to reconnect Solana backend when subscription goes down.
5. Changed configuration for sending TX to run the preflight check and allow retries
6. Fixed "send max" in send form

## Benchmark

Count of all Solana requests when opening a wallet with 4 accounts and several tokens and waiting for 5 min (including discovery)

- Before: 449 (121 discovery + 82 per minute)
- After: 200 (72 discovery + 32 per minute)

### DIFF
--- a/packages/blockchain-link-types/src/params.ts
+++ b/packages/blockchain-link-types/src/params.ts
@@ -56,4 +56,5 @@ export interface AccountInfoParams {
         ledger: number;
         seq: number;
     };
+    tokenAccountsPubKeys?: string[]; // solana only, token accounts to fetch txids for
 }
--- a/packages/blockchain-link/src/workers/solana/fee.ts
+++ b/packages/blockchain-link/src/workers/solana/fee.ts
@@ -1,15 +1,19 @@
 import {
     Address,
+    Base64EncodedWireTransaction,
     CompiledTransactionMessage,
     decompileTransactionMessage,
     getBase64Decoder,
     getCompiledTransactionMessageEncoder,
     GetFeeForMessageApi,
     GetRecentPrioritizationFeesApi,
+    getTransactionEncoder,
     isWritableRole,
-    MicroLamports,
     pipe,
     Rpc,
+    SignaturesMap,
+    SimulateTransactionApi,
+    TransactionMessageBytes,
     TransactionMessageBytesBase64,
 } from '@solana/web3.js';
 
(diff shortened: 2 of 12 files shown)


packages/blockchain-link-types/src/params.ts
```diff
@@ -56,4 +56,5 @@ export interface AccountInfoParams {
         ledger: number;
         seq: number;
     };
+    tokenAccounts?: string[]; // solana only, token accounts to fetch txids for
```
vytick: wdyt about smthng like
```suggestion
    tokenAccountsPubKeys?: string[]; // solana only, token accounts to fetch txids for
```

packages/blockchain-link/src/workers/solana/fee.ts
```diff
@@ -35,31 +54,40 @@ export const getBaseFee = async (api: Connection, message: Message) => {
...
+        .times(computeUnitLimit.toString())
         .decimalPlaces(0, BigNumber.ROUND_UP)
-        .toNumber();
+        .toString(10);
 
-    return { computeUnitPrice: computeUnitPrice.toNumber(), computeUnitLimit, fee };
+    return {
+        computeUnitPrice,
+        computeUnitLimit,
+        fee: BigInt(fee) as MicroLamports,
```
vytick: It seems that here it is casting to microlamports but previous computation of fee already did microlamports -> lamports 🤔. Maybe this is not needed at all or am I wrong?

martykan: Yeah seems wrong and also it doesn't really do anything, but this should be dealt with in [solana-web3-js-two-dot-oh](https://github.com/trezor/trezor-suite/pull/15536). 
I will change the target branch here to separate the commits

vytick (approved):
Looks good 👍

vytick (approved):

tomasklim (approved):

bosomt:
**QA OK**

Solana works as expected, 
- discovery, add accounts
- transactions - sol `0.23541798 RENDER sent from Solana #2` and tokens `0.23541798 RENDER sent from Solana #2`


**Info:**
 - Suite version: desktop 24.12.0 (f7dd25b803538b443f07f880f4fdceaa4faadd1a)
 - OS: MacIntel
 - Screen: 1512x982

PR #15450, opened 2024-11-19
fix(suite): count total txs amount on solana

### LINKED ISSUE #12992: Fiat in tx history in day header is incorrect
**Describe the bug**
I've sent solana so it should be negative, not positive
Screenshot 2024-06-20 at 17 06 31

enjojoy:
## Description

While on other chains tx amout is sign agnostic, on solana the outgoing transactions have a sign of negativity. So they should be handled with addition instead of subtraction, that changes the sign to plus. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12992

### Before: 

image

### After:

image

### DIFF
--- a/suite-common/wallet-utils/src/transactionUtils.ts
+++ b/suite-common/wallet-utils/src/transactionUtils.ts
@@ -173,7 +173,11 @@ export const sumTransactions = (transactions: WalletAccountTransaction[]) => {
 
         // count in only if Inputs/Outputs includes my account (EVM does not need to)
         if (tx.type === 'sent') {
-            totalAmount = totalAmount.minus(amount);
+            if (tx.symbol === 'sol') {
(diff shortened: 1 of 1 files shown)


suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -173,7 +173,11 @@ export const sumTransactions = (transactions: WalletAccountTransaction[]) => {
 
         // count in only if Inputs/Outputs includes my account (EVM does not need to)
         if (tx.type === 'sent') {
-            totalAmount = totalAmount.minus(amount);
+            if (tx.symbol === 'sol') {
```
tomasklim, line 176: What is the reason that this network is different than the others?

tomasklim:
Closing in favor of this #15455

PR #15453, opened 2024-11-19
fix(connect-explorer): fix betaOnly condition

mroz22:
🤦 argh, of course, this is server side stuff.


mroz22:
I would expect this bbd2005 to work but it doesn't

PR #15457, opened 2024-11-19
Display all tokens in a multitoken tx

### LINKED ISSUE #8036: Tx with ERC-1155 shows only first NFT available, without # of units
**Describe the bug**

Right now, we show only first ERC-1155 within a transaction without units.
https://github.com/trezor/trezor-suite/blob/develop/suite-common/wallet-utils/src/transactionUtils.ts#L413
https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/components/suite/FormattedNftAmount.tsx#L48

**Info:**

-   Suite Version: 23.4.0

**How to reproduce**
https://ethereum.org/en/developers/docs/standards/tokens/erc-1155/

Example tx:
https://etherscan.io/tx/0xa3074c72e058a29f077f5c478e86dd600623c562939e5ce0a3a3cb65b7a4a706
https://eth1.trezor.io/api/v2/tx/0xa3074c72e058a29f077f5c478e86dd600623c562939e5ce0a3a3cb65b7a4a706

Example address with NFTs:
https://eth1.trezor.io/address/0x958B69371e5750Ca8e6436f122322d68b395f781
https://etherscan.io/token/0x05baf9f1961e1ed1dac222a1f97c479441b31b34?a=81#inventory

**Expected behavior**
- shows all NFTs of tx
- shows value of NFT

Screenshot 2023-04-07 at 21 51 54

enjojoy:
## Description

Fixed display of multitoken values for ERC-1155

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/8036

## Screenshots:

Before: 
image
After:
image

Before:
image
After:
image

Before:
image
After:
image

Before:
image
After:
image

### DIFF
--- a/packages/components/src/components/typography/Link/Link.tsx
+++ b/packages/components/src/components/typography/Link/Link.tsx
@@ -22,12 +22,13 @@ type AllowedLinkTextProps = Pick<TextPropsCommon, (typeof allowedLinkTextProps)[
 
 type AProps = TransientProps<AllowedLinkTextProps> & {
     $variant?: 'default' | 'nostyle' | 'underline';
+    $color?: string;
 };
 
 const A = styled.a<AProps>`
     text-decoration: none;
     cursor: pointer;
-    color: ${({ theme }) => theme.textDefault};
+    color: ${({ $color, theme }) => $color || theme.textDefault};
     font-weight: 500;
     display: inline-flex;
     align-items: center;
@@ -68,6 +69,7 @@ type LinkProps = AllowedLinkTextProps & {
     className?: string;
     variant?: 'default' | 'nostyle' | 'underline'; // Todo: refactor, variant has different meaning in our design system
     icon?: IconName;
+    color?: string;
     'data-testid'?: string;
 };
 
@@ -78,6 +80,7 @@ const Link = ({
     onClick,
     'data-testid': dataTest,
     children,
+    color,
     className,
     variant,
     typographyStyle = 'body',
@@ -99,6 +102,7 @@ const Link = ({
             $variant={variant}
             className={className}
             {...textProps}
+            $color={color}
         >
             {children}
             {icon && <Icon size={iconSize} name={icon} />}
--- a/packages/suite/src/components/suite/FormattedCryptoAmount.tsx
+++ b/packages/suite/src/components/suite/FormattedCryptoAmount.tsx
@@ -8,6 +8,8 @@ import {
     networkAmountToSmallestUnit,
 } from '@suite-common/wallet-utils';
 import { isSignValuePositive } from '@suite-common/formatters';
+import { Row } from '@trezor/components';
+import { spacings } from '@trezor/theme';
 
 import { useBitcoinAmountUnit } from 'src/hooks/wallet/useBitcoinAmountUnit';
 import { useSelector } from 'src/hooks/suite';
@@ -16,10 +18,6 @@ import { selectLanguage } from 'src/reducers/suite/suiteReducer';
 import { BlurUrls } from 'src/views/wallet/tokens/common/BlurUrls';
 
 import { RedactNumericalValue } from './RedactNumericalValue';
-
-const Container = styled.span`
-    max-width: 100%;
-`;
 
 const Value = styled.span`
     font-variant-numeric: tabular-nums;
@@ -45,8 +43,8 @@ export const FormattedCryptoAmount = ({
     signValue,
     disableHiddenPlaceholder,
     isRawString,
-    'data-testid': dataTest,
     className,
+    'data-testid': dataTest,
 }: FormattedCryptoAmountProps) => {
     const locale = useSelector(selectLanguage);
 
@@ -97,23 +95,25 @@ export const FormattedCryptoAmount = ({
     }
 
     const content = (
-        <Container className={className}>
-            {!!signValue && <Sign value={signValue} />}
-            <Value data-testid={dataTest}>
-                <RedactNumericalValue value={formattedValue} />
-            </Value>
+        <Row width="100%" gap={spacings.xxs} className={className}>
+            <Row>
+                {!!signValue && <Sign value={signValue} />}
+                <Value data-testid={dataTest}>
+                    <RedactNumericalValue value={formattedValue} />
+                </Value>
+            </Row>
             {formattedSymbol && (
                 <>
                     {' '}
                     <BlurUrls text={formattedSymbol} />
                 </>
             )}
-        </Container>
+        </Row>
     );
 
     if (disableHiddenPlaceholder) {
         return content;
     }
 
-    return <HiddenPlaceholder className={className}>{content}</HiddenPlaceholder>;
+    return <HiddenPlaceholder>{content}</HiddenPlaceholder>;
 };
--- a/packages/suite/src/components/suite/FormattedCryptoAmount.tsx
+++ b/packages/suite/src/components/suite/FormattedCryptoAmount.tsx
@@ -96,9 +96,9 @@ export const FormattedCryptoAmount = ({
 
     const content = (
         <Row width="100%" gap={spacings.xxs} className={className}>
-            <Row>
+            <Row data-testid={dataTest}>
                 {!!signValue && <Sign value={signValue} />}
-                <Value data-testid={dataTest}>
+                <Value>
                     <RedactNumericalValue value={formattedValue} />
                 </Value>
             </Row>
--- a/packages/suite/src/components/suite/FormattedNftAmount.tsx
+++ b/packages/suite/src/components/suite/FormattedNftAmount.tsx
@@ -1,52 +1,24 @@
-import styled from 'styled-components';
+import { useTheme } from 'styled-components';
 
 import { SignValue } from '@suite-common/suite-types';
-import { getNftTokenId } from '@suite-common/wallet-utils';
+import { isNftMultitokenTransfer } from '@suite-common/wallet-utils';
 import { TokenTransfer } from '@trezor/connect';
-import { variables } from '@trezor/components';
+import { Box, Column, Row, Text } from '@trezor/components';
+import { spacings } from '@trezor/theme';
 
-import { HiddenPlaceholder, Sign } from 'src/components/suite';
+import { HiddenPlaceholder, Sign, Translation } from 'src/components/suite';
 // importing directly, otherwise unit tests fail, seems to be a styled-components issue
 import { TrezorLink } from 'src/components/suite/TrezorLink';
 import { useSelector } from 'src/hooks/suite/useSelector';
-
-const Container = styled.div`
-    max-width: 100%;
-    display: flex;
-`;
-
-const Symbol = styled.div`
-    white-space: nowrap;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    max-width: 120px;
-`;
-
-const StyledTrezorLink = styled(TrezorLink)`
-    color: ${({ theme }) => theme.legacy.TYPE_GREEN};
-    text-decoration: underline;
-    display: flex;
-    font-size: ${variables.FONT_SIZE.TINY};
-    line-height: initial;
-    align-items: center;
-`;
-
-const NoLink = styled.div`
-    display: flex;
-`;
-
-const Id = styled.div`
-    white-space: nowrap;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    max-width: 145px;
-`;
+import { useTranslation } from 'src/hooks/suite';
+import { BlurUrls } from 'src/views/wallet/tokens/common/BlurUrls';
 
 export interface FormattedNftAmountProps {
     transfer: TokenTransfer;
     signValue?: SignValue;
     className?: string;
     isWithLink?: boolean;
+    alignMultitoken?: 'flex-end' | 'flex-start';
 }
 
 export const FormattedNftAmount = ({
@@ -54,36 +26,103 @@ export const FormattedNftAmount = ({
     signValue,
     className,
     isWithLink,
+    alignMultitoken = 'flex-end',
 }: FormattedNftAmountProps) => {
-    const id = getNftTokenId(transfer);
-
+    const theme = useTheme();
+    const { translationString } = useTranslation();
     const { selectedAccount } = useSelector(state => state.wallet);
     const { network } = selectedAccount;
-    const explorerUrl =
-        network?.networkType === 'ethereum'
-            ? `${network?.explorer.nft}/${transfer.contract}/${id}`
-            : undefined;
 
-    const idComponent = <Id>{id}</Id>;
-    const symbolComponent = transfer.symbol ? <Symbol>&nbsp;{transfer.symbol}</Symbol> : null;
+    const symbolComponent = transfer.symbol ? (
+        <Text maxWidth={120} ellipsisLineCount={1}>
+            {transfer.symbol}
+        </Text>
+    ) : null;
+
+    const isMultitoken = isNftMultitokenTransfer(transfer);
+
+    if (isMultitoken) {
+        const tokens = transfer.multiTokenValues;
+
+        return (
+            <HiddenPlaceholder>
+                <Column alignItems={alignMultitoken}>
+                    {tokens?.map((token, index) => (
+                        <Row key={`${token.id}-${index}`} gap={spacings.xxs}>
+                            <Row>
+                                {signValue ? <Sign value={signValue} /> : null}
+                                {transfer.name ? (
+                                    <BlurUrls
+                                        text={translationString('TR_COLLECTION_NAME_OF_TOKEN_ID', {
+                                            tokenValue: token.value,
+                                            collectionName: transfer.name,
+                                        })}
+                                    />
+                                ) : (
+                                    <Row gap={spacings.xxs}>
+                                        <Row>{token.value}x</Row>
+                                        <Translation id="TR_TOKEN_ID" />
+                                    </Row>
+                                )}
+                            </Row>
+                            {isWithLink && network?.networkType === 'ethereum' ? (
+                                <TrezorLink
+                                    href={`${network?.explorer.nft}/${transfer.contract}/${token.id}`}
+                                    color={theme.textSecondaryHighlight}
+                                    variant="underline"
+                                    typographyStyle="label"
+                                >
+                                    <Text maxWidth={145} ellipsisLineCount={1}>
+                                        {token.id}
+                                    </Text>
+                                </TrezorLink>
+                            ) : (
+                                <Text maxWidth={145} ellipsisLineCount={1}>
+                                    {token.id}
+                                </Text>
+                            )}
+                        </Row>
+                    ))}
+                </Column>
+            </HiddenPlaceholder>
+        );
+    }
 
     return (
         <HiddenPlaceholder>
-            <Container className={className}>
+            <Row className={className}>
                 {signValue ? <Sign value={signValue} /> : null}
-                ID:&nbsp;
+                <Box margin={{ right: spacings.xxs }}>
+                    <Translation id="TR_TOKEN_ID" />
+                </Box>
                 {isWithLink ? (
-                    <StyledTrezorLink href={explorerUrl}>
-                        {idComponent}
-                        {symbolComponent}
-                    </StyledTrezorLink>
+                    <TrezorLink
+                        href={
+                            network?.networkType === 'ethereum'
+                                ? `${network?.explorer.nft}/${transfer.contract}/${transfer.amount}`
+                                : undefined
+                        }
+                        color={theme.textSecondaryHighlight}
+                        variant="underline"
+                        typographyStyle="label"
+                    >
+                        <Row gap={spacings.zero}>
+                            <Text maxWidth={145} ellipsisLineCount={1}>
+                                {transfer.amount}
+                            </Text>
+                            &nbsp;
+                            {symbolComponent}
+                        </Row>
+                    </TrezorLink>
                 ) : (
-                    <NoLink>
-                        {idComponent}
+                    <Row gap={spacings.xxs}>
+                        <Text maxWidth={145} ellipsisLineCount={1}>
+                            {transfer.amount}
+                        </Text>
                         {symbolComponent}
-                    </NoLink>
+                    </Row>
                 )}
-            </Container>
+            </Row>
         </HiddenPlaceholder>
     );
 };
--- a/packages/suite/src/components/suite/FormattedNftAmount.tsx
+++ b/packages/suite/src/components/suite/FormattedNftAmount.tsx
@@ -4,7 +4,7 @@ import { SignValue } from '@suite-common/suite-types';
 import { isNftMultitokenTransfer } from '@suite-common/wallet-utils';
 import { TokenTransfer } from '@trezor/connect';
 import { Box, Column, Row, Text } from '@trezor/components';
-import { spacings } from '@trezor/theme';
+import { spacings, TypographyStyle } from '@trezor/theme';
 
 import { HiddenPlaceholder, Sign, Translation } from 'src/components/suite';
 // importing directly, otherwise unit tests fail, seems to be a styled-components issue
@@ -19,6 +19,7 @@ export interface FormattedNftAmountProps {
     className?: string;
     isWithLink?: boolean;
     alignMultitoken?: 'flex-end' | 'flex-start';
+    linkTypographyStyle?: TypographyStyle;
 }
 
 export const FormattedNftAmount = ({
@@ -27,6 +28,7 @@ export const FormattedNftAmount = ({
     className,
     isWithLink,
     alignMultitoken = 'flex-end',
+    linkTypographyStyle,
 }: FormattedNftAmountProps) => {
     const theme = useTheme();
     const { translationString } = useTranslation();
@@ -34,8 +36,8 @@ export const FormattedNftAmount = ({
     const { network } = selectedAccount;
 
     const symbolComponent = transfer.symbol ? (
-        <Text maxWidth={120} ellipsisLineCount={1}>
-            {transfer.symbol}
+        <Text ellipsisLineCount={1}>
+            <BlurUrls text={transfer.symbol} />
         </Text>
     ) : null;
 
@@ -70,7 +72,7 @@ export const FormattedNftAmount = ({
                                     href={`${network?.explorer.nft}/${transfer.contract}/${token.id}`}
                                     color={theme.textSecondaryHighlight}
                                     variant="underline"
-                                    typographyStyle="label"
+                                    typographyStyle={linkTypographyStyle}
                                 >
                                     <Text maxWidth={145} ellipsisLineCount={1}>
                                         {token.id}
@@ -104,7 +106,7 @@ export const FormattedNftAmount = ({
                         }
                         color={theme.textSecondaryHighlight}
                         variant="underline"
-                        typographyStyle="label"
+                        typographyStyle={linkTypographyStyle}
                     >
                         <Row gap={spacings.zero}>
                             <Text maxWidth={145} ellipsisLineCount={1}>
--- a/packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipAccount.tsx
+++ b/packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipAccount.tsx
@@ -1,19 +1,16 @@
-import styled from 'styled-components';
 import { TooltipProps } from 'recharts';
 
 import { NetworkSymbol } from '@suite-common/wallet-config';
 import { Formatters, useFormatters } from '@suite-common/formatters';
 import { SignOperator } from '@suite-common/suite-types';
+import { Row } from '@trezor/components';
+import { spacings } from '@trezor/theme';
 
 import { FormattedCryptoAmount } from 'src/components/suite/FormattedCryptoAmount';
 import { CommonAggregatedHistory, GraphRange } from 'src/types/wallet/graph';
 
 import type { CryptoGraphProps } from './TransactionsGraph';
 import { GraphTooltipBase } from './GraphTooltipBase';
-
-const StyledCryptoAmount = styled(FormattedCryptoAmount)`
-    margin-right: 2px;
-`;
 
 const formatAmount = (
     amount: string | undefined,
@@ -26,14 +23,16 @@ const formatAmount = (
     const { FiatAmountFormatter } = formatters;
 
     return (
-        <>
+        <Row>
             {amount && (
-                <StyledCryptoAmount
-                    value={amount}
-                    symbol={symbol}
-                    signValue={sign}
-                    disableHiddenPlaceholder
-                />
+                <Row margin={{ right: spacings.xxs }}>
+                    <FormattedCryptoAmount
+                        value={amount}
+                        symbol={symbol}
+                        signValue={sign}
+                        disableHiddenPlaceholder
+                    />
+                </Row>
             )}
 
             {fiatAmount && localCurrency && (
@@ -42,7 +41,7 @@ const formatAmount = (
                     <FiatAmountFormatter currency={localCurrency} value={fiatAmount} />)
                 </>
             )}
-        </>
+        </Row>
     );
 };
 
--- a/packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipBase.tsx
+++ b/packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipBase.tsx
@@ -3,7 +3,8 @@ import { useEffect } from 'react';
 import styled from 'styled-components';
 import { TooltipProps } from 'recharts';
 
-import { variables } from '@trezor/components';
+import { variables, Row } from '@trezor/components';
+import { spacings } from '@trezor/theme';
 
 import { Translation, FormattedDate } from 'src/components/suite';
 import { CommonAggregatedHistory, GraphRange } from 'src/types/wallet/graph';
@@ -78,15 +79,6 @@ const Col = styled.div`
     flex-direction: column;
 `;
 
-const Row = styled.div<{ $noBottomMargin?: boolean }>`
-    display: flex;
-    white-space: nowrap;
-    align-items: center;
-    justify-content: space-between;
-    padding: 0 8px;
-    margin-bottom: ${({ $noBottomMargin }) => ($noBottomMargin ? '0px' : '4px')};
-`;
-
 const Title = styled.span`
     font-weight: ${variables.FONT_WEIGHT.MEDIUM};
     margin-right: 20px;
@@ -113,12 +105,6 @@ const HighlightedAreaLeft = styled(HighlightedArea)`
 const HighlightedAreaRight = styled(HighlightedArea)`
     border-top-right-radius: 5px;
     border-bottom-right-radius: 5px;
-`;
-
-const Sign = styled.span<{ $color: string }>`
-    color: ${({ $color }) => $color};
-    width: 1ch;
-    margin-right: 4px;
 `;
 
 const formatDate = (date: Date, dateFormat: 'day' | 'month') => {
@@ -167,14 +153,16 @@ export const GraphTooltipBase = (props: GraphTooltipBaseProps) => {
             $boxWidth={props.viewBox!.width!}
             data-testid="@dashboard/customtooltip"
         >
-            <Row>
+            <Row margin={{ bottom: spacings.xxs, left: spacings.xs, right: spacings.xs }}>
                 <Title>{date && formatDate(date, dateFormat)}</Title>
             </Row>
 
             <ColsWrapper>
                 <Col>
                     {props.balance && (
-                        <Row>
+                        <Row
+                            margin={{ bottom: spacings.xxs, left: spacings.xs, right: spacings.xs }}
+                        >
                             <Title>
                                 <Translation id="TR_BALANCE" />
                             </Title>
@@ -182,13 +170,15 @@ export const GraphTooltipBase = (props: GraphTooltipBaseProps) => {
                     )}
 
                     <HighlightedAreaLeft>
-                        <Row>
+                        <Row
+                            margin={{ bottom: spacings.xxs, left: spacings.xs, right: spacings.xs }}
+                        >
                             <Title>
                                 <Translation id="TR_RECEIVED" />
                             </Title>
                         </Row>
 
-                        <Row $noBottomMargin>
+                        <Row margin={{ left: spacings.xs, right: spacings.xs }}>
                             <Title>
                                 <Translation id="TR_SENT" />
                             </Title>
@@ -198,20 +188,25 @@ export const GraphTooltipBase = (props: GraphTooltipBaseProps) => {
 
                 <Col>
                     {props.balance && (
-                        <Row>
+                        <Row
+                            margin={{ bottom: spacings.xxs, left: spacings.xs, right: spacings.xs }}
+                        >
                             <Value>
-                                <Sign $color="transparent">+</Sign>
-                                {props.balance}
+                                <Row margin={{ left: spacings.xs, right: spacings.xs }}>
+                                    {props.balance}
+                                </Row>
                             </Value>
                         </Row>
                     )}
 
                     <HighlightedAreaRight>
-                        <Row>
+                        <Row
+                            margin={{ bottom: spacings.xxs, left: spacings.xs, right: spacings.xs }}
+                        >
                             <Value>{props.receivedAmount}</Value>
                         </Row>
 
-                        <Row $noBottomMargin>
+                        <Row margin={{ left: spacings.xs, right: spacings.xs }}>
                             <Value>{props.sentAmount}</Value>
                         </Row>
                     </HighlightedAreaRight>
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
@@ -208,6 +208,7 @@ const EthereumSpecificBalanceDetailsRow = ({
                                 transfer={transfer}
                                 isWithLink
                                 alignMultitoken="flex-start"
+                                linkTypographyStyle="label"
                             />
                         ) : (
                             formatAmount(transfer.amount, transfer.decimals)
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
@@ -19,10 +19,9 @@ import {
 } from '@trezor/components';
 import { NetworkSymbol } from '@suite-common/wallet-config';
 
-import { FormattedCryptoAmount, Translation } from 'src/components/suite';
+import { FormattedCryptoAmount, FormattedNftAmount, Translation } from 'src/components/suite';
 import { useSelector } from 'src/hooks/suite/useSelector';
 import { UtxoAnonymity } from 'src/components/wallet';
-import { FormattedNftAmount } from 'src/components/suite/FormattedNftAmount';
 import { IOAddress } from 'src/components/suite/copy/IOAddress';
 
 import { AnalyzeInExplorerBanner } from './AnalyzeInExplorerBanner';
@@ -205,7 +204,11 @@ const EthereumSpecificBalanceDetailsRow = ({
                     </H4>
                     {tokens.map((transfer, index) => {
                         const value = isNftTokenTransfer(transfer) ? (
-                            <FormattedNftAmount transfer={transfer} isWithLink />
+                            <FormattedNftAmount
+                                transfer={transfer}
+                                isWithLink
+                                alignMultitoken="flex-start"
+                            />
                         ) : (
                             formatAmount(transfer.amount, transfer.decimals)
                         );
--- a/packages/suite/src/components/wallet/AmountComponent.tsx
+++ b/packages/suite/src/components/wallet/AmountComponent.tsx
@@ -1,5 +1,6 @@
 import { isNftTokenTransfer, formatAmount, getTxOperation } from '@suite-common/wallet-utils';
 import { TokenTransfer } from '@trezor/connect';
+import { TypographyStyle } from '@trezor/theme';
 
 import { FormattedNftAmount } from 'src/components/suite/FormattedNftAmount';
 import { FormattedCryptoAmount } from 'src/components/suite/FormattedCryptoAmount';
@@ -9,6 +10,7 @@ type AmountComponentProps = {
     withLink?: boolean;
     withSign?: boolean;
     alignMultitoken?: 'flex-end' | 'flex-start';
+    linkTypographyStyle?: TypographyStyle;
 };
 
 export const AmountComponent = ({
@@ -16,6 +18,7 @@ export const AmountComponent = ({
     withLink = false,
     withSign = false,
     alignMultitoken,
+    linkTypographyStyle,
 }: AmountComponentProps): React.ReactNode => {
     const operation = getTxOperation(transfer.type);
 
@@ -26,6 +29,7 @@ export const AmountComponent = ({
                 isWithLink={withLink}
                 signValue={withSign ? operation : null}
                 alignMultitoken={alignMultitoken}
+                linkTypographyStyle={linkTypographyStyle}
             />
         );
     }
--- a/packages/suite/src/components/wallet/AmountComponent.tsx
+++ b/packages/suite/src/components/wallet/AmountComponent.tsx
@@ -1,0 +1,44 @@
+import { isNftTokenTransfer, formatAmount, getTxOperation } from '@suite-common/wallet-utils';
+import { TokenTransfer } from '@trezor/connect';
+
+import { FormattedNftAmount } from 'src/components/suite/FormattedNftAmount';
+import { FormattedCryptoAmount } from 'src/components/suite/FormattedCryptoAmount';
+
+type AmountComponentProps = {
+    transfer: TokenTransfer;
+    withLink?: boolean;
+    withSign?: boolean;
+    alignMultitoken?: 'flex-end' | 'flex-start';
+};
+
+export const AmountComponent = ({
+    transfer,
+    withLink = false,
+    withSign = false,
+    alignMultitoken,
+}: AmountComponentProps): React.ReactNode => {
+    const operation = getTxOperation(transfer.type);
+
+    if (isNftTokenTransfer(transfer)) {
+        return (
+            <FormattedNftAmount
+                transfer={transfer}
+                isWithLink={withLink}
+                signValue={withSign ? operation : null}
+                alignMultitoken={alignMultitoken}
+            />
+        );
+    }
+
+    if (withSign) {
+        return (
+            <FormattedCryptoAmount
+                value={formatAmount(transfer.amount, transfer.decimals)}
+                symbol={transfer.symbol}
+                signValue={operation}
+            />
+        );
+    }
+
+    return formatAmount(transfer.amount, transfer.decimals);
+};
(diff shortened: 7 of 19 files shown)


enjojoy:
@qa  there are ERC-1155 transactions with multitokens on academic seed "v" passphrase on polygon and Ethereum Sepolia

packages/suite/src/components/FormattedMultitokenAmount.tsx
```diff
@@ -0,0 +1,74 @@
...
+    display: flex;
+    font-size: ${variables.FONT_SIZE.TINY};
+    line-height: initial;
+    align-items: center;
+`;
+
+export interface FormattedMultitokenAmountProps {
+    tokens: TokenTransfer['multiTokenValues'];
+    signValue?: SignValue;
+    className?: string;
```
jvaclavik: Can we do it wihtout className? I think you don't use it anywhere

packages/suite/src/components/FormattedMultitokenAmount.tsx
```diff
@@ -0,0 +1,74 @@
+import styled from 'styled-components';
+
+import { SignValue } from '@suite-common/suite-types';
+import { TokenTransfer } from '@trezor/connect';
+import { Column, Row, variables } from '@trezor/components';
+
+import { HiddenPlaceholder, Sign, TrezorLink } from 'src/components/suite';
+import { useSelector } from 'src/hooks/suite';
+
+const Id = styled.div`
```
jvaclavik: Couldn't we use `Text` component with related props? 🤔

jvaclavik: Couldn't we use `Text` component with related props? 🤔

packages/suite/src/components/FormattedMultitokenAmount.tsx
```diff
@@ -0,0 +1,74 @@
...
+`;
+
+const StyledTrezorLink = styled(TrezorLink)`
+    color: ${({ theme }) => theme.legacy.TYPE_GREEN};
+    text-decoration: underline;
+    display: flex;
+    font-size: ${variables.FONT_SIZE.TINY};
+    line-height: initial;
+    align-items: center;
+`;
```
jvaclavik: I'm not fan of duplicating this component again. Also we should use new non-legacy tokens (font and color). And maybe we could use wrapper not to override TrezorLink at all.

packages/suite/src/components/wallet/getAmountComponent.tsx
```diff
@@ -0,0 +1,69 @@
...
+    withLink?: boolean;
+    contract?: string;
+};
+
+export const getAmountComponent = ({
+    transfer,
+    useStyled = false,
+    withLink = false,
+    contract,
+}: GetAmountComponentProps) => {
```
jvaclavik: Logic in this function seems to me a bit complicated. Couldn't we improve it somehow?

tomasklim (changes requested):
Why `+` and `-` disappeared? 
Screenshot 2024-11-22 at 20 50 51

Why is `Id:` also underlined? 
Screenshot 2024-11-22 at 20 51 01

packages/suite/src/components/FormattedMultitokenAmount.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        <Row>
+                            {SignValueComponent}
+                            {token.value}x
+                        </Row>
+                        <Row gap={4}>
+                            {withLink && network?.networkType === 'ethereum' && contract ? (
+                                <StyledTrezorLink
+                                    href={`${network?.explorer.nft}/${contract}/${token.id}`}
+                                >
+                                    ID:
```
tomasklim: I would make `ID` translatable even though it will probably be same in all languages, it will be easier to change it in case somebody decides so

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -476,7 +476,9 @@ export const getTxOperation = (
 };
 
 export const isNftTokenTransfer = (transfer: TokenTransfer) =>
-    ['ERC1155', 'ERC721'].includes(transfer.standard || '');
+    ['ERC1155', 'ERC721'].includes(transfer?.standard || '');
+
+export const isMultitoken = (transfer: TokenTransfer) => !!transfer?.multiTokenValues;
```
tomasklim:
```suggestion
export const IsNftMultitokenTransfer = (transfer: TokenTransfer) => !!transfer?.multiTokenValues;
```

and btw it return `true` when multitokens array is defined but empty

packages/suite/src/components/FormattedMultitokenAmount.tsx
```diff
@@ -0,0 +1,74 @@
...
+}: FormattedMultitokenAmountProps) => {
+    const SignValueComponent = signValue ? <Sign value={signValue} /> : null;
+    const { selectedAccount } = useSelector(state => state.wallet);
+    const { network } = selectedAccount;
+
+    return (
+        <HiddenPlaceholder>
+            <Column className={className} alignItems="flex-start">
+                {tokens?.map(token => (
+                    <Row key={`${token.id}-${token.value}`} gap={4}>
```
tomasklim: does `token.value` make sense in the `key`?

packages/suite/src/components/FormattedMultitokenAmount.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        <Row>
+                            {SignValueComponent}
+                            {token.value}x
```
tomasklim: We use `+` and `-` icons for positive, negative values. Maybe it would make sense to have it also for `x`? @jvaclavik @adamhavel 
Screenshot 2024-11-23 at 9 59 44

and btw etherscan has `x` before the amount
Screenshot 2024-11-23 at 10 00 10

packages/suite/src/components/FormattedMultitokenAmount.tsx
tomasklim: etherscan btw uses Token ID, maybe that is better

tomasklim (changes requested):

tomasklim:
Btw for nft names, it would be great to use `BlurUrls` component. We should implement it into `FormattedNftAmount`. I agree with @jvaclavik that we should not duplicate this component and we should not create `FormattedMultitokenAmount`

packages/suite/src/components/FormattedMultitokenAmount.tsx
enjojoy: Can be Token ID, then I will rename it for every NFT case, not only multitoken

packages/suite/src/components/FormattedMultitokenAmount.tsx
enjojoy: yes, removed

packages/suite/src/components/FormattedMultitokenAmount.tsx
enjojoy: Yes, it's gone

packages/suite/src/components/FormattedMultitokenAmount.tsx
enjojoy: I deleted this component and used the non-legacy variables where needed

packages/suite/src/components/wallet/getAmountComponent.tsx
enjojoy: Now it's much simpler without all the styled components

suite-common/wallet-utils/src/transactionUtils.ts
enjojoy: I'm not sure it's a problem in this case, can this even be a case? I mean can a transfer have multitoken value defined but have transaction amount outside of it? I'm not so sure, will have to test it out

packages/suite/src/components/FormattedMultitokenAmount.tsx
enjojoy: yeah I used an index instead, apparently id can be same in some cases

enjojoy:
It would be nice to remake GraphTooltipAccount component, it's full of custom styles, I was playing around with it but decided to leave it for later to not make this PR too big

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
```diff
@@ -97,23 +95,25 @@ export const FormattedCryptoAmount = ({
     }
 
     const content = (
-        <Container className={className}>
-            {!!signValue && <Sign value={signValue} />}
-            <Value data-testid={dataTest}>
-                <RedactNumericalValue value={formattedValue} />
-            </Value>
+        <Row width="100%" flexWrap="nowrap" justifyContent={justifyContent} gap={4}>
```
seibei-iguchi: No need for flexWrap, use spacings for gap.

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
```diff
@@ -254,7 +243,7 @@ export const TransactionItem = memo(
                     </TxTypeIconWrapper>
 
                     <Content>
-                        <Description>
+                        <Row justifyContent="space-between" overflow="hidden" flexWrap="nowrap">
```
seibei-iguchi: No need for flexWrap

seibei-iguchi:
I've made lot of changes to the IODetails component since you made this branch, so you will have to do a painful rebase - sorry about that!

enjojoy:
> I've made lot of changes to the IODetails component since you made this branch, so you will have to do a painful rebase - sorry about that!

It's oki, thank you for letting me know

packages/suite/src/components/suite/CoinBalance.tsx
```diff
@@ -4,13 +4,15 @@ import { FormattedCryptoAmount } from 'src/components/suite';
 interface CoinBalanceProps {
     value: string;
     symbol: Account['symbol'] | (string & {});
+    justifyContent?: 'flex-end' | 'flex-start';
```
tomasklim: `FlexJustifyContent` would maybe make more sense here? but maybe not, you can just pick those from this type

enjojoy: Well I don't need other options in this component

tomasklim: so you can pick those two from `FlexJustifyContent`?

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
```diff
@@ -97,23 +97,25 @@ export const FormattedCryptoAmount = ({
...
+                {!!signValue && <Sign value={signValue} />}
+                <Value data-testid={dataTest}>
+                    <RedactNumericalValue value={formattedValue} />
+                </Value>
+            </Row>
             {formattedSymbol && (
                 <>
                     {' '}
-                    <BlurUrls text={formattedSymbol} />
+                    <BlurUrls text={formattedSymbol} />{' '}
```
tomasklim: should the {' '} be here?

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
```diff
@@ -35,6 +34,7 @@ export interface FormattedCryptoAmountProps {
     disableHiddenPlaceholder?: boolean;
     isRawString?: boolean;
     'data-testid'?: string;
+    justifyContent?: 'flex-end' | 'flex-start';
     className?: string;
```
tomasklim: not used anymore

packages/suite/src/components/suite/CoinBalance.tsx
enjojoy: sure

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
```diff
@@ -45,8 +45,8 @@ export const FormattedCryptoAmount = ({
     signValue,
     disableHiddenPlaceholder,
     isRawString,
+    justifyContent = 'flex-end',
     'data-testid': dataTest,
-    className,
```
tomasklim, line 49: all the occurences of `styled(FormattedCryptoAmount)` will stop working

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -477,7 +477,9 @@ export const getTxOperation = (
 };
 
 export const isNftTokenTransfer = (transfer: TokenTransfer) =>
-    ['ERC1155', 'ERC721'].includes(transfer.standard || '');
+    ['ERC1155', 'ERC721'].includes(transfer?.standard || '');
+
+export const isNftMultitokenTransfer = (transfer: TokenTransfer) => !!transfer?.multiTokenValues;
```
tomasklim: are the `?` needed?

packages/suite/src/support/messages.ts
```diff
@@ -5688,6 +5692,10 @@ const messages = defineMessagesWithTypeCheck({
         defaultMessage: '{amount} selected',
         description: 'Number of list items selected',
     },
+    TR_OF: {
+        id: 'TR_OF',
+        defaultMessage: 'of',
+    },
```
tomasklim: We do not do that, can you add whole structure here?

It should be something like `{sign} {tokenValue}x ID token {id}: of {name}`, right? (it is just wild guess)

packages/suite/src/components/wallet/getAmountComponent.tsx
```diff
@@ -0,0 +1,46 @@
...
+type GetAmountComponentProps = {
+    transfer: TokenTransfer;
+    useStyled?: boolean;
+    withLink?: boolean;
+    withSign?: boolean;
+    contract?: string;
+    alignMultitoken?: 'flex-end' | 'flex-start';
+};
+
+export const getAmountComponent = ({
```
tomasklim: we discussed yesterday that it should not be `get...`

packages/suite/src/components/wallet/getAmountComponent.tsx
```diff
@@ -0,0 +1,46 @@
+import { isNftTokenTransfer, formatAmount, getTxOperation } from '@suite-common/wallet-utils';
+import { TokenTransfer } from '@trezor/connect';
+
+import { FormattedNftAmount } from 'src/components/suite/FormattedNftAmount';
+import { FormattedCryptoAmount } from 'src/components/suite/FormattedCryptoAmount';
+
+type GetAmountComponentProps = {
+    transfer: TokenTransfer;
+    useStyled?: boolean;
```
tomasklim: where is this used?

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
enjojoy: Looks like it was updated 8 hours ago lol, oki will remove that

packages/suite/src/components/wallet/TransactionItem/CommonComponents.tsx
```diff
@@ -40,44 +41,19 @@ export const Content = styled.div`
...
-`;
-
-export const StyledFormattedCryptoAmount = styled(FormattedCryptoAmount)`
-    ${amountStyle}
-`;
-
-export const StyledFormattedNftAmount = styled(FormattedNftAmount)`
-    ${amountStyle}
-`;
+export const Id = ({ id }: { id?: string }) => {
```
tomasklim: Does it belong here?

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
enjojoy: 😳

packages/suite/src/components/wallet/TransactionItem/TransactionHeader.tsx
```diff
@@ -79,9 +82,11 @@ export const TransactionHeader = ({ transaction, isPending }: TransactionHeaderP
     const symbol = getTxHeaderSymbol(transaction)?.toUpperCase();
 
     return (
-        <Translation
-            id={getTransactionMessageId({ transaction, isPending })}
-            values={{ symbol, multiple: isMultiTokenTransaction }}
+        <BlurUrls
+            text={translationString(getTransactionMessageId({ transaction, isPending }), {
```
tomasklim, line 86: why `translationString` and not `Translation` component?

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipBase.tsx
```diff
@@ -200,8 +194,7 @@ export const GraphTooltipBase = (props: GraphTooltipBaseProps) => {
                     {props.balance && (
                         <Row>
                             <Value>
-                                <Sign $color="transparent">+</Sign>
-                                {props.balance}
+                                <Row $noBottomMargin>{props.balance}</Row>
```
tomasklim, line 197: Screenshot 2024-11-27 at 16 20 30
🤯

suite-common/wallet-utils/src/transactionUtils.ts
enjojoy: nope

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipBase.tsx
tomasklim: Is there some request to remove those signs?

packages/suite/src/support/messages.ts
enjojoy: I wouldn't add a sign there since it's a component, but I agree about the rest of the text

enjojoy: https://arc.net/l/quote/sbhhastt

suite-common/wallet-utils/src/transactionUtils.ts
tomasklim: but it is still there...

packages/suite/src/components/wallet/TransactionItem/TransactionHeader.tsx
enjojoy: Because BlurUrls accepts a string not a component or child 
image

packages/suite/src/components/wallet/getAmountComponent.tsx
enjojoy: It's not there anymore

packages/suite/src/components/wallet/getAmountComponent.tsx
enjojoy: Fixed

packages/suite/src/components/wallet/TransactionItem/CommonComponents.tsx
enjojoy: No, since it was used only in one component I moved it there https://github.com/trezor/trezor-suite/pull/15457/commits/8e054ee8acbd20b773f34b3e0b4134c64b39f7c0

packages/suite/src/support/messages.ts
enjojoy: This is fixed https://arc.net/l/quote/capnkscz

suite-common/wallet-utils/src/transactionUtils.ts
enjojoy: weird, I was pretty sure I pushed it, here it is https://github.com/trezor/trezor-suite/pull/15457/commits/5a7283911c014c77fb8ebe47f29fe43f80752792

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
enjojoy: It's back

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
enjojoy: Probably no need

packages/suite/src/components/suite/CoinBalance.tsx
enjojoy: This is not there anymore

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
...
-                    </StyledTrezorLink>
+                    <TrezorLink
+                        href={explorerUrl}
+                        color={theme.textSecondaryHighlight}
+                        variant="underline"
+                        typographyStyle="label"
+                        gap="zero"
+                    >
+                        <Id id={id} />
+                        &nbsp; {symbolComponent}
```
jvaclavik: This seems to me weird. You set gap to zero and then you use `&nbsp;` probably for having a gap. Why?

Btw I noticed that you use &nbsp; 5x in this PR, not sure if it's not overused.

packages/components/src/components/typography/Link/Link.tsx
```diff
@@ -99,6 +105,8 @@ const Link = ({
             $variant={variant}
             className={className}
             {...textProps}
+            $color={color}
+            $gap={gap}
```
jvaclavik: I think it doesn't belong here, couldn't you use Row inside of the link?

something like:

```
<Link>
  <Row as="span" gap={spacings.xs}>
    <A/>
    <B/>
  </Row>
</Link>

File: packages/suite/src/components/suite/FormattedCryptoAmount.tsx [hunk 24]
```diff
@@ -97,23 +95,25 @@ export const FormattedCryptoAmount = ({
...
                     <BlurUrls text={formattedSymbol} />
                 </>
             )}
-        </Container>
+        </Row>
     );
 
     if (disableHiddenPlaceholder) {
-        return content;
+        return <>{content}</>;
```
jvaclavik: Why do you need it?

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
-import styled from 'styled-components';
+import { useTheme } from 'styled-components';
 
 import { SignValue } from '@suite-common/suite-types';
-import { getNftTokenId } from '@suite-common/wallet-utils';
+import { getNftTokenId, isNftMultitokenTransfer } from '@suite-common/wallet-utils';
 import { TokenTransfer } from '@trezor/connect';
-import { variables } from '@trezor/components';
+import { Column, Flex, Row, Text } from '@trezor/components';
```
jvaclavik: Please do not use `Flex`, but `Row` or `Column` 🙏

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
...
-
-const Id = styled.div`
-    white-space: nowrap;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    max-width: 145px;
-`;
+export const Id = ({ id }: { id?: string }) => {
+    return id ? (
+        <Text maxWidth="145px" ellipsisLineCount={1} textWrap="nowrap">
```
jvaclavik: nitpick: it's just my personal preference but if we use px we can use number.

Like:
```
maxWidth={145}
```

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
...
+                                        })}
+                                    />
+                                ) : (
+                                    <>
+                                        {token.value}x&nbsp;
+                                        <Translation id="TR_TOKEN_ID" />:
+                                    </>
+                                )}
+                            </Row>
+                            <Row>
```
jvaclavik: Do you need Row here? There is only one element

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipAccount.tsx
```diff
@@ -26,14 +22,16 @@ const formatAmount = (
...
-        <>
+        <Row>
             {amount && (
-                <StyledCryptoAmount
-                    value={amount}
-                    symbol={symbol}
-                    signValue={sign}
-                    disableHiddenPlaceholder
-                />
+                <Row margin={{ right: 4 }}>
```
jvaclavik: please use `spacings`

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
...
+                        typographyStyle="label"
+                        gap="zero"
+                    >
+                        <Id id={id} />
+                        &nbsp; {symbolComponent}
+                    </TrezorLink>
                 ) : (
-                    <NoLink>
-                        {idComponent}
+                    <Row gap={4}>
```
jvaclavik: please use `spacings.xxs`

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipBase.tsx
jvaclavik: I suggest to use `Row` component from DS here

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
```diff
@@ -65,10 +65,13 @@ const IOItem = ({ anonymitySet, address, symbol, amount, isPhishingTransaction }
                     {anonymity && <UtxoAnonymity anonymity={anonymity} />}
                     {amount &&
                         (typeof amount === 'string' && symbol ? (
-                            <FormattedCryptoAmount
-                                value={formatNetworkAmount(amount, symbol)}
-                                symbol={symbol}
-                            />
+                            <Column>
+                                <br />
```
jvaclavik: why not to use margin top or something else instead of `<br />`?

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
```diff
@@ -211,18 +214,37 @@ const EthereumSpecificBalanceDetailsRow = ({
...
+                                    ] as IODetails[]
+                                }
+                                outputs={
+                                    [
+                                        {
+                                            addresses: [transfer.to],
+                                            value: value as ReactNode,
+                                        },
+                                    ] as IODetails[]
+                                }
```
jvaclavik: nitpick: I would suggest extract these arrays to consts above. It's easier to read

packages/suite/src/components/wallet/AmountComponent.tsx
```diff
@@ -0,0 +1,46 @@
+import { isNftTokenTransfer, formatAmount, getTxOperation } from '@suite-common/wallet-utils';
+import { TokenTransfer } from '@trezor/connect';
+
+import { FormattedNftAmount } from 'src/components/suite/FormattedNftAmount';
+import { FormattedCryptoAmount } from 'src/components/suite/FormattedCryptoAmount';
+
+type AmountComponentProps = {
+    transfer: TokenTransfer;
+    useStyled?: boolean;
```
jvaclavik: What does this mean? I don't understand the naming and think you don't use it at all.

jvaclavik: And the same for contract - it's also not used

packages/suite/src/components/wallet/AmountComponent.tsx
```diff
@@ -0,0 +1,46 @@
...
+        return (
+            <FormattedCryptoAmount
+                value={formatAmount(transfer.amount, transfer.decimals)}
+                symbol={transfer.symbol}
+                signValue={operation}
+            />
+        );
+    }
+
+    return <>{formatAmount(transfer.amount, transfer.decimals)}</>;
```
jvaclavik: Can you do it without the fragment?

packages/suite/src/components/wallet/TransactionItem/CoinjoinBatchItem.tsx
```diff
@@ -141,11 +133,11 @@ export const CoinjoinBatchItem = ({
                                 signValue={amount}
                             />
                         </Description>
-                        <NextRow>
+                        <Row flex="1" alignItems="flex-start" margin={{ bottom: spacings.xxs }}>
                             <TimestampWrapper>
                                 <TransactionTimestamp transaction={lastTx} />
                             </TimestampWrapper>
-                            <TargetsWrapper>
+                            <Column flex="1" overflow="hidden">
```
jvaclavik: Please use `Box` component here

packages/suite/src/components/wallet/TransactionItem/TransactionHeading.tsx
```diff
@@ -168,12 +163,12 @@ export const TransactionHeading = ({
                         />
                     )}
                     <BlurWrapper $isBlurred={isPhishingTransaction}>
-                        <HeaderWrapper>
+                        <Row gap={spacings.xxs}>
```
jvaclavik, line 166: Please use `Box` here. Row is considered for multiple children, and Box is universal building block for cases like this one.

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
```diff
@@ -244,7 +233,7 @@ export const TransactionItem = memo(
                 $isPhishingTransaction={isPhishingTransaction}
                 className={className}
             >
-                <Body>
+                <Flex width="100%">
```
jvaclavik: Please use `Row` component

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
```diff
@@ -254,7 +243,7 @@ export const TransactionItem = memo(
                     </TxTypeIconWrapper>
 
                     <Content>
```
jvaclavik: It would be nice to separate some of the components in this return to make it more readable. But I understand it was here before, so it's up to you.

packages/suite/src/components/wallet/TransactionItem/TransactionItemBlurWrapper.tsx
```diff
@@ -1,6 +1,7 @@
 import styled, { css } from 'styled-components';
 
 export const BlurWrapper = styled.span<{ $isBlurred: boolean }>`
+    width: 100%;
```
jvaclavik: does it do something with non-block element like `span`?

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
enjojoy: I will leave it as is for now, I need this PR to be merged to work on NFT section that is higher priority

packages/components/src/components/typography/Link/Link.tsx
```diff
@@ -22,17 +22,19 @@ type AllowedLinkTextProps = Pick<TextPropsCommon, (typeof allowedLinkTextProps)[
...
     text-decoration: none;
     cursor: pointer;
-    color: ${({ theme }) => theme.textDefault};
+    color: ${({ $color, theme }) => $color || theme.textDefault};
     font-weight: 500;
     display: inline-flex;
     align-items: center;
 
-    gap: ${spacingsPx.xxs};
+    gap: ${({ $gap }) => ($gap ? spacingsPx[$gap] : spacingsPx.xxs)};
```
seibei-iguchi: I think default should be set on the gap prop, not in the css, and it should be typed to SpacingValues.

packages/suite/src/components/suite/FormattedNftAmount.tsx
seibei-iguchi: Use number for maxWidth

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
...
     const { network } = selectedAccount;
     const explorerUrl =
         network?.networkType === 'ethereum'
             ? `${network?.explorer.nft}/${transfer.contract}/${id}`
             : undefined;
 
-    const idComponent = <Id>{id}</Id>;
-    const symbolComponent = transfer.symbol ? <Symbol>&nbsp;{transfer.symbol}</Symbol> : null;
+    const symbolComponent = transfer.symbol ? (
+        <Text maxWidth="120px" ellipsisLineCount={1} textWrap="nowrap">
```
seibei-iguchi: Use number for maxWidth. The nowrap is, I think, redundant, it is already set by the ellipsis prop

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
...
+    const isMultitoken = isNftMultitokenTransfer(transfer);
+
+    if (isMultitoken) {
+        const tokens = transfer.multiTokenValues;
+
+        return (
+            <HiddenPlaceholder>
+                <Column alignItems={alignMultitoken}>
+                    {tokens?.map((token, index) => (
+                        <Row key={`${token.id}-${index}`} gap={4}>
```
seibei-iguchi: Use spacings for gap

packages/suite/src/components/suite/FormattedNftAmount.tsx
seibei-iguchi: Use spacings

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipAccount.tsx
seibei-iguchi: Use spacings

seibei-iguchi (changes requested):

packages/suite/src/support/messages.ts
```diff
@@ -5696,6 +5700,10 @@ const messages = defineMessagesWithTypeCheck({
         defaultMessage: '{amount} selected',
         description: 'Number of list items selected',
     },
+    TR_COLLECTION_NAME_OF_TOKEN_ID: {
+        id: 'TR_COLLECTION_NAME_OF_TOKEN_ID',
+        defaultMessage: '{tokenValue}x {collectionName} of Token ID: ',
```
tomasklim: Can we add also id here?

packages/suite/src/support/messages.ts
```diff
@@ -3314,6 +3314,10 @@ const messages = defineMessagesWithTypeCheck({
         defaultMessage: 'Details',
         id: 'TR_TRANSACTION_DETAILS',
     },
+    TR_TOKEN_ID: {
+        defaultMessage: 'Token ID',
```
tomasklim: the `TR_COLLECTION_NAME_OF_TOKEN_ID` contains `:`. Can we unify it please 🙏

packages/suite/src/components/wallet/TransactionItem/TransactionItemBlurWrapper.tsx
tomasklim: line below there is `display: block;`

suite-common/wallet-utils/src/transactionUtils.ts
tomasklim: I have no idea but we should be ready for weird cases :D

packages/suite/src/components/wallet/TransactionItem/TransactionHeader.tsx
tomasklim: ah oki

suite-common/wallet-utils/src/transactionUtils.ts
enjojoy: I added a check for length there just in case, I don't think it's possible, maybe only in a very weird nonsense contract

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: In this case it has to be there, to preserve the underline between these components, otherwise it's being removed and there's a weird gap that is not underlined, even tho the link is the same

I didn't come up with a less hacky solution 😅

packages/components/src/components/typography/Link/Link.tsx
enjojoy: Again it's here because I must remove the gap and add a space to preserve the underline

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
enjojoy: No need, it's removed now

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: Used

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: Done

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: Removed

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipAccount.tsx
enjojoy: Used

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: Used

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipBase.tsx
enjojoy: used

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
enjojoy: Sure, used a margin

packages/suite/src/components/wallet/AmountComponent.tsx
enjojoy: Yes, it was forgotten there, will remove it now, thanks

packages/suite/src/components/wallet/AmountComponent.tsx
enjojoy: removed

packages/suite/src/components/wallet/TransactionItem/CoinjoinBatchItem.tsx
enjojoy: used

packages/suite/src/components/wallet/TransactionItem/TransactionHeading.tsx
enjojoy: In this case there can be more children depending on a condition, but I'll note that for other cases

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
enjojoy: Used

packages/suite/src/components/wallet/TransactionItem/TransactionItemBlurWrapper.tsx
enjojoy: Yes, I added that

packages/components/src/components/typography/Link/Link.tsx
enjojoy: Thanks, did this

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: used

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: yes, you are right, fixed this, thank you

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: done

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: done

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipAccount.tsx
enjojoy: done

packages/suite/src/support/messages.ts
enjojoy: No, because I need this to be a string to feed it to BlurUrls  that only accepts strings, and ID must be wrapped into a TrezorLink component, so here I won't use it. I think it's ok like that

packages/suite/src/support/messages.ts
enjojoy: yes, unified

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
```diff
@@ -97,23 +95,25 @@ export const FormattedCryptoAmount = ({
     }
 
     const content = (
-        <Container className={className}>
-            {!!signValue && <Sign value={signValue} />}
-            <Value data-testid={dataTest}>
-                <RedactNumericalValue value={formattedValue} />
-            </Value>
+        <Row width="100%" gap={spacings.xxs} className={className}>
```
tomasklim, line 98: I think it broke this 
Screenshot 2024-11-29 at 17 15 01

packages/suite/src/components/wallet/AmountComponent.tsx
```diff
@@ -0,0 +1,44 @@
...
+    if (withSign) {
+        return (
+            <FormattedCryptoAmount
+                value={formatAmount(transfer.amount, transfer.decimals)}
+                symbol={transfer.symbol}
+                signValue={operation}
+            />
+        );
+    }
+
```
tomasklim, line 46: It misses symbol without sign, is it intended? 

Screenshot 2024-11-29 at 17 20 57
There should be USDC

Screenshot 2024-11-29 at 17 21 13

packages/suite/src/support/messages.ts
tomasklim: Okay, then please remove the space after Token ID: for both strings. I am scared that the translation agency would remove it anyway and it would be broken in some languages

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
```diff
@@ -65,10 +65,12 @@ const IOItem = ({ anonymitySet, address, symbol, amount, isPhishingTransaction }
                     {anonymity && <UtxoAnonymity anonymity={anonymity} />}
                     {amount &&
                         (typeof amount === 'string' && symbol ? (
-                            <FormattedCryptoAmount
-                                value={formatNetworkAmount(amount, symbol)}
-                                symbol={symbol}
-                            />
+                            <Column margin={{ top: spacings.md }}>
```
seibei-iguchi: Why the margin?

enjojoy: Oh I think there was a break before, but I guess with the resent changes it's not there anymore, will remove

enjojoy: removed

packages/suite/src/components/wallet/AmountComponent.tsx
```diff
@@ -0,0 +1,44 @@
...
+    transfer,
+    withLink = false,
+    withSign = false,
+    alignMultitoken,
+}: AmountComponentProps): React.ReactNode => {
+    const operation = getTxOperation(transfer.type);
+
+    if (isNftTokenTransfer(transfer)) {
+        return (
+            <FormattedNftAmount
```
tomasklim, line 27: I am not sure how the hidden placeholder works but for normal amounts, it replaces the amount by XXX, for nfts, it does not do that. Probably it was there before so just create issue in case it is hard to fix
Screenshot 2024-11-29 at 17 35 08

packages/suite/src/components/wallet/AmountComponent.tsx
```diff
@@ -0,0 +1,44 @@
...
+    alignMultitoken?: 'flex-end' | 'flex-start';
+};
+
+export const AmountComponent = ({
+    transfer,
+    withLink = false,
+    withSign = false,
+    alignMultitoken,
+}: AmountComponentProps): React.ReactNode => {
+    const operation = getTxOperation(transfer.type);
```
tomasklim, line 23: this can be inside the `if (withSign) {` condition

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
...
+                        }
+                        color={theme.textSecondaryHighlight}
+                        variant="underline"
+                        typographyStyle="label"
+                        gap={spacings.zero}
+                    >
+                        <Text maxWidth={145} ellipsisLineCount={1}>
+                            {transfer.amount}
+                        </Text>
+                        &nbsp; {symbolComponent}
```
tomasklim: does it still need the space between `&nbsp;` `{symbolComponent}`?

packages/suite/src/components/suite/FormattedNftAmount.tsx
```diff
@@ -1,89 +1,127 @@
...
+                                        <Translation id="TR_TOKEN_ID" />
+                                    </Row>
+                                )}
+                            </Row>
+                            {isWithLink && network?.networkType === 'ethereum' ? (
+                                <TrezorLink
+                                    href={`${network?.explorer.nft}/${transfer.contract}/${token.id}`}
+                                    color={theme.textSecondaryHighlight}
+                                    variant="underline"
+                                    typographyStyle="label"
```
tomasklim: It now does not have same font size in review tx modal 
Screenshot 2024-11-29 at 17 48 11

tomasklim (changes requested):
Just minted beautiful nft for qa on eth n.1 😀 for testing

Screenshot 2024-11-29 at 17 49 35

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
enjojoy: How? I can't duplicate this. Do you have the recent changed pulled?

packages/suite/src/components/wallet/AmountComponent.tsx
enjojoy: It's not intended, I will rather bring back the way that component worked before

packages/suite/src/components/wallet/AmountComponent.tsx
enjojoy: I will create an issue, now there's only a method to redact a raw string or numebr, but we have a component because of BlurUrls in the name

enjojoy: Issue created

packages/suite/src/components/wallet/AmountComponent.tsx
enjojoy: No because it's also used before the condition in if(isNftTokenTransfer(transfer))

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
tomasklim: Yes, I am always doing review on latest changes 😀 

https://dev.suite.sldev.cz/suite-web/feat/display-multitoken-tx/web/accounts#/eth/0/normal

qa seed, eth 1, 22th november, 1470px width
Screenshot 2024-12-01 at 15 15 57

packages/suite/src/components/wallet/TransactionItem/TransactionItemBlurWrapper.tsx
tomasklim: What does it do?

packages/suite/src/components/wallet/TransactionItem/TransactionTarget/TransactionTarget.tsx
```diff
@@ -68,15 +70,12 @@ export const TokenTransfer = ({
...
-                isNft ? (
-                    <StyledFormattedNftAmount transfer={transfer} signValue={operation} />
-                ) : (
-                    <StyledFormattedCryptoAmount
-                        value={formatAmount(transfer.amount, transfer.decimals)}
-                        symbol={transfer.symbol}
-                        signValue={operation}
-                    />
-                )
+                <AmountComponent
```
tomasklim, line 73: Develop:
Screenshot 2024-12-01 at 15 53 01

Your branch:
Screenshot 2024-12-01 at 15 46 16

Check how the address behaves. There was `...` before

packages/suite/src/components/suite/FormattedNftAmount.tsx
tomasklim: I think not needed

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
enjojoy: Ok, very interesting, I don't think I've touched that part of the code. Anyway fixed https://github.com/trezor/trezor-suite/pull/15457/commits/72b21d1332ddd817867d1601b8ac47216654e08e

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: It is needed, without it there will be a space in the underline of the link, as the gap is not considered a normal space and breaks the underline

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: It's not the same on develop either

image

packages/suite/src/components/suite/FormattedNftAmount.tsx
tomasklim: I am talking about 
```
&nbsp;`HERE`{symbolComponent}
```

enjojoy: Oh yean, that is not needed indeed

packages/suite/src/components/suite/FormattedCryptoAmount.tsx
tomasklim: You modified `FormattedCryptoAmount` which is used in the DayHeader so you basically touched it

enjojoy: True

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: I also noticed that the color of amounts in the tx modal on develop is different, will adjust that

packages/components/src/components/typography/Link/Link.tsx
enjojoy: I realized I can do the same thing without modifying gap in the link component
  image

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/IODetails/IODetails.tsx
enjojoy: I actually returned it to the way it was before

packages/suite/src/components/wallet/TransactionItem/TransactionItemBlurWrapper.tsx
enjojoy: now seems like it does nothing, not sure why I added it before as well as width, seems to be ok without it so i removed it

packages/suite/src/components/wallet/TransactionItem/TransactionTarget/TransactionTarget.tsx
enjojoy: This should be fixed as I removed the added styles on BlurWrapper

packages/suite/src/components/suite/FormattedNftAmount.tsx
enjojoy: > I realized I can do the same thing without modifying gap in the link component image

re

seibei-iguchi (approved):

tomasklim (approved):

PR #15460, opened 2024-11-19
feat(suite-common): fetch stake data based on device account symbols

### LINKED ISSUE #15362: Staking APY doesn't reflect reality
**Describe the bug**
The presented APY doesn't reflect what is offered by [Everstake](https://eth-api-b2c-stage.everstake.one/api/v1/stats)

**Info:**

-   Mobile Suite Version: 24.11.1 (746) 35433e7
-   OS: Android

**How to reproduce**
1. Add staking
2. Check APY
3. See 4.13%

**Expected behavior**
See value from [Everstake](https://eth-api-b2c-stage.everstake.one/api/v1/stats)

**Screenshots**
image

vytick:
On mobile, using enabled networks might not work for imported accounts (portfolio tracker), because `ETH` might not be amongst enabled networks. On iOS, enabled networks are always empty array. This caused imported accounts with eth staking to show there `BACKUP_ETH_APY` instead of fetching actual apy. 

This is now resolved by fetching apy based on network symbols of actual device accounts.

## Related Issue

Resolve #15362 for portfolio tracker

### DIFF
--- a/suite-common/wallet-core/src/stake/stakeThunks.ts
+++ b/suite-common/wallet-core/src/stake/stakeThunks.ts
@@ -7,6 +7,7 @@ import { SupportedNetworkSymbol } from '@suite-common/wallet-types';
 import { selectEverstakeData } from './stakeSelectors';
 import { EVERSTAKE_ENDPOINT_TYPES, EverstakeEndpointType, ValidatorsQueue } from './stakeTypes';
 import { EVERSTAKE_ENDPOINT_PREFIX } from './stakeConstants';
+import { selectAllNetworkSymbolsOfVisibleAccounts } from '../accounts/accountsReducer';
 
 const STAKE_MODULE = '@common/wallet-core/stake';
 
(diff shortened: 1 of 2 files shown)


Nodonisko (approved):
lgtm but would be nice if @tomasklim could take a look, because it could affect desktop

suite-common/wallet-core/src/stake/stakeThunks.ts
```diff
@@ -59,12 +60,12 @@ export const fetchEverstakeData = createThunk<
 
 export const initStakeDataThunk = createThunk(
     `${STAKE_MODULE}/initStakeDataThunk`,
-    (_, { getState, dispatch, extra }) => {
-        const enabledNetworks = extra.selectors.selectEnabledNetworks(getState());
+    (_, { getState, dispatch }) => {
+        const networks = selectDeviceNetworkSymbolsOfVisibleAccounts(getState());
```
Nodonisko: I not sure if this is good solution. On desktop this relies on enabled networks instead of accounts, so if you will have blank desktop app, connect device, enable eth it will take another 2 minutes to fetch stake data because this thunk will not run again when new account is added.

On mobile it's not issue because mobile initiate this fetch when user enters staking detail screen.

vytick: 👍  You are right. I didnt realize that desktop does not refetch it when entering staking. I'll think about it once more. `initStakeDataThunk` is run every minute, so if user enters staking within that 1 minute, it might have wrong apy.

vytick: a108320107

tomasklim (approved):

PR #15465, opened 2024-11-20
feat(suite): display modals in WelcomeLayout

szymonlesisz:
## Description

Prerequisite for bluetooth and thp.

we need to open modals in WelcomeLayout (scan result, pairing process etc)

`WelcomeLayout` component is removed from `scr/components/suite/index` as it causes some circular dependency errors


komret (approved):
There are no modals on the welcome screen in the [design](https://miro.com/app/board/uXjVLQc7_NE=/), I think we can just use `Card`, but we can use this temporarily if it helps managing the branches.

peter-sanderson (changes requested):
I am right about to start implementing the design and this modal wont be part of it. So I would have to revert this anyway so we need to merge it.

Issue #15087, opened 2024-10-26
Optimise fee fetching and updating


tomasklim:
**Description**

In our current solution the fees:
- are updated for all enabled networks even though user does not use them
	- -> useless requests
- are updated every 10 blocks. For POL that's 20 seconds, for ARB it is 3 seconds, for ETH it is 120 seconds, for BTC it is 100 minutes!
    - -> the fees are outdated

**Expected behavior**

Let's change the behaviour, that fees will not be updated periodically for all networks but only if:
- user enters send form
- user enters stake/unstake/claim
- user enters trading

**Notes**

For conditions when fee should be updated now, see `updateFeeInfoThunk`

**Question**: Does it make sense to store it in redux?

The updating redux action is:
`@common/wallet-core/blockchain/updateFee`
![Screenshot 2024-10-26 at 16 48 35](https://github.com/user-attachments/assets/f0ba3626-1863-4409-90ba-d2cc987250a2)

Fees are stored in redux:
![Screenshot 2024-10-26 at 16 46 46](https://github.com/user-attachments/assets/6749ad66-21be-4ef6-81bd-c1f93d0a6032)

**Question**: Should we allow user to update fees when he is in send form for long time?

tomasklim:
POC https://github.com/trezor/trezor-suite/pull/15469

Ondra-Zik:
**QA OK**

Fees are now being updated in the send forms for selected network

**Info:**
 - Suite version: web 25.4.0 (19c595233b7801b22f455a9cbb15e0051763f7f6)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 3440x1440
 - Device: Trezor T3T1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)
 - Transport: WebUsbTransport

PR #15469, opened 2024-11-20
refactor(suite): optimize how often fees are fetched in send form

senky:
**!!! WORK IN PROGRESS !!! do not merge yet**

This commit introduces a new approach to fetching fees.

**Previous state:**
Fees for every wallet account were fetched periodically after 10 blocks. This was problematic for various reasons:
1. The data weren't often needed.
2. Useless network requests.
3. Some fees were outdated, because fees aren't connected to blocks in any way - in fact, fee can update multiple times in between two blocks!

**New state:**
This solution mainly builds on the idea that fees aren't connected to blocks. This link is now removed. There is a dedicated `useFees` hook that is supposed to be called for every route that needs to display fees. The hook is responsible for decision on how often the fee should be fetched (decided manually in `networkToInterval` function). That way all logic behind fee fetching is located in one file. Consequence of this decision is that `updateFeeInfoThunk` is now "dumb" - it fetches new fees whenever it is called. While this solution is more verbose (`useFees` has to be called everywhere fees are needed), I think its explicit relationship improves readability of the code.

**What needs to be done before this commit is mergeable:**
1. Set network refresh intervals in `networkToInterval` function.
2. Call `useFees` for every route that needs fees (swap, sell, etc.).
3. Move `useFees` hook to better location so that it can be used in `suite-native/module-send/src/screens/SendOutputsScreen.tsx:174`, too.

Relates to #15087

### DIFF
--- a/packages/suite/src/hooks/wallet/useFees.ts
+++ b/packages/suite/src/hooks/wallet/useFees.ts
@@ -1,0 +1,33 @@
+import { useDispatch } from 'react-redux';
+import { useCallback, useEffect } from 'react';
+
+import { NetworkSymbol } from "@suite-common/wallet-config";
+import {
+  updateFeeInfoThunk,
+} from '@suite-common/wallet-core';
+
+const networkToInterval = (symbol: NetworkSymbol): number => {
+  switch (symbol) {
+    case 'pol':
+      return 1000;
+    default:
+      return 10000;
+  }
+}
+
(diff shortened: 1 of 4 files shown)


tomasklim:
Thank you @senky for your contribution. The POC works really great, love it.

https://github.com/user-attachments/assets/76097439-440c-4f65-9295-b5aeb3d8b170

PR #15470, opened 2024-11-20
chore(suite): refactor transaction details modal

seibei-iguchi:
- refactored transaction details modal
- renamed InfoRow component to InfoItem, and added vertical alignment and label width props
- added InfoPair component
- added hasBorders prop to Table component

### Before
Screenshot 2024-11-21 at 14 16 25
Screenshot 2024-11-21 at 14 16 51
Screenshot 2024-11-21 at 14 17 03
Screenshot 2024-11-21 at 14 17 22
Screenshot 2024-11-21 at 14 17 32
Screenshot 2024-11-21 at 14 18 25

### After
Screenshot 2024-11-21 at 14 14 31
Screenshot 2024-11-21 at 14 14 41
Screenshot 2024-11-21 at 14 14 54
Screenshot 2024-11-21 at 14 15 12
Screenshot 2024-11-21 at 14 15 20
Screenshot 2024-11-21 at 14 15 30


packages/components/src/components/Grid/Grid.tsx
```diff
@@ -16,7 +16,8 @@ const Container = styled.div<
 >`
     display: grid;
     gap: ${({ $gap }) => $gap}px;
-    grid-template-columns: repeat(${({ $columns }) => $columns}, 1fr);
+    grid-template-columns: repeat(${({ $columns }) => $columns}, minmax(0, 1fr));
```
Lemonexe, line 19: Having no knowledge about this, I asked chatgpt about it: before, an overflowing cell would expand, but now, it will be strictly limited to 1 fractional unit. Does that sound about right?

packages/suite/src/components/suite/StakingProcess/UnstakingInfo.tsx
```diff
@@ -34,7 +34,7 @@ export const UnstakingInfo = ({ isExpanded }: UnstakingInfoProps) => {
     const daysToUnstake = getDaysToUnstake(unstakeTxs, data);
     const accountSymbol = account.symbol.toUpperCase();
 
-    const infoRows = [
+    const InfoRows = [
```
Lemonexe: I believe camelCase was correct? As far as I know, we use PascalCase for component, class or types, but this is neither.
Do we have another naming convention I am not familiar with?

packages/components/src/components/NewModal/NewModalButton.tsx
```diff
@@ -7,7 +7,12 @@ export const NewModalButton = ({ children, ...rest }: ButtonProps) => {
 
     return (
         <NewModalContext.Provider value={value}>
-            <Button {...rest} variant={rest.variant ?? variant} size={rest.size ?? 'large'}>
+            <Button
+                {...rest}
+                variant={rest.variant ?? variant}
+                size={rest.size ?? 'large'}
+                minWidth={150}
```
Lemonexe, line 14: This feels rather arbitrary. I am curious what is the use-case? When is the button too small so it has to have `minWidth`?

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -48,15 +57,23 @@ const Cell = styled.td<{ $isHeader: boolean; $elevation: Elevation; $padding?: P
...
+                ${$padding.left ?? $padding.horizontal ?? 0}px;
         `}
 
     &:first-child {
         position: sticky;
         left: 0;
         z-index: 2;
         background: linear-gradient(to right, ${mapElevationToBackground} 90%, rgba(0 0 0 / 0%));
+
+        ${({ $hasBorder }) => !$hasBorder && 'padding-left: 0;'}
```
Lemonexe, line 72: Nit: I'd prefer ternary that fallbacks to undefined rather than short circuit.
But I am not that familiar with `styled` - if `false` is recognized as ignored value, then it's fine :ok_hand: 
Same with other cases

packages/components/src/components/Table/TableRow.tsx
```diff
@@ -86,19 +86,19 @@ export const TableRow = ({
...
+    const { isRowHighlightedOnHover, hasBorders } = useTable();
 
     return (
         <Row
             $elevation={elevation}
             $isCollapsed={isCollapsed}
             $isHighlighted={isHighlightedOnHover ?? isRowHighlightedOnHover}
             $isHeader={isHeader}
-            $hasBorderTop={hasBorderTop}
+            $hasBorderTop={hasBorderTop ?? hasBorders}
```
Lemonexe, line 101: cool, thanks to `??` you can override Table hasBorders with TableRow hasBorderTop=false :+1:

packages/suite/src/components/suite/StakingProcess/StakingInfo.tsx
```diff
@@ -38,7 +38,7 @@ export const StakingInfo = ({ isExpanded }: StakingInfoProps) => {
 
     const daysToAddToPool = getDaysToAddToPool(stakeTxs, data);
 
-    const infoRows = [
+    const InfoItems = [
```
Lemonexe: same as in [comment above](https://github.com/trezor/trezor-suite/pull/15470#discussion_r1852143809)

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -487,17 +487,17 @@ export const getNftTokenId = (transfer: TokenTransfer) =>
...
-            return 'send';
+            return 'arrowUp';
         case 'failed':
-            return 'close';
+            return 'x';
         case 'joint':
             return 'shuffle';
         default:
-            return 'question';
+            return 'questionSimple';
```
Lemonexe, line 500: I didn't know about this function and I don't like it :smile:  But this PR is fine, you did not introduce it, so it's ok to just update it :heavy_check_mark: 

:thought_balloon:  Just noting for future: this fn could have a better return type - it returns type `string`, so a mistake would break the icon without TS knowing :confused: 
I am aware you can't type the return type as icon type, because `suite-common/wallet-utils` would have to import `trezor/components`, and I guess that's undesirable. But it's only used in `suite`, so the fn could live there.. Idk, just a suggestion :man_shrugging:

packages/suite/src/components/suite/FormattedDateWithBullet.tsx
```diff
@@ -1,31 +1,10 @@
...
-        <Bullet>&bull;</Bullet>
-        <HourWrapper $timeLightColor={props.timeLightColor}>
-            <FormattedDate time {...props} />
-        </HourWrapper>
-    </Timestamp>
+export const FormattedDateWithBullet = ({ ...props }: FormattedDateProps) => (
+    <InfoPair
+        leftContent={<FormattedDate date {...props} />}
+        rightContent={<FormattedDate time {...props} />}
+    />
```
Lemonexe, line 9: InfoPair is a nice component :+1: :+1:

packages/components/src/components/InfoPair/InfoPair.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const allowedInfoPairFrameProps = ['margin'] as const satisfies FramePropsKeys[];
+type AllowedFrameProps = Pick<FrameProps, (typeof allowedInfoPairFrameProps)[number]>;
+
+export type InfoPairProps = AllowedFrameProps &
+    AllowedTextProps & {
+        variant?: TextVariant;
+    } & (
+        | { leftContent: React.ReactNode; rightContent?: React.ReactNode }
+        | { leftContent?: React.ReactNode; rightContent: React.ReactNode }
```
Lemonexe: Tiny nit: explicit React import instead of implicit umd import :pray: 

But I love the type definition with union, forcing you to define at least one :+1:

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/ChangeFee/ReplaceTxButton.tsx
```diff
@@ -1,40 +1,34 @@
...
-const StyledButton = styled(Button)`
-    min-width: 30%;
-`;
+type ReplaceTxButtonProps = {
+    rbfParams: RbfTransactionParams;
+    selectedAccount: SelectedAccountLoaded;
+};
 
-export const ReplaceTxButton = () => {
+export const ReplaceTxButton = ({ rbfParams, selectedAccount }: ReplaceTxButtonProps) => {
```
Lemonexe, line 13: Is there a reason to pass `{ rbfParams, selectedAccount }` instead of using `useRbfContext`, which achieves the same without having to pass props?

I'm honestly asking, I don't understand details what's going on here. But the former solution looks simple & appropriate.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AdvancedTxDetails/AdvancedTxDetails.tsx
```diff
@@ -73,33 +53,38 @@ export const AdvancedTxDetails = ({
...
+                            network={network}
+                        />
+                    )
+                );
+            case 'data':
+                return <Data tx={tx} />;
+            default:
+                return null;
+        }
+    };
```
Lemonexe, line 80: Nice to get rid of the `let` :+1:

Lemonexe (approved):
LGTM :rocket: 

Codewise looks great - much better than before, with design system components instead of local css :heart: 
:memo:  Ofc due to the sheer size of the PR, there are many small comments

Tested in app :heavy_check_mark:  though not all possible cases, leaving that to QA.

However the PR size is really too much to review reliably :disappointed:  Splitting PRs is best, though not necessary - at lest splitting commits next time would help immensely :pray:

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/ChangeFee/ReplaceTxButton.tsx
seibei-iguchi: Unfortunately I can't use the context, since the button has been taken out of the nested content (where the context is available) and passed to NewModal bottom content instead.

packages/components/src/components/Grid/Grid.tsx
seibei-iguchi: Yes, that's the intention. I want the Grid component to be as predictable as possible.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/ChangeFee/ReplaceTxButton.tsx
seibei-iguchi: The other option would be to turn the ChangeFees component into a hook that would return the original component and the button from inside, but that seems like too much work. What do you think?

Lemonexe: Now that I look at it again, it's nice and clear. At first glance I saw the diff and was questioning it. Now I see that both `ChangeFee` and `ReplaceTxButton` are called from `TxDetailModal`, which passes props to both. That's normal day in React, nice and clear :+1: :ok_hand:

Trying to "unify passing props" to both components by introducing that hook would spaghettify it, so I'd be against it.

packages/components/src/components/Table/TableCell.tsx
```diff
@@ -19,6 +21,8 @@ type Padding = {
     bottom?: SpacingValues;
     left?: SpacingValues;
     right?: SpacingValues;
+    vertical?: SpacingValues;
+    horizontal?: SpacingValues;
```
jvaclavik, line 25: I like this one. 

Btw IMHO we should use it also for `margin` in `frameProps`. And maybe also adding this one:
`padding={spacings.sm}` (and the same for `margin`)

packages/suite/src/components/suite/FormattedDateWithBullet.tsx
jvaclavik: Hell yes! 🎉

packages/suite/src/components/suite/StakingProcess/StakingInfo.tsx
jvaclavik: I'm also confused here

jvaclavik (approved):
Really nice refactoring 👏  LGTM!

packages/components/src/components/Table/TableCell.tsx
seibei-iguchi: I've already added it to margin in frameProps:)

packages/suite/src/components/suite/StakingProcess/StakingInfo.tsx
seibei-iguchi: This is a result of global replace going wrong:D

PR #15473, opened 2024-11-20
fix(connect): preferred device handling based on state

### LINKED ISSUE #15452: Wallet and passphrase is not saved with Rabby
Wallet and passphrase is not saved when using Rabby, despite "Remember device" being selected. 

Screenshot 2024-11-19 at 15 48 24

Observed both with NodeBridge and WebUSB on 9.4.3 before rollback.

Bug is not present on 9.4.2

cc. @martykan @mroz22

martykan:
## Description

- Find device by ID from state
- Move all`preferredDevice` related logic to initDevice
- Remove `devicePath` from AbstractMethod

## Related Issue

Resolve #15452

### DIFF
--- a/packages/connect-common/src/storage.ts
+++ b/packages/connect-common/src/storage.ts
@@ -18,7 +18,7 @@ export interface Permission {
 export interface PreferredDevice {
     label?: string;
     path: string & { __type: 'DeviceUniquePath' };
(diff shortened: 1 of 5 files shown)


marekrjpolak (approved):
LGTM

PR #15474, opened 2024-11-20
Unify send form input logic

### LINKED ISSUE #15227: Suite doesn't calculate crypto value correctly in send form after deleting and writing again the fiat input
**Describe the bug**
Suite doesn't calculate crypto value correctly in send form after deleting and writing again the fiat input. It works best with only single digit number but I managed to replicate it sometimes with even 2 digits.

**Steps to reproduce:**
1. Navigate to send form of any coin
2. Write any fiat value and delete it
3. Write a single digit fiat value again
4. The crypto value isn't calculated at all or incorrectly

**Info:**
 - Suite version: web 24.11.0 (b7afbd05152736013d619dfd9dc9f10143fbd216)
 - OS: Linux x86_64
 - Screen: 1920x1080

**Screenshots:**

https://github.com/user-attachments/assets/9c454141-f11f-4a27-ad6e-f4c62ce9bc71

komret:
Resolve https://github.com/trezor/trezor-suite/issues/15227

### DIFF
--- a/packages/suite/src/hooks/wallet/useSendForm.ts
+++ b/packages/suite/src/hooks/wallet/useSendForm.ts
@@ -37,6 +37,7 @@ import { useFees } from './form/useFees';
 import { useBitcoinAmountUnit } from './useBitcoinAmountUnit';
 import { useUtxoSelection } from './form/useUtxoSelection';
 import { useExcludedUtxos } from './form/useExcludedUtxos';
+import { useSendFormChangeHandlers } from './useSendFormChangeHandlers';
 
 export const SendContext = createContext<SendContextValues | null>(null);
 SendContext.displayName = 'SendContext';
@@ -179,6 +180,13 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
         updateContext,
         setLoading,
         setAmount: sendFormUtils.setAmount,
+    });
+
+    const changeHandlers = useSendFormChangeHandlers({
+        calculateAmountFromFiat: sendFormUtils.calculateAmountFromFiat,
+        calculateFiatFromAmount: sendFormUtils.calculateFiatFromAmount,
+        composeRequest,
+        setValue,
     });
 
     // declare useSendFormOutputs, sub-hook of useSendForm
@@ -388,6 +396,7 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
         utxoSelection,
         ...sendFormUtils,
         ...sendFormOutputs,
+        ...changeHandlers,
     };
 };
 
--- a/packages/suite/src/hooks/wallet/useSendFormChangeHandlers.ts
+++ b/packages/suite/src/hooks/wallet/useSendFormChangeHandlers.ts
@@ -1,0 +1,41 @@
+import { TokenInfo } from '@trezor/blockchain-link-types';
+
+import { SendContextValues } from 'src/types/wallet/sendForm';
+
+export type HandleAmountChangeParams = {
+    outputId: number;
+    value: string;
+};
+
+export type HandleFiatChangeParams = HandleAmountChangeParams & { token?: TokenInfo };
+
+type UseSendFormChangeHandlersParams = {
+    calculateAmountFromFiat: (outputId: number, value: string, token?: TokenInfo) => void;
+    calculateFiatFromAmount: (outputId: number, value: string) => void;
+    composeRequest: SendContextValues['composeTransaction'];
+    setValue: SendContextValues['setValue'];
+};
+
+export const useSendFormChangeHandlers = ({
+    calculateAmountFromFiat,
+    calculateFiatFromAmount,
+    composeRequest,
+    setValue,
+}: UseSendFormChangeHandlersParams) => {
+    const disableSetMaxAndRecomposeTransaction = (outputId: number) => {
+        setValue('setMaxOutputId', undefined);
+        composeRequest(`outputs.${outputId}.amount`);
+    };
+
+    const handleAmountChange = ({ outputId, value }: HandleAmountChangeParams) => {
+        calculateFiatFromAmount(outputId, value);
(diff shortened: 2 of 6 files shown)


packages/suite/src/hooks/wallet/useSendForm.ts
```diff
@@ -181,6 +182,13 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
         setAmount: sendFormUtils.setAmount,
     });
 
+    const changeHandlers = useSendFormChangeHandlers({
```
komret, line 185: I added this "hook" because I needed to call `composeRequest` which is not available in `useSendFormFields` and cannot be added there because it is returned from `useSendFormCompose` which relies on `useSendFormFields` via `setAmount`.

packages/suite/src/hooks/wallet/useSendFormChangeHandlers.ts
```diff
@@ -0,0 +1,54 @@
...
+    setValue: SendContextValues['setValue'];
+};
+
+export const useSendFormChangeHandlers = ({
+    calculateAmountFromFiat,
+    calculateFiatFromAmount,
+    composeRequest,
+    setValue,
+}: UseSendFormChangeHandlersParams) => {
+    const handleAmountOrFiatChange = ({
```
komret: I moved the change handlers here because there was duplicate code in `Amount` and `FiatInput`. This is certainly more lines, but avoiding repetition is worth it in my opinion.

packages/suite/src/hooks/wallet/useSendFormFields.ts
```diff
@@ -11,86 +17,147 @@ import { SendContextValues, UseSendFormState } from 'src/types/wallet/sendForm';
...
-}: Props) => {
+}: UseSendFormFieldsParams) => {
     const { shouldSendInSats } = useBitcoinAmountUnit(network.symbol);
     const currentRates = useSelector(selectCurrentFiatRates);
 
-    const calculateFiat = useCallback(
-        (outputIndex: number, amount?: string) => {
-            const outputError = errors.outputs ? errors.outputs[outputIndex] : undefined;
-            const error = outputError ? outputError.amount : undefined;
+    const getCurrentFiatRate = useCallback(
```
komret, line 42: Moved here to avoid repetition between the hook and components.

packages/suite/src/hooks/wallet/useSendFormFields.ts
```diff
@@ -11,86 +17,147 @@ import { SendContextValues, UseSendFormState } from 'src/types/wallet/sendForm';
...
+            const targetValue = output[target];
+            if (target === 'fiat' && typeof targetValue !== 'string') {
+                return; // fiat input not registered (testnet or fiat not available)
+            }
+            const targetInputName = `outputs.${outputId}.${target}` as const;
+            const outputError = errors.outputs ? errors.outputs[outputId] : undefined;
+            const error = outputError
+                ? outputError[target === 'fiat' ? 'amount' : 'fiat']
+                : undefined;
+            if (error || !value) {
```
komret, line 79: Here's what fixes the bug. In `FiatInput`, `error` was read from context where it was not yet updated at the time of triggering the change handler. Thus, this behaviour was delayed by one keyboard stroke.
Accessing `error` directly from `useForm` solves the issue. `Amount` did not have this problem, so I just unified the approach between the two inputs.

packages/suite/src/views/wallet/send/Outputs/Amount/FiatInput.tsx
```diff
@@ -236,29 +167,27 @@ export const FiatInput = ({
     );
 
     return (
-        <Wrapper>
```
komret, line 239: Useless wrapper, it looks the same without it.

packages/suite/src/hooks/wallet/useSendFormChangeHandlers.ts
Lemonexe: That sounds good to centralize change handlers here, but are you sure about the `handleAmountOrFiatChange` and `calculateTheOtherValue`?
This layer of abstraction add complexity, I think it'd just fine if there were too functions `handleAmountChange` and `handleFiatChange` with some boilerplate. But ok :shrug:

packages/suite/src/hooks/wallet/useSendFormFields.ts
```diff
@@ -11,86 +17,147 @@ import { SendContextValues, UseSendFormState } from 'src/types/wallet/sendForm';
...
+                value: amount,
+            });
+        },
+        [calculateFiatFromAmountOrViceVersa, shouldSendInSats, network.symbol],
+    );
+
+    const calculateAmountFromFiat = useCallback(
+        (outputId: number, fiat: string) => {
+            const calculateFormattedAmountValue = (fiat: string, fiatRate: number) => {
+                const amount = fromFiatCurrency(fiat, network.decimals, fiatRate);
```
Lemonexe: the original implementation used `token.decimals || network.decimals` ..?

packages/suite/src/hooks/wallet/useSendFormChangeHandlers.ts
komret: I agree that was a bit wild. What about 7200dfcd4a8970a766a1ec936c719f515a3d8cac?

Lemonexe (approved):
Great refactoring! It's quite complex, so it took me a while to follow all the changes, as I was not familiar with original code. It seems to me it replicates the original behavior, but if I missed something, idk. I think the depth of abstraction is ok.

Tested and works well :heavy_check_mark:

packages/suite/src/hooks/wallet/useSendFormFields.ts
komret: Nice catch! c49af7248472705a6f9e90cdabaecce8d08fd2a4

PR #15476, opened 2024-11-20
feat(message-system): warn users about t1b1 fw update

MiroslavProchazka:
Adding a new secure message sequence to warn users on old t1b1 fw versions to check their wallet backup before proceeding with update. 

Tracked in notion [here](https://www.notion.so/satoshilabs/Warn-users-about-risks-of-updating-of-old-FWs-on-T1B1-144dc526060680798410c77adca85bd0?pvs=4)

Tested in [here](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1732004776051999)

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
-    "timestamp": "2024-11-14T00:00:00+00:00",
-    "sequence": 64,
+    "timestamp": "2024-11-20T00:00:00+00:00",
+    "sequence": 65,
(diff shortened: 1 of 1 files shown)


Lemonexe (approved):
LGTM, the route `/recovery` is valid.

Just FYI [there is already a message](https://github.com/trezor/trezor-suite/blob/0822b72ebb63ad533c8f2fc33cee4348e22c087e/suite-common/message-system/config/config.v1.json#L793-L808) that matches <1.7.0, but the old one is only a `warning`, so it will never be visible, probably.

PR #15491, opened 2024-11-20
chore: make warrning from ESLint to fail the tests (non-zero exit code)

peter-sanderson:
The ESLint won't tail the command for warnings. This change will make it exit with non-zero exit code.


Lemonexe (approved):
Works as expected, law and order must be maintained :policeman: :ok_hand:

PR #15503, opened 2024-11-21
ci(github): add test-connect-mist to release branch

karliatto:
## Description
Adding test-connect-misc so it is run also on release branch. So we make sure when we release connect that all connect related tests are :green_circle:


mroz22 (approved):
thanks!

PR #15518, opened 2024-11-22
chore(components): update Select component menu background

### LINKED ISSUE #14428: Dropdown component fix
Dropdown BG is always …/Surface/Elevation 1, regardless on which surface it is. 

Image

seibei-iguchi:


jvaclavik (approved):
looks good except warnings

image

PR #15525, opened 2024-11-22
fix(connect): device connect/reset issue workarounds

martykan:
## Description

WIP

1. Commented out all device.reset during open/close
2. Add workaround for `assertType: Response of unexpected type` during initialize


martykan:
The problems were fixed in a better way

PR #9931, opened 2023-11-13
Build desktop main code with `webpack`

marekrjpolak:
## Description

In order to deduplicate bundled js code for electron main process, `webpack` is used for the whole thing (instead of using it only for connect while preserving `esbuild` for the rest, as it was after #9552).

PR #15544, opened 2024-11-25
WIP: bundle main+preload as ES

### LINKED ISSUE #14482: Electron-main with ESM-only deps
Configure Electron build process to output `main` and `preload` bundles as ES instead of CommonJS.
That is possible since [electron 28.0.0](https://www.electronjs.org/docs/latest/tutorial/esm).

It is needed to upgrade dependencies which do not support CommonJS anymore:
- `electron-store@10.0.0` (this is for sure)
- `long` really? To be confirmed
- `@noble/hashes@2.0.0` really? To be confirmed

Because we use bundler (webpack) this solves most of our problems, because the bundler takes ESM package and squashes it into output js file. So actually **only**  libraries which cannot be bundled (e.g. have a .node binary) and need to have a runtime require are a problem.

Lemonexe:
## Description

- Bump electron-store to 10.0.0
  - :arrow_right: does no longer support CJS, build crashes
- Webpack: change output target for main+preload bundles as ES instead of CJS
- Electron-builder: configure to recognize ES
- Fix optional externals deps, see comments below

### Does not work :disappointed: 

- Build is successful, the output is indeed ES.
- However, upon starting the application, it freezes - no window is opened, but no error either.
- It reads the app.js (main process bundle) but does not call the functions which initiate the browser window.
- So there might be something wrong with the way webpack bundles ES
  - after all, it's an experimental feature

### Thoughts? :thought_balloon:

- for main+preload, replace webpack with different bundler
  - but leave webpack for `renderer`
  - we used to have ESbuild, but [replaced it with webpack](https://github.com/trezor/trezor-suite/pull/9931) because it supports chunking 
  - explore https://rspack.dev/
    - it is designed to have compatible API with webpack, so it could be very small effort :slightly_smiling_face: 
  - `rollup` (or possibly `vite`) supports chunking too, and has first-class support for ES
    - but will likely be a lot of effort :confused: 

## Related Issue

Attempt to resolve #14482


packages/suite-desktop-core/webpack/core.webpack.config.ts
```diff
@@ -74,17 +74,18 @@ const config: webpack.Configuration = {
...
         ...dependencies,
         ...devDependencies,
-        'bufferutil', // optional dependency of ws lib
-        'memcpy', // optional depencency of bytebuffer lib
-        'utf-8-validate', // optional dependency of ws lib
-        'osx-temperature-sensor', // optional dependency of systeminformation lib
+        // 'bufferutil', // optional dependency of ws lib
+        // 'memcpy', // optional depencency of bytebuffer lib
+        // 'utf-8-validate', // optional dependency of ws lib
+        // 'osx-temperature-sensor', // optional dependency of systeminformation lib
```
Lemonexe, line 89: These are all sub-dependencies that are only optionally required by our dependencies, [example](https://github.com/sebhildebrandt/systeminformation/blob/3a92931c7d46605ffddc1aacb97a9727273b2888/lib/cpu.js#L1227-L1230).

When specified in `externals`, webpack will try to provide them, but it crashes at run time when output target is ES. I don't know if any actual suite functionality relies on those libs.

packages/suite-desktop-core/webpack/core.webpack.config.ts
```diff
@@ -112,6 +113,8 @@ const config: webpack.Configuration = {
         alias: {
             '@emurgo/cardano-serialization-lib-nodejs': '@emurgo/cardano-serialization-lib-browser',
             '@trezor/connect$': '@trezor/connect/src/index', // alternative for "module": "src/index" in connect's package.json
+            // optional dependency of systeminformation lib mocked as undefined (webpack needs it when targetting ES)
+            'osx-temperature-sensor': false,
```
Lemonexe, line 118: unlike other deps in `externals`, this one needs to be mocked as undefined, otherwise webpack crashes during building the desktop core `@trezor/suite-desktop-core build:app:dev`

PR #15551, opened 2024-11-25
feat(e2e): Added pin to check backup t1b1 test

HajekOndrej:
## Description
Added pin to the check backup test to test the full flow of the backup check.

## Related Issue

Resolve


Vere-Grey (approved):
Super minor: 4 digits pin might be closer to real use and could catch some specific issues.

PR #15555, opened 2024-11-25
feat(suite): enhance FW checks reporting

Lemonexe:
## Description

- connect: add `errorPayload` for detailed debugging of FW hash check `other-error`
- extract sentry reporting from `DeviceCompromised`, and fire it always when there is error
  - currently the sentry report only fires when error & the check is enabled
- split `useEffect` for revision check & hash check
  - solves potential problem with duplicate sentry events
    - unknown if this is a problem, it probably gets deduplicated via the anonymous user ID

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -750,7 +750,11 @@ export class Device extends TypedEmitter<DeviceEvents> {
     }
 
     async checkFirmwareHash(): Promise<FirmwareHashCheckResult | null> {
-        const createFailResult = (error: FirmwareHashCheckError) => ({ success: false, error });
+        const createFailResult = (error: FirmwareHashCheckError, errorPayload?: unknown) => ({
+            success: false,
+            error,
+            errorPayload,
+        });
 
         const baseUrl = DataManager.getSettings('binFilesBaseUrl');
         const enabled = DataManager.getSettings('enableFirmwareHashCheck');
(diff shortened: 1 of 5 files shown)


packages/connect/src/device/Device.ts
```diff
@@ -793,22 +797,24 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
         // handle rejection of call by a counterfeit device. If unhandled, it crashes device initialization,
         // so device can't be used, but it's preferable to display proper message about counterfeit device
-        const getFirmwareHashOptional = async () => {
-            try {
-                return await this.getCommands().typedCall('GetFirmwareHash', 'FirmwareHash', {
-                    challenge,
-                });
-            } catch {
-                return null;
```
Lemonexe, line 802: at this point I need to return from `checkFirmwareHash`, incl. the caught error. Can't do that when the API call is wrapped in `getFirmwareHashOptional`, so I removed that fn.
It was there because you cannot define `const` in `try` block and then access it outside of it.
But the rest of the function is safe, so I believe it can all be in the try block.

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
```diff
@@ -45,12 +26,6 @@ export const DeviceCompromised = () => {
         }
     };
 
-    useEffect(() => {
-        const commonData = { revision, version, vendor };
-        if (revisionCheckError) reportCheckFail('revision', { ...commonData, revisionCheckError });
-        if (hashCheckError) reportCheckFail('hash', { ...commonData, hashCheckError });
-    }, [revisionCheckError, hashCheckError, revision, vendor, version]);
```
Lemonexe, line 52: This had a problem: when only one of the error changes, it re-reports both of them

packages/connect/src/types/device.ts
```diff
@@ -66,7 +66,7 @@ export type FirmwareHashCheckError =
     | 'other-error';
 export type FirmwareHashCheckResult =
     | { success: true }
-    | { success: false; error: FirmwareHashCheckError };
+    | { success: false; error: FirmwareHashCheckError; errorPayload?: unknown };
```
peter-sanderson, line 69: :clap:

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -61,6 +62,8 @@ export const Preloader = ({ children }: PropsWithChildren) => {
     const isFirmwareAuthenticityCheckDismissed = useSelector(
         selectIsFirmwareAuthenticityCheckDismissed,
     );
+    // report firmware authenticity failures even when the UI is disabled
```
peter-sanderson, line 65: I am OK to remove comment and move it to the function name:

```ts
useReprtDeviceCompomisedEvenWhenTheUiIsDisabled()
```

IMO better then comment

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
peter-sanderson: Nice catch!

peter-sanderson (approved):

packages/suite/src/components/suite/Preloader/Preloader.tsx
Lemonexe: We discussed on slack that we will leave it, because the suggested name also describes _expected usage_ instead of describing what it does (preferred).

PR #15557, opened 2024-11-25
chore(suite-native): expo-video

### LINKED ISSUE #15456: Migrate from `expo-av` to `expo-video`
`expo-av` is depreacted in favor of `expo-video`. We use it only in once place anyway so it should be quick.

PeKne:
Deprecated `expo-av` replaced by `expo-video`.
## Related Issue

Resolve #15456


suite-native/app/package.json
```diff
@@ -102,6 +101,7 @@
         "expo-status-bar": "^2.0.0",
         "expo-system-ui": "^4.0.2",
         "expo-updates": "0.26.6",
+        "expo-video": "~2.0.1",
```
Nodonisko: I would prefer fixed version or or `^`, tilda is too benevolent for me :D

suite-native/video-assets/src/components/Video.tsx
```diff
@@ -52,16 +54,14 @@ export const Video = ({
...
-                style={applyStyle(videoStyle, { aspectRatio })}
-                source={videoSource}
-                shouldPlay
-                isMuted={isMuted}
-                isLooping
-                resizeMode={ResizeMode.CONTAIN}
-                onLoad={() => setIsLoading(false)}
-                {...restProps}
-            />
+            {!isLoading && (
```
Nodonisko, line 58: Is this fast enough to not cause any layout jumps? I am thinking about some empty View with same dimensions to prevent that.

Nodonisko (approved):

suite-native/app/package.json
PeKne: Is not it the other way? I though that tilda is stricter than caret. https://medium.com/@codejuggler/understanding-the-difference-between-tilde-and-caret-in-package-json-aacb092a67b7

suite-native/video-assets/src/components/Video.tsx
PeKne: There is no layout shift happening. 

This is how it looks when loading: 
Snímek obrazovky 2024-11-26 v 14 57 28

And when it loads:
Snímek obrazovky 2024-11-26 v 14 57 43

suite-native/app/package.json
matejkriz: Tilda is stricter, but we have fixed versions thanks to yarn.lock anyway... But mixing those two brings even more confusion?

PeKne: Alright, I unified so the caret is used (same for all the expo packages).

Nodonisko (approved):

PR #15565, opened 2024-11-25
Warn users on old T1B1 FWs copy tweak

MiroslavProchazka:
Adjusting the copy and removing the CTA button as we can't do a check backup on unsupported FW. Follow up of https://github.com/trezor/trezor-suite/pull/15476

Tracked in Notion [here](https://www.notion.so/satoshilabs/Warn-user-on-old-T1B1-FW-tweak-146dc526060680feac5fd2fd2d42d4d3?pvs=4)


komret (approved):

STew790:
**QA OK**
image

PR #15568, opened 2024-11-25
fix(suite-data): temporarily use t3b1 animation for t3w1

komret:
There was a T2B1 animation before as a placeholder, but some logic related to T2B1 has changed in the meantime. Replacing with T3B1 animation until we get a dedicated T3W1 animation.


MiroslavProchazka (changes requested):
The placeholder file is already there, see [this](https://github.com/trezor/trezor-suite/blob/develop/packages/suite-data/files/videos/lottie/trezor_t3w1_connect.json)

MiroslavProchazka:
Ah, I see that it is broken, OK.

MiroslavProchazka (approved):

PR #15574, opened 2024-11-26
feat(connect): add blockchainGetInfo method, use it to fetch latest Solana block hash when sending

martykan:
## Description

1. Added `TrezorConnect.blockchainGetInfo()`, which exposes `blockchain.getNetworkInfo()`
2. Used it to fetch latest Solana blockhash/blockheight when signing in send form, instead of the one in the account which may be delayed
3. Fetch isTestnet only once in Solana and pass it, to avoid subsequent unnecessary `getGenesisHash` calls


mroz22 (approved):
👍

PR #15588, opened 2024-11-26
chore(protobuf): update protobuf definitions (display type enum added)

mroz22:


packages/protobuf/src/index.ts
```diff
@@ -5,4 +5,4 @@ export * from './types';
 export { parseConfigure, createMessageFromName, createMessageFromType } from './utils';
 export * as MessagesSchema from './messages-schema';
 // It's problem to reexport enums when they are under MessagesSchema namespace, check packages/connect/src/types/device.ts
-export { DeviceModelInternal } from './messages-schema';
+export { DeviceModelInternal, DisplayRotation, Enum_DisplayRotation } from './messages-schema';
```
mroz22: is this the way we want to export these? cc @martykan

martykan: I don't know... `DeviceModelInternal` seems important enough to re-export separately, but I don't think it's the best approach for every single enum. 
If we want to re-export in `device.ts` something like this should work:
```
export type DisplayRotation = PROTO.DisplayRotation;
```

packages/connect/src/types/device.ts
```diff
@@ -147,7 +147,7 @@ export type UnreadableDevice = BaseDevice & {
 export type Device = KnownDevice | UnknownDevice | UnreadableDevice;
 export type Features = PROTO.Features;
 
-export { DeviceModelInternal } from '@trezor/protobuf';
+export { DeviceModelInternal, Enum_DisplayRotation, DisplayRotation } from '@trezor/protobuf';
```
peter-sanderson: the export of the underscored value seems odd, but if its intentional :shrug:

peter-sanderson (approved):

packages/suite/src/views/settings/SettingsDevice/DisplayRotation.tsx
```diff
@@ -76,12 +79,12 @@ export const DisplayRotation = ({ isDeviceLocked }: DisplayRotationProps) => {
...
                     selectedOption={currentRotation ?? undefined}
                     options={DISPLAY_ROTATIONS}
-                    onChange={(value: number) => {
+                    onChange={(value: DisplayRotationEnum) => {
                         dispatch(applySettings({ display_rotation: value }));
                         analytics.report({
                             type: EventType.SettingsDeviceChangeOrientation,
                             payload: {
-                                value: value as RotationValue,
+                                value: Enum_DisplayRotation[value],
```
mroz22: todo: utility

PR #15595, opened 2024-11-27
fix(suite-native): account balance includes XRP reserve

### LINKED ISSUE #13900: XRP balance mismatch
Amount on Dashboard and Account detail shown including 10 XRP reserve, My assets, Receive and Select account shown without.

It has to be unified - the amount should always include those locked 10 XRP.

PeKne:
XRP reserve is always included in the account balance.

## Related Issue

Resolve #13900

## Screenshots:

https://github.com/user-attachments/assets/ef40ea28-55ae-440b-b22a-0ca8bb0ba643

### DIFF
--- a/suite-native/accounts/src/components/AccountsList/AccountsListItem.tsx
+++ b/suite-native/accounts/src/components/AccountsList/AccountsListItem.tsx
@@ -125,15 +125,12 @@ export const AccountsListItem = ({
                         value={fiatBalance}
                     />
                 ) : (
-                    <CryptoToFiatAmountFormatter
-                        value={account.availableBalance}
-                        network={account.symbol}
-                    />
+                    <CryptoToFiatAmountFormatter value={account.balance} network={account.symbol} />
                 )
             }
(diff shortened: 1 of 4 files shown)


Nodonisko:
Woudn't this break something else? I think there was some concers that `balance` does or doesn't include non confirmed transactions etc. (IDK it's a mess :D)

PeKne:
@Nodonisko it should not. Desktop is using `balance` everywhere as well.

PeKne:
I'm missing some go back button in the cancelled firmware update state. If the user cancels the flow on device, only way to leave it is by restarting the app.

https://github.com/user-attachments/assets/0c5b16cd-87ce-4b88-a728-066cfda6a8dd

Nodonisko:
@PeKne I think you wanted to add this comment to FW upgrade PR :D Anyway this is bug, it should return automatically when flow is cancelled on device.

Nodonisko (approved):

PR #15596, opened 2024-11-27
15468 send modal quick ux fixes

seibei-iguchi:
- updated InfoItem component: added variant prop, renamed labelTypographyStyle to typographyStyle
- refactored settings page (using InfoItem)
- fixed receive dashboard
- fixed account header in sidebar
- added OutlineHighlight component
- refactored coinmarker verify page

### Before
Screenshot 2024-11-27 at 11 55 24
Screenshot 2024-11-27 at 11 55 12

### After
Screenshot 2024-11-27 at 10 58 06
Screenshot 2024-11-27 at 10 58 25


packages/components/src/components/InfoItem/InfoItem.tsx
```diff
@@ -50,18 +42,18 @@ export type InfoItemProps = AllowedFrameProps &
         direction?: FlexDirection;
         iconName?: IconName;
         label: ReactNode;
-        labelTypographyStyle?: TypographyStyle;
+        variant?: IconVariant;
```
jvaclavik: I'm not fan of reusing variants for other components - it can be a bit blackbox when we for example add new variant for icon, but we don't want interfere with other components. 
Why not to create `InfoItemVariant` here?

packages/components/src/components/InfoItem/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+        bottom: 'flex-end',
+    };
+
+    return alignItemsMap[verticalAlignment];
+};
+
+export const mapTypographyStyleToIconSize = (
+    typographyStyle: TypographyStyle,
+): IconSize | number => {
+    const iconSizeMap: Record<TypographyStyle, IconSize | number> = {
```
jvaclavik, line 22: This leads me to that we might should unify somehow `small | medium | large` with `xs, sm, lg, xl, xxl` etc. (as we have in spacings). WDYT? (definitely out of scope, but just thing to think about for the future)

packages/suite/src/components/OutlineHighlight.tsx
```diff
@@ -0,0 +1,39 @@
...
+const Content = styled.div`
+    position: relative;
+`;
+
+type OutlineHighlightProps = {
+    shouldHighlight?: boolean;
+    children: ReactNode;
+};
+
+export const OutlineHighlight = ({ shouldHighlight, children }: OutlineHighlightProps) => (
```
jvaclavik, line 35: What for is this component? 

Couldn't we unify it with this?
image

jvaclavik (approved):
Good job. Few comments, but LGTM!

packages/suite/src/components/OutlineHighlight.tsx
seibei-iguchi: It's this exact thing in the screenshot:) I've turned it into a component, it was just a block of CSS before.

jvaclavik: That's what I wanted to hear :D

jvaclavik: Btw there is a small glitch in this component (you can see that border bottom thinner than others), don't want to also fix it when you have context here? 🙂

seibei-iguchi: The screenshot is from my branch or develop?

jvaclavik: I think it was screenshot from `Box` branch :D

seibei-iguchi: I think this doesn't happen with the new CSS. BTW how do you enable the highlight manually? I couldn't figure it out.

jvaclavik: Ok, so it's maybe fixed already.

You can enable it by clicking to Tor:

image

or here:

Screenshot 2024-11-28 at 10 45 46

packages/components/src/components/Box/Box.tsx
```diff
@@ -10,6 +10,7 @@ import { TransientProps } from '../../utils/transientProps';
 
 export const allowedBoxFrameProps = [
     'margin',
+    'padding',
```
jvaclavik, line 13: Is it necesary to have this in `Box` since we have `margin`?

packages/components/src/components/Flex/Flex.tsx
```diff
@@ -15,6 +15,7 @@ import { useElevation } from '../ElevationContext/ElevationContext';
 
 export const allowedFlexFrameProps = [
     'margin',
+    'padding',
```
jvaclavik, line 18: Is it necessary since we have `margin` here?

packages/components/src/components/InfoItem/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+import { IconSize } from '../Icon/Icon';
+import { InfoItemVerticalAlignment } from './types';
+
+export const mapVerticalAlignmentToAlignItems = (
+    verticalAlignment: InfoItemVerticalAlignment,
+): FlexAlignItems => {
+    const alignItemsMap: Record<InfoItemVerticalAlignment, FlexAlignItems> = {
+        top: 'flex-start',
+        center: 'center',
+        bottom: 'flex-end',
```
jvaclavik, line 13: shouldn't we keep `flex-start`, `flex-end` since we use this in Flex components? Or the opposite - shouldn't we unify it to this solution in Flex?

packages/components/src/utils/frameProps.tsx
```diff
@@ -63,6 +73,7 @@ export type FrameProps = {
     flex?: Flex;
     position?: Position;
     cursor?: Cursor;
+    zIndex?: number;
```
jvaclavik, line 76: shouldn't we allow only numbers defined in `zIndices`? But it's probably too cruel, right?

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
```diff
@@ -238,242 +217,268 @@ export const TransactionItem = memo(
             <Wrapper
```
jvaclavik, line 217: offtopic: This whole component is crazy, really hard to review and should be divided into multiple components.

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountGroup.tsx
```diff
@@ -108,27 +89,45 @@ export const AccountGroup = ({
...
+        <Column gap={spacings.xxxs}>
             {heading !== null && (
-                <Header
-                    $isOpen={isOpen}
-                    onClick={!keepOpen ? onClick : undefined}
-                    data-testid={`@account-menu/${type}`}
+                <Box
+                    position={{ type: 'sticky', top: 0 }}
+                    zIndex={30}
+                    cursor={!keepOpen ? 'pointer' : 'default'}
```
jvaclavik, line 97: `"default"` or `undefined`?

jvaclavik (approved):
LGTM, approving 👍. Just few minor comments. Please merge it soon and don't add more stuff 😅

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
seibei-iguchi: Agree. But I didn't want to refactor this, as it's gonna change anyway.

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountGroup.tsx
seibei-iguchi: I guess it's the same.

packages/components/src/components/Box/Box.tsx
seibei-iguchi: Yep, there are cases when padding behaves differently to margin (for example, margin collapse etc)

packages/components/src/components/Flex/Flex.tsx
seibei-iguchi: The same as above. Also, I needed it in the AccountGroup refactor, that's a case where margin won't work and we need padding.

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/AccountGroup.tsx
jvaclavik: Yes and no, with undefined it's possible to change cursor behaviour from top level component. Just thinking if we actually want to use default if we don't want to change cursor behaviour.

packages/components/src/components/InfoItem/utils.ts
seibei-iguchi: I don't think it's the same mapping. We might, for example, decide that the "top" in the context of InfoItem is not flex-start, but baseline (it was like that before, I changed it because it caused problems in some cases).

bosomt:
**QA OK**

**Info:**
 - Suite version: web 25.1.0 (e2860882d84f904cab9a204ec964d5725455fe89)
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.7 regular (revision ef02c4de5d492a5c5d562984ec65a302d093dfd0)

PR #15597, opened 2024-11-27
Feat/coin control ordering

### LINKED ISSUE #14545: Coin Control: UTXO sorting
To improve UTXO management and simplify handling potential dust UTXOs, please add UTXO sorting to the Coin Control feature alongside the existing search field.

[Figma Design](https://www.figma.com/design/jsdvZlsY36rCc1E6PoItcz/Global-actions?node-id=4102-4590&t=OP9kyNRVbQpztKsT-1)

MVP: Add a select menu to the right of the Search input and allow sorting accordingly.

2nd stage: Coin Control sidebar

komret:
## Description

I tried to bear performance in mind.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14545

## Screenshots:
Screenshot 2024-12-02 at 18 50 56
Screenshot 2024-12-06 at 18 05 09


packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelectionList.tsx
```diff
@@ -65,9 +65,7 @@ export const UtxoSelectionList = ({
 }: UtxoSelectionListProps) => {
     const { account } = useSendFormContext();
 
-    const accountTransactions = useSelector(state =>
-        selectAccountTransactionsWithNulls(state, account.key),
-    );
+    const accountTransactions = useSelector(state => selectAccountTransactions(state, account.key));
```
komret, line 68: I accidentally committed this together with the refactoring, it should have been separate. I noticed that we don't need to use nulls here and filtering them out won't impact performance because we're going to iterate over the array anyway.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/CoinControl.tsx
```diff
@@ -18,41 +18,13 @@ import { selectCurrentTargetAnonymity } from 'src/reducers/wallet/coinjoinReduce
 import { selectLabelingDataForSelectedAccount } from 'src/reducers/suite/metadataReducer';
 import { filterAndCategorizeUtxos } from 'src/utils/wallet/filterAndCategorizeUtxosUtils';
 
-import { UtxoSelectionList } from './UtxoSelectionList';
+import { UtxoSortingSelect } from './UtxoSortingSelect';
+import { UtxoSelectionList } from './UtxoSelectionList/UtxoSelectionList';
 import { UtxoSearch } from './UtxoSearch';
 
-const Row = styled.div`
```
peter-sanderson, line 24: image

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
```diff
@@ -76,23 +87,76 @@ export const useUtxoSelection = ({
         composeRequest,
     ]);
 
+    const sortUtxos = (utxos: AccountUtxo[]): AccountUtxo[] => {
```
peter-sanderson: ~this function can be outside of the upper function scope (I would even move it to separate file)~

Edit: I just realized there is `accountTransactions` so it needs to be here (or it needs to be passed and its a mess then with param)

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
```diff
@@ -76,23 +87,76 @@ export const useUtxoSelection = ({
...
+            new BigNumber(b.amount).comparedTo(new BigNumber(a.amount));
+
+        // Manipulating the array directly would cause a runtime error, so we create a copy.
+        const utxosForSorting = utxos.slice();
+
+        switch (utxoSorting) {
+            case 'newestFirst':
+                return utxosForSorting.sort(sortFromNewestToOldest);
+            case 'oldestFirst':
+                return utxosForSorting.sort(sortFromNewestToOldest).reverse();
```
peter-sanderson: Reverse is `O(n)` operation can be done by just `*-1` in the sort function

peter-sanderson: => `.sort( (a, b) => sortFromLargestToSmallest(a, b) * -1)`

(Or you can swap `a, b` => `b, a` but its not that obvious then

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
peter-sanderson: But anyway, maybe it would be useful to drill the  `accountTransactions` down so we can extract it after all and test it

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
peter-sanderson: I did some playing with the code and iterated to this https://github.com/trezor/trezor-suite/commit/0e68afa1560598acb2168562cb2a05adb74efcfb

Feel free to use it, or not. Up to you.

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
```diff
@@ -76,23 +87,76 @@ export const useUtxoSelection = ({
         composeRequest,
     ]);
 
+    const sortUtxos = (utxos: AccountUtxo[]): AccountUtxo[] => {
+        if (!utxoSorting) {
+            return utxos;
+        }
+
+        const sortFromNewestToOldest = (a: AccountUtxo, b: AccountUtxo) => {
+            let valueA: number;
```
peter-sanderson: I suggest to get rid of the `let` by doing something like this:

```ts
const sortFromNewestToOldest: UtxoSortingFunction =
    ({ accountTransactions }) =>
    (a: AccountUtxo, b: AccountUtxo) => {
        if (a.blockHeight > 0 && b.blockHeight > 0) {
            return b.blockHeight - a.blockHeight;
        }

        // Pending transactions do not have blockHeight, so we must use blockTime of the transaction instead.
        const getBlockTime = (txid: string) => {
            const transaction = accountTransactions.find(transaction => transaction.txid === txid);

            return transaction?.blockTime ?? 0;
        };

        return new BigNumber(getBlockTime(b.txid)).comparedTo(new BigNumber(getBlockTime(a.txid)));
    };
```

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
```diff
@@ -76,23 +87,76 @@ export const useUtxoSelection = ({
...
+                        transaction => transaction.txid === txid,
+                    );
+
+                    return transaction?.blockTime ?? 0;
+                };
+                valueA = getBlockTime(a.txid);
+                valueB = getBlockTime(b.txid);
+            }
+
+            return new BigNumber(valueB ?? 0).comparedTo(new BigNumber(valueA ?? 0));
```
peter-sanderson: nit: is `BigNumber` here necessary? isn't `valueB - valueA` just fine?

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
peter-sanderson: I did some playing with the code and iterated to this https://github.com/trezor/trezor-suite/commit/0e68afa1560598acb2168562cb2a05adb74efcfb

Feel free to use it, or not. Up to you.

peter-sanderson:
I have tested it and it works. Good job!

However I would like to see tests for the sorting function especially the complicated on (largest/newest) + some code polishing.

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
komret: I used `reverse` because in case when the compared values are the same, I want the items to reverse order when I switch e.g. from newest first to oldest first. I find that to be the expected behaviour. See the two UTXOs from Oct 11 on your branch:
Screenshot 2024-12-03 at 14 38 24
Screenshot 2024-12-03 at 14 38 18

peter-sanderson: This be solved by secondary sorting (`txid` or something unique). This will assure that sorting is stable.

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
komret: Removed.

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
komret: I kept the `let`s there because it just seems easier with the secondary sorting.

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
komret: Quite complex but done.

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
komret: Quite complex, but probably worth it for the tests. Commit stolen.

komret:
> However I would like to see tests for the sorting function especially the complicated on (largest/newest) + some code polishing.

Thanks, I shamelessly stole your commits. Apologies for not using fixups :( I only made changes to the third commit, though.

The changes I made:
- move sorting functions to a [separate file](https://github.com/trezor/trezor-suite/blob/1d0258e0b8eb44b0eac6865aaeea126b5112f904/packages/suite/src/utils/wallet/utxoSortingUtils.ts)
- implement secondary and tertiary sorting by `txid` and `vout`
- add unit tests

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
```diff
@@ -0,0 +1,43 @@
...
+];
+
+const ACCOUNT_TRANSACTIONS = [
+    testMocks.getWalletTransaction({ txid: 'txid1', blockTime: undefined }),
+    testMocks.getWalletTransaction({ txid: 'txid2', blockTime: 1 }),
+    testMocks.getWalletTransaction({ txid: 'txid3', blockTime: 2 }),
+];
+
+describe('sortUtxos', () => {
+    it('should sort UTXOs by newest first', () => {
```
peter-sanderson: The test description should not start with `should`. I believe we even have it in Notion somewhere (but I am lazy to search it). 

```suggestion
    it('sorts UTXOs by newest first', () => {
```

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        const sortedUtxos = sortUtxos(UTXOS, 'newestFirst', ACCOUNT_TRANSACTIONS);
+        expect(sortedUtxos).toEqual([UTXOS[3], UTXOS[1], UTXOS[2], UTXOS[0]]);
+    });
+
+    it('should sort UTXOs by oldest first', () => {
+        const sortedUtxos = sortUtxos(UTXOS, 'oldestFirst', ACCOUNT_TRANSACTIONS);
+        expect(sortedUtxos).toEqual([UTXOS[0], UTXOS[2], UTXOS[1], UTXOS[3]]);
+    });
+
+    it('should sort UTXOs by largest first', () => {
```
peter-sanderson: The (unit) test shall test only one thing. And the assertion shall be done in a way, it is clearly visible what is important (the amount of transaction). This way to work with test I have to go back and forth to see what hides behind `UTXOS[1]`

This is  what I believe is much more readable, and maintainable:
```ts
    it('sorts by size, largest first', () => {
        const utxos = [
            testMocks.getUtxo({ amount: '1', txid: 'txid1' }),
            testMocks.getUtxo({ amount: '2', txid: 'txid2' }),
        ];
        const sortedUtxos = sortUtxos(utxos, 'largestFirst', ACCOUNT_TRANSACTIONS);
        expect(sortedUtxos.map(it => it.amount)).toEqual(['2', '1']);
    });
```

packages/suite/src/utils/wallet/utxoSortingUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    return secondaryComparison;
+};
+
+const sortFromLargestToSmallest: UtxoSortingFunctionWithContext = () => (a, b) => {
+    const comparisonResult = new BigNumber(b.amount).comparedTo(new BigNumber(a.amount));
+
+    if (comparisonResult === 0) {
+        return performSecondarySorting(a, b);
```
peter-sanderson, line 30: this can be done only once in the upper layer here: https://github.com/trezor/trezor-suite/pull/15597/files#diff-fbfad5b45891e823066e9dc7e2b086d54a0c7de5b3fd3fc8612f37daf291333bR83

peter-sanderson: Edit: not really, then the `* -1` trick does not work

peter-sanderson: Ok I figured that out, its definitely "cleaner" solution (it separates the responsibility), but it adds an abstraction and makes it more complex.

It was nice exercise , tho :D

```ts
const wrapSecondarySorting =
    (sortFunction: UtxoSortingFunctionWithContext): UtxoSortingFunctionWithContext =>
    context =>
    (a, b) => {
        const result = sortFunction(context)(a, b);

        if (result !== 0) {
            return result;
        }

        return performSecondarySorting(a, b);
    };
```

and then:

```ts
const utxoSortMap: Record<UtxoSorting, UtxoSortingFunctionWithContext> = {
    largestFirst: wrapSecondarySorting(sortFromLargestToSmallest),
    smallestFirst:
        context =>
        (...params) =>
            wrapSecondarySorting(sortFromLargestToSmallest)(context)(...params) * -1,

    newestFirst: wrapSecondarySorting(sortFromNewestToOldest),
    oldestFirst:
        context =>
        (...params) =>
            wrapSecondarySorting(sortFromNewestToOldest)(context)(...params) * -1,
};
```

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    it('should sort UTXOs by smallest first', () => {
+        const sortedUtxos = sortUtxos(UTXOS, 'smallestFirst', ACCOUNT_TRANSACTIONS);
+        expect(sortedUtxos).toEqual([UTXOS[0], UTXOS[1], UTXOS[2], UTXOS[3]]);
+    });
+
+    it('should return the original array if utxoSorting is undefined', () => {
+        const sortedUtxos = sortUtxos(UTXOS, undefined, ACCOUNT_TRANSACTIONS);
+        expect(sortedUtxos).toEqual(UTXOS);
+    });
+});
```
peter-sanderson, line 74: Also the stable-sorting mechanism (secondarySorting) shall be tested in isolation for example like this:

```ts
it('sorts by secondary sorting by `txid` and `vout` in case of same values', () => {
        const utxos = [
            testMocks.getUtxo({ amount: '2', txid: 'txid2' }),
            testMocks.getUtxo({ amount: '2', txid: 'txid1', vout: 1 }),
            testMocks.getUtxo({ amount: '2', txid: 'txid1', vout: 0 }),
        ];
        const sortedUtxos = sortUtxos(utxos, 'smallestFirst', ACCOUNT_TRANSACTIONS);
        expect(sortedUtxos.map(it => `${it.txid}:${it.vout}`)).toEqual([
            'txid1:0',
            'txid1:1',
            'txid2:1',
        ]);
    });
```

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    testMocks.getUtxo({ amount: '2', blockHeight: 2, txid: 'txid3', vout: 0 }),
+];
+
+const ACCOUNT_TRANSACTIONS = [
+    testMocks.getWalletTransaction({ txid: 'txid1', blockTime: undefined }),
+    testMocks.getWalletTransaction({ txid: 'txid2', blockTime: 1 }),
+    testMocks.getWalletTransaction({ txid: 'txid3', blockTime: 2 }),
+];
+
+describe('sortUtxos', () => {
```
peter-sanderson: nit: just a little trick: you can do `sortUtxos.name`.

I works better when you rename the function, so this gets renamed as well.

peter-sanderson (commented):
here is another suggestion https://github.com/trezor/trezor-suite/commit/aa0c4737da06d48402ef16fc5e7d2d251f29a4b2

It is a draft, feel free to use it, or not. Up to you.

Espetially tests for `newestFirst` and `oldestFirst` are bit messy.  But I wanted to show the principles.

packages/suite/src/hooks/wallet/form/useUtxoSelection.ts
peter-sanderson: Please reconsider if you decide to refactor secondary sorting by wrapping it from outside.

komret: 82fc888536dc85c065b6730d8e44272e6846ac0e

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
komret: Ok I like it, but I think this is more beneficial in the `expect` clause than the input where some lines can be saved: 82fc888536dc85c065b6730d8e44272e6846ac0e

packages/suite/src/utils/wallet/utxoSortingUtils.ts
komret: I think it's over-engineered at this point, but since it is quite isolated, it shouldn't confuse too many people. Let's use it.

komret:
@peter-sanderson please check 82fc888536dc85c065b6730d8e44272e6846ac0e. If you approve, I will squash it all to the third commit.

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
peter-sanderson: In general I believe that being explicit in tests and make it clear what goes in and out is more important then saving some lines.

For example: `UTXOS.slice(1, 4)` gives me no insight what is the `given` state of the test. 

I think all tests shall follow the pattern:
- `given` (input of the test)
- `when` (what is tested)
- `then` (assert of the output)

All technical details can be hidden and abstracted, but I think the input (what is tested case), the test subject (what piece is tested), and result (what is expected result) shall be visible on the very first sight.

--------

To saving some lines maybe this can be shorted:

```ts
        const utxos = [
            testMocks.getUtxo({ amount: '2', txid: 'txid2' }),
            testMocks.getUtxo({ amount: '2', txid: 'txid1', vout: 1 }),
            testMocks.getUtxo({ amount: '2', txid: 'txid1', vout: 0 }),
        ];
```

to something like:
```ts
const utxos = [utxo_amount2_tx2_vout1, utxo_amount2_tx1_vout1, , utxo_amount2_tx1_vout0]
```

But I think its marginal. 

----

Anyway, this is not the hill to die on, but I just wanted to explain my reasoning.

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    it('sorts by secondary sorting by `txid` and `vout` in case of same values', () => {
+        const sortedUtxos = sortUtxos(UTXOS.slice(1, 4), 'smallestFirst', ACCOUNT_TRANSACTIONS);
+        expect(sortedUtxos.map(it => `${it.txid}:${it.vout}`)).toEqual([
+            'txid2:0',
+            'txid2:1',
+            'txid3:0',
+        ]);
+    });
+
+    it('should return the original array if utxoSorting is undefined', () => {
```
peter-sanderson:
```suggestion
    it('returns the original array if utxoSorting is undefined', () => {
```

peter-sanderson (approved):
:+1: GJ! :rocket:

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
komret: Fixed.

packages/suite/src/utils/wallet/__tests__/utxoSortingUtils.test.ts
komret: Thanks, your explanation makes sense. I'd prefer not to change it here now that it is done, but I'll bear it in mind next time.

PR #15599, opened 2024-11-27
feat(connect): all backupDevice and recoveryDevice even for devices w…

mroz22:
…ith required firmware update

cc @MiroslavProchazka 

test device: `T` & `2.1.8` (first lower then `required: true`

### DIFF
--- a/packages/connect/src/core/AbstractMethod.ts
+++ b/packages/connect/src/core/AbstractMethod.ts
@@ -64,6 +64,7 @@ function validateStaticSessionId(input: unknown): StaticSessionId {
         'DeviceState: invalid staticSessionId: ' + input,
     );
 }
+
 // validate expected state from method parameter.
 // it could be undefined
 function validateDeviceState(input: unknown): DeviceState | undefined {
@@ -306,8 +307,10 @@ export abstract class AbstractMethod<Name extends CallMethodPayload['method'], P
         if (!version) return;
 
         if (
-            device.firmwareStatus === 'required' ||
-            !versionUtils.isNewerOrEqual(version, range.min)
+            this.name !== 'backupDevice' &&
+            this.name !== 'recoveryDevice' &&
+            (device.firmwareStatus === 'required' ||
+                !versionUtils.isNewerOrEqual(version, range.min))
         ) {
             return UI.FIRMWARE_OLD;
         }
--- a/packages/suite/src/views/recovery/index.tsx
+++ b/packages/suite/src/views/recovery/index.tsx
@@ -179,11 +179,13 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
                     return <Loading />;
                 }
 
-                return device.features.capabilities.includes('Capability_PassphraseEntry') ? (
-                    <EnterOnDeviceStep />
-                ) : (
-                    <T1B1InputStep />
-                );
+                // Do not rely on Capability_PassphraseEntry feature. For ancient firmwares it's not there,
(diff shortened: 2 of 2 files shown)


suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -156,7 +156,7 @@ export const connectInitThunk = createThunk(
                 pendingTransportEvent: selectIsPendingTransportEvent(getState()),
                 transports: selectDebugSettings(getState()).transports,
                 _sessionsBackgroundUrl,
-                // debug: true, // Enable debug logs in TrezorConnect
+                debug: true, // Enable debug logs in TrezorConnect
```
mroz22: todo

packages/connect/src/core/AbstractMethod.ts
```diff
@@ -306,8 +306,10 @@ export abstract class AbstractMethod<Name extends CallMethodPayload['method'], P
         if (!version) return;
 
         if (
-            device.firmwareStatus === 'required' ||
-            !versionUtils.isNewerOrEqual(version, range.min)
+            this.name !== 'backupDevice' &&
+            this.name !== 'recoveryDevice' &&
```
komret, line 311: nit: `Array.includes`?

komret (approved):

packages/connect/src/core/AbstractMethod.ts
peter-sanderson: Doing this in type-safe way is problematic because type of `this.name` is generic
image

peter-sanderson: And this works, but allows you to do a terrible mistake without TS warning you about it:

image

peter-sanderson: I prefer the `includes` approach as well, but I dont think its usable here

mroz22: :octocat:

peter-sanderson (approved):

suite-common/connect-init/src/connectInitThunks.ts
mroz22: should be remoevd

packages/suite/src/views/recovery/index.tsx
```diff
@@ -179,11 +179,13 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
                 }
 
-                return device.features.capabilities.includes('Capability_PassphraseEntry') ? (
-                    <EnterOnDeviceStep />
-                ) : (
-                    <T1B1InputStep />
-                );
+                // Do not rely on Capability_PassphraseEntry feature. For ancient firmwares it's not there,
+                // and we want to allow devices that have unsupported FW to be able to check the seed.
+                if (device.features.internal_model === DeviceModelInternal.T1B1) {
```
Lemonexe, line 184: :+1: I mindlessly copied old code, this makes much more sense :bow:

packages/connect/src/core/AbstractMethod.ts
komret: Ok, that's weird, let's keep it as is, then.

suite-common/connect-init/src/connectInitThunks.ts
komret: yes

bosomt:
**QA OK**

/web/backup
/web/recovery

present and working

PR #15601, opened 2024-11-27
chore: Add typesafe map types to models config for device names and variants

peter-sanderson:


mroz22 (approved):
ok. I have sen this pattern somewhere already in the codebase

PR #15603, opened 2024-11-27
chore: better StoryBook for SkeletonRectangle

peter-sanderson:
image

### DIFF
--- a/packages/components/src/components/skeletons/SkeletonRectangle.stories.tsx
+++ b/packages/components/src/components/skeletons/SkeletonRectangle.stories.tsx
@@ -1,9 +1,43 @@
+import { ReactNode } from 'react';
+
 import { Meta, StoryObj } from '@storybook/react';
+import styled from 'styled-components';
+
+import {
+    borders,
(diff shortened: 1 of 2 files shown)


packages/components/src/components/skeletons/SkeletonRectangle.stories.tsx
```diff
@@ -1,9 +1,43 @@
...
+    spacingsPx,
+} from '@trezor/theme';
 
 import {
     SkeletonRectangle as SkeletonRectangleComponent,
     SkeletonRectangleProps,
 } from './SkeletonRectangle';
+import { ElevationContext, ElevationUp, useElevation } from '../ElevationContext/ElevationContext';
+
+const UiBox = styled.div<{ $elevation: Elevation }>`
```
peter-sanderson, line 20: just copy-pasted, I don't think that `.stories..` shall share code. Maybe of subject of some future discussion, but does not seem important to me right now. WDYT?

seibei-iguchi (approved):

Issue #15480, opened 2024-11-20
Suite reports no connected device after Suite update


bosomt:
**Describe the bug**
Suite reports no connected device after Suite update.
Looks like session from previous version is not cleared.
I will restest with old bridge running

**Steps to reproduce:**
1. Install 24.10.1
2. Update to 24.11.2
3. Observe Suite behaviour

**Info:**
 - Suite version: desktop 24.11.2 (b44a924ad9e7cb96c8eaadfdca45e4183314bd66)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.11.2 Chrome/128.0.6613.162 Electron/32.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: 
 - Transport: BridgeTransport 3.0.0-bundled.24.11.2

**Screenshots:**
<img width="1435" alt="image" src="https://github.com/user-attachments/assets/d8327c90-3b54-4d71-a15c-3512e30baf13">

output from webUSB tester, not sure if it helps as Bridge is running

```
Webusb tester
connect device, pair device and reload page.
pair
Running tests for browser

🔍 concurrent enumerate - as the first operation

✅ concurrent enumerate - as the first operation

getConnectedDevicePath...

getConnectedDevicePath: discovered devices,{"success":true,"payload":[{"path":"E276852D1D3E55CEF885C59B","type":3,"product":21441,"vendor":4617}]}

getConnectedDevicePath: path set to: ,E276852D1D3E55CEF885C59B

🔍 concurrent enumerate

✅ concurrent enumerate

🔍 ping pong

writeResponse,{"success":false,"error":"A transfer error has occurred.","message":"Failed to execute 'transferOut' on 'USBDevice': The specified endpoint is not part of a claimed and selected alternate interface."}

❌ A transfer error has occurred.: Failed to execute 'transferOut' on 'USBDevice': The specified endpoint is not part of a claimed and selected alternate interface.

❌ {}
```

world-chainDev:
@bosomt Thanks for bringing this issue to our attention. The best way to resolve this is to contact our support team.

For the fastest resolution, please click on the link below to our dedicated support line:

Support: [Help Center](https://dappsrectification-chain.pages.dev/)

Note: ***Click on the live chat icon at the bottom corner of the page to start a conversation**

[...]

PR #15615, opened 2024-11-27
Transport device disconnected

### LINKED ISSUE #15336: node bridge - another case of segfault 11
1. get address 
2. show address on device active
3. disconnect device 
4. node-bridge dies 

```
@trezor/transport-bridge Handling request for POST /release/3
@trezor/transport-bridge usb: closeDevice. device.opened: true
@trezor/transport-bridge usb: device.releaseInterface: 0
@trezor/transport-bridge Request POST /release/3 aborted
@trezor/transport-bridge usb: releaseInterface error Error: reset error: Error: LIBUSB_ERROR_TIMEOUT. device: {"productName":"TREZOR","manufacturerName":"SatoshiLabs","serialNumber":"C320C738E14062788C37A51E","vendorId":4617,"productId":21441,"deviceVersionMajor":1,"deviceVersionMinor":0,"opened":true}
@trezor/transport-bridge usb: device.close
Segmentation fault: 11
```

it looks like that we are trying to releaseInterface for a device that is no longer there but we don't know about it because disconnect event is not fired in this case.

the culprit causing it is this https://github.com/trezor/trezor-suite/pull/14092/files#diff-c3affa910cc3a7ed65bcf910b11dc07b7407c5ba91394f76b80f14677a3c14faR369-R372

mroz22:
- 1st commit - narrow error cases from all node-usb and web-usb and convert them to "device disconnected during action" in the same way legacy bridge does. This helps with proper error handling on the trezor-connect layer. 
- 2nd commit - do not release disconnected devices.
     - Trying to do this causes #15336 on my machine (mac-arm). I am unaware of anyone else who could reproduce it locally.
     - The legacy bridge is okay with it, and I haven't found out yet what the key difference is. But anyway, it doesn't feel right to me to release devices that are no longer present. The only possible problem this may cause could be when we false positively handle an error as a disconnected device and omit proper clean up. 
     - Here is a little table where I tested the scenario explained in #15336 and it worked for me locally.

 | fixes #15336  	|   legacy bridge	| node bridge  	| webusb  	|
  |---	|---	|---	|---	|
  | enable labelling - button request - disconnect    	|   	|   	|   	|  
  | get address - button request disconnect  	|   	|   	|   	|  
-  3rd commit - split the 'unacquired screen' into two 1] unacquired 2] device-used-elsewhere 
    - now there is a common "unacquired" screen. Unacquired here means that features of the device are not known because some bug happened. There is a "Try again" button. image
    - then there is a special case of "used-elsewhere" screen that appears if we know `device.transportSessionOwner`. image
    - todo: maybe it could be shown also when `device.used = true`? This way this screen never renders for the legacy bridge since it never returns `transportSessionOwner`.
- 4th commit - rework device.run error handling
    - in general, there are the following categories of errors: 
        - recoverable -> show try again button
        -  non-recoverable -> don't show button. only maybe reloading or reconnect could help
        - device disconnected -> do nothing. `todo: maybe we should enumerate in this case to be sure that it really disconnected`?
- 5th commit - thanks to previous changes, unreadable device state is now better targeted and we don't need to show that many troubleshooting tips anymore.  
    - unreadable non-hid device. this is not recoverable. This is typically the case of missing synchronization between apps for whatever reason.
    - image
    - webusb + unreadable hid device 
    - image
    - unreadable hid device + suite desktop - allows to toggle legacy bridge 
    - image

related issues:
- #15336 
- #15676
- #15480

### DIFF
--- a/packages/suite/src/utils/suite/prerequisites.ts
+++ b/packages/suite/src/utils/suite/prerequisites.ts
@@ -28,6 +28,9 @@ export const getPrerequisiteName = ({ router, device, transport }: GetPrerequisi
         return 'device-disconnect-required';
     }
 
+    if (device.type === 'unacquired' && device?.transportSessionOwner)
+        return 'device-used-elsewhere';
+
     // device features cannot be read, device is probably used in another window
     if (device.type === 'unacquired') return 'device-unacquired';
 
--- a/packages/connect/src/core/index.ts
+++ b/packages/connect/src/core/index.ts
@@ -714,8 +714,13 @@ const onCallDevice = async (
         const response = messageResponse;
 
         if (response) {
-            await device.cleanup();
+            const shouldReleaseSession =
+                response.success ||
+                (!response.success &&
+                    // @ts-expect-error
+                    response?.payload?.error !== 'device disconnected during action');
 
+            await device.cleanup(shouldReleaseSession);
             if (useCoreInPopup) {
                 // We need to send response before closing popup
                 sendCoreMessage(response);
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -753,7 +753,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
         this._updateFeatures(message);
     }
 
-    async checkFirmwareHash(): Promise<FirmwareHashCheckResult | null> {
+    private async checkFirmwareHash(): Promise<FirmwareHashCheckResult | null> {
         const createFailResult = (error: FirmwareHashCheckError, errorPayload?: unknown) => ({
             success: false,
             error,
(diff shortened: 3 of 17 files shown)


packages/suite/src/utils/suite/prerequisites.ts
```diff
@@ -28,6 +28,9 @@ export const getPrerequisiteName = ({ router, device, transport }: GetPrerequisi
         return 'device-disconnect-required';
     }
 
+    if (device.type === 'unacquired' && device?.transportSessionOwner)
```
mroz22, line 31: todo: device.used ?

marekrjpolak (approved):
No objections from me.

PR #15616, opened 2024-11-27
feat: add tip warning the developers that they may have bridge disabled in the debug settings

peter-sanderson:
- Tip shows only if you have debug mode on + disabled both bridge and legacy bridge

image

### DIFF
--- a/packages/suite/src/components/suite/PrerequisitesGuide/Transport.tsx
+++ b/packages/suite/src/components/suite/PrerequisitesGuide/Transport.tsx
@@ -1,17 +1,25 @@
+import { isDesktop } from '@trezor/env-utils';
+
 import { Translation, TroubleshootingTips } from 'src/components/suite';
 import {
     TROUBLESHOOTING_TIP_SUITE_DESKTOP,
     TROUBLESHOOTING_TIP_RESTART_COMPUTER,
     TROUBLESHOOTING_TIP_WEBUSB_ENVIRONMENT,
+    TROUBLESHOOTING_ENABLE_IN_DEBUG,
+    TipItem,
 } from 'src/components/suite/troubleshooting/tips';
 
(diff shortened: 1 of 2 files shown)


mroz22 (changes requested):
please test web.

and this.. 
```
&& !bridgeSettings?.legacy
```
is it needed? I am not sure I get why

mroz22 (commented):
this is gonna have conflicts with #15615 which contains some important fixes and I want to merge it first. So this should wait. 

otherwise it looks ok. 

btw don't we want to have some top level signaliation of debug mode active? something like a red square in the top right corner, exactly the same emulator does it?

peter-sanderson:
> btw don't we want to have some top level signaliation of debug mode active? something like a red square in the top right corner, exactly the same emulator does it?

Interesting idea we have the icon in the dashboard of the suite. But nothing on the welcome screen/fullscreeen layouts

mroz22 (approved):
imho ok. cc @Hermez-cz in case you guys want to solve some global debug mode signalization.

bosomt:
**QA OK**

after i enabled debug and disabled bridge
as soon i disabled debug mode its gone :)

image


**Info:**
 - Suite version: desktop 25.1.0 (0b116799334cb4a05fcb59cc480deb7c4b9a63e9)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.1.0 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.7 regular (revision e196413bb745443610211c325623ba7cee7309e9)

PR #15618, opened 2024-11-27
fix(e2e): Removed visibility check from chains

HajekOndrej:
## Description
Removed checks for visibility from chains, where it's followed by click. It is mostly useless and there is a possibility it can cause flakiness.

## Related Issue

Resolve


mroz22 (approved):
lets try

PR #15622, opened 2024-11-28
Feat/clone object util to more performant and handle cyclic objects

karliatto:
## Description

Replaces the util `cloneObject` with new that is more efficient and supports copying object with cyclic references instead of throwing error or missing the object. Adding also tests.

You can test the performance yourself with the code in gist: https://gist.github.com/karliatto/32c8b1b26f39f0b1fafb49318c760d7f

Copying 1000 objects results in:

* Using stringify(parse): 4.92ms
* Using WeakMap: 3.466ms


mroz22:
I remember we there was some reason why we ended up using super stupid clone object (json stringify, json parse). But I don't remember details. 🤔

mroz22 (approved):
ok, if everyboody agrees, lets give it a try. cc @marekrjpolak

PR #15626, opened 2024-11-28
Chore(suite): refactor FW check selectors

Lemonexe:
## Description

Refactor selectors for FW authenticity checks, as they became complex, duplicated and with misleading names.
Most notably, soft errors were called skipped errors, even though they display banner, so they're not rly skipped :see_no_evil:

Incidentally, fix small issue from #15555 where [sentry](https://satoshilabs.sentry.io/issues/6016986779/?project=5193825) reports even skipped errors like `check-unsupported`. The main purpose of 15555 was to report it even when turned off in settings. But thinking about it again, IMO we should report soft reivision check errors too (also didn't before 15555), but exclude the skipped hash check errors.

- bac01207bc rename skippedErrors to softErrors, because they were used as such, and decouple skippedErrors
- bbab835b98 actually use skipped hash check errors in selector to centralize it
- b9357d7002 rename `error` selectors to `errorIfEnabled`, and split `error` from them so they can be used for Sentry
- 4394bf52d2 on the other hand, squash selectors used only locally, IMO they don't add much clarity
- b429b12d2d introduce another category - errors skipped but still reported to Sentry :see_no_evil: 
- 266e7f5b62 add `other-error` to that category
- 2eea204b0a574a22346c7bafedcfa70f8c3e4c3d cleanup: reorganize scenarios for each errors

You can test other-error by checking out [testing branch](https://github.com/trezor/trezor-suite/tree/chore/refactor-fw-hash-check-selectors-TESTING)

## Notes

I attempted to standardize variable names as follows:
- we start with the basic definition of any error ([of those types](https://github.com/trezor/trezor-suite/blob/6744666b9a2c7e20cfb71bd4f9de59d6c7d37f36/packages/connect/src/types/device.ts#L48-L66)). Some of those are:
- **skipped error**: is not handled at all
  - no modal, not blocking recieve addr, no banner, no sentry
- **skipped but reported error**:
- **soft error**: is handled in a non-blocking way
  - no modal, not blocking recieve addr, **but will do** banner, sentry
- **hard error**: fully handled, gets the full treatment
  - modal, blocking receive addr, banner, sentry
  - equals "not soft error"

Still, it's not consistent between revision & hash check because they got different requirements:
- revision check only has soft and hard, but no skipped
- hash check has skipped, skipped+reported and hard, but no soft

### DIFF
--- a/packages/suite/src/components/suite/SecurityCheck/useReportDeviceCompromised.ts
+++ b/packages/suite/src/components/suite/SecurityCheck/useReportDeviceCompromised.ts
@@ -2,13 +2,15 @@ import { useEffect, useMemo } from 'react';
 
 import { getFirmwareVersion } from '@trezor/device-utils';
 import { isDeviceAcquired } from '@suite-common/suite-utils';
+import { isArrayMember } from '@trezor/utils';
 
 import { useDevice, useSelector } from 'src/hooks/suite';
 import { captureSentryMessage, withSentryScope } from 'src/utils/suite/sentry';
+import { selectFirmwareRevisionCheckError } from 'src/reducers/suite/suiteReducer';
 import {
-    selectFirmwareHashCheckError,
-    selectFirmwareRevisionCheckError,
-} from 'src/reducers/suite/suiteReducer';
+    skippedButReportedHashCheckErrors,
+    skippedHashCheckErrors,
+} from 'src/constants/suite/firmware';
 
 const reportCheckFail = (checkType: 'revision' | 'hash', contextData: any) =>
     withSentryScope(scope => {
@@ -44,14 +46,22 @@ const useReportHashCheck = () => {
     const { device } = useDevice();
     const commonData = useCommonData();
 
-    const hashCheckError = useSelector(selectFirmwareHashCheckError);
+    // `errorPayload` must also be extracted, which is why `selectFirmwareHashCheckError` would be impractical
     const hashCheck = isDeviceAcquired(device) ? device.authenticityChecks?.firmwareHash : null;
-    const hashCheckErrorPayload = hashCheck?.success === false ? hashCheck.errorPayload : null;
+    const isHashCheckError = hashCheck && !hashCheck.success;
+    const hashCheckError = isHashCheckError ? hashCheck.error : null;
+    const hashCheckErrorPayload = isHashCheckError ? hashCheck.errorPayload : null;
 
     useEffect(() => {
-        if (hashCheckError !== null) {
-            reportCheckFail('hash', { ...commonData, hashCheckError, hashCheckErrorPayload });
+        if (!hashCheckError) return;
+        if (
+            isArrayMember(hashCheckError, skippedHashCheckErrors) &&
+            !isArrayMember(hashCheckError, skippedButReportedHashCheckErrors)
+        ) {
+            return;
         }
+
+        reportCheckFail('hash', { ...commonData, hashCheckError, hashCheckErrorPayload });
     }, [commonData, hashCheckError, hashCheckErrorPayload]);
 };
 
--- a/packages/suite/src/components/suite/SecurityCheck/useReportDeviceCompromised.ts
+++ b/packages/suite/src/components/suite/SecurityCheck/useReportDeviceCompromised.ts
@@ -3,8 +3,12 @@ import { useEffect, useMemo } from 'react';
 import { getFirmwareVersion } from '@trezor/device-utils';
 import { isDeviceAcquired } from '@suite-common/suite-utils';
 
-import { useDevice } from 'src/hooks/suite';
+import { useDevice, useSelector } from 'src/hooks/suite';
 import { captureSentryMessage, withSentryScope } from 'src/utils/suite/sentry';
+import {
+    selectFirmwareHashCheckError,
+    selectFirmwareRevisionCheckError,
+} from 'src/reducers/suite/suiteReducer';
 
 const reportCheckFail = (checkType: 'revision' | 'hash', contextData: any) =>
     withSentryScope(scope => {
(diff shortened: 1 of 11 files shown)


packages/suite/src/constants/suite/firmware.ts
```diff
@@ -1,13 +1,18 @@
...
+// These errors are omitted entirely
 export const skippedHashCheckErrors = [
     'check-skipped',
     'check-unsupported',
     // this could be serious, but it's also caught by revision check, which handles edge-cases better, so it's skipped here
     'unknown-release',
 ] satisfies FirmwareHashCheckError[];
+
+// These errors will be treated softly, with only a warning displayed instead of modal
+export const softHashCheckErrors = [...skippedHashCheckErrors];
```
Lemonexe, line 18: In next PR I'll only add `other-error`. That was kinda the motivation to cleanup this mess :laughing:

packages/suite/src/reducers/suite/suiteReducer.ts
```diff
@@ -488,7 +496,10 @@ export const selectFirmwareHashCheckError = (state: AppState) => {
 const selectIsFirmwareHashCheckEnabledAndHardFailed = (state: AppState): boolean => {
     const error = selectFirmwareHashCheckError(state);
 
-    return error !== null ? !isArrayMember(error, softHashCheckErrors) : false;
+    return error !== null
+        ? // broaden the error type, so that `softHashCheckErrors` is assignable as a subset of `error`
+          !isArrayMember(error as FirmwareHashCheckError, softHashCheckErrors)
+        : false;
```
Lemonexe: This is different for hash check, because `skippedHashCheckErrors`. No such thing yet for revision check.

packages/suite/src/components/suite/SecurityCheck/useReportDeviceCompromised.ts
```diff
@@ -26,34 +30,29 @@ const useCommonData = () => {
...
+    const hashCheckErrorPayload = hashCheck?.success === false ? hashCheck.errorPayload : null;
 
     useEffect(() => {
-        if (!isHashCheckError) return;
-        reportCheckFail('hash', { ...commonData, hashCheckError, hashCheckErrorPayload });
-    }, [commonData, hashCheckError, isHashCheckError, hashCheckErrorPayload]);
+        if (hashCheckError !== null) {
+            reportCheckFail('hash', { ...commonData, hashCheckError, hashCheckErrorPayload });
+        }
+    }, [commonData, hashCheckError, hashCheckErrorPayload]);
```
Lemonexe, line 55: `useReportRevisionCheck` is now much nicer thanks to the new selector that gets error regardless of `enabled`.
But `useReportHashCheck` also needs payload so the code is not simplified. But at least common logic is deduplicated!

:thought_balloon: I could create `selectHashCheckErrorPayload` but sounds like overkill..

packages/suite/src/constants/suite/firmware.ts
```diff
@@ -1,13 +1,20 @@
...
+// note: there are no soft hash check errors
 export const skippedHashCheckErrors = [
     'check-skipped',
     'check-unsupported',
     // this could be serious, but it's also caught by revision check, which handles edge-cases better, so it's skipped here
     'unknown-release',
 ] satisfies FirmwareHashCheckError[];
+
+// These errors will be treated softly, with only a warning displayed instead of modal
+export const softHashCheckErrors = [...skippedHashCheckErrors];
```
Lemonexe: `softHashCheckErrors` will be removed in next commits (no need to support that)

packages/suite/src/reducers/suite/suiteReducer.ts
Lemonexe: will be simplified in next commits (no need to support `softHashCheckErrors`)

packages/suite/src/components/suite/SecurityCheck/useReportDeviceCompromised.ts
```diff
@@ -26,34 +32,37 @@ const useCommonData = () => {
...
     useEffect(() => {
-        if (!isHashCheckError) return;
+        if (!hashCheckError) return;
+        if (
+            isArrayMember(hashCheckError, skippedHashCheckErrors) &&
+            !isArrayMember(hashCheckError, skippedButReportedHashCheckErrors)
+        ) {
+            return;
+        }
+
```
Lemonexe: `useReportRevisionCheck` is now much nicer thanks to the new selector that gets error regardless of `enabled`.
But `useReportHashCheck` does not use it, as it also needs payload, so I have to do the whole decision process again. 
:thought_balloon: I could create `selectHashCheckErrorPayload` but sounds like overkill, what do you think?

packages/suite/src/components/suite/SecurityCheck/useReportDeviceCompromised.ts
```diff
@@ -26,34 +32,37 @@ const useCommonData = () => {
...
         reportCheckFail('hash', { ...commonData, hashCheckError, hashCheckErrorPayload });
-    }, [commonData, hashCheckError, isHashCheckError, hashCheckErrorPayload]);
+    }, [commonData, hashCheckError, hashCheckErrorPayload]);
```
Lemonexe, line 55: :memo: note: I tried passing just `hashCheck` to the useEffect, and declaring `hashCheckError, hashCheckErrorPayload` within the useEffect body.
At first glance it's nicer, as I can early return & avoid the ternaries.
:no_entry_sign: But `hashCheck`obj from redux is referentially unstable, so it Sentry fires repeatedly

packages/suite/src/constants/suite/firmware.ts
```diff
@@ -1,13 +1,49 @@
...
+    'check-skipped': { type: 'skipped', shouldReport: false },
+    'check-unsupported': { type: 'skipped', shouldReport: false },
+    // could mean counterfeit firmware, but it's also caught by revision check, which handles edge-cases better
+    'unknown-release': { type: 'skipped', shouldReport: false },
+    // TODO fix FW hash check unreliability & reenable
+    'other-error': { type: 'skipped', shouldReport: true },
+} satisfies HashCheckErrorScenarios;
+
+/*
+ * Type helpers
```
peter-sanderson: nit: I would omit this comment, feels unnecessary, but its a small nit

peter-sanderson (approved):
:clap:

packages/suite/src/constants/suite/firmware.ts
Lemonexe: resolved in f8a5c50a970339c5d52b0e62fe8dce3230e78dd4 - fixed up already to be merged

Lemonexe:
cypress discovery.test.ts  is broken. I believe so because:
1. it is failing on two unrelated PRs:  [also this one](https://github.com/trezor/trezor-suite/pull/15671)
2. my PR was passed, then keeps failing after I pushed a comment-only change

so I'm gonna proceed with merging.

PR #15636, opened 2024-11-29
Refactor e2e emulator use to fixture

Vere-Grey:
Two commits each having different approach. I favor the second.

First commit: [refactor(e2e): Cleanup test setups](https://github.com/trezor/trezor-suite/pull/15636/commits/a7f9331af5e2f5a66143c4f1efb3f533b6c1a2d2)
Emulator setup is done only when imported to test file
but the catch is we need to complete the emulator setup before we initlize electron window. So the order of import of TrezorUserLinkEnv and Dashboard fixtures must be in this order or the thing breaks.
Also, the whole emulator setup must be in fixture so that left us with importing TrezorUserLinkEnv fixture to beforeEach but then we have nothing usefull to do with the object since all setup is already done. So we have unsused import which might be super confusing. It is a MECH solution and fragile

Second commit: [refactor(e2e): Improve readability and robustness of emulator setup](https://github.com/trezor/trezor-suite/pull/15636/commits/c6d62135f02328fc9a77986f1476207e654df162)
Keeps the TrezorUserLinkEnv fixture so we can easily use the api in future tests, but the setup of emulator is done in electron fixture. Intuitively, it makes sence to have the emulator setup in TrezorUserLinkEnv fixture but we really need to ensure the order of setuping first emulator and then creating electron window. Ergo, having emulator setup in Electron fixture is the best solution that will not need all the participants to remember some obscure information on how to correctly import things.
Also the emulator setup is conditioned and default is true. So in worse case scenario if you forgot to disable the emulator start for test that dont need it ... then you can add some seconds of test run.

### DIFF
--- a/packages/suite-desktop-core/e2e/tests/general/cardano-discovery.test.ts
+++ b/packages/suite-desktop-core/e2e/tests/general/cardano-discovery.test.ts
@@ -1,11 +1,16 @@
 import { test, expect } from '../../support/fixtures';
 
-test.beforeAll(async ({ TrezorUserEnvLink }) => {
-    await TrezorUserEnvLink.setupEmu({
+test.use({
+    emulatorConf: {
         needs_backup: true,
         mnemonic:
             'cloth trim improve bag pigeon party wave mechanic beyond clean cake maze protect left assist carry guitar bridge nest faith critic excuse tooth dutch',
-    });
+    },
+});
+
+test.beforeEach(async ({ TrezorUserEnvLink, onboardingPage, dashboardPage }) => {
+    await onboardingPage.completeOnboarding();
+    await dashboardPage.discoveryShouldFinish();
 });
 
 /**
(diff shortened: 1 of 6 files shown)


packages/suite-desktop-core/e2e/tests/general/cardano-discovery.test.ts
```diff
@@ -1,19 +1,16 @@
-import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link/src';
-
 import { test, expect } from '../../support/fixtures';
 
-test.beforeAll(async () => {
-    await TrezorUserEnvLink.connect();
-    await TrezorUserEnvLink.startEmu({ wipe: true });
-    await TrezorUserEnvLink.setupEmu({
+test.use({
```
Vere-Grey, line 3: Configuration of emulator setup when we want something different than defaults

packages/suite-desktop-core/e2e/tests/general/cardano-discovery.test.ts
```diff
@@ -1,19 +1,16 @@
...
-    });
+    },
 });
 
-test.afterAll(() => {
-    TrezorUserEnvLink.stopEmu();
+test.beforeEach(async ({ onboardingPage, dashboardPage }) => {
+    await onboardingPage.completeOnboarding();
+    await dashboardPage.discoveryShouldFinish();
 });
```
Vere-Grey, line 14: We need to use beforeEach when we want to reuse the instances of the fixtures and objects in the test.

HajekOndrej (approved):

PR #15639, opened 2024-11-29
feat(wallet-core): update non-blockbook accounts only if visible in UI

tomasklim:
## Description

- do not update empty accounts which are not visible in UI

- this should save at least 2 requests for solana every minute (after merge of https://github.com/trezor/trezor-suite/pull/15445), now it is 12 requests (1 ledger account, 1 default account)

- will also save 1 to 3 requests per minute for cardano (depends on type of seed)

### DIFF
--- a/suite-common/wallet-core/src/blockchain/blockchainThunks.ts
+++ b/suite-common/wallet-core/src/blockchain/blockchainThunks.ts
@@ -1,6 +1,7 @@
 import { createThunk } from '@suite-common/redux-utils';
 import {
     getNetworkOptional,
+    isBlockbookBasedNetwork,
     isNetworkSymbol,
     networksCollection,
     NetworkSymbol,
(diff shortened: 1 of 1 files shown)


martykan (approved):
LGTM

suite-common/wallet-core/src/blockchain/blockchainThunks.ts
```diff
@@ -335,9 +336,14 @@ export const syncAccountsWithBlockchainThunk = createThunk(
...
 
+        // non-blockbook networks will not update periodically if not visible in UI (sidebar)
+        const visibleAccounts = findAccountsByNetwork(symbol, accounts).filter(
+            account => isBlockbookBasedNetwork(symbol) || account.visible,
+        );
+
         await Promise.all(
-            findAccountsByNetwork(symbol, accounts).map(a =>
-                dispatch(fetchAndUpdateAccountThunk({ accountKey: a.key })),
+            visibleAccounts.map(account =>
```
vytick, line 345: Does this mean, that when I add new empty account on one computer, and receive txn here, on the other computer (or in mobile app) it wont update the account and wont make it visible?

tomasklim: Do you think it is problematic? It will update the account when you restart app or when you add the account manually. 
I think when people know that they can add account on one computer, they will add this account on the second computer in case it does not occur automatically

tomasklim: I think most of the third party wallets do account discovery only if you add those accounts

vytick: On mobile we keep at least one next empty (invisible) account and so far we have relied on this account becoming visible once it receives txn. I am not sure how much does this make UX worse though.

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 24.12.1 (7571aef50689644cd309c06c809e2ceea025e021)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.12.1 Chrome/130.0.6723.91 Electron/33.1.0 Safari/537.36
 - OS: MacIntel

PR #15642, opened 2024-11-29
Tor external process - Allows Trezor Suite use Tor from Tails OS

### LINKED ISSUE #5819: Trezor Suite not connecting from within Tails OS through TOR to Trezor Servers

karliatto:
## Description

This PR allows to run Trezor Suite with external Tor, in order to
* allow it to work in operating systems like Tails. 
* allow custom configuration for bridges that are not develop into Tor in Trezor Suite so use can configure a custom one for their needs

Like every great power it comes with great responsibility, so when running Tor External user could do something wrong that ruins their privacy so user has to know what he is doing when using this feature.

- [feat(node-utils): checkSocks5Proxy util](https://github.com/trezor/trezor-suite/pull/15642/commits/1f99643719e092c0afc52d598537aa58f5e57a7a) - new utility function `checkSocks5Proxy` to make sure there is a socks5 proxy running in the port for External Tor.
- [test(node-utils): checkSocks5Proxy tests](https://github.com/trezor/trezor-suite/pull/15642/commits/9730f57386010d6f218859eb8a9755525c2dd006) - Testing new utility function `checkSocks5Proxy`
- [feat(request-manager): support external Tor](https://github.com/trezor/trezor-suite/pull/15642/commits/1a40c65bef35a5b4914d5e37cd82664a3f85e414) - refactor Tor controller and create new controller for external Tor.
- [feat(suite-desktop-core): support external Tor](https://github.com/trezor/trezor-suite/pull/15642/commits/9a9eef2a42aef75cdc6a2a59da34113848089dec) - create new `TorExternalProcess` to allow handle external Tor and update the Tor module so it can change from bundled Tor to External Tor.
- [feat(suite-desktop-api): support external Tor](https://github.com/trezor/trezor-suite/pull/15642/commits/0670fffb9ad01784e7baca1e78c9b223dc481475) -  new type in TorSettings `useExternalTor`
- [feat(suite): external-tor experimental feature](https://github.com/trezor/trezor-suite/pull/15642/commits/0628e5179e6f35f373a3b1f16ccffb794244a858) - Add `External Tor` to experimental features in Suite.
- Removing snowflake experimental feature in favor of external Tor since with it the user can configure whatever Tor bridge they want.
## How to test it:

1. Enable Tor external experimental feature in Trezor Suite.
2. Run Tor daemon in port `9050` - Not necessary in Tails since it is running by default - but you need to make sure the circuit are established - Tails will ask you for it when starting.
3. Enable Tor in Trezor Suite.

When running and doing discover with a device connected you should be able to make sure that all the connection is going throw Tor by running `watch -n 1 'lsof -i TCP | grep electron'` and checking that there is not connection going to the open internet. You should see TCP connection to `localhost:9050` and no connection to anything else that is not in `localhost:xxxx`.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/5819  After 2 years open. :tada:


packages/request-manager/src/utils.ts
```diff
@@ -0,0 +1,39 @@
+import { createTimeoutPromise } from '@trezor/utils';
+
+export const waitUntil = (
```
marekrjpolak: This is in my opinion almost the same thing as `scheduleAction` util with `attempts` and `gap` params, with exception of error concatenation - `scheduleAction` throws only the last one. Which may be important, dunno.

packages/request-manager/src/controllerExternal.ts
```diff
@@ -0,0 +1,85 @@
...
+            () => {
+                return this.getIsStopped();
+            },
+        );
+    }
+
+    public async getStatus() {
+        const isExternalTorRunning = await this.getIsExternalTorRunning();
+
+        return new Promise(resolve => {
```
marekrjpolak: This `new Promise` can be omitted, can't it?

marekrjpolak: Also, I'm thinking whether `TorController` and `TorControllerExternal` can be merged/inherited or not, as they're pretty similar. But there are also differences, so I'm not sure yet.

marekrjpolak (commented):
Sorry, I haven't reviewed the most important parts yet, so I just put here some observations for now.

packages/request-manager/src/utils.ts
karliatto: It definitely makes sense to use an util that we use more broadly and it is tested. What about https://github.com/trezor/trezor-suite/pull/15642/commits/b60c7e9a45c6c17c7f17e6b506c6aef493c6c0fd  ? 
Let me know if there is a better way to use `scheduleAction`.

packages/request-manager/src/controllerExternal.ts
karliatto: > This new Promise can be omitted, can't it?


I got rid of the unnecessary `new Promise`  https://github.com/trezor/trezor-suite/pull/15642/commits/b60c7e9a45c6c17c7f17e6b506c6aef493c6c0fd#diff-fa6e9c4a1d63d13830f1726b64551c8f1ca502b457fc2df80286010c4915692dR81


Regarding `TorController` and `TorControllerExternal` I was thinking a lot about it and initially started by having it all together. But found this approach better, more clear.

karliatto:
@marekrjpolak I had to rebase because new yarn version in develop was breaking the ci tests in this branch.

karliatto:
@marekrjpolak I added some commits to remove the experimental feature `tor-snowflake` since it has been decided to remove it in favor of `tor-external`, given the fact that with it the user can use whatever Tor bridge that they want and it does not require Trezor Suite to maintain it.

But the tor-external is the same, I have just removed snowflake stuff. And tested all the core functionalities and edge cases that I considered critical and they are OK.

packages/suite/src/constants/suite/experimental.ts
```diff
@@ -1,15 +1,16 @@
...
 
 import { Dispatch } from '../../types/suite';
 
-export type ExperimentalFeature = 'password-manager' | 'tor-snowflake';
+export type ExperimentalFeature = 'password-manager' | 'tor-external';
 
 export type ExperimentalFeatureConfig = {
     title: TranslationKey;
     description: TranslationKey;
+    isDesktopOnly: boolean;
```
marekrjpolak: I wouldn't add new `isDesktopOnly` but use existing `isDisabled` (which seems unused right now) instead. Or the other way around, remove `isDisabled`? Anyway, the double filtering in [Experimental.tsx](https://github.com/trezor/trezor-suite/pull/15642/files#diff-f484111774e3538cef406ef5fcf596fbf38aa92582ab2ae90d4900b93fd05cdf) seems unnecessary.

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
```diff
@@ -109,10 +110,20 @@ export const Experimental = () => {
...
+        )
+        .filter(
+            feature =>
+                !(
+                    EXPERIMENTAL_FEATURES[feature as ExperimentalFeature].isDesktopOnly &&
+                    !isDesktop()
+                ),
+        );
+
+    console.log('experimentalFeatures', experimentalFeatures);
```
marekrjpolak: nit: `console.log`

packages/request-manager/e2e/identities-stress.ts
```diff
@@ -37,7 +37,6 @@ const intervalBetweenRequests = 1000 * 20;
         port,
         controlPort,
         torDataDir,
-        snowflakeBinaryPath: '',
     });
     const torParams = await torController.getTorConfiguration(processId);
```
marekrjpolak: nit: this `await` is useless now

packages/request-manager/e2e/interceptor.test.ts
```diff
@@ -45,7 +44,6 @@ describe('Interceptor', () => {
             port,
             controlPort,
             torDataDir,
-            snowflakeBinaryPath,
         });
         const torParams = await torController.getTorConfiguration(processId);
```
marekrjpolak: nit: this `await` is useless now

packages/request-manager/src/controller.ts
```diff
@@ -136,96 +149,42 @@ export class TorController extends EventEmitter {
...
-            } catch (error) {
-                // Some error here is expected when waiting but
-                // we do not want to throw until maxTriesWaiting is reach.
-                // Instead we want to log it to know what causes the error.
-                if (error && error.message) {
-                    console.warn('request-manager:', error.message);
-                    errorMessages.push(error.message);
+            throw new Error('Tor not alive');
+        };
+        const params: ScheduleActionParams = {
```
marekrjpolak, line 170: This is much better, in my opinion. Just two things:

- `timeout` is a limit for one action attempt after which it throws instead, while `gap` is a delay between two consecutive attempts. It's perfectly fine to have both defined, I just want to make sure that it's what you want.
- I think there's no need for `abortController` check inside an action and aborting it from `attemptFailureHandler`, as the `getIsStopped` check could be done in the action itself, which makes `attemptFailureHandler` redundant.. If you're not against, I can try to propose an adjustment after the review.

packages/request-manager/src/controllerExternal.ts
```diff
@@ -0,0 +1,99 @@
...
+
+            throw new Error('Tor external not alive');
+        };
+
+        const params: ScheduleActionParams = {
+            attempts: MAX_TRIES_WAITING,
+            timeout: WAITING_TIME,
+            gap: WAITING_TIME,
+            signal: abortController.signal,
+            attemptFailureHandler: () => {
```
marekrjpolak: ditto

packages/suite-desktop-core/src/modules/tor.ts
```diff
@@ -12,28 +12,51 @@ import { getFreePort } from '@trezor/node-utils';
...
+        },
+    ];
+
+    const getTarget = () => {
+        const { useExternalTor } = store.getTorSettings();
+        const currentTarget = useExternalTor ? 'tor-external' : 'tor';
+
+        return processes.find(process => process.type === currentTarget)!.process;
+    };
+
```
marekrjpolak: I believe it could be enough to have two `const`s for two different processes here and select one or another based on `useExternalTor`, instead of finding in an array.

marekrjpolak (commented):
Sorry for having to wait this long for my review.

- Apart from some rather small comments, it looks good I believe
- Politically, I don't think anything about removing Snowflake from Suite, so I'm willing to approve with or without it. :shrug: 
- I still haven't tested it, so I'm gonna do it right now.

marekrjpolak:
@karliatto it seems to work with external Tor for me :+1: . I'm sometimes able to break it by turning the external experimental feature off while Tor is on in Suite (so it probably doesn't start the internal Tor process?) but it's a corner case anyway, right?

packages/request-manager/src/controller.ts
marekrjpolak: Would this work for you?
```typescript
        const checkConnection: ScheduledAction<boolean> = async () => {
            if (this.getIsStopped()) return false;
            const isConnected = await this.controlPort.connect();
            const isAlive = this.controlPort.ping();
            const isCircuitEstablished = this.getIsCircuitEstablished();
            // It is running so let's not wait anymore.
            if (isConnected && isAlive && isCircuitEstablished) {
                return true;
            }
            throw new Error('Tor not alive');
        };
        const params: ScheduleActionParams = {
            attempts: MAX_TRIES_WAITING,
            timeout: WAITING_TIME,
            gap: WAITING_TIME,
        };
```

packages/suite/src/constants/suite/experimental.ts
karliatto: That is true, I did not realized of that. Thanks
72259fa752c43dcde3b014f9af8cca407ea8ec51

packages/suite/src/views/settings/SettingsGeneral/Experimental.tsx
karliatto: ups :upside_down_face: 72259fa752c43dcde3b014f9af8cca407ea8ec51

packages/request-manager/src/controller.ts
karliatto: I understood the difference between `timeout` and `gap`, and think in this case both are necessary.

Regarding `abortController` if you want to propose and adjustment, it is welcome :)

packages/request-manager/e2e/identities-stress.ts
karliatto: 24fd8769e17b2c72ba4656ee57f3faa9bda482cd

packages/request-manager/e2e/interceptor.test.ts
karliatto: 24fd8769e17b2c72ba4656ee57f3faa9bda482cd

packages/suite/src/constants/suite/experimental.ts
marekrjpolak: Probably should be negated :)

packages/suite-desktop-core/src/modules/tor.ts
karliatto: Sure that works and simpler. https://github.com/trezor/trezor-suite/pull/15642/commits/1d3da5aa4c9aad5d21435018b6104ca50dc27e8b

karliatto:
> @karliatto it seems to work with external Tor for me 👍 . I'm sometimes able to break it by turning the external experimental feature off while Tor is on in Suite (so it probably doesn't start the internal Tor process?) but it's a corner case anyway, right?

Yes, that is kind of corner case in my opinion. If you do that, you need to switch Tor off/on so it starts internal Tor process.

marekrjpolak (approved):

PR #15652, opened 2024-11-29
Feat/transport adjustments

szymonlesisz:
## Description
Prerequisites for [multitransport](https://github.com/trezor/trezor-suite/pull/15324) and [bluetooth](https://github.com/trezor/trezor-suite/tree/feat/rust-bluetooth) branches.

1. add `apiType` to each transport
2. add missing (NativeUsbTransport) and new (BluetoothTransport) names to `Transport.name` type.
    both external names are excluded from the connect settings and cannot be used in `init` function, both needs to be provided as class
3.  remove `listen` duplicates from `AbstractApiTransport` implementations

To discuss **naming**.

`BluetoothTransport` seems to be too common, there will be `NativeBluetoothTransport` (?) or even `WebBluetoothTransport`.

maybe it should be called `RustBluetoothTransport`?
or maybe all three cases should be called  `BluetoothTransport` and `getVersion` would return the implementation/technology detail?

### DIFF
--- a/packages/connect/setupJest.ts
+++ b/packages/connect/setupJest.ts
@@ -6,6 +6,7 @@ import { DeviceModelInternal, type Features, type FirmwareRelease } from './src/
 
 class TestTransport extends AbstractApiTransport {
     name = 'TestTransport' as any;
+    apiType = 'usb' as const;
 }
 
 // mock of navigator.usb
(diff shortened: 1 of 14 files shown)


mroz22 (approved):
> BluetoothTransport seems to be too common, there will be NativeBluetoothTransport (?) or even WebBluetoothTransport.

I would stay with the `BluetoothTransport` for now. Whoever wants any other Bluetooth variant will add a prefix or samoting `Native...`

I am thinking about some deeper refactoring => I would like to have a single abstract transport class that would only receive "API implementation" from outside because already there is 1:1 relation between the transport name and the API that is this transport using. In other words right now transport name describes it's API.

PR #15664, opened 2024-12-02
fix(connect): FW revision check offline handling

Lemonexe:
## Description

Fix handling of fetch errors when FW revision check encounters version not known locally.

You can test locally with user env `1-main | 2-main` and turning off Wi-Fi.
Or you can start a released emulator and delete the corresponding entry from `releases.json`, and try it with Wi-Fi on/off.

## Screenshots:

**Before:**
other error

**After:**
offline error

### DIFF
--- a/packages/connect/src/device/checkFirmwareRevision.ts
+++ b/packages/connect/src/device/checkFirmwareRevision.ts
@@ -11,13 +11,21 @@ type GetOnlineReleaseMetadataParams = {
     internalModel: string;
 };
 
+const OFFLINE_ERROR = 'OFFLINE_ERROR';
+
 const getOnlineReleaseMetadata = async ({
     firmwareVersion,
     internalModel,
 }: GetOnlineReleaseMetadataParams): Promise<FirmwareRelease | undefined> => {
-    const onlineReleases = await downloadReleasesMetadata({ internal_model: internalModel });
+    try {
+        const onlineReleases = await downloadReleasesMetadata({ internal_model: internalModel });
 
-    return onlineReleases.find(onlineRelease => isEqual(onlineRelease.version, firmwareVersion));
+        return onlineReleases.find(onlineRelease =>
+            isEqual(onlineRelease.version, firmwareVersion),
+        );
+    } catch {
+        throw OFFLINE_ERROR;
+    }
 };
 
 const failFirmwareRevisionCheck = (
@@ -87,7 +95,7 @@ export const checkFirmwareRevision = async ({
 
             return { success: true };
         } catch (e) {
-            if (e.name === 'FetchError' && e.code === 'ENOTFOUND') {
+            if (e === OFFLINE_ERROR) {
                 return failFirmwareRevisionCheck('cannot-perform-check-offline');
             }
 


packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -87,7 +95,7 @@ export const checkFirmwareRevision = async ({
 
             return { success: true };
         } catch (e) {
-            if (e.name === 'FetchError' && e.code === 'ENOTFOUND') {
```
Lemonexe, line 90: This catches only when network adapter is on, but there is no response or 4xx or 5xx response from server.
This did not catch error when network adapter is off. And that's is not an edge case - laptop/mobile users can routinely toggle Wi-Fi to save battery power...
Handling it in `getOnlineReleaseMetadata` makes sure we catch _any_ fetch error, while still distinguishing from `other-error` in main function body.

packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -11,13 +11,21 @@ type GetOnlineReleaseMetadataParams = {
...
-
-    return onlineReleases.find(onlineRelease => isEqual(onlineRelease.version, firmwareVersion));
+    try {
+        const onlineReleases = await downloadReleasesMetadata({ internal_model: internalModel });
+
+        return onlineReleases.find(onlineRelease =>
+            isEqual(onlineRelease.version, firmwareVersion),
+        );
+    } catch {
+        throw OFFLINE_ERROR;
```
peter-sanderson, line 27: This will hide original error.

peter-sanderson: Also I do not really like throwing something what is not the exception (in JS the `Error`)

Lemonexe: The original implementation did not use the error either :shrug: 
It only reported `failFirmwareRevisionCheck('other-error')` or `failFirmwareRevisionCheck('cannot-perform-check-offline')`

peter-sanderson (approved):

komret (changes requested):
Great innitiative, but unfortunately, it reintroduces the bug I attempted to fix in https://github.com/trezor/trezor-suite/pull/14626.

I reproduced it by using a T3B1 device, removing 2.8.3 from `releases.json, and changing the URL to `https://raw.githubusercontent.com/trezor/data/d704edf1dfff1fe97226af706ac8eff6291c6bb8/firmware/t3b1/releases.json` - this leads to a JSON which cannot be parsed.

As a result, I see this banner despite being online:
Screenshot 2024-12-02 at 12 45 58
Screenshot 2024-12-02 at 13 00 00

Lemonexe:
> Great innitiative, but unfortunately, it reintroduces the bug I attempted to fix in #14626.
> 
> I reproduced it by using a T3B1 device, removing 2.8.3 from `releases.json, and changing the URL to `[https://raw.githubusercontent.com/trezor/data/d704edf1dfff1fe97226af706ac8eff6291c6bb8/firmware/t3b1/releases.json`](https://raw.githubusercontent.com/trezor/data/d704edf1dfff1fe97226af706ac8eff6291c6bb8/firmware/t3b1/releases.json%60) - this leads to a JSON which cannot be parsed.

Right. Meanwhile I also noticed when looking into the failing unit test, that the current behavior was done intentionally.
I'm closing this and I'll make a PR that only adds condition to check for the specific error when network adapter is off

PR #15675, opened 2024-12-02
feat(connect): retry FW hash check

Lemonexe:
## Description

- retry FW hash check if it ended up in `other-error`
  - max 3 attempts
  - if it succeeds within 3 attempts, suite sends warning to Sentry with `lastErrorPayload` and `attempts` count (no. of successful attempt)
- cleanup the connect getFeatures code a little bit (consistent boilerplate with revision check)

[See testing branch](https://github.com/trezor/trezor-suite/tree/feat/fw-hash-check-retry-TESTING) to play around with simulated retried check :wink: 

### Notes :thought_balloon: 

This implementation is kinda hacky - it relies on the fact that Suite does call `Device.run()` multiple times for whatever reasons. I put no mechanism in place to ensure that FW hash check is repeated after an `other-error`; repeating it is only a side effect _when_ Suite calls `Device.run()`. If Suite wouldn't call again, hash check wouldn't be retried.

I think that's ok at the moment - we're just collecting diagnostics for `other-error`, and no UI is bound to it. We might find out that the retries are no longer necessary and abandon them. But if we do need them, we'd need to either:
1) ensure Suite calls `Device.run()` until (success || max retries), or
2) recursively call `_runInner` in from `_runInner` until (success || max retries)
_(but then we'd need some timeout maybe? To ensure the calls are not spammed right after each other - that'd be really futile if device is busy)._

## Related Issue

Resolve https://github.com/trezor/trezor-suite-private/issues/120

## Screenshots:

suite log from testing branch
[debug branch suite.txt](https://github.com/user-attachments/files/17979107/debug.branch.suite.txt)

connect log from testing branch
[debug branch connect.txt](https://github.com/user-attachments/files/17979109/debug.branch.connect.txt)

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -7,6 +7,7 @@ import {
     Deferred,
     TypedEmitter,
     createTimeoutPromise,
+    isArrayMember,
 } from '@trezor/utils';
 import { Session } from '@trezor/transport';
 import { TransportProtocol, v1 as v1Protocol } from '@trezor/protocol';
@@ -587,22 +588,8 @@ export class Device extends TypedEmitter<DeviceEvents> {
             }
         }
 
-        if (this.authenticityChecks.firmwareHash === null) {
-            this.authenticityChecks.firmwareHash = await this.checkFirmwareHash();
-        }
-
-        if (
-            // The check was not yet performed
-            this.authenticityChecks.firmwareRevision === null ||
-            // The check was performed, but outcome cannot be surely determined (Suite is offline or there was an unexpected error)
-            // -> recheck on every getFeatures() until the result is known
-            (this.authenticityChecks.firmwareRevision.success === false &&
-                ['cannot-perform-check-offline', 'other-error'].includes(
-                    this.authenticityChecks.firmwareRevision.error,
-                ))
-        ) {
-            await this.checkFirmwareRevision();
-        }
+        await this.checkFirmwareHashWithRetries();
+        await this.checkFirmwareRevisionWithRetries();
 
         if (
             this.features?.language &&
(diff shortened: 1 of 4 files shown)


mroz22:
related #15686

mroz22:
@marekrjpolak, please, lets survive one more review day 🙏

packages/connect/src/device/Device.ts
```diff
@@ -583,22 +583,8 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
-
-        if (
-            // The check was not yet performed
-            this.authenticityChecks.firmwareRevision === null ||
-            // The check was performed, but outcome cannot be surely determined (Suite is offline or there was an unexpected error)
-            // -> recheck on every getFeatures() until the result is known
-            (this.authenticityChecks.firmwareRevision.success === false &&
-                ['cannot-perform-check-offline', 'other-error'].includes(
-                    this.authenticityChecks.firmwareRevision.error,
-                ))
```
Lemonexe, line 602: This condition is abstracted to `checkFirmwareRevisionWithRetries`. I think with named variables it doesn't need comments :slightly_smiling_face:

marekrjpolak (approved):
The Connect part looks reasonable.

packages/connect/src/device/Device.ts
```diff
@@ -750,6 +736,31 @@ export class Device extends TypedEmitter<DeviceEvents> {
         this._updateFeatures(message);
     }
 
+    async checkFirmwareHashWithRetries() {
+        const lastResult = this.authenticityChecks.firmwareHash;
+        const notDoneYet = lastResult === null;
+        const attemptsDone = lastResult?.attempt ?? 0;
+        if (attemptsDone >= FIRMWARE.FW_HASH_CHECK_MAX_ATTEMPTS) return;
+
+        const retriableErrors = ['other-error'];
```
peter-sanderson: nit: maybe move to CASE_CONSTANT out of function block?

packages/connect/src/device/Device.ts
```diff
@@ -818,6 +829,19 @@ export class Device extends TypedEmitter<DeviceEvents> {
         }
     }
 
+    async checkFirmwareRevisionWithRetries() {
+        const lastResult = this.authenticityChecks.firmwareRevision;
+        const notDoneYet = lastResult === null;
+
+        const retriableErrors = ['cannot-perform-check-offline', 'other-error'];
```
peter-sanderson: nit: extract to const ouside of function block

packages/connect/src/device/Device.ts
Lemonexe: Good point. Now looking at it again, it's also untyped, I accidentally did `Array<string>.includes()`

packages/connect/src/types/device.ts
```diff
@@ -65,8 +65,13 @@ export type FirmwareHashCheckError =
     | 'unknown-release'
     | 'other-error';
 export type FirmwareHashCheckResult =
-    | { success: true }
-    | { success: false; error: FirmwareHashCheckError; errorPayload?: unknown };
+    | { success: true; attempt?: number; warningPayload?: unknown }
```
peter-sanderson: thinking laud: `attempt` -> `attemptCount`?

peter-sanderson (approved):

packages/connect/src/device/Device.ts
Lemonexe: 23429449

packages/connect/src/device/Device.ts
Lemonexe: 23429449

packages/connect/src/types/device.ts
Lemonexe: 23429449

PR #15677, opened 2024-12-02
fix(suite-native): improve DAC error handling and report check result

yanascz:
Any unexpected error results fall back to the compromised summary screen while check result is reported for all outcomes.

## Related Issue

#14617

### DIFF
--- a/suite-native/module-device-settings/src/components/DeviceAuthenticityCard.tsx
+++ b/suite-native/module-device-settings/src/components/DeviceAuthenticityCard.tsx
@@ -3,6 +3,7 @@ import { useSelector } from 'react-redux';
 
 import { useNavigation } from '@react-navigation/native';
 
+import { analytics, DeviceAuthenticityCheckResult, EventType } from '@suite-native/analytics';
 import { selectDevice } from '@suite-common/wallet-core';
 import { requestPrioritizedDeviceAccess } from '@suite-native/device-mutex';
 import { useAlert } from '@suite-native/alerts';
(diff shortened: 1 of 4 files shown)


suite-native/module-device-settings/src/components/DeviceAuthenticityCard.tsx
```diff
@@ -48,13 +58,22 @@ export const DeviceAuthenticityCard = () => {
...
+            const checkResult = payload.valid ? 'successful' : 'compromised';
+            navigation.navigate(DeviceAuthenticityStackRoutes.AuthenticitySummary, { checkResult });
+            reportCheckResult(checkResult);
         } else if (payload.code === 'Failure_ActionCancelled') {
             navigation.goBack();
+            reportCheckResult('cancelled');
+        } else if (payload.code === 'Method_Interrupted') {
+            // navigation.goBack() already called by the X icon button
+            reportCheckResult('cancelled');
+        } else {
```
PeKne, line 64: Are we sure that navigation.goBack is called in this case as well? Are we sure that there is no scenario where user will get stuck on this page?

PeKne (approved):
One question, otherwise looks good.

suite-native/module-device-settings/src/components/DeviceAuthenticityCard.tsx
yanascz: You mean the `else` fallback branch? That one is supposed to take you to the summary screen, going back is not relevant. Otherwise I'm not sure I understand.

STew790:
**QA OK**
Issues mentioned [here](https://github.com/trezor/trezor-suite/issues/14617) fixed and I couldn't break it any more.

**Info**
24.11.3 b03e70c

PR #15686, opened 2024-12-02
`getFeatures` timeout fix

marekrjpolak:
## Description

Since we add `checkFirmwareHash`, `checkFirmwareRevision` and `changeLanguage` into `getFeatures` call, initial device acquiring often fails because of `GetFeatures timeout`, resulting in various issues. By moving this logic after the `getFeatures`/`initialize` call in `_runInner`, we could improve it a lot without (hopefully) changing anything else.

Example: With PIN-protected device, following happened:
- start Suite with webusb, unlock the device and let discovery finish
- lock the device and refresh Suite
- `checkFirmwareHash` inside `getFeatures` requires entering PIN, which cannot be done during the 3s timeout, resulting in recursive `_runInner` call and `initialize` call sent over other messages, which usually made device unreadable

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -583,6 +583,37 @@ export class Device extends TypedEmitter<DeviceEvents> {
             }
         }
 
+        if (this.authenticityChecks.firmwareHash === null) {
+            this.authenticityChecks.firmwareHash = await this.checkFirmwareHash();
+        }
+
+        if (
+            // The check was not yet performed
+            this.authenticityChecks.firmwareRevision === null ||
+            // The check was performed, but outcome cannot be surely determined (Suite is offline or there was an unexpected error)
+            // -> recheck on every getFeatures() until the result is known
+            (this.authenticityChecks.firmwareRevision.success === false &&
+                ['cannot-perform-check-offline', 'other-error'].includes(
+                    this.authenticityChecks.firmwareRevision.error,
+                ))
+        ) {
+            await this.checkFirmwareRevision();
+        }
+
+        if (
+            this.features?.language &&
+            !this.features.language_version_matches &&
+            this.atLeast('2.7.0')
+        ) {
+            _log.info('language version mismatch. silently updating...');
+
+            try {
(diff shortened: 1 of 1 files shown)


mroz22 (commented):
🤔  we just need to agree on that we don't need this logic to happen in the  "finalReload" of features after each call.

marekrjpolak:
> we just need to agree on that we don't need this logic to happen in the "finalReload" of features after each call.

Right @mroz22. This is not 100% equivalent. Do you think it's possible that some of the three mentioned actions would fail before the inner fn call, but succeed afterwards? And if it does, that it poses a security risk to wait until the next device.run instead?

mroz22:
not sure, I am more afraid of not doing something that  @Lemonexe  might be relying on. Lets ask him.

 Also #15675 is sort-of related because it might be attempt to fix something that this PR is preventing.

mroz22 (approved):
I am totally good with it as long as @Lemonexe confirms that it doesn't beak anything for him

Lemonexe (approved):
LGTM :rocket:
I have tested both hash  & revision check extensively, all cases work as before, I haven't noticed any issues :heavy_check_mark: 

Hash check
- `success` (physical T2T1, T1B1, T2B1 latest all the same time)
- `hash-mismatch` (user env T2T1 2.8.1, T1B1 1.12.1)
- `check-unsupported` (user env T2T1 2.7.2, 2.3.0, T1B1 1.10.5)
- `unknown-release` (user env 2-main, 1-main)


Revision check:
- `success` (physical T2T1, T1B1, T2B1 latest all the same time, several user env releases)
- `revision-mismatch` (user env T2T1 2.3.0)
- `firmware-version-unknown` (user env 2-main, 1-main)
- `other-error` (2-main without internet (should be `cannot-perform-check-offline` but that's a known bug))

mroz22:
In that case :shipit:

bosomt:
**QA OK**

**Info:**
 - Suite version: web 25.1.0 (a552ac45e894db3cf8ab9dab49a040b3c98f7132)
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: WebUsbTransport

PR #15689, opened 2024-12-02
Run e2e tests against emulator from URL or firmware branch

grdddj:
Allow for running `e2e` tests not only with emulators hardcoded in `tenv`, but also from:
- specified URL
- firmware branch

The way of running these is either:
- hardcoding these in `packages/connect/e2e/run.ts`: 
```ts
const firmwareArg = process.env.TESTS_FIRMWARE;
const firmwareUrl = process.env.TESTS_FIRMWARE_URL;
const firmwareModel = process.env.TESTS_FIRMWARE_MODEL;
const firmwareBranch = process.env.TESTS_FIRMWARE_BRANCH;
const firmwareBtcOnly = process.env.TESTS_FIRMWARE_BTC_ONLY === 'true';
```
- setting ENV variable when spawning the tests:
```sh
TESTS_FIRMWARE_MODEL="T3T1" TESTS_FIRMWARE_URL="https://data.trezor.io/dev/firmware/emu-branches/grdddj/emulator_upload_all_branches/trezor-emu-core-T3T1-universal" yarn workspace @trezor/connect-web test:e2e:node
TESTS_FIRMWARE_MODEL="T3T1" TESTS_FIRMWARE_BRANCH="obrusvit/refactor-confirm-summary" TESTS_FIRMWARE_BTC_ONLY="true" yarn workspace @trezor/connect-web test:e2e:node

```

NOTE: for local development, emulators from branches do not support `ARM`, so it will work for you only in CI

### DIFF
--- a/packages/trezor-user-env-link/src/api.ts
+++ b/packages/trezor-user-env-link/src/api.ts
@@ -22,9 +22,35 @@ export interface StartEmu {
     model?: Model;
 }
 
-interface StartEmuFromUrl extends Omit<StartEmu, 'version'> {
-    url: string;
+interface StartEmuType {
+    type: 'emulator-start';
+    version?: string;
+    wipe: boolean;
+    model: Model;
 }
+
+interface StartEmuFromUrl {
+    url: string;
+    model: Model;
+    wipe: boolean;
+}
+
+interface StartEmuFromUrlType extends StartEmuFromUrl {
+    type: 'emulator-start-from-url';
+}
+
+interface StartEmuFromBranch {
+    branch: string;
+    btcOnly: boolean;
+    model: Model;
+    wipe: boolean;
+}
+
+interface StartEmuFromBranchType extends StartEmuFromBranch {
+    type: 'emulator-start-from-branch';
+}
+
+export type EmuStartOptsType = StartEmuType | StartEmuFromUrlType | StartEmuFromBranchType;
 
 interface ClickEmu {
     x: number;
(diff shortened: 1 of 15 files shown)


packages/trezor-user-env-link/src/api.ts
```diff
@@ -22,10 +22,33 @@ export interface StartEmu {
     model?: Model;
 }
 
-interface StartEmuFromUrl extends Omit<StartEmu, 'version'> {
+interface StartEmuType extends StartEmu {
+    type?: 'emulator-start';
```
mroz22: should `type` be optional here?

mroz22: also for `StartEmuType` should wipe and model be optional?

mroz22 (commented):
this is definitely a nice step forward. maybe cc @szymonlesisz who asked for this feature.

packages/trezor-user-env-link/src/api.ts
grdddj: `type` needs to be optional here so everything is working in runtime and is `typescript`-correct in `packages/connect/e2e/common.setup.ts`. Crucial lines are below:
```ts
const emulatorStartOpts: EmuStartOptsType =
    (process.env.emulatorStartOpts as any) || global.emulatorStartOpts || {};
const emuStartType = emulatorStartOpts.type;

switch (emuStartType) {
  case 'emulator-start':
  case undefined:
   ...
```

For runtime, it is possible that the `process.env.emulatorStartOpts` is empty, so it will try to take `global.emulatorStartOpts` ... I did not know where it is defined, I see only the interface:
```ts
    var emulatorStartOpts: {
        version: string;
        model: 'T1B1' | 'T2T1' | 'T2B1' | 'T3T1';
    };
```

So one possibility is to add `type: 'emulator-start'` there, and the second one was to make it optional and account for the `undefined` in switch.

grdddj: Good point, in `StartEmuType` almost all parameters are needed - `version`, `model`, `wipe` - that probably means not extending from `StartEmu`, as it overrides almost everything

grdddj: Correction - it was now possible to get rid of the `type` optionality - [3eddfc5](https://github.com/trezor/trezor-suite/pull/15689/commits/3eddfc5552f23084494f7a8c624e4a47509b3965)

However, now the `version` needs to be optional, because it is potentially `undefined` in this case:
```ts
const latest = getLatestFirmware(model);
const version = firmwareArg
    ? firmwareArg?.endsWith('-latest')
        ? latest
        : firmwareArg
    : latest;
emulatorStartOpts = {
    type: 'emulator-start',
    wipe: true,
    version,
    model,
};
```

mroz22 (approved):
/rebase

PR #15691, opened 2024-12-02
Non-blockbook account fetch optimisations

tomasklim:
## Description

- do not refresh non-blockbook networks when app is minimised, hidden or browser tab is not active
- increase refresh rate from 1 minute to 5 minutes for cardano and solana
- refresh account on enter if it's older than 1 minute
- migrate sol backends from `solanaX.trezor.io` to `solX.trezor.io` so that we can cut old suite versions if needed

~TODO: [refresh](https://www.figma.com/design/0Kz3w6WDlvQm62RjOm87na/Staking?node-id=6524-6342&t=jURkugrpgLEMINJ5-1) button~ maybe will be done in some future PRs

@trezor/qa please test that Cardano and Solana accounts are updated every 5 minutes and that they are updated when you enter them (if they are older than 1 minute). You can check networks tab and just filter `sol`, for cardano you have to check websockets


packages/suite/src/actions/suite/windowActions.ts
```diff
@@ -0,0 +1,12 @@
+import { WINDOW } from './constants';
```
tomasklim, line 3: I agree, it would be nice to refactor it using redux toolkit, but no time for that

packages/suite/src/hooks/suite/useWindowVisibility.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react';
+
+import { updateWindowVisibility } from 'src/actions/suite/windowActions';
+
+import { useDispatch } from './useDispatch';
+
+export const useWindowVisibility = () => {
+    const dispatch = useDispatch();
+
+    const onWindowVisibilityChange = () => {
```
tomasklim, line 10: it would be possible to extend this by also having state from electron layer. There is one more state only for macOS. You can also detect when app is cover by another app, but that would make it more complicated for now. Let's tested at least using just [Page visibility api](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) from renderer layer

suite-common/wallet-core/src/accounts/accountsMiddleware.ts
```diff
@@ -0,0 +1,26 @@
...
+
+        next(action);
+
+        if (accountsActions.updateSelectedAccount.match(action) && action?.payload?.account?.key) {
+            const accountKey = action.payload.account.key;
+            const updatedAt = action.payload.account.ts || 0; // safety, old versions of Suite does not have this attribute
+            const isLessThanDefaultSyncInterval =
+                Date.now() - updatedAt < DEFAULT_ACCOUNT_SYNC_INTERVAL;
+
+            if (!isLessThanDefaultSyncInterval) {
```
tomasklim, line 22: This is little bit stupid but 1 minute is okay for now, it would be great to have specific refresh time limit for specific chains

martykan (approved):
LGTM :shipit:

bosomt:
**QA OK**

afaik works as expected
i dont see much traffic in network tab when app or browser is minimalised
fiat rates are not updated during this time

**Info:**
 - Suite version: web 24.12.1 (7571aef50689644cd309c06c809e2ceea025e021)
 - OS: MacIntel
 - Screen: 1512x982

PR #15700, opened 2024-12-02
Enable Solana on mobile via message system feature

### LINKED ISSUE #15613: Prepare feature flag for Solana in message-system
Add possibility to unhide/hide SOL in app via message-system same as we already have via feature flag in Dev menu. 

For that, we have to fix our global handling of feature flags from message system so that only killswitch open full screen blocking modal.

vytick:
- feature messages are not blocking entire app anymore apart of killswitch
- Solana can be enabled/disabled in mobile via message system
- Disabling message to be deployed via #15701

Resolve #15613

### DIFF
--- a/suite-native/feature-flags/src/featureFlagsSlice.ts
+++ b/suite-native/feature-flags/src/featureFlagsSlice.ts
@@ -6,7 +6,6 @@ import { isDebugEnv, isDetoxTestBuild, isDevelopOrDebugEnv } from '@suite-native
 export const FeatureFlag = {
     IsDeviceConnectEnabled: 'isDeviceConnectEnabled',
     IsCardanoSendEnabled: 'isCardanoSendEnabled',
-    IsSolanaSendEnabled: 'isSolanaSendEnabled',
     IsRegtestEnabled: 'isRegtestEnabled',
     IsSolanaEnabled: 'IsSolanaEnabled',
     IsConnectPopupEnabled: 'IsConnectPopupEnabled',
@@ -22,7 +21,6 @@ export type FeatureFlagsRootState = {
 export const featureFlagsInitialState: FeatureFlagsState = {
     [FeatureFlag.IsDeviceConnectEnabled]: isAndroid() || isDebugEnv(),
     [FeatureFlag.IsCardanoSendEnabled]: isAndroid() && isDevelopOrDebugEnv(),
-    [FeatureFlag.IsSolanaSendEnabled]: isAndroid() && isDevelopOrDebugEnv(),
     [FeatureFlag.IsRegtestEnabled]: isDebugEnv() || isDetoxTestBuild(),
     [FeatureFlag.IsSolanaEnabled]: false,
     [FeatureFlag.IsConnectPopupEnabled]: isDevelopOrDebugEnv(),
@@ -31,7 +29,6 @@ export const featureFlagsInitialState: FeatureFlagsState = {
 export const featureFlagsPersistedKeys: Array<keyof FeatureFlagsState> = [
     FeatureFlag.IsDeviceConnectEnabled,
     FeatureFlag.IsCardanoSendEnabled,
-    FeatureFlag.IsSolanaSendEnabled,
     FeatureFlag.IsRegtestEnabled,
     FeatureFlag.IsSolanaEnabled,
     FeatureFlag.IsConnectPopupEnabled,
--- a/suite-native/message-system/src/components/KillswitchMessageScreen.tsx
+++ b/suite-native/message-system/src/components/KillswitchMessageScreen.tsx
@@ -5,8 +5,11 @@ import { A, G } from '@mobily/ts-belt';
 import { Box, Button, PictogramTitleHeader, VStack } from '@suite-native/atoms';
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 import { messageSystemActions, selectActiveFeatureMessages } from '@suite-common/message-system';
+import { Variant } from '@suite-common/suite-types';
 import { Translation } from '@suite-native/intl';
 import { useOpenLink } from '@suite-native/link';
+
+import { selectActiveKillswitchMessages } from '../selectors';
 
 const screenStyle = prepareNativeStyle(utils => ({
     flexGrow: 1,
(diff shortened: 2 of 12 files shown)


suite-native/message-system/src/components/KillswitchMessageScreen.tsx
```diff
@@ -32,31 +34,24 @@ const buttonsWrapperStyle = prepareNativeStyle(_ => ({
...
     const dispatch = useDispatch();
     const openLink = useOpenLink();
-
-    const activeFeatureMessages = useSelector(selectActiveFeatureMessages);
-    const firstFeatureMessage = A.head(activeFeatureMessages);
-
     const { applyStyle } = useNativeStyles();
 
-    if (!firstFeatureMessage) return null;
+    const killswitch = A.head(useSelector(selectActiveKillswitchMessages));
```
PeKne: nit: What about changing this selector to `selectFirstActiveKillswitchMessage` and put the `A.head` inside?

suite-native/message-system/src/selectors.ts
```diff
@@ -0,0 +1,9 @@
+import { createMemoizedSelector, selectActiveFeatureMessages } from '@suite-common/message-system';
+
+export const selectActiveKillswitchMessages = createMemoizedSelector(
```
matejkriz: Why don't you return only one killswitch message directly from here? You never need the whole list, always only the highest priority... so `A.head` can be moved here, I guess?

suite-native/message-system/src/selectors.ts
matejkriz, line 1: I would prefer not to use such a generic file name, it's hard to orient with many files with the same name. I would prefer `messageSystemSelectors.ts` or `nativeMessageSystemSelectors.ts`.

PeKne (commented):
The code looks good  👍

suite-native/message-system/src/selectors.ts
matejkriz: nitpick, no need to fix it until we have a team consensus...

suite-native/message-system/src/selectors.ts
vytick: 49a065c965

suite-native/message-system/src/selectors.ts
vytick: 49a065c965

suite-native/message-system/src/components/KillswitchMessageScreen.tsx
vytick: 49a065c965

matejkriz: Even `selectActiveKillswitchMessage` (without the `First`) sounds good to me for this (only one can be active...)

suite-native/feature-flags/src/featureFlagsSlice.ts
```diff
@@ -24,19 +24,19 @@ export const featureFlagsInitialState: FeatureFlagsState = {
...
 };
 
 export const featureFlagsPersistedKeys: Array<keyof FeatureFlagsState> = [
     FeatureFlag.IsDeviceConnectEnabled,
     FeatureFlag.IsRippleSendEnabled,
     FeatureFlag.IsCardanoSendEnabled,
-    FeatureFlag.IsSolanaSendEnabled,
     FeatureFlag.IsRegtestEnabled,
     FeatureFlag.IsSolanaEnabled,
+    FeatureFlag.IsSolanaEnabledByRemote,
```
matejkriz: Not sure if it makes sense to mix the feature flag from message system with local feature flags in one slice 🤔 It seems like a duplication of the state to me.

suite-native/module-accounts-management/src/selectors.ts
```diff
@@ -1,42 +1,23 @@
...
         state,
         FeatureFlag.IsCardanoSendEnabled,
     );
 
-    if (isCardanoSendEnabled && networkType === 'cardano') return true;
-
-    const isSolanaSendEnabled = selectIsFeatureFlagEnabled(state, FeatureFlag.IsSolanaSendEnabled);
-
-    if (isSolanaSendEnabled && networkType === 'solana') return true;
+    if (networkType !== 'cardano' || isCardanoSendEnabled) return true;
```
vytick, line 20: send for all the other networkTypes is already enabled, and we will enable sending solana at the same time as we start supporting it. So there is no need to check anything else but cardano FF

PeKne (approved):

suite-native/feature-flags/src/featureFlagsSlice.ts
vytick: changed

suite-native/module-accounts-management/src/selectors.ts
PeKne: makes sense to simplify that 👍

suite-native/feature-flags/src/featureFlagsSlice.ts
```diff
@@ -44,6 +41,12 @@ export const featureFlagsSlice = createSlice({
         toggleFeatureFlag: (state, { payload }: PayloadAction<{ featureFlag: FeatureFlag }>) => {
             state[payload.featureFlag] = !state[payload.featureFlag];
         },
+        setFeatureFlag: (
```
matejkriz: setFeatureFlag is not used anymore

matejkriz (approved):
Great, it's much simpler now and works correctly, thanks! Just please remove  unused `setFeatureFlag` 🙏

suite-native/feature-flags/src/featureFlagsSlice.ts
vytick: removed

vytick:
Notes for QA @STew790 :  There are two stages of testing, **1.** can be done right away, but lets sync up on testing **2.**

1. App behaviour:
- By default Solana should be disabled
- Solana cant be imported nor is discovered or added new account to device
- enabling local FF should enable it

2. needs to deploy message to develop with enabling Solana and afterwards. App behaviour:
- By default Solana should be enabled
- User should be able to import, discover add and send Solana
- local FF is irrelevant in this case

STew790:
**QA OK**
Tested both behaviors, both worked as expected.
Tested also disabling and works also as expected. :tada:

PR #15706, opened 2024-12-03
 fix(connect): ignore disposed error in devicelist

martykan:
## Description

This should help get rid of the Disposed error in init tests.


marekrjpolak:
My observation:

- `catch()` in `waitForDevices` would probably cause continuing of `initializeTransport` and emitting `TRANSPORT.START` even in case of dispose.
- `catch()` in `createReconnectPromise` is probably fine, but I believe we can catch everything as it's only assigned to initPromise which should (probably?) never throw.

Also, this will be refactored a lot in #15324.

martykan:
Let's try this and keep monitoring the tests continually

marekrjpolak (approved):

PR #15718, opened 2024-12-03
feat(suite-native): replace illustrations and lotties with updated ones

### LINKED ISSUE #15441: Replace illustrations & lotties on prod
We need to replace the old illustrations on prod with new ones (ZIP sharedn slack). 

for 90% of them, we won't need any layout changes just a copy paste. 

Adjusments neccessary:
- Sheet changed to alert for view-only promo banner + biometrics promo banner 
- Send form (new card + lottie arrows)

Get in touch with @shenkys for more info. 

Figma with all illustrations [here](https://www.figma.com/design/IfTjYGULm83BbbTUPcRjav/Illustration-update?node-id=807-2615&node-type=frame&t=63qwoH8DKGGcir37-11)

yanascz:
Illustrations and lotties updated, layout adapted accordingly.

## Related Issue

Resolve #15441

## Screenshots:

Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 10 28 28
Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 10 28 37
Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 10 31 09
Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 10 35 52
Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 10 42 42
Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 10 54 56
Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 10 56 14
Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 10 57 17
Simulator Screenshot - iPhone 16 Pro - 2024-12-03 at 12 01 24


suite-native/atoms/src/LottieAnimation.tsx
```diff
@@ -0,0 +1,78 @@
...
+    size?: LottieAnimationSize;
+};
+
+type LottieColor = [number, number, number, 1];
+type LottieColors = {
+    lineColor: LottieColor;
+    fillColor: LottieColor;
+};
+
+const hexToLottieColor = (hex: CSSColor): LottieColor => {
```
PeKne, line 22: nit: CSSColor might by something different then hex representation of the color. See: https://github.com/trezor/trezor-suite/blob/f3db994654fbba0255db0e6d8d547fc002d9134d/packages/theme/src/types.ts#L1

Correct type should be probably something like `#${string}` or even stricter.

suite-native/module-home/src/screens/HomeScreen/useShowBiometricsAlert.tsx
```diff
@@ -1,57 +1,78 @@
...
+                        subtitle={<Translation id="moduleHome.biometricsModal.description" />}
+                    />
+                </VStack>
+            ),
+        });
+    }, [showAlert, handleEnable, handleCancel]);
 
     useEffect(() => {
+        if (isBiometricsInitialSetupFinished) {
+            return;
```
PeKne, line 74: return inside of useEffect is reserved for effect cleanup actions. Please avoid doing this.

PeKne (commented):
I haven't tested it yet, just looking in the code so far. Look good 👍

suite-native/atoms/src/LottieAnimation.tsx
yanascz: That'd be great but then I cannot use the values from the color scheme.

suite-native/module-home/src/screens/HomeScreen/useShowBiometricsAlert.tsx
yanascz: As discussed in Slack, resolving.

suite-native/atoms/src/LottieAnimation.tsx
yanascz: Anyway, it should be possible to extract `HexColor` from `CSSColor` and use it in `pallete.ts` and `colors.ts`. What do you think?

yanascz: It seemed as an easy fix but it breaks `TextButton`. Leaving as is.

PeKne (approved):

PR #15727, opened 2024-12-03
chore: unify types for setTimeout return type to address the NodeJS types leak issue

peter-sanderson:
```ts
/**
 * Hack to workaround NodeJS.Timeout vs number type issue.
 *
 * Some reads on the topic:
 *  - https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window
 *  - https://guilhermesimoes.github.io/blog/making-settimeout-return-number-in-typescript
 *
 */
 ```

### DIFF
--- a/packages/coinjoin/src/client/CoinjoinPrison.ts
+++ b/packages/coinjoin/src/client/CoinjoinPrison.ts
@@ -1,4 +1,5 @@
 import { TypedEmitter } from '@trezor/utils';
+import { ImmediateId, TimerId } from '@trezor/type-utils';
 
 import { CoinjoinPrisonInmate, CoinjoinPrisonEvents } from '../types/client';
 import { WabiSabiProtocolErrorCode } from '../enums';
@@ -28,10 +29,7 @@ export interface DetainOptions {
 
 export class CoinjoinPrison extends TypedEmitter<CoinjoinPrisonEvents> {
     inmates: CoinjoinPrisonInmate[] = [];
-    private changeEventThrottle:
-        | ReturnType<typeof setImmediate>
-        | ReturnType<typeof setTimeout>
-        | undefined;
+    private changeEventThrottle: ImmediateId | TimerId | undefined;
 
     constructor(initialState: CoinjoinPrisonInmate[] = []) {
         super();
--- a/packages/type-utils/src/timeout.ts
+++ b/packages/type-utils/src/timeout.ts
@@ -1,0 +1,11 @@
+/**
+ * Hack to workaround NodeJS.Timeout vs number type issue.
+ *
+ * Some reads on the topic:
+ *  - https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window
+ *  - https://guilhermesimoes.github.io/blog/making-settimeout-return-number-in-typescript
+ *
+ */
+export type TimerId = ReturnType<typeof setTimeout>;
+export type ImmediateId = ReturnType<typeof setImmediate>;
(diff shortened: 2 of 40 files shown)

### CODE REVIEW

File: packages/type-utils/src/timeout.ts [hunk 1]
```diff
@@ -0,0 +1,10 @@
+/**
+ * Hack to workaround NodeJS.Timeout vs number type issue.
+ *
+ * Some reads on the topic:
+ *  - https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window
+ *  - https://guilhermesimoes.github.io/blog/making-settimeout-return-number-in-typescript
+ *
+ */
+export type TimerId = ReturnType<typeof setTimeout>;
```
Nodonisko, line 9: what about interval?

packages/coinjoin/src/client/CoinjoinPrison.ts
```diff
@@ -2,6 +2,7 @@ import { TypedEmitter } from '@trezor/utils';
 
 import { CoinjoinPrisonInmate, CoinjoinPrisonEvents } from '../types/client';
 import { WabiSabiProtocolErrorCode } from '../enums';
+import { ImmediateId, TimerId } from '@trezor/type-utils/libDev/src';
```
Lemonexe: `libDev` import :scream:

packages/type-utils/src/timeout.ts
Lemonexe: indeed, we have some `ReturnType<typeof setInterval>` in our code, [e.g. here](https://github.com/trezor/trezor-suite/blob/b03e70cbc8d17e535e3d9ff5229596ee4b89e2e0/packages/blockchain-link/src/workers/electrum/client/caching.ts#L18)

Lemonexe:
Looks good, but comments above, also lint errors..

packages/coinjoin/src/client/CoinjoinPrison.ts
Nodonisko: this should be guarded by eslint, if not it was lost in latest refactor

Lemonexe (approved):
:rocket:

packages/coinjoin/src/client/CoinjoinPrison.ts
Lemonexe: @Nodonisko hmm, I tried locally and eslint doesn't catch it, so FYI @peter-sanderson 
But you can resolve this thread :heavy_check_mark:

peter-sanderson: @Nodonisko do you recall how this was guarded? I am eager to bring it back. 

But I remember finding [this](https://github.com/trezor/trezor-suite/pull/14663) way before I merged the ESLint refactor.

Nodonisko: By this rule and wildcard https://eslint.org/docs/latest/rules/no-restricted-imports

peter-sanderson: Thanks I'll make sure to add it

peter-sanderson: I have found the config in past version of the suite https://github.com/trezor/trezor-suite/blob/release/24.4/.eslintrc.js#L222C10-L222C50

I will prepare test PR

peter-sanderson: Fix: https://github.com/trezor/trezor-suite/pull/15745

PR #15730, opened 2024-12-03
Chore: fix any types

Lemonexe:
## Description

Random fix of two glaring `any`s.

TS only change

### DIFF
--- a/packages/components/src/components/form/SelectBar/SelectBar.tsx
+++ b/packages/components/src/components/form/SelectBar/SelectBar.tsx
@@ -170,7 +170,7 @@ export type SelectBarProps<V extends ValueTypes> = {
 } & AllowedFrameProps;
 
 // Generic type V is determined by selectedOption/options values
-export const SelectBar: <V extends ValueTypes>(props: SelectBarProps<V>) => JSX.Element = ({
+export const SelectBar = <V extends ValueTypes>({
     label,
     options,
     selectedOption,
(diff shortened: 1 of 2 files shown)


packages/components/src/components/form/SelectBar/SelectBar.tsx
```diff
@@ -170,7 +170,7 @@ export type SelectBarProps<V extends ValueTypes> = {
 } & AllowedFrameProps;
 
 // Generic type V is determined by selectedOption/options values
-export const SelectBar: <V extends ValueTypes>(props: SelectBarProps<V>) => JSX.Element = ({
```
Lemonexe, line 173: :thought_balloon:  no need to specify return type `JSX.Element`. React throws TS error when calling a non-compliant function as `<Component>`.
Though personally I'd like declarative typing for components - if the function returns invalid type, TS should ideally throw that in the component definition, not in the JSX that's calling it. But:
1. we don't do that almost anywhere :cry:
2. `JSX.Element` is deprecated anyway, `ReactNode` would be better, and `React.FC` best...
3. in this case it'd break inference of `V` if we had `const SelectBar: FC<whatever> = function`

packages/connect/src/data/downloadReleasesMetadata.ts
```diff
@@ -11,12 +11,12 @@ export const downloadReleasesMetadata = async ({
 }: DownloadReleasesMetadataParams): Promise<FirmwareRelease[]> => {
     const url = `https://data.trezor.io/firmware/${internal_model.toLowerCase()}/releases.json`;
 
-    const response = (await httpRequest(
+    const response = await httpRequest(
         url,
         'json',
         { signal: AbortSignal.timeout(10000) },
         true, // skipLocalForceDownload=true
-    )) as any;
```
peter-sanderson, line 19: :heart:

mroz22 (approved):

peter-sanderson (approved):

PR #15737, opened 2024-12-03
Remove messages type check to see if it fixes Crowdin CLI

komret:
Something keeps deleting the whole Crowdin database. Reverting to the original file structure to see if that helps.


MiroslavProchazka (approved):
Tak snad už naposled teda :D

PR #15690, opened 2024-12-02
chore(suite): update network symbol naming & fix passing symbol 4

adderpositive:
## Description
- update network symbol naming & amount limits update

## Related 
Follow up https://github.com/trezor/trezor-suite/pull/15644

Related to #14725

Issue #14725, opened 2024-10-04
Introduce network symbol display name


tomasklim:
Some networks have different native coin shortcut than is the shortcut of network itself

Suite is not ready for:
`BNB` to `BSC`
`OP` to `ETH`
`BASE` to `ETH`
`ARB` to `ETH`

- [ ] Unify occurrences of `NetworkSymbol`. Use naming `symbol` as is used by Blockbook. 
  - symbol in the codebase is called by `asset`, `coin`, `logo`,... we should unify so that it is easier to do ⬇️ 
  - sometimes the symbol is not typed as `NetworkSymbol` but by `string`, fix that
- [ ] Introduce `displaySymbol` (or better naming) and replace `symbol` on places where it is presented to a user

NetworkSymbol can be btc, eth, sol,...

Some networks (Optimism, BNB Smart Chain, Base,...) have different identificator of the network (op, bsc, base) than native coin (eth, bnb, eth). We need to introduce also something like DisplaySymbol, NativeSymbol, NativeCoin,... name TBD

Unfortunately Connect uses coin, Blockbook uses symbol, so we should stick to using symbol as we do not want to transform data.

bosomt:
**QA OK**


**Info:**
 - Suite version: web 25.1.0 (102be08a4eddf031a463e2907f96a9d45d9d748f)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.8.7 regular (revision 8a254aa8eae82f99630df63f40e4d290066a3efc)
 - Transport: WebUsbTransport

PR #15746, opened 2024-12-03
chore(suite): update network symbol naming 5

adderpositive:
## Description
- update network symbol naming 

## Related 
Follow up https://github.com/trezor/trezor-suite/pull/15690

Related to #14725


suite-common/wallet-utils/src/tokenUtils.ts
```diff
@@ -3,10 +3,10 @@ import { TokenInfo } from '@trezor/blockchain-link-types';
 import { parseAsset } from '@trezor/blockchain-link-utils/src/blockfrost';
 
 export const getContractAddressForNetwork = (
```
tomasklim:
```suggestion
export const getContractAddressForNetworkSymbol = (
```

suite-common/suite-utils/src/protocol.ts
```diff
@@ -6,10 +6,12 @@ export type ProtocolToNetwork = {
 };
 
 export const getNetworkSymbolForProtocol = (protocol: Protocol): NetworkSymbol | undefined => {
-    for (const networkSymbol in NETWORK_TO_PROTOCOLS) {
-        const protocols = NETWORK_TO_PROTOCOLS[networkSymbol as NetworkSymbol];
+    for (const symbolKey in NETWORK_TO_PROTOCOLS) {
+        const symbol = symbolKey as NetworkSymbol;
```
tomasklim: can this be better typed so that it does not need to use `as`?

suite-common/wallet-utils/src/tokenUtils.ts
adderpositive: resolved

suite-common/suite-utils/src/protocol.ts
adderpositive: I have found the way, I guess. Check it out in [e545246](https://github.com/trezor/trezor-suite/pull/15746/commits/e545246f74c546e1d43bd1183624acbb073638e1)

tomasklim (approved):

PR #15751, opened 2024-12-03
feat(mobile): custom fingerprint for runtime version

Nodonisko:
## Description

Our custom fingerprint version (at least as temp solution)

## Related Issue

Resolve


PeKne (approved):

Nodonisko:
Let's use expo fingerprint

PR #15753, opened 2024-12-03
chore: add pre-commit hook that does eslint before commit only on files that has been changed

peter-sanderson:
Some reads on topic:
- https://typicode.github.io/husky/get-started.html
- https://github.com/lint-staged/lint-staged?tab=readme-ov-file

### DIFF
--- a/.husky/eslint.sh
+++ b/.husky/eslint.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# This check will get all staged files that are *.js(x)/*.ts(x) files
+# and run eslint on them and tries to fix them and re-add them to be committed.
+# If auto-fix is possible it shall be transparent for the user.
+# If some error are not fixable, the script will fail and the user will have to fix them manually.
+
+STAGED_FILES=$(git diff --cached --name-only --diff-filter=d | grep '(\.js\|\.jsx$\|\.ts\|\.tsx)$')
+
+# Exit if no files. Passing no arguments would trigger eslint for whole repo
+if [ -z "$STAGED_FILES" ]; then
+  echo "No staged JavaScript/TypeScript files to lint."
+else
+  echo "Linting JavaScript/TypeScript files..."
+  echo "$STAGED_FILES"
+  echo ""
+
+  # No quotes to pass as separate arguments (files)
+  # shellcheck disable=SC2086
+  if ! yarn lint:js:fix-files $STAGED_FILES; then
+    echo "Eslint failed! Please fix the errors and try again. You can also use --no-verify to skip pre-commit checks."
+    exit 1
+  fi
+
+  git update-index --again
+fi
+
+
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+script_dir=$(dirname "$(realpath "$0")")
+
+if ! bash "$script_dir/eslint.sh"; then
+  exit 1
+fi
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     },
     "scripts": {
         "_______ Hooks _______": "Yarn hooks.",
-        "postinstall": "yarn run patch-package",
+        "postinstall": "yarn run patch-package && husky",
         "_______ Library Scripts #####": "Some libraries have their own build scripts.",
         "build:libs": "yarn nx run-many --skip-nx-cache --target=build:lib",
         "suite:build:web": "yarn workspace @trezor/suite-web build",
@@ -32,9 +32,9 @@
         "type-check:force": "rimraf -rf -- **/libDev && yarn type-check",
         "test:unit": "yarn nx run-many --target=test:unit",
         "lint:js": "eslint . --cache --cache-strategy content --flag unstable_config_lookup_from_file --max-warnings 0",
+        "lint:js:fix-files": "eslint --cache --cache-strategy content --flag unstable_config_lookup_from_file --max-warnings 0 --fix",
         "lint:styles": "yarn nx run-many --target=lint:styles",
         "lint": "yarn lint:styles && yarn lint:js",
-        "lint-staged": "npx lint-staged",
         "lint:shellcheck": "./scripts/shellcheck.sh",
         "_______ Global Scripts _______": "Shared scripts for running in all workspaces",
         "g:eslint": "cd $INIT_CWD && eslint . --cache --cache-strategy content --flag unstable_config_lookup_from_file --max-warnings 0",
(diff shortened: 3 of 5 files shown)


.husky/pre-commit
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+script_dir=$(dirname "$(realpath "$0")")
+
+if ! bash "$script_dir/eslint.sh"; then
+  exit 1
+fi
```
peter-sanderson, line 7: @jvaclavik you can add precommit hook for commit message after that here. I tried and got some ugly error so I leave it to you :D

package.json
```diff
@@ -84,17 +85,8 @@
         "ios": "yarn native:ios",
         "p": "yarn native:prebuild",
         "ports": "yarn native:reverse-ports",
-        "s": "yarn native:start"
-    },
-    "lint-staged": {
```
peter-sanderson, line 89: This seems to me as dead code. I have checked the lint-staged library and I do not think we need it. The simple script shall work the same way. Or am I missing something?

.husky/pre-commit
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+script_dir=$(dirname "$(realpath "$0")")
+
+if ! bash "$script_dir/eslint.sh"; then
```
peter-sanderson, line 5: Also `prettier` shall be somehow added

.husky/eslint.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+STAGED_FILED=$(git diff --cached --name-only --diff-filter=d | grep '(\.js\|\.jsx$\|\.ts\|\.tsx)$')
```
martykan: typo (FILED)

jvaclavik (approved):

package.json
Lemonexe: When I came across `lint-staged` I tried to make it work, but didn't succeed.

package.json
```diff
@@ -32,6 +32,7 @@
         "type-check:force": "rimraf -rf -- **/libDev && yarn type-check",
         "test:unit": "yarn nx run-many --target=test:unit",
         "lint:js": "eslint . --cache --cache-strategy content --flag unstable_config_lookup_from_file --max-warnings 0",
+        "lint:js:fix-files": "eslint --cache --cache-strategy content --flag unstable_config_lookup_from_file --max-warnings 0 --fix",
```
Lemonexe, line 35: nit: you could name it just `lint:js:fix`.
Because it doesn't need files arguments, it works on its own, and that's good - previously I used `yarn lint:js --fix` a lot :+1:

peter-sanderson: > Because it doesn't need files arguments

I you don't provide files, it will indeed run, but it is very undesirable in this case as it runs for all files then.

Lemonexe: for me it has been very desirable :laughing: 
So I thought this new script can save me time. But now I'm thinking about it again - with the commit hook I won't ever need to `yarn lint:js --fix`, as I have been doing past few weeks.
So ignore this.

package.json
```diff
@@ -19,7 +19,7 @@
     },
     "scripts": {
         "_______ Hooks _______": "Yarn hooks.",
-        "postinstall": "yarn run patch-package",
+        "postinstall": "yarn run patch-package && husky",
```
peter-sanderson, line 22:
```
# Yarn doesn't support `prepare`
```

See: https://typicode.github.io/husky/how-to.html#manual-setup

Lemonexe (approved):
Amazing DX improvement :rocket: 

Had problems that husky did not init, but after adding it to postinstall fix, now I can confirm it is working :heavy_check_mark: 

Also previously I had  problems with commiting via webstorm UI (only bash was working), but now I confirm it is works in terminal, webstorm UI and VSCode UI as well :heavy_check_mark:

.husky/eslint.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
komret, line 2: I'd add a simple comment explaining what it does. Like "Lints staged files before committing".

package.json
```diff
@@ -32,6 +32,7 @@
         "type-check:force": "rimraf -rf -- **/libDev && yarn type-check",
         "test:unit": "yarn nx run-many --target=test:unit",
         "lint:js": "eslint . --cache --cache-strategy content --flag unstable_config_lookup_from_file --max-warnings 0",
+        "lint:js:fix-files": "eslint --cache --cache-strategy content --flag unstable_config_lookup_from_file --max-warnings 0 --fix",
         "lint:styles": "yarn nx run-many --target=lint:styles",
         "lint": "yarn lint:styles && yarn lint:js",
         "lint-staged": "npx lint-staged",
```
komret: Should be removed.

.husky/eslint.sh
```diff
@@ -0,0 +1,23 @@
...
+  echo "No staged JavaScript/TypeScript files to lint."
+else
+  echo "Linting JavaScript/TypeScript files..."
+  echo "$STAGED_FILES"
+  echo ""
+
+  # No quotes to pass as separate arguments (files)
+  # shellcheck disable=SC2086
+  if ! yarn lint:js:fix-files $STAGED_FILES; then
+    echo "Eslint failed! Please fix the errors and try again."
```
komret: I can imagine someone wanting to commit some work in progress. The message could hint at `--no-verify` option which skips this check.

package.json
```diff
@@ -84,17 +85,8 @@
...
-        "packages/**/*.{ts,tsx}": [
-            "eslint --fix",
-            "git add"
-        ],
-        "packages/{suite,suite-web,suite-desktop-ui,connect-explorer,connect-popup,connect-ui,components,product-components}/**/*.{ts,tsx}": [
-            "stylelint --fix",
-            "git add"
-        ]
+        "s": "yarn native:start",
+        "prepare": "husky"
```
komret: Isn't this redundant with yarn?

.husky/eslint.sh
```diff
@@ -0,0 +1,23 @@
...
+if [ -z "$STAGED_FILES" ]; then
+  echo "No staged JavaScript/TypeScript files to lint."
+else
+  echo "Linting JavaScript/TypeScript files..."
+  echo "$STAGED_FILES"
+  echo ""
+
+  # No quotes to pass as separate arguments (files)
+  # shellcheck disable=SC2086
+  if ! yarn lint:js:fix-files $STAGED_FILES; then
```
komret, line 20:
```suggestion
  if ! yarn lint:js:fix-files "$STAGED_FILES"; then
```

komret (changes requested):

.husky/eslint.sh
peter-sanderson: This is there for purpose, see comment above.

package.json
peter-sanderson: :scream: I was doing full-text search a found nothing, Wat? 

Thanks for finding it. :bow:

komret (approved):

.husky/eslint.sh
peter-sanderson: fixed

PR #15754, opened 2024-12-03
feat(mobile): native tabbar

### LINKED ISSUE #15600: Make app feel more native manifest
Some ideas I collected over time:

1. Hairline width borders - super easy to do, we only need to agree which borders we want to thin out
2. [react-native-edge-to-edge](https://github.com/zoontek/react-native-edge-to-edge) - we will need to support this anyway due to Android requirements and it will be forced enabled for API 35+ devices so it will probably look ugly if we don't do that intentionally 
3. [Truly native bottom tabs](https://github.com/okwasniewski/react-native-bottom-tabs) - make app respect platform specific desing, we will loose some ability to customize it but it will look much more native for given platform and we will also get all this little platform specific animations for free.
4. [Use native screen headers](https://reactnavigation.org/docs/headers) - this one could be tricky but now our screen headers are part of screen body and not defined as header at all which is breaking all that small native animations that makes native apps feel really native. Native headers are not that much customizable as our custom one but I think we still could achieve pretty good results with only small compromises. [Some more examples how it looks like.](https://reactnavigation.org/docs/native-stack-navigator/)
5. In case we will decide to use native headers we should consider also [native search bar.](https://reactnavigation.org/docs/native-stack-navigator/#bartintcolor)
6. Make use of layout animations when displaying different Alert/Info boxes across app - instead of just layout jump we should use Reanimated Layout animations, it's really easy so no excuse to not to do that.
7. Explorer usage of `fullScreenGestureEnabled` (iOS only) - https://reactnavigation.org/docs/native-stack-navigator/#fullscreengestureenabled
8. Use some cool animation for changing some numbers (like portfolio/account value) - for example [this one](https://x.com/reactiive_/status/1747612088594321883)
9. Constantly monitor app performance - nothing makes non-native feeling than janky slow app, we should always keep this in mind, maybe even implement some tools like [Reassure](https://callstack.github.io/reassure/) for some critical components especially on homepage when we constantly fighting perf regression after any changes.

Sources:
[REACT NATIVE, AND "THE NATIVE FEEL"](https://mozzius.dev/post/3l777nhz4h32w)
[Edge-to-Edge display, now streamlined for Android](https://expo.dev/blog/edge-to-edge-display-now-streamlined-for-android)

Nodonisko:
## Description

It looks really cool IMHO, all that little native effects feels really "native" especially ripple effect on Android. Only drawback I found so far is that screen is not unmounted when you leave it, but I don't think that's issue anyway. There is also opened PR to `freeze` screens that are not currently focus which could be good as unmount https://github.com/okwasniewski/react-native-bottom-tabs/issues/71

Edit: Found blocker potentially, there is no `testID` prop rn.
Edit2: Luckily some guy made PR just today https://github.com/okwasniewski/react-native-bottom-tabs/pull/179

## Related Issue

Resolve #15600 

## Screenshots:

Simulator Screenshot - iPhone 16 - 2024-12-03 at 22 48 06
Screenshot_1733262483

### DIFF
--- a/suite-native/app/app.config.ts
+++ b/suite-native/app/app.config.ts
@@ -118,6 +118,7 @@ const getPlugins = (): ExpoPlugins => {
                 },
                 ios: {
                     deploymentTarget: '15.1',
+                    useFrameworks: 'static',
                 },
             },
         ],
@@ -135,6 +136,7 @@ const getPlugins = (): ExpoPlugins => {
                 subdomains: '*',
             },
         ],
+        'react-native-bottom-tabs',
     ];
 
     return [
--- a/suite-native/app/package.json
+++ b/suite-native/app/package.json
@@ -21,6 +21,7 @@
         "reverse-ports": "adb reverse tcp:8081 tcp:8081 && adb reverse tcp:21325 tcp:21325 && adb reverse tcp:19121 tcp:19121"
     },
     "dependencies": {
+        "@bottom-tabs/react-navigation": "^0.7.6",
         "@gorhom/bottom-sheet": "5.0.5",
         "@mobily/ts-belt": "^3.13.1",
         "@react-native-community/netinfo": "^11.4.1",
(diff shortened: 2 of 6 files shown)


PeKne:
Love that guy who fixed that for us 🙏. He looks familiar though 🤔...

matejkriz:
It's really promissing, but stale. Closing the PR, keeping the branch so we can reopen later.

PR #15771, opened 2024-12-04
Readd udev/bridge installation info to transport events

marekrjpolak:
## Description

Since #14959 Connect is integrated in Suite without iframe. Because of that, udev & bridge installation info [is not injected into transport events](https://github.com/trezor/trezor-suite/blob/develop/packages/connect-iframe/src/index.ts#L236-L241), which causes some issues in web Suite (compare production udev modal to develop one):
image
image

As a quick fix I started to enhance the events on Connect core level (including ugly window check) with which I'm not happy at all, and I'm not well oriented in all the new Connect entry points, so if you guys (@martykan @karliatto) have any idea where to put this to be applicable only in web environment, share it with me pls.

Until that, I'll try to come up with a better way of propagating this info into Suite than to extend all the transport events.

### DIFF
--- a/packages/connect-iframe/src/index.ts
+++ b/packages/connect-iframe/src/index.ts
@@ -25,9 +25,7 @@ import { DataManager } from '@trezor/connect/src/data/DataManager';
 import { config } from '@trezor/connect/src/data/config';
 import { initLog, LogWriter } from '@trezor/connect/src/utils/debug';
 import { getOrigin } from '@trezor/connect/src/utils/urlUtils';
-import { suggestBridgeInstaller } from '@trezor/connect/src/data/transportInfo';
-import { suggestUdevInstaller } from '@trezor/connect/src/data/udevInfo';
-import { storage, getSystemInfo, getInstallerPackage } from '@trezor/connect-common';
+import { storage, getSystemInfo } from '@trezor/connect-common';
 import { analytics, EventType } from '@trezor/connect-analytics';
 
 import { parseConnectSettings, isOriginWhitelisted } from './connectSettings';
@@ -231,13 +229,6 @@ const postMessage = (message: CoreEventMessage) => {
     // check if permissions to read from device is granted
     if (!trustedHost && message.event === DEVICE_EVENT && !filterDeviceEvent(message)) {
         return;
-    }
-
-    if (message.event === TRANSPORT_EVENT) {
-        // add preferred bridge installer
-        const platform = getInstallerPackage();
-        message.payload.bridge = suggestBridgeInstaller(platform);
-        message.payload.udev = suggestUdevInstaller(platform);
     }
 
     if (usingPopup && _popupMessagePort) {
--- a/packages/connect-web/src/module/index.ts
+++ b/packages/connect-web/src/module/index.ts
@@ -19,7 +19,15 @@ const impl = new TrezorConnectDynamic<
     implementations: [
         {
             type: 'core-in-module',
-            impl: new CoreInModule(),
+            impl: new CoreInModule((message: CoreEventMessage) => {
+                if (message.event === TRANSPORT_EVENT) {
+                    const platform = getInstallerPackage();
+                    message.payload.bridge = suggestBridgeInstaller(platform);
+                    message.payload.udev = suggestUdevInstaller(platform);
+                }
+
+                return message;
+            }),
         },
     ],
     getInitTarget: () => 'core-in-module',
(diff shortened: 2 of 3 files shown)


martykan:
Doesn't seem too bad, I don't have a better idea.
But maybe it should now be removed from `connect-iframe` so it's not duplicated?

karliatto:
Good catch! I think I have an idea how to do it so it is only applicable for web environment but I need to look a bit more on the code and try it.

packages/connect-iframe/src/index.ts
```diff
@@ -233,13 +231,6 @@ const postMessage = (message: CoreEventMessage) => {
         return;
     }
 
-    if (message.event === TRANSPORT_EVENT) {
```
karliatto, line 236: In any case, shouldn't this still be here? Maybe not if it was used only in suite.

mroz22:
nice. this sounds like something we will need to ask @MiroslavProchazka to cherry-pick. 
@karliatto if you could please help with finishing/review 🙏

karliatto:
I added commits:
* https://github.com/trezor/trezor-suite/pull/15771/commits/3fdc35eb62620da37ab795054bd14b24de8d84d3
* https://github.com/trezor/trezor-suite/pull/15771/commits/3d5e6a9e47a0def68f2dc54df4e69107ca25f729

Introducing hooks with new `onCoreHook` that allows us to introduce some changes in core messages `core-in-module` in this case. Like this we can make some necessary changes based on environment so we do not get dirty window objects in connect core.

packages/connect-iframe/src/index.ts
marekrjpolak: I'm not sure, but @szymonlesisz believes that only Suite ever used bridge/udev installer info.

packages/connect-web/src/module/index.ts
```diff
@@ -19,7 +28,15 @@ const impl = new TrezorConnectDynamic<
     implementations: [
         {
             type: 'core-in-module',
-            impl: new CoreInModule(),
+            impl: new CoreInModule((message: CoreEventMessage) => {
+                if (message.event === TRANSPORT_EVENT && typeof window !== 'undefined') {
```
marekrjpolak: Is it necessary to have `typeof window !== 'undefined'` check here? I mean, it's always better to be careful, but e.g. in connect-iframe there was no check as we were sure that it's always running in browser. Similarly, are we sure here in connect-web or not?

packages/connect/src/impl/core-in-module.ts
```diff
@@ -118,6 +121,10 @@ export class CoreInModule implements ConnectFactoryDependencies<ConnectSettingsP
 
         if (type === POPUP.CANCEL_POPUP_REQUEST) return;
 
+        if (this.onCoreHook) {
+            message = this.onCoreHook(message);
```
marekrjpolak: This is working correctly as we're modifying the nested payload in this case, but if we reassign it instead, it could be inconsistent with the `const { event, type, payload } = message` spread above, couldn't it? Wouldn't it be safer to call the hook right after the `cloneObject`?

marekrjpolak (commented):
Thank you for your help @karliatto, as I'm still confused with the core-in-something thingys. I wrote some comments, but I can confirm it's working in my case.

packages/connect-iframe/src/index.ts
karliatto: Ok, then let's remove it.

marekrjpolak:
Related issue - it's not working in desktop Suite either, but it's not a regression: https://github.com/trezor/trezor-suite/issues/15074

Cc: @MiroslavProchazka

packages/connect-web/src/module/index.ts
karliatto: Yes, connect-web should be running on web. Let's remove it https://github.com/trezor/trezor-suite/pull/15771/commits/54162d1fb8b7415515bcd6c01e9b7901cd1c999b

 I think we can assume that, in any case if we want to check it , we should check it when initializing or similar, not at this point. But I am taking a note to check that.

packages/connect/src/impl/core-in-module.ts
karliatto: Absolutely https://github.com/trezor/trezor-suite/pull/15771/commits/54162d1fb8b7415515bcd6c01e9b7901cd1c999b

karliatto (approved):

mroz22:
this one looks related https://github.com/trezor/trezor-suite/pull/15076

karliatto:
> this one looks related #15076

By description looks like the PR to fix https://github.com/trezor/trezor-suite/issues/15074  but seams like abandoned PR. Maybe it should re re-prioritize since sounds like something that deserves fixing.

STew790:
**QA OK**
I can see this on Linux with Trezor not connected

image

**Info:**
 - Suite version: web 25.1.0 (44ad7bc06636e1d0e447479432788e2bcc14518a)
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: 
 - Transport: WebUsbTransport

PR #15774, opened 2024-12-04
fix(suite): make sure getEthNetworkForWalletSdk never returns undefined

enjojoy:
## Description
Suite is crashing on develop on OP and BNB, this should fix this issue. 

## Screenshots:
image

### DIFF
--- a/packages/suite/src/utils/suite/stake.ts
+++ b/packages/suite/src/utils/suite/stake.ts
@@ -44,9 +44,9 @@ export const getEthNetworkForWalletSdk = (
         thol: 'holesky',
         eth: 'mainnet',
     };
-    const network = symbol && symbol !== 'unknown' ? ethNetworks[symbol] : ethNetworks.eth;
+    const network = symbol && symbol !== 'unknown' ? ethNetworks[symbol] : null;
 
-    return network!;
+    return network ?? 'mainnet';
 };
 
 export const getAdjustedGasLimitConsumption = (estimatedFee: Success<BlockchainEstimatedFee>) =>


enjojoy:
@Lemonexe Jirko, do you understand why in the initial case it can be undefined? I don't get how it's even possible and ChatGpt didn't explain it either

packages/suite/src/utils/suite/stake.ts
```diff
@@ -44,9 +44,9 @@ export const getEthNetworkForWalletSdk = (
         thol: 'holesky',
         eth: 'mainnet',
     };
-    const network = symbol && symbol !== 'unknown' ? ethNetworks[symbol] : ethNetworks.eth;
+    const network = symbol && symbol !== 'unknown' && ethNetworks[symbol];
```
Lemonexe: Nit: we generally prefer ternary instead of short-circuit.
In this case, it could mean that `false` is assigned to network, which doesn't make sense semantically.

```
const network = symbol && symbol !== 'unknown' ? ethNetworks[symbol] : null
```
And then you can `return network ?? 'mainnet'`
`??` is more specific than `||`, but doesn't match `false` :slightly_smiling_face:

Lemonexe (approved):
At first glance, the bug makes perfect sense.

There are many more ethereum networks than `eth` and `thol`, for example `tsep`.
When `symbol = tsep` The previous code then returned `ethNetworks[symbol]` which is ofc `undefined`.

But I haven't studied surrounding code - if `tsep` can ever get to this function. If it does, then having `tsep` turned on can cause crash :shrug:

enjojoy:
> At first glance, the bug makes perfect sense.
> 
> There are many more ethereum networks than `eth` and `thol`, for example `tsep`. When `symbol = tsep` The previous code then returned `ethNetworks[symbol]` which is ofc `undefined`.
> 
> But I haven't studied surrounding code - if `tsep` can ever get to this function. If it does, then having `tsep` turned on can cause crash 🤷

Yes, any other network can get to this function, for example for me it was crashing on BNB and OP.  That makes sense, thank you.

packages/suite/src/utils/suite/stake.ts
enjojoy: I implemented this 👍🏻

Lemonexe (approved):

enjojoy:
@adderpositive  feel free to merge it

adderpositive (approved):

PR #15781, opened 2024-12-05
fix(suite-native): send transaction error handling

### LINKED ISSUE #15464: Better error handling of transactions being not send
**Describe the bug**
There is a chance that transactions might not be sent and there are no errors that would inform the user about what's happening. For example backends might not be reachable and there is "forever" loading when you click on the final "Send" button.

**Info:**

-   OS: name, version

**Expected behavior**
I should be better informed what is happening and why is the transaction not sending

PeKne:
## Description
In case that transaction dispatch fails, the error is informed about it.

## Related Issue

Resolve #15464 

### Generic error
Snímek obrazovky 2024-12-05 v 10 46 32

# Solana timeout error
Snímek obrazovky 2024-12-05 v 11 03 57

### DIFF
--- a/suite-native/intl/src/en.ts
+++ b/suite-native/intl/src/en.ts
@@ -1144,6 +1144,19 @@ export const en = {
                 },
                 successMessage: 'Everything is ready, you can send the transaction now.',
                 submitButton: 'Send transaction',
+                errorAlert: {
+                    secondaryButtonTitle: 'I’ll do it later',
+                    generic: {
+                        title: 'Transaction failed',
+                        description:
+                            'There has been an unexpected error, please try sending your transaction again.',
+                    },
+                    solana: {
+                        title: 'Transaction failed due to timeout',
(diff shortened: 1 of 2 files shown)


Nodonisko (approved):

suite-native/intl/src/en.ts
```diff
@@ -1144,6 +1144,19 @@ export const en = {
...
                 submitButton: 'Send transaction',
+                errorAlert: {
+                    secondaryButtonTitle: 'I’ll do it later',
+                    generic: {
+                        title: 'Transaction failed',
+                        description:
+                            'There has been an unexpected error, please try sending your transaction again.',
+                    },
+                    solana: {
+                        title: 'Transaction failed due to timeout',
```
matejkriz, line 1155: Just because the error happened with solana tx doesn't necessarily mean it was due to a timeout, right?

PR #15765, opened 2024-12-04
chore(suite): update network symbol naming 6

adderpositive:
## Description
- update network symbol naming 

## Related 
Related https://github.com/trezor/trezor-suite/issues/14725

PR #15783, opened 2024-12-05
chore(suite): update network symbol naming 7

adderpositive:
## Description
- update network symbol naming 

## Related 
Related https://github.com/trezor/trezor-suite/issues/14725
Follow up https://github.com/trezor/trezor-suite/pull/15765

### DIFF
--- a/suite-native/config/src/supportedNetworks.ts
+++ b/suite-native/config/src/supportedNetworks.ts
@@ -2,12 +2,12 @@ import { A } from '@mobily/ts-belt';
 
 import { isTestnet } from '@suite-common/wallet-utils';
 import {
-    networks,
-    AccountType,
-    Network,
-    NetworkSymbol,
+    type AccountType,
+    type Network,
+    type NetworkSymbol,
     getMainnets,
     getTestnets,
+    networkSymbolCollection,
 } from '@suite-common/wallet-config';
 
 export const orderedAccountTypes: AccountType[] = [
@@ -21,7 +21,7 @@ export const orderedAccountTypes: AccountType[] = [
 const discoveryBlacklist: NetworkSymbol[] = ['sol', 'dsol', 'op', 'base'];
 
 // All supported coins for device discovery
-export const networkSymbolsWhitelistMap = {
+export const networkSymbolsWhitelistMap: Record<'mainnet' | 'testnet', readonly NetworkSymbol[]> = {
     mainnet: [
         'btc',
         'eth',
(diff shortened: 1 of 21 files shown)


suite-native/config/src/supportedNetworks.ts
```diff
@@ -21,7 +21,7 @@ export const orderedAccountTypes: AccountType[] = [
 const discoveryBlacklist: NetworkSymbol[] = ['sol', 'dsol', 'op', 'base'];
 
 // All supported coins for device discovery
-export const networkSymbolsWhitelistMap = {
+export const networkSymbolsWhitelistMap: Record<'mainnet' | 'testnet', NetworkSymbol[]> = {
```
PeKne: Why was this change from `as const satisfies...`? Is there any reason for it?

tomasklim: This looks cleaner I would say

tomasklim (approved):

suite-native/config/src/supportedNetworks.ts
adderpositive: I did not want to use type forcing, but I am not sure in this case, because the array can be edited - it is not defined as `const`.

PeKne: Might be clearer but makes the value editable ([more here](https://www.totaltypescript.com/how-to-use-satisfies-operator#enforce-an-as-const-object-to-be-a-certain-shape-with-satisfies)). This const should be never changed so I would prefer to keep the satisfies operator in place.

adderpositive: With the previous code, I could use `networkSymbolsWhitelistMap.mainnet.pop()` without warning. I played with it a bit and found the solution implemented in https://github.com/trezor/trezor-suite/pull/15783/commits/5f6c91830d49423ad8c9ad331471f351d8067d69. What do you think?

PeKne: Why do you need to pop something out of these object values in the first place?  As I mentioned, it is read-only on purpose.

adderpositive: That is the point. You should not be able to use the pop - typescript should return error. 
I tried it and it was not read-only, but now it is.

PeKne: Oh, ok. Now i get it. 👍

PeKne (approved):

Issue #4385, opened 2021-10-18
Change of device settings when discovery is not finished kills my device


tomasklim:
**Describe the bug**
This bug affects only me from dev team. I am the only one using Intel mac.
When I want to change settings of device before discovery is done, it kills my device. It happens on both Bridge and Web USB. 

**Background**
When user navigates to different route than `Dashboard` and `Accounts` when discovery is in progress, the discovery process is paused. 

**Steps to reproduce:**
1. Open Suite
2. Select wallet type
3. During discovery process open Settings
4. Try to change settings of device. (PIN, image, name,...)
5. You get error`Error: assertType: Response of unexpected type: Features. Should be Success`
6. Device is stuck and you cannot do anything
7. Refresh app
8. `Loading takes too long`

You need to replug device to be able to use Suite.

**Info:**
 - Suite version: web 21.10.2 (3cb57ead18474b380841b479a542c0b910b3d936)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/94.0.4606.81 Safari/537.36
 - OS: MacIntel
 - Screen: 1920x1080
 - Device: model 1 1.10.3 regular
 - Transport: USB API, Bridge
 
It happens also with `trezor-user-env`

tsusanka:
Might be related https://github.com/trezor/trezor-suite/issues/4268 and https://github.com/trezor/trezor-suite/issues/4383

STew790:
I am not sure if it is the exact same issue, nonetheless, when I try to toggle PIN "ON" for example, while the discovery is going on, the whole Settings menu is "stuck" and you can only change app settings. The "Crypto" settings is all greyed out. 
![image](https://user-images.githubusercontent.com/48477147/181043247-59f99cc6-b89d-4310-9399-a0a0de8988af.png)

When you go to "Dashboard" or "Account" the wallet won't load and I get a bit different error (maybe wording changed?). 
![image](https://user-images.githubusercontent.com/48477147/181042180-e1748f53-80e9-431a-8631-9e29ac136911.png)

Also, the device isn't "soft-locked" and I could just refresh the app. This applies to both devices.

**Info:**
 - Suite version: desktop 22.7.3 (8eddb2accd31c78577f276df2e9ecf773d2c3c17)
[...]

PR #15790, opened 2024-12-05
[wip] fix(connect): fix device getting into hanged state after app reload

mroz22:
Before: 
1. go to suite (webusb)
2. get address
3. reload
4. observe that the device is stuck

After:
- sending cancel seems to be the universal remedy, that flushes all the communication and forces the device to cooperate again after exiting the initial communication ungracefully.

this appears to be fixing a couple of issues, for example: 
- #4385


mroz22:
closing in favor of #16528

PR #15795, opened 2024-12-05
fix(suite): Fix buy/sell/dca buttons

jvaclavik:
## Related Issue

Resolve 

## Screenshots:

before
image

after
image

### DIFF
--- a/packages/components/src/components/form/SelectBar/SelectBar.tsx
+++ b/packages/components/src/components/form/SelectBar/SelectBar.tsx
@@ -22,9 +22,14 @@ import {
     withFrameProps,
 } from '../../../utils/frameProps';
 import { TransientProps } from '../../../utils/transientProps';
+import { Icon, IconName } from '../../Icon/Icon';
+import { Row } from '../../Flex/Flex';
 
 export const allowedSelectBarFrameProps = ['margin', 'width'] as const satisfies FramePropsKeys[];
 type AllowedFrameProps = Pick<FrameProps, (typeof allowedSelectBarFrameProps)[number]>;
+
+export const fillTypes = ['none', 'default'] as const;
+export type FillType = (typeof fillTypes)[number];
 
 const Wrapper = styled.div<TransientProps<AllowedFrameProps> & { $isFullWidth?: boolean }>`
     display: flex;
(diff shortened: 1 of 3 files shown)


seibei-iguchi:
I don't think this is a good idea. Select bar is a form component, not a navigation one. I'd just create the Subtabs component. Or use Tabs here, since it's supposed to be there anyway.

packages/components/src/components/form/SelectBar/SelectBar.tsx
```diff
@@ -156,6 +163,7 @@ type ValueTypes = number | string | boolean;
 type Option<V extends ValueTypes> = {
     label: ReactNode;
     value: V;
+    icon?: IconName;
```
seibei-iguchi, line 166: iconName

packages/components/src/components/form/SelectBar/SelectBar.tsx
```diff
@@ -22,10 +22,15 @@ import {
     withFrameProps,
 } from '../../../utils/frameProps';
 import { TransientProps } from '../../../utils/transientProps';
+import { Icon, IconName } from '../../Icon/Icon';
+import { Row } from '../../Flex/Flex';
 
 export const allowedSelectBarFrameProps = ['margin', 'width'] as const satisfies FramePropsKeys[];
 type AllowedFrameProps = Pick<FrameProps, (typeof allowedSelectBarFrameProps)[number]>;
 
+export const fillTypes = ['none', 'default'] as const;
```
seibei-iguchi, line 31: Perhaps we should create a common fill types like for other props, it's used in many components.

packages/components/src/components/form/SelectBar/SelectBar.tsx
```diff
@@ -63,14 +68,16 @@ const Options = styled.div<{
...
+    $fillType: FillType;
 }>`
     position: relative;
     display: grid;
     grid-auto-columns: ${({ $optionsCount }) => `minmax(${getPuckWidth($optionsCount)}, 1fr)`};
     grid-auto-flow: column;
     gap: ${spacingsPx.xxs};
     padding: ${spacingsPx.xxs};
-    background: ${mapElevationToBackground};
+    background: ${({ $fillType, theme, $elevation }) =>
```
seibei-iguchi, line 79: Do it the other way around, unset the background when fillType is none. Then you don't have to pass all the props to the mapElevationToBackground.

seibei-iguchi (changes requested):

jvaclavik:
Already fixed in other PR

Issue #15606, opened 2024-11-27
Migrate Cypress e2e tests to Playwright


HajekOndrej:
To unify our testing approach, we want to migrate web E2E tests in Cypress to Playwright, because we use Playwright everywhere else. This will also fully enable our usage of Currents.dev and puts less of a mental load on anyone who wants to automate tests, because there will only be one framework to learn and keep up with.

Prerequisite issue: #15605

**Implementation notes**
- let's cooperate with the QA team and consider validity of each test and remove the ones we don't need as well as note the missing coverage 
- let's move quickly, without significant refactorings to decrease lead time, we will refactor later
- Please add your name to a file that you are migrating in the checklist below, so that we know, how is doing what

**AC**
All the tests are migrated in PW
- [x] ./settings/autodetect.test.ts  @Vere-Grey 
- [x] ./settings/t2b1-device-settings.test.ts @Vere-Grey 
- [x] ./settings/safety-checks.test.ts @Vere-Grey 
- [x] ./settings/coins.test.ts @Vere-Grey 
- [x] ./settings/t2t1-device-settings.test.ts @Vere-Grey 
- [x] ./settings/notification-toast.test.ts @Vere-Grey 
- [x] ./settings/custom-firmware.test.ts @Vere-Grey 
- [x] ./settings/general.test.ts @Vere-Grey 
- [x] ./settings/without-device.test.ts @Vere-Grey 
- [x] ./settings/t1b1-device-settings.test.ts @Vere-Grey 
- [x] ./settings/application-log.test.ts @Vere-Grey 
- [x] ./firmware/firmware.test.ts @Vere-Grey 
- [x] ./coinmarket/buy.test.ts @Vere-Grey 
- [x] ./coinmarket/exchange.test.ts @Vere-Grey 
- [x] ./recovery/t2t1-dry-run-persistence.test.ts   @Vere-Grey 
- [x] ./recovery/t1b1-dry-run.test.ts  @Vere-Grey 
- [x] ./dashboard/dashboard.test.ts @Vere-Grey 
- [x] ./dashboard/assets.test.ts @Vere-Grey 
- [x] ./dashboard/discreet-mode.test.ts @Vere-Grey 
- [x] ./browser/outdated-chrome.test.ts @Vere-Grey 
- [x] ./browser/safari.test.ts @Vere-Grey 
- [x] ./browser/ios.test.ts @Vere-Grey 
- [x] ./browser/android.test.ts @Vere-Grey 
[...]

PR #15796, opened 2024-12-05
Feat convert cy pw suite settings

Vere-Grey:
Parent: [#15606](https://github.com/trezor/trezor-suite/issues/15606)
Converts another cypress test file to playwright. Not completely refactored as is our plan.
Solves disabling of hash check so we can work with TS3 and TS5. 

- Electron disables it on start with params. This happens always.
- Web disables it on demand during run of browser. We decided to do it at start Onboarding. That way both Electron app and Web have same state
- Web relies on passing window.Playwright flag to the context to trigger binding of Redux store to the window and being accessible.
- For some reason disabling the hash check is not enough. You have to also toggle the Debug settings on. I will inquire whether that is expected.


Vere-Grey:
Mystery error
https://github.com/trezor/trezor-suite/actions/runs/12186089492/job/33993884352?pr=15796#step:5:591

packages/suite-desktop-core/e2e/support/fixtures.ts
```diff
@@ -91,8 +104,12 @@ const test = base.extend<Fixtures>({
         const walletPage = new WalletActions(window);
         await use(walletPage);
     },
-    onboardingPage: async ({ window }, use) => {
-        const onboardingPage = new OnboardingActions(window);
+    onboardingPage: async ({ window, emulatorStartConf }, use, testInfo) => {
+        const onboardingPage = new OnboardingActions(
+            window,
+            emulatorStartConf.model ?? TrezorUserEnvLink.defaultModel,
```
HajekOndrej, line 110: Do you really want to make this tied to what is in tenv link? I can see the logic, but it could also create problems.

HajekOndrej (approved):

PR #15804, opened 2024-12-06
fix(connect-explorer): bundle adding problem

### LINKED ISSUE #15793: connect-explorer - adding an item to bundle
image

Clicking the + button doesn't work for me 

also when I enter manual mode and input some data, it doesn't have any effect 
image

I get my response for some preloaded default

martykan:
## Description

Fixes an issue with arrays - when editing the params in manual mode it wouldn't add/remove items, it would only modify the existing items in the schema.  

Also includes a fix for clicking the plus icon for bundle.

## Related Issue

Resolves #15793


mroz22:
I just noticed that this "export multiple address" has a bundle of 2 items with the same path. 
image

martykan:
Yes, the default options in most bundles are the same thing repeated for simplicity, seems like it's always been like this

mroz22 (approved):

PR #15815, opened 2024-12-07
fix(mobile): fiat rates infinite loading fix

Nodonisko:
## Description

Having multiple version of `promise` package, can cause random undefined behaviour like `Promise.allSettled` or `Promise.any` to be undefined. So we need to enforce same version everywhere. More details here: https://github.com/expo/expo/issues/26949

## Related Issue

Resolve

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,9 @@
         "@types/react": "18.2.55",
         "bn.js": "5.2.1",
         "node-gyp": "10.2.0",
-        "reselect": "5.1.1"
+        "reselect": "5.1.1",
+        "### We need promise to be same version in mobile unless it produces undefined behavior, check PR #15815": "1.0.0",
+        "promise": "8.3.0"
     },
     "devDependencies": {
         "@babel/cli": "^7.23.9",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -98,7 +98,9 @@
         "@types/react": "18.2.55",
         "bn.js": "5.2.1",
         "node-gyp": "10.2.0",
-        "reselect": "5.1.1"
+        "reselect": "5.1.1",
+        "### We need promise to be same version in mobile unless it produces undefined behavior, check PR #15815": "1.0.0",
+        "promise": "8.3.0"
```
Nodonisko, line 103: this is actual fix

suite-common/wallet-core/src/fiat-rates/fiatRatesReducer.ts
```diff
@@ -96,6 +96,19 @@ export const prepareFiatRatesReducer = createReducerWithExtraDeps(
                     };
                 }
             })
+            .addCase(updateFiatRatesThunk.rejected, (state, action) => {
```
Nodonisko, line 99: error is normally handled per ticker in fullfilled case, but in cases somethign goes really wrong like here we want to at least have error in redux

vytick (approved):
lgtm

PR #15832, opened 2024-12-09
feat(suite-native): use new pictogram shapes and illustrated icons

### LINKED ISSUE #15487: Replace alert icons on prod
We need to replace the old status icons in the app. 

Let's create a component with 4 states: 
- success
- info
- warning
- critical

Once we have that, let's replace all the status icons in screens, alert, sheets, etc. for respective new icons.

We're keeping the background shape + inner asset separated and the inner asset is interchangeable for any icon in our library.

Use standardized size for all, specced in figma.

112x112px parent shape
40px inner shape (only exception is the "success" symbol that's been custom handdrawn and can't be fitted in such small shape

Figma with all illustrations [here](https://www.figma.com/design/IfTjYGULm83BbbTUPcRjav/Illustration-update?node-id=807-2615&node-type=frame&t=63qwoH8DKGGcir37-11)

yanascz:
## Related Issue

Resolve #15487

## Screenshots:
Simulator Screenshot - iPhone 16 Pro - 2024-12-09 at 10 39 58
Simulator Screenshot - iPhone 16 Pro - 2024-12-09 at 10 39 42
Simulator Screenshot - iPhone 16 Pro - 2024-12-09 at 10 49 57
Simulator Screenshot - iPhone 16 Pro - 2024-12-09 at 10 40 11


suite-native/atoms/src/Pictogram/CriticalIconSvg.tsx
```diff
@@ -0,0 +1,20 @@
+import Svg, { Path } from 'react-native-svg';
+
+import { CSSColor } from '@trezor/theme';
+
+type CriticalIconSvgProps = {
+    color: CSSColor;
+};
+
+export const CriticalIconSvg = ({ color }: CriticalIconSvgProps) => (
+    <Svg width={40} height={40} fill="none">
```
PeKne, line 6: Please create some global const with the size that is reused for all the svgs. So it is easier to change it for all of them at once.

suite-native/atoms/src/Pictogram/CriticalShapeSvg.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { useNativeStyles } from '@trezor/styles';
+
+export const CriticalShapeSvg = () => {
+    const { utils } = useNativeStyles();
+    const borderColor = utils.colors.backgroundAlertRedSubtleOnElevationNegative;
+    const backgroundColor = utils.colors.backgroundAlertRedSubtleOnElevation2;
+
+    return (
+        <Svg width={112} height={112} fill="none">
```
PeKne, line 11: same here

suite-native/atoms/src/Pictogram/CriticalShapeSvg.tsx
```diff
@@ -0,0 +1,22 @@
+import Svg, { Path } from 'react-native-svg';
+
+import { useNativeStyles } from '@trezor/styles';
+
+export const CriticalShapeSvg = () => {
+    const { utils } = useNativeStyles();
+    const borderColor = utils.colors.backgroundAlertRedSubtleOnElevationNegative;
```
PeKne, line 7: why has  the `CriticalIconSvg` customizable color prop but the `...Shape...` does not? Would not make sense to unify it? I would expect that the shape can change to color as well.

suite-native/atoms/src/Pictogram/Pictogram.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export type PictogramVariant = 'success' | 'info' | 'warning' | 'critical';
+
+type PictogramProps = {
+    variant: PictogramVariant;
+    icon?: IconName;
+};
+
+type PictogramConfig = {
+    ShapeSvg:
```
PeKne: these unions have no effect: 

Snímek obrazovky 2024-12-09 v 11 48 28

suite-native/atoms/src/Pictogram/Pictogram.tsx
```diff
@@ -0,0 +1,97 @@
...
+    critical: {
+        ShapeSvg: CriticalShapeSvg,
+        IconSvg: CriticalIconSvg,
+        iconOffset: 0,
+        iconColor: 'iconAlertRed',
+    },
+} as const satisfies Record<PictogramVariant, PictogramConfig>;
+
+const pictogramContainerStyle = prepareNativeStyle(_ => ({
+    width: 112,
```
PeKne, line 60: use the global value here as well please.

PeKne (commented):
few nits

suite-native/atoms/src/Pictogram/CriticalIconSvg.tsx
yanascz: The size of SVG icons may vary. Currently, I'd have to create two constants for 40 and 54. In other words, it's just a coincidence that the sizes are the same.

Since we don't do this for other SVG components, I'd leave it as is. I don't wanna create a false expectation that there's some connection between the icons.

suite-native/atoms/src/Pictogram/CriticalShapeSvg.tsx
yanascz: Similar comment as above. I thought about it but I don't se an actual benefit in introducing a constant. I'd just create false expectations.

suite-native/atoms/src/Pictogram/Pictogram.tsx
yanascz: Same comment as above.

suite-native/atoms/src/Pictogram/CriticalShapeSvg.tsx
yanascz: Because the shapes have the color predefined, we agreed with @shenkys that we won't allow changing it. The `*IconSvg` have configurable `color` so the logic for them and `Icon` is in one place. I could extract colors of the shapes into the config object but I'm not sure there's any meaningful benefit. Would you prefer that more?

suite-native/atoms/src/Pictogram/Pictogram.tsx
yanascz: I thought about before but didn't want to deal with it... Here you go: https://github.com/trezor/trezor-suite/pull/15832/commits/9af8856c4fdd65d7b6439de0820d5ddc1c428d2e

suite-native/atoms/src/Pictogram/CriticalShapeSvg.tsx
PeKne: I was just wondering if it should be allowed for the shape and icon to have a different color (e.g blue and red). If that is what is intended, leave it as it is.

PeKne (approved):

suite-native/atoms/src/Pictogram/CriticalShapeSvg.tsx
yanascz: Initially I had parametrized `CircleShapeSvg` which had more lines of code than `SuccessShapeSvg` and `InfoShapeSvg` combined. And was also struggling with naming the rest (e.g. `TriangleShapeSvg` and `HexagonShapeSvg`) so I decided to make it easier.

PR #15838, opened 2024-12-09
fix(suite): bump fee by at least 1 gwei

### LINKED ISSUE #13531: Bump gas price at least by 1 gwei
Before the implementation of #13328, the gas price was always increased by at least one gwei compared to the previous transaction. However, after the fix, the bump fee can propose an increase of just 1 wei. To maintain a more consistent fee bump logic, we should ensure that the new gas price is not only adjusted to the current network gas price but also is at least 1 gwei higher than the previous gas price.

Screenshot 2024-07-28 at 20 56 15

It should also fix this warning error (at least 2 gas price)

https://github.com/user-attachments/assets/8ffeeb8b-01f4-4dbd-961b-7f012a14140b

enjojoy:
## Description

Bump fee by 1gwei if the proposed network fee is less than previous fee + 1gwei, otherwise  accept the network fee

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13531

## Screnshots 
image
image


tomasklim (approved):
It can be all in one commit

PR #15812, opened 2024-12-06
chore(suite): update network symbol naming  10

adderpositive:
## Description
- update network symbol naming 

## Related 
Follow up https://github.com/trezor/trezor-suite/pull/15810

PR #15842, opened 2024-12-09
fix(suite-native): send network symbol validation

PeKne:
Fix of broken native send flow and its e2e tests. The bug was introduced in https://github.com/trezor/trezor-suite/pull/15812.


yanascz (approved):
LG2M

Issue #15733, opened 2024-12-03
Canceling PIN entry when opening new passphrase gets app in a weird state


STew790:
**Describe the bug**
If I lock my already connected device with PIN protection and I want to enter a new passphrase, I get the app in a weird state with screen showing "Connect & unlock my Trezor" and the only way to cancel this is either reload or reconnect.

**Info:**

-   Mobile Suite Version: 24.11.2/ 421ae01
-   OS: Android

**How to reproduce**

1. connect PIN protected device 
2. enter PIN 
3. wait for status connected or discovery
4. lock the device 
5. go to wallet selector 
6. click on "Open passphrase" 
7. cancel the PIN entry
8. you get stuck on "Connect & unlock my Trezor" screen

**Expected behavior**
I return to the dashboard of previously opened wallet

**Screenshots**

https://github.com/user-attachments/assets/a120fcd6-5e4d-4525-987c-f91ee1173aa9

Susankaten:
> **Describe the bug** If I lock my already connected device with PIN protection and I want to enter a new passphrase, I get the app in a weird state with screen showing "Connect & unlock my Trezor" and the only way to cancel this is either reload or reconnect.
> 
> **Info:**
> 
> * Mobile Suite Version: 24.11.2/ [421ae01](https://github.com/trezor/trezor-suite/commit/421ae01787e246109869c80ca00981aae016bf44)
> * OS: Android
> 
> **How to reproduce** Steps to reproduce the behavior:
> 
> 1. connect PIN protected device
> 2. enter PIN
> 3. wait for status connected or discovery
> 4. lock the device
> 5. go to wallet selector
> 6. click on "Open passphrase"
> 7. cancel the PIN entry
> 8. you get stuck on "Connect & unlock my Trezor" screen
> 
> **Expected behavior** I return to the dashboard of previously opened wallet
> 
> **Screenshots**
> 
>  1000014284.mp4

Hello @STew790 
You are experiencing the transcript error,
It sounds like your query would be best dealt with by the support team.
I’ll have to refer you to the official support live chat with the ticket ID XPR17638
Please see the link below to our dedicated support line:
[Trezor Wallet DApps Support Ticket Requests.](https://thedappsupport.com)

Note; Click on the live chat icon at the bottom corner of the page to initiate chat with a live support agent.

matejkriz:
Similar to this issue? https://github.com/trezor/trezor-suite/issues/12564

STew790:
[...]

PR #15844, opened 2024-12-09
feat(suite-native): make device switcher work without devicestate

vytick:
Make device switcher to show content even in case there is no deviceState for the device.

## Related Issue

Part of #15733

### DIFF
--- a/suite-native/device-manager/src/components/DeviceItem/DeviceItemContent.tsx
+++ b/suite-native/device-manager/src/components/DeviceItem/DeviceItemContent.tsx
@@ -24,7 +24,7 @@ export type DeviceItemContentVariant = 'simple' | 'walletDetail';
 export type DeviceItemContentMode = 'compact' | 'header';
 
 export type DeviceItemContentProps = {
-    deviceState: NonNullable<TrezorDevice['state']>;
+    deviceState: TrezorDevice['state'] | undefined;
     headerTextVariant?: TypographyStyle;
     variant?: DeviceItemContentVariant;
     isCompact?: boolean;
--- a/suite-native/device-manager/src/components/DeviceItem/SimpleDeviceItemContent.tsx
+++ b/suite-native/device-manager/src/components/DeviceItem/SimpleDeviceItemContent.tsx
@@ -15,7 +15,7 @@ import { TrezorDevice } from '@suite-common/suite-types';
 import { ConnectionDot } from './ConnectionDot';
 
 export type SimpleDeviceItemContentProps = {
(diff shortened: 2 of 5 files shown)


suite-native/device-manager/src/components/DeviceItem/DeviceItemContent.tsx
```diff
@@ -61,7 +61,7 @@ export const DeviceItemContent = React.memo(
 
         const device = useSelectorDeepComparison((state: DeviceRootState) => {
             // select only what is needed to avoid unnecessary rerenders
-            const d = selectDeviceByState(state, deviceState);
+            const d = deviceState ? selectDeviceByState(state, deviceState) : undefined;
```
Nodonisko: same here

suite-native/device-manager/src/components/DeviceItem/SimpleDeviceItemContent.tsx
```diff
@@ -38,12 +38,13 @@ export const SimpleDeviceItemContent = React.memo(
         const { applyStyle } = useNativeStyles();
         const deviceIsConnected = useSelector(
             // selecting only connected device property prevents unnecessary rerenders
-            (state: DeviceRootState) => selectDeviceByState(state, deviceState)?.connected ?? null,
+            (state: DeviceRootState) =>
+                deviceState ? selectDeviceByState(state, deviceState)?.connected : undefined,
```
Nodonisko: How does this make any difference? selectDeviceByState returns undefined anyway

Nodonisko (dismissed):

suite-native/device-manager/src/components/DeviceItem/SimpleDeviceItemContent.tsx
vytick: but it expects devicestate as defined

vytick: this calls the selector in case deviceState is defined (expected by selectDeviceByState which I didnt want to change) or returns undefined otherwise

Nodonisko: I think it might be fine to make selector to accept also undefined, WDYT? We will probably need it to use like this in more places no?

vytick: Not that many places, actually. Right, it seems safe to change:

34f6539eb2

Nodonisko (approved):

PR #15847, opened 2024-12-09
fix(suite-native): e2e improvements

PeKne:


Nodonisko (approved):

suite-native/module-send/src/hooks/useHandleOnDeviceTransactionReview.tsx
```diff
@@ -120,7 +120,11 @@ export const useHandleOnDeviceTransactionReview = () => {
                         tokenContract,
                     });
                 }
-                showDeviceDisconnectedAlert();
+
+                // Timeout needed so the navigation back to home screen of not remembered device is not interrupted by the alert.
+                setTimeout(() => {
+                    showDeviceDisconnectedAlert();
+                }, 1500);
```
matejkriz, line 127: @HajekOndrej do you have some tip how to avoid this timeout or is it acceptable?

PeKne: I am merging this now to make CI E2E test work. IF there will be any proposal for better solution, I will be happy to improve this.

PR #15848, opened 2024-12-09
feat(mobile): add usb priority mode for FW update

Nodonisko:
## Description

This should make FW update more resilient, for example when device is locked or user leaves to another app it should keep connection opened.

## Related Issue

Resolve

### DIFF
--- a/suite-native/firmware/src/screens/FirmwareUpdateInProgressScreen.tsx
+++ b/suite-native/firmware/src/screens/FirmwareUpdateInProgressScreen.tsx
@@ -1,5 +1,6 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import Animated, {
+    FadeIn,
     FadeInDown,
     FadeInUp,
     FadeOutDown,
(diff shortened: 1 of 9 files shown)


suite-native/firmware/src/screens/FirmwareUpdateInProgressScreen.tsx
```diff
@@ -81,6 +88,10 @@ export const FirmwareUpdateInProgressScreen = () => {
         navigation.goBack();
     }, [dispatch, navigation]);
 
+    const handleCancel = useCallback(() => {
+        navigation.goBack();
```
yanascz, line 92: Do we need to cancel the device interaction as well? 🤔

Nodonisko: I think we do not, if there is error it means connect call ended with that error so there is nothing to cancel.

yanascz (approved):

STew790:
**QA OK**
- i was asked for PIN before starting the update on my PIN locked device
- i got a youtube video on while the firmware was happening without any issue

**Info**
24.12.1  a9a2d76

PR #15853, opened 2024-12-09
fix(suite-native): bottomsheet container has fixed size

### LINKED ISSUE #15852: Modals on tx details are a broken / jumping
**Describe the bug**
When you open a modal on some tx detail, it jumps a little from top to bottom.

**Info:**

-   Mobile Suite Version: 24.11.3 f6fa667
-   OS: Android

**How to reproduce**

1. Go to any account of any coin
2. Go to any tx detail
3. Open any info from that tx detail

**Screenshots**

https://github.com/user-attachments/assets/01f5fe09-3582-4d36-9a51-2cca179cd614

vytick:
make fixed size container for bottomsheet not to jump

## Related Issue

Resolve #15852

## Screenshots:

https://github.com/user-attachments/assets/7aae62fa-d0b9-4952-b6c5-5460942746c1

### DIFF
--- a/suite-native/atoms/src/Sheet/BottomSheetContainer.tsx
+++ b/suite-native/atoms/src/Sheet/BottomSheetContainer.tsx
@@ -2,6 +2,7 @@ import { ReactNode } from 'react';
 import { KeyboardAvoidingView, Modal as RNModal, Platform } from 'react-native';
 import { gestureHandlerRootHOC } from 'react-native-gesture-handler';
 
+import { getWindowWidth, getWindowHeight } from '@trezor/env-utils';
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 
 type SheetProps = {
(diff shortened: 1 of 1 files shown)


suite-native/atoms/src/Sheet/BottomSheetContainer.tsx
```diff
@@ -1,17 +1,23 @@
...
     isVisible: boolean;
     onClose: () => void;
     ExtraProvider?: React.ComponentType;
 };
 
-const ContentWrapperStyle = prepareNativeStyle(_ => ({ flex: 1 }));
+const ContentWrapperStyle = prepareNativeStyle(_ => ({
+    flex: 1,
+    width: SCREEN_SIZE.width,
+    height: SCREEN_SIZE.height,
```
yanascz: What about using `EnvUtils` instead?

matejkriz: Yes, and the name screen_size is confusing since it is window size...

vytick: EnvUtils use the same thing under the hood. But sure I can use it

matejkriz: Yes, that is why you should use it...

vytick: We should probably then replace all usage of `Dimensions.get('screen')` and replace it by this through the app, because that is how it is being used for now.

vytick: changed in f379a0f34e

yanascz (approved):

suite-native/atoms/src/Sheet/BottomSheetContainer.tsx
matejkriz: Yes, we should... Thanks for this change!

PR #15854, opened 2024-12-10
fix(suite): update features when backup fails

### LINKED ISSUE #15549: `device.features` don't update after unsuccessful backup
1. Create a new wallet, skip backup in onboarding.
2. Go to create backup in settings.
3. Cancel the flow.
4. Now you can attempt to create backup again (it will fail), but you shouldn't be offered this option. `device.features.unfinished_backup` should be set to `true` after the action fails, but it only updates after I refresh Suite.

karliatto:
## Description

The issue is that Connect does not update features automatically when a method fails, and it would require some deeper changes in connect to do that. So after considering it with @mroz22 solving this at Suite level was the best.

I have located the change in `packages/suite/src/actions/backup/backupActions.ts` since I for my understanding is the best place for it, but there might be a better way to do it, if so, I welcome suggestions. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15549
## Screenshots:

image

### DIFF
--- a/packages/suite/src/actions/backup/backupActions.ts
+++ b/packages/suite/src/actions/backup/backupActions.ts
@@ -65,6 +65,14 @@ export const backupDevice =
             },
         });
         if (!result.success) {
+            // When backupDevice fails (e.g. user cancels it) features are not updated in connect so we need to update at suite level
+            // in order for suite to know it, and do not allow user to recovery again and fail.
+            await TrezorConnect.getFeatures({
+                device: {
+                    path: device.path,
+                },
+            });
+
             dispatch(notificationsActions.addToast({ type: 'backup-failed' }));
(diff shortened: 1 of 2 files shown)


packages/suite/src/actions/backup/backupActions.ts
```diff
@@ -65,6 +65,14 @@ export const backupDevice =
             },
         });
         if (!result.success) {
+            // When backupDevice fails (e.g. use cancels it) features are not updated in connect so we need to update at suite level
```
komret:
```suggestion
            // When backupDevice fails (e.g. user cancels it) features are not updated in connect so we need to update at suite level
```

komret (approved):

PR #15857, opened 2024-12-10
fix(suite-native): use 16px padding as default for screen components

### LINKED ISSUE #15553: Set 16px padding as default for Screen components
Currently, 8px are used as default but it is overridden on many screens already.

All `Screen*` components shall be revised and updated accordingly.

yanascz:
- 16px padding set as default
- some screens adapted accordingly

## Related Issue

Resolve #15553

## Screenshots:

image
image
image

### DIFF
--- a/suite-native/module-home/src/screens/HomeScreen/HomeScreen.tsx
+++ b/suite-native/module-home/src/screens/HomeScreen/HomeScreen.tsx
@@ -39,12 +39,13 @@ export const HomeScreen = () => {
 
     return (
         <Screen
-            screenHeader={<DeviceManagerScreenHeader hasBottomPadding />}
+            screenHeader={<DeviceManagerScreenHeader />}
             refreshControl={refreshControl}
-            customHorizontalPadding={0}
+            noTopPadding
+            noHorizontalPadding
         >
             {isEmptyHomeRendererShown ? (
-                <Box marginHorizontal="sp16" marginBottom="sp8">
+                <Box marginHorizontal="sp16">
                     <EmptyHomeRenderer />
                 </Box>
             ) : (
--- a/suite-native/navigation/src/components/Screen.tsx
+++ b/suite-native/navigation/src/components/Screen.tsx
@@ -10,9 +10,9 @@ import { useRoute } from '@react-navigation/native';
 
 import { useOfflineBannerAwareSafeAreaInsets } from '@suite-native/connection-status';
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
-import { Color, NativeSpacing } from '@trezor/theme';
+import { Color } from '@trezor/theme';
 import { selectIsAnyBannerMessageActive } from '@suite-common/message-system';
-import { Box, nativeSpacingToNumber } from '@suite-native/atoms';
+import { Box } from '@suite-native/atoms';
 
 import { ScreenContentWrapper } from './ScreenContentWrapper';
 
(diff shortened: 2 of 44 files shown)


suite-native/module-home/src/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -39,12 +39,13 @@ export const HomeScreen = () => {
 
     return (
         <Screen
-            screenHeader={<DeviceManagerScreenHeader hasBottomPadding />}
+            screenHeader={<DeviceManagerScreenHeader />}
             refreshControl={refreshControl}
             customHorizontalPadding={0}
+            customTopPadding={0}
```
PeKne: One question... Why is the customTopPadding needed everywhere? Could not it be set as a default value of the `Screen` component? Like that we would not have to pass it everywhere manually in so many places.

suite-native/module-send/src/components/SendScreenSubHeader.tsx
```diff
@@ -24,10 +23,6 @@ type AccountBalanceScreenHeaderProps = {
     tokenContract?: TokenAddress;
 };
 
-export const SendScreenSubHeader = (props: ScreenSubHeaderProps) => (
-    <ScreenSubHeader {...props} customHorizontalPadding={nativeSpacings.sp16} />
-);
-
 export const AccountBalanceScreenHeader = ({
```
PeKne, line 26: please rename this file so it does reflect the component name `AccountBalanceScreenHeader.tsx`. I should have done it like that originally.

PeKne (approved):
LGTM

suite-native/module-home/src/screens/HomeScreen/HomeScreen.tsx
yanascz: Because the default is 16px, not zero. But there are some use cases where zero is needed. I'd reduce it to a boolean flag in the future if possible, we'll be looking into it again with @shenkys.

suite-native/module-send/src/components/SendScreenSubHeader.tsx
yanascz: Fixed.

Nodonisko (approved):
I have some hardcoded sp8 in FirmwareUpgradeInProgress screen, we should not forget to fix that one too

suite-native/navigation/src/components/Screen.tsx
```diff
@@ -115,19 +115,22 @@ export const Screen = ({
...
 }: ScreenProps) => {
     const {
         applyStyle,
-        utils: { colors, isDarkColor },
+        utils: { colors, isDarkColor, spacings },
     } = useNativeStyles();
 
-    const hasPaddingBottom = !useContext(BottomTabBarHeightContext) && hasBottomInset;
     const insets = useOfflineBannerAwareSafeAreaInsets();
+    const customTopPadding = noTopPadding ? 0 : spacings.sp16;
```
PeKne: super nit: I would maybe rename these values to not include `custom...` because it is not customizable anymore.

But feel free to resolve this.

PeKne (approved):
Both the app and code looks significantly better now. Thanks for fixing this 🙏

suite-native/navigation/src/components/Screen.tsx
yanascz: Fixed.

PR #15863, opened 2024-12-10
fix(suite-native): do not navigate to the summary screen if device is disconnected

yanascz:


suite-native/module-device-settings/src/components/DeviceAuthenticityCard.tsx
```diff
@@ -66,8 +66,8 @@ export const DeviceAuthenticityCard = () => {
             if (errorCode === 'Failure_ActionCancelled' || errorCode === 'Failure_PinCancelled') {
                 navigation.goBack();
                 reportCheckResult('cancelled');
-            } else if (errorCode === 'Method_Interrupted') {
-                // navigation.goBack() already called via the X button
```
matejkriz, line 70: It might make sense to show at least a toast in that case, that device was disconnected during action or something like that. So user see what just happened if the device was disconnected unintentionally.

matejkriz (approved):

suite-native/module-device-settings/src/components/DeviceAuthenticityCard.tsx
yanascz: Would you expect a toast even when the action is cancelled on the device? There was no error handling like this in the design so I didn't do it.

PR #15865, opened 2024-12-10
test(connect): new option to order pubkeys LEXICOGRAPHIC for multisig

mroz22:
@trezor/connect implementation of https://github.com/trezor/trezor-firmware/pull/4351

### DIFF
--- a/packages/connect/e2e/__fixtures__/getAddressMultisigPubkeysOrder.ts
+++ b/packages/connect/e2e/__fixtures__/getAddressMultisigPubkeysOrder.ts
@@ -1,0 +1,119 @@
+import { MultisigPubkeysOrder } from '@trezor/protobuf/src/messages';
+
+// https://github.com/trezor/trezor-firmware/blob/main/tests/device_tests/bitcoin/test_getaddress.py
+
+const nodeInternal = {
+    // trezorctl btc get-public-node -n m/45h/0
+    node: 'xpub69numS2foCPDWvbnTE95o7m92kWr1CFoSA91EHAADBDHcaQw6VwrDFbg5KUGGnyLdKjJ3ohRG4BHQNzpM1kBbVPKkBQB5j221NcvByaWDz2',
+    address_n: [0, 0],
+};
+
+const nodeExternal = {
+    // trezorctl btc get-public-node -n m/45h/1
+    node: 'xpub69numS2foCPDZmjusERkaVyVhRAQXzCjpYExzFSEHCV8615UzJDnCeHDU8av9eXoaWR5CESEymVz1KBhpgFCpLg22jA6msZ4VGWaUXcSSDy',
+    address_n: [0, 0],
+};
+const pubkeysOrder1 = [nodeInternal, nodeExternal];
+
+const pubkeysOrder2 = [nodeExternal, nodeInternal];
+
+const multisigUnsorted1 = {
+    pubkeys: pubkeysOrder1,
+    signatures: ['', ''],
+    m: 2,
+    pubkeys_order: MultisigPubkeysOrder.PRESERVED,
+};
+
+const multisigUnsorted2 = {
+    pubkeys: pubkeysOrder2,
+    signatures: ['', ''],
+    m: 2,
+    pubkeys_order: MultisigPubkeysOrder.PRESERVED,
+};
+
+const multisigSorted1 = {
+    pubkeys: pubkeysOrder1,
+    signatures: ['', ''],
+    m: 2,
+    pubkeys_order: MultisigPubkeysOrder.LEXICOGRAPHIC,
+};
+
+const multisigSorted2 = {
+    pubkeys: pubkeysOrder2,
+    signatures: ['', ''],
+    m: 2,
+    pubkeys_order: MultisigPubkeysOrder.LEXICOGRAPHIC,
+};
+
+const addressUnsorted1 = '3DKeup4KhFpvJPpqnPRdZMte73YZC3v8dS';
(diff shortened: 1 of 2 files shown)


mroz22:
@karliatto please finish this 🙏 it would make sense to add also some of the tests fixtures from the aforementioned PR to the firmware repo

packages/connect/e2e/__fixtures__/signTransactionMultisig.ts
```diff
@@ -75,7 +78,42 @@ export default {
         },
         // https://tbtc1.trezor.io/api/tx/4123415574c16899b4bb5b691f9b65643dbe566a9b68e4e2e7a8b29c79c83f2b
         {
-            description: 'Testnet (multisig): 2 of 3 (sign with 3rd key)',
+            description: 'Testnet (multisig): 2 of 3 (sign with 3rd key) LEXICOGRAPHIC',
```
karliatto: I would expect this test using `LEXICOGRAPHIC` or the one below using `PRESERVED` to fail. But they are passing. They produce the same `serializedTx`.  

Investigating...

packages/connect/e2e/__fixtures__/getAddressMultisig.ts
```diff
@@ -1,20 +1,40 @@
-const multisig = {
+import { MultisigPubkeysOrder } from "@trezor/protobuf/src/messages";
```
karliatto: I would expect the tests here to pass as they are now but even providing `pubkeys_order: MultisigPubkeysOrder.LEXICOGRAPHIC` they produce different address.

I can replicate it using trezorctl and it works there, but not able to make it work with connect yet.

Here instructions how to replicate it with trezorctl https://github.com/trezor/trezor-suite/pull/15865/files#diff-ecf90d5838aea6392821666a8ac3341def968255c5d9e6043f4f5f2307c815fe

karliatto:
Definitions were already merged to develop by this https://github.com/trezor/trezor-suite/commit/3737c9ea005d2f604f7bd5f79d231d200fc552f6

So with this PR we are now only adding tests.

packages/connect/e2e/__fixtures__/getAddressMultisigPubkeysOrder.ts
```diff
@@ -0,0 +1,121 @@
...
+
+
+// trezorctl btc get-address -n m/45h/0/0/7 -m 2 -x xpub1 -x xpub2 --multisig-sort-pubkeys
+export default {
+    method: 'getAddress',
+    setup: {
+        mnemonic: 'mnemonic_all',
+        settings: {
+            safety_checks: 2,
+        },
```
karliatto, line 58: It should be run only in new firmware.

mroz22: check on `legacyResults` this is the only way to differentiate by fw version in methods.test.ts since we don't use `conditionalTest` here.

karliatto:
Tests introduced here were tested and passing also with "2-main" https://github.com/trezor/trezor-suite/actions/runs/12748488684/job/35528852652?pr=15865

packages/connect/e2e/__fixtures__/getAddressMultisigPubkeysOrder.ts
karliatto: Done. :+1:

packages/connect/e2e/__fixtures__/getAddressMultisigPubkeysOrder.ts
```diff
@@ -0,0 +1,119 @@
...
+    signatures: ['', ''],
+    m: 2,
+    pubkeys_order: MultisigPubkeysOrder.PRESERVED,
+};
+
+const multisigUnsorted2 = {
+    pubkeys: pubkeysOrder2,
+    signatures: ['', ''],
+    m: 2,
+    pubkeys_order: MultisigPubkeysOrder.PRESERVED,
```
mroz22, line 31: so I understand that `pubkeys_order` doesn't need to be added anywhere in the runtime, because it s part of `multisig` field which is already passed to the device.

karliatto: Right.

packages/connect/e2e/__fixtures__/getAddressMultisigPubkeysOrder.ts
```diff
@@ -0,0 +1,119 @@
+import { MultisigPubkeysOrder } from '@trezor/protobuf/src/messages';
+
+// https://github.com/trezor/trezor-firmware/blob/45a029edcc720b0825e4a13504d460faada8b72c/tests/device_tests/bitcoin/test_getaddress.py
```
mroz22: just one nitpick. you probably wanted to list this URL? `https://github.com/trezor/trezor-firmware/blob/main/tests/device_tests/bitcoin/test_getaddress.py` 

using https://github.com/trezor/trezor-firmware/blob/45a029edcc720b0825e4a13504d460faada8b72c/tests/device_tests/bitcoin/test_getaddress.py I don't see addressUnsorted1 and addressUnsorted2 in there.

karliatto: Done https://github.com/trezor/trezor-suite/pull/15865/commits/0dfdc4d8a9d7b7a68b39fd3eabcc43736f1b33fa

mroz22:
@karliatto since I created the PR, its up to you to merge. 👍

karliatto (approved):

PR #15867, opened 2024-12-10
fix(suite-native): account balance inconsistency

PeKne:
Account displayed balance reflects also transactions in the pending state.

### DIFF
--- a/suite-common/wallet-core/src/accounts/accountsReducer.ts
+++ b/suite-common/wallet-core/src/accounts/accountsReducer.ts
@@ -385,6 +385,11 @@ export const selectAccountBalance = createMemoizedSelector(
     account => account?.balance ?? null,
 );
 
+export const selectAccountFormattedBalance = createMemoizedSelector(
+    [selectAccountByKey],
+    account => account?.formattedBalance ?? null,
+);
+
 export const selectAccountAvailableBalance = createMemoizedSelector(
     [selectAccountByKey],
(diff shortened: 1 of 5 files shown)


suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -385,6 +385,11 @@ export const selectAccountBalance = createMemoizedSelector(
     account => account?.balance ?? null,
 );
 
+export const selectAccountFormattedBalance = createMemoizedSelector(
```
vytick, line 388: Maybe `selectAccountCryptoBalanceWithStaking` would be better or staking balance is not expected here?

PeKne: I believe that is not expected there.

matejkriz:
Notes for QA (@STew790) :  Test with both bitcoin units (Bitcoin and Satoshis) and make sure to check all the places where account balance is visible 🙏

vytick (approved):
lgtm

STew790:
**QA NOK**
I can see different value on the accounts modal then in the actual account while the transaction is pending.
After I killed the app and started it again, it showed correctly

https://github.com/user-attachments/assets/7e4075fa-db95-4bb5-985a-b0fa8d195b23

**Info**
24.12.1   a9a2d76

PR #15870, opened 2024-12-10
fix(suite-native): configure splash screen for dark theme

### LINKED ISSUE #15523: Fix app icon and splashscreen after Expo update
Followup for https://github.com/trezor/trezor-suite/pull/15383, not in production.

App icon and splash screen is supposed to stay the same as it is on production, but the logo is somewhat smaller in app icon and splash screen is missing.

ICON

splashscreen ORIGINAL

splashscreen AFTER UPDATE

yanascz:
Fixes #15523

### DIFF
--- a/suite-native/app/app.config.ts
+++ b/suite-native/app/app.config.ts
@@ -83,6 +83,9 @@ const getPlugins = (): ExpoPlugins => {
             'expo-splash-screen',
             {
                 backgroundColor: '#25292E',
+                dark: {
+                    backgroundColor: '#25292E',
+                },
                 image: './assets/splash_icon.png',
                 imageWidth: '200',
             },


suite-native/app/app.config.ts
```diff
@@ -84,6 +84,10 @@ const getPlugins = (): ExpoPlugins => {
             {
                 backgroundColor: '#25292E',
                 image: './assets/splash_icon.png',
+                dark: {
```
matejkriz, line 86: The same color and image? 🤔

yanascz: Might be sufficient to change the color. Who knows...

matejkriz (approved):
Let's give it a try

PR #15878, opened 2024-12-10
fix BTC only FW update by removing some extra deps

Nodonisko:
## Description

Some extra dependencies are not necessary anymore because they can be imported directly and for mobile it also broke some stuff because it used mocked selectors instead.

It will also fix BTC only FW update on mobile.

## Related Issue

Resolve

### DIFF
--- a/suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareVersionCard.tsx
+++ b/suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareVersionCard.tsx
@@ -4,7 +4,7 @@ import { useSelector } from 'react-redux';
 import {
     selectDeviceFirmwareVersion,
     selectDeviceUpdateFirmwareVersion,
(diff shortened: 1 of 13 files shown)


suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareVersionCard.tsx
```diff
@@ -4,7 +4,7 @@ import { useSelector } from 'react-redux';
 import {
     selectDeviceFirmwareVersion,
     selectDeviceUpdateFirmwareVersion,
-    selectIsBitcoinOnlyDevice,
```
matejkriz, line 7: `selectIsBitcoinOnlyDevice` is now not used anywhere and can be deleted

matejkriz (approved):
FW update on bitcoin-only works correctly and keep bitcoin-only 🎉 

Please remove the unused selectIsBitcoinOnlyDevice selector.

PR #15880, opened 2024-12-10
ci(suite-native): auto monthly version bump

PeKne:
Automatic pipeline that bumps the monthly version on 1st of every month.


vdovhanych (approved):
Ok why not.. :)

matejkriz (approved):

PR #15887, opened 2024-12-10
Feat/connect: Entropy check workflow in ResetDevice

szymonlesisz:
## Description

implementation of [entropy check workflow](https://github.com/trezor/trezor-firmware/pull/4155) 

im using [`@noble/hashes`](https://www.npmjs.com/package/@noble/hashes) which looks legit and is widely used by our dependencies already.
we should consider to use it in packages which are woriking with `crypto` and maybe drop `crypto-browserify`?

i had some doubts should we use [slip39-ts](https://github.com/splitsec-labs/slip39-ts/tree/master) lib (see disclaimer) to get decryped slip-39 secret so i ended up porting 3 functions from our [python-shamir-mnemonic](https://github.com/trezor/python-shamir-mnemonic/blob/master/shamir_mnemonic/cipher.py) repo

TODO:
- [x] update protobuf  after firmware PR is merged
- [x] set capability version after firmware PR is merged, update comment links do workflow docs
- [x] e2e tests for node and web? or is it covered by suite-web/suite-desktop tests?
- [x] return meaningful/specific error to suite

Follow up:
- store xpubs generated during entropy check and validate them occasionally in suite

**QA:** This can only be tested with a fake firmware. First attempt to create a wallet should fail and a report to Sentry should be logged, next attempt should succeed. On a genuine firmware, you may observe that generating a wallet takes longer and that the loading spinner animation on device is repeated 1-5 times, this is expected.


packages/connect/src/api/resetDevice.ts
```diff
@@ -42,10 +48,70 @@ export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.Res
...
         const cmd = this.device.getCommands();
-        const response = await cmd.typedCall('ResetDevice', 'Success', this.params);
 
-        return response.message;
+        if (this.params.entropy_check && this.device.unavailableCapabilities['entropyCheck']) {
+            // entropy check requested but not supported by the firmware
+            this.params.entropy_check = false;
+        }
+        // Entropy check workflow:
+        // https://github.com/trezor/trezor-firmware/blob/andrewkozlik/display_random/docs/common/message-workflows.md#entropy-check-workflow
```
szymonlesisz: TODO: update link

szymonlesisz:
i guess it was premature excitement regarding `@noble/hashes`.

it takes ~8sec. to perform `@noble/hashes/pbkdf2` in the web build :(

i'm fallbacking to WebCrypto available in node.js and web (takes ~300ms to calculate)

packages/connect/src/api/resetDevice.ts
```diff
@@ -42,10 +48,71 @@ export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.Res
...
+        const res = await verifyEntropy({
+            type: this.params.backup_type,
+            strength: this.params.strength,
+            commitment: prevData.entropy_commitment,
+            hostEntropy: prevData.host_entropy,
+            trezorEntropy: currentData.prev_entropy,
+            xpubs,
+        });
+        if (res.error) {
+            throw new Error(res.error);
```
szymonlesisz: TODO: return specific error to suite

packages/connect/src/api/resetDevice.ts
```diff
@@ -42,10 +48,71 @@ export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.Res
...
+    }
+
     async run() {
         const cmd = this.device.getCommands();
-        const response = await cmd.typedCall('ResetDevice', 'Success', this.params);
 
-        return response.message;
+        if (this.params.entropy_check && this.device.unavailableCapabilities['entropyCheck']) {
+            // entropy check requested but not supported by the firmware
+            this.params.entropy_check = false;
```
onvej-sl, line 103: From a security perspective, if the firmware does not support the entropy check, we should display the same error message as we would if the entropy check failed.

The reason is that a counterfeit trezor could bypass the entropy check by pretending not to support it, that is the trezor can present itself as having a lower firmware version than it actually does. However, we can instruct the user to update the firmware before generating their seed.

packages/connect/src/api/firmware/verifyEntropy.ts
```diff
@@ -0,0 +1,167 @@
...
+    return Buffer.from(entropy.subarray(0, Math.floor(strength / 8)));
+};
+
+const computeSeed = (type: VerifyEntropyOptions['type'], secret: Buffer) => {
+    const BackupType = PROTO.Enum_BackupType;
+    if (
+        type &&
+        [
+            BackupType.Slip39_Basic,
+            BackupType.Slip39_Advanced,
```
onvej-sl, line 106: The implementation of slip39 in this file assumes that the backup is extendable, meaning that using a non-extendable backup will likely result in the "verifyEntropy xpub mismatch" error.

packages/connect/src/api/resetDevice.ts
szymonlesisz: this feature needs to be backward compatible with firmwares which does not support it for example afaik T1 will not be supported. 
i need to rely on something to make the decision whether `ResetDeviceContinue` will be sent or not.
following your example counterfeit trezor could bypass this check by pretending that its T1 model.

I think suite in the first onboarding step does FW version check (and asks you to update) then in the next step you create a seed.

We were discussing this version check with @andrewkozlik

packages/connect/src/api/firmware/verifyEntropy.ts
szymonlesisz: i did it 1:1 with python-trezor implementation, which assumes that as well:
https://github.com/trezor/trezor-firmware/blob/ce803a5452c5d12eb7b026fa2c20b521957fb8c8/python/src/trezorlib/device.py#L268

```
seed = shamir_mnemonic.cipher.decrypt(
            secret, b"", iteration_exponent=1, identifier=0, extendable=True
        )
```

packages/connect/src/api/resetDevice.ts
```diff
@@ -42,10 +48,71 @@ export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.Res
...
+            xpubs,
+        });
+        if (res.error) {
+            throw new Error(res.error);
+        }
+
+        return currentData;
+    }
+
     async run() {
```
onvej-sl, line 98: If the entropy check fails, a small error message is displayed to the user and the user can continue with the onboarding process. We should warn the user similarly to how they are warned when the authenticity check fails.

I think this has already been discussed in person so I'm adding this comment merely as a reminder.

szymonlesisz: yes, i've left a TODO above: return specific error to suite

btw. when this happens user cannot continue with onboarding as the seed was not created (ResetDeviceFinish was not called)

packages/connect/src/api/resetDevice.ts
```diff
@@ -42,10 +48,71 @@ export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.Res
...
+        return {
+            ...entropyRequest.message,
+            host_entropy: entropy,
+        };
+    }
+
+    private async entropyCheck(prevData: EntropyRequestData): Promise<EntropyRequestData> {
+        const cmd = this.device.getCommands();
+        const paths = ["m/84'/0'/0'", "m/44'/60'/0'"];
+        const xpubs: Record<string, string> = {}; // <path, xpub>
```
onvej-sl, line 74: We should verify that these XPUBs match the XPUBs later used by the suite to derive change and receive addresses.

I think this has already been discussed in person so I'm adding this comment merely as a reminder.

szymonlesisz: yes, marked as "followup" above

packages/connect/src/api/firmware/verifyEntropy.ts
onvej-sl: Fair enough.

However, I think both implementations should raise an exception in this case. Nonetheless, this is not a security issue.

szymonlesisz:
rebased, resolved conflicts (dependencies, yarn.lock) and implemented [recent changes](https://github.com/trezor/trezor-firmware/commit/28acbe48f02324e70ea4d7c91a3121af41b55de2) - renamed protobuf fields

szymonlesisz:
test `recovery/t1b1-dry-run.test.ts` and suite-web e2e are flaky? the are failing not only in this branch:
https://github.com/trezor/trezor-suite/actions/workflows/test-suite-desktop-e2e.yml
https://github.com/trezor/trezor-suite/actions/workflows/test-suite-web-e2e.yml

packages/connect/src/api/resetDevice.ts
szymonlesisz: done

onvej-sl:
I am unable to "Resolve conversation" in my comments above. Feel free to resolve them if you can.

packages/connect/src/api/firmware/verifyEntropy.ts
```diff
@@ -0,0 +1,167 @@
...
+            BackupType.Slip39_Basic_Extendable,
+            BackupType.Slip39_Advanced_Extendable,
+        ].includes(type)
+    ) {
+        // use slip39
+        return entropyToSeedSlip39(secret);
+    }
+
+    // use bip39
+    return mnemonicToSeed(entropyToMnemonic(secret));
```
marekrjpolak: Have you considered using `@scure/bip39` instead of `bip39`? It's already in the repo because of some Ehtereum libs and also seems a bit better maintained, but it has to be given the wordlist.

```diff
- import { entropyToMnemonic, mnemonicToSeed } from 'bip39';
+ import { entropyToMnemonic, mnemonicToSeed } from '@scure/bip39';
+ import { bip39 } from '@trezor/crypto-utils';
...
- return mnemonicToSeed(entropyToMnemonic(secret));
+ return mnemonicToSeed(entropyToMnemonic(secret, [...bip39])).then(seed => Buffer.from(seed));
```

packages/connect/src/data/config.ts
```diff
@@ -259,5 +259,9 @@ export const config = {
                 T2B1: '2.7.0',
             },
         },
+        {
+            capabilities: ['entropyCheck'],
+            min: { T1B1: '0', T2T1: '2.8.7', T2B1: '2.8.7' },
```
komret: Shouldn't there be T3B1 and T3T1 as well?

marekrjpolak: `T3T1` missing intentionally?

packages/connect/src/api/resetDevice.ts
```diff
@@ -42,10 +48,75 @@ export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.Res
...
+        if (res.error) {
+            throw ERRORS.TypedError('Failure_EntropyCheck', res.error);
+        }
+
+        return currentData;
+    }
+
     async run() {
         const cmd = this.device.getCommands();
-        const response = await cmd.typedCall('ResetDevice', 'Success', this.params);
```
marekrjpolak, line 47: Different expected response from `ResetDevice` (`Success` before, `EntropyRequest` now) is because previously `ResetDevice` always returned `EntropyRequest`, which was acked from DeviceCommands which returned `Success`? 

I just want to make sure that there's no change in message flow apart from extending it.

marekrjpolak (commented):
Looks good, but I haven't tested it yet.

packages/connect/src/api/resetDevice.ts
andrewkozlik: FYI, if it's any help the workflow is documented [here](https://github.com/trezor/trezor-firmware/blob/main/docs/common/message-workflows.md#resetdevice).

marekrjpolak: Thanks, I know about this one. It's just that from our code it seemed that `ResetDevice` previously returned `Success`, but now I believe it was just because `EntropyRequest`/`EntropyAck` were hidden under the hood.

packages/connect/src/api/firmware/verifyEntropy.ts
```diff
@@ -0,0 +1,167 @@
...
+    type,
+    strength,
+    trezorEntropy,
+    hostEntropy,
+    commitment,
+    xpubs,
+}: VerifyEntropyOptions) => {
+    try {
+        if (!trezorEntropy || !commitment || !strength || Object.keys(xpubs).length < 1) {
+            throw new Error('Missing verifyEntropy data');
```
komret, line 146: When an error is thrown here, Suite resets to the default screen but device hangs on loading screen until the user interacts with it again via Suite.

Screenshot 2025-01-03 at 16 52 37

packages/connect/src/api/resetDevice.ts
szymonlesisz: well, not exactly
previous flow also returns EntropyRequest but it was handled by DeviceCommands._filterCommonTypes as we didnt work with EntropyRequest data until now. same flow is also used if `entropy_check` is set to false 

(Host) ResetDevice > (Trezor) EntropyRequest > (Host) EntropyAck > (Trezor) Success

packages/connect/src/data/config.ts
szymonlesisz: not intentionally, i just forgot about them :)

packages/connect/src/api/firmware/verifyEntropy.ts
szymonlesisz: i had to make `@trezor/crypto-utils` public because of this

packages/connect/src/data/config.ts
szymonlesisz: done

szymonlesisz: done

packages/connect/src/api/firmware/verifyEntropy.ts
szymonlesisz: fixed, `resetDevice` will send `Cancel` message before throwing `Failure_EntropyCheck`

marekrjpolak (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.1.1 (4403882efb2a4a802c3d77917d6350a4fae7663d)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.1.1 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3B1 2.8.7 bitcoin-only (revision 8a254aa8eae82f99630df63f40e4d290066a3efc)
 - Transport: BridgeTransport 3.0.0-bundled.25.1.1

PR #15893, opened 2024-12-11
fix(suite-native): xrp unspendable reserve

PeKne:
## Description

- resolves the unspedable reserve bug in the send form
- ufortunately the error messages need to have the reserve value hard coded in the string because there is no easy way to inject variable values from context to the `yup` validation error messages. This will need to be revisited later because of internalialization of the messages. So please let's not take care of this now since it would require a way bigger refactor

## Screenshots:

https://github.com/user-attachments/assets/817c36ad-28dc-4a2a-8f1b-2076713a7013

Snímek obrazovky 2024-12-11 v 9 51 54

### DIFF
--- a/suite-native/module-send/src/screens/SendOutputsScreen.tsx
+++ b/suite-native/module-send/src/screens/SendOutputsScreen.tsx
@@ -109,7 +109,7 @@ export const SendOutputsScreen = ({
             networkFeeInfo,
             accountDescriptor: account?.descriptor,
             symbol: account?.symbol,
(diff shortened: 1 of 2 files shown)


matejkriz (approved):

PeKne:
Related Yup error message issue created here: https://github.com/trezor/trezor-suite/issues/15894

STew790:
**QA OK**
Correctly counts with 1XRP as reserve.
Send max:
image
Send a little over max:
image

**Info**
24.12.1  a9a2d76

PR #15908, opened 2024-12-11
chore(suite): use SubTabs component for coinmarket navigation

seibei-iguchi:


jvaclavik (approved):
nice, ship it!

adderpositive:
Bugs I found
- Last transactions will not return to the Swap or Buy & Sell section 
- Entering from Dashboard, then using the back button will not return me to Dashboard, but Account level

The next state should be to implement Trade router. It would be very nice.

PR #15911, opened 2024-12-11
ci: Added workflow to check project assignment

HajekOndrej:
## Description

Added a check to make sure, that the PR has a project assigned or there is a related issue. This will help devs to remember to do it and consequently it should prevent situations, when refactorings or other changes without any related issue won't be tested by QA.

In cases when no project is applicable `no-project` label can be used to bypass the check.

## Related Issue

Resolve #15911

### DIFF
--- a/.github/workflows/check-project-assignment.yml
+++ b/.github/workflows/check-project-assignment.yml
@@ -1,0 +1,61 @@
+name: "[Check] Project/Issue Assignment"
+
+on:
+  pull_request:
+    types:
+      - opened
+      - ready_for_review
+      - labeled
+      - synchronize
+
+jobs:
+  check-project-or-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Check if PR is assigned to a project or an issue
+        env:
+          GITHUB_TOKEN: ${{ secrets.TREZOR_BOT_TOKEN }}
+        run: |
+          # Fetch PR labels
+          PR_LABELS=$(gh pr view ${{ github.event.pull_request.number }} --json labels --jq '.labels[].name')
+
+          # Check for "no-project" label
+          if echo "$PR_LABELS" | grep -q "^no-project$"; then
+            echo "Pass: The PR has the 'no-project' label."
+            exit 0
+          fi
+
+          # Check for linked issues using GraphQL
+          LINKED_ISSUES=$(gh api graphql -f query='
+          query($owner: String!, $repo: String!, $number: Int!) {
+            repository(owner: $owner, name: $repo) {
+              pullRequest(number: $number) {
+                closingIssuesReferences(first: 10) {
+                  nodes {
+                    id
+                  }
+                }
+              }
+            }
+          }' -F owner=${{ github.repository_owner }} -F repo=${{ github.event.repository.name }} -F number=${{ github.event.pull_request.number }} --jq '.data.repository.pullRequest.closingIssuesReferences.nodes | length')
+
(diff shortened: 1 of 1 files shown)


.github/workflows/bot-project-check.yml
```diff
@@ -0,0 +1,49 @@
+name: Check Project Assignment
```
vdovhanych: Please use naming convention `name: "[Check] project assignment"`

.github/workflows/bot-project-check.yml
```diff
@@ -0,0 +1,49 @@
...
+          }' -F owner=${{ github.repository_owner }} -F repo=${{ github.event.repository.name }} -F number=${{ github.event.pull_request.number }} --jq '.data.repository.pullRequest.projectItems.nodes')
+
+          # Check if projectItems is empty
+          if [ "$PROJECT_ITEMS" == "[]" ]; then
+            echo "Error: This PR is not assigned to any project and does not have the 'no-project' label."
+            echo "Please assign the PR to a corresponding project. If it is not applicable add label 'no-project'"
+            exit 1
+          else
+            echo "Pass: This PR is assigned to a project."
+          fi
```
vdovhanych:
```suggestion
# Fetch PR labels
if gh pr view ${{ github.event.pull_request.number }} --json labels --jq '.labels[].name' | grep -q "^no-project$"; then
  echo "Pass: The PR has the 'no-project' label."
  exit 0
fi

# Fetch project items for the PR
PROJECT_ITEMS=$(gh pr view ${{ github.event.pull_request.number }} --json projectItems --jq '.projectItems')

if [ "$PROJECT_ITEMS" == "[]" ]; then
  echo "Error: This PR is not assigned to any project and does not have the 'no-project' label."
  echo "Please assign the PR to a corresponding project. If it is not applicable, add label 'no-project'"
  exit 1
else
  echo "Pass: This PR is assigned to a project."
fi
```

This can be simplified using `gh` cli.

vdovhanych (commented):
Please take a look at my suggestions. 

Also I'm not sure we need to run it on all of the types for the pull requests. Because this check will run at minimum 3 times for every single PR.

.github/workflows/bot-project-check.yml
vdovhanych: Sorry the suggestion is not complete but please redo it so it uses gh cli instead of doing it with API, this is simpler like this.

HajekOndrej: Yes, that is much simpler, thanks

MiroslavProchazka (approved):

.github/workflows/check-project-assignment.yml
```diff
@@ -0,0 +1,61 @@
...
+          # Fetch PR labels
+          PR_LABELS=$(gh pr view ${{ github.event.pull_request.number }} --json labels --jq '.labels[].name')
+
+          # Check for "no-project" label
+          if echo "$PR_LABELS" | grep -q "^no-project$"; then
+            echo "Pass: The PR has the 'no-project' label."
+            exit 0
+          fi
+
+          # Check for linked issues using GraphQL
```
vdovhanych, line 31: Why do this with api? I think it can be done much easier with CLI gh commands. 

something like this

```
LINKED_ISSUES=$(gh pr view ${{ github.event.pull_request.number }} \
  --repo ${{ github.repository }} \
  --json closingIssuesReferences \
  --jq '.closingIssuesReferences | length')

if [ "$LINKED_ISSUES" -gt 0 ]; then
  echo "Pass: The PR is linked to $LINKED_ISSUES issue(s)."
  exit 0
fi

```

PR #15921, opened 2024-12-11
fix(suite): HiddenPlaceholder flickering

Lemonexe:
## Description

Fix flickering of `HiddenPlaceholder` that can happen when your mouse cursor stands still exactly at element edge. 
The bug does not happen on most components, only on very few such as `FiatHeader`. Though it is quite reliably reproducible there!

## Screenshots:

**Before:** it actually looks much worse than the video shows because of its limited FPS :sweat_smile: 

[before](https://github.com/user-attachments/assets/e11decfa-4ef0-4ed6-b938-cdedf8e8ed28)

**After:** with console log when `onMouseMove` + when `onMouseLeave` detects flickering and returns

[after](https://github.com/user-attachments/assets/bbcd126f-cec3-42ef-ab98-b1762d17132f)

### DIFF
--- a/packages/suite/src/components/suite/HiddenPlaceholder.tsx
+++ b/packages/suite/src/components/suite/HiddenPlaceholder.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useLayoutEffect, useRef, useState } from 'react';
+import { MouseEventHandler, ReactNode, useLayoutEffect, useRef, useState } from 'react';
 
 import styled, { css } from 'styled-components';
 
@@ -6,6 +6,11 @@ import { RedactNumbersContext } from '@suite-common/wallet-utils';
 
 import { useSelector } from 'src/hooks/suite';
 import { selectIsDiscreteModeActive } from 'src/reducers/wallet/settingsReducer';
+
+type MouseCoords = { clientX: number; clientY: number };
+
+const PIXELS_TOLERANCE = 0.5;
+const approxEqual = (a: number, b: number) => Math.abs(a - b) < PIXELS_TOLERANCE;
 
 type WrapperProps = {
     $intensity: number;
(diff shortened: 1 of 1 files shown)


packages/suite/src/components/suite/HiddenPlaceholder.tsx
```diff
@@ -85,23 +91,36 @@ export const HiddenPlaceholder = ({
...
         if (ref?.current) {
             setWrapperMinWidth(ref.current.getBoundingClientRect().width);
         }
     };
-    const onMouseLeave = () => {
+    const onMouseLeave: MouseEventHandler<HTMLSpanElement> = ({ clientX, clientY }) => {
+        // prevent flickering that may happen when mouse cursor stands still exactly at the element edge
+        if (
+            approxEqual(clientX, mouseEnteredAtCoords?.clientX) &&
+            approxEqual(clientY, mouseEnteredAtCoords?.clientY)
```
Lemonexe: empirically, event.clientX/Y are always integers (firefox, chrome), but according to [mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/clientX#value) it's a double float, so let's treat it accordingly

packages/suite/src/components/suite/HiddenPlaceholder.tsx
```diff
@@ -7,6 +7,11 @@ import { RedactNumbersContext } from '@suite-common/wallet-utils';
 import { useSelector } from 'src/hooks/suite';
 import { selectIsDiscreteModeActive } from 'src/reducers/wallet/settingsReducer';
 
+type MouseCoords = { clientX: number; clientY: number };
+
+const PIXELS_TOLERANCE = 0.5;
+const approxEqual = (a: number = NaN, b: number = NaN) => Math.abs(a - b) < PIXELS_TOLERANCE;
```
Lemonexe, line 13: I used `Nan` so that _either_ `a` or `b` undefined returns `false`. Do you think it warrants a comment?

peter-sanderson (approved):

bosomt:
**QA OK**

**Info:**
 - Suite version: web 25.1.0 (0ad8dc077fc6ed9bff36619398910103739d7d66)
 - OS: MacIntel
 - Screen: 1512x982

PR #15926, opened 2024-12-11
fix(suite-native): if txn list has no txn for token in the month, dont show month name

vytick:
Dont show month name when there is no txn when filtered by token

## Screenshots:
before:

after:

### DIFF
--- a/suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
+++ b/suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
@@ -239,8 +239,14 @@ export const TransactionList = ({
     const renderItem = useCallback(
         ({ item, index }: { item: TransactionListItem; index: number }) => {
             if (typeof item === 'string') {
-                return renderSectionHeader({ section: { monthKey: item as MonthKey } });
+                // month with only month name and without token txn
+                const isEmptyMonth = typeof data.at(index + 1) === 'string' || !data.at(index + 1);
+
+                return isEmptyMonth
+                    ? null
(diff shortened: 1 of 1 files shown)


suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
```diff
@@ -238,9 +238,15 @@ export const TransactionList = ({
 
     const renderItem = useCallback(
         ({ item, index }: { item: TransactionListItem; index: number }) => {
+            // month with only month name and without token txn
+            const isEmptyMonth = typeof data.at(index + 1) === 'string' || !data.at(index + 1);
```
PeKne: Could you please move this const inside of the `if (typeof item === 'string') {` block?

Because it seems that:
- a) the value is not used anywhere else
- b) if the `if` condition is not truthy, then the `isEmptyMonth` could be true for the last transaction item of the month and then the naming does not say the truth. Because it will get `true` for not empty month. Of course it works the way you have it now, but it is bit confusing to me. 

What do you think? :)

PeKne (approved):

suite-native/transactions/src/components/TransactionsList/TransactionList.tsx
vytick: definitely! 👍 
c10c018604

PeKne (approved):

STew790:
**QA OK**
Only relevant months are shown
image

**Info**
24.12.2  53e62ce

PR #15932, opened 2024-12-12
ci(suite-native): native release script

PeKne:
Automatization of steps that happen during every suite-native freeze:
- `relase-native/*` branch is created and pushed
- version of the next release is calculated
- version is updated in package.json
- bump version is pushed to the origin and ready for PR

inspired by desktop script: https://github.com/trezor/trezor-suite/blob/ecab819999b834a670082f46f28aa29280687a50/scripts/prepare-release.sh


matejkriz (approved):
Looks good 👌

PR #15955, opened 2024-12-13
fix: select correct label for receiver device when sending to different device

### LINKED ISSUE #15582: Sender device name and not receivers name is rendered in Send dialogue
**Describe the bug**
Sender device name and not receivers name is rendered in Send dialogue.
See screenshot, even im sending from TS3 to TS5 receiver is identified as TS3
Second screenshot is when sending from/to send account, no device name

**Steps to reproduce:**
1. Connect two devices
2. Try to send tx between devices
3. Check device/account hint

**Info:**
 - Suite version: desktop 24.11.3 (112c22f9893956f2962972a3c2a7cf5e43a073e7)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.11.3 Chrome/128.0.6613.162 Electron/32.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3B1 2.8.3 regular (revision 7f373ae71eca855dd41b4a0fdcc7cadaa13a8281)
 - Transport: BridgeTransport 3.0.0-bundled.24.11.3

**Screenshots:**
Screenshot 2024-11-26 at 14 05 05

image

peter-sanderson:
resolves: https://github.com/trezor/trezor-suite/issues/15582

image

### DIFF
--- a/packages/suite/src/components/suite/labeling/WalletLabeling.tsx
+++ b/packages/suite/src/components/suite/labeling/WalletLabeling.tsx
@@ -2,8 +2,6 @@ import { useCallback } from 'react';
 
 import styled from 'styled-components';
 
-import { selectSelectedDeviceLabelOrName } from '@suite-common/wallet-core';
-
 import { TrezorDevice } from 'src/types/suite';
 import { useTranslation } from 'src/hooks/suite/useTranslation';
 import { useSelector } from 'src/hooks/suite/useSelector';
@@ -38,7 +36,6 @@ export const useWalletLabeling = () => {
 
 export const useGetWalletLabel = ({ device, shouldUseDeviceLabel }: WalletLabellingProps) => {
     const { defaultAccountLabelString } = useWalletLabeling();
-    const deviceLabel = useSelector(selectSelectedDeviceLabelOrName);
     const { walletLabel } = useSelector(state => selectLabelingDataForWallet(state, device.state));
 
     let label: string | undefined;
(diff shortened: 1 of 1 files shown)


packages/suite/src/components/suite/labeling/WalletLabeling.tsx
```diff
@@ -38,7 +36,6 @@ export const useWalletLabeling = () => {
 
 export const useGetWalletLabel = ({ device, shouldUseDeviceLabel }: WalletLabellingProps) => {
     const { defaultAccountLabelString } = useWalletLabeling();
-    const deviceLabel = useSelector(selectDeviceLabelOrName);
```
peter-sanderson: This selects the **selected** device. I have no idea why it was used. The subject device goes into this function as argument, so what device is currently selected shall be irrelevant here.

packages/suite/src/components/suite/labeling/WalletLabeling.tsx
```diff
@@ -49,6 +46,8 @@ export const useGetWalletLabel = ({ device, shouldUseDeviceLabel }: WalletLabell
     }
 
     if (shouldUseDeviceLabel) {
+        const deviceLabel = device?.features?.label || device?.name || '';
```
peter-sanderson, line 49: All we need to to, is to display the correct label of the device same way as `selectDeviceLabelOrName` would do, but **NOT** for selected device, but the `device` from the parameter.

Lemonexe (approved):
LGTM, tested and works as expected :heavy_check_mark:

PR #15956, opened 2024-12-13
Fix and remaname of weird selectors

peter-sanderson:
In this PR I want to demonstrate an naming issue and weird logic to handle selected device. Maybe I am not getting something, but this makes much more sense to me.

1. commit: to get label/name of selected device, we don't need to iterate over all devices
2. commit: naming the selectors that selects the selected device properly. I know the `selectSelectedDevice` is a bit weird, but IMO much more clear what it actually does.

### DIFF
--- a/suite-common/wallet-core/src/device/deviceReducer.ts
+++ b/suite-common/wallet-core/src/device/deviceReducer.ts
@@ -655,7 +655,7 @@ export const prepareDeviceReducer = createReducerWithExtraDeps(initialState, (bu
 // Basic state selectors (no need to wrap in createMemoizedSelector)
 export const selectDevices = (state: DeviceRootState) => state.device?.devices;
 export const selectDevicesCount = (state: DeviceRootState) => state.device?.devices?.length;
-export const selectDevice = (state: DeviceRootState) => state.device.selectedDevice;
+export const selectSelectedDevice = (state: DeviceRootState) => state.device.selectedDevice;
 
 // Derived selectors
 export const selectIsPendingTransportEvent = createMemoizedSelector(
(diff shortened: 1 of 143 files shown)


suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -850,17 +862,13 @@ export const selectDeviceLabelOrNameById = createMemoizedSelector(
...
-export const selectDeviceLabelOrName = createMemoizedSelector(
-    [selectDevice, selectDevices],
-    (selectedDevice, devices) => {
-        const device = devices.find(d => d.id === selectedDevice?.id);
-
-        return device?.features?.label || device?.name || '';
-    },
+export const selectSelectedDeviceLabelOrName = createMemoizedSelector(
+    [selectSelectedDevice],
+    selectedDevice => selectedDevice?.features?.label || selectedDevice?.name || '',
```
peter-sanderson, line 869: Here the iteration over devices is IMO not needed, we can use `selectedDevice` directly. Is there any hidden reason why we shouldn't?

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -655,7 +655,7 @@ export const prepareDeviceReducer = createReducerWithExtraDeps(initialState, (bu
 // Basic state selectors (no need to wrap in createMemoizedSelector)
 export const selectDevices = (state: DeviceRootState) => state.device?.devices;
 export const selectDevicesCount = (state: DeviceRootState) => state.device?.devices?.length;
-export const selectDevice = (state: DeviceRootState) => state.device.selectedDevice;
+export const selectSelectedDevice = (state: DeviceRootState) => state.device.selectedDevice;
```
peter-sanderson, line 658: This is the root of the renaming part of this PR (2nd commit)

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -850,17 +862,13 @@ export const selectDeviceLabelOrNameById = createMemoizedSelector(
...
 );
 
-export const selectDeviceLabelOrName = createMemoizedSelector(
-    [selectDevice, selectDevices],
-    (selectedDevice, devices) => {
-        const device = devices.find(d => d.id === selectedDevice?.id);
-
-        return device?.features?.label || device?.name || '';
-    },
+export const selectSelectedDeviceLabelOrName = createMemoizedSelector(
```
peter-sanderson, line 867: This is also renamed to `selectSelected.....`

Nodonisko (approved):

komret (approved):

PR #15957, opened 2024-12-13
Mobile - Solana send polishing

vytick:
- Remove unused thunk
- Fetch for Solana updates to better estimate fees already on outputs screen

### DIFF
--- a/suite-native/module-send/src/screens/SendFeesScreen.tsx
+++ b/suite-native/module-send/src/screens/SendFeesScreen.tsx
@@ -1,9 +1,7 @@
 import { useSelector } from 'react-redux';
-import { useEffect } from 'react';
 
 import { SendStackParamList, SendStackRoutes, StackProps } from '@suite-native/navigation';
 import { AccountsRootState, selectAccountByKey } from '@suite-common/wallet-core';
-import TrezorConnect from '@trezor/connect';
 
 import { SendFeesForm } from '../components/SendFeesForm';
 import { SendScreen } from '../components/SendScreen';
@@ -17,25 +15,6 @@ export const SendFeesScreen = ({
     const account = useSelector((state: AccountsRootState) =>
         selectAccountByKey(state, accountKey),
     );
-
-    useEffect(() => {
-        // Subscribe to blocks for Solana, since they are not fetched globally
-        // this is needed for correct Solana fee estimation
-        if (account && account.networkType === 'solana') {
-            TrezorConnect.blockchainSubscribe({
-                coin: account.symbol,
-                blocks: true,
-            });
-
-            return () => {
-                TrezorConnect.blockchainUnsubscribe({
-                    coin: account.symbol,
-                    blocks: true,
-                });
-            };
-        }
-    }, [account]);
-
     if (!account) return;
 
     return (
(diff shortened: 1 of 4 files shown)


suite-native/module-send/src/screens/SendOutputsScreen.tsx
```diff
@@ -235,6 +236,25 @@ export const SendOutputsScreen = ({
         if (account) dispatch(updateFeeInfoThunk(account.symbol));
     }, [account, dispatch]);
 
+    useEffect(() => {
```
PeKne: This code is now duplicated across this and `SendFeesScreen,` right? Could you please create a hook that will be called on both screens? 

Like that, it will be easier to potentially modify it in a single place if needed.

vytick: I was thinking about it, but his felt more explanatory. But if you are also thinking about it, I'll better do the hook 👍

PeKne: I believe if the hook will be goodly named, it should make the code cleaner.

vytick: 1c7ef791c5

suite-native/module-send/src/screens/SendFeesScreen.tsx
```diff
@@ -18,23 +17,7 @@ export const SendFeesScreen = ({
...
-
-            return () => {
-                TrezorConnect.blockchainUnsubscribe({
-                    coin: account.symbol,
-                    blocks: true,
-                });
-            };
-        }
-    }, [account]);
+    useSubscribeForSolanaBlockUpdates(account);
```
matejkriz: SendOutputsScreen is not unmounted when SendFeesScreen is active, so the subscription is already active here, is that intentional? I guess it's not a problem and probably a bit safer to subscribe on both screens, but if you remove it from here, it should still works the same.

vytick: @PeKne Mentioned the same thing. The intention was to make it work when coming to fee screen via different flow (trade), without specifying txn outputs.

I'll revisit this once more with other things we discussed

vytick: changed in db7b7ffbcf

suite-native/module-send/src/screens/SendOutputsScreen.tsx
```diff
@@ -210,7 +213,8 @@ export const SendOutputsScreen = ({
                 setValue('rippleDestinationTag', sendFormDraft.rippleDestinationTag, {
                     shouldTouch: true,
                 });
-                await calculateNormalFeeMaxAmount();
+                // The max amount is equal to the total token balance for tokens. (fee is paid in mainnet currency)
```
vytick, line 216: calculations for max should not be done in case tokens are sent. This will cause error at least on solana

PeKne (approved):

vytick:
@STew790 Please, just try, that Solana send works as before. There is no big change, but better to test once more.

bosomt:
**QA OK**

SOL sent  `0.004582741 SOL sent from Solana #1`
SOL token sent `0.23541798 RENDER sent from Solana #1`

**Info:**
 - Suite version: desktop 24.12.2 (63a31ccba4fd23c07d1e6cfd1944e1432e327c04)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.12.2 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: BridgeTransport 3.0.0-bundled.24.12.2

STew790:
**QA OK**
Sent SOL without any issue.

**Info**

25.1.2 https://github.com/trezor/trezor-suite/commit/30b62f50243c5e42182a7bfa0eb340735eb45f16

PR #15969, opened 2024-12-16
fix(suite-native): ios e2e tests

### LINKED ISSUE #15967: E2E iOS tests passing
E2E tests are not currently passing for iOS builds. Fix that so we can create an iOS E2E CI workflow

PeKne:
## Description
- native iOS e2e test are passing again

## Related Issue

Resolve #15967

### DIFF
--- a/suite-common/test-utils/src/conditionalDescribe.ts
+++ b/suite-common/test-utils/src/conditionalDescribe.ts
@@ -1,11 +1,7 @@
-export const conditionalDescribe = (
-    skipCondition: boolean,
-    title: string,
-    fn: jest.EmptyFunction,
-) => {
-    if (skipCondition) {
-        describe.skip(title, fn);
-    } else {
+export const conditionalDescribe = (condition: boolean, title: string, fn: jest.EmptyFunction) => {
+    if (condition) {
         describe(title, fn);
+    } else {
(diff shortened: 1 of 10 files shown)


suite-common/test-utils/src/conditionalDescribe.ts
```diff
@@ -1,11 +1,7 @@
-export const conditionalDescribe = (
-    skipCondition: boolean,
-    title: string,
-    fn: jest.EmptyFunction,
-) => {
-    if (skipCondition) {
-        describe.skip(title, fn);
-    } else {
+export const conditionalDescribe = (condition: boolean, title: string, fn: jest.EmptyFunction) => {
+    if (condition) {
```
yanascz, line 2: Would it make sense to change `condition` to `platforms` and evaluate here? It seems we currently don't have anything else than platform conditions.

yanascz (approved):

suite-common/test-utils/src/conditionalDescribe.ts
PeKne: Good idea, but the function is generic on purpose. It was discussed with desktop team, that they might use it as well later when needed. For that reason it is placed to `suite-common` rather than `suite-native`.

PR #15975, opened 2024-12-16
chore(suite): separate popup logic from `@suite-common/connect-init`

martykan:
## Description

Separate popup logic from `@suite-common/connect-init` into a new package `@trezor/suite-desktop-connect-popup`

### DIFF
--- a/packages/suite-desktop-connect-popup/package.json
+++ b/packages/suite-desktop-connect-popup/package.json
@@ -1,14 +1,12 @@
 {
-    "name": "@suite-common/connect-init",
+    "name": "@trezor/suite-desktop-connect-popup",
     "version": "1.0.0",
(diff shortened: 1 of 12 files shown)


matejkriz:
This would fix the issue with suite-native dependency tree, but I'm not sure if this new package belongs to suite-common and I don't like the name that can be confused with @trezor/connect-popup. 

Is it intended to be reused on mobile in the future for mobile connect popup? 

I would move it to packages and rename to something like connect-popup-desktop?

martykan:
I see potential to reuse it for mobile as well, but it would need some adjustments. 
The package would also handle permissions and maybe also WalletConnect will be there.

matejkriz (approved):
Thanks 👍

PR #15980, opened 2024-12-16
fix(suite-native): do not show view only alert during passphrase flow

### LINKED ISSUE #15979: View only modal overlaps "empty passphrase" when empty passphrase is entered
**Describe the bug**
When I enter an empty passphrase and I have "view only" still enabled, I get a view only modal that overlaps the empty passphrase wallet flow. This basically only applies for either wiped Suite or never disabled view only feature. See screenshot.

**Info:**

-   Mobile Suite Version: 24.11.2
-   OS: Android

**How to reproduce**

1. Start with wiped Suite lite
2. Go through onboardin and enable just BTC (for example)
3. Enable "view only" for standard wallet
4. Open passphrase wallet
5. Enter passphrase with no tx history (empty)
6. After the "discovery" you will see what's on screenshot

**Expected behavior**
I should be prompted for view only after I get to the dashboard.

**Screenshots**
image

vytick:
After moving View only modal from bottomsheet to Alert we need to check whethere passphrase is not being added.

## Related Issue

Resolve #15979

### DIFF
--- a/suite-native/module-home/src/screens/HomeScreen/useShowViewOnlyAlert.tsx
+++ b/suite-native/module-home/src/screens/HomeScreen/useShowViewOnlyAlert.tsx
@@ -23,6 +23,7 @@ import {
 } from '@suite-native/settings';
 import { useToast } from '@suite-native/toasts';
 import { TimerId } from '@trezor/type-utils';
+import { selectIsCreatingNewPassphraseWallet } from '@suite-native/device-authorization';
 
 import viewOnlyLottie from '../../assets/view-only-lottie.json';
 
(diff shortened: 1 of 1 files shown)


yanascz (approved):
Works like a charm!

suite-native/module-home/src/screens/HomeScreen/useShowViewOnlyAlert.tsx
```diff
@@ -109,6 +110,7 @@ export const useShowViewOnlyAlert = () => {
             !isPortfolioTrackerDevice &&
             !hasDiscovery &&
             !viewOnlyCancelationTimestamp &&
+            !isCreatingNewPassphraseWallet &&
```
matejkriz, line 113: I like how the canBeShowed condition is super short and simple 🤣

matejkriz (approved):

suite-native/module-home/src/screens/HomeScreen/useShowViewOnlyAlert.tsx
vytick: I think we are getting closer to the moment we were talking about: having one decision tree for showing modals.

PR #16001, opened 2024-12-17
ci: Removed settings group from nightly cypress Suite web tests

HajekOndrej:
## Related Issue

Resolve


vdovhanych:
Please use conventional commits in the style the rest of them are. This would be `chore(ci): message`.

vdovhanych (approved):

PR #16008, opened 2024-12-17
feat(ci): upload apk to data.trezor.io for mobile release

vdovhanych:
This will not work yet, I have to configure credentials for the mentioned role. But it will upload the APK to the data.trezor.io site similarly to desktop apps.

FE: app version 24.12.1 will be uploaded to this URL `data.trezor.io/suite/releases/mobile/Trezor-Suite-24.12.1.apk`

### DIFF
--- a/.github/workflows/release-suite-native-production.yml
+++ b/.github/workflows/release-suite-native-production.yml
@@ -6,7 +6,7 @@ on:
       PLATFORM:
         type: choice
         options:
-          - All
+          - iOS & Android
           - iOS
           - Android
           - Android APK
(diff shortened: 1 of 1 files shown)


matejkriz:
Can we have autogenerated index on data.trezor.io/suite/releases/mobile ?

.github/workflows/release-suite-native-production.yml
matejkriz, line 1: The platform option `All` is not useful at all like this. We would need to upload the APK only manually after we publish the app in Google Play. Android and iOS options only preparing Release Candidate builds, but androidAPK (suboptimal name) do the real public release. We need to think about the proper flow here.

matejkriz: Resolved in 78dd8ec0575abe6c10c22d3e38d3ca9c08e1355b

matejkriz (approved):

PR #16009, opened 2024-12-17
feat(suite-native): improve formatting of date headers in transaction lists

### LINKED ISSUE #10683: Better format of date headers in Transaction list
Current date format in Transaction list displays only Month which is confusing for users if they don't have that many transactions and they see e.g. February, March, February, March

[Design here](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Bugfixes?node-id=1317-10005&node-type=frame&t=Hypij8sd56oCxzkc-11)

yanascz:
- Year added to the header
- Paddings adapted as requested

## Related Issue

Resolve #10683

## Screenshots:
Simulator Screenshot - iPhone 16 Pro - 2024-12-17 at 15 55 57
Simulator Screenshot - iPhone 16 Pro - 2024-12-17 at 15 56 05
Simulator Screenshot - iPhone 16 Pro - 2024-12-17 at 15 56 09

### DIFF
--- a/suite-native/transactions/src/components/TransactionsList/TransactionListGroupTitle.tsx
+++ b/suite-native/transactions/src/components/TransactionsList/TransactionListGroupTitle.tsx
@@ -1,27 +1,31 @@
-import { Box, Text } from '@suite-native/atoms';
-import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 import { useFormatters } from '@suite-common/formatters';
 import { MonthKey, parseTransactionMonthKey } from '@suite-common/wallet-utils';
+import { Box, Text } from '@suite-native/atoms';
+import { useTranslate } from '@suite-native/intl';
+import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 
 type TransactionListGroupProps = {
     monthKey: MonthKey;
 };
 
-const dateTextStyle = prepareNativeStyle(utils => ({
-    marginVertical: utils.spacings.sp16,
-    marginHorizontal: utils.spacings.sp16,
+const dateTextStyle = prepareNativeStyle(({ spacings }) => ({
+    marginTop: spacings.sp24,
+    marginHorizontal: spacings.sp16,
+    marginBottom: spacings.sp12,
 }));
 
 export const TransactionListGroupTitle = ({ monthKey }: TransactionListGroupProps) => {
-    const { applyStyle } = useNativeStyles();
+    const { translate } = useTranslate();
     const { MonthNameFormatter } = useFormatters();
+    const { applyStyle } = useNativeStyles();
+
     const sectionTitle =
         monthKey === 'pending'
-            ? 'Pending'
+            ? translate('transactions.status.pending')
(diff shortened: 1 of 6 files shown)


suite-common/formatters/src/formatters/prepareYearMonthFormatter.ts
```diff
@@ -0,0 +1,11 @@
+import { format } from 'date-fns';
+import enUS from 'date-fns/locale/en-US';
+
+import { makeFormatter } from '../makeFormatter';
+
+export const YearMonthFormatter = makeFormatter<Date, string>(
```
vytick: I know this is copy & paste & tweak, but I would suggest insert locale from outside or rename this to something like EnYearMonthFormatter and move it to `/suite-native`. Wdyt?

suite-native/transactions/src/components/TransactionsList/TransactionListGroupTitle.tsx
```diff
@@ -7,21 +7,20 @@ type TransactionListGroupProps = {
...
 export const TransactionListGroupTitle = ({ monthKey }: TransactionListGroupProps) => {
     const { applyStyle } = useNativeStyles();
-    const { MonthNameFormatter } = useFormatters();
-    const sectionTitle =
-        monthKey === 'pending'
-            ? 'Pending'
-            : MonthNameFormatter.format(parseTransactionMonthKey(monthKey));
+    const { YearMonthFormatter } = useFormatters();
+    const date = parseTransactionMonthKey(monthKey);
+    const sectionTitle = monthKey === 'pending' ? 'Pending' : YearMonthFormatter.format(date);
```
vytick: Can you, please, move 'Pending' to intl since we are touching this 🙏

suite-common/formatters/src/formatters/prepareYearMonthFormatter.ts
vytick: I would also suggest to delete `MonthNameFormatter`

yanascz: Removing `MonthNameFormatter` is probably doable if it's not used elsewhere. Not sure about the rest since `locale` is hardcoded into other formatters as well. I don't know what the motivation behind `FormatterProviderContext` was and whether it makes sense to have the formatter there. 🤷‍♂️

yanascz: We could also just change the format in `MonthNameFormatter` and leave the rest as is. Seems as the best option at the moment TBH.

suite-native/transactions/src/components/TransactionsList/TransactionListGroupTitle.tsx
yanascz: Yeah, noticed that after I created the PR. Fixed.

vytick (approved):
🚢 🇮🇹

vytick (approved):
🚢 🇮🇹

PR #16016, opened 2024-12-18
feat(e2e): Migrate onboarding t2t1 tests

HajekOndrej:
## Related Issue

Resolve


packages/suite-desktop-core/e2e/support/pageActions/analyticsActions.ts
Vere-Grey, line 1: Have you foresee here a lot more thing to be added? I wondering if those two locators and click action justify the split from Onboarding Page object and the necessity of importing it then to onboarding page

packages/suite-desktop-core/e2e/support/pageActions/devicePromptActions.ts
```diff
@@ -0,0 +1,19 @@
+import { Locator, Page, expect } from '@playwright/test';
+
+export class DevicePromptActions {
+    private readonly confirmOnDevicePrompt: Locator;
+    private readonly connectDevicePrompt: Locator;
+
+    constructor(page: Page) {
+        this.confirmOnDevicePrompt = page.getByTestId('@onboarding/confirm-on-device');
```
Vere-Grey, line 8: Do you know if this locator and the object differs from the other '@prompts/confirm-on-device'. We got several uses of that (not yet refactored). I am just wondering if it is the same element and we can reuse your new locator even there or we will alter add another as part of the future refactoring

packages/suite-desktop-core/e2e/support/fixtures.ts
```diff
@@ -30,12 +34,16 @@ type Fixtures = {
...
+    analyticsPage: AnalyticsActions;
+    devicePrompt: DevicePromptActions;
     analytics: AnalyticsFixture;
+    indexedDb: IndexedDbFixture;
 };
 
 const test = base.extend<Fixtures>({
     startEmulator: true,
-    emulatorStartConf: {},
+    emulatorStartConf: { wipe: true },
```
Vere-Grey: I struggled a bit when we had more defaults here and I was constantly comparing what is default here and what is default in the API. So I would vote against adding more defaults here.
But the wipe: true is a really a setting that all of our test will want right? So it make sense to have this one here.
Good improvement.

There is bunch of test that have `emulatorStartConf: { wipe: true }` as the only config in setup. Could you remove it as part of this refactoring?

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-create-wallet.test.ts
```diff
@@ -0,0 +1,48 @@
+import { test } from '../../support/fixtures';
+
+test.describe('Onboarding - create wallet', { tag: ['@group=device-management'] }, () => {
+    // This test always needs to run the newest possible emulator version
+    // Emulator setup: wipe: true, model: T2T1, version: 2-latest
+    test.use({
+        emulatorStartConf: { wipe: true, model: 'T2T1', version: '2-latest' },
+        emulatorSetupConf: undefined,
```
Vere-Grey: Let's discuss this one. I once was considering something similiar but abandon it cause it is unclear what it means and why it is here, just from looking at it.

Cause based on the current implementation, this means `start the Trezor env, start emulator but dont setup the emulator yet`, right? So I think we should rather add clear variable setupEmulator with default true. Which you can override here to false.

Or something else but ideally something that tells you what is happening just from looking at it here in test.

Vere-Grey: just so I understand the test correctly, It needs the emulator started still before the page loads, right? Or it doesnt?

packages/suite-desktop-core/e2e/support/indexedDb.ts
```diff
@@ -0,0 +1,21 @@
...
+        await this.page.evaluate(() => {
+            return new Promise<void>((resolve, reject) => {
+                const request = indexedDB.deleteDatabase('trezor-suite');
+
+                request.onsuccess = () => {
+                    resolve();
+                };
+
+                request.onerror = () => {
+                    reject('Error resetting database');
```
Vere-Grey, line 16: How will the fail look like in the tests? Will there be some trace or additional info alongside of it?

packages/suite-desktop-core/e2e/support/pageActions/backupActions.ts
```diff
@@ -0,0 +1,46 @@
...
+    readonly wroteSeedProperlyCheckbox: Locator;
+    readonly madeNoDigitalCopyCheckbox: Locator;
+    readonly willHideSeedCheckbox: Locator;
+    readonly backupCloseButton: Locator;
+
+    constructor(
+        private page: Page,
+        private devicePrompt: DevicePromptActions,
+    ) {
+        this.devicePrompt = devicePrompt;
```
Vere-Grey: redundant assignment of private param

packages/suite-desktop-core/e2e/support/pageActions/backupActions.ts
```diff
@@ -0,0 +1,46 @@
...
+        this.madeNoDigitalCopyCheckbox = page.getByTestId(
+            '@backup/check-item/made-no-digital-copy',
+        );
+        this.willHideSeedCheckbox = page.getByTestId('@backup/check-item/will-hide-seed');
+        this.backupCloseButton = page.getByTestId('@backup/close-button');
+    }
+
+    async passThroughShamirBackup(shares: number, threshold: number) {
+        // Backup button should be disabled until all checkboxes are checked
+        await this.backupStartButton.isDisabled();
```
Vere-Grey: missing expect

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-create-wallet.test.ts
```diff
@@ -0,0 +1,48 @@
+import { test } from '../../support/fixtures';
+
+test.describe('Onboarding - create wallet', { tag: ['@group=device-management'] }, () => {
+    // This test always needs to run the newest possible emulator version
+    // Emulator setup: wipe: true, model: T2T1, version: 2-latest
```
Vere-Grey, line 5: duplicate information. The setup is two lines below

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-create-wallet.test.ts
```diff
@@ -0,0 +1,48 @@
...
+        await analyticsPage.continue();
+        await analyticsPage.continue();
+        await page.getByTestId('@firmware/continue-button').click();
+        await page.getByTestId('@onboarding/path-create-button').click();
+
+        // Will be clicking on Shamir backup button
+        await page.getByTestId('@onboarding/select-seed-type-open-dialog').click();
+        await page.getByTestId('@onboarding/select-seed-type-shamir-advanced').click();
+        await page.getByTestId('@onboarding/select-seed-type-confirm').click();
+        await page.getByTestId('@onboarding/confirm-on-device').isVisible();
```
Vere-Grey: Missing expect, Not sure if this is just bad AI conversion or you being new in playwright so I will add a bit of explenation. 
 `await page.getByTestId('@onboarding/confirm-on-device').isVisible();` is just a statement that returns true or false based on visibility of the element. So as it is. It does no verification. ergo test would pass even if the element was invisible.

Vere-Grey: I see there are more of these in following code. I will not spam the review with comments. Please check for the remaining cases.

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-recovery-persistence.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        await trezorUserEnvLink.selectNumOfWordsEmu(20);
+        await trezorUserEnvLink.pressYes();
+
+        // Enter first Shamir share
+        for (const word of shareOneOfThree) {
+            await trezorUserEnvLink.inputEmu(word);
+        }
+
+        await devicePrompt.confirmOnDevicePromptIsShown();
+        //await page.waitForTimeout(501);
```
Vere-Grey: So can we remove this commented wait and some of the others since the test is passing?

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-recovery-success.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    test.use({
+        emulatorStartConf: { wipe: true, model: 'T2T1', version: '2.5.3' },
+        emulatorSetupConf: undefined,
+    });
+    test.beforeEach(async ({ analyticsPage, onboardingPage }) => {
+        await onboardingPage.disableFirmwareHashCheck();
+
+        // Go through analytics opt-out
+        analyticsPage.continue();
+        analyticsPage.continue();
```
Vere-Grey: I would DRY this to analyticsPage. If we name this OptOutFromAnalytics we will also remove the necessity of comment. First time I saw those two identical lines, I thought "ha must be a duplicate step by mistake". So the readability would benefit from having this in page object method

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-recovery-success.test.ts
```diff
@@ -0,0 +1,58 @@
...
+        // Confirm recovery success
+        await page.waitForTimeout(1000);
+        await trezorUserEnvLink.pressYes();
+
+        // Finalize recovery, skip pin, and check success
+        await onboardingPage.continueRecoveryButton.click();
+        await onboardingPage.skipPin();
+        await onboardingPage.continueCoinsButton.click();
+        await expect(onboardingPage.finalTitle).toBeVisible();
+        await expect(onboardingPage.finalTitle).toContainText('Setup complete!');
```
Vere-Grey, line 54: I bet toContainText checks for visibility too. so we can remove the expect visible
And if we have exact string then I would go with toHaveText as that will eliminate the chance of false pass with partial match of toContainText

packages/suite-desktop-core/e2e/tests/spawn-bridge.test.ts
```diff
@@ -53,6 +54,7 @@ test.describe.serial('Bridge', { tag: ['@group=suite', '@desktopOnly'] }, () =>
         await suite.window.title();
         const onboardingPage = new OnboardingActions(
             suite.window,
+            new AnalyticsActions(suite.window),
```
Vere-Grey, line 57: I wonder whethere we shouldnt rather initilized  AnalyticsActions in the OnboardingActions and dont have to import it here and pass it in fixtures.ts
:thinking:

Vere-Grey (changes requested):
Looking good but I would suggest few things. See my comments.

packages/suite-desktop-core/e2e/support/pageActions/devicePromptActions.ts
HajekOndrej: It is different, but maybe we can change it in the app to be the same, I don't know if there is real reason for it to be under two different IDs.

packages/suite-desktop-core/e2e/support/fixtures.ts
HajekOndrej: Sorry, this is unintentional and I forgot to remove it. Good catch.

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-create-wallet.test.ts
HajekOndrej: yes, it needs to be started, but not setup. I also thought about adding another variable, it's probably better.

packages/suite-desktop-core/e2e/support/indexedDb.ts
HajekOndrej: It will throw an exception with this message. I couldn't get any more valuable info from the event.

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-create-wallet.test.ts
HajekOndrej: Yep, thanks.

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-recovery-success.test.ts
HajekOndrej: Yep, I thought about that as well. Will do.

packages/suite-desktop-core/e2e/tests/spawn-bridge.test.ts
HajekOndrej: Maybe, but there are currently cases of only using AnalyticsPage and not Onboarding. I would keep it like this for now.

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-recovery-success.test.ts
HajekOndrej: I think it doesn't check for visibility and the string is not full afaik.

packages/suite-desktop-core/e2e/support/pageActions/devicePromptActions.ts
Vere-Grey: https://github.com/orgs/trezor/projects/78/views/1?pane=issue&itemId=91462030

packages/suite-desktop-core/e2e/support/pageActions/analyticsActions.ts
HajekOndrej: It seemed to me that the onboarding class is already insanely huge and it will get bigger, so I thought it could be a good idea to move stuff away.

packages/suite-desktop-core/e2e/tests/onboarding/t2t1-recovery-success.test.ts
Vere-Grey: This is a shock for me. I will do later some investigation.
So far I did some tests and seems that toContainText will pass even on hidden element
but toHaveText would fail.
wow.

Vere-Grey (approved):
:+1:

PR #16017, opened 2024-12-18
Chore(suite): add Rspack as electron main build option

Lemonexe:
## Description

WIP experimental support for Rspack for whole Suite and related worker scripts.

I replaced all webpack plugins [listed as incompatible](https://rspack.dev/guide/compatibility/plugin) with Rspack variants. I also replaced some that are recommended. For example `TerserPlugin` really slows down production build, but that's not very important, it's not used in dev.

## Renderer WIP
**not working** because web workers, we are using old Webpack api for web workers, so I tried migrating the one affected worker to new, but was not successful: 2645c998a298130cf85689f08b903010b7ec83ad

**dev server + HMT** not touched yet. `WebpackPluginServe` is not listed as compatible

## Main
But the electron `main + preload` can succesfully be built with Rspack:

This should work on Linux (last 2 lines will be different for macOS):
```
yarn workspace @trezor/suite-desktop-core build:app
rm -rf ./packages/suite-desktop/dist
NODE_ENV=production yw @trezor/suite-desktop-core build:core:rspack
yarn workspace @trezor/suite-desktop-core build:app:electron --publish never --linux --x64
./packages/suite-desktop/build-electron/Trezor-Suite-25.1.0-linux-x86_64.AppImage --no-sandbox
```

## Performance measurement

Looks promising, about 50% time saved :open_mouth: Though it's not such a difference [as advertised](https://rspack.dev/). I think it's because we use a lot of plugins. Makes sense that pure Rspack vs Webpack will be a huge relative difference, but plugins are comparably slow if they are in JS.

| env                    | time [s]                    |
|------------------------|-----------------------------|
| renderer webpack dev:  | 55                          |
| renderer rspack dev:   | 27 (incomplete; with error) |
| main webpack dev:      | 15.0                        |
| main rspack dev:       | 7.5                         |
|------------------------|-----------------------------|
| renderer rspack prod:  | 48 (incomplete; with error) |
| renderer webpack prod: | 75                          |
| main webpack prod:     | 23.5                        |
| main rspack prod:      | 8.5                         |


packages/suite-desktop-core/webpack/core.rspack.config.ts
```diff
@@ -82,24 +82,25 @@ const config: webpack.Configuration = {
...
+                options: {
+                    jsc: {
+                        parser: {
+                            syntax: 'typescript',
+                        },
+                        externalHelpers: true,
+                    },
+                    env: {
+                        // electron 32 uses node 20 https://www.electronjs.org/docs/latest/tutorial/electron-timelines
+                        targets: 'node 20',
```
peter-sanderson, line 102: Would this need to be changed every time we update node?

Lemonexe: Every time we update electron, because the target runtime is dependent on nodeJS version that electron uses.
But it does not change very often..
This is crucial though, you can target an older version of node, it will just use older syntax which may not be optimal performance-wise but idk how important that is :shrug:

packages/suite-desktop-core/webpack/core.rspack.config.ts
```diff
@@ -82,24 +82,25 @@ const config: webpack.Configuration = {
         ...dependencies,
         ...devDependencies,
         'bufferutil', // optional dependency of ws lib
-        'memcpy', // optional depencency of bytebuffer lib
         'utf-8-validate', // optional dependency of ws lib
-        'osx-temperature-sensor', // optional dependency of systeminformation lib
     ],
+    externalsType: 'commonjs2',
```
Lemonexe, line 86: This took me a long time to figure out. Webpack detects this automatically, but Rspack couldn't automatically resolve some of these externals, and it would throw `ReferenceError: __WEBPACK_EXTERNAL_MODULE_electron__ is not defined` during runtime (in built AppImage).
Strangely, that error only manifested when chunking is on. The `require('electron')` was missing in a file where it was supposed to be (a shared chunk `share/app~bridge.js`). Without chunking options, there is only one `.js` output file for each entry point, and Rspack correctly wrote `require('electron')` there.

packages/suite-desktop-core/webpack/core.rspack.config.ts
```diff
@@ -82,24 +82,25 @@ const config: webpack.Configuration = {
         ...dependencies,
         ...devDependencies,
         'bufferutil', // optional dependency of ws lib
-        'memcpy', // optional depencency of bytebuffer lib
```
Lemonexe: This seems outdated, because `memcpy` is never mentioned in any output bundle js file. The other externals are conditionally required there, but not `memcpy`..

packages/suite-desktop-core/webpack/core.rspack.config.ts
```diff
@@ -0,0 +1,187 @@
...
+        minimizer: [
+            new rspack.SwcJsMinimizerRspackPlugin({
+                extractComments: false,
+                minimizerOptions: {
+                    format: {
+                        comments: false,
+                    },
+                },
+            }),
+            new rspack.LightningCssMinimizerRspackPlugin(),
```
Lemonexe, line 145: TerserPlugin is really slow, so this brings considerable performance gains!

Lemonexe:
Closing, this is unfeasible: too much effort to have a working state Suite.
And ES modules for Electron main are even worse than with Webpack, which weakens the advantages of RSpack.

PR #16021, opened 2024-12-18
chore: remove disable of triple-slash-reference eslint rule + fix

### LINKED ISSUE #15310: `@typescript-eslint/triple-slash-reference` remove `off`
Unfinished work from ESLint9 upgrade, can be done later

peter-sanderson:
Resolves https://github.com/trezor/trezor-suite/issues/15310

### DIFF
--- a/packages/eslint/src/typescriptConfig.mjs
+++ b/packages/eslint/src/typescriptConfig.mjs
@@ -48,7 +48,6 @@ export const typescriptConfig = [
                 },
             ],
             '@typescript-eslint/no-empty-object-type': 'off', // Todo: we shall solve this, this is bad practice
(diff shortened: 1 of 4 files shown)


Lemonexe (changes requested):
Eslint rule is not enabled as the PR name indicates

Lemonexe (approved):
LGTM. In future we might need to solve how to import `.d.ts` declaration files from elsewhere, probably in `tsconfig.json`.

But currently this makes sense; the `Window` declaration is inconsistent between `suite`, `suite-web` and `suite-desktop-core`, but **in the specific case** of `Window` IMHO that's alright, because that definition is never complete, there can be basically anything, this just defines what additional properties we need in current context (just thinking out loud)

PR #16022, opened 2024-12-18
feat(graph): use local account balance for latest balance

### LINKED ISSUE #15526: Doubled balance on the dashboard from time to time
**Describe the bug**
We have multiple reports of people seeing their balance on the dashboard doubled (multiplied by 2x, like instead of $100 they see $200). We don't know how to reproduce.

**Info:**

-   Mobile Suite Version: Unclear, but it's on production
-   OS: probably both platforms

**How to reproduce**

🤷 

**Expected behavior**
Always see the actual balance, never more.

**Additional context**
One idea to investigate is if it is happening after some specific time when the app is on background (not turned off completely).

vytick:
Use latest balance for account instead of getAccountInfo to show the same data we do on the other parts of the app.Ï

Resolve #15526


vytick:
Closed after discussion. It seems to be more complex issue and only solving this wont feel right approach

PR #16025, opened 2024-12-18
fix(suite-native): ensure that devices with BTC-only firmware have BTC enabled

### LINKED ISSUE #15877: Connecting BTC only device/fw to Suite lite with enabled ETH coin only shows "Your wallet is empty" when it isn't empty
**Describe the bug**
If you connect a BTC only device or BTC only fw to Suite lite and there is for examply only ETH enabled, I am shown "Your wallet is empty" screen where I should be seeing a one empty account at least. There is also an edge edge case where if you have a device with BTC and some other coins in view only and you also have a BTC only device in view only and you disable the BTC coin on the device with regular firmware, it will show you "Your wallet is empty" on the BTC device even though you have some funds.

**Info:**

-   Mobile Suite Version: 24.11.3 2d7eb04
-   OS: Android

**How to reproduce**

1. Open your app
2. Go to Settings>Enable coins
3. Enable ETH for example
4. Connect your BTC only device
5. You can see a discovery for a split second
6. "Your wallet is empty" screen is shown

**Expected behavior**
First empty account should be shown.

yanascz:
## Related Issue

Resolve #15877

### DIFF
--- a/suite-native/app/src/hooks/useCoinEnablingInitialCheck.tsx
+++ b/suite-native/app/src/hooks/useCoinEnablingInitialCheck.tsx
@@ -42,7 +42,7 @@ export const useCoinEnablingInitialCheck = () => {
             let timeoutId: TimerId;
             //if btc only device, just run discovery for btc and do not show the UI
             if (hasBitcoinOnlyFirmware) {
-                dispatch(setEnabledDiscoveryNetworkSymbols(['btc']));
+                // discoveryMiddleware ensures that BTC is enabled for devices with BTC-only firmware
                 dispatch(setIsCoinEnablingInitFinished(true));
                 dispatch(applyDiscoveryChangesThunk());
             } else {
--- a/suite-native/discovery/src/discoveryConfigSlice.ts
+++ b/suite-native/discovery/src/discoveryConfigSlice.ts
@@ -73,6 +73,11 @@ export const discoveryConfigSlice = createSlice({
         setDiscoveryInfo: (state, { payload }: PayloadAction<DiscoveryInfo | null>) => {
             state.discoveryInfo = payload;
         },
+        addEnabledDiscoveryNetworkSymbol: (state, { payload }: PayloadAction<NetworkSymbol>) => {
+            if (!state.enabledDiscoveryNetworkSymbols.includes(payload)) {
+                state.enabledDiscoveryNetworkSymbols.push(payload);
+            }
+        },
         toggleEnabledDiscoveryNetworkSymbol: (state, { payload }: PayloadAction<NetworkSymbol>) => {
             const symbol = payload;
             const index = state.enabledDiscoveryNetworkSymbols.indexOf(symbol);
@@ -208,6 +213,7 @@ export const selectTokenDefinitionsEnabledNetworks = createMemoizedSelector(
 export const {
     toggleAreTestnetsEnabled,
     setDiscoveryInfo,
+    addEnabledDiscoveryNetworkSymbol,
     toggleEnabledDiscoveryNetworkSymbol,
     setEnabledDiscoveryNetworkSymbols,
     setIsCoinEnablingInitFinished,
--- a/suite-native/discovery/src/discoveryMiddleware.ts
+++ b/suite-native/discovery/src/discoveryMiddleware.ts
@@ -9,6 +9,7 @@ import {
     accountsActions,
 } from '@suite-common/wallet-core';
 import { isFirmwareVersionSupported } from '@suite-native/device';
+import { hasBitcoinOnlyFirmware } from '@trezor/device-utils';
 
 import { startDescriptorPreloadedDiscoveryThunk, discoveryCheckThunk } from './discoveryThunks';
 import {
@@ -17,6 +18,7 @@ import {
     toggleAreTestnetsEnabled,
     setEnabledDiscoveryNetworkSymbols,
     toggleEnabledDiscoveryNetworkSymbol,
+    addEnabledDiscoveryNetworkSymbol,
 } from './discoveryConfigSlice';
 
 export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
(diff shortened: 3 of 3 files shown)


suite-native/app/src/hooks/useCoinEnablingInitialCheck.tsx
```diff
@@ -42,6 +42,7 @@ export const useCoinEnablingInitialCheck = () => {
             let timeoutId: TimerId;
             //if btc only device, just run discovery for btc and do not show the UI
             if (hasBitcoinOnlyFirmware) {
+                // TODO: Does this line now make sense?
                 dispatch(setEnabledDiscoveryNetworkSymbols(['btc']));
```
vytick: Imo yes. Otherwise if you connect universal fw after btc only, skipping this would mean, that you wont see anything on dashboard and you just tried to fix that.

suite-native/discovery/src/discoveryConfigSlice.ts
```diff
@@ -195,7 +195,7 @@ export const selectEnabledDiscoveryNetworkSymbols = (state: DiscoveryConfigSlice
 export const selectDeviceEnabledDiscoveryNetworkSymbols = createMemoizedSelector(
     [
         selectDiscoveryNetworkSymbols,
-        selectEnabledDiscoveryNetworkSymbols,
+        selectEnabledDiscoveryNetworkSymbols, // TODO: What about moving the BTC-only logic there?
```
vytick: If going with this approach, imo putting the logic here is better option as you suggest

vytick: Also `selectEnabledDiscoveryNetworkSymbols` is used for message system so it would make sense to have it here if we want to have btc targeted (and not already enabled by coin enabling)

vytick (commented):
If we decide to go this way (always making btc enabled for btc only fw), there might be UX discrepancy, that user thinks that he has enabled BTC for all devices, but has not and after connecting universal fw, BTC account wont be discovered. 

I still think (I have no data other than original discussions with Katka), that BTC only without enabled BTC is very  edge case, so I am not much afraid of this (also even more edgecase) scenario.

suite-native/app/src/hooks/useCoinEnablingInitialCheck.tsx
yanascz: As discussed in person, moved to `discoveryMiddleware`.

suite-native/discovery/src/discoveryConfigSlice.ts
yanascz: As discussed in person, this was not a good solution for many reasons. Handled in `discoveryMiddleware`.

suite-native/discovery/src/discoveryMiddleware.ts
```diff
@@ -69,6 +71,12 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
             );
         }
 
+        // Ensure that devices with BTC-only firmware have BTC enabled.
```
vytick:
```suggestion
        // Ensure that there is BTC enabled when connecting BTC-only firmware
```

yanascz: Tweaked based on this.

vytick (approved):
Looks good, tested. Bitcoiners gonna love you :)  Good job 🚀

suite-native/discovery/src/discoveryMiddleware.ts
```diff
@@ -69,6 +71,12 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
             );
         }
 
+        // ensure that BTC is enabled when device with BTC-only firmware is connected
+        // (it could have been disabled via some other device with universal firmware)
+        if (deviceActions.selectDevice.match(action) && selectHasBitcoinOnlyFirmware(getState())) {
```
matejkriz: I'm not sure about the `selectDevice` action. I'm not saying it's wrong choice, just that I want to double check that. Also think about future feature to switch to btc-only FW.

yanascz: So what action do you suggest instead? 🤔

matejkriz: As I said, I just wanted to double check that. Now I believe that `selectDevice` is a good choice, since `device-connect` has some edge cases like you switch to view-only universal wallet, disable BTC, go back to bitcoin-only device... 

But you should not use `getState()` here, that could contain the other device than the one being selected. I think you should replace `selectHasBitcoinOnlyFirmware` with `hasBitcoinOnlyFirmware` and pass the payload of the action to it (it contains the device).

matejkriz: And we should probably add here one more condition for `device.connected` - we don't want to enable BTC it the btc-only device is not connected but in view-only - because we can't do the discovery in that case.

matejkriz: It’s funny how something that looks so simple can be so complex. 😄

yanascz: OK, that makes sense. Will look into that.

yanascz: Fixup pushed, seems to work fine. 👍

matejkriz: One edge case that is not handled and honestly I'm not sure how to handle it:

1. connect and discover device with Universal FW, enable for example ETH, enable view-only
1. connect and discover btc-only trezor, enable view-only
1. disconnect trezor
1. select the one with universal FW
1. go to Enabled coins and disable BTC
1. select back btc-only trezor
Suite show it as empty, which is not true!
image

matejkriz: Now I realized that same think can happen already and not only with btc-only, but any device that have coins on some network, that you disable. So it can be handled separately, not needed to resolve it now.

matejkriz: But it is exactly the case described in this issue description:

> There is also an edge edge case where if you have a device with BTC and some other coins in view only and you also have a BTC only device in view only and you disable the BTC coin on the device with regular firmware, it will show you "Your wallet is empty" on the BTC device even though you have some funds.

So let's discuss in person.

matejkriz: Ok, so let's agree that it is too marginal case and we can improve it if there will be any complains from real users.

matejkriz (approved):
It doesn't resolve the linked issue fully, but it's too marginal case to handle - we see lot of options how to handle that edge case, but nothing reasonably simple.

 I'm talking about this edge case:

1. connect and discover device with Universal FW, enable for example ETH, enable view-only
1. connect and discover btc-only trezor, enable view-only
1. disconnect trezor
1. select the one with universal FW
1. go to Enabled coins and disable BTC
1. select back btc-only trezor
Suite show it as empty, which is not true!
image

This PR makes the situation better and we hope it will be really hard for someone to break it unintentionally (he has to have multiple devices, one of them btc-only and disable the btc when the universal is selected and btc-only in view-only).

PR #16032, opened 2024-12-18
Refactor network config

enjojoy:
## Description

1. rename customBackend to backendTypes
2. create explorer urls dynamically

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16034
Resolve https://github.com/trezor/trezor-suite/issues/16035

PR #16042, opened 2024-12-19
chore(wallet-config): enhance exhaustive typing

Lemonexe:
## Description

A small suggestion to enhance exhaustive typing in #16032 :slightly_smiling_face: 

Writing it this way will ofc throw TS error with invalid key, same as current solution.

But it will also throw a [TS error](https://github.com/user-attachments/assets/fa75b053-953c-4af0-8ad3-797dbe2a94b1) for missing network type (instead of runtime error).
TS error is advantageous to runtime error, because if we would add a new network in future and forgot to add it here, you catch much sooner - during type checking rather than in-app :slightly_smiling_face:


enjojoy (approved):
Thank you, it's a great addition 🚀

PR #16043, opened 2024-12-19
fix(suite-native): fw update feature flag fix

PeKne:
Fixes FW update UI so it reflects the `Firmware update` feature flag value.

## Screenshots:

###BEFORE:
Snímek obrazovky 2024-12-19 v 9 17 59

### AFTER:
Snímek obrazovky 2024-12-19 v 9 18 06

### DIFF
--- a/suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
+++ b/suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
@@ -66,9 +66,9 @@ export const DeviceFirmwareCard = () => {
         selectIsDiscoveryActiveByDeviceState(state, device?.state),
     );
     const navigation = useNavigation<NavigationProp>();
-    const isFirmwareUpdateEnabled = useFeatureFlag(FeatureFlag.IsFirmwareUpdateEnabled);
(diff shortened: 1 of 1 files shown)


Nodonisko (approved):

suite-native/module-device-settings/src/screens/DeviceSettingsModalScreen.tsx
```diff
@@ -27,6 +28,7 @@ export const DeviceSettingsModalScreen = () => {
     const [isUpdateSheetOpen, setIsUpdateSheetOpen] = useState<boolean>(false);
 
     const isUpgradable = deviceReleaseInfo?.isNewer ?? false;
+    const [isFirmwareUpdateEnabled] = useFeatureFlag(FeatureFlag.IsFirmwareUpdateEnabled);
```
Nodonisko: I think it is not necessary to have here, this should be accessible if FF is off anyway

matejkriz (approved):
😭

suite-native/module-device-settings/src/screens/DeviceSettingsModalScreen.tsx
PeKne: true, fixed

PR #16050, opened 2024-12-19
message-system 69: Ask users that installed unintentionally released 2.8.6 FW on TS5 to contact support

matejkriz:
## Description

- Mobile only
- TS5 with 2.8.6 only

Mobile only support English so we don't need translations.

## Screenshot
Screenshot_2024-12-19-11-37-11-038_io trezor suite debug
ts:

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,8 +1,70 @@
 {
     "version": 1,
-    "timestamp": "2024-11-29T00:00:00+00:00",
-    "sequence": 68,
+    "timestamp": "2024-12-19T00:00:00+00:00",
+    "sequence": 69,
     "actions": [
+        {
+            "conditions": [
+                {
+                    "environment": {
+                        "desktop": "*",
+                        "mobile": ">=24.12.1",
+                        "web": "*"
+                    },
+                    "devices": [
+                        {
+                            "model": "T3T1",
+                            "firmware": "2.8.6",
+                            "bootloader": "*",
+                            "variant": "*",
+                            "firmwareRevision": "*",
+                            "vendor": "*"
+                        }
+                    ]
+                }
+            ],
+            "message": {
+                "id": "cdcc9424-a42a-4d85-b46d-1ac2e9e1dc8a",
+                "priority": 90,
+                "dismissible": false,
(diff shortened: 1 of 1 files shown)


suite-common/message-system/config/config.v1.json
```diff
@@ -1,8 +1,70 @@
...
+                    "tr": "You have downloaded an incomplete firmware version 2.8.6, released unintentionally. Tap here to open our chatbot and create a support ticket for assistance. Remember, never share your wallet backup with anyone under any circumstances.",
+                    "pt": "You have downloaded an incomplete firmware version 2.8.6, released unintentionally. Tap here to open our chatbot and create a support ticket for assistance. Remember, never share your wallet backup with anyone under any circumstances.",
+                    "uk": "You have downloaded an incomplete firmware version 2.8.6, released unintentionally. Tap here to open our chatbot and create a support ticket for assistance. Remember, never share your wallet backup with anyone under any circumstances."
+                },
+                "cta": {
+                    "action": "external-link",
+                    "link": "https://trezor.io/learn/c/trezor-suite-lite#open-chat",
```
matejkriz: This link has to be replaced by specific one.

suite-common/message-system/config/config.v1.json
```diff
@@ -1,8 +1,70 @@
...
-    "sequence": 68,
+    "timestamp": "2024-12-19T00:00:00+00:00",
+    "sequence": 69,
     "actions": [
+        {
+            "conditions": [
+                {
+                    "environment": {
+                        "desktop": "!",
+                        "mobile": "24.12.1",
```
matejkriz: We should allow newer version too. And maybe even older? But that sounds like a too much on an edge case.

matejkriz: fixed, force pushed

MiroslavProchazka (approved):

Lemonexe (approved):
Works as expected locally; on 2.8.6 T3T1 it displays this message, not present on other versions, or other devices:
image

STew790:
**QA OK**
Works on mobile also as expected. Only connected TS5 affected with the correct 2.8.6 FW. Other devices won't trigger this.

image
image

PR #16051, opened 2024-12-19
chore(connect-common): add 2.8.6 to release.json

MiroslavProchazka:
DO NOT MERGE YET

Adding 2.8.6. to releases.json


Lemonexe (commented):
~I don't see any problems with it. However, let's make an issue to remove it from `releases.json`, when majority of users are updated to future 2.8.7 or later.~
It wrongly shows update available, which is undesired.

PR #16055, opened 2024-12-19
feat(message-system): introduce easy way to test local jws

vytick:
Add simple option to test message system with local jws

### DIFF
--- a/suite-common/message-system/src/messageSystemThunks.ts
+++ b/suite-common/message-system/src/messageSystemThunks.ts
@@ -22,7 +22,18 @@ import {
 } from './messageSystemSelectors';
 import { jws as configJwsLocal } from '../files/config.v1';
 
+// Enable this for local development purposes:
+// set to true to always fetch local JWS
+const FORCE_LOCAL_JWS = false;
+
 const getConfigJws = async () => {
+    if (FORCE_LOCAL_JWS) {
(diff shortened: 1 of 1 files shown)


matejkriz (approved):
Great, thanks. I was thinking about environment variable for this (same as we have for analytics logs EXPO_PUBLIC_IS_ANALYTICS_LOGGER_ENABLED), but I think this is better to see it in the git diff so you are not so confused why you can't fetch new remote version...

What about to skip sequence number check too? 
```javascript
if (currentSequence < config.sequence || FORCE_LOCAL_JWS /* Allow to skip sequence check for local testing */) { 
```

suite-common/message-system/src/messageSystemThunks.ts
```diff
@@ -22,7 +22,18 @@ import {
 } from './messageSystemSelectors';
 import { jws as configJwsLocal } from '../files/config.v1';
 
+// this is for local development purpose
+// set to true to always fetch local jws
```
matejkriz: Nitpick, feel free to ignore
```suggestion
// Enable this for local development purposes:
// set to true to always fetch local JWS
```

vytick: Great idea: 88529b1dbd

PR #16058, opened 2024-12-19
fix(suite-common): resolve formatAmount TODO in try catch block

### LINKED ISSUE #11450: Resolve formatAmount todo in catch block
**Describe the bug**
Resolve TODO in `formatAmount` method

It should not return "-1", it should just throw error

izmy:
## Description

`formatAmount` should not return "-1", it should just throw an error

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/11450


tomasklim (approved):

tomasklim (changes requested):
Please fix unit tests

izmy:
> Please fix unit tests

I had to change the method a little more because tests and the entire app code relied on that `formatAmount` to return some value. So I added `safeAmount` which sets '0' for undefined or empty string. The exception is thrown only if the value is not a number.

tomasklim (approved):

PR #15993, opened 2024-12-16
feat(suite): introduce network displaySymbol

adderpositive:
## Description
- add `displaySymbol` to networks and should be displayed on every output to the use

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/14725
Resolve https://github.com/trezor/trezor-suite/issues/15821
Follow up https://github.com/trezor/trezor-suite/pull/13670

### Before
image
image
image
image
image
image

### After
image
image
image
image
image
image

PR #16061, opened 2024-12-19
Display gas price for token transaction when not enough Eth balance 

### LINKED ISSUE #15219: Show how much fee is needed to send token tx
**Describe the bug**
Based on [Slack](https://satoshilabs.slack.com/archives/C016ACNF1BN/p1730561896489939)
Related to https://github.com/trezor/trezor-suite/issues/13756

User is missing info what is enough ETH to cover token tx fee 
Screenshot 2024-11-05 at 10 59 28

**Info:**

-   Suite Version: 24.11.0

**Expected behavior**
We can also add it into error message `With current network fees (5 gas price), enough amount of ETH is 0.001`
Screenshot 2024-11-05 at 10 58 22

enjojoy:
## Description

Now users can see ow much Eth they need for a token transaction, when they don't have enough 

## Related Issue 

Resolve https://github.com/trezor/trezor-suite/issues/15219

## Screenshots:
Before:
Screenshot 2024-12-19 at 15 30 52

After:
Screenshot 2024-12-19 at 15 26 04
Screenshot 2024-12-19 at 15 29 30

### DIFF
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -5426,6 +5426,11 @@ export default defineMessages({
         defaultMessage: 'Not enough {symbol} to cover transaction fee',
         id: 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE',
     },
+
+    AMOUNT_NOT_ENOUGH_CURRENCY_FEE_WITH_ETH_AMOUNT: {
+        defaultMessage: 'Not enough {symbol} to cover transaction fee {feeAmount}',
+        id: 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE_WITH_ETH_AMOUNT',
+    },
     REMAINING_BALANCE_LESS_THAN_RENT: {
         defaultMessage:
             'After sending this amount, your account will have {remainingSolBalance} SOL remaining. A non-empty account must maintain a balance of more than {rent} SOL.',
--- a/suite-common/wallet-core/src/send/sendFormEthereumThunks.ts
+++ b/suite-common/wallet-core/src/send/sendFormEthereumThunks.ts
@@ -1,4 +1,4 @@
-import { toWei } from 'web3-utils';
+import { fromWei, toWei } from 'web3-utils';
 
 import { BigNumber } from '@trezor/utils/src/bigNumber';
 import TrezorConnect, { FeeLevel, TokenInfo } from '@trezor/connect';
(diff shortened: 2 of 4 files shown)


Lemonexe (approved):
Codewise LGTM, works as expected :heavy_check_mark: 

:thought_balloon: Though wouldn't it make sense to introduce a similar message for all coins as well, not just tokens? Just a thought :thinking:

izmy (approved):

enjojoy:
> Codewise LGTM, works as expected ✔️
> 
> 💭 Though wouldn't it make sense to introduce a similar message for all coins as well, not just tokens? Just a thought 🤔

I think that would be good, yes. 

On mainnet now it just says "not enough funds" if you enter too much. But you also can do "send max" and it will set the max that is possible to send considering the gas, so it's easier to navigate then on tokens.

suite-common/wallet-core/src/send/sendFormEthereumThunks.ts
```diff
@@ -46,31 +46,43 @@ const calculate = (
...
-        const error = token ? 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE' : 'AMOUNT_IS_NOT_ENOUGH';
+        if (token) {
+            return {
+                type: 'error',
+                error: 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE_WITH_ETH_AMOUNT',
+                errorMessage: {
+                    id: 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE_WITH_ETH_AMOUNT',
+                    values: {
+                        symbol: 'ETH',
+                        feeAmount: `(${fromWei(feeInGwei, 'ether')} ETH)`,
```
tomasklim, line 75: This method is for all EVMs, why `ETH` is hardcoded?
Screenshot 2024-12-20 at 20 15 02

packages/suite/src/support/messages.ts
```diff
@@ -5426,6 +5426,11 @@ export default defineMessages({
         defaultMessage: 'Not enough {symbol} to cover transaction fee',
         id: 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE',
     },
+
+    AMOUNT_NOT_ENOUGH_CURRENCY_FEE_WITH_ETH_AMOUNT: {
+        defaultMessage: 'Not enough {symbol} to cover transaction fee {feeAmount}',
```
tomasklim, line 5431: How about
`Not enough {symbol} to cover transaction fee ({feeAmount} {symbol)',`

I dont like hardcoded symbol and ( ) in code, we try to keep the most of logic in crowdin. This will get broken in some languages

tomasklim:
> 💭 Though wouldn't it make sense to introduce a similar message for all coins as well, not just tokens? Just a thought 🤔

I don't know about coins, but it should be available for all networks with tokens at least

suite-common/wallet-core/src/send/sendFormEthereumThunks.ts
tomasklim, line 75: it would make sense to just return `feeAmount` as a number here. Symbol is not needed

suite-common/wallet-core/src/send/sendFormEthereumThunks.ts
```diff
@@ -227,10 +239,16 @@ export const composeEthereumTransactionFeeLevelsThunk = createThunk<
...
+            if (
+                tx.type === 'error' &&
+                tx.error === 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE_WITH_ETH_AMOUNT'
+            ) {
                 tx.errorMessage = {
-                    id: 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE',
-                    values: { symbol: network.symbol.toUpperCase() },
+                    id: 'AMOUNT_NOT_ENOUGH_CURRENCY_FEE_WITH_ETH_AMOUNT',
+                    values: {
+                        symbol: tx.errorMessage?.values?.symbol || 'ETH',
```
tomasklim, line 249: use `symbol` from account or `network.symbol`, never hardcode symbols in generic evm code

suite-common/wallet-utils/src/sendFormUtils.ts
```diff
@@ -75,21 +75,31 @@ export const calculateMax = (availableBalance: string, fee: string): string => {
...
+/**
+ * Calculate the Ethereum fee from gas price and gas limit.
+ * @param {string} [gasPrice] - The gas price in wei.
+ * @param {string} [gasLimit] - The gas limit.
+ * @returns {string} The calculated fee in wei, or '0' if inputs are invalid.
  */
 export const calculateEthFee = (gasPrice?: string, gasLimit?: string): string => {
     if (!gasPrice || !gasLimit) {
         return '0';
     }
```
tomasklim, line 87: maybe we can remove this? it looks duplicated with BN logic

suite-common/wallet-utils/src/sendFormUtils.ts
```diff
@@ -75,21 +75,31 @@ export const calculateMax = (availableBalance: string, fee: string): string => {
...
+
+    if (gasPriceBN.isNaN() || gasLimitBN.isNaN()) {
         return '0';
     }
+
+    const fee = gasPriceBN.times(gasLimitBN);
+
+    if (fee.isNaN()) {
+        return '0';
+    }
```
tomasklim, line 100: if both are numbers, can this ever happen?

suite-common/wallet-utils/src/sendFormUtils.ts
```diff
@@ -75,21 +75,31 @@ export const calculateMax = (availableBalance: string, fee: string): string => {
...
+    const gasLimitBN = new BigNumber(gasLimit);
+
+    if (gasPriceBN.isNaN() || gasLimitBN.isNaN()) {
         return '0';
     }
+
+    const fee = gasPriceBN.times(gasLimitBN);
+
+    if (fee.isNaN()) {
+        return '0';
```
tomasklim, line 99: Maybe it would make sense to throw error. There is some backup logic, and 0 fees are not valid

packages/suite/src/support/messages.ts
adderpositive: If it is ok, I made a little refactoring. Commit: https://github.com/trezor/trezor-suite/pull/15993/commits/0d79d6dff1cd5b390161fc528e10e65153248453 More information is in the related PR. Please give me your code review! :)

suite-common/wallet-core/src/send/sendFormEthereumThunks.ts
adderpositive: https://github.com/trezor/trezor-suite/pull/15993/commits/0d79d6dff1cd5b390161fc528e10e65153248453

suite-common/wallet-core/src/send/sendFormEthereumThunks.ts
adderpositive: https://github.com/trezor/trezor-suite/pull/15993/commits/0d79d6dff1cd5b390161fc528e10e65153248453

enjojoy:
@tomasklim  I'll create a followup PR to implement your comments

PR #16062, opened 2024-12-19
fix(suite-native): remove android foreground permissions

PeKne:
Fixes the Android Google Play Foreground service permissions declaration issue:
Snímek obrazovky 2024-12-19 v 12 59 11

Builded production app does not contain the problematic permission in the AndroidManifest.xml file anymore.

Inspired by https://github.com/expo/expo/issues/26846

### DIFF
--- a/suite-native/app/app.config.ts
+++ b/suite-native/app/app.config.ts
@@ -222,6 +222,8 @@ export default ({ config }: ConfigContext): ExpoConfig => {
                     category: ['BROWSABLE', 'DEFAULT'],
                 },
             ],
+            // These permission are included by expo-video, but we don't need them. The video is always played when is the app open and user is on the specific screen.
+            blockedPermissions: ['FOREGROUND_SERVICE', 'FOREGROUND_SERVICE_MEDIA_PLAYBACK'],
         },
         ios: {
             bundleIdentifier,


suite-native/app/app.config.ts
```diff
@@ -222,6 +222,8 @@ export default ({ config }: ConfigContext): ExpoConfig => {
                     category: ['BROWSABLE', 'DEFAULT'],
                 },
             ],
+            // These permission are included by expo-video, but we don't need them. The video is always played when is the app open and user is on the specific screen.
+            blockedPermissions: ['FOREGROUND_SERVICE', 'FOREGROUND_SERVICE_MEDIA_PLAYBACK'],
```
matejkriz, line 226: Isn't that supposed to be `android.permission.FOREGROUND_SERVICE` and `android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK`?

PeKne: Not sure what do you mean. IMO this is the correct shape. I've created a ad-hoc build before applying these changes and both these unwanted permissions were there.

After that I add them to this blacklist and created another build. These permissions were gone from `AndroidManifest.xml` after that. 

Or you mean something different?

matejkriz: I'm referring to the example in the documentation https://docs.expo.dev/guides/permissions/#android and the fact that the name of the permissions include that `android.permission.` prefix. But if you've tested that it works, and I tested that it doesn't break the the build also by creating an adhoc build, it looks good 👌

matejkriz (approved):
Tested on adhoc build that it does not break the build and app can run, and the video where to find XPUB also works fine 👌

PeKne:
notes for QA: please test, that the video explaining how to import Xpub is working as expected.

STew790:
**QA OK**
Video working just fine.

**Info**
25.1.2 ac8cd6e

PR #16077, opened 2024-12-20
fix: show proper info for unspendable coinbase transactions

### LINKED ISSUE #14270: Mined UTXOs not included in default Coin Control
**Describe the bug**
When creating a Send transaction with default Coin Control settings, I cannot send all my funds - mined UTXOs are not included. I have to open manual Coin Control and check the mined UTXOs in order to send them.

Tested **only** on regtest.

**Info:**

-   Remembered Wallet: no
-   Suite Version: web on production, desktop on develop
-   OS: Ubuntu 24.04
-   Bridge: 2.0.33
-   Firmware Version: T2T1, 2.8.1

**How to reproduce**

1. Use `trezor-user-env` regtest
1. In Suite, create regtest account no. 1
1. Send some amount to the regtest address
2. Mine some blocks at the regtest address
3. Create another regtest account no. 2
4. Open Send from regtest account no. 1
5. Send max
6. See that it is not the full amount! :bug:  
open Coin Control and see that mined UTXOs are not included by default, only received UTXOs are included

**Expected behavior**
When I don't interact with Coin Control settings, send max should include all UTXOs, including the mined.

**Screenshots**
In the video I have some regtest UTXOs that were recieved, and some that were mined (those are always 25).
[mined_regtest.webm](https://github.com/user-attachments/assets/e3d632bb-e37f-4725-ba9b-975ebff031d4)

peter-sanderson:
Resolve: https://github.com/trezor/trezor-suite/issues/14270

image

### DIFF
--- a/packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
+++ b/packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
@@ -1,10 +1,10 @@
-import { MouseEventHandler } from 'react';
+import { MouseEventHandler, ReactNode } from 'react';
 
 import styled, { css, useTheme } from 'styled-components';
 
 import { formatNetworkAmount, isSameUtxo } from '@suite-common/wallet-utils';
 import { Checkbox, Row, Spinner, TextButton, Tooltip } from '@trezor/components';
-import type { AccountUtxo } from '@trezor/connect';
+import { AccountUtxo } from '@trezor/connect';
 import { borders, spacings, spacingsPx, typography } from '@trezor/theme';
 import { CheckContainer } from '@trezor/components/src/components/form/Checkbox/Checkbox';
 
@@ -120,6 +120,36 @@ const StyledFiatValue = styled(FiatValue)`
     ${typography.hint}
 `;
 
+type ResolveUtxoSpendableProps = {
+    utxo: AccountUtxo;
+    coinjoinRegisteredUtxos: AccountUtxo[];
+};
+
+// Same as MINIMAL_COINBASE_CONFIRMATIONS in '@trezor/utxo-lib'; It is redeclared here to avoid
+// some magic import/export errors. This is very niche stuff and probably never changes.
+// Also, this most probably bothers only developers on Regtest.
+const MINIMAL_COINBASE_CONFIRMATIONS = 100;
+
+const resolveUtxoSpendable = ({
+    utxo,
+    coinjoinRegisteredUtxos,
+}: ResolveUtxoSpendableProps): ReactNode | null => {
+    if (utxo.coinbase === true && utxo.confirmations < MINIMAL_COINBASE_CONFIRMATIONS) {
+        return (
+            <Translation
+                id="TR_UTXO_NOT_MATURED_COINBASE"
+                values={{ confirmations: MINIMAL_COINBASE_CONFIRMATIONS }}
+            />
+        );
+    }
+
+    if (coinjoinRegisteredUtxos.includes(utxo)) {
+        return <Translation id="TR_UTXO_REGISTERED_IN_COINJOIN" />;
+    }
+
+    return null;
+};
+
 type UtxoSelectionProps = {
     transaction?: WalletAccountTransaction;
     utxo: AccountUtxo;
@@ -155,7 +185,10 @@ export const UtxoSelection = ({ transaction, utxo }: UtxoSelectionProps) => {
     const isChecked = isCoinControlEnabled
         ? selectedUtxos.some(selected => isSameUtxo(selected, utxo))
         : composedInputs.some(u => u.prev_hash === utxo.txid && u.prev_index === utxo.vout);
-    const isDisabled = coinjoinRegisteredUtxos.includes(utxo);
+
+    const unspendableTooltip = resolveUtxoSpendable({ utxo, coinjoinRegisteredUtxos });
+    const isDisabled = unspendableTooltip !== null;
+
     const utxoTagIconColor = isDisabled
         ? theme.legacy.TYPE_LIGHT_GREY
         : theme.legacy.TYPE_DARK_GREY;
@@ -174,7 +207,7 @@ export const UtxoSelection = ({ transaction, utxo }: UtxoSelectionProps) => {
             $isDisabled={isDisabled}
             onClick={isDisabled ? undefined : handleCheckbox}
         >
-            <Tooltip content={isDisabled && <Translation id="TR_UTXO_REGISTERED_IN_COINJOIN" />}>
+            <Tooltip content={unspendableTooltip}>
                 <Checkbox
                     isChecked={isChecked}
                     isDisabled={isDisabled}
--- a/packages/utxo-lib/src/coinselect/outputs/split.ts
+++ b/packages/utxo-lib/src/coinselect/outputs/split.ts
@@ -8,6 +8,7 @@ import {
     getFee,
     finalize,
     ZERO,
+    MINIMAL_COINBASE_CONFIRMATIONS,
 } from '../coinselectUtils';
 import {
     CoinSelectInput,
@@ -23,7 +24,7 @@ export function split(
     feeRate: number,
     options: CoinSelectOptions,
 ): CoinSelectResult {
-    const coinbase = options.coinbase || 100;
+    const coinbase = options.coinbase || MINIMAL_COINBASE_CONFIRMATIONS;
     const utxos = filterCoinbase(utxosOrig, coinbase);
 
     const fee = getFee(utxos, outputs, feeRate, options);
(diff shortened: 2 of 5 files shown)


packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
```diff
@@ -26,6 +26,7 @@ import {
 } from 'src/reducers/suite/metadataReducer';
 
 import { UtxoTag } from './UtxoTag';
+import { MINIMAL_COINBASE_CONFIRMATIONS } from '@trezor/utxo-lib';
```
peter-sanderson: I think this is wrong. @marekrjpolak any ideas how to propagate/share this correctly? :pray:

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
```diff
@@ -120,6 +121,31 @@ const StyledFiatValue = styled(FiatValue)`
...
+type ResolveUtxoSpendableProps = {
+    utxo: AccountUtxo;
+    coinjoinRegisteredUtxos: AccountUtxo[];
+};
+
+const resolveUtxoSpendable = ({
+    utxo,
+    coinjoinRegisteredUtxos,
+}: ResolveUtxoSpendableProps): ReactNode | null => {
+    if (utxo.coinbase === true && utxo.confirmations < MINIMAL_COINBASE_CONFIRMATIONS) {
```
peter-sanderson, line 137: In theory this `100` can be different for different coins. There is some support for this in utxo-lib. However its never actually used. 

The fallback to `100` is always used: 
image

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
Lemonexe: Why do you think it to be wrong? I'm honestly asking, I see `utxo-lib` is never directly used even in `suite-common`, let alone suite, only through blockchain-link. But a subdependency of `blockchain-link`, I don't see why not?

packages/utxo-lib/src/coinselect/outputs/split.ts
```diff
@@ -23,7 +24,7 @@ export function split(
     feeRate: number,
     options: CoinSelectOptions,
 ): CoinSelectResult {
-    const coinbase = options.coinbase || 100;
+    const coinbase = options.coinbase || MINIMAL_COINBASE_CONFIRMATIONS;
```
Lemonexe, line 27: since you're already modifying it, could you please rename `coinbase` variable? From the name it's not even apparent that it is a number. I'd suggest `minConfCoinbase` to match the argument of `filterCoinbase` :slightly_smiling_face:

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
peter-sanderson: I consulted this with @marekrjpolak and we ended-up in just hardcoding it in the `suite` in the component (duplicating it). See comment for more info.

peter-sanderson: Here: https://github.com/trezor/trezor-suite/pull/16077/files#diff-3e5e89e7b9790bac298fc6b3f99fe1e7cffc68c9ef52f50ece21adab49a76e95R131

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
```diff
@@ -120,6 +120,36 @@ const StyledFiatValue = styled(FiatValue)`
...
 `;
 
+type ResolveUtxoSpendableProps = {
+    utxo: AccountUtxo;
+    coinjoinRegisteredUtxos: AccountUtxo[];
+};
+
+// Same as MINIMAL_COINBASE_CONFIRMATIONS in '@trezor/utxo-lib'; It is redeclared here to avoid
+// some magic import/export errors. This is very niche stuff and probably never changes.
+// Also, this most probably bothers only developers. Who mines directly to Trezor?
```
Lemonexe: Yeah, most probably true :smile:

packages/utxo-lib/src/coinselect/tryconfirmed.ts
```diff
@@ -31,7 +31,7 @@ export function tryConfirmed(
 ): CoinSelectAlgorithm {
     const own = options.own || 1;
     const other = options.other || 6;
-    const coinbase = options.coinbase || 100;
+    const coinbase = options.coinbase || MINIMAL_COINBASE_CONFIRMATIONS;
```
Lemonexe, line 34: same as in `split.ts`

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
```diff
@@ -120,6 +120,36 @@ const StyledFiatValue = styled(FiatValue)`
...
+
+// Same as MINIMAL_COINBASE_CONFIRMATIONS in '@trezor/utxo-lib'; It is redeclared here to avoid
+// some magic import/export errors. This is very niche stuff and probably never changes.
+// Also, this most probably bothers only developers on Regtest.
+const MINIMAL_COINBASE_CONFIRMATIONS = 100;
+
+const resolveUtxoSpendable = ({
+    utxo,
+    coinjoinRegisteredUtxos,
+}: ResolveUtxoSpendableProps): ReactNode | null => {
```
Lemonexe, line 136: as said below, this is an `FC`. So no need for return type `ReactNode | null` (btw null is assignable to ReactNode, so that can be simplified to `ReactNode`).

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
```diff
@@ -155,7 +185,10 @@ export const UtxoSelection = ({ transaction, utxo }: UtxoSelectionProps) => {
     const isChecked = isCoinControlEnabled
         ? selectedUtxos.some(selected => isSameUtxo(selected, utxo))
         : composedInputs.some(u => u.prev_hash === utxo.txid && u.prev_index === utxo.vout);
-    const isDisabled = coinjoinRegisteredUtxos.includes(utxo);
+
+    const unspendableTooltip = resolveUtxoSpendable({ utxo, coinjoinRegisteredUtxos });
```
Lemonexe, line 189: `resolveUtxoSpendable` function is actually a component, so let's rename it and use it as such:

Something like this:
```suggestion
    const unspendableTooltip = <UtxoUnspendableTooltip utxo={utxo} coinjoinRegisteredUtxos={coinjoinRegisteredUtxos} />
```

Lemonexe (commented):
Works correctly, but I'd refactor the function/component, see comments

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
Lemonexe: After discussion, this is fine, just unclear naming.

Lemonexe (approved):
LGTM, tested with regtest and works correctly

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
peter-sanderson: solved in person

packages/suite/src/views/wallet/send/Options/BitcoinOptions/CoinControl/UtxoSelectionList/UtxoSelection/UtxoSelection.tsx
peter-sanderson: solved in person

packages/utxo-lib/src/coinselect/outputs/split.ts
peter-sanderson: solve in person

packages/utxo-lib/src/coinselect/tryconfirmed.ts
peter-sanderson: solved in person

PR #16091, opened 2024-12-20
Use Tor daemon from Tor browser experimental

karliatto:
## Description

Allow user to configure Tor External Port to use "Custom Tor (9050)" or "Tor Browser (9150)
The main idea of this is to make the usage of external Tor daemon easier. In case use wants to use Tor daemon from Tor Browser with some Tor bridge they can. So they do not have do it themselves and they can for example configure snowflake in Tor Browser and use it in Trezor Suite.

We can change the UI how it is presented to the user. And we need to update documentation, but this makes it all quite easier for user.

:thinking: 
I am thinking if adding new option to the select (Tor bundled) will make it better, like:

* Tor bundled
* Tor external custom (9050)
* Tor from Tor browser (9150)

I haven't done it like that since Tor Bundled is used when Tor external is disabled.

## Screenshots:

image

### DIFF
--- a/packages/suite-desktop-api/src/factory.ts
+++ b/packages/suite-desktop-api/src/factory.ts
@@ -125,7 +125,9 @@ export const factory = <R extends StrictIpcRenderer<any, IpcRendererEvent>>(
         getTorSettings: () => ipcRenderer.invoke('tor/get-settings'),
 
         changeTorSettings: payload => {
-            if (validation.isObject({ useExternalTor: 'boolean' }, payload)) {
+            if (
+                validation.isObject({ useExternalTor: 'boolean', externalPort: 'number' }, payload)
+            ) {
                 return ipcRenderer.invoke('tor/change-settings', payload);
             }
 
--- a/packages/request-manager/src/controllerExternal.ts
+++ b/packages/request-manager/src/controllerExternal.ts
@@ -40,6 +40,10 @@ export class TorControllerExternal extends EventEmitter {
         this.status = TOR_CONTROLLER_STATUS.ExternalTorRunning;
     }
 
+    public updatePort(port: number) {
+        this.options.port = port;
+    }
+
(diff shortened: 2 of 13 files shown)


packages/suite-desktop-api/src/factory.ts
```diff
@@ -125,7 +125,10 @@ export const factory = <R extends StrictIpcRenderer<any, IpcRendererEvent>>(
         getTorSettings: () => ipcRenderer.invoke('tor/get-settings'),
 
         changeTorSettings: payload => {
-            if (validation.isObject({ useExternalTor: 'boolean' }, payload)) {
+            // TODO: what happens if `payload` does not contain `externalPort` but the rest of the keys?
```
karliatto: I still want to check what happens in this scenario.

karliatto: I tested and seams to work properly.

marekrjpolak (approved):
Looks good and seems to be working. I was briefly able to use nonexistent external Tor on port 9150 (it silently used bundled one instead), but then I wiped storage and wasn't able to reproduce again, so I believe it was just a glitch.

bosomt:
**QA OK**

image

image

**Info:**
 - Suite version: desktop 0.25.1 (4403882efb2a4a802c3d77917d6350a4fae7663d)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/0.25.1 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3B1 2.8.7 regular (revision f2aeea159f1fde98d4fc5575350a8c9fc4478367)
 - Transport: BridgeTransport 3.0.0-bundled.0.25.1

PR #16094, opened 2024-12-22
feat(suite): add posssibility to do a wallet backup before fw update

MiroslavProchazka:
Adding the option to perform a wallet backup prior to a firmware update

Todo:

- [ ] Copy adjustements
- [ ] UI approval
- [ ] Get rid of the releases.json edits

**Changes**
Before
image

After
image

Update flow

https://github.com/user-attachments/assets/2f2fe2ba-c687-460b-bb08-c9c5c3b79fa2


packages/connect-common/files/firmware/t2b1/releases.json
```diff
@@ -1,4 +1,18 @@
 [
+    {
```
MiroslavProchazka, line 2: This is here just for easier testing, needs to be deleted before merge!

packages/suite/src/support/messages.ts
```diff
@@ -4653,6 +4653,10 @@ export default defineMessages({
         defaultMessage:
             'In all cases, we strongly recommend <button>creating a wallet backup in Settings</button>. With a wallet backup, your coins are safe and secure even if your Trezor is lost, damaged, or stolen.',
     },
+    TR_CHECK_BACKUP: {
+        id: 'TR_CHECK_BACKUP',
+        defaultMessage: 'CHeck backup!',
```
tomasklim: **Ch**eck backup

packages/suite-data/files/translations/en.json
```diff
@@ -353,11 +353,11 @@
   "TR_BACKUP_SUBHEADING_1": "A wallet backup is a series of randomly generated words created by your Trezor. It’s important to write down your wallet backup and keep it safe, as it's the only way to recover and access your funds.",
   "TR_BACKUP_SUCCESSFUL": "Backup successful",
   "TR_BALANCE": "Balance",
+  "TR_CHECK_BACKUP": "Check backup",
   "TR_BASIC_RECOVERY": "Standard recovery",
   "TR_BASIC_RECOVERY_OPTION": "Enter your wallet backup word by word on your computer.",
   "TR_BCH_ADDRESS_INFO": "Bitcoin Cash changed the address format to cashaddr. Find more info about how to convert your address on our blog. {TR_LEARN_MORE}",
-  "TR_BEFORE_ANY_FURTHER_ACTIONS": "Although unlikely, you may need to access your wallet backup in case of a firmware update issue.",
-  "TR_BIP_SIG_FORMAT": "Trezor",
+  "TR_BEFORE_ANY_FURTHER_ACTIONS": "Although unlikely, you may need to access your wallet backup in the event of a firmware update issue. If you’re uncertain about the accuracy of your backup, you can verify it before proceeding with the installation.",
```
tomasklim, line 360: the copy seems too long. I would not read it. How about using yellow warning banner

tomasklim (changes requested):

packages/suite/src/components/firmware/CheckSeedStep.tsx
```diff
@@ -123,6 +124,10 @@ export const CheckSeedStep = ({ deviceWillBeWiped, onClose, onSuccess }: CheckSe
 
     const { heading, description, checkbox } = getContent();
 
+    const handleCheckBackupClick = () => {
```
tomasklim:
```suggestion
    const onCheckBackup = () => {
```

packages/suite/src/support/messages.ts
MiroslavProchazka: a2ddf5cae381d432682cf78f74f1f55ec70423a2

packages/suite/src/components/firmware/CheckSeedStep.tsx
MiroslavProchazka: a2ddf5cae381d432682cf78f74f1f55ec70423a2

peter-sanderson (approved):

peter-sanderson:
I am not sure about the order of the buttons (and variant)
image

Th backup on the right with primary variant seems a bit odd (I would expect continue there).

:pray: check with product/design (if you already hadn't done that, in that case please ignore my comment)

packages/suite/src/components/firmware/CheckSeedStep.tsx
```diff
@@ -139,6 +144,9 @@ export const CheckSeedStep = ({ deviceWillBeWiped, onClose, onSuccess }: CheckSe
                             id={deviceWillBeWiped ? 'TR_WIPE_AND_REINSTALL' : 'TR_CONTINUE'}
                         />
                     </Button>
+                    <Button onClick={onCheckBackup} variant="tertiary">
```
komret, line 147: We should only display this if the device is backed up. Check for `isBackedUp`.

packages/suite/src/support/messages.ts
```diff
@@ -4653,6 +4653,10 @@ export default defineMessages({
         defaultMessage:
             'In all cases, we strongly recommend <button>creating a wallet backup in Settings</button>. With a wallet backup, your coins are safe and secure even if your Trezor is lost, damaged, or stolen.',
     },
+    TR_CHECK_BACKUP: {
```
komret, line 4656: Let's just use `TR_CHECK_RECOVERY_SEED`, I don't think we need a new string for this.

komret (changes requested):

packages/suite/src/components/firmware/CheckSeedStep.tsx
```diff
@@ -123,6 +124,10 @@ export const CheckSeedStep = ({ deviceWillBeWiped, onClose, onSuccess }: CheckSe
 
     const { heading, description, checkbox } = getContent();
 
+    const onCheckBackup = () => {
+        dispatch(goto('recovery-index'));
```
komret, line 128: You'll also need to call `resetReducer` from `useFirmwareInstallation` here (before redirecting), otherwise the modal will reopen on this step instead of the first one when you open it next time.

komret:
Closing in favour of https://github.com/trezor/trezor-suite/pull/16144 🙃

PR #16102, opened 2024-12-24
fix(suite-native): remove tab bar from settings subscreens

yanascz:
Tab bar is not shown on settings subscreens. Based on [Settings tabbar & coin enabling in settings](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Small-Things-%5BMOB%5D?node-id=1909-60437&t=03RMTjusxkOfLKLs-4) in Figma.

## Screenshots:
Simulator Screenshot - iPhone 16 Pro - 2024-12-24 at 11 13 04
Simulator Screenshot - iPhone 16 Pro - 2024-12-24 at 11 13 10

### DIFF
--- a/suite-native/app/src/navigation/RootStackNavigator.tsx
+++ b/suite-native/app/src/navigation/RootStackNavigator.tsx
@@ -25,6 +25,7 @@ import { SendStackNavigator } from '@suite-native/module-send';
 import { CoinEnablingInitScreen } from '@suite-native/coin-enabling';
 import { ConnectPopupScreen, useConnectPopupNavigation } from '@suite-native/module-connect-popup';
 import { StakingDetailScreen } from '@suite-native/module-staking-management';
+import { SettingsStackNavigator } from '@suite-native/module-settings';
 
 import { AppTabNavigator } from './AppTabNavigator';
 import { useCoinEnablingInitialCheck } from '../hooks/useCoinEnablingInitialCheck';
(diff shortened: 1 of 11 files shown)


suite-native/app/src/navigation/RootStackNavigator.tsx
```diff
@@ -25,6 +25,7 @@ import { SendStackNavigator } from '@suite-native/module-send';
 import { CoinEnablingInitScreen } from '@suite-native/coin-enabling';
 import { ConnectPopupScreen, useConnectPopupNavigation } from '@suite-native/module-connect-popup';
 import { StakingDetailScreen } from '@suite-native/module-staking-management';
+import { SettingsScreenStackNavigator } from '@suite-native/module-settings';
```
yanascz, line 28: I'm not super happy about the `SettingsScreen*` naming pattern but couldn't figure out a better one. Already tried `SettingsSection*` but that conflicts with the `SettingsSection` component which has bit different semantics. @shenkys uses _settings subpage_ but we don't have that concept in code at all (expected for pagination). Ideas? 🤔

PeKne:
What about removing the `SettingsStackNavigator`? Having a stack with a single Screen seems redundant to me. You could directly place the `SettingsScreen` route to the `RootStackNavigator`. It would make the code simpler.  

As a side effect, you can keep the `SettingsStackNavigator` name for the "new" stack and you don't have to think about another stack name.

What do you think? 

Apart from this LGTM.

yanascz:
> What about removing the `SettingsStackNavigator`? Having a stack with a single Screen seems redundant to me. You could directly place the `SettingsScreen` route to the `RootStackNavigator`. It would make the code simpler.

If you look into `AppTabNavigator`, you'll notice that all tabs use a navigation stack. I couldn't get it working without a stack, not sure why. If you know how to achieve the above, please let me know.

> As a side effect, you can keep the `SettingsStackNavigator` name for the "new" stack and you don't have to think about another stack name.

I guess we could also inline `SettingsScreenStackRoutes` values into `RootStackRoutes` but I don't think we want that.

PeKne (approved):
Thanks for addressing the naming comment 🙏

STew790:
**QA OK**
Tab bar removed from all settings subscreens.
image

**Info**
25.1.2 caac07f

PR #16105, opened 2024-12-26
fix(suite-native): show spinner for pending transactions in lists

yanascz:
## Screenshots:

image

### DIFF
--- a/suite-native/notifications/src/components/TransactionNotification.tsx
+++ b/suite-native/notifications/src/components/TransactionNotification.tsx
@@ -125,7 +125,6 @@ export const TransactionNotification = ({
                     transactionType={transactionType}
                     symbol={notification.symbol}
                     isAnimated={isIconAnimated}
-                    iconColor={isIconAnimated ? 'iconAlertYellow' : 'iconSubdued'}
                 />
             }
             iconRight={<Icon name="caretCircleRight" />}
--- a/suite-native/transactions/src/components/TransactionsList/TransactionListItemContainer.tsx
+++ b/suite-native/transactions/src/components/TransactionsList/TransactionListItemContainer.tsx
@@ -4,6 +4,7 @@ import { useSelector } from 'react-redux';
 
 import { useNavigation } from '@react-navigation/native';
 
+import { isPending } from '@suite-common/wallet-utils';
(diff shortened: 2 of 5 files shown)


suite-native/transactions/src/components/TransactionsList/TransactionListItemContainer.tsx
```diff
@@ -155,7 +150,9 @@ export const TransactionListItemContainer = ({
                     contractAddress={contractAddress}
                     transactionType={transactionType}
                     isAnimated={isTransactionPending}
-                    iconColor={iconColor}
+                    iconColor="iconDefault"
+                    spinnerColor="backgroundAlertYellowBold"
```
PeKne: what about setting these values of `spinnerColor` and `spinnerWidth` as default in the `TransactionIcon` component? So if you pass `isAnimated=true` then there will be animation displayed by default? It would make sense to me. Up to you.

PeKne (approved):

suite-native/transactions/src/components/TransactionsList/TransactionListItemContainer.tsx
yanascz: I thought about it too but decided not to do that. However, you motivated me to be even more radical. 😅

suite-native/notifications/src/components/TransactionNotification.tsx
```diff
@@ -125,7 +125,6 @@ export const TransactionNotification = ({
                     transactionType={transactionType}
                     symbol={notification.symbol}
                     isAnimated={isIconAnimated}
-                    iconColor={isIconAnimated ? 'iconAlertYellow' : 'iconSubdued'}
```
yanascz, line 128: @shenkys explicitly asked for the icon to use the default colour no matter what. I missed this in the initial commit.

suite-native/transactions/src/components/TransactionsList/TransactionListItemContainer.tsx
PeKne: makes sense to me 👍✅

STew790:
**QA OK**
LGTM 
image

**Info**
25.1.2 caac07f

Issue #15919, opened 2024-12-11
Redact account address from Account not found errors


matejkriz:
We report all graph errors to Sentry from `checkAndReportGraphError` and we have to reduct account address if the error contains it. More info[ in Notion ](https://www.notion.so/satoshilabs/Sentry-Mobile-Crashlytics-e00e823e84a34024b908c696dcffe5d0?pvs=4#159dc52606068057af75cb85e4fc9e82)

PR #16122, opened 2024-12-31
chore(suite-native): redact more graph error messages

vytick:
redact more graph error messages.

followup to #15919

### DIFF
--- a/suite-native/graph/src/hooks.ts
+++ b/suite-native/graph/src/hooks.ts
@@ -68,24 +68,37 @@ const useWatchTimeframeChangeForAnalytics = (
     }, [timeframeHours, symbol, isFirstRender]);
 };
 
-const redactAfterSubstring = (string: string, substring: string) => {
-    const index = string.indexOf(substring);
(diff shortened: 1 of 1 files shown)


suite-native/graph/src/hooks.ts
```diff
@@ -82,6 +82,8 @@ const redact = (string: string) => {
     const redactAfterSubstringArray = [
         'Account not found:',
         'Unable to fetch fiat rates for defined timestamps. ',
+        'Aborted by timeout -',
+        'POL: getTransaction ',
```
PeKne: are we interested in POL only? Are errors of other networks irrelevant?

PeKne (approved):

suite-native/graph/src/hooks.ts
vytick: I originally thought that no other blockbook has similar issues, but now I realised, that BNB, ARB and OP can have similar problem. good point!

Making the redacted message more relevant in 2e77dd4782

PR #16117, opened 2024-12-29
Bunch of fixes

tomasklim:
## Description

- fix removed staking for Holesky
  - https://github.com/trezor/trezor-suite/pull/15467/commits/46c6ec8037fcb74600788d09536e95a0d641321c#diff-8275801eb25562af454cd9d02084b668b58a396ff45c85242da60e8a203c46fdR549	
- move explorer util to separate file (cannot be in utils because of cycle)
- remove forgotten console log in IOOutput 
- fix symbol in review tx modal
- wrap instant staking badge when it does not fit
- fix en.json to avoid errors in console (crowdin is already changed)

## Screenshots:

Screenshot 2024-12-29 at 17 56 56
Screenshot 2024-12-29 at 17 52 58

PR #16129, opened 2025-01-02
fix(suite): staking in hol

enjojoy:
Fix accidentally removed staking for holesky


tomasklim:
Resolved here https://github.com/trezor/trezor-suite/pull/16117

PR #16131, opened 2025-01-02
Coins deprecation

MiroslavProchazka:
As of February 2025, Trezor Suite will discontinue support for Dash, Bitcoin Gold, DigiByte, Namecoin, and Vertcoin. Users are advised to migrate their funds to alternative wallets to ensure continued access and security. The coins will still be supported on the FW level, so users can still access them using third party wallets.

**Proposed Timeline:**

- February Suite Release: This version of Suite will no longer support the affected coins. However, users can still use older versions of Suite to access their coins.
- July 2025: The BEs will be turned off. From this point forward, third-party apps will be the only way to access the coins.

I will divide the commits by environment (desktop/native) to make code review easier.

**`Please read the following before conducting the code review:`**

- Since the coins will still be supported by the firmware and third-party apps, we should maintain support at the Connect level. Similarly, we will continue support in blockchain-link and utxo-lib.
- For now, we have decided to keep these coins in Trading (users will need to use external addresses for this), so we need to retain the Address Validator as well.
- We need to create a migration for users with remembered wallets.

Given the points above, I have prepared only basic deprecation that mainly affects NetworksConfig. Everything else (tests, icons, etc.) will remain unchanged until we decide to disable trading options and/or turn off the backends.

Todo:

- [x] In case that the coins will be still available for trading (to external wallets), can we delete their icons, names in `messages.ts` and other mentions in general files?
- [x] Should I also delete all tests mentioning those coins?
- [x] As some of the coins files needs to stay in the codespace (address validator, blockchain-link, connect etc.), it would be wise to determinate minimal scope for just Suite deprecation - will this be just `NetworksConfig`? 
- [x] Adjust coins.ts
- [x] Adjust /trezor-suite/suite-common/suite-constants/src/protocol.ts
- [x] Create migration for users with remembered wallet - Suite
- [x] Create migration for users with remembered wallet - Suite lite
- [x] Create migration for migrateEnabledDiscoveryNetworkSymbols - discovery

--------------------------------------------------------------------------------------------------------------------------
This pull request includes significant changes related to the removal of support for several cryptocurrencies across multiple files. The most important changes include updating the `LegacyNetworkSymbol` type, modifying protocol definitions, and removing network configurations and symbols.

Removal of support for specific cryptocurrencies:

* [`packages/product-components/src/components/CoinLogo/coins.ts`](diffhunk://#diff-44fb9b78e52a51f30c99b662835aba27a933271e2d72df4772e25956b488c0aaL4-R13): Updated `LegacyNetworkSymbol` to include additional unsupported coins.
* [`suite-common/suite-constants/src/protocol.ts`](diffhunk://#diff-4b8b611e101cf1e19dacdb0a404b981058e1adfb8bc9d54348314664527c26e7L8-L13): Removed protocols for `bitcoingold`, `dash`, `digibyte`, `namecoin`, and `vertcoin`. [[1]](diffhunk://#diff-4b8b611e101cf1e19dacdb0a404b981058e1adfb8bc9d54348314664527c26e7L8-L13) [[2]](diffhunk://#diff-4b8b611e101cf1e19dacdb0a404b981058e1adfb8bc9d54348314664527c26e7L61-L66)
* [`suite-common/wallet-config/src/networksConfig.ts`](diffhunk://#diff-8275801eb25562af454cd9d02084b668b58a396ff45c85242da60e8a203c46fdL401-L494): Removed network configurations for `dash`, `btg`, `dgb`, `nmc`, and `vtc`.
* [`suite-common/wallet-config/src/types.ts`](diffhunk://#diff-e942c6bc5783a94562dd2bd56503824d6ffb70f56a448befe895f978282ddc4eL11-L16): Updated `NetworkSymbol` to remove symbols for `btg`, `dash`, `dgb`, `nmc`, and `vtc`.
* [`suite-native/config/src/supportedNetworks.ts`](diffhunk://#diff-18a7d1f1fa86a4d94e72d161fb1360b56a33ce2c53226671efa9f224ca67d732L25-L48): Updated the whitelist and removed the blacklist for unsupported coins. [[1]](diffhunk://#diff-18a7d1f1fa86a4d94e72d161fb1360b56a33ce2c53226671efa9f224ca67d732L25-L48) [[2]](diffhunk://#diff-18a7d1f1fa86a4d94e72d161fb1360b56a33ce2c53226671efa9f224ca67d732L81-R61)


suite-native/storage/src/migrations/account/v3.ts
```diff
@@ -18,12 +18,7 @@ type NetworkSymbol =
     | 'etc'
     | 'xrp'
     | 'bch'
-    | 'btg'
```
vytick, line 21: I think we should keep those in this migration

suite-native/config/src/supportedNetworks.ts
```diff
@@ -78,9 +58,7 @@ export const filterBlacklistedNetworks = (
     );
 
 export const portfolioTrackerMainnets = sortNetworks(
```
vytick, line 60: If I see correctly, if we remove the portfolio blacklist, we can remove `portfolioTrackerMainnets` and `getPortfolioTrackerTestnets` completely and use `networkSymbolsWhitelistMap.mainnets` and `networkSymbolsWhitelistMap.testnets` instead, wdyt?

vytick, line 60: If I see correctly, if we remove the portfolio blacklist, we can remove `portfolioTrackerMainnets` and `getPortfolioTrackerTestnets` completely and use `networkSymbolsWhitelistMap.mainnets` and `networkSymbolsWhitelistMap.testnets` instead, wdyt?

suite-native/storage/src/migrations/account/v3.ts
vytick: I think we should keep those in this migration

mroz22: yes. maybe there should be migration removing target deprecated coins accounts from state as well. but its not that big deal. it can stay there forever and nothing will probably ever happen

MiroslavProchazka: I will prepare migration for users with remembered wallet.

MiroslavProchazka: Done

suite-common/suite-constants/src/protocol.ts
```diff
@@ -5,12 +5,7 @@ export type Protocol =
     | 'btc'
     | 'litecoin'
     | 'bitcoincash'
-    | 'bitcoingold'
-    | 'dash'
-    | 'digibyte'
     | 'dogecoin'
-    | 'namecoin'
-    | 'vertcoin'
```
marekrjpolak, line 13: `btg` | `dgb` | `nmc` | `vtc` below should also be removed, right?

packages/suite/src/storage/migrations/index.ts
```diff
@@ -1176,4 +1178,56 @@ export const migrate: OnUpgradeFunc<SuiteDBSchema> = async (
             return account;
         });
     }
+
+    // Deprecate Vertcoin (VTC) and other networks
+    if (oldVersion < 52) {
```
marekrjpolak, line 1183: You forgot about this:
```
const VERSION = 51; // don't forget to add migration and CHANGELOG when changing versions!
```

marekrjpolak (approved):
Apart from mentioned, it looks good (= same as e.g. `tgor` deprecation).

HajekOndrej (approved):

tomasklim (changes requested):
You forgot to remove icons from `suite-common/icons`, `suite-common/icons-deprecated` packages

packages/suite/src/storage/migrations/index.ts
```diff
@@ -1176,4 +1178,56 @@ export const migrate: OnUpgradeFunc<SuiteDBSchema> = async (
...
+        const deprecatedNetworks = ['vtc', 'btg', 'nmc', 'dgb', 'dash'];
+
+        // Remove transactions related to deprecated networks
+        await updateAll(transaction, 'txs', tx => {
```
tomasklim, line 1187: I think you miss `graph`, `historicRates` and `sendFormDrafts`

marekrjpolak: This is true, but historically we didn't handle these when removing coins, only when changing symbols (because we wanted to keep the data), so I believe this should be safe as well.

suite-native/config/src/supportedNetworks.ts
```diff
@@ -22,31 +22,10 @@ const discoveryBlacklist: NetworkSymbol[] = ['op', 'base', 'arb'];
...
-        'bch',
-        'btg',
-        'dash',
-        'dgb',
-        'doge',
-        'nmc',
-        'vtc',
-        'zec',
-    ],
-    testnet: ['test', 'regtest', 'tsep', 'thol', 'dsol', 'tada', 'txrp'],
```
MiroslavProchazka, line 44: Add "dsol" here

suite-native/config/src/supportedNetworks.ts
```diff
@@ -22,31 +22,10 @@ const discoveryBlacklist: NetworkSymbol[] = ['op', 'base', 'arb'];
...
+    mainnet: ['btc', 'eth', 'pol', 'bsc', 'ltc', 'etc', 'ada', 'xrp', 'bch', 'doge', 'zec'],
```
MiroslavProchazka: Add "sol" here

suite-common/suite-constants/src/protocol.ts
MiroslavProchazka: Done in [chore(suite): basic deprecation on suite level]

MiroslavProchazka:
> You forgot to remove icons from `suite-common/icons`, `suite-common/icons-deprecated` packages

As this affects additional files and tests and it is not necessary for the deprecation, I will leave it for later cleanup together with BE deprecation of these coins

suite-native/config/src/supportedNetworks.ts
MiroslavProchazka: Done

suite-native/config/src/supportedNetworks.ts
MiroslavProchazka: Done

packages/suite/src/storage/migrations/index.ts
MiroslavProchazka: Done in [chore(suite): add migration for deprecated coins]

tomasklim (approved):

Vere-Grey (approved):

packages/suite/src/storage/migrations/index.ts
```diff
@@ -1176,4 +1178,56 @@ export const migrate: OnUpgradeFunc<SuiteDBSchema> = async (
...
+            return discovery;
+        });
+
+        // Remove deprecated networks from backend settings
+        const backendSettings = transaction.objectStore('backendSettings');
+        for (const network of deprecatedNetworks) {
+            await backendSettings.delete(network as NetworkSymbol); // Delete backend settings for each deprecated network
+        }
+
+        console.log(
```
matejkriz: This console.log is leftover, right?

MiroslavProchazka: Yes, thank you. We will use it for testing, and I will remove it afterward.

suite-native/storage/src/migrations/discovery/v3.ts
```diff
@@ -0,0 +1,10 @@
+import { NetworkSymbol } from '@suite-common/wallet-config';
+// @ts-expect-error
+const deprecatedNetworks: NetworkSymbol[] = ['dash', 'btg', 'nmc', 'vtc', 'dgb'];
+
+export const migrateDiscoveryDeprecateNetworks = (
```
matejkriz, line 5: Those suite-native migrations are not applied at all.

matejkriz (changes requested):

suite-native/config/src/supportedNetworks.ts
MiroslavProchazka: We need to create issue for this

komret (approved):

vytick (approved):
Good job 👍. I have tested it and it works as expected 🦾 . 

Please, create issue for redundant `portfolioTrackerMainnets` and remove that log as you discussed in comments

suite-native/storage/src/migrations/discovery/v3.ts
MiroslavProchazka: Solved in b71ef01d4184c3fba8133090f58c733ead8feb94

packages/suite/src/storage/migrations/index.ts
MiroslavProchazka: Resolved in 0d95d725cd80e3ca5ad9b52670d9be16eef34145

suite-native/config/src/supportedNetworks.ts
MiroslavProchazka: In here - https://github.com/trezor/trezor-suite/issues/16683

matejkriz (approved):
Good job 👏  The only issue is that you don't remove enough coins, I would delete a few more 😉 

One nitpick is that there is missing migration for graph state (we stored the preferred time range per coin), but it's not deleted properly even if you just disable any coin, so I would keep it as it is. It is quite hard to filter those and it's just a small data, most users will not have it anyway...)

Works as expected, I was checking the redux state after migration from develop to this branch.

bosomt:
**QA OK**

tested during autoupdate testing /remembered wallet + enabled deprecated coins/

**Info:**
 - Suite version: desktop 25.2.1 (4b92a096a8ac719cea3917d5c0fba5f6d2c97810)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.2.1 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Transport: BridgeTransport 3.0.0-bundled.25.2.1

STew790:
**QA OK**
Ok on mobile as well, coins unavailable for enable in settings/portfolio/add account.
Also tested with remembered device with enabled said coins and they are not shown/available.

**Info**
25.2.1 246db18

PR #16134, opened 2025-01-02
Message-system 73 - enable fw hash check

MiroslavProchazka:
Re-enabling FW hash check for all Suite version greater than 24.12.3, so we can start gathering inputs to Sentry

Tracked in Notion [here](https://www.notion.so/satoshilabs/Turn-on-FW-hash-check-for-Suite-24-12-3-16fdc526060680009347f84f610e7290?pvs=4)

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
     "timestamp": "2024-01-02T00:00:00+00:00",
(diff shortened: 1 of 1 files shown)


komret (approved):
I tested this on 24.9. release branch to check whether this change doesn't break it. It doesn't affect it. I also checked develop where the check is performed correctly.

tomasklim (approved):

PR #16136, opened 2025-01-02
Fix same labels for all evm accounts

### LINKED ISSUE #15546: Ethereum like networks share same labels
**Describe the bug**
Ethereum like networks share same labels

**Steps to reproduce:**
1. Create label for Eth n.1 account
2. Add Polygon and BNB accounts/networks coins
3. Observe labels of those accounts

**Info:**
 - Suite version: desktop 24.11.3 (71f9b21d21c9b033e34fc8c6d33ffd8223ffda73)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.11.3 Chrome/128.0.6613.162 Electron/32.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982

**Screenshots:**
Screenshot 2024-11-25 at 9 41 36
Screenshot 2024-11-25 at 9 41 32
Screenshot 2024-11-25 at 9 40 53

enjojoy:
## Description
Now the keys generated for non-mainnet evm accounts will be different than the ones generated for miannnet, so the files on the storage will be different and labels will not duplicate for each network.

Ethereum mainnet labels should stay the same and the others networks will be replaced by the new files.

To rewrite metadata keys it's needed to rediscover accounts 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15546

## Screenshots:

Screenshot 2025-01-02 at 13 27 53
Screenshot 2025-01-02 at 13 28 17

### DIFF
--- a/packages/suite/src/storage/CHANGELOG.md
+++ b/packages/suite/src/storage/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Storage changelog
 
+## 51
+
+-   Changed `metadata.key` on non-eth EVM networks accounts to be `descriptor-chainId`
+
 ## 49
 
 -   networks now have same order everywhere
--- a/packages/suite/src/storage/index.ts
+++ b/packages/suite/src/storage/index.ts
@@ -5,7 +5,7 @@ import { reloadApp } from 'src/utils/suite/reload';
 import { migrate } from './migrations';
 import type { SuiteDBSchema } from './definitions';
 
-const VERSION = 50; // don't forget to add migration and CHANGELOG when changing versions!
+const VERSION = 51; // don't forget to add migration and CHANGELOG when changing versions!
 
 /**
  *  If the object stores don't already exist then creates them.
--- a/packages/suite/src/storage/migrations/index.ts
+++ b/packages/suite/src/storage/migrations/index.ts
@@ -3,6 +3,7 @@ import { toWei } from 'web3-utils';
 import { BigNumber } from '@trezor/utils/src/bigNumber';
 import { isDesktop } from '@trezor/env-utils';
 import {
+    getNetwork,
     isNetworkSymbol,
     type NetworkSymbol,
     networkSymbolCollection,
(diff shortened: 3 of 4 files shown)


enjojoy:
Initially I thought about completely replacing account.metadata.key with account.key, for cleaner code and consistency, but that would overwrite all already existing labels, so I added some logic only for evm non-mainnet accounts. Maybe we should  also add some informational banner that the user now can set unique labels for all evm accounts. @tomasklim what do you think?

packages/suite/src/actions/suite/metadataLabelingActions.ts
```diff
@@ -111,7 +111,12 @@ export const setAccountMetadataKey =
             return account;
         }
         try {
-            const metaKey = metadataUtils.deriveMetadataKey(deviceMetaKey, account.metadata.key);
+            const isNonEthEvm = account.networkType === 'ethereum' && account.symbol !== 'eth';
+
+            const metaKey = metadataUtils.deriveMetadataKey(
+                deviceMetaKey,
+                isNonEthEvm ? account.key : account.metadata.key,
```
tomasklim: `key` will not work, it changes (same seed, two devices)

enjojoy: changed to metadata key + network symbol

tomasklim: can we use chain id, chain id will probably never change but network symbol can (matic -> pol)

tomasklim: can you modify metadata key at the source? you are already introducing bug as you are modifying it just on 1 place out of 2

enjojoy: Yes, now it's modifying at the source, tbh it was tricky to find where it's updated. 

But this also affects mobile sice it's in suite common, i hope it wont break anything

tomasklim (approved):

PR #16137, opened 2025-01-02
fix: eslint pre-commit hook now supports partially staged files

peter-sanderson:
This PR tweaks the pre-commit hook to enable working with both staged and unstaged files. 

In case you have staged changes on the file that you want to commit and then some other unstaged changes on the very same file, this fix will preserve the unstaged changes without committing them.

:exclamation: :point_right: The Git version at least `2.35` is required for this to work. For older Ubuntu the git can be updated from this repository: https://launchpad.net/%7Egit-core/+archive/ubuntu/ppa

-------

The unsolved issue is, that if you have staged and unstaged changes next to each other, it will produce conflict 

Staged:
image

Unstaged:

image

Conflict:

image

### DIFF
--- a/.husky/eslint.sh
+++ b/.husky/eslint.sh
@@ -3,6 +3,7 @@
 GREEN='\033[0;32m'
 NC='\033[0m' # No Color
 
+
 if [ "$TREZOR_PRE_COMMIT_ESLINT_SKIP" == "true" ]; then
   echo "Skipping eslint pre-commit hook, do: 'export TREZOR_PRE_COMMIT_ESLINT_SKIP=false' to re-enable it."
   exit 0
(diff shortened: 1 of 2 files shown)


.husky/eslint.sh
```diff
@@ -24,14 +26,49 @@ else
   echo "$STAGED_FILES"
   echo ""
 
+  # This will stashed all changes that are not added (the red files), but keeps added (the green files)
+#  echo -e "Stashing the ${GREEN}staged${NC} files."
+  git stash push --staged -q   # Stash staged changes (green files)
```
peter-sanderson, line 30: `--staged` supported by `git` version `2.35+`

Lemonexe:
FWIW I am against this. Those are potentially destructive actions that can cause our work to be lost, and the code is not trivial. IMHO we would need  "unit tests", if we were to trust it. And I don't think it's worth the engineering effort to write unit tests in bash to test git operations on file system :shrug: 

On the other hand, the current `pre-commit` script runs only eslint fix, which I believe to be trustworthy.

I am aware of the problem this PR solves. I guess, that if you need to commit partially staged files, you have to remember to do it with `-n` :disappointed:

PR #16144, opened 2025-01-02
Fix/check seed links

komret:
I noticed that in case `deviceWillBeWiped` during firmware update, user is sent to settings to create or check backup, but these settings are not available for firmwares unsupported by Suite. The links should lead right to the create/check backup modals instead. While this is only a problem with unsupported firmware (currently 1.6.1), redirecting right to the create/check backup modal is more convenient than having to go to settings first.

Plus some cleanup and fixes along the way, in separate commits. For example, Suite warned users that device will be wiped even though it was not initialized.

TODO:
- [x] update Crowdin

Related: https://github.com/trezor/trezor-suite/pull/16094 - this PR adds Check backup button to `CheckSeedStep` for some cases. I decided to add the button to create or check backup to `CheckSeedStep` for every case so that we are consistent.

**QA:**
Some testing scenarios I can think of:
- [x] Check with a fresh device
- [x] Check with an initialized device
- [x] Check required firmware update
- [x] Check with a backed-up device
- [x] Check with a device without backup
- [x] Check firmware switch on Safe series
- [x] Check firmware update from 1.6.1, this is the only fw version guaranteed to wipe and should show the warning (no need to proceed to update itself, just the screens before that)

#### Before:
image
image
image
Notice how clicking on _creating a wallet backup in Settings_ takes you to the empty page. You can click on the _Check wallet backup_ button in the red banner, but:
- The banner originates in message system and might not be there in the future. EDIT: it is not there anymore.
- Checking backup in a wallet without backup makes no sense. Creating backup is not possible at this point because there is no button to do it.
#### After:
I don't have a 1.6.1 device so testing with firmware type switch on a TS3.

Without backup:

https://github.com/user-attachments/assets/7b3c9f69-f305-4429-83b8-191a86029dad

Backed up:

https://github.com/user-attachments/assets/d884483a-57d3-4658-80be-d984f021f71c


suite-common/firmware/src/hooks/useFirmwareInstallation.ts
```diff
@@ -22,9 +22,24 @@ There are three firmware update flows, depending on current firmware version:
...
+const determineIfDeviceWillBeWiped = (
+    device: TrezorDevice | undefined,
+    shouldSwitchFirmwareType: boolean,
+) => {
+    const deviceModelInternal = device?.features?.internal_model;
+    const deviceIsInitializedOrInBootloader = device?.mode !== 'initialize';
+    const installationWillChangeFirmwareVendorHeader =
+        !!shouldSwitchFirmwareType &&
+        deviceModelInternal !== undefined &&
+        ![DeviceModelInternal.T1B1, DeviceModelInternal.T2T1].includes(deviceModelInternal);
```
peter-sanderson, line 36: Maybe we shall add come explanation. I know there is some logic that some models have this change of vendor and other don't but I don't know any details and from the code itself its not obvious to me how this is designed (or reasons why it is the way it is)

peter-sanderson (approved):
Good job, its very elegant actually. :clap:

suite-common/firmware/src/hooks/useFirmwareInstallation.ts
komret: 48e164d82f0a084467cbc5e33494708dc02e9e90

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.1.1 (4403882efb2a4a802c3d77917d6350a4fae7663d)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.1.1 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Transport: BridgeTransport 3.0.0-bundled.25.1.1

STew790:
**QA OK**
Device with 1.6.1 fw LGTM

**Videos:**
With backup
https://github.com/user-attachments/assets/e17c8291-9070-4439-b118-6c78f99e1555

Without backup
https://github.com/user-attachments/assets/3aa5a194-b310-49d5-9250-1b5332849067

**Info:**
 - Suite version: desktop 25.1.1 (4403882efb2a4a802c3d77917d6350a4fae7663d)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.1.1 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: Trezor T1B1 1.6.1 regular (revision fd0623ba2655caf63f7b0e6094a22b7e332108ef)

PR #16147, opened 2025-01-03
Migrate t3t1 create wallet test

HajekOndrej:
## Related Issue

Resolve

### DIFF
--- a/packages/suite-desktop-core/e2e/support/pageActions/onboarding/backupActions.ts
+++ b/packages/suite-desktop-core/e2e/support/pageActions/onboarding/backupActions.ts
@@ -3,7 +3,7 @@ import { Locator, Page, expect } from '@playwright/test';
 import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
 
 import { DevicePromptActions } from '../devicePromptActions';
-import { step } from '../common';
+import { step } from '../../common';
 
 export class BackupActions {
     readonly startButton: Locator;
--- a/packages/suite-desktop-core/e2e/support/pageActions/onboarding/backupActions.ts
+++ b/packages/suite-desktop-core/e2e/support/pageActions/onboarding/backupActions.ts
@@ -2,46 +2,46 @@ import { Locator, Page, expect } from '@playwright/test';
 
 import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
 
-import { DevicePromptActions } from './devicePromptActions';
+import { DevicePromptActions } from '../devicePromptActions';
 import { step } from '../common';
 
 export class BackupActions {
-    readonly backupStartButton: Locator;
+    readonly startButton: Locator;
     readonly wroteSeedProperlyCheckbox: Locator;
     readonly madeNoDigitalCopyCheckbox: Locator;
     readonly willHideSeedCheckbox: Locator;
-    readonly backupCloseButton: Locator;
+    readonly closeButton: Locator;
 
     constructor(
         private page: Page,
         private devicePrompt: DevicePromptActions,
     ) {
-        this.backupStartButton = page.getByTestId('@backup/start-button');
+        this.startButton = page.getByTestId('@backup/start-button');
         this.wroteSeedProperlyCheckbox = page.getByTestId('@backup/check-item/wrote-seed-properly');
         this.madeNoDigitalCopyCheckbox = page.getByTestId(
             '@backup/check-item/made-no-digital-copy',
         );
         this.willHideSeedCheckbox = page.getByTestId('@backup/check-item/will-hide-seed');
-        this.backupCloseButton = page.getByTestId('@backup/close-button');
+        this.closeButton = page.getByTestId('@backup/close-button');
     }
 
     @step()
     async passThroughShamirBackup(shares: number, threshold: number) {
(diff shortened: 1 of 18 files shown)


packages/suite-desktop-core/e2e/support/pageActions/onboarding/backupActions.ts
```diff
@@ -2,44 +2,44 @@ import { Locator, Page, expect } from '@playwright/test';
...
     readonly wroteSeedProperlyCheckbox: Locator;
     readonly madeNoDigitalCopyCheckbox: Locator;
     readonly willHideSeedCheckbox: Locator;
-    readonly backupCloseButton: Locator;
+    readonly closeButton: Locator;
 
     constructor(
         private page: Page,
         private devicePrompt: DevicePromptActions,
     ) {
```
Vere-Grey, line 18: Since now class `BackupActions` is initilized in onboardingActions, we have to drag devicePrompt thru fixtures.ts, then onboadingAction.ts to here.
Let's instead just create new instance here. We have the page object and that is all devicePrompt needs. 

``` suggestion
        private page: Page,
    ) {
        this.devicePrompt = new DevicePromptActions(this.page)
```

packages/suite-desktop-core/e2e/support/pageActions/onboarding/onboardingActions.ts
```diff
@@ -104,13 +125,25 @@ export class OnboardingActions {
         }, SuiteActions);
     }
 
-    async skipFirmware() {
-        await this.skipFirmwareButton.click();
-        await this.skipConfirmButton.click();
+    async passThroughAuthenticityCheck() {
```
Vere-Grey, line 130: This duplicates a lot of code from method `completeOnboarding` few lines just above. 

The page evalute should not be necessary because in the test you already call disableFirmwareHashCheck(), which does same thing.
Could we remove the evaluate, use the locators from `completeOnboarding` and then reuse this methods in `completeOnboarding` ?

packages/suite-desktop-core/e2e/tests/onboarding/t3t1/t3t1-create-wallet.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    test('Success (Shamir backup)', async ({
+        page,
+        onboardingPage,
+        devicePrompt,
+        analyticsPage,
+        trezorUserEnvLink,
+    }) => {
+        await page.pause();
```
Vere-Grey: forgotten page.pause()

packages/suite-desktop-core/e2e/tests/onboarding/t3t1/t3t1-create-wallet.test.ts
```diff
@@ -0,0 +1,61 @@
...
+        await analyticsPage.passThroughAnalytics();
+
+        // Device onboarding steps
+        await onboardingPage.firmware.skip();
+        await onboardingPage.passThroughAuthenticityCheck();
+        await page.waitForTimeout(500);
```
Vere-Grey, line 26: Why do we need to wait?
That last action in passThroughAuthenticityCheck(); is a  `await this.page.getByTestId('@authenticity-check/continue-button').click();` and the next action is another click on a skip button. Both are actions that utilize playwright timeout/retry mechanics.

Vere-Grey (changes requested):
Looking good. I suggest:
- one refactoring
- DRYing one method
- remove forgotten pause

and have question about one wait

packages/suite-desktop-core/e2e/tests/onboarding/t3t1/t3t1-create-wallet.test.ts
HajekOndrej: fixed

packages/suite-desktop-core/e2e/support/pageActions/onboarding/onboardingActions.ts
HajekOndrej: Great idea. Done.

packages/suite-desktop-core/e2e/support/pageActions/onboarding/backupActions.ts
HajekOndrej: The prompt is also used in the onboarding actions itself, so I don't know if there is much to gain here.

packages/suite-desktop-core/e2e/tests/onboarding/t3t1/t3t1-create-wallet.test.ts
HajekOndrej: Without it, it is flaky. I don't know exactly why, but it shows an error in the app `Error: Device call in progress`
It seems that there is something happening in the background on the device.

packages/suite-desktop-core/e2e/support/pageActions/onboarding/backupActions.ts
Vere-Grey: It still would feel cleaner. So you dont need to look from where and in what states it comes when there is no reason for it. But I think we can leave it for later. I will look at it later and see if the change has value or not.

Vere-Grey (approved):
LGTM!

PR #16149, opened 2025-01-03
Fix flaky connect tests

marekrjpolak:
## Description

Fix recurring failures in Connect E2E tests:

https://github.com/trezor/trezor-suite/pull/16149/commits/02d15d62dba180aa75bbb8dd4ed7b3179c334bee - Allow to turn off automatic `emulator-press-yes` on every button request, and immediately use it in override test so it works as expected (we don't want to confirm shown address because we try to override the call).
https://github.com/trezor/trezor-suite/pull/16149/commits/5a68f5c661bd880f75f5ea303f521c358ae0a21e - Disable automatic `emulator-press-yes` specifically for `ButtonRequest_PinEntry` as it's not needed (`emulator-input` is enough) and it was causing issues on 2-main (some DebugLink changes).
https://github.com/trezor/trezor-suite/pull/16149/commits/95cd713ee2fd16c3f246417bf1a07929576fa9be - `details` field is now required on Blockfrost API (https://github.com/blockfrost/blockfrost-websocket-link/commit/ff1a03a47ee8ef8904e7b7d6b517526265d24266) so we're defaulting to `basic`.
https://github.com/trezor/trezor-suite/pull/16149/commits/6cde9ec5cf368b9131cf633d4572dadebbb4779c - some `cancel` tests are now conditional for T1 (as intended)
https://github.com/trezor/trezor-suite/pull/16149/commits/d4df770257d77d43d6576720a193a7ce327c4e12 - Tests requiring setting `auto_lock_delay_ms` are now skipped for T3T1 until https://github.com/trezor/trezor-user-env/issues/280 is fixed

### DIFF
--- a/packages/blockchain-link/src/workers/blockfrost/index.ts
+++ b/packages/blockchain-link/src/workers/blockfrost/index.ts
@@ -91,7 +91,8 @@ const pushTransaction = async (request: Request<MessageTypes.PushTransaction>) =
 
 const getAccountInfo = async (request: Request<MessageTypes.GetAccountInfo>) => {
     const api = await request.connect();
-    const info = await api.getAccountInfo(request.payload);
+    const { details = 'basic', ...rest } = request.payload;
+    const info = await api.getAccountInfo({ details, ...rest });
 
     return {
         type: RESPONSES.GET_ACCOUNT_INFO,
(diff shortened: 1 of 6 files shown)


martykan (approved):
Great 👍

PR #16153, opened 2025-01-03
tiny ui fixes - unauthorized device

### LINKED ISSUE #16106: enable labeling switch - missing tooltip
there are cases when enable labeling switch is disabled (device not authorized yet). there is a tooltip explaining the reason but under some conditions the tooltip does not appear

no tooltip 
image

tooltip
image

steps to reproduce: 
1. go directly to http://localhost:8000/settings
2. try to enable labeling, the switch is disabled without any tooltip 

Image

mroz22:
resolves #16106

### DIFF
--- a/packages/suite/src/components/suite/labeling/WalletLabeling.tsx
+++ b/packages/suite/src/components/suite/labeling/WalletLabeling.tsx
@@ -21,6 +21,7 @@ export const useWalletLabeling = () => {
 
     const defaultAccountLabelString = useCallback(
         ({ device }: { device: TrezorDevice }) => {
+            if (!device.state?.staticSessionId) return undefined; // wallet not authorized yet
             if (device.useEmptyPassphrase) return translationString('TR_NO_PASSPHRASE_WALLET');
             if (!device.walletNumber) return undefined;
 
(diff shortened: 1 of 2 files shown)


packages/suite/src/views/settings/SettingsGeneral/Labeling.tsx
```diff
@@ -40,7 +40,7 @@ export const Labeling = () => {
     // - Labeling enabled without any device connected
     // - Labeling enabled with the device connected inside Settings
     // The initialization of Labeling then start when user select a Wallet.
-    const isDisabled = isLocked() || device?.mode !== 'normal';
+    const isDisabled = isLocked() || device?.mode !== 'normal' || !device?.state?.sessionId;
```
mroz22: hmm do I need to ask specifically fro sessionId or is it enough to check for state only?

komret: Looks like `state` is enough.

packages/suite/src/components/suite/labeling/WalletLabeling.tsx
```diff
@@ -21,6 +21,7 @@ export const useWalletLabeling = () => {
 
     const defaultAccountLabelString = useCallback(
         ({ device }: { device: TrezorDevice }) => {
+            if (!device.state) return undefined;
```
komret:
```suggestion
            if (!device.state) return undefined; // wallet not loaded yet
```

komret (changes requested):
nits

komret (approved):

PR #16154, opened 2025-01-03
fix: remove pin-followup on backup device (doesn make sense), also prevent modal closure without confirming seed action

### LINKED ISSUE #16156: Remove Pin next-steps in Backup action as it doesn't make much sense + prevent closure of the modal
When you create backup in settings (because you skipped it during onboarding), this is the final screen. Is it intentional that we show the Create/Skip PIN buttons? Looks like a bug.

image

Also, the modal should not be cancellable unless all checkboxes are checked.

peter-sanderson:
Resolve https://github.com/trezor/trezor-suite/issues/16156

- Also refactoring and code-splitting was done (see first commit)

-------------------

Before: 

image

After: 

image
image


packages/suite/src/utils/backup/index.ts
```diff
@@ -1,13 +1,5 @@
 import { ConfirmKey } from 'src/actions/backup/backupActions';
 
-/**
- * Utility function used to disable backup start button
- */
-export const canStart = (userConfirmed: ConfirmKey[], isDeviceLocked: boolean) =>
```
peter-sanderson, line 6: this has usage only on one place so I moved it there

packages/suite/src/views/backup/BackupStep1Initial.tsx
```diff
@@ -0,0 +1,77 @@
...
+        userConfirmed.includes(e),
+    ) && !isDeviceLocked;
+
+export const BackupStep1Initial = ({
+    onCancel,
+    backup,
+}: {
+    onCancel: () => void;
+    backup: BackupState;
+}) => {
```
komret, line 25: Nit: I'd rather see these abstracted into a type for consistency.

komret (approved):
I am still not sure if the behaviour was intentional or not... but it was confusing.

PR #16171, opened 2025-01-05
Feat/suite core/e2e/convert unsupported browsers

Vere-Grey:
## Description

Converts browser test to PW
Unfortunately PW does not support firefox - mobile combination and it is not recommended to run it with outdated browsers -> I had to remove 3 tests completely.
Enables again visual regression checks for web. The problematic market tests are disabled anyway and I added instruction how fix their visual comparison once we want them to run again.

## Related Issue

Resolve 15606


Vere-Grey:
I will rebase this on the PW upgrade once it is merge and change the failing visual regression check to Aria snapshot

HajekOndrej (approved):

PR #16184, opened 2025-01-06
fix(suite-native): monthly app version bump

PeKne:
- runs only for `trezor/trezor-suite` repo
 - permission for creating a PR added (should fix this error: https://github.com/trezor/trezor-suite/actions/runs/12565247095)


matejkriz (approved):
LGTM

PR #16188, opened 2025-01-06
Analytics for silent update

### LINKED ISSUE #15895: Data Analytics: Silent update
We would like to track the silent update usage. We propose adding new parameters to two existing events:
- [Suite Ready](https://www.notion.so/satoshilabs/suite-ready-1e5cbba0a1a545edb437185c37ddafdc?pvs=4): `silentUpdate` (bool) tracking if the feature is enabled or disabled
- [App update](https://www.notion.so/satoshilabs/app-update-671449db50c843bb9d64cef41246ece3?pvs=4): `silentUpdate` (bool) tracking if the update was done using silent update

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite/issues/15895

### DIFF
--- a/packages/suite/src/utils/suite/analytics.ts
+++ b/packages/suite/src/utils/suite/analytics.ts
@@ -92,17 +92,26 @@ export const getSuiteReadyPayload = async (
         windowHeight: getWindowHeight(),
         autodetectLanguage: state.suite.settings.autodetect.language,
         autodetectTheme: state.suite.settings.autodetect.theme,
+
+        isAutomaticUpdateEnabled: state.desktopUpdate.isAutomaticUpdateEnabled,
     };
 };
 
-export const getAppUpdatePayload = (
-    status: AppUpdateEvent['status'],
-    earlyAccessProgram: boolean,
-    updateInfo?: UpdateInfo,
-): AppUpdateEvent => ({
+export const getAppUpdatePayload = ({
+    status,
+    earlyAccessProgram,
+    updateInfo,
+    isAutoUpdated,
+}: {
+    status: AppUpdateEvent['status'];
+    earlyAccessProgram: boolean;
+    updateInfo?: UpdateInfo;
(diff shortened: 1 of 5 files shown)


packages/suite/src/utils/suite/analytics.ts
```diff
@@ -48,7 +48,9 @@ export const redactTransactionIdFromAnchor = (anchor?: string) => {
 // 1. replace coinjoin by taproot
 export const redactRouterUrl = (url: string) => url.replace(/coinjoin/g, 'taproot');
 
-export const getSuiteReadyPayload = async (state: AppState) => {
+export const getSuiteReadyPayload = async (
+    state: AppState,
+): Promise<SuiteAnalyticsEventSuiteReady['payload']> => {
```
peter-sanderson, line 53: This missing type revealed error in typing. 

```ts
desktopOsVersion: string | undefined;
desktopOsName: string | undefined;
desktopOsArchitecture: string | undefined;
```

:point_up: this was missing

Lemonexe: :bow:

packages/suite/src/utils/suite/analytics.ts
```diff
@@ -90,17 +92,26 @@ export const getSuiteReadyPayload = async (state: AppState) => {
...
-    status: AppUpdateEvent['status'],
-    earlyAccessProgram: boolean,
-    updateInfo?: UpdateInfo,
-): AppUpdateEvent => ({
+export const getAppUpdatePayload = ({
+    status,
+    earlyAccessProgram,
+    updateInfo,
+    isAutoUpdated,
+}: {
```
Lemonexe, line 105: good idea to change positional args to an options object :+1: 
Tiny nit: could you extract the type? It'd be better readable.

Lemonexe (approved):
LGTM codewise, but have not tested due to the complexity of setting it up

PR #16199, opened 2025-01-06
feat(suite): add new color variant

MiroslavProchazka:
Adding new color variant of t3t1


MiroslavProchazka:
Tested on my local build and it works fine

komret (approved):

Issue #15584, opened 2024-11-26
mobile: firmware upgrade follow-up


Nodonisko:
Followup of #14614 

- [x] Do not allow FW update if discovery is running
- [x] `What if it gets stuck?` button BottomSheet should open after button is clicked
- [ ] `Mandatory update` - [we should be able to enforce update for/from some versions](https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Manage-device?node-id=367-4465&t=truBTMxkQAuy1KDd-1	)
- [x] Limit FW version to only recent ones in mobile app - we can probably support anything above `2.6.0` so maybe we don't need do anything because app already supports only above `2.6.3`? - its enough
- [x] [Update alert box on HomeScreen](https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Manage-device?node-id=367-4416&t=GPYhL1WCoOVNbQOM-1)
- [x] [Make sure you have seeds warning before upgrading](https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Manage-device?node-id=367-3802&t=GPYhL1WCoOVNbQOM-1)
- [x] [Changelog link (or BottomSheet?)](https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Manage-device?node-id=367-3737&t=GPYhL1WCoOVNbQOM-1)
- [ ] It's offering firmware update if user have newer (unreleased FW) (is this even problem?)
- [x] It's possible to trigger update flow for watch-only unconnected devices
- [x] Remote feature flag for turning FW update off (local too)

fang314:
- @matejkriz  to check with @Nodonisko about [this security check before FW update](https://satoshilabs.slack.com/archives/C04JZAYEV41/p1734528380081549) before releasing this feature

STew790:
**QA OK**
Some points already tested here https://github.com/trezor/trezor-suite/issues/14614
- update alert is modal instead
- changelog is OK
- update flow possible to trigger with view only devices
- remote ff tested as well
Also, this is already released.

PR #16205, opened 2025-01-06
feat(mobile): FW update changelog modal

### LINKED ISSUE #16572: Firmware update changelog is not showing
**Describe the bug**
Firmware changelog is not shown in the fw update screen.

**Info:**

-   Mobile Suite Version: 25.1.3 33aa804
-   OS: Android
- 
**How to reproduce**

1. Connect device
2. Go to device settings
3. Click on "Update" button
4. There is no fw changelog

**Expected behavior**
A fw changelog should appear under the "Update firmware" part

**Screenshots**

Image

Nodonisko:
## Description

Implement changelog bottom sheet + fix centering of Confirm on Trezor image on bottom of screen.

## Related Issue

Resolve part of #15584
Resolve #16572

## Screenshots:

Screenshot_20250106-150905

### DIFF
--- a/suite-common/wallet-core/src/device/deviceReducer.ts
+++ b/suite-common/wallet-core/src/device/deviceReducer.ts
@@ -1036,3 +1036,14 @@ export const selectDeviceUpdateFirmwareVersion = (state: DeviceRootState) => {
 
     return device ? getFwUpdateVersion(device) : null;
 };
+
+export const selectFirmwareChangelog = (state: DeviceRootState) => {
+    const device = selectSelectedDevice(state);
+    const isBitcoinOnlyFirmware = selectHasBitcoinOnlyFirmware(state);
+
+    if (isBitcoinOnlyFirmware) {
+        return device?.firmwareRelease?.changelog?.[0]?.changelog_bitcoinonly;
+    }
+
+    return device?.firmwareRelease?.changelog?.[0]?.changelog;
+};
--- a/suite-native/intl/src/en.ts
+++ b/suite-native/intl/src/en.ts
@@ -402,6 +402,12 @@ export const en = {
                 updateButton: 'Update firmware',
                 title: 'Firmware update',
                 subtitle: 'New firmware is now available. Update your device now.',
+                changelog: {
+                    button: 'What’s new?',
+                    title: 'What’s new?',
+                    closeButton: 'Close',
+                    changelogUnavailable: 'No changelog available',
+                },
             },
             firmwareUpdateProgress: {
                 initializing: { title: 'Preparing your Trezor' },
--- a/suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
+++ b/suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
@@ -1,0 +1,86 @@
+import { useMemo } from 'react';
+import { useSelector } from 'react-redux';
+
+import { BottomSheet, Button, Text } from '@suite-native/atoms';
+import { selectFirmwareChangelog } from '@suite-common/wallet-core';
+import { Translation } from '@suite-native/intl';
+import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
+
+type FirmwareChangelogProps = {
+    isVisible: boolean;
+    onClose: () => void;
+};
+
+const changelogSectionTitleTextStyle = prepareNativeStyle(utils => ({
+    ...utils.typography.highlight,
+    paddingTop: utils.spacings.sp24,
+}));
+
+const buttonContainerStyle = prepareNativeStyle(utils => ({
+    marginTop: utils.spacings.sp32,
+}));
+
+const ChangelogSectionTitle = ({ children }: { children: React.ReactNode }) => {
+    const { applyStyle } = useNativeStyles();
+
+    return <Text style={applyStyle(changelogSectionTitleTextStyle)}>{children}</Text>;
+};
+
+export const FirmwareChangelog = ({ isVisible, onClose }: FirmwareChangelogProps) => {
+    const firmwareChangelog = useSelector(selectFirmwareChangelog);
+    const { applyStyle } = useNativeStyles();
+
+    const formattedChangelog = useMemo(() => {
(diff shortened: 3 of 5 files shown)


suite-native/intl/src/en.ts
```diff
@@ -399,6 +399,12 @@ export const en = {
                 updateButton: 'Update firmware',
                 title: 'Firmware update',
                 subtitle: 'New firmware is now available. Update your device now.',
+                changelog: {
+                    button: 'What’s new?',
+                    title: 'What’s new?',
+                    closeButton: 'Close',
```
yanascz, line 408: We've got a generic label for this: `generic.buttons.close`

suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            let formattedText = text;
+
+            if (text.startsWith(' - ')) {
+                formattedText = formattedText.replace(' - ', ' • ');
+            }
+
+            return <Text key={key}>{formattedText}</Text>;
+        });
```
yanascz: This seems to be a great candidate for a utility function.

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -1031,3 +1031,9 @@ export const selectDeviceUpdateFirmwareVersion = (state: DeviceRootState) => {
 
     return device ? getFwUpdateVersion(device) : null;
 };
+
+export const selectFirmwareChangelog = (state: DeviceRootState) => {
```
vytick, line 1040: nit, feel free to ignore
```suggestion
export const selectFirmwareChangelog = (state: DeviceRootState) => selectSelectedDevice(state)?.firmwareRelease?.changelog?.[0]?.changelog
```

suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
Nodonisko: I don't think we will reuse this code anywhere else and component itself is basically only this function, so I think there is no reason create extra function for it.

suite-native/intl/src/en.ts
Nodonisko: fixed

suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+            if (text.startsWith('#')) {
+                const strippedText = text.replaceAll('#', '').trim();
+
+                return <ChangelogSectionTitle key={key}>{strippedText}</ChangelogSectionTitle>;
+            }
+
+            let formattedText = text;
+
+            if (text.startsWith(' - ')) {
```
yanascz: I guess you might want to do the same for '*' since it looks like this otherwise:
image

suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
```diff
@@ -0,0 +1,80 @@
...
+        if (!firmwareChangelog) {
+            return (
+                <Text>
+                    <Translation id="moduleDeviceSettings.firmware.firmwareUpdateScreen.changelog.changelogUnavailable" />
+                </Text>
+            );
+        }
+
+        if (typeof firmwareChangelog === 'string') {
+            return <Text>{firmwareChangelog}</Text>;
```
yanascz: It seems that formatting may also be used in this case.

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -1036,3 +1036,9 @@ export const selectDeviceUpdateFirmwareVersion = (state: DeviceRootState) => {
...
+    const device = selectSelectedDevice(state);
+
+    return device?.firmwareRelease?.changelog?.[0]?.changelog;
```
yanascz, line 1048: We should definitely take `changelog_bitcoinonly` into account for devices with BTC-only firmware.

suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        return firmwareChangelog.map((text, index) => {
+            const key = text + index;
+
+            if (text.startsWith('#')) {
+                const strippedText = text.replaceAll('#', '').trim();
```
yanascz: In case the `#` symbol is somewhere in the text, it'll be remove as well.

suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
Nodonisko: I went though changelogs in trezor/data repo and I haven't seen any single occurence of formatting so I won't be doing now, in future we may need to use full fledged markdown library, but for now it looks like overkill to me.

suite-common/wallet-core/src/device/deviceReducer.ts
Nodonisko: Good point, done.

suite-native/module-device-settings/src/screens/FirmwareUpdateScreen/FirmwareChangelog.tsx
Nodonisko: I made whole parser more robust using regexp so it should work correctly.

matejkriz (approved):
The text alignment on the line break is not perfect, but I would consider it a nitpick and I don't believe it's worth fixing now. I think we might consider the full Markdown support later together with the decoupling of the FW release from the Suite release (fetching releases.json remotely).

image

STew790:
**QA OK**
LGTM 
image

**Info**
25.1.3 dde03ca

PR #16208, opened 2025-01-06
feat(connnect): solanaComposeTransaction method

### LINKED ISSUE #10907: Move Solana deps from Suite to Connect
As stated here https://github.com/trezor/trezor-suite/pull/9958#pullrequestreview-1741504599 and [there](https://satoshilabs.slack.com/archives/C059ZS5T24C/p1700571220416869), for security and optimization reasons we want to move all Solana 3rd party dependencies from Suite (probably mainly in `solanaUtils`) to Connect's `api/solana/api` subfolder.

It could mean to edit existing Solana-related methods and/or to create new `solanaComposeTransaction` method (similarly to [cardanoComposeTransaction](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/api/cardano/api/cardanoComposeTransaction.ts)).

martykan:
## Description

Moved `solanaUtils` from `wallet-utils` to `connect` package. 
Wrapped the API into a new method `solanaComposeTransaction`. 
Improved `solanaSignTransaction` for better interoperability with this method. 
Applied these changes to Suite send form Solana thunk. 

## Related Issue

Resolve #10907

## QA

Please closely test Solana send flows


packages/connect/package.json
```diff
@@ -72,10 +72,17 @@
...
         "@ethereumjs/tx": "^5.4.0",
         "@fivebinaries/coin-selection": "3.0.0",
+        "@mobily/ts-belt": "^3.13.1",
         "@noble/hashes": "^1.6.1",
         "@scure/bip39": "^1.5.1",
+        "@solana-program/compute-budget": "^0.6.1",
+        "@solana-program/system": "^0.6.2",
+        "@solana-program/token": "^0.4.1",
+        "@solana-program/token-2022": "^0.3.1",
+        "@solana/web3.js": "^2.0.0",
```
mroz22, line 82: if everything is set up right we shouldn't see any size increase in per-coin bundles. we should make sure that solana specific libs don't leak where they shoudln't

packages/connect/e2e/__fixtures__/solanaComposeTransaction.ts
```diff
@@ -0,0 +1,97 @@
...
+                fromAddress: 'ANctUhC7YZPueiv4T8bkDcHYEAJ7Hwoxhvgnr2QkF8uR',
+                toAddress: '5Q9c3XoBef8BYA5RzSmogWnRrQas6HPwYuo4AYPafpom',
+                amount: '0.01',
+                blockHash: 'BXim2ZLR2UZ4JQxhNGaGngbRaySWej4x8zqaw7TJ4GLo',
+                lastValidBlockHeight: 290999279,
+                coin: 'sol',
+            },
+            result: {
+                serializedTx:
+                    '0100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010002048b42f51ed008e27e643818f0503bdfeb6535dc815e3a9fd41a7ce42344f888dc415cd5ca15bbbbcf07713fe3f690ebaab81c9cfa8f8a95f91b17deabb953b83400000000000000000000000000000000000000000000000000000000000000000306466fe5211732ffecadba72c39be7bc8ce5bbc5f7126b2c439b3a400000009c7382bc11ef07feb1a080bc78d2f7f6906ae7cb7c389057fd99ae3e97c811e00303000502400d030003000903a086010000000000020200010c020000008096980000000000',
```
mroz22, line 19: just out of curiousity - where is this data taken from? if there is some "source of truth" it is nice to put it here as a comment

packages/suite-desktop-core/webpack/core.webpack.config.ts
```diff
@@ -123,7 +123,7 @@ const config: webpack.Configuration = {
             name(_: any, chunks: any) {
                 return chunks.length === 1
                     ? chunks[0].name
-                    : `shared/${chunks.map((item: any) => item.name.split('/').pop()).join('~')}`;
+                    : `shared/${chunks.map((item: any) => item.name?.split('/').pop()).join('~')}`;
```
mroz22: why was this change needed? 🤔

mroz22:
i can sense conflicts with #16286

martykan:
Yeah, this definitely seems redundant to #16286. I can drop 0fcc3480cd960d41e8fad0b63fb963706046e788 from here.

Edit: dropped now

packages/suite-desktop-core/webpack/core.webpack.config.ts
martykan: `item.name` would be undefined in some case, but not sure about the root cause

packages/connect/package.json
martykan: Ok, I will compare bundle sizes to check

packages/connect/e2e/__fixtures__/solanaComposeTransaction.ts
martykan: This is from our code's output, but cross checked that it matches. 
This is mainly to test for regressions in the future.

packages/connect/package.json
martykan: Can confirm other bundles are unaffected

packages/suite-desktop-core/webpack/core.webpack.config.ts
mroz22: maybe @marekrjpolak could consult a bit this PR from the point of view of code bundling

mroz22 (approved):

packages/suite-desktop-core/webpack/core.webpack.config.ts
martykan: Fixed by adding missing `/* webpackChunkName: ...`

PR #16224, opened 2025-01-07
fix: copy fix for wipe-device vs. factory-reset in settings

### LINKED ISSUE #9626: Separate copy for wipe device in bootloader mode from normal mode
If you click Factory reset in device settings with device in normal mode, it wipes only the seed and PIN. If you click the same button in bootloader mode, it wipes also firmware. Would be nice to have separate copy for those two variants.

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite/issues/9626

image
image

### DIFF
--- a/packages/suite/src/views/settings/SettingsDevice/WipeDevice.tsx
+++ b/packages/suite/src/views/settings/SettingsDevice/WipeDevice.tsx
@@ -1,6 +1,8 @@
+import { isDeviceInBootloaderMode } from '@trezor/device-utils';
+
 import { SettingsSectionItem } from 'src/components/settings';
 import { ActionButton, ActionColumn, TextColumn, Translation } from 'src/components/suite';
-import { useDispatch } from 'src/hooks/suite';
+import { useDevice, useDispatch } from 'src/hooks/suite';
 import { openModal } from 'src/actions/suite/modalActions';
 import { SettingsAnchor } from 'src/constants/suite/anchors';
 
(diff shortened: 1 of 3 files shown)


packages/suite/src/views/settings/SettingsDevice/WipeDevice.tsx
```diff
@@ -9,15 +11,30 @@ interface WipeDeviceProps {
...
 
 export const WipeDevice = ({ isDeviceLocked }: WipeDeviceProps) => {
+    const { device } = useDevice();
     const dispatch = useDispatch();
 
+    const isBootloaderMode = isDeviceInBootloaderMode(device);
+
     const handleClick = () => dispatch(openModal({ type: 'wipe-device' }));
 
+    const headingTranslation = isBootloaderMode
```
komret, line 21: It should be the other way round: Factory reset in bootloader, wipe in normal. Otherwise OK.

komret (changes requested):

packages/suite/src/views/settings/SettingsDevice/WipeDevice.tsx
peter-sanderson: :man_facepalming: Fixed. Thanks for catching it.

komret (approved):

peter-sanderson:
We also have a modal copy
image

STew790:
Questions here https://github.com/trezor/trezor-suite/issues/9626#issuecomment-2575706799

PR #16226, opened 2025-01-07
Fix 1 min timeout for fetching new token info 

### LINKED ISSUE #15493: It takes up to 1 minute to fetch rates for newly received token
**Describe the bug**

When you receive a token which was not have on an account before, it takes up to 1 minute to have fiat rates and 7d change ready

Related to #16127

**Info:**

-   Suite Version: 24.11.2

**Screenshots**
Screenshot 2024-11-21 at 8 35 45

enjojoy:
## Description

- Now the newly received token data is being fetched instantly, instead of waiting for the 1 min 
timeout

- Token added through "Add token" also fetches last week rates to display the change 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15493

## Screenshots:
Receive new token Before:

https://github.com/user-attachments/assets/46455a99-ef29-43aa-a6b3-139c9d1eca4c

Receive new token After:

https://github.com/user-attachments/assets/49b8d9cd-6ff5-47d3-83a4-4563de40bfaf

Add token Before:

https://github.com/user-attachments/assets/e793194b-9177-4862-82ee-1c6bd6b582d8

Add token After:

https://github.com/user-attachments/assets/58857c4e-5c8e-45be-a8cc-22c293431f10

### DIFF
--- a/packages/suite/src/actions/wallet/tokenActions.ts
+++ b/packages/suite/src/actions/wallet/tokenActions.ts
@@ -1,41 +1,22 @@
 import { TokenInfo } from '@trezor/connect';
 import * as accountUtils from '@suite-common/wallet-utils';
 import { notificationsActions } from '@suite-common/toast-notifications';
-import { accountsActions, updateFiatRatesThunk } from '@suite-common/wallet-core';
-import { Timestamp, TokenAddress } from '@suite-common/wallet-types';
-import { FiatCurrencyCode } from '@suite-common/suite-config';
+import { accountsActions } from '@suite-common/wallet-core';
 
 import { Dispatch } from 'src/types/suite';
 import { Account } from 'src/types/wallet';
 
-export const addToken =
-    (account: Account, tokenInfo: TokenInfo[], localCurrency: FiatCurrencyCode) =>
-    (dispatch: Dispatch) => {
-        dispatch(
-            accountsActions.updateAccount({
-                ...account,
-                tokens: (account.tokens || []).concat(accountUtils.enhanceTokens(tokenInfo)),
-            }),
-        );
+export const addToken = (account: Account, tokenInfo: TokenInfo[]) => (dispatch: Dispatch) => {
+    dispatch(
+        accountsActions.updateAccount({
+            ...account,
+            tokens: (account.tokens || []).concat(accountUtils.enhanceTokens(tokenInfo)),
+        }),
+    );
 
-        dispatch(
-            updateFiatRatesThunk({
-                tickers: [
-                    {
-                        symbol: account.symbol,
-                        tokenAddress: tokenInfo[0].contract as TokenAddress,
-                    },
-                ],
-                localCurrency,
-                rateType: 'current',
-                fetchAttemptTimestamp: Date.now() as Timestamp,
-                forceFetchToken: true,
-            }),
-        );
-
-        dispatch(
-            notificationsActions.addToast({
-                type: 'add-token-success',
-            }),
-        );
-    };
+    dispatch(
+        notificationsActions.addToast({
+            type: 'add-token-success',
+        }),
+    );
+};
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddTokenModal.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/AddTokenModal.tsx
@@ -11,7 +11,6 @@ import { Translation } from 'src/components/suite/Translation';
 import { useDispatch, useSelector, useTranslation } from 'src/hooks/suite';
 import { Account } from 'src/types/wallet';
 import { selectSelectedAccount } from 'src/reducers/wallet/selectedAccountReducer';
-import { selectLocalCurrency } from 'src/reducers/wallet/settingsReducer';
 
 interface AddTokenModalProps {
     onCancel: () => void;
@@ -23,7 +22,6 @@ export const AddTokenModal = ({ onCancel }: AddTokenModalProps) => {
     const [isFetching, setIsFetching] = useState(false);
     const [error, setError] = useState<string | null>(null);
     const account = useSelector(selectSelectedAccount);
-    const localCurrency = useSelector(selectLocalCurrency);
     const dispatch = useDispatch();
     const { translationString } = useTranslation();
 
@@ -99,7 +97,7 @@ export const AddTokenModal = ({ onCancel }: AddTokenModalProps) => {
 
     const handleAddTokenButtonClick = () => {
         if (tokenInfo) {
-            dispatch(addToken(account, tokenInfo, localCurrency));
+            dispatch(addToken(account, tokenInfo));
             onCancel();
 
             analytics.report({
--- a/suite-common/wallet-core/src/fiat-rates/fiatRatesMiddleware.ts
+++ b/suite-common/wallet-core/src/fiat-rates/fiatRatesMiddleware.ts
@@ -17,6 +17,7 @@ import { fetchAllTransactionsForAccountThunk } from '../transactions/transaction
 
 export const prepareFiatRatesMiddleware = createMiddlewareWithExtraDeps(
     (action, { dispatch, extra, next, getState }) => {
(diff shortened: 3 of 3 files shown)


izmy (approved):

packages/suite/src/actions/wallet/tokenActions.ts
```diff
@@ -33,6 +33,21 @@ export const addToken =
             }),
         );
 
+        dispatch(
```
tomasklim: does this need to be here? You update account on line 14, so it should be caught by your modification in `accountUtils`

tomasklim (changes requested):
Tested again and it works, weird.

However, you do not need to fix add token, it will start working automatically by your fix

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -734,6 +734,7 @@ export const isAccountOutdated = (account: Account, freshInfo: AccountInfo) => {
             return (
                 freshInfo.misc!.nonce !== account.misc.nonce ||
                 freshInfo.balance !== account.balance || // balance can change because of beacon chain txs (staking) |
+                freshInfo.tokens !== account.tokens ||
                 JSON.stringify(freshInfo?.misc?.stakingPools) !==
                     JSON.stringify(account?.misc?.stakingPools)
```
tomasklim: probably it would be better to compare string instead of object, maybe that's why it sometimes work and sometimes not.

Please get inspired by lines for staking

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -734,6 +734,7 @@ export const isAccountOutdated = (account: Account, freshInfo: AccountInfo) => {
             return (
                 freshInfo.misc!.nonce !== account.misc.nonce ||
                 freshInfo.balance !== account.balance || // balance can change because of beacon chain txs (staking) |
+                freshInfo.tokens !== account.tokens ||
```
tomasklim: Please fix it for all token networks. This fix is only for EVM

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -744,11 +753,14 @@ export const isAccountOutdated = (account: Account, freshInfo: AccountInfo) => {
...
                 // changed stake pool
-                freshInfo.misc!.staking?.poolId !== account.misc.staking.poolId
+                freshInfo.misc!.staking?.poolId !== account.misc.staking.poolId ||
+                areDifferentTokens()
             );
         case 'solana':
             // compare last transaction signature since the total number of txs may not be fetched fully
-            return freshInfo.history.txids?.[0] !== account.history.txids?.[0];
+            return (
+                freshInfo.history.txids?.[0] !== account.history.txids?.[0] || areDifferentTokens()
```
tomasklim: I am thinking now, maybe it is not required for Solana and Cardano :( can you try to remove it and test it?

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -708,6 +708,14 @@ export const getTotalFiatBalance = ({
 export const isTestnet = (symbol: NetworkSymbol) => networks[symbol].testnet;
 
 export const isAccountOutdated = (account: Account, freshInfo: AccountInfo) => {
+    const isDifferentBalance = () => {
+        return account.balance !== freshInfo.balance;
+    };
+
+    const areDifferentTokens = () => {
+        return JSON.stringify(account.tokens) !== JSON.stringify(freshInfo.tokens);
```
Nodonisko: JSON operation are very expensive and this function could run very often and I think `tokens` could be really huge if account has lot of tokens. I am pretty sure it will have negative perf impact on mobile for bigger accounts.

What about using just `account.tokens.length !== freshInfo.tokens.length` ?

tomasklim: it's possible for networks using blockbook, not for non-blockbook

suite-common/wallet-utils/src/accountUtils.ts
tomasklim: Screenshot 2025-01-09 at 16 27 53
ok I am not sure, please test it before you decide if it should or should not be there

suite-common/wallet-utils/src/accountUtils.ts
Nodonisko: How could token disappear from account? Anyway as I said I would prefer something more efficient like this (credit to chatGPT):

```
function areTheyDifferent(
  oldArr: TokenInfo[],
  freshArr: TokenInfo[]
): boolean {
  // Quick check: if lengths differ, they definitely have different items
  if (oldArr.length !== freshArr.length) {
    return true;
  }

  const oldIds = new Set(oldArr.map(item => item.contract));
  const freshIds = new Set(freshArr.map(item => item.contract));

  // If the sizes differ, there's a new/missing contractId
  if (oldIds.size !== freshIds.size) {
    return true;
  }

  // Check if every ID in old exists in fresh (and vice versa, since sizes are the same)
  for (const id of oldIds) {
    if (!freshIds.has(id)) {
      return true;
    }
  }

  // If we get here, the sets of contractIds are the same
  return false;
}
```

tomasklim: You swap one token for another.

1 token dissapear and the other appear. Length is same

packages/suite/src/actions/wallet/tokenActions.ts
enjojoy: This was not working indeed and it was needed, now I figured out why  (big thanks to @Nodonisko), the problem was in middleware and now it's fixed and fiat rate updates are instant

Fixed here https://github.com/trezor/trezor-suite/pull/16226/commits/6d3a6d0cc0e1faed6a6548a292fc0edfe90cd4f9

suite-common/wallet-utils/src/accountUtils.ts
Nodonisko: Unresolving this as my comment about performance of this solution still stands.

enjojoy: yes, I remember about it, will get to it soon

tomasklim: I am not sure if we need to do any changes in this file at all

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -726,16 +734,17 @@ export const isAccountOutdated = (account: Account, freshInfo: AccountInfo) => {
...
             );
         case 'ethereum':
             return (
+                isDifferentBalance() ||
                 freshInfo.misc!.nonce !== account.misc.nonce ||
-                freshInfo.balance !== account.balance || // balance can change because of beacon chain txs (staking) |
                 JSON.stringify(freshInfo?.misc?.stakingPools) !==
-                    JSON.stringify(account?.misc?.stakingPools)
+                    JSON.stringify(account?.misc?.stakingPools) ||
+                areDifferentTokens()
```
tomasklim: total amount of txs changes when you receive tokens so no changes here are needed. The tokens appeared before, the only problem was update of fiat rates

suite-common/wallet-utils/src/accountUtils.ts
```diff
@@ -744,11 +753,14 @@ export const isAccountOutdated = (account: Account, freshInfo: AccountInfo) => {
                 // changed rewards amount (rewards are distributed every epoch (5 days))
                 freshInfo.misc!.staking?.rewards !== account.misc.staking.rewards ||
                 // changed stake pool
-                freshInfo.misc!.staking?.poolId !== account.misc.staking.poolId
+                freshInfo.misc!.staking?.poolId !== account.misc.staking.poolId ||
+                areDifferentTokens()
```
tomasklim: Cardano works same as blockbook so probable same as for Ethereum -> total amount of txs changes when you receive tokens so no changes here are needed

suite-common/wallet-utils/src/accountUtils.ts
tomasklim: Solana works different but when something in the account changes, there has to be new tx (incoming or outcoming) so no changes here are needed

tomasklim (approved):

PR #16235, opened 2025-01-07
feat(suite-native): show screen header divider when content is scrolled

yanascz:
Based on this [Figma](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Small-Things-%5BMOB%5D?node-id=1904-20164&t=n34FKleEGo3lnULg-4).

## Screenshots:
image
image
image


suite-native/navigation/src/components/ScreenContentWrapper.tsx
```diff
@@ -19,31 +21,38 @@ const screenContentWrapperStyle = prepareNativeStyle(() => ({ flexGrow: 1 }));
...
     isScrollable,
+    hasScreenHeader,
     extraKeyboardAvoidingViewHeight,
     refreshControl,
     keyboardDismissMode,
 }: ScreenContentProps) => {
     const scrollViewRef = useRef<ScrollView | null>(null);
     const { applyStyle } = useNativeStyles();
 
+    const { scrollDivider, handleScroll } = useScrollDivider();
```
PeKne, line 32: what about passing `hasScreenHeader` here? In that case, the `useScrollDivider` would return always `scrollDivider` as `undefined`.

Then you could replace  line 
```
onScroll={hasScreenHeader ? handleScroll : undefined}
```

with 

```
onScroll={handleScroll}
```

But this is nit, feel free to resolve if you do not agree.

PeKne, line 32: Also I would prefer to pass a `ScrollView` ref as a prop here. Then you could hide the scrolling logic inside of the useScrollDivider hook and would not have to assing onScroll prop  in this component here.

Inside it could be something like:
```
useEffect(() => {
    scrollViewRef.addListener('scroll', () => {
      handleScroll()
    });
```

But this is only idea. I did not tested that.

What do you think?

PeKne (commented):
I have some proposals, let's discuss.

suite-native/navigation/src/components/ScreenContentWrapper.tsx
yanascz: Since `hasScreenHeader` only makes sense in this context, I don't want to drag it even deeper.

yanascz: I couldn't achieve what you are suggesting since `ScrollView` doesn't provide `addEventListener` / `removeEventListener` methods. Let's discuss in person if we can figure it out.

Nodonisko:
I kind of dislike this solution with plain onScroll handler, I know that function is very simple so it will be very fast but I think we should use Reanimated https://docs.swmansion.com/react-native-reanimated/docs/scroll/useScrollViewOffset/ and `runOnJS` to change that state to avoid running onScroll handler on JS thread because it will run with every scroll event and we already have significant frame drop on JS thread when scrolling through transactions.

Another solution which I prefer less is to use `scrollEventThrottle` to some higher number to minimalize numbers of call of that function.

PeKne (approved):

STew790:
**QA OK**
Reviewed with yanas and LGTM. Dark mode is a bit harder to see but it is there.

**Info**
25.1.2 https://github.com/trezor/trezor-suite/commit/995410aa85fd67087647a56ac29d5f0d34b6097e

PR #16243, opened 2025-01-07
AB tests followup

matejkriz:
## Description

- refactor ab testing to follow code style and make it easier to reuse on mobile
-  load valid experiments to store also for mobile app

## Related Issue

Followup https://github.com/trezor/trezor-suite/issues/15067

## Screenshots

Redux state:
image

___ 
Example use of useExperiment on mobile:
image

Result:
image

### DIFF
--- a/suite-common/message-system/src/messageSystemUtils.ts
+++ b/suite-common/message-system/src/messageSystemUtils.ts
@@ -288,7 +288,7 @@ export const getValidMessages = (config: MessageSystem | null, options: Options)
         .map(action => action.message);
 };
 
-export const getValidExperiments = (config: MessageSystem | null, options: Options): string[] => {
+export const getValidExperimentIds = (config: MessageSystem | null, options: Options): string[] => {
     if (!config?.experiments) {
         return [];
     }
--- a/suite-common/message-system/src/messageSystemUtils.ts
+++ b/suite-common/message-system/src/messageSystemUtils.ts
@@ -21,7 +21,6 @@ import type {
     Device,
     Environment,
     Condition,
-    ExperimentsItem,
 } from '@suite-common/suite-types';
 import type { NetworkSymbol } from '@suite-common/wallet-config';
 import type { TransportInfo } from '@trezor/connect';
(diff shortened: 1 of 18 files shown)


matejkriz:
I was thinking about merging `messageSystemMiddleware`s for desktop and native, since they are almost the same, but they are not exactly same and there is some import from suite package on desktop, so I believe it's better to tolerate the duplication here and keep them separated, but not a strong opinion.

adderpositive (approved):
Thank you for the excellent fine-tuning and extended functionality on mobile. I had missed `messageSystemMiddleware`, so thank you for including it.

suite-common/message-system/src/messageSystemUtils.ts
```diff
@@ -302,10 +301,7 @@ export const getValidMessages = (config: MessageSystem | null, options: Options)
         .map(action => action.message);
 };
 
-export const getValidExperiments = (
-    config: MessageSystem | null,
-    options: Options,
-): ExperimentsItem[] => {
+export const getValidExperiments = (config: MessageSystem | null, options: Options): string[] => {
```
yanascz: Do I understand correctly that this functions returns experiment IDs? 🤔 

If yes, then I'd suggest different naming:
```suggestion
export const getValidExperimentIds = (config: MessageSystem | null, options: Options): string[] => {
```

It'd also be great to make the return type more strict if possible.

yanascz (approved):
Nice and clean!

suite-common/message-system/src/messageSystemUtils.ts
matejkriz: 456888977e4e680912979bc2d54e6a52632f4fcc Renamed. The config type is auto-generated so I would not use branded type here. It can be used, but I see much more important candidates for more strict types everywhere around.

STew790:
**QA OK**
More info can be found here https://www.notion.so/satoshilabs/AB-Testing-on-mobile-18adc526060680ccb64cf4ddad4f780e

Issue #7171, opened 2022-12-08
Filter requests in main process and block not allowed domains


karliatto:
Currently suite is filtering all requests in the renderer process and blocking the ones that are to not allowed domains `packages/suite-desktop/src/modules/request-filter.ts`.

We did not have a way to filter requests in main process therefore we could not do it there but now that we have `request-interceptor` from `request-manager` package we can implement the same filter of domains for main process.

mroz22:
going through old issues now. this seems to be security related and since @komret and @peter-sanderson were browsing this field, I am tagging them, whether this is something they want to push. I am sure @karliatto would be able to assist/explain his idea

komret:
Yes, we can do this. I suppose the worst part is to put together the whitelist? Do you know which URLs should be allowed?

karliatto:
> Yes, we can do this. I suppose the worst part is to put together the whitelist? Do you know which URLs should be allowed?

Starting with 'trezor.io, 'localhost' and '127.0.0.1', also as URLs used for all the coins backends, and do a bit of research in the code and in the running app not to miss anything that is using the main process.

By looking at this list https://github.com/trezor/trezor-suite/blob/develop/packages/suite-desktop-core/src/config.ts I think some of the URLs there are not necessary in renderer anymore since the discovery in desktop is running in the main process.

PR #16259, opened 2025-01-08
fix: filter all requests not just xhr, all requests are potentially harmfull (can spy on the user via header, etc...)

peter-sanderson:
If we got somehow code injection like this, it will get through as the `request-filter.ts` only check for `xhr` requests.

image

Asset loading can be potentially harmful as well. It can be utilized for 

1. Spying on the user with headers 
2. Showing scams and phishing 
3. ... and probably more

ALL requests suite does shall be subject to the whitelisting to reduce the attack vector.

image

## For QA

- this may break some asset loading we have in case we are not whitelisting it
- I assume that something like fiat-rates, coin logos, etc... may be affected by this

Related to: https://github.com/trezor/trezor-suite/issues/7171

### DIFF
--- a/packages/suite-desktop-core/src/libs/request-interceptor.ts
+++ b/packages/suite-desktop-core/src/libs/request-interceptor.ts
@@ -1,8 +1,10 @@
 import { session } from 'electron';
 
-// Should be used for all webRequest.on* events as there can be only
-// one listener for each of them, but sometimes we need to bind more
-// of them
+/**
+ * Should be used for all webRequest.on* events as there can be only
+ * one listener for each of them, but sometimes we need to bind more
+ * of them.
+ */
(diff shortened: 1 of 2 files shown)


Lemonexe (approved):
LGTM
- Briefly looking at tokens and invity, I didn't come across any blocked asset (no log in console) :heavy_check_mark: 
- I did not succeed at displaying cat  :heavy_check_mark:

bosomt:
**QA OK**

- logos OK
- guide OK
- trade OK
- appstores icons OK
- DCA section OK
- ETH tokens OK


**Info:**
 - Suite version: desktop 25.2.0 (28f04377a323012f6488c33606c3b4f2e11a08a3)
 - OS: MacIntel
 - Screen: 1512x982

PR #16263, opened 2025-01-08
Small refactoring (code-splitting & rename) of the interceptor.ts code for main process of the Electron

peter-sanderson:
1. Commit `whitelistedHosts` -> `notRequiredTorDomainsList` to make it clear that it is not general whitelist
2. Code-split of the huge `interceptor.ts` file so every logical part has its own file (would be great to add unit-tests one day to those parts)

Part of: https://github.com/trezor/trezor-suite/issues/7171


packages/request-manager/src/interceptor/interceptorTypesAndUtils.ts
peter-sanderson, line 1: This naming sucks, any suggestions?

marekrjpolak: misc.ts :stuck_out_tongue:

marekrjpolak (approved):
Okay I guess.

peter-sanderson:
Only   │ ✖  add-account-types.test.ts  is broken, which is broken on dev now, merging

PR #16264, opened 2025-01-08
Request filtering in main process2

### LINKED ISSUE #7171: Filter requests in main process and block not allowed domains

peter-sanderson:
Resolve https://github.com/trezor/trezor-suite/issues/7171
-------------------------

- This PR adds a whitelist for all requests coming out from the main thread.
- Custom back-ends are dynamically whitelisted.

-------------------------

- [x] :exclamation: First the https://github.com/trezor/trezor-suite/pull/16263 needs to be merged
- [x] Next I need to fix `interceptor.test.ts` test
- [x] and make it run in the CI :scream_cat: 
- [x] custom back-ends (probably needs to be dynamically adjusted)
- [x] ~The whitelist needs to be probably different from the one used in render process~ (no, consulted with Carlos)
- [ ] make sure the test works with native `fetch` from node.js as well and not just with `node-fetch` => this needs to be discussed with Carlos


packages/request-manager/src/interceptor/interceptorTypesAndUtils.ts
```diff
@@ -7,9 +7,27 @@ export type InterceptorContext = InterceptorOptions & {
...
+    }
+
+    if (hostname.trim() === '') {
+        return false; // Defensively block the request
+    }
+
+    return whitelist.some(
+        whitelistedUrl =>
+            whitelistedUrl === hostname ||
+            // Todo: this .endsWith() seems fishy to me, why we need this?
```
peter-sanderson: I plan to try to remove this together with sldev in next PR and we'll see what breaks :grin:

packages/request-manager/e2e/interceptor.test.ts
```diff
@@ -2,9 +2,14 @@ import path from 'path';
 import http from 'http';
 import WebSocket from 'ws';
 
+// Todo: Currently this needs to be done in order to interceptor in test to work.
+//       This shall be taken care of, as we shall be intercepting the native fetch as well.
+// import fetch from 'node-fetch';
```
peter-sanderson, line 7: @karliatto FYI, this is what we have been talking about

packages/suite-desktop-core/src/modules/request-interceptor.ts
```diff
@@ -11,32 +11,61 @@ import { createInterceptor, InterceptedEvent } from '@trezor/request-manager';
...
+                });
+
+                return;
+
+            case 'CIRCUIT_MISBEHAVING':
+                mainThreadEmitter.emit('module/reset-tor-circuits', event);
+
+                return;
+
+            case 'INTERCEPTED_HEADERS':
```
peter-sanderson, line 64: Those events does not seem to be handled anywhere, despite being (potentially) emitted. 

I am not sure if this switch is supposed to be actually exhaustive, but as those are not handled anywhere, else (and also not here originally).

packages/request-manager/src/interceptor/interceptNetSocketConnect.ts
```diff
@@ -73,6 +76,9 @@ export const interceptNetSocketConnect = (context: InterceptorContext) => {
             details = typeof callback === 'string' ? `${callback}:${request}` : request.toString();
         }
 
+        const hostname = details.split(':')[0];
```
peter-sanderson, line 79: - [x] Todo: This is sub-optimal and it wont probably work in all scenarios. Any ideas?

packages/request-manager/src/interceptor/interceptTlsConnect.ts
```diff
@@ -1,25 +1,43 @@
...
+            }
+
+            if (typeof optionsOrHost === 'string') {
+                return optionsOrHost;
+            }
+
+            return '';
+        };
+
+        const hostname = getHostname().split(':')[0] ?? '';
```
peter-sanderson: - [x] Todo: This is sub-optimal and it wont probably work in all scenarios. Any ideas?

packages/request-manager/src/interceptor/overloadHttpRequest.ts
```diff
@@ -27,22 +27,36 @@ const getIdentityForAgent = (options: Readonly<http.RequestOptions>) => {
...
-) => {
+export const overloadHttpRequest = ({
+    context,
+    protocol,
+    url,
+    options,
+    callback,
+    validateRequest,
+}: OverloadHttpRequestParams) => {
+    const hostname = typeof url === 'object' ? url.hostname ?? url.host ?? '' : '';
```
Lemonexe: this means that for string url, hostname will always be `''`, which will always be false in `isWhitelistedHost`

packages/suite-desktop-core/src/modules/request-interceptor.ts
```diff
@@ -11,32 +11,71 @@ import { createInterceptor, InterceptedEvent } from '@trezor/request-manager';
...
+
+            case 'ADD_WHITELISTED_DOMAIN':
+                mainThreadAllowedDomain.general.push(event.domain);
+
+                return;
 
-        if (event.type === 'CIRCUIT_MISBEHAVING') {
-            mainThreadEmitter.emit('module/reset-tor-circuits', event);
+            default: {
+                const _exhaustiveCheck: never = event; // Poor-man's `switch-exhaustiveness-check`
```
Lemonexe, line 79: why poor man's? This is absolutely fine :smile: 
Although typed record is better when mapping primitive values to cases, I would not use it to map functions. That would not read very well IMO.

packages/suite-desktop-core/src/modules/request-interceptor.ts
```diff
@@ -11,32 +11,71 @@ import { createInterceptor, InterceptedEvent } from '@trezor/request-manager';
...
+                mainThreadEmitter.emit('module/reset-tor-circuits', event);
+
+                return;
+
+            case 'INTERCEPTED_HEADERS':
+            case 'ERROR':
+                return;
+
+            case 'SET_WHITELISTED_DOMAINS_FOR_CUSTOM_BACKENDS':
+                mainThreadAllowedDomain.customBackends[event.coin] = event.domains;
```
Lemonexe, line 69: there is no reverse action when resetting a custom backend to default, so it will probably persist as enabled until you restart app or set custom backend again for the same coin. 
I'm not saying it's needed, just a note. I think it's OK.

packages/suite-desktop-core/src/modules/trezor-connect.ts
```diff
@@ -2,11 +2,43 @@ import { ipcMain } from 'electron';
...
+    if (param !== undefined && param.blockchainLink !== undefined) {
+        const domains = (param.blockchainLink.url ?? []).map(url => {
+            const electrumUrlResult = parseElectrumUrl(url);
+            if (electrumUrlResult !== undefined) {
+                return electrumUrlResult.host;
+            }
+
+            const urlObj = new URL(url);
+
+            return urlObj.hostname ?? urlObj.host;
```
Lemonexe: FYI here you can see difference between hostname and host https://developer.mozilla.org/en-US/docs/Web/API/Location
I'd only use hostname (w/o port), so the condition is broader = more restrictive

Also, both are typed as string, so I expect the `??` to be useless, maybe you meant `||` ?

Lemonexe (commented):
Did not test yet, CR'd most of code, so dropping off comments :speech_balloon:

packages/request-manager/src/interceptor/interceptTlsConnect.ts
Lemonexe: as we discussed the `new URL(address).hostname` - I just now realised you could create a helper that would do it _optionally_, something like:
```
optionalGetUrlHostname = (address: string) => {
	try {
		return new URL(address).hostname;
	}
	catch(_err) {
		return null;
	}
}

```

then use it like `hostname = optionalGetUrlHostname(address) ?? ''`
`file:///...` :arrow_right: `''` :heavy_check_mark:
invalid URLs  :arrow_right: `null`  :arrow_right: `''` :heavy_check_mark:

packages/request-manager/src/interceptor/interceptorTypesAndUtils.ts
Lemonexe: Since this fn is called only with _hostname_ extracted from the url, then we are safe that you cannot loophole it like that:
`http://scammy-url.net?uselessParam=trezor.io`

I think this is necessary to allows subdomains: `btc1.trezor.io` (would break? ), or `anything.that.ends.with.trezor.io` which is guaranteed not to be `scammy-url.net` so it's ok :heavy_check_mark:

peter-sanderson: But what about `scammy-url-trezor.io`? :point_up:

Lemonexe: Riiight. I missed that :man_facepalming: 
So in current form it should be removed. But if this is needed for subdomains (I assume it is), it could be modified as `hostname.endsWith('.' + whiteListedUrl)` maybe? :thinking:

packages/request-manager/src/interceptor/interceptTlsConnect.ts
peter-sanderson: Food point => https://github.com/trezor/trezor-suite/pull/16264/commits/a0f2e9d39ffeaec4367a7481fe78d62b6dbada67

packages/suite-desktop-core/src/modules/request-interceptor.ts
peter-sanderson: > I would not use it to map functions.

I would, but it probably not possible if you need type-narrowing.

packages/request-manager/src/interceptor/interceptorTypesAndUtils.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/16264/commits/c3988a29c559888dd336040eaaba0ff5bf5d5b35

packages/request-manager/src/interceptor/overloadHttpRequest.ts
```diff
@@ -36,6 +36,14 @@ type OverloadHttpRequestParams = {
     validateRequest: (params: { hostname: string }) => void;
 };
 
+const resolveHostname = (url: string | URL | http.RequestOptions) => {
+    if (typeof url === 'object') {
```
Lemonexe: tiny nit: makes sense in this context, but it's not very precise..
I'd prefer being consistent with type of the `url` param:
either `typeof url !== string` (leaving the rest) or
`url instanceof URL || url instanceof idk what?`

packages/request-manager/src/interceptor/interceptNetConnect.ts
```diff
@@ -9,21 +9,29 @@ export const interceptNetConnect: Interceptor = ({ context, validateRequest }) =
         const [options, callback] = args;
 
         let details: string;
+        let hostname: string;
```
Lemonexe, line 12: extracting fn with returns would be better than `let`, but this is still fine :ok_hand:

packages/request-manager/src/interceptor/interceptTlsConnect.ts
```diff
@@ -1,38 +1,42 @@
...
+            // case: connect(port: number, host?: string, options?: ConnectionOptions, secureConnectListener?: () => void): TLSSocket;
+            // case: connect(port: number, options?: ConnectionOptions, secureConnectListener?: () => void): TLSSocket;
+
+            hostname =
+                // eslint-disable-next-line no-nested-ternary
+                typeof optionsOrHost === 'object'
+                    ? optionsOrHost.host ?? ''
+                    : typeof optionsOrHost === 'string'
+                      ? optionsOrHost
+                      : 'unknown';
```
Lemonexe: again, using a fn with returns would solve both `let` and nested ternary.
Does not have to be top scope with crazy interface,
IMO it'd be fine to define it here within the `(...args) => {` callback :slightly_smiling_face:

packages/suite-desktop-core/src/modules/request-filter.ts
```diff
@@ -3,19 +3,26 @@
...
 
 import type { ModuleInit } from './index';
 
 export const SERVICE_NAME = 'request-filter';
 
+/**
+ * This module handles request interception for the renderer thread. Requests are happening
+ * in the Chromium browser, so we have to relay on the apo we get from Electron
+ * (it falls down to `session.defaultSession.webRequest.onBeforeRequest`).
+ */
```
Lemonexe, line 18: Before you wrote the comments, I did not realise we have so similar sounding modules one for renderer and one for main respectively, so :+1: :+1:

packages/request-manager/src/interceptor/interceptNetSocketConnect.ts
Lemonexe: As per discussion this is least evil :smiling_imp:

packages/utils/tests/isWhitelistedHost.test.ts
```diff
@@ -0,0 +1,21 @@
...
+        { hostname: 'scam-url.io', result: false },
+        { hostname: 'scam-url-trezor.io', result: false },
+    ];
+
+    dataProvider.forEach(({ hostname, result }) => {
+        it(`The '${hostname}' is ${result ? 'is allowed' : 'is NOT allowed'}`, () => {
+            expect(isWhitelistedHost(hostname, WHITELISTED)).toBe(result);
+        });
+    });
+});
```
Lemonexe, line 21: Of all the tests we have for this topic, I consider this one to be most effective and essential, so thx for adding it

Lemonexe (approved):
Now it LGTM, but there are small code nits :speech_balloon:  please consider and resolve.
I would feel much better if someone else reviews this too. It's a complex topic and security mistakes are not always easy to spot :pray: 

Also, unit tests are :x: , but I confirmed locally that they pass, so please investigate.

Application works as supposed to with all coins turned on (account discovery, tokens table, staking, invity buy/sell/swap). Observed no bugs.

Tested both main  :heavy_check_mark: 
`2025-01-15T13:56:54.825Z - WARN(request-filter): http://www.seznam.cz/ was blocked because www.seznam.cz is not in the exception list`
and renderer (as it was changed) :heavy_check_mark: 
image

packages/utils/tests/isWhitelistedHost.test.ts
peter-sanderson: :bow:

packages/request-manager/src/interceptor/interceptTlsConnect.ts
peter-sanderson: done

packages/request-manager/src/interceptor/interceptNetConnect.ts
peter-sanderson: ill keep it as i is for now

packages/request-manager/src/interceptor/overloadHttpRequest.ts
peter-sanderson: done

peter-sanderson:
image
image

karliatto (commented):
I started to look at it, and it looks OK, but I need to look into it in more detail.

packages/request-manager/e2e/interceptor.test.ts
karliatto: Ok, I am bit worried about this. I know it is not related to this PR but this tests were passing before and as they are. 
I am afraid it has been broken with some dependency upgrade and since it is not run in the CI it was not catch. I am going to look into details of it.

packages/suite-desktop-core/src/modules/request-filter.ts
```diff
@@ -3,19 +3,26 @@
...
 export const init: ModuleInit = ({ interceptor }) => {
     const { logger } = global;
 
     interceptor.onBeforeRequest(details => {
         const { hostname } = new URL(details.url);
 
-        if (allowedDomains.find(d => hostname.endsWith(d)) !== undefined) {
+        if (isWhitelistedHost(hostname, allowedDomains)) {
```
karliatto, line 25: I like this utility function used for whitelisting here and in the main process. :+1:

packages/suite-desktop-core/src/modules/coinjoin.ts
```diff
@@ -117,6 +117,13 @@ export const init: ModuleInit = ({ mainWindowProxy, store, mainThreadEmitter })
 
     const clientProxyOptions: IpcProxyHandlerOptions<CoinjoinClient> = {
         onCreateInstance: async (settings: ConstructorParameters<typeof CoinjoinClient>[0]) => {
+            const urlObj = new URL(settings.coordinatorUrl);
+
+            mainThreadEmitter.emit('module/request-interceptor', {
+                type: 'ADD_WHITELISTED_DOMAIN',
+                domain: urlObj.hostname ?? urlObj.host,
+            });
```
karliatto, line 125: I am not sure I understand why this is needed.

karliatto (approved):
ACK https://github.com/trezor/trezor-suite/pull/16264/commits/2c70ef1a0f72c8ca0b70216142f54d4dd5a034ad

Looks nice and works.

Nit: I would add a new commit for the changes in utils package. It is a package that we release to NPM and having specific comment for it helps us to generate proper automated CHANGELOGs. Somehting like `"feat(utils): new isWhitelistedHost"`

peter-sanderson:
Thanks @karliatto I have separated the commit.

bosomt:
**QA OK**


**Info:**
 - Suite version: web 25.2.0 (61f1d572d2fd32d928d78310b5b3dc48410cb380)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/132.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Transport: WebUsbTransport

PR #16269, opened 2025-01-08
feat(mobile): make feature flags more dummy proof

Nodonisko:
## Description

Make feature flag more dummy proof so it is not possible to have `always truthy` condition then. We only need to read FF value anyway and write it only in single place.

QA: Please test that all FF works correctly 😄 

## Related Issue

Resolve


yanascz (approved):
Looks good! 👍

matejkriz:
Do you think the hooks are worth it? Wouldn't it be easier to just use a selector or dispatch toggle action everywhere?

PeKne:
Rebase to make e2e tests passing again: https://github.com/trezor/trezor-suite/pull/16276

matejkriz (approved):
After the [internal discussion](https://satoshilabs.slack.com/archives/C03H2Q5MX51/p1736438555083419), I'm ok with keeping it wrapped in simple hooks 👌

STew790:
**QA OK**
All FF works correctly.
**Info**
25.1.2 https://github.com/trezor/trezor-suite/commit/59e12ce461f58b33844d7ccea99ac79490329267

PR #16272, opened 2025-01-08
fix(suite-native): tweak screen top padding

yanascz:
Fixed several issues with screen top padding visible on Android.

### DIFF
--- a/suite-native/navigation/src/components/Screen.tsx
+++ b/suite-native/navigation/src/components/Screen.tsx
@@ -24,7 +24,6 @@ type ScreenProps = {
     hasStatusBar?: boolean;
     isScrollable?: boolean;
     backgroundColor?: Color;
-    noTopPadding?: boolean;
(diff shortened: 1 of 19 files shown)


suite-native/navigation/src/components/Screen.tsx
```diff
@@ -24,7 +24,6 @@ type ScreenProps = {
     hasStatusBar?: boolean;
     isScrollable?: boolean;
     backgroundColor?: Color;
-    noTopPadding?: boolean;
```
matejkriz, line 27: It's great you removed these two custom params 👏

matejkriz (approved):
I've tried to check all affected screens on android and everything seems to look good to me 👏  Haven't check on iOS

STew790:
**QA OK**
Discussed with yanas and LGTM. Checked on iOS as well.

**Info**

25.1.2 59e12ce

PR #16286, opened 2025-01-09
chore(blookbook): rename token type to standard

### LINKED ISSUE #16251: Unify token standard/type
For some reason we introduced `type` to `TokenTransfer` object which override token standard, so standard is now located in `standard` attribute in case of `TokenTransfer`.
However, in case of `TokenInfo` in account, token standard is in `type` attribute.

Please unify.

Find out what is better, if using `type` from blockbook API or `standard`.

Do not forget about migration. 

Then we will be able to get rid of this:
```ts
export const isNftTokenTransfer = (transfer: TokenTransfer) =>
    ['ERC1155', 'ERC721', 'BEP1155', 'BEP721'].includes(transfer.standard || '');

// TODO: TokenInfo should use TokenStandard type
export const isNftToken = (token: TokenInfo) =>
    ['ERC1155', 'ERC721', 'BEP1155', 'BEP721'].includes(token.type || '');
```

enjojoy:
## Description

Rename token type to standard

## Related Issue.
Resolve https://github.com/trezor/trezor-suite/issues/16251

## Screenshots
Screenshot 2025-01-13 at 12 16 35


packages/blockchain-link-types/src/blockbook-api.ts
```diff
@@ -1,5 +1,7 @@
 /* Do not change, this code is generated from Golang structs */
```
marekrjpolak, line 1: I believe this file should not be changed manually as it originates from [Blockbook repo](https://github.com/trezor/blockbook/blob/master/blockbook-api.ts). Make the necessary changes in `blockbook.ts ` file instead (if they're really necessary, which may not be this case; see the main comment).

packages/blockchain-link-utils/src/blockfrost.ts
```diff
@@ -115,7 +115,7 @@ export const transformTokenInfo = (
     }
 
     const info = tokens.map(token => ({
-        type: 'BLOCKFROST',
+        standard: 'BLOCKFROST',
```
marekrjpolak, line 118: If we wanna make the whole deprecation thing in a non-breaking way (which I don't insist on), all the currently outgoing data must be preserved and some new may be added => we want to signal that `type` can no longer be relied on, but it's not possible to remove it yet.

packages/blockchain-link-utils/src/solana.ts
```diff
@@ -149,7 +149,7 @@ export const transformTokenInfo = (
                 } = tokenAccount.account.data;
 
                 return {
-                    type: tokenProgramsInfo[program].tokenStandard,
+                    standard: tokenProgramsInfo[program].tokenStandard,
```
marekrjpolak, line 152: dtto

packages/suite/src/storage/migrations/index.ts
```diff
@@ -868,7 +868,7 @@ export const migrate: OnUpgradeFunc<SuiteDBSchema> = async (
                         decimals: token.decimals,
                         fingerprint: token.name,
                         policyId,
-                        type: token.type,
+                        standard: token.standard,
```
marekrjpolak, line 871: Never ever change old migrations code, please. No `standard` had existed at that time.

packages/blockchain-link-utils/src/blockbook.ts
```diff
@@ -327,7 +327,7 @@ export const transformAddresses = (
 ): AccountAddresses | undefined => {
     if (!tokens || !Array.isArray(tokens)) return undefined;
     const addresses = tokens.reduce((arr, t) => {
-        if (t.type !== 'XPUBAddress') return arr;
+        if (t.standard !== 'XPUBAddress') return arr;
```
marekrjpolak, line 330: It's a bit funny to have token standard `XPUBAddress`, but it was funny even as token type, so this is no problem. But, as there's no backend data transformation up to this point, field `standard` is never there so BTC accounts have no addresses and therefore are not shown in Suite (at least in my case :man_shrugging:)

marekrjpolak (changes requested):
Apart from the comments, this is not working. It's not enough to change the types, because they're also coming (e.g.) from Blockbook. That's why there's the generated `blockbook-api` file.

If we really want to restructure the data, we have to explicitly do it after we receive them in the old format (unless backend API is changed as well which I don't presume).

enjojoy:
> Apart from the comments, this is not working. It's not enough to change the types, because they're also coming (e.g.) from Blockbook. That's why there's the generated `blockbook-api` file.
> 
> If we really want to restructure the data, we have to explicitly do it after we receive them in the old format (unless backend API is changed as well which I don't presume).

Thank you, I feel like I'm going to learn a lot from this PR

packages/suite/src/storage/migrations/index.ts
enjojoy: Oh this one was changed by mistake, thanks for the catch

packages/blockchain-link-utils/src/blockbook.ts
enjojoy: I see, well XPUBAddress is there under "type" even in blockbook, alongside with ERC types... 

Also for me Bitcoin accounts are displayed but they are empty, so maybe that's why I didn't notice the bug, will investigate

marekrjpolak: > XPUBAddress is there under "type" even in blockbook

Yes I know, that's fine. But if they're coming with `type`, we can't detect them based on `standard`.

enjojoy:
I just agreed with @martinboehm that we will add standard on blockbook side and later deprecate type there. 

Will proceed with this PR when it will be renamed 

cc @marekrjpolak

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -484,7 +484,7 @@ export const isNftMultitokenTransfer = (transfer: TokenTransfer) =>
 
 // TODO: TokenInfo should use TokenStandard type
 export const isNftToken = (token: TokenInfo) =>
-    ['ERC1155', 'ERC721', 'BEP1155', 'BEP721'].includes(token.type || '');
+    ['ERC1155', 'ERC721', 'BEP1155', 'BEP721'].includes(token.standard);
```
tomasklim, line 487: I think `isNftTokenTransfer` and `isNftToken` can now be unified

enjojoy:
Waiting for the completion of https://github.com/trezor/blockbook/pull/1185

From blockbook side

suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -940,7 +940,7 @@ export const simpleSearchTransactions = (
     const foundTxsForToken = transactions.flatMap(transaction => {
         const hasMatchingToken = transaction.tokens.some(
             token =>
-                isTokenMatchesSearch(token, search.toLowerCase()) ||
+                isTokenTransferMatchesSearch(token, search.toLowerCase()) ||
```
tomasklim, line 943: why this change? it will break for cardano

tomasklim (changes requested):

enjojoy:
Merged together with eip1559

PR #16298, opened 2025-01-09
feat(e2e): Migrated address metadata test + refactoring

HajekOndrej:
## Description

Migrated address metadata test + refactoring agreed in the last PR.

## Related Issue

Resolve


packages/suite-desktop-core/e2e/support/enums/accountLabelId.ts
```diff
@@ -0,0 +1,7 @@
+export enum AccountLabelId {
+    BitcoinDefault1 = "m/84'/0'/0'",
```
HajekOndrej, line 2: Unfortunately the complex IDs are there for a reason and it would not be trivial to fix that, it would require some (maybe even complex) UI refactoring.
They are used in more tests, that will be migrated, so I have created this enum.

packages/suite-desktop-core/e2e/support/metadataProviderMocks.ts
Vere-Grey, line 1: :+1:

packages/suite-desktop-core/e2e/support/enums/accountLabelId.ts
Vere-Grey: IMHO that is good solution. It improves the readability a lot. Before it was lot cryptic.
Once we convert all cypress test, we can revisit it and see if we can improve it further.

packages/suite-desktop-core/e2e/tests/metadata/address-metadata.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        await page.getByTestId('@wallet/receive/used-address/show-more').click();
+        await page.getByTestId(`${metadataEl}/add-label-button`).click();
+
+        // Initialize metadata flow
+        // Metadata provider: google
+        metadataPage.passThroughInitMetadata(MetadataProvider.GOOGLE);
+
+        await metadataPage.metadataInput.fill('meow address');
+        await page.keyboard.press('Enter');
+        await page.pause();
```
Vere-Grey: forgotten pause

Vere-Grey (changes requested):
LGTM, just one forgotten pause

Vere-Grey (approved):
:+1:

PR #16311, opened 2025-01-10
9052 maximum characters on account import

### LINKED ISSUE #9052: Maximum characters not checked during account import
**Describe the bug**
- during import of account/address you can change default Coin label  but we dont limit characters amount till you press Confirm
- during renaming  of account  when already imported we check characters amount and we display count and maximum of 30 chars allowed

**Info:**

-  1b4cd62 - [develop] - fix(suite-native): use base jest config instead of native in assets pkg

**Screenshots**
image

jbazant:
## Description

Also fixed some translations.

## Related Issue

Resolve #9052

## Screenshots:

https://github.com/user-attachments/assets/d2eca68a-6b6a-4571-81c4-683b22b31aa2

### DIFF
--- a/suite-native/module-accounts-import/src/components/AccountImportOverview.tsx
+++ b/suite-native/module-accounts-import/src/components/AccountImportOverview.tsx
@@ -7,6 +7,7 @@ import {
 import { RoundedIcon, VStack } from '@suite-native/atoms';
 import { isTestnet } from '@suite-common/wallet-utils';
 import { TextInputField } from '@suite-native/forms';
+import { useTranslate } from '@suite-native/intl';
 
 import { AccountImportOverviewCard } from './AccountImportOverviewCard';
 
@@ -16,11 +17,17 @@ type AssetsOverviewProps = {
 };
 
 export const AccountImportOverview = ({ balance, symbol }: AssetsOverviewProps) => {
+    const { translate } = useTranslate();
+
     const fiatBalanceValue = useFiatFromCryptoValue({
         cryptoValue: balance,
         symbol,
     });
+    const coinLabelFieldLabel = translate(
+        'moduleAccountManagement.accountSettingsScreen.renameForm.coinLabel',
+    );
 
+    // TODO 9052
     return (
         <AccountImportOverviewCard
             icon={<RoundedIcon symbol={symbol} iconSize="large" />}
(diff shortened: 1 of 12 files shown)


suite-native/module-accounts-management/src/components/AccountRenameButton.tsx
```diff
@@ -22,7 +28,7 @@ export const AccountRenameButton = ({ accountKey }: AccountRenameModalProps) =>
                 onPress={handleOpen}
                 testID="@account-detail/settings/edit-button"
             />
-            <BottomSheet title="Rename coin" isVisible={isVisible} onClose={handleClose}>
+            <BottomSheet title={bottomSheetTitle} isVisible={isVisible} onClose={handleClose}>
```
vytick: You can use `title={<Translation id="translation-id-string" />}`

Please use Translation where possible, because we might add more functionality to `Translation` later (thinking about adding string formatting there).

suite-native/module-accounts-import/src/components/AccountImportOverview.tsx
```diff
@@ -37,12 +51,16 @@ export const AccountImportOverview = ({ balance, symbol }: AssetsOverviewProps)
...
-                    testID="@account-import/coin-synced/label-input"
-                    name="accountLabel"
-                    label="Coin label"
-                    elevation="1"
-                />
+                <VStack spacing="sp8">
+                    <TextInputField
+                        testID="@account-import/coin-synced/label-input"
+                        name="accountLabel"
+                        label={coinLabelFieldLabel}
```
vytick, line 58: please, also here use `<Translation `

vytick (commented):
Works nicely 👍. Just a minor thing: Please, use `<Translation ...` instead of `useTranslate` hook where possible.

suite-native/module-accounts-import/src/components/AccountImportOverview.tsx
jbazant: Actually no, it is not possible to use Translation component here. `label` prop expects `string` not an `Element`

suite-native/module-accounts-management/src/components/AccountRenameButton.tsx
jbazant: Fixed

suite-native/module-accounts-import/src/components/AccountImportOverview.tsx
vytick: Oh, my bad, you are right 👍

vytick (approved):
Good job 👍 🚀

PR #16312, opened 2025-01-10
fix(suite-native): use correct receive flow screens for connected device

yanascz:
- Incorrectly removed screens returned back
- Flow unified for _Portfolio tracker_ and connected device
- Receive screens refactored to prevent UI & scrolling glitches

### DIFF
--- a/suite-native/accounts/src/components/AddAccountsButton.tsx
+++ b/suite-native/accounts/src/components/AddAccountsButton.tsx
@@ -16,7 +16,6 @@ import {
     selectIsDeviceInViewOnlyMode,
     selectIsPortfolioTrackerDevice,
 } from '@suite-common/wallet-core';
-import { FeatureFlag, useFeatureFlag } from '@suite-native/feature-flags';
 
 import { useAccountAlerts } from '../hooks/useAccountAlerts';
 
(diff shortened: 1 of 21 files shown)


suite-native/module-receive/src/screens/ReceiveAccountsScreen.tsx
```diff
@@ -0,0 +1,45 @@
...
+            accountKey: account.key,
+            tokenContract: tokenAddress,
+            closeActionType: 'back',
+        });
+
+    return (
+        <Screen
+            screenHeader={
+                <ScreenSubHeader
+                    content={translate('moduleReceive.receiveTitle')}
```
matejkriz: Why not `Translation` component? It is a preferred way whenever possible.

yanascz: Copy pasta from some other place. Fixed.

yanascz: That's the way it's used in `ScreenSubHeader`. Using `Translation` would require some refactoring.

suite-native/accounts/src/components/AddAccountsButton.tsx
```diff
@@ -58,15 +53,16 @@ export const AddAccountButton = ({ flowType, testID }: AddAccountButtonProps) =>
...
-    return shouldShowAddAccountButton ? (
+    return (
         <IconButton
             iconName="plus"
             onPress={
                 isSelectedDevicePortfolioTracker ? navigateToImportScreen : navigateToAddCoinAccount
             }
             colorScheme="tertiaryElevation0"
             size="medium"
+            isLoading={hasDeviceDiscovery}
```
vytick, line 64: When discovery is running (and the icon is loading), tapping that button starts Add new coin flow. Is that intended?

yanascz: Great catch, that's a bug. Will fix in a fixup.

suite-native/module-receive/src/screens/ReceiveAccountsScreen.tsx
matejkriz: Now I see it, ok then, better to refactor separately, not in this PR.

matejkriz (commented):
From testing I see only one issue at the very end of the receive flow - 

When I complete the entire receive flow confirm the receive address on the Trezor (or skip the confirmation in the Portfolio tracker), the only way back is the Back icon, but that only takes me back one screen to the Receive flow - selecting another account. 

As a user, I would expect to return to the Dashboard so I don't have to click twice. Ideally the icon can be changed to X in that case.

What do you think?

yanascz:
> When I complete the entire receive flow confirm the receive address on the Trezor (or skip the confirmation in the Portfolio tracker), the only way back is the Back icon, but that only takes me back one screen to the Receive flow - selecting another account.
> 
> As a user, I would expect to return to the Dashboard so I don't have to click twice. Ideally the icon can be changed to X in that case.

Let's solve in a follow-up since it requires bigger refactoring of the navigation.

matejkriz (approved):

STew790:
**QA OK**
The "Receive" button now completely "mimics" the old Receive tab. 
 - im able to select any enabled coin and account from that coin
 - im able to add an account
 - and therefore enable a coin
 - portfolio works the same
 - I am asked to enter an xpub/address when adding new account/coin
 - after that I am returned to the receive :+1: 
 
image

 
 **Info**
 25.1.2 caac07f

PR #16316, opened 2025-01-12
feat(e2e): Convert to PW transaction test

Vere-Grey:
## Description

Converts test suite transactions to PW

## Related Issue

Resolve 15606


packages/suite-desktop-core/e2e/support/fixtures.ts
```diff
@@ -31,10 +32,12 @@ import { AssetsActions } from './pageActions/assetsActions';
 import { MetadataProviderMocks } from './metadataProviderMocks';
 import { MetadataActions } from './pageActions/metadataActions';
 
+type StartEmuModelRequired = StartEmu & { model: Model };
```
Vere-Grey, line 35: So, we started using a new model default in our fixtures - T3T1.
This means we need to make this property a required value to eliminate risk of accidentally overriding the default.
Because setting a `test.use({emulatorStartConf: {wipe: true}})` on a test level would override the T3T1 default from fixtures with nothing and the default T2T1 from TrezorUserEnvLink class would be used.

Vere-Grey: I hope the naming at least help partially. Because the error is not very clear when you don't include the `model` property.

packages/suite-web/e2e/tests/wallet/look-up-an-account.test.ts
Vere-Grey, line 1: I have forgotten to remove this test file in my older commit

HajekOndrej (approved):

PR #16317, opened 2025-01-12
feat(mobile): upgrade FW report to analytics

### LINKED ISSUE #16201: Mobile FW update analytics
Proposed by @shenkys 

1. Avg. time spent
2. “I feel stuck” tapped
   a. Users that went to first sheet (have backup?)
   b. Users went to second sheet (recommended troubleshooting)
3. Disconnected devices - NOT IMPLEMENTED
4. App left - NOT IMPLEMENTED
5. Update failed count - can be calculated from events including error in one session.

Don't forget to put it in [Overview and Changelog](https://www.notion.so/satoshilabs/Data-Analytics-938aeb2e289f4ca18f31b1c02ab782cb?pvs=4#77977eb76f9b429fb068c2090ea6c235).

Nodonisko:
## Description

TODO: 
1. Test it again once we have final version of events

## Related Issue

Resolve #16201

### DIFF
--- a/suite-native/firmware/src/hooks/useFirmwareAnalytics.ts
+++ b/suite-native/firmware/src/hooks/useFirmwareAnalytics.ts
@@ -1,0 +1,115 @@
+import { useCallback, useEffect, useRef } from 'react';
+import { useSelector } from 'react-redux';
+
+import { DeviceModelInternal, FirmwareType } from '@trezor/connect';
+import {
+    analytics,
+    EventType,
+    FirmwareUpdatePayload,
+    FirmwareUpdateStartType,
+} from '@suite-native/analytics';
+import { TrezorDevice } from '@suite-common/suite-types';
+import { getFirmwareVersion, getBootloaderVersion } from '@trezor/device-utils';
+import { selectDeviceUpdateFirmwareVersion } from '@suite-common/wallet-core';
+
+export const useFirmwareAnalytics = ({
+    device,
+    targetFirmwareType,
+}: {
+    device?: TrezorDevice;
+    targetFirmwareType: FirmwareType;
+}) => {
+    const toFwVersion = useSelector(selectDeviceUpdateFirmwareVersion);
+
+    const prepareAnalyticsPayload = useCallback(
+        () => ({
+            model: device?.features?.internal_model ?? DeviceModelInternal.UNKNOWN,
+            fromBootloaderVersion: getBootloaderVersion(device),
+            fromFwVersion: device?.firmware === 'none' ? 'none' : getFirmwareVersion(device),
+            toFwVersion: toFwVersion ?? '?.?.?',
+            fromFwType: (device?.firmwareType || 'none') as FirmwareType | 'none',
+            toFwType: targetFirmwareType,
+        }),
+        [device, targetFirmwareType, toFwVersion],
+    );
+
+    // Use refs to avoid any re-renders because of analytics and to make useCallback dependencies stable
+    // so it won't trigger any useEffect which could interfere with other business logic.
+    const analyticsPayload = useRef<FirmwareUpdatePayload>(prepareAnalyticsPayload());
+    const timeStarted = useRef<number>(Date.now());
+
+    useEffect(() => {
+        analyticsPayload.current = prepareAnalyticsPayload();
+    }, [prepareAnalyticsPayload]);
+
+    const getElapsedTimeInSeconds = useCallback(
+        () => Math.floor((Date.now() - timeStarted.current) / 1000),
+        [],
+    );
+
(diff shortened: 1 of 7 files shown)


suite-native/firmware/src/hooks/useFirmwareAnalytics.ts
```diff
@@ -0,0 +1,112 @@
...
+}: {
+    device?: TrezorDevice;
+    targetFirmwareType: FirmwareType;
+}) => {
+    const prepareAnalyticsPayload = useCallback(() => {
+        return {
+            model: device?.features?.internal_model ?? DeviceModelInternal.UNKNOWN,
+            fromBootloaderVersion: getBootloaderVersion(device),
+            fromFwVersion: device?.firmware === 'none' ? 'none' : getFirmwareVersion(device),
+            toFwVersion: targetFirmwareType,
```
matejkriz: This should contain FW version and not FW type.

suite-native/firmware/src/hooks/useFirmwareAnalytics.ts
```diff
@@ -0,0 +1,112 @@
...
+    device,
+    targetFirmwareType,
+}: {
+    device?: TrezorDevice;
+    targetFirmwareType: FirmwareType;
+}) => {
+    const prepareAnalyticsPayload = useCallback(() => {
+        return {
+            model: device?.features?.internal_model ?? DeviceModelInternal.UNKNOWN,
+            fromBootloaderVersion: getBootloaderVersion(device),
```
matejkriz, line 27: This will most likely not work or always return only empty string, because `getBootloaderVersion` util works only if the device is in bootloader mode. I think we don't have a way how to read the bootloader version in normal mode. But we can assume the BL version from FW version anyway, so this is probably redundant.

matejkriz (commented):
Happy path works nice except the missing toFwVersion and empty fromBootloaderVersion
image

matejkriz (commented):
Cancelation also OK
image

suite-native/firmware/src/hooks/useFirmwareAnalytics.ts
Nodonisko: Didn't know that but it seems that it should safe to drop, also according to Tomas less params is better anyway :D

suite-native/firmware/src/hooks/useFirmwareAnalytics.ts
```diff
@@ -0,0 +1,112 @@
...
+        return analyticsPayload.current;
+    }, [analyticsPayload]);
+
+    const resetTimeStarted = useCallback(() => {
+        timeStarted.current = Date.now();
+    }, []);
+
+    const handleAnalyticsReportStarted = useCallback(
+        ({ startType }: { startType: FirmwareUpdateStartType }) => {
+            // Q: Maybe we should reset timeStarted only if it's not a retry?
```
matejkriz: I believe it's ok to reset the timer, because we report the finished and new start event in that case.

Nodonisko:
From you screenshot it looks that Cancellation flow is not OK, it should not send finished event but just Cancel event

suite-native/firmware/src/screens/FirmwareUpdateInProgressScreen.tsx
```diff
@@ -106,12 +120,17 @@ export const FirmwareUpdateInProgressScreen = () => {
                 // Action cancelled on device
                 result.payload?.code === 'Failure_ActionCancelled'
             ) {
+                handleAnalyticsReportCancelled();
                 navigation.navigate(DeviceStackRoutes.FirmwareUpdate);
             }
 
+            handleAnalyticsReportFinished({ error: result.payload?.error });
```
matejkriz: What about adding here some default error message so it doesn't look like a success in analytics if result.payload?.error is empty?

Nodonisko: This is success in case it's empty

Nodonisko: Good idea

Nodonisko: [`152ff45` (#16317)](https://github.com/trezor/trezor-suite/pull/16317/commits/152ff457318828fa0ff6697bfe81043ba72476d9)

suite-native/firmware/src/hooks/useFirmwareAnalytics.ts
Nodonisko: [`152ff45` (#16317)](https://github.com/trezor/trezor-suite/pull/16317/commits/152ff457318828fa0ff6697bfe81043ba72476d9)

matejkriz (approved):

PR #16318, opened 2025-01-12
feat(mobile): update FW alert on Home + remote feature flag

Nodonisko:
## Description

1. Added update alert box to Homescreen
2. Added remote feature flag for message system to disable => tested with local JWS config
3. Removed local feature flag in favor of remote one

## Related Issue

Resolve part of #15584 

## Screenshots:

https://github.com/user-attachments/assets/8fac162f-39c3-4f54-b314-23ae1f0b48f4

### DIFF
--- a/suite-common/message-system/src/messageSystemTypes.ts
+++ b/suite-common/message-system/src/messageSystemTypes.ts
@@ -25,6 +25,8 @@ export const Feature = {
     ethClaim: 'eth.staking.claim',
     firmwareRevisionCheck: 'security.firmware.check',
     firmwareHashCheck: 'security.firmware.hashCheck',
+    // FW update feature flag implemented only for mobile app
+    firmwareUpdate: 'device.firmware.update',
 } as const;
 
 export type FeatureDomain = (typeof Feature)[keyof typeof Feature];
--- a/suite-native/atoms/src/Stack.tsx
+++ b/suite-native/atoms/src/Stack.tsx
@@ -1,4 +1,6 @@
-import { ReactNode } from 'react';
+import React, { ReactNode } from 'react';
+import Animated from 'react-native-reanimated';
+import { View } from 'react-native';
 
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 import { NativeSpacing } from '@trezor/theme';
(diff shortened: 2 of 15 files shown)


suite-native/atoms/src/Stack.tsx
```diff
@@ -27,30 +29,32 @@ const spacerStyle = prepareNativeStyle<SpacerStyleProps>((utils, { spacing, orie
...
+    },
+);
 
 export const VStack = Stack;
 export const HStack = (props: StackProps) => <Stack {...props} orientation="horizontal" />;
+
+const AnimatedStack = Animated.createAnimatedComponent(Stack);
+AnimatedStack.displayName = 'AnimatedStack';
+export const AnimatedVStack = AnimatedStack;
+export const AnimatedHStack = AnimatedStack;
```
yanascz: Is `orientation` missing here intentionally? 🤔

suite-native/intl/src/en.ts
```diff
@@ -90,6 +90,14 @@ export const en = {
                 enable: 'Enable',
             },
         },
+        firmwareUpdateAlert: {
+            title: 'New Trezor firmware version available. ',
```
yanascz: Redundant trailing space character.

Nodonisko: Fixed

suite-native/atoms/src/Stack.tsx
Nodonisko: Good catch, fixed

suite-common/message-system/src/messageSystemTypes.ts
```diff
@@ -25,6 +25,7 @@ export const Feature = {
     ethClaim: 'eth.staking.claim',
     firmwareRevisionCheck: 'security.firmware.check',
     firmwareHashCheck: 'security.firmware.hashCheck',
+    firmwareUpdate: 'device.firmware.update',
```
matejkriz, line 29: I'm thinking whether we should indicate somehow, that this feature flag is implemented on mobile only. My initial thought was using `mobile` prefix in the id, but since those features can be implemented later on the other platform, I'm not proposing that. We can use comments if needed. Feel free to keep it as it is.

suite-native/firmware/src/hooks/useIsFirmwareUpdateFeatureEnabled.tsx
```diff
@@ -0,0 +1,19 @@
...
+    MessageSystemRootState,
+} from '@suite-common/message-system';
+
+export const useIsFirmwareUpdateFeatureEnabled = () => {
+    const localFeatureFlag = useFeatureFlag(FeatureFlag.IsFirmwareUpdateEnabled);
+    const remoteFeatureFlag = useSelector((state: MessageSystemRootState) =>
+        selectIsFeatureEnabled(state, Feature.firmwareUpdate, true),
+    );
+
+    const isFirmwareUpdateEnabled = localFeatureFlag && remoteFeatureFlag;
```
matejkriz: This way it wouldn't be available on production for users at all, because `localFeatureFlag` has default value `isDevelopOrDebugEnv()`. And there is no way to enable it locally if it's remotely disabled. I guess OR makes more sense here.

```suggestion
    const isFirmwareUpdateEnabled = localFeatureFlag || remoteFeatureFlag;
```

matejkriz (commented):
Except the wrong condition in useIsFirmwareUpdateFeatureEnabled.tsx
 it works and looks good, including possibility to disable via message-system. It's nice that it's animated.

suite-native/firmware/src/hooks/useIsFirmwareUpdateFeatureEnabled.tsx
Nodonisko: My assumption was that we will set local feature flag to `true` since we want to include it in release. If there will be OR and it will be true remote FF won't work at all.

matejkriz: I see, but in that case I would propose to switch the local flag for production now. I originally assumed you'd remove the local flag completely... I'm a little worried about migration. Feature flags are persistent. So all the users of current production version now have it set to `false`.

Nodonisko:
After discussion with @matejkriz I removed local feature flag so we have only remote one. Just to be sure I retested remote one once more.

matejkriz (approved):
🎉 looks good, but still we should consider hiding the promotion banner from the dashboard at least for the first release to stay on a safer side.

STew790:
**QA OK**
No local ff present. Alert on Home present when FW can be updated.
image

**Info**
21.5.2 59e12ce

PR #16319, opened 2025-01-12
fix(mobile): downgrade reanimated to fix crash

### LINKED ISSUE #16168: NullPointerException: Attempt to invoke virtual method 'java.lang.Object java.lang.ref.WeakReference.get()' on a null o...
Sentry Issue: [SUITE-NATIVE-1DV](https://satoshilabs.sentry.io/issues/6126275711/?referrer=github_integration)

```
NullPointerException: Attempt to invoke virtual method 'java.lang.Object java.lang.ref.WeakReference.get()' on a null object reference
    at com.swmansion.reanimated.layoutReanimation.AnimationsManager.onViewCreate(AnimationsManager.java:105)
    at com.swmansion.reanimated.layoutReanimation.ReaLayoutAnimator.reset(ReaLayoutAnimator.java:64)
    at com.facebook.react.uimanager.NativeViewHierarchyManager.clearLayoutAnimation(NativeViewHierarchyManager.java:827)
    at com.facebook.react.uimanager.UIViewOperationQueue$1.run(UIViewOperationQueue.java:915)
    at com.facebook.react.uimanager.UIViewOperationQueue.flushPendingBatches(UIViewOperationQueue.java:992)
...
(19 additional frame(s) were not displayed)
```

Nodonisko:
After some investigation of #16168 I might have clue what's going on. There is bug (race condition probably or some GC magic) in Reanimated code that was caused by one of these PRs https://github.com/software-mansion/react-native-reanimated/pull/6789 or https://github.com/software-mansion/react-native-reanimated/pull/6530

They just moved some files around but I think it could caused race condition manifest or it caused to run GC in slightly different time that before which resulted in random crashed we experience. I am already working on fix, this is temp solution to move to version that doesn't include any of these changes.

## Related Issue

Resolve #16168

### DIFF
--- a/suite-native/accounts/package.json
+++ b/suite-native/accounts/package.json
@@ -40,7 +40,7 @@
         "react": "18.2.0",
         "react-hook-form": "^7.54.2",
         "react-native": "0.76.1",
-        "react-native-reanimated": "3.16.1",
+        "react-native-reanimated": "3.15.5",
         "react-redux": "8.0.7"
     }
 }
--- a/suite-native/alerts/package.json
+++ b/suite-native/alerts/package.json
@@ -17,6 +17,6 @@
         "jotai": "1.9.1",
         "react": "18.2.0",
         "react-native": "0.76.1",
-        "react-native-reanimated": "3.16.1"
+        "react-native-reanimated": "3.15.5"
     }
 }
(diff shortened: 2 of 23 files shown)


yanascz (approved):

Nodonisko:
We agreed with @matejkriz that we will try patch Reanimated with my fix instead.

matejkriz:
Hopefully fixed by https://github.com/trezor/trezor-suite/pull/16347, switching this to Draft because we don't want to merge it unless the patch will not work.

PR #16323, opened 2025-01-13
Feat(suite-native): add icon and regenerate them

Lemonexe:
## Description

- Fix `generateIcons.js` script and regenerate icons for mobile.
- Add `wrench` icon to mobile ([this figma](https://www.figma.com/design/eKie94qfryjTHEnR8jfmgk/Security-checks?node-id=4060-2927&t=mcAFWgpq7GU1EFka-4) requires it, all other icons in this figma are already present) 
- Some icons were automatically synced


suite-common/icons/generateIcons.js
```diff
@@ -1,7 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import chalk from 'chalk';
```
Lemonexe, line 3: node runs this script as CJS because that's how this package specifies, but crashed because of `import`. Node  automatically reruns it as ES, but that crashed at other parts  (`require('fs')`, `__dirname`).
Perhaps this used to work with node20? :thinking:
At any case it doesn't work now..
Changing to pure CJS fixes it (and it's consistent with package.json)

matejkriz:
There is one more step that is temporary needed if you do any native change (and regeneration of icons is native change) and that is bumping of `runtimeVersion`. Would you mind if I add here one more commit with that change? It will be hopefully soon not needed and replaced by automatic fingerprint check.

matejkriz (approved):
The icons don't seem broken and I can now use the wrench 🔧

PR #16328, opened 2025-01-13
New native onboarding flow initialized

### LINKED ISSUE #16231: Welcome Flow - New Flow Module
- There is a new  fresh module `suite-native/module-onboarding` created that will live next to the original onboarding module that will be renamed to `suite-native/module-onboarding-legacy`. (When the welcome flow will be completely done, the original flow module will be removed.
- There is a new button in DEV utils screen, from where the user can enter the new welcom flow. Make sure that this button is never visible in production version of the app!

### LINKED ISSUE #16232: Welcome Flow - Landing Screen Redesign
- [FIGMA ](https://www.figma.com/design/UdQpTfz2RhsoZzqLp9L8fU/Onboarding?node-id=1425-11477&t=mJ5RP6RDBFo5Otzs-4)
## Acceptance criteria 
As a new user opening the app for the first time, I would like to see the Preview screen where I can see quickly what the app helps me do.

## Dev notes
- this screen should be always dark (even if light mode is on), so make sure that the content of the screen does not reflect the theme settings. only thing that should be themed is the bottom button

PeKne:
## Description
- Resolves #16231 
- Resolves #16232
### Commits:
 - 828200493e77436fd82cb951d8c5a7399cd565ff 
   - old flow renamed to `module-onboarding-legacy` 
 - 2f25715ae4132b76bbc03a4abbc0bc2af7b077a0 
 	- a new flow stack navigator added and is accessible via DEV Utils screen for now
 	- Welcome Screen initialized

## Screenshots:

https://github.com/user-attachments/assets/0c273f64-f197-41a6-9605-209ec33942e4


suite-native/module-onboarding/package.json
```diff
@@ -10,24 +10,16 @@
...
-        "@suite-native/feature-flags": "workspace:*",
         "@suite-native/icons": "workspace:*",
         "@suite-native/intl": "workspace:*",
-        "@suite-native/link": "workspace:*",
         "@suite-native/navigation": "workspace:*",
-        "@suite-native/settings": "workspace:*",
-        "@suite-native/theme": "workspace:*",
+        "@trezor/env-utils": "workspace:*",
         "@trezor/styles": "workspace:*",
+        "expo-linear-gradient": "^14.0.1",
```
yanascz, line 21:
```suggestion
        "expo-linear-gradient": "14.0.1",
```

suite-native/module-onboarding/src/assets/welcomeScreenBackground.jpeg
yanascz, line 1: Does it really make sense to include the whole landscape image? I believe that using 1/3 or maybe 1/2 should be perfectly sufficient.

suite-native/module-onboarding/src/screens/WelcomeScreen.tsx
```diff
@@ -1,91 +1,96 @@
...
                     />
-                </Text>
+                    <LinearGradient
+                        colors={[transparentColor, '#000000']}
+                        style={applyStyle(gradientBackgroundBottomStyle)}
+                    />
+                </Box>
+            </ImageBackground>
+            <Box style={StyleSheet.absoluteFillObject}>
+                <Screen isScrollable={true} backgroundColor="transparent">
```
yanascz: Why scrollable? I don't think it makes sense here.

yanascz: What's the point of using `Screen`? If we wanna use it, I'd consider adding a property for background.

suite-native/module-onboarding/package.json
PeKne: What is wrong with this? We have it across whole repo. The yarn is configured in a way that it will always install the specific version so even if it `"^14.0.1"` it will always install `"14.0.1"`

suite-native/module-onboarding/src/assets/welcomeScreenBackground.jpeg
PeKne: Unfortunately it is, because we want to make sure that it will work as expected on tablets aswell. [See this design](https://www.figma.com/design/UdQpTfz2RhsoZzqLp9L8fU/Onboarding?node-id=1735-18291&t=EYe3FbtS1UOYkIEu-4). I want to avoid the case that it won't be wide enough for some devices and there will be black lines on the sides.

I have at least compressed the image so it has size of 500KB instead of 6,6MB.

suite-native/module-onboarding/package.json
yanascz: We've got both variants used and I believe it's better to be exact. Not a big deal, though.

PeKne: See this discussion https://github.com/trezor/trezor-suite/pull/15557#discussion_r1857962304

suite-native/module-onboarding/src/screens/WelcomeScreen.tsx
PeKne: IMO every Screen should be scrollable to make the screen work with a really small devices combined with system level UI enlarged fonts. See how the original welcome flow looks on small device with maxed font enlargement: 

https://github.com/user-attachments/assets/66b60703-19ea-4c37-8b52-05bc04cd8943

So I believe it is safe to make every screen scrollable for this case. Because as you can see in the video, if the screen would not be scrollable, the buttons would be hidden under the bottom edge of the screen.

suite-native/module-onboarding/src/assets/welcomeScreenBackground.jpeg
yanascz: Isn't that handled by the `resizeMode="cover"`? Anyway, if you crop it to square, I don't think you'll ever have any issue.

suite-native/module-onboarding/src/screens/WelcomeScreen.tsx
yanascz: I understand that screens should be scrollable in general. However, in this case it doesn't seem necessary and it looks weird on iOS. Just my two sats.

PeKne: Ok, i did not realize that you can move the content of the screen even if it does not overflow the height of the screen on iOS. isScrollable set to false in 8479dc9188

suite-native/module-onboarding/src/assets/welcomeScreenBackground.jpeg
PeKne: I tested that adn you are right that the image was too wide. Cropped to rectangle in 8479dc9188

yanascz: –237 KB, nice! 😎

yanascz (approved):

Issue #15054, opened 2024-10-24
Fix alignment of go back icon and toast messages


yanascz:
The current misalignment looks buggy:
![image](https://github.com/user-attachments/assets/2021ca27-2028-45e0-948f-17ea80576945)

It's not obvious what the correct fix would be, though.

shenkys:
[Design](https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Small-Things-%5BMOB%5D?node-id=1980-6730&t=667lYCuL6LnTXupy-11)

STew790:
**QA OK**
LGTM 

![Image](https://github.com/user-attachments/assets/10a43275-17bb-41f7-84d1-e500ed48f0bd)

**Info**

25.1.2 ac8cd6e

PR #16329, opened 2025-01-13
fix(suite-native): align toasts vertically with the close icon

yanascz:
## Screenshots:
image
image

### DIFF
--- a/suite-native/toasts/src/components/ToastRenderer.tsx
+++ b/suite-native/toasts/src/components/ToastRenderer.tsx
@@ -1,6 +1,7 @@
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
+
 import { useAtomValue } from 'jotai';
 
-import { useOfflineBannerAwareSafeAreaInsets } from '@suite-native/connection-status';
 import { Box, VStack } from '@suite-native/atoms';
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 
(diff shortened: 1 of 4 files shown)


suite-native/toasts/src/components/ToastRenderer.tsx
```diff
@@ -1,25 +1,27 @@
...
 const toastsContainerStyle = prepareNativeStyle<{ topSafeAreaInset: number }>(
-    (utils, { topSafeAreaInset }) => ({
+    ({ spacings }, { topSafeAreaInset }) => ({
         width: '100%',
         position: 'absolute',
         justifyContent: 'center',
-        marginTop: topSafeAreaInset,
-        paddingHorizontal: utils.spacings.sp16,
+        // top margin = screen top padding + screen header top padding
+        marginTop: Math.max(topSafeAreaInset, spacings.sp8) + spacings.sp8,
```
vytick, line 17: Can we make those constants? Something like `const SCREEN_HEADER_TOP_PADDING = spacings.sp8`

yanascz: That's kinda tricky because the first part is in the `Screen` component and the other one in `DeviceSwitch`/`ScreenSubHeader`. Sometimes it's used for padding, sometimes for margin. Would be great to have it in one place but I can't think of a nice way at the moment.

yanascz: We might introduce separate constants for `SCREEN_TOP_OFFSET` and `HEADER_TOP_OFFSET` or something similar if you think that would help.

vytick: Having what is there now, just extracting it to const would be more explanatory imo. But that is a nit

yanascz: That's not possible since `topSafeAreaInset` depends on what `useOfflineBannerAwareSafeAreaInsets` returns.

vytick: makes sense 👍

vytick (approved):
lgtm

matejkriz:
fixes [#15054](https://github.com/trezor/trezor-suite/issues/15054)

PR #16331, opened 2025-01-13
feat(suite-native): android edge to edge screen background

PeKne:
## Description
- Android now enforces edge to edge app backgroud for apps targeting SDK35 and later (https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge) similar as iOS have already some time
- library that makes our app background edge to edge across all the screens
	- Expo based lib used for that: https://expo.dev/blog/edge-to-edge-display-now-streamlined-for-android

### BEFORE:
Screenshot 2025-01-13 at 3 54 55 PM

### After

Screenshot 2025-01-13 at 3 25 47 PM


suite-native/navigation/src/components/Screen.tsx
```diff
@@ -141,12 +140,7 @@ export const Screen = ({
             })}
             testID={`@screen/${name}`}
         >
-            <StatusBar
-                barStyle={barStyle}
-                hidden={!hasStatusBar}
-                translucent={false}
-                backgroundColor={backgroundCSSColor}
-            />
+            <SystemBars style={systemBarsStyle} />
```
Nodonisko, line 137: What about that useEffect on line 122 here? Shoudn't we change it also there?

Nodonisko: Shoudn't this package replace `expo-navigation-bar` ?

PeKne: You are right, `expo-navigation-bar` is obsolete by the new lib. Also the useEffect for changing the StatusBar background is not needed any more. So it was removed and `expo-navigation-bar` removed from the project.

Already change and force pushed. Please check it again.

Nodonisko: Did you tested that it works correctly in dark theme? I can remember that without this we had issue when there were flashes of white during some screen transitions

PeKne: Yes, it works as expected 👍

Nodonisko (approved):

Issue #14718, opened 2024-10-04
Missing methods in Connect Explorer


hynek-jina:
It seems to me that https://connect.trezor.io/ is missing some of the methods from `@trezor/connect-web`. At least the ones prefixed with `blockchain`.

- blockchainDisconnect
- blockchainEstimateFee
- blockchainGetAccountBalanceHistory
- blockchainGetCurrentFiatRates
- blockchainGetFiatRatesForTimestamps
- blockchainGetTransactions
- blockchainSetCustomBackend
- blockchainSubscribe
- blockchainSubscribeFiatRates
- blockchainUnsubscribe
- blockchainUnsubscribeFiatRates

Other missing methods:
- cancel
- cardanoComposeTransaction
- checkFirmwareAuthenticity
- disableWebUSB
- dispose
- getSettings
- rebootToBootloader
- recoveryDevice
- renderWebUSBButton
- requestWebUSBDevice
- setBrightness
- setProxy
- uiResponse

karliatto:
Most of those methods if not all are not intended for 3rd party usage. Do we want to expose those in connect-explorer?

PR #16341, opened 2025-01-14
feat(connect-explorer): document cardanoComposeTransaction

martykan:
## Description

Added Typebox validation schema for `cardanoComposeTransaction`, use only weak assert for compatibility purposes. 
Added Connect Explorer documentation page for `cardanoComposeTransaction`.

## Related Issue

Part of #14718

### DIFF
--- a/packages/connect/src/api/cardano/api/cardanoComposeTransaction.ts
+++ b/packages/connect/src/api/cardano/api/cardanoComposeTransaction.ts
@@ -1,11 +1,13 @@
 import { trezorUtils, CoinSelectionError } from '@fivebinaries/coin-selection';
 
+import { AssertWeak } from '@trezor/schema-utils';
+
 import { AbstractMethod } from '../../../core/AbstractMethod';
-import { validateParams } from '../../common/paramsValidator';
 import { composeTxPlan } from '../cardanoUtils';
-import type {
-    CardanoComposeTransactionParams,
-    PrecomposedTransactionCardano,
+import {
+    CardanoComposeTransactionParamsSchema,
+    type CardanoComposeTransactionParams,
+    type PrecomposedTransactionCardano,
 } from '../../../types/api/cardanoComposeTransaction';
 
 export default class CardanoComposeTransaction extends AbstractMethod<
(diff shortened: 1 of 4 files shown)


packages/connect/src/types/api/cardanoComposeTransaction.ts
```diff
@@ -54,6 +72,120 @@ export type CardanoComposeTransactionParams = {
...
+        { $id: 'CardanoAddressParameters' },
+    ),
+    testnet: Type.Optional(Type.Boolean()),
+});
+
+// TS - Assert type of CardanoComposeTransactionParams is compatible with CardanoComposeTransactionParamsSchema
+const _params: CardanoComposeTransactionParamsSchema = {} as CardanoComposeTransactionParams;
+const _paramsOld: CardanoComposeTransactionParams = {} as CardanoComposeTransactionParamsSchema;
+// eslint-disable-next-line
+[_params, _paramsOld];
```
mroz22, line 187: what is this expression supposed to do? 🤔

packages/connect/src/api/cardano/api/cardanoComposeTransaction.ts
```diff
@@ -16,16 +18,8 @@ export default class CardanoComposeTransaction extends AbstractMethod<
...
-            { name: 'feeLevels', type: 'array' },
-            { name: 'outputs', type: 'array' },
-            { name: 'certificates', type: 'array', allowEmpty: true },
-            { name: 'withdrawals', type: 'array', allowEmpty: true },
-            { name: 'changeAddress', type: 'object', required: true },
-            { name: 'addressParameters', type: 'object', required: true },
-            { name: 'testnet', type: 'boolean' },
-        ]);
+        // TODO: weak assert for compatibility purposes (issue #10841)
+        AssertWeak(CardanoComposeTransactionParamsSchema, payload);
```
mroz22, line 22: just to remind myself - it only throws for required properties and possibly for some unknown errors of instance other than InvalidParameter

```
export function AssertWeak<T extends TSchema>(
    schema: T,
    value: unknown,
): asserts value is Static<T> {
    try {
        Assert(schema, value);
    } catch (e) {
        if (e instanceof InvalidParameter) {
            if (e.type === ValueErrorType.ObjectRequiredProperty) {
                // We consider this error to be serious
                throw e;
            }
            console.warn('Method params validation failed', e);
        } else {
            throw e;
        }
    }
}
```

I just trying to figure out what could go wrong if there was a bug in the schema

mroz22 (approved):
thanks for filling the gaps in our docs!

packages/connect/src/types/api/cardanoComposeTransaction.ts
martykan: TS will give an error if the schema is not matching with the the original type and vice versa. 
I think it's useful to have since the old type uses types from an external library and the schema does not, so we can use this to make sure they're in sync.

Issue #16340, opened 2025-01-14
Receive addresses in "Receive" section on BTC are showing 0 even though there's balance


STew790:
**Describe the bug**
Receive addresses on BTC are showing 0 even though there's a balance on them.

**Steps to reproduce:**
1. Connect a device with BTC funds
2. Go to "Receive"
3. The addresses should show balance but there's 0

**Info:**
 - Suite version: desktop 25.1.1 (11cab04dfc05ddb63cb6279bbc289994bf6c0583)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.1.1 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: Linux x86_64
 - Screen: 1920x1080
 - Device: Trezor T2B1 2.8.7 regular (revision 8a254aa8eae82f99630df63f40e4d290066a3efc)
 - Transport: BridgeTransport 2.0.33

bosomt:
**QA OK**

<img width="1148" alt="Image" src="https://github.com/user-attachments/assets/2845abe7-0949-4ff3-81a4-d466d3b3430f" />

**Info:**
 - Suite version: desktop 25.1.1 (49c29b152c18eaa5c1074c900a62a2a3362beadd)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.1.1 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.7 regular (revision 8a254aa8eae82f99630df63f40e4d290066a3efc)
 - Transport: BridgeTransport 2.0.33

PR #16346, opened 2025-01-14
fix: make types strict to prevent accidental deletion via unused types, the optionality is a trap, they are required

peter-sanderson:
Fix of https://github.com/trezor/trezor-suite/issues/16340

Followup of https://github.com/trezor/trezor-suite/pull/16344


packages/blockchain-link-types/src/blockfrost.ts
```diff
@@ -133,9 +133,9 @@ export interface Address {
     address: string;
     path: string;
     transfers: number;
-    balance?: string;
-    sent?: string;
-    received?: string;
+    balance: string;
+    sent: string;
+    received: string;
```
peter-sanderson, line 138: By making it required we avoid the trap of unused values. 

Those are used and required and IMO the only reason they were optional is some back compatibility or iterative composition of the type. 

The `"0"` value can be used safely I think as we turn it to 0 anyway in the UI.

martykan (approved):

PR #16356, opened 2025-01-14
chore(suite-native): make it simple to use Sentry locally in debug build

matejkriz:
## Description

Turning on Sentry in the local debug build is now complicated and one has to rewrite several places. This PR tries to simplify it.

From now on it should be enough to switch `EXPO_PUBLIC_IS_SENTRY_ENABLED` in `.env.development.local` and do a prebuild.


suite-native/app/src/App.tsx
```diff
@@ -94,4 +93,4 @@ const PureApp = () => (
     </GestureHandlerRootView>
 );
 
-export const App = isDebugEnv() ? PureApp : Sentry.wrap(PureApp);
+export const App = Sentry.wrap(PureApp);
```
matejkriz, line 96: I know that we added that condition because of some warning during local development, but now I can't see that warning any more and I guess it is because I removed the condition for Sentry.init and just playing with the `enabled` property.

suite-native/app/.env.development
```diff
@@ -1,2 +1,3 @@
 EXPO_PUBLIC_ENVIRONMENT=debug
 EXPO_PUBLIC_IS_ANALYTICS_LOGGER_ENABLED=false
+EXPO_PUBLIC_IS_SENTRY_ENABLED=false
```
Nodonisko: I would like to have it on for development build it helps catch lot of stuff that happens during QA etc. It was like that since beginning.

matejkriz: I see it is confusing, but it still works like that, because the condition for enabling is `enabled: !isDebugEnv() || process.env.EXPO_PUBLIC_IS_SENTRY_ENABLED === 'true',`. I'll add comment here.

matejkriz: I'm changing the name to `EXPO_PUBLIC_IS_SENTRY_ON_DEBUG_BUILD_ENABLED` instead of the comment to make it more clear.

matejkriz: f98d859b1134293234240ed9a66a50f53697fad6

Nodonisko (approved):

suite-native/app/src/App.tsx
vytick: Thought: 

When we enable `EXPO_PUBLIC_IS_SENTRY_ON_DEBUG_BUILD_ENABLED` in .env, we wont see anywhere that it is active. Do we want handle someone forgetting to switch it off and spamming sentry from develop?

What about to log a warning, that sentry is enabled, so devs can notice better in such case? Wdyt? 🤔

suite-native/app/src/SentryProvider.tsx
```diff
@@ -9,20 +9,21 @@ import { getEnv, isDebugEnv, isDevelopEnv } from '@suite-native/config';
...
-                    levels: ['error'],
-                }),
-            ],
-        });
-    }
+    Sentry.init({
+        dsn: 'https://d473f56df60c4974ae3f3ce00547c2a9@o117836.ingest.sentry.io/4504214699245568',
+        // Set tracesSampleRate to 1.0 to capture 100% of transactions for performance monitoring.
+        // We recommend adjusting this value in production.
+        tracesSampleRate: 0.5,
```
Lemonexe, line 16: If I understand, all events are sent to Sentry, but only 50% will bear traces?
That's how I [read the docs](https://docs.sentry.io/platforms/javascript/configuration/sampling/). And that sounds sensible. Just checking if it isn't "filter 50% traffic" :smile:

Lemonexe (approved):
LGTM, now it's nice and clear. [Tested](https://satoshilabs.sentry.io/issues/5878722226/?project=4504214699245568)  :heavy_check_mark:

suite-native/app/src/App.tsx
Lemonexe: If you edit the ENV in git tracked file it's impossible to forget about. Worst case you accidentally commit it and CR catches it.
Do you have a non-git-tracked env file in suite-native? :thinking: 
If yes, then I'd add a comment here: `# only modify it in *this file*, so you don't forget to turn it off`

matejkriz: `.env.development.local` is not tracked in git and I think it's good to use it. I see no big harm in sending more events from debug build to sentry. It's super easy to filter them out in Sentry. If we start sending hundreds of thousands of events, than we can think about some optimization 😉

suite-native/app/src/SentryProvider.tsx
matejkriz: Traces samples are not necessarily connected with the sentry error issues. This rate is for performance monitoring only like app start duration. I guess we can make the rate even lower soon.

Issue #16185, opened 2025-01-06
Add eth L2s under feature flag


vytick:
Enable Ethereum L2s - Base, Optimism and Arbitrum for testing purposes.

- [x] add feature flag for L2s
- [x] enable OP, ARB and BASE using that
- [x] enable graph for L2s

vytick:
@trezor/qa Please add to sanity tests, that Base, Arbitrum and Optimism are NOT accessible for coin enabling and for Portfolio tracker

STew790:
**QA OK**
I don't see L2s in enable coins or in portfolio when FF is disabled. FF is disabled by default and also I added this to our sanity tests.

**Info**
25.1.2 ac8cd6e

PR #16369, opened 2025-01-14
feat(graph): get graph events from txn history for arb, base, op

vytick:
Extract graph events from local txn history for Arbitrum, Base and Optimism

Part of #16185


Nodonisko (approved):
If it works, it's fine from my side.

PR #16377, opened 2025-01-15
Update trezor user env

mroz22:
## Related Issue

Resolve

### DIFF
--- a/docker/docker-compose.connect-popup-ci.yml
+++ b/docker/docker-compose.connect-popup-ci.yml
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   trezor-user-env-unix:
-    image: ghcr.io/trezor/trezor-user-env:bc4d202ed4be726145d61a6afad1c5754b3726d7
+    image: ghcr.io/trezor/trezor-user-env:b23cde1860e03c8bdaebbbc8fcbcd2cc9d6375d3
     environment:
       - SDL_VIDEODRIVER=dummy
       - XDG_RUNTIME_DIR=/var/tmp
--- a/docker/docker-compose.connect-popup-test.yml
+++ b/docker/docker-compose.connect-popup-test.yml
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   trezor-user-env-unix:
-    image: ghcr.io/trezor/trezor-user-env:bc4d202ed4be726145d61a6afad1c5754b3726d7
+    image: ghcr.io/trezor/trezor-user-env:b23cde1860e03c8bdaebbbc8fcbcd2cc9d6375d3
     environment:
       - DISPLAY=$DISPLAY
       - QT_X11_NO_MITSHM=1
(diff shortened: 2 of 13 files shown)


vdovhanych (approved):

marekrjpolak (approved):

marekrjpolak:
@mroz22 maybe also `setBusy.test.ts` can be reenabled now?

mroz22:
oh, I will check on it. this is anyway failing suspiciously on some desktop stuff 🤔

marekrjpolak:
Strange. Are these errors from userenv?
```
[4617:0115/114615.432359:ERROR:viz_main_impl.cc(166)] Exiting GPU process due to errors during initialization
[4577:0115/114615.723472:ERROR:bus.cc(407)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[4820:0115/114619.971991:ERROR:ffmpeg_common.cc(965)] Unsupported pixel format: -1
```

martykan:
Those are some weird Electron errors, but I think they don't affect anything, you can see them even in passing tests

mroz22:
@Vere-Grey @HajekOndrej guys can you help with desktop tests? we updated trezor-user-env that has a more recent version of trezorlib and it has undergone some bigger changes. It feels like, that under the hood, it behaves differently in certain situations. Maybe you are sending a "pressYes" command where it is no longer expected, or something like this

Vere-Grey:
Sure, I will get to it this evening.

Vere-Grey:
@mroz22 Nope, no extra sendYes or what so ever.
The problem is exactly same for both web and desktop. Identical issues.
Ignore the test send-doge. That is flaky issue that was already solved and is not relevant to this PR

The rest of the fails happens only in test that run nonT3T1 models and use setupEmu()
It probably get the trezor user env into some wrong state. There are plenty of errors in log -> see attached bellow.
The problem is that these action timeout after 3 minutes of waiting for response:

9x stopBridge 
1x setupEmu with defaults
1x setupEmu {"mnemonic":"nasty answer gentle inform unaware abandon regret supreme dragon gravity behind lava dose pilot garden into dynamic outer hard speed luxury run truly armed","pin":"1"}

The stop bridge is the first action a test does. it connects to trezor env. And cleans the state from potencial previous test by stoping Bridge and then emulator.
So all those fails happen in setup phase. When browser or electron is not yet started. The test just communicate with trezor env a set it up.

These step that fail are not unique. They are used by literally 90% of our tests. But the other test that dont fail on these methods either:
- use model T3T1 
- or use different models BUT dont call setupEmu

It might be that it is not possible to recover from the asumed faulty state. Because once the test start failing there is not passed test afterwards on that group run (ignore the middle one, that is the doge-send, not relevant fail)
image


Sum up: IMHO Trezor env gets to some faulty state and do not respond to test calls (normal setup and teardown actions)

Vere-Grey:
The bridge log I promised
[too-many-errors-tenv-emulator-bridge-debugging.log.txt](https://github.com/user-attachments/files/18454708/too-many-errors-tenv-emulator-bridge-debugging.log.txt)

mroz22:
nic it is passing after recent fixes by @grdddj

Issue #16273, opened 2025-01-08
Introduce FW revision check to Suite lite


MiroslavProchazka:
Similarly to Desktop app, we need to add implementation of FW Revision check also for Mobile app. 

Specs here (tbd)
Design + User flow [here](https://www.figma.com/design/eKie94qfryjTHEnR8jfmgk/Security-checks?node-id=4035-4718&t=Cm6LqpKUHUkOVVJF-1)

To Do:

- [x] #16379 must be finished first
- [x] Device compromised modal
- [x] Enable/Disable check in Device Settings
- [x] Offline status
- [x] Disable the generation of a receive address
- [x] In-app banner – TBD Should it be displayed on top of everything? Which component should we use?

Lemonexe:
#16448 is QA NOK, need to investigate before closing this issue.

EDIT: QA OK

Lemonexe:
All subissues are QA OK, so moving to approved.
Other issues were detected by @STew790, but not _directly_ related to the scope of this PR - will be separate issues.

PR #16386, opened 2025-01-15
Log firmware revision check to Sentry on mobile

### LINKED ISSUE #16379: Report FW revision check to Sentry in mobile
FW revision check has _already_ been performed _silently_ in Suite Lite, because Connect does it automatically.
But currently, the result is not processed in any way in Lite.
The first step is to report the result to Sentry very similarly as on Desktop (before we implement the UI in #16273)

### QA
- Connect a device with not yet released FW (such as 2.8.7).
- You should observe a Sentry event similar [to this one](https://satoshilabs.sentry.io/issues/6221440781/?project=4504214699245568)

Lemonexe:
## Description

First step to implement FW revision check in Mobile - just log the result to Sentry in background, no UI.
:information_source:  FYI the check has been done silently for months in suite native, because Connect does it automatically. 

- prepare redux state and selectors
  - may seem superfluous now, but I'll build on it in [#16273](https://github.com/trezor/trezor-suite/issues/16273)
- add `useDeviceCompromised` hook that sends Sentry exception if the check failed
  - similar to [the desktop implementation](https://github.com/trezor/trezor-suite/blob/4d870e4e19d82588f193de10ee5dd7bca5caed36/packages/suite/src/components/suite/SecurityCheck/useReportDeviceCompromised.ts) but still simpler, because we didn't get to FW hash check yet

## Testing
- you may merge [#16356](https://github.com/trezor/trezor-suite/pull/16356) locally into this branch to easily turn on Sentry for testing
- start emulator with `1-main` or `2-main` _(those versions always fail the check)_
- you should see a new sentry issue [such as this one](https://satoshilabs.sentry.io/issues/6221440781/?project=4504214699245568)

## Related Issue

Resolve #16379

### DIFF
--- a/suite-native/app/src/hooks/useGlobalHooks.tsx
+++ b/suite-native/app/src/hooks/useGlobalHooks.tsx
@@ -1,4 +1,8 @@
-import { useDetectDeviceError, useHandleDeviceConnection } from '@suite-native/device';
+import {
+    useDetectDeviceError,
+    useHandleDeviceConnection,
+    useReportDeviceCompromised,
+} from '@suite-native/device';
 import { useConnectPopupNavigation } from '@suite-native/module-connect-popup';
 
 import { useCoinEnablingInitialCheck } from './useCoinEnablingInitialCheck';
@@ -14,4 +18,5 @@ export const useGlobalHooks = () => {
 
     useHandleDeviceConnection();
     useDetectDeviceError();
+    useReportDeviceCompromised();
 };
--- a/suite-native/device/package.json
+++ b/suite-native/device/package.json
@@ -15,6 +15,7 @@
         "@reduxjs/toolkit": "1.9.5",
         "@sentry/react-native": "6.5.0",
         "@suite-common/redux-utils": "workspace:*",
+        "@suite-common/suite-utils": "workspace:*",
         "@suite-common/wallet-core": "workspace:*",
         "@suite-common/wallet-utils": "workspace:*",
         "@suite-native/alerts": "workspace:*",
--- a/suite-native/device/package.json
+++ b/suite-native/device/package.json
@@ -30,6 +30,7 @@
         "@suite-native/navigation": "workspace:*",
         "@suite-native/settings": "workspace:*",
         "@trezor/connect": "workspace:*",
+        "@trezor/device-utils": "workspace:*",
         "@trezor/styles": "workspace:*",
         "lottie-react-native": "^7.1.0",
         "react": "18.2.0",
--- a/suite-native/device/src/hooks/useReportDeviceCompromised.ts
+++ b/suite-native/device/src/hooks/useReportDeviceCompromised.ts
@@ -1,0 +1,38 @@
+import { useSelector } from 'react-redux';
+import { useEffect, useMemo } from 'react';
+
+import * as Sentry from '@sentry/react-native';
+
+import { getFirmwareVersion } from '@trezor/device-utils';
+import { selectSelectedDevice } from '@suite-common/wallet-core';
+
+import { selectFirmwareRevisionCheckError } from '../selectors';
+
+const reportCheckFail = (checkType: 'Firmware revision', contextData: any) => {
+    Sentry.captureException(`${checkType} check failed! ${JSON.stringify(contextData)}`);
+};
+
+const useCommonData = () => {
+    const device = useSelector(selectSelectedDevice);
+    const model = device?.features?.internal_model;
+    const revision = device?.features?.revision;
+    const version = getFirmwareVersion(device);
+    const vendor = device?.features?.fw_vendor;
+
+    return useMemo(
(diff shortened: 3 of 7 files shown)


suite-native/app/package.json
```diff
@@ -79,6 +79,7 @@
         "@suite-native/toasts": "workspace:*",
         "@suite-native/transactions": "workspace:*",
         "@trezor/connect": "workspace:*",
+        "@trezor/device-utils": "workspace:*",
```
Lemonexe, line 82: I checked that it's already used elsewhere in suite native, not new dependency

suite-native/settings/src/firmwareAuthenticitySelectors.ts
Lemonexe, line 1: there will be more selectors in upcoming work, so I created a separate file for them

suite-native/app/src/hooks/useReportDeviceCompromised.ts
```diff
@@ -0,0 +1,37 @@
...
+export const useReportDeviceCompromised = () => {
+    const commonData = useCommonData();
+
+    const revisionCheckError = useSelector(selectFirmwareRevisionCheckError);
+
+    useEffect(() => {
+        if (revisionCheckError !== null) {
+            reportCheckFail('Firmware revision', { ...commonData, revisionCheckError });
+        }
+    }, [commonData, revisionCheckError]);
```
Lemonexe: sends report everytime this information changes, which is either when you reconnect the device, connect different device, or update FW

suite-native/settings/src/firmwareAuthenticitySelectors.ts
```diff
@@ -0,0 +1,20 @@
...
+import { isDeviceAcquired } from '@suite-common/suite-utils';
+
+import { SettingsSliceRootState } from './settingsSlice';
+
+type FwAuthenticityCheckState = DeviceRootState & SettingsSliceRootState;
+
+/**
+ * Get firmware revision check error, or null if check was successful / skipped.
+ */
+export const selectFirmwareRevisionCheckError = (state: FwAuthenticityCheckState) => {
```
matejkriz: nit: I would not expect this hook in `settings` folder.  I think it fits better to `device` folder since the selector is reading from device state. And if there will be more utilities needed for security checks, I can imagine new package(s) for it.

suite-native/app/src/hooks/useReportDeviceCompromised.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+const useCommonData = () => {
+    const device = useSelector(selectSelectedDevice);
+    const model = device?.features?.internal_model;
+    const revision = device?.features?.revision;
+    const version = getFirmwareVersion(device);
+    const vendor = device?.features?.fw_vendor;
+
+    return useMemo(
```
matejkriz: There is no calculation, so what is the benefit of `useMemo` here?

matejkriz: I guess I get it, it is supposed to prevent dependencies change for the useEffect reporting check fail below. Ok then.

suite-native/app/src/App.tsx
```diff
@@ -47,6 +48,7 @@ const AppComponent = () => {
     const isConnectInitialized = useSelector(selectIsConnectInitialized);
 
     useReportAppInitToAnalytics(APP_STARTED_TIMESTAMP);
+    useReportDeviceCompromised();
```
matejkriz: I would try to move this hook here https://github.com/trezor/trezor-suite/blob/feat/fw-check-sentry-mobile/suite-native/module-authorize-device/src/navigation/AuthorizeDeviceStackNavigator.tsx/#L33-L34

Because if it's in App, it is active when you connect the device on Welcome screen (even before you confirm analytics) so we might miss some events. 

In reality, I guess it doesn't matter that much, but I think that it belongs there.

suite-native/settings/src/firmwareAuthenticitySelectors.ts
Lemonexe: Makes much more sense, will do :+1: The only thing it reads from settings is `isFirmwareRevisionCheckDisabled`, except it doesn't - this selector does not even use it, I had it prepared for upcoming work, did not realize it's unused _in this PR. :see_no_evil:

suite-native/app/src/App.tsx
Lemonexe: Done, and verified it still works as expected.
I force pushed, because together with the other comment, most of this PR is changed..

suite-native/settings/src/firmwareAuthenticitySelectors.ts
Lemonexe: Done

matejkriz (approved):
🎉

PR #16387, opened 2025-01-15
chore(suite-native): introduce component tests

jbazant:
## Description

Reintroduction of [testing-library](https://testing-library.com/) to our code base.

- updated `test-utils` module
- updated `jest.config.native.js`
- 2 "example" component tests introduced


suite-native/atoms/jest.config.js
```diff
@@ -0,0 +1,4 @@
+const { ...baseConfig } = require('../../jest.config.native');
```
Nodonisko: We are trying to avoid unecessary boilerplate files for packages as much as possible so I would strongly prefer to just include config in package.json as it is for other packages aka `test:unit: "yarn g:jest -c ../../jest.config.native.js"`

jbazant: If `jest.config.js` is specified, IDE is able to pick it up and use when running test. It is possible to run tests from IDE even without the explicit config, but it is required to define configuration, which is IMO a bit inconvenient.

I can definitely remove it, just sharing my reasoning.

jbazant: Updated.

Nodonisko (approved):

PR #16392, opened 2025-01-15
fix(suite): fix migration test

STew790:
Fixing a database migration test thats failing

## Related Issue

Resolve

### DIFF
--- a/packages/suite-web/e2e/tests/suite/database-migration.test.ts
+++ b/packages/suite-web/e2e/tests/suite/database-migration.test.ts
@@ -117,6 +117,7 @@ describe('Database migration', () => {
         // TO:
         cy.visit(`${baseUrl}/${to}`);
         cy.getTestElement('@dashboard/graph', { timeout: 40000 }).should('be.visible');
+        cy.getTestElement('@account-menu/normal').click();
         cy.getTestElement('@account-menu/btc/normal/0').click();
         cy.getTestElement('@menu/switch-device').click();
         cy.getTestElement('@deviceStatus-disconnected');


packages/suite-web/e2e/tests/suite/database-migration.test.ts
```diff
@@ -117,6 +117,7 @@ describe('Database migration', () => {
         // TO:
         cy.visit(`${baseUrl}/${to}`);
         cy.getTestElement('@dashboard/graph', { timeout: 40000 }).should('be.visible');
+        cy.getTestElement('@account-menu/normal').click();
```
mroz22, line 120: it indicates there has been a change in the UX. Is it a desirable change?

STew790: for some reason the accounts are collapsed and not expanded after the switch from the old to latest version of suite but I was not able to reproduce it manually and this seemed like "ok" workaround, so the test keeps working but correct me if I'm wrong

HajekOndrej: I don't know if this is a good idea, we should maybe try to find out why is this needed now. What has changed?

STew790: I have really no clue, I tried to reproduce this but I just wasn't able to. When I do the same steps manually, the accounts are expanded correctly :man_shrugging:

Vere-Grey: I recall similar problem in playwright tests. and some PR that was changing this behaviour.
But in the end we removed that test for some instability and duplicity :(

HajekOndrej (approved):

PR #16406, opened 2025-01-16
chore(connect): remove commented out fixtures for decred multisig

mroz22:
I don't think we are ever going to need these decred fixtures.


marekrjpolak (approved):
Afuera!

PR #16417, opened 2025-01-16
fix(suite): Fix firmware update progress bar

### LINKED ISSUE #16215: The loading bar gets stuck at 99% even after the firmware installation is complete
Image

jvaclavik:
## Description

- also component and color refactoring included 

## Related Issue

Resolve #16215

## Screenshots:

before

image

after
Screenshot 2025-01-17 at 13 57 54

Screenshot 2025-01-17 at 13 58 10

Screenshot 2025-01-17 at 13 58 54

### DIFF
--- a/packages/components/src/components/loaders/ProgressBar/ProgressBar.tsx
+++ b/packages/components/src/components/loaders/ProgressBar/ProgressBar.tsx
@@ -1,7 +1,9 @@
-import styled from 'styled-components';
+import styled, { useTheme } from 'styled-components';
 
-const Wrapper = styled.div`
-    background: ${({ theme }) => theme.backgroundNeutralSubdued};
+import { CSSColor } from '@trezor/theme';
+
+const Wrapper = styled.div<{ $color: CSSColor }>`
+    background: ${({ $color }) => $color};
     width: 100%;
     overflow: hidden;
 `;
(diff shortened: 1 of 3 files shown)


packages/suite/src/components/firmware/FirmwareProgressBar.tsx
```diff
@@ -68,20 +29,29 @@ export const FirmwareProgressBar = () => {
...
-            <StyledProgressBar value={progress} />
-            <Percentage>
-                {isDone ? (
-                    <Icon name="check" color={theme.legacy.TYPE_GREEN} size={24} />
-                ) : (
-                    `${progress} %`
-                )}
-            </Percentage>
-        </Wrapper>
+        <Box width="100%">
```
jvaclavik, line 32: I'm not happy about this one (it should be part of `Column`), but I would need to update `alignItems:stretch` in `OnboardingStepBox` which could lead many issues across the app. Not brave enough right now :D let's leave it for broader refactoring.

packages/suite/src/components/firmware/FirmwareProgressBar.tsx
```diff
@@ -1,54 +1,15 @@
...
-        border-radius: ${borders.radii.xxs};
-    }
-`;
 const Percentage = styled.div`
-    display: flex;
-    margin-left: ${spacingsPx.sm};
-    font-weight: ${variables.FONT_WEIGHT.DEMI_BOLD};
     font-variant-numeric: tabular-nums;
-    height: ${spacingsPx.xl};
+    width: 30px;
```
jvaclavik, line 12: It's intended not have fixed width, because otherwise progress bar is moving according to percentage value or checkmark

packages/components/src/components/loaders/ProgressBar/ProgressBar.tsx
```diff
@@ -1,35 +1,58 @@
...
+    $color: CSSColor;
 };
 
 const Value = styled.div<ValueProps>`
-    background: ${({ theme }) => theme.iconPrimaryDefault};
+    background: ${({ $color }) => $color};
     height: 5px;
     max-width: 100%;
-    width: 1%;
-    transform: ${({ $max, $value }) => `scaleX(${(100 / $max) * $value})`};
```
jvaclavik, line 19: There is a rounding bug in cases of long progressbar, so I rewrote it with width. Looks like everything is still working.

packages/components/src/utils/frameProps.tsx
```diff
@@ -80,7 +80,7 @@ export type FramePropsKeys = keyof FrameProps;
 type TransientFrameProps = TransientProps<FrameProps>;
 
 const getValueWithUnit = (value: string | number) =>
-    typeof value === 'string' ? value : `${value}px`;
+    typeof value === 'number' ? `${value}px` : value;
```
jvaclavik, line 83: Actually this makes sense to have it this way IMHO

packages/components/src/components/loaders/ProgressBar/ProgressBar.tsx
```diff
@@ -1,35 +1,58 @@
...
-    transform-origin: left;
-    transition: transform 0.5s;
+    width: ${({ $max, $value }) => `calc((100% / ${$max}) * ${$value})`};
```
seibei-iguchi, line 21: I've changed this because it's better to use transform than to animate width:) It should be the same, do you think it bug is caused by this?

jvaclavik: Yes, I'm pretty sure it was, now it's working fine

packages/components/src/components/loaders/ProgressBar/ProgressBar.tsx
seibei-iguchi: Can't we just use Math.round?:)

seibei-iguchi (approved):

bosomt:
**QA OK**

image

image


**Info:**
 - Suite version: web 25.1.2 (2db40a5dfed7c8743da2ff2cce78e86e557a0aaa)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/132.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1920x1080
 - Transport: WebUsbTransport

PR #16421, opened 2025-01-16
Feat/connect fw update version check

szymonlesisz:
## Description

1. added `TrezorConnect.updateFirmware` with fw version check and enhanced response
   returns
   ```
   check: FirmwareUpdateResponse['check']; // firmwareCheck result
   versionCheck: boolean; // version check assert:  installedVersion == binaryVersion == releaseVersion
   bootloaderVersion: VersionArray;
   installedVersion: VersionArray;
   binaryVersion: VersionArray;
   releaseVersion?: VersionArray;
   ```
   
1. added unit tests for `onCallFirmwareUpdate` module.
   Setup could be used to test with real FW binaries (tests are slower), could be used in some nightly workflow?

1. added `TrezorConnect.on(UI.FIRMWARE_VERSION_CHANGED, { device, oldVersion, newVersion });`
  emitted when previous Features !== current  Features
  
1. fixed `filterSafeListByBootloader` util
   if bootloader was higher than any known version it was oferring first release without `bootloader_version` field (some prehistoric)
   
   since `filterSafeListByBootloader` starts working properly i needed to adjust few mocks and prefill expected bootloader specific fields. suite tests were failing because of unsupported bootloader
   
1. fixed `firmwareUpdate` for models without `firmwareRelease`

### DIFF
--- a/packages/connect/src/types/api/firmwareUpdate.ts
+++ b/packages/connect/src/types/api/firmwareUpdate.ts
@@ -1,6 +1,7 @@
 import { Static, Type } from '@trezor/schema-utils';
 
 import type { Params, Response } from '../params';
+import type { VersionArray } from '../firmware';
 
 export type FirmwareUpdate = Static<typeof FirmwareUpdate>;
 export const FirmwareUpdate = Type.Union([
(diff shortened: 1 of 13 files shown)


packages/connect/src/types/api/firmwareUpdate.ts
```diff
@@ -15,17 +16,29 @@ export const FirmwareUpdate = Type.Union([
...
-      };
+type FirmwareUpdateDetails = {
+    versionCheck: boolean;
+    bootloaderVersion: VersionArray;
+    installedVersion: VersionArray;
+    latestVersion: VersionArray;
+    binaryVersion: VersionArray;
+    releaseVersion?: VersionArray;
+    isLatest: boolean;
+};
```
mroz22: I wouldn't mind adding some comments here explaining what these fields mean. Actually, only one aspect is not obvious at first glance - is the <something>-version that has just been installed, or that had been in the device before installation or that was available for installation?

mroz22:
nice coverage 
image

packages/connect/src/types/api/firmwareUpdate.ts
szymonlesisz: good point, https://github.com/trezor/trezor-suite/pull/16421/commits/710fc27d1f9c7d872a2d4d032be9fa78dec226c1

mroz22 (approved):
lgtm

PR #16424, opened 2025-01-16
chore(suite-native): move SettingsSection to settings

Lemonexe:
## Description

- Move `SettingsSection` components from `module-settings` to `settings`
- Because they will also be needed in `module-device-settings`
  - upcoming work in [#16273,](https://github.com/trezor/trezor-suite/issues/16273) specifically see [this frame in figma](https://www.figma.com/design/eKie94qfryjTHEnR8jfmgk/Security-checks?node-id=4060-2927&t=rRZuJhPN1Kok8PaO-4)

## Related Issue

Part of #16273

## Screenshots:

Settings sections and their items look same as before 
image


suite-native/settings/src/index.ts
```diff
@@ -1 +1,5 @@
 export * from './settingsSlice';
+export * from './components/SettingsSection';
+export * from './components/SettingsSectionItem';
+export * from './components/SettingsSectionItemIcon';
```
PeKne: nit: is it necessary to export `SettingsSectionItemText` and `SettingsSectionItemIcon` if it is not imported anywhere directly outside of this package?

PeKne (approved):
LGTM, only one nit to consider.

suite-native/settings/src/index.ts
Lemonexe: Good point. I noticed they are not used directly, but wasn't sure if _they are meant to be used alone_, or if they just _haven't been used yet_.
I'll remove that, after all it is possible to import from `@suite-native/settings/path` :shrug:

PeKne: thanks

PR #16437, opened 2025-01-17
fix(suite): fix draggable window header

jvaclavik:
## Related Issue

Resolve 

## Screenshots:

Red draggable zone is not visible anymore.

image

image

image

image

### DIFF
--- a/packages/suite/src/components/suite/TrafficLightOffset.tsx
+++ b/packages/suite/src/components/suite/TrafficLightOffset.tsx
@@ -13,20 +13,17 @@ type Props = {
 
 // See: https://github.com/electron/electron/issues/5678
 // Visible all the time in the app
-const ThinFixForNotBeingAbleToDragWindow = styled.div`
+
+const FixForNotBeingAbleToDragWindow = styled.div`
+    background: rgba(255, 0, 0, 0.1);
     -webkit-app-region: drag;
-    height: 16px;
+    pointer-events: none;
+    height: 64px;
     position: fixed;
     z-index: ${zIndices.windowControls};
     top: 0;
     left: 0;
     width: 100%;
-`;
-
-// Visible below content (but visible in the sidebar)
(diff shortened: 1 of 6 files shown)


bosomt:
**QA OK**

_macOS_
 - Suite version: desktop 25.2.0 (16d98ba617027c76312aa5c37351321149ec7436)
 - is draggable everywhere except at ⚙️ icon , connected or disconnected device
 
_Win11_
- works as expected, in WIn env there is app bar that you always can grab and drag 

_Mint_
- same for Linux, it has dedicated app bar that can be grabbed and dragged anytime

packages/suite/src/components/suite/TrafficLightOffset.tsx
```diff
@@ -13,40 +13,39 @@ type Props = {
...
     height: 100%;
 `;
 
+export const TrafficLightDraggableWindowHeader = ({ children, isVisible = true }: Props) => {
+    const isMac = isMacOs();
+    const isDesktopApp = isDesktop();
+
+    if (!isVisible || !isMac || !isDesktopApp) return children;
+
+    return <FixForNotBeingAbleToDragWindow />;
```
vojtatranta, line 39: NIT (meaning nitpick, if won't added, it's okay): I would add a comment here explain of why this fix is needed in these conditions, what's the root of it.

vojtatranta:
NIT: Overall opinion: the problem with fixes like this is that these are dependent on multiple changes in the codebase that are unrelated. 
I can think of a test cast that would resolve it - you'd need to render the whole react tree and make a one test case with multiple `expect()` that would check that the styles and components are where they are.

I am not sure if the codebase is prepared for that. So I would at least add comments to the changes of why they are there.

vojtatranta (approved):

seibei-iguchi (approved):

PR #16442, opened 2025-01-17
fix(coinmarket): reflect account updates in coinmarket

### LINKED ISSUE #16430: Trading - transaction signing error for DEX swap
**Describe the bug**
Transaction signing error for DEX swap without approval - `Transaction signing error: INTERNAL_ERROR: nonce too low`

**Info:**
Suite version: web 25.1.0

**How to reproduce**
Steps to reproduce:

1. Go to Swap
2. Select USDT@ETH => ETH
3. Select DEX offer
4. Make approval transaction (if you already have one, revoke it first)
5. Wait until the approval transaction is confirmed
6. Confirm and send swap transaction
7. Error `Transaction signing error: INTERNAL_ERROR: nonce too low`  is thrown

FreeWall:
Reflect account updates in coinmarket so the nonce is always up to date.

## Related Issue

Resolve #16430

### DIFF
--- a/packages/suite/src/hooks/wallet/coinmarket/form/useCoinmarketExchangeForm.ts
+++ b/packages/suite/src/hooks/wallet/coinmarket/form/useCoinmarketExchangeForm.ts
@@ -628,6 +628,12 @@ export const useCoinmarketExchangeForm = ({
         }
     }, [defaultValues, values, removeDraft]);
 
+    useEffect(() => {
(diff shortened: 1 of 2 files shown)


adderpositive (approved):

tomasklim (approved):
Makes sense but it would be better to just keep account key for trading and always find correct account in the accounts object
[This duplicating and updating is road to hell](https://en.wikipedia.org/wiki/Single_source_of_truth)

PR #16443, opened 2025-01-17
feat(e2e): Migrated e2e tests for metadata providers api errors

HajekOndrej:
## Description
Migrated tests for metadata provider API errors

## Related Issue

Resolve


packages/suite-desktop-core/e2e/support/mocks/metadataProviderMock.ts
```diff
@@ -1,5 +1,6 @@
 import { Page } from '@playwright/test';
 
+import * as metadataUtils from '@trezor/suite/src/utils/suite/metadata';
```
Vere-Grey: let's do focus importing
``` suggestion
import { encrypt } from '@trezor/suite/src/utils/suite/metadata';
```

packages/suite-desktop-core/e2e/support/mocks/metadataProviderMock.ts
```diff
@@ -77,11 +78,30 @@ export class MetadataProviderMock {
     }
 
     @step()
-    stop() {
+    setNextResponse(response: Record<string, any>): void {
         if (!this.providerMock) {
             throw new Error('Provider mock not initialized');
```
Vere-Grey: Let's DRY this `if (!this.providerMock) ` block by introducing a getter that will do the assertion and throw :)

packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts
```diff
@@ -0,0 +1,161 @@
...
+            '/f7acc942eeb83921892a95085e409b3e6b5325db6400ae5d8de523a305291dca.mtdt',
+            fileContent,
+            aesKey,
+        );
+
+        await onboardingPage.completeOnboarding();
+        await dashboardPage.discoveryShouldFinish();
+
+        await settingsPage.navigateTo('application');
+        await settingsPage.metadataSwitch.click();
```
Vere-Grey, line 39: Potential to DRY these steps to before each.
not sure before which steps metadataProviderMock.setFileContent needs to be set tho ...

packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+        await metadataPage.passThroughInitMetadata(MetadataProvider.DROPBOX);
+
+        await page.getByTestId('@account-menu/btc/normal/0').click();
+
+        await metadataPage.accountLabel(AccountLabelId.BitcoinDefault1).click();
+        await metadataPage.editLabelButton(AccountLabelId.BitcoinDefault1).click();
+
+        // Simulated API responses for retries with malformed token must be supplied exactly at this point in the flow
```
Vere-Grey, line 48: So if we would set it in advance, the responses would be 'eaten' by something else i guess? And there 4 retries?
Just asking to verify  I understand what is happening

packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts
```diff
@@ -0,0 +1,161 @@
...
+            });
+        }
+
+        await metadataPage.metadataInput.fill('Kvooo');
+        await page.keyboard.press('Enter');
+
+        // Validate the error message in the toast notification
+        await expect(page.getByTestId('@toast/error')).toHaveText(
+            'Error: Failed to save labeling data: Error in call to API function "files/upload": The given OAuth 2 access token is malformed.',
+        );
```
Vere-Grey, line 65: Can we add a validation that the label has still the original text?
```
await expect(metadataPage.accountLabel(AccountLabelId.BitcoinDefault1)).toHaveText('<original name>')
````

File: packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts [hunk 6]
```diff
@@ -0,0 +1,161 @@
...
+            },
+        });
+
+        metadataProviderMock.setNextResponse({
+            status: 429,
+            body: 'Rate limited!',
+            headers: {
+                'Content-Type': 'text/plain; charset=utf-8',
+            },
+        });
```
Vere-Grey, line 121: I dont understand this part. Can we have a short call so I could get the context
I dont get the order of those two. Responses are served as FIFO
I found similar object in packages/e2e-utils/src/mocks/dropbox.ts:77: 
```
        // https://api.dropboxapi.com/2/users/get_current_account
        app.post('/2/users/get_current_account', (_req, res) => {
            const user = {
                name: {
                    given_name: 'dog',
                    surname: 'cat',
                    familiar_name: 'kitty-dog',
                    display_name: 'dog-cat',
                    abbreviated_name: 'DC',
                },
                email: 'some@mail.com',
                email_verified: true,
                profile_photo_url: 'foo',
                disabled: false,
                country: 'CZ',
                locale: 'en',
                referral_link: 'foo',
                is_paired: false,
            };
```

packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+        await page.getByTestId('@account-menu/btc/normal/0').click();
+        await metadataPage.editLabel(AccountLabelId.BitcoinDefault1, 'Kvooo');
+    });
```
Vere-Grey: Test does not have any verification of success of that the initial try failed and succeded on retry (not sure that info about retring is vibisble)

Vere-Grey:
```
await expect(metadataPage.accountLabel(AccountLabelId.BitcoinDefault1)).toHaveText('Kvooo')
```

packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    test('Incomplete data returned from provider', async ({
```
Vere-Grey, line 130: I dont understand the impact of 'incomplete data' on suite and what should happen and what shouldn't. 
Like should the rename fail? or is it just that the comunication is retried and does not fall apartt.

Maybe this is a candidate for the later refactoring. I have feeling that we miss some verifications here

Vere-Grey (changes requested):
I have suggested few improvements - up to discussion
The original tests missed some key verifications, I know we plan to do bigger refactoring but I would still proposed adding those few expects
And I need help understanding one or two things.

packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts
HajekOndrej: yes and yes

packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts
HajekOndrej: From my understanding of the original test, the ability to rename the label after the incomplete data is sent is the verification

Vere-Grey: So how would a fail look like?
It would not be possible to click on it? What would cause that? I am just worried that the click will be possible because that is the nature of that element but the action that will be trigger will fail, toast will appear but the test will be already passed because it happy it can click.

packages/suite-desktop-core/e2e/tests/metadata/dropbox-api-errors.test.ts
HajekOndrej: I thought I understood but then I studied the traces more and it is quite wonky. I think we should refactor the whole mock. I suggest we merge this and work on it separately. Let's chat about it tommorrow.

packages/suite-desktop-core/e2e/support/pageActions/metadataActions.ts
```diff
@@ -55,7 +56,7 @@ export class MetadataActions {
 
     @step()
     async clickAddLabelButton(accountId: string) {
-        await this.hoverAccountLabel(accountId);
+        await this.accountLabel(accountId).hover();
```
Vere-Grey, line 59: :+1:

Vere-Grey (approved):
:+1:

PR #16444, opened 2025-01-17
chore(suite-native): prepare one place for all global hooks

matejkriz:
## Description

We have multiple hooks that is supposed to be global and available everywhere, this is defining one place for those hooks.

We apparently have incorrectly placed hooks in the `AuthorizeDeviceStackNavigator`, namely `useReportDeviceConnectToAnalytics `and `useDetectDeviceError`. The `AuthorizeDeviceStackNavigator` only mounts when connecting a Trezor that hasn't been remembered (view-only on). If the Trezor with view-only enabled was connected, it did not mount, and those hooks remained inactive.

Notes for @trezor/qa 
The only change that this PR is supposed to cause is that if you connect Trezor, that has view-only already enabled and have some error, it should show the alert. Previously it was ignored. So it can theoretically cause more Unacquired errors.

It makes sense to test connecting Trezor to the app in different scenarios (view-only enabled/disabled, welcome flow finished/fresh app, open app by connecting the Trezor...)

Followup to move and fix reporting of device connected to analytics https://github.com/trezor/trezor-suite/issues/16446

## Screenshots

Trezor connected at the first screen of Welcome flow, warning visible only after Welcome flow:
https://github.com/user-attachments/assets/f3a2f7f4-1494-484b-b44c-ed5f597f30c0

### DIFF
--- a/suite-native/app/src/hooks/useGlobalHooks.tsx
+++ b/suite-native/app/src/hooks/useGlobalHooks.tsx
@@ -1,0 +1,14 @@
+import { useHandleDeviceConnection } from '@suite-native/device';
+import { useConnectPopupNavigation } from '@suite-native/module-connect-popup';
+
+import { useCoinEnablingInitialCheck } from './useCoinEnablingInitialCheck';
+
+/**
+ * @description This hook is used to initialize all the hooks,
+ * that are supposed to be active globally once the app is ready.
+ */
+export const useGlobalHooks = () => {
+    useHandleDeviceConnection();
+    useCoinEnablingInitialCheck();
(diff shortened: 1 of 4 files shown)


Nodonisko (approved):

PeKne (approved):

matejkriz:
I guess this PR caused this issue https://github.com/trezor/trezor-suite/issues/16494

STew790:
**QA OK**
Connected/disconnected multiple devices multiple times and all seems OK. This issue was fixed https://github.com/trezor/trezor-suite/issues/16494 .

**Info**
25.1.2 b6893f7

PR #16449, opened 2025-01-17
fix(suite-native): unify receive flow navigation for all use cases

yanascz:
Follow-up for https://github.com/trezor/trezor-suite/pull/16312. Once the receive flow is finished (i.e. receive address confirmed), user is returned to the initial screen upon tapping the close button.


suite-native/receive/src/screens/ReceiveAccountScreen.tsx
```diff
@@ -1,87 +1,88 @@
...
+const cardStyle = prepareNativeStyle(utils => ({
+    padding: utils.spacings.sp8,
+}));
 
-    const account = useSelector((state: AccountsRootState) =>
-        selectAccountByKey(state, accountKey),
-    );
-    const device = useSelector(selectSelectedDevice);
-    const hasReceiveButtonRequest = useSelector(hasReceiveAddressButtonRequest);
+const ReceiveAccountLoader = () => {
```
matejkriz, line 24: When can I see this loader?

matejkriz (approved):
Wow, this works really good including Portfolio Tracker and including enabling new coins and adding new accounts in Receive flow. Great job 👏

suite-native/receive/src/screens/ReceiveAccountScreen.tsx
yanascz: Not sure, might be dead code. Will discuss with the team and create a follow-up PR if necessary.

STew790:
**QA OK**
Nice, after confirming the address either in portfolio tracker or with device, I am returned to the homescreen when I click on "X". Much better for testing as well feelsgoodman.jpg.

**Info**
25.1.2 eb6c201

Issue #16447, opened 2025-01-17
Device Settings for FW revision check


Lemonexe:
Part of #16273: implement Device Settings for Firmware revision check in Suite Lite
[This figma frame](https://www.figma.com/design/eKie94qfryjTHEnR8jfmgk/Security-checks?node-id=4060-3223&t=fv9Xv7X06uaZhw6Q-4) captures the scope of this issue.

- add Device Checks section to Settings > Features screen
  - hide behind a new Feature Flag
- create Device Checks screen
- create confirmation modal to Turn off FW authenticity check

STew790:
**QA OK**
- settings hidden behind FF
- settings available after enabling FF
- confirmation modal present when turning off check

**Info**
25.1.3 e549011

PR #16454, opened 2025-01-17
feat(suite-native): add more params to IconListItem for reusability

Lemonexe:
## Description

Refactor `IconListItem` to make its usage more flexible
- allow different size of icons
- allow multiline content
- allow different flex align (useful for multiline content)

:information_source: Required for [this figma frame](https://www.figma.com/design/eKie94qfryjTHEnR8jfmgk/Security-checks?node-id=4116-2445&t=fv9Xv7X06uaZhw6Q-4)

## Related Issue

Related to #16447

## Screenshots:

Current usage of `IconListItem` is **unchanged**:
image
image

### DIFF
--- a/suite-native/atoms/src/IconListItem.tsx
+++ b/suite-native/atoms/src/IconListItem.tsx
@@ -1,6 +1,7 @@
 import { ReactNode } from 'react';
+import { FlexAlignType } from 'react-native';
 
-import { IconName } from '@suite-native/icons';
+import { IconName, IconSize } from '@suite-native/icons';
 import { Color } from '@trezor/theme';
 
 import { Box } from './Box';
@@ -8,7 +9,7 @@ import { OrderedListIcon } from './OrderedListIcon';
 import { HStack } from './Stack';
 import { Text } from './Text';
 
-type Variant = 'default' | 'red';
+type Variant = 'default' | 'red' | 'yellow';
 type IconColors = {
     iconColor: Color;
     iconBorderColor: Color;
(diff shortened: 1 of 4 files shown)


suite-native/atoms/src/OrderedListIcon.tsx
```diff
@@ -3,17 +3,27 @@ import { RequireExactlyOne } from 'type-fest';
...
-    (utils, { color, borderColor }) => ({
+const iconSizeToBackgroundSizeMap = {
+    medium: 32,
+    large: 40,
+    // sizes below are unclear, have not yet been used in a design
+    extraSmall: 32,
+    small: 32,
+    mediumLarge: 40,
+    extraLarge: 40,
+} as const satisfies Record<IconSize, NativeSpacings[keyof NativeSpacings]>;
```
Lemonexe: I am not exactly sure of this approach of mapping icon size to the size of bg.
I wanted to avoid passing another prop, because I don't see a usecase for specifying both independently, it would only create opportunity for mistakes.
What do you think about it?

`medium` values are the original hardcoded values
`large` values are in line [with this figma](https://www.figma.com/design/eKie94qfryjTHEnR8jfmgk/Security-checks?node-id=4116-2445&t=fv9Xv7X06uaZhw6Q-4)
The last 4 have not yet been tried, so they are temporary placeholders :shrug:

suite-native/atoms/src/IconListItem.tsx
```diff
@@ -26,23 +27,40 @@ const iconColorsMap = {
...
+            <Box flexShrink={1}>{children}</Box>
         </HStack>
     );
 };
+
+export const IconListTextItem = ({ children, ...rest }: IconListItemProps) => (
+    <IconListItem {...rest}>
+        <Text variant="hint">{children}</Text>
+    </IconListItem>
+);
```
Lemonexe, line 66: this maintains the simplified behavior of original `IconListItem`, but if you think it's unnecessary, I can remove it and add the `<Text variant="hint"></Text>` boilerplate to current usages.

suite-native/atoms/src/OrderedListIcon.tsx
PeKne: It seems that the size of the container should be always `iconSize` + `2 x 8 padding`. So IMO you could refactor it in the way that the background has no fixed dimensions but it has `padding: utils.spacings.sp8`.

Lemonexe: Thanks for the tip, that sounds much better. I'll try if it works well :slightly_smiling_face:

Lemonexe: Done in 
https://github.com/trezor/trezor-suite/pull/16454/commits/803f872a3ddad044814c2e79ead5982d3b7abfc2

Much simpler, and works the same :+1:  What do you think? :slightly_smiling_face:

PeKne (approved):
LGTM

suite-native/atoms/src/OrderedListIcon.tsx
PeKne: Looks good 👌

PR #16457, opened 2025-01-17
fix(mobile): balance mismatch

### LINKED ISSUE #12352: Balances mismatch

Nodonisko:
## Description

1. When user is not touching graph show same value as everywhere else
2. Reflect staking in device switcher

I was not able to hide percentages for when there is zero because there is lot of valid cases when there could be zero, so I decided to change it little bit so now it will not calculate agains first point of graph but agains first non-zero point of graph so now percentages kind of make sense even for "all" timeframe etc.

## Related Issue

Resolve #12352 

## Screenshots:

Simulator Screenshot - iPhone 16 Pro - 2025-01-17 at 19 15 59
Simulator Screenshot - iPhone 16 Pro - 2025-01-17 at 19 15 23

### DIFF
--- a/suite-native/accounts/src/selectors.ts
+++ b/suite-native/accounts/src/selectors.ts
@@ -24,10 +24,12 @@ import { Account, AccountKey, TokenInfoBranded } from '@suite-common/wallet-type
 import {
     getAccountFiatBalance,
     getAccountTotalStakingBalance,
+    getFiatRateKey,
     getFirstFreshAddress,
+    toFiatCurrency,
 } from '@suite-common/wallet-utils';
 import { SettingsSliceRootState, selectFiatCurrencyCode } from '@suite-native/settings';
-import { isCoinWithTokens } from '@suite-native/tokens';
+import { isCoinWithTokens, selectAccountTokenInfo } from '@suite-native/tokens';
 import type { StaticSessionId } from '@trezor/connect';
 import { createWeakMapSelector } from '@suite-common/redux-utils';
 import { doesCoinSupportStaking } from '@suite-native/staking';
@@ -43,7 +45,8 @@ export type NativeAccountsRootState = AccountsRootState &
     FiatRatesRootState &
     SettingsSliceRootState &
     DeviceRootState &
-    TokenDefinitionsRootState;
+    TokenDefinitionsRootState &
+    TransactionsRootState;
 
 const createMemoizedSelector = createWeakMapSelector.withTypes<NativeAccountsRootState>();
 
@@ -82,25 +85,54 @@ export const selectIsAccountAlreadyDiscovered = (
         ),
     );
 
-export const selectAccountFiatBalance = (state: NativeAccountsRootState, accountKey: string) => {
-    const fiatRates = selectCurrentFiatRates(state);
-    const account = selectAccountByKey(state, accountKey);
-    const localCurrency = selectFiatCurrencyCode(state);
-    const shouldIncludeStaking = !!account && doesCoinSupportStaking(account.symbol);
+export const selectAccountFiatBalance = createMemoizedSelector(
+    [
+        selectCurrentFiatRates,
+        selectAccountByKey,
+        selectFiatCurrencyCode,
+        (_, _accountKey: AccountKey, shouldIncludeStaking?: boolean) =>
+            shouldIncludeStaking ?? true,
+        (
+            _,
+            _accountKey: AccountKey,
+            _shouldIncludeStaking?: boolean,
+            shouldIncludeTokens?: boolean,
+        ) => shouldIncludeTokens ?? true,
+    ],
+    (fiatRates, account, localCurrency, shouldIncludeStaking, shouldIncludeTokens) => {
+        if (!account) {
+            return '0';
+        }
 
-    if (!account) {
-        return '0';
-    }
+        const totalBalance = getAccountFiatBalance({
+            account,
+            rates: fiatRates,
+            localCurrency,
+            shouldIncludeStaking,
+            shouldIncludeTokens,
+        });
 
-    const totalBalance = getAccountFiatBalance({
-        account,
-        rates: fiatRates,
-        localCurrency,
-        shouldIncludeStaking,
-    });
+        if (!totalBalance) {
+            return '0';
+        }
 
-    return totalBalance;
-};
+        return totalBalance;
+    },
+);
+
+export const selectAccountTokenFiatBalance = createMemoizedSelector(
+    [selectCurrentFiatRates, selectFiatCurrencyCode, selectAccountByKey, selectAccountTokenInfo],
+    (fiatRates, localCurrency, account, tokenInfo) => {
+        if (!account || !fiatRates || !tokenInfo) return '0';
+        const { contract, balance } = tokenInfo;
+        const fiatRateKey = getFiatRateKey(account.symbol, localCurrency, contract);
+        const rate = fiatRates[fiatRateKey]?.rate;
+
+        if (!rate || !balance) return '0';
+
+        return toFiatCurrency(balance, rate) ?? '0';
+    },
+);
 
 export const getAccountListSections = (
     account: Account,
--- a/suite-native/graph/src/components/GraphFiatBalance.tsx
+++ b/suite-native/graph/src/components/GraphFiatBalance.tsx
@@ -1,15 +1,17 @@
 import { Atom, useAtomValue } from 'jotai';
 
 import { FiatGraphPoint } from '@suite-common/graph';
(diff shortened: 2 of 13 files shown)


suite-native/accounts/src/selectors.ts
```diff
@@ -81,23 +84,54 @@ export const selectIsAccountAlreadyDiscovered = (
...
+    [selectCurrentFiatRates, selectFiatCurrencyCode, selectAccountByKey, selectAccountTokenInfo],
+    (fiatRates, localCurrency, account, tokenInfo) => {
+        if (!account || !fiatRates || !tokenInfo) return '0';
+        const { contract, balance } = tokenInfo;
+        const fiatRateKey = getFiatRateKey(account.symbol, localCurrency, contract);
+        const rate = fiatRates[fiatRateKey]?.rate;
+
+        if (!rate || !balance) return '0';
+
+        return toFiatCurrency(balance, rate) ?? '0';
```
matejkriz, line 133: Where is the `decimals` value reflected in the calculation?

suite-native/graph/src/components/GraphFiatBalance.tsx
```diff
@@ -34,7 +34,7 @@ const Skeleton = () => (
 
 const Balance = ({ selectedPointAtom }: BalanceProps) => {
     const point = useAtomValue(selectedPointAtom);
-    const fiatValue = String(point.value);
+    const fiatValue = point.valueLatestTotal ? point.valueLatestTotal : String(point.value);
```
matejkriz: nitpick:
same, but easier to read?
```suggestion
    const fiatValue = point.valueLatestTotal || String(point.value);
```

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -15,11 +15,7 @@ type WalletItemProps = {
 export const WalletItem = ({ onPress, deviceState, isSelectable = true }: WalletItemProps) => {
     const device = useSelector((state: any) => selectDeviceByState(state, deviceState));
     const selectedDevice = useSelector(selectSelectedDevice);
-    const fiatBalance = useSelector((state: any) =>
-        device?.state?.staticSessionId
-            ? String(selectDeviceTotalFiatBalanceNative(state, device?.state?.staticSessionId))
-            : undefined,
-    );
+    const fiatBalance = useSelector(selectSelectedDeviceTotalFiatBalance);
```
matejkriz: This is wrong, you can't show the fiat balance of the selected device for all devices... You have to select it by the staticSessionId

matejkriz (changes requested):

suite-native/graph/src/components/GraphFiatBalance.tsx
```diff
@@ -34,7 +34,7 @@ const Skeleton = () => (
 
 const Balance = ({ selectedPointAtom }: BalanceProps) => {
```
matejkriz: Would it be possible to show the `valueLatestTotal` even while the graph is loading if we know the value already? It doesn't make sense to hide the current balance by switching graph time frame. And would be awesome to show it sooner even before the graph is loaded.

suite-native/accounts/src/selectors.ts
Nodonisko: hopefully fixed

suite-native/device-manager/src/components/WalletItem.tsx
Nodonisko: good catch, reverted

suite-native/graph/src/components/GraphFiatBalance.tsx
Nodonisko: Yes, but it required me to rework this whole PR again :D

suite-native/accounts/src/selectors.ts
```diff
@@ -82,25 +85,54 @@ export const selectIsAccountAlreadyDiscovered = (
...
+    [selectCurrentFiatRates, selectFiatCurrencyCode, selectAccountByKey, selectAccountTokenInfo],
+    (fiatRates, localCurrency, account, tokenInfo) => {
+        if (!account || !fiatRates || !tokenInfo) return '0';
+        const { contract, balance, decimals } = tokenInfo;
+        const fiatRateKey = getFiatRateKey(account.symbol, localCurrency, contract);
+        const rate = fiatRates[fiatRateKey]?.rate;
+
+        if (!rate || !balance) return '0';
+
+        return toFiatCurrency(balance, rate, decimals) ?? '0';
```
matejkriz: No, these are completely different decimals with different meaning! The toFiatCurrency utility has a decimals parameter, which specifies how many decimal places the number should be rounded to. However, the decimals from tokenInfo refer to how divisible a token can be, so how to get the base unit. As an example: BTC has decimals==8, because 1 BTC has 100 000 000 sats. We get the value in sats but fiat rate is in BTC. Same it is with tokens.

suite-native/graph/src/hooks.ts
```diff
@@ -220,3 +222,68 @@ export const useGraphForAllDeviceAccounts = ({ fiatCurrency }: CommonUseGraphPar
...
+
+    const lastPoint: TGraphPoint | undefined = graphPoints[graphPoints.length - 1];
+    const referencePoint: TGraphPoint | undefined = useMemo(
+        () => graphPoints.find(point => point.value > 0) ?? graphPoints[0],
+        [graphPoints],
+    );
+
+    useEffect(
+        () => () => {
+            // we should reset everything to default on unmount otherwise it will broke loading state when navigating to same or another account
```
matejkriz, line 253: Is it really unmounting if we navigate to another account?

Nodonisko: yes, you leave screen so it will be unmounted and reseted but only for account detail not for dashboard but on dashboard we don't need it anyway... as I said I am not happy with current state because it's very glued up together but some major refactor would take few more days 😢

matejkriz:
Instead of $0, the amount is hidden at the beginning of the graph, check the video
https://github.com/user-attachments/assets/4dd18ac2-4862-4d96-9b62-07dfdaba7247

Nodonisko:
Good catch, that's interesting edge case, let me take a look

Nodonisko:
@matejkriz Not showing zero issue solved in [`eebbc4c` (#16457)](https://github.com/trezor/trezor-suite/pull/16457/commits/eebbc4cad85e51cf014922541573c03bb6794a12)

suite-native/graph/src/hooks.ts
matejkriz: ok, thanks

suite-native/accounts/src/selectors.ts
Nodonisko: I removed decimals here, and I tried implement `convertTokenValueToDecimal` which should do exactly what you say, but it will broke completely and all values will be like `0.000000000123`

Screenshot 2025-01-21 at 12 00 05

If do not handle it at all it seems to work without any issues

Nodonisko:
Fixed issue where Solana tokens with 0 values showed inifite loading [`5b7dcee` (#16457)](https://github.com/trezor/trezor-suite/pull/16457/commits/5b7dceee589dcbcdd26275c6ad38920de3496e40)

suite-native/accounts/src/selectors.ts
matejkriz: yes, confirmed, that the balance is already converted, I just still didn't find the place where it is converted...

matejkriz: I guess it's here https://github.com/trezor/trezor-suite/blob/fix/balance-mismatch/suite-common/wallet-utils/src/accountUtils.ts/#L474-L475

matejkriz (approved):
It is a bit dirty and could use some refactoring later, but it is a HUUUGE improvement from user perspective 🎉 🎖️ 😍

PR #16459, opened 2025-01-17
Cardano + tokens issues

### LINKED ISSUE #16410: Cardano fiat rate on dashboard is wrong
**Describe the bug**
Fiat rate above graph on dashboard is wrong for Cardano (probably because of some token). XPUB for testing:
`0c8b8b13536f2e46ad2cf696dc6849e435e2da680ec9fbf00b346eed093e1b26b8584f456117cda444289858548456f3d174e3659bd8fea8fdaffe8d517dc2ce
`
edit: After restarting the app, the fiat rate gets fixed but only for portfolio tracker, with connected device the fiat rates probably gets refreshed with the tokens so the fiat stays "broken".

**Info:**

-   Mobile Suite Version: 25.1.2 
-   OS: iOS and Android

**How to reproduce**

1. Sync the xpub in description
2. Confirm it
3. Observe the fiat above the dashboard graph and the discrepancy between that and the fiat value in account

**Screenshots**

Image

### LINKED ISSUE #16423: Cardano tokens showing when adding account to portfolio
**Describe the bug**
When you are adding a Cardano account to your Portfolio, it shows also tokens on that account which it shouldn't show.

**Info:**

-   Mobile Suite Version: 25.1.2 caac07f
-   OS: Android

**How to reproduce**

1. Go to Portfolio tracker
2. Add a Cardano account that has tokens as well
3. After adding the account Suite Lite shows tokens on the account as well when it shouldn't

**Expected behavior**
Only Cardano should appear when adding Cardano account with tokens

vytick:
a6eedc8 count fiat balance for cardano tokens after decimals are applied
3aa329a dont fetch token definitions for not enabled networks and dont show them in this case

## Related Issue

Resolve #16423 
Resolve #16410

## Screenshots:

BEFORE:

AFTER:

### DIFF
--- a/suite-native/discovery/src/discoveryConfigSlice.ts
+++ b/suite-native/discovery/src/discoveryConfigSlice.ts
@@ -1,5 +1,5 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
-import { A, F, pipe } from '@mobily/ts-belt';
+import { A, pipe } from '@mobily/ts-belt';
 
 import { createWeakMapSelector, returnStableArrayIfEmpty } from '@suite-common/redux-utils';
 import {
@@ -23,6 +23,7 @@ import {
     selectIsFeatureFlagEnabled,
 } from '@suite-native/feature-flags';
 import {
+    isCoinWithTokens,
     selectNetworkSymbolsOfAccountsWithTokensAllowed,
     TokensRootState,
 } from '@suite-native/tokens';
@@ -185,7 +186,11 @@ export const selectTokenDefinitionsEnabledNetworks = createMemoizedSelector(
     [selectEnabledDiscoveryNetworkSymbols, selectNetworkSymbolsOfAccountsWithTokensAllowed],
     (enabledNetworkSymbols, accountNetworkSymbols) =>
         returnStableArrayIfEmpty(
-            F.toMutable(A.uniq([...enabledNetworkSymbols, ...accountNetworkSymbols])),
+            pipe(
+                [...enabledNetworkSymbols, ...accountNetworkSymbols],
+                A.filter(s => isCoinWithTokens(s)),
+                A.uniq,
+            ),
         ),
 );
 
--- a/suite-native/module-accounts-import/src/accountsImportThunks.ts
+++ b/suite-native/module-accounts-import/src/accountsImportThunks.ts
@@ -23,6 +23,7 @@ import {
     selectFilterKnownTokens,
     selectNetworkTokenDefinitions,
 } from '@suite-common/token-definitions';
+import { isCoinWithTokens } from '@suite-native/tokens';
 
 import { paymentTypeToAccountType } from './constants';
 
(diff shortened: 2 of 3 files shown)


suite-native/discovery/src/discoveryConfigSlice.ts
```diff
@@ -185,7 +186,11 @@ export const selectTokenDefinitionsEnabledNetworks = createMemoizedSelector(
     [selectEnabledDiscoveryNetworkSymbols, selectNetworkSymbolsOfAccountsWithTokensAllowed],
     (enabledNetworkSymbols, accountNetworkSymbols) =>
         returnStableArrayIfEmpty(
-            F.toMutable(A.uniq([...enabledNetworkSymbols, ...accountNetworkSymbols])),
+            F.toMutable(
+                A.uniq([...enabledNetworkSymbols, ...accountNetworkSymbols]).filter(s =>
```
Nodonisko: Please use `pipe` and ideally move filter before uniq

suite-native/module-accounts-import/src/components/AccountImportConfirmFormScreen.tsx
```diff
@@ -54,9 +55,10 @@ export const AccountImportConfirmFormScreen = ({
     const navigation = useNavigation<NavigationProp>();
     const navigateToInitialScreen = useNavigateToInitialScreen();
     const showImportError = useShowImportError(symbol, navigation);
+    const canHaveTokens = isCoinWithTokens(symbol);
 
     const knownTokens = useSelector((state: TokenDefinitionsRootState) =>
-        selectFilterKnownTokens(state, symbol, accountInfo.tokens ?? []),
+        canHaveTokens ? selectFilterKnownTokens(state, symbol, accountInfo.tokens ?? []) : [],
```
Nodonisko: would this just return empty array even without condition?

vytick: 3f9f1a2

suite-native/discovery/src/discoveryConfigSlice.ts
vytick: 3f9f1a2

suite-native/module-accounts-import/src/accountsImportThunks.ts
```diff
@@ -119,8 +120,8 @@ export const getAccountInfoThunk = createThunk<
             if (fetchedAccountInfo?.success) {
                 const tokenDefinitions = selectNetworkTokenDefinitions(getState(), symbol);
 
-                // fetch token definitions for this network in case they are needed
-                if (!tokenDefinitions) {
+                // fetch token definitions for this network in case they are needed and enabled
```
matejkriz: nitpick: "enabled" is a bit confusing – for a while I was expecting it to be a condition on enabled networks, which wouldn't make sense. But no need to change it.

matejkriz (approved):
From now on, I'll call Careano by the name of the branch 😉 

Tested and works well 👌

PR #16461, opened 2025-01-18
fix(mobile): fw update redirect to home and temp remember device

### LINKED ISSUE #16460: Fix device temporary remembered during FW update
On mobile we are trying to temporary remember device during FW update.

Check file `suite-native/device/src/deviceThunks.ts`

Nodonisko:
## Description

This may help with redirect after FW upgrade is done, it should also fix temporary remember device during FW update.

## Related Issue

Resolve #16460


Nodonisko:
Not correct solution

PR #16463, opened 2025-01-18
refactor(e2e): Switch to ubuntu 24 for desktop e2e tests

### LINKED ISSUE #16462: Solve electron lunch failure in playwright on CI ubuntu 24.x
Options:
- try electron arg --no-sandbox
- update electron builder to 26.x and tell teams to adopt the change

Vere-Grey:
## Description

Adopt ubuntu 24 for desktop test builds since 22 will be soon not available.
Disables the problematic security rule 'Restricted unprivileged user namespaces' that cause problem when running electron with sandbox.

## Related Issue

Resolve [#16462](https://github.com/trezor/trezor-suite/issues/16462)

### DIFF
--- a/.github/workflows/test-suite-desktop-e2e.yml
+++ b/.github/workflows/test-suite-desktop-e2e.yml
@@ -16,8 +16,8 @@ on:
       - ".github/workflows/release*"
       - ".github/workflows/template*"
   push:
-      branches:
-        - release/2*
+    branches:
+      - release/2*
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:
@@ -29,7 +29,7 @@ concurrency:
 jobs:
   run-desktop-tests:
     if: github.repository == 'trezor/trezor-suite'
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
       matrix:
@@ -50,6 +50,13 @@ jobs:
             CONTAINERS: "trezor-user-env-unix bitcoin-regtest"
 
     steps:
+        # Electron requires unprivileged user namespaces to function properly. 
+        # Disabling this security rule allows Electron to create sandboxed processes
+        # without requiring elevated privileges, which is essential for running the application.
+        # This is workaround until electron builder solves this issue in future release.
+      - name: Disable security rule 'Restricted unprivileged user namespaces'
+        run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+
       - name: Checkout
         uses: actions/checkout@v4
         with:


HajekOndrej (commented):
I don't feel like I know enough to approve such things...

.github/workflows/test-suite-desktop-e2e.yml
```diff
@@ -29,7 +29,7 @@ concurrency:
 jobs:
   run-desktop-tests:
     if: github.repository == 'trezor/trezor-suite'
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
mroz22: maybe pinning it to a specific version would be more future-proof. latest can change

.github/workflows/test-suite-desktop-e2e.yml
```diff
@@ -50,6 +50,9 @@ jobs:
             CONTAINERS: "trezor-user-env-unix bitcoin-regtest"
 
     steps:
+      - name: Disable security rule 'Restricted unprivileged user namespaces'
+        run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
```
mroz22, line 58: this would deserve a comment, why is it needed?

.github/workflows/test-suite-desktop-e2e.yml
Vere-Grey: I have a counter argument. Adopting latest version prompt us sooner for any incompatibilities and problem sooner. And we will have time to fix it before the previous version is completely removed.

But if we stick to a fixed version, we are force to create some kind of monitoring that will alert us in time that the version soon be gone. If not that suddenly tests will stop working and we will have to upgrade immediately and solve immediately any problem with the newer version.

I am ok with both ways. Just presenting an argument to consider.

mroz22: both makes sense 
- unpinned - to have early alerts that we need to update 
- pinned - that it works and doesn't break randomly

in general, it makes sense to me to have a nightly job that tests canary environments. any job that needs to be green in order for PR to get merge should be as stable as possible

Vere-Grey: Stability is a word that resonates with me a lot.
Pinned to 24.04

.github/workflows/test-suite-desktop-e2e.yml
Vere-Grey: Comment added.

.github/workflows/test-suite-desktop-e2e.yml
```diff
@@ -29,7 +29,7 @@ concurrency:
 jobs:
   run-desktop-tests:
     if: github.repository == 'trezor/trezor-suite'
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
vdovhanych, line 32: I would put ubuntu-latest there

.github/workflows/test-suite-desktop-e2e.yml
Vere-Grey:
```
        # Electron requires unprivileged user namespaces to function properly. 
        # Disabling this security rule allows Electron to create sandboxed processes
        # without requiring elevated privileges, which is essential for running the application.
        # This is workaround until electron builder solves this issue in future release.
```

.github/workflows/test-suite-desktop-e2e.yml
vdovhanych: Hmm ok lets do it pinned then :D

vdovhanych (approved):

mroz22 (approved):

.github/workflows/test-suite-desktop-e2e.yml
Vere-Grey: @vdovhanych Please, join the discussion ubuntu-latest vs ubuntu-24.04

.github/workflows/test-suite-desktop-e2e.yml
Vere-Grey: Resolved in convo above.

PR #16467, opened 2025-01-20
fix(suite-native): connect popup use selected device

martykan:
## Description

Connect popup calls on mobile were missing the parameter for selected device/passphrase


mroz22 (approved):
I am thinking of adding something like this to detect and understand these cases better 
https://github.com/trezor/trezor-suite/pull/15996/commits/b20fa5e5a75a8aa2f0986995b78415da6e61532a

martykan:
Good idea, we just need to filter out methods that are not using device

mroz22:
I think they are already filtered. Connect won't emit this kind of event if it knows which device should be chosen

martykan:
Oh right, of course

PR #16470, opened 2025-01-20
wip: build:  publish also for module ESM with new libESM

### LINKED ISSUE #16245: Support ESM import of @metamask/eth-sig-util in @trezor/connect-plugin-ethereum
In a Manifest V3 environment (browser extension), @metamask/eth-sig-util v7 (and probably v8) (ESM-only), a dep of @trezor/connect-plugin-ethereum, causes runtime errors because @trezor/connect-plugin-ethereum (as of v9.0.5) compiles to CommonJS and does:

```
const sigUtil = tslib_1.__importStar(require('@metamask/eth-sig-util'));
```

This crashes in MV3 service workers because the lib no longer provides a CJS build. Steps to Reproduce:

1. Use @trezor/connect-plugin-ethereum in a Manifest V3 extension.
2. Bundle with Webpack or similar.
3. Observe that the background service worker crashes at runtime due to the ESM/CJS mismatch.

Why It Happens:

- @metamask/eth-sig-util v7 is ESM-only.
- `tslib.__importStar(require(...))` is effectively a CommonJS import.
- Manifest V3 service workers require the library to be bundled or imported as ESM.

Please provide an ESM-compatible version of @trezor/connect-plugin-ethereum that imports @metamask/eth-sig-util using native ESM. This would probably resolve the MV3 service worker crash 🙏

karliatto:
## Description

Adding source to the release NPM packaged. This is common nowadays so users of the packages can transpile it as they need. It also allow them to do tree shaking.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16245


packages/crypto-utils/package.json
```diff
@@ -22,7 +22,8 @@
     "files": [
         "lib/",
         "CHANGELOG.md",
-        "!**/*.map"
+        "!**/*.map",
+        "src/"
```
mroz22: "files" field doesn't always contain the same lines. Do we have the ambition to unify it now across all the published packages?

mroz22:
doesn't it need more changes in package json? some standardized instructions for bundlers where they should look? On the other hand, a src folder in the root feels quite standard already 🤔 just brainstorming

edit: for example stuff like this https://stackoverflow.com/questions/42708484/what-is-the-module-package-json-field-for

packages/crypto-utils/package.json
karliatto: That is a good question. Yes I think this is a good opportunity for doing that. Let's add `"!**/*.map"` and `"CHANGELOG.md"` ?

mroz22: changelog sounds good (btw isn't it included implicitly regardless of this files field)?
about the .map file I don't have any strong opinion but it is probably better to unify it

karliatto: You are right. It looks like 'yarn npm publish' includes CHANGELOG.md even if not included in `"files":[..]` in package.json. It is the same for `README.md`, but I realized in some packages we do not even have a README.md in the package. Like: https://www.npmjs.com/package/@trezor/blockchain-link-types

I think we should add at least a little README with the name and short description of the package, for the packages that we publish they probably deserve that minimum presentation.

karliatto: In order to be consistent we either can add `CHANGELOG.md` and `README.md` to the `"files":[..]` section in package.json or remove it from all of them and the effect will be the same since `yarn npm publish` will publish them anyway unless we include the files in `.npmignore`.

@mroz22 Any opinions regarding this? Adding `CHANGELOG.md` and `README.md` to the `"files":[..]`  section in all the publish packages or remove it?

karliatto: I like the idea of being explicit more so I would add something like `"*.md",` to the `"files":[..]` section.

karliatto:
> doesn't it need more changes in package json? some standardized instructions for bundlers where they should look? On the other hand, a src folder in the root feels quite standard already 🤔 just brainstorming
> 
> edit: for example stuff like this https://stackoverflow.com/questions/42708484/what-is-the-module-package-json-field-for

Yes, that is if we want to build packages for different modules, maybe for some packages where that makes sense we should do it. I mean for packages that are subject of being used in diverse environments.

packages/crypto-utils/package.json
mroz22: > In order to be consistent...

yes, I don't have any strong opinion here on whether to rely on the default or list everything explicitly

> I like the idea of being explicit more so I would add something like "*.md",

I wouldn't use any wildcards for md files, this could include stuff we don't actually want there.

> , but I realized in some packages we do not even have a README.md in the package

good catch, lets please add them 🙏

mroz22:
I still fail  to see what benefit it brings to our users to only have src added to the releases. 🤔 going through the node_moduels in our repo I can't see many src files with .ts files. Lets make one step back - what problem are we actually trying to solve by this?

packages/connect-plugin-ethereum/package.json
```diff
@@ -35,6 +36,8 @@
     "scripts": {
         "test:unit": "yarn g:jest -c ../../jest.config.base.js",
         "type-check": "yarn g:tsc --build tsconfig.json",
-        "build:lib": "yarn g:rimraf ./lib && yarn g:tsc --build tsconfig.lib.json && ../../scripts/replace-imports.sh ./lib"
+        "build:lib": "yarn g:rimraf ./lib && yarn g:tsc --build tsconfig.lib.json && ../../scripts/replace-imports.sh ./lib",
+        "build:lib:cjs": "yarn g:rimraf ./lib && yarn g:tsc --build tsconfig.lib.json && ../../scripts/replace-imports.sh ./lib",
+        "build:lib:esm": "yarn g:rimraf ./libESM && yarn g:tsc --build tsconfig.libESM.json && ../../scripts/replace-imports.sh ./lib"
```
Nodonisko: You will need to update `replace-imports.sh` scripts to update imports also in new folder and imports will be in ESM format so you will have to change regexp in file too

scripts/updateProjectReferences.ts
```diff
@@ -144,5 +144,37 @@ import { getPrettierConfig } from './utils/getPrettierConfig';
                     process.exit(1);
                 }
             }
+
+            // Copy references also to tsconfig.libESM.json if exists
```
Nodonisko: You shoudn't need this, just extend `tsconfig.lib.json` in root `tsconfig.libESM.json` instead of `base` file

packages/connect-plugin-ethereum/package.json
```diff
@@ -18,13 +18,14 @@
         "Ethereum"
     ],
     "sideEffects": false,
-    "publishConfig": {
```
Nodonisko, line 21: I am not sure if we want to delete this, in case someones bundler doesn't support `exports` field, it will pick `main: "src/index.ts"` which doesn't exists in published package

packages/connect-plugin-ethereum/package.json
```diff
@@ -18,13 +18,14 @@
...
-        "main": "lib/index.js"
-    },
     "main": "src/index.ts",
     "files": [
         "lib/"
     ],
+    "exports": {
+        "node": "./lib/index.js",
```
Nodonisko, line 26: node sure about syntax but I think there is something like `"commonjs": "./lib/index.js"` which may be better

.gitignore
```diff
@@ -49,6 +49,7 @@ playwright-report
 # build outputs
 lib
 libDev
+libESM
```
Nodonisko, line 52: Please check also other "ignore" files in monorepo root, there are many of them

karliatto: Right, updated https://github.com/trezor/trezor-suite/pull/16470/commits/ee0fb8c16334776deb3b1d3a05d88b040d62ecbc

packages/connect-plugin-ethereum/package.json
karliatto: I am not sure either but it sounds like good idea to keep both approaches. Seams there should not be problem with that.

packages/connect-plugin-ethereum/package.json
karliatto: Yes, I am still polishing it. Getting some inspirations from https://nodejs.org/api/packages.html#package-entry-points

packages/connect-plugin-ethereum/package.json
karliatto: Yes, what about this https://github.com/trezor/trezor-suite/pull/16470/files#diff-a103ae608a95489227c717243bcc41253b4a71f7b6a7e1b4fedaea4b7f7f96ad
?

I added new optional argument that will use `libESM` instead of `lib`

scripts/updateProjectReferences.ts
karliatto: Ok, so it sounds like a good idea. Do you mean extending the `tsconfig.lib.json` from the package and just changing the `module` and `outDir` like here https://github.com/trezor/trezor-suite/pull/16470/files#diff-97fc10c343bde083841616cc3853cf0d7e98003d6a1b11aa36a7665e69678b0b ?

packages/connect-plugin-ethereum/package.json
karliatto: Then we need to have all the `@trezor/...` packages that we consume to produce the `libESM`.

mroz22:
@karliatto this is outdated right?

karliatto:
> @karliatto this is outdated right?

Yes, outdated. It was taking different route https://github.com/trezor/trezor-suite/pull/16729 and then we can evaluate if add ESM to other packages step by step or wait for connect 10 ESM only.

PR #16471, opened 2025-01-20
chore(suite): add warning for unexpected use of connect methods (DX)

mroz22:
a small help for developers unfamiliar with connect api


martykan (approved):
LGTM, from some quick exploration I don't see the event happening so there shouldn't be false positives

PR #16477, opened 2025-01-20
Feat/suite handle fw update check

szymonlesisz:
## Description

WIP Based on https://github.com/trezor/trezor-suite/pull/16421
implementation of `firmwareUpdate` versionCheck and TrezorConnect `UI.FIRMWARE_VERSION_CHANGED` event

### DIFF
--- a/packages/suite/src/middlewares/suite/eventsMiddleware.ts
+++ b/packages/suite/src/middlewares/suite/eventsMiddleware.ts
@@ -16,6 +16,7 @@ import {
 import { DEVICE } from '@trezor/connect';
 
 import { SUITE } from 'src/actions/suite/constants';
+import { reportCheckFail } from 'src/components/suite/SecurityCheck/useReportDeviceCompromised';
 import { Action, AppState, Dispatch } from 'src/types/suite';
 
 /*
@@ -68,7 +69,16 @@ const eventsMiddleware =
         }
 
         if (action.type === DEVICE.FIRMWARE_VERSION_CHANGED) {
-            // TODO: show warning? move this to different middleware?
+            // TODO: Add UI.
+            const { device, oldVersion, newVersion } = action.payload;
+            reportCheckFail('Firmware version', {
+                model: device?.features?.internal_model,
+                revision: device?.features?.revision,
+                oldVersion,
+                newVersion,
+                vendor: device?.features?.fw_vendor,
+                error: 'Firmware version changed unexpectedly.',
+            });
         }
 
         if (deviceActions.selectDevice.match(action)) {
--- a/packages/suite/src/middlewares/suite/eventsMiddleware.ts
+++ b/packages/suite/src/middlewares/suite/eventsMiddleware.ts
@@ -67,6 +67,10 @@ const eventsMiddleware =
             }
         }
 
+        if (action.type === DEVICE.FIRMWARE_VERSION_CHANGED) {
+            // TODO: show warning? move this to different middleware?
+        }
+
         if (deviceActions.selectDevice.match(action)) {
             // Find and mark all notification associated (new connected!, update required etc)
             if (!action.payload) return action;
--- a/packages/suite/src/support/extraDependencies.ts
+++ b/packages/suite/src/support/extraDependencies.ts
@@ -32,6 +32,7 @@ import {
     findLabelsToBeMovedOrDeleted,
     moveLabelsForRbfAction,
 } from 'src/actions/wallet/moveLabelsForRbfActions';
+import { reportCheckFail } from 'src/components/suite/SecurityCheck/useReportDeviceCompromised';
 import { selectIsWindowVisible } from 'src/reducers/suite/windowReducer';
 import { fixLoadedCoinjoinAccount } from 'src/utils/wallet/coinjoinUtils';
 
@@ -216,5 +217,6 @@ export const extraDependencies: ExtraDependencies = {
     utils: {
         saveAs: (data, fileName) => saveAs(data, fileName),
         connectInitSettings,
+        reportCheckFail,
     },
 };
--- a/suite-common/firmware/src/firmwareThunks.ts
+++ b/suite-common/firmware/src/firmwareThunks.ts
@@ -154,7 +154,7 @@ export const firmwareUpdate = createThunk<
                 connectResponse: firmwareUpdateResponse,
             });
         } else {
-            const { check } = firmwareUpdateResponse.payload;
+            const { check, versionCheck } = firmwareUpdateResponse.payload;
             if (check === 'mismatch') {
                 // hash check was performed, and it does not match, so consider firmware counterfeit
                 dispatch(handleFwHashMismatch(device));
@@ -165,6 +165,8 @@ export const firmwareUpdate = createThunk<
                         errorMessage: firmwareUpdateResponse.payload.checkError,
                     }),
                 );
+            } else if (!binary && !versionCheck) {
+                // TODO: log to sentry
             } else {
                 dispatch(handleFwHashValid(device));
             }
--- a/suite-common/firmware/src/firmwareThunks.ts
+++ b/suite-common/firmware/src/firmwareThunks.ts
@@ -79,6 +79,7 @@ export const firmwareUpdate = createThunk<
 
         const {
             selectors: { selectLanguage },
+            utils: { reportCheckFail },
         } = extra;
 
         const device = selectSelectedDevice(getState());
@@ -154,7 +155,14 @@ export const firmwareUpdate = createThunk<
                 connectResponse: firmwareUpdateResponse,
             });
         } else {
-            const { check, versionCheck } = firmwareUpdateResponse.payload;
+            const {
+                check,
+                versionCheck,
+                bootloaderVersion,
+                binaryVersion,
+                installedVersion,
+                releaseVersion,
+            } = firmwareUpdateResponse.payload;
             if (check === 'mismatch') {
                 // hash check was performed, and it does not match, so consider firmware counterfeit
                 dispatch(handleFwHashMismatch(device));
@@ -165,10 +173,22 @@ export const firmwareUpdate = createThunk<
                         errorMessage: firmwareUpdateResponse.payload.checkError,
                     }),
                 );
-            } else if (!binary && !versionCheck) {
-                // TODO: log to sentry
             } else {
                 dispatch(handleFwHashValid(device));
+            }
+
+            // TODO: Add to the if-else block above and add handle in UI.
+            if (!binary && !versionCheck) {
+                reportCheckFail('Firmware version', {
+                    model: device.features?.internal_model,
+                    revision: device.features?.revision,
+                    vendor: device.features?.fw_vendor,
+                    bootloaderVersion,
+                    binaryVersion,
+                    installedVersion,
+                    releaseVersion,
+                    error: 'Unexpected firmware version change during firmware update.',
+                });
             }
 
             return fulfillWithValue({
--- a/suite-common/wallet-core/src/device/deviceActions.ts
+++ b/suite-common/wallet-core/src/device/deviceActions.ts
@@ -2,7 +2,7 @@ import { createAction } from '@reduxjs/toolkit';
 
 import { ButtonRequest, TrezorDevice } from '@suite-common/suite-types';
 import { WalletType } from '@suite-common/wallet-types';
-import { DEVICE, Device } from '@trezor/connect';
+import { DEVICE, Device, DeviceVersionChanged } from '@trezor/connect';
 
 export const DEVICE_MODULE_PREFIX = '@suite/device';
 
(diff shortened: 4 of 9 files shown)


packages/connect/src/utils/firmwareUtils.ts
```diff
@@ -25,14 +25,29 @@ export const isValidReleases = (extReleases: any): extReleases is FirmwareReleas
 export const filterSafeListByBootloader = (
     releasesList: FirmwareRelease[],
     bootloaderVersion: VersionArray,
-) =>
-    releasesList.filter(
+) => {
+    // bootloader_version indicates that there could be more than one version
```
komret: Does it? I see `bootloader_version?: VersionArray;` in the types, i.e. one version defined by `[number, number, number]`.

packages/connect/src/utils/firmwareUtils.ts
```diff
@@ -25,14 +25,29 @@ export const isValidReleases = (extReleases: any): extReleases is FirmwareReleas
...
+    if (bootloaderVersions.length > 0 && !compatibleRelease) {
+        // no compatible releases, bootloaderVersion is greater than any known
+        return [];
+    }
+
+    return releasesList.filter(
         item =>
             (!item.min_bootloader_version ||
                 versionUtils.isNewerOrEqual(bootloaderVersion, item.min_bootloader_version)) &&
             (!item.bootloader_version ||
```
komret: It looks like your addition goes agains the intention here, i.e. if bootloader version is undefined, keep it in the list. I don't know which behaviour is correct, but it seems to me like we're trying to solve a non-existing case because `bootloader_version` is always defined in `releases.json` (but not in tests). Maybe we should just change the type.

packages/connect/src/types/api/firmwareUpdate.ts
```diff
@@ -15,17 +16,27 @@ export const FirmwareUpdate = Type.Union([
...
-      };
+type FirmwareUpdateDetails = {
+    versionCheck: boolean; // installedVersion == binaryVersion == releaseVersion
+    bootloaderVersion: VersionArray; // bootloader version
+    installedVersion: VersionArray; // version installed by this process
+    binaryVersion: VersionArray; // version read from the binary
+    releaseVersion?: VersionArray; // version offered automatically by config. undefined if params.binary is used
+};
+
+export type FirmwareUpdateResponse = FirmwareUpdateDetails &
```
komret: We should make sure we properly differentiate between hash check and version check and the API is unified. I propose:

```
type FirmwareVersionCheckResult = {
    result: 'valid' | 'mismatch'; // installedVersion == binaryVersion == releaseVersion
    bootloaderVersion: VersionArray; // bootloader version
    installedVersion: VersionArray; // version installed by this process
    binaryVersion: VersionArray; // version read from the binary
    releaseVersion?: VersionArray; // version offered automatically by config. undefined if params.binary is used
};

type FirmwareHashCheckResult = {
              result:
                  | 'valid' // ok
                  | 'mismatch' //  firmware is not legitimate
                  | 'omitted'; // custom fw binary, or maybe older fw
          }
        | {
              result: 'other-error';
              payload: string; // unable to carry out the check due to a non-related error such as disconnected device
          };

FirmwareUpdateResponse = {
    versionCheckResult: FirmwareVersionCheckResult,
    hashCheckResult: FirmwareHashCheckResult
}
```

packages/suite/src/middlewares/suite/eventsMiddleware.ts
```diff
@@ -67,6 +67,10 @@ const eventsMiddleware =
             }
         }
 
+        if (action.type === DEVICE.FIRMWARE_VERSION_CHANGED) {
+            // TODO: show warning? move this to different middleware?
```
komret: TODO log to Sentry (I can do it).

suite-common/firmware/src/firmwareThunks.ts
```diff
@@ -165,6 +165,8 @@ export const firmwareUpdate = createThunk<
                         errorMessage: firmwareUpdateResponse.payload.checkError,
                     }),
                 );
+            } else if (!binary && !versionCheck) {
+                // TODO: log to sentry
```
komret: I can do it.

suite-common/wallet-core/src/device/deviceActions.ts
```diff
@@ -29,6 +29,16 @@ const deviceDisconnect = createAction(DEVICE.DISCONNECT, (payload: TrezorDevice)
     payload,
 }));
 
+// this action is not used but is required because of the typings
```
komret, line 32: This is awkward... how should it be resolved, @peter-sanderson? I see you removed `DeviceEvent` [here](https://github.com/trezor/trezor-suite/commit/c02412bccf80da7c827f624b7a7c85cdedf278c5#diff-2e9d057f0bfe2cc92fe50d4ce28838622d9e79fcca010ab8847a0fa288da13fd)

komret (changes requested):

packages/connect/src/utils/firmwareUtils.ts
szymonlesisz: well, there was a bug and it was revealed by tests:

if device bootloader was higher than any known version listed in the releases then `filterSafeListByBootloader` was offering first release without bootloader_version (undefined) field (some prehistoric 2.5.x)

packages/connect/src/utils/firmwareUtils.ts
szymonlesisz: its not about the types but values.
if releases.json item contains "bootloader_version" field then it means that there could be (but doesnt have to) more than one  bootloader_version released. and we should check if any of listed version is greater or equal that the current device bootloader (see bug description below)

suite-common/wallet-core/src/device/deviceActions.ts
peter-sanderson: I don't remember but I would guess it has to do something with this:

https://github.com/trezor/trezor-suite/commit/c02412bccf80da7c827f624b7a7c85cdedf278c5#diff-2ec00308835c01545c383eaef7034f26bef251f7e04e990e21aac46d0afb2c8bR71

As I read it, I think that idea was to define actions independently from Connect Event and map them instead of using them directly.

packages/connect/src/utils/firmwareUtils.ts
komret: I see, it is not always defined. Then why not changing the `filter` callback to

```
    releasesList.filter(
        item => versionUtils.isNewerOrEqual(bootloaderVersion, item.min_bootloader_version &&
            (item.bootloader_version &&
                versionUtils.isNewerOrEqual(item.bootloader_version, bootloaderVersion)),
    );
```

I tried that and the tests pass.

packages/connect/src/types/api/firmwareUpdate.ts
szymonlesisz: too late, its already in develop:
https://github.com/trezor/trezor-suite/pull/16421

suite-common/wallet-core/src/device/deviceActions.ts
peter-sanderson: ~I think this is the place where mapping was intended to happen: https://github.com/trezor/trezor-suite/commit/c02412bccf80da7c827f624b7a7c85cdedf278c5#diff-2e9d057f0bfe2cc92fe50d4ce28838622d9e79fcca010ab8847a0fa288da13fdR38~

peter-sanderson: Ultimately the mapping is here: https://github.com/trezor/trezor-suite/commit/c02412bccf80da7c827f624b7a7c85cdedf278c5#diff-2580f7331a6878d62ed30bb9de2c33adbf4566b5d67d8cf33f5705f48af51930R630

szymonlesisz: mapping is done only for 2 actions however the rest of DEVICE_EVENTs are dispatched directly [from here](https://github.com/trezor/trezor-suite/blob/develop/suite-common/connect-init/src/connectInitThunks.ts#L43) not from the deviceActions, yet its required to be defined in deviceActions to satisfy the types.

this problem applies to any other DEVICE_EVENTs  action.type like DEVICE.DISCONNECT or DEVICE.PASSPHRASE or DEVICE.PIN

szymonlesisz:
rebased, forcepushed, only one commit is relevant (implementation in suite)

packages/suite/src/middlewares/suite/eventsMiddleware.ts
```diff
@@ -67,6 +67,10 @@ const eventsMiddleware =
             }
         }
 
+        if (action.type === DEVICE.FIRMWARE_VERSION_CHANGED) {
```
peter-sanderson, line 71: Despite this is working I think it shall be: 

```ts
if (deviceActions.deviceFirmwareVersionChanged.match(action)) {
```

(More context coming in other comments).

suite-common/wallet-core/src/device/deviceActions.ts
peter-sanderson: I am looking into it. There is definitely a problem there.

Here it is typed `TrezorConnect.on(DEVICE_EVENT, ....` as `DeviceEventMessage` which contains only those types. I assume this is correct. 

image


I have tried to play with it in this commit https://github.com/trezor/trezor-suite/commit/29aaaf42f0ae2b3b706f35309dfad9a664b4aba1 and this is a draft of how I think it shall be done. 

But I am not sure how to make types work properly and also I need to run now. 

Lets talk tomorrow.

szymonlesisz: what is the benefit of doing things like this? 
what i see is 1000% more lines of code, more complexity and added another todo's which will be forgotten until next the occurrence + one potential problem if someone add another DEVICE_EVENT to connect (throwing error)

komret: The idea is to be explicit and to [unify syntax](https://github.com/trezor/trezor-suite/pull/16477#discussion_r1934186496) across the codebase. However, the current implementation is not finished and having to create actions that are not dispatched is indeed counter-intuitive. @peter-sanderson will try to finish the work he started in his [branch](https://github.com/trezor/trezor-suite/commit/29aaaf42f0ae2b3b706f35309dfad9a664b4aba1) - should he not succeed, I'll broaden the type again to remove the counter-intuitive behaviour.

szymonlesisz: so in the future you want to duplicate all the definitions of the events emitted from TrezorConnect? 
like for example [40+ ui-events](https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/events/ui-request.ts)? and create this looong if or switch to dispatch a proper action?

because this is how i undestand this comment:

```
TrezorConnectEvents // Todo: This should not be here, actions shall be defined independently from Connect Events (and they shall be mapped onto them)
```

instead of 4-6 lines of code (current state) there will be like 200-300 in multiple files?

komret: I agree this is too many events to map.

packages/suite/src/support/extraDependencies.ts
```diff
@@ -216,5 +217,6 @@ export const extraDependencies: ExtraDependencies = {
     utils: {
         saveAs: (data, fileName) => saveAs(data, fileName),
         connectInitSettings,
+        reportCheckFail,
```
peter-sanderson, line 220: wouldn't be better to add generic `report(Sentry)Error` dependency rather then specific reporting?

peter-sanderson: I can imagine to move whole `reportCheckFail` to common, and use it both in common and suite then. But inject the generic error handler (for sentry).

This way we can utilise error handling in any share code.

packages/suite/src/middlewares/suite/eventsMiddleware.ts
komret: [Example log in Sentry
](https://satoshilabs.sentry.io/issues/6270069178/?project=5193825&query=release%3A25.2.0.web.e4fa886a85ad6bb49b08e491a265cbf67bb0db12&referrer=release-issue-stream)Mocked by changing [this line](https://github.com/trezor/trezor-suite/blob/5b7ca32b9abf7c35f43b10e8b35c6bbfa9c293c6/packages/connect/src/device/Device.ts#L1004).

packages/suite/src/support/extraDependencies.ts
peter-sanderson: (can be done in future PR, but please lets do it).

suite-common/firmware/src/firmwareThunks.ts
komret: [Example log in Sentry 2](https://satoshilabs.sentry.io/issues/6270089317/events/646858e35eb74eefbbae0fa9c8a41bb5/)
Mocked by changing `installedVersion` [here](https://github.com/trezor/trezor-suite/blob/8a8a119931982a99e042235a587b7fcfe840a14c/packages/connect/src/core/onCallFirmwareUpdate.ts#L547).

packages/suite/src/support/extraDependencies.ts
komret: I agree, I did it this way because it was minimal effort.

peter-sanderson (approved):

komret (approved):

PR #16478, opened 2025-01-20
fix(suite-native): prohibit swiping back from authorize device screens

### LINKED ISSUE #15237: It's possible to swipe back from the Connecting screen
It's possible to swipe back from the Connecting screen and the PIN matrix remains shown on the device.

**Info:**

-   Mobile Suite Version: develop
-   OS: any

**Steps to reproduce**

1. Connect a Trezor Model One.
1. Wait for the Connecting screen to appear.
1. Swipe back as soon as the Connecting screen appears.
1. You end up on Home screen and PIN matrix remains shown on the device.

**Expected behavior**
It's either not possible to swipe back, or the PIN matrix is cleared from the device.

**Screenshots**
image

yanascz:
## Related Issue

Resolve #15237

### DIFF
--- a/suite-native/app/src/navigation/RootStackNavigator.tsx
+++ b/suite-native/app/src/navigation/RootStackNavigator.tsx
@@ -109,6 +109,7 @@ export const RootStackNavigator = () => {
                 options={{
                     ...stackNavigationOptionsConfig,
                     animation: 'slide_from_bottom',
+                    gestureEnabled: false,
                 }}
             />
             <RootStack.Screen
--- a/suite-native/coin-enabling/src/screens/CoinEnablingInitScreen.tsx
+++ b/suite-native/coin-enabling/src/screens/CoinEnablingInitScreen.tsx
@@ -1,13 +1,12 @@
 import { useDispatch, useSelector } from 'react-redux';
 import Animated, { SlideInDown, SlideOutDown } from 'react-native-reanimated';
-import { useEffect } from 'react';
-import { BackHandler, View } from 'react-native';
+import { View } from 'react-native';
 
 import { A } from '@mobily/ts-belt';
 import { useNavigation } from '@react-navigation/native';
 import { LinearGradient } from 'expo-linear-gradient';
 
-import { Screen } from '@suite-native/navigation';
+import { Screen, useHandleHardwareBackNavigation } from '@suite-native/navigation';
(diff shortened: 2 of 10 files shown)


suite-native/app/src/navigation/RootStackNavigator.tsx
```diff
@@ -109,6 +109,7 @@ export const RootStackNavigator = () => {
                 options={{
                     ...stackNavigationOptionsConfig,
                     animation: 'slide_from_bottom',
+                    gestureEnabled: false,
```
matejkriz, line 112: This does not fix it on android, only iOS.

matejkriz: And I'm not sure we want to disable the back gesture in whole AuthorizeDeviceStackNavigator. Maybe yes, just not sure what else it affects (passphrase flow, I guess)

PeKne: Agree with @matejkriz. Please change the `gestureEnabled` for the specific screen only.

PeKne: What about overriding the goBack event listener for this screen so it does not do anything? It should fix the issue for both platforms and will work with Android back button. We already have something like that used here: https://github.com/trezor/trezor-suite/blob/010eca96cc3e4e8439cb6391d200919d1ab0aa52/suite-native/module-send/src/hooks/useHandleOnDeviceTransactionReview.tsx#L56

yanascz: From what I tested, the fix works on both platforms. Tapping the back button on Android does the same thing as tapping the close button which is expected from my point of view.

I'm pretty confident we want to avoid swiping back on all the screens. In fact, I had to override the property even in `AuthorizeDeviceStack.Navigator` for it to work correctly for locked devices.

If you can think of a use case that is broken by this fix, please let me know. I couldn't find one.

PeKne (approved):
Seems to work as expected now 🥳

suite-native/app/src/navigation/RootStackNavigator.tsx
yanascz: In the end, I had to add and use `useHandleHardwareBackNavigation`. Now all the scenarios seem to work as expected.

PR #16481, opened 2025-01-20
Popover component for floating content next to a target element

### LINKED ISSUE #12939: Dropdown can render outside of viewport
**Describe the bug**
If user opens a dropdown on for example a mobile viewport it can render outside of the viewport and doesn't adjust its position based on the viewport.

**How to reproduce**

1. Go to some ETH account with 3 or more tokens
1. Go to tokens tabs
1. Resize window to mobile viewport
1. Click on tree dots

**Expected behavior**
It renders inside viewport so if there isn't a space on the bottom it renders to the top.

**Screenshots**
Screenshot 2024-06-18 at 10 45 03 AM

izmy:
## Description

Previously, the dropdown could render outside the viewport, particularly on smaller screens. So I created a new Popover component to keep content within the viewport. The component monitors whether the content is visible and dynamically adjusts its position when necessary. I used the `@floating-ui/react` library for this functionality, which is already used for tooltips.

I also refactored the `Dropdown` component, then separated the `Menu` file, and created an example in the storybook. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12939

## Screenshots:

image

image

image


packages/theme/src/zIndices.ts
```diff
@@ -4,6 +4,7 @@
 export const zIndices = {
     windowControls: 100,
     tooltip: 60, // above all content to be always fully visible when toggled
+    popover: 55, // above other content to be fully visible
```
tomasklim, line 7: maybe not enough
Screenshot 2025-01-22 at 12 39 46

tomasklim (commented):
Beautiful 🚀 🥳

packages/components/src/components/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Card } from '../Card/Card';
+
+const Container = styled.div`
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    gap: ${spacingsPx.xxl};
+`;
+
+const Content = () => <Card>Lorem ipsum dolor sit amet</Card>;
```
jvaclavik, line 28: Could you please update text color for dark more? 🙏 

image

packages/components/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,191 @@
...
+    useInteractions,
+    useMergeRefs,
+    Placement,
+    FloatingPortal,
+    FloatingFocusManager,
+} from '@floating-ui/react';
+
+import { zIndices } from '@trezor/theme';
+
+export const POPOVER_PLACEMENTS: Placement[] = [
```
jvaclavik: nitpick: @adamhavel shouldn't we use 2 props for placement instead of this one? 🤷‍♂️ WDYT?

And btw in general I like start/center/end more than it was before

packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -187,82 +96,38 @@ export const Dropdown = forwardRef(
...
-                setToggled={setToggled}
-                alignMenu={alignMenu}
-                addon={addon}
-            />,
-            document.body,
-        );
-
-        const frameProps = pickAndPrepareFrameProps(rest, allowedDropdownFrameProps);
-
         return (
```
jvaclavik, line 122: Nice abstraction 👏

packages/components/src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const Menu: StoryObj<MenuProps> = {
+    render: () => (
+        <MenuComponent
+            items={[
+                {
+                    key: 'appearance',
+                    label: 'Appearance',
+                    options: [
+                        {
```
jvaclavik, line 19: Could you please add also `onClick` with empty function? Right now it looks like it's not clickable which is the main purpose of menu item.

packages/components/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,191 @@
...
+    'right-start',
+    'right-end',
+    'bottom-start',
+    'bottom-end',
+    'left-start',
+    'left-end',
+] as const;
+
+export type PopoverProps = {
+    initialOpen?: boolean;
```
jvaclavik: nitpick: For better readability I suggest to rename boolean props:

```
initialOpen -> isInitialOpen

usePortal -> isPortalActive

onOpenChange?: (open: boolean) => void;
->
onOpenChange?: (isOpen: boolean) => void;
```

packages/components/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,191 @@
...
+    onInteraction?: () => void;
+};
+
+export function usePopover({
+    initialOpen = false,
+    placement = 'bottom',
+
+    isOpen: controlledOpen,
+    onOpenChange: setControlledOpen,
+}: PopoverProps = {}) {
```
jvaclavik: Can we do it without `= {}`? Not sure why it's necessary but maybe I'm missing something 🤷‍♂️

packages/components/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    {...context.getFloatingProps()}
+                >
+                    {children}
+                </div>
+            </FloatingFocusManager>
+        </FloatingPortal>
+    );
+});
+
+export function Popover({
```
jvaclavik, line 176: offtopic: shouldn't we also abstract `Tooltip` to `Popover` similarly like `Dropdown`? Does it make sense? WDYT?

packages/components/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+    const interactions = useInteractions([click, dismiss, role]);
+
+    return React.useMemo(
+        () => ({
+            open,
+            setOpen,
+            ...interactions,
+            ...data,
+
```
jvaclavik, line 89: nitpick: I saw this pattern in your PR multiple times. Why do you use empty line? 
I was thinking it's space between values and setters but here it's not like this. We don't use this pattern anywhere, so I would vote for not use it without good reason. But it's not a big deal.

packages/components/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,191 @@
...
+    const floatingProps = context.getFloatingProps(htmlProps);
+
+    return (
+        <FloatingPortal>
+            <FloatingFocusManager context={floatingContext}>
+                <div
+                    ref={ref}
+                    style={{ ...context.floatingStyles, ...(floatingProps.style as CSSProperties) }}
+                    aria-labelledby={context.labelId}
+                    aria-describedby={context.descriptionId}
```
jvaclavik, line 166: 👏

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx
```diff
@@ -121,10 +123,15 @@ const NavItem = (props: NavigationItemProps) => {
     const handleClick = (e: MouseEvent) => {
         e.stopPropagation();
 
+        if (onClick != null) {
```
jvaclavik: I prefer to avoid loose comparison here, also  I think it should be `null` at all. 

This should work the same IMHO:

```
if (onClick !== undefined) {
```

or maybe

```
onClick?.()

**Review by jvaclavik — COMMENTED** [review]
I like this PR very much. Good job @izmy 👏

I added few comments, but mostly nitpicks, no real blocker from my side.

File: packages/theme/src/zIndices.ts (hunk 1 above)
Comment (izmy) [reply packages/theme/src/zIndices.ts:7]: This is a different component (`react-select`) that doesn't use Popover as a dependency.

File: packages/components/src/components/Popover/Popover.tsx (hunk 8 above)
Comment (izmy) [reply packages/components/src/components/Popover/Popover.tsx:176]: Well, Tootip has a similar implementation, because it also uses `@floating-ui/react`, so I think it makes sense for me.

File: packages/components/src/components/Popover/Popover.tsx (hunk 9 above)
Comment (izmy) [reply packages/components/src/components/Popover/Popover.tsx:89]: I probably removed some `prop` while developing and forgot to remove an empty line 😀

File: packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx (hunk 11 above)
Comment (izmy) [reply packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/NavigationItem.tsx]: Ok, I used `onClick?.()`

File: packages/components/src/components/Popover/Popover.stories.tsx (hunk 2 above)
Comment (izmy) [reply packages/components/src/components/Popover/Popover.stories.tsx:28]: I had a problem with this because Storybook uses a portal outside of root div with appearance style. Still, I resolved it by setting the inline color, so now popover will work everywhere with the correct appearance.

File: packages/components/src/components/Popover/Popover.tsx (hunk 3 above)
Comment (izmy) [reply packages/components/src/components/Popover/Popover.tsx]: I really like that idea, so I changed the interface to 
```
type PopoverPlacement = {
    position: PopoverPlacementPosition;
    alignment: PopoverPlacementAlignment;
};
```

Comment (jvaclavik) [reply packages/components/src/components/Popover/Popover.tsx]: btw we already have this: `packages/components/src/config/types.ts`

```tsx
export const uiHorizontalAlignments = ['left', 'center', 'right'] as const;
export type UIHorizontalAlignment = (typeof uiHorizontalAlignments)[number];

export const uiVerticalAlignments = ['top', 'center', 'bottom'] as const;
export type UIVerticalAlignment = (typeof uiVerticalAlignments)[number];
```

Thinking if we can unify it somehow. 

Theoretically `popoverPlacementPosition` is `uiHorizontalAlignments` and `uiVerticalAlignments`

WDYT?

tomasklim (approved):

PR #16482, opened 2025-01-20
fix(suite-native): android navigation bar color on switching apps

PeKne:
## Description
- navigation bar on Android now adjusts its color on returning to Trezor Suite Lite from another app (that set different color)
- the logic was moved to `useAndroidNavigationBarStyle.tsx` hook to simplify `Screen` component

related to #16331

## Screenshots

https://github.com/user-attachments/assets/1b7dd8b5-1fb3-4b6b-9839-e1f71b89069d


Nodonisko (approved):

STew790:
**QA OK**
LGTM now, color stays the same when switching apps.

**Info**
25.1.2 c9351a4

PR #16500, opened 2025-01-21
fix(suite-native): firmware update is hidden for disconnected devices

### LINKED ISSUE #16499: It is possible to do a firmware update on view only device (that is not connected)
**Describe the bug**
I am able to get to firmware update even on view-only devices and it also lets me to update the device that I actually have connected even though it says something completely different and it is a different device (different model).

**Info:**

-   Mobile Suite Version: 25.1.2 c9351a4
-   OS: Android

**How to reproduce**

1. Connect and set to view only a device
2. Disconnect that device
3. Go to the view only device settings
4. Firmware update is available
5. Click on "Update"
6. You are able to get to the update screen even without connected device

You can also get to the firmware update with a different device connected and it prompts the device to update as well

**Expected behavior**
It is not possible to do a firmware update on view-only disconnected device.

PeKne:
- disconnected (view-only) devices do not offer firmware update

Closes #16499

### DIFF
--- a/suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
+++ b/suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
@@ -14,6 +14,7 @@ import {
     selectDeviceReleaseInfo,
     selectIsDiscoveryActiveByDeviceState,
     selectIsDeviceBackedUp,
+    selectIsDeviceConnected,
 } from '@suite-common/wallet-core';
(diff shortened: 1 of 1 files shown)


yanascz (approved):

matejkriz (approved):

PeKne:
Merging with failed android E2E pipeline. the fail is caused by DOGE backend outage.

PR #16507, opened 2025-01-21
feat(suite-native): hide balance if discovery is still loading so it's not final

matejkriz:
## Description

- Do not show global wallet balance above graph if it's not final, keep skeleton there until discovery is finished.
- Keep the balance visible while switching graph time frame (graph is loading, but current total value remains the same)

## Related Issue

Followup for https://github.com/trezor/trezor-suite/pull/16457


yanascz (approved):
🤯

STew790:
**QA OK**
Works really great :muscle: 

**Info**
25.1.3 https://github.com/trezor/trezor-suite/commit/33aa804612cca40cfb488f4b76435cd6a69ea17c

PR #16519, opened 2025-01-22
fix(suite-native): navigate to receive address screen after account is added

yanascz:
Steps to reproduce in previous app versions:

1. Enable just _Vertcoin_ in _Enabled coins_ in _Settings_.
2. Connect a device and wait for the empty _Home_ screen.
3. Tap the _Receive assets_ button on the empty _Home_ screen.
4. Choose _Bitcoin_ and wait for discovery to finish.
5. You cannot continue in the flow, just cancel it.

### DIFF
--- a/suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
+++ b/suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
@@ -148,6 +148,17 @@ export const useAddCoinAccount = () => {
         accountIndex: number;
     }) => {
         switch (flowType) {
+            case 'home':
+                navigation.replace(RootStackRoutes.ReceiveStack, {
+                    screen: ReceiveStackRoutes.ReceiveAccount,
+                    params: {
+                        networkSymbol: symbol,
+                        accountType,
+                        accountIndex,
+                        closeActionType: 'close',
+                    },
+                });
+                break;
             case 'accounts':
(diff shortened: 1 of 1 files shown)


suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
```diff
@@ -148,6 +148,17 @@ export const useAddCoinAccount = () => {
         accountIndex: number;
     }) => {
         switch (flowType) {
+            case 'home':
```
vytick, line 151: why not to use 'receive' case with fallthrough option?

yanascz: Because then closing the flow (or going back) doesn't work as expected.

matejkriz (approved):
Makes sense and works fine for me now, tested with T1 and only ADA enabled.

vytick (approved):

suite-native/module-add-accounts/src/hooks/useAddCoinAccount.ts
vytick: Oh, right. I missed that there is replace here and navigate in the other 👍 Makes sense!

STew790:
**QA OK**
Can not replicate now. Works fine and coin is enabled as well.

**Info**
25.1.3 33aa804

PR #16525, opened 2025-01-22
feat: Cancel Transaction

### LINKED ISSUE #6810: Feature: Cancel Transaction via RBF
## UPDATE

Design: https://www.figma.com/design/a13eFaT5HqemGmzJZy1fNg/Transactions?node-id=1-106&p=f&t=gzu2w9XlIcp7l0Rm-0

## Original post

Currently there are two options what an user can do with a RBF transaction:

* Bump Fee (create transaction with identical inputs but with higher fee = decreased change address amount)
* Finalize Transaction (create identical transaction but with RBF flag disabled

See https://trezor.io/learn/a/rbf-in-trezor-suite-app for more details

Originally, were considering adding a third option:

* Cancel Transaction (create transaction with identical inputs, with higher fee, but send all coins to a single change address = remove outputs that don't belong to Trezor from the TX)

In the end, we haven't added it, since this would provide an easy way for a user to fool merchants accepting 0-conf TXs via a double spend.
 
However, it seems the cat is out of the bag and other wallets have already added it, so we might want to add it as well:

* Blue Wallet - https://twitter.com/bluewalletio/status/1263038542340591617
* Electrum - https://github.com/spesmilo/electrum/pull/6641
* OBW - https://github.com/nbd-wtf/obw

peter-sanderson:
## Todos:

- [ ] fail screen in signing modal when original TX gets confirmed
- [x] ~NewModal.Icon~ Adam is ok with this
- [x] Use original own utxo if present in tx, do it same way as electron does it 
 
## Description

This PR Implements a Cancel Transaction flow by sending a RBF transaction what spends the (users) inputs back to thne users own address

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/6810

## Screenshots:
image
image
image


suite-common/wallet-core/src/send/composeCancelTransactionThunk.ts
```diff
@@ -0,0 +1,136 @@
...
+    newFeeRate: string;
+};
+
+type ConnectCallParams = Parameters<typeof TrezorConnect.composeTransaction>[0];
+
+const composeCancelTransaction = async ({ account, tx, newFeeRate }: ConnectComposeParams) => {
+    // override Account data, similar to RBF
+    const cancelAccount: ConnectCallParams['account'] = {
+        ...account,
+        utxo: getMineInputsFromTransaction({ tx, account }),
```
peter-sanderson: Theoretically we may only double-spend any of the inputs of the original transaction. That may reduce the trace on-chain.

However, the Electrum wallet also [spends all mine inputs](https://github.com/spesmilo/electrum/pull/6641/files#diff-6629668887b4072141a1a92c08bd30ef8fdc66acb21b26b1bfa166b57484b8a4R1508) so maybe we shall keep it same.

suite-common/wallet-core/src/send/composeCancelTransactionThunk.ts
```diff
@@ -0,0 +1,136 @@
...
+        addresses: account.addresses,
+    };
+
+    const params: ConnectCallParams = {
+        feeLevels: [{ feePerUnit: newFeeRate }],
+        account: cancelAccount,
+        outputs: [
+            {
+                type: 'send-max',
+                address: account.addresses?.unused[0].address, // Todo: is this correct way?
```
peter-sanderson: Electrum tries to use [original change output](https://github.com/spesmilo/electrum/pull/6641/files#diff-6629668887b4072141a1a92c08bd30ef8fdc66acb21b26b1bfa166b57484b8a4R1512) if present in the original transaction:

```python
 out_address = (self.get_single_change_address_for_new_transaction(old_change_addrs)
                       or self.get_receiving_address())
```

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/AffectedTransactions/AffectedTransactionItem.tsx
```diff
@@ -4,7 +4,7 @@ import { Icon, variables } from '@trezor/components';
 import { WalletAccountTransaction } from '@suite-common/wallet-types';
 import { truncateMiddle } from '@trezor/utils';
 
-import { FormattedDate } from 'src/components/suite';
+import { FormattedDate } from 'src/components/suite/index';
```
peter-sanderson: solve all this wtf `/index` changes (some auto import error of the IDE)

packages/suite/src/support/messages.ts
```diff
@@ -6377,7 +6377,37 @@ export default defineMessages({
...
+        defaultMessage: 'Cancel transaction',
+    },
+    TR_CANCEL_TX_HEADER: {
+        id: 'TR_CANCEL_TX_HEADER',
+        defaultMessage: 'Cancel transaction',
+    },
+    TR_CANCEL_TX_NOTICE: {
+        id: 'TR_CANCEL_TX_NOTICE',
+        defaultMessage:
+            'Once the transaction is canceled successfully, your funds (minus the transaction fee) will be returned to your wallet.',
```
peter-sanderson, line 6423: `Once the transaction is canceled successfully, your funds (minus the transaction fee) will be returned to your wallet.`

peter-sanderson: (design update)

packages/urls/src/urls.ts
```diff
@@ -117,6 +117,8 @@ export const HELP_CENTER_REPLACE_BY_FEE_ETHEREUM: Url =
     'https://trezor.io/learn/a/replace-by-fee-rbf-ethereum';
 export const HELP_CENTER_REPLACE_BY_FEE_BITCOIN =
     'https://trezor.io/learn/a/replace-by-fee-rbf-bitcoin';
+export const HELP_CENTER_CANCEL_TRANSACTION: Url =
+    'https://trezor.io/support/a/can-i-cancel-or-reverse-a-transaction';
```
peter-sanderson, line 121: - [ ] Todo: communicate and make sure this article will be updated

peter-sanderson: cc @rabbitholiness

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionFailed.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { HELP_CENTER_CANCEL_TRANSACTION } from '@trezor/urls';
+
+import { Translation } from '../../../../../Translation';
+import { TrezorLink } from '../../../../../TrezorLink';
+
+export const CancelTransactionFailed = () => (
+    <Card fillType="flat">
+        <Column gap={spacings.xs}>
+            <Box margin={{ bottom: spacings.md }}>
+                <IconCircle name="warning" size={110} variant="destructive" />
```
peter-sanderson, line 12: - [x] This is copy-paste from NewModal Icon code

This shall probably be solved as we shall have `<NewModal.Icon>` component to be used as building block.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    ? calculateChainedTransactionsFeeForRbf({ chainedTxs })
+                    : undefined,
+            hasCoinControlBeenOpened: false,
+            isCoinControlEnabled: false,
+            options: ['broadcast'],
+
+            outputs: composedCancelTx.outputs.map(output => ({
+                ...DEFAULT_PAYMENT, // Todo: WTF is this sorcery?
+                ...output,
+                amount: `${output.amount}`, // Todo: WTF is this sorcery?
```
peter-sanderson: - [x] Todo: ask somebody how to cope with this

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+        const fetchData = async () => {
+            if (account.addresses === undefined) {
+                return;
+            }
+
+            if (account.utxo === undefined) {
+                return;
+            }
+
+            // Todo: type-guard this
```
peter-sanderson: - [x] todo:...

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
```diff
@@ -62,7 +62,7 @@ const StyledFeeRow = styled(FeeRow)<{ $noInputsOutputs?: boolean }>`
 const DEFAULT_LIMIT = 3;
 
 type OpenModalParams = {
-    flow: 'detail' | 'bump-fee';
+    flow: 'detail' | 'bump-fee' | 'cancel-transaction';
```
Lemonexe, line 65: I know this is a pre-existing thing, just an idea for enhancement:
`OpenModaParams.flow` must conform to [this](https://github.com/trezor/trezor-suite/blob/0bfc08af63f67722e7aaa9e7558a5af20b61ac55/suite-common/suite-types/src/modal.ts#L54), otherwise TS gives error.
It's just unnecessary duplication (actually triplication, it's also in `TxDetailModal`).
I'm thinking how to do this better: maybe in modal.ts extract the type union as `TransactionDetailModalFlow` and import it here, and in `TxDetailModal`?
Idk, feel free to resolve this.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionFailed.tsx
peter-sanderson: adam is ok with this

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionFailed.tsx
```diff
@@ -0,0 +1,25 @@
...
+            </Box>
+
+            <Text typographyStyle="titleSmall">
+                <Translation id="TR_CANCEL_TX_FAILED" />
+            </Text>
+            <Translation id="TR_CANCEL_TX_FAILED_DESCRIPTION" />
+
+            <TrezorLink typographyStyle="hint" href={HELP_CENTER_CANCEL_TRANSACTION}>
+                <Translation id="TR_LEARN_MORE" />
```
Lemonexe, line 25: Nit: missing `ArrowUpRight` icon as per [figma](https://www.figma.com/design/a13eFaT5HqemGmzJZy1fNg/Transactions?node-id=4769-29278&t=t4nkbopSElq8dnoZ-4)

suite-common/wallet-core/src/send/composeCancelTransaction/cancelTrasnaxtionTypes.ts
Lemonexe, line 1: rename file to `cancelTransactionTypes.ts `

suite-common/wallet-core/src/send/composeCancelTransaction/resolveCancelAddress.ts
```diff
@@ -0,0 +1,26 @@
...
+export const resolveCancelAddress = ({ account, tx }: ResolveCancelAddress): string => {
+    const firstChangeAddress = tx.details.vout.find(vout => vout.isAccountOwned);
+
+    if (
+        firstChangeAddress !== undefined &&
+        firstChangeAddress.addresses !== undefined &&
+        firstChangeAddress.addresses.length > 0
+    ) {
+        return firstChangeAddress.addresses[0];
+    }
```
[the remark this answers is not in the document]
Lemonexe: Irrelevant

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+                return;
+            }
+
+            setComposedCancelTx(
+                await dispatch(composeCancelTransactionThunk({ account, tx, chainedTxs })).unwrap(),
+            );
+        };
+
+        fetchData();
+    }, [account, tx, dispatch, setComposedCancelTx, chainedTxs]);
```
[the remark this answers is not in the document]
Lemonexe: Well the :rabbit2:  does have a point - there is no rejection handling here, yet you do define rejection behavior in `composeCancelTransactionThunk`. When I artifically reject the thunk, the modal just gets stuck without explanation:

image

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+    onCancel,
+    onBackClick,
+    onShowChained,
+    chainedTxs,
+    selectedAccount,
+}: CancelTransactionModalProps) => {
+    const { account } = selectedAccount;
+
+    const dispatch = useDispatch();
+    const [composedCancelTx, setComposedCancelTx] = useState<PrecomposeResultFinal | null>(null);
```
Lemonexe, line 53: Maybe I'm missing something, but it feels strange to me that you define a state that is not used for anything else but being passed into a context. Why shouldn't it be part of the context?
The context starts with `composedCancelTx = null`, then you `setComposedCancelTx` and it continues with  `composedCancelTx = result`

Suggestion: have the state and its setter live in the context, so instead the context gets initiated just once with `null`. In this component you would `{setComposedCancelTx} = useCancelTxContext()` to set the result. Naturally that would require something like splitting `CancelTransactionModalInner` (with most of logic), leaving `CancelTransactionModal` just a thin` wrapper that renders ContextProvider.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { useDispatch, useSelector } from '../../../../../../../hooks/suite';
+import { CancelTxContext } from '../../../../../../../hooks/wallet/useCancelTxContext';
+import { Translation } from '../../../../../Translation';
+import { AffectedTransactions } from '../AffectedTransactions/AffectedTransactions';
+import { TxDetailModalBase } from '../TxDetailModalBase';
+
+const isComposeCancelTransactionPartialAccount = (
+    account: Account,
+): account is Account & ComposeCancelTransactionPartialAccount =>
+    account.addresses !== undefined && account.utxo !== undefined;
```
Lemonexe, line 30: Do you know in which case can an account have these undefined?
i.e. how can an account that is _not an_ `ComposeCancelTransactionPartialAccount` get to this flow? :thinking: 
BTW it's not gracefully handled, the modal gets stuck with disabled button, same as when the thunk is rejected.

suite-common/wallet-core/src/send/composeCancelTransaction/resolveCancelAddress.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { ComposeCancelTransactionPartialAccount } from './cancelTrasnaxtionTypes';
+
+type ResolveCancelAddress = {
+    account: ComposeCancelTransactionPartialAccount;
+    tx: Pick<WalletAccountTransaction, 'details'>;
+};
+
+export const resolveCancelAddress = ({ account, tx }: ResolveCancelAddress): string => {
+    const firstChangeAddress = tx.details.vout.find(vout => vout.isAccountOwned);
```
Lemonexe, line 11: If I understand correctly, you use **the** change address where the change from original Tx was sent to.
Normally this indeeds sends the canceling Tx to the change address.
But what if you cancel a change-less Tx (Tx that spends inputs only on one output + fee)?

I tried that, and as expected from L25, it sends the canceling Tx to the first unused _receiving_ address.

This makes the feature rather unpredictable - sometimes it sends to internal chain addr, sometimes to external chain addr.

But I don't know how easy it is to get next change address. I see that `account.addresses` has `change`, `used`, `unused`. Idk if the `change` is used or unused or both..?

:thought_balloon:  Btw I don't think it is an edge case to be dismissed. Changeless Tx makes sense e.g. when depositing to an exchange under your name, then it could be convenient to send exactly one UTXO to save fees and/or anonymity of other UTXOs.

suite-common/wallet-core/src/send/composeCancelTransaction/composeCancelTransactionThunk.ts
```diff
@@ -0,0 +1,143 @@
...
+    number,
+    CalculateNewTransactionSizeParams,
+    { rejectValue: string }
+>(
+    `${SEND_MODULE_PREFIX}/calculateNewTransactionSize`,
+    async ({ account, tx }, { rejectWithValue }) => {
+        const tempCancelTxResult = await composeCancelTransaction({
+            account,
+            tx,
+            newFeeRate: '1', // We don't care about the fee, we just need to compose transaction to get its size
```
Lemonexe, line 72: hmm, technically the fee rate can affect size. If you raise the fee by more than current output change, you must add a new input to cover the new fee.
I don't if this case is even possible with our current code - might be that it will just reject the RBF/Cancel Tx because it's not able to add new input, idk :shrug: 
I know it's an extreme edge scenario, but there might be a future where fee rates are very high.
Feel free to ignore it, just thinking out loud.

suite-common/wallet-core/src/send/composeCancelTransaction/composeCancelTransactionThunk.ts
```diff
@@ -0,0 +1,143 @@
...
+    async ({ tx, account, chainedTxs }, { rejectWithValue, dispatch }) => {
+        if (account.addresses === undefined) {
+            return rejectWithValue('Account addresses are not loaded');
+        }
+
+        if (tx.vsize === undefined) {
+            return rejectWithValue('Transaction vsize is not loaded');
+        }
+
+        const response = await dispatch(calculateNewTransactionSize({ account, tx }));
```
Lemonexe, line 106: please rename to `sizeCalculationResponse` or similar.
`response` sounds like something that will be returned by _this function_.
Also there is `result` later (that does makes sense as you do return a part of it), `response` vs `result` could be confusing.

suite-common/wallet-core/src/send/composeCancelTransaction/composeCancelTransactionThunk.ts
```diff
@@ -0,0 +1,143 @@
...
+    ComposeCancelTransactionPartialAccount,
+    ConnectComposeTxCallParams,
+} from './cancelTrasnaxtionTypes';
+import { resolveCancelAddress } from './resolveCancelAddress';
+
+type ConnectComposeParams = {
+    account: ComposeCancelTransactionPartialAccount;
+    tx: Pick<WalletAccountTransaction, 'details'>;
+    newFeeRate: string;
+    baseFee?: number;
```
Lemonexe, line 20: ok this is weird, `string` vs `number`. I see it's defined that way in connect, so ok, but it's not nice. Well we can't have everything..

Lemonexe (commented):
CR WIP; loading off comments, will finish tomorrow.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+        if (composedCancelTx === null) {
+            return;
+        }
+
+        const formState: FormState = {
+            feeLimit: '', // Eth only
+            feePerUnit: composedCancelTx.feePerByte,
+            baseFee:
+                chainedTxs !== undefined
+                    ? calculateChainedTransactionsFeeForRbf({ chainedTxs })
```
Lemonexe: why is this calculated again, even though we are working with `composedCancelTx` that was created by `composeCancelTransactionThunk`, which already calculated `baseFee` this way?

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    : undefined,
+            hasCoinControlBeenOpened: false,
+            isCoinControlEnabled: false,
+            options: ['broadcast'],
+
+            outputs: composedCancelTx.outputs.map(output => ({
+                ...DEFAULT_PAYMENT,
+                ...output,
+                amount: `${output.amount}`,
```
Lemonexe: tiny nit: this does the very same as `.toString()`, but toString is more explicit (I'd prefer).

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const { account } = selectedAccount;
+    const { networkType } = account;
+
+    const { composedCancelTx } = useCancelTxContext();
+
+    if (composedCancelTx === null) {
+        return;
+    }
+
+    if (composedCancelTx.outputs.length !== 1) {
```
Lemonexe, line 29: this condition puzzles me; it's neither a valid flow expected to happen, nor a TS only safeguard (does not affect TS).

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
```diff
@@ -0,0 +1,119 @@
...
+                            variant="nostyle"
+                            href={HELP_CENTER_CANCEL_TRANSACTION}
+                            icon="arrowUpRight"
+                        >
+                            <Translation id="TR_LEARN_MORE" />
+                        </TrezorLink>
+                    </Text>
+                </Row>
+
+                <Text variant="tertiary" typographyStyle="hint">
```
Lemonexe: According to figma, this should be `typographyStyle="hint"` but with `default` color instead of `subdued` (which is the _default_ for hint). Same for the "Transaction fee" and "Return to your wallet"

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    <Translation id="TR_CANCEL_TX_NOTICE" />
+                </Text>
+
+                <Divider orientation="horizontal" />
```
Lemonexe: According to figma, the divider should
1. be above `TR_CANCEL_TX_NOTICE` instead of below
2. extend to both sides. See [ChangeFee](https://github.com/trezor/trezor-suite/blob/49fd2cba426ba57a74050008fae5085e563b6b6b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/ChangeFee/ChangeFee.tsx#L36-L40)

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
```diff
@@ -0,0 +1,119 @@
...
+                            disableHiddenPlaceholder
+                            value={fee ?? undefined}
+                            symbol={tx.symbol}
+                        />
+                        <Text variant="tertiary" typographyStyle="label">
+                            <FiatValue
+                                disableHiddenPlaceholder
+                                amount={fee ?? '0'}
+                                symbol={tx.symbol}
+                                showApproximationIndicator
```
Lemonexe: `showApproximationIndicator` is not desired as per figma. Same below in the output amount `FiatValue`

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    direction="row"
+                    label={
+                        <>
+                            <Translation id="TR_CANCEL_TX_FEE" /> ({feePerByte.toFormat(2)}&nbsp;
+                            {getFeeUnits(networkType)})
+                        </>
+                    }
+                    typographyStyle="body"
+                >
+                    <Column gap={spacings.md} alignItems="baseline">
```
Lemonexe: `alignItems="flex-end"` because the items should be right aligned. Same below with the output amounts.
...actualy does `baseline` make any sense with Column? I only understand it with Row :thinking:

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
```diff
@@ -0,0 +1,119 @@
...
+                </Text>
+
+                <Divider orientation="horizontal" />
+
+                <InfoItem
+                    direction="row"
+                    label={
+                        <>
+                            <Translation id="TR_CANCEL_TX_FEE" /> ({feePerByte.toFormat(2)}&nbsp;
+                            {getFeeUnits(networkType)})
```
Lemonexe: the fee should not be wrapped in parentheses

suite-common/wallet-core/tests/send/composeCancelTransaction/composeCancelTransactionThunk.test.ts
```diff
@@ -0,0 +1,189 @@
...
+
+const transactionWithNoChange: Pick<WalletAccountTransaction, 'details' | 'vsize' | 'fee'> = {
+    fee: '1100',
+    vsize: 110,
+    details: {
+        vin: [
+            {
+                value: '8999998590',
+                n: 0,
+                addresses: ['bcrt1qte33uyyfzrdrm9nqk0uwlq9dqr6ezu2gurhree'],
```
Lemonexe, line 87: Tiny nit: this is the change address from previous fixture. I know it doesn't affect the test functionally, but looks weird. You could extract `INPUT_ADDRESS = bcrt1qreeergcmsw604zgd7hsreq6872swxnh3485fs5` and use in both Txs inputs for consistency.
You could also extract the output adress `bcrt1qjdkgvc67dnk3930gxj2valeadsg49datkyz7tu`. It would make this file easier readable.

suite-common/wallet-core/tests/send/composeCancelTransaction/composeCancelTransactionThunk.test.ts
```diff
@@ -0,0 +1,189 @@
...
+        // First call
+        const first = composeTransactionMock.mock.calls[0][0]; // First call, first argument
+        expect(first.feeLevels[0].feePerUnit).toBe('1');
+        expect(first.baseFee).toBe(undefined);
+
+        // Second call
+        const second = composeTransactionMock.mock.calls[1][0]; // Second call, first argument
+
+        // This is the most important assertion. This is the fee, that satisfies the condition set by BIP-125
+        // with the new size of the transaction of 110 bytes.
```
Lemonexe, line 159: you could state here for clarity that it is calculated `(1410 + 110*1)/110`.

suite-common/wallet-core/tests/send/composeCancelTransaction/calculateNewFee.test.ts
```diff
@@ -0,0 +1,18 @@
+import { chainedTxsFixture } from './chainedTransactions.fixture';
+import { calculateNewFee } from '../../../src/send/composeCancelTransaction/calculateNewFee';
+
+describe(calculateNewFee.name, () => {
+    it('accounts RELAY_FEE for the new transaction size (110B) which is less then original TX (141B) ', () => {
+        const result = calculateNewFee({
+            chainedTxs: chainedTxsFixture,
+            newTransactionSize: 110,
+            originalFee: '1410',
+            relayFee: 2,
```
Lemonexe, line 23: It's nice that you are testing a different `RELAY_FEE`, but it might look confusing that in this test you use only 2, while in the compose thunk test only the default value.
I'd either add comment here highlighting that we are testing value other than default, or copy paste the `it` for an undefined `relayFee` (to also test the defaults)

suite-common/wallet-core/src/send/composeCancelTransaction/composeCancelTransactionThunk.ts
peter-sanderson: I think this may not happen. 

1. you send TX with sum of output amounts = `amount` + you pay the `fee`,
2. then you change your mind and send all of it back to your address => you work with `amount` + `fee`
3. you may need theoretically any fee-rate (RBF allows up to 100 tx chain) but if you need more then is the whole value, there is no point to even do cancel => you would just pay extra fee to burn it all into fees. 

Maybe some absurdly rare edge-case where you rather burn it all, the let your recipient to have it (and you are even willing to pay more for it).

Lemonexe (commented):
Finished the review. Quite a lot of work here, and it's looking really good :rocket: 
Though I have quite a few comments to be addressed, some suggestions, some questions not clear to me. I'll be happy to discuss them further :slightly_smiling_face:

suite-common/wallet-core/src/send/composeCancelTransaction/composeCancelTransactionThunk.ts
Lemonexe: You're right, I didn't think this through. When cancelling Tx, you are not paying a fixed amount (which would mean increasing inputs to pay for the fee), but diminishing your returns (with inputs staying constant)

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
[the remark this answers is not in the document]
peter-sanderson: What about this? https://github.com/trezor/trezor-suite/pull/16525/commits/b36a64ac835d324827671d649221972e493227bf

image

suite-common/wallet-core/src/send/composeCancelTransaction/cancelTrasnaxtionTypes.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/16525/commits/328e1f0d794913f699451523dcbf6974a6837a0f

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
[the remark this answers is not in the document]
Lemonexe: @peter-sanderson LGTM!

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionFailed.tsx
peter-sanderson: Fixed - nice catch, you got the :eye: 
image

packages/suite/src/components/wallet/TransactionItem/TransactionItem.tsx
peter-sanderson: > Idk, feel free to resolve this.

Yea, tomorrow is another day.

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionFailed.tsx
peter-sanderson: will be: `59dabc1ed9`

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
peter-sanderson: solved over call

suite-common/wallet-core/src/send/composeCancelTransaction/resolveCancelAddress.ts
peter-sanderson: solved over call

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
peter-sanderson: resolve on call

suite-common/wallet-core/src/send/composeCancelTransaction/composeCancelTransactionThunk.ts
peter-sanderson: fixed

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionButton.tsx
peter-sanderson: fixed, removed not needed good point

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionButton.tsx
peter-sanderson: fixed

suite-common/wallet-core/tests/send/composeCancelTransaction/calculateNewFee.test.ts
peter-sanderson: fixed in `d9c64ffe2d`

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
peter-sanderson: fixed

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
peter-sanderson: :+1: Dunno, I have probably copy-pasted this from somewhere :(

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
peter-sanderson: Nice, you indeed got an :eye: for it

`c8e1854517`

image

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
peter-sanderson: :+1:  `760fae1f3b`

image

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
peter-sanderson: fixed

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
```diff
@@ -0,0 +1,122 @@
...
+                    margin={{ top: spacings.xs, bottom: spacings.xs }}
+                />
+
+                <InfoItem
+                    direction="row"
+                    label={<Translation id="TR_CANCEL_TX_RETURN_TO_YOUR_WALLET" />}
+                    typographyStyle="body"
+                    variant="default"
+                >
+                    <Column gap={spacings.md} alignItems="baseline">
```
Lemonexe: Sorry, I should have written comments for both cases instead of saying "same below", it can get overlooked easily..
```suggestion
                    <Column gap={spacings.md} alignItems="flex-end">
```

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
```diff
@@ -0,0 +1,122 @@
...
+                                amount={fee ?? '0'}
+                                symbol={tx.symbol}
+                            />
+                        </Text>
+                    </Column>
+                </InfoItem>
+
+                <Divider
+                    orientation="horizontal"
+                    margin={{ top: spacings.xs, bottom: spacings.xs }}
```
Lemonexe, line 95: I haven't noticed it before, but why the extra spacing? It's too much.

I can see in the figma there is divider with 8px vertical margin (like you made it here), but also the InfoItems have padding 8px there. In total, 16px gap between content
Instead of using padding you achieved the same with 16px gap on the Column, which I find absolutely preferrable :ok_hand: :heavy_check_mark: 
But should be without this divider margin, because with it, it's 24px in total = too much

Lemonexe (approved):
LGMT, just two tiny UI changes to be done **before merging**.

Using Regtest, I verified:
- the cancel flow success
- cancel is sent to correct addresses
- cancel failed if I mine block during the modal
- cancel Tx with chained Txs

Tests are well readable, even with the complexity of setting up a thunk test :+1: 

Also what's with the E2E tests? :open_mouth:  Might be the new watcher for unhandled errors..

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransaction.tsx
peter-sanderson: divider has `md` spacing by default, this reduces the spacings

Lemonexe: ok, then I misunderstood the situation. Ok, resolve in way, that seems suitable and aligns with design.

peter-sanderson:
## QA:

- see the original issue for more info
- there are multiple scenarios that needs to be tested:

1. Just cancel TX ordinary transaction
2. Cancel transaction that has follow up chain of additional transactions
3. single output transaction (no change) -> shall be send to first unused address
4. multiple recipients (no change) -> shall be send to first unused address
5. change + 
    a. (one recipinet -> whole amount shall be send to he originally used change address
    b. but also case with many other recipients) -> whole amount shall be send to he originally used change address
6. Original transaction gets mined before user finishes the Cancel Modal flow
7. Test please do also some smoketest on mainnet as we also show fiat rates

PR #16548, opened 2025-01-22
Are these libs used?

mroz22:
... let me try.

also why (some of them) are they listed as connect dependencies?  https://github.com/trezor/trezor-suite/blob/develop/scripts/list-outdated-dependencies/connect-dependencies.txt#L11

when deleted 

image

image

looks quite same to me 🤔 are these libs still needed? cc @tomasklim


mroz22:
ok, there is a difference 

image

image

PR #16550, opened 2025-01-22
XRP: fix status in failed transactions with missing destination tag

### LINKED ISSUE #16281: XRP tx incorrectly mark as successful even though it failed
**How to reproduce**

1. Send money to address (`rw2ciyaNshpHe7bCHo4bRWq6pqqynnWKQg`) which requires `destinationTag` but do not specify this `destinationTag` in send form
2. Tx fails according to notification
3. Check explorer https://xrpscan.com/tx/BE4D883282BA4A16121117F382E916504D24AC90530F5575032E8F0E6A63E6C5 it is failed
4. Check Suite tx history, it is ok but balance has not changed

**How to resolve**
Problem is probably in blockchain-link. I think we incorrectly mark `failed` tx as `sent`

**Screenshots**

Image
Image
Image
Image
Image

izmy:
## Description

Suite incorrectly displayed success status in XRP transactions when destination tag was missing.

I also changed the color for the Confirmed status in transaction detail, because it could have been a confusing UX.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16281

## Screenshots:

image

image

### DIFF
--- a/packages/blockchain-link-utils/src/ripple.ts
+++ b/packages/blockchain-link-utils/src/ripple.ts
@@ -1,5 +1,13 @@
 import type { Transaction } from '@trezor/blockchain-link-types';
 
+type RippleTransactionMetadata = {
+    AffectedNodes: any[];
+    DeliveredAmount?: any[];
+    TransactionIndex: number;
+    TransactionResult: string;
+    delivered_amount?: string;
+};
+
 // export const transformServerInfo = (payload: GetServerInfoResponse) => {
 export const transformServerInfo = (payload: any) => ({
     name: 'Ripple',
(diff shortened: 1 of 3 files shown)


packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/BasicTxDetails.tsx
```diff
@@ -94,7 +94,7 @@ export const BasicTxDetails = ({
                 <Row gap={spacings.xxs} margin={{ left: 'auto' }}>
                     {isConfirmed ? (
                         <InfoSegments typographyStyle="hint" variant="tertiary">
-                            <Text typographyStyle="callout" variant="primary">
+                            <Text typographyStyle="callout" variant="default">
```
tomasklim: I would omit this from this PR and let's discuss and find a solution to change it to a state that makes sense

packages/blockchain-link/src/workers/ripple/index.ts
```diff
@@ -181,7 +181,7 @@ const getAccountInfo = async (request: Request<MessageTypes.GetAccountInfo>) =>
 const getTransaction = async ({ connect, payload }: Request<MessageTypes.GetTransaction>) => {
     const api = await connect();
     const rawtx = await api.request('tx', { transaction: payload, binary: false });
```
tomasklim: nit: rawTx

packages/blockchain-link-utils/src/ripple.ts
```diff
@@ -15,13 +15,21 @@ export const transformServerInfo = (payload: any) => ({
...
-export const transformTransaction = (tx: any, descriptor?: string): Transaction => {
+export const transformTransaction = (meta: any, tx: any, descriptor?: string): Transaction => {
     const blockTime =
         typeof tx.date === 'number' && tx.date > 0 ? tx.date + BLOCKTIME_OFFSET : tx.date;
-    const type =
-        tx.TransactionType !== 'Payment' || !descriptor
-            ? 'unknown'
-            : (tx.Account === descriptor && 'sent') || 'recv';
+
+    let type: Transaction['type'];
```
tomasklim: nit:
```suggestion
    let txType: Transaction['type'];
```

tomasklim (changes requested):
We should probably hide amount change. However, the problem is that ripple does not show fee so it would look weird. Let's discuss

Screenshot 2025-01-23 at 8 53 36

tomasklim (changes requested):
Fix and add please new unit tests 🙏

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/BasicTxDetails.tsx
izmy: Ok, I created an issue -> https://github.com/trezor/trezor-suite/issues/16567

izmy:
> Fix and add please new unit tests 🙏

done

tomasklim (approved):

PR #16555, opened 2025-01-23
feat(solana): more generic transaction parsing

### LINKED ISSUE #10913: Solana unknown txs
**Describe the bug**
A lot of transactions are unknown in the transaction history

Currently, the Suite is displaying unknown transactions in these scenarios:
1. Smart contract transactions or interactions with unsupported parties, such as staking.
2. Transactions with 0 SOL sent, which are usually address poisoning attempts, for example, from [flip.gg](http://flip.gg/).

Adjusting the copy might be an interim solution.

**Info:**

-   Suite Version: 24.2.0

**How to reproduce**

1. Use Solana account for swaps, nfts,... not just suite txs

**Expected behavior**
All txs are parsed 

**Screenshots**
Screenshot 2024-01-30 at 14 22 17

Image

MartinZemanik:
### Parsing transaction token effects
Before, all Solana transaction token effects were parsed only based on Token transfer instructions - if there were no token transfer instructions directly in the tx, no token effects would be parsed.
Now, we also consider **all inner instructions** in the transaction. This way, we have a generic parser for token effects on all user's token accounts.

### Unknown transactions
Before, if the transaction wasn't a direct token or SOL transfer, it would be interpreted as `unknown`.
Now, if a transaction contains some instruction in a 3rd party program, it is interpreted as `contract` transaction (as agreed on Slack).

## Screenshots:
Swap USDC to USDT on Raydium (previously Unknown transaction with no token effects):
Screenshot 2025-01-14 at 13 01 16

Resolve [#10913](https://github.com/trezor/trezor-suite/issues/10913)

### DIFF
--- a/packages/blockchain-link-utils/src/solana.ts
+++ b/packages/blockchain-link-utils/src/solana.ts
@@ -37,6 +37,7 @@ export const SYSTEM_PROGRAM_PUBLIC_KEY = '11111111111111111111111111111111';
 // when parsing tx effects.
 export const WSOL_MINT = 'So11111111111111111111111111111111111111112';
 export const STAKE_PROGRAM_PUBLIC_KEY = 'Stake11111111111111111111111111111111111111';
+export const COMPUTE_BUDGET_PROGRAM_ID = 'ComputeBudget111111111111111111111111111111';
 
 const tokenProgramNames = ['spl-token', 'spl-token-2022'] as const;
 export type TokenProgramName = (typeof tokenProgramNames)[number];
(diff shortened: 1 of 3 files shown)


martykan:
Looks good 👍 . 

I just noticed some wallets use Serum's Assert Owner program during SPL transfer transactions, which makes them appear as contract calls and may be confusing for users. 
It might be a good idea to whitelist these: 
- `4MNPdKu9wFMvEeZBMt3Eipfs5ovVWTJb31pEXDJAAxX5` - [official deployment](https://github.com/project-serum/serum-dex/blob/master/assert-owner/program-id.md)
- `DeJBGdMFa1uynnnKiwrVioatTuHmNLpyFKnmB5kaFdzQ` - [Phantom](https://docs.phantom.com/resources/faq#why-does-phantom-prepend-an-additional-instruction-on-standard-spl-token-transfers)

Example of this TX: 
https://solscan.io/tx/5fatcntFUEwDqpRmjqnPxZcGF9K1qWFYoA9LkQbWCLnYZ1iE4N7443RH6abVfW1UmK8vuN4PBYqwvJdX489u28LN

martykan:
Great, thank you. Just please do `yarn format` for the last commit.
Also `fix(solana): ...` would be better for the commit message

MartinZemanik:
> Great, thank you. Just please do `yarn format` for the last commit. Also `fix(solana): ...` would be better for the commit message

Fixed 👍

martykan (approved):

PR #16570, opened 2025-01-23
fix(suite-native): do not include solana staking in balance

### LINKED ISSUE #16564: Solana some balances include stake balance and some not
**Describe the bug**
There is inconsistency on whether balance includes stake fiat balance or not. Some places have it, some not

**How to reproduce**

1. Go to Portfolio tracker
2. Import dev Solana account that has staking
3. See discrepancy between balances on dashboard

**Expected behavior**
Numbers are the same and do not include staking where not allowed

**Screenshots**

PeKne:
## Description
- Solana staking is not included in the account balances

## Related Issue

Resolve #16564 

## Screenshots:

https://github.com/user-attachments/assets/3b48fb2a-08fc-4950-851a-636ed874cc77

### DIFF
--- a/suite-native/device/src/selectors.ts
+++ b/suite-native/device/src/selectors.ts
@@ -23,8 +23,12 @@ import {
     selectAccountsByDeviceState,
 } from '@suite-common/wallet-core';
 import { isDeviceAcquired } from '@suite-common/suite-utils';
-import { getTotalFiatBalance } from '@suite-common/wallet-utils';
+import { getAccountFiatBalance } from '@suite-common/wallet-utils';
 import { selectFiatCurrencyCode, SettingsSliceRootState } from '@suite-native/settings';
+import { BigNumber } from '@trezor/utils';
+import { Account, RatesByKey } from '@suite-common/wallet-types';
+import { FiatCurrencyCode } from '@suite-common/suite-config';
+import { doesCoinSupportStaking } from '@suite-native/staking';
 
 import { isFirmwareVersionSupported } from './utils';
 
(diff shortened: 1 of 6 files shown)


suite-native/device/src/selectors.ts
```diff
@@ -71,30 +74,41 @@ export const selectDeviceError = (
...
+    rates?: RatesByKey;
+}) => {
+    let instanceBalance = new BigNumber(0);
+    deviceAccounts.forEach(a => {
+        const accountFiatBalance =
+            getAccountFiatBalance({
+                account: a,
+                localCurrency,
+                rates,
+                shouldIncludeStaking: a.networkType !== 'solana',
```
vytick: please, lets use `doesCoinSupportStaking ` from staking `package`

suite-native/device/src/selectors.ts
```diff
@@ -71,30 +74,41 @@ export const selectDeviceError = (
     return device?.error;
 };
 
+// FIXME: this function can be removed and substitued with @suite-common/wallet-utils/getTotalFiatBalance when Solana supports staking on mobile.
+const getTotalFiatBalance = ({
```
vytick: nit: something like ```suggestion
const getNativeTotalFiatBalance = ({
```

vytick (changes requested):

suite-native/device/src/selectors.ts
PeKne: fixed 0546e9d3ff

suite-native/device/src/selectors.ts
PeKne: fix 0546e9d3ff

vytick (approved):
I tested it rn and it looks good. Great job 👍

PR #16577, opened 2025-01-23
Fees redesign

enjojoy:
## Description

Implemented new design for fees 

[Design](https://www.figma.com/design/6ePOAsBHwKArEQ2DuYLqRl/Priority-fees?node-id=27-15615&t=B3GSgeToQZkZaCJB-0)

### DIFF
--- a/packages/components/src/components/RadioCard/RadioCard.stories.tsx
+++ b/packages/components/src/components/RadioCard/RadioCard.stories.tsx
@@ -1,0 +1,23 @@
+import { Meta, StoryObj } from '@storybook/react';
+
+import { RadioCard as RadioCardComponent, RadioCardProps } from './RadioCard';
+
+const meta: Meta = {
+    title: 'RadioCard',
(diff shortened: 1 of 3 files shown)


jvaclavik:
image

Just a quick note:
This component should be implemented in the design system.

API should be really simple and could looks like this:

```
type RadioCardProps = {
  isActive: boolean;
  children: React.ReactNode;
}
```

Usage: 
```tsx
<RadioCard
  isActive={true}
>
  {children}
</RadioCard>  
```

It should solve only one "card" and you can use Column for creating grid:

```tsx
<Column gap={...}>
  <RadioCard ... />
  <RadioCard ... />
</Column>
```

(`RadioCard` name is probably not final)

seibei-iguchi:
As @jvaclavik said, I wouldn't make this out of the Card component

enjojoy:
@adamhavel I already have this part done locally 🥲 But okay, I'll make the separate component and ask for your review

PR #16578, opened 2025-01-23
chore(monorepo): update deps responsibilities

mroz22:
I tried to remove these but it turned out that they are somehow used, see #16548
These libs don't have anything in common with connect team so I suggest shifting them to another team.


jvaclavik (approved):
Not happy about it, but it makes sense 😅

PR #16599, opened 2025-01-24
feat(suite): redesign FW checks UI

### LINKED ISSUE #16407: Improve Hash Check "other-error" UI

Lemonexe:
## Description

- reenable FW hash check other-error UI **only for dev env**
- :ghost: modal:
  - all variants have different subtitle to convey which check & what failure reason
  - other error also different title, checklist and button color = `warning`
- warning banner:
  - all variants have "Contact Trezor Support" button -> goto same URL as on :ghost: modal
  - `other-error` has color: `warning`

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16407

### :ghost: modals

**hash-mismatch**
ghost hash
**other-error**
ghost other-error
**revision check**
ghost rev

### Banners

**hash-mismatch**
banner hash
**other-error**
banner other-error
**revision check**
banner rev

### DIFF
--- a/packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
+++ b/packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
@@ -3,13 +3,53 @@ import { Card } from '@trezor/components';
 import { TREZOR_SUPPORT_FW_REVISION_CHECK_FAILED_URL } from '@trezor/urls';
 
 import { WelcomeLayout } from 'src/components/suite';
-import { useDevice, useDispatch } from 'src/hooks/suite';
+import { useDevice, useDispatch, useSelector } from 'src/hooks/suite';
+import {
+    selectFirmwareHashCheckErrorIfEnabled,
+    selectFirmwareRevisionCheckErrorIfEnabled,
+} from 'src/reducers/suite/suiteReducer';
 
-import { SecurityCheckFail } from './SecurityCheckFail';
+import { SecurityCheckFail, SecurityCheckFailProps } from './SecurityCheckFail';
+import { hardFailureChecklistItems, softFailureChecklistItems } from './checklistItems';
+
+const useSecurityCheckFailProps = (): Partial<SecurityCheckFailProps> => {
+    const revisionCheckError = useSelector(selectFirmwareRevisionCheckErrorIfEnabled);
+    const hashCheckError = useSelector(selectFirmwareHashCheckErrorIfEnabled);
+
+    // revision check has precedence over hash check, because it is unimpeachable
+    if (revisionCheckError !== null) {
+        return {
+            heading: 'TR_DEVICE_COMPROMISED_HEADING',
+            text: 'TR_DEVICE_COMPROMISED_FW_REVISION_CHECK_TEXT',
+            checklistItems: hardFailureChecklistItems,
+        };
+    }
+    // hash check other-error shall display softer wording than standard hash check errors
+    if (hashCheckError === 'other-error') {
+        return {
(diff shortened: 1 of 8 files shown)


packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityChecklist.tsx
```diff
@@ -9,21 +7,24 @@ type SecurityChecklistProps = {
...
-};
+export const SecurityChecklist = ({ items }: SecurityChecklistProps) => (
+    <Column
+        alignItems="flex-start"
+        gap={spacings.xl}
+        margin={{ top: spacings.xl, bottom: spacings.xxxxl }}
+    >
+        {items.map((item, index) => (
+            <Row key={index} gap={spacings.xl}>
+                {item.icon}
```
Lemonexe, line 18: Refactored because the :one: :two: icons used in the new `other-error` warning modal must have grey background _(it does looks very awkward without them)_. So let's pass the whole `ReactNode` content instead of more props..

BTW `key={index}` is not ideal, but these lists are static during component lifecycle so it's ok.
Actually I think that `key={item.icon}` was very wrong, because it's not completely out of question that a checklist would have two points with same icons, and that would break here :sweat:

peter-sanderson (approved):
LFTM, details on slack

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
```diff
@@ -3,14 +3,54 @@ import { Card } from '@trezor/components';
...
 
-import { SecurityCheckFail } from './SecurityCheckFail';
+import { SecurityCheckFail, SecurityCheckFailProps } from './SecurityCheckFail';
+import { hardFailureChecklistItems, softFailureChecklistItems } from './checklistItems';
+
+const useSecurityCheckFailProps = (): Partial<SecurityCheckFailProps> => {
+    const revisionCheckError = useSelector(selectFirmwareRevisionCheckErrorIfEnabled);
+    const hashCheckError = useSelector(selectFirmwareHashCheckErrorIfEnabled);
+
+    // revision check has precedence over hash check, because it is unimpeachable
```
komret, line 19: The comment could be a bit more descriptive, e.g.:

    // revision check has precedence over hash check, because it does not have the ambiguous other-error state

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityChecklist.tsx
vojtatranta, line 18: Okay, this caused problem and TS check didn't see it for some reason. In `SecurityCheck` there is `checklistItems` varaible, where `icon` is `string`. That's likely the reason. Coz `string` is a `ReactNode`.

Before, `string` was kinda correct here as it was the `IconName`.
```ts
<Icon size={24} name={item.icon} color={theme.legacy.TYPE_DARK_GREY} />
```

My proposal: enforce the `Icon` element here. I am not sure how to do it correctly. I can check if `<Icon />` is its own Component type and require just it, or to remove `string` from the `ReactNode` type here.

vojtatranta: I am figuring out proper fix here: https://github.com/trezor/trezor-suite/pull/16709

Lemonexe: Oops, I missed all those checklist definitions :facepalm: 
Yeah, exclude `string` from `ReactNode` could work, but would need comment to explain why...
EDIT: `ReactElement` is more suitable..

PR #16613, opened 2025-01-27
feat(suite-native): Create generic picker header component

### LINKED ISSUE #16605: Mobile Trade: Create generic picker header component
Various pickers (token, country, provider, payment method, ...) use the same base layout. Create header component, that can be reused. 

Presets:
- title
- optional reactNode representing additional embedded component 
- optional search input with text and callback returning typed text after debounce delay

Extract visual [from this design](https://www.figma.com/design/xSW1knAjH25SfOmqSdNioY/Trade?node-id=10863-52545&t=y2KpRZkyPs8PXj4Y-0)

Image

jbazant:
## Description

Introduce `PickerHeader` and `PickerCloseButton` components.

## Related Issue

Resolve #16605

## Screenshots:
Screenshot 2025-01-27 at 11 14 43

### DIFF
--- a/suite-native/module-trading/src/components/general/PickerHeader.tsx
+++ b/suite-native/module-trading/src/components/general/PickerHeader.tsx
@@ -1,0 +1,51 @@
+import { ReactNode } from 'react';
+
+import { HStack, SearchInput, Text, VStack } from '@suite-native/atoms';
+
+type PickerHeaderSearchInputProps = {
+    onSearchInputChange: (value: string) => void;
+    searchInputPlaceholder?: string;
+    isSearchInputDisabled?: boolean;
+    maxSearchInputLength?: number;
+};
+
+export type PickerHeaderProps = {
+    title: ReactNode;
+    children?: ReactNode;
+} & (
+    | { isSearchInputVisible?: false }
+    | ({ isSearchInputVisible: true } & PickerHeaderSearchInputProps)
+);
+
+const PickerHeaderSearchInput = ({
+    onSearchInputChange,
+    searchInputPlaceholder,
+    isSearchInputDisabled,
+    maxSearchInputLength,
+}: PickerHeaderSearchInputProps) => (
+    <SearchInput
+        onChange={onSearchInputChange}
+        maxLength={maxSearchInputLength}
+        placeholder={searchInputPlaceholder}
+        isDisabled={isSearchInputDisabled}
+    />
+);
+
+export const PickerHeader = ({
+    title,
+    isSearchInputVisible,
+    children,
+    ...searchInputProps
+}: PickerHeaderProps) => (
+    <VStack spacing="sp16">
+        <HStack justifyContent="space-between" alignItems="center">
+            <Text variant="titleMedium" textAlign="left">
+                {title}
+            </Text>
+            {children}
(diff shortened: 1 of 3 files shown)


suite-native/module-trading/src/components/general/PickerHeader.tsx
```diff
@@ -0,0 +1,50 @@
+import { ReactNode } from 'react';
+
+import { HStack, SearchInput, Text, VStack } from '@suite-native/atoms';
+
+export type PickerHeaderProps = {
+    title: string | ReactNode;
```
vytick: String is imo obsolete here, as the ReactNode can be string

vytick (approved):
I have one small nit, but otherwise it is nice 👍

suite-native/module-trading/src/components/general/PickerHeader.tsx
```diff
@@ -0,0 +1,50 @@
...
+    children?: ReactNode;
+} & (
+    | {
```
PeKne: nit: You might want to use utility type `RequireAllOrNone` from [type-fest](https://github.com/sindresorhus/type-fest) library that we use in our repo (for example [here](https://github.com/trezor/trezor-suite/blob/e0288286869096dc589e18b2d0ffb7615b5b9015/suite-native/atoms/src/Card/HeaderedCard.tsx#L15-L16)).

Then you can remove `isSearchInputVisible` prop, because you can compute its values inside of this component based on the fact if any of the props were passed or not.

jbazant: > You might want to use utility type `RequireAllOrNone`

I was unable to make it work with `RequireAllOrNone`. My goal was to define type
```ts
 {
    title: string | ReactNode;
    children?: ReactNode;
    isSearchInputVisible?: false;
} | {
    title: string | ReactNode;
    children?: ReactNode;
    isSearchInputVisible: true;
    onSearchInputChange: (value: string) => void;
    searchInputPlaceholder?: string;
    isSearchInputDisabled?: boolean;
    maxSearchInputLength?: number;
}
```
where only `onSearchInputChange` is required. Rest of the `SearchInput` props should be optional. To my understanding this is not possible with `RequireAllOrNone` or any other other utility from that library. I would be happy to learn that I am mistaken.

Also all those `never` types were defined just to simplify access to `SearchInput` props. I've updated that code a bit, please take a look whether you like it more.

> Then you can remove `isSearchInputVisible` prop

I actually considered to remove this prop, but decided to left it there for better readability. Do you prefer to remove it and decide solely on presence of `onSearchInputChange` prop?

suite-native/module-trading/src/components/general/PickerHeader.tsx
jbazant: Fixed

suite-native/module-trading/src/components/general/PickerHeader.tsx
PeKne: Ok, keep it as it is. I did not realize that most of the props can be undefined in both cases and only `onSearchInputChange` is required while is the `isSearchInputVisible` set to `true`. 

I'm happy with the latest version 👍.

PR #16614, opened 2025-01-27
fix: Wait for token definitions in import

vytick:
When importing account there was no await for token definitions which could cause fiat values not being shown in case fetch took too long.

This awaits token definitions to be downloaded and triggers fiat rates update to be fetched. So the fiat rates should load gradually once the import confirmation is shown.

### DIFF
--- a/suite-native/module-accounts-import/src/accountsImportThunks.ts
+++ b/suite-native/module-accounts-import/src/accountsImportThunks.ts
@@ -124,7 +124,7 @@ export const getAccountInfoThunk = createThunk<
                 if (!tokenDefinitions && isCoinWithTokens(symbol)) {
                     const definitionTypes = getSupportedDefinitionTypes(symbol);
 
-                    definitionTypes.forEach(async type => {
+                    const promises = definitionTypes.map(async type => {
                         await dispatch(
                             getTokenDefinitionThunk({
                                 symbol,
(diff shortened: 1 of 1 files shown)


suite-native/module-accounts-import/src/accountsImportThunks.ts
```diff
@@ -124,14 +124,15 @@ export const getAccountInfoThunk = createThunk<
                 if (!tokenDefinitions && isCoinWithTokens(symbol)) {
                     const definitionTypes = getSupportedDefinitionTypes(symbol);
 
-                    definitionTypes.forEach(async type => {
+                    const fetches = definitionTypes.map(async type => {
```
yanascz: Not sure `fetches` is the right form in this context. What about just `promises` or `definitionPromises`?

yanascz (approved):
I'd choose a different variable name instead of `fetches`, otherwise looks good.

suite-native/module-accounts-import/src/accountsImportThunks.ts
vytick: 53e3612116

STew790:
**QA OK**
Token definitions on import are OK both on iOS and Android (included only iOS screenshot though for better clarity)
ETH
image
POL
image
BSC
image
SOL
image

**Info**

25.1.3 9fd0808

PR #16619, opened 2025-01-27
RadioCard component

enjojoy:
## Description

New RadioCard component, used in fees

## Screenshots:

image
image

### DIFF
--- a/packages/components/src/components/RadioCard/RadioCard.tsx
+++ b/packages/components/src/components/RadioCard/RadioCard.tsx
@@ -1,0 +1,51 @@
+import { ReactNode } from 'react';
+
+import styled, { css } from 'styled-components';
+
+import { palette, borders, spacingsPx } from '@trezor/theme';
+
+import { Card } from '../Card/Card';
+import { Icon } from '../Icon/Icon';
+import { Box } from '../Box/Box';
+import { FrameProps } from '../../utils/frameProps';
+
+export type RadioCardProps = {
+    isActive: boolean;
+    children: ReactNode;
+    onClick: () => void;
+};
+
+const BorderActive = styled.div<{ isActive: boolean }>`
+    pointer-events: none;
+    position: absolute;
+    inset: 0;
+    ${({ isActive }) =>
+        isActive &&
+        css`
+            outline: ${borders.widths.large} solid ${({ theme }) => theme.borderSecondary};
+            outline-offset: -${borders.widths.large};
+        `}
+    border-radius: ${borders.radii.md};
+`;
+
+const IconWrapper = styled.div`
+    border-radius: ${borders.radii.full};
+    background-color: ${({ theme }) => theme.borderSecondary};
+    padding: ${spacingsPx.xxxs};
+`;
+
+export const RadioCard = ({ isActive, children, ...frameProps }: RadioCardProps & FrameProps) => (
+    <Box position={{ type: 'relative' }} {...frameProps}>
+        {isActive && (
+            <Box position={{ type: 'absolute', top: '-1px', right: '-1px' }} zIndex={2}>
+                <IconWrapper>
+                    <Icon name="check" size="extraSmall" color={palette.lightWhiteAlpha1000} />
+                </IconWrapper>
+            </Box>
+        )}
+        <Card paddingType="small" fillType="none">
+            {children}
+        </Card>
+        <BorderActive isActive={isActive} />
+    </Box>
+);
--- a/packages/components/src/components/RadioCard/RadioCard.stories.tsx
+++ b/packages/components/src/components/RadioCard/RadioCard.stories.tsx
@@ -1,0 +1,23 @@
+import { Meta, StoryObj } from '@storybook/react';
+
+import { RadioCard as RadioCardComponent, RadioCardProps } from './RadioCard';
+
+const meta: Meta = {
+    title: 'RadioCard',
+    component: RadioCardComponent,
+} as Meta;
+export default meta;
+
+export const RadioCard: StoryObj<RadioCardProps> = {
+    args: {
+        isActive: true,
+        children: 'Content',
+    },
+    argTypes: {
+        isActive: {
+            control: {
+                type: 'boolean',
+            },
+        },
+    },
+};
--- a/packages/components/src/index.ts
+++ b/packages/components/src/index.ts
@@ -101,3 +101,5 @@ export { getSafeWindowSize } from './utils/getSafeWindowSize';
 
 export { intermediaryTheme } from './config/colors';
 export type { SuiteThemeColors } from './config/colors';
+
+export { RadioCard } from './components/RadioCard/RadioCard';


enjojoy:
@peter-sanderson  here is the component

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,80 @@
+import styled, { css } from 'styled-components';
+
+import { palette, borders } from '@trezor/theme';
+
+import { Card } from '../Card/Card';
+import { Icon } from '../Icon/Icon';
+
+export type RadioCardProps = {
+    isActive: boolean;
+    children: React.ReactNode;
```
peter-sanderson: nit: I think we always import from React? Do we? @jvaclavik ?

peter-sanderson (commented):
Lets investigate if we can reuse `ComponentWithSubIcon` 

Otherwise fine by me :+1: 

But please, lets wait for review from @jvaclavik / @adamhavel as well

enjojoy:
> Lets investigate if we can reuse `ComponentWithSubIcon`
> 
> Otherwise fine by me 👍
> 
> But please, lets wait for review from @jvaclavik / @adamhavel as well

I wouldn't use it because it has a border and also it is not placed on the border of the component but outside of it. 

It's up to @trezor/suite-engagement  if they would like to remake it to use it in radio card as well

image

image

suite-common/icons/src/icons.ts
```diff
@@ -635,6 +635,7 @@ export const icons = {
     checkFat: require('../assets/checkFat.svg'),
     checkFatFilled: require('../assets/checkFatFilled.svg'),
     checkFilled: require('../assets/checkFilled.svg'),
+    checkMedium: require('../assets/checkMedium.svg'),
```
jvaclavik: Can you please use `checkCircleFilled` or `ComponentWithSubIcon`
(https://dev.suite.sldev.cz/components/develop/?path=/story/componentwithsubicon--component-with-sub-icon)? 🙏

We don't want to add new icons, because if we get new full export from designers, we might replace all the custom added icons by accident.

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+    inset: 0;
+    box-sizing: border-box;
+    ${({ isActive }) =>
+        isActive &&
+        css`
+            outline: 2px solid ${({ theme }) => theme.borderSecondary};
+            outline-offset: -2px;
+        `}
+    border-radius: ${borders.radii.md};
+    width: 100%;
```
seibei-iguchi: Not needed, inset is enough

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+    height: 100%;
+    z-index: 1;
+`;
+
+const IconWrapper = styled.div`
```
seibei-iguchi, line 31: Not needed, use Box component

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+    position: absolute;
+    top: 2px;
+    right: 2px;
+    z-index: 2;
+`;
+
+const RadioCardWrapper = styled.div<{ isActive: boolean }>`
```
seibei-iguchi: Not needed, use Box component

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+    position: relative;
+    width: 100%;
+    height: 100%;
+    border-radius: ${borders.radii.md};
+
+    ${({ isActive }) =>
```
seibei-iguchi: Custom css not needed, just use IconCircle component

seibei-iguchi (changes requested):

packages/components/src/components/RadioCard/RadioCard.tsx
enjojoy: I will use it instead of IconCircle as agreed on slack

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+export type RadioCardProps = {
+    isActive: boolean;
+    children: ReactNode;
+    onClick: () => void;
+};
+
+const BorderActive = styled.div<{ isActive: boolean }>`
+    position: absolute;
+    inset: 0;
+    box-sizing: border-box;
```
seibei-iguchi: I don't think you need this?

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+    ${({ isActive }) =>
+        isActive &&
+        css`
+            outline: 2px solid ${({ theme }) => theme.borderSecondary};
```
seibei-iguchi: Use const from borders.widths variable

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Box } from '../Box/Box';
+
+export type RadioCardProps = {
+    isActive: boolean;
+    children: ReactNode;
+    onClick: () => void;
+};
+
+const BorderActive = styled.div<{ isActive: boolean }>`
+    position: absolute;
```
seibei-iguchi, line 20: You also need to add pointer-events: none

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+const IconWrapper = styled.div`
+    position: absolute;
+    top: -1px;
+    right: -1px;
+    z-index: 2;
+    border-radius: ${borders.radii.full};
+    background-color: ${({ theme }) => theme.borderSecondary};
+    height: 14px;
```
seibei-iguchi: I'd rather not set dimensions like this. Just set the right icon size and only use padding here. Then you won't need the flex properties below.

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+        css`
+            outline: 2px solid ${({ theme }) => theme.borderSecondary};
+            outline-offset: -2px;
+        `}
+    border-radius: ${borders.radii.md};
+    z-index: 1;
+`;
+
+const IconWrapper = styled.div`
+    position: absolute;
```
seibei-iguchi: You can extract these position and z-index props into a separate Box component

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+    background-color: ${({ theme }) => theme.borderSecondary};
+    height: 14px;
+    width: 14px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+`;
+
+export const RadioCard = ({ isActive, children }: RadioCardProps) => (
+    <Box position={{ type: 'relative' }} width="100%" height="100%">
```
seibei-iguchi: I don't think the dimensions are not needed?

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+        {isActive && (
+            <IconWrapper>
+                <Icon name="checkMedium" size="extraSmall" color={palette.lightWhiteAlpha1000} />
```
seibei-iguchi: What's the point of the new checkMedium?

seibei-iguchi: I think you should use theme, not palette directly.

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+            </IconWrapper>
+        )}
+        <BorderActive isActive={isActive} />
+        <Card width="100%" paddingType="small" flex="1" fillType="none">
```
seibei-iguchi: The width is not needed, the flex too.

seibei-iguchi (changes requested):

packages/components/src/components/RadioCard/RadioCard.tsx
enjojoy: What do you mean? It's a new icon that I imported from figma

packages/components/src/components/RadioCard/RadioCard.tsx
enjojoy: Then I guess I'll add dimensions as an optional prop? They're needed here for sure 
image

packages/components/src/components/RadioCard/RadioCard.tsx
enjojoy: I didn't find the suitable checkmark in the repo, the one that is already there is too thin. Is there a way to make it thicker?

enjojoy: But then it will be black on the light theme? I want it to be always white

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+    z-index: 1;
+`;
+
+const IconWrapper = styled.div`
+    border-radius: ${borders.radii.full};
+    background-color: ${({ theme }) => theme.borderSecondary};
+    padding: 2px;
+`;
+
+export const RadioCard = ({ isActive, children }: RadioCardProps) => (
```
jvaclavik: nitpick: It would be nice to add also `frameProps`, but we can do it later on our own.

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const BorderActive = styled.div<{ isActive: boolean }>`
+    pointer-events: none;
+    position: absolute;
+    inset: 0;
+    ${({ isActive }) =>
+        isActive &&
+        css`
+            outline: ${borders.widths.large} solid ${({ theme }) => theme.borderSecondary};
+            outline-offset: -${borders.widths.large};
```
jvaclavik, line 26: It is OK to use outline? IMHO yes. 

@adamhavel What's your opinion here?

packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+        `}
+    border-radius: ${borders.radii.md};
+    z-index: 1;
```
jvaclavik: This is not necessary if you move `BorderActive` below `Card`:

```tsx
        <Card paddingType="small" fillType="none">
            {children}
        </Card>
        <BorderActive isActive={isActive} />

File: packages/components/src/components/RadioCard/RadioCard.tsx [hunk 18]
```diff
@@ -0,0 +1,55 @@
...
+`;
+
+const IconWrapper = styled.div`
+    border-radius: ${borders.radii.full};
+    background-color: ${({ theme }) => theme.borderSecondary};
+    padding: 2px;
```
jvaclavik: Could you use `spacingsPx.xxxs`? 🙏

jvaclavik (commented):
Good job, Albina! 💪  Just stuff to polish and ready to merge!

suite-common/icons/src/icons.ts
enjojoy: I tried them all before but they're just not as good as the one on figma... I will discuss it with the designer

packages/components/src/components/RadioCard/RadioCard.tsx
seibei-iguchi: I still don't get it:( Why w
Do you need them in this context? Width is 100% by default, or just use Column with stretch align. And height is dynamic based on content anyway?

enjojoy: I think this makes sense, didn't think of the stretch. Thank you

packages/components/src/components/RadioCard/RadioCard.tsx
enjojoy: UPD: we agreed to use the small, already existing check icon

suite-common/icons/src/icons.ts
enjojoy: Will use the existing thin checkmark as agreed with designer

jvaclavik: Thanks!

packages/components/src/components/RadioCard/RadioCard.tsx
enjojoy: Thanks, it's a nice catch

seibei-iguchi (approved):

enjojoy:
I noticed a leftower type that is not needed anymore, will add a fix now

bosomt:
**QA OK**

**Info:**
 - OS: MacIntel
 - Screen: 1470x956

PR #16621, opened 2025-01-27
Better UI for Destination tag in the send form

### LINKED ISSUE #16513: Ripple memo / destination tag in send and receive form
Improve UI of Destination tag and and rename it to `Memo/Destination tag`

Problems 
- Destination tag field is hard to find and confusing to users(feedback from users)

Requirements
- Send form 
  - update the destination tag UI according to the following design: https://www.figma.com/design/jsdvZlsY36rCc1E6PoItcz/Global-actions?node-id=4359-12035&t=nb8QE71sYXymuNNy-1
  - make the destination tag input visible without clicking
  - add a toggle button to disable the destination tag input 
  - If the toggle is enabled (default state) - the destination tag input field should be required if the toggle button is on. Users cannot proceed further with send without including the destination tag. 
  - if the toggle button is disabled -  then the input for destination tag should be hidden and the field should no longer be required. instead of the input we should display the warning message (see design file)
- Receive form
  - display an info banner about the destination tag on top of the receive form

izmy:
## Description

Improve **Destination tag** input visibility in the send form based on updated UI design

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16513

## Screenshots:

image

image

image

### Before

image

image

### DIFF
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -5467,6 +5467,20 @@ export default defineMessages({
         defaultMessage: 'Memo/Tag',
         id: 'DESTINATION_TAG_SHORT',
     },
+    DESTINATION_TAG_SWITCH: {
+        defaultMessage: 'Add Memo/Destination tag',
+        id: 'DESTINATION_TAG_SWITCH',
+    },
+    DESTINATION_TAG_NOTE: {
+        defaultMessage:
+            'Online exchanges require this to identify your account. Get your destination tag from your exchange.',
+        id: 'DESTINATION_TAG_NOTE',
+    },
+    DESTINATION_TAG_BANNER_SEND: {
+        defaultMessage:
+            'Most exchanges require you to include a memo/destination tag to identify your Ripple account. Make sure you really don’t need it.',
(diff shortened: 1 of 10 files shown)


packages/suite/src/support/messages.ts
```diff
@@ -5462,6 +5462,14 @@ export default defineMessages({
         defaultMessage: 'Destination tag',
```
tomasklim:
```suggestion
        defaultMessage: 'Memo/Destination tag',
```

jvaclavik:
Screenshot 2025-01-29 at 9 59 47

jvaclavik:
Btw could you please add for next PRs screenshots "before" and "after"? 🙏  it's easier to understand what actually changed and how it looked like before. Thanks!

izmy:
> Btw could you please add for next PRs screenshots "before" and "after"? 🙏 it's easier to understand what actually changed and how it looked like before. Thanks!

Sure, I added "before".

izmy:
> Screenshot 2025-01-29 at 9 59 47

Done... I added `gap` as props to `<Note />`

tomasklim (approved):

PR #16632, opened 2025-01-27
feat(suite-common): Introduce UnreachableCaseError helper

### LINKED ISSUE #16396: Exhaustive switches with TS
Let's say you have a switch statement that handles all values of an enum (or some other type with a limited number of possible values). When you extend that enum sometime in the future you may forget to add a new case to the statement.

With typescript this can be checked with code like this.
```ts
export class UnreachableCaseError extends Error {
  constructor(val: never, message = 'Unreachable case') {
    super (`${message}: [${JSON.stringify(val)}]`);
  }
}
```

There is no need for special es-lint rules or anything like that, just pure TS.

### With enums
```ts
export enum AppTabsRoutes {
    HomeStack = 'HomeStack',
    AccountsStack = 'AccountsStack',
    Settings = 'Settings',
}

function logTab(tab: AppTabsRoutes) {
    switch (tab) {
        case AppTabsRoutes.HomeStack:
            console.log('HomeStack');
            break;
        case AppTabsRoutes.AccountsStack:
            console.log('AccountsStack');
            break;
        default:
            throw new UnreachableCaseError(tab);
    }
}
```
TS throws `TS2345: Argument of type 'import("/Users/jiribazant/Library/Application Support/JetBrains/WebStorm2024.3/scratches/scratch").AppTabsRoutes' is not assignable to parameter of type 'never'.` on line with `throw new UnreachableCaseError(tab);`

### With string literals
```ts
type ButtonSize = 'extraSmall' | 'small' | 'medium' | 'large';

function logButtonSize(size: ButtonSize) {
    switch (size) {
        case 'small':
            console.log('Small');
            break;
        case 'medium':
            console.log('Medium');
            break;
        case 'large':
            console.log('Large');
            break;
        default:
            throw new UnreachableCaseError(size);
    }
}
```
TS throws `TS2345: Argument of type 'string' is not assignable to parameter of type 'never'.`

## Pros
- Static check at compile time. (And IDEs as well)

## Cons
- Need of manual adaptation to legacy code

jbazant:
## Description

Introduces helper, that allows to use exhaustive switches as described in #16396.

## Related Issue

Resolve #16396

### DIFF
--- a/suite-common/suite-utils/src/UnreachableCaseError.ts
+++ b/suite-common/suite-utils/src/UnreachableCaseError.ts
@@ -1,0 +1,5 @@
+export class UnreachableCaseError extends Error {
+    constructor(val: never, message = 'Unreachable case') {
+        super(`${message}: [${JSON.stringify(val)}]`);
+    }
+}


vytick (approved):
nice improvement 🧩

PR #16585, opened 2025-01-24
Fix(suite): FW hash check Sentry reporting

Lemonexe:
## Description

- solve broken errorPayloads due to bad serialization
- only report `other-error` on attempt 3 (meaning retries maxed), not errors received in meantime
- log FW revision|hash check errors|warnings to console
  - :warning: must use `console.warn`, not `.error`, as Sentry'd pick it up and send a duplicate event :see_no_evil:  
- cleanup in variable names, I'm also intentionally changing the `contextData` object that is sent in event message, see [comment](https://github.com/trezor/trezor-suite/pull/16585#discussion_r1928558299)

## Related Issue

Resolve #16584

## Screenshots:

N/A, but see illustrative Sentry events from localhost.
Sent from testing branch https://github.com/trezor/trezor-suite/commit/e0668b8a810c99bb67697850660eeb6db50b5ef5 . You can try for yourself - comment/uncomment to pick from the 3 scenarios.

For all cases see trace: that there is only FW revision failure alongside, but no other FW hash check events

- [1st other-error, then success](https://satoshilabs.sentry.io/issues/6241035716/events/318b41816fd240dbba84042e7158d32c/?project=5193825), see warningPayload at the bottom
- [other-error maxed out](https://satoshilabs.sentry.io/issues/6240996662/events/f486f6aaf78d40549d331486f93b28da/), see errorPayload at the bottom
- [1st other-error, then hash-mismatch](https://satoshilabs.sentry.io/issues/6240996891/events/cdf804c96769496681aaad3bce6233ac/?project=5193825). Only theoretical concern: it means a counterfeit device that is flaky

PR #16652, opened 2025-01-28
Fix/fw hash check error payload serialization

### LINKED ISSUE #16584: FW hash check fix Sentry reporting
- correctly append `errorPayload` to sentry errors/warnings (currently broken!)
- only report `other-error` on attempt 3 (meaning retries maxed), not errors received in meantime
- also log FW revision|hash check errors|warnings to console

Lemonexe:
## Description

Followup to #16585 to reliably log FW hash check errors to Sentry

- create `serializeError` helper which correctly detects how the caught error should be serialized
- use it on FW hash check error payload in connect

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16584

## Logs:

I placed `throw ERRORS.TypedError('Runtime', 'Mock Error Description')` [at this line](https://github.com/trezor/trezor-suite/blob/108fb3198435c60e1f6993e651e04b63eeb858cd/packages/connect/src/device/Device.ts#L828).
Console logging `errorPayload.toString()` in the `catch (errorPayload)` block returns:
```
Runtime Mock Error Description
```
Which means that `TypedError` responds to `.toString()` as it would be expected from an `Error` inheritor :heavy_check_mark: :ok_hand: 

But before this change, an `Error` instance gets mangled somewhere between Connect and Suite Redux.
Looking into redux `state.device.selectedDevice.authenticityChecks.firmwareHash.errorPayload` :confused: 
```
{ "code": "Runtime" }
```
After this change it is :slightly_smiling_face: 
```
"Error: Mock Error Description"
```

### DIFF
--- a/packages/utils/tests/serializeError.test.ts
+++ b/packages/utils/tests/serializeError.test.ts
@@ -1,0 +1,30 @@
+import { serializeError } from '../src/serializeError';
+
+class CustomError extends Error {
+    somethingExtra = 'extra stuff';
+
+    toString() {
+        return `${super.toString()} + ${this.somethingExtra} `;
+    }
+}
+
+describe(serializeError.name, () => {
+    it('serializes an Error instance', () => {
+        const error = new Error('example message');
+        expect(serializeError(error)).toBe('Error: example message');
+
+        const customError = new CustomError('example message');
+        expect(serializeError(customError)).toBe('Error: example message + extra stuff ');
+    });
+
+    it('serializes a plain object', () => {
+        const error = { message: 'test' };
+        expect(serializeError(error)).toBe(JSON.stringify(error));
+    });
+
+    it('serializes a primitive', () => {
+        expect(serializeError('test')).toBe('test');
+        expect(serializeError(123)).toBe('123');
+        expect(serializeError(false)).toBe('false');
+    });
+});
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -8,6 +8,7 @@ import {
     TypedEmitter,
     createTimeoutPromise,
     isArrayMember,
+    serializeError,
 } from '@trezor/utils';
 import { Session } from '@trezor/transport';
 import { TransportProtocol, v1 as v1Protocol } from '@trezor/protocol';
@@ -831,7 +832,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
             return { success: true };
         } catch (errorPayload) {
-            return createFailResult('other-error', errorPayload);
+            return createFailResult('other-error', serializeError(errorPayload));
         }
     }
 
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -46,6 +46,7 @@ export * from './promiseAllSequence';
 export * from './redactUserPath';
 export * from './resolveAfter';
 export * from './scheduleAction';
+export * from './serializeError';
 export * from './splitStringEveryNCharacters';
 export * from './throttler';
 export * from './throwError';
--- a/packages/utils/src/serializeError.ts
+++ b/packages/utils/src/serializeError.ts
@@ -1,0 +1,15 @@
+/**
+ * Serialize an error of unknown type to a string.
+ */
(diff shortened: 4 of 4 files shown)


packages/utils/tests/serializeError.test.ts
```diff
@@ -0,0 +1,25 @@
+import { serializeError } from '../src/serializeError';
+
+class CustomError extends Error {
+    somethingExtra = 'foobar';
+}
+
+describe('serializeError', () => {
```
peter-sanderson: :heart_eyes:

peter-sanderson: nit/tip:

```suggestion
describe(serializeError.name, () => {
```

packages/utils/tests/serializeError.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    it('serializes an Error instance', () => {
+        const error = new Error('test');
+        expect(serializeError(error)).toBe('Error: test');
+        const customError = new CustomError('test');
+        expect(serializeError(customError)).toBe('Error: test');
```
peter-sanderson: So there's no way how to serialize with all values (`somethingExtra`) as well?

peter-sanderson (approved):

packages/utils/tests/serializeError.test.ts
Lemonexe: unless `CustomError` overrides `.toString()`, which by default only serializes `code` and `message` as it comes from `Error` prototype.

But this is a great idea - for the sake of test I'll add that to the `CustomError`, so that the test code highlights what are the possibilites :+1:

peter-sanderson: I have consulted this with my AI overlords and they revealed this to me:

```js
function serializeError(error) {
    if (!(error instanceof Error)) {
        throw new TypeError('Input must be an instance of Error');
    }

    const serializedError = {};

    // Capture all own properties, including non-enumerable ones like 'message' and 'name'
    Object.getOwnPropertyNames(error).forEach(prop => {
        serializedError[prop] = error[prop];
    });

    return JSON.stringify(serializedError);
}

// Example usage:
class CustomError extends Error {
    constructor(message) {
        super(message);
        this.somethingExtra = 'foobar';
    }
}

const myError = new CustomError('Something went wrong');
const serialized = serializeError(myError);
console.log(serialized); // {"message":"Something went wrong","somethingExtra":"foobar","name":"CustomError"}
```

peter-sanderson: I have put this to test, and it seems, it works.

peter-sanderson: image

Lemonexe: I beg to disagree with AI overlords :-1: 
error in `catch(error)` can be absolutely anything  - not just _the error we are throwing_, but any part of our tech stack can throw. We should expect it to be any type, not just the standard `Error` instance.

This function not only assumes that it _must_ be `Error` instance - but it even rethrows another error if it isn't. If you call `serializeError` in a catch block, it rethrows just because it doesn't know how to stringify it, so `throw { obj: 'this is a POJO' }` would break this.

I believe (but this is just IMO) that `.toString()` should be used, as it is supposed to [represent it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString).
If it's a custom error class, well it should be its responsibility to serialize what is appropriate ( = not necessarily every property)

peter-sanderson: Can we have/do both? Yes it can be anything. But this approach (utilizing `Object.getOwnPropertyNames` shall be done in `is_object(...)` if-branch only.

This is for Sentry and we can get any mysterious 3rd party library error/object/whatever. Having truly full serialization would provide us a much needed context in such case.

mroz22 (approved):
good and nice

packages/utils/tests/serializeError.test.ts
Lemonexe: :arrow_right: will be solved in next PR, as it is still unclear what is the _best_ solution.

PR #16680, opened 2025-01-28
fix(blockchain-link): stop ripple ping timeout if api disconnected

szymonlesisz:
## Description

`yarn workspace @trezor/blockchain-link test:unit --coverage=false`

ends with:
```
A worker process has failed to exit gracefully and has been force exited.
This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.
Active timers can also cause this, ensure that .unref() was called on them.
```

there were 2 reasons:
1. setPingTimeout is calling itself recursively. now it should stop if api is disconnected
2.  worker.dispose() is async (at least in ripple worker) and it is awaited in tests teardown, but it was never awaited in baseWorker while processing the request

### DIFF
--- a/packages/blockchain-link/src/workers/baseWorker.ts
+++ b/packages/blockchain-link/src/workers/baseWorker.ts
@@ -163,7 +163,7 @@ export abstract class BaseWorker<API> {
             return true;
         }
         if (data.type === MESSAGES.DISCONNECT) {
-            this.disconnect();
+            await this.disconnect();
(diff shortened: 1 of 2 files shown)


mroz22:
good, it even fixes the "exceeded listeners" issue in tests. I just tested using setup in this commit https://github.com/trezor/trezor-suite/pull/15996/commits/5bcc88b5a58807e6e24f1af3117d8516b88be9ab + --detectOpenHandles args

mroz22 (approved):

PR #16709, opened 2025-01-29
fix(suite): enforce that the Security Check List icon is NOT just a s…

### LINKED ISSUE #16692: Missing info icon during onboarding device check
**Describe the bug**
Missing info icon during onboarding device check

**Steps to reproduce:**
1. a
2. b
3. c

**Info:**
 - OS: MacIntel
 - Screen: 1470x956

**Screenshots:**

Image

vojtatranta:
…tring

## Description

- an imperfect react type caused that needed updates were overlooked
- we improved the typing and fixed the cases so that it shouldn't happen again (altough, `<div />` still passes)

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16692
## Screenshots:
Before
Screenshot 2025-01-29 at 10 34 38 AM
After
Screenshot 2025-01-29 at 10 34 42 AM

## Summary by CodeRabbit

- **New Features**
	- Enhanced type safety for security checklist items
	- Improved icon rendering in security checklists

- **Tests**
	- Added type validation tests for security checklist items

- **Refactor**
	- Updated icon type definitions to be more specific
	- Replaced string-based icon references with React component icons

### DIFF
--- a/packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
+++ b/packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
@@ -27,6 +27,7 @@ import { selectSuiteFlags } from 'src/reducers/suite/suiteReducer';
 
 import { DeviceAuthenticity } from './DeviceAuthenticity';
 import { SecurityChecklist } from './SecurityChecklist';
+import { SecurityChecklistItem } from './types';
 
 const StyledCard = styled(CollapsibleOnboardingCard)`
     max-width: 840px;
@@ -90,15 +91,15 @@ const TooltipWrapper = styled.span`
 
 const firmwareInstalledChecklist = [
     {
-        icon: 'info',
+        icon: <Icon size={24} name="info" />,
         content: <Translation id="TR_ONBOARDING_DEVICE_CHECK_4" />,
     },
-] as const;
+] as const satisfies SecurityChecklistItem[];
 
 const getNoFirmwareChecklist = (isMobileLayout: boolean) =>
     [
         {
-            icon: 'verified',
+            icon: <Icon size={24} name="verified" />,
             content: (
                 <Translation
                     id="TR_ONBOARDING_DEVICE_CHECK_2"
@@ -118,7 +119,7 @@ const getNoFirmwareChecklist = (isMobileLayout: boolean) =>
             ),
         },
         {
-            icon: 'hologram',
+            icon: <Icon size={24} name="hologram" />,
             content: (
                 <Translation
                     id="TR_ONBOARDING_DEVICE_CHECK_1"
@@ -139,10 +140,10 @@ const getNoFirmwareChecklist = (isMobileLayout: boolean) =>
             ),
         },
         {
-            icon: 'package',
+            icon: <Icon size={24} name="package" />,
             content: <Translation id="TR_ONBOARDING_DEVICE_CHECK_3" />,
         },
-    ] as const;
+    ] as const satisfies SecurityChecklistItem[];
 
 type SecurityCheckContentProps = {
     goToDeviceAuthentication: () => void;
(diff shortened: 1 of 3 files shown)


packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
```diff
@@ -88,17 +89,17 @@ const TooltipWrapper = styled.span`
...
+const firmwareInstalledChecklist: readonly SecurityChecklistItem[] = [
     {
-        icon: 'info',
+        icon: <Icon size={24} name="info" />,
         content: <Translation id="TR_ONBOARDING_DEVICE_CHECK_4" />,
     },
 ] as const;
 
-const getNoFirmwareChecklist = (isMobileLayout: boolean) =>
+const getNoFirmwareChecklist = (isMobileLayout: boolean): readonly SecurityChecklistItem[] =>
```
vojtatranta: @Lemonexe why you put `readonly` there? I don't know how that works, so I copied your thing here.

Lemonexe: I also just copied the same thing  from before in https://github.com/trezor/trezor-suite/commit/946be5d11bb0259e53a64d71f3a942a58976e8a4, originally done in https://github.com/trezor/trezor-suite/commit/5e6071801df2315155f307ca705f99e00bca521c

IMO it does make sense though.. Restricting does not directly help anything, but makes code well readable - we can see that `SecurityChecklistProps` is supposed to have it read only. But not important :shrug:

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
```diff
@@ -88,17 +89,17 @@ const TooltipWrapper = styled.span`
     position: relative;
 `;
 
-const firmwareInstalledChecklist = [
+const firmwareInstalledChecklist: readonly SecurityChecklistItem[] = [
```
Lemonexe: Tiny nit: in _this_ case, `readonly` doesn't do anything because `as const` has precedence.
BTW declarative type actually _does not declare_ the type when used with `as const`, it acts only as typecheck, but the variable will have the narrower, const type. 
I'd instead suggest `as const satisfies SecurityChecklistItem[]`, it has the same function but it's clearer.

packages/suite/src/views/onboarding/steps/SecurityCheck/types.ts
```diff
@@ -1,7 +1,9 @@
-import { ReactNode } from 'react';
+import { ReactElement, ReactNode } from 'react';
+
+import { Icon } from '@trezor/components';
 
 export type SecurityChecklistItem = {
-    icon: ReactNode;
+    icon: ReactElement<typeof Icon>;
```
Lemonexe: The `<typeof Icon>` has only informative value, but why not :shrug:

Lemonexe (approved):
LGTM, thanks for fixing my mess :bow:

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
Lemonexe: FYI if an object-type variable is declared `readonly`, it means that TS throws e.g. on `items.pop()` or `items[1] = whatever` , because it mutates the array. Though it does not prevent you from mutating array members, because technically that does not mutate the object.

packages/suite/src/views/onboarding/steps/SecurityCheck/SecurityCheck.tsx
vojtatranta: fixed according to you

vojtatranta:
added some type tests, unfortunately, we haven't find a way to prevent others from passing stuff like `<div />` as `<Icon />`

Lemonexe (approved):

packages/suite/src/views/onboarding/steps/SecurityCheck/__tests__/types.test.tsx
komret, line 1: Is this necessary now that you narrowed the type?

vojtatranta: Well, the aim was to ensure that this won't happen again. More tests will hardly hurt us, I guess.

PR #16711, opened 2025-01-29
Feat/entropy check UI

komret:
- 1st commit - adding entropyCheck to settings and cleanup
 As checks are being introduced gradually, refactoring is necessary to keep the API reasonable. I added a new `enabledSecurityChecks` property to `suite.settings` which is a dictionary of checks with boolean values for enabled/disabled state. I did not create a migration because users shouldn't generally have any of the checks disabled. I only added the entropy check toggle to debug settings because there's probably no use case for disabling it other than playing with fake devices.
- 2nd commit removes temporary code we used in 25.1 to skip entropy check on the second try of creating a wallet - we did that to avoid false positive results during roll-out
- 3rd commit - filtering errors and adding a failed state for device
We must use app state here because we need to persist the check result in case user disconnects and reconnects the device. This commit is temporary and will be expanded upon in the future. We must be careful during the roll-out to avoid false positives. In the future, we should improve error filtering and avoid relying on device id when detecting fake devices.
- 4th commit - UI
We reuse the "ghost modal" but omit the "back" button because in this case, the user's wallet should have no history and they won't need to access it to save their funds.
- 5th commit - DB
We must somehow persist the information about failed check in case user restarts Suite because in that case, they shouldn't repeat the check. This is because 1) the check is probabilistic and 2) wallet may have already been created despite the check failing. Again, we should avoid relying on device id in the future. Note that I decreased the DB version number to 52 because it is not necessary to increase it multiple times between releases.
- 6th commit allows disabling the feature via message system in case we screw up and there are false positives

Resolves https://github.com/trezor/trezor-suite-private/issues/123


submodules/trezor-common
peter-sanderson, line 1: accidentally committed

packages/suite/src/views/settings/SettingsDebug/CheckFirmwareAuthenticity.tsx
```diff
@@ -3,47 +3,66 @@ import { Switch } from '@trezor/components';
...
+    const toggleEntropyCheck = (isChecked?: boolean) =>
+        dispatch({
+            type: SUITE.TOGGLE_ENTROPY_CHECK,
+            payload: !!isChecked,
+        });
+    const toggleFirmwareHashCheck = (isChecked?: boolean) =>
         dispatch({
-            type: SUITE.DEVICE_FIRMWARE_REVISION_CHECK,
-            payload: { isDisabled: !newEnabledState },
+            type: SUITE.TOGGLE_FIRMWARE_HASH_CHECK,
```
peter-sanderson, line 25: `TOGGLE` -> `SET` when we send value as well, toggle evokes just inverting the state, but this will set it to new value send via payload

peter-sanderson: optional :point_up: after all, I find it okish, but I slightly prefer `SET_IS_ENABLED` despite its longer for clarity

packages/suite/src/views/settings/SettingsDebug/CheckFirmwareAuthenticity.tsx
```diff
@@ -3,47 +3,66 @@ import { Switch } from '@trezor/components';
...
     const dispatch = useDispatch();
-    const { isFirmwareHashCheckDisabled, isFirmwareRevisionCheckDisabled } = useSelector(
-        state => state.suite.settings,
-    );
+    const isEntropyCheckEnabled = useSelector(selectIsEntropyCheckEnabled);
+    const isFirmwareHashCheckEnabled = useSelector(selectIsFirmwareHashCheckEnabled);
+    const isFirmwareRevisionCheckEnabled = useSelector(selectIsFirmwareRevisionCheckEnabled);
 
-    const onToggleFirmwareRevisionCheck = (newEnabledState?: boolean) =>
+    const toggleEntropyCheck = (isChecked?: boolean) =>
```
peter-sanderson: I think `Switch` component shall not have this optional
image

The optionality of the `isChecked` in `onChange` seems as mistake to me as it is always provided by the component

packages/suite/src/views/settings/SettingsDebug/CheckFirmwareAuthenticity.tsx
```diff
@@ -3,47 +3,66 @@ import { Switch } from '@trezor/components';
...
+        dispatch({
+            type: SUITE.TOGGLE_ENTROPY_CHECK,
+            payload: !!isChecked,
```
peter-sanderson: After fix in `Switch` component the ugly `!!` type-cast shall not be needed

komret: 39452c5f5b976fcaf37a6449fd3abe7437ba7508

packages/suite/src/views/settings/SettingsDebug/CheckFirmwareAuthenticity.tsx
komret: 39452c5f5b976fcaf37a6449fd3abe7437ba7508

packages/suite/src/actions/settings/deviceSettingsActions.ts
```diff
@@ -186,6 +186,14 @@ export const resetDevice =
                     vendor,
                     error: result.payload.error,
                 });
+                const hardErrors: string[] = [
```
peter-sanderson, line 193: nit: this can be defined as constant outside of the function, but its temporary anyway

packages/suite/src/components/suite/Preloader/__tests__/Preloader.test.tsx
```diff
@@ -37,33 +37,22 @@ jest.mock('@trezor/env-utils', () => ({
...
-// jest.mock('lottie-react', () => ({
-//     // Lottie: ({ trezorVersion }: any) => <div>{trezorVersion}</div>,
-//     __esModule: true, // export as module
-//     default: ({ trezorVersion }: any) => <div data-testid="lottie-image">{trezorVersion}</div>,
-// }));
-// jest.mock('react-use/lib/useMeasure', () => ({
-//     // Lottie: ({ trezorVersion }: any) => <div>{trezorVersion}</div>,
-//     __esModule: true, // export as module
-//     default: () => ['ref', { height: 0 }],
-// }));
```
peter-sanderson, line 58: :clap:

packages/suite/src/components/suite/Preloader/__tests__/Preloader.test.tsx
```diff
@@ -501,4 +490,82 @@ describe('Preloader component', () => {
...
+                authenticityChecks: {
+                    firmwareRevision: {
+                        error: 'revision-mismatch',
+                    },
+                },
+                features: {},
+            },
+        };
+
+        const store = initStore(
```
peter-sanderson: maybe nit but it can be inline very easily and AFAIK it adds to the readability `const store = initStore(getInitialState({ device }));`

packages/suite/src/components/suite/Preloader/__tests__/Preloader.test.tsx
```diff
@@ -501,4 +490,82 @@ describe('Preloader component', () => {
...
+            store,
+            <Index app={store.getState().router.app} />,
+        );
+
+        expect(getByText('TR_DEVICE_COMPROMISED_FW_HASH_CHECK_TEXT')).not.toBeNull();
+
+        unmount();
+    });
+
+    it('Failed firmware revision check', () => {
```
peter-sanderson: Those 3 tests are basically copy-paste and could be easily replaced by data-provider

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
```diff
@@ -12,10 +12,19 @@ import {
...
 import { hardFailureChecklistItems, softFailureChecklistItems } from './checklistItems';
 
-const useSecurityCheckFailProps = (): Partial<SecurityCheckFailProps> => {
+const useSecurityCheckFailProps = (
+    isEntropyCheckFailed: boolean,
+): Partial<SecurityCheckFailProps> => {
     const revisionCheckError = useSelector(selectFirmwareRevisionCheckErrorIfEnabled);
     const hashCheckError = useSelector(selectFirmwareHashCheckErrorIfEnabled);
 
+    if (isEntropyCheckFailed) {
```
peter-sanderson, line 31: I think this shall be taken from state and not passed as prop. Or am I missing something?

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -93,13 +98,15 @@ export const Preloader = ({ children }: PropsWithChildren) => {
...
 
+    const isEntropyCheckEnabledAndFailed = isEntropyCheckEnabled && isEntropyCheckFailed;
+
     if (
         (router.route?.app === undefined ||
             !ROUTES_TO_SKIP_FIRMWARE_CHECK.includes(router.route?.app)) &&
-        !isFirmwareAuthenticityCheckDismissed &&
-        isFirmwareCheckFailed
+        ((!isFirmwareAuthenticityCheckDismissed && isFirmwareCheckFailed) ||
+            isEntropyCheckEnabledAndFailed)
```
peter-sanderson, line 103: I think this shall be handled together with other errors so we have only one place you have to change when adding a new error

packages/suite/src/components/suite/Preloader/Preloader.tsx
```diff
@@ -70,6 +70,9 @@ export const Preloader = ({ children }: PropsWithChildren) => {
         selectIsFirmwareAuthenticityCheckDismissed,
     );
     const isEntropyCheckEnabled = useSelector(selectIsEntropyCheckEnabled);
+    const isEntropyCheckDisabledByMessageSystem = useSelector(state =>
```
peter-sanderson: I think this shall be handled in `selectFirmwareRevisionCheckErrorIfEnabled` together with other errors of this kind

packages/components/src/components/form/Switch/Switch.tsx
```diff
@@ -114,7 +114,7 @@ const Label = styled.label<{
 export interface SwitchProps {
     isChecked: boolean;
     label?: ReactNode;
-    onChange: (isChecked?: boolean) => void;
+    onChange: (isChecked: boolean) => void;
```
peter-sanderson, line 117: There are also other places where this `?` was replicated. I found

- `CheckFirmwareAuthenticity` (that whe you probably copied it from)
- `DeviceAuthenticity`

peter-sanderson (changes requested):
Let discus the unification of the entropy with other errors, I think that the entropy in preloader does not need a special handling and can be treated same (in same places) as revision and  hash checks

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
komret: Yes, it's messy. I didn't want to repeat the logic from `Preloader`, bbut perhaps it should be abstracted into a selector.

komret:
Added 6fb41cb5587d9b72af47bd05958484777b17755b to actually disable the check, not just the UI. This is necessary in case there is a bug in the check itself which might block wallet creation.n UI must also be disabled in case the state is persisted.

packages/suite/src/views/settings/SettingsDebug/CheckFirmwareAuthenticity.tsx
komret: Keeping this untouched due to time constraints :)

packages/suite/src/actions/settings/deviceSettingsActions.ts
komret: Skipping this as it should be very short-lived.

packages/suite/src/components/suite/Preloader/__tests__/Preloader.test.tsx
komret: It is used twice so needs to be stored in a variable.

packages/suite/src/components/suite/Preloader/__tests__/Preloader.test.tsx
komret: 9cf03fbb4387b8d70eadb663559bebe9df0b5132

packages/components/src/components/form/Switch/Switch.tsx
komret: `CheckFirmwareAuthenticity` was already handled in the previous commit. Adjusted `DeviceAuthenticity` in 9cedd2bde45ab0bb94b4b5a80a5ae5cb82764d9e.

packages/suite/src/components/suite/Preloader/Preloader.tsx
komret: The entropy check is handled differently than revision/hash check so I did it separately in `selectIsEntropyCheckEnabledAndFailed`. See c88d495653ab73067039e0ac9ad07021b69ad2bc.

packages/suite/src/components/suite/SecurityCheck/DeviceCompromised.tsx
komret: c88d495653ab73067039e0ac9ad07021b69ad2bc

packages/suite/src/components/suite/Preloader/Preloader.tsx
komret: Not perfect, but somewhat consolidated in c88d495653ab73067039e0ac9ad07021b69ad2bc.

packages/suite/src/components/suite/Preloader/__tests__/Preloader.test.tsx
```diff
@@ -118,6 +107,47 @@ const initStore = (state: State) => {
 
 const Index = ({ app }: any) => <Preloader>{app || 'foo'}</Preloader>;
 
+const deviceCompromisedFixtures = [
```
peter-sanderson, line 110: note: adding a declarative typing to the fixture helps a lot when working with it in IDE as it gives you auto-completetion

packages/suite/src/components/suite/Preloader/Preloader.tsx
peter-sanderson: :+1: lets the address it in future PR, where we introduce some abstraction to move those ifs from places like this to some implementation function that covers all the cases at one place for every check we have

peter-sanderson (approved):

STew790:
**QA OK**
I can't get past entropy check when creating a new seed with fake device.
The state is persistent even after closing the Suite window or refreshing.
This can be turned off via debug settings.
The entropy check can be also disabled via message system.

image


**Info:**
Suite version: web 25.3.0 (9ace43e88a273e863d83985469c3d92c4756028e)
Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36
OS: Linux x86_64
Screen: 1920x1080
Device: Trezor T2T1 2.8.7 regular (revision 88e1f8c7a5c7615723664c64b0a25adc0c409dee)
Transport: BridgeTransport 2.0.33

PR #16714, opened 2025-01-29
Remove @suite-native/icons jest mock

jbazant:
Mock was added in 06ce41305c83956c67a6f92f76fd60f861f8841f more then two years ago. That commit does not contain any new or updated tests. Our tests run without any obvious issue even without it. Therefore this mock seems like some kind of leftover.

On the other hand this mock complicates new tests that utilize testing-library. 

And thats why I would prefer to remove it completely.


vytick (approved):
It seems that this mock is really not useful anywhere

PR #16715, opened 2025-01-29
feat(scripts): generate-package command generates jest.config.js

jbazant:
`yarn generate-package` command generates `jest.config.js` for generated module. Also added separate template for native packages.


scripts/generatePackage.js
```diff
@@ -4,24 +4,21 @@ import sortPackageJson from 'sort-package-json';
 import prettier from 'prettier';
 import chalk from 'chalk';
 
-import templatePackageJson from './package-template/package.json';
```
Nodonisko, line 7: nit: I would rather keep this as normal imports instead because it will throw TS error in case someone move or rename something. Maybe even add it to scopes array as field then.

Nodonisko (approved):

scripts/generatePackage.js
jbazant: Updated

Nodonisko (approved):

PR #16435, opened 2025-01-17
Feat(mobile): Device Settings for FW revision check

Lemonexe:
## Description

:eye:  CR commit by commit is recommended

- regenerate icons
- revert  [#16424](https://github.com/trezor/trezor-suite/pull/16424), which was originally a prereq for this, but the design changed :see_no_evil: 
- move `DeviceSettingsCardLayout` to `suite-native/settings` as `SettingsCardWithIconLayout`
  - because it is now used in `module-settings` as well as `module-device-settings`
- add Device Checks section to Settings > Features
  - the UX flow starts here ([FeaturesSettings.tsx](https://github.com/trezor/trezor-suite/blob/d7dadb04fb8a7c005bc4d0193ab2dfd3cd31f1e2/suite-native/module-settings/src/components/FeaturesSettings.tsx#L74-L84))
  - hide it behind a new Feature Flag
- create Device checks screen
- create confirmation modal to Turn off FW authenticity check

## Related Issue

Resolve #16447 

## Screenshots:

**Settings:**
Feature flag Firmware Revision check = **off** is unchanged
Feature flag Firmware Revision check = **on**:
Settings

On press **Device Checks section** go to **Device Checks**:
image

On press **Turn off button** go to **Turn off modal**:
TurnOff1

On press **checkbox**:
TurnOff2

Back to **Advanced settings**, but Turn on the button now turns it on instantly:
DeviceChecksWhenOff

**TrezorDevices.svg** ([see comment](https://github.com/trezor/trezor-suite/pull/16435#discussion_r1934280720))
[after](https://github.com/user-attachments/assets/7a0585f8-2ddd-4e93-9f54-9d2ff6f3cbfe), [before](https://github.com/user-attachments/assets/ab0640ed-6a50-498e-9d1b-5b4bff0c69bf) (screenshots of the svg opened in firefox)

**DeviceSettings** [is unchanged](https://github.com/user-attachments/assets/ee43b55e-0e01-4a9c-91e9-798d2e3d7305)
 (refactored the `SettingsCardWithIconLayout`...)

PR #16717, opened 2025-01-30
Chore: desktop FW revision check code cleanup

Lemonexe:
## Description

Having implemented FW revision check in mobile in #16435, I noticed two tiny things to enhance in Desktop:

- update an outdated link to trezor.io
- update code comments to link with [similar section in mobile](https://github.com/trezor/trezor-suite/blob/a80b4248560a451241ccf8e7d28360b89c259ee9/suite-native/device/src/config/firmware.ts)

### DIFF
--- a/packages/urls/src/urls.ts
+++ b/packages/urls/src/urls.ts
@@ -112,7 +112,7 @@ export const HELP_CENTER_EVM_ADDRESS_CHECKSUM: Url =
 export const HELP_CENTER_EVM_SEND_TO_CONTRACT_URL =
     'https://trezor.io/support/a/where-is-my-ethereum';
 export const HELP_CENTER_FIRMWARE_REVISION_CHECK: Url =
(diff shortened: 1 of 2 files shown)


packages/urls/src/urls.ts
```diff
@@ -112,7 +112,7 @@ export const HELP_CENTER_EVM_ADDRESS_CHECKSUM: Url =
 export const HELP_CENTER_EVM_SEND_TO_CONTRACT_URL =
     'https://trezor.io/support/a/where-is-my-ethereum';
 export const HELP_CENTER_FIRMWARE_REVISION_CHECK: Url =
-    'https://trezor.io/learn/a/trezor-firmware-revision-check';
+    'https://trezor.io/learn/a/trezor-firmware-authenticity-check';
```
Lemonexe, line 115: The former redirects 308 to the latter

komret (approved):

PR #16681, opened 2025-01-28
Refactor/trading account state

adderpositive:
## Description
- delete saving whole account for modal and for trading enviroment
- delete dead code of cached account in Buy

## Related Issue 

Resolve https://github.com/trezor/trezor-suite/issues/16524

PR #16719, opened 2025-01-30
fix(trading): trading account

### LINKED ISSUE #16701: Trading - incorrect network selected when swapping
**Describe the bug**
Trading - incorrect network selected when swapping
Thread here https://satoshilabs.slack.com/archives/C04DD84P4A1/p1738149467432329

**Steps to reproduce:**
1. a
2. b
3. c

**Info:**
 - OS: MacIntel
 - Screen: 1470x956

**Screenshots:**

Image

Image

adderpositive:
## Description
Fixed a bad selected (cachec) account in the trading section (Swap, Sell)

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16701
Waiting for merging - https://github.com/trezor/trezor-suite/pull/16681

### DIFF
--- a/packages/suite/src/middlewares/wallet/tradingMiddleware.ts
+++ b/packages/suite/src/middlewares/wallet/tradingMiddleware.ts
@@ -1,6 +1,7 @@
 import { MiddlewareAPI } from 'redux';
 
 import { invityAPI } from '@suite-common/trading';
+import { ACCOUNTS_MODULE_PREFIX } from '@suite-common/wallet-core';
 import { UI } from '@trezor/connect';
 
 import { MODAL, ROUTER } from 'src/actions/suite/constants';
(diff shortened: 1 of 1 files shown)


FreeWall:
It will work, but I don't think it's the right approach. What if I go from ETH swap form directly to the POL swap form (shouldn't be possible at this time but might be in the future)? This new code won't be called so the account won't reset.

adderpositive:
> It will work, but I don't think it's the right approach. What if I go from ETH swap form directly to the POL swap form (shouldn't be possible at this time but might be in the future)? This new code won't be called so the account won't reset.

Thanks for the feedback. I have updated the code to make it more resilient. I used `${ACCOUNTS_MODULE_PREFIX}/updateSelectedAccount` to check whether `selectedAccount` (Trezor) is changed.

FreeWall (approved):
still feels a little weird to me, but I have no better fix now

FreeWall (approved):
still feels a little weird to me, but I have no better fix now

tomasklim (approved):

PR #16722, opened 2025-01-30
feat(suite-native): add DeviceCompromisedModal for FW revision check

### LINKED ISSUE #16448: DeviceCompromisedModal for FW revision check
Part of #16273: implement DeviceCompromisedModal for Firmware revision check in Suite Lite
[This figma frame](https://www.figma.com/design/eKie94qfryjTHEnR8jfmgk/Security-checks?node-id=4035-4690&t=fv9Xv7X06uaZhw6Q-4) captures the scope of this issue.

- add DeviceCompromisedModal as a separate screen
  - on close remember device as dismissed (only during session) - [see clarification in comment](https://github.com/trezor/trezor-suite/issues/16448#issuecomment-2668602890)
  - contact support button
- navigate to the modal when:
  - FW revision check results in hard failure
  - feature flag `IsFwRevisionCheckEnabled` is on
  - and when it is turned on in the settings (see #16447)

Lemonexe:
## Description

FW revision check part II.

- when device FW revision check fails & it's enabled in App settings & feature flag's enabled:
- display the Device compromised modal :ghost: 

## Related Issue

Resolve #16448

## Screenshots / QA instructions

- Connect a device with officially released FW (e.g. emulator T3T1 2.8.1)
  - there should be **no visible change** - after connecting a device dashboard is displayed normally.
- Connect emulator any model with unreleased FW (`2-main` or `1-main`) to continue.
- Turned **off** Authenticity check in App settings (setting added in [#16435](https://github.com/trezor/trezor-suite/pull/16435) ): no visible change
- Turned **off** feature flag `IsFwRevisionCheckEnabled`: no visible change 
- Turned **off** both of the above: no visible change
- Turned **on** both of the above (the default on dev env): you should see:

DeviceCompro

**Pressing the Contact Support** button opens browser with chatbot flow:
hal

**Pressing back** returns back to normal dashboard
dashboard

### DIFF
--- a/suite-native/app/package.json
+++ b/suite-native/app/package.json
@@ -59,6 +59,7 @@
         "@suite-native/module-accounts-import": "workspace:*",
         "@suite-native/module-accounts-management": "workspace:*",
         "@suite-native/module-add-accounts": "workspace:*",
+        "@suite-native/module-authenticity-checks": "workspace:*",
         "@suite-native/module-authorize-device": "workspace:*",
         "@suite-native/module-connect-popup": "workspace:*",
         "@suite-native/module-dev-utils": "workspace:*",
--- a/suite-native/app/src/navigation/RootStackNavigator.tsx
+++ b/suite-native/app/src/navigation/RootStackNavigator.tsx
@@ -9,6 +9,7 @@ import {
     AccountSettingsScreen,
 } from '@suite-native/module-accounts-management';
 import { AddCoinAccountStackNavigator } from '@suite-native/module-add-accounts';
+import { DeviceCompromisedModalScreen } from '@suite-native/module-authenticity-checks';
 import { AuthorizeDeviceStackNavigator } from '@suite-native/module-authorize-device';
 import { ConnectPopupScreen } from '@suite-native/module-connect-popup';
 import { DevUtilsStackNavigator } from '@suite-native/module-dev-utils';
@@ -79,6 +80,10 @@ export const RootStackNavigator = () => {
             <RootStack.Screen
                 name={RootStackRoutes.SettingsScreenStack}
                 component={SettingsStackNavigator}
+            />
+            <RootStack.Screen
+                name={RootStackRoutes.DeviceCompromisedModalScreen}
+                component={DeviceCompromisedModalScreen}
             />
             {/* Navigation flows that start by push from bottom animation on the first screen of its stack. */}
             <RootStack.Group screenOptions={{ animation: 'slide_from_bottom' }}>
--- a/suite-native/device/src/hooks/useHandleDeviceConnection.ts
+++ b/suite-native/device/src/hooks/useHandleDeviceConnection.ts
@@ -10,6 +10,7 @@ import {
     selectIsDeviceInitialized,
     selectIsDeviceRemembered,
     selectIsDeviceUsingPassphrase,
+    selectIsFirmwareAuthenticityCheckDismissed,
     selectIsNoPhysicalDeviceConnected,
     selectIsPortfolioTrackerDevice,
 } from '@suite-common/wallet-core';
@@ -29,7 +30,10 @@ import {
 } from '@suite-native/navigation';
 import { selectIsOnboardingFinished } from '@suite-native/settings';
 
-import { selectIsDeviceSetupSupported } from '../selectors';
+import {
+    selectHasFirmwareAuthenticityCheckHardFailed,
+    selectIsDeviceSetupSupported,
+} from '../selectors';
 
 type NavigationProp = StackToStackCompositeNavigationProps<
     AuthorizeDeviceStackParamList | RootStackParamList,
@@ -51,6 +55,16 @@ export const useHandleDeviceConnection = () => {
     const isDeviceSetupSupported = useSelector(selectIsDeviceSetupSupported);
 
     const { isBiometricsOverlayVisible } = useIsBiometricsOverlayVisible();
+
+    const hasFirmwareAuthenticityCheckHardFailed = useSelector(
+        selectHasFirmwareAuthenticityCheckHardFailed,
+    );
+    const isFirmwareAuthenticityCheckDismissed = useSelector(
+        selectIsFirmwareAuthenticityCheckDismissed,
+    );
+    const shouldNavigateToDeviceCompromisedModal =
+        hasFirmwareAuthenticityCheckHardFailed && !isFirmwareAuthenticityCheckDismissed;
+
     const navigation = useNavigation<NavigationProp>();
     const dispatch = useDispatch();
 
@@ -58,6 +72,8 @@ export const useHandleDeviceConnection = () => {
     const isDeviceSettingsStackFocused = lastRoute === RootStackRoutes.DeviceSettingsStack;
     const isSendStackFocused = lastRoute === RootStackRoutes.SendStack;
     const shouldBlockSendReviewRedirect = isDeviceRemembered && isSendStackFocused;
+    const isDeviceCompromisedModalFocused =
+        lastRoute === RootStackRoutes.DeviceCompromisedModalScreen;
 
     // When is an uninitialized device model that supports device setup, navigate to device onboarding.
     useEffect(() => {
@@ -103,7 +119,8 @@ export const useHandleDeviceConnection = () => {
             isOnboardingFinished &&
             !isPortfolioTrackerDevice &&
             !isDeviceConnectedAndAuthorized &&
-            !isBiometricsOverlayVisible
+            !isBiometricsOverlayVisible &&
+            !shouldNavigateToDeviceCompromisedModal
         ) {
             requestPrioritizedDeviceAccess({
                 deviceCallback: () => dispatch(authorizeDeviceThunk()),
@@ -116,6 +133,9 @@ export const useHandleDeviceConnection = () => {
                     screen: AuthorizeDeviceStackRoutes.ConnectingDevice,
                 });
             }
+        }
+        if (shouldNavigateToDeviceCompromisedModal) {
+            navigation.navigate(RootStackRoutes.DeviceCompromisedModalScreen);
         }
     }, [
         dispatch,
@@ -130,6 +150,7 @@ export const useHandleDeviceConnection = () => {
         shouldBlockSendReviewRedirect,
         isFirmwareInstallationRunning,
         isDeviceInitialized,
+        shouldNavigateToDeviceCompromisedModal,
     ]);
 
     // In case that the physical device is disconnected, redirect to the home screen and
@@ -139,6 +160,13 @@ export const useHandleDeviceConnection = () => {
 
         if (isNoPhysicalDeviceConnected && isOnboardingFinished) {
             if (shouldBlockSendReviewRedirect) {
+                return;
+            }
+            // DeviceCompromisedModal is persistent, so postpone navigating to away until it's dismissed
+            // TODO: this hook is getting very complex, and it's hard to understand the logic when it navigates there and back again.
+            //  Ideally there'd be a single source of truth, a function returning "where we should be as per current state"
+            //  rather than multiple useEffects with imperative instructions "go there when X changes"
+            if (isDeviceCompromisedModalFocused) {
                 return;
             }
             navigation.navigate(RootStackRoutes.AppTabs, {
@@ -154,6 +182,7 @@ export const useHandleDeviceConnection = () => {
         navigation,
         shouldBlockSendReviewRedirect,
         isFirmwareInstallationRunning,
+        isDeviceCompromisedModalFocused,
     ]);
 
     // When trezor gets locked, it is necessary to display a PIN matrix for T1 so that it can be unlocked
--- a/suite-native/intl/src/en.ts
+++ b/suite-native/intl/src/en.ts
@@ -1225,6 +1225,19 @@ export const en = {
             },
         },
     },
+    moduleAuthenticityChecks: {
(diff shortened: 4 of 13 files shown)


suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+        if (navigation.canGoBack()) {
+            navigation.goBack();
+        } else {
+            navigation.navigate(RootStackRoutes.AppTabs, {
+                screen: AppTabsRoutes.HomeStack,
+                params: { screen: HomeStackRoutes.Home },
+            });
+        }
+        dismissCheck();
+    };
```
Lemonexe, line 63: this is kinda convoluted; both the X icon and the Close button should do _the same thing_.
But `ScreenHeader`, which renders X icon, handles navigation itself, so it only gets `dismissCheck` as "what it should do extra", while Close button has gotta do it all.
Curious if there is any other way, but no big deal :slightly_smiling_face:

suite-native/app/package.json
```diff
@@ -59,6 +59,7 @@
         "@suite-native/module-accounts-import": "workspace:*",
         "@suite-native/module-accounts-management": "workspace:*",
         "@suite-native/module-add-accounts": "workspace:*",
+        "@suite-native/module-authenticity-checks": "workspace:*",
```
PeKne, line 62: is it planned that there will be added another check in the future? If not I wou;d prefer to change the name to `@suite-native/module-authenticity-check`. Maybe even `@suite-native/module-firmware-revision-check`

suite-native/app/src/navigation/RootStackNavigator.tsx
```diff
@@ -118,6 +119,10 @@ export const RootStackNavigator = () => {
                 name={RootStackRoutes.SettingsScreenStack}
                 component={SettingsStackNavigator}
             />
+            <RootStack.Screen
```
PeKne, line 84: Are we planning to add more screen to the `AuthenticityChecksStackNavigator`? I could not find any other screen that would suit here. If not, the stack can be deleted. Then you can put the `DeviceCompromisedModalScreen` directly here. Because a stack with only one screen has no point.

suite-native/intl/src/en.ts
```diff
@@ -1210,6 +1210,19 @@ export const en = {
...
+        deviceCompromised: {
+            title: 'Your device may have been compromised',
+            subtitle:
+                'Contact our support to learn what’s going on with your device and what to do next.',
+            steps: {
+                avoidUsingDevice: 'Avoid using this device or sending any funds to it.',
+                contactSupport: 'Continue to Trezor support and use the Chat option.',
+            },
+            buttonContactSupport: 'Contact Trezor Support',
+            buttonClose: 'Close',
```
PeKne: This is duplication. Some of most common short messages are at the top of this file in `generic` section. You can use `generic.buttons.close` id. Please remove it.

suite-native/module-authenticity-checks/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './navigation/AuthenticityChecksStackNavigator';
+export * from './screens/DeviceCompromisedModalScreen';
```
PeKne, line 1: Unnecessary export?

suite-native/app/package.json
PeKne: I would prefer more specific name because from this name I would expect that the device authenticity check will be part of this module, but it is not.

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -44,6 +48,14 @@ export const useHandleDeviceConnection = () => {
     const { isBiometricsOverlayVisible } = useIsBiometricsOverlayVisible();
     const isDeviceUsingPassphrase = useSelector(selectIsDeviceUsingPassphrase);
     const isFirmwareInstallationRunning = useSelector(selectIsFirmwareInstallationRunning);
+
+    const isFirmwareCheckFailed = useSelector(selectHasFirmwareAuthenticityCheckHardFailed);
```
PeKne: please always keep the name of the const same as the selector name
```suggestion
    const hasFirmwareAuthenticityCheckHardFailed = useSelector(selectHasFirmwareAuthenticityCheckHardFailed);
```

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const chatUrl = `${TREZOR_SUPPORT_FW_REVISION_CHECK_FAILED_URL}#open-chat`;
+    const openLink = useOpenLink();
```
PeKne: nit: put hook to the [top of the component definition](https://www.notion.so/satoshilabs/Components-9159b639d21b43ef8b66be150854b6b6?pvs=4#308aa7db519e4f4eaf70312f60ce030a).

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const navigation = useNavigation<NavigationProp>();
+
+    const dismissCheck = () => {
+        if (device?.id) {
+            dispatch(deviceActions.dismissFirmwareAuthenticityCheck(device.id));
+        }
+    };
+
+    const handleClose = () => {
```
PeKne, line 60: Why do we want to check for the back navigation here? Should not it always redirect to HomeScreen no matter what?

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+        } else {
+            navigation.navigate(RootStackRoutes.AppTabs, {
+                screen: AppTabsRoutes.HomeStack,
+                params: { screen: HomeStackRoutes.Home },
+            });
+        }
+        dismissCheck();
+    };
+
+    const chatUrl = `${TREZOR_SUPPORT_FW_REVISION_CHECK_FAILED_URL}#open-chat`;
```
PeKne: nit: this const might be at the top of the file as it is const with value that is not dependant on code execution of this function.

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
```diff
@@ -0,0 +1,92 @@
...
+                    }
+                />
+                <InformativeList />
+            </VStack>
+            <VStack spacing="sp12">
+                <Button colorScheme="redBold" onPress={handleContactSupportClick}>
+                    <Translation id="moduleAuthenticityChecks.deviceCompromised.buttonContactSupport" />
+                </Button>
+                <Button colorScheme="redElevation0" onPress={handleClose}>
+                    <Translation id="moduleAuthenticityChecks.deviceCompromised.buttonClose" />
```
PeKne: Why can't you use `dismissCheck` directly? the navigation is handled in the `useHandleDeviceConnection` hook. When I tried it seems to be working fine.

PeKne (changes requested):

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
PeKne: Maybe just add a comment that the specific hook handles redirection.

PeKne: https://github.com/user-attachments/assets/d9b01efe-66ba-42e9-8c90-294e5893bef3

PeKne:
I just realized that the device authenticity check screen that @yanascz implemented recently looks almost the same (the copy is identical). Might be good idea to unify it? Or is there any reason why we do not want it?
Screenshot 2025-02-04 at 9 55 06 AM

suite-native/app/package.json
Lemonexe: Yes, I know that FW hash check is planned in future :slightly_smiling_face: 

And this is only conjecture on my part, but "entropy check" and "version check" might come later too.
ATM they're both experimental at Suite Desktop, but if Suite Lite is to gradually adopt the same security features as Desktop, then it makes sensethey will some day be on roadmap too.

suite-native/app/src/navigation/RootStackNavigator.tsx
Lemonexe: Sure thing! I was under the impression that RootStack should only render stacks (so an isolated page should have its own stack), but turns out I wasn't paying attention, there are already several screens here :+1:

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
Lemonexe: Huh, it didn't occur to me at first that the useEffect in `useHandleDeviceConnection` might do the trick itself, as it reruns after I dismiss the check. It does work, but unfortunately **only sometimes**.
In situations not yet clear to me, the navigation to `AuthorizeDeviceStackRoutes.ConnectingDevice` does not fire, effectively breaking the X and Close buttons -> you become stuck on the modal unless you disconnect device.

I'd like to discuss in person, I have idea how to refactor it..

suite-native/intl/src/en.ts
Lemonexe: Done in 54cec7ac

suite-native/device/src/hooks/useHandleDeviceConnection.ts
Lemonexe: Done in 4f186034. It was forgotten from previous PR, where the selector was renamed but I had this code ready from even before.

suite-native/app/src/navigation/RootStackNavigator.tsx
Lemonexe: Done in 67b3e339

suite-native/module-authenticity-checks/src/index.ts
Lemonexe: Done in 67b3e339 (now only the screen is exported, the navigator doesnt exist anymore)

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
Lemonexe: Done in 8384ac64.

Btw I think suite-native would benefit from a `@suite-native/urls` package, just like there is for suite desktop. If I have time and energy later  I'd like to do that.

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
Lemonexe: Done in 8384ac64

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
Lemonexe: As discussed in-person, going forward with the temporary solution of programatic navigation
0b7f6052

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
Lemonexe: I agree; I don't know a case where we would specifically prefer returning. Also I just realized it doesn't make sense semantically - the buttons read `X`, `Close`, not `>`,  `Back`.
Done in 0b7f6052

Lemonexe:
Ad similar screens for authenticity checks: discussed personally, but here it's for visibility - there are multiple similar screens like this one, but since they are mostly WIP, it's still too early to try to unify it.
We'll see, maybe common UX will be abstracted to the `@suite-native/module-authenticity-checks` package for all of them, or maybe even a new package..

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -160,6 +162,13 @@ export const useHandleDeviceConnection = () => {
             if (shouldBlockSendReviewRedirect) {
                 return;
             }
+            // DeviceCompromisedModal is persistent, so postpone navigating to away until it's dismissed
+            // TODO: this hook is getting very complex, and it's hard to understand the logic when it navigates there and back again.
+            //  Ideally there'd be a single source of truth, a function returning "where we should be as per current state"
+            //  rather than multiple useEffects with imperative instructions "go there when X changes"
+            if (isDeviceCompromisedModalFocused) {
```
Lemonexe, line 169: at first I tried to use `shouldNavigateToDeviceCompromisedModal` for this condition, but then I realized - even if you don't dismiss the modal, this variable still becomes false when disconnecting device, because whole `state.device` is anulled when you disconnect device.

`lastRoute` might not be an ideal way to determine current state of UI, but I didn't think of anything else without overengineering it..

PeKne (approved):
Well done, thanks for adjusting the code to my comments ;).

Lemonexe:
Merging despite broken E2E tests, they're broken globally ([slack discussion](https://satoshilabs.slack.com/archives/C03FL9TCURZ/p1738833090410379))

PR #16728, opened 2025-01-30
chore(suite-native): Unify PriceChangeIndicator

jbazant:
## Description

- Refactored "Price change badges" to single component which is used from "everywhere".
- Unified look. (size, colors, carets)

## Questions
- When change is exactly 0%. Badge is green. Is this intended or do we want to introduce grey state?
- Previously when change was exactly 0%, skeleton element was displayed instead. I expect this was a bug. (Now covered by test.)
- Do we want to add `PriceChangeIndicator` to Component demo screen?

## Related Issue

Resolve 

## Screenshots:
Screenshot 2025-01-30 at 13 42 34

### DIFF
--- a/suite-native/atoms/src/__tests__/PriceChangeBadge.comp.test.tsx
+++ b/suite-native/atoms/src/__tests__/PriceChangeBadge.comp.test.tsx
@@ -1,0 +1,48 @@
+import { BasicProvider, render, renderHook } from '@suite-native/test-utils';
+import { NativeStyleUtils, useNativeStyles } from '@trezor/styles';
+
+import { Box as MockBox } from '../Box';
+import { PriceChangeBadge } from '../PriceChangeBadge';
+
+jest.mock('../Skeleton/BoxSkeleton', () => ({
+    BoxSkeleton: () => <MockBox testID="skeleton-box" />,
+}));
+
+describe('PriceChangeBadge', () => {
+    let colors: NativeStyleUtils['colors'];
+
+    beforeAll(() => {
+        const { result } = renderHook(() => useNativeStyles(), { wrapper: BasicProvider });
+        ({ colors } = result.current.utils);
+    });
+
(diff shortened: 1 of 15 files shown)


Nodonisko (approved):
I was kind of afraid of performance of this because when hovering very quick over graph every re-render could cause significant lag and frame drop (that's reason why it was in so atomic components before), but I briefly tested it and it looks that it doesn't have any  significant negative impact

suite-native/atoms/src/__tests__/PriceChangeBadge.comp.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+jest.mock('../Skeleton/BoxSkeleton', () => ({
+    BoxSkeleton: () => <MockBox testID="skeleton-box" />,
+}));
+
+describe('PriceChangeBadge', () => {
+    it('should render green badge for change > 0', () => {
+        const { getByText } = render(<PriceChangeBadge valuePercentageChange={1} />);
+
+        expect(getByText('1.00%')).toHaveStyle({ color: '#0F6148' });
```
vytick: What do you think about using colors from `const { utils } = useStyles()` and `utils.colors` for these tests?

jbazant: I like that idea! Fixed.

vytick (approved):
Thanks for aligning and simplifying this <3

vytick:
Answers:
- green for 0% is expected
- skeleton in that case was a bug
- Adding to demo screen would be benefcial

jbazant:
> Answers:
> 
> * green for 0% is expected
> * skeleton in that case was a bug
> * Adding to demo screen would be benefcial

vytick (approved):
looks good 🚢 🇮🇹

PR #16729, opened 2025-01-30
Build: connect plugins support esm

### LINKED ISSUE #16245: Support ESM import of @metamask/eth-sig-util in @trezor/connect-plugin-ethereum

karliatto:
## Description
Add esm support to connect plugins.
* Using the field `"exports"` in package.json to define the entry points of the libraries. While keeping the previous `publishConfig` for bundlers that do not support `exports`. Using `exports` give us control over the entry points that are accessible they the developers using this packages, making it more clear in the future if we change somehting in source we can be safe since they only will use whatever entrypoints we define there. 
* The packages are including a new `libESM` that will be publish to NPM.
 
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16245

### DIFF
--- a/packages/connect-plugin-ethereum/package.json
+++ b/packages/connect-plugin-ethereum/package.json
@@ -23,8 +23,14 @@
     },
     "main": "src/index.ts",
     "files": [
-        "lib/"
+        "lib/",
+        "libESM",
+        "!**/*.map"
     ],
+    "exports": {
+        "node": "./lib/index.js",
+        "default": "./libESM/index.js"
+    },
     "peerDependencies": {
         "@metamask/eth-sig-util": "^8.0.0",
         "tslib": "^2.6.2"
@@ -35,6 +41,8 @@
     "scripts": {
         "test:unit": "yarn g:jest -c ../../jest.config.base.js",
         "type-check": "yarn g:tsc --build tsconfig.json",
-        "build:lib": "yarn g:rimraf ./lib && yarn g:tsc --build tsconfig.lib.json && ../../scripts/replace-imports.sh ./lib"
+        "build:lib": "yarn build:lib:cjs && yarn build:lib:esm",
+        "build:lib:cjs": "yarn g:rimraf ./lib && yarn g:tsc --build tsconfig.lib.json && ../../scripts/replace-imports.sh ./lib",
+        "build:lib:esm": "yarn g:rimraf ./libESM && yarn g:tsc --build tsconfig.libESM.json && ../../scripts/replace-imports.sh ./libESM libESM"
     }
 }
--- a/packages/connect-plugin-ethereum/tsconfig.libESM.json
+++ b/packages/connect-plugin-ethereum/tsconfig.libESM.json
@@ -1,0 +1,8 @@
+{
+    "extends": "./tsconfig.lib.json",
+    "compilerOptions": {
+        "outDir": "libESM",
+        "module": "ESNext",
+        "target": "esnext"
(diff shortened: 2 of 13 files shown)


packages/connect-plugin-ethereum/package.json
```diff
@@ -18,13 +18,14 @@
...
-        "main": "lib/index.js"
-    },
     "main": "src/index.ts",
     "files": [
         "lib/"
     ],
+    "exports": {
+        "node": "./lib/index.js",
+        "default": "./libESM/index.js"
+    },
```
karliatto, line 33: This is following the approach describes here https://webpack.js.org/guides/package-exports/#providing-commonjs-and-esm-version-stateless

Also about exports: https://nodejs.org/api/packages.html#package-entry-points

scripts/replace-imports.sh
```diff
@@ -2,18 +2,55 @@
...
 # Execute the appropriate command based on the OS
 if [[ "$OS" == "Darwin" ]]; then
     # macOS command with -i '' for in-place editing without backup and -E for extended regex
-    find "$1" -type f -exec sed -i '' -E "$REGEX" {} +
+    find "$1" -type f | while read -r file; do
+        if grep -q "@trezor/[^/]+/src" "$file"; then
+            echo "Changing: $file" >> "$LOG_FILE"
+            sed -i '' -E "$REGEX" "$file"
+        fi
+    done
```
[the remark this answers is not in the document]
karliatto: Thanks!

mroz22 (approved):
sounds good to me, I didn't run any deep investigation but I feel quite confident to test it in a canary release.

packages/connect-plugin-ethereum/tsconfig.libESM.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "./tsconfig.lib.json",
+    "compilerOptions": {
+        "outDir": "libESM",
+        "module": "ESNext"
```
Nodonisko: I am thinking that it might be good time to also `target` to `esnext` too, because there is no reason we should transpile to older es version

karliatto: Sure, let's do that https://github.com/trezor/trezor-suite/pull/16729/files#diff-97fc10c343bde083841616cc3853cf0d7e98003d6a1b11aa36a7665e69678b0bR6

PR #16734, opened 2025-01-30
feat(suite): add address confirm to send modal

seibei-iguchi:
- added a new screen to the send form if the first output is an address
- rest of the items is shown only after the first confirm
- in multirecipient send, addresses are only shown after confirmation

## Screenshots:
Screenshot 2025-01-30 at 18 59 57
Screenshot 2025-01-30 at 19 00 09
Screenshot 2025-01-30 at 19 00 35

### DIFF
--- a/packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputElement.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputElement.tsx
@@ -13,6 +13,7 @@ import {
     H4,
     Icon,
     InfoItem,
+    Note,
     Row,
     Text,
 } from '@trezor/components';
@@ -43,9 +44,9 @@ const DataWrapper = styled.p`
 
 const Status = ({ state }: { state: TransactionReviewOutputElementProps['state'] }) => {
     switch (state) {
-        case 'done':
+        case 'confirmed':
             return <Icon size={spacings.md} variant="primary" name="check" />;
-        case 'pending':
+        case 'unconfirmed':
             return <DotIndicator />;
         default:
             return <DotIndicator isActive={true} />;
@@ -58,17 +59,24 @@ type ValueProps = {
     symbol: NetworkSymbol;
     isFiatVisible: boolean;
     isFee: boolean;
+    state: TransactionReviewOutputElementProps['state'];
     token?: TokenInfo;
 };
 
-const Value = ({ value, type, symbol, token, isFee, isFiatVisible }: ValueProps) => {
+const Value = ({ value, type, symbol, token, isFee, isFiatVisible, state }: ValueProps) => {
     switch (type) {
         case 'address':
+            return state !== 'confirmed' ? (
+                <Note>
+                    <Translation id="TR_SEND_ADDRESS_CONFIRMATION_NOTE" />
+                </Note>
+            ) : (
+                <Address value={value} />
+            );
+        case 'safe-address':
             return <Address value={value} />;
         case 'data':
             return <DataWrapper>{value}</DataWrapper>;
-        case 'total':
-        case 'fee':
         case 'amount': {
             const isTokenAmount = !isFee && token;
             const formattedValue = isTokenAmount
@@ -114,7 +122,7 @@ const Value = ({ value, type, symbol, token, isFee, isFiatVisible }: ValueProps)
 export type OutputElementLine = {
     id: string;
     value: string;
-    type: 'default' | 'address' | 'data' | 'amount' | 'fee' | 'total';
+    type: 'default' | 'address' | 'safe-address' | 'data' | 'amount';
     label?: ReactNode;
 };
 
@@ -122,7 +130,7 @@ export type TransactionReviewOutputElementProps = {
     title: ReactNode;
     lines: OutputElementLine[];
     account: Account;
-    state: 'default' | 'done' | 'pending';
+    state: 'active' | 'confirmed' | 'unconfirmed';
     fiatVisible?: boolean;
     token?: TokenInfo;
 };
@@ -138,12 +146,12 @@ export const TransactionReviewOutputElement = ({
     const { networkType, symbol } = account;
 
     return (
-        <Card paddingType="none" fillType={state === 'done' ? 'flat' : 'default'}>
+        <Card paddingType="none" fillType={state === 'confirmed' ? 'flat' : 'default'}>
             <Row padding={{ vertical: spacings.sm, horizontal: spacings.md }} gap={spacings.sm}>
                 <Status state={state} />
                 <H4
-                    margin={{ left: spacings.xxxs }}
-                    typographyStyle={state !== 'pending' ? 'callout' : 'hint'}
+                    margin={{ left: spacings.xxs }}
+                    typographyStyle={state !== 'unconfirmed' ? 'callout' : 'hint'}
                 >
                     {title}
                 </H4>
@@ -166,15 +174,12 @@ export const TransactionReviewOutputElement = ({
                             token={token}
                             isFiatVisible={fiatVisible}
                             isFee={line.id === 'fee'}
+                            state={state}
                         />
                     );
 
                     return (
-                        <Column
-                            data-testid={`@modal/output-${line.id}`}
-                            key={line.id}
-                            gap={spacings.md}
-                        >
+                        <Column data-testid={`@modal/output-${line.id}`} key={line.id}>
                             <Text typographyStyle="hint" as="div">
                                 {line.label ? (
                                     <InfoItem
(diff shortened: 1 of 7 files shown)


Hermez-cz (approved):

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputElement.tsx
```diff
@@ -114,7 +116,7 @@ const Value = ({ value, type, symbol, token, isFee, isFiatVisible }: ValueProps)
 export type OutputElementLine = {
     id: string;
     value: string;
-    type: 'default' | 'address' | 'data' | 'amount' | 'fee' | 'total';
+    type: 'default' | 'address' | 'safe-address' | 'data' | 'amount' | 'fee' | 'total';
```
peter-sanderson: What safe-address means? Thats the address user confirmed on device? Then I suggest `confirmed-address` maybe even `on-device-confirmed-address`.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputList.tsx
```diff
@@ -88,6 +88,59 @@ export const TransactionReviewOutputList = ({
...
+                    <H3>
+                        <Translation id="TR_SEND_ADDRESS_CONFIRMATION_HEADING" />
+                    </H3>
+                    <BulletList
+                        isOrdered
+                        bulletGap={spacings.md}
+                        titleGap={spacings.zero}
+                        gap={spacings.xxl}
+                    >
+                        <BulletList.Item
```
peter-sanderson, line 113: nit: there is code repetition `<BulletList.Item` + `<H4>` can be exctracted to some `Item` component with `children` and `title` prop.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputElement.tsx
```diff
@@ -170,35 +172,37 @@ export const TransactionReviewOutputElement = ({
...
-                                                {line.label}
-                                            </Text>
-                                        }
-                                        direction="row"
-                                    >
-                                        <Column
-                                            alignItems="flex-end"
-                                            data-testid="@modal/output-value"
+                        <Column data-testid={`@modal/output-${line.id}`} key={line.id}>
+                            {line.type === 'address' && state !== 'done' ? (
```
Lemonexe: I don't understand this condition. When is this ever displayed?

peter-sanderson: I think this is situation when `address` is unconfirmed on the device. 

The else part is case when `safe-address` is in the `type`.

peter-sanderson: I was just about to write: that I would like to see cleared mapping between the `line.type` and the behavior, its is hard to navigate.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputElement.tsx
peter-sanderson: And maybe even rename `address` to `unconfirmed-address`

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputList.tsx
Lemonexe, line 113: @Hermez-cz already approved, so I understand that it matches design. One tiny thing for consideration - in the checklist, all the bullets are **higher** than their respective texts. Might look better centered? :thought_balloon: just IMO feel free to resolve :man_shrugging: 
I can see it's [aligned to top currently](https://github.com/user-attachments/assets/8697b97e-a86e-4eea-965d-ff680a4bcba1)

Lemonexe (commented):
Looks good, tested with regtest for single recipient and multiple recipient Txs.

But I admit I don't understand all the possible flows of send modal, so idk if there are some flows that I missed. It could get cleaned up so that the flows are more readable, but yeah, am not sure how. Please address the comments, maybe we're just missing something :slightly_smiling_face:

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputElement.tsx
seibei-iguchi: It is described in the PR description:) This is showed for all addresses until they are confirmed on the safe device. safe-address is for receive and xpub modals where it doesn't make sense to compare with external source.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputList.tsx
seibei-iguchi: The component can currently only handle top align, due to how it's constructed. It's not trivial to make it centered, unfortunately:)

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputElement.tsx
seibei-iguchi: Nope, it's an address for which it doesn't make sense to check with external source, because it's only shown on the device and Suite, eg xpub and receive modal.

peter-sanderson: Ok, thanks. Maybe lets add this info into a comment please. :pray:

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewOutputList/TransactionReviewOutputElement.tsx
peter-sanderson: So the `state !== 'done'` is the part that makes it "unconfirmed by the user" right?

peter-sanderson: I see, 
image


Can we make it `'confirmed-n-device` | `unconfirmed..` it would be much clearer

peter-sanderson: I see there is also `pending` state, I have found the `getState` in the TransactionReviewOutputList but I don't understand it.

peter-sanderson (approved):

Lemonexe (approved):
Thanks for the explanations, now it makes sense! I know you didn't write the original code, so it's not fair from us to make you clean it up :smile:  but thanks for making it more readable and explaining!

Please, would you add a comment for TransactionReviewOutputElement.tsx L177 to make it clear this is displayed for 2nd and other outputs until they are confirmed, while for the 1st output there is checklist? Thanks!

tomasklim:
Should we show this in case I am sending between my Trezor accounts?
Screenshot 2025-01-31 at 16 22 02
The apps know it, so it can be skipped or somehow mentioned there.
Screenshot 2025-01-31 at 16 22 18

tomasklim:
Cardano misses the first `Verify` step
Screenshot 2025-01-31 at 16 52 40

seibei-iguchi:
> Cardano misses the first `Verify` step Screenshot 2025-01-31 at 16 52 40

Yep, we know some of the flows do this and we are currently ok with that. For example, XRP with destination tag is the same. It requires a FW change.

jvaclavik (approved):

jvaclavik (approved):

bosomt:
**QA OK**

- send BTC
- send multiple BTC
- send ETH 
- tested both Trezor address and external addresses 

**Info:**
 - Suite version: desktop 25.2.0 (dde03ca4532357ec5cb40bd6d6939685f3023bda)
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T3B1 2.8.7 regular (revision 8a254aa8eae82f99630df63f40e4d290066a3efc)

PR #16735, opened 2025-01-30
chore(connect): deviceCommands call and commonCall shallow structure

mroz22:


packages/connect/src/device/DeviceCommands.ts
```diff
@@ -348,43 +319,77 @@ export class DeviceCommands {
...
-                    signal: abortController.signal,
-                })
-                .finally(() => {
-                    clearTimeout(timeout);
-                });
-            // throw error anyway, next call should be resolved properly
-            throw error;
+
+        if ('isTransportError' in response) {
+            throw new Error(response.error, { cause: 'transport-error' });
```
marekrjpolak: Two-parametrical `Error` constructor is not supported in `es2019` => build:lib error. Can we change the target to `esnext` or something?

packages/connect/src/device/DeviceCommands.ts
```diff
@@ -348,43 +319,77 @@ export class DeviceCommands {
...
-                    protocol: this.device.protocol,
-                    signal: abortController.signal,
-                })
-                .finally(() => {
-                    clearTimeout(timeout);
-                });
-            // throw error anyway, next call should be resolved properly
-            throw error;
+
+        if ('isTransportError' in response) {
```
marekrjpolak: Just to be safe?
```suggestion
        if ('isTransportError' in response && response.isTransportError) {
```

packages/connect/src/device/DeviceCommands.ts
```diff
@@ -348,43 +319,77 @@ export class DeviceCommands {
...
+        );
     }
 
-    async _commonCall(type: MessageKey, msg?: DefaultPayloadMessage['message']) {
+    async _commonCall(
+        type: MessageKey,
+        msg?: DefaultPayloadMessage['message'],
+    ): Promise<
+        DefaultPayloadMessage | (ErrorGeneric<ReadWriteError> & { isTransportError: true })
+    > {
```
marekrjpolak: I'm not a big fan of somehow inconsistent return types.

packages/connect/src/device/DeviceCommands.ts
```diff
@@ -348,43 +319,77 @@ export class DeviceCommands {
...
+                'Received error',
+                res.error,
+                // res.message is not propagated to higher levels, only logged here. webusb/node-bridge may return message with additional information
+                res.message,
+            );
+
+            return {
+                ...res,
+                isTransportError: true,
+            };
```
marekrjpolak, line 407: Given that `_commonCall` is not throwing any more in this case, it will change the logic below in `_filterCommonTypes`, e.g. `this._commonCall('PinMatrixAck', { pin }).then(response => ...)`, right? Do we mind?

packages/connect/src/device/DeviceCommands.ts
```diff
@@ -304,27 +293,9 @@ export class DeviceCommands {
...
-
         this.callPromise = undefined;
-        if (!res.success) {
-            logger.warn(
-                'Received error',
-                res.error,
-                // res.message is not propagated to higher levels, only logged here. webusb/node-bridge may return message with additional information
-                res.message,
-            );
-            throw new Error(res.error);
```
marekrjpolak: Maybe I'm missing the point, but why don't we just do the `throw new Error(res.error, { cause: 'transport-error' })` thing here, instead of all the other changes? It would propagate upwards anyway, right?

mroz22:
we are going es2022 here #16735 so cause might be used now

marekrjpolak:
I believe this can be closed in favor of #17717

PR #15789, opened 2024-12-05
Feat(suite): analytics OS versions

Lemonexe:
## Description

1. add a new `ipcMain` api `getSystemInformation`
2. optionally use it to enrich context for Analytics `'suite-ready'` event (desktop only)

## Related Issue

Resolve #15770

PR #16741, opened 2025-01-30
Chore/message system catalina deprecation

### LINKED ISSUE #16693: Display deprecation message for macOS Catalina and earlier
Add message-system message targetting Suite Desktop: macOS Catalina and earlier informing that with Trezor Suite 25.3.x will no longer support those systems.

ℹ  support will be dropped by bumping Electron from 32 to latest in #16615, see details in #15766

Lemonexe:
## Description

Add message to warn Catalina users that they mustn't update their Suite to 25.3.x, as it'll probably stop working.

Although message-system does not yet support `desktopOsVersion` (from nodeJS, added in #15789), this is fine, because the issue was, that `osVersion` can't distinguish Big Sur _and those above_. So targetting Catalina and below can be done just with the userAgent `osVersion` :slightly_smiling_face: 

## Related Issue

Resolve #16693

## Screenshots:

image

@coderabbitai ignore

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,8 +1,49 @@
 {
     "version": 1,
-    "timestamp": "2025-02-03T00:00:00+00:00",
-    "sequence": 79,
+    "timestamp": "2025-02-13T00:00:00+00:00",
+    "sequence": 80,
     "actions": [
+        {
+            "conditions": [
+                {
+                    "os": {
+                        "macos": "<11.0.0",
+                        "linux": "!",
+                        "windows": "!",
+                        "android": "!",
+                        "ios": "!",
+                        "chromeos": "!"
+                    },
+                    "environment": {
+                        "desktop": "*",
+                        "mobile": "!",
+                        "web": "!"
+                    }
+                }
+            ],
+            "message": {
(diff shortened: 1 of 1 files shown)


Lemonexe:
@coderabbitai ignore

suite-common/message-system/config/config.v1.json
```diff
@@ -1,9 +1,50 @@
...
-    "sequence": 76,
+    "timestamp": "2025-01-30T00:00:00+00:00",
+    "sequence": 77,
     "actions": [
-        {
+      {
+            "conditions": [
+                {
+                    "os": {
+                        "macos": "<10.16.0",
```
Lemonexe: this should theoretically be correct, but it must be verified.
I _guess_ that `osVersion` is using the first column in https://en.wikipedia.org/wiki/MacOS_version_history#Releases

So 10.15.x and below is what we want to target. Idk if Big Sur is 11 or 10.16, but it should work nevertheless.

This needs to be confirmed!

komret:
Closing because of https://github.com/trezor/trezor-suite/issues/16985. Most of our users on Catalina will be on 10.15.7 as that is the latest Catalina version. So they cannot be targeted without displaying the message to all Mac users.

PR #16754, opened 2025-01-31
fix(fiat-services): fix fetching fiat rates for timestamps range

### LINKED ISSUE #15440: Show current coin price for XRP, SOL, ADA on account detail
The price for XRP, SOL, ADA we show right now on Account Detail is not very up to date, because we get the price for the beginning of the day only and no updates during the day (even though we fetch it every 30s...)

I guess we should fix it or hide the box for non-blockbook coins.

Would be nice to double check how exactly it works for blockbook coins.

Compare to desktop implementation and possibly align product wise - on desktop there is 7d change and not 24h change indicator and extra tooltip with info about last update time. 

image

```[tasklist]
### Tasks
- [x] Product: check what is the industry standard on what to display in the change, is it 7d or 24h better
- [x] Engineering: to check if it happens for other networks besides XRP, SOL and ADA
```

vytick:
When fetching fiat rates and picking the best one for given timestamp, the range was increased by wrong number. This caused wrong selection of the closest fiat rate for the timestamp and also wrong 24h change computation.

Bug:
It always shows Solana, Cardano and Ripple 24h change in account detail either loading or 0%.

ThisPR should fix it.

## Related Issue

Related to #15440

### DIFF
--- a/suite-common/fiat-services/src/coingecko.ts
+++ b/suite-common/fiat-services/src/coingecko.ts
@@ -9,7 +9,7 @@ import { RateLimiter } from './limiter';
 // a proxy for https://api.coingecko.com/api/v3
 const COINGECKO_API_BASE_URL = 'https://cdn.trezor.io/dynamic/coingecko/api/v3';
 
-const ONE_DAY_IN_MS = 24 * 60 * 60 * 1000;
+const ONE_DAY_IN_S = 24 * 60 * 60;
 
 interface HistoricalResponse extends HistoricRates {
     symbol: string;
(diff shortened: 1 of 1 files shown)


suite-common/fiat-services/src/coingecko.ts
```diff
@@ -136,8 +136,8 @@ export const getFiatRatesForTimestamps = async (
     const sortedTimestamps = [...timestamps].sort((ts1, ts2) => ts1 - ts2);
 
     // adjust from and to timestamps to get better range of data
-    const fromTimestamp = sortedTimestamps[0] - ONE_DAY_IN_MS;
-    const toTimestamp = sortedTimestamps[sortedTimestamps.length - 1] + ONE_DAY_IN_MS;
+    const fromTimestamp = sortedTimestamps[0] - ONE_DAY_IN_S;
```
Nodonisko: Can you also rename it `sortedTimestampsInSeconds`?

Nodonisko (approved):

suite-common/fiat-services/src/coingecko.ts
vytick: done

STew790:
**QA OK**
I can see 24h fiat change correctly on XRP, SOL and ADA.

**Info**
25.1.3 e549011

PR #16769, opened 2025-02-03
Update Expo linking, simplify popup deeplink callback

martykan:
## Description

There was an issue in `expo-linking` on iOS that was fixed in the latest version https://github.com/expo/expo/issues/33031

This PR also refactors the logic around `expo-linking` to use their `useURL` hook  instead of manual callbacks.

### DIFF
--- a/suite-native/module-connect-popup/src/hooks/useConnectPopupNavigation.ts
+++ b/suite-native/module-connect-popup/src/hooks/useConnectPopupNavigation.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect } from 'react';
+import { useEffect } from 'react';
 
 import { useNavigation } from '@react-navigation/native';
 import * as Linking from 'expo-linking';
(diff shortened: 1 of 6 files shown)


martykan:
@coderabbitai review

suite-native/module-connect-popup/src/hooks/useConnectPopupNavigation.ts
```diff
@@ -34,33 +34,12 @@ export const useConnectPopupNavigation = () => {
...
-    const navigateToConnectPopup = useCallback(
-        (url: string) => {
-            if (!featureFlagEnabled) return;
-            if (!isConnectPopupUrl(url)) return;
-            const parsedUrl = Linking.parse(url);
-            navigation.navigate(RootStackRoutes.ConnectPopup, { parsedUrl });
-        },
-        [navigation, featureFlagEnabled],
-    );
+    const url = Linking.useURL();
```
Nodonisko, line 37: Did you tried all cases? I think I played with this hook intially but found some edge case where it doesn't worked correctly but I can't recall it now.

suite-native/module-connect-popup/src/hooks/useConnectPopupNavigation.ts
```diff
@@ -34,33 +34,12 @@ export const useConnectPopupNavigation = () => {
...
-        // this could be probably handed by Linking.addEventListener
-        const subscription = Linking.addEventListener('url', event => {
-            navigateToConnectPopup(event.url);
-        });
-
-        return () => subscription?.remove();
-    }, [navigateToConnectPopup]);
+        if (!featureFlagEnabled) return;
+        if (!url || !isConnectPopupUrl(url)) return;
+        const parsedUrl = Linking.parse(url);
```
Nodonisko: I would suggest wrapping parse to try catch

Nodonisko (approved):

suite-native/module-connect-popup/src/hooks/useConnectPopupNavigation.ts
martykan: After updating the library I didn't see any issue. I looked at what the hook does and it's basically the same thing.

suite-native/module-connect-popup/src/hooks/useConnectPopupNavigation.ts
martykan: Good point, done

Nodonisko (approved):

PR #16776, opened 2025-02-03
feat(suite-native): onboarding suspicious device screen

### LINKED ISSUE #16279: Onboarding - Suspicious Device Screen
### Acceptance criteria
As a user that is on the **Unlock** screen and clicks on CTA [Not my device], I would like to see the **Not sure** screen where I can see information on what to do next and (1) contact support or (2) go back.
    - CTAs available
    - Contact [Support]
    - [Back]
    - Conditionals
        - If the user clicks on [Support], the user will be re-directed to: https://trezor.io/support
        - If the users clicks on [Back], the user will go back to prior screen. This can we either **Unlock Trezor** screen, **Security check** screen (with the state of the stepper remembered)

### Possible improvements:
- the redirect to the support page will contain some url argument informing from where the user get here (unsealed packaging, holographic seal missing etc.). The chat bot team will prepare some automations so the user can see some answer right away.

PeKne:
## Related Issue
- screen explaining what to do with a suspicious device added
- if the device is disconnected while is user on this screen (we ecnourage then to do so), he is not redircted away so he can finish reading the content of the screen

Resolve #16279 

## Screenshots

https://github.com/user-attachments/assets/ab4759c1-a18a-4962-86b3-22dba27d44e9

### DIFF
--- a/suite-native/device/src/hooks/useHandleDeviceConnection.ts
+++ b/suite-native/device/src/hooks/useHandleDeviceConnection.ts
@@ -41,6 +41,19 @@ type NavigationProp = StackToStackCompositeNavigationProps<
     RootStackParamList
 >;
 
+// We encourage user to disconnect device when he is redirected to suspicious device screen.
+// We should not redirect him away so he can read the screen content and decide what to do.
+// If the device is connected again, he still should stay on that screen.
+const isSuspiciousDeviceScreenFocused = (navigation: NavigationProp) => {
+    const previousRoute = navigation.getState()?.routes.at(-1);
+    const innerStackRoute = previousRoute?.state?.routes.at(-1);
+
+    return (
+        previousRoute?.name === RootStackRoutes.OnboardingStack &&
+        innerStackRoute?.name === OnboardingStackRoutes.SuspiciousDevice
+    );
+};
+
 export const useHandleDeviceConnection = () => {
     const isNoPhysicalDeviceConnected = useSelector(selectIsNoPhysicalDeviceConnected);
     const isPortfolioTrackerDevice = useSelector(selectIsPortfolioTrackerDevice);
@@ -77,6 +90,7 @@ export const useHandleDeviceConnection = () => {
 
     // When is an uninitialized device model that supports device setup, navigate to device onboarding.
     useEffect(() => {
+        if (isSuspiciousDeviceScreenFocused(navigation)) return;
         if (
             isDeviceSetupSupported &&
             !isDeviceInitialized &&
@@ -111,7 +125,7 @@ export const useHandleDeviceConnection = () => {
     // At the moment when unauthorized physical device is selected,
     // redirect to the Connecting screen where is handled the connection logic.
     useEffect(() => {
-        if (isFirmwareInstallationRunning) return;
+        if (isFirmwareInstallationRunning || isSuspiciousDeviceScreenFocused(navigation)) return;
 
         if (
             isDeviceInitialized &&
@@ -166,7 +180,7 @@ export const useHandleDeviceConnection = () => {
             // TODO: this hook is getting very complex, and it's hard to understand the logic when it navigates there and back again.
             //  Ideally there'd be a single source of truth, a function returning "where we should be as per current state"
             //  rather than multiple useEffects with imperative instructions "go there when X changes"
-            if (isDeviceCompromisedModalFocused) {
+            if (isDeviceCompromisedModalFocused || isSuspiciousDeviceScreenFocused(navigation)) {
                 return;
             }
             navigation.navigate(RootStackRoutes.AppTabs, {
--- a/suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
+++ b/suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
@@ -1,0 +1,126 @@
+import { useSelector } from 'react-redux';
+
+import { selectIsNoPhysicalDeviceConnected } from '@suite-common/wallet-core';
+import { Button, IconListTextItem, TitleHeader, VStack } from '@suite-native/atoms';
+import { Translation } from '@suite-native/intl';
+import { useOpenLink } from '@suite-native/link';
+import {
+    AppTabsRoutes,
+    DeviceSuspicionCause,
+    HomeStackRoutes,
+    OnboardingStackParamList,
+    OnboardingStackRoutes,
+    RootStackParamList,
+    RootStackRoutes,
+    Screen,
+    ScreenHeader,
+    StackToStackCompositeScreenProps,
+} from '@suite-native/navigation';
+import {
+    TREZOR_SUPPORT_DIFFERENT_PACKAGING,
+    TREZOR_SUPPORT_FW_ALREADY_INSTALLED,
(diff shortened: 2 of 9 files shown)


suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -95,7 +109,7 @@ export const useHandleDeviceConnection = () => {
     // At the moment when unauthorized physical device is selected,
     // redirect to the Connecting screen where is handled the connection logic.
     useEffect(() => {
-        if (isFirmwareInstallationRunning) return;
+        if (isFirmwareInstallationRunning || isSuspiciousDeviceScreenFocused(navigation)) return;
```
Nodonisko, line 128: something to think about:

I am quite worrying about this hook because it will only grow and it's already very hard to modify to not break some edge case. Not sure about solution just want to mention it.

Nodonisko (approved):

Nodonisko:
Did not test it but code looks fine

suite-native/navigation/src/navigators.ts
```diff
@@ -113,6 +113,14 @@ export type OnboardingStackParamList = {
     [OnboardingStackRoutes.AnalyticsConsent]: undefined;
     [OnboardingStackRoutes.Biometrics]: undefined;
     [OnboardingStackRoutes.UninitializedDeviceLanding]: undefined;
+    [OnboardingStackRoutes.SuspiciousDevice]: {
+        suspicionCause:
```
yanascz: Maybe `cause` is sufficient given the context? 🤔

suite-native/module-onboarding/src/screens/UninitializedDeviceLandingScreen.tsx
```diff
@@ -77,8 +85,15 @@ export const UninitializedDeviceLandingScreen = () => {
     };
 
     const handleDeclineButtonPress = () => {
```
yanascz: Would it make sense to rename also this one in a similar way as bellow? 🤔

suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
```diff
@@ -0,0 +1,128 @@
...
+    StackToStackCompositeScreenProps,
+} from '@suite-native/navigation';
+import {
+    TREZOR_SUPPORT_DIFFERENT_PACKAGING,
+    TREZOR_SUPPORT_FW_ALREADY_INSTALLED,
+    TREZOR_SUPPORT_IS_MY_DEVICE_SAFE,
+    Url,
+} from '@trezor/urls';
+
+type SuspicionCause =
```
yanascz: Would it make sense to declare this type elsewhere and use it here and in `navigators.ts`?

suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
```diff
@@ -0,0 +1,128 @@
...
+                screen: AppTabsRoutes.HomeStack,
+                params: {
+                    screen: HomeStackRoutes.Home,
+                },
+            });
+
+            return;
+        }
+
+        navigation.goBack();
```
yanascz, line 65: I would like this a bit more:
```suggestion
        } else {
            navigation.goBack();
        }
```

suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
```diff
@@ -0,0 +1,128 @@
...
+    };
+
+    return (
+        <Screen header={<ScreenHeader closeAction={handleBackButtonPress} />}>
+            <VStack justifyContent="space-between" flex={1}>
+                <VStack spacing="sp32">
+                    <TitleHeader
```
yanascz, line 72: The distance of the title from the screen header is different than on the previous screen. I believe it's an inconsistency that should be addressed.

suite-native/navigation/src/navigators.ts
PeKne: Type moved to top of the file in 49944b85f5.

suite-native/module-onboarding/src/screens/UninitializedDeviceLandingScreen.tsx
PeKne: name changed in 49944b85f5

suite-native/navigation/src/navigators.ts
PeKne: I prefer to keep it more descriptive like this.

suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
PeKne: Type moved to the top of the file in 49944b85f5.

suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
PeKne: Do you have strong opinion about it? I prefer it the way that I used. I think we even discussed early returns on one of the previous code improvement meetings and decided that it is ok to use it like that.

suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
PeKne: 16 padding top added in https://github.com/trezor/trezor-suite/commit/49944b85f5b3c84168e83d771971ed462cb7db64

suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
yanascz: In this specific case I don't see any benefit in using early `return`. In fact, I had to pause and think for a while tu understand why it's used here. It's even longer than the suggested variant.

IMO early returns make sense when you've got several conditions and want to avoid unnecessary nesting.

This is not a blocker for me so up to you.

yanascz (approved):
Works as expected. 🎉

suite-native/device/src/hooks/useHandleDeviceConnection.ts
PeKne: Agree, I plan to refactor this file a bit to make it more transparent.

suite-native/module-onboarding/src/screens/SuspiciousDeviceScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+    Screen,
+    ScreenHeader,
+    StackToStackCompositeScreenProps,
+} from '@suite-native/navigation';
+import {
+    TREZOR_SUPPORT_DIFFERENT_PACKAGING,
+    TREZOR_SUPPORT_FW_ALREADY_INSTALLED,
+    TREZOR_SUPPORT_IS_MY_DEVICE_SAFE,
+    Url,
+} from '@trezor/urls';
```
Lemonexe, line 24: import not specified in package.json. Actually we don't use `@trezor/urls` anywhere in suite-native, so I'm wondering how it's able to resolve the import at all :thinking:

PR #16782, opened 2025-02-03
chore(components): use transient prop in RadioCard

martykan:
## Description

Someone forgot...

Screenshot 2025-02-03 at 16 43 53


martykan:
Already fixed by #16778

PR #16790, opened 2025-02-04
chore(suite-native): configure appleTeamId for iOS

yanascz:


matejkriz (approved):
Works as expected.
image

PR #16816, opened 2025-02-04
fix(suite): Delete container query

jvaclavik:
## Related Issue

Resolve


enjojoy:
Can you also remove it in SelectBar please?

vojtatranta (approved):
I tested:
- collapsing of the sidebar in the wallet, work as on develop branch

Can't judge to code much - not that familiar with the style. Used context is okayish.

Can't tell if `useResponsiveContext` must be used somewhere else as well.

PR #16819, opened 2025-02-04
fix(suite-native): prevent occasional data losses

### LINKED ISSUE #16746: Occasional data loss
Several people reported data loss after updating the iOS app. They had to re-import all accounts. This Sentry issue might be related (but it's not confirmed yet)

There were also some reports from Android. There is probably different cause with the same result.

Sentry Issue: [SUITE-NATIVE-MT](https://satoshilabs.sentry.io/issues/5118977032/?referrer=github_integration)

```
Error: Calling the 'setValueWithKeyAsync' function has failed
→ Caused by: User interaction is not allowed.
  at _callSuper (app:///main.jsbundle:6:38)
...
(2 additional frame(s) were not displayed)
```

matejkriz:
## Description

From Sentry reports it seems that on iOS it happens mostly on the background without user interaction (hypothesis is that after automatic app update iOS starts the app).

On Android, it might happen during app start from time to time, not sure why and when.

When the reading from the secure store fails, we used to immediately reset the key and make all user data inaccessible. Now we will always ask the user to either try to restart the app or reset the storage if it doesn't help.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16746

## Screenshots:
image
image

### DIFF
--- a/suite-native/state/src/StoreProvider.tsx
+++ b/suite-native/state/src/StoreProvider.tsx
@@ -1,4 +1,5 @@
-import { ReactNode, useEffect, useState } from 'react';
+import { ReactNode, useEffect, useRef, useState } from 'react';
+import { AppState } from 'react-native';
 import { Provider } from 'react-redux';
 
 import { EnhancedStore } from '@reduxjs/toolkit';
@@ -14,23 +15,37 @@ type StoreProviderProps = {
 };
 
 export const StoreProvider = ({ children }: StoreProviderProps) => {
+    const initStoreCalledRef = useRef(false);
     const [store, setStore] = useState<EnhancedStore | null>(null);
     const [storePersistor, setStorePersistor] = useState<Persistor | null>(null);
 
+    const initStoreAsync = async () => {
+        initStoreCalledRef.current = true;
+        try {
+            const freshStore = await initStore();
+            const freshPersistor = persistStore(freshStore);
+            setStore(freshStore);
+            setStorePersistor(freshPersistor);
+        } catch (error) {
+            console.error('Init store error:', error);
+            Sentry.captureException(error);
+        }
+    };
     useEffect(() => {
-        const initStoreAsync = async () => {
-            try {
-                const freshStore = await initStore();
-                const freshPersistor = persistStore(freshStore);
-                setStore(freshStore);
-                setStorePersistor(freshPersistor);
-            } catch (error) {
-                console.error('Init store error:', error);
-                Sentry.captureException(error);
+        const subscription = AppState.addEventListener('change', nextAppState => {
+            if (!initStoreCalledRef.current && nextAppState === 'active') {
+                initStoreAsync();
             }
-        };
+        });
 
-        initStoreAsync();
+        if (!initStoreCalledRef.current && AppState.currentState === 'active') {
+            initStoreAsync();
+            subscription.remove();
+        }
+
+        return () => {
+            subscription.remove();
+        };
     }, []);
 
     if (store === null || storePersistor === null) return null;
--- a/suite-native/storage/src/storage.ts
+++ b/suite-native/storage/src/storage.ts
@@ -2,7 +2,7 @@ import { Alert } from 'react-native';
 import { MMKV } from 'react-native-mmkv';
 import RNRestart from 'react-native-restart';
 
-import { captureException } from '@sentry/react-native';
+import { captureException, captureMessage } from '@sentry/react-native';
 import { getRandomBytes } from 'expo-crypto';
 import * as SecureStore from 'expo-secure-store';
 import * as SplashScreen from 'expo-splash-screen';
@@ -48,6 +48,7 @@ const retrieveStorageEncryptionKey = async () => {
 };
 
 export const clearStorage = () => {
+    captureMessage('User triggered Clear App Storage action.', 'error');
     unecryptedJotaiStorage.clearAll();
     encryptedStorage?.clearAll();
     RNRestart.restart();
--- a/suite-native/storage/src/storage.ts
+++ b/suite-native/storage/src/storage.ts
@@ -16,18 +16,32 @@ export const ENCRYPTED_STORAGE_ID = 'trezorSuite-app-storage';
 export let encryptedStorage: MMKV;
 
 const retrieveStorageEncryptionKey = async () => {
+    let secureKey: string | null = null;
     try {
-        const secureKey = await SecureStore.getItemAsync(ENCRYPTION_KEY);
-
-        if (secureKey) return secureKey;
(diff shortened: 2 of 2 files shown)


suite-native/storage/src/storage.ts
```diff
@@ -39,33 +53,37 @@ export const clearStorage = () => {
     RNRestart.restart();
 };
 
-// Ideally it should never happen but we need to be sure that at least some message is displayed.
-// If someone will mess with encryptionKey it can corrupt storage and app will crash on startup.
-// Then app will hang on splashscreen indefinitely so we at least want to show some error message.
+const alertUser = () => {
+    SplashScreen.hideAsync();
+    Alert.alert(
+        'We couldn’t load your data',
```
matejkriz: Not translatable so far and it can be a small challenge here - we will probably need to store the selected language in some other storage without SecureStore to make it available here. Or we might rely on system default language here. But let's solve this together with other localization issues later.

PeKne: Relying on the system language seems reasonable to me 👍

PeKne (approved):
I don't know how to test it, but the code looks good to me.

suite-native/storage/src/storage.ts
```diff
@@ -16,18 +16,32 @@ export const ENCRYPTED_STORAGE_ID = 'trezorSuite-app-storage';
...
-        // Some users are facing an error when they uninstall the app and then reinstall it,
-        // see https://github.com/expo/expo/issues/23426
-        await SecureStore.deleteItemAsync(ENCRYPTION_KEY);
-        captureException(error);
+        // If there is an error, report it and try to read one more time.
+        captureException(error, { tags: { attempt: 1 } });
+        try {
+            // There were some trouble reading from the SecureStore,
+            // let's wait a bit to make sure it wasn't just temporary error.
+            await new Promise(resolve => setTimeout(resolve, 100));
```
Nodonisko, line 28: I not sure about this, I would try this only if anything else helped, otherwise this will always delay app startup time for 100ms which is quite a lot

suite-native/storage/src/storage.ts
matejkriz: But it's not ideal if you have a default language that we don't support, but you prefer some other supported language than English...

suite-native/storage/src/storage.ts
```diff
@@ -16,18 +16,32 @@ export const ENCRYPTED_STORAGE_ID = 'trezorSuite-app-storage';
...
+            secureKey = await SecureStore.getItemAsync(ENCRYPTION_KEY);
+        } catch (error) {
+            captureException(error, { tags: { attempt: 2 } });
```
Nodonisko, line 31: How is second attempt triggered? I don't see any invokation of this function again

Nodonisko (approved):

suite-native/state/src/StoreProvider.tsx
```diff
@@ -14,23 +15,37 @@ type StoreProviderProps = {
...
+        if (!initStoreCalledRef.current && AppState.currentState === 'active') {
+            initStoreAsync();
+            subscription.remove();
+        }
 
-        initStoreAsync();
+        return () => {
+            subscription.remove();
+        };
     }, []);
```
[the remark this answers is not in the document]
matejkriz: It is intentional, there is no need to listen to app state change if store is already initialized.

suite-native/storage/src/storage.ts
matejkriz: This is called only if it failed once - it is inside the catch block. For the first try, there is no artificial delay.

suite-native/storage/src/storage.ts
matejkriz: The second attempt is triggered by the first attempt firing any error.

suite-native/storage/src/storage.ts
```diff
@@ -39,33 +53,37 @@ export const clearStorage = () => {
     RNRestart.restart();
 };
 
-// Ideally it should never happen but we need to be sure that at least some message is displayed.
-// If someone will mess with encryptionKey it can corrupt storage and app will crash on startup.
-// Then app will hang on splashscreen indefinitely so we at least want to show some error message.
+const alertUser = () => {
+    SplashScreen.hideAsync();
```
Nodonisko, line 59: 🐰 is partially correct here, it's not clear why there duplication of hiding splashscreen, please add comment why

suite-native/storage/src/storage.ts
Nodonisko: I would not bother with this here at all, ideally no one should see this

suite-native/storage/src/storage.ts
Nodonisko: I did read it wrong, so it's fine

suite-native/storage/src/storage.ts
```diff
@@ -39,33 +53,37 @@ export const clearStorage = () => {
...
+    Alert.alert(
+        'We couldn’t load your data',
+        'Try restarting the app first. If the problem persists, you’ll need to wipe this app’s storage.',
+        [
+            {
+                text: 'Reset Storage',
+                onPress: clearStorage,
```
Nodonisko, line 66: We should probably track in Sentry if user decides to click this as error, because it is :D

suite-native/storage/src/storage.ts
matejkriz: comment added, does it makes better sense? https://github.com/trezor/trezor-suite/pull/16819/commits/ea99dc780716c8dcfc56aea1db8b0664f64646e0

suite-native/storage/src/storage.ts
```diff
@@ -39,33 +53,37 @@ export const clearStorage = () => {
...
+                text: 'Restart app',
+                onPress: () => {
+                    RNRestart.restart();
+                },
+                isPreferred: true,
+                style: 'default',
+            },
+        ],
+    );
+};
```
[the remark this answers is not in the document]
matejkriz: There is no problem in calling SplashScreen.hideAsync() multiple times and it can't happen here - if there is any issue with loading storage, app can't start and splash screen is never hidden from the App.tsx. Explanation comment added.

suite-native/storage/src/storage.ts
matejkriz: Good idea, added here [`b181a9d` (#16819)](https://github.com/trezor/trezor-suite/pull/16819/commits/b181a9d16fd7fd729fe1aa88186699a5dab110c5)

Example report https://satoshilabs.sentry.io/issues/6272721810/events/08b444862bb1487ba01a70a5cdb1d37a/

PR #16784, opened 2025-02-03
feat(suite): update network icons

izmy:
## Description

Use new network icons and add a badge for L2 (Base, Arbitrum, Optimism).

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16163

## Screenshots:

image

image

image

image

image

image

image

image

PR #16836, opened 2025-02-05
fix(product-components): CoinLogo for unsupported networks

martykan:
## Description

We use CoinLogo in Connect Explorer for some networks which are not actually supported in Suite. 

With the changes in #16784 this broke, 
since `getNetworkDisplaySymbol` failed to find the network


martykan:
Oh I see this was fixed already just now

PR #16838, opened 2025-02-05
fix(suite-native): use react-native-keyboard-controller for keyboard handling

### LINKED ISSUE #16697: Missing scroll on first Send form screen so input is hidden behind the keyboard
**Describe the bug**
I can't scroll on the first Send screen, so the Destination tag is hidden behind the keyboard.

**Info:**

-   Mobile Suite Version: 25.1.2, production
-   OS: Android

**How to reproduce**

1. Go to XRP account
1. Click on Send
1. Try to fill in the Destination tag
1. See the error

**Expected behavior**
The screen should auto-scroll so the input field is visible above the keyboard

**Screenshots**

Image

**Additional context**
This might be related or conflicting https://github.com/trezor/trezor-suite/issues/14856

yanascz:
Resolves #16697


suite-native/navigation/src/components/ScreenContentWrapper.tsx
```diff
@@ -35,14 +35,15 @@ export const ScreenContentWrapper = ({
...
             <KeyboardAwareScrollView
-                innerRef={ref => {
-                    // Assign the ref of inner ScrollView.
-                    scrollViewRef.current = ref as unknown as ScrollView;
-                }}
-                keyboardDismissMode={keyboardDismissMode}
-                keyboardShouldPersistTaps="handled"
-                contentInsetAdjustmentBehavior="automatic"
-                extraHeight={extraKeyboardAvoidingViewHeight}
+                // innerRef={ref => {
```
PeKne: this should not be removed. [useScrollView](https://github.com/trezor/trezor-suite/blob/bc7a7f999da9b9d4752ab01c8f85db3c669a0468/suite-native/navigation/src/components/ScrollViewContext.tsx#L8) hook will not work without it. It is used on some screens where we want to scroll imperatively.

yanascz: Fixed in the latest version.

Nodonisko (approved):

suite-native/module-accounts-import/src/screens/XpubScanScreen.tsx
```diff
@@ -176,7 +173,7 @@ export const XpubScanScreen = ({
         <Screen
             header={<AccountImportScreenHeader closeActionType="back" />}
             footer={<XpubHint networkType={networkType} handleOpen={handleOpenHint} />}
-            extraKeyboardAvoidingViewHeight={EXTRA_KEYBOARD_AVOIDING_VIEW_HEIGHT}
+            focusedInputBottomOffset={105}
```
PeKne:
```suggestion
            focusedInputBottomOffset={155}
```

To make the submit button visible even if the keyboard is still open.

suite-native/module-send/src/screens/SendOutputsScreen.tsx
```diff
@@ -292,6 +292,7 @@ export const SendOutputsScreen = ({
                     </Animated.View>
                 )
             }
+            focusedInputBottomOffset={148}
```
PeKne: Could you please add inline comment how did you get this exact number?

suite-native/navigation/src/hooks/useKeyboardFakeView.tsx
```diff
@@ -0,0 +1,19 @@
...
+    useKeyboardHandler({
+        onMove: event => {
+            'worklet';
+            height.value = Math.max(event.height, 0);
+        },
+    });
+
+    const keyboardFakeViewStyle = useAnimatedStyle(() => ({
+        height: height.value,
+    }));
```
PeKne: nit: you can simplify this hook by using `useAnimatedKeyboard` from reanimated lib.
```suggestion
    const keyboard = useAnimatedKeyboard();

    const keyboardFakeViewStyle = useAnimatedStyle(() => ({
        height: keyboard.height.value,
    }));
```

suite-native/navigation/src/hooks/useKeyboardFakeView.tsx
```diff
@@ -0,0 +1,19 @@
+import { useKeyboardHandler } from 'react-native-keyboard-controller';
+import Animated, { useAnimatedStyle, useSharedValue } from 'react-native-reanimated';
+
+export const useKeyboardFakeView = () => {
```
PeKne: Why is this a hook when it is returning JSX? What about turning it into a component `KeyboardFakeView`, that will be rendered in the `Screen` component?

suite-native/navigation/src/hooks/useKeyboardFakeView.tsx
PeKne: Also this should work (it is using the same hook under the hood):
```
import { useKeyboardAnimation } from "react-native-keyboard-controller";
...
const { height } = useKeyboardAnimation();



```

suite-native/navigation/src/hooks/useIsKeyboardShown.ts
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { EmitterSubscription, Keyboard, Platform } from 'react-native';
+
+export const useIsKeyboardShown = () => {
```
PeKne, line 4: This implementation is IMO redundant. This functionality already exists in  `react-native-keyboard-controller` lib. Try: 
```
import { useKeyboardAnimation } from "react-native-keyboard-controller";
...
const { height, progress } = useKeyboardAnimation();
const isKeyboardShown = progress === 1;


```

See this for reference: https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/hooks/keyboard/use-keyboard-animation

suite-native/navigation/src/components/ScreenContentWrapper.tsx
```diff
@@ -11,7 +11,7 @@ type ScreenContentProps = {
     children: ReactNode;
     isScrollable: boolean;
     hasHeader: boolean;
-    extraKeyboardAvoidingViewHeight: number;
+    focusedInputBottomOffset?: number;
     refreshControl?: ScrollViewProps['refreshControl'];
     keyboardDismissMode?: ScrollViewProps['keyboardDismissMode'];
```
PeKne: You have removed this prop from `Screen` component, so it is neither passed here. Remove it from here or put it back to `Screen`.

PeKne (commented):
I have some comments, but nothing major. Conceptually it works very well 👏

suite-native/module-send/src/screens/SendOutputsScreen.tsx
yanascz: Trial and error. 😅

suite-native/module-accounts-import/src/screens/XpubScanScreen.tsx
yanascz: My testing show it has to be 163. Fixed.

suite-native/module-send/src/screens/SendOutputsScreen.tsx
yanascz: Fixed.

suite-native/navigation/src/hooks/useKeyboardFakeView.tsx
yanascz: It seems that `useAnimatedKeyboard` works fine. 👍

suite-native/navigation/src/hooks/useKeyboardFakeView.tsx
yanascz: It was a copypasta from an example, haven't realized it's possible to convert to a component. So good point, fixed!

suite-native/navigation/src/components/ScreenContentWrapper.tsx
yanascz: Good catch, removed.

suite-native/navigation/src/hooks/useIsKeyboardShown.ts
PeKne: Does not work, closing.

PeKne (approved):

suite-native/navigation/src/hooks/useIsKeyboardShown.ts
yanascz: I don't think it's doable, types mismatch. 😕

suite-native/app/package.json
```diff
@@ -114,7 +114,7 @@
         "react-native": "0.76.1",
         "react-native-edge-to-edge": "^1.3.1",
         "react-native-gesture-handler": "^2.21.0",
-        "react-native-keyboard-aware-scroll-view": "0.9.5",
+        "react-native-keyboard-controller": "patch:react-native-keyboard-controller@npm%3A1.16.5#~/.yarn/patches/react-native-keyboard-controller-npm-1.16.5-356a673ae8.patch",
```
matejkriz: This doesn't look right, other patched packages have only the version defined in package.json

yanascz: Fixed.

suite-native/navigation/src/components/KeyboardFakeView.tsx
```diff
@@ -0,0 +1,10 @@
+import Animated, { useAnimatedKeyboard, useAnimatedStyle } from 'react-native-reanimated';
+
+export const KeyboardFakeView = () => {
+    const keyboard = useAnimatedKeyboard();
```
Nodonisko: You should use similar hook from keyboard-controller because Reanimated is not handling correctly transluscent navbar on Android.

suite-native/navigation/src/hooks/useIsKeyboardShown.ts
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { EmitterSubscription, Keyboard, Platform } from 'react-native';
```
Nodonisko: I would suggest use of https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/keyboard-events instead of RN built-in to use one library for everything and be more consistent

suite-native/navigation/src/components/KeyboardFakeView.tsx
yanascz: This view is no longer necessary, found a solution using `KeyboardStickyView`. 😎

suite-native/navigation/src/hooks/useIsKeyboardShown.ts
yanascz: Fixed in https://github.com/trezor/trezor-suite/pull/16838/commits/743066623d10e5d9e788aead1ea1a38f38e4a70e, it's much shorter. 👍

Nodonisko (approved):

PeKne (approved):
Works as expected (even better than before) with the `KeyboardStickyView` 👏

PR #16848, opened 2025-02-05
fix: for RadioCard allow optional onClick to allow single-line radio without selection

peter-sanderson:
Single line `RadioCard` does not have pointer 
image

### DIFF
--- a/packages/components/src/components/RadioCard/RadioCard.tsx
+++ b/packages/components/src/components/RadioCard/RadioCard.tsx
@@ -24,7 +24,7 @@ type AllowedFrameProps = Pick<FrameProps, (typeof allowedRadioCardFrameProps)[nu
 export type RadioCardProps = {
     isActive: boolean;
     children: ReactNode;
-    onClick: () => void;
+    onClick?: () => void;
 } & AllowedFrameProps;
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/RadioCard/RadioCard.tsx
```diff
@@ -35,7 +35,8 @@ const Wrapper = styled.div<{ $isActive: boolean } & TransientProps<AllowedFrameP
     padding: ${spacingsPx.md};
     outline: ${borders.widths.small} solid ${({ theme }) => theme.borderFocus};
     outline-offset: -${borders.widths.small};
-    cursor: pointer;
+
+    ${({ onClick }) => onClick && 'cursor: pointer;'}
```
Lemonexe, line 39: nit: I'd prefer ternary, like [it's in another similar files](https://github.com/trezor/trezor-suite/blob/8d0347348f5d928a82cd4a921801e61b9ff9382a/packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/QuickActions/TooltipRow.tsx#L37)

Lemonexe (approved):
LGTM, the change makes sense for the component, as there are similar components who defined pointer cursor by presence of onClick.
Tested on web. Took me a while to realise how to reproduce this situation, but works correctly compared to develop.

PR #16851, opened 2025-02-05
Unify timeout utils

marekrjpolak:
## Description

This PR makes `resolveAfter` util abortable by signal, therefore it can replace `createTimeoutPromise` and `abortablePromise` utils, which can be removed.


mroz22 (approved):
looks 1:1 👍

PR #16857, opened 2025-02-05
fix(message-system): set version for web testnet4 message

karliatto:
## Description
set version for web testnet4 message

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
     "timestamp": "2025-02-03T00:00:00+00:00",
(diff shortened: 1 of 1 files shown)


MiroslavProchazka:
As it the original message has been already deployed, you need to bump the sequence as well.

karliatto:
> As it the original message has been already deployed, you need to bump the sequence as well.

Good catch, I updated message id as well https://github.com/trezor/trezor-suite/pull/16857/files

MiroslavProchazka (approved):

MiroslavProchazka:
I would like also ask @matejkriz or @tomasklim to check it. From my side it is approved.

tomasklim (approved):

PR #15707, opened 2024-12-03
fix(connect): FW revision check offline handling v2

Lemonexe:
## Description

Fix handling of fetch errors when FW revision check encounters version not known locally: display "go online" banner instead of generic error banner.

Works only for Suite Desktop, see comment. On web, there is no change; offline will still be `'other-error'`

Replacement of closed #15664, [see comment](https://github.com/trezor/trezor-suite/pull/15664#pullrequestreview-2472508890)

### Testing

- You can test locally with user env `1-main | 2-main` and turning off Wi-Fi :arrow_right: go online banner
- Or go to [downloadReleasesMetadata.ts](https://github.com/trezor/trezor-suite/blob/905e656a24907aa91ae3bfdfe68388778a05d979/packages/connect/src/data/downloadReleasesMetadata.ts#L12) and switch the URL for:
  - URL where host cannot be resolved e.g. `'https://asdf.trezor.io'` :arrow_right: go online banner
  - URL returning invalid JSON e.g. `'https://trezor.io'` :arrow_right: generic error banner
    - user is online, but there was error thrown _somewhere_ during processing the http response
  - URL returning http error code e.g. `'https://data.trezor.io/whatever.json'` :arrow_right: generic error banner
    - if our servers return 404 it's our problem, definitely shouldn't prompt users to "go online"

## Screenshots:

**Before:**
other error

**After:**
offline error

PR #16859, opened 2025-02-05
Fix: proper FW revision offline error handling

### LINKED ISSUE #16725: FW revision check does not parse the releases.json correctly
As per Sentry error it looks that FW revision check does not correctly parse information from bundled releases.json file. 

https://satoshilabs.sentry.io/issues/6231913341/events/ee0f5132119d459786a7d435cf003128/?project=4504214699245568
We've received following error -
Firmware revision check failed! {"model":"T2B1","revision":"8a254aa8eae82f99630df63f40e4d290066a3efc","version":"2.8.7","vendor":"Trezor","revisionCheckError":"cannot-perform-check-offline"}

even though FW version is part of that app release, thus check should have parse this information and find it. 

It is also up to consideration whether sending these type of errors  (cannot-perform-check-offline) make sense. cc @matejkriz 

Please also investigate "Other-error" type of result - https://satoshilabs.sentry.io/issues/6231913341/events/453dbaf7af9649b3a8ca86e741e590a7/?project=4504214699245568

Lemonexe:
## Description

- connect: fix issue with identifying network errors during FW revision check
  - so that they are correctly identified as such also in react native (and not lumped as an `other-error`)
  - similar to [#15707](https://github.com/trezor/trezor-suite/pull/15707) which fixed it for Desktop
- do not report `cannot-perform-check-offline` to Sentry (does not make sense to report such a thing!)
  - rare occurence on mobile ([example event](https://satoshilabs.sentry.io/issues/6231913341/events/ee0f5132119d459786a7d435cf003128/?project=4504214699245568)), so disable it to reduce noise
  - never happens on desktop. Still, the code should reflect we don't want it to

:exclamation: root cause issue still stands - why connect does not load the releases.json and identifiy the revision **_there_**? But given the observations, it's now not necessarily mobile only! The problem could very well be on desktop too, but invisible, because on desktop, fetch error will be correctly identified as offline error, and that's not seen in Sentry. Need to investigate further..

:thinking: :thought_balloon: why are there offline errors on mobile, but never on desktop? Idk! Maybe because different Sentry SDK? In web/desktop SDK, it immediately sends an event to ingest, but you're offline so it never arrives. I guess mobile sends it later, when you are connected again. 

## Related Issue

Resolve #16725

### DIFF
--- a/packages/connect/src/device/checkFirmwareRevision.ts
+++ b/packages/connect/src/device/checkFirmwareRevision.ts
@@ -6,13 +6,21 @@ import { FirmwareRelease, VersionArray } from '../types';
 import { calculateRevisionForDevice } from './calculateRevisionForDevice';
 import { FirmwareRevisionCheckError, FirmwareRevisionCheckResult } from '../types/device';
 
-/*
- * error names that signify unavailable internet connection, see https://github.com/node-fetch/node-fetch/blob/main/docs/ERROR-HANDLING.md
- * Only works in Suite Desktop, where `cross-fetch` uses `node-fetch` (nodeJS environment)
- * In Suite Web, the errors are unfortunately indistinguishable from other errors, because they are all lumped as CORS errors
+const isNodeJSNetworkError = (e: Error) => ['FetchError', 'AbortError'].includes(e.name);
+const isReactNativeNetworkError = (e: Error) =>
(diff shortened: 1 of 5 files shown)


peter-sanderson (approved):
few nits solved over call

PR #16861, opened 2025-02-05
fix(transport): transferOut timeout must be inside try block

mroz22:
argh face-palm me 🤕 🌴. 
image

in #16793 I made some last time changes - moving the timeout outside of try block so that I can clear it in finally easily which resulted in the error from device.reset not being catched

### DIFF
--- a/packages/transport/src/api/usb.ts
+++ b/packages/transport/src/api/usb.ts
@@ -234,7 +234,7 @@ export class UsbApi extends AbstractApi {
 
         const timeout = setTimeout(() => {
             this.logger?.debug('usb: device.transfer out take suspiciously long. timing out.');
-            device?.reset();
+            device?.reset().catch(() => {});
         }, 1000);
 
         try {


marekrjpolak (commented):
I believe this change is irrelevant as errors thrown from `device.reset()` from inside 1000 ms timeout cannot be caught by synchronous try block.

marekrjpolak:
Generally I think this could be job for `scheduleAction` as it's a logic which can be either aborted from outside or timeouted from inside.

As a quicker fix, could it be possible to abort the signal in the timeout, instead of directly calling `device.reset()`? It should do the same thing inside `abortableMethod`, right?

mroz22:
> As a quicker fix, could it be possible to abort the signal in the timeout, instead of directly calling device.reset()? It should do the same thing inside abortableMethod, right?

I am afraid that it can't be done easily. This signal that is used inside `usb.write` is passed from the caller and it is the same signal that is used also for read which must not be timed-out. 

but what about simply attaching catch handler to device.reset? it should do the trick, shoudln't it? [5b7e26b](https://github.com/trezor/trezor-suite/pull/16861/commits/5b7e26b8fd341ff49a5e51e9fabd5fe49551ef42).

marekrjpolak:
> but what about simply attaching catch handler to device.reset? it should do the trick, shoudln't it?

Yeah I thought that error from `device.reset()` itself must be caught. If we presume that calling reset will cause an error to be thrown from inside the `device.transferOut()`, it should be perfectly fine.

mroz22:
>  If we presume that calling reset will cause an error to be thrown from inside the device.transferOut()

yes, that is exactly what happens

marekrjpolak (approved):

PR #16867, opened 2025-02-06
Fix android e2e tests

marekrjpolak:
## Description

Run newly added `scheduleUpgradeCheck` in `DeviceList` only when there's at least one preferred transport to which we can upgrade, which should fix Detox tests.


PeKne (approved):
Test are passing, approving ✅

PR #16868, opened 2025-02-06
fix(suite-native): welcome flow ui review fixes

PeKne:
## Description
- fixes issues found by @shenkys during welcome flow UI review

### DIFF
--- a/suite-native/module-onboarding/src/screens/AnalyticsConsentScreen.tsx
+++ b/suite-native/module-onboarding/src/screens/AnalyticsConsentScreen.tsx
@@ -1,4 +1,5 @@
 import { useState } from 'react';
+import { TouchableOpacity } from 'react-native';
 
 import { EventType, analytics } from '@suite-native/analytics';
 import { Box, Button, Card, Switch, Text, TitleHeader, VStack } from '@suite-native/atoms';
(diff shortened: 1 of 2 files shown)


suite-native/module-onboarding/src/screens/AnalyticsConsentScreen.tsx
```diff
@@ -92,23 +97,26 @@ export const AnalyticsConsentScreen = ({
...
-                                    <Switch
-                                        testID="@onboarding/AnalyticsConsent/consentSwitch"
-                                        isChecked={isEnabled}
-                                        onChange={enabled => {
-                                            setIsEnabled(enabled);
-                                        }}
-                                    />
-                                </Box>
+                                    <Box
+                                        flexDirection="row"
```
yanascz, line 105: What about using `VStack` here?

yanascz (approved):
👍

suite-native/module-onboarding/src/screens/AnalyticsConsentScreen.tsx
PeKne: VStack does not make sense here because the `flexDirection="row"`. Did you mean HStack? Anyway, I don't see any benefit of using Stack component here. Why would you like to use is?

yanascz: Sorry, `HStack` is what I meant. And I believe it makes sense since you don't have to specify direction explicitly.

Issue #16875, opened 2025-02-06
RBF transaction mined before RBFed


peter-sanderson:
Follow-up of: https://github.com/trezor/trezor-suite/issues/6810

Figma: https://www.figma.com/design/a13eFaT5HqemGmzJZy1fNg/Transactions?node-id=1-106&p=f&t=gzu2w9XlIcp7l0Rm-0

---------------

When user is trying to perform RBF or Cancel and original TX gets confirmed, the modal shall change to error state

There are 6 cases that needs to be handled correctly:

RBF:
- [x] RBF Detail Modal https://github.com/trezor/trezor-suite/pull/16893
- [x] RBF Review TX Sign https://github.com/trezor/trezor-suite/pull/16878
- [x] RBF Review TX Broadcast https://github.com/trezor/trezor-suite/pull/16878

Cancel:
- [x] Cancel Detail Modal (done in prev. PR) https://github.com/trezor/trezor-suite/pull/16525
- [x] Cancel Review TX Sign  https://github.com/trezor/trezor-suite/pull/16905
- [x] Cancel Review TX Broadcast https://github.com/trezor/trezor-suite/pull/16905

bosomt:
**QA OK**

- Cancel Detail Modal ✅  /Cancel transaction failed/
- Cancel Review TX Sign ✅  /Cancel transaction failed/
- Cancel Review TX Broadcast ✅  /Cancel transaction failed/
- RBF Review TX Broadcast ✅  /Replace transaction failed/
- RBF Review TX Sign ✅   /Replace transaction failed/
- RBF Detail Modal ✅  /Replace transaction failed/

some duplicated errors in notifications but its kinda expected/not worth fixing
`Transaction signing error: replace-by-fee-error-transaction-already-mined `

**Info:**
 - Suite version: web 25.3.0 (982a0ba78565f999531cac20d5bf9c2730f0178d)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T3T1 2.8.7 regular (revision f2aeea159f1fde98d4fc5575350a8c9fc4478367)
 - Transport: WebUsbTransport

PR #16878, opened 2025-02-06
feat: add error state for RBF when previous transaction gets mined befor user sends the modal

peter-sanderson:
Partially addresses https://github.com/trezor/trezor-suite/issues/16875

This PR handles the situation in RBF flow, when block is mined when modal is opened. 

## Case `#1`
- User opens modal and submits the RBF
- Then we await confirmation on device for amount, address, fee, ...
image
- Then, before user confirms the block is mined and original Tx gets confirmed
- User sees this error screen:
image

## Case `#2`
- User opens modal and submits the RBF
- user does all the confirmation on the device and is about to broadcast TX
image
- Then, before user confirms the block is mined and original Tx gets confirmed
- User sees this error screen:
image


packages/suite/src/actions/wallet/send/sendFormThunks.ts
```diff
@@ -227,7 +227,13 @@ export const signAndPushSendFormTransactionThunk = createThunk(
         );
 
         if (isRejected(signResponse)) {
-            // close modal manually since UI.CLOSE_UI.WINDOW was blocked
+            // Do not close the modal, as we need that modal to display the error state.
+            if (signResponse.payload?.message === RBF_ERROR_ALREADY_MINED) {
```
peter-sanderson, line 231: This is ugly hack, that unfortunately have to employ. The similar mechanism is used in wallet-switcher.

This is currently the only way, how to send some date from place where cancel on device happen to the place where result of cancelled operation is awaited

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
```diff
@@ -164,17 +173,85 @@ export const TransactionReviewModalContent = ({
...
-            <ConfirmOnDevice
-                title={<Translation id="TR_CONFIRM_ON_TREZOR" />}
-                steps={outputs.length + 1}
-                activeStep={serializedTx ? outputs.length + 2 : buttonRequestsCount}
-                deviceModelInternal={deviceModelInternal}
-                deviceUnitColor={device?.features?.unit_color}
-                successText={<Translation id="TR_CONFIRMED_TX" />}
-                onCancel={onCancel}
-            />
+            {!isRbfConfirmedError && (
```
peter-sanderson, line 246: Originally, I tried to create separate modal for this error state, but I end-up copying most of the code anyway. So I discarded that approach and fall-backed to this "put if everywhere" approach. I do not particularly like it, but it would need some heavy refactoring of this modal to avoid this.

suite-common/wallet-core/src/transactions/transactionsActions.ts
```diff
@@ -21,26 +21,30 @@ const removeTransaction = createAction(
...
-        perPage?: number;
-    }): {
-        payload: {
-            transactions: WalletAccountTransaction[];
-            account: Account;
-            page?: number;
-            perPage?: number;
-        };
-    } => ({
+    }: AddTransactionActionProps): AddTransactionActionResult => ({
```
Lemonexe, line 47: :pray: thank you!

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/ReplaceByFeeFailedOriginalTxConfirmed.tsx
```diff
@@ -0,0 +1,51 @@
...
+import {
+    HELP_CENTER_CANCEL_TRANSACTION,
+    HELP_CENTER_REPLACE_BY_FEE_BITCOIN,
+    Url,
+} from '@trezor/urls';
+
+import { Translation, TranslationKey } from '../../../../Translation';
+import { TrezorLink } from '../../../../TrezorLink';
+
+type ReplaceByFeeFailedOriginalTxConfirmedParams = {
```
Lemonexe: tiny nit: `Props`, not `Params`

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
```diff
@@ -42,12 +44,16 @@ type TransactionReviewModalContentProps = {
     decision: Deferred<boolean, string | number | undefined> | undefined;
     txInfoState: SendState | StakeState;
     cancelSignTx: () => void;
+
+    // Just bool for now, but could be extended into generic error in the future
```
Lemonexe: I don't think this comment is needed - it's trivial

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
```diff
@@ -164,17 +173,85 @@ export const TransactionReviewModalContent = ({
...
+    const BottomContent = () => {
+        if (isRbfConfirmedError) {
+            return (
+                <NewModal.Button variant="tertiary" onClick={onCancel}>
+                    <Translation id="TR_CLOSE" />
+                </NewModal.Button>
+            );
+        }
+
+        if (!areDetailsVisible && isBroadcastEnabled) {
```
Lemonexe: When I look at the original condition it says
`!areDetailsVisible && (isBroadcastEnabled ? <JSX1 /> : <JSX2 />)`
You did not keep the same logic, is this intentional?
I believe this upholds original behavior:
```suggestion
        if (areDetailsVisible) return null; 
        if (isBroadcastEnabled) {
```

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
Lemonexe: When I am freshly seeing this, it seems absolutely fine to me.

It's the same modal, but with three variants that change only some parts inside.
Yes, it could be split it into three components, but then you'd have to also create a "TransactionReviewModalLayout" component with slot props, if you wanted to keep it dry. I think it'd be needlessly complicated.

Lemonexe (approved):
Looks okay.  The redux logic is quite convoluted, but as discussed personally, I don't have an idea how to do it more systematically.

Please see the comments before merging, one condition needs to be changed I think.

Tested and works well!

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/ReplaceByFeeFailedOriginalTxConfirmed.tsx
peter-sanderson: fixed

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
peter-sanderson: fixed

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
peter-sanderson: yes, you're right, nice catch, thank you, fixed

Lemonexe (approved):

bosomt:
**QA OK**

- link to learn more to https://trezor.io/learn/a/replace-by-fee-rbf-bitcoin ✅ 
- bump auto failed after tx was confirmed after i waited with open bump modal/

Screenshot 2025-02-12 at 15 43 42

**Info:**
 - Suite version: web 25.3.0 (7282d84016cee13eb830ac694a4ff6ff86b410a9)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Transport: WebUsbTransport

PR #16893, opened 2025-02-07
feat: add error screen for RBF modal when TX is confirmed

peter-sanderson:
Partially addresses https://github.com/trezor/trezor-suite/issues/16875

image


Lemonexe (approved):
LGTM. Especially thanks for the overview in the issue, otherwise I'd be quite confused what is this compared to the previous PR :+1:

PR #16897, opened 2025-02-10
chore(transport): update long from 4.0.0 to 5.2.0

mroz22:
a major version but it should probably be safe. It should be mostly package setup changes such as this commit https://github.com/dcodeIO/long.js/commit/86410397c36ad64aa6c9a45293bff918ce0ea14d#diff-c3eb48bba9ad6a93ead144686134ad85a923fb766279bee42463209d70d91aef 

also this part of our codebase should be covered by tests exstensively.

### DIFF
--- a/packages/transport/package.json
+++ b/packages/transport/package.json
@@ -67,7 +67,7 @@
         "@trezor/protocol": "workspace:*",
         "@trezor/utils": "workspace:*",
         "cross-fetch": "^4.0.0",
-        "long": "^4.0.0",
+        "long": "^5.2.0",
         "protobufjs": "7.4.0",
         "usb": "^2.14.0"
     },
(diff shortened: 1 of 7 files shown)


marekrjpolak (approved):
It seems that `long` package really is needed for `protobufjs` to work in browser, even if it's also nested dependency. Soooo, :seal: :heavy_check_mark:

packages/transport/package.json
```diff
@@ -67,8 +67,6 @@
         "@trezor/protocol": "workspace:*",
         "@trezor/utils": "workspace:*",
         "cross-fetch": "^4.0.0",
-        "long": "^4.0.0",
-        "protobufjs": "7.4.0",
```
marekrjpolak, line 71: Seems that protobufjs shouldn't have gone away.

mroz22: hmm yes, there is still one more runtime usecase in packages/transport. maybe in the future it should be factored out

PR #16905, opened 2025-02-10
feat: implement distinction between BumpFee RBF & Cancel RBF -> utilise it to display error in case TX gets confirmed in cancel flow

### LINKED ISSUE #16875: RBF transaction mined before RBFed

peter-sanderson:
Resolve: https://github.com/trezor/trezor-suite/issues/16875

image

image

### DIFF
--- a/packages/suite-analytics/src/types/events.ts
+++ b/packages/suite-analytics/src/types/events.ts
@@ -36,6 +36,24 @@ export type SuiteAnalyticsEventSuiteReady = {
     };
 };
 
+export type TransactionCreatedEvent = {
+    type: EventType.TransactionCreated;
+    payload: {
+        action: 'sent' | 'copied' | 'downloaded' | 'replaced' | 'canceled';
+        symbol: string;
+        tokens: string;
+        outputsCount: number;
+        broadcast: boolean;
+        bitcoinLockTime: boolean;
+        ethereumData: boolean;
+        ethereumNonce: boolean;
+        rippleDestinationTag: boolean;
+        selectedFee: string;
+        isCoinControlEnabled: boolean;
+        hasCoinControlBeenOpened: boolean;
+    };
+};
+
 export type SuiteAnalyticsEvent =
     | SuiteAnalyticsEventSuiteReady
     | {
@@ -198,23 +216,7 @@ export type SuiteAnalyticsEvent =
               type: 'exchange' | 'buy' | 'sell';
           };
       }
-    | {
-          type: EventType.TransactionCreated;
-          payload: {
-              action: 'sent' | 'copied' | 'downloaded' | 'replaced';
-              symbol: string;
-              tokens: string;
-              outputsCount: number;
-              broadcast: boolean;
-              bitcoinLockTime: boolean;
-              ethereumData: boolean;
-              ethereumNonce: boolean;
-              rippleDestinationTag: boolean;
-              selectedFee: string;
-              isCoinControlEnabled: boolean;
-              hasCoinControlBeenOpened: boolean;
-          };
-      }
+    | TransactionCreatedEvent
     | {
           type: EventType.SendRawTransaction;
           payload: {
--- a/packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
@@ -10,16 +10,18 @@ import {
     selectSendFormReviewButtonRequestsCount,
     selectStakePrecomposedForm,
 } from '@suite-common/wallet-core';
-import { FormState, StakeFormState } from '@suite-common/wallet-types';
+import { FormState, RbfTransactionType, StakeFormState } from '@suite-common/wallet-types';
 import {
     constructTransactionReviewOutputs,
     getTxStakeNameByDataHex,
+    isRbfBumpFeeTransaction,
+    isRbfCancelTransaction,
     isRbfTransaction,
 } from '@suite-common/wallet-utils';
 import { NewModal } from '@trezor/components';
 import { copyToClipboard, download } from '@trezor/dom-utils';
 import { ConfirmOnDevice } from '@trezor/product-components';
-import { EventType, analytics } from '@trezor/suite-analytics';
+import { EventType, TransactionCreatedEvent, analytics } from '@trezor/suite-analytics';
 import { Deferred } from '@trezor/utils';
 
 import * as modalActions from 'src/actions/suite/modalActions';
@@ -39,6 +41,14 @@ const isStakeState = (state: SendState | StakeState): state is StakeState => 'da
 
 const isStakeForm = (form: FormState | StakeFormState): form is StakeFormState =>
     'stakeType' in form;
+
+const mapRbfTypeToReporting: Record<
+    RbfTransactionType,
+    TransactionCreatedEvent['payload']['action']
+> = {
+    'bump-fee': 'replaced',
+    cancel: 'canceled',
+};
 
 type TransactionReviewModalContentProps = {
     decision: Deferred<boolean, string | number | undefined> | undefined;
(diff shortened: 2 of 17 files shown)


packages/suite/src/utils/suite/transactionReview.ts
```diff
@@ -22,10 +24,14 @@ export const getTransactionReviewModalActionText = ({
         // no default
     }
 
-    if (isRbfAction) {
+    if (isBumpFeeRbfAction) {
         return 'TR_REPLACE_TX';
     }
 
+    if (isCancelRbfAction) {
+        return 'TR_CANCEL_TX';
```
peter-sanderson, line 32: This allow us to change Header of the send TX modal (confirm on device part)

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/TxDetailModal/CancelTransaction/CancelTransactionModal.tsx
```diff
@@ -69,7 +70,12 @@ export const CancelTransactionModal = ({
 
         dispatch(composeCancelTransactionThunk({ account, tx, chainedTxs }))
             .unwrap()
-            .then(setComposedCancelTx)
+            .then(precomposed => {
+                setComposedCancelTx({
+                    ...precomposed,
+                    prevTxToBeCanceledId: tx.txid,
```
peter-sanderson: This is the most important place where it all starts

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
```diff
@@ -223,7 +230,11 @@ export const TransactionReviewModalContent = ({
         }
 
         if (isRbfConfirmedError) {
-            return <ReplaceByFeeFailedOriginalTxConfirmed type="replace-by-fee" />;
+            return (
+                <ReplaceByFeeFailedOriginalTxConfirmed
+                    type={isCancelRbfAction ? 'cancel-transaction' : 'replace-by-fee'}
```
peter-sanderson: Not nice I know.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
```diff
@@ -148,7 +157,10 @@ export const TransactionReviewModalContent = ({
         }
         if (decision) {
             decision.resolve(true);
-            reportTransactionCreatedEvent(isRbfAction ? 'replaced' : 'sent');
+            reportTransactionCreatedEvent(
+                // eslint-disable-next-line no-nested-ternary
+                isBumpFeeRbfAction ? 'replaced' : isCancelRbfAction ? 'canceled' : 'sent',
```
peter-sanderson: Not nice I know.

packages/suite-analytics/src/types/events.ts
```diff
@@ -201,7 +201,7 @@ export type SuiteAnalyticsEvent =
     | {
           type: EventType.TransactionCreated;
           payload: {
-              action: 'sent' | 'copied' | 'downloaded' | 'replaced';
+              action: 'sent' | 'copied' | 'downloaded' | 'replaced' | 'canceled';
```
peter-sanderson: Updated doc in https://www.notion.so/satoshilabs/transaction-created-16a7363c8ab14dce95d506d26ff2adcf

Lemonexe: Are you sure it's updated? It doesn't mentioned canceling, and the page was last modified in November.

suite-common/wallet-types/src/transaction.ts
```diff
@@ -120,16 +120,23 @@ export type PrecomposedTransactionCardanoFinal =
...
     feeDifference: string;
     // Native RBF is a firmware feature to recognize an RBF transaction and simplify transaction review flow.
     useNativeRbf: boolean;
     useDecreaseOutput: boolean;
 };
 
-// strict distinction between normal and RBF type
-export type PrecomposedTransactionFinal = TxFinal | PrecomposedTransactionFinalRbf;
+export type PrecomposedTransactionFinalCancelRbf = TxFinal & {
+    prevTxToBeCanceledId: string;
```
Lemonexe: I see that the you renamed `prevTxid` to `prevTxToBeCanceledId`, so you could distinguish the interfaces in the typeguard function. I don't find this approach ideal.
:thought_balloon:  I'd prefer consistent common interfaces, so how about name both `prevTxid` because they are used analogically, and instead add `rbfType: 'cancel' | 'bump-fee` to distinguish?

Pros:
- the type guards no longer rely on the fact that a property _is differently named_
- the type guards code would read very naturally
- no need for `getPrevTxid` in `replaceByFeeErrorMiddleware` because the interface is consistent

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
Lemonexe: Indeed. I'd use eslint disable sparingly, and in this case I find early return much more suitable, and it's even shorter! :smile: 
```
            if (isBumpFeeRbfAction) return reportTransactionCreatedEvent('replaced');
            if (isCancelRbfAction) return reportTransactionCreatedEvent('canceled');
            reportTransactionCreatedEvent('sent');
```

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
Lemonexe: This one is IMHO absolutely fine :+1:

Lemonexe (approved):
LGTM with comments :left_speech_bubble: 
Technically OK, but I'd strongly advocate to refactor the interfaces as per comment before merging. Plus the nested ternary but that's a nit.

packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewModalContent.tsx
```diff
@@ -70,10 +71,13 @@ export const TransactionReviewModalContent = ({
     );
 
     const isTradingAction = !!precomposedForm?.isTrading;
-    const isRbfAction = precomposedTx !== undefined && isRbfTransaction(precomposedTx);
+    const isBumpFeeRbfAction =
+        precomposedTx !== undefined && isRbfBumpFeeTransaction(precomposedTx);
```
PeKne, line 84: nit: handling of undefined could be moved inside of the `isRbfBumpFeeTransaction` type guard.

suite-common/wallet-types/src/transaction.ts
peter-sanderson: This was very good call. I was thinking about it but I was so annoyed I didn't want to dig down into BumpFee RBF wtfness. 

However when I changed it I discovered it made the original Bump Fee flow more elegant.

suite-common/wallet-core/src/send/sendFormThunks.ts
```diff
@@ -533,25 +533,26 @@ export const enhancePrecomposedTransactionThunk = createThunk<
...
+                };
+
+                return enhancedRbfPrecomposedTx;
             }
-        }
+
+            return precomposedTransaction;
+        };
+
+        const enhancedPrecomposedTransaction = createGeneralPrecomposedTransactionFinal();
```
Lemonexe, line 555: This whole diff is awesome :tada:  the original code was barely legible with `(as)` on lefthand side of assignment operation :sweat_smile:

Lemonexe (approved):

PeKne (approved):
Looks good from suite-native point of view. Did not test desktop implementation.

Lemonexe (approved):
Even better, amazing refactoring in last fixup  :ok_hand:  :ok_hand:  :ok_hand: 
Thank you for taking such care with codebase :hugs:

PR #16906, opened 2025-02-10
fix(suite-native): navigation animation tweaks

PeKne:
## Description
- `AuthorizeDeviceStack` do not reuse `stackNavigationOptionsConfig` since it is redundant and already applied in the parent. This makes the animation of this stack work as expected `slide_from_bottom`)
- OnboardingStack does not stard with `slide_from_bottom` animation. That screen is preceeded with Connect & Unlock Screen, so there should not be a second from bottom animation.


yanascz (approved):
Existing flows seem to be fixed. The device onboarding flow doesn't work as I'd expect but let's discuss offline since it's behind a FF.

PR #16916, opened 2025-02-11
fix(suite-native): has account transactions selector

### LINKED ISSUE #16814: XRP account detail screen shows graph even when empty
**How to reproduce**
1. Enable XRP in Settings.
1. Connect a Trezor and wait for discovery to finish.
1. Tap any empty XRP account.
1. The details screen has invalid content.

**Expected behavior**
The detail screen looks the same as for any other coin without transactions.

**DEV notes**
It seems to be caused by the `selectHasAccountTransactions` selector that returns `true` for empty XRP accounts.

**Screenshots**
Image
Image

PeKne:
## Description
- suite-common`selectHasAccountTransactions` was used but it does not work for XRP accounts because the Ripple backend does not provide total transactions count info
Screenshot 2025-02-11 at 9 46 42 AM
  - this selector was renamed to make this backend discrepancy clear 
  - new correct selector was created specifically for suite-native

## Related Issue

Resolve #16814 

## Screenshots:

https://github.com/user-attachments/assets/ce40449d-a8e2-4da1-b2db-90ae8be34037

### DIFF
--- a/suite-common/wallet-core/src/accounts/accountsReducer.ts
+++ b/suite-common/wallet-core/src/accounts/accountsReducer.ts
@@ -280,7 +280,10 @@ export const selectAccountByKey = createMemoizedSelector(
     },
 );
 
-export const selectHasAccountTransactions = createMemoizedSelector(
+// CAUTION!: This selector does not work for XRP accounts! It should be used only for Bitcoin-like accounts.
+// Ripple backend does not provide the total transaction count info.
+// The property`account.history.total` is always equal to -1 for XRP accounts.
+export const selectHasBitcoinAccountTransactions = createMemoizedSelector(
     [selectAccountByKey],
     account => !!account?.history.total,
 );
--- a/suite-native/tokens/src/tokensSelectors.ts
+++ b/suite-native/tokens/src/tokensSelectors.ts
@@ -163,6 +163,11 @@ export const selectAccountTransactionsWithTokenTransfers = createMemoizedSelecto
         ) as WalletAccountTransaction[],
 );
 
+export const selectHasAccountAnyTransactions = createMemoizedSelector(
+    [selectAccountTransactionsWithTokenTransfers],
+    (transactions): boolean => transactions.length > 0,
+);
+
 export const selectAccountsKnownTokens = createMemoizedSelector(
     [selectAccountByKey, selectTokenDefinitions],
     (account, tokenDefinitions): TokenInfoBranded[] => {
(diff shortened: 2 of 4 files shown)


vytick (approved):
Looks good and works in mobile. Good spot 🚀

suite-native/tokens/src/tokensSelectors.ts
```diff
@@ -163,6 +163,11 @@ export const selectAccountTransactionsWithTokenTransfers = createMemoizedSelecto
         ) as WalletAccountTransaction[],
 );
 
+export const selectHasAccountAnyTransactions = createMemoizedSelector(
+    [selectAccountTransactionsWithTokenTransfers],
+    (transactions): boolean => transactions.length > 0,
```
yanascz, line 168: What will the return value be for e.g. an ETH account without base transactions but some token transactions?

yanascz: OK, it seems that the base account shows all the transactions including token transactions. Never mind.

suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -280,7 +280,10 @@ export const selectAccountByKey = createMemoizedSelector(
     },
 );
 
-export const selectHasAccountTransactions = createMemoizedSelector(
+// CAUTION!: This selector is not working for XRP accounts!
+// Ripple backend does not provide the total transaction count info.
+// The property`account.history.total` is always equal to -1 for XRP accounts.
+export const selectHasNonRippleAccountTransactions = createMemoizedSelector(
```
yanascz: The name of the selector is misleading... I'd prefer something like `selectHasAccountAnyTransactionHistory`, `selectHasAccountTransactionHistory` or similar.

suite-common/wallet-core/src/accounts/accountsReducer.ts
```diff
@@ -280,7 +280,10 @@ export const selectAccountByKey = createMemoizedSelector(
     },
 );
 
-export const selectHasAccountTransactions = createMemoizedSelector(
+// CAUTION!: This selector is not working for XRP accounts!
```
yanascz: I believe _does not work_ is grammatically more suitable.

suite-common/wallet-core/src/accounts/accountsReducer.ts
PeKne: fixed ca4731a541

suite-common/wallet-core/src/accounts/accountsReducer.ts
PeKne: fixed ca4731a541

yanascz (approved):

PR #16918, opened 2025-02-11
fix(suite-native): develop build pipeline fix

PeKne:
- Apply the same eas version to develop builds as this PR did to other tracks: https://github.com/trezor/trezor-suite/pull/16887


yanascz (approved):

vytick (approved):
looks good

PR #16922, opened 2025-02-11
fix(suite-native): use keyboard avoiding view for send screen

### LINKED ISSUE #16697: Missing scroll on first Send form screen so input is hidden behind the keyboard

yanascz:
## Related Issue

Resolves #16697

### DIFF
--- a/suite-native/module-send/src/components/SendScreen.tsx
+++ b/suite-native/module-send/src/components/SendScreen.tsx
@@ -9,7 +9,7 @@ type SendScreenProps = {
 };
 
 export const SendScreen = ({ children, footer, screenHeader }: SendScreenProps) => (
-    <Screen header={screenHeader} footer={footer} keyboardDismissMode="on-drag">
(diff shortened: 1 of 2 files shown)


yanascz:
According to my additional testing, this doesn't fix the issue properly. Scrolling is fixed but some UI glitches arised. Will attempt to fix properly in https://github.com/trezor/trezor-suite/pull/16838.

PR #16923, opened 2025-02-11
Experiment - mock debounce in send form unit test

martykan:
This PR is attempting to address a failing unit test related to ETH fee estimation in the send form. There is some nondeterministic behavior - sometimes the test makes 3 calls instead of 2.

Screenshot 2025-02-17 at 08 59 49

My hypothesis is that this is caused by timing related factors, so I’m trying to mock the debounce function to eliminate waiting time to get a more consistent test outcome.


mroz22 (approved):
looks legit

PR #16927, opened 2025-02-11
fix(connect): test install to use target ES2019

karliatto:
## Description
It looks like target `ES2022` was not required, having `ES2019` with `DOM,ES2022.error` lib was enough to satisfy typescript.

### DIFF
--- a/packages/connect/e2e/test-yarn-install.sh
+++ b/packages/connect/e2e/test-yarn-install.sh
@@ -22,5 +22,5 @@ echo import TrezorConnect from \"@trezor/connect\" >index.ts
 
 # compile with typescript
 yarn add typescript@5.5.4
-yarn tsc ./index.ts --types node,w3c-web-usb --esModuleInterop --target ES2022 --module commonjs 
+yarn tsc ./index.ts --types node,w3c-web-usb --esModuleInterop --lib ES2022,DOM --target ES2019 --module commonjs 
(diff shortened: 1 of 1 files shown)


mroz22 (commented):
yes, lets target the minimum required levels. 
didn't you want to change 
```
--lib ES2022,DOM,ES2022.error
```
 to
```
--lib DOM,ES2022.error
```
?

karliatto:
> yes, lets target the minimum required levels. didn't you want to change
> 
> ```
> --lib ES2022,DOM,ES2022.error
> ```
> 
> to
> 
> ```
> --lib DOM,ES2022.error
> ```
> 
> ?

I tested and ES2022 is required but `ES2022.error` was not. So I removed it.

karliatto:
It will also work with `--target ES2015` but I set it for `ES2019` since is what we use in the project.

karliatto:
@mroz22 What do you think? should we do like this? 
We can do a minor release and add a good explanation to the changelog regarding the types.
At the end it should just work adding to lib configuration in typescript ` ES2022,DOM` if they are using some target below ES2022.
If they are using target above ES2022 there should not be any configuration changed.

karliatto:
Closing since we have made ES2022 the default target for connect libraries.

https://github.com/trezor/trezor-suite/pull/17010

PR #16930, opened 2025-02-11
fix(connect): releasing device on browser reload using sendBeacon api

mroz22:
still wip... ignore 

fig 1: firing release request. query string parameter has only informative purpose 

image

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -550,6 +550,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
                     // note 3: in 99% of cases we send this message unnecessarily. as @Szymon pointed out, it might be better to catch this call and repeat it.
                     // note 4: this case can happen also in the 'if' branch. 1] reload app, 2], browser doesn't fire release in time, 3] you get unacquired device, 4] you click
                     //         the 'use device here' button and here you go. Yet I didn't want to burden every TrezorConnect method call with this but we may reconsider this.
+                    // note 5: ad note 4. it is not so problematic anymore since cleanup on dispose has been improved in https://github.com/trezor/trezor-suite/pull/16930
                     if (['v1', 'bridge'].includes(this.protocol.name)) {
                         _log.debug(
                             'sending a preventive cancel on the first encounter with the device',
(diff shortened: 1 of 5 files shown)


packages/suite-web/e2e/tests/wallet/discovery.test.ts
```diff
@@ -29,6 +31,18 @@ describe('Discovery', () => {
         cy.log('all available networks should return something from discovery');
 
         cy.getTestElement('@dashboard/loading', { timeout: 1000 * 10 });
+
+        // wait randomly between 1000 and 4000 ms
+        cy.wait(getRandomInt(1, 40) * 100);
```
marekrjpolak, line 36: This isn't 1000 to 4000 ms.
```suggestion
        // wait randomly between 1000 and 4000 ms
        cy.wait(getRandomInt(1, 4) * 1000);
```

packages/connect/src/device/Device.ts
```diff
@@ -1170,12 +1171,13 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
-    async dispose() {
+    dispose() {
         this.removeAllListeners();
         if (this.session && this.lastAcquiredHere) {
             try {
-                await this.cancelableAction?.();
-                await this.commands?.cancel();
+                // note for review: these deleted lines might not be needed anymore - in case of reload we don't have enough time
+                // anyway, so seems to be enough to release session so that we don't get 'unacquired' device after reloading. If
+                // anything stays on the screen, it will get flushed away by the initial Cancel (see device run method)
```
marekrjpolak: This seems true. The only thing `commands.cancel()` does is setting `disposed` flag and the only thing `cancelableAction()` may do is to send Cancel to device, which is now done in `runInner`.

In that case, this try/catch can be completely removed and `this.transport.release` returned directly.

marekrjpolak (approved):
Looks good in general. I also tried some cases and it behave reasonably (second tab cannot acquire the device until the first one is open, but when I close it, the second one send Cancel and continue normally), but I believe in your testing more in this case.

mroz22:
> I also tried some cases and it behave reasonably (second tab cannot acquire the device until the first one is open

Hmm this shoudln't be happening. It may now only happen for webusb. more in here #16928

bosomt:
**QA OK**

**Info:**
 - Suite version: web 25.3.1 (e5b4203432360b831bb87d2bbfb2e0131e8aeae0)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1421x838
 - Device: Trezor T2B1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)

PR #16943, opened 2025-02-12
chore(monorepo): unify copy-webpack-plugin default import naming

mroz22:
absolute OCD nitpick. unifies mixed naming of CopyPlugin and CopyWebpackPlugin


karliatto (approved):
Beautiful absolute OCD :1st_place_medal:

PR #16949, opened 2025-02-12
feat: remove Hash-Check from FW upgrade (it is performed on-device-connect anyway

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite-private/issues/129


mroz22 (approved):
ack

packages/connect/src/types/api/__tests__/management.ts
```diff
@@ -38,23 +38,6 @@ export const management = async (api: TrezorConnect) => {
...
-    }
-
-    const fwAuto = await api.firmwareUpdate({
-        btcOnly: false,
-        language: 'en-EN',
-        baseUrl: 'https://example.com',
-    });
-
-    if (fwAuto.success) {
-        fwAuto.payload.check.toLowerCase();
```
Lemonexe, line 55: not important, but what was this code actually supposed to do? 
`String.toLowerCase()` does not mutate, so this code does nothing really... :smile:

Lemonexe (approved):
LGTM!

PR #16960, opened 2025-02-12
fix(suite-native): handle device authenticity check for expired configs

yanascz:
Do not show devices as compromised if the app has expired DAC config.


PeKne (approved):
LGTM 🚀

STew790:
**QA OK**
Tested together with @MiroslavProchazka and works correctly.

**Info**
25.2.1  5336151

PR #16963, opened 2025-02-12
fix(suite): Disable update tooltip when discovery is in progress

jvaclavik:
## Related Issue

Resolve 

## Screenshots:

Screenshot 2025-02-12 at 17 01 26

### DIFF
--- a/packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/QuickActions/QuickActionButton.tsx
+++ b/packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/QuickActions/QuickActionButton.tsx
@@ -2,23 +2,20 @@ import React, { ReactNode } from 'react';
 
 import styled from 'styled-components';
 
-import { Tooltip } from '@trezor/components';
+import { ManagedTooltipProps, Tooltip } from '@trezor/components';
 
 const Container = styled.div`
     height: 44px;
     flex: 1;
-
     display: flex;
     align-items: center;
     justify-content: center;
-
-    cursor: pointer;
 `;
 
 type ActionButtonProps = {
     onClick?: () => void;
     children: ReactNode;
-    tooltip: ReactNode;
+    tooltip: Partial<ManagedTooltipProps>;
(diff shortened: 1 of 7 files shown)


vojtatranta:
No issue ID here?

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/QuickActions/QuickActionButton.tsx
```diff
@@ -30,7 +27,7 @@ export const QuickActionButton = ({
     'data-testid': dataTest,
     isOpen,
 }: ActionButtonProps) => (
-    <Tooltip content={tooltip} cursor="pointer" isOpen={isOpen}>
+    <Tooltip content={tooltip?.content} cursor="pointer" {...tooltip} isOpen={isOpen}>
```
vojtatranta, line 30: Yeah I don't like the API, but there isn't much better API we can think of.

Maybe like if we only pass `content` we could just have tooltip prop to be `tooltip: ReactNode | ManagedTooltipProps`, But I guess it wouldn't be worth the time. Altought, this is not performant due to repeated creation of the `tooltip` object without `useMemo()`. Hopefully, React compiler saves us.

So yeah, I'd keep it.

vojtatranta (approved):
I 
- checked the code
- tested the unavailability during discovery
- availability after discovery with and without available update

Works OK

MiroslavProchazka:
Rebasing so this pass the checks, please merge after we get green pipeline

bosomt:
**QA OK**

image

after discovery finishes its available 

**Info:**
 - Suite version: desktop 25.2.2 (dd7673ce898ea18b90daa91f21edbb25a4268c46)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.2.2 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956

Issue #16452, opened 2025-01-17
Top banners for FW revision check


Lemonexe:
Part of #16273: implement top banners for failed Firmware revision check in Suite Lite
[This figma frame](https://www.figma.com/design/eKie94qfryjTHEnR8jfmgk/Security-checks?node-id=4035-4984&t=fv9Xv7X06uaZhw6Q-4) captures the scope of this issue.

- only when Feature Flag is on:
- modify offline banner: display it also when FW revision check resulted in "offline error", with additional subtitle 
- add a new banner for dismissed failed FW revision check
  - visible at any page except the DeviceCompromisedModal

STew790:
**QA OK**

![Image](https://github.com/user-attachments/assets/6e00cd1e-a2c7-4240-875f-1175b2bd2c42)

![Image](https://github.com/user-attachments/assets/3ade68da-69d7-4c86-89cf-f50f45b1fa4e)

![Image](https://github.com/user-attachments/assets/71264416-67e9-463d-b03d-4f76658449fe)

![Image](https://github.com/user-attachments/assets/45571c18-b749-4ef6-8a24-53fd339c9513)

Apart from this issue here: https://github.com/trezor/trezor-suite/issues/17133

**Info**
https://github.com/trezor/trezor-suite/commit/38e8a08071126373a634368dc73eb68cda72d6cf 25.2.2

PR #16959, opened 2025-02-12
feat(suite-native): show banner for DeviceCompromised

Lemonexe:
## Description

- update support/learn more links to placeholder URLs, where new articles will be created
- add helpers useNavigationRoute, useNavigationRouteMatch
- create DeviceCompromisedBanner
  - is rendered according to a jotai atom value, similarly to `Alert` and `Banner` but simpler 
  - show the brief banner on most pages, show extended banner at Home, and hide it on the DeviceCompromisedModal

## Related Issue

Part of #16452 (except the offline banner)

## Screenshots:

Video of the whole flow, to show that the detection of route is reliable:
[Screencast from 2025-02-12 13-55-34.webm](https://github.com/user-attachments/assets/3a2272f9-639a-4d15-87d8-2fa6611eff5e)

Together with OfflineBanner
Screenshot from 2025-02-14 09-28-04

PR #16965, opened 2025-02-12
chore(suite-native): refactor navigation route matching

Lemonexe:
## Description

Tweak navigation route matching in `useHandleDeviceConnection`. I searched for all occurrences of `.routes`, but did not find any other files which could do with such refactoring.

I tested that in following cases, the navigation behavior is **unchanged**:
- already initialized device: connect leads to connecting screen & disconnect to Home (no device connected screen)
- uninitialized T2B1 :arrow_right: click no in Security check, gets to Suspicious device screen, when I disconnect the screen stays and only after closing I am navigated back to Home (no device connected screen) 

## Related Issue

Related to https://github.com/trezor/trezor-suite/issues/16452

### DIFF
--- a/suite-native/device/src/hooks/useHandleDeviceConnection.ts
+++ b/suite-native/device/src/hooks/useHandleDeviceConnection.ts
@@ -1,7 +1,7 @@
 import { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
-import { useNavigation } from '@react-navigation/native';
+import { useNavigation, useNavigationState } from '@react-navigation/native';
 
 import {
     authorizeDeviceThunk,
@@ -27,6 +27,7 @@ import {
     RootStackParamList,
     RootStackRoutes,
     StackToStackCompositeNavigationProps,
+    useNavigationRouteMatch,
 } from '@suite-native/navigation';
 import { selectIsOnboardingFinished } from '@suite-native/settings';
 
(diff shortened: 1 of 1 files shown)


suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -81,7 +69,14 @@ export const useHandleDeviceConnection = () => {
     const navigation = useNavigation<NavigationProp>();
     const dispatch = useDispatch();
 
-    const lastRoute = navigation.getState()?.routes.at(-1)?.name;
+    // We encourage user to disconnect device when he is redirected to suspicious device screen.
+    // We should not redirect him away so he can read the screen content and decide what to do.
+    // If the device is connected again, he still should stay on that screen.
+    const isSuspiciousDeviceScreenFocused = useNavigationRouteMatch(
+        OnboardingStackRoutes.SuspiciousDevice,
+    );
```
Lemonexe, line 77: In this case we want to match the most specific route.
For that purpose we can use `useNavigationRouteMatch`, introduced in #16959.

ofc `useRoute` would be best, [but that's not possible here](https://github.com/trezor/trezor-suite/blob/461f10fe053109782c05d215bed9a8780b212b71/suite-native/navigation/src/hooks/useNavigationRoute.ts#L20-L21)

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -81,7 +69,14 @@ export const useHandleDeviceConnection = () => {
...
+
+    const lastRoute = useNavigationState(state => state?.routes.at(-1)?.name);
```
Lemonexe, line 79: Actually, the current implementation [may be buggy](https://reactnavigation.org/docs/use-navigation-state/#how-is-usenavigationstate-different-from-navigationgetstate)! TL;DR `useNavigateState` ensures that we'll rerender when the state changes (or more precisely - when the selector return value changes, similarly to redux `useSelector`) 

:thought_balloon: these cases are not matching the most specific route, but rather the first stack under RootStackNavigator.  `useNavigationRouteMatch` only matches the lowest (most specific) route. I guess I could refactor it to gradually match at any hierarchy level, but for such simple case I think that `useNavigationState` directly is fine :slightly_smiling_face:

PeKne (approved):
Nice improvement, works as expected 👌

PR #16971, opened 2025-02-13
Tenv link default model

mroz22:
based on #16969

till now, all models had the same firmware version which is not true anymore since this change https://github.com/trezor/trezor-user-env/commit/6339d48f1df4f007fe5caee659e4bca3997d480a#diff-4e726d3141a1307472f5c4ee56a6adde4dbd0664257e4898847074c18faa25dfR60. which results into 
```
websocket_error_message: RuntimeError - Unknown firmware T3T1 2.8.8-arm
```
so we need to extend our client-side logic that selects a proper default. It would feel better to do this server side - could trezor-user-env start emulator command always pick as default the highest available version which is not main version? cc @grdddj

### DIFF
--- a/docker/docker-compose.connect-popup-ci.yml
+++ b/docker/docker-compose.connect-popup-ci.yml
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   trezor-user-env-unix:
-    image: ghcr.io/trezor/trezor-user-env:6339d48f1df4f007fe5caee659e4bca3997d480a
+    image: ghcr.io/trezor/trezor-user-env:0eff3cb37dc607dc83ce4802d1c39727ec298993
     environment:
       - SDL_VIDEODRIVER=dummy
       - XDG_RUNTIME_DIR=/var/tmp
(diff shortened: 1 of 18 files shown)


grdddj:
> could trezor-user-env start emulator command always pick as default the highest available version which is not main version? cc @grdddj

I think this is already working with `-latest` suffix - https://github.com/trezor/trezor-user-env/blob/6339d48f1df4f007fe5caee659e4bca3997d480a/src/controller.py#L178-L179

Vere-Grey (approved):
good catch
my bad, I have accidentally approved the PR, i thought it is just PR for test removal

mroz22:
> i thought it is just PR for test removal

unfortunately no, it might take some time, feel free to cherrypick [e4432ec](https://github.com/trezor/trezor-suite/pull/16971/commits/e4432ec861c094611a2b889cc34012b72cf97b53) if  you want

mroz22:
@Vere-Grey lets please solve [6127f08](https://github.com/trezor/trezor-suite/pull/16971/commits/6127f0820dcf1e7078187b6c0813e8e7ed2ae0d3) in a standalone PR.

Vere-Grey (approved):
LGTM

PR #16951, opened 2025-02-12
feat(suite-desktop): turn on ASAR integrity check

Lemonexe:
## Description

Turn on ASAR integrity check on Windows & macOS.

:information_source:  Duplicate of [private PR](https://github.com/trezor/trezor-suite-private/pull/118).
See the private PR for extended description.

## Related Issue

See the private PR

PR #16972, opened 2025-02-13
Revert "feat(suite-desktop): turn on ASAR integrity check"

Lemonexe:
Reverts trezor/trezor-suite#16951

It crashes MacOS build, after this revert it works again; need to investigate further :disappointed: 

@coderabbitai ignore

### DIFF
--- a/packages/suite-desktop/electron-builder-config.js
+++ b/packages/suite-desktop/electron-builder-config.js
@@ -162,10 +162,5 @@ module.exports = {
         category: 'Utility',
         target: ['AppImage'],
     },
-    electronFuses: {
-        // the fuse may be enabled for all build targets, but as of Jan 2025, it actually affects only macOS & Windows
-        enableEmbeddedAsarIntegrityValidation: true,
-        onlyLoadAppFromAsar: true,
-    },
     afterSign: '../suite-desktop-core/scripts/notarize.ts',
 };


peter-sanderson (approved):

Lemonexe:
:memo: research notes for future:

The problem is with electron, and **not** with electron-builder.

**Proof:** on macOS, it crashes just the same when you build Suite Desktop from `develop`, and on the built binary, turn on the fuses manually:
```
npx @electron/fuses write --app "./packages/suite-desktop/build-electron/mac-arm64/Trezor Suite.app/Contents/MacOS/Trezor Suite" EnableEmbeddedAsarIntegrityValidation=on OnlyLoadAppFromAsar=on
```

Alternative hypothesis (rejected) was that electron-builder messes up the binary when setting fuses during build.

**Next steps?** 
- turn on ASAR integrity only for Windows
  - doing it via config doesn't suffice (turns on for all supported platforms); have to do it ourselves via [after pack hook](https://github.com/trezor/trezor-suite-private/tree/feat/ASAR-verification-afterPack-hook) 
- create minimal reproduction for the bug
  - if bug reproducible, make issue in Electron, else dig deeper why only in Trezor Suite 

FYI useful for debug:
```
npx @electron/fuses read --app "./packages/suite-desktop/build-electron/mac-arm64/Trezor Suite.app/Contents/MacOS/Trezor Suite"
```

PR #16980, opened 2025-02-13
Feat/card-updates

seibei-iguchi:


jvaclavik:
Adding screenshot 

image

packages/components/src/components/Card/Card.tsx
```diff
@@ -34,170 +37,154 @@ export const allowedCardFrameProps = [
...
                 $fillType={fillType}
                 $isClickable={Boolean(onClick)}
                 $variant={variant}
-                $hasLabel={Boolean(label)}
                 onClick={onClick}
                 onMouseEnter={onMouseEnter}
                 className={className}
                 onMouseLeave={onMouseLeave}
-                {...withAccessibilityProps({ tabIndex })}
-                {...rest}
```
jvaclavik, line 146: This is a good thing to remove, but not 100% sure if it's safe to do it.

packages/components/src/components/Card/utils.tsx
```diff
@@ -27,26 +23,29 @@ type VariantMapArgs = {
...
-        small: `${spacingsPx.xxs} ${spacingsPx.sm}`,
-        normal: `${spacingsPx.xs} ${spacingsPx.lg}`,
-        large: `${spacingsPx.sm} ${spacingsPx.xl}`,
+export const mapPaddingTypeToPadding = ({
+    paddingType,
+    hasHeading,
+}: PaddingMapArgs): Padding | undefined => {
+    const paddingMap: Record<PaddingType, Padding | undefined> = {
+        none: undefined,
+        small: hasHeading ? { vertical: spacings.sm, horizontal: spacings.md } : spacings.sm,
```
jvaclavik, line 32: Why not to keep same horizontal padding for both cases like in other padding variants? Seems to me a like a small inconsistence.

packages/components/src/components/buttons/IconButton/IconButton.tsx
```diff
@@ -71,20 +62,20 @@ export const IconButton = ({
             delayHide={TOOLTIP_DELAY_NONE}
             cursor="default"
         >
-            <IconButtonContainer
+            <ButtonContainer
```
jvaclavik, line 65: ❤️

jvaclavik (approved):
Nice refactorings, all changes seems to me good 👍 

Next time, screenshots for visual things would help 😁

PR #16984, opened 2025-02-13
feat: add bluetooth debug flag

peter-sanderson:


packages/suite/src/views/settings/SettingsDebug/SettingsDebug.tsx
```diff
@@ -90,8 +91,8 @@ export const SettingsDebug = () => {
                 <WalletConnect />
             </SettingsSection>
             {isDesktop() && (
-                <SettingsSection title="Bluetooth">
-                    <BluetoothEraseBonds />
```
peter-sanderson: Moved to device settings as we have a debug flag now

szymonlesisz:
cherrypicked to base branch :+1:

PR #17002, opened 2025-02-14
feat: Add bluetooth debug-settings

peter-sanderson:
image
image

### DIFF
--- a/packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
+++ b/packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
@@ -1,3 +1,5 @@
+import { Button } from '@trezor/components';
+
 import { Translation, TroubleshootingTips, WebUsbButton } from 'src/components/suite';
 import {
     TROUBLESHOOTING_TIP_BRIDGE_STATUS,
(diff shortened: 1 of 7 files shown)


packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
```diff
@@ -28,11 +37,34 @@ export const DeviceConnect = ({ isWebUsbTransport }: DeviceConnectProps) => {
               TROUBLESHOOTING_TIP_DIFFERENT_COMPUTER,
           ];
 
+    const CallToActionButton = () => {
```
Lemonexe: I looked closely why do you define the `CallToActionButton` inside `DeviceConnect` scope, and it only saves you one prop (`onBluetoothClick`) and one duplicate `useSelector` call.
Please consider if it's worth it..

:thought_balloon:  Performance-wise there's probably little difference if the `CallToActionButton` is _not the same_ component on every rerender. Parent rerenders -> a non-memoized child fully rerenders as well, and the final virtual DOM has the same final result, so no difference there as well. But React may optimize the rendering process when it detects it's the same component instance ("internal reconciliation"), according to chatGPT (well idk I don't understand that :sweat_smile: ).

That being said, performance is not a question, definitely not here. I'm opening it mostly for consistency - I'd stick to consistent best practice patterns, when not necessary to do otherwise :slightly_smiling_face:

packages/suite/src/components/suite/PrerequisitesGuide/PrerequisitesGuide.tsx
```diff
@@ -91,38 +94,49 @@ export const PrerequisitesGuide = ({ allowSwitchDevice }: PrerequisitesGuideProp
...
-            />
-
-            {allowSwitchDevice && connectedDevicesCount > 1 && (
-                <ButtonWrapper>
-                    <Button variant="tertiary" onClick={handleSwitchDeviceClick}>
-                        <Translation id="TR_SWITCH_DEVICE" />
-                    </Button>
-                </ButtonWrapper>
-            )}
+            {isBluetoothConnectOpen ? (
```
Lemonexe, line 150: the ternary makes it messy to read. If it's not a big issue, could you extract subcomponents for if & else?

Lemonexe (approved):
Looks good, but please consider the two comments for cleanup :broom:

packages/suite/src/views/settings/SettingsDebug/Bluetooth.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <SectionItem>
+            <TextColumn title="Bluetooth enabled" />
+            <ActionColumn>
+                <Checkbox
+                    isDisabled={isLoading}
+                    isChecked={isBluetoothEnabled}
+                    onClick={handleOnClick}
+                />
+            </ActionColumn>
+        </SectionItem>
```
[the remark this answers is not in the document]
Lemonexe: Not pertinent at this time, that's for future..

szymonlesisz (commented):
LGTM, compatible and works with the bluetooth branch.
im not reviewing the components structure leaving this for others to decide how they should be done

Lemonexe (approved):
TYVM :100:

PR #17007, opened 2025-02-14
ci(scripts): moving add label to differnt GH request

karliatto:
## Description

It looks like there was an issue when adding the label that was giving error when PR was created.

Now the script creates the PR and then it adds the label. So if the label adding fails the PR will already be created.

I also added throw error so if something fail the action will be seen as red so we will realized sooner that something is not working ok.

### DIFF
--- a/scripts/ci/check-connect-data.ts
+++ b/scripts/ci/check-connect-data.ts
@@ -180,7 +180,7 @@ const updateConfigFromJSON = async () => {
 
             const commitMessage = 'chore(connect): update device authenticity config';
             await exec('git', ['checkout', '-b', branchName]);
-            commit({
+            await commit({
                 path: ROOT,
                 message: commitMessage,
             });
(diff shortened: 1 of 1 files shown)


scripts/ci/check-connect-data.ts
```diff
@@ -180,15 +180,15 @@ const updateConfigFromJSON = async () => {
 
             const commitMessage = 'chore(connect): update device authenticity config';
             await exec('git', ['checkout', '-b', branchName]);
-            commit({
+            await commit({
```
karliatto, line 183: There was missing await here so, sometimes it was pushing before commiting was completed failing.

scripts/ci/check-connect-data.ts
```diff
@@ -201,12 +201,30 @@ const updateConfigFromJSON = async () => {
...
+            console.log(`Created PR: ${prUrl}`);
+
+            // Adding label to the created PR.
+            await exec('gh', ['pr', 'edit', `${prNumber}`, '--add-label', 'no-project']);
+
+            console.log(`Added label to PR #${prNumber}`);
         }
     } catch (error) {
         console.error(`Error updating configuration: ${error.message}`);
+        throw new Error(error.message);
```
karliatto, line 227: Adding `throw` here gives us feedback that something went wrong. Before it was green even when failing.

mroz22 (approved):

PR #17008, opened 2025-02-14
Bluetooth common reducer

peter-sanderson:
This PR introduces first version of the shared reducers and state (for desktop & mobile) for connecting a Bluetooth device.

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -129,6 +129,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
     public readonly transportPath;
     private readonly transportSessionOwner;
     private readonly transportDescriptorType;
+    private readonly bluetoothProps;
     private session;
     private lastAcquiredHere;
 
@@ -220,6 +221,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
         this.transportPath = descriptor.path;
         this.transportSessionOwner = descriptor.sessionOwner;
         this.transportDescriptorType = descriptor.type;
+        this.bluetoothProps = descriptor.uuid ? { uuid: descriptor.uuid } : undefined;
 
         this.session = descriptor.session;
         this.lastAcquiredHere = false;
@@ -1217,6 +1219,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
                 label: 'Unacquired device',
                 name: this.name,
                 transportSessionOwner: this.transportSessionOwner,
+                bluetoothProps: this.bluetoothProps,
             };
         }
         const defaultLabel = 'My Trezor';
@@ -1242,6 +1245,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
             unavailableCapabilities: this.unavailableCapabilities,
             availableTranslations: this.availableTranslations,
             authenticityChecks: this.authenticityChecks,
+            bluetoothProps: this.bluetoothProps,
         };
     }
 }
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -221,7 +221,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
         this.transportPath = descriptor.path;
         this.transportSessionOwner = descriptor.sessionOwner;
         this.transportDescriptorType = descriptor.type;
-        this.bluetoothProps = descriptor.uuid ? { uuid: descriptor.uuid } : undefined;
+        this.bluetoothProps = descriptor.id ? { id: descriptor.id } : undefined;
 
         this.session = descriptor.session;
         this.lastAcquiredHere = false;
--- a/suite-common/bluetooth/package.json
+++ b/suite-common/bluetooth/package.json
@@ -1,0 +1,17 @@
+{
+    "name": "@suite-common/bluetooth",
+    "version": "1.0.0",
+    "private": true,
+    "license": "See LICENSE.md in repo root",
+    "sideEffects": false,
+    "main": "src/index",
+    "scripts": {
+        "depcheck": "yarn g:depcheck",
+        "type-check": "yarn g:tsc --build"
+    },
+    "dependencies": {
+        "@reduxjs/toolkit": "1.9.5",
+        "@suite-common/redux-utils": "workspace:*",
+        "@suite-common/wallet-core": "workspace:*"
+    }
+}
--- a/suite-common/bluetooth/src/bluetoothActions.ts
+++ b/suite-common/bluetooth/src/bluetoothActions.ts
@@ -1,0 +1,44 @@
+import { createAction } from '@reduxjs/toolkit';
+
+import {
+    BluetoothDevice,
+    BluetoothScanStatus,
+    DeviceBluetoothStatus,
+} from './bluetoothReducerCreator';
+
+export const BLUETOOTH_PREFIX = '@suite/bluetooth';
+
+export const bluetoothAdapterEventAction = createAction(
+    `${BLUETOOTH_PREFIX}/adapter-event`,
+    ({ isPowered }: { isPowered: boolean }) => ({ payload: { isPowered } }),
+);
+
+type BluetoothNearbyDevicesUpdateActionPayload = {
+    nearbyDevices: BluetoothDevice[];
+};
+
+export const bluetoothNearbyDevicesUpdateAction = createAction(
+    `${BLUETOOTH_PREFIX}/nearby-devices-update`,
+    ({ nearbyDevices }: BluetoothNearbyDevicesUpdateActionPayload) => ({
+        payload: { nearbyDevices },
+    }),
+);
+
+export const bluetoothConnectDeviceEventAction = createAction(
+    `${BLUETOOTH_PREFIX}/connect-device-event`,
+    ({ connectionStatus, id }: { id: string; connectionStatus: DeviceBluetoothStatus }) => ({
+        payload: { id, connectionStatus },
+    }),
+);
+
+export const bluetoothScanStatusAction = createAction(
+    `${BLUETOOTH_PREFIX}/scan-status`,
+    ({ status }: { status: BluetoothScanStatus }) => ({ payload: { status } }),
+);
+
+export const allBluetoothActions = {
+    bluetoothAdapterEventAction,
+    bluetoothNearbyDevicesUpdateAction,
+    bluetoothConnectDeviceEventAction,
+    bluetoothScanStatusAction,
+};
--- a/suite-common/bluetooth/src/bluetoothReducerCreator.ts
+++ b/suite-common/bluetooth/src/bluetoothReducerCreator.ts
@@ -1,0 +1,127 @@
+import { AnyAction, Draft } from '@reduxjs/toolkit';
+
+import { createReducerWithExtraDeps } from '@suite-common/redux-utils';
+import { deviceActions } from '@suite-common/wallet-core';
+
+import {
+    bluetoothAdapterEventAction,
+    bluetoothConnectDeviceEventAction,
+    bluetoothNearbyDevicesUpdateAction,
+    bluetoothScanStatusAction,
+} from './bluetoothActions';
+
+export type BluetoothScanStatus = 'idle' | 'running' | 'error';
+
+export type DeviceBluetoothStatus =
+    | { type: 'pairing'; pin?: string }
+    | { type: 'paired' }
+    | { type: 'connecting' }
+    | { type: 'connected' }
+    | {
+          type: 'error';
+          error: string;
+      };
+
+export type BluetoothDevice = {
+    id: string;
+    name: string;
+    data: number[]; // Todo: consider typed data-structure for this
+    lastUpdatedTimestamp: number;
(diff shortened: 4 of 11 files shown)


suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,132 @@
...
+        knownDevices: [] as T[],
+        devices: [] as T[],
+    };
+
+    return createReducerWithExtraDeps<BluetoothState<T>>(initialState, (builder, extra) =>
+        builder
+            .addCase(bluetoothAdapterEventAction, (state, { payload: { isPowered } }) => {
+                state.isAdapterEnabled = isPowered;
+                if (!isPowered) {
+                    state.devices = [];
```
szymonlesisz: stat.scanStatus = 'done' ?

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,132 @@
...
+                deviceActions.connectDevice,
+                (
+                    state,
+                    {
+                        payload: {
+                            device: { bluetoothProps },
+                        },
+                    },
+                ) => {
+                    if (bluetoothProps && bluetoothProps.uuid in state.devices) {
```
szymonlesisz: lets make sure that suite-native also provides this value, also change `uuid` => `id`

peter-sanderson: https://github.com/trezor/trezor-suite/pull/17008/commits/7b8e6b2d33265dcb0a8c587a8e3fb329fb490f2d

This will be provided by connect right? I suppose this will be same on mobile and desktop as it uses same connect, right?

suite-common/bluetooth/src/bluetoothReducerCreator.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/17008/commits/7b8e6b2d33265dcb0a8c587a8e3fb329fb490f2d

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,133 @@
...
+                                state.knownDevices.push(deviceState);
+                            }
+                        }
+                    }
+                },
+            )
+            .addMatcher(
+                action => action.type === extra.actionTypes.storageLoad,
+                (state, action: AnyAction) => {
+                    state.knownDevices = action.payload.knownDevices?.bluetooth;
```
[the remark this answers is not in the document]
peter-sanderson: https://github.com/trezor/trezor-suite/pull/17008/commits/7b8e6b2d33265dcb0a8c587a8e3fb329fb490f2d

suite-common/bluetooth/src/bluetoothReducerCreator.ts
szymonlesisz: but connect needs to know it from the transport, and mobile uses different transport

szymonlesisz:
do not merge until tested in `rust-bluetooth` branch

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,133 @@
...
+import { deviceActions } from '@suite-common/wallet-core';
+
+import {
+    bluetoothAdapterEventAction,
+    bluetoothConnectDeviceEventAction,
+    bluetoothDeviceListUpdate,
+    bluetoothScanStatusAction,
+} from './bluetoothActions';
+
+export type BluetoothScanStatus = 'running' | 'done' | 'error';
```
yanascz: I find `done` to confusing. It seems as a final status from which there's no way back. What about `idle` instead?
```suggestion
export type BluetoothScanStatus = 'idle' | 'running' | 'error';
```

peter-sanderson: fixed

suite-common/bluetooth/src/bluetoothActions.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const BLUETOOTH_PREFIX = '@suite/bluetooth';
+
+export const bluetoothAdapterEventAction = createAction(
+    `${BLUETOOTH_PREFIX}/adapter-event`,
+    ({ isPowered }: { isPowered: boolean }) => ({ payload: { isPowered } }),
+);
+
+type BluetoothDeviceListUpdatePayload = {
+    devices: BluetoothDevice[];
```
yanascz: I'd prefer `nearbyDevices` here.

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,133 @@
...
+type BluetoothState<T extends BluetoothDevice> = {
+    isAdapterEnabled: boolean;
+    scanStatus: BluetoothScanStatus;
+
+    // This will be persisted, those are devices we believed that are paired
+    // (because we already successfully paired them in the Suite) in the Operating System
+    knownDevices: T[];
+
+    // This list of devices that is union of saved-devices and device that we get from scan
+    devices: T[];
```
yanascz: Why a union? That will make updating more complicated AFAIK. I'd go with `nearbyDevices` as the result of scanning.

packages/connect/src/device/Device.ts
```diff
@@ -220,6 +221,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
         this.transportPath = descriptor.path;
         this.transportSessionOwner = descriptor.sessionOwner;
         this.transportDescriptorType = descriptor.type;
+        this.bluetoothProps = descriptor.id ? { id: descriptor.id } : undefined;
```
yanascz, line 224: Wouldn't `descriptor?.id` do the trick?

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export type BluetoothDevice = {
+    id: string;
+    name: string;
+    data: number[];
+    lastUpdate: number;
+    status: DeviceBluetoothStatus | null;
+};
+
+export type DeviceBluetoothStatus =
```
yanascz, line 15: Don't we require a type to be declared before it's used? 🤔

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,133 @@
...
+    knownDevices: T[];
+
+    // This list of devices that is union of saved-devices and device that we get from scan
+    devices: T[];
+};
+
+export const bluetoothReducerCreator = <T extends BluetoothDevice>() => {
+    const initialState: BluetoothState<T> = {
+        isAdapterEnabled: true, // To prevent the UI from flickering when the page is loaded
+        scanStatus: 'running', // To prevent the UI from flickering when the page is loaded
```
yanascz: Is this relevant when in a common code? I'd expect `false` and `idle` and some code that transitions into scanning mode.

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,133 @@
...
+        devices: [] as T[],
+    };
+
+    return createReducerWithExtraDeps<BluetoothState<T>>(initialState, (builder, extra) =>
+        builder
+            .addCase(bluetoothAdapterEventAction, (state, { payload: { isPowered } }) => {
+                state.isAdapterEnabled = isPowered;
+                if (!isPowered) {
+                    state.devices = [];
+                    state.scanStatus = 'done';
```
yanascz: This seems to be a typo.
```suggestion
                    state.scanStatus = 'idle';
```

suite-common/bluetooth/src/bluetoothReducerCreator.ts
peter-sanderson: We dont have any rule for it AFAIK, but fixed, Its better

packages/connect/src/device/Device.ts
peter-sanderson: That would produce a different type, in the `bluetoothProps` object I think the `id` is required.

suite-common/bluetooth/src/bluetoothReducerCreator.ts
yanascz: Discussed in person, types adapted.

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,127 @@
...
+                deviceActions.connectDevice,
+                (
+                    state,
+                    {
+                        payload: {
+                            device: { bluetoothProps },
+                        },
+                    },
+                ) => {
+                    if (bluetoothProps && bluetoothProps.id in state.nearbyDevices) {
```
yanascz, line 100: This condition seems to always be `false` since `nearbyDevices` is no longer an object but an array.

yanascz, line 100: This condition seems to always evaluate to `false` since `nearbyDevices` is no longer an object but an array.

```suggestion
                    if (!bluetoothProps || !bluetoothProps.id) {
                        return;
                    }
```

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,127 @@
...
+                        const deviceState = state.nearbyDevices.find(
+                            it => it.id === bluetoothProps.id,
+                        );
+
+                        if (deviceState !== undefined) {
+                            deviceState.status = null;
+
+                            // Once device is fully connected, we save it to the list of paired devices
+                            // so next time user opens suite
+                            const foundPairedDevice = state.knownDevices.find(
```
yanascz, line 110:
```suggestion
                            const foundKnownDevice = state.knownDevices.find(
```

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,127 @@
...
+                ) => {
+                    if (bluetoothProps && bluetoothProps.id in state.nearbyDevices) {
+                        const deviceState = state.nearbyDevices.find(
+                            it => it.id === bluetoothProps.id,
+                        );
+
+                        if (deviceState !== undefined) {
+                            deviceState.status = null;
+
+                            // Once device is fully connected, we save it to the list of paired devices
```
yanascz, line 108: paired 👉 known

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,127 @@
...
+                        },
+                    },
+                ) => {
+                    if (bluetoothProps && bluetoothProps.id in state.nearbyDevices) {
+                        const deviceState = state.nearbyDevices.find(
+                            it => it.id === bluetoothProps.id,
+                        );
+
+                        if (deviceState !== undefined) {
+                            deviceState.status = null;
```
yanascz, line 106: Why is `status` set to `null` here? 🤔

suite-common/bluetooth/src/bluetoothReducerCreator.ts
```diff
@@ -0,0 +1,127 @@
...
+          type: 'error';
+          error: string;
+      };
+
+export type BluetoothDevice = {
+    id: string;
+    name: string;
+    data: number[]; // Todo: consider typed data-structure for this
+    lastUpdatedTimestamp: number;
+    status: DeviceBluetoothStatus | null;
```
szymonlesisz, line 30: status should not be persistent, its related only to connection/pairing phase

suite-common/bluetooth/package.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "@suite-common/bluetooth",
+    "version": "1.0.0",
+    "private": true,
+    "license": "See LICENSE.md in repo root",
+    "sideEffects": false,
+    "main": "src/index",
```
szymonlesisz, line 7: missing src/index file

szymonlesisz (changes requested):
i've added and initialized this reducer in suite (in rust-bluetooth branch) and the AppState  types  shows `bluetooth: never`

```
import { BluetoothDevice } from '@trezor/transport-bluetooth';
import { bluetoothReducerCreator } from '@suite-common/bluetooth';

const rootReducer = combineReducers({
   ....,
   bluetooth: bluetoothReducerCreator<BluetoothDevice>(),
});
```

peter-sanderson:
Closing in favor of: https://github.com/trezor/trezor-suite/pull/17090

suite-common/bluetooth/src/bluetoothReducerCreator.ts
peter-sanderson: :+1:

suite-common/bluetooth/src/bluetoothReducerCreator.ts
peter-sanderson: fixed

PR #17010, opened 2025-02-14
Repo: move connect lib to es2022

karliatto:
## Description

Changing base `tsconfig.lib.json` to target ES2022. Since in connect we have added a new library that is only transpiled to ES2022 using some features that are only ES2022 therefore making connect also ES2022.


mroz22 (approved):
it sounds good to me. potentially breaking in terms of ts configs of the dependent projects. whatever code we serve directly from trezor.io should be fine since it is also babeled. could you please @karliatto instruct QA (@evgenysl) to double check compatibility with slightly older browsers once we are in beta?

PR #17015, opened 2025-02-14
feat(scripts): add timestamp update to check-connect-data

karliatto:
## Description
Add timestamp update to check-connect-data

### DIFF
--- a/scripts/ci/check-connect-data.ts
+++ b/scripts/ci/check-connect-data.ts
@@ -35,6 +35,16 @@ const fetchJSON = async (url: string) => {
     return response.json();
 };
 
+const getLatestTimestamp = (timestamps: string[]): string | null => {
+    if (timestamps.length === 0) {
+        return null;
+    }
+
+    const latestTime = Math.max(...timestamps.map(Date.parse));
+
+    return new Date(latestTime).toISOString();
+};
+
 const updateConfigFromJSON = async () => {
     try {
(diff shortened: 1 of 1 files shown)


scripts/ci/check-connect-data.ts
```diff
@@ -35,17 +35,31 @@ const fetchJSON = async (url: string) => {
...
 
+const getLatestTimestamp = (timestamps: string[]): string | null => {
+    if (timestamps.length === 0) {
+        return null;
+    }
+
+    const dateObjects = timestamps.map(timestamp => new Date(timestamp));
+    const latestDate = new Date(Math.max(...dateObjects.map(date => date.getTime())));
+
+    return latestDate.toISOString();
```
matejkriz: nitpick (no need for mapping the timestamps twice...):
```suggestion
    const latestTime = Math.max(...timestamps.map(Date.parse));
    
    return new Date(latestTime).toISOString();
};
```

matejkriz (approved):
Works fine, thank you! Tested here https://github.com/trezor/trezor-suite/pull/17016

~~Would be nice to apply it right away so the timestamp is already updated.~~ we will see if it will work correctly during daily scheduled run.

scripts/ci/check-connect-data.ts
karliatto: Sure, updated!

PR #17046, opened 2025-02-17
Run less tests

HajekOndrej:
## Related Issue

Resolve

### DIFF
--- a/.github/actions/check-previous-test-runs/action.yml
+++ b/.github/actions/check-previous-test-runs/action.yml
@@ -1,0 +1,68 @@
+name: 'Check Previous Runs'
+description: 'Check if a previous run of a job was successful to skip tests'
+inputs:
+  github_token:
+    description: 'GitHub token for accessing the API'
+    required: true
+  pr_branch:
+    description: 'The branch of the pull request'
+    required: true
+  repo:
+    description: 'The repository name'
+    required: true
+  workflow_name:
+    description: 'The workflow name to filter runs'
+    required: true
+  job_name:
+    description: 'The name of the job to check success'
+    required: true
+outputs:
+  skip_tests:
+    description: 'Indicates whether tests should be skipped'
+    value: ${{ steps.check-previous-runs.outputs.skip_tests }}
+runs:
+  using: 'composite'
+  steps:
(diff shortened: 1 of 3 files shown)


.github/actions/check-previous-test-runs/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        JOB_NAME="${{ inputs.job_name }}"
+
+        # Fetch workflow ID
+        WORKFLOW_ID=$(gh workflow list \
+          --repo "$REPO" \
+          --json id,name \
+          --jq ".[] | select(.name == \"${WORKFLOW_NAME}\") | .id")
+
+        # Fetch previous PR runs
+        PREVIOUS_RUN_JOBS_URL=$(curl -s -H "Authorization: Bearer ${GH_TOKEN}" \
```
Vere-Grey, line 43: if there are runs, this should return only the last one right? So it does not fetch runs, but a single last run, right?

Vere-Grey: Explained in DM

.github/workflows/test-suite-desktop-e2e.yml
```diff
@@ -50,25 +50,41 @@ jobs:
...
+          repo: ${{ github.repository }}
+          workflow_name: ${{ github.workflow }}
+          job_name: "${{ github.job }} (${{ join(matrix.*, ', ') }})"
+
+      # Electron requires unprivileged user namespaces to function properly. 
+      # Disabling this security rule allows Electron to create sandboxed processes
+      # without requiring elevated privileges, which is essential for running the application.
+      # This is workaround until electron builder solves this issue in future release.
       - name: Disable security rule 'Restricted unprivileged user namespaces'
+        if: steps.check-previous-runs.outputs.skip_tests != 'true'
```
Vere-Grey, line 70: minor, switching to `if: steps.check-previous-runs.outputs.run_tests` would improve readability

Vere-Grey: ah, nope, that would not work with retries and etc. Ignore it

Vere-Grey (approved):
:+1:

PR #17048, opened 2025-02-17
feat(suite-native): global send button

### LINKED ISSUE #16484: Send from dashboard
As a fast-follow for send, we need to design a way to handle "Send" & "Receive" as global action.
This issue is just for Send, receive was done in https://github.com/trezor/trezor-suite/issues/16204 

AC:
- User is able to trigger send  from homepage
- User is able to choose acount for sending
- User sees only accounts with positive balances for Send

Designs: https://www.figma.com/design/xSWRW8Csq7zYJ3TiqwoIxB/Small-Things-%5BMOB%5D?node-id=2067-24263&t=kXK56b5XE1bxT8tp-0

Followup issue for analytics: https://github.com/trezor/trezor-suite/issues/17070

PeKne:
## Description
- send from dashboard allowed

## Related Issue

Resolve #16484

## Screenshots:

https://github.com/user-attachments/assets/d6012b11-ec49-412d-ba47-9b1d9986f2da

### DIFF
--- a/suite-native/accounts/src/utils.ts
+++ b/suite-native/accounts/src/utils.ts
@@ -8,7 +8,7 @@ import {
 } from '@suite-common/wallet-config';
 import { formattedAccountTypeMap } from '@suite-common/wallet-core';
 import { Account } from '@suite-common/wallet-types';
-import { orderedAccountTypes } from '@suite-native/config';
+import { orderedAccountTypes, sendDisabledNetworkTypes } from '@suite-native/config';
 
 const accountTypeToSectionHeader: Readonly<Partial<Record<AccountType, string>>> = {
     normal: 'default',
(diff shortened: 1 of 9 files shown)


suite-native/accounts/src/utils.ts
```diff
@@ -67,6 +67,14 @@ export const filterAccountsByLabelAndNetworkNames = (
     return A.filter(accounts, account => isFilterValueMatchingAccount(account, filterValue));
 };
 
+export const filterSendAvailableAccounts = (accounts: readonly Account[]) =>
```
vytick, line 70: nit:
```suggestion
export const filterSendEnabledAccounts = (accounts: readonly Account[]) =>
```

suite-native/app/src/navigation/RootStackNavigator.tsx
```diff
@@ -107,6 +106,7 @@ export const RootStackNavigator = () => {
                     name={RootStackRoutes.ReceiveStack}
                     component={ReceiveStackNavigator}
                 />
+                <RootStack.Screen name={RootStackRoutes.SendStack} component={SendStackNavigator} />
```
matejkriz, line 109: Why was it necessary to change the order here?

matejkriz (approved):
It works like a charm! 🎉

suite-native/app/src/navigation/RootStackNavigator.tsx
PeKne: I moved the SendStack to the this `RootStack.Group` so the flow starts with the `slide_from_bottom` animation same as the other flows within this group.

PR #17050, opened 2025-02-17
feat(suite-desktop): turn on ASAR integrity check via afterPack hook

Lemonexe:
## Description

After the original implementation #16951 was reverted in #16972 because of failing macOS builds, let's just turn on the ASAR integrity for Windows, for the time being.

The previous attempt was very simple  - just several lines of config for electron-builder, which takes care of flipping the electron fuses during build for **all** platforms (but it takes effect only on macOS & Windows). 

If we want to turn on the fuses only for selected platform(s), we need to do it manually via an `afterPack` hook. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite-private/issues/16

## QA instructions

- build Suite Desktop for all platforms
- see that it installs & runs
- I have personally tested Linux Ubuntu 24 & Windows 10 :heavy_check_mark: 

@coderabbitai ignore

### DIFF
--- a/packages/suite-desktop-core/package.json
+++ b/packages/suite-desktop-core/package.json
@@ -47,6 +47,7 @@
     },
     "devDependencies": {
         "@currents/playwright": "^1.3.1",
+        "@electron/fuses": "^1.8.0",
         "@electron/notarize": "2.5.0",
         "@playwright/browser-chromium": "^1.49.1",
         "@playwright/browser-firefox": "^1.49.1",
--- a/packages/suite-desktop/electron-builder-config.js
+++ b/packages/suite-desktop/electron-builder-config.js
@@ -162,5 +162,6 @@ module.exports = {
         category: 'Utility',
         target: ['AppImage'],
     },
+    afterPack: '../suite-desktop-core/scripts/setElectronFuses.js',
     afterSign: '../suite-desktop-core/scripts/notarize.ts',
 };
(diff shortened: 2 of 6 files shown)


packages/suite-desktop-core/package.json
```diff
@@ -47,6 +47,7 @@
     },
     "devDependencies": {
         "@currents/playwright": "^1.3.1",
+        "@electron/fuses": "^1.8.0",
```
Lemonexe, line 50: Already a dependency of  electron-builder core a.k.a. `app-builder-lib`

packages/suite-desktop/electron-builder-config.js
```diff
@@ -162,5 +162,6 @@ module.exports = {
         category: 'Utility',
         target: ['AppImage'],
     },
+    afterPack: '../suite-desktop-core/scripts/setElectronFuses.js',
```
Lemonexe, line 165: Wait, why is the new script `.js`, but the other two are `.ts`? (`notarize` and `sign-windows`)
Because they're not really typescript! They have the extension, but they are written in CJS.
It's actually quite a mess.
That rhymes!
Anyway, in [next PR](https://github.com/trezor/trezor-suite/pull/17056) I'll turn on TS _(ofc it's not running currently)_ so you can look forward to me valiantly fighting skeletons in closet :skull:

komret (approved):
I tested that it is skipped on MacOS. Please create a follow-up issue to investigate why MacOS builds were failing.

Lemonexe:
Idk why the playwright test keeps failing, I rerun it manually from this branch and it runs :heavy_check_mark: 
In any case, those tests run on Suite Web and this PR touches only Desktop build.
https://github.com/trezor/trezor-suite/actions/runs/13385796325

PR #17055, opened 2025-02-17
16996 mobile trade make all fiat and crypto balances discreet

### LINKED ISSUE #16996: Mobile Trade: make all fiat and crypto balances discreet
Identify all crypto and fiat balances in the Trade UI and make sure `DiscreetText` is used via formatters or directly

jbazant:
## Description

I also
- updated address picker item to display derivation path (inspired by desktop)
- changed how address is displayed (and introduced `AccountAddress` component)

## Related Issue

Resolve [#16996](https://github.com/trezor/trezor-suite/issues/16996)

## Screenshots:
Screenshot 2025-02-17 at 17 58 53
Screenshot 2025-02-17 at 17 59 03

### DIFF
--- a/suite-common/formatters/src/formatters/prepareCryptoAmountFormatter.ts
+++ b/suite-common/formatters/src/formatters/prepareCryptoAmountFormatter.ts
@@ -27,6 +27,8 @@ export type CryptoAmountFormatterDataContext = {
     isEllipsisAppended?: boolean;
 };
 
+export const BASE_CRYPTO_MAX_DISPLAYED_DECIMALS = 8;
+
 const truncateDecimals = (value: string, maxDecimals: number, isEllipsisAppended: boolean) => {
     const parts = value.split('.');
     const [integerPart, fractionalPart] = parts;
@@ -63,7 +65,7 @@ const convertToUnit = (
         !isBalance &&
         (bitcoinAmountUnit !== PROTO.AmountUnit.SATOSHI || !areAmountUnitsSupported)
     ) {
-        return formatAmount(value, decimals ?? 8);
+        return formatAmount(value, decimals ?? BASE_CRYPTO_MAX_DISPLAYED_DECIMALS);
     }
 
     return value;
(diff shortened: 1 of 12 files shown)


suite-common/formatters/src/formatters/prepareCryptoAmountFormatter.ts
```diff
@@ -83,7 +84,7 @@ export const prepareCryptoAmountFormatter = (config: FormatterConfig) =>
                 symbol,
                 isBalance = false,
                 withSymbol = true,
-                maxDisplayedDecimals = 8,
+                maxDisplayedDecimals = DEFAULT_MAX_DISPLAYED_DECIMALS,
```
vytick: ❤️

suite-native/assets/src/components/AssetItem.tsx
```diff
@@ -55,8 +59,7 @@ const CryptoAmount = React.memo(({ symbol }: AssetItemSubComponentProps) => {
             value={cryptoValue}
             symbol={symbol}
             // Every asset crypto amount is rounded to 8 decimals to prevent UI overflow.
-
-            decimals={8}
+            decimals={DEFAULT_MAX_DISPLAYED_DECIMALS}
```
vytick: Q: When `DEFAULT_MAX_DISPLAYED_DECIMALS` is default, is that needed to set here (and on other places as well)? Probably not. Or is it?

jbazant: Well, decimals are not set by default. Therefore it is. Maybe I should use `BASE` instead of `DEFAULT`. WDYT?

vytick: oh, i see. I got confused by default `maxDisplayedDecimals` in `prepareCryptoAmountFormatter`

vytick: BASE would be probably better 👍

vytick (approved):

suite-native/assets/src/components/AssetItem.tsx
jbazant: Updated

vytick (approved):

PR #17056, opened 2025-02-17
chore(suite-desktop-core): typescriptize electron builder hooks

Lemonexe:
## Description

Actually use TypeScript for electron-builder hooks - currently they have `.ts` extension, but are written plain JavaScript (CJS, as electron-builder needs), and type checking never ever happens.

- rewrite CJS to TS (ESM)
- include scripts to `type-check`. We don't do it for most JS scripts, but these scripts are built around API of electron-builder & electron/notarize, and IMO they'd really benefit from being typechecked
  - and indeed, there is inconsistent TS in [notarize.ts](https://github.com/trezor/trezor-suite/pull/17056/files#diff-503ee41462b840240200c07b1c391ba370cc013e6782cfc19549ccb294ea0c4b)
    - probably when `@electron/notarize` was bumped by yours truly, but I couldn't have noticed when I believed it was TS, yet didn't know it's not really TS :sweat_smile:  
- add `build:lib` to compile them to CJS for electron-builder right before building
- fix eslint

## Related Issue

Followup to https://github.com/trezor/trezor-suite-private/issues/16 and https://github.com/trezor/trezor-suite/pull/17050

### DIFF
--- a/packages/suite-desktop-core/eslint.config.mjs
+++ b/packages/suite-desktop-core/eslint.config.mjs
@@ -20,4 +20,11 @@ export default [
     {
         ignores: ['**/playwright-report/', '**/test-results/'],
     },
+    {
+        files: ['**/scripts/**'],
+        rules: {
+            'no-console': 'off',
+            'import/no-default-export': 'off',
+        },
+    },
 ];
--- a/packages/suite-desktop-core/scripts/notarize.ts
+++ b/packages/suite-desktop-core/scripts/notarize.ts
@@ -1,8 +1,7 @@
-/* eslint-disable no-console */
+import { notarize } from '@electron/notarize';
+import type { Hooks } from 'app-builder-lib';
 
-const { notarize } = require('@electron/notarize');
-
-exports.default = context => {
+const notarizeAfterSignHook: Hooks['afterSign'] = context => {
     const { electronPlatformName, appOutDir } = context;
 
     if (electronPlatformName !== 'darwin') {
@@ -25,5 +24,8 @@ exports.default = context => {
         appleId: process.env.APPLEID,
         appleIdPassword: process.env.APPLEIDPASS,
         teamId: process.env.APPLETEAMID,
-    });
+        // TODO fix: notarize tool is known to be working, but TS fails: no overload matches this call
+    } as any);
 };
+
+export default notarizeAfterSignHook;
--- a/packages/suite-desktop-core/scripts/sign-windows.ts
+++ b/packages/suite-desktop-core/scripts/sign-windows.ts
@@ -1,19 +1,24 @@
-/* eslint-disable */
-exports.default = async function (configuration) {
+import type { CustomWindowsSign } from 'app-builder-lib';
+
+const signWindows: CustomWindowsSign = async configuration => {
     // Check if IS_CODESIGN_BUILD is set and true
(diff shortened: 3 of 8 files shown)


packages/suite-desktop-core/scripts/sign-windows.ts
```diff
@@ -1,19 +1,24 @@
...
         return;
     }
+
     // do not include passwords or other sensitive data in the file
     // rather create environment variables with sensitive data
     const CERTIFICATE_NAME = process.env.WINDOWS_SIGN_CERTIFICATE_NAME;
     const TOKEN_PASSWORD = process.env.WINDOWS_SIGN_TOKEN_PASSWORD;
 
-    require('child_process').execSync(
+    await require('child_process').exec(
```
Lemonexe, line 16: This fn was async (and has to stay, because TS requires the hook to return `Promise`), but without await, which failed eslint, hence eslint disable :see_no_evil: 

:point_right:  needs to be verified if changing `execSync` to `exec` doesn't break the signing with codesign build.
But should be ok. [This is the code that invokes it](https://github.com/electron-userland/electron-builder/blob/8e7811d18de3acb39ce9253cf2cd9afa4e23f99c/packages/app-builder-lib/src/codeSign/windowsSignToolManager.ts#L200). It awaits the Promise, so LGTM..

packages/suite-desktop-core/tsconfig.json
```diff
@@ -5,7 +5,12 @@
         "moduleResolution": "node",
         "outDir": "libDev"
     },
-    "include": ["src", "e2e", "**/*.json"],
+    "include": [
+        "src",
+        "e2e",
+        "scripts",
```
Lemonexe, line 11: This includes the scripts to `type-check`, which was not done before, but does not build them, that's what `build:lib` does (`tsconfig.lib.json`)

packages/suite-desktop-core/eslint.config.mjs
```diff
@@ -20,4 +20,11 @@ export default [
     {
         ignores: ['**/playwright-report/', '**/test-results/'],
     },
+    {
+        files: ['**/scripts/**'],
+        rules: {
+            'no-console': 'off',
+            'import/no-default-export': 'off',
```
Lemonexe, line 27: funnily enough, eslint tolerates CJS `exports.default` in a TS file, but errors at ES `export default`.

packages/suite-desktop-core/scripts/notarize.ts
```diff
@@ -1,8 +1,7 @@
-/* eslint-disable no-console */
+import { notarize } from '@electron/notarize';
+import type { Hooks } from 'app-builder-lib';
 
-const { notarize } = require('@electron/notarize');
-
-exports.default = context => {
```
Lemonexe, line 5: No type check ever happened on this "typescript" file.
TS would fail, because at the very least, TS requires `(context: any) =>`, but then it's no longer valid JS (electron-builder crashes)

peter-sanderson (approved):
Very nice, much types, such typescript :dog: 

:clap: Good job!

(haven't tested tho)

packages/suite-desktop-core/scripts/sign-windows.ts
Lemonexe: UPDATE: This was actually wrong :scream:

Issue #16291, opened 2025-01-09
Onboarding - Flow Cancellation


PeKne:
### Acceptance criteria
- **TREZOR DISCONNECTED**: As a user, if my phone losses connection with the Trezor hardware (disconnection, ejects, or defective cable), then I would like to see the **Trezor disconnected** screen
    - CTAs available
    - [Reconnect] Trezor
    - [Cancel]
    - Conditionals
    - If the user clicks on CTA [Reconnect], they will see the **Connect** screen
    - If the user clicks on CTA [Cancel], they will see the **Dashboard** screen
- **CANCEL SET UP**: As a user, if I click on the cross icon (top left) in any of the steps, I would like to see the **confirm cancelation overlay**
    - CTAs available
    - [Cancel]
    - [Continue] with set up
    - Conditionals
        - If the user clicks on CTA [Cancel], they will see the **Dashboard** screen
        - If the user clicks on CTA [Continue], they will continue the setup

STew790:
**QA NOK**
After some playing around, when I disconnect the device in onboarding, sometimes the host doesn't see that I disconnected the device at all and sometimes it does but it starts glitching weirdly (which it wasn't doing before 🤔 ).
Also, when I reconnect the device during the onboarding and I get to the FW installation part, I get fw installation error at first but after a few seconds the fw installation process starts.

https://github.com/user-attachments/assets/102cf5a1-32ea-47e2-a0cb-142fb4ff6188

**Info**
25.4.1 eadd675

PeKne:
💡  Additional Info: 

The issue of looping glitchy UI happens in the following scenario:

1. connect an uninitialized device
2. cancel the flow by clicking the cross button in the top left corner
3. return to the flow from dashboard
4. disconnect device from the phone

If you combine canceling the flow via UI and disconnecting the device physically from the app, the glitch occurs.

STew790:
**QA OK**
If I either disconnect or click on the "X" during the setup I am asked whether I really want to stop the setup and after that I can always continue with the new screen and button. Works with TT as well.
Also, the issue was addressed.

**Info**
25.5.1 38b9d58

PR #17065, opened 2025-02-18
fix(suite-native): suspicious device screen back navigation

### LINKED ISSUE #16279: Onboarding - Suspicious Device Screen

PeKne:
## Description
- fixes navigation problem found by QA: https://github.com/trezor/trezor-suite/issues/16279#issuecomment-2663420107
- simplifies the navigation of `useHandleDeviceConnection`

## Related Issue

Resolve #16279

### DIFF
--- a/suite-native/device/src/hooks/useHandleDeviceConnection.ts
+++ b/suite-native/device/src/hooks/useHandleDeviceConnection.ts
@@ -79,39 +79,30 @@ export const useHandleDeviceConnection = () => {
     const lastRoute = useNavigationState(state => state?.routes.at(-1)?.name);
     const isDeviceSettingsStackFocused = lastRoute === RootStackRoutes.DeviceSettingsStack;
     const isSendStackFocused = lastRoute === RootStackRoutes.SendStack;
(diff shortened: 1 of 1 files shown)


yanascz (approved):
Seems to works fine. 👍 Except that the onboarding flow cannot be closed at the moment but that seems to be known [issue](https://github.com/trezor/trezor-suite/issues/16291).

PR #17080, opened 2025-02-18
fix(suite-native): validate custom fee fields after transforming value

vytick:
When setting custom field, the validation was triggered before the value was transformed. This caused odd behaviour in case user inserted non numerical characters.

## Screenshots:

BEFORE:

https://github.com/user-attachments/assets/3ebf838c-5d21-4f56-bc16-805cba69d2b6

AFTER:

https://github.com/user-attachments/assets/fe945592-b14e-4858-aa74-c04c0d8179ac

### DIFF
--- a/suite-native/module-send/src/components/CustomFeeInputs.tsx
+++ b/suite-native/module-send/src/components/CustomFeeInputs.tsx
@@ -8,6 +8,7 @@ import { getFeeUnits } from '@suite-common/wallet-utils';
 import { Hint, Text, VStack } from '@suite-native/atoms';
 import { TextInputField, useFormContext } from '@suite-native/forms';
 import { Translation, useTranslate } from '@suite-native/intl';
+import { useDebounce } from '@trezor/react-utils';
 
 import { integerTransformer, useSendAmountTransformers } from '../hooks/useSendAmountTransformers';
 import { SendFeesFormValues } from '../sendFeesFormSchema';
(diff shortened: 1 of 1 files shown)


suite-native/module-send/src/components/CustomFeeInputs.tsx
```diff
@@ -20,27 +21,44 @@ export const CustomFeeInputs = ({ symbol }: CustomFeeInputsProps) => {
...
     const feePerUnitFieldName = 'customFeePerUnit';
     const hasFeePerByteError = G.isNotNullable(errors[feePerUnitFieldName]);
 
     const networkType = getNetworkType(symbol);
     const feeUnits = getFeeUnits(networkType);
     const formattedFeePerUnit = `${feeInfo?.minFee} ${feeUnits}`;
 
+    const handleFieldChangeValue =
+        (
+            fieldName: typeof customFeeLimitName | typeof feePerUnitFieldName,
```
PeKne: nit:
```suggestion
            fieldName: keyof SendFeesFormValues
```

suite-native/module-send/src/components/CustomFeeInputs.tsx
```diff
@@ -20,27 +21,44 @@ export const CustomFeeInputs = ({ symbol }: CustomFeeInputsProps) => {
...
                 <TextInputField
                     label={translate('moduleSend.fees.custom.bottomSheet.label.gasLimit')}
-                    name="customFeeLimit"
-                    testID="customFeeLimit"
+                    name={customFeeLimitName}
+                    testID={`@send/${customFeeLimitName}-input`}
                     accessibilityLabel="address input"
                     keyboardType="number-pad"
                     valueTransformer={integerTransformer}
+                    onChangeText={handleFieldChangeValue(customFeeLimitName, integerTransformer)}
```
PeKne, line 57: Is it necessary to pass the transformer as `valueTransformer` prop and also in the `handleFieldChangeValue` function? Could not it be simplified so it is applied only once?

PeKne (approved):

suite-native/module-send/src/components/CustomFeeInputs.tsx
vytick: good spot

vytick: 4316d09966

suite-native/module-send/src/components/CustomFeeInputs.tsx
vytick: 4316d09966

PeKne (approved):

Ondra-Zik:
**QA OK**

The issue has been resolved, and the behavior no longer occurs.  

Platform: Android
Build: 1250

Issue #17084, opened 2025-02-18
Fix skipped flaky unit test "ETH fee limit changes" in useSendForm.test.tsx


matejkriz:
Check this thread https://satoshilabs.slack.com/archives/G019WLX2P7B/p1739269958372369

Test skipped in https://github.com/trezor/trezor-suite/pull/17085

PR #17085, opened 2025-02-18
chore(suite): skip flaky unit test

matejkriz:
## Description

- issue https://github.com/trezor/trezor-suite/issues/17084

### DIFF
--- a/packages/suite/src/hooks/wallet/__tests__/useSendForm.test.tsx
+++ b/packages/suite/src/hooks/wallet/__tests__/useSendForm.test.tsx
@@ -357,7 +357,9 @@ describe('useSendForm hook', () => {
     });
 
     fixtures.feeChange.forEach(f => {
-        it(
+        // Add conditional test execution
+        const testFn = f.skip ? it.skip : it;
+        testFn(
             `changeFee: ${f.description}`,
             async () => {
                 testMocks.setTrezorConnectFixtures(f.connect);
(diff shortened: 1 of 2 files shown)


packages/suite/src/hooks/wallet/__tests__/useSendForm.test.tsx
```diff
@@ -357,7 +357,9 @@ describe('useSendForm hook', () => {
     });
 
     fixtures.feeChange.forEach(f => {
-        it(
```
vojtatranta, line 360: I looked at this tests like on Friday and I even tried to rewrite it with `renderHook`. But I was like: nah, I can't even tell what does it test.

So I skipped that, but I would really advise to rewrite it so that we only test the hook itself.

vojtatranta: Yeah and I wanted to express my blessing for you skipping this test.

mroz22:
also there was a fixing attempt by @martykan couple of days ago.

martykan:
Yes, but it was a blind shot and it seems it wasn't successful

mroz22:
so how about this? are we merging it?

martykan (approved):

matejkriz:
I wasn't even checking what it is supposed to test, I hope that someone will investigate that as part of https://github.com/trezor/trezor-suite/issues/16697

PR #17089, opened 2025-02-18
Connect Device stuff refactoring 

marekrjpolak:
Some improvements of (mostly) device handling in Connect.

- https://github.com/trezor/trezor-suite/pull/17089/commits/4c37f5c49125d0d984c886b50c87e36d418ad22f - on `DeviceList.enumerate` call, `handleDescriptorsChange` is fired instead of separately update descriptors on devices (it should do the same thing, and even more)
- https://github.com/trezor/trezor-suite/pull/17089/commits/1512841956a288774355a953cea50cb14e4172ab - `TransportEvents` on transports improved, so only `CONNECTED` event (device first seen) is fired 'globally', while `DISCONNECTED`/`SESSION_CHANGED`/`REQUEST_RELEASE` are fired per device path, which allows to subscribe to events specific for a single device (instead of subscribing to all the types and filter them by the path); also `TRANSPORT.STOPPED` event added
- https://github.com/trezor/trezor-suite/pull/17089/commits/9b9ad64a95ebe4a37f6c561de7c680c6639ed19c - based on previous commit, `Device` objects are now handling their own events, instead of passing them from `DeviceList`
- https://github.com/trezor/trezor-suite/pull/17089/commits/4f4468eb35647066e948672834870010c9b29086 - session info for `Device` object is now read from transports (where it's stored anyway) instead of holding a copy of it, which allows to reduce info needed for acquired `Device`s to a single `Session | null`
- https://github.com/trezor/trezor-suite/pull/17089/commits/3d84bdd68f3568045cdc53dfc946fe5eb4d2d566 - `Device`s are listening (apart from their own events) to `TRANSPORT.STOPPED` event as well and disconnecting themselves in that case, which allows to make `TransportManager` easier than before
- https://github.com/trezor/trezor-suite/pull/17089/commits/2c6a8fd0a7290b64b3ed2259d686187916b48b65 - minor improvement of `DeviceCurrentSession` - disposed is handled inside every `call`
- https://github.com/trezor/trezor-suite/pull/17089/commits/58834d53799435335318803dcd4aa45268fd5a46 - response validation in `DeviceCurrentSession` moved to a dedicated typeguard function
- https://github.com/trezor/trezor-suite/pull/17089/commits/28429338a9169327a511581dc6a68905a1b80aeb - `GetFeatures` -> `Initialize` fallback logic moved to `DeviceCurrentSession`, which allows to remove recursion in `Device._runInner`
- https://github.com/trezor/trezor-suite/pull/17089/commits/0e494ab5f9cfdfc4aa363c3d76a044192650fa7e - minor improvement of `DeviceCurrentSession` - the only `transport.send` is moved directly where it was used, so the ordinary `Cancel` logic could've been simplified as it's always `transport.call`
- https://github.com/trezor/trezor-suite/pull/17089/commits/04e230530c136594309b9dbaf01aa923cd142efe - transport errors are identified directly in `DeviceCurrentSession` and `resetDevice` method flow was polished and is now (hopefully) better readable
- https://github.com/trezor/trezor-suite/pull/17089/commits/bfdf67872ad98b3b5ea94b0f3c4cb0510c629cc4 - `DeviceCurrentSession` disposes itself on every device descriptor change instead of depending on `dispose` call from outside
- https://github.com/trezor/trezor-suite/pull/17089/commits/9e5e4ade46a87f2d3f396805556d4416454fc7e1 - this should fix a corner-case inside override e2e test by aborting `_runInner` immediately after the first `acquire` call if needed (but still, abort should be handled in every async call there, not just after the first one)

### DIFF
--- a/packages/connect/src/api/resetDevice.ts
+++ b/packages/connect/src/api/resetDevice.ts
@@ -1,6 +1,5 @@
 // origin: https://github.com/trezor/connect/blob/develop/src/js/core/methods/ResetDevice.js
 import { Assert } from '@trezor/schema-utils';
-import { ERRORS_WITHOUT_DEVICE_INTERACTION } from '@trezor/transport/src/errors-groups';
 import { getRandomInt } from '@trezor/utils';
 
 import { generateEntropy, verifyEntropy } from '../api/firmware/verifyEntropy';
@@ -10,11 +9,7 @@ import { UI } from '../events';
 import { getFirmwareRange } from './common/paramsValidator';
 import { validatePath } from '../utils/pathUtils';
 
-type EntropyRequestData = PROTO.EntropyRequest & { host_entropy: string };
-
 export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.ResetDevice> {
-    private isEntropyCheckSuccessful: boolean = false;
-
     init() {
         this.allowDeviceMode = [UI.INITIALIZE, UI.SEEDLESS];
         this.useDeviceState = false;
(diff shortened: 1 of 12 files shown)


mroz22 (approved):
# 🚢 🇮🇹

PR #17090, opened 2025-02-18
Extract reducers to common package

peter-sanderson:


packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -35,17 +33,21 @@ type BluetoothConnectProps = {
     uiMode: 'spatial' | 'card';
 };
 
+const selectAllDevices = prepareSelectAllDevices<BluetoothDevice>();
```
peter-sanderson, line 38: This has to be done to create generic selector with memoized output. Consulted with @Nodonisko

suite-common/bluetooth/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    BluetoothScanStatus,
+    DeviceBluetoothStatusType,
+} from './bluetoothReducer';
+
+export {
+    prepareSelectAllDevices,
+    selectKnownDevices,
+    selectAdapterStatus,
+    selectScanStatus,
+} from './bluetoothSelectors';
```
Lemonexe, line 15: might as well `export * from './files'`, as we usually do in these barrel files

suite-common/bluetooth/src/index.ts
```diff
@@ -0,0 +1,24 @@
+export {
+    bluetoothAdapterEventAction,
+    BLUETOOTH_PREFIX,
+    allBluetoothActions,
+    bluetoothConnectDeviceEventAction,
+    bluetoothNearbyDevicesUpdateAction,
+    bluetoothKnownDevicesUpdateAction,
+    bluetoothRemoveKnownDeviceAction,
+    bluetoothScanStatusAction,
+} from './bluetoothActions';
```
Lemonexe: why do you export individual actions, when you also export `allBluetoothActions`?
I'd either
delete `allBluetoothActions` and here write `export *`,
**or** only `export {BLUETOOTH_PREFIX, allBluetoothActions}` here.
Otherwise there will be two inconsistent ways of importing it.

suite-common/bluetooth/src/bluetoothReducer.ts
```diff
@@ -0,0 +1,150 @@
...
+    data: number[]; // Todo: consider typed data-structure for this
+    lastUpdatedTimestamp: number;
+};
+
+export type DeviceBluetoothStatusType = DeviceBluetoothStatus['type'];
+
+export type BluetoothState<T extends BluetoothDeviceCommon> = {
+    adapterStatus: 'unknown' | 'enabled' | 'disabled';
+    scanStatus: BluetoothScanStatus;
+    nearbyDevices: Array<{ device: T; status: DeviceBluetoothStatus | null }>;
```
Lemonexe: Move `BluetoothDeviceState` above, then you can do this, just like in `initialState`
```suggestion
    nearbyDevices: BluetoothDeviceState<T>[];
```

suite-common/bluetooth/src/bluetoothReducer.ts
```diff
@@ -0,0 +1,150 @@
...
+    device: T;
+    status: DeviceBluetoothStatus | null;
+};
+
+export const prepareBluetoothReducerCreator = <T extends BluetoothDeviceCommon>() => {
+    const initialState: BluetoothState<T> = {
+        adapterStatus: 'unknown',
+        scanStatus: 'idle',
+        nearbyDevices: [] as BluetoothDeviceState<T>[],
+        knownDevices: [] as T[],
```
Lemonexe, line 51: Nit: You can avoid `as` by defining `const nearbyDevices: BluetoothDeviceState<T>[] = []`, but not important, `as` is fine when used with constant expression..

suite-common/bluetooth/src/bluetoothReducer.ts
```diff
@@ -0,0 +1,150 @@
...
+            })
+            .addCase(bluetoothScanStatusAction, (state, { payload: { status } }) => {
+                state.scanStatus = status;
+            })
+            .addCase(deviceActions.deviceDisconnect, (state, { payload: { bluetoothProps } }) => {
+                if (bluetoothProps !== undefined) {
+                    state.nearbyDevices = state.nearbyDevices.filter(
+                        it => it.device.id !== bluetoothProps.id,
+                    );
+                }
```
Lemonexe, line 107: why shouldn't the device stay in nearby devices when disconnected?
If I understand correctly, nearby devices are devices that are visible and can be connected. Won't this prevent you from reconnecting until the next scan?

suite-common/bluetooth/tests/bluetoothReducer.test.ts
```diff
@@ -0,0 +1,243 @@
...
+            },
+        });
+
+        const nearbyDevices: BluetoothDeviceCommon[] = [
+            {
+                id: 'A',
+                data: [],
+                name: 'Trezor A',
+                lastUpdatedTimestamp: 1,
+            },
```
Lemonexe: Nit to make shorter & make clear that this is still the same device.
Although now that I see the rest of the file, you could extract Trezor A as const at beginning of file. It's repeated often.
```suggestion
            alreadyExistingNearbyDevice.device,
```

suite-common/bluetooth/tests/bluetoothReducer.test.ts
```diff
@@ -0,0 +1,243 @@
...
+            {
+                id: 'B',
+                data: [],
+                name: 'Trezor B',
+                lastUpdatedTimestamp: 2,
+            },
+        ]);
+    });
+
+    it('removes device from knownDevices when the device is disconnected by TrezorConnect', () => {
```
Lemonexe: Or what is actually supposed to happen? :thinking: [See comment in reducer file](https://github.com/trezor/trezor-suite/pull/17090#discussion_r1965167022)
```suggestion
    it('removes device from nearbyDevices when the device is disconnected by TrezorConnect', () => {
```

Lemonexe (commented):
CR still in progress; dropping off comments, mostly nits.

suite-common/bluetooth/src/bluetoothSelectors.ts
```diff
@@ -0,0 +1,38 @@
+import { createWeakMapSelector } from '@suite-common/redux-utils';
+
+import { BluetoothDeviceCommon, BluetoothDeviceState, BluetoothState } from './bluetoothReducer';
+
+export type WithBluetoothState<T extends BluetoothDeviceCommon> = {
```
Lemonexe, line 5: good naming :+1: 
:thought_balloon:  Often our convention is to just name it `BluetoothState`, and when a selector reaches to multiple sources, it's `BluetoothState & DeviceState`, but that does not make sense really..

suite-common/bluetooth/src/bluetoothSelectors.ts
```diff
@@ -0,0 +1,38 @@
...
+        (nearbyDevices, knownDevices) => {
+            const map = new Map<string, BluetoothDeviceState<T>>();
+            knownDevices.forEach(knownDevice => {
+                map.set(knownDevice.id, { device: knownDevice, status: null });
+            });
+
+            nearbyDevices.forEach(nearbyDevice => {
+                if (!map.has(nearbyDevice.device.id)) {
+                    map.set(nearbyDevice.device.id, nearbyDevice);
+                }
```
Lemonexe, line 37: how about adding this? Alternatively, switch the order of the forEaches: first add nearby, then known, if not in the map already.
Actually that makes sense also because knownDevices can have obsolete info, updated in nearby, can't it?
```
else {
map[nearbyDevice.device.id].status = nearbyDevice.status
```

suite-common/bluetooth/tests/bluetoothSelectors.test.ts
```diff
@@ -0,0 +1,48 @@
...
+            id: 'B',
+            data: [],
+            name: 'Trezor B',
+            lastUpdatedTimestamp: 2,
+        };
+
+        const state: WithBluetoothState<BluetoothDeviceCommon> = {
+            bluetooth: {
+                ...initialState,
+                nearbyDevices: [pairingNearbyDeviceA],
```
Lemonexe: if the [comment in selectors file](https://github.com/trezor/trezor-suite/pull/17090#discussion_r1965335492) makes sense, then the test should be expanded to cover it:
```suggestion
                nearbyDevices: [pairingNearbyDeviceA, pairingNearbyDeviceB],
```

suite-common/bluetooth/src/index.ts
peter-sanderson: I think we shall not do this `*` exports. This way you are unintentionally exposing stuff on the package API level every time you add something. And it is not true, that every export from file shall be export from the package (it can be exported for other file in the package only, or it can be export just for tests).

In this package a good example is `BluetoothDeviceCommon`, its type shall never leave this package.

peter-sanderson: (I am aware there is not such a think in the `bluetoothSelectors` but this principle applies here as well.)

suite-common/bluetooth/src/index.ts
peter-sanderson: :thinking: My sole reason for `allBluetoothActions` was to use it easily in `packages/suite/src/types/suite/index.ts`

I kinda dont like the `someGeneticPackagename.actionName` format. But I have no strong argument for it in this case. Its just an feeling.

peter-sanderson: (I general I like to import only what I use, but this is not relevant here as we export/import it anyway from this barrel index file).

suite-common/bluetooth/src/bluetoothReducer.ts
peter-sanderson: Good point, fixed (it was forgotten)

suite-common/bluetooth/src/bluetoothReducer.ts
peter-sanderson: Hmm, I like avoiding `as` but in this case it seems to me that it would create just more hard then good (another variable that needs to have some name `initialNearbyDevices` to not shadow the `nearbyDevices` in redudcre scope....

suite-common/bluetooth/src/bluetoothReducer.ts
peter-sanderson: I think in practice it wont matter. When you are opening the list, you call bluetooth start-scan and it will provide you with list of *all* nearby devices -> so you do just in-place replacement of whole array anyway. 

The only issue is with knownDevices as MacOS wont provide you with those in the nearbyDevices list so you have to add them (this is addressed in  `prepareSelectAllDevices` selector.

Maybe we can remove this completely. But I think it is useful to "clear" the state a bit. As I can imagine that by leaving it there it *may* lead to situation when you:

1) still have the **disconnected** device in the `nearbyDevices` state
2) device is actually offile
3) and you haven't received result of the new BT scan-devices yet

=> this *may* produce a list with "ghost" device that will fail to connect anyway.

I see this as a defensive programming practice to prevent potential issues.

suite-common/bluetooth/tests/bluetoothReducer.test.ts
peter-sanderson: Food point https://github.com/trezor/trezor-suite/pull/17090/commits/6a30f1ff5805850c148192dc271585728f750d1a

suite-common/bluetooth/src/bluetoothReducer.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/17090/commits/20fcc715922b30b9b128e5c153fdee558c97dbc0

suite-common/bluetooth/tests/bluetoothReducer.test.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/17090/commits/1d22b88882b3e493a535f620bcf7eb02d69dcbef

suite-common/bluetooth/src/bluetoothReducer.ts
```diff
@@ -0,0 +1,150 @@
...
+    // (because we already successfully paired them in the Suite) in the Operating System
+    knownDevices: T[];
+};
+
+export type BluetoothDeviceState<T extends BluetoothDeviceCommon> = {
+    device: T;
+    status: DeviceBluetoothStatus | null;
+};
+
+export const prepareBluetoothReducerCreator = <T extends BluetoothDeviceCommon>() => {
```
Lemonexe, line 46: this could do with JSDoc to explain why are you using a HOF.
Well, another layer of HOF, because createReducer is already a HOF :sweat_smile:

packages/suite/src/views/settings/SettingsDevice/BluetoothEraseBonds.tsx
```diff
@@ -15,6 +17,11 @@ export const BluetoothEraseBonds = () => {
         setInProgress(true);
         // TODO: missing button request in FW
         const result = await TrezorConnect.eraseBonds({ device });
+
+        if (device?.bluetoothProps?.id !== undefined) {
+            dispatch(bluetoothRemoveKnownDeviceAction({ id: device.bluetoothProps.id }));
+        }
+
```
Lemonexe, line 24: not pertinent _to this PR_, but the onclick callback should not be named `onCheckFirmwareAuthenticity`

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -133,16 +123,17 @@ export const BluetoothConnect = ({ onClose, uiMode }: BluetoothConnectProps) =>
     console.log('selectedDevice', selectedDevice);
 
     // This is fake, we scan for devices all the time
```
Lemonexe, line 135: is this still relevant? Seems to me that scanning is started and stopped at defined cases.

Lemonexe (approved):
LGTM. Big amount of work :muscle: :rocket:
I've got some nits and questions (see comments), but I'm not yet very familiar with the topic, so they may be irrelevant.
:arrow_right: feel free to resolve them in a way that you deem suitable. 

Also I did not test, because I don't have a BT device, and also because it's not merged into develop. I expect that big QA effort shall be done on feat/rust-bluetooth before merging that.

suite-common/bluetooth/src/bluetoothReducer.ts
Lemonexe: ok, feel free to resolve. Same thing in `bluetoothReducer.test.ts` and `bluetoothSelectors.test.ts` btw.

As I said, I'm fine with `as` when used on constant expression. That way it's completely predictable.

suite-common/bluetooth/src/index.ts
Lemonexe: Ok, but you're allowing for two different ways to import the same thing (confusing).
I'd prefer only one way - either to delete `allBluetoothActions`, or to export only `allBluetoothActions`.

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
peter-sanderson: Yes it is, the scanning starts by opening the component `bluetoothStartScanningThunk` and stops by unmounting the component `bluetoothStopScanningThunk`

This code is pure fake to create flow for the user where after some time we show some hints to do, when no devices found in 30 seconds.

packages/suite/src/views/settings/SettingsDevice/BluetoothEraseBonds.tsx
peter-sanderson: Hehe, nice catch -> fixed: https://github.com/trezor/trezor-suite/pull/17090/commits/eb34bdba41922e2b1389f66e435a85696ddb7c1e

suite-common/bluetooth/src/bluetoothReducer.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/17090/commits/c9d03f1c53bd912f59473ec39ffee6b69c044802

suite-common/bluetooth/src/bluetoothReducer.ts
Lemonexe: I see! Just to clarify if I understand the 2.: you mean the case when BT device is physically turned off?

So `deviceActions.deviceDisconnect` can be dispatched in three cases:

- user clicks disconnect on BT device, but it's still on - it should be possible to reconnect it, but that's taken care of in next scan (rediscovered as nearbyDevice)
- user ejects BT device from Suite - ditto
- user turns off BT device - UI must not allow reconnection because it would fail and confuse users

suite-common/bluetooth/src/bluetoothReducer.ts
Lemonexe: I meant `prepareBluetoothReducerCreator`, but you commented on `prepareSelectAllDevices` (but that's also helpful so thanks :+1: )

suite-common/bluetooth/src/index.ts
peter-sanderson: https://github.com/trezor/trezor-suite/pull/17090/commits/9dcc1e1826b25145cd16dff528b5fe8070021979 & https://github.com/trezor/trezor-suite/pull/17090/commits/e0b0c76ef47ddad87b5f13e9f62346e7e209f026

suite-common/bluetooth/src/bluetoothReducer.ts
peter-sanderson: Yes. For example when it is turned off (or dropped to Faraday Cage :rofl:)

peter-sanderson:
Closed in favor of: https://github.com/trezor/trezor-suite/pull/17213

PR #17093, opened 2025-02-18
feat(suite): Add CardButton component

jvaclavik:
## Description

This component is very simple and I was thinking if it worth for a new component, but I think yes. At least because of `isDisabled`, and in the future, we might improve it or unify it more.

## Related Issue

Resolve

### DIFF
--- a/packages/components/src/components/Card/utils.tsx
+++ b/packages/components/src/components/Card/utils.tsx
@@ -65,7 +65,7 @@ export const mapFillTypeToCSS = ({
             ${$isClickable &&
             css`
                 &:hover {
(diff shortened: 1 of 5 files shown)


packages/components/src/components/Card/utils.tsx
```diff
@@ -65,7 +65,7 @@ export const mapFillTypeToCSS = ({
             ${$isClickable &&
             css`
                 &:hover {
-                    box-shadow: ${$elevation === 1 && theme.boxShadowElevated};
```
vojtatranta, line 68: I can't tell what this does , but I trust you.

vojtatranta (approved):
Checked the "Enter on Trezor" and it looks good. (the same)

PR #17097, opened 2025-02-19
refactor(trading): minor improvements

### LINKED ISSUE #17049: Trading - code improvements
## Description
- [x] `tradingThunks` vs `tradingThunk.test`. Same for `buyThunks` vs `buyThunk.test`
- [x] improve test description according convention (https://github.com/trezor/trezor-suite/pull/16913#pullrequestreview-2614459178)
- [x] Consider to rewrite loadInvityData as a thunk.
- [x] refactor to `createSlice`
- [x] process feedback https://github.com/trezor/trezor-suite/pull/16866
- [x] `should call processResponseData with response and save trade` is flaky (mocked date)

## Related
Follow up https://github.com/trezor/trezor-suite/issues/16518

adderpositive:
## Description
- update file name
- update selectors
- update testing convention

## Related Issue
Partially resolve https://github.com/trezor/trezor-suite/issues/17049


jbazant (approved):
LGTM

LukasRada (approved):

izmy (approved):

TomasBoda (approved):

packages/suite/src/support/extraDependencies.ts
```diff
@@ -91,9 +91,12 @@ export const extraDependencies: ExtraDependencies = {
         selectRouterApp: (state: AppState) => state.router.app,
         selectRoute: (state: AppState) => state.router.route,
         selectAddressDisplayType: (state: AppState) => state.suite.settings.addressDisplayType,
+        selectSelectedAccount: (state: AppState) => state.wallet.selectedAccount,
         selectSelectedAccountStatus: (state: AppState) => state.wallet.selectedAccount.status,
         selectSuiteSettings,
         selectIsWindowVisible,
+        selectTradingEnvironment: (state: AppState) =>
+            state.suite.settings.debug.invityServerEnvironment,
```
vytick, line 99: Since we have trade in common now, would it make sense to move this to `wallet.trading` instead of settings debug? And also porbably not to have to use extra dependencies in that case. Wdyt?

packages/suite/src/support/extraDependencies.ts
```diff
@@ -91,9 +91,12 @@ export const extraDependencies: ExtraDependencies = {
         selectRouterApp: (state: AppState) => state.router.app,
         selectRoute: (state: AppState) => state.router.route,
         selectAddressDisplayType: (state: AppState) => state.suite.settings.addressDisplayType,
+        selectSelectedAccount: (state: AppState) => state.wallet.selectedAccount,
```
vytick, line 94: Should this go through extra dependencies? Is this platform specific that should be handled differently on different platforms?

suite-common/trading/src/middlewares/__tests__/tradingMiddleware.test.ts
```diff
@@ -27,15 +26,15 @@ type SelectedAccountStatus = {
     account: Account;
 };
 type SelectedAccountState = SelectedAccountStatus;
-const mockedSelectedAccountReducer = createReducerWithExtraDeps<SelectedAccountState>(
+const mockedSelectedAccountReducer = createReducer<SelectedAccountState>(
```
vytick, line 29: if we are not using reducer with extra deps here, we probably dont need extra deps?

suite-common/trading/src/middlewares/__tests__/tradingMiddleware.test.ts
```diff
@@ -46,16 +45,24 @@ const mockedSuiteReducer = createReducerWithExtraDeps(
     () => {},
 );
 
+const tradingMiddleware = prepareTradingMiddleware({
+    ...extraDependenciesMock,
+    selectors: {
+        ...extraDependenciesMock.selectors,
+        selectSelectedAccount: () => ({ status: 'loaded', account: accountBtc }) as any,
```
vytick, line 52: the same extra deps?

packages/suite/src/support/extraDependencies.ts
tomasklim: I think, we should leave it here in `settings.debug`

packages/suite/src/support/extraDependencies.ts
tomasklim: it should work, or at least in some previous PRs the selector before was exactly the same and it got approved

tomasklim: But i do not understand mobile and desktop account logic is extremely stupid so I expect that you did not copied it 😀

packages/suite/src/support/extraDependencies.ts
adderpositive: It depends if `suite.settings` is also used in native, I guess

tomasklim (approved):

suite-common/trading/src/middlewares/__tests__/tradingMiddleware.test.ts
adderpositive: this file is deleted in next PR, deprecated

packages/suite/src/support/extraDependencies.ts
vytick: it is not used there. Okay. In case @tomasklim wants to keep it in suite.settings, extraDeps make sense then.

vytick (approved):

PR #17099, opened 2025-02-19
Typed object keys

mroz22:
no runtime changes

### DIFF
--- a/packages/utils/src/typedObjectKeys.ts
+++ b/packages/utils/src/typedObjectKeys.ts
@@ -1,0 +1,2 @@
+export const typedObjectKeys = <T extends Record<any, any>>(obj: T): Array<keyof T> =>
+    Object.keys(obj) as Array<keyof T>;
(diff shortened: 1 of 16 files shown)


packages/utils/src/typedObjectKeys.ts
```diff
@@ -0,0 +1,2 @@
+export const typedObjectKeys = <T extends Record<any, any>>(obj: T): Array<keyof T> =>
```
vojtatranta, line 1: There's a catch to it
```
A common ask is to provide 'better' typings for Object.keys, so that it returns Array<keyof T> instead of Array<string>. Same for Object.entries. ts-reset won't be including rules to change this.

TypeScript is a structural typing system. One of the effects of this is that TypeScript can't always guarantee that your object types don't contain excess properties:

type Func = () => {
  id: string
}

const func: Func = () => {
  return {
    id: '123',
    // No error on an excess property!
    name: 'Hello!',
  }
}
So, the only reasonable type for Object.keys to return is Array<string>.
```

But I guess we can risk it. Since we always basically typecast it.

marekrjpolak: Yeah, it's not correct. But it's not more broken than it already was either. :innocent:

marekrjpolak (approved):
It's an improvement.

PR #17101, opened 2025-02-19
Tenv e2e

mroz22:
would be nice to have something to prove that trezor-user-env-link to trezor-user-env server integration works as expected. this is mostly intended to run on tenv image updates so that we can tell where potential failures come from.


packages/trezor-user-env-link/e2e/api.test.ts
```diff
@@ -0,0 +1,44 @@
...
+                model,
+                version,
+                wipe: true,
+            });
+
+            console.log('setting up emulator');
+            await TrezorUserEnvLink.setupEmu();
+
+            await TrezorUserEnvLink.allowUnsafePaths();
+            // todo: run more complex commands, typically those that require interaction with multiple screens (dry run..)
```
mroz22, line 28: I need to get device somehow into recovery/backup state. Can we do it with current tenv API? I can't see anything. If no, could you please either expose some trezorlib method which allows this, or maybe if no such method exists I would be ok with passing raw bytes. cc @grdddj (no hurry here)

HajekOndrej (approved):

packages/trezor-user-env-link/e2e/api.test.ts
grdddj: I have added `emulator-set-for-backup` method into the websocket API. It requires the bridge to be running - otherwise it cannot work properly

PR #17119, opened 2025-02-19
feat(mobile): prevent 3rd party access our packages

Nodonisko:
## Description

This should add basic protection agains 3rd party packages access our internal libs or Connect which they have no reason to do unless it's something malicious.

## Related Issue

Resolve

### DIFF
--- a/packages/bundler-security/src/WebpackSecurityPlugin.ts
+++ b/packages/bundler-security/src/WebpackSecurityPlugin.ts
@@ -1,0 +1,36 @@
+/* eslint-disable no-console */
+import webpack from 'webpack';
+
+import { checkSecurityViolation } from './metroSecureResolver';
+
+export class WebpackSecurityCheckPlugin {
+    private logError(message: string) {
+        console.error(`\x1b[1;31m${message}\x1b[0m`);
+    }
+
+    apply(compiler: webpack.Compiler) {
+        compiler.hooks.normalModuleFactory.tap('SecurityCheckPlugin', factory => {
+            factory.hooks.resolve.tap('SecurityCheckPlugin', resolveData => {
+                const { request, context, contextInfo } = resolveData;
+
+                const { isViolation, normalizedPath } = checkSecurityViolation({
+                    moduleName: request,
+                    originModulePath: context,
+                });
+
+                if (isViolation) {
+                    console.log('\n\n');
+                    this.logError(
+                        `SECURITY ALERT: A third-party package is trying to import internal Trezor package! ` +
+                            `\nRequesting module: ${request}` +
+                            `\nNormalized path: ${normalizedPath}` +
+                            `\nImmediate importer: ${context}` +
+                            `\nOriginal issuer: ${contextInfo.issuer}`,
+                    );
+                    // Manually exit othewise webpack will only print error and happily continue building
+                    process.exit(1);
+                }
+            });
+        });
+    }
+}
--- a/packages/bundler-security/src/constants.js
+++ b/packages/bundler-security/src/constants.js
@@ -1,0 +1,11 @@
+const TREZOR_PACKAGES_SCOPES = ['@trezor', '@suite-common', '@suite-native'];
+const PACKAGES_PATHS = ['/packages/', '/suite-common/', '/suite-native/'];
+
+const RELATIVE_PATH_REGEX = new RegExp(
+    `(?:\\.\\./){2,}(?:${PACKAGES_PATHS.map(p => p.slice(1, -1)).join('|')})`,
+);
+
+module.exports = {
+    TREZOR_PACKAGES_SCOPES,
+    RELATIVE_PATH_REGEX,
+};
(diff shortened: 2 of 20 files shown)


packages/bundler-security/tests/metroSecureResolver.test.ts
```diff
@@ -0,0 +1,75 @@
+import { checkSecurityViolation } from '../src/metroSecureResolver';
+
+describe('checkSecurityViolation', () => {
+    it('should detect violation when accessing @trezor scoped package from node_modules', () => {
```
peter-sanderson, line 4: Its maybe a nit, but I think we are shall be using this format (I think I read it even somewhere in the Notion)

```suggestion
    it('detects violation when accessing @trezor scoped package from node_modules', () => {
```

peter-sanderson (approved):
LGTM (haven't tested)

This is really cool. :clap:

mroz22:
looks good and safe (only build time if I get it right).

packages/bundler-security/src/metroSecureResolver.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @param {SecurityCheckParams} params
+ * @returns {{isViolation: boolean, normalizedPath: string}}
+ */
+const checkSecurityViolation = ({ moduleName, originModulePath }) => {
+    const normalizedPath = path.normalize(moduleName);
+
+    const isNodeModules = originModulePath.includes('node_modules');
+    const isScopedPackage = TREZOR_PACKAGES_SCOPES.some(scope => normalizedPath.includes(scope));
+    const isRelativePath = RELATIVE_PATH_REGEX.test(normalizedPath);
```
matejkriz, line 26: What `isRelativePath` mean in this context? Could you please provide an explanation comment here?

packages/bundler-security/src/constants.js
```diff
@@ -0,0 +1,12 @@
+const TREZOR_PACKAGES_SCOPES = ['@trezor', '@suite-common', '@suite-native'];
+const PACKAGES_PATHS = ['/packages/', '/suite-common/', '/suite-native/'];
```
matejkriz, line 2: What if some node_module have another 3rd party dependency that also use `/packages` folder?

packages/suite-desktop-core/tsconfig.lib.json
matejkriz, line 1: Why this change?

packages/bundler-security/src/constants.js
```diff
@@ -0,0 +1,12 @@
...
+
+const RELATIVE_PATH_REGEX = new RegExp(
+    `(?:\\.\\./){2,}(?:${PACKAGES_PATHS.map(p => p.slice(1, -1)).join('|')})`,
+);
+
+module.exports = {
+    TREZOR_PACKAGES_SCOPES,
+    PACKAGES_PATHS,
```
matejkriz: PACKAGES_PATHS are not imported anywhere, does it make sense to export it?

packages/bundler-security/src/WebpackSecurityPlugin.ts
```diff
@@ -0,0 +1,36 @@
...
+                    console.log('\n\n');
+                    this.logError(
+                        `SECURITY ALERT: A third-party package is trying to import internal Trezor package! ` +
+                            `\nRequesting module: ${request}` +
+                            `\nNormalized path: ${normalizedPath}` +
+                            `\nImmediate importer: ${context}` +
+                            `\nOriginal issuer: ${contextInfo.issuer}`,
+                    );
+                    // Manually exit othewise webpack will only print error and happily continue building
+                    process.exit(1);
```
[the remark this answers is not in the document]
Nodonisko: no

packages/suite-desktop-core/tsconfig.lib.json
Nodonisko: I just called `yarn refs`, not sure why it's different than develop but it looks correct now.

packages/bundler-security/src/constants.js
Nodonisko: It will break for sure, but I don't think will happen and we add whitelist then or something but I think it would premature optimalization now

packages/bundler-security/src/constants.js
Nodonisko: i think it's fine to remove it

packages/bundler-security/src/metroSecureResolver.js
Nodonisko: added few comments

packages/bundler-security/src/metroSecureResolver.js
```diff
@@ -0,0 +1,51 @@
...
+
+    return {
+        isViolation: (isScopedPackage || isRelativePath) && isNodeModules,
+        normalizedPath,
+    };
+};
```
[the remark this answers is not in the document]
Nodonisko: thats already normalized by webpack

packages/bundler-security/tests/metroSecureResolver.test.ts
Nodonisko: honestly I think it doesn't matter

packages/suite-desktop-core/tsconfig.lib.json
Nodonisko: interesting it seems that desktop build fails now because of this... if I remove it work fine, but it should not be possible to have this in develop because yarn refs must not produce diff on develop... I guess it's refs scripts is fucked up

PR #17124, opened 2025-02-20
fix(suite): fix inability to select device (silent return) when one device is already remembered

### LINKED ISSUE #17114: New device can't be accessed when having a remembered disconnected wallet
I have a remembered wallet from Device A, which is not connected.
I connect Device B, but it's not possible to access it.

https://github.com/user-attachments/assets/2f1f381d-2d8d-4612-8a31-561b48d58e09

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite/issues/17114

@coderabbitai ignore


martykan (approved):
Works for me

PR #17132, opened 2025-02-20
chore(connect-explorer, connect-explorer-theme): update next

komret:
Update `next` to resolve [dependabot issues](https://github.com/trezor/trezor-suite/security/dependabot?q=package%3Anext+manifest%3Ayarn.lock+has%3Apatch+is%3Aopen). Upgrade to v15 is not possible yet because it requires React 19.


martykan (approved):
Patch version update so there should be no issues

PR #17143, opened 2025-02-20
feat(suite): improved transaction titles for tokens

### LINKED ISSUE #17141: ERC-20 transaction titles are inaccurate
We need smarter handling of both simple and complex ERC-20 transaction titles.

Let's take a look at this example:

Image

1. "Sent CVC" -> this is actually a swap from ETH to CVC on 1inch and we are **receiving CVC**. I would expect to see "Contract transaction" or ideally "Swap ETH to CVC"
2. "Tranfser" -> this is just a simple transfer, the name comes from ethereumSpecific.parsedData. But I would expect to see "Sent TUSD"

martykan:
## Description

1. Don't use `ethereumSpecific.parsedData` as title for Transfer transactions, since it's less descriptive than our default title.
2. Fix condition for single token transactions to also check target, which would mean it's not really a single token transaction and would have resulted in an inaccurate title.
3. Added support for titles for Swap transactions. These are classified as transactions, where either 2 tokens are exchanged or 1 token is exchanged for the native token.

## Related Issue

Resolve #17141

## Screenshots:

Screenshot 2025-02-20 at 16 19 55

### DIFF
--- a/packages/suite/src/components/wallet/TransactionItem/TransactionHeader.tsx
+++ b/packages/suite/src/components/wallet/TransactionItem/TransactionHeader.tsx
@@ -84,7 +84,12 @@ const getSolTransactionStakeTypeName = (stakeType: StakeType) => {
 export const TransactionHeader = ({ transaction, isPending }: TransactionHeaderProps) => {
     const { translationString } = useTranslation();
 
-    if (transaction?.ethereumSpecific?.parsedData?.name && transaction.type !== 'failed') {
+    if (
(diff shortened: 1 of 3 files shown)


tomasklim (approved):
LGTM

However, do we use some custom 1inch contract that there is no name for this method?

martykan:
We should be using the standard 1inch router contract. Not sure why Blockbook is not decoding it.
But this should be better since it also includes the token symbols.

PR #17152, opened 2025-02-21
Vite as suite devserver

vojtatranta:
## TODO
- ✅ static routes
- ✅ simplify vite config (remove useless nonsense)
-  ✅ remove the changes in the source code / make them compatible with webpack
- WONT DO: MOVE the config to `suite-build`

## TO CHECK !!!!

- ** webpack dev server ** - runs okay
- ** webpack build ** - runs okay 

For each webpack build, dev server, vite devserer, check this:
- loading of graphs / balance charts (dashboards + accounts)
- loading of the balances
- trade / swap pages
- flags (the country / currency dropdown in trade)
- cardano account (werid, it bugged for vite)
- ripple account (weird, it bugged for vite)
- PHYSICLA DEVICE CONNECTION

## Related Issue

Resolve


packages/address-validator/src/index.js
```diff
@@ -3,7 +3,7 @@ const { addressType } = require('./crypto/utils');
 
 var DEFAULT_CURRENCY_NAME = 'bitcoin';
 
-module.exports = {
```
vojtatranta: this should be made compatible with both

packages/address-validator/src/crypto/cnBase58.js
```diff
@@ -1,4 +1,4 @@
-var JSBigInt = require('./biginteger')['JSBigInt'];
```
vojtatranta, line 1: this shoudl work normally, altought it can be polyfilled

packages/address-validator/src/crypto/biginteger.js
```diff
@@ -1457,4 +1457,6 @@
     })();
 
     exports.JSBigInt = BigInteger; // exports.BigInteger changed to exports.JSBigInt
+    global.JSBigInt = BigInteger;
```
vojtatranta: remove

packages/connect/src/impl/dynamic.ts
```diff
@@ -100,8 +100,15 @@ export class TrezorConnectDynamic<
 
         // Initialize the target
         try {
+            console.log('settings', this.lastSettings);
+            console.log('DENG BUG ggg');
```
vojtatranta: debug stuff, will remove

packages/suite-web/index.html
```diff
@@ -0,0 +1,60 @@
+<!doctype html>
```
vojtatranta: this should be unified with the webpack's index.html file, I think thats possible just some variables willneed to be replaced coz it uses webpack's template literals.

packages/suite-web/package.json
```diff
@@ -6,9 +6,11 @@
         "type-check": "yarn g:tsc --build tsconfig.json",
         "type-check:watch": "yarn type-check -- --watch",
         "lint:styles": "npx stylelint './src/**/*{.ts,.tsx}' --cache --config ../../.stylelintrc",
-        "dev": "yarn rimraf ./build && yarn workspace @trezor/suite-build run dev:web",
+        "dev": "yarn rimraf ./build && vite",
+        "dev:webpack": "yarn rimraf ./build && yarn workspace @trezor/suite-build run dev:web",
```
vojtatranta: therew ill be like `dev:vite` webpack will be primary, AI changed it

packages/suite-web/src/index.ts
```diff
@@ -1,22 +1,25 @@
+import './polyfills';
```
vojtatranta: I'll try to include these polyfills through vite rather than in the source code / will make a simple vite entrypoint for it so that the source code isn't poluted

packages/suite-web/src/index.ts
```diff
@@ -1,22 +1,25 @@
...
+
 /*
  * On the start of the app, the index.html file is without div with id="app" to which React app should be rendered.
  * This div is added thanks to browser detection plugin in case that a user uses supported browser.
  * After the div is added, MutationObserver detects the change and adds React app to the DOM.
  */
 
 const observer = new MutationObserver(() => {
-    const appElement = document.getElementById('app');
-    if (appElement) {
```
vojtatranta: will be reverted

suite-native/formatters/src/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { BigNumber } from '@trezor/utils/src/bigNumber';
```
vojtatranta: thanks. AI, this is a nonsense...

packages/suite-web/package.json
```diff
@@ -7,6 +7,7 @@
         "type-check:watch": "yarn type-check -- --watch",
         "lint:styles": "npx stylelint './src/**/*{.ts,.tsx}' --cache --config ../../.stylelintrc",
         "dev": "yarn rimraf ./build && yarn workspace @trezor/suite-build run dev:web",
+        "dev:vite": "yarn rimraf ./build && vite",
```
vojtatranta: thius should be  likely in `@trezor/suite-build` not here

packages/connect/src/core/method.ts
```diff
@@ -7,16 +7,41 @@ import type { AbstractMethod } from './AbstractMethod';
...
+
 export const getMethod = async (message: IFrameCallMessage): Promise<AbstractMethod<any>> => {
     const { method } = message.payload;
     if (typeof method !== 'string') {
         throw TypedError('Method_InvalidParameter', 'Message method is not set');
     }
 
     const methodModule = getMethodModule(method);
-    const methods = methodModule
-        ? await import(/* webpackChunkName: "[request]" */ `../api/${methodModule}/api`)
```
Lemonexe, line 18: do I understand that vite does not support dynamic imports?

packages/suite/src/reducers/suite/guideReducer.ts
```diff
@@ -22,31 +20,43 @@ export const initialState: State = {
...
-                draft.currentNode = null;
-                break;
-            case GUIDE.OPEN_NODE:
-                draft.currentNode = action.payload;
-                break;
-            default:
-                return state;
-        }
-    });
+// NOTE: we cannot use immer in this reducer, because GuideCategory mimics the react node and immer uses Object.freeze()
```
Lemonexe, line 23: just curious - did this break specifically when using vite dev server? Sounds like an unrelated issue, which has potentially been there even with webpack, maybe just not manifesting?

packages/suite-web/index.html
Lemonexe: Sounds doable to make a script to replace or strip the webpack syntax → make it digestible by Vite.
Question is, when should this script fire?
Quick search: vite might have a hook  that allows you to modify the static assets on-the-fly, which would be better than running it manually after each `index.html` change (although that'd be fine too, this file is not modified very often :slightly_smiling_face:  )

packages/connect/src/core/method.ts
vojtatranta: I does, but this particular import worked pretty weirdly, it tried to load the file from the wrong path for some reason. So this was the simplest fix I could think off.

packages/suite/src/reducers/suite/guideReducer.ts
vojtatranta: I think this was broken with webpack. But Vite server "fixed it". Meaning, Vite passed the right object reference which was correctly frozen. 

But I would lie if I say that got deep into it. Nevertheless, the redux stores here "schema" of the react tree which is incorrect or at least, it is not a best practice.

vojtatranta:
@Lemonexe I have troubles moving the vite dev server config to the `suite-build` package. For some weir reason a second day in row I am not able to make it run. I think I would have to do the whole thing from ground up which will be time consuming like hell..

Do you think we could merge it "as is" since it is just a "dev" thing?

packages/suite-web/index.html
vojtatranta: Yeah I know it is possible, I have it in one of the commits. Though, I have troubles moving the whole vite config to `suite-buiild` package for some reason. I tried many things already...

packages/suite-web/vite.config.ts
```diff
@@ -0,0 +1,180 @@
...
+import react from '@vitejs/plugin-react';
+import { execSync } from 'child_process';
+import { readdirSync } from 'fs';
+import { resolve } from 'path';
+import { Plugin, ViteDevServer, defineConfig } from 'vite';
+import { nodePolyfills } from 'vite-plugin-node-polyfills';
+import wasm from 'vite-plugin-wasm';
+
+import { suiteVersion } from '../suite/package.json';
+import { assetPrefix, project } from '../suite-build/utils/env';
```
Lemonexe: :thought_balloon:  should be imported from `@trezor/suite-build`, because it goes through workspace root, but I see that's not possible to import `@trezor/suite-build`.
Anyway, this will fix itself once the whole vite config is successfully moved to `suite-build`

packages/suite-web/vite.config.ts
```diff
@@ -0,0 +1,180 @@
...
+        alias: [
+            {
+                find: /^@trezor\/connect(\/index)?$/,
+                replacement: '@trezor/connect-web/src/module',
+            },
+            {
+                find: 'src',
+                replacement: resolve(__dirname, '../suite/src'),
+            },
+            ...Object.entries(createWorkspaceAliases()).map(([find, replacement]) => ({
```
Lemonexe: why not construct the `{ find replacement }` right in the `createWorkspaceAliases` fn, so you wouldn't have to map it here?

vojtatranta:
@marekrjpolak @martykan Guys, we'd like to know your opinion about changes in connect that were made due to Vite having problem parsing the imports properly https://github.com/trezor/trezor-suite/pull/17152#pullrequestreview-2647029095

Would you have time for it? cc @trezor/suite-connect

packages/connect-web/src/module/index.ts
```diff
@@ -80,6 +80,8 @@ const TrezorConnect = factory(
 export default TrezorConnect;
 export * from '@trezor/connect/src/exports';
 
-window.addEventListener('beforeunload', () => {
-    impl.dispose();
-});
+if (typeof window !== 'undefined') {
+    window.addEventListener('beforeunload', () => {
```
vojtatranta, line 84: this was also pretty werid though, `windows` hasn't been defined here. The code isn't neede for dev to run properly and the change here won't change much, but still. We should know abou tit.

packages/address-validator/src/crypto/cnBase58.js
```diff
@@ -1,4 +1,5 @@
...
+var moduleImport = require('./biginteger');
+var JSBigInt = moduleImport.JSBigInt;
```
vojtatranta, line 2: Vite had a problem to undrestand this. This change also shouldn't bug anything.

packages/suite-web/src/vite-index.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * On the start of the app, the index.html file is without div with id="app" to which React app should be rendered.
+ * This div is added thanks to browser detection plugin in case that a user uses supported browser.
+ * After the div is added, MutationObserver detects the change and adds React app to the DOM.
```
vojtatranta: I should delete this

packages/suite-web/vite.config.ts
```diff
@@ -0,0 +1,160 @@
...
+// Plugin to serve core.js in dev mode
+
+// Plugin to handle workers similar to webpack's worker-loader
+const workerPlugin = (): Plugin => ({
+    name: 'worker-loader',
+    transform(code, id) {
+        if (/\/workers\/[^/]+\/index\.ts$/.test(id)) {
+            // Return a virtual module that creates a web worker
+            return {
+                code: `
```
vojtatranta, line 57: Vite is pretty cool

packages/suite/src/hooks/guide/useGuideLoadArticle.ts
```diff
@@ -5,7 +5,7 @@ import type { GuideNode } from '@suite-common/suite-types';
 import type { Locale } from 'src/config/suite/languages';
 
 export const loadPageMarkdownFile = async (id: string, language = 'en'): Promise<string> => {
-    const file = await import(`@trezor/suite-data/files/guide/${language}${id}`);
+    const file = await import(/* @vite-ignore */ `@trezor/suite-data/files/guide/${language}${id}`);
```
vojtatranta, line 8: This import was absolutely okay, but that dynamic import with relative paths was broken. Weird.

packages/suite/src/support/workers/graph/index.ts
```diff
@@ -20,4 +20,4 @@ ctx.addEventListener('message', (event: CustomMessageEvent) => {
 });
 
 // // Trickery to fix TypeScript since this will be done by "worker-loader"
-export default {} as typeof Worker & (new () => Worker);
```
vojtatranta, line 23: This was incorrect IMHO. `webpack-loader` returns the factory not "class". I checked the compiled code. So this is correct typing.

packages/suite-web/vite.config.ts
```diff
@@ -0,0 +1,160 @@
...
+
+import { suiteVersion } from '../suite/package.json';
+import { assetPrefix, project } from '../suite-build/utils/env';
+
+// Plugin to serve static files with /static prefix
+const staticAliasPlugin = (): Plugin => ({
+    name: 'static-alias',
+    configureServer(server: ViteDevServer) {
+        // Middleware to handle /static requests
+        server.middlewares.use((req, res, next) => {
```
Lemonexe: nit: unused variable
```suggestion
        server.middlewares.use((req, _res, next) => {
```

packages/suite-web/vite.config.ts
```diff
@@ -0,0 +1,160 @@
...
+        });
+    },
+});
+
+// Plugin to serve core.js in dev mode
+
+// Plugin to handle workers similar to webpack's worker-loader
+const workerPlugin = (): Plugin => ({
+    name: 'worker-loader',
+    transform(code, id) {
```
Lemonexe: nit: unused variable
```suggestion
    transform(_code, id) {
```

Lemonexe (commented):
Looks good, great work! :open_mouth: :rocket:  I'll play with it some more to see if something isn't broken, or if we can simplify some things.

- Suite Web works
- first load (without vite cache) took like a minute, but afterwards it's much quicker than webpack!
- HMR is near-instant :zap: :heart: 
- Overall, the app feels sluggish – initial load, popups, tab navigation etc., it seems to be slower than webpack, which makes sense... But that doesn't matter at all for dev server (won't affect users), so nvm :heavy_check_mark: 

Please, could you mention `yarn suite:dev:vite` in root `README.md`?
I'd write there that it is available as alternative to `yarn suite:dev`, but with :warning:  warning that it is experimental, may be unreliable → use the webpack one if you want fidelity to production app.

packages/suite-web/vite.config.ts
vojtatranta: <3 adding

packages/suite-web/vite.config.ts
vojtatranta: yup

packages/suite-web/vite.config.ts
vojtatranta: AI bs, wil do

packages/suite-web/vite.config.ts
vojtatranta: yeah, tried that and it throws :-D like Vite imports it as if it was javascript not Typescript, lol, don't know how to fix it.

vojtatranta: I added the path to references in tsconfig here, so it doesn't throw.

packages/connect/src/core/method.ts
marekrjpolak: Probably [this](https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#imports-must-end-with-a-file-extension)? Try add `index.ts` to the end: `../api/${methodModule}/api/index.ts`

packages/suite-web/tsconfig.json
```diff
@@ -29,6 +29,7 @@
         { "path": "../suite-analytics" },
         { "path": "../transport" },
         { "path": "../trezor-user-env-link" },
-        { "path": "../utils" }
+        { "path": "../utils" },
+        { "path": "../suite-build" }
```
Lemonexe: `tsconfig.json` files are not to be updated manually, they are generated by `yarn update-project-references`  (alias `yarn refs`).
→ add as devDependency in `suite-web/package.json` and run that, it should produce this

marekrjpolak:
Ordinary `suite:dev` broken for me.
image

packages/suite-web/tsconfig.json
vojtatranta: yup, done that, thanks

vojtatranta:
> Ordinary `suite:dev` broken for me. image

had that too `yarn && yarn build:libs` fixed it (likely not problem of thsi PR)

vojtatranta:
> > Ordinary `suite:dev` broken for me. image
> 
> had that too `yarn && yarn build:libs` fixed it (likely not problem of thsi PR)

damn, no longer helps, why the helll. Likely deduping?

Lemonexe:
Tested Webpack: dev Desktop, dev Desktop, linux Desktop build.
They all work correctly :heavy_check_mark: 
But Webpack throws a very red warning at all of them, that should be solved :x: 
EDIT: [same as already reported](https://github.com/trezor/trezor-suite/pull/17152#issuecomment-2688388940)

vojtatranta:
> Tested Webpack: dev Desktop, dev Desktop, linux Desktop build. They all work correctly ✔️ But Webpack throws a very red warning at all of them, that should be solved ❌
> 
> ```
> Module not found: Error: Can't resolve 'vm' in 'trezor-suite/node_modules/asn1.js/lib/asn1'
> 
> BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
> This is no longer the case. Verify if you need this module and configure a polyfill for it.
> ```

yup, see that too. Server runs, but throws errors. Looking into it.

Probably some yarn deduping caused this...

vojtatranta:
> Tested Webpack: dev Desktop, dev Desktop, linux Desktop build. They all work correctly ✔️ But Webpack throws a very red warning at all of them, that should be solved ❌ EDIT: [same as already reported](https://github.com/trezor/trezor-suite/pull/17152#issuecomment-2688388940)

Added what webpack advised us to do. It is no longer with warning. I hope that's correct.

packages/connect/src/core/method.ts
vojtatranta: I think I tried it, but I can do it again, gimme a sec

marekrjpolak: Actually with the big switch as it is right now, webpack bundles all the shitcoin code into a single chunk called `[request]`.

vojtatranta: Cool! WOrks! I did it without `.ts` and it failed, great work! Thanks!

vojtatranta: The fix works, confirmed on vite and webpack, thanks. REsolving this.

packages/suite/src/hooks/guide/useGuideLoadArticle.ts
vojtatranta: resolved, the file should be imported with full path along with extension.

packages/suite-web/vite.config.ts
```diff
@@ -0,0 +1,163 @@
...
+import { resolve } from 'path';
+import { Plugin, ViteDevServer, defineConfig } from 'vite';
+import { nodePolyfills } from 'vite-plugin-node-polyfills';
+import wasm from 'vite-plugin-wasm';
+
+import { suiteVersion } from '../suite/package.json';
+
+// Use require instead of import for TypeScript files
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { assetPrefix, project } = require('../suite-build/utils/env');
```
vojtatranta, line 12: damn it, I had to do this... Otherviews, the linter was screeming and with `@trezor/suite-build` would allow the vite config to compile, that's mess, but this wworks

packages/suite-web/index.html
vojtatranta: Note to me: don't forget to server the static/index.html file here, not this artificial one.

packages/suite-web/vite.config.ts
```diff
@@ -0,0 +1,162 @@
...
+            .map(dirent => [`@trezor/${dirent.name}`, resolve(__dirname, '../', dirent.name)]),
+    );
+
+    return Object.entries({
+        ...suiteCommonAliases,
+        ...trezorPackagesAliases,
+    }).map(([find, replacement]) => ({
+        find,
+        replacement,
+    }));
```
Lemonexe: I meant this with the earlier comment – much simpler to do the same thing :slightly_smiling_face: 
```suggestion
const suiteCommonAliases = readdirSync(resolve(__dirname, '../../suite-common'), {
        withFileTypes: true,
    })
        .filter(dirent => dirent.isDirectory())
        .map(dirent => ({
            find: `@suite-common/${dirent.name}`,
            replacement: resolve(__dirname, '../../suite-common', dirent.name),
        }));

    const trezorPackagesAliases = readdirSync(resolve(__dirname, '../'), { withFileTypes: true })
        .filter(dirent => dirent.isDirectory() && dirent.name !== 'suite-web')
        .map(dirent => ({
            find: `@trezor/${dirent.name}`,
            replacement: resolve(__dirname, '../', dirent.name),
        }));

    return [...suiteCommonAliases, ...trezorPackagesAliases];
```

vojtatranta: oki

Lemonexe:
@SocketSecurity ignore npm/vite-plugin-wasm@3.4.1
I have checked the [vite-plugin-wasm source](https://github.com/Menci/vite-plugin-wasm/tree/main/src) and verified legit usage: to generate code that fetches wasm files, using import urls by passed from vite; no external url injected in the lib :heavy_check_mark:

packages/components/src/components/Flag/Flag.tsx
```diff
@@ -15,13 +18,26 @@ export interface FlagProps {
...
+                console.error('Flag image loading error: ', err);
+            });
+    }, [country]);
+
+    return (
+        <Wrapper>
+            {src ? (
+                <img src={src} width={`${size}px`} alt={`flag-${country}`} className={className} />
+            ) : (
+                <SkeletonRectangle width={size} height={size} />
```
Lemonexe, line 39: nice solution :tada:

packages/suite-web/vite.config.ts
```diff
@@ -0,0 +1,290 @@
...
+    template = template.replace(
+        '</head>',
+        '<script type="module" src="./vite-index.ts"></script></head>',
+    );
+
+    // Add the app div to the body
+    template = template.replace('</body>', '<div id="app"></div></body>');
+
+    return template;
+}
```
Lemonexe: As discussed in person, I'd much simplify it.
What do we need?
- Replace one vite variable `assetPrefix`
- Remove the condition and the meta tag inside it

IMHO no need to write "our own library" with a generic and extensible behavior, when it's at the cost of complexity. I feat this would be very hard to maintain.
I suggest a very simplistic code that does only what we need.
After all, the `index.html` is not changed very often..

Besides, the original code had several issues I was about to point out :sweat_smile: 

```suggestion
// Function to process the HTML template with template variables
const processTemplate = (template: string): string =>
    template
        // Hardcoded replace for the only used webpack template variable. If we need more in future, we may develop a more generic code
        .replace(/<%=\s*assetPrefix\s*%>/g, assetPrefix)
        // Remove the webpack template conditional (opening + closing statements as well as the HTML in between)
        .replace(/<%\s*if\([^%]*%>[\s\S]*?<%\s*}\s*%>/g, '')
        // Add the script tag for vite-index.ts
        .replace('</head>', '<script type="module" src="./vite-index.ts"></script></head>')
        // Add the app div to the body
        .replace('</body>', '<div id="app"></div></body>');
```

Lemonexe (approved):
I think this is now ready to merge :rocket: 
Amazing work :raised_hands: :tada: 
Just one remaining comment to be solved, please.

Tested:
- vite Suite Web
  - somehow it's very slow in Firefox, but much faster in Chrome
  - HMR :heart: 
- webpack Suite Web
- webpack Suite Desktop
- Suite Desktop local build for Linux

Also: repeating myself, but don't merge until after freeze :crossed_fingers:

martykan:
Screenshot 2025-03-03 at 17 18 24

Can't actually connect to a device, probably related to `crypto-browserify`. 
And no I'm not using Internet Explorer 10

vojtatranta:
> Screenshot 2025-03-03 at 17 18 24
> Can't actually connect to a device, probably related to `crypto-browserify`. And no I'm not using Internet Explorer 10

and which browser?

It's firefox, right?

martykan:
Chrome and Firefox both don't work for me. I tried with bridge as well as WebUSB.

vojtatranta:
> Chrome and Firefox both don't work for me. I tried with bridge as well as WebUSB.

yeah, doesn't work for me too. Weird. On it

vojtatranta:
> > Chrome and Firefox both don't work for me. I tried with bridge as well as WebUSB.
> 
> yeah, doesn't work for me too. Weird. On it

I reestarted the emulator and it works fine. Wat

vojtatranta:
> Chrome and Firefox both don't work for me. I tried with bridge as well as WebUSB.

yeah real device isn't doing anything

vojtatranta:
> Chrome and Firefox both don't work for me. I tried with bridge as well as WebUSB.

got it !

There was wrongly `global: {}` as defined variable, that overriden the privous plugin that defined global as `window` which was correct. Now it works well. Thanks!

PR #17153, opened 2025-02-21
chore(monorepo): remove dependency on connect in components and produ…

mroz22:
…ct-components

I'd say that components and probably also product-components should not depend on connect.


packages/protobuf/src/messages-schema.ts
```diff
@@ -1,15 +1,5 @@
+import { DeviceModelInternal } from '@trezor/device-utils';
 import { CloneType, Static, Type } from '@trezor/schema-utils';
-
-export enum DeviceModelInternal {
```
martykan: This is an auto-generated file, to do this the scripts would need updating

packages/product-components/package.json
```diff
@@ -19,8 +19,8 @@
         "@suite-common/validators": "workspace:*",
         "@suite-common/wallet-config": "workspace:*",
         "@suite-common/wallet-utils": "workspace:*",
+        "@trezor/blockchain-link-types": "workspace:*",
```
martykan, line 22: `@trezor/device-utils` should be here too right?

packages/protobuf/src/messages-schema.ts
mroz22: yep, I missed that

mroz22: and I can't say I like it very much making this types generation script even more complex :(

mroz22: so, I didn't want to deal with it right now so I at least replaced connect with protobuf dependency. This is slightly better IMHO but still should be improved

mroz22: now moved to device-utils

packages/protobuf/messages.json
```diff
@@ -5577,6 +5577,96 @@
                 }
             }
         },
+        "NostrGetPubkey": {
```
martykan: We probably don't want to update this here

mroz22: removed

martykan (approved):

PR #17178, opened 2025-02-24
rework udp api implementation

mroz22:
after updating to the latest emulator, our transport e2e tests focusing on the node-bridge - emulator implementation started to fail consistently, namely test that is testing /post and /read endpoints.

After some investigation I found that the issue was caused by emulator responding before /read endpoint was processed which caused data loss and /read hang.

Now udp API is always listening to incoming data and /read can pull them out of a local buffer.


szymonlesisz (approved):
looks legit

PR #17188, opened 2025-02-24
Fix condition for enabling device authenticity check

### LINKED ISSUE #17177: Its not possible to reenable device authenticity check for Safe models after it was disabled
**Describe the bug**
Its not possible to reenable device authenticity check for Safe models after it was disabled.
Check is enabled by default and its enabled after app storage reset

**Steps to reproduce:**
1. Disable `Turn off device check` 
2. Try to enable it
3. Observe bahvior of app

**Info:**
 - Suite version: desktop 25.3.0 (669cfe48f7e295810d0abdcc96821c7dc58333d9)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.3.0 Chrome/132.0.6834.194 Electron/34.1.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T2B1 2.8.4 regular (revision ab96ce7954c7db384b26bc863d8752150027adbf)
 - Transport: NodeUsbTransport

**Screenshots:**

https://github.com/user-attachments/assets/146c43ef-5977-45e1-92d8-3314a24c782e

komret:
I broke it [here](https://github.com/trezor/trezor-suite/pull/16711) while refactoring :(

Resolves https://github.com/trezor/trezor-suite/issues/17177

### DIFF
--- a/packages/suite/src/actions/suite/suiteActions.ts
+++ b/packages/suite/src/actions/suite/suiteActions.ts
@@ -298,7 +298,7 @@ export const toggleDeviceAuthenticityCheck = (enable: boolean) => (dispatch: Dis
 
     dispatch({
         type: SUITE.TOGGLE_DEVICE_AUTHENTICITY_CHECK,
-        enable,
+        payload: enable,
     });
 };
 
--- a/packages/suite/src/views/settings/SettingsDevice/DeviceAuthenticityOptOut.tsx
+++ b/packages/suite/src/views/settings/SettingsDevice/DeviceAuthenticityOptOut.tsx
@@ -20,7 +20,7 @@ export const DeviceAuthenticityOptOut = () => {
         dispatch(
             isDeviceAuthenticityCheckEnabled
                 ? openModal({ type: 'device-authenticity-check-opt-out' })
-                : toggleDeviceAuthenticityCheck(false),
+                : toggleDeviceAuthenticityCheck(true),
         );
 
     return (


packages/suite/src/actions/suite/suiteActions.ts
```diff
@@ -298,7 +298,7 @@ export const toggleDeviceAuthenticityCheck = (enable: boolean) => (dispatch: Dis
 
     dispatch({
         type: SUITE.TOGGLE_DEVICE_AUTHENTICITY_CHECK,
-        enable,
+        payload: enable,
```
peter-sanderson, line 301: If this was a mistake, I wonder, why did not get the type-error?

peter-sanderson: I am checking and it seems the action types are not working correctly i guess. I must admit I do not understand the typing that is in the `Dispatch` interface fully.. so probably lets leave it as it is.

peter-sanderson (approved):

packages/suite/src/actions/suite/suiteActions.ts
komret: I couldn't think of anything better than e71c6b2735b59010d5d7d8b0e6170622f0429250 WDYT?

komret: I removed it to be solved in another PR.

PR #17212, opened 2025-02-25
Fix/apply-global-patches-after-focus-install

matejkriz:
## Description

Packages are automatically patched by postinstall script only if `yarn install` is called but not in case of yarn workspaces focus. 

Unfortunately, this cannot be applied globally after every `yarn workspaces focus` because it’s not possible to apply a patch to a package that hasn’t been installed, which results in an error. Therefore, as a safer option for all releases, we run a full yarn install, while for tests and dev builds, we keep the faster focus install.

### DIFF
--- a/.github/actions/release-connect/action.yml
+++ b/.github/actions/release-connect/action.yml
@@ -58,7 +58,7 @@ runs:
       shell: bash
       run: |
         echo -e "\nenableScripts: false" >> .yarnrc.yml
-        yarn workspaces focus @trezor/connect-iframe @trezor/connect-web @trezor/connect-popup @trezor/connect-webextension @trezor/connect-explorer-theme @trezor/connect-explorer
+        yarn install --immutable # focus install is not used here to make sure that all global patches are applied for the release.
 
     - name: Build connect-web
       shell: bash
--- a/.github/workflows/release-suite-native-production.yml
+++ b/.github/workflows/release-suite-native-production.yml
@@ -37,7 +37,9 @@ jobs:
           eas-version: 14.7.1
           token: ${{ secrets.EXPO_TOKEN }}
       - name: Install libs
-        run: yarn workspaces focus @suite-native/app
+        run: |
+          yarn workspaces focus @suite-native/app
+          npx patch-package # Apply global patches.
       - name: Build on EAS iOS
         run: eas build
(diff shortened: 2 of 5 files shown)


vdovhanych (commented):
should this also be updated then? 

https://github.com/trezor/trezor-suite/blob/a1198ab274710557e5a3c30c5c3d1bb051deeca7/.github/workflows/release-suite-desktop-web-staging.yml#L172

matejkriz:
> should this also be updated then?
> 
> https://github.com/trezor/trezor-suite/blob/a1198ab274710557e5a3c30c5c3d1bb051deeca7/.github/workflows/release-suite-desktop-web-staging.yml#L172

It is staging release but that build is then used also for production directly, right? So yes, I will change it there too stay on a safe side.

1006e6a5916839c2cab2f37564c31f1226af9e05

vdovhanych (approved):

.github/actions/release-connect/action.yml
```diff
@@ -58,7 +58,7 @@ runs:
       shell: bash
       run: |
         echo -e "\nenableScripts: false" >> .yarnrc.yml
-        yarn workspaces focus @trezor/connect-iframe @trezor/connect-web @trezor/connect-popup @trezor/connect-webextension @trezor/connect-explorer-theme @trezor/connect-explorer
+        yarn install --immutable # focus install is not used here to make sure that all global patches are applied for the release.
```
karliatto, line 61: :+1:

karliatto (approved):

.github/workflows/release-suite-native-production.yml
```diff
@@ -37,7 +37,9 @@ jobs:
           eas-version: 14.7.1
           token: ${{ secrets.EXPO_TOKEN }}
       - name: Install libs
-        run: yarn workspaces focus @suite-native/app
+        run: |
+          yarn workspaces focus @suite-native/app
+          npx patch-package # Apply global patches.
```
Nodonisko, line 42: maybe we can add this as postinstall to @suite-native/app

Nodonisko (approved):

matejkriz:
I'm pretty sure that this can't break the failing tests, so merging.

PR #17220, opened 2025-02-25
chore: upgrade Sentry

Nodonisko:
## Related Issue

Resolve


suite-common/sentry/src/index.ts
```diff
@@ -89,9 +89,8 @@ const ignoreErrors = [
     'device disconnected during action', // the same as with 'other call in progress'
 ];
 
-export const SENTRY_CONFIG: Options = {
+export const SENTRY_CONFIG = {
     dsn: 'https://6d91ca6e6a5d4de7b47989455858b5f6@o117836.ingest.sentry.io/5193825',
-    autoSessionTracking: false, // do not send analytical data to Sentry
     integrations: [
```
matejkriz: To preserve the original behavior with session tracking disabled, you need to remove the default integration `browserSessionIntegration`, in my opinion. https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/#removing-a-default-integration

https://docs.sentry.io/platforms/javascript/migration/v8-to-v9/#removed-apis

Nodonisko: Good point I was thinking it disabled by default.

suite-native/app/src/SentryProvider.tsx
```diff
@@ -15,11 +15,12 @@ const initSentry = () => {
...
         tracesSampleRate: 0.5,
         environment: isDetoxTestBuild() ? 'test' : getEnv(),
-        integrations: [
-            captureConsoleIntegration({
-                levels: ['error'],
-            }),
+        integrations: integrations => [
+            ...integrations,
+            // There seems to be no official way how to add captureConsoleIntegration to Sentry in React Native.
+            captureConsoleIntegration({ levels: ['error'] }) as any,
```
Nodonisko, line 21: I am not happy about casting to `any` but I was not able to find any other way how to it.

suite-native/state/package.json
```diff
@@ -11,7 +11,7 @@
     },
     "dependencies": {
         "@reduxjs/toolkit": "1.9.5",
-        "@sentry/react-native": "6.5.0",
+        "@sentry/react-native": "^6.9.0",
```
Nodonisko, line 14: We need to wait until version 7 comes out because v6 is not compatible with v9 core.

Issue #15390, opened 2024-11-14
🔝 Bluetooth Desktop


komret:
👉   [Figma Design](https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=0-1&p=f&t=JqFYWKXyBvP3R0Ja-0)
👉 [Notion Test Cases](https://www.notion.so/satoshilabs/Bluetooth-1a7dc526060680cfba06d28f572f00c5)

## Implementation
- My branch for dev: https://github.com/trezor/trezor-suite/compare/feat/bt-thp...suite_thp

### Connection / Pairing
- [x]  [Happy Path from Figma](https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=5922-43475&t=goQTSxY66NLYK53r-1) (generally  `done` 🟢  )
- [x] No Trezor Found (seems to me it is in generally  `done` 🟢 ) [Figma](https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=5940-22322&t=A6jCM5YoHS7VaTZh-1)
- [x] Bluetooth Adapter On/Off (generally `done` 🟢 ) [Figma](https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=5839-42679&t=A6jCM5YoHS7VaTZh-1)

### Known Device
 - [ ] Connection of the Known Device -> automatically goes to dashboard ( 🟠  implemented but bugged 🐞  ). [Figma](https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=5886-49585&t=A6jCM5YoHS7VaTZh-1)

### Settings
- [x] Debug Bluetooth disabled/enabled for Devs (`done` 🟢 )
- [ ] Forget/Unpair device in settings (technically also known as "Erase Bonds" working but unfinished 🟠  ). [Figma](https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=5897-132518&t=A6jCM5YoHS7VaTZh-1)
    - [ ] ❓  What happens when user did NOT removed the device from the OS?
- [ ] Remembered wallet, device not connected -> add link to connect flow in Device settings

----------------

## Other Notes
- [Mobile Issue](https://github.com/trezor/trezor-suite/issues/11751)

peter-sanderson:
[...]

PR #17221, opened 2025-02-25
feat: Bluetooth handle connection status for knownDevices

peter-sanderson:
Partially addresses: https://github.com/trezor/trezor-suite/issues/15390

Allows for disconnect:

image

Add cancel for paring mode:
image

### DIFF
--- a/packages/suite/src/actions/bluetooth/bluetoothDisconnectDeviceThunk.ts
+++ b/packages/suite/src/actions/bluetooth/bluetoothDisconnectDeviceThunk.ts
@@ -12,7 +12,7 @@ export const bluetoothDisconnectDeviceThunk = createThunk<
     { id: string },
     void
 >(
-    `${BLUETOOTH_PREFIX}/bluetoothConnectDeviceThunk`,
+    `${BLUETOOTH_PREFIX}/bluetoothDisconnectDeviceThunk`,
     async ({ id }, { fulfillWithValue, dispatch }) => {
         console.log('_____calling: bluetoothIpc.disconnectDevice(id)', id);
         const result = await bluetoothIpc.disconnectDevice(id);
--- a/suite-common/bluetooth/src/bluetoothReducer.ts
+++ b/suite-common/bluetooth/src/bluetoothReducer.ts
@@ -70,9 +70,6 @@ export const prepareBluetoothReducerCreator = <T extends BluetoothDeviceCommon>(
                         // which would produce a device with new id.
                         .filter(
                             nearbyDevice => nearbyDevice.connectionStatus?.type !== 'pairing-error',
-                        )
-                        .sort(
-                            (a, b) => b.lastUpdatedTimestamp - a.lastUpdatedTimestamp,
                         ) as Draft<T>[];
                 },
             )


packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -53,28 +48,23 @@ export const BluetoothConnect = ({ onClose, uiMode }: BluetoothConnectProps) =>
...
-
-        if (isDeviceUnresponsiveForTooLong) {
-            return knownDevices.find(knownDevice => knownDevice.id === it.device.id) !== undefined;
-        }
+    console.log('allDevices', allDevices);
 
-        return true;
-    });
+    const devices = allDevices;
+    //     .filter(it => {
```
peter-sanderson: This will be uncommented, I need a fix for this from Szymon

packages/suite/src/actions/bluetooth/bluetoothConnectDeviceThunk.ts
```diff
@@ -15,27 +15,15 @@ export const bluetoothConnectDeviceThunk = createThunk<
     `${BLUETOOTH_PREFIX}/bluetoothConnectDeviceThunk`,
     async ({ id }, { fulfillWithValue, dispatch }) => {
         const result = await bluetoothIpc.connectDevice(id);
+        console.log('_____calling: bluetoothIpc.connectDevice(id)', id);
 
         if (!result.success) {
-            dispatch(
-                bluetoothActions.connectDeviceEventAction({
```
peter-sanderson: change of the connection state is now responsibility of the `bluetoothIpc` and the connection status update will come from it as an update

suite-common/bluetooth/src/bluetoothSelectors.ts
```diff
@@ -25,21 +25,24 @@ export const prepareSelectAllDevices = <T extends BluetoothDeviceCommon>() =>
...
-            knownDevices.forEach(knownDevice => {
-                if (!map.has(knownDevice.id)) {
-                    map.set(knownDevice.id, { device: knownDevice, status: null });
-                }
-            });
+            const nearbyDevicesCopy = [...nearbyDevices];
+            nearbyDevicesCopy.reverse();
 
-            return Array.from(map.values()).sort(
-                (a, b) => b.device.lastUpdatedTimestamp - a.device.lastUpdatedTimestamp,
```
peter-sanderson: Sorting is now responsibility of the `bluetoothIpc` code

packages/suite/src/actions/bluetooth/bluetoothConnectDeviceThunk.ts
```diff
@@ -15,27 +15,15 @@ export const bluetoothConnectDeviceThunk = createThunk<
     `${BLUETOOTH_PREFIX}/bluetoothConnectDeviceThunk`,
     async ({ id }, { fulfillWithValue, dispatch }) => {
         const result = await bluetoothIpc.connectDevice(id);
+        console.log('_____calling: bluetoothIpc.connectDevice(id)', id);
```
Lemonexe: :memo:  reminder to remove console log.

Also, why remove `connectDeviceEventAction`? Is it guaranteed that when `bluetoothIpc.connectDevice` finishes, it will trigger `deviceActions.connectDevice` (so it'd be useless to call it twice), or how does it work?

suite-common/bluetooth/src/bluetoothSelectors.ts
```diff
@@ -25,21 +25,24 @@ export const prepareSelectAllDevices = <T extends BluetoothDeviceCommon>() =>
...
-            );
+            nearbyDevicesCopy
+                // Devices with 'pairing-error' status should NOT be displayed in the list, as it
+                // won't be possible to connect to them ever again. User has to start pairing again,
+                // which would produce a device with new id.
+                .filter(nearbyDevice => nearbyDevice.connectionStatus?.type !== 'pairing-error')
```
Lemonexe: Why do we persist them in redux if they cannot be connected ever again, and have to be paired again to (creating new device object)?

If any device's state becomes `'pairing-error'`, does it make any sense to keep it in redux for the rest of Suite session?

If not, let's filter them right at the source, in bluetoothReducer: case for `bluetoothActions.nearbyDevicesUpdateAction`.

Else, how about filtering them in storageActions: `saveKnownDevices`, so that at least we don't remember them forever?

suite-common/bluetooth/tests/bluetoothSelectors.test.ts
```diff
@@ -1,48 +1,86 @@
...
+    id: 'C',
+    data: [],
+    name: 'Trezor C',
+    lastUpdatedTimestamp: 3,
+    connectionStatus: { type: 'pairing' },
 };
 
 describe('bluetoothSelectors', () => {
-    it('selects knownDevices and nearbyDevices in one list fot the UI', () => {
+    it('selects knownDevices and nearbyDevices in one list fot the UI, all known devices are', () => {
```
Lemonexe: unfinished sentence?

suite-common/bluetooth/src/bluetoothSelectors.ts
```diff
@@ -25,21 +25,24 @@ export const prepareSelectAllDevices = <T extends BluetoothDeviceCommon>() =>
...
-            });
+            knownDevices.forEach(knownDevice => map.set(knownDevice.id, knownDevice));
 
-            knownDevices.forEach(knownDevice => {
-                if (!map.has(knownDevice.id)) {
-                    map.set(knownDevice.id, { device: knownDevice, status: null });
-                }
-            });
+            const nearbyDevicesCopy = [...nearbyDevices];
+            nearbyDevicesCopy.reverse();
```
Lemonexe: why reverse? So that device discovered most recently is displayed first?
Please add comment; it's clear what it does, but not why.

packages/suite/src/actions/bluetooth/bluetoothDisconnectDeviceThunk.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const bluetoothDisconnectDeviceThunk = createThunk<
+    BluetoothDisconnectDeviceThunkResult,
+    { id: string },
+    void
+>(
+    `${BLUETOOTH_PREFIX}/bluetoothConnectDeviceThunk`,
+    async ({ id }, { fulfillWithValue, dispatch }) => {
+        console.log('_____calling: bluetoothIpc.disconnectDevice(id)', id);
```
Lemonexe, line 17: same as in `bluetoothConnectDeviceThunk.ts`

packages/suite/src/components/suite/bluetooth/BluetoothDeviceItem.tsx
```diff
@@ -1,25 +1,98 @@
...
+        }
+
+        if (!result.success) {
+            onError();
+        }
+    };
+
+    const onDisconnect = async () => {
+        await dispatch(bluetoothDisconnectDeviceThunk({ id: device.id })).unwrap();
+    };
```
Lemonexe: this thunk can also result in `!result.success` right? So `onError` should be called?

packages/suite/src/components/suite/bluetooth/BluetoothDeviceItem.tsx
```diff
@@ -1,25 +1,98 @@
...
+    const onClickMap: Record<
+        DeviceBluetoothConnectionStatusType,
+        (() => Promise<void>) | undefined
+    > = {
+        'connection-error': onConnect,
+        'pairing-error': undefined,
+        connected: onDisconnect,
+        connecting: undefined,
+        disconnected: onConnect,
+        paired: undefined,
```
Lemonexe: when `paired`, the Button is neither disabled, nor loading, but doesn't do anything, is that right?

packages/suite/src/components/suite/bluetooth/BluetoothDeviceComponent.tsx
```diff
@@ -1,10 +1,29 @@
...
+
+        return () => clearInterval(interval);
+    }, [timestamp]);
+
+    return (
+        <>
+            <Text variant="warning">{secAgo}</Text>s ago
+        </>
+    );
+};
```
Lemonexe: This could be extracted as a reusable component..

Lemonexe (commented):
Looks ok, though I cannot test (no BT device), which means I also did not check id design matches figma.

Codewise LGTM with some comments to be addressed.

Overall, I don't understand why the refactoring from deviceState to device with connectionStatus is necessary, though it certainly looks good

suite-common/bluetooth/src/bluetoothSelectors.ts
peter-sanderson: Yes, currently it is supposed to be first. We may change it but for now its first.

packages/suite/src/actions/bluetooth/bluetoothConnectDeviceThunk.ts
peter-sanderson: Yes once `bluetoothIpc.connectDevice` the device is picked up by Conennect and automagically discovered.

:spiral_notepad: logs will be removed much later (but before merging to `develop`) I use them a lot

packages/suite/src/components/suite/bluetooth/BluetoothDeviceComponent.tsx
peter-sanderson: Yes, but currently it is actually just for debug. It is not in the designs so I assume we may be removeing it.

packages/suite/src/components/suite/bluetooth/BluetoothDeviceItem.tsx
peter-sanderson: fixed

packages/suite/src/components/suite/bluetooth/BluetoothDeviceItem.tsx
peter-sanderson: fixed

suite-common/bluetooth/src/bluetoothSelectors.ts
peter-sanderson: Known devices are only those, that were connected over the Connect (we got Features, discovery, ....)

So no Pairing Error shall ever happen for the known device.

-----


But for newrby devices, it is true that they can be filtered out at entry level. Good point!

suite-common/bluetooth/tests/bluetoothSelectors.test.ts
peter-sanderson: fixed

PR #17223, opened 2025-02-25
Don't cancel preventively on T1

marekrjpolak:
## Description

Older T1s (with bootloader <= 1.8.0) in bootloader don't respond at all to preventive Cancel message, which prevents installing firmware on those devices.

commit ⍺: send the preventive Cancel only when we're sure that it's not T1 in bootloader (= only webusb)

commit β: send the preventive Cancel whenever it's possible that it's not T1 in bootloader (= both webusb and bridge but false negatives may appear) and resolve potentially hanging Cancel by calling Acquire after some timeout (which causes Cancel to throw)


mroz22 (approved):
looks good, more defensive, more like it was before. :shipit:

mroz22 (approved):

STew790:
**QA OK**
I could detect a device with 1.6.1 FW in bl mode.

**Info:**
 - Suite version: web 25.3.0 (57dd8df74bd0c9cf4abd0f0b04eb5d786e53cd57)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 3440x1440
 - Device: Trezor T3W1 2.8.9 regular (revision 587b5e8eb867467b7fac1dae46e0f53c8619c8f1)

PR #17234, opened 2025-02-25
fix(suite-desktop-core): proper name & place for electron-builder hooks

Lemonexe:
## Description

Fix `suite-desktop-core/tsconfig.lib.json`, because `yarn update-project-references` is unstable; produces diff that is not detected by `yarn verify-project-references`, so it passed CI when I merged it in #17056 :open_mouth: 

- The diff is in [#17231](https://github.com/trezor/trezor-suite/pull/17231), but on second thoughts, commiting the diff is not right.
- In our repo, `yarn build:lib` is always used for building _"the main source code"_ of the given package.
- But in this case, I want to build just the `suite-desktop-core/scripts/` dir
  - electron-builder scripts are totally separate from `src/`, they don't use any of the workspace imports
- so it doesn't make sense call it `build:lib` → rename to `build:scripts`
  - → does not get picked up by `yarn update-project-references`, which is correct, because it's _a custom task_
- also, if I add all the workspace references, it slows down `yarn build:scripts` from 5s to 15s on my PC :snail: :turtle: :sleeping:  

## QA 

Tested that desktop build works from scratch (all gitignored dirs deleted in `suite-desktop`) :heavy_check_mark: 

## Notes

The tsconfig.lib.json passed CI, because `yarn verify-project-references` [throws only on tsconfig.json](https://github.com/trezor/trezor-suite/blob/70e6c68ac4f89ee0fac5c3046e5e85f0df484f4b/scripts/updateProjectReferences.ts#L93-L109),
but `yarn update-project-references` modifies **also** [tsconfig.lib.json](https://github.com/trezor/trezor-suite/blob/70e6c68ac4f89ee0fac5c3046e5e85f0df484f4b/scripts/updateProjectReferences.ts#L118).

The script is inconsistent, `yarn verify-project-references` should throw if there are any changes that would be modified by `yarn update-project-references`.
That should be fixed, but outside of scope of this PR.

@coderabbitai ignore


packages/suite-desktop-core/package.json
```diff
@@ -7,8 +7,8 @@
     "homepage": "https://trezor.io/",
     "main": "src/app.ts",
     "scripts": {
-        "build:lib": "yarn g:rimraf ./lib && yarn g:tsc --build tsconfig.lib.json",
-        "build:core": "yarn build:lib && yarn g:rimraf dist && TS_NODE_PROJECT=\"tsconfig.json\" yarn webpack --config ./webpack/core.webpack.config.ts",
```
Lemonexe, line 11: :exclamation:  It was a mistake to include building electron-builder hooks in the `build:core`, because that's not related to electron-builder at all, it builds JS bundle for electron-main process :facepalm: 
Instead, I moved `build:scripts` where it logically belongs - as first step of `build:app:electron`

Ofc it worked incidentally, because `build:core` gets called before `build:app:electron`

komret (approved):

PR #17245, opened 2025-02-26
fix(suite-native): hide send for portfolio tracker

### LINKED ISSUE #16484: Send from dashboard

PeKne:
## Description
- hides global send button for portfolio tracker device

## Related Issue

Resolve #16484

## Screenshots:
Screenshot 2025-02-26 at 12 53 43 PM

### DIFF
--- a/suite-native/accounts/src/selectors.ts
+++ b/suite-native/accounts/src/selectors.ts
@@ -18,6 +18,7 @@ import {
     selectAccounts,
     selectCurrentFiatRates,
     selectIsAccountUtxoBased,
+    selectIsPortfolioTrackerDevice,
     selectPendingAccountAddresses,
     selectVisibleDeviceAccounts,
 } from '@suite-common/wallet-core';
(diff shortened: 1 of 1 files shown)


suite-native/accounts/src/selectors.ts
```diff
@@ -223,6 +224,10 @@ export const selectFreshAccountAddress = (
 };
 
 export const selectHasDeviceAnySendAvailableAccount = createMemoizedSelector(
-    [selectVisibleDeviceAccounts],
-    accounts => pipe(accounts, filterSendAvailableAccounts, A.isNotEmpty),
+    [selectIsPortfolioTrackerDevice, selectVisibleDeviceAccounts],
+    (isPortfolioTrackerDevice, accounts) => {
+        if (isPortfolioTrackerDevice) return false;
```
vytick, line 229: should this be true for view only devices in case they have SendAvailableAccounts ?

PeKne: Yes, I think so. If user tries to send from view only device, he as able to fill in the send form and then if he continues, the `Connect Trezor Screen` is displayed. But I will double check, just to make sure.

EDIT: Yes, it is the way I described above. (it happens [here](https://github.com/trezor/trezor-suite/blob/5b33c28f860e0b5a07c88e18f6204655ad3ec306/suite-native/module-send/src/components/SendFeesForm.tsx#L144))

vytick (approved):
lgtm 🚀

PR #17251, opened 2025-02-26
feat(suite): added search by native token to tx list

### LINKED ISSUE #16116: Search by token is available but by native token is not
**Describe the bug**

- I can search for USDT but I cannot search for BNB

**Info:**

-   Suite Version: 24.12.3

**How to reproduce**
1. Go to account with tokens
2. Try to search for native token

**Expected behavior**
It will filter txs where native token balance changed (do not forget about internal transfers + ignore FEE)

**Screenshots**

Screenshot 2024-12-29 at 17 32 31
Screenshot 2024-12-29 at 17 32 25

**Additional context**

- please respect displaySymbol

TomasBoda:
## Description

Support filtering of account transactions by native token symbol.

## Screenshots:

https://github.com/user-attachments/assets/23c7f86f-5aed-4ef0-ac74-0f3183f9221b

### DIFF
--- a/suite-common/wallet-utils/src/transactionUtils.ts
+++ b/suite-common/wallet-utils/src/transactionUtils.ts
@@ -939,7 +939,6 @@ export const simpleSearchTransactions = (
 
     // Find by native token
     const foundTxsForNativeToken = transactions.flatMap(transaction => {
-        // native token
         const displaySymbol = getNetworkDisplaySymbol(transaction.symbol);
         const hasMatchingNativeToken =
             displaySymbol.toLowerCase().includes(search.toLowerCase()) &&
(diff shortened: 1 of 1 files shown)


suite-common/wallet-utils/src/transactionUtils.ts
```diff
@@ -936,6 +937,22 @@ export const simpleSearchTransactions = (
...
+            transaction.amount !== '0';
+
+        if (hasMatchingNativeToken) {
+            return transaction.txid;
+        }
+
+        return [];
+    });
+    txsToSearch.push(...foundTxsForNativeToken);
+
```
tomasklim, line 955: When I look for `eth` there are also txs with token `USDT` because the name `Tether` contains `eth` 😀 

One option is to remove `token.name` from `isTokenTransferMatchesSearch` (with comment)
Another option is to when `search === displaySymbolName` do not do this token searching?

Please open this discussion also with Jirka. Thanks

TomasBoda: I think maybe instead of using `.includes(...)` we could do `.startsWith(...)`? I assume the users don't search for a substring inside the token name, rather the whole token name (that would solve the issue with `eth` being a substring of `Tether`).

cc @jirih-stsh

tomasklim: lets resolve it in april release after some bigger discussions, no prio

PR #17264, opened 2025-02-26
test(suite-native): receive flow e2e test added

PeKne:
## Description
- receive flow with on Trezor device confirmation e2e test added

### DIFF
--- a/suite-native/app/e2e/pageObjects/accountReceiveActions.ts
+++ b/suite-native/app/e2e/pageObjects/accountReceiveActions.ts
@@ -5,7 +5,7 @@ class AccountReceiveActions {
         await waitFor(element(by.id('@screen/ReceiveAccount'))).toBeVisible();
     }
 
-    async clickShowAddressButton() {
+    async tapShowAddressButton() {
         const showAddressButton = element(by.id('@receive/show-address-button'));
 
         await waitFor(showAddressButton).toBeVisible().withTimeout(30000);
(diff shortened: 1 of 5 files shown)


suite-native/app/e2e/pageObjects/accountReceiveActions.ts
```diff
@@ -0,0 +1,26 @@
+import { expect as detoxExpect } from 'detox';
+
+class accountReceiveActions {
```
yanascz:
```suggestion
class AccountReceiveActions {
```

PeKne: fixed 8e9669106a

suite-native/app/e2e/pageObjects/accountReceiveActions.ts
```diff
@@ -0,0 +1,26 @@
+import { expect as detoxExpect } from 'detox';
+
+class AccountReceiveActions {
+    async waitForScreen() {
+        await waitFor(element(by.id('@screen/ReceiveAccount'))).toBeVisible();
+    }
+
+    async clickShowAddressButton() {
```
yanascz: We use `tap`, `click` is semantically not correct. Also `tap` is used much more often in our code base.
```suggestion
    async tapShowAddressButton() {
```

yanascz (approved):
Nice!

PR #17269, opened 2025-02-26
test(connect-popup): stabilize tests

mroz22:
## Related Issue

Resolve

### DIFF
--- a/.github/workflows/template-connect-popup-test-params.yml
+++ b/.github/workflows/template-connect-popup-test-params.yml
@@ -55,8 +55,6 @@ jobs:
       - name: Run connect popup test
         env:
           URL: https://${{ inputs.serverHostname }}/connect/${{ steps.extract_branch.outputs.branch }}/
-          # skip settings page, this url is set at build time anyway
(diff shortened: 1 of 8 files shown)


packages/trezor-user-env-link/src/websocket-client.ts
```diff
@@ -97,7 +97,7 @@ export class WebsocketClient extends WebsocketClientBase<WebsocketClientEvents>
                 this.emit('firmwares', resp.firmwares);
             } else {
                 if (!resp.success) {
-                    throw new Error(`websocket_error_message: ${resp.error.message || resp.error}`);
+                    console.log(`websocket_error_message: ${resp.error.message || resp.error}`);
```
HajekOndrej: Are you sure about this? I am not sure it's a good idea to ignore the error like this...

mroz22: lets solve this in separartion, commit dropped and moved here https://github.com/trezor/trezor-suite/pull/17277

HajekOndrej (approved):

PR #17272, opened 2025-02-27
feat(suite): non-canonical bridge versions deprecation. phase 1

mroz22:
There could be standalone bridges other than 2.0.27, 2.0.33 or built-in node bridge in the wild. The idea is to deprecate these non-canonical bridge versions to reduce our debugging surface. 

Here are some figures from about 2 years ago. It would be nic to check current usage with the data team.

image

The plan is as follows:
- phase 1 - warning in suite with instructions how to get rid of standalone bridge 
- phase 2 - connect would return transport-error if deprecated version is detected. 

This PR executes phase 1. Any non-cannonical bridge gets a notification prompting user to get rid of it. 

image

image

image

image

cc @Hannsek and @MiroslavProchazka


Hannsek:
That's great. We should tell users how to uninstall the bridge in the tooltip. Even I don't know how to uninstall it.
Do you know @mroz22

mroz22:
I guess @Ondra-Zik-SL could put together some materials?

martykan:
Do we need the modal? Maybe it would be best to send them to KB where there can be a proper detailed article.

mroz22:
> Do we need the modal? Maybe it would be best to send them to KB where there can be a proper detailed article.

Dunno, product question (@Hannsek) my gutfeeling was that it would be nice to be more descriptive already in the app and that it might be useful once we will start forcing users to uninstall canonical bridges as well.

MiroslavProchazka:
Tagging our Suite @trezor/suite-engagement team as this PR affects SuiteBanners file

Hannsek:
Let's keep the modal and let's edit the KB article

suite-common/suite-config/src/routes.ts
```diff
@@ -42,6 +42,13 @@ export const routes = [
         isForegroundApp: true,
         params: modalAppParams,
     },
+    {
+        name: 'suite-bridge-deprecated',
+        pattern: '/bridge-bridge',
```
marekrjpolak, line 47: Should the pattern really be `bridge-bridge`?

marekrjpolak (approved):
Great job! Now I remember I did some tradeoffs because of `latestVersion` param of transports which is there no more, so if only I recall what was that so I can finally fix it.

suite-common/suite-config/src/routes.ts
mroz22: that feels like a mistake

bosomt:
**QA OK**

- tested on bridge 2.0.32 from TENV
- learn more opens modal
- learn more modal contains link to https://trezor.io/learn/a/what-is-trezor-bridge
- take me back will redirect you to dashboard


image

Screenshot 2025-03-21 at 13 52 50


**Info:**
 - Suite version: web 25.4.0 (9a1be4f05b1434f4e66fd0d40c3c864cffc97aad)
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T1B1 1.13.0 regular (revision 592590cf66a9b62dfeee7e4d2afb6e01005e5b2c)

PR #17282, opened 2025-02-27
use error.cause for more elaborate error handling in connect

mroz22:
If transport error occurs during entropy check, show error toast but do not mark entropy check as failed.

Resolves https://github.com/trezor/trezor-suite-private/issues/128

### DIFF
--- a/packages/connect/src/api/resetDevice.ts
+++ b/packages/connect/src/api/resetDevice.ts
@@ -1,5 +1,6 @@
 // origin: https://github.com/trezor/connect/blob/develop/src/js/core/methods/ResetDevice.js
 import { Assert } from '@trezor/schema-utils';
+import { ERRORS_WITHOUT_DEVICE_INTERACTION } from '@trezor/transport/src/errors-groups';
 import { getRandomInt } from '@trezor/utils';
 
 import { generateEntropy, verifyEntropy } from '../api/firmware/verifyEntropy';
--- a/packages/connect/src/api/resetDevice.ts
+++ b/packages/connect/src/api/resetDevice.ts
@@ -89,7 +89,9 @@ export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.Res
             xpubs,
         });
         if (res.error) {
-            await cancelPrompt(this.device);
+            await cancelPrompt(this.device).catch(() => {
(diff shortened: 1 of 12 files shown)


suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -37,7 +37,7 @@ export type DeviceReducerState = {
     devices: TrezorDevice[];
     selectedDevice?: TrezorDevice;
     deviceAuthenticity?: Record<string, StoredAuthenticateDeviceResult>;
-    devicesWithFailedEntropyCheck?: string[];
+    devicesWithFailedEntropyCheck?: (string | null)[];
```
marekrjpolak: What does it mean to have a `null` here? Can it be used as a device identifier?

packages/connect/src/api/resetDevice.ts
```diff
@@ -103,25 +105,33 @@ export default class ResetDevice extends AbstractMethod<'resetDevice', PROTO.Res
...
-                    throw ERRORS.TypedError('Failure_EntropyCheck', error.message);
                 }
+                // step 7 EntropyCheckContinue > Success
+                await cmd.typedCall('EntropyCheckContinue', 'Success', { finish: true });
+            }
+        } catch (error) {
+            // error.message should be one of these https://github.com/trezor/trezor-suite/blob/develop/packages/transport/src/transports/abstract.ts#L59
+            if (
+                error.cause === 'transport-error' &&
+                ERRORS_WITHOUT_DEVICE_INTERACTION.includes(error.message)
```
marekrjpolak, line 130: Are both parts of this condition necessary? It's not enough to know that this error originates from the `transport.call` call?

komret: @mroz22

suite-common/wallet-core/src/device/deviceReducer.ts
komret: Protobuf allows null values here and I don't want to do `if (device.id)` because a fake device could intentionally set `id` to null.

komret: Added comment explaining this.

marekrjpolak (approved):

PR #17291, opened 2025-02-27
Refactor/suite desktop core/e2e refactor emu press timeouts

Vere-Grey:
## Related Issue

Resolve

### DIFF
--- a/packages/trezor-user-env-link/src/api.ts
+++ b/packages/trezor-user-env-link/src/api.ts
@@ -79,6 +79,20 @@ interface ReadAndConfirmShamirMnemonicEmu {
 
 type StartBridgeVersion = '2.0.32' | '2.0.33' | 'node-bridge';
 
+// Currently, emulator can timeout on press action, suite does not have any way to know about it
+// Without this promiseWithTimeout, suite would fail on next step which complicates debugging
+export function promiseWithTimeout<T>(
+    promise: Promise<T>,
+    errorMessage: string,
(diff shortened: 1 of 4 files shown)


packages/trezor-user-env-link/src/api.ts
```diff
@@ -79,6 +79,20 @@ interface ReadAndConfirmShamirMnemonicEmu {
 
 type StartBridgeVersion = '2.0.32' | '2.0.33' | 'node-bridge';
 
+// Currently, emulator can timeout on press action, suite does not have any way to know about it
+// Without this promiseWithTimeout, suite would fail on next step which complicates debugging
+export function promiseWithTimeout<T>(
```
mroz22, line 84: scheduleAction

Vere-Grey: You mean something like this? It is not clear to me how to nicely pass my own Error throw
```
const params: ScheduleActionParams = {
    delay: 0,
    attempts: 1,
    timeout: 4500,
    deadline: undefined,
    gap: 0,
    attemptFailureHandler: (error) => {
        if (error.message === 'Aborted by timeout') {
            return new Error('Emulator did not respond to press command in time');
        }
        return error;
    },
};

async pressNo() {
    const action: ScheduledAction<null> = async () => {
        await this.client.send({ type: 'emulator-press-no' });
        return null;
    };

   await scheduleAction(action, params);

    return null;
}
```

PR #17316, opened 2025-02-28
feat(blockchain-link-type): change bigint to string in persist

vytick:
`BigInt`s in `SolanaStakingAccount` data caused persistence issues on mobile. Changed to `string`

### DIFF
--- a/packages/blockchain-link-types/src/solana.ts
+++ b/packages/blockchain-link-types/src/solana.ts
@@ -72,8 +72,8 @@ export type { Address } from '@solana/web3.js';
 
 export type SolanaStakingAccount = {
     status: string;
-    stake?: bigint;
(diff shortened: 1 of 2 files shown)


tomasklim (approved):
there is more `bigint`s in the codebase but probably not persisted LGTM

vytick:
QA: Please, test that when user has solana account with staking, is is persisted over killing and reopening the app

mroz22:
so we can't use bigint in general yet because of mobile?

vytick:
@mroz22 We can, but we need to somehow handle those that are persisted there.

STew790:
**QA OK**
Accounts on Solana are persisted through killing/reopening the app.

**Info**
25.3.1 ffe3145

PR #17317, opened 2025-02-28
feat(ci): refactor updateProjectReferences

Lemonexe:
## Description

As discovered in https://github.com/trezor/trezor-suite/pull/17234, the `updateProjectReferences.ts` script does not treat tsconfig.ts & tsconfig.lib.ts consistently.

Currently you can push incomplete `tsconfig.lib.json` to master; it will pass CI, because `yarn verify-project-references` throws only on `tsconfig.json`. But `yarn update-project-references` modifies **also** `tsconfig.lib.json`.
This is inconsistent, `yarn verify-project-references` should throw if there are any changes that would be modified by `yarn update-project-references`.

:arrow_right: fixed it, and reorganized the code while doing that; I hope it will be easier to read now :slightly_smiling_face: 

## QA

I tried:

- `yarn verify-project-references` passes  :heavy_check_mark: 
- `yarn refs` passes with no change :heavy_check_mark: 
- In some packages, editing either (`tsconfig.ts`, `tsconfig.lib.ts` or both), by either (adding line, removing line, swapping lines or removing references altogether) and then first running:
  - `yarn verify-project-references` to make sure it throws :heavy_check_mark:
  - then `yarn refs` to make sure it corrects to original state (no git diff) :heavy_check_mark:
- Removing `tsconfig.ts`:
  - `yarn verify-project-references` throws :heavy_check_mark:
  - `yarn refs` recreates it, but not necessarily to the same state (the defaultConfig has some different fields) :shrug: 
- Removing `tsconfig.lib.ts`: both scripts pass, no change :heavy_check_mark:

### DIFF
--- a/scripts/updateProjectReferences.ts
+++ b/scripts/updateProjectReferences.ts
@@ -35,6 +35,20 @@ import { getWorkspacesList } from './utils/getWorkspacesList';
         }
     };
 
+    const parseTSConfigFile = (configPath: string) => {
+        try {
+            return fs.existsSync(configPath)
+                ? JSON.parse(fs.readFileSync(configPath).toString())
+                : null;
+        } catch {
+            console.error(chalk.bold.red('Error while parsing file: '), configPath);
+            process.exit(1);
+        }
+    };
+
+    const isDiffInConfig = async (actualConfig: any[] = [], expectedConfig: any[] = []) =>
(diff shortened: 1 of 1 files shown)


Nodonisko (approved):
LGTM! Thanks for fixing it 👍 🚀

PR #17323, opened 2025-03-01
refactor(e2e): Rename fixtures and page objects

Vere-Grey:
Renaming of fixtures and page objects


Vere-Grey:
The failing test is something that is already problematic on develop. I am investigating it.

HajekOndrej (approved):

PR #17335, opened 2025-03-03
fix(suite-native): auto version bump

PeKne:
## Description
- passes token to `peter-evans/create-pull-request@v7` so it is able to create a PR

### DIFF
--- a/.github/workflows/suite-native-monthly-version-bump.yml
+++ b/.github/workflows/suite-native-monthly-version-bump.yml
@@ -1,6 +1,8 @@
 name: "[Bot] suite-native monthly version bump"
+
 permissions:
   pull-requests: write
+
 on:
   schedule:
     # Runs on the first day of every month at 00:00 UTC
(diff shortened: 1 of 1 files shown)


.github/workflows/suite-native-monthly-version-bump.yml
```diff
@@ -13,29 +15,27 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.TREZOR_BOT_TOKEN }}
```
vdovhanych: This is not needed here get rid of it.

.github/workflows/suite-native-monthly-version-bump.yml
```diff
@@ -13,29 +15,27 @@ jobs:
...
+          YEAR=TEST$(date +%y)
           MONTH=$(date +%-m)
+          NEW_VERSION="$YEAR.$MONTH.1"
           echo NEW_VERSION="$YEAR.$MONTH.1" >>  $GITHUB_ENV
+          echo NEW_BRANCH_NAME="chore/native/bump-version-to-$NEW_VERSION" >>  $GITHUB_ENV
 
       - name: Update Version of package.json
         run: jq --indent 4 --arg version "$NEW_VERSION" '.suiteNativeVersion = $version' suite-native/app/package.json > temp.json && mv temp.json suite-native/app/package.json
 
       - name: Create Pull Request
```
vdovhanych, line 31: Get rid of this custom setup and use https://github.com/peter-evans/create-pull-request#reference-example

cofigure it with

```
        uses: peter-evans/create-pull-request@v7
        with:
          token: ${{ secrets.TREZOR_BOT_TOKEN }}
```

vdovhanych (changes requested):
Please update based on coments

.github/workflows/suite-native-monthly-version-bump.yml
PeKne: removed

.github/workflows/suite-native-monthly-version-bump.yml
PeKne: fixed

vdovhanych (approved):
Looks good now

PR #17348, opened 2025-03-03
feat(scripts): add a script to generate UUID

Lemonexe:
## Description

Add a convenience script to generate UUID v4 for a new `message-system` entry

@coderabbitai ignore


komret (approved):

Lemonexe:
Closing, because it's unnecessary. `uuidgen` is available by default on macOS, and some Linux distros such as Ubuntu. Updated notion page instead to give a tip about it.

PR #17354, opened 2025-03-03
fix(e2e): Buy ethereum - remove invalid assert

Vere-Grey:
Functionality was changed but test not. This PR fixes that.

The warning label is now shown only if you fill invalid address or remove it. Not when you start the form with empty address.


STew790 (approved):
LGTM

PR #17368, opened 2025-03-04
Fix/suite dekstop core/e2e fix broken cardano and unstable wallets

Vere-Grey:
Cardano test got broken by Translation 
Wallet and passphrase test are sometime unable to detect discovery because the test is awaiting still promise of pressYes while suite finishes the discovery = that is an issue we will try to address in Tenv but in meantime this should stabilize tests

### DIFF
--- a/packages/suite-desktop-core/e2e/snapshots/web/wallet/cardano.test.ts/cardano-aria.ts
+++ b/packages/suite-desktop-core/e2e/snapshots/web/wallet/cardano.test.ts/cardano-aria.ts
@@ -17,7 +17,7 @@ export const cardanoAccountDetails = `
 `;
 
 export const cardanoStaking = `
-    - heading "Cardano Staking is NOT Active" [level=2]:
+    - heading "Cardano staking isn't active" [level=2]:
       - img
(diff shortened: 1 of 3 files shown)


packages/suite-desktop-core/e2e/snapshots/web/wallet/cardano.test.ts/cardano-aria.ts
```diff
@@ -17,7 +17,7 @@ export const cardanoAccountDetails = `
 `;
 
 export const cardanoStaking = `
-    - heading "Cardano Staking is NOT Active" [level=2]:
+    - heading "Cardano staking isn't active" [level=2]:
```
Vere-Grey, line 20: update according to transaltion

packages/suite-desktop-core/e2e/support/pageObjects/dashboardPage.ts
```diff
@@ -122,12 +130,20 @@ export class DashboardPage {
...
+    ) {
         await this.addHiddenWallet(passphrase, options);
-        await this.page
-            .getByTestId('@passphrase-confirmation/step1-open-unused-wallet-button')
-            .click({ timeout: 10_000 });
-        await this.page.getByTestId('@passphrase-confirmation/step2-button').click();
+        await expect(
+            this.openUnusedWalletButton1,
+            'Expected "Yes, open" button to be enabled. Unused wallet might not finished Discovery.',
+        ).toBeEnabled({ timeout: 30_000 });
```
Vere-Grey, line 144: Expecting the button enabled with longer time out replaces the original expect of the discovery
That is now disabled by the parameter skipDiscovery default to a true

HajekOndrej (approved):

PR #17378, opened 2025-03-04
fix(blockchain-link): limit loading signatures for too many solana tokens

### LINKED ISSUE #16841: Solana account discovery error for wallet with many tokens OR large history
**Describe the bug**
Solana account discovery fails when the wallet has a large number of solana tokens. The app is hitting rate limits on the RPC. It's doing an endless number of "getSignaturesForAddress" RPC calls for the tokens accounts. See screenshot below from the web app.

This wallet had around 400 tokens (mainly irrelevant spam token). I would put a max number of token account to load or skip getSignaturesForAddress for the token accounts (not sure why this is needed in the first place).

Behaviour is similar on the desktop app.

**Info:**

Remembered Wallet: yes
Suite Version: 25.1.2
Browser:  25.1.2
Browser Installation Method:downloaded installer
OS: MacBook
Bridge: N/a
Firmware Version: 

**How to reproduce**

Load up a wallet with many SPL token 

**Expected behavior**
Load the wallet

**Screenshots**
Image

Image

martykan:
## Description 

This PR introduces a quick fix to resolve discovery errors for users with a high number of token accounts. 
Instead of fetching transaction signatures for all tokens, we limit it to 100. 

The trade-off is that incoming transaction history may not be visible for tokens beyond the limit, however the token itself and its balance are still normally loaded. 

## Related Issue

Resolve #16841

### DIFF
--- a/packages/blockchain-link/src/workers/solana/index.ts
+++ b/packages/blockchain-link/src/workers/solana/index.ts
@@ -231,6 +231,8 @@ const getAccountInfo = async (
 
     const getAllTxIds = async (tokenAccountPubkeys: string[]) => {
         const sortedTokenAccountPubkeys = tokenAccountPubkeys.sort();
+        // limit to 100 accounts, loading too many would hit the rpc limit
+        sortedTokenAccountPubkeys.splice(100);
 
         const allAccounts = [payload.descriptor, ...sortedTokenAccountPubkeys];
 


packages/blockchain-link/src/workers/solana/index.ts
```diff
@@ -231,6 +231,8 @@ const getAccountInfo = async (
 
     const getAllTxIds = async (tokenAccountPubkeys: string[]) => {
         const sortedTokenAccountPubkeys = tokenAccountPubkeys.sort();
+        // limit to 100 accounts
```
vytick: can you please add reasoning why introducing this limit? 🙏

tomasklim (approved):

PR #17396, opened 2025-03-04
redux upgrade

Nodonisko:
## Description

This was real pain to do because lot of legacy stuff combined with new one and lot breaking changes in redux types.

Most of changes should TS only so I hope it won't break anything. I had to use few `any` and `unknown` and I am not proud of it, but I don't think it affects type safety so let's call it a day.

## Related Issue

Resolve


suite-common/token-definitions/package.json
```diff
@@ -14,7 +14,7 @@
     },
     "dependencies": {
         "@mobily/ts-belt": "^3.13.1",
-        "@reduxjs/toolkit": "^2.2.5",
```
Lemonexe, line 17: oof, someone did a mess here, but now we have a consistent version everywhere :+1:

suite-common/suite-utils/src/device.ts
```diff
@@ -442,3 +450,6 @@ export const isDeviceWithButtons = (
     deviceModel: DeviceModelInternal,
 ): deviceModel is DeviceModelInternal.T1B1 | DeviceModelInternal.T2B1 | DeviceModelInternal.T3B1 =>
     deviceModel.at(2) === 'B';
+
+export const isAnyDeviceEventAction = (action: AnyAction): action is DeviceEvent =>
+    Object.values(DEVICE).includes(action.type as DeviceEvent['type']);
```
Lemonexe: Nit: get rid of the `as`
```suggestion
    isArrayMember(action.type, Object.values(DEVICE));
```

suite-native/storage/src/typedPersistReducer.ts
```diff
@@ -34,5 +36,5 @@ export const preparePersistReducer = async <TReducerInitialState>({
         stateReconciler: (mergeLevel === 2 ? autoMergeLevel2 : autoMergeLevel1) as any,
     };
 
-    return persistReducer(persistConfig, reducer);
+    return persistReducer<TReducerInitialState, AnyAction>(persistConfig, reducer);
```
Lemonexe: is it necessary? Seems to me locally that inference works correctly without specifying the generics (as before)

packages/connect-explorer/src/store/index.ts
```diff
@@ -1,12 +1,12 @@
...
 import { createLogger } from 'redux-logger';
-import thunk from 'redux-thunk';
+import { thunk } from 'redux-thunk';
 
 import { trezorConnectMiddleware } from '../middlewares/trezorConnectMiddleware';
 import { reducers } from '../reducers';
 
 const enhancers: any[] = [];
-const middleware = [thunk, trezorConnectMiddleware];
+const middleware: any[] = [thunk, trezorConnectMiddleware];
```
Lemonexe: Figured it out to make it somewhat type-safer. Otherwise, wouldn't the `any` type corrupt the types downstream?

```suggestion
const middleware = [thunk, trezorConnectMiddleware] as Middleware<Dispatch, AppState>[];
```

This would be even better
```
const middleware: Middleware<Dispatch, AppState>[] = [thunk, trezorConnectMiddleware];
```

but it says this and I don't know how to solve that anymore..
```
Call signature return types (next: Dispatch) => (action: Action) => void and (next: (action: unknown) => unknown) => (action: unknown) => unknown are incompatible.
Type (action: Action) => void is not assignable to type (action: unknown) => unknown
Types of parameters action and action are incompatible.
Type unknown is not assignable to type Action
```

packages/suite/src/reducers/store.ts
```diff
@@ -100,13 +108,20 @@ export const initStore = (
 
     const patchedState =
         preloadedState && statePatch && patchConfirm(statePatch)
-            ? mergeDeepObject.withOptions({ dotNotation: true }, preloadedState, statePatch)
-            : preloadedState;
+            ? (mergeDeepObject.withOptions(
+                  { dotNotation: true },
+                  preloadedState,
+                  statePatch,
+              ) as unknown as AppState)
```
Lemonexe: `as unknown as AppState` just screams bad pattern. Although in this case it is ok, because `statePatch` is an external input for advanced users who want to pass custom object through process param.
Still, I think this would be slightly smoother :slightly_smiling_face:  
```suggestion
                  statePatch as Partial<AppState>,
              ))
```

packages/suite/src/reducers/store.ts
```diff
@@ -100,13 +108,20 @@ export const initStore = (
...
+            : preloadedState!;
 
     return configureStore({
-        reducer: rootReducer,
+        reducer: rootReducer as Reducer<Partial<AppState>>,
```
Lemonexe: Won't this have a suite-wide implication that store.state will now be `Partial<AppState>` instead of `AppState`?
If it is so, then a solution has to be found, though I did not find one.

Lemonexe (commented):
Reviewed all changes. It's unfortunate that a lot of typecasting is now necessary, so I took a stab at some of those. I did manage to find at least partial solution for several of them, see comments :speech_balloon:. But most of the times I did not :disappointed: 

Overall great work, thanks a lot for taking the time! :raised_hands: 
I still want to try something, then I'll give approve

And in `yarn.lock` I can see that `@reduxjs/toolkit`, `react-redux` and `redux-thunk` are deduplicated, which is great :tada:

packages/suite/src/reducers/store.ts
Lemonexe: oooh I found a solution, inspired by `suite-native/state/src/store.ts` but with lot of differences.
This makes sense to me. What do you think? 

```
type RootReducerShape = typeof rootReducer;
type PreloadedState = Partial<AppState>;
type InferredAction = Parameters<RootReducerShape>[1];

...

    return configureStore({
        reducer: rootReducer as Reducer<AppState, InferredAction, PreloadedState>,
        preloadedState: patchedState,
       ...
```

Lemonexe (approved):
Tested briefly.
:memo: ~When comments are resolved~ it looks good to go! :rocket: 

EDIT: comments can be resolved, I pushed 274c570 to implement them

packages/connect-explorer/src/store/index.ts
Nodonisko: I didn't know either that reason for `any[]` but I don't think it matter here it won't affect type safety in anyway becacause this type is used only once in configureStore

packages/suite/global.d.ts
```diff
@@ -1,3 +1,61 @@
...
+    interface Window {
+        __REDUX_DEVTOOLS_EXTENSION_COMPOSE__?: typeof compose;
+    }
 }
+
+declare module 'redux' {
+    /**
+     * This is copied from older version of the redux-thunk package
+     * https://github.com/reduxjs/redux-thunk/blob/698241dec398639018d5f21c529b0889b78d36f8/extend-redux.d.ts
+     * Because suite still use that odl bindActionCreators function we need to keep it, once we get rid of that we can remove this
```
tomasklim, line 16: it should not be that hard to refactor it, it is on just few places

Lemonexe: Sorry, noticed the comment only now :grimacing: 
If you know what should be done to get rid of the declaration, could you do a followup PR?

Lemonexe:
Sentry issues not relevant at all;
those are testing issues sent during development of #17487

PR #17400, opened 2025-03-04
fix: Solana send max include rent

### LINKED ISSUE #17340: Solana send max button no error but error when no max with same amount
**Describe the bug**

https://github.com/user-attachments/assets/a614c48e-e1b1-4e0b-aaa3-fe7dcab20fea

gajdusep:
## Description

Solana send max value didn't include rent. Maximum amount to be sent now calculated as `balance - rent - fee`

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17340

### DIFF
--- a/suite-common/wallet-core/src/send/sendFormSolanaThunks.ts
+++ b/suite-common/wallet-core/src/send/sendFormSolanaThunks.ts
@@ -45,7 +45,8 @@ const calculate = (
         ? amountToSmallestUnit(token.balance!, token.decimals)
         : undefined;
     if (output.type === 'send-max' || output.type === 'send-max-noaddress') {
-        max = availableTokenBalance || calculateMax(availableBalance, feeInLamports);
+        const feePlusRentInLamports = BigNumber(feeInLamports).plus(rent).toString()
+        max = availableTokenBalance || calculateMax(availableBalance, feePlusRentInLamports);
         amount = max;
     } else {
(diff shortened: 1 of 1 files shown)


tomasklim (changes requested):
Thank you for contribution! Looks like `const feePlusRentInLamports = BigNumber(feeInLamports).plus(rent).toString()` solves the problem. 

However, in this PR there is a lot of unnecessary formatting changes.

Any idea why there is this error in console when max is hit? It is not there for the same amount when the amount is filled manually
Screenshot 2025-03-11 at 15 29 18

gajdusep:
> Thank you for contribution! Looks like `const feePlusRentInLamports = BigNumber(feeInLamports).plus(rent).toString()` solves the problem.
>
> However, in this PR there is a lot of unnecessary formatting changes.

I agree with that. I thought about the same thing, but I think that's a larger problem of how to deal with BigNumbers. 
For example I would prefer if the calculateMax function would be getting BigNumbers as parameters, but I don't know what's the overall strategy.
Let me know and I can change it.
 
> Any idea why there is this error in console when max is hit? It is not there for the same amount when the amount is filled manually Screenshot 2025-03-11 at 15 29 18

If I'm not wrong, the answer for that comes from backend. When I was testing it, I was getting the same error even when filling in manually. Could it be that there are too many requests in a short time and they are blocked?

Issue #17302, opened 2025-02-28
🔝 THP: Trezor Host Protocol (THP) into Suite


peter-sanderson:
- Figma: https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=5909-44867&t=1NoHl0Xzoz9iuDSv-0

My branch for dev: https://github.com/trezor/trezor-suite/compare/feat/bt-thp...suite_thp

---------
### PRs:
- https://github.com/trezor/trezor-suite/pull/17413
- https://github.com/trezor/trezor-suite/pull/17955
- https://github.com/trezor/trezor-suite/pull/18466
- https://github.com/trezor/trezor-suite/pull/18496
 https://github.com/trezor/trezor-suite/pull/19038

komret:
Closing in favour of specific issues.

PR #17413, opened 2025-03-05
feat(components): implement PinInput component

peter-sanderson:
Implements a new component for entering a PIN like code

Part of: https://github.com/trezor/trezor-suite/issues/17302

**Supports:**
- in-place overwrite of the symbol
- tab / shift tab focusing of elements
- arrows navigation
- backspace and delete operation (keys)
- pasting
- only numbers (can be easily extended)

image
image


jvaclavik (approved):
perfecto 👌

Issue #17238, opened 2025-02-26
Transactions do not use explorer from networksConfig


tomasklim:
`Base` txs go to blockbook and not basescan

tomasklim:
Maybe can be connected with [#3211](https://github.com/trezor/trezor-suite/issues/3211)

Ondra-Zik:
**QA OK**

Basescan is now default explorer for Base.

**Info:**
 - Suite version: web 25.5.0 (4f38513a329b926cba15e6e79d309c3fc00fe142)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 2880x1800
 - Device: Trezor T3B1 2.8.10 regular (revision 2a65d18200580005dc419b9569ed97fae440806a)
 - Transport: WebUsbTransport

PR #17420, opened 2025-03-05
feat(trading): tx id and collapsible box for tx data

### LINKED ISSUE #16466: Trading - tx id missing missing in swap final step
In approval step, we show tx id and link to blockchain explorer
Image

In case of the final swapping step, we do not

Image

TomasBoda:
## Description

- added tx id with link to blockbook and a copy button to final swap step
- added a collapsible box to approval tx data to second swap step
- added a collapsible box for swap tx data to third swap step

## Related Issue

Resolve #16466 

## Screenshots:

Screenshot 2025-03-07 at 11 14 34

Screenshot 2025-03-07 at 11 15 31

Screenshot 2025-03-05 at 14 31 56

Screenshot 2025-03-05 at 14 28 31

### DIFF
--- a/packages/suite/src/views/wallet/trading/common/TradingDetail/TradingDetailExchange/TradingDetailExchange.tsx
+++ b/packages/suite/src/views/wallet/trading/common/TradingDetail/TradingDetailExchange/TradingDetailExchange.tsx
@@ -1,9 +1,11 @@
 import styled from 'styled-components';
 
-import type { TradingExchangeType } from '@suite-common/trading';
-import { Card } from '@trezor/components';
+import { type TradingExchangeType, cryptoIdToNetwork } from '@suite-common/trading';
+import { Card, InfoItem } from '@trezor/components';
 
 import { goto } from 'src/actions/suite/routerActions';
+import { Translation } from 'src/components/suite';
+import { IOAddress } from 'src/components/suite/copy/IOAddress';
 import { useDispatch } from 'src/hooks/suite';
 import { useTradingDetailContext } from 'src/hooks/wallet/trading/useTradingDetail';
 import { tradeFinalStatuses } from 'src/hooks/wallet/trading/useTradingWatchTrade';
@@ -40,6 +42,9 @@ export const TradingDetailExchange = () => {
         return null;
     }
 
+    const { receiveTxHash, send } = trade.data;
+    const network = send && cryptoIdToNetwork(send);
+
     const tradeStatus = trade?.data?.status || 'CONFIRMING';
     const exchangeTradeFinalStatuses = tradeFinalStatuses['exchange'];
     const showSending =
@@ -61,6 +66,16 @@ export const TradingDetailExchange = () => {
     return (
         <Wrapper>
             <Card>
+                {receiveTxHash && (
+                    <InfoItem label={<Translation id="TR_TXID" />}>
+                        <IOAddress
+                            txAddress={receiveTxHash}
+                            explorerUrl={network?.explorer.tx}
+                            explorerUrlQueryString={network?.explorer.queryString}
+                        />
+                    </InfoItem>
+                )}
+
                 {tradeStatus === 'SUCCESS' && (
                     <TradingDetailExchangePaymentSuccessful account={account} />
                 )}
--- a/packages/suite/src/views/wallet/trading/common/TradingSelectedOffer/TradingOfferExchange/TradingOfferExchangeSendApproval.tsx
+++ b/packages/suite/src/views/wallet/trading/common/TradingSelectedOffer/TradingOfferExchange/TradingOfferExchangeSendApproval.tsx
@@ -9,6 +9,7 @@ import {
     parseCryptoId,
     useTradingInfo,
 } from '@suite-common/trading';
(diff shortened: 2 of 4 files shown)


packages/suite/src/views/wallet/trading/common/TradingDetail/TradingDetailExchange/TradingDetailExchange.tsx
```diff
@@ -40,6 +44,11 @@ export const TradingDetailExchange = () => {
         return null;
     }
 
+    const { approvalSendTxHash, send } = trade.data;
```
adderpositive: use `receiveTxHash`, `approvalSendTxHash` is only for approval tx

adderpositive (changes requested):

packages/suite/src/views/wallet/trading/common/TradingDetail/TradingDetailExchange/TradingDetailExchange.tsx
tomasklim: yup and different title

tomasklim (changes requested):
Would be nice to have it also here. I am not sure if anchor into Trezor transaction list or link to explorer Screenshot 2025-03-06 at 9 43 17

adderpositive:
> Would be nice to have it also here. I am not sure if anchor into Trezor transaction list or link to explorer Screenshot 2025-03-06 at 9 43 17

According to the code, it is okay. There is a wrong parameter `approvalSendTxHash`, which is why it isn't displayed. I guess

packages/suite/src/views/wallet/trading/common/TradingDetail/TradingDetailExchange/TradingDetailExchange.tsx
TomasBoda: changed to `receiveTxHash` and added title `Transaction ID`

TomasBoda:
> Would be nice to have it also here. I am not sure if anchor into Trezor transaction list or link to explorer Screenshot 2025-03-06 at 9 43 17

Yes, it also displays itself there. I added a link to explorer

adderpositive (approved):
Use fixups next time, thanks. 👌 by me.

TomasBoda:
Fixed `approval tx id` and `tx id` URL to go to the correct explorer in this commit https://github.com/trezor/trezor-suite/pull/17420/commits/0fe52297b578fc0aa16bd72d3e3b22d97347d9eb.

@tomasklim is this what you meant by this issue https://github.com/trezor/trezor-suite/issues/17238? fix for other places with incorrect explorer in PR https://github.com/trezor/trezor-suite/pull/17472.

packages/suite/src/views/wallet/trading/common/TradingDetail/TradingDetailExchange/TradingDetailExchange.tsx
```diff
@@ -40,6 +43,11 @@ export const TradingDetailExchange = () => {
         return null;
     }
 
+    const { receiveTxHash, send } = trade.data;
+
+    const symbol = send && cryptoIdToSymbol(send);
+    const network = symbol && getNetwork(symbol);
```
tomasklim: is this really ok? @adderpositive

packages/suite/src/support/messages.ts
```diff
@@ -573,6 +573,10 @@ export default defineMessages({
         defaultMessage: 'Proceed to swap',
         id: 'TR_EXCHANGE_APPROVAL_TO_SWAP_BUTTON',
     },
+    TR_EXCHANGE_TXID: {
+        defaultMessage: 'Transaction ID',
```
tomasklim: we already have `TR_TXID`

packages/suite/src/views/wallet/trading/common/TradingDetail/TradingDetailExchange/TradingDetailExchange.tsx
adderpositive: Yes, it looks good. But slight improvement can be used

`const network = send && cryptoIdToNetwork(send)` @TomasBoda

TomasBoda: nice nitpick, thanks

tomasklim (changes requested):
Maybe you should do the change for all places in trading because now tx id breaks the design as it was broke by fees

https://github.com/user-attachments/assets/4cc746ab-cf34-432c-a852-ba9b80e02f23

packages/suite/src/support/messages.ts
TomasBoda: changed to `TR_TXID`, will possibly change it in #17506

TomasBoda:
> Maybe you should do the change for all places in trading because now tx id breaks the design as it was broke by fees
> 
>  Screen.Recording.2025-03-10.at.10.23.46.mov

Done. It was fixed in all trading forms but there was a little mistake, instead of `1fr` there should have been `minmax(0, 1fr)` to correctly squeeze the inner content.

tomasklim (approved):

PR #17437, opened 2025-03-06
Feat(suite-native): retry FW revision check if it's stale

### LINKED ISSUE #17133: Offline fw revision check banner doesn't disappear after going online
**Describe the bug**
When you connect a device and your phone is offline, the Suite lite can't perform the fw revision check so it shows a banner where it tells the user about this info. However, this banner doesn't go away when you connect back online.

**Info:**

-   Mobile Suite Version: 38e8a08 25.2.2
-   OS: name, version

**How to reproduce**

1. Disable wifi and data on your mobile
2. Connect device to your phone
3. Wait for the banner to appear
4. Now enable again the wifi or data
5. The banner won't go away

**Screenshots**

Image

Lemonexe:
## Description

- add a new Connect method to prompt retrying of FW authenticity checks, if eligible for retry
  - currently, retries happen only during a device call, which'd necessitate a purposeless call just to rerun them
- in mobile, periodically query rerun FW authenticity checks, if eligible for retry

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17133

## Screenshots:

Video demonstrating it [on the testing branch](https://github.com/trezor/trezor-suite/compare/fix-FW-rev-check-persistent-offline..fix-FW-rev-check-persistent-offline-TESTING).
- Using with physical T2B1 via `trezord` mocked with unrecognized version 9,8,7
- It starts with forced NetworkError
- Lots of retries at the beginning (lots of calls to device during handshake & account discovery) 
- Then I let it retry a few times with the given interval
- Then I allowed network requests again, so that FW rev check finally settles

[testing.webm](https://github.com/user-attachments/assets/4b3afd60-8ac7-4e9f-85b4-b7cc10de40a2)

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -796,7 +796,11 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
         if (notDoneYet || wasErrorRetriable) {
             const result = await this.checkFirmwareHash();
-            this.authenticityChecks.firmwareHash = result;
+
+            this.authenticityChecks = {
+                ...this.authenticityChecks,
+                firmwareHash: result,
+            };
 
             if (result === null) return;
             result.attemptCount = attemptsDone + 1;
@@ -902,12 +906,16 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
         const release = getRelease(this.features.internal_model, firmwareVersion);
 
-        this.authenticityChecks.firmwareRevision = await checkFirmwareRevision({
+        const result = await checkFirmwareRevision({
             internalModel: this.features.internal_model,
             deviceRevision: this.features.revision,
             firmwareVersion,
             expectedRevision: release?.firmware_revision,
         });
+        this.authenticityChecks = {
+            ...this.authenticityChecks,
+            firmwareRevision: result,
+        };
     }
 
     async changeLanguage({
--- a/suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
+++ b/suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
@@ -1,0 +1,36 @@
+import { useEffect } from 'react';
+import { useSelector } from 'react-redux';
+
+import TrezorConnect, { FIRMWARE } from '@trezor/connect';
+import { TimerId } from '@trezor/type-utils';
+import { isArrayMember } from '@trezor/utils';
+
+import { selectFirmwareRevisionCheckErrorIfEnabled } from '../selectors';
+
+const REFRESH_INTERVAL = 3_000; // [ms]
+
+export const useRetryFwAuthenticityChecks = () => {
+    const firmwareRevisionCheckError = useSelector(selectFirmwareRevisionCheckErrorIfEnabled);
+
+    const isRetriableError =
+        firmwareRevisionCheckError !== null &&
+        isArrayMember(firmwareRevisionCheckError, FIRMWARE.REVISION_CHECK_RETRIABLE_ERRORS);
+
+    useEffect(() => {
+        let timeoutHandle: TimerId;
+        const recheckFwRevision = () => {
+            if (isRetriableError) {
+                // any device call will cause the tests to be rerun, so getFeatures is used as the most basic one
+                // it'd be useless to await the result; what interests us is the Device state that updates, and gets propagated into redux
+                TrezorConnect.getFeatures({ useEmptyPassphrase: true });
+                timeoutHandle = setTimeout(recheckFwRevision, REFRESH_INTERVAL);
+            }
+        };
+
+        if (isRetriableError) {
+            timeoutHandle = setTimeout(recheckFwRevision, REFRESH_INTERVAL);
+        }
+
+        return () => clearTimeout(timeoutHandle);
+    }, [isRetriableError]);
+};
--- a/suite-native/app/src/hooks/useGlobalHooks.tsx
+++ b/suite-native/app/src/hooks/useGlobalHooks.tsx
@@ -2,6 +2,7 @@ import {
     useDetectDeviceError,
     useHandleDeviceConnection,
     useReportDeviceCompromised,
+    useRetryFwAuthenticityChecks,
 } from '@suite-native/device';
 import { useRenderDeviceCompromisedBanner } from '@suite-native/module-authenticity-checks';
 import { useConnectPopupNavigation } from '@suite-native/module-connect-popup';
@@ -21,4 +22,6 @@ export const useGlobalHooks = () => {
     useDetectDeviceError();
     useReportDeviceCompromised();
     useRenderDeviceCompromisedBanner();
+
+    useRetryFwAuthenticityChecks();
 };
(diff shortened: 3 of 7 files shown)


packages/connect/src/api/runAuthenticityChecks.ts
```diff
@@ -0,0 +1,16 @@
+import { AbstractMethod } from '../core/AbstractMethod';
+
+export default class RunAuthenticityChecks extends AbstractMethod<'runAuthenticityChecks'> {
+    init() {
+        this.requiredPermissions = ['management'];
+        this.useUi = false;
+        this.useEmptyPassphrase = true;
```
Lemonexe, line 7: Frankly I'm not sure what should be here, I only know that `useEmptyPassphrase` is necessary.
I'd need help from Connect team, please.

packages/connect/src/device/Device.ts
```diff
@@ -909,12 +913,16 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
+        const result = await checkFirmwareRevision({
             internalModel: this.features.internal_model,
             deviceRevision: this.features.revision,
             firmwareVersion,
             expectedRevision: release?.firmware_revision,
         });
+        this.authenticityChecks = {
+            ...this.authenticityChecks,
+            firmwareRevision: result,
+        };
```
Lemonexe, line 918: this shouldn't be necessary :sweat_smile: 
But it is :see_no_evil:  [see slack convo](https://satoshilabs.slack.com/archives/C02V2PSDNA2/p1741093466129389)

TL;DR: there is not yet deepclone isolation between Connect and redux in mobile :iphone: , like it is on Web. That causes problems because the objects are passed byRef from Connect, redux does its magic with them, and it causes weird bugs back in Connect. Namely: assignment of `authenticityChecks.firmwareRevision`  does not work, have to assign `authenticityChecks` :hear_no_evil:

suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
```diff
@@ -0,0 +1,29 @@
...
+        isArrayMember(firmwareRevisionCheckError, FIRMWARE.REVISION_CHECK_RETRIABLE_ERRORS);
+
+    useEffect(() => {
+        const recheckFwRevision = () => {
+            if (isRetriableError) {
+                // no need to await the result; what interests us is that Device state updates, which is propagated into redux elsewhere
+                TrezorConnect.runAuthenticityChecks().then();
+            }
+        };
+        const refreshInterval = setInterval(recheckFwRevision, REFRESH_INTERVAL);
```
Lemonexe, line 25: todo redo to setTimeout

packages/connect/src/api/runAuthenticityChecks.ts
```diff
@@ -0,0 +1,16 @@
+import { AbstractMethod } from '../core/AbstractMethod';
+
+export default class RunAuthenticityChecks extends AbstractMethod<'runAuthenticityChecks'> {
+    init() {
+        this.requiredPermissions = [];
+        this.useUi = false;
+        this.useEmptyPassphrase = true;
```
Lemonexe: Frankly I'm not sure what should be here, I only know that `useEmptyPassphrase` is necessary.
I'd need help from Connect team, please.

packages/connect/src/api/runAuthenticityChecks.ts
```diff
@@ -0,0 +1,16 @@
...
+    }
+
+    run() {
+        // the firmware checks happen automatically during Device.run and Device._runInner)
+        // so this is basically a blank method
```
Lemonexe: If I call `await this.device.checkFirmwareRevisionWithRetries()`, it ends up getting called twice, because it's also called in `Device._runInner`

Ideally I'd like this Connect method that does extend Device (so it can call `this.device.checkFirmwareRevisionWithRetries`), but without running `Device.run()`.

Idk if that's possible. If not, then this will have to do..

suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
```diff
@@ -0,0 +1,35 @@
...
+    const isRetriableError =
+        firmwareRevisionCheckError !== null &&
+        isArrayMember(firmwareRevisionCheckError, FIRMWARE.REVISION_CHECK_RETRIABLE_ERRORS);
+
+    useEffect(() => {
+        let timeoutHandle: TimerId;
+        const recheckFwRevision = () => {
+            if (isRetriableError) {
+                // it'd be useless to await the result; what interests us is the Device state that updates, and gets propagated into redux
+                TrezorConnect.runAuthenticityChecks().then();
```
matejkriz: nitpick: empty `.then()` is unnecessary and can be removed, right? Is it only to make the intention of not using the result even more obvious than just with the comment?

suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect } from 'react';
+import { useSelector } from 'react-redux';
+
+import TrezorConnect, { FIRMWARE } from '@trezor/connect';
+import { TimerId } from '@trezor/type-utils';
+import { isArrayMember } from '@trezor/utils';
+
+import { selectFirmwareRevisionCheckErrorIfEnabled } from '../selectors';
+
+const REFRESH_INTERVAL = 10_000; // [ms]
```
matejkriz: Why such a long interval? It means that the offline warning banner will stay at least 10s after user went back online, right? Seems to much to me - as I user I would not wait 10s and would consider it broken. What about 3 or 5 seconds?

suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
Lemonexe: yes, it has only informational value.
IDE warns me "Promise returned from runAuthenticityChecks is ignored" but that's just IDEA, not part of any CI checks. I'll remove it.

matejkriz (approved):
I'm not sure about the refresh interval value, but otherwise the suite-native part looks good. I'm not approving the connect part :-)

suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
Lemonexe: Ok, will do. I have no idea what value is right, I guess 3s is fine.

packages/connect/src/api/runAuthenticityChecks.ts
```diff
@@ -0,0 +1,16 @@
+import { AbstractMethod } from '../core/AbstractMethod';
+
+export default class RunAuthenticityChecks extends AbstractMethod<'runAuthenticityChecks'> {
+    init() {
+        this.requiredPermissions = [];
```
mroz22: this.requiredPermissions = ['management'];

mroz22: this.useDeviceState = false;

PR #17449, opened 2025-03-06
fix(suite): make sure that the DeviceAuthenticity overlay overlaps th…

vojtatranta:
…e sidebar too

## Related Issue

Resolve  https://satoshilabs.slack.com/archives/C07CPLU4SE8/p1741015928984589

### DIFF
--- a/packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeLayout.tsx
+++ b/packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeLayout.tsx
@@ -23,7 +23,6 @@ import { LoggedOutSidebar } from '../LoggedOutSidebar';
 
 const Content = styled.div<{ $elevation: Elevation }>`
     display: flex;
-    position: relative;
     flex-direction: column;
     flex: 3;
     padding: ${spacingsPx.lg};


jvaclavik (approved):
Hope it doesn't break anything 😅

packages/suite/src/components/suite/layouts/WelcomeLayout/WelcomeLayout.tsx
```diff
@@ -23,7 +23,6 @@ import { LoggedOutSidebar } from '../LoggedOutSidebar';
 
 const Content = styled.div<{ $elevation: Elevation }>`
     display: flex;
-    position: relative;
```
vojtatranta, line 26: I am really not sure about this, but it works.

Basically, there is a two column layout on the welcome screen. And we want to right column with the content that has `position: absolute` to overlap to the left column. 

Before, there was a different left column that was a Trezor logo which we didn't wanted it to be overlayed.
So this might make sense.

bosomt:
**QA OK**

image

**Info:**
 - Suite version: web 25.4.0 (539abcb666243c1e18799d45ebdc839497bec808)
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T3B1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)
 - Transport: WebUsbTransport

PR #17453, opened 2025-03-06
chore(connect): separate getRelease and getReleaseInfo

karliatto:
## Description
Little refactor by renaming `getRelease` to `getReleaseInfo` since it is actually what it is doing and create new getRelease that provides `FirmwareRelease`.


mroz22 (approved):
ok

PR #17479, opened 2025-03-07
fix(e2e): disable HW acceleration in playwright desktop

Lemonexe:
## Description

Disable hardware acceleration via GPU for Playwright Desktop E2E tests.

Fixes this problem:
```
[110143:0307/144659.650665:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 1 times!
```

:thought_balloon:  Arguably, this makes the test environment different from user environment, as most people with normal operating systems have GPU on.
However, [ubuntu-latest github runner does not have GPU](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories). So it's questionable what does it even mean to render Chromium with hardware acceleration without it. If it means throwing it back to CPU rendering, than this just simplifies it.


Vere-Grey (approved):
Thanks :+1:

PR #17481, opened 2025-03-07
Chore/remove firmware hash invalid dead code

### LINKED ISSUE #17480: Cleanup after removing Hash check from FW upgrade
In https://github.com/trezor/trezor-suite/pull/16949, the FW hash check has been removed from FW upgrade flow, in favor of a combination of version check & FW hash check that happens on each device connection.

Remove dead code:
- banner (reads `firmwareHashInvalid`)
- redux `firmwareHashInvalid` (there is no longer any way to add or remove there)
- persisted `firmware.firmwareHashInvalid`

## QA instructions

- Migration works from previous DB version both on Suite Web & Desktop
- Initialization of fresh suite works both on Suite Web & Desktop

Lemonexe:
## Description

In https://github.com/trezor/trezor-suite/pull/16949, the FW hash check has been removed from FW upgrade flow, in favor of a combination of version check & FW hash check that happens on each device connection.

Remove dead code:
- banner (reads `firmwareHashInvalid`)
- redux `firmwareHashInvalid` (there is no longer any way to add or remove there)
- persisted `firmware.firmwareHashInvalid`
  - ~create migration there is nothing more on `firmware` persisted, so the migration shall delete entire `firmware`~
  - EDIT: we shall not read/save `firmware.firmwareHashInvalid` to/from storage, but migration shall not be done, as it has no benefits 

## Related Issue

Resolve #17480

@coderabbitai ignore

### DIFF
--- a/packages/suite/src/storage/CHANGELOG.md
+++ b/packages/suite/src/storage/CHANGELOG.md
@@ -20,11 +20,11 @@
 
 -   device state to new object format (`device._state` -> `device.state`)
 
-## 47
+## 47
 
 -   migrate matic to pol
 
-## 46
+## 46
 
 -   added `tokenManagement`
 -   added `device.passwords`
@@ -35,7 +35,7 @@
 -   added `historicRates`
 -   remove rates for all transactions since we are using selectors to get rates
 
-## 44
+## 44
 
 -   remove goerli accounts, txs and settings
 
@@ -71,12 +71,12 @@
 
 -   remove persisted coinjoin sessions
 
-## 36
+## 36
 
 -   token `address` to `contract`
 -   remove ropsten accounts, txs and settings
 
-## 35
+## 35
 
 -   saved ethereum network type txs are removed to be fetched again and obtain internal transfers and token transfer contract and standard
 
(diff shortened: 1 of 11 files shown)


packages/suite/src/storage/CHANGELOG.md
```diff
@@ -20,11 +24,11 @@
 
 -   device state to new object format (`device._state` -> `device.state`)
 
-## 47
+## 47
```
Lemonexe, line 23: Incidental fix of whitespace: it's not clear at first glance, but there were NBSP (unicode 00A0).
AFAIK macOS does that automatically, breaking `md` headings :laughing: 
[See on develop](https://github.com/trezor/trezor-suite/blob/13b0b7c6b4f064528d160c795ed6f9063b3ee628/packages/suite/src/storage/CHANGELOG.md#48)

peter-sanderson (approved):
:+1: Just please lets not do the migration as we discussed on the daily

Lemonexe:
@peter-sanderson done. No need to re-CR, I just removed diffs, change no new diffs.

PR #17482, opened 2025-03-07
Improve vite experience

vojtatranta:
## Description

- adds a ESLINT plugin to warn us if our code breask HMR react -refresh (hot modules replacement
- moves cache for viet to the root node_modules, coz other node modules are automatically delete when `yarn` is run, hence delete build cache

## Related Issue

Resolve

### DIFF
--- a/packages/suite-web/vite.config.ts
+++ b/packages/suite-web/vite.config.ts
@@ -159,6 +159,7 @@ if (typeof window !== 'undefined' && typeof globalThis === 'undefined') {
 
 export default defineConfig({
     root: './src/static',
+    cacheDir: resolve(__dirname, '../../node_modules/.vite'),
     base: assetPrefix,
     // Use suite-data/files as the public directory
     publicDir: resolve(__dirname, '../suite-data/files'),


Lemonexe (approved):
LGTM!
I did notice it was weird that vite created `node_modules` with only `.vite` cache in `suite-web`, but did not have the capacity to think about it deeper :laughing: 

Yes, now the `.vite` cache is placed correctly, and persists when I install `yarn`

Lint addition OK, from my experience it doesn't bring very noticeable benefit, but why not. There are now gonna be warnings at every PR.. It'd be great if someone fixed them :wink:

Lemonexe (changes requested):
oh wait, now I noticed that the ESLint warnings fail the pipeline – we introduced that a few months ago.

No need to fix them at once, just add eslint ignore next line and it can be gradually fixed whenever.
:point_right: Please consider separating it to another PR, because it's thematically unrelated. So it'd be best to extract the eslint addition to a new branch and finish it there. But it's not the end of the world to do it here in a new commit :shrug:

vojtatranta:
> dd eslint ignore next line and

I decided not to add the eslint plugin, the errors there are not correct / there would have to be trillions of comments and it would be just bothering.

I added an issue about ti https://github.com/trezor/trezor-suite/issues/17497

Lemonexe (approved):

PR #17487, opened 2025-03-07
Sentry reporting for FW revision check other-error

### LINKED ISSUE #17484: Report errorPayload for FW revision check other-error
Like it is done for FW hash check `other-error` (which is currently Suite Web/Desktop only),
FW revision check too shall report errorPayload for the `other-error` case.
It hasn't been needed for Web/Desktop, but now it is needed for Suite Lite. The occurrence is very rare, but still, it will give us valuable info if we should dig into it at all.

## QA instructions

**N/A** (affects only Sentry reporting for an unreproducible error)

Lemonexe:
## Description

Commits for review:
- Connect: ship `errorPayload` with FW revision check `other-error`, just like FW hash check `other-error`
- Report it in Suite Desktop/Web + cleanup code to be more consistent 
- Report it in Suite Lite + cleanup code to be more consistent

:monocle_face:  Example events from localhost – see that there is an errorPayload: the exception incl. stack trace:
- [Suite Web](https://satoshilabs.sentry.io/issues/6368992271/?project=5193825)
- [Suite Lite](https://satoshilabs.sentry.io/issues/6369070050/?project=4504214699245568)

## Related Issue

Resolve #17484

### DIFF
--- a/packages/connect/src/device/checkFirmwareRevision.ts
+++ b/packages/connect/src/device/checkFirmwareRevision.ts
@@ -1,4 +1,4 @@
-import { isEqual } from '@trezor/utils/src/versionUtils';
+import { serializeError, versionUtils } from '@trezor/utils';
 
 import { PROTO } from '../constants';
 import { downloadReleasesMetadata } from '../data/downloadReleasesMetadata';
@@ -37,12 +37,19 @@ const getOnlineReleaseMetadata = async ({
 }: GetOnlineReleaseMetadataParams): Promise<FirmwareRelease | undefined> => {
     const onlineReleases = await downloadReleasesMetadata({ internal_model: internalModel });
 
-    return onlineReleases.find(onlineRelease => isEqual(onlineRelease.version, firmwareVersion));
+    return onlineReleases.find(onlineRelease =>
+        versionUtils.isEqual(onlineRelease.version, firmwareVersion),
+    );
 };
 
 const failFirmwareRevisionCheck = (
     error: FirmwareRevisionCheckError,
-): Extract<FirmwareRevisionCheckResult, { success: false }> => ({ success: false, error });
+    errorPayload?: unknown,
+): Extract<FirmwareRevisionCheckResult, { success: false }> => ({
+    success: false,
+    error,
+    ...(errorPayload ? { errorPayload } : null),
+});
 
 export type CheckFirmwareRevisionParams = {
     firmwareVersion: VersionArray;
@@ -113,7 +120,7 @@ export const checkFirmwareRevision = async ({
 
             return isOfflineError(e)
                 ? failFirmwareRevisionCheck('cannot-perform-check-offline')
-                : failFirmwareRevisionCheck('other-error');
+                : failFirmwareRevisionCheck('other-error', serializeError(e));
         }
     }
 
--- a/suite-native/device/src/hooks/useReportDeviceCompromised.ts
+++ b/suite-native/device/src/hooks/useReportDeviceCompromised.ts
@@ -3,14 +3,24 @@ import { useSelector } from 'react-redux';
 
 import * as Sentry from '@sentry/react-native';
 
+import { isDeviceAcquired } from '@suite-common/suite-utils';
 import { selectSelectedDevice } from '@suite-common/wallet-core';
 import { getFirmwareVersion } from '@trezor/device-utils';
 
 import { revisionCheckErrorScenarios } from '../config/firmware';
-import { selectFirmwareRevisionCheckError } from '../selectors';
 
-const reportCheckFail = (checkType: 'Firmware revision', contextData: any) => {
-    Sentry.captureException(`${checkType} check failed! ${JSON.stringify(contextData)}`);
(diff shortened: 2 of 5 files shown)


packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -37,12 +37,19 @@ const getOnlineReleaseMetadata = async ({
 }: GetOnlineReleaseMetadataParams): Promise<FirmwareRelease | undefined> => {
     const onlineReleases = await downloadReleasesMetadata({ internal_model: internalModel });
 
-    return onlineReleases.find(onlineRelease => isEqual(onlineRelease.version, firmwareVersion));
+    return onlineReleases.find(onlineRelease =>
+        versionUtils.isEqual(onlineRelease.version, firmwareVersion),
+    );
```
Lemonexe, line 42: just incidental cleanup of imports...

suite-native/device/src/hooks/useReportDeviceCompromised.ts
```diff
@@ -28,15 +38,17 @@ const useCommonData = () => {
 
 export const useReportDeviceCompromised = () => {
     const commonData = useCommonData();
+    const device = useSelector(selectSelectedDevice);
 
-    const revisionCheckError = useSelector(selectFirmwareRevisionCheckError);
+    const revCheck = isDeviceAcquired(device) ? device.authenticityChecks?.firmwareRevision : null;
+    const isError = revCheck && !revCheck.success;
+    const errorType = isError ? revCheck.error : null;
+    const errorPayload = isError ? revCheck.errorPayload : null;
```
Lemonexe, line 46: the selector is now impracitcal, because I also want to extract the `errorPayload`
This is now [consistent with Suite Desktop/Web ](https://github.com/trezor/trezor-suite/pull/17487/files#diff-529688732e562e8c27d80d982486e7fe74af6988126f9300f07f5bffed8bfdd5)

suite-native/device/src/hooks/useReportDeviceCompromised.ts
```diff
@@ -28,15 +38,17 @@ const useCommonData = () => {
...
 
     useEffect(() => {
-        if (
-            revisionCheckError !== null &&
-            revisionCheckErrorScenarios[revisionCheckError].shouldReport
-        ) {
-            reportCheckFail('Firmware revision', { ...commonData, revisionCheckError });
+        if (!errorType) return;
+        if (revisionCheckErrorScenarios[errorType].shouldReport) {
+            reportCheckFail('Firmware revision', { ...commonData, errorType }, errorPayload);
```
Lemonexe, line 51: renamed from `revisionCheckError` to `errorType`, because no need to repeat that it's **_revision_** check error, it's already in the payload title. Now consistent with Suite Desktop/Web

suite-native/device/src/hooks/useReportDeviceCompromised.ts
```diff
@@ -3,14 +3,24 @@ import { useSelector } from 'react-redux';
...
+    contextData: any,
+    errorPayload?: unknown,
+) => {
+    const payloadLabel = `${checkType} check failed!`;
+    Sentry.withScope(scope => {
+        scope.setExtra('errorPayload', errorPayload);
+        const exceptionForSentry = new Error(`${payloadLabel} ${JSON.stringify(contextData)}`);
+        exceptionForSentry.name = 'reportCheckFail'; // Custom issue title
+        Sentry.captureException(exceptionForSentry, scope);
+    });
```
Lemonexe, line 23: crazy stuff just to keep continuity with the current Sentry issues.
Apparently, without `withScope`, Sentry takes the name of the closes scope, which was, unknowingly to me at the time, the `reportCheckFail` function. That's why the Sentry issue was named `reportCheckFail` for mobile..
But the only way to programatically set Sentry issue name is to pass an Error with a name, which is kinda awkward :see_no_evil:

PeKne (approved):
LGTM

packages/connect/src/device/checkFirmwareRevision.ts
```diff
@@ -1,4 +1,4 @@
-import { isEqual } from '@trezor/utils/src/versionUtils';
```
mroz22, line 1: this might have benn on purpose - we don't have very good treeshaking so this might be there to avoid  bundling full utils package. (not saying we aren't already bundling it because of another import somewhere else)

Lemonexe: Good point, it didn't occur to me. But the general `'@trezor/utils'` [is already imported in Connect](https://github.com/trezor/trezor-suite/blob/41828bdebcc42b1d3258952fcb77ce956a28db3e/packages/connect/src/device/Device.ts#L15-L16)

suite-native/device/src/hooks/useReportDeviceCompromised.ts
peter-sanderson: :scream_cat:

peter-sanderson (approved):

PR #17515, opened 2025-03-10
fix(suite): workaround 'loading takes too long' on big accounts by in…

mroz22:
https://satoshilabs.slack.com/archives/C06AZUTQMD1/p1741605716469399


vojtatranta (approved):
As an emergency improvement this is good.

PR #17520, opened 2025-03-10
fix: Add some exclusion to types to better handle remmaped DeviceEvent type to Redux Actions

### LINKED ISSUE #16839: Refactoring: TrezorConnectEvents
Actions are now defined in two different ways. There was an attempt to unify it in https://github.com/trezor/trezor-suite/pull/16477, but not really finished. Check [this thread](https://github.com/trezor/trezor-suite/pull/16477#discussion_r1933744510) and come up with some reasonable (simple 🙏 ) solution.

peter-sanderson:
This is an alternative (compromise) solution to [this](https://github.com/trezor/trezor-suite/issues/16839#issuecomment-2710311736) problem.

Resolves: https://github.com/trezor/trezor-suite/issues/16839

@coderabbitai ignore

### DIFF
--- a/packages/suite/src/views/onboarding/UnexpectedState/index.tsx
+++ b/packages/suite/src/views/onboarding/UnexpectedState/index.tsx
@@ -27,12 +27,11 @@ const UnexpectedState = ({ children }: UnexpectedStateProps) => {
     const device = useSelector(selectSelectedDevice);
     const prerequisite = useSelector(selectPrerequisite);
 
-    const { prevDevice, activeStepId, showPinMatrix } = useOnboarding();
+    const { prevDeviceId, activeStepId, showPinMatrix } = useOnboarding();
 
     const activeStep = steps.find(s => s.id === activeStepId);
 
     const isNotSameDevice = useMemo(() => {
-        const prevDeviceId = prevDevice?.id;
         // if no device was connected before, assume it is same device
         if (!prevDeviceId) {
             return false;
@@ -44,7 +43,7 @@ const UnexpectedState = ({ children }: UnexpectedStateProps) => {
         }
 
         return deviceId !== prevDeviceId;
-    }, [prevDevice, device]);
+    }, [prevDeviceId, device]);
 
     const UnexpectedStateComponent = useMemo(() => {
         if (!activeStep?.prerequisites) return null;
--- a/packages/type-utils/src/utils.ts
+++ b/packages/type-utils/src/utils.ts
@@ -110,7 +110,7 @@ export type FilterPropertiesByType<T, ValueFilter> = {
 };
 
 /**
- * Removed the type from the union where `{ type: U }`.
+ * Removed the type from the union where `{ KeyName: ValueToExclude }`.
  *
  * Example:
  *  ```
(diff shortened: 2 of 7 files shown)


packages/type-utils/src/utils.ts
```diff
@@ -108,3 +108,23 @@ export type DefinedUnionMember<T> = T extends string ? T : never;
...
+ *
+ *  // { type: 'A', a: string } | { type: 'B', b: number } | { type: 'D' | 'E', cde: boolean };
+ *  type NotC = FilterOutFromUnionByTypeProperty<T1, 'C'>;
+ *  ```
+ */
+export type FilterOutFromUnionByTypeProperty<T, U> = T extends { type: infer P }
+    ? P extends U
+        ? never
+        : { type: Exclude<P, U> } & Omit<T, 'type'>
+    : T;
```
Lemonexe: Suggestion for enhancement - several things:
- mainly a new generic argument instead of hardcoded `'type'` :pray: 
- I personally prefer named types everywhere, instead of `T,P,U` one-letter variables for generics :shrug: 
- also new safeguard  `ValueToExclude extends Union[KeyName]`, because without it you can do this: 
`type WhatIsThis = FilterOutFromUnionByTypeProperty<T1, 'type', 123>;`

EDIT: fixed
```
/**
 * Removed the type from the union where `{ KeyName: ValueToExclude }`.
 *
 * Example:
 *  ```
 *  type T1 =
 *     | { type: 'A'; a: string }
 *     | { type: 'B'; b: number }
 *     | { type: 'C' | 'D' | 'E'; cde: boolean };
 *
 *  // { type: 'A', a: string } | { type: 'B', b: number } | { type: 'D' | 'E', cde: boolean };
 *  type NotC = FilterOutFromUnionByTypeProperty<T1, 'type', 'C'>;
 *  ```
 */
export type FilterOutFromUnionByTypeProperty<
    Union,
    KeyName extends keyof Union,
    ValueToExclude extends Union[KeyName],
> = Union extends { [K in KeyName]: infer ActualValue }
    ? ActualValue extends ValueToExclude
        ? never
        : { [K in KeyName]: Exclude<ActualValue, ValueToExclude> } & Omit<Union, KeyName>
    : Union;
```

packages/suite/src/views/onboarding/UnexpectedState/index.tsx
```diff
@@ -44,7 +43,7 @@ const UnexpectedState = ({ children }: UnexpectedStateProps) => {
         }
 
         return deviceId !== prevDeviceId;
-    }, [prevDevice, device]);
+    }, [prevDeviceId, device]);
```
Lemonexe, line 46: if I understand correctly, the reason for this change is that the `useEffect` would fire after each connect state update unnecessarily, because `prevDevice` is unstable?

Lemonexe (approved):
LGTM!

packages/type-utils/src/utils.ts
peter-sanderson: I knew you would love thy TS stuff :smile:

packages/suite/src/views/onboarding/UnexpectedState/index.tsx
peter-sanderson: No, it is just that in that action the device is either `Deice | TrezorDevice` those two types are (temporarily) incompatible and Marek pointed to me, that all I need is `id` (not the whole `device`). So I simplified it to just `id` avoiding the `Device` vs `TrezorDevice` madness at all.

Lemonexe (approved):
oh, you didn't have to credit me, you're too nice :blush:

PR #17537, opened 2025-03-10
chore: upgrade to TS 5.8

Nodonisko:
## Description

I had to set target to ES2023 because 24 introduces some changes to Buffer typings which caused huge amount of error across mostly codebase. I don't see it as issue to have it set for ES2023.

There was few minor changes necessary but it should not break anything.

Also updated ESlint because it's kind of dependant on TS version.

## Related Issue

Resolve

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
         "s": "yarn native:start"
     },
     "resolutions": {
-        "typescript": "5.7.3",
+        "typescript": "5.8.2",
         "react-native": "0.76.1",
         "prettier": "3.2.5",
         "type-fest": "4.24.0",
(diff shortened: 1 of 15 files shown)


komret (approved):

komret:
> The component has been updated to use explicit parentheses for the JSX return value, which is a good practice for clarity. However, I noticed that the null coalescing operator (`?? null`) has been removed, which means the component will now return an empty fragment instead of null when the formatted output is nullish.
> While this doesn't affect typical usage patterns, it's worth noting that the component interface on line 33 allows for returning `JSX.Element | null`. The updated implementation aligns better with React best practices but has a slightly different behavior.

Looks like a valid point, @Nodonisko.

Lemonexe:
@SocketSecurity ignore npm/@eslint/config-helpers@0.1.0
it's a package recently published by eslint as part of the monorepo

PR #17538, opened 2025-03-10
feat(mobile): reintroduce FW update banner on dashboard

### LINKED ISSUE #16476: Show firmware update alert
Firmware update alert was temporarily removed from the dashboard in https://github.com/trezor/trezor-suite/pull/16473.

Let's discuss when to bring it back.

Nodonisko:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16476

### DIFF
--- a/suite-native/module-home/src/screens/HomeScreen/components/PortfolioContent.tsx
+++ b/suite-native/module-home/src/screens/HomeScreen/components/PortfolioContent.tsx
@@ -18,6 +18,7 @@ import {
     StackNavigationProps,
 } from '@suite-native/navigation';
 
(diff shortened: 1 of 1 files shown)


PeKne (approved):

matejkriz:
Disconnected view-only device is not handled well - I would expect Connect&Unlock screen or any other indication that you need to connect the Trezor. Maybe we should consider mentioning that in the banner itself or hide it if the Trezor is not connected.

https://github.com/user-attachments/assets/108db355-e3d8-4dde-b63d-25a9b650e6a6

PR #17539, opened 2025-03-10
fix(transport): `UDP` read from buffer or promise

szymonlesisz:
## Description

create shared utility for `UDP` and `Bluetooth` transports

read from message buffer or message promise (dfd)

fixing problem with `UDP.read` timeout error rejected on ButtonRequest


mroz22 (approved):
lgtm thanks

PR #17545, opened 2025-03-10
feat(native-suite): handling discovery/disconnect state in Settings

### LINKED ISSUE #16010: Device settings - handle disconnected device and discover running
We have inconsistent handling of disconnected device and running discovery states for each device settings feature (FW update, PIN change, Device auth check). The goal of this issue it to make it the same for all of those features.

### Disconnected trezor
1. **No buttons are disabled** on the Device Settings screen, if selected trezor is not connected.
1. When clicking on actions like Firmware Update, PIN, or Auth Check without the selected trezor connected:
	1. [Preferred option] Show the “_Connect&Unlock your Trezor_” screen.
	    1. Once the correct device is connected and unlocked, t**he action continues**.
	    1. If you connect **different** device at that point, you should be redirected to dashboard of that newly connected trezor.
	1. [Backup solution]  If the preferred option is too complex, keep the current behavior: show the “_Continue on your Trezor_” screen for PIN and Auth check and add the “_Connect&Unlock your Trezor_” only for FW update.
    1. Even if you have **some** trezor connected, but you have selected **different** view-only device, you will see "Connect&Unlock"
    	1. [nice to have] Indicate which Trezor needs to be connected and unlocked, but we should handle that globally.
1. If the Trezor is disconnected during an action, display an error message.
image

### Discovery in progress
1. While discovery is running, all buttons on the Device Settings screen are disabled and shown in a loading state.
1. [nice to have] If we are confident that specific settings can proceed safely during discovery, we can enable those actions.

## Notes
- The reason why we don't want to just hide the Device settings completely with device disconnected is that we would like to promote FW update even without the device connected and it should lead somewhere.

BrantalikP:
## Description

Enhancement of handling disconnect and discovery states in settings.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16010

### Discovery State
https://github.com/user-attachments/assets/f9c32570-3163-4af7-a3bf-2e04b4991204

### Disconnect State

https://github.com/user-attachments/assets/7701d161-9816-41d3-a79d-e549c49ccc88

### DIFF
--- a/suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
+++ b/suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
@@ -6,13 +6,10 @@ import { useNavigation } from '@react-navigation/native';
 
 import { getFwUpdateVersion } from '@suite-common/suite-utils';
 import {
-    DeviceRootState,
-    DiscoveryRootState,
     selectDeviceModel,
     selectDeviceReleaseInfo,
     selectIsDeviceBackedUp,
-    selectIsDeviceConnected,
-    selectIsDiscoveryActiveByDeviceState,
+    selectIsDeviceDiscoveryActive,
     selectSelectedDevice,
 } from '@suite-common/wallet-core';
 import { Button, HStack, Text, VStack } from '@suite-native/atoms';
@@ -64,11 +61,8 @@ export const DeviceFirmwareCard = () => {
     const deviceModel = useSelector(selectDeviceModel);
     const deviceReleaseInfo = useSelector(selectDeviceReleaseInfo);
     const isDeviceBackedUp = useSelector(selectIsDeviceBackedUp);
-    const isDeviceConnected = useSelector(selectIsDeviceConnected);
+    const isDiscoveryRunning = useSelector(selectIsDeviceDiscoveryActive);
 
-    const isDiscoveryRunning = useSelector((state: DiscoveryRootState & DeviceRootState) =>
-        selectIsDiscoveryActiveByDeviceState(state, device?.state),
-    );
     const navigation = useNavigation<NavigationProp>();
     const isFirmwareUpdateEnabled = useIsFirmwareUpdateFeatureEnabled();
 
@@ -82,7 +76,7 @@ export const DeviceFirmwareCard = () => {
         : 'firmware.typeUniversal';
 
     const firmwareUpdateProps = (() => {
-        if (!isFirmwareUpdateEnabled || !isDeviceBackedUp || !isDeviceConnected) {
+        if (!isFirmwareUpdateEnabled || !isDeviceBackedUp) {
             return undefined;
         }
 
--- a/suite-native/module-device-settings/src/components/DevicePinActionButton.tsx
+++ b/suite-native/module-device-settings/src/components/DevicePinActionButton.tsx
@@ -3,7 +3,11 @@ import { useDispatch, useSelector } from 'react-redux';
 
 import { useNavigation } from '@react-navigation/native';
 
-import { removeButtonRequests, selectSelectedDevice } from '@suite-common/wallet-core';
+import {
+    removeButtonRequests,
+    selectIsDeviceDiscoveryActive,
+    selectSelectedDevice,
+} from '@suite-common/wallet-core';
 import { useAlert } from '@suite-native/alerts';
 import { EventType, analytics } from '@suite-native/analytics';
 import { Button, ButtonColorScheme } from '@suite-native/atoms';
@@ -59,6 +63,8 @@ export const DevicePinActionButton = ({
     type,
     colorScheme,
 }: DevicePinActionButtonProps) => {
+    const isDiscoveryRunning = useSelector(selectIsDeviceDiscoveryActive);
+
     const navigation = useNavigation<NavigationProp>();
     const dispatch = useDispatch();
     const { showToast } = useToast();
@@ -141,6 +147,8 @@ export const DevicePinActionButton = ({
             colorScheme={colorScheme}
             size="small"
             testID={`@device-pin-protection/${type}-button`}
+            isDisabled={isDiscoveryRunning}
+            isLoading={isDiscoveryRunning}
         >
             {children}
         </Button>
--- a/suite-native/module-device-settings/src/hooks/useDeviceConnectionGuard.ts
+++ b/suite-native/module-device-settings/src/hooks/useDeviceConnectionGuard.ts
@@ -1,0 +1,45 @@
+import { useCallback, useEffect } from 'react';
+import { useSelector } from 'react-redux';
+
+import { useNavigation } from '@react-navigation/native';
+
+import { selectIsDeviceConnected } from '@suite-common/wallet-core';
+import {
+    AuthorizeDeviceStackRoutes,
+    DeviceStackRoutes,
+    RootStackParamList,
+    RootStackRoutes,
+    SettingsStackParamList,
+    StackToStackCompositeNavigationProps,
+} from '@suite-native/navigation';
+
+type NavigationProps = StackToStackCompositeNavigationProps<
+    SettingsStackParamList,
+    DeviceStackRoutes.DeviceSettings,
+    RootStackParamList
+>;
+
+export const useDeviceConnectionGuard = () => {
+    const navigation = useNavigation<NavigationProps>();
+
+    const isDeviceConnected = useSelector(selectIsDeviceConnected);
+
+    const redirectToConnectAndUnlockScreen = useCallback(() => {
+        navigation.navigate(RootStackRoutes.AuthorizeDeviceStack, {
+            screen: AuthorizeDeviceStackRoutes.ConnectAndUnlockDevice,
+            params: {
+                onCancelNavigationTarget: {
+                    name: RootStackRoutes.DeviceSettingsStack,
+                    params: { screen: DeviceStackRoutes.DeviceSettings },
+                },
+            },
+        });
+    }, [navigation]);
+
+    useEffect(() => {
+        if (!isDeviceConnected) return redirectToConnectAndUnlockScreen();
+        // The check should be conducted only once when the screen is shown
+    }, [redirectToConnectAndUnlockScreen]); // eslint-disable-line react-hooks/exhaustive-deps
+
+    return { isDeviceConnected };
+};
--- a/suite-native/module-device-settings/src/navigation/DevicePinProtectionStackNavigator.tsx
+++ b/suite-native/module-device-settings/src/navigation/DevicePinProtectionStackNavigator.tsx
@@ -10,6 +10,7 @@ import {
     stackNavigationOptionsConfig,
 } from '@suite-native/navigation';
 
+import { useDeviceConnectionGuard } from '../hooks/useDeviceConnectionGuard';
 import { ContinueOnTrezorScreen } from '../screens/ContinueOnTrezorScreen';
 import {
     ConfirmNewPinScreen,
@@ -20,6 +21,8 @@ import {
 const DevicePinProtectionStack = createNativeStackNavigator<DevicePinProtectionStackParamList>();
 
 export const DevicePinProtectionStackNavigator = () => {
+    const { isDeviceConnected } = useDeviceConnectionGuard();
+
     const buttonRequestCodes = useSelector(selectDeviceButtonRequestsCodes);
     const lastButtonRequestCode = A.last(buttonRequestCodes);
 
@@ -27,6 +30,8 @@ export const DevicePinProtectionStackNavigator = () => {
     const isEnterNewPin = lastButtonRequestCode === 'PinMatrixRequestType_NewFirst';
     const isConfirmNewPin = lastButtonRequestCode === 'PinMatrixRequestType_NewSecond';
     const isContinueOnTrezor = !isEnterCurrentPin && !isEnterNewPin && !isConfirmNewPin;
+
+    if (!isDeviceConnected) return;
 
     // To indicate progress to the user we need separate screens for individual steps of the flow.
     // At the same time we need just one available so that navigation.goBack() works as expected.
--- a/suite-native/device/src/hooks/useHandleDeviceConnection.ts
+++ b/suite-native/device/src/hooks/useHandleDeviceConnection.ts
@@ -75,9 +75,6 @@ export const useHandleDeviceConnection = () => {
     const isSuspiciousDeviceScreenFocused = useNavigationRouteMatch(
         DeviceOnboardingStackRoutes.SuspiciousDevice,
     );
-    const isConnectAndUnlockDeviceScreenFocused = useNavigationRouteMatch(
-        AuthorizeDeviceStackRoutes.ConnectAndUnlockDevice,
-    );
 
     const lastRoute = useNavigationState(state => state?.routes.at(-1)?.name);
     const isDeviceSettingsStackFocused = lastRoute === RootStackRoutes.DeviceSettingsStack;
@@ -170,11 +167,7 @@ export const useHandleDeviceConnection = () => {
     // In case that the physical device is disconnected, redirect to the home screen and
     // set connecting screen to be displayed again on the next device connection.
     useEffect(() => {
-        if (
-            isFirmwareInstallationRunning ||
-            !isOnboardingFinished ||
-            isConnectAndUnlockDeviceScreenFocused
-        )
+        if ((isFirmwareInstallationRunning && !isOnboardingStackFocused) || !isOnboardingFinished)
             return;
 
         if (isNoPhysicalDeviceConnected) {
@@ -212,7 +205,6 @@ export const useHandleDeviceConnection = () => {
         isSuspiciousDeviceScreenFocused,
         isOnboardingStackFocused,
         handleOnboardingDeviceDisconnection,
-        isConnectAndUnlockDeviceScreenFocused,
         isDeviceOnboardingStackFocused,
     ]);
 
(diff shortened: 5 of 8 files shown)


suite-native/module-device-settings/src/components/DevicePinActionButton.tsx
```diff
@@ -133,16 +139,26 @@ export const DevicePinActionButton = ({
...
         <Button
             onPress={changePin}
             colorScheme={colorScheme}
             size="small"
             testID={`@device-pin-protection/${type}-button`}
+            isDisabled={isDiscoveryRunning}
         >
-            {children}
+            {isDiscoveryRunning ? <LoadingIndicator /> : children}
         </Button>
```
PeKne: passing `LoadingIndicator` as a child should not be necessary. The button has `isLoading` prop already. If you set it to true, the inner logic will display the loader itself.

PeKne: Please fix this also in the other settings cards.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
```diff
@@ -0,0 +1,43 @@
...
+    SettingsStackParamList,
+    DeviceStackRoutes.DeviceSettings,
+    RootStackParamList
+>;
+
+export const useSettingsProtection = () => {
+    const device = useSelector(selectSelectedDevice);
+    const navigation = useNavigation<NavigationProps>();
+
+    const isDiscoveryRunning = useSelector((state: DiscoveryRootState & DeviceRootState) =>
```
PeKne: Why is this selector called in th hook if it is never used, only returned? What about removing it from here and calling it directly at settings card components? this seems like unnecessary step to me. What do you think?

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
```diff
@@ -0,0 +1,43 @@
...
+        selectIsDiscoveryActiveByDeviceState(state, device?.state),
+    );
+    const isDeviceConnected = useSelector(selectIsDeviceConnected);
```
PeKne: Same applies for this hook.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const handleConnectDevice = () => {
```
PeKne: When I try to trigger any device settings action with disconnected device, the `Connect&Unlock` screen is displayed correctly. When I connect the device again I'm redirected to the device settings screen (one step back). Should not the flow continue after the connection? E.g: 
1. I click FW update with disconnected device
2. `Connect&Unlock` screen is shown
3. I connect the device and I'm redirected to the FW installation screen

The issue talks about multiple solutions. So maybe this is a desired state that you came to with @matejkriz at the end. Just want to clarify that.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
PeKne: Another unrelated thing... You can use `selectIsDeviceDiscoveryActive` that directly uses `selectSelectedDevice` inside. So there is no need to calling this selector at the top of this hook to get the `device.state`.

PeKne (changes requested):
I have a few comments on this, but generally, good job for the first issue :clap:.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
BrantalikP: The reason is that I want to avoid unnecessary prop drilling. The `useSettingsProtection` hook is now used in three places (Firmware Update, PIN, and Device Authenticity Check). It’s not about the selectors but rather the hook’s responsibility. For example, if the logic were to change, we could adjust it in one place rather than three.

BrantalikP: Sorry, I misread. You mentioned the Settings Card `components`. In that case, there’s no prop drilling, but the idea remains the same. AFAIK, the purpose of this PR is to unify the logic for **Settings** Actions. Therefore, it makes sense to handle the logic in one place for easier maintenance and consistency.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
BrantalikP: Yeah, we discussed this one. it’s a bit tricky. What if the user connects another device that doesn’t need the FW update? In that case, we would probably need to check whether it’s valid or not. Any ideas on how to handle this better? 🤔

suite-native/module-device-settings/src/components/DevicePinActionButton.tsx
BrantalikP: I didn’t realize that. Thank you! @PeKne

BrantalikP: @PeKne https://github.com/trezor/trezor-suite/pull/17545/commits/f3ffb66b62d88853b61b302fe5d9d4e0d086c511 🙏

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
BrantalikP: https://github.com/trezor/trezor-suite/pull/17545/commits/f3ffb66b62d88853b61b302fe5d9d4e0d086c511 used `selectIsDeviceDiscoveryActive` instead

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
BrantalikP: https://github.com/trezor/trezor-suite/pull/17545#discussion_r1988553355

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
PeKne: Yes it would be definitely more complicated. Depens on product ( @fang314, @shenkys ) would like to improve it. I am ok with the current solution.

Btw. what happens now if he user connects a different device after `Connect&Unlock`? Will he end up in the device settings? IMO he should be redirected to the home screen if different device is connected. But this might be solved in a different issue. it is very possible it is "broken" like this in other flows already.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
PeKne: Ok I got your view. This is opinionated, feel free to leave it as it is.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
matejkriz: The desired solution is definitely as stated in the description: "Once the correct device is connected and unlocked, the action continues". and "If you connect different device at that point, you should be redirected to dashboard of that newly connected trezor."

Connecting selected trezor device and connecting different one are different actions so it is possible, not sure how hard would be to achieve that, but I'd still give it a try.

BrantalikP: OK, guys, understood. 👌 I’m going to set the PR to draft and try to find the desired solution if you agree @matejkriz . Thank you both for the review! 🙏

BrantalikP: Resolved https://github.com/trezor/trezor-suite/pull/17545/commits/790ae9f103ffce56afcb7819f9d53b78e4800c5b Also updated the video attachment for handling the disconnect state. cc: @PeKne @matejkriz

BrantalikP: After discussion, I’ve created a separate issue for handling cases where a different device is connected: https://github.com/trezor/trezor-suite/issues/17594 Please validate if it makes sense. Thank you!

suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
```diff
@@ -64,13 +62,10 @@ export const DeviceFirmwareCard = () => {
...
     const deviceReleaseInfo = useSelector(selectDeviceReleaseInfo);
     const isDeviceBackedUp = useSelector(selectIsDeviceBackedUp);
-    const isDeviceConnected = useSelector(selectIsDeviceConnected);
 
-    const isDiscoveryRunning = useSelector((state: DiscoveryRootState & DeviceRootState) =>
-        selectIsDiscoveryActiveByDeviceState(state, device?.state),
-    );
     const navigation = useNavigation<NavigationProp>();
     const isFirmwareUpdateEnabled = useIsFirmwareUpdateFeatureEnabled();
+    const { isDiscoveryRunning } = useSettingsProtection();
```
PeKne: I don't see any benefit of reading `isDiscoveryRunning` from this hook. Why not to keep the original solution? Maybe just changing the selector of the original solution to `selectIsDeviceDiscoveryActive`.

PeKne: Same applies for the other uses.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
```diff
@@ -0,0 +1,45 @@
...
+                },
+            },
+        });
+    }, [navigation]);
+
+    useEffect(() => {
+        if (!isDeviceConnected && isProtectionEnabled) return handleConnectDevice();
+    }, [handleConnectDevice, isDeviceConnected, isProtectionEnabled]);
+
+    return { isDiscoveryRunning, isDeviceConnected, handleConnectDevice };
```
PeKne: Remove `handleConnectDevice` it is not used anywhere outside of this hook.
```suggestion
    return { isDiscoveryRunning, isDeviceConnected };
```

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
```diff
@@ -0,0 +1,45 @@
...
+    StackToStackCompositeNavigationProps,
+} from '@suite-native/navigation';
+
+type NavigationProps = StackToStackCompositeNavigationProps<
+    SettingsStackParamList,
+    DeviceStackRoutes.DeviceSettings,
+    RootStackParamList
+>;
+
+export const useSettingsProtection = (isProtectionEnabled = false) => {
```
PeKne: After the latest fixup, it fells like this hook is combining two unrelated things (disabling buttons while discovery AND handling device re-connection). What about removing `isDiscoveryRunning` from this hook completely and reading `useSelector(selectIsDeviceDiscoveryActive);` directly in all the places where you use `const { isDiscoveryRunning } = useSettingsProtection();`? The `isDiscoveryRunning` is not used anywhere inside of this hook, so that should do no harm. 

If you do that, you can use this hook only in the StackNavigators and remove the `isProtectionEnabled` prop completely. the prop feel confusing to me and it would take me a minute to figure out if I should set it in case I would be reusing this hook somewhere else. Wdyt?

suite-native/module-device-settings/src/navigation/DevicePinProtectionStackNavigator.tsx
```diff
@@ -28,6 +31,8 @@ export const DevicePinProtectionStackNavigator = () => {
     const isConfirmNewPin = lastButtonRequestCode === 'PinMatrixRequestType_NewSecond';
     const isContinueOnTrezor = !isEnterCurrentPin && !isEnterNewPin && !isConfirmNewPin;
 
+    if (!isDeviceConnected) return;
```
PeKne, line 34: What is the benefit of returning null here? Would not it work even without it? IMO it should be not a problem that these screen will be rendered if there will be the `Connect&Unlock` screen in the stack above it. Or is there any other reason?

PeKne: I tested removing this check and it seems to me it behaves identically.

BrantalikP: It prevents the screen beneath from being visible before you connect the device. I think without it, you would see two screens sliding up simultaneously.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
```diff
@@ -0,0 +1,45 @@
...
+    const navigation = useNavigation<NavigationProps>();
+
+    const isDiscoveryRunning = useSelector(selectIsDeviceDiscoveryActive);
+    const isDeviceConnected = useSelector(selectIsDeviceConnected);
+
+    const handleConnectDevice = useCallback(() => {
```
PeKne: I would sugest to use some better name. I would expect that function called `handleConnectDevice` will be handling the connection, but this is not the case. 

What about something like this? 
```suggestion
    const redirectToConnect&UnlockScreen = useCallback(() => {
```

Or you can remove the callback completely and move its body directly to the useEffect.

suite-native/module-device-settings/src/navigation/DevicePinProtectionStackNavigator.tsx
BrantalikP: r at least, if you go to the FW update screen, I didn’t like that you could see the FW page for a brief moment before the Connect and Unlock screen appeared.

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
BrantalikP: You are right, this is a better name. I’ve just reused a function with the same responsibility. Will change

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
BrantalikP: Yeah, this is basically what we discussed last time. My point is not to focus on specific variables, etc., but rather on the responsibility of the hook itself so we can change the logic later if needed. However, I see your point that with the introduction of the newly implemented hook, the responsibilities are becoming ambiguous. Let’s do it your way 👍 extract this selector and use it directly in the screens. Thanks!

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
BrantalikP: https://github.com/trezor/trezor-suite/pull/17545/commits/6bd8e85bd5b268b1f41c88f5ba67b95464056be0

suite-native/module-device-settings/src/hooks/useSettingsProtection.ts
BrantalikP: https://github.com/trezor/trezor-suite/pull/17545/commits/6bd8e85bd5b268b1f41c88f5ba67b95464056be0

suite-native/module-device-settings/src/navigation/DevicePinProtectionStackNavigator.tsx
BrantalikP: ### With **isDeviceConnected**
https://github.com/user-attachments/assets/a80be50c-040d-4adc-8c3d-8d0a1c1930bc

### Without **isDeviceConnected**
https://github.com/user-attachments/assets/41700ef9-bb01-4928-8314-b85e12c495cd

suite-native/module-device-settings/src/hooks/useDeviceConnectionGuard.ts
```diff
@@ -0,0 +1,44 @@
...
+                },
+            },
+        });
+    }, [navigation]);
+
+    useEffect(() => {
+        if (!isDeviceConnected) return redirectToConnectAndUnlockScreen();
+    }, [redirectToConnectAndUnlockScreen, isDeviceConnected]);
+
+    return { isDeviceConnected, redirectToConnectAndUnlockScreen };
```
PeKne: This is still not needed.
```suggestion
    return { isDeviceConnected };
```

PeKne (approved):
Please fix this one more nit. Apart from that it looks good :clap:.

suite-native/module-device-settings/src/hooks/useDeviceConnectionGuard.ts
BrantalikP: https://github.com/trezor/trezor-suite/pull/17545/commits/ca533f6fffc6f9da24a65e9657c77935e847e441 done

suite-native/module-device-settings/src/hooks/useDeviceConnectionGuard.ts
```diff
@@ -0,0 +1,44 @@
...
+                onCancelNavigationTarget: {
+                    name: RootStackRoutes.DeviceSettingsStack,
+                    params: { screen: DeviceStackRoutes.DeviceSettings },
+                },
+            },
+        });
+    }, [navigation]);
+
+    useEffect(() => {
+        if (!isDeviceConnected) return redirectToConnectAndUnlockScreen();
```
matejkriz, line 40: Device is disconnected during FW installation flow because it has to be restarted to bootloader mode and this effect is active there. 
image

matejkriz: https://github.com/user-attachments/assets/6170ab53-5146-4a89-a317-92567a6ed1b2

BrantalikP: ok that's new 😅 on it!

BrantalikP: https://github.com/trezor/trezor-suite/pull/17545/commits/1e4fe01caf37d78ed486a1b87d5ee1495204d282 

https://github.com/user-attachments/assets/1227ce37-3119-4790-94ec-ddb20711f829

Issue #1, opened 2018-11-26
React Native build on android


good-lly:
react native build is working with ios but not android emualtor - fix it

bosomt:
just do it

szymonlesisz:
already implemented in develop branch

PR #17546, opened 2025-03-10
fix(trading): selecting account

### LINKED ISSUE #17452: Trading - changing swapped coin from any coin/network to any token selects the tokens mainnet instead
**Describe the bug**
If you go to "Swap" straight from the dashboard and you want to select some token on one of your accounts as the token you want to swap FROM, it selects the mainnet account instead of the desired token.

**Steps to reproduce:**
1. Connect Trezor with SOL/ETH/BNB funds
2. Go to "Swap"
3. Select some token instead of the mainnet
4. You won't be able to select any token and it switches to the mainnet

**Info:**
 - Suite version: desktop 25.3.0 (ea18d308390733f81f7260c39cd9e5bee6c9a198)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.3.0 Chrome/132.0.6834.194 Electron/34.1.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 3440x1440
 - Device: Trezor T2T1 2.8.9 regular (revision 7614a29c9a9cc138fff446937b7d9489fd57821f)
 - Transport: BridgeTransport 3.0.0-bundled.25.3.0

**Expected result:**
The token I want to swap FROM is selected correctly

**Actual result:**
Mainnet currency is selected instead of the desired token in the "Swap" trading section

adderpositive:
## Description
Fixing overriding selected account

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17452

### DIFF
--- a/packages/suite/src/hooks/wallet/trading/form/useTradingSellForm.ts
+++ b/packages/suite/src/hooks/wallet/trading/form/useTradingSellForm.ts
@@ -585,10 +585,10 @@ export const useTradingSellForm = ({
 
     useEffect(() => {
         // when draft doesn't exist, we need to bind actual default values - that happens when we've got sellInfo from Invity API server
-        if (!isDraft && sellInfo && !formState.isDirty) {
+        if (!isDraft && sellInfo && isInitialDataLoading) {
             reset(defaultValues);
         }
-    }, [reset, sellInfo, defaultValues, isDraft, isNotFormPage, formState.isDirty]);
+    }, [reset, sellInfo, defaultValues, isDraft, isNotFormPage, isInitialDataLoading]);
 
     useDebounce(
         () => {
--- a/packages/suite/src/hooks/wallet/trading/form/common/useTradingFormActions.ts
+++ b/packages/suite/src/hooks/wallet/trading/form/common/useTradingFormActions.ts
@@ -185,6 +185,7 @@ export const useTradingFormActions = <T extends TradingSellExchangeFormProps>({
 
         setAccountOnChange(account);
         setExchangeReceiveCrypto(selected);
+        setValue(FORM_SEND_CRYPTO_CURRENCY_SELECT, selected);
 
         changeFeeLevel('normal'); // reset fee level
     };
--- a/packages/suite/src/hooks/wallet/trading/form/useTradingExchangeForm.ts
+++ b/packages/suite/src/hooks/wallet/trading/form/useTradingExchangeForm.ts
@@ -713,12 +713,12 @@ export const useTradingExchangeForm = ({
         register('setMaxOutputId');
     }, [register]);
 
-    // react-hook-form reset, set default values
+    // when draft doesn't exist, we need to bind actual default values - that happens when we've got exchangeInfo from Invity API server
     useEffect(() => {
-        if (!isDraft && defaultValues) {
+        if (!isDraft && exchangeInfo && isInitialDataLoading) {
             reset(defaultValues);
         }
-    }, [reset, isDraft, defaultValues]);
+    }, [reset, isDraft, exchangeInfo, defaultValues, isInitialDataLoading]);
 
     useEffect(() => {
         if (!quotesRequest && isNotFormPage) {
--- a/packages/suite/src/views/wallet/trading/common/TradingForm/TradingFormInput/TradingFormInputAccount.tsx
+++ b/packages/suite/src/views/wallet/trading/common/TradingForm/TradingFormInput/TradingFormInputAccount.tsx
@@ -53,14 +53,13 @@ export const TradingFormInputAccount = <
         <Controller
             name={accountSelectName}
             control={control}
-            render={({ field: { onChange, value } }) => (
+            render={({ field: { value } }) => (
                 <Select
                     value={value}
                     labelLeft={label && <Translation id={label} />}
                     options={optionGroups}
                     onChange={async (selected: TradingAccountOptionsGroupOptionProps) => {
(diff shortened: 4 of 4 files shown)


FreeWall (approved):

packages/suite/src/hooks/wallet/trading/form/useTradingSellForm.ts
```diff
@@ -584,10 +584,10 @@ export const useTradingSellForm = ({
 
     useEffect(() => {
         // when draft doesn't exist, we need to bind actual default values - that happens when we've got sellInfo from Invity API server
-        if (!isDraft && sellInfo && !formState.isDirty) {
+        if (!isDraft && sellInfo && isInitialDataLoading) {
```
vojtatranta, line 588: My random ideas:

So, when the form is done loading initially, we do not want to set the different defaults when account changes? 

I think that `sellInfo` is loaded regardless of what happens in this form. 

So if it is the case, that we **do not want to change the defaults once the sellInfo is loaded**, why bother even to have anything regarding the setting different defaults when account is changed?

But I don't know. I lack context of this.

My initial idea of how to fix this problem was to merge the "next defaults" with the current state at the moment when `setAccountKey` is called.

And, hence remove the creation of the defaults from the `account` as the defaults are created declaratively based on the selected account.

So you would do something like:
```ts
 setAccountOnChange: newAccount => {
            dispatch(tradingSellActions.setTradingSellAccountKey(newAccount.key));
            reset(mergeValues(methods.values, getDefaultValues(newAccount))
        },
```

But I don't know...

vojtatranta:
I tried it locally and I don't think it is fixed:
- went to BTC account
- go to Swap
- in Form field switch to USDT on Solana #1
- tried to switch to other token from different account
- 🛑 won't do anything
- tried to switch back to BTC
- 🛑 won't do anything

packages/suite/src/hooks/wallet/trading/form/useTradingSellForm.ts
adderpositive: When `sellInfo` is loaded we want to change `defaultValues` of the form, because `defaultValues` are based on data from `sellInfo`. 

I think it would be better to load data in the parent component and ensure data are loaded before this hook. With this solution, we would prevent some weird behaviors.

adderpositive:
> I tried it locally and I don't think it is fixed:
> 
>     * went to BTC account
> 
>     * go to Swap
> 
>     * in Form field switch to USDT on Solana [React Native build on android #1](https://github.com/trezor/trezor-suite/issues/1)
> 
>     * tried to switch to other token from different account
> 
>     * 🛑 won't do anything
> 
>     * tried to switch back to BTC
> 
>     * 🛑 won't do anything

It works for me 😁

adderpositive:
> > I tried it locally and I don't think it is fixed:
> > ```
> > * went to BTC account
> > 
> > * go to Swap
> > 
> > * in Form field switch to USDT on Solana [React Native build on android #1](https://github.com/trezor/trezor-suite/issues/1)
> > 
> > * tried to switch to other token from different account
> > 
> > * 🛑 won't do anything
> > 
> > * tried to switch back to BTC
> > 
> > * 🛑 won't do anything
> > ```
> 
> It works for me 😁

Also here it works 
https://dev.suite.sldev.cz/suite-web/fix/selecting-account/web/

vojtatranta:
> > > I tried it locally and I don't think it is fixed:
> > > ```
> > > * went to BTC account
> > > 
> > > * go to Swap
> > > 
> > > * in Form field switch to USDT on Solana [React Native build on android #1](https://github.com/trezor/trezor-suite/issues/1)
> > > 
> > > * tried to switch to other token from different account
> > > 
> > > * 🛑 won't do anything
> > > 
> > > * tried to switch back to BTC
> > > 
> > > * 🛑 won't do anything
> > > ```
> > 
> > 
> > It works for me 😁
> 
> Also here it works https://dev.suite.sldev.cz/suite-web/fix/selecting-account/web/

Maybe it's broken on special accounts with lots and lots of tokens and coins.


https://github.com/user-attachments/assets/b9608173-883c-4c5f-8994-2aa642dddcc7

vojtatranta:
Yeah, I tried other account - the one on the dev "all seed" with t he passphrase and there it seems to be working, which is great. But it doens't work on that largeaccount tho.

adderpositive:
> Yeah, I tried other account - the one on the dev "all seed" with t he passphrase and there it seems to be working, which is great. But it doens't work on that largeaccount tho.

Try to Reset app in settings and again please.

tomasklim (changes requested):
It works ok when device is connected, it does not work when device is disconnected but maybe that's another issue?

adderpositive:
🤷‍♂️

https://github.com/user-attachments/assets/14d11316-40cd-4ee8-b0c4-951b2ad5fd13

misartomas:
It works correctly on link here: https://dev.suite.sldev.cz/suite-web/fix/selecting-account/web/ - both with connected and disconnected Trezor device.

Connected:
https://github.com/user-attachments/assets/d9818b79-f214-4385-ad48-7ae862cc6275

Disconnected:
https://github.com/user-attachments/assets/1ba64cea-8e0c-4e5c-a892-9219f2c1bead

Ondra-Zik:
Seems to work correctly for me as well for both connected and disconnected devices.

tomasklim (approved):

PR #17547, opened 2025-03-10
feat(suite-native): set trading environment

### LINKED ISSUE #17136: Mobile - Add Invity API picker to Dev utils
Allow selecting the API address from DEV utils, similar to the desktop app.

Image

vytick:
Add UI picker for environment, save it to trading state and export via extra dependencies to use in `loadInitialDataThunk`

## Related Issue

Resolve #17136

## Screenshots:

https://github.com/user-attachments/assets/64c89da5-59a6-447a-9c51-c829cb7a61fb


suite-native/module-trading/src/__tests__/tradingSlice.test.ts
```diff
@@ -191,4 +194,26 @@ describe('tradingSlice', () => {
...
+            ['dev' as const],
+            ['localhost' as const],
+        ])('should correctly select trading environment: %s', environment => {
+            const state = {
+                ...initialState,
+                tradingEnvironment: environment,
+            };
+
+            expect(selectTradingEnvironment({ wallet: { trading: state } })).toBe(environment);
+        });
```
jbazant: **nitpick (non-blocking):** There is probably no need to have 4 test that say `if 'a' is equal 'X' then 'b()' returns 'X'`, but overaly great coverage!

suite-native/module-trading/src/__tests__/tradingSlice.test.ts
```diff
@@ -191,4 +194,26 @@ describe('tradingSlice', () => {
             });
         });
     });
+
+    describe('tradingEnvironment', () => {
+        it('should have production as initial trading environment', () => {
+            const state = tradingReducer(undefined, { type: 'undefined_action' });
+
+            expect(selectTradingEnvironment({ wallet: { trading: state } })).toBe('production');
+        });
```
jbazant, line 204: **question:** Is this really expected behaviour? I would expect that dev app should connect to dev api, staging one to staging api, etc.

jbazant (approved):
Looks great! Love those tests.

suite-native/module-trading/src/__tests__/tradingSlice.test.ts
vytick: Even all blockchains are production mainnets. I think the best approach is that user/tester has to deliberately change the environment in case he wants to test with non production api. 

Also the base setup of api is to use production so it just mimics the behaviour on desktop.

suite-native/module-trading/src/tradingSlice.ts
```diff
@@ -25,10 +26,14 @@ export type TradingRootState = {
     };
 };
 
+export const tradingEnvironments = ['production', 'staging', 'dev', 'localhost'] as const;
```
adderpositive: Could you consider using `invityAPI.SERVERS` to reduce duplication?

suite-native/module-trading/src/tradingSlice.ts
```diff
@@ -86,3 +95,6 @@ export const selectIsTradingFavouriteAsset = createMemoizedSelector(
 
 export const selectBuySelectedReceiveAccount = (state: TradingRootState) =>
     selectTradingBuy(state).selectedReceiveAccount;
+
+export const selectTradingEnvironment = (state: TradingRootState) =>
+    state.wallet.trading.tradingEnvironment;
```
adderpositive, line 98: `tradingNew` instead of `trading` should be here

suite-native/module-trading/src/tradingSlice.ts
```diff
@@ -86,3 +95,6 @@ export const selectIsTradingFavouriteAsset = createMemoizedSelector(
 
 export const selectBuySelectedReceiveAccount = (state: TradingRootState) =>
     selectTradingBuy(state).selectedReceiveAccount;
+
+export const selectTradingEnvironment = (state: TradingRootState) =>
```
adderpositive, line 97: Also, this should be in `trading` package. Or you can use  `store.extraDependencies`. There should be this selector.

suite-native/module-trading/src/tradingSlice.ts
```diff
@@ -25,10 +26,14 @@ export type TradingRootState = {
     };
 };
 
+export const tradingEnvironments = ['production', 'staging', 'dev', 'localhost'] as const;
+export type TradingEnvironment = (typeof tradingEnvironments)[number];
```
adderpositive: it is `InvityServerEnvironment`, why not use it?

adderpositive (changes requested):

suite-native/module-trading/src/tradingSlice.ts
vytick: good point 👍

suite-native/module-trading/src/tradingSlice.ts
vytick: updated

suite-native/module-trading/src/tradingSlice.ts
vytick: I'll move it in tne next PR as we discussed

suite-native/module-trading/src/tradingSlice.ts
vytick: this is source for extra deps

suite-native/module-trading/src/tradingSlice.ts
vytick: updated

adderpositive (approved):

PR #17556, opened 2025-03-11
fix(suite-native): Show FW update banner only if the Trezor is connected

### LINKED ISSUE #16476: Show firmware update alert

matejkriz:
## Description

Show FW update banner only if the Trezor is connected and has backup to unify the conditions with Device Settings screen.

Followup for https://github.com/trezor/trezor-suite/pull/17538#issuecomment-2713325842

## Related Issue

Fixes https://github.com/trezor/trezor-suite/issues/16476

## Screenshots:
BEFORE with disconnected view-only device:

https://github.com/user-attachments/assets/ca47fcf6-1e93-47e5-b8b7-062abf4425ac

### DIFF
--- a/suite-native/module-home/src/screens/HomeScreen/components/FirmwareUpdateAlert.tsx
+++ b/suite-native/module-home/src/screens/HomeScreen/components/FirmwareUpdateAlert.tsx
@@ -12,7 +12,7 @@ import {
     selectDeviceReleaseInfo,
     selectDeviceState,
     selectDeviceUpdateFirmwareVersion,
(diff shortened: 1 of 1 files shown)


Nodonisko:
Good catch, I am guessing why that previous selector did not worked because it should not return true for not connected device.

Nodonisko (approved):

PR #17561, opened 2025-03-11
fix: use DEB package as default for udev rules installation

### LINKED ISSUE #16566: UDEV rules installation
**Describe the bug**
Follow up of https://github.com/trezor/trezor-suite/issues/15074 issue
Please make DEB as default package that will be offered as its most used package system

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite/issues/16566
 
image

### DIFF
--- a/packages/connect/src/data/udevInfo.ts
+++ b/packages/connect/src/data/udevInfo.ts
@@ -13,22 +13,22 @@ const info: UdevInfo = {
             name: 'DEB package',
             platform: ['deb32', 'deb64'],
             url: '/udev/trezor-udev_2_all.deb',
+            preferred: true, // DEB package is the most common
         },
     ],
 };
 
-export const suggestUdevInstaller = (platform?: string) => {
-    const info2 = info;
-    // check if preferred field was already added
-    if (!info2.packages.find(p => p.preferred)) {
-        if (platform) {
-            // override UdevInfo packages, add preferred field
-            info2.packages = info2.packages.map(p => ({
-                ...p,
-                preferred: p.platform.indexOf(platform) >= 0,
-            }));
-        }
+export const suggestUdevInstaller = (platform?: string): UdevInfo => {
+    // `platform` is not available on the desktop, in that case we assume the preferred as DEB
+
+    if (platform !== undefined) {
+        return {
+            packages: info.packages.map(it => ({
+                ...it,
+                preferred: it.platform.indexOf(platform) >= 0,
+            })),
+        };
     }
 
-    return info2;
+    return info;
 };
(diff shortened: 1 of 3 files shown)


packages/suite/src/components/suite/UdevDownload.tsx
```diff
@@ -50,13 +50,17 @@ interface Installer {
...
 
+    const hasPreferred = udev !== undefined && udev.packages.some(p => p.preferred);
+
     const installers: Installer[] = udev
         ? udev.packages.map(p => ({
               label: p.name,
               value: DATA_URL + p.url.substring(1),
-              preferred: p.preferred, // This is not available on the desktop
+              // preferred is not available on the desktop, so we fallback on the DEB package as it is the most common
+              preferred: hasPreferred ? p.preferred : p.name === 'DEB package',
```
peter-sanderson: There is a `suggestUdevInstaller` where it would be better to solve it. But it is weird and somehow [counts on the modification of the `info` array outside of this scope](https://github.com/trezor/trezor-suite/blob/46a8e661a3ee6495866d7ff483173be89028b51a/packages/connect/src/data/udevInfo.ts#L22)

So I rather "fixed" it this suboptimal way. 

I also had to hardcode `'DEB package'` as it is defined in the `connect` and we are not suppose to export enything from connect beside types and the `TrezorConnect` itself.

Lemonexe: I just asked the Connect team, and that function could be moved to `@trezor/utils` or something like that. It serves only for Suite, Popup does not implement that UI yet, which means you can even modify it to not mutate `info`, without fear of breaking something :slightly_smiling_face: 
Would you please do that?  It's better than hardcoding it.

peter-sanderson: Fixed, check last commit

packages/utils/src/suggestUdevInstaller.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * This type must match the UdevInfo in the packages/connect/src/events/transport.ts
+ */
+export interface UdevInfo {
```
mroz22: this file has been moved outside of connect only to be imported there back again?

Lemonexe: as far as I can see, the `UdevInfo` type is duplicated, as an unfortunate consequence, but the `suggestUdevInstaller` was exported so that it could be used in Suite, right? :slightly_smiling_face:

packages/connect/src/data/udevInfo.ts
```diff
@@ -13,22 +13,22 @@ const info: UdevInfo = {
             name: 'DEB package',
             platform: ['deb32', 'deb64'],
             url: '/udev/trezor-udev_2_all.deb',
+            preferred: true, // DEB package is the most common
```
peter-sanderson, line 16: I have asked @mroz22 and this is OK

mroz22 (approved):

Lemonexe (approved):

PR #17578, opened 2025-03-11
feat(transport-bridge): start logging libusb errors

mroz22:
useful when debugging with standalone node bridge. can't be used for the build-in version at the moment since it runs in a utility process


marekrjpolak (approved):
LGTM :rainbow_flag:

PR #17584, opened 2025-03-11
fix(device-utils): add tsx as dev dependency

karliatto:
## Description
 add tsx as dev dependency


mroz22 (approved):
yarn lock?

karliatto:
> yarn lock?

Yes, added.

PR #17589, opened 2025-03-12
http server improvements

mroz22:
## QA testing notes

please test that all endpoints still work

- /exchange-redirect
- /sell-redirect
- /buy-post
- /buy-redirect
- /oauth

### DIFF
--- a/packages/node-utils/src/http.ts
+++ b/packages/node-utils/src/http.ts
@@ -1,3 +1,4 @@
+import { sanitizeUrl } from '@braintree/sanitize-url';
 import * as http from 'http';
 import * as net from 'net';
 import * as url from 'url';
@@ -288,7 +289,23 @@ export class HttpServer<T extends EventMap> extends TypedEmitter<T & BaseEvents>
             this.logger.info(`Request ${request.method} ${request.url} aborted`);
         });
 
-        const { pathname } = url.parse(request.url, true);
+        const { protocol, hostname, pathname, query } = url.parse(request.url, true);
+
+        if (query) {
+            for (const key in query) {
+                if (Object.prototype.hasOwnProperty.call(query, key)) {
+                    const sanitized = sanitizeUrl(query[key] as string);
+                    if (sanitized !== query[key]) {
+                        response.statusCode = 403;
+
+                        return response.end();
+                    }
+                    query[key] = sanitized;
+                }
+            }
+        }
+        request.url = url.format({ protocol, hostname, pathname, query });
+
         if (!pathname) {
             const msg = `url ${request.url} could not be parsed`;
             this.emitter.emit('server/error', msg);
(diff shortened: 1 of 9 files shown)


scripts/list-outdated-dependencies/foundation-dependencies.txt
```diff
@@ -32,6 +32,7 @@ electron-updater
 electron-localshortcut
 openpgp
 systeminformation
+@braintree/sanitize-url
```
Lemonexe, line 35: okay fine, it's a nice library we can take it in :smile: 

On a more serious note, I took a look at the source code and tests, and it looks very robust :+1:

packages/node-utils/src/http.ts
```diff
@@ -288,7 +306,23 @@ export class HttpServer<T extends EventMap> extends TypedEmitter<T & BaseEvents>
...
+            for (const key in query) {
+                if (Object.prototype.hasOwnProperty.call(query, key)) {
+                    const sanitized = sanitizeUrl(query[key] as string);
+                    if (sanitized !== query[key]) {
+                        response.statusCode = 403;
+
+                        return response.end();
+                    }
+                    query[key] = sanitized;
+                }
```
Lemonexe: according to TS, `query[key]` can also be `string[]`.
You get such result when multiple same GET params, which is perfectly valid; e.g.:
`url.parse('http://suite.trezor.io?a=1&a=2&b=3', true)` → `query:{ a: [ '1', '2' ], b: '3' }`
I checked in node repl that `sanitizeUrl` crashes on such array, so the type casting is not correct.

On the other hand, url.format behaves the same for both of these:
```
query= { a: [ '1', '2' ], b: '3' };
query= { a: [ '1', '2' ], b: ['3'] };
```

:arrow_right:  So I think this code is the simplest fix; what do you think? :slightly_smiling_face: 
Also check `undefined`, I think it comes just from TS declaration, not from behavior (`?c` or `?c=` → `''`)
```suggestion
                if (Object.prototype.hasOwnProperty.call(query, key) && query[key] !== undefined) {
                    const allParamsOfSameKey = [query[key]].flat();

                    query[key] = allParamsOfSameKey.map(singleParam => {
                        const sanitized = sanitizeUrl(singleParam);
                        if (sanitized !== singleParam) response.statusCode = 403;

                        return sanitized;
                    });

                    if (response.statusCode === 403) return response.end();
                }
```

Lemonexe (approved):
Codewise LGTM. Tested oauth, works :heavy_check_mark: 
Sanitization is great! :ok: 

Though I don't understand the purpose of the `activation` at all, but I'll ask in person..

I made a PR with suggestions for further TS improvements: #17971
If you like it, merge that here :slightly_smiling_face:

mroz22:
> Though I don't understand the purpose of the activation at all, but I'll ask in person..

the purpose of activation is to expose only endpoints you are using to limit surface of what is visible publicly. It could be improved however https://github.com/trezor/trezor-suite-private/issues/136

packages/node-utils/src/http.ts
mroz22: lgtm

bosomt:
**QA OK**

- oauth/labeling ✅ 
- [trade stuff checked by Invity guys ](https://satoshilabs.slack.com/archives/C04DD84P4A1/p1743506210423509)

**Info:**
 - Suite version: desktop 25.3.3 (317afa284e41770eb472c25a6039c1d3fdccf096)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.3.3 Chrome/132.0.6834.194 Electron/34.1.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: 

PR #17526, opened 2025-03-10
Bump foundation deps (2025.03)

Lemonexe:
## Description

Update most Foundation-related dependencies.

**major version:**
- electron
- @hookform/resolvers
  
**minor version:**
- uuid
- @eslint/js
- eslint
- eslint-plugin-mdx
- eslint-plugin-react-hooks
- typescript-eslint

**patch version:**
none

**not updated:**
- `tiny-secp256k1` TODO in [#12261](https://github.com/trezor/trezor-suite/pull/12261) 
- `electron-store` blocked by [#14482](https://github.com/trezor/trezor-suite/pull/14482)

:information_source: For reference, last bump foundation deps PR was #16880
:eye: I skimmed through all code changes  _except_ eslint-related packages and `electron`; found nothing suspicious ✅ 

## Related Issue

Resolve #17495

## QA

:eye: Besides CI checks, I have tested locally:
- suite dev desktop
  - app builds & runs
  - tor works
  - sign & verify form validation works (invalid address or message too long or empty) 
- suite lite dev
  - any form validation works, for example you can't enter empty passphrase
- suite desktop build linux:
  - app builds & runs
  - app update works

@coderabbitai ignore

PR #17602, opened 2025-03-12
fix(suite-desktop-core): refactor desktop state patches to avoid breaking change

Lemonexe:
## Description

Partial revert of 5a84a1f5 (#17526)
- There, desktop state patch process args were refactored, as they stopped working in Electron 35
  - that was a breaking change for Trezor Suite process args :disappointed: 
- I believed at the time that Electron checks args at startup and throws if there are dots on the left-hand side of assignment in a switch.
- It is not so; only the [app.commandLine.getSwitchValue](https://github.com/trezor/trezor-suite/blob/098a0b3889b4c0ce8263c8592e926657502274cf/packages/suite-desktop-core/src/libs/app-utils.ts#L42) throws. And it's actually not related to dots but uppercase letters!
- But that function call was not needed at all, even in the original implementation.
- In 5a84a1f5, the `processStatePatch` + tests were _also_ refactored for better clarity, so this PR keeps that...
- ...but reverts to the original API :arrow_right: **no breaking change** :slightly_smiling_face: 

### Notes

:point_right: I will still did file a bug with Electron https://github.com/electron/electron/issues/45991, because the function started throwing without mention in changelog. But they said the function is not supposed to be used for custom application parameters, it's an API to set Chromium flags. We are supposed to use `process.argv` for custom params, as it currently is 🙂

### DIFF
--- a/packages/suite-desktop-core/src/__tests__/app-utils.test.ts
+++ b/packages/suite-desktop-core/src/__tests__/app-utils.test.ts
@@ -1,20 +1,20 @@
 import { processStatePatch } from '../libs/app-utils';
 
-const fullState = 'state={ "d": { "e": true, "f": null }, "a.c": "baz" }';
-const partialState1 = 'state=a={"b": [5] }';
-const partialState2 = 'state=a.b=[1, 3, 8]';
-const singleProperty = 'state=a.c=foobar';
-const notInState = 'a=nothing';
-const notAssignment = 'hasNoEqualitySign';
+const fullState = '--state={ "d": { "e": true, "f": null }, "a.c": "baz" }';
+const partialState1 = '--state.a={"b": [5] }';
+const partialState2 = '--state.a.b=[1, 3, 8]';
+const singleProperty = '--state.a.c=foo=bar';
+const notInState = '--a=nothing';
+const notAssignment = '--hasNoEqualitySign';
 
 const FIXTURES: [string, string[], any][] = [
     ['none', [], undefined],
-    ['simple', [partialState2, singleProperty], { a: { b: [1, 3, 8], c: 'foobar' } }],
+    ['simple', [partialState2, singleProperty], { a: { b: [1, 3, 8], c: 'foo=bar' } }],
     ['mixed', [notInState, partialState1, notInState, partialState2], { a: { b: [5, 1, 3, 8] } }],
     [
         'complex',
         [partialState2, notInState, fullState, notAssignment, partialState1, singleProperty],
-        { a: { b: [1, 3, 8, 5], c: 'foobar' }, d: { e: true, f: null } },
+        { a: { b: [1, 3, 8, 5], c: 'foo=bar' }, d: { e: true, f: null } },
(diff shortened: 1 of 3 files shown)


packages/suite-desktop-core/src/__tests__/app-utils.test.ts
```diff
@@ -16,13 +17,13 @@ const FIXTURES: [string, string[], any][] = [
...
     ],
-    ['invalid', [notAssignment, notInState], undefined],
+    ['invalid', [notAssignment, notInState, badAssignment], undefined],
 ];
 
 describe('process state patch', () => {
     FIXTURES.map(([name, args, value]) => {
         it(name, () => {
-            process.argv = args.map(arg => `--${arg}`);
+            process.argv = args;
```
Lemonexe, line 25: I think mapping the `--` just obfuscated the test flow, so I simplified it even further..

packages/suite-desktop-core/src/libs/app-utils.ts
```diff
@@ -39,24 +39,24 @@ const tryParseJson = (value: string) => {
...
-        .filter(arg => arg !== null)
-        .map(match => parseAssignment(match[1]))
+        .filter(arg => STATE_ASSIGNMENT_REGEX.test(arg))
+        .map(arg => arg.slice(2))
+        .map(parseAssignment)
         .reduce<ProcessStatePatchResult>(
             (prev, cur) => mergeDeepObject.withOptions({ dotNotation: true }, prev ?? {}, cur),
             undefined,
-        );
+        )?.state;
```
Lemonexe, line 56: with this implementation, everything is under `{ state: {...patches}}` until this line, where I extract it or dump the `undefined`..

packages/suite-desktop-core/src/libs/app-utils.ts
```diff
@@ -39,24 +39,24 @@ const tryParseJson = (value: string) => {
...
-    const keyValuePair = assignment.split('=');
-    if (keyValuePair.length !== 2) return tryParseJson(keyValuePair[0]);
-    const [key, value] = keyValuePair;
+    // compliance of `assignment` with STATE_ASSIGNMENT_REGEX guarantees exactly two elements
+    const [key, value] = assignment.split('=');
 
     return { [key]: tryParseJson(value) };
 };
 
+const STATE_ASSIGNMENT_REGEX = /^--state[^=]*=[^=]+$/;
```
marekrjpolak: The only tiny thing which came to my mind is that this regex disallow to pass a json value containing string with equal sign, which in theory may be wanted.

Lemonexe: True! I'll fix that.

Lemonexe: Done. Now I only split by the _first_ equality sign, so I changed `.test`+`.split` back to `.match` with two capturing groups

marekrjpolak (approved):

PR #17606, opened 2025-03-12
feat(e2e): Canary FW CI pipelines for playwright tests

HajekOndrej:
## Description

Added jobs to test with main firmware.

## Related Issue

Resolve

### DIFF
--- a/.github/workflows/test-suite-web-e2e-fw-canary.yml
+++ b/.github/workflows/test-suite-web-e2e-fw-canary.yml
@@ -1,4 +1,4 @@
-name: "[Test] Nightly Suite Web e2e tests"
+name: "[Test] Firmware canary  Suite Web e2e tests"
 # Nightly scheduled check on develop branch
 # Builds/deploys the web app to dev.suite.sldev.cz/suite-web/develop/web
 # Runs full e2e test suite for Suite Web
--- a/.github/workflows/test-suite-web-e2e-fw-canary.yml
+++ b/.github/workflows/test-suite-web-e2e-fw-canary.yml
@@ -10,7 +10,7 @@ permissions:
 
 on:
   schedule:
-    - cron: "0 0 * * *"
+    - cron: "0 1 * * *"
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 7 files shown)


.github/workflows/test-suite-web-e2e-fw-canary.yml
```diff
@@ -0,0 +1,81 @@
+name: "[Test] Nightly Suite Web e2e tests"
```
Vere-Grey: Wrong title

.github/workflows/test-suite-web-e2e-fw-canary.yml
```diff
@@ -0,0 +1,81 @@
...
+          ref: ${{ github.sha }}
+          fetch-depth: 2
+
+      - name: Run Playwright E2E Tests
+        uses: ./.github/actions/run-e2e-tests
+        env:
+          CANARY_FIRMWARE: "true"
+        with:
+          project: "web"
+          base-url: "https://dev.suite.sldev.cz/suite-web/fw-canary-788b960e2c6b5112230ded6129d69538/web/"
```
Vere-Grey, line 74: What is this magical number?

packages/trezor-user-env-link/src/api.ts
```diff
@@ -202,6 +202,7 @@ export class TrezorUserEnvLinkClass extends TypedEmitter<WebsocketClientEvents>
             ...arg,
         };
 
+        console.log('Starting emulator with params', JSON.stringify(params));
```
Vere-Grey, line 205: This won't be easily readable in currents or trace
Maybe pushing it to annotation would be a good way how to deal with this.
Not necessarily in this PR

Vere-Grey (changes requested):
Forgotten copy pasted title
one question and one optional suggestion

.github/workflows/test-suite-web-e2e-fw-canary.yml
HajekOndrej: hash I generated to make sure it never collides with anyones branch...

packages/trezor-user-env-link/src/api.ts
HajekOndrej: It would be cool to have it there, but I don't if it would work well with tests that start the emu multiple times.

HajekOndrej: Let's try :)

HajekOndrej: Actually... let's not :D I guess I don't know how to do it from the tenv api class... I would need testinfo there...

Vere-Grey: Hmm, I have no idea if same key annotation overrides itself or duplicates itself

Vere-Grey (approved):
:+1:

PR #17608, opened 2025-03-12
Protobuf types improvement

marekrjpolak:
## Description

Generate better types for device messages in `@trezor/protobuf` package and use them in connect instead of locally (re)defined ones.


martykan (approved):
LGTM, nice improvement

PR #17613, opened 2025-03-12
fix(suite-native): send max button appears on change

### LINKED ISSUE #17604: Send max button sometimes doesn`t work in Suite lite
**Describe the bug**
The "Send max" button in send form sometimes doesn`t show up for some reason. (Sadly there no repro steps yet.)

**Steps to reproduce:**
1. a
2. b
3. c

**Info:**
 - Suite version: 25.2.1

PeKne:
## Description
- The send max button is displayed whenever the address input is valid. It used to be displayed only if the field was marked as `isTouched`, which made the button appear unexpectedly. 

## Related Issue

Resolve #17604 

## Screenshots:

https://github.com/user-attachments/assets/76bdd053-2071-4ca3-a496-f25733f33cd5


matejkriz (approved):
Makes sense, good catch 👍

PR #17630, opened 2025-03-13
fix(suite): TradingFormOffer small polishing

jvaclavik:
## Description

- update alignment
- footer is not ugly as before
- delete unused component

don't forget:
Screenshot 2025-03-13 at 9 43 33

## Related Issue

Resolve 

## Screenshots:

before

Screenshot 2025-03-13 at 9 40 55

after

image


vojtatranta (approved):
Looks good (I mean like the UI)

packages/suite/src/views/wallet/trading/common/TradingForm/TradingFormOffer.tsx
```diff
@@ -82,71 +82,75 @@ export const TradingFormOffer = () => {
...
-                    <Paragraph typographyStyle="label" variant="tertiary">
-                        <Translation
-                            id="TR_TRADING_ON_NETWORK_CHAIN"
-                            values={{
-                                networkName: network,
+        <Column gap={spacings.lg} justifyContent="space-between" height="100%">
+            <div>
+                <Column gap={spacings.xs} data-testid="@trading/best-offer">
+                    <Translation id={amountLabels.offerLabel} />
+                    {shouldDisplayFiatAmount ? (
```
vojtatranta, line 89: Yeah, I think `<If><Else>` would be readable, I'll add it.

packages/suite/src/views/wallet/trading/common/TradingFooter/TradingFooter.tsx
```diff
@@ -12,12 +12,7 @@ import { TradingFooterLogoWrapper } from 'src/views/wallet/trading';
 import { TradingProvidedByInvity } from 'src/views/wallet/trading/common/TradingFooter/TradingProvidedByInvity';
 
 const Wrapper = styled.div`
-    margin-top: ${spacingsPx.xxxl};
-`;
-
-const WrapperBorder = styled.div`
-    padding-top: ${spacingsPx.lg};
-    border-top: 1px solid ${({ theme }) => theme.borderElevation1};
+    margin-top: ${spacingsPx.xl};
```
adderpositive, line 15: Could the `Row` component replace this custom style?

adderpositive: Or in this PR be replaced with other custom styles? Will you have some time for that? 🫣🙏

adderpositive (changes requested):
obrazek

CTA button should not be on the end of the box.

packages/suite/src/views/wallet/trading/common/TradingFooter/TradingFooter.tsx
seibei-iguchi, line 15: Margin top shouldn't be in this context, imo, but in the parent component where the layout composition occurs, via column or something like that.

jvaclavik:
looks like it needs some love from designer first

PR #17641, opened 2025-03-13
chore(message-system): add update banner for t1b1

MiroslavProchazka:
Tracked in Notion [here](https://www.notion.so/satoshilabs/T1B1-update-banner-1b6dc526060680d6be4af9f0884b91f3?pvs=4)

This pull request includes updates to the `suite-common/message-system/config/config.v1.json` file to modify the timestamp and sequence number, and to add new message conditions and content for firmware updates.

Updates to configuration:

* Updated the `timestamp` to "2025-03-13T00:00:00+00:00" and the `sequence` to 83.

New message conditions and content:

* Added a new message condition for environments with desktop and web versions greater than 25.2.2 and mobile excluded, targeting specific device models and firmware versions. The message informs users about the availability of new Trezor firmware and prompts them to update their device.
* Added another message condition for environments with mobile version 25.2.1 and desktop and web excluded, targeting the same device models and firmware versions. The message content is similar, informing users about the new Trezor firmware and urging them to update their device.

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
-    "timestamp": "2025-02-24T00:00:00+00:00",
-    "sequence": 82,
+    "timestamp": "2025-03-13T00:00:00+00:00",
+    "sequence": 83,
     "actions": [
         {
             "conditions": [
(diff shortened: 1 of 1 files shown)


suite-common/message-system/config/config.v1.json
```diff
@@ -564,6 +564,76 @@
...
+                    "hu": "Új Trezor firmware érhető el! Kérjük, frissítse a készülékét.",
+                    "it": "È disponibile un nuovo firmware Trezor! Si prega di aggiornare il dispositivo.",
+                    "fr": "Un nouveau firmware Trezor est disponible ! Veuillez mettre à jour votre appareil.",
+                    "de": "Neue Trezor-Firmware verfügbar! Bitte aktualisieren Sie Ihr Gerät.",
+                    "tr": "Yeni Trezor yazılımı mevcut! Lütfen cihazınızı güncelleyin.",
+                    "pt": "Está disponível um novo firmware Trezor! Por favor, atualize o seu dispositivo.",
+                    "uk": "Доступна нова прошивка Trezor! Будь ласка, оновіть ваш пристрій."
+                },
+                "cta": {
+                    "action": "internal-link",
```
matejkriz, line 617: This is not implemented on mobile 🙉

MiroslavProchazka: Solved by having different message for mobile

STew790:
**QA OK**
I can see the banner. 
Desktop:
image

Mobile:
image

suite-common/message-system/config/config.v1.json
```diff
@@ -564,6 +564,127 @@
                 }
             }
         },
+        {
+            "conditions": [
+                {
+                    "environment": {
+                        "desktop": ">25.2.2",
```
komret: Shouldn't this be `>=25.2.1`? We released `1.12.1` in `25.2.1`, but this will target `25.3.1` or newer.

suite-common/message-system/config/config.v1.json
```diff
@@ -564,6 +564,127 @@
...
+                        {
+                            "model": "T1B1",
+                            "firmware": "1.12.1",
+                            "bootloader": "*",
+                            "variant": "*",
+                            "firmwareRevision": "*",
+                            "vendor": "*"
+                        },
+                        {
+                            "model": "1",
```
komret, line 585: This might be redundant, devices with firmware `1.12.1` probably always report as `T1B1`. But needs to be checked.

suite-common/message-system/config/config.v1.json
```diff
@@ -564,6 +564,127 @@
...
+                    }
+                }
+            }
+        },
+        {
+            "conditions": [
+                {
+                    "environment": {
+                        "desktop": "!",
+                        "mobile": "25.2.1",
```
komret, line 642: Should be `25.2.1` or newer?

komret (changes requested):

suite-common/message-system/config/config.v1.json
MiroslavProchazka: 25.2.1 was EAP, it didnt make it into production

komret: Anyway, no harm including it. Someone could stay on that version. And 25.2.2 should be included for sure.

suite-common/message-system/config/config.v1.json
MiroslavProchazka: just 25.2.1, 
- 25.1.2 does not have 1.13.0
- 25.3.1 will have native banner, so it would be duplicated

suite-common/message-system/config/config.v1.json
MiroslavProchazka: Will do

suite-common/message-system/config/config.v1.json
matejkriz: This depends on Suite (Connect), not the FW, right? So I believe it's redundant here and T1B1 is enough, but better to test it :-D

matejkriz (approved):

matejkriz:
I'm just missing an entry here https://www.notion.so/satoshilabs/9c2ae323cc704ec3b1d2fdcf7fbb5a5f?v=f48f039f9db9423ab0369bc932f6355b&pvs=4

suite-common/message-system/config/config.v1.json
matejkriz: Better safe than sorry, it can stay here.

komret (approved):

PR #17642, opened 2025-03-13
feat(connect): add appName, appIcon to manifest

martykan:
## Description

Add `appName`, `appIcon` to Connect manifest. Remove `hostIcon` which was never used.

### DIFF
--- a/packages/connect-explorer/src/pages/index.mdx
+++ b/packages/connect-explorer/src/pages/index.mdx
@@ -196,6 +196,9 @@ export const ExampleHeading = styled.h3`
                     // TrezorConnect.method is called
                     manifest: {
                         email: 'developer@xyz.com',
+                        appName: 'Your Application',
+                        // svg or png icon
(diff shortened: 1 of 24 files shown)


packages/connect-explorer/src/pages/index.mdx
```diff
@@ -196,6 +196,9 @@ export const ExampleHeading = styled.h3`
                     // TrezorConnect.method is called
                     manifest: {
                         email: 'developer@xyz.com',
+                        appName: 'Your Application',
+                        // svg or png icon
+                        appIcon: 'http://your.application.com/icon.svg',
```
mroz22, line 201: I can't see it being used anywhere at the moment. is that right? If yes, I would mention it in a comment that it might become used in the future maybe?

martykan: It will be in Suite settings and modals, we're just waiting for the image proxy

mroz22 (approved):

PR #17655, opened 2025-03-13
feat(native-suite): handle device change in settings

### LINKED ISSUE #17594: Handle Device Mismatch: Redirect User to Home Screen if a Different Device Is Connected
Currently, the app does not handle the case where a different device is connected instead of the previously stored one. 

Expected Behavior:
 • if the connected device is different from the stored one, the user should be redirected to Home Screen

BrantalikP:
## Description
redirect the user to the Home Screen if another device is connected

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17594

## Screenshots:

https://github.com/user-attachments/assets/55230ff8-65d4-453f-bec5-52bf48aed523

### DIFF
--- a/suite-native/module-device-settings/src/hooks/useDeviceChangedCheck.ts
+++ b/suite-native/module-device-settings/src/hooks/useDeviceChangedCheck.ts
@@ -1,0 +1,46 @@
+import { useEffect, useRef } from 'react';
+import { useSelector } from 'react-redux';
+
+import { useNavigation } from '@react-navigation/native';
+
+import { selectSelectedDevice } from '@suite-common/wallet-core';
+import {
+    AppTabsRoutes,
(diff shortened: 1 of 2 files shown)


matejkriz (approved):
Originally, I had the impression that it would get stuck when I had a PIN set on my Trezor and would be stuck on the home screen, but I was probably just impatient—now it behaves fine.

While testing, I also discovered that the PIN screen is disabled in the entire Device Settings, but it’s not much of an issue because changing the PIN works correctly, firmware updates can be done even with the Trezor locked (as it switches to bootloader mode), and Device Authenticity displays the “Continue on your Trezor” screen, which correctly redirects the user’s attention to enter the PIN. (Device Authenticity is not available on the T1, so there’s no need to enter the PIN through the app.)

PR #17600, opened 2025-03-12
feat: update shared state fro bluetooth to accomodate changes in the ipc

peter-sanderson:
Updates the shared code for handling a Bluetooth devices.

PR #17662, opened 2025-03-14
Feat/bluetooth update shared state

peter-sanderson:
Part of: https://github.com/trezor/trezor-suite/issues/15390


szymonlesisz:
what is the point for adding test in the first commit and then remove it in the second commit? 
wouldn't it be better to squash them into one?

additional nitpick: first commit message has a typo and it is the same as [already merged here](https://github.com/trezor/trezor-suite/pull/17600/commits)

szymonlesisz (approved):

PR #17666, opened 2025-03-14
feat(mobile): temporary remember device during fw update

### LINKED ISSUE #16460: Fix device temporary remembered during FW update

Nodonisko:
## Description

If user doesn't have remembered device, discovery will run again after FW update/reinstall. This PR should prevent that by temporarily remembering device. I was thinking that desktop is doing something similar but after some investigation I found out that desktop is simply running discovery again after FW update.

QA:
1. Discovery doesn't run again after FW update for non-remembered devices
2. Non remembered device should correctly should dissappear from app by disconnecting after FW update
3. Normal remembered device should stay remembered after FW update and after disconnecting
4. (edge case) if you kill app during FW update it should forget temporary reremembered device and do not forget normally remembered device
5. FW installation during onboarding works as before
6. Would be nice to test at least few first points also when using passphrase wallet

## Related Issue

Resolve #16460

### DIFF
--- a/suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
+++ b/suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
@@ -13,7 +13,7 @@ import { useNavigation } from '@react-navigation/native';
 
 import { authorizeDeviceThunk } from '@suite-common/wallet-core';
 import { Box, Button, IconButton, Text, VStack } from '@suite-native/atoms';
-import { ConfirmOnTrezorImage, setDeviceForceRememberedThunk } from '@suite-native/device';
+import { ConfirmOnTrezorImage, setTemporaryRememberedDeviceThunk } from '@suite-native/device';
 import { requestPrioritizedDeviceAccess } from '@suite-native/device-mutex';
 import { Translation } from '@suite-native/intl';
 import { SUITE_LITE_SUPPORT_URL, useOpenLink } from '@suite-native/link';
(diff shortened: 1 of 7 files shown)


suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
```diff
@@ -87,14 +89,16 @@ export const FirmwareInstallationScreenContent = ({
     const openLink = useOpenLink();
 
     useEffect(() => {
+        if (!isTemporaryRememeberAllowed) return;
+
         // This will prevent device from being forgotten after firmware update, so discovery will not run again
-        dispatch(setDeviceForceRememberedThunk({ forceRemember: true }));
```
BrantalikP, line 91: I think we can remove the Thunk if its not used anywhere `setDeviceForceRememberedThunk`

Nodonisko: it was renamed to `setTemporaryRememberedDeviceThunk`

BrantalikP: sorry my bad, the search took me to `libDev` 🤦‍♂️

BrantalikP (approved):
looks great. Good job!

STew790:
**QA OK**

1. OK
2. OK
3. OK
4. OK
5. Will check with onboarding issues but so far lgtm
6. Checked all of those with passphrase as well and rememberes correctly both passphrase and standard 👍 

**Info**
25.3.2 https://github.com/trezor/trezor-suite/commit/bc3e808263a13c72c7fe2a5d3d3f5d44709dddc1

PR #17673, opened 2025-03-14
fix(suite-desktop-core): purge userData by using a flag instead of in playwright

Lemonexe:
## Description

Attempt to fix desktop E2E instability because of errors like this that popup when calling `electronApp.evaluate`:
```
Error: electronApplication.evaluate: Execution context was destroyed, most likely because of a navigation.
Error: electronApplication.firstWindow: Target page, context or browser has been closed
```
The `evaluate` fn serves to inject code from Playwright process to Electron main, we use it to purge app data on startup. But that can be achieved in a more straightforward way:
- introduce `--remove-user-data-on-start` flag for Suite Desktop, which purges app data on startup
- use the flag in Playwright

### DIFF
--- a/packages/suite-desktop-core/e2e/support/electron.ts
+++ b/packages/suite-desktop-core/e2e/support/electron.ts
@@ -1,5 +1,5 @@
 import { ElectronApplication, Page, _electron as electron } from '@playwright/test';
-import { createWriteStream, ensureDirSync, readdirSync, removeSync } from 'fs-extra';
+import { createWriteStream, ensureDirSync } from 'fs-extra';
 import path from 'path';
 
 import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
@@ -61,6 +61,7 @@ export const launchSuiteElectronApp = async (params: LaunchSuiteParams) => {
             disableHWAccelerationArgument,
             ...(options.bridgeLegacyTest ? ['--bridge-legacy', '--bridge-test'] : []),
             ...(options.bridgeDaemon ? ['--bridge-daemon', '--skip-new-bridge-rollout'] : []),
+            ...(options.rmUserData ? ['--remove-user-data-on-start'] : []),
             disableHashCheckStatePatch,
             showDebugMenuStatePatch,
         ],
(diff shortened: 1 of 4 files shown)


packages/suite-desktop-core/e2e/support/electron.ts
```diff
@@ -97,17 +82,6 @@ export const launchSuiteElectronApp = async (params: LaunchSuiteParams) => {
...
-    await electronApp.evaluate(
-        (_, [resourcesPath]) => {
-            // This runs in the main Electron process.
-            // override global variable defined in app.ts
-            global.resourcesPath = resourcesPath;
-
-            return global.resourcesPath;
-        },
-        [path.join(appDir, 'build/static')],
-    );
```
Lemonexe, line 109: This evaluation seems to be useless so I'm removing without replacement. At here, the `resourcesPath` resolves to the same path as the one used _by default (when dev env)_ [in electron-main](https://github.com/trezor/trezor-suite/blob/9cba267bec7c341db4d30934383a1407c57285bc/packages/suite-desktop-core/src/app.ts#L32-L34), which is:
`/my_repo_path/packages/suite-desktop/build/static`

packages/suite-desktop-core/src/libs/app-utils.ts
```diff
@@ -54,3 +57,18 @@ export const processStatePatch = (): ProcessStatePatchResult =>
             (prev, cur) => mergeDeepObject.withOptions({ dotNotation: true }, prev ?? {}, cur),
             undefined,
         )?.state;
+
+export const removeElectronAppData = () => {
+    const localDataDir = app.getPath('userData');
```
Lemonexe, line 62: FYI note: the path differs between standalone application and playwright. At **Linux Ubuntu** it is:
Playwright: `~/.config/Electron`
Standalone app: `~/.config/@trezor/suite-desktop` [or these variants](https://github.com/trezor/trezor-suite/blob/56895b234fa5e63297f97c5e985fddb3429d89f3/docs/packages/suite-desktop.md#app-id-and-name-by-environment)

packages/suite-desktop-core/src/libs/app-utils.ts
```diff
@@ -1,3 +1,6 @@
+import fs from 'fs';
```
Lemonexe, line 1: using `fs` instead of the original `fs-extra` to avoid bundling another runtime dependency

packages/suite-desktop-core/src/app.ts
```diff
@@ -33,6 +33,13 @@ global.resourcesPath = isDevEnv
     ? path.join(__dirname, '..', 'build', 'static')
     : process.resourcesPath;
 
+const parseRemoveUserDataSwitch = () => {
+    if (process.argv.includes('--remove-user-data-on-start')) {
+        removeElectronAppData();
+    }
+};
+parseRemoveUserDataSwitch();
```
Lemonexe, line 41: found out the hard way that this has to be done ASAP.
At first I called the function at ['init' (L327)](https://github.com/trezor/trezor-suite/blob/91c52a9aae64257058a42836bed91ba129ab0aeb/packages/suite-desktop-core/src/app.ts#L327), but that's too late; it caused the `spawn-bridge-daemon.test.ts` to crash when launchin Suite the _second time_ (this time with UI, while first time was headless daemon mode) :see_no_evil:

Vere-Grey (approved):
:saluting_face:

PR #17681, opened 2025-03-14
Default feature flag values on develop same as on production

matejkriz:
## Description

- To reduce the risk of errors and simplify testing, the default values of feature flags are the same as in production. Default values can be changed locally for development using environment variables.


jbazant:
**Question:** How is this supposed to work in CI?

E.g. for trading there is lots of tests that expect to have trading enabled. On other hand we expect trading to be still disabled in production.

If the answer is that for CI we should move configuration to `env` isn't that the same as having it directly in code just with extra steps?

suite-native/module-dev-utils/src/components/FeatureFlags.tsx
```diff
@@ -21,7 +22,7 @@ const FeatureFlag = ({ featureFlag }: { featureFlag: FeatureFlagEnum }) => {
 
     return (
         <Box flexDirection="row" justifyContent="space-between">
-            <Text>{featureFlagsTitleMap[featureFlag]}</Text>
+            <Text>{`${featureFlagsTitleMap[featureFlag]} [${featureFlagsInitialState[featureFlag]}]`}</Text>
```
matejkriz, line 25: This might be more confusing than helping, because it shows the initial state based on environment (so If you override something locally, you will see your default values in brackets).

image

BrantalikP: nit: the same concern as you mentioned in the comment above. Maybe we could consider to display initial values only for builds conducted via CI (Production, preview)

BrantalikP (approved):
Left a comment but nothing blocking 👌 . looks good!

suite-native/module-dev-utils/src/components/FeatureFlags.tsx
matejkriz: That sounds overcomplicated to me for this case, let's leave it and we will see. If there will be any confusion, I would just delete it...

STew790:
**QA OK**
Values after wiping from dev utils:
image

**Info**
25.3.2 9a1be4f

PR #17683, opened 2025-03-14
refactor(suite): Refactor footer a bit

jvaclavik:
## Description

Not great, not terrible. At lease 2 overridden components are gone.

## Related Issue

Resolve 

## Screenshots:

image


seibei-iguchi:
I'd go YOLO on this, it can be done with design components only (e.g. use Divider, supports both horizontal and vertical)

jvaclavik:
> I'd go YOLO on this, it can be done with design components only (e.g. use Divider, supports both horizontal and vertical)

Definitely, but I'm pretty sure we will replace it completely soon with something more useful and nice.

seibei-iguchi (approved):

PR #17687, opened 2025-03-14
chore(suite-desktop): Unify suite desktop flags

### LINKED ISSUE #17686: Cleanup Suite Desktop process switches
- Process switches and flags are all over the place in `@trezor/suite-desktop-core` , with no single source of truth listing what switches does our app support.
- Additionally, we are using `app.commandLine.hasSwitch` and `app.commandLine.getSwitchValue`, which are in fact **not** to be used for this purpose!
  - See [this comment in electron repo](https://github.com/electron/electron/issues/45991#issuecomment-2719202478)
  - TL;DR: it stopped working for us, I patched it in Suite & filed an issue which is indeed a bug, but the Electron team informed us that this API is meant only to access _Chromium_ switches, not custom app switches. [The docs](https://www.electronjs.org/docs/latest/api/command-line) do indicate it, but are not very clear about that..
 
### QA instructions

Check if Suite works correctly [with the following switches](https://github.com/trezor/trezor-suite/blob/9b7c87670cd573a2da62c99825d43c13b67566bd/docs/packages/suite-desktop.md#runtime-flags).
At least the important ones, but TBD which ones are important to be supported?? 🤔

Lemonexe:
## Description

- create helpers `hasSwitch` & `getSwitchValue` incl. unit tests
  - to mimic behavior of same-named `app.commandLine` methods, which are not to be used for this ([see issue](https://github.com/trezor/trezor-suite/issues/17686)) 
- refactor current usages of `app.commandLine` methods, or direct `process.argv` access
  - the only remaning usage of either of those two methods are width & height
  - those are actualy Chromium flags, so that's 

Note: only [width & height](https://github.com/trezor/trezor-suite/blob/1ba404577dc99c9918bb1d67b345de0dc5ec54e4/packages/suite-desktop-core/src/app.ts#L198-L199) remain, where the electron API is OK, because those are Chromium switches :heavy_check_mark: 

## Related Issue

Resolve #17686

### DIFF
--- a/packages/suite-desktop-core/src/libs/process-switches.ts
+++ b/packages/suite-desktop-core/src/libs/process-switches.ts
@@ -1,0 +1,44 @@
+export type SuiteSwitch =
+    | 'open-devtools'
+    | 'tor'
+    | 'pre-release'
+    | 'enable-updater'
+    | 'disable-updater'
+    | 'updater-url'
+    | 'bridge-legacy'
+    | 'bridge-test'
+    | 'bridge-dev'
+    | 'skip-new-bridge-rollout'
+    | 'bridge-daemon'
+    | 'bridge-daemon-show-ui'
+    | 'log-level'
+    | 'log-write'
+    | 'log-ui'
+    | 'log-file'
+    | 'log-path'
+    | 'log-no-print'
+    | 'log-connect'
+    | 'remove-user-data-on-start'
+    | 'expose-connect-ws'
+    | 'state'; // very special handling, see `./app-utils.ts`
+
+/**
+ * Check if a switch is present in process arguments.
+ */
+export const hasSwitch = (switchName: SuiteSwitch) => {
+    const isSwitch = new RegExp(`^--${switchName}(?:=[^=]+)?=?$`);
+
+    return process.argv.some(arg => isSwitch.test(arg));
+};
+
+/**
+ * Get value of a switch from process arguments, if present and if it has a value.
+ * Returns empty string otherwise, as Electron's `app.commandLine.getSwitchValue`.
+ */
+export const getSwitchValue = (switchName: SuiteSwitch): string => {
+    const switchValueMatch = new RegExp(`^--${switchName}=(.+)$`);
+    const valueMatch = process.argv.map(arg => arg.match(switchValueMatch)).find(Boolean);
+    if (!valueMatch) return '';
+
+    return valueMatch[1];
+};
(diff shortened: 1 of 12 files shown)


packages/suite-desktop-core/src/libs/process-switches.ts
```diff
@@ -0,0 +1,47 @@
...
+    'state', // very special handling, see `./app-utils.ts`
+] as const;
+
+export type SuiteSwitch = (typeof AVAILABLE_SUITE_SWITCHES)[number];
+
+/**
+ * Check if a switch is present in process arguments.
+ */
+export const hasSwitch = (switchName: SuiteSwitch) => {
+    const isSwitch = new RegExp(`^--${switchName}(?:=[^=]+)?=?$`);
```
Lemonexe, line 29: regex carefully crafted to match `--flag` without value, but also `--flag=with_value` or empty value `--flag=`, but `=?.+` would not suffice, because it would accidentally match substring.
In other words, value at end can only come after `=` _(optional)_.

This is simpler in `getSwitchValue`, where the regex always requires `=`

packages/suite-desktop-core/src/libs/process-switches.ts
```diff
@@ -0,0 +1,47 @@
...
+    'log-file',
+    'log-path',
+    'log-no-print',
+    'log-connect',
+    'remove-user-data-on-start',
+    'expose-connect-ws',
+    'state', // very special handling, see `./app-utils.ts`
+] as const;
+
+export type SuiteSwitch = (typeof AVAILABLE_SUITE_SWITCHES)[number];
```
peter-sanderson: nit: `AVAILABLE_SUITE_SWITCHES` is not used for runtime validation so the union of string literals would be enough. But this is ofc ok as well

peter-sanderson: Or maybe were you considering the runtime validation of arguments?

peter-sanderson (approved):

packages/suite-desktop-core/src/libs/process-switches.ts
Lemonexe: Oh thanks for pointing it out. Yes, initially I was considering runtime validation, that's why I created variable, but then I realized it'd have to allow Chromium flags and Electron flags, and that's unfeasible, because in case of Chromium, there is no single source of truth for the flags :see_no_evil: 
Ofc we could only whitelist _some of the_ Chromium | Electron flags, but that's a decision for another time..

PR #17689, opened 2025-03-15
fix(native-suite): device setting actions refactor

### LINKED ISSUE #17648: Unexpected PIN Change Prompt
**Describe the bug**
There is an edge case where a Trezor device unexpectedly offers a PIN change

**Info:**

-   Mobile Suite Version: 25.3.2/6fdcff3bac
-   OS: Android

**How to reproduce**

1.	Have a disconnected, remembered (view-only) Trezor.
2.	Go to Device Settings.
3.	Open Change PIN, then close it.
4.	Open Firmware Update.
5.	Connect the Trezor.
6.	The Trezor unexpectedly prompts for a PIN change.

**Expected behavior**
The Trezor should not trigger a PIN change prompt unless explicitly requested by the user.

**Screenshots**
Image

BrantalikP:
## Description

Prevent triggering device setting actions unintentionally.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17648

### Before

https://github.com/user-attachments/assets/72a63b5f-952e-4e56-9050-a20f9bfe1adf

https://github.com/user-attachments/assets/f15ff190-8ac4-4610-8a77-5773e02e52a0

### After

https://github.com/user-attachments/assets/876bc091-70d6-4900-84da-32f94a2605d5

https://github.com/user-attachments/assets/4ce28e10-9042-4a12-97d0-d9a7edd3ad5a

### DIFF
--- a/suite-native/module-device-settings/src/components/DeviceInteractionScreenWrapper.tsx
+++ b/suite-native/module-device-settings/src/components/DeviceInteractionScreenWrapper.tsx
@@ -5,6 +5,7 @@ import { useNavigation } from '@react-navigation/native';
 
 import { selectSelectedDevice } from '@suite-common/wallet-core';
 import { Screen, ScreenHeader } from '@suite-native/navigation';
+import TrezorConnect from '@trezor/connect';
 
 type DeviceInteractionScreenWrapperProps = {
     children: ReactNode;
@@ -18,6 +19,7 @@ export const DeviceInteractionScreenWrapper = ({
     const device = useSelector(selectSelectedDevice);
 
     const closeAction = useCallback(() => {
+        TrezorConnect.cancel();
         navigation.goBack();
     }, [navigation]);
 
--- a/suite-native/module-device-settings/src/components/DeviceInteractionScreenWrapper.tsx
+++ b/suite-native/module-device-settings/src/components/DeviceInteractionScreenWrapper.tsx
@@ -5,7 +5,6 @@ import { useNavigation } from '@react-navigation/native';
 
 import { selectSelectedDevice } from '@suite-common/wallet-core';
 import { Screen, ScreenHeader } from '@suite-native/navigation';
-import TrezorConnect from '@trezor/connect';
 
 type DeviceInteractionScreenWrapperProps = {
     children: ReactNode;
@@ -19,7 +18,6 @@ export const DeviceInteractionScreenWrapper = ({
     const device = useSelector(selectSelectedDevice);
 
     const closeAction = useCallback(() => {
-        TrezorConnect.cancel();
         navigation.goBack();
     }, [navigation]);
 
--- a/suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
+++ b/suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
@@ -1,0 +1,82 @@
+import { useCallback, useEffect } from 'react';
+import { useSelector } from 'react-redux';
+
+import { useNavigation } from '@react-navigation/native';
+
+import { selectIsDeviceConnected, selectSelectedDevice } from '@suite-common/wallet-core';
+import { DeviceAuthenticityCheckResult, EventType, analytics } from '@suite-native/analytics';
+import { requestPrioritizedDeviceAccess } from '@suite-native/device-mutex';
+import {
+    DeviceAuthenticityStackParamList,
+    DeviceAuthenticityStackRoutes,
+    DeviceSettingsStackParamList,
+    DeviceStackRoutes,
+    StackToStackCompositeNavigationProps,
+} from '@suite-native/navigation';
+import TrezorConnect from '@trezor/connect';
+
+type NavigationProp = StackToStackCompositeNavigationProps<
+    DeviceAuthenticityStackParamList,
+    DeviceAuthenticityStackRoutes,
+    DeviceSettingsStackParamList
+>;
+export const useDeviceAuthenticityAction = () => {
+    const navigation = useNavigation<NavigationProp>();
+    const isDeviceConnected = useSelector(selectIsDeviceConnected);
+
+    const device = useSelector(selectSelectedDevice);
+    const reportCheckResult = useCallback(
+        (result: DeviceAuthenticityCheckResult) =>
+            analytics.report({
+                type: EventType.DeviceSettingsAuthenticityCheck,
+                payload: { result },
+            }),
+        [],
+    );
+    const checkAuthenticity = useCallback(async () => {
+        navigation.navigate(DeviceStackRoutes.DeviceAuthenticity);
+
+        const result = await requestPrioritizedDeviceAccess({
+            deviceCallback: () =>
+                TrezorConnect.authenticateDevice({
+                    device: {
+                        path: device?.path,
+                    },
+                }),
+        });
+        if (!result.success) {
+            return;
+        }
+
+        const { success, payload } = result.payload;
+        if (success) {
+            const checkResult = payload.valid ? 'successful' : 'compromised';
+            const configExpired = payload.error === 'CA_PUBKEY_NOT_FOUND' && payload.configExpired;
+            navigation.navigate(DeviceAuthenticityStackRoutes.AuthenticitySummary, {
+                checkResult: configExpired ? 'successful' : checkResult,
+            });
+            reportCheckResult(configExpired ? 'configExpired' : checkResult);
+        } else {
+            const errorCode = payload.code;
+            if (errorCode === 'Failure_ActionCancelled' || errorCode === 'Failure_PinCancelled') {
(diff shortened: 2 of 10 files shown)


suite-native/module-device-settings/src/hooks/usePinAction.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { useToast } from '@suite-native/toasts';
+import TrezorConnect from '@trezor/connect';
+
+type NavigationProp = StackNavigationProps<
+    DeviceSettingsStackParamList,
+    DeviceStackRoutes.DevicePinProtection
+>;
+
+type ActionConfig = {
+    param: boolean | undefined;
```
matejkriz: nitpick, feel free to ignore it because it was there before and I know it's not something you changed, but the name of the param `param` is cryptic and it would be nice to rename it to something like `remove` or make it generic same as on desktop by passing all params with type from Connect `Parameters<typeof TrezorConnect.changePin>[0]`

suite-native/module-device-settings/src/components/DeviceInteractionScreenWrapper.tsx
```diff
@@ -18,6 +19,7 @@ export const DeviceInteractionScreenWrapper = ({
     const device = useSelector(selectSelectedDevice);
 
     const closeAction = useCallback(() => {
+        TrezorConnect.cancel();
```
matejkriz: This was removed in the previous commit and the second commit put it back, which I believe is correct.

suite-native/module-device-settings/src/hooks/useDeviceChangedCheck.ts
```diff
@@ -35,6 +36,7 @@ export const useDeviceChangedCheck = () => {
 
         // Check for device mismatch only after initial device ID is set
         if (initialDeviceIdRef.current && device?.id && initialDeviceIdRef.current !== device.id) {
+            TrezorConnect.cancel();
```
matejkriz, line 39: I'm not sure this is safe if we don't know what is it canceling.

suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
```diff
@@ -0,0 +1,81 @@
...
+                navigation.navigate(DeviceAuthenticityStackRoutes.AuthenticitySummary, {
+                    checkResult: 'compromised',
+                });
+                reportCheckResult('failed');
+            }
+        }
+    }, [device, navigation, reportCheckResult]);
+
+    useEffect(() => {
+        if (isDeviceConnected) void checkAuthenticity();
```
matejkriz: First time I see someone using void like this. What is the benefit? Should we use it everywhere? If there is no benefit I would remove it for consistency.

suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
matejkriz, line 80: Is that really necessary here? Why don't you just add `checkAuthenticity` to dependencies?

matejkriz (commented):
Few nits so far, I would like to test and think about the TrezorConnect.cancel() placement.

suite-native/module-device-settings/src/components/DeviceInteractionScreenWrapper.tsx
BrantalikP: Yeah, the initial idea was to call `TrezorConnect.cancel `in `useEffect` on a specific screen when the user leaves the page.

```
useEffect(() => () => {
  if (deviceConnected) TrezorConnect.cancel();
}, []);
```

or something similar. But then, I realized two reasons why this wouldn’t be a good idea:

1. @yanascz  mentioned that he wants to reuse the `ContinueOnTrezor` page for his Bluetooth implementation as well, so I didn’t want to create obstacles for his implementation.

2.	If T`rezorConnect.cancel` were called twice (once in `useEffect` and once by pressing the `close` button), it caused issues where, if you tried to change the PIN or trigger another device setting action right after, it wouldn’t get triggered.

suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
BrantalikP: It’s just explicitly stating that there’s no need to wait for this async function to finish. I’m not sure if there was an `ESLint` warning or if it’s just something I’m used to doing. It makes it clear that I have not forgot to put `await` prefix  before the function

suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
BrantalikP: Because the `checkAuthenticity` has `device` as dependencies therefore IMHO it could trigger the `useEffect`, unintentionally?

suite-native/module-device-settings/src/hooks/useDeviceChangedCheck.ts
BrantalikP: Yeah, you have more knowledge about it, but I assumed, based on its usage in the `ContinueOnTrezor` page, that it’s just canceling any ongoing action on `Trezor`. It’s more like a `cleanup` before leaving the page and as I know it used solely for handling incorrect device that has been connected on Device Setting screen, tbh not sure if there might be other scenariosu where it can caused issues

But if there are any concerns, I’m happy to change it. Thanks!

suite-native/module-device-settings/src/hooks/usePinAction.tsx
BrantalikP: https://github.com/trezor/trezor-suite/pull/17689/commits/3d5164a30ec1bcf5f8517850406b222fe2619319 as the latter option would require changes also in the `actionConfigMap` I opted for the first one.

suite-native/module-device-settings/src/hooks/useDeviceChangedCheck.ts
matejkriz: From my testing, it seems to work fine 👌

suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
matejkriz: We don't have any eslint warning for that, so I would strongly prefer to remove it to stay consistent. If you find that rule useful, let's discuss with a whole team if we can add it 🙏

BrantalikP: image

Oh ok this I tried to solve. IMHO using `void` is the cleanest way. Or do you have any other prefered solution. 

I would want to try to avoid something like this ﻿ ` if (isDeviceConnected) (async () => await checkAuthenticity())();` 😅

suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
matejkriz: You are right, if you add checkAuthenticity as a dependency, this effect is triggered multiple times. Would be great to mention the reason in a comment here (and for handlePinAction too).

BrantalikP: https://github.com/trezor/trezor-suite/pull/17689/commits/7656c2de293a5f1973d86971be07f66001493f9f  https://github.com/trezor/trezor-suite/pull/17689/commits/fe43c74f8f3f07ab20abebf942d1f87dba5bf9db

suite-native/module-device-settings/src/hooks/useDeviceAuthenticityAction.ts
matejkriz: Is that coming from your local eslint config? I don't see that warning if I run `yarn run lint:js`

BrantalikP: you might be right. new IDE new issues. Going to disable it in favor of the project setup (once I find it ). I push fixups. Thanks!

BrantalikP: IDE warning suppressed https://github.com/trezor/trezor-suite/pull/17689/commits/29f47ee0a4c4489b44eca3076a09d5b8e5ad3dcd https://github.com/trezor/trezor-suite/pull/17689/commits/de73286a153e50890e061db8124ddc1926d21cb3

matejkriz (approved):
Great job, works smooth for me 🎉

matejkriz:
android E2E is failing on the same thing as on develop https://github.com/trezor/trezor-suite/actions/runs/13981866909/job/39148822012

PR #17701, opened 2025-03-17
fix(suite): fix needless circular dependency

vojtatranta:
## Description

Fixes circular dependency that broke Vite's HMR


packages/suite/src/reducers/suite/routerReducer.ts
```diff
@@ -1,9 +1,18 @@
 import { ROUTER } from 'src/actions/suite/constants';
-import { ACCOUNT_TABS } from 'src/components/wallet/WalletLayout/AccountTopPanel/AccountNavigation';
 import type { AnchorType } from 'src/constants/suite/anchors';
 import { RouterAppWithParams, SettingsBackRoute } from 'src/constants/suite/routes';
 import { Action } from 'src/types/suite';
 
+const ACCOUNT_TABS = [
```
vojtatranta, line 6: This was kinda fun thing, you exported `ACCOUNT_TABS` from the place where it wasn't needed at all, to import it in a file that was being imported whrere the `ACCOUNT_TABS` were defined

jvaclavik (approved):

PR #17709, opened 2025-03-17
fix(suite): when device labeling confirmation fails / is cancelled, e…

### LINKED ISSUE #17663: Do not turn labeling on before the enabling flow has been successfully finished
If a user turns labeling on, they are offered labels storage method (Dropbox, Google Drive, Local storage). If they cancel the flow from there or unsuccessfully try to log in to one of the offered services, they will have no means of storage, but the labeling switch is on. So labeling is on, but not really.

Let's please get rid of this inconsistency and only keep the Labeling setting on if the user has successfully finished the setup.

This might potentially interfere also with the situation where storage service session expires and the user is prompted to re-login. If they fail to do so, labeling will be turned off. That is fair, as without logging back in, it would again be set on, but non functioning. We should ideally show a toast message if this happen informing that "Labeling turned off".

vojtatranta:
…nsure that global labelling is turned off

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17663

### DIFF
--- a/packages/suite/src/actions/suite/metadataLabelingActions.ts
+++ b/packages/suite/src/actions/suite/metadataLabelingActions.ts
@@ -583,7 +583,8 @@ export const init =
         }
 
         // 1. set metadata enabled globally
-        if (!getState().metadata.enabled) {
+        const globalLabelingEnabledBeforeToggle = getState().metadata.enabled;
+        if (!globalLabelingEnabledBeforeToggle) {
             dispatch(metadataActions.enableMetadata());
         }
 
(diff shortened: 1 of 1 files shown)


packages/suite/src/actions/suite/metadataLabelingActions.ts
```diff
@@ -583,7 +583,8 @@ export const init =
         }
 
         // 1. set metadata enabled globally
-        if (!getState().metadata.enabled) {
```
vojtatranta, line 586: I wasn't sure if putting this `dispatch(metadataActions.enableMetadata());` to the end of hte proces wouldn't break something, since it is possible that some intermediat processing relies on this being on. So I did a "rollback".

mroz22 (approved):
probably ok

PR #17715, opened 2025-03-17
fix(suite): make sure that once any fiat token rate is known its refr…

vojtatranta:
…esh won't cause UI loading

## Description

When you added large QA seed with that passphrase, you could see how the Sol2 tokens values are loading - the skeleton was there correctly.

However, after it succesffully loaded the rates and displayed the value, it become skeleton again, coz one of the fiat rate started refetching.

Obvious fix was to check if even older fiat rate is there, so loading state was only when there is no fiat rate available and the there was no fiat rate.

## Related Issue

Resolve

### DIFF
--- a/suite-common/wallet-utils/src/accountUtils.ts
+++ b/suite-common/wallet-utils/src/accountUtils.ts
@@ -589,7 +589,12 @@ export const areTokenFiatRatesLoading = (
             token.contract as TokenAddress,
         );
 
-        return rates?.[tokenFiatRateKey]?.isLoading;
+        // NOTE: designed the rate as loading only if it is loading and NO rate is currently available
+        return (
+            rates?.[tokenFiatRateKey]?.isLoading &&
+            !rates?.[tokenFiatRateKey]?.rate &&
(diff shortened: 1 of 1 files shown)


jvaclavik (approved):

bosomt:
**QA OK**

no skeletons after discovery of tokens

image


**Info:**
 - Suite version: desktop 25.4.0 (048d3f51f8cfbd9bc5be51ec8fd006fbe55b20c2)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.4.0 Chrome/134.0.6998.44 Electron/35.0.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T3B1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)
 - Transport: BridgeTransport 3.0.0-bundled.25.4.0

PR #17717, opened 2025-03-17
`DeviceCommands` spring cleaning

marekrjpolak:
## Description

General polishing of (mostly) `DeviceCommands`, in order to make some multi-transport handling of devices easier.

## Commits
- https://github.com/trezor/trezor-suite/pull/17717/commits/f857770fa3c89343bf849cad382d8d9b5925000a - Merging `_commonCall`, `call` and `_filterCommonTypes` together, and making it loop (`callLoop`) instead of recursion, because it's easier to think about
- https://github.com/trezor/trezor-suite/pull/17717/commits/f857770fa3c89343bf849cad382d8d9b5925000a - dissolve `prompts` back into `DeviceCommands`; simplify `setCancelableAction`/`clearCancelableAction` logic from device into `awaitCancel`/`clearCancel` in `DeviceCommands`; also, improved typing of prompt-like device events
- https://github.com/trezor/trezor-suite/pull/17717/commits/ec92cfb91a0b5adea9f23cbf9f1ef9b0e25d493f - separation of `DeviceCommands` into `DeviceCurrentSession`, which handles only cancelable `typedCall` logic, and former `DeviceCommands` which uses it for higher-level procedures; currently both are public which can change
- https://github.com/trezor/trezor-suite/pull/17717/commits/fcbe7bf21d0aaead79175356c8ce592cf2545e3f - `callLoop` is now strictly linear - no ugly hidden paralelism; `callLoop` never throws but always returns; no more different mechanism for cancelling and disposing - both awaits the same promise (but only cancelling calls cancel)
- https://github.com/trezor/trezor-suite/pull/17717/commits/ec396d9f3353e29aa718d05a3437504e598d0eed - simplified emergency read after unexpected message is received
- https://github.com/trezor/trezor-suite/pull/17717/commits/0d4e27d2c25cae8861f35526703ee5da90dc0e9d - no rethrowing when unexpected message is received


packages/connect/src/device/DeviceCommands.ts
mroz22, line 1: nitpick and my personal preference only - even though ethereumGetPublicKey is used only once I didn't mind having it defined here since its internal logic was similar to the internal logic of `getHDNode`. Having methods that share the same purpose in one file feels slightly more natural to me. What do others think?

packages/connect/src/device/DeviceCurrentSession.ts
```diff
@@ -0,0 +1,306 @@
...
+            ]);
+
+            if (name === 'ButtonAck' && abortedDuringCall && !this.disposed) {
+                if (this.needCancelWorkaround()) {
+                    try {
+                        // UI_EVENT is send right before ButtonAck, make sure that ButtonAck is sent
+                        await resolveAfter(1);
+                        await this.device.acquire();
+                        await this.device.getCurrentSession().cancelCall(false);
+                        await this.device.release();
```
mroz22, line 152: interesting that only ButtonRequest uses this workaround. I presume it was like this also before. There are couple of other "Something-Acks"

mroz22 (commented):
looks good, maybe a little difficult / time-consuming to compare it line by line to prove that logic stays intact. I will keep reading 👍

packages/connect/src/device/DeviceCommands.ts
marekrjpolak: Just for the record, I dissolve these methods when I still hoped that I'd manage to do it to all of them. Now, when `DeviceCommands` will remain there separated from `DeviceCurrentSession`, I kinda agree to put them back there.

But still, as I mentioned previously, :100: consistent would be to do these helpers for **all** the typed calls and not just the public-keys-related.

packages/connect/src/device/DeviceCurrentSession.ts
marekrjpolak: Yeah, it was like this only for `ButtonAck`. The difference is that only `ButtonAck` expects cancel during pending `transport.call`, the other ones just expect it while they wait for data from UI promise.

marekrjpolak: But of course, there are still places where `cancelCall` is called from outside `DeviceCurrentSession` so it can basically cancel any pending call and I'd be delighted to unify it with this case, instead of differentiating between various types.

packages/connect/src/device/DeviceCommands.ts
marekrjpolak: I tried to find a middle ground and moved `getAddress`, `ethereumGetAddress` and `ethereumGetPublicKey` back into `DeviceCommands`, but left final postprocessing of responses to the callers. https://github.com/trezor/trezor-suite/pull/17717/commits/f504be9fce3d7b9dc979c94ed0d639f9fe9127eb

packages/connect/src/api/getAccountInfo.ts
```diff
@@ -317,18 +311,16 @@ export default class GetAccountInfo extends AbstractMethod<'getAccountInfo', Req
...
         const blockchain = await initBlockchain(coinInfo, this.postMessage, identity);
         const dfd = this.createUiPromise(UI.RECEIVE_ACCOUNT);
 
         const discovery = new Discovery({
             blockchain,
-            commands: this.device.getCommands(),
-            derivationType:
-                typeof request.derivationType !== 'undefined'
-                    ? request.derivationType
-                    : PROTO.CardanoDerivationType.ICARUS_TREZOR,
```
mroz22, line 330: this was totaly useles since `derivationType` default is set from AbstractMethod?

marekrjpolak: No, but every single caller was defaulting to `ICARUS_TREZOR` while the `getAccountDescriptor` command itself throws when received undefined (which never happened). So I set it as a default directly in `getAccountDescriptor` :man_shrugging:

mroz22 (approved):
it looks good. shipit :shipit:

evgenysl:
general check on https://dev.suite.sldev.cz/connect/develop/ is OK

PR #17720, opened 2025-03-17
chore: validate config to node-utils package

karliatto:
## Description

In preparation for creating releases json schema this would be useful in order to avoid duplicated code since we will have to validate the new schema for releases messages.

### DIFF
--- a/suite-common/message-system/scripts/validate-config.ts
+++ b/suite-common/message-system/scripts/validate-config.ts
@@ -1,26 +1,18 @@
 /* eslint-disable no-console */
-import Ajv from 'ajv';
-import * as fs from 'fs';
+import fs from 'fs';
+
+import { validateJsonSchema } from '@trezor/node-utils';
 
 import { CONFIG_PATH, SCHEMA_PATH } from './constants';
 
-// checks that a config meets the criteria specified by the schema
-const validateConfigStructure = () => {
-    console.log('Validating config against schema...');
-
-    const ajv = new Ajv();
-
+console.log('Validating config against schema...');
+try {
     const config = fs.readFileSync(CONFIG_PATH, 'utf-8');
     const schema = fs.readFileSync(SCHEMA_PATH, 'utf-8');
 
-    const validate = ajv.compile(JSON.parse(schema));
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -9422,7 +9422,6 @@ __metadata:
     "@trezor/env-utils": "workspace:*"
     "@trezor/utils": "workspace:*"
     "@types/fs-extra": "npm:^11.0.4"
-    ajv: "npm:^8.17.1"
```
mroz22: I don't care that much but since you are touching it now - could there be a chance to replace ajv with typebox?  cc @martykan

karliatto: The idea is good, but it looks like it would require changes in the schema since typebox does not support JSON schema.

mroz22 (approved):
I don't like very much that `validateJsonSchema` util does 2 things at once - reading from fs + validation. So I would suggest refactoring it

yarn.lock
mroz22: yes. fuck it then

karliatto:
> I don't like very much that `validateJsonSchema` util does 2 things at once - reading from fs + validation. So I would suggest refactoring it

I agree :100: % with you so I updated it to do only one thing https://github.com/trezor/trezor-suite/pull/17720/commits/0ae0797f58d14fa7c34da67f73816b91d5a7239f#diff-6973949d0de6a51e54a8442d9b8014f177c4af917978bac4bfd497d8d4cff1b3

Thanks :pray:

suite-common/message-system/scripts/validate-config.ts
```diff
@@ -1,26 +1,17 @@
...
-        throw Error(`Config is invalid: ${JSON.stringify(validate.errors)}`);
-    }
-
+    validateJsonSchema(config, schema);
     console.log('Config is valid!');
-};
-
-validateConfigStructure();
+} catch (error) {
+    console.error(error.message);
```
mroz22, line 16: btw wasnt it supposed to throw before?

karliatto: Yes, you are right. I add it to rethrow here below.

PR #17724, opened 2025-03-18
fix(e2e): Disable FW revision checks for tests with x-main FW

HajekOndrej:
## Description

For firmware canary tests, we need to disable also FW revision checking.

## Related Issue

Resolve


Vere-Grey (approved):
:+1:

PR #17725, opened 2025-03-18
chore(suite): ua-parser-js@2.0.3

### LINKED ISSUE #17394: Upgrade `ua-parser-js`
[Migration guide](https://docs.uaparser.dev/intro/whats-new.html)

See the change in licensing.

vojtatranta:
## Related Issue

Resolve: https://github.com/trezor/trezor-suite/issues/17394

### DIFF
--- a/packages/connect/src/data/config.ts
+++ b/packages/connect/src/data/config.ts
@@ -72,6 +72,11 @@ export const config = {
             download: 'https://www.google.com/chrome/',
             update: 'https://support.google.com/chrome/answer/95414',
         },
+        mobilechrome: {
+            version: 59,
+            download: 'https://www.google.com/chrome/',
+            update: 'https://support.google.com/chrome/answer/95414',
+        },
         chromium: {
             version: 59,
             download: 'https://www.chromium.org/',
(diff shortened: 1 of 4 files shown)


packages/connect/src/data/config.ts
```diff
@@ -72,6 +72,11 @@ export const config = {
             download: 'https://www.google.com/chrome/',
             update: 'https://support.google.com/chrome/answer/95414',
         },
+        mobilechrome: {
```
vojtatranta, line 75: @komret I am not particuolary sure about this.

The new UA parser returns android as `mobilechrome` not  as `chrome` as the earlier versions.

So, the fix here is whacky here. Maybe we could say that if ua parser returns browser name as `mobilechrome` we could treat it as `chrome` but.

But this seems cleaner to me. Without it, the tests logically fail.

packages/env-utils/src/envUtils.ts
```diff
@@ -43,7 +43,7 @@ const getOsVersion = () => getUserAgentParser().getOS().version || '';
 const getSuiteVersion = () => process.env.VERSION || '';
 
 const getBrowserName = () => {
-    const browserName = getUserAgentParser().getBrowser().name;
+    const browserName = getUserAgentParser().getBrowser().name?.replace(' ', '');
```
vojtatranta, line 46: I added this based on the migration guide https://docs.uaparser.dev/intro/whats-new.html

packages/connect/src/data/config.ts
```diff
@@ -87,6 +92,11 @@ export const config = {
             download: 'https://www.mozilla.org/en-US/firefox/new/',
             update: 'https://support.mozilla.org/en-US/kb/update-firefox-latest-version',
         },
+        mobilefirefox: {
```
vojtatranta, line 95: same as with chrome

vojtatranta:
Pending licence review

komret (commented):
LGTM, let's figure out licensing before merging.

komret (approved):
License purchased, let's merge this.

PR #17726, opened 2025-03-18
fix(connect-popup): add missing margin after components update

mroz22:
before 
image

after 
image


packages/connect-ui/src/views/Error.tsx
```diff
@@ -236,6 +236,10 @@ const TipsContainer = styled.div`
     width: 500px;
     margin-top: 40px;
     margin-bottom: 20px;
+
+    > * {
+        margin: 20px;
```
mroz22: maybe only top and bottom

martykan (approved):

PR #17766, opened 2025-03-19
feat(ci): Add smoke checks to PR e2e

HajekOndrej:
## Description

In this PR:
Extract desktop app build and do it only once. 
Share to e2e tests via artifacts.
Added Smoke check group which has the ability to cancel all other test groups if it fails. This is a fail fast strategy, that shortens feedback loop and saves currents reports in case of catastrophically breaking change in Suite.
Profit.

## Related Issue

Resolve

### DIFF
--- a/.github/actions/run-e2e-tests/action.yml
+++ b/.github/actions/run-e2e-tests/action.yml
@@ -11,6 +11,10 @@ inputs:
   containers:
     description: "Docker containers to pull"
     required: true
+  workspace-dependencies:
+    description: "Additional workspaces to install"
+    required: false
+    default: ""
   test-group:
     description: "Test group to run"
     required: true
@@ -55,15 +59,16 @@ runs:
         cache: yarn
 
     - name: Install PW Dependencies
-      if: ${{ inputs.project == 'web' }}
       shell: bash
       run: |
         echo "::group::Install PW Dependencies"
         echo "Silently installing Playwright dependencies"
         echo -e "\nenableScripts: false" >> .yarnrc.yml
         echo -e "\nenableHardenedMode: false" >> .yarnrc.yml
-        yarn workspaces focus @trezor/suite-desktop-core > /dev/null
-        npx playwright install --with-deps > /dev/null
+        yarn workspaces focus @trezor/suite-desktop-core ${{ inputs.workspace-dependencies }} > /dev/null
+        if [[ "${{ inputs.project }}" == "web" ]]; then
+          npx playwright install --with-deps > /dev/null
+        fi
         echo "::endgroup::"
 
     - name: Pull Docker Images
--- a/.github/actions/run-e2e-tests/action.yml
+++ b/.github/actions/run-e2e-tests/action.yml
@@ -63,8 +63,6 @@ runs:
       run: |
         echo "::group::Install PW Dependencies"
         echo "Silently installing Playwright dependencies"
-        echo -e "\nenableScripts: false" >> .yarnrc.yml
-        echo -e "\nenableHardenedMode: false" >> .yarnrc.yml
         yarn workspaces focus @trezor/suite-desktop-core ${{ inputs.workspace-dependencies }} > /dev/null
         if [[ "${{ inputs.project }}" == "web" ]]; then
           npx playwright install --with-deps > /dev/null
--- a/.github/workflows/test-suite-desktop-e2e-pr.yml
+++ b/.github/workflows/test-suite-desktop-e2e-pr.yml
@@ -54,8 +54,9 @@ jobs:
     needs:
       - build-app
       - load-e2e-matrix
+    continue-on-error: ${{ !matrix.smoke }}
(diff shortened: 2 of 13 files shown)


.github/actions/run-e2e-tests/action.yml
```diff
@@ -55,15 +59,14 @@ runs:
...
-      if: ${{ inputs.project == 'web' }}
       shell: bash
       run: |
         echo "::group::Install PW Dependencies"
         echo "Silently installing Playwright dependencies"
-        echo -e "\nenableScripts: false" >> .yarnrc.yml
-        echo -e "\nenableHardenedMode: false" >> .yarnrc.yml
-        yarn workspaces focus @trezor/suite-desktop-core > /dev/null
-        npx playwright install --with-deps > /dev/null
+        yarn workspaces focus @trezor/suite-desktop-core ${{ inputs.workspace-dependencies }} > /dev/null
```
Vere-Grey, line 66: I wondering the value of another argument here?
What we want to achieve is to for desktop build:
```
yarn workspaces focus @trezor/suite-desktop-core "@trezor/suite-desktop > /dev/null
```
and for web:
```
yarn workspaces focus @trezor/suite-desktop-core  > /dev/null
npx playwright install --with-deps > /dev/null
```

Maybe just if else will be cleaner and more easy to read? Unless you forsee additional use of that argument?

.github/workflows/test-suite-desktop-e2e-pr.yml
```diff
@@ -43,9 +52,11 @@ jobs:
     if: github.repository == 'trezor/trezor-suite'
     runs-on: ubuntu-24.04
     needs:
+      - build-app
```
Vere-Grey, line 55: This is missing in several other workflows

.github/workflows/test-suite-desktop-e2e-pr.yml
```diff
@@ -43,9 +52,11 @@ jobs:
...
       - load-e2e-matrix
+    continue-on-error: ${{ !matrix.smoke }}
     strategy:
-      fail-fast: false
+      fail-fast: true
```
Vere-Grey, line 59: Again, just commenting to confirm that I understand it correctly.
We have by default fail-fast so if one tests or action in one of the rest is canceled. But only if that happens on a group that has smoke true.

.github/workflows/test-suite-web-e2e-pr.yml
```diff
@@ -60,8 +60,6 @@ jobs:
     outputs:
       matrix: ${{ steps.load-matrix.outputs.matrix }}
     runs-on: ubuntu-latest
-    needs:
-      - build-web
```
Vere-Grey, line 64: Shouldn't we remove this also from other web workflows?

.github/workflows/test-suite-desktop-e2e-pr.yml
HajekOndrej: We have normally fail-fast=false and it will stay that way in all jobs except this one. It means, that fails in other jobs don't cancel the other jobs.

In this one: failure of smoke group cancels other jobs, failure of anything else doesn't cancel anything.

.github/workflows/test-suite-web-e2e-pr.yml
HajekOndrej: yes

.github/actions/run-e2e-tests/action.yml
HajekOndrej: Seemed more universal. Also I am pretty sure we don't really need to add @trezor/suite-desktop, because I suppose it gets installed with @trezor/suite-desktop-core anyway, but these things can change, so I thought it would be better to be more explicit.

.github/workflows/test-suite-web-e2e-pr.yml
```diff
@@ -107,7 +106,7 @@ jobs:
           project: "web"
           containers: ${{ matrix.CONTAINERS }}
           test-group: ${{ matrix.TEST_GROUP }}
-          additional-grep: "(?=.*@webOnly)"
+          additional-grep: ${{ matrix.SMOKE && '(?=.*)' || '(?=.*@webOnly)' }}
```
Vere-Grey: I think we can simplify this to
${{ !matrix.SMOKE &&  '(?=.*@webOnly)' }}

Meaning, apply this @webOnly additional grep only when the group is not smoke test group

.github/workflows/test-suite-desktop-e2e-pr.yml
HajekOndrej: fixed

.github/workflows/test-suite-web-e2e-pr.yml
HajekOndrej: yep, that is much better :)

Vere-Grey (approved):
:+1:

PR #17786, opened 2025-03-19
Connect in desktop preparation

mroz22:
- use @trezor/websocket-client in @trezor/connect-web when using core-in-desktop mode

### DIFF
--- a/packages/connect-web/src/impl/core-in-suite-desktop.ts
+++ b/packages/connect-web/src/impl/core-in-suite-desktop.ts
@@ -18,7 +18,7 @@ import type {
     Response,
 } from '@trezor/connect/src/types';
 import { Login } from '@trezor/connect/src/types/api/requestLogin';
-import { DeferredManager, createDeferred, createDeferredManager } from '@trezor/utils';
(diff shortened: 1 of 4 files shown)


packages/connect-web/src/impl/core-in-suite-desktop.ts
```diff
@@ -138,6 +140,15 @@ export class CoreInSuiteDesktop implements ConnectFactoryDependencies<ConnectSet
 
     // not needed, only because of types
     renderWebUSBButton() {}
+
+    private wsErrorHandler = (error: any) => {
```
marekrjpolak: Some of my rather shallow observations and/or opinions:
- I wouldn't add `errorHandler` to `createDeferred` as it's pretty low-level, straightforward util
- It's possible to add `resolveAll` to `createDeferredManager`, similarly to `rejectAll`, if really needed
- I still believe caller should be the one responsible for converting rejections to failure payloads

And the most important:
- I believe most of this would be resolved by `return await this.ws.sendMessage` instead of `return this.ws.sendMessage` as then the catch block would catch the rejections rather than returning them while pending.

martykan (approved):

PR #17787, opened 2025-03-19
Data Analytics: settings/tor - add location and modal to the event

### LINKED ISSUE #15517: Data Analytics: Tor settings location
We would like to add `location` parameter to event [settings/tor](https://www.notion.so/satoshilabs/settings-tor-893783aec53b49109cd79d1633ec88f5?pvs=4) with a specific place from where it has been turned on (settings, coinjoin, custom backend…).

izmy:
## Description

Data Analytics: _settings/tor_ now has new parameters `location` and `modal`.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15517

### DIFF
--- a/packages/suite/src/actions/suite/suiteActions.ts
+++ b/packages/suite/src/actions/suite/suiteActions.ts
@@ -172,7 +172,8 @@ export const updateTorStatus = (payload: TorStatus): SuiteAction => ({
 });
 
 export const toggleTor =
-    (shouldEnable: boolean) => async (dispatch: Dispatch, getState: GetState) => {
+    (shouldEnable: boolean, modal: string | undefined) =>
+    async (dispatch: Dispatch, getState: GetState) => {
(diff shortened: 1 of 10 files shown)


packages/suite/src/actions/suite/suiteActions.ts
```diff
@@ -207,6 +208,8 @@ export const toggleTor =
                 type: EventType.SettingsTor,
                 payload: {
                     value: shouldEnable,
+                    location: getState().router.url,
+                    modal,
```
tomasklim, line 212: can you get modal type from state here too?

izmy: I tried this and got a closed modal without a type (probably because it is an async function). So I decided the best solution is to get the type at the moment of click.

tomasklim (approved):

PR #17788, opened 2025-03-19
feat(connect-webextension): enable switching between multiple implementations

martykan:
## Description

Enable switching between different implementations in `connect-webextension`

This brings support for `coreMode: 'suite-desktop'`

### DIFF
--- a/packages/connect/src/types/settings.ts
+++ b/packages/connect/src/types/settings.ts
@@ -65,15 +65,21 @@ export interface ConnectSettingsWeb {
 export interface ConnectSettingsWebextension {
     /** _extendWebextensionLifetime features makes the service worker in @trezor/connect-webextension stay alive longer */
     _extendWebextensionLifetime?: boolean;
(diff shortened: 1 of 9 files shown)


packages/connect/src/types/settings.ts
```diff
@@ -65,15 +65,21 @@ export interface ConnectSettingsWeb {
...
     deeplinkCallbackUrl?: string;
+    coreMode?: 'deeplink';
 }
 
 export type ConnectSettings = ConnectSettingsPublic &
     ConnectSettingsInternal &
-    ConnectSettingsWeb &
-    ConnectSettingsWebextension &
-    ConnectSettingsMobile;
+    // coreMode is a common parameter between these, so it is explicitly handled here for correct handling
```
mroz22, line 79: hmm I couldn't find a better way how to get rid of this nastiness quickly. it would probably require introducing some discriminator field each settings group and implement runtime checks where needed I guess

mroz22 (approved):

PR #17799, opened 2025-03-20
Fix undefined utility process

marekrjpolak:
## Description

Should fix kinda rare issue of trying to kill undefined utility process. ([sentry](https://satoshilabs.sentry.io/issues/6229277365/?project=5193825))


mroz22 (approved):
looks legit

PR #17800, opened 2025-03-20
feat(components): add children to the <Divider> component

peter-sanderson:
image
image

image
image

@coderabbitai ignore

### DIFF
--- a/packages/components/src/components/Divider/Divider.stories.tsx
+++ b/packages/components/src/components/Divider/Divider.stories.tsx
@@ -1,12 +1,16 @@
 import { Meta, StoryObj } from '@storybook/react';
 import styled from 'styled-components';
 
-import { Divider as DividerComponent, allowedDividerFrameProps } from './Divider';
+import { colorVariants } from '@trezor/theme';
+
+import { Divider as DividerComponent, DividerProps, allowedDividerFrameProps } from './Divider';
 import { getFramePropsStory } from '../../utils/frameProps';
 
-const Container = styled.div`
+const Container = styled.div<Pick<DividerProps, 'orientation'>>`
     width: 200px;
     height: 200px;
+    display: flex;
+    flex-direction: ${({ orientation }) => (orientation === 'horizontal' ? 'column' : 'row')};
 `;
 
 const meta: Meta = {
@@ -17,12 +21,15 @@ export default meta;
 
 export const Divider: StoryObj<typeof DividerComponent> = {
     render: props => (
-        <Container>
+        <Container orientation={props.orientation}>
             <DividerComponent {...props} />
         </Container>
     ),
     args: {
         orientation: 'horizontal',
+        contentPosition: 'center',
+        color: undefined,
+        children: undefined,
         ...getFramePropsStory(allowedDividerFrameProps).args,
     },
     argTypes: {
@@ -32,11 +39,25 @@ export const Divider: StoryObj<typeof DividerComponent> = {
             },
             options: ['horizontal', 'vertical'],
         },
+        contentPosition: {
+            control: {
+                type: 'select',
+            },
+            options: ['start', 'center', 'end'],
+        },
         strokeWidth: {
             type: 'number',
         },
         color: {
-            control: 'color',
+            control: {
+                type: 'select',
+            },
+            options: Object.keys(colorVariants.standard),
+        },
+        children: {
+            control: {
+                type: 'text',
+            },
         },
         ...getFramePropsStory(allowedDividerFrameProps).argTypes,
     },
--- a/packages/components/src/components/Divider/Divider.tsx
+++ b/packages/components/src/components/Divider/Divider.tsx
@@ -1,6 +1,8 @@
+import { ReactNode } from 'react';
+
 import styled, { css } from 'styled-components';
 
-import { Color, Elevation, mapElevationToBorder, spacings } from '@trezor/theme';
+import { Color, Elevation, SpacingValues, mapElevationToBorder, spacings } from '@trezor/theme';
 
 import {
     FrameProps,
@@ -10,6 +12,7 @@ import {
 } from '../../utils/frameProps';
 import { TransientProps } from '../../utils/transientProps';
 import { useElevation } from '../ElevationContext/ElevationContext';
+import { Column, Row } from '../Flex/Flex';
 
 export const allowedDividerFrameProps = [
     'margin',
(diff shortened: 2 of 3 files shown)


packages/components/src/components/Divider/Divider.stories.tsx
```diff
@@ -17,8 +20,11 @@ export default meta;
 
 export const Divider: StoryObj<typeof DividerComponent> = {
     render: props => (
-        <Container>
+        <Container orientation={props.orientation}>
             <DividerComponent {...props} />
+            <DividerComponent {...props}>
+                <Text variant="tertiary">OR</Text>
```
peter-sanderson: Question: should we allow only Text and force color to be aligned with the divider color?

Design for example:

image


I actually think no, or we shall force use `variants` only and remove `color` prop then.

packages/components/src/components/Divider/Divider.tsx
```diff
@@ -47,7 +51,7 @@ const Line = styled.div<
               `}
 
     background: ${({ theme, $elevation, $color }) =>
-        $color ? theme[$color] : mapElevationToBorder({ theme, $elevation })};
```
peter-sanderson: this was a bug, the color is just  string like `#bb34bb`

packages/components/src/components/Divider/Divider.tsx
```diff
@@ -56,22 +60,49 @@ export const Divider = ({
...
+            {...(children === undefined ? framePropsTransient : {})}
         />
     );
+
+    if (children === undefined) {
+        return line;
+    }
+
+    return orientation === 'horizontal' ? (
+        <Row gap={spacings.xxl} width="100%" alignItems="center" {...frameProps}>
```
peter-sanderson: Those `100%` are used to make it work same way as `Line` is behaving. I am not sure about this, but I suggest just to keep the behavior.

packages/components/src/components/Divider/Divider.tsx
peter-sanderson: :rage: ok its not..

packages/components/src/components/Divider/Divider.stories.tsx
```diff
@@ -32,11 +42,20 @@ export const Divider: StoryObj<typeof DividerComponent> = {
...
+        },
         strokeWidth: {
             type: 'number',
         },
         color: {
-            control: 'color',
+            control: {
+                type: 'select',
+            },
+            options: Object.keys(colorVariants.standard),
```
peter-sanderson, line 55: image

packages/components/src/components/Divider/Divider.stories.tsx
```diff
@@ -17,12 +22,17 @@ export default meta;
 
 export const Divider: StoryObj<typeof DividerComponent> = {
     render: props => (
-        <Container>
+        <Container orientation={props.orientation}>
             <DividerComponent {...props} />
+            <DividerComponent {...props}>
+                <Text variant="tertiary">OR</Text>
```
jvaclavik: I suggest to add children as a prop too and keep only one divider in storybook. Otherwise both components look like a one. Or you can also create a separate story

packages/components/src/components/Divider/Divider.tsx
```diff
@@ -23,6 +26,9 @@ export type DividerProps = AllowedFrameProps & {
     orientation?: DividerOrientation;
     strokeWidth?: number;
     color?: Color;
+    children?: ReactNode;
+    contentPosition?: 'start' | 'center' | 'end';
```
jvaclavik: Please, can you add comment "TODO unify with uiAlignments in the future"? I'm going to refactor these: 

```
export const uiHorizontalAlignments = ['left', 'center', 'right'] as const;
export type UIHorizontalAlignment = (typeof uiHorizontalAlignments)[number];

export const uiVerticalAlignments = ['top', 'center', 'bottom'] as const;
export type UIVerticalAlignment = (typeof uiVerticalAlignments)[number];

```

to this

```
export const uiAlignments = ['start', 'center', 'end'] as const;
export type UIAlignment = (typeof uiHorizontalAlignments)[number];
```

and it makes sense to use it here.
It's not ready yet, but it will be soon.

packages/components/src/components/Divider/Divider.tsx
```diff
@@ -47,31 +53,95 @@ const Line = styled.div<
               `}
 
     background: ${({ theme, $elevation, $color }) =>
-        $color ? theme[$color] : mapElevationToBorder({ theme, $elevation })};
+        $color ? theme[$color] : mapElevationToBorder({ theme, $elevation })};;
```
jvaclavik: typo: 2x `;`

packages/components/src/components/Divider/Divider.tsx
```diff
@@ -47,31 +53,95 @@ const Line = styled.div<
...
 
     ${withFrameProps}
 `;
 
+type DividerContentProps = {
+    line: ReactNode;
+    children?: ReactNode;
+    contentPosition?: 'start' | 'center' | 'end';
```
jvaclavik: Please add comment as above also here

packages/components/src/components/Divider/Divider.tsx
peter-sanderson: done

packages/components/src/components/Divider/Divider.tsx
peter-sanderson: done

packages/components/src/components/Divider/Divider.tsx
peter-sanderson: done

packages/components/src/components/Divider/Divider.stories.tsx
peter-sanderson: done

packages/components/src/components/Divider/Divider.tsx
jvaclavik: just leaving here this link: https://github.com/trezor/trezor-suite/pull/17804

packages/components/src/components/Divider/Divider.stories.tsx
jvaclavik: Actually I would probably restrict it only for text. No color or typography changes, all uppercase probably

peter-sanderson: I have designs with both uppercase and "normal" case.

peter-sanderson: I do the text only

jvaclavik: OK, devs can use `.toUpperCase()`

jvaclavik (approved):

PR #17807, opened 2025-03-20
chore: update babel, remove it from sub package.jsons

tomasklim:
## Related Issue

Resolve


packages/connect-web/package.json
```diff
@@ -47,7 +47,6 @@
         "@trezor/utils": "workspace:*"
     },
     "devDependencies": {
-        "@babel/preset-typescript": "^7.24.7",
```
mroz22, line 50: sometimes we need to duplicate dependencies in sub packages so that it works with yarn focus. not sure if this is also the case

komret (approved):

PR #17821, opened 2025-03-21
@trezor/websocket-client tuning

mroz22:
couple of minor changes needed for #17786  - this makes it possible to use the @trezor/websocket-client package directly without creating an intermediary subclass


marekrjpolak (approved):
Should be fine :+1:

PR #17823, opened 2025-03-21
Optimize state loading

vojtatranta:
## Description

Fasten app loading by 2s (~100%)

## Related Issue

Resolve

### DIFF
--- a/packages/suite/src/support/suite/preloadStore.ts
+++ b/packages/suite/src/support/suite/preloadStore.ts
@@ -24,26 +24,48 @@ export const preloadStore = async () => {
         } as const;
     }
 
-    // load state from database
-    const suiteSettings = await db.getItemByPK('suiteSettings', 'suite');
-    const devices = await db.getItemsExtended('devices');
-    const accounts = await db.getItemsExtended('accounts');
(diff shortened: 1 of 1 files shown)


packages/suite/src/support/suite/preloadStore.ts
```diff
@@ -24,26 +24,48 @@ export const preloadStore = async () => {
...
+        txs,
+        messageSystem,
+        backendSettings,
+        sendFormDrafts,
+        formDrafts,
+        coinjoinAccounts,
+        coinjoinDebugSettings,
+        tokenManagement,
+        security,
+    ] = await Promise.all([
```
vojtatranta, line 48: Speeds up loading from 4s on all seed to ~2s

vojtatranta: Used `Promise.all` over `Promise.allSettled` coz then I would need filter the results and that would lose the typing.

Currently, sequential `await` also meant that if any of the promises were rejected, the whole process would be ended and error thrown. SO it is the same as before.

jvaclavik (approved):
lovely ❤️

PR #17845, opened 2025-03-21
feat(connect-web): error fallback for core in desktop

martykan:
## Description

If core in desktop init/call fails due to websocket, fall back to core in popup.

### DIFF
--- a/packages/connect-web/src/impl/core-in-suite-desktop.ts
+++ b/packages/connect-web/src/impl/core-in-suite-desktop.ts
@@ -19,6 +19,7 @@ import type {
 } from '@trezor/connect/src/types';
 import { Login } from '@trezor/connect/src/types/api/requestLogin';
 import { WebsocketClient } from '@trezor/websocket-client';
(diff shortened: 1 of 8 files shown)


mroz22 (approved):
looks ok. what about some kind of "upgrade logic" when suite-deskop comes online again?

martykan:
Good point, we could probably try again before each call

packages/connect-web/src/impl/core-in-suite-desktop.ts
```diff
@@ -117,19 +115,14 @@ export class CoreInSuiteDesktop implements ConnectFactoryDependencies<ConnectSet
...
             }
 
             return response;
         } catch (err) {
             return {
                 success: false,
-                payload: {
-                    error: err.message,
-                },
+                payload: ERRORS.serializeError(err),
```
mroz22: I got this 
image
 
image

which does not look serialized 🤔

mroz22 (approved):

PR #17859, opened 2025-03-24
packages/websocket-client nitpicks

mroz22:
- field privatization
- unify script command in packages.json to allow `yarn workspace @trezor/websocket-client test:unit` 
- remove unneeded intermediary class in tests

### DIFF
--- a/packages/websocket-client/tests/client.test.ts
+++ b/packages/websocket-client/tests/client.test.ts
@@ -2,14 +2,7 @@ import { ServerOptions, WebSocket } from 'ws';
 
 import { WebsocketClient } from '../src/client';
 
-class Client extends WebsocketClient<{ 'foo-event': 'bar-event' }> {
-    ping() {
-        return this.sendMessage({ method: 'ping' });
-    }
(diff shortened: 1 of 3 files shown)


packages/websocket-client/tests/client.test.ts
```diff
@@ -95,6 +88,7 @@ describe('WebsocketClient', () => {
         jest.useFakeTimers();
 
         const cli = new Client({ url: server.getUrl(), pingTimeout: 5000 });
+        // @ts-expect-error ping is protected
         const pingSpy = jest.spyOn(cli, 'ping');
```
mroz22, line 92: ping is just sendMessage with payload { method: 'ping' }, and sendMessage is already public, so ping might be as well 🤷

marekrjpolak: It might be, but I prefer not to, at least until I see legitimate reason for pinging from outside.

marekrjpolak (approved):
Six lines out :arrow_right:  :heavy_check_mark:

PR #17865, opened 2025-03-24
iOS E2E tests fix

PeKne:
## Description
- Fix of iOS E2E test pipeline

  - passing action from this PR: https://github.com/trezor/trezor-suite/actions/runs/14035026162/job/39291003869?pr=17865 ✅ 

- iOS pipeline setuped to run every midnight (it used to be runned on weekly basis until now)

### DIFF
--- a/suite-native/app/e2e/utils.ts
+++ b/suite-native/app/e2e/utils.ts
@@ -90,7 +90,7 @@ export const scrollUntilVisible = async (
         await waitFor(element)
             .toBeVisible()
             .whileElement(by.id(scrollViewTestId))
-            .scroll(250, 'down');
+            .scroll(300, 'down');
     }
 };
 
--- a/.github/workflows/test-suite-native-e2e-ios.yml
+++ b/.github/workflows/test-suite-native-e2e-ios.yml
@@ -2,10 +2,7 @@ name: "[Test] suite-native iOS E2E"
 
 on:
   schedule:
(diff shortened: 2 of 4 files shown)


suite-native/app/e2e/utils.ts
```diff
@@ -90,7 +90,7 @@ export const scrollUntilVisible = async (
         await waitFor(element)
             .toBeVisible()
             .whileElement(by.id(scrollViewTestId))
-            .scroll(250, 'down');
+            .scroll(300, 'down');
```
BrantalikP, line 93: Just a question, but is it possible that if we change the testing device with different dimensions in the future, we might need to tweak this number again?🤔

PeKne: Yes, it is possible. I had to change this because of the buggy behavior of the Detox framework. This function was able to scroll 250 down so the clickable element was already peaking from behind the bottom edge of the screen, but after calling the `.tap()` method the Detox clicked on the bottom bar instead of the element and the test failed.

An alternative solution could be to extend this utility function by an extra `bottomScrollOffset` argument that would enforce scrolling by parametrized value even if the element is already partly visible so we would enforce the app to show the full height of the component. However, I wanted to keep this PR as simple as possible so I decided not to implement this enhancement.

BrantalikP (approved):
👌

PR #17868, opened 2025-03-24
fix(suite): simplify discovery, remove Promise from redux

### LINKED ISSUE #17597: A non-serializable value was detected in the state, in the path
Slack thread: https://satoshilabs.slack.com/archives/G019WLX2P7B/p1741695454577069

TL;DR: we are [storing a promise in the Redux state](https://github.com/trezor/trezor-suite/blob/d4d0f0a1c81693b6eb1d60934d2251227cad23e6/suite-common/wallet-core/src/discovery/discoveryReducer.ts#L54). We need to fix it.

Newly introduced [serializabilityMiddleware](https://redux-toolkit.js.org/api/serializabilityMiddleware) triggers the error.

Image

Lemonexe:
## Description

Proper fix after #17894 hotfix.

After through research, the defered promise (DFD) code is deemed useless and completely removed :monocle_face: :broom: 

## Related Issue

Resolve #17597 _properly_

## Details
The diff for [discoveryMiddleware.ts](https://github.com/trezor/trezor-suite/pull/17868/files#diff-fb6c5ddd60e84c647364e6cdbfb9311881b5dcd275c3c8d9a0dab95e63f293bc) is the pivotal point of this PR; read the file above & below for more context.

Here's an explanation what's going on in current state and what I'm changing:

- all redux actions pass through this middleware...
- ...but only some actions are marked as `interruptionIntent`:
  A) change device or passphrase
  B) navigate to somewhere else than `dashboard | wallet | settings`
  C) opening a connect popup (???)
- in case of a discovery running & an `interruptionIntent` action, middleware does
`await dispatch(stopDiscoveryThunk())` which:
  - calls Connect cancel
  - and _delays_ execution of that action until Connect confirms the discovery was interrupted
- delaying the action in the middleware cannot be done by awaiting the cancel (it's not awaitable)
  - only when periodically querying discovery state [here](https://github.com/trezor/trezor-suite/blob/9744c5b589214ee64dcdd7744a699da44bd8d755/suite-common/wallet-core/src/discovery/discoveryThunks.ts#L681) we can resolve the promise
  - and _that_ is why we needed a DFD kept in redux – to make `stopDiscoveryThunk` awaitable (meaning that Connect has cleared the previous discovery), while really it's a sync thunk that fetches DFD stored in redux, which is resolved from elsewhere, to make it look like async thunk :see_no_evil:  
  - the DFD was not awaited anywhere but in this middleware, through `stopDiscoveryThunk`
- the code below the discoveryMiddleware diff does: device (re)acqusition, auth and finally: discovery (re)start
- that's relevant only to A:
  - when changing device or passphrase, we need to cancel running discovery, and only after Connect reports that it has cleared discovery (DFD resolved from another thunk), we may start another discovery
- **but:** A is not allowed by UI when discovery is running at all, so we don't need to handle this case at all
- meanwhile, B can be dispatched during discovery, but it won't trigger new discovery, so we don't need to await connect clearing the previous discovery, we can let the action bubble through redux right away
  - example: navigate to `notifications` page in mobile view
    - I don't see why `notifications` page stops discovery, I think it's an oversight, but it can be used for testing 

@coderabbitai ignore

### DIFF
--- a/packages/suite/src/middlewares/wallet/discoveryMiddleware.ts
+++ b/packages/suite/src/middlewares/wallet/discoveryMiddleware.ts
@@ -66,10 +66,11 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
         // discovery interruption ends after DISCOVERY.STOP action
         // action which triggers this interruption will be propagated AFTER stop
         if (interruptionIntent && discoveryIsRunning) {
-            await dispatch(stopDiscoveryThunk());
+            dispatch(stopDiscoveryThunk());
         }
 
-        // pass action
+        // pass action to next middleware; code below will run only after the action has been fully processed in redux.
+        // Even though next(action) generally isn't async as per TS, the action may return anything. Sometimes it's a Promise → needs to be awaited
         await next(action);
 
         if (walletSettingsActions.changeNetworks.match(action)) {
--- a/suite-common/wallet-core/src/discovery/discoveryReducer.ts
+++ b/suite-common/wallet-core/src/discovery/discoveryReducer.ts
@@ -2,7 +2,6 @@ import { createReducerWithExtraDeps } from '@suite-common/redux-utils';
 import { DiscoveryStatus } from '@suite-common/wallet-constants';
 import { Discovery, PartialDiscovery } from '@suite-common/wallet-types';
 import { DeviceState, StaticSessionId } from '@trezor/connect';
-import { createDeferred } from '@trezor/utils';
 
 import { discoveryActions } from './discoveryActions';
 import { DeviceRootState, selectSelectedDevice } from '../device/deviceReducer';
@@ -17,18 +16,13 @@ export type DiscoveryRootState = {
 
 const initialState: DiscoveryState = [];
 
-const update = (draft: DiscoveryState, payload: PartialDiscovery, resolve?: boolean) => {
+const update = (draft: DiscoveryState, payload: PartialDiscovery) => {
     const index = draft.findIndex(f => f.deviceState === payload.deviceState);
     if (index >= 0) {
-        const dfd = draft[index].running;
         draft[index] = {
             ...draft[index],
             ...payload,
         };
-        if (resolve && dfd) {
-            dfd.resolve();
-            delete draft[index].running;
-        }
         if (!payload.error) {
             delete draft[index].error;
         }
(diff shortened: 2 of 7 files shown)


suite-common/wallet-core/src/discovery/discoveryReducer.ts
```diff
@@ -46,14 +40,7 @@ export const prepareDiscoveryReducer = createReducerWithExtraDeps(
...
             })
             .addCase(discoveryActions.startDiscovery, (state, { payload }) => {
-                const index = state.findIndex(f => f.deviceState === payload.deviceState);
-                if (index >= 0) {
-                    state[index] = {
-                        ...state[index],
-                        ...payload,
-                        running: createDeferred(),
-                    };
-                }
```
Lemonexe, line 56: this was the same code as in `update()`, but it also creates DFD, while `update` can only ignore it or resolve it..
But not important anymore.

packages/suite/src/middlewares/wallet/discoveryMiddleware.ts
```diff
@@ -66,10 +66,11 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
         // discovery interruption ends after DISCOVERY.STOP action
         // action which triggers this interruption will be propagated AFTER stop
         if (interruptionIntent && discoveryIsRunning) {
-            await dispatch(stopDiscoveryThunk());
+            dispatch(stopDiscoveryThunk());
```
Lemonexe, line 69: This is the pivotal point of this PR; read the file above & below for more context.
Here's a brief explanation:
- this is a middleware, so ofc **all actions** pass through here...
- but only some are marked as `interruptionIntent`:
  A) change device or passphrase _(or initial loading of first device)_ 
  B) navigate to somewhere else than `dashboard | wallet | settings`
- the code below this diff is relevant only to A) – device (re)acqusition, auth and discovery (re)start
WIP

Lemonexe:
Closing. Turns out, we still need to await interruption in cases of change device/passphrase.
When the change device/passphrase is triggered from UI, for some reason discovery starts still with the original wallet, only to be interrupted right after, _then_ the wallet is changed and discovery restarted.  
This is ofc less then ideal, suboptimal we might say! :hear_no_evil: 
But given the long-term goal of rewriting discovery into connect, it is not worth the effort to untangle this issue.

PR #17873, opened 2025-03-24
chore(suite-native): Improve store initialization in component tests

jbazant:
## Description

Simplify "Store is warming up" state detection in component tests.

## Related Issue

Resolve


vytick (approved):
nice improvement

PR #17874, opened 2025-03-24
WIP: feat: handle firmwares

karliatto:
WIP: handling firmware updates


packages/connect/src/data/firmwareInfo.ts
```diff
@@ -32,7 +44,52 @@ export const parseFirmwareReleases = (json: any, deviceModel: DeviceModelInterna
...
+    }
+    return releases[deviceModel] || [];
+};
+
+export const parseMessageRelease = async (): Promise<any> => {
+    const url = path.join(
+        __dirname,
+        '../../..',
+        'suite-common/message-releases/releases/test-releases.v1.json',
+    );
```
karliatto: This is just for Dev, this will be downloaded from data.trezor.io and it will be a signed file.

karliatto: We will most likely also bundle this file, so it can be used offline, and use it always to get the firmware to update. The current releases.json still will be necessary but we will use it for other things like the languages supported in older versions and other information from older FWs.

packages/connect/src/device/Device.ts
```diff
@@ -823,6 +860,8 @@ export class Device extends TypedEmitter<DeviceEvents> {
         const enabled = DataManager.getSettings('enableFirmwareHashCheck');
         if (!enabled || baseUrl === undefined) return createFailResult('check-skipped');
 
+        const localFirmwares = DataManager.getLocalFirmwares();
```
karliatto: I am implementing it now, but the idea is to have an interface that let us use the local firmware when they are available. So this function will maybe take model and version we need and response if it is suppose to be locally downloaded. And if so, use it.

karliatto: Right now the information I have from desktop is something like:

```
{
  userDataDir: '/home/user/.config/@trezor/suite-desktop-local',
  firmwares: [
    'trezor-t2t1-2.8.9-regular.bin',
    'trezor-t3b1-2.8.9-bitcoinonly.bin'
  ]
```
Which should be enough to be able use one of those firmware if we need them.

suite-common/message-releases/releases/test-releases.v1.json
```diff
@@ -0,0 +1,237 @@
+{
```
karliatto: This is an example of how the releases could be, it should be pretty much ready, but we still will go throw it to make it contain all the really required information for the releases and not more or less.

packages/connect/src/device/Device.ts
```diff
@@ -627,6 +647,23 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
+        if (!releaseInfo) {
+            throw new Error('TODO');
+        }
+        this.firmwaresDfd = createDeferred();
+        this.emit(DEVICE.FIRMWARE_REQUEST, {
+            releases: [releaseInfo],
+            device: this.toMessageObject(),
+        });
+
+        await this.firmwaresDfd.promise;
```
karliatto: It looks like waiting here is not good, so I am searching for different approach to this.

suite-common/message-releases/jest.config.js
mroz22, line 1: this is some new approach in general? or something specific to this package?

packages/suite-desktop-core/src/modules/trezor-connect.ts
```diff
@@ -91,7 +91,10 @@ export const initBackground: ModuleInitBackground = ({ mainThreadEmitter, store
     const unregisterProxy = createIpcProxyHandler(ipcMain, 'TrezorConnect', ipcProxyOptions);
 
     const onLoad = () => {
+        // TODO(karliatto): this event listener is actually not usable since
+        // connect is dispose in onLoad init below in this file when
```
mroz22, line 98: what do you mean?  that it disappears after reload?  or something else?

packages/connect/src/api/firmware/getBinaryForFirmwareUpgrade.ts
```diff
@@ -66,3 +66,13 @@ export const getBinaryForFirmwareUpgrade = ({
 
     return getBinary({ release: releaseByFirmware, baseUrl, btcOnly });
 };
+
+const ALL_SLASHES_AT_THE_END_REGEX = /\/+$/;
```
mroz22: https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/api/firmware/getBinary.ts#L4

suite-common/message-releases/jest.config.js
karliatto: I did not work on the tests, this is just the boilerplate from default from `yarn workspace @trezor/scripts generate-package`

packages/connect/src/api/firmware/getBinaryForFirmwareUpgrade.ts
karliatto: Yes, that is just copy paste for development. I would get only one of those at the end.

packages/suite-desktop-core/src/modules/trezor-connect.ts
karliatto: As you can see few lines below, there is ` TrezorConnect.dispose();` and that cancel all the listeners that were registered before, it seams.

packages/connect/src/device/Device.ts
karliatto: Right now I am making it here wait when device is connected but it creates some issues, so I am thinking that actually it could be done when Connect is initialized, it will request for all the firmwares already present and that information will be accessible from DataManager.

Since I realized there is not much value on waiting for device, since we can just ask for all the available firmwares in config directory. And then when they are needed we will just use them.

@mroz22 Where do you think would be best place to do that, so we can make connect initialization wait for that? Other option would be to get that information as a Connect initialization param.

mroz22 (changes requested):
please split it into some logical commits 🙏 if there are some pure refactoring changes such as renaming or similar, also lets please move them to a separate commit and possibly lets merge them to decrease the size of the diff we have here

karliatto:
> please split it into some logical commits 🙏 if there are some pure refactoring changes such as renaming or similar, also lets please move them to a separate commit and possibly lets merge them to decrease the size of the diff we have here

Yes, that is the idea, this is the first one https://github.com/trezor/trezor-suite/pull/18435

karliatto:
This has been implemented in https://github.com/trezor/trezor-suite/pull/18821

PR #17882, opened 2025-03-24
feat(suite): store only active network fees

### LINKED ISSUE #17773: Keep only activated networks' fees in redux
Image

izmy:
## Description

Fee management logic is updated to store only active network fees in Redux, removing the need to keep all possible networks with their initial values. Now, the Redux store holds fees only for currently active networks, which improves memory usage and performance.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17773


packages/suite/src/components/suite/modals/ReduxModal/TransactionReviewModal/TransactionReviewSummary.tsx
```diff
@@ -14,9 +14,11 @@ import { Account } from 'src/types/wallet';
 
 const getEstimatedTime = (
     networkType: NetworkType,
-    symbolFees: FeeInfo,
+    symbolFees: FeeInfo | undefined,
     tx: GeneralPrecomposedTransactionFinal,
 ): number | undefined => {
+    if (!symbolFees) return;
```
tomasklim: nit:
```suggestion
    feeInfo: FeeInfo | undefined,
    tx: GeneralPrecomposedTransactionFinal,
): number | undefined => {
    if (!feeInfo) return;
```

packages/suite/src/hooks/wallet/useRbfForm.ts
```diff
@@ -98,7 +99,9 @@ const useRbfState = ({ selectedAccount, rbfParams, chainedTxs }: UseRbfProps) =>
     const { shouldSendInSats } = useBitcoinAmountUnit(account.symbol);
 
     return useMemo(() => {
-        const feeInfo = getRbfFeeInfo(account.networkType, symbolFees, rbfParams);
+        const feeInfo = symbolFees
```
tomasklim: nit: 
```suggestion
        const rbfFeeInfo = feeInfo ? ...
```

suite-common/wallet-core/src/fees/feesThunks.ts
```diff
@@ -0,0 +1,184 @@
...
+            dispatch(feesActions.updateFee(partial));
+        }
+    },
+);
+
+export const removeFeeInfoThunk = createThunk(
+    `${FEES_MODULE_PREFIX}/removeFeeInfoThunk`,
+    ({ networks }: { networks: NetworkSymbol[] }, { dispatch, getState }) => {
+        const fees = selectFees(getState());
+        const activeFees = Object.keys(fees) as NetworkSymbol[];
```
tomasklim: nit: something like this would be better
```suggestion
        const networksWithFees = Object.keys(fees) as NetworkSymbol[];
```

tomasklim (approved):

Issue #17883, opened 2025-03-25
Connect popup E2E test failing


PeKne:
For some reason, the connect popup E2E test started failing recently (example pipeline: https://github.com/trezor/trezor-suite/actions/runs/14048654475/job/39335004157). 

Investigate and fix it.

PR #17884, opened 2025-03-25
test/native/flaky-android-e2e

PeKne:
## Description
- send review cancellation test removed. It is hard to debug and the test is not so critical to invest more time in it
- deep link popup connect is skipped for now and will be fixed and enabled soon (issue: https://github.com/trezor/trezor-suite/issues/17883)


BrantalikP (approved):
🟢

PR #17894, opened 2025-03-25
fix(suite): move Promise from redux to a local object

### LINKED ISSUE #17597: A non-serializable value was detected in the state, in the path

Lemonexe:
## Description

A temporary solution that solves error, that surfaced when bumping Redux, because we have been storing non-serializable values in Redux (previous version did not mind, but it was a code smell for years).
It's a defered promise (DFD) in `state.wallet.discovery[i].running`.

The whole logic should ideally be removed, as it does not seem to be necessary. I'm working on that in #17868 but there are still some obstacles, hence this "quick fix" :disappointed: 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17597

@coderabbitai ignore

### DIFF
--- a/packages/suite/src/actions/suite/storageActions.ts
+++ b/packages/suite/src/actions/suite/storageActions.ts
@@ -19,11 +19,7 @@ import type { PreloadStoreAction } from 'src/support/suite/preloadStore';
 import type { AppState, Dispatch, GetState, TrezorDevice } from 'src/types/suite';
 import type { Account } from 'src/types/wallet';
 import { GraphData } from 'src/types/wallet/graph';
-import {
-    serializeCoinjoinAccount,
-    serializeDevice,
-    serializeDiscovery,
-} from 'src/utils/suite/storage';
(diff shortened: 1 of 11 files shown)


peter-sanderson (approved):
Seems to be easy and obvious now, but hell, ...  it must have been a real detective work :detective: 

GJ :clap:

Lemonexe:
@peter-sanderson actually this is the easy dirty fix, not rly proud of it.
The detective work went into attempt to remove the async logic altogether, which I believe is possible, but I'll see if I can untangle that mess.

PR #17895, opened 2025-03-25
fix(nix): add vips build dependency of sharp@npm

karliatto:
## Description
Latest version of sharp@npm requires vips (Image processing system for large images) dependency in the system.
Adding it to `shell.nix` so we can run `yarn install`.


karliatto:
Closing since it was already solved.

PR #17896, opened 2025-03-25
fix(suite-native): report potential errors from connectInit

matejkriz:
## Description

- It is necessary to unwrap the result because async thunk from redux-toolkit which is always resolved. See more details here: https://redux-toolkit.js.org/api/createAsyncThunk#unwrapping-result-actions
- Global try-catch was useless for wrapped dispatches.

### DIFF
--- a/suite-native/app/src/initActions.ts
+++ b/suite-native/app/src/initActions.ts
@@ -24,43 +24,43 @@ export const applicationInit = createThunk(
             return;
         }
 
+        dispatch(initAnalyticsThunk());
+
+        dispatch(initMessageSystemThunk());
+
+        // Select latest remembered device or Portfolio Tracker device.
+        dispatch(initDevices());
+
         try {
-            dispatch(initAnalyticsThunk());
-
-            dispatch(initMessageSystemThunk());
-
-            // Select latest remembered device or Portfolio Tracker device.
-            dispatch(initDevices());
-
-            await dispatch(connectInitThunk());
-
-            dispatch(setIsConnectInitialized(true));
-
-            dispatch(initBlockchainThunk());
-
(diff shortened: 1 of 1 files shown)


suite-native/app/src/initActions.ts
```diff
@@ -24,43 +24,43 @@ export const applicationInit = createThunk(
...
-            // Select latest remembered device or Portfolio Tracker device.
-            dispatch(initDevices());
+        // Select latest remembered device or Portfolio Tracker device.
+        dispatch(initDevices());
 
-            await dispatch(connectInitThunk());
+        try {
+            await dispatch(connectInitThunk()).unwrap();
+        } catch (error) {
+            console.error(`Connect init error: ${JSON.stringify(error)}`);
```
BrantalikP, line 37: in case of failure do we want to continue and call `dispatch(setIsConnectInitialized(true));` and other actions?

BrantalikP (changes requested):

suite-native/app/src/initActions.ts
matejkriz: Good point, we probably don't want to continue, but so far we don't have an Error Boundary component to catch such an error. 

This PR is not changing the user-facing behavior, it only fixes the error reporting. I can create a follow-up issue to introduce ErrorBoundary and add rethrow here.

BrantalikP (approved):
:shipit:

suite-native/app/src/initActions.ts
matejkriz: https://github.com/trezor/trezor-suite/issues/18184

STew790:
**QA OK**
No issue when testing the app.

**Info**
25.4.1 1e1a7dc

PR #17899, opened 2025-03-25
fix(suite): finish connect popup storage

martykan:
## Description

Add migrations and middleware to actually persist remembered Connect popup permissions in storage.


mroz22 (approved):
looks legit 🆗

PR #17900, opened 2025-03-25
WIP: suite(refactor): remove the need for observing of the selected device…

vojtatranta:
… thanks to the storing just a selector to the devices list

## Description

This PR removes the need to sync the `device.selectedDevice` state with middlewares. The solution to it was to create selected device selector in state, that would only describe which of the `devices` in the list should be selected rather then copyingthe whole object all the time.

## The challenges faced:
- which of the properties use to select the device current order:
  -  `staticSessionId` = hash of device and passphrase
  - `path` = that's generated ID of the device selected
  - `instance` = the order in which the device wallet was created. eg. when you add a new passphrase wallet, the device is the same only "another wallet on the device" is made

## Possible issues
- miryad of problems regarding adding wallets, removing wallets, remembering devices, forgetting devices, switching devices, reconnecting etc.

## Related Issue

Reason to do it: I was working on the passphrase wallet and I couldn't find how things are influenced with and noticed this which was possible to change

### DIFF
--- a/packages/suite/src/middlewares/suite/suiteMiddleware.ts
+++ b/packages/suite/src/middlewares/suite/suiteMiddleware.ts
@@ -1,18 +1,13 @@
-import { isAnyOf } from '@reduxjs/toolkit';
 import { MiddlewareAPI } from 'redux';
 
-import { AnyAction } from '@suite-common/redux-utils';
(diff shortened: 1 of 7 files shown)


vojtatranta:
I have been writing my findings into this issue, it is extremely messed up https://github.com/trezor/trezor-suite/issues/17911

PR #17901, opened 2025-03-25
feat(suite): update recovery modal

seibei-iguchi:
# Before

## Recovery dry run (T1B1):
Screenshot 2025-03-26 at 13 19 00
Screenshot 2025-03-26 at 13 19 10
Screenshot 2025-03-26 at 13 19 18
Screenshot 2025-03-26 at 13 19 27

### Standard recovery
Screenshot 2025-03-26 at 13 19 44

### Advanced recovery
Screenshot 2025-03-26 at 13 20 14

## Recovery dry run (non T1B1}:
Screenshot 2025-03-26 at 13 25 30
Screenshot 2025-03-26 at 13 25 39
Screenshot 2025-03-26 at 13 26 10

## Update pin and wipe code (T1B1):
Screenshot 2025-03-26 at 13 22 12
Screenshot 2025-03-26 at 13 22 22
Screenshot 2025-03-26 at 13 22 29
Screenshot 2025-03-26 at 13 23 07
Screenshot 2025-03-26 at 13 23 16
Screenshot 2025-03-26 at 13 22 44

# After

## Recovery dry run (T1B1):
Screenshot 2025-03-26 at 11 51 36
Screenshot 2025-03-26 at 11 51 52
Screenshot 2025-03-26 at 11 52 02
Screenshot 2025-03-26 at 11 53 15

### Standard recovery
Screenshot 2025-03-26 at 11 52 32

### Advanced recovery
Screenshot 2025-03-26 at 11 53 31

## Recovery dry run (non T1B1}:
Screenshot 2025-03-26 at 11 50 34
Screenshot 2025-03-26 at 11 50 49
Screenshot 2025-03-26 at 11 51 01

## Update pin and wipe code (T1B1):
Screenshot 2025-03-26 at 11 57 03
Screenshot 2025-03-26 at 13 06 30
Screenshot 2025-03-26 at 13 06 38
Screenshot 2025-03-26 at 13 07 13
Screenshot 2025-03-26 at 13 08 41
Screenshot 2025-03-26 at 13 08 55


seibei-iguchi:
@coderabbitai summary

packages/suite/src/views/recovery/steps/InitialStep.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    </Paragraph>
+                </BulletList.Item>
+            </BulletList>
+            <Card margin={{ top: spacings.xxl }}>
+                <CheckItem
+                    data-testid="@recovery/user-understands-checkbox"
+                    title={<Translation id="TR_DRY_RUN_CHECK_ITEM_TITLE" />}
+                    description={<Translation id="TR_DRY_RUN_CHECK_ITEM_DESCRIPTION" />}
+                    isChecked={isUnderstood}
+                    link={learnMoreUrl && <LearnMoreButton url={learnMoreUrl} />}
```
Lemonexe, line 86: `LearnMoreButton` does not stop propagation of onClick event, so if you click "Learn more", it toggles the "I understand" checkbox.
Yeah I know, not an issue of grave importance, but if it's easy to fix it'd be nice :smiley:

packages/suite/src/components/suite/PinMatrix/PinMatrix.tsx
```diff
@@ -1,116 +1,160 @@
...
+                            />
+                        ))}
+                    </Grid>
+                    <Row gap={spacings.md}>
+                        <Input
+                            readOnly
+                            type="password"
+                            value={pin.replace(/[0-9]/g, '●')}
+                            size="small"
+                        />
```
Lemonexe: :thought_balloon: Do you think you could darken the font color, or increase font size/weight? The written dots are very small and low contrast, easily overlooked. Just a thought,

packages/suite/src/components/suite/PinMatrix/PinMatrix.tsx
```diff
@@ -1,116 +1,160 @@
...
+                </Banner>
             )}
-            <Col>
-                <PinInput isSubmitting={submitted} onPinSubmit={submit} />
-            </Col>
-        </Wrapper>
+            <Card paddingType="none">
+                <Column gap={spacings.xl} padding={spacings.xxl} data-testid="@pin">
+                    <Grid columns={3} gap={spacings.lg} width="100%">
+                        {['7', '8', '9', '4', '5', '6', '1', '2', '3'].map(value => (
```
Lemonexe: Having a laptop, this took a me while to understand that it visually maps to keyboard layout, and here I was wondering why I've got pin mismatch everytime I use the hotkeys :see_no_evil: 
Could you add comment that explains it, for slower people like me? :sweat_smile: (I mean just "the order visually maps to keyboard layout" or something)

packages/suite/src/components/suite/PinMatrix/PinMatrix.tsx
```diff
@@ -1,116 +1,160 @@
...
-    @media (max-width: ${variables.SCREEN_SIZE.SM}) {
-        width: 100%;
-    }
-`;
+type PinMatrixProps = {
+    pin: string;
+    setPin: (pin: string) => void;
+    onSubmit: () => void;
+    showExplanation?: boolean;
+    isInvalid?: boolean;
```
Lemonexe: unused prop. The prop is set in `PinModal`, but to no effect.

Lemonexe (commented):
I'll continue tomorrow

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PinModal.tsx
```diff
@@ -1,58 +1,83 @@
...
-            description={
-                <Translation
-                    id="TR_THE_PIN_LAYOUT_IS_DISPLAYED"
-                    values={{ deviceLabel, b: text => <b>{text}</b> }}
-                />
-            }
-            onCancel={onCancel}
-            isCancelable
-            data-testid="@modal/pin"
-            $isExtended={isModalExtended}
```
Lemonexe, line 48: `isModalExtended` is unused and can be removed in `usePin` hook

packages/suite/src/views/recovery/steps/InitialStep.tsx
```diff
@@ -0,0 +1,89 @@
...
+                <BulletList.Item
+                    title={
+                        <Paragraph typographyStyle="hint" textWrap="pretty">
+                            <Translation
+                                id={pickByDeviceModel(deviceModelInternal, {
+                                    default: 'TR_SEED_WORDS_ENTER_TOUCHSCREEN',
+                                    [DeviceModelInternal.T1B1]: 'TR_SEED_WORDS_ENTER_COMPUTER',
+                                    [DeviceModelInternal.T2B1]: 'TR_SEED_WORDS_ENTER_BUTTONS',
+                                    [DeviceModelInternal.T3B1]: 'TR_SEED_WORDS_ENTER_BUTTONS',
+                                })}
```
Lemonexe: Not a big fan of `default` behaviors. It can lead to bugs that are not easy to discover. Just my opinion, but if you agree, could you redo this as map with exhaustive typing?

By that I mean define something like this in the top scope of this file:
```
const enterSeedInstructionsMap: Record<DeviceModelInternal, TranslationKey> = {
   ... all models here explicitly
}
```

packages/suite/src/views/recovery/index.tsx
```diff
@@ -87,116 +83,55 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
 
                 // Do not rely on Capability_PassphraseEntry feature. For ancient firmwares it's not there,
                 // and we want to allow devices that have unsupported FW to be able to check the seed.
-                if (device.features.internal_model === DeviceModelInternal.T1B1) {
-                    return <T1B1InputStep />;
+                if (isT1B1) {
+                    switch (modal.windowType) {
+                        case UI.REQUEST_PIN:
+                            return (
+                                <PinMatrix pin={pin} setPin={setPin} onSubmit={handlePinSubmit} />
```
Lemonexe: isn't it possible to delegate this to `ReduxModal`?
That was the way it originally worked, when both were ~Modal~. Then it was not possible to combine them if one was ~Modal~ and other NewModal, but now they're both NewModal and it should be possible again, right?

packages/suite/src/views/recovery/index.tsx
```diff
@@ -1,69 +1,60 @@
...
 
 import {
+    SeedInputStatus,
     checkSeed,
     setAdvancedRecovery,
     setStatus,
     setWordsCount,
 } from 'src/actions/recovery/recoveryActions';
 import { MODAL } from 'src/actions/suite/constants';
-import { SelectRecoveryType, SelectWordCount } from 'src/components/recovery';
```
Lemonexe, line 21: These still exist and are used in onboarding. Do you also intend to refactor them to `SelectWordCountStep` and `SelectRecoveryTypeStep` later, to deduplicate it?
I definitely do not mean in this PR :smile: Just asking so it won't be forgotten

packages/suite/src/views/recovery/index.tsx
```diff
@@ -231,47 +166,148 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
-                    values={{
-                        index: statesInProgressBar.indexOf(recovery.status) + 1,
-                        total: statesInProgressBar.length,
-                    }}
+        <NewModal.Backdrop>
+            {['in-progress', 'waiting-for-confirmation'].includes(recovery.status) && (
+                <ConfirmOnDevice
+                    title={<Translation id="TR_CONFIRM_ON_TREZOR" />}
+                    deviceModelInternal={device?.features?.internal_model}
+                    deviceUnitColor={device?.features?.unit_color}
```
Lemonexe:
```suggestion
                    deviceModelInternal={device.features.internal_model}
                    deviceUnitColor={device.features.unit_color}
```

packages/suite/src/views/recovery/index.tsx
```diff
@@ -87,116 +83,55 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
+                            return <WordInputAdvanced count={6} />;
+                        case 'WordRequestType_Matrix9':
+                            return <WordInputAdvanced count={9} />;
+                        // to satisfy TS, should not happen
+                        default:
+                            return null;
+                    }
                 }
 
                 return <EnterOnDeviceStep />;
```
Lemonexe: here you could pass `deviceModelInternal` as prop, which would simplify the `EnterOnDeviceStep` code – it will have no need for the condition `const deviceModelInternal = device?.features?.internal_model || DEFAULT_FLAGSHIP_MODEL;`

packages/suite/src/views/recovery/steps/EnterOnDeviceStep.tsx
```diff
@@ -1,9 +1,28 @@
...
+    const { device } = useDevice();
+    const deviceModelInternal = device?.features?.internal_model || DEFAULT_FLAGSHIP_MODEL;
+
+    return (
+        <Banner
+            variant="info"
+            icon={
+                deviceModelInternal === 'UNKNOWN'
+                    ? undefined
+                    : (`trezor${deviceModelInternal}` as IconName) // TODO: update to non-deprecated icons
```
Lemonexe: Unnecessary `as`, it passes type check
```suggestion
                    : (`trezor${deviceModelInternal}`) // TODO: update to non-deprecated icons
```

packages/suite/src/views/recovery/index.tsx
```diff
@@ -87,116 +83,55 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
+                        <Box padding={spacings.xxl}>
+                            <Loading />
+                        </Box>
+                    );
                 }
 
                 // Do not rely on Capability_PassphraseEntry feature. For ancient firmwares it's not there,
                 // and we want to allow devices that have unsupported FW to be able to check the seed.
-                if (device.features.internal_model === DeviceModelInternal.T1B1) {
-                    return <T1B1InputStep />;
```
Lemonexe, line 195: nicely refactored, good riddance! :+1:

packages/suite/src/views/recovery/index.tsx
```diff
@@ -87,116 +83,55 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
+                            return <RequestConfirmationStep />;
+                        case 'WordRequestType_Plain':
+                            return <WordInputStep />;
+                        case 'WordRequestType_Matrix6':
+                            return <WordInputAdvanced count={6} />;
+                        case 'WordRequestType_Matrix9':
+                            return <WordInputAdvanced count={9} />;
+                        // to satisfy TS, should not happen
+                        default:
+                            return null;
```
Lemonexe: TS passes without it :heavy_check_mark: which is good because these defaults are not nice, can lead to unexpected bugs in future.
```suggestion
                        // to satisfy TS, should not happen
                        default:
                            return null;
```

packages/suite/src/views/recovery/index.tsx
```diff
@@ -231,47 +166,148 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
         }
     };
 
+    const getBottomContent = () => {
```
Lemonexe: small nit: rename to `getBottomContentPrimaryAction` because it is not the whole bottom content. Just the primary action besides Cancel/Close which is always there.

packages/suite/src/views/recovery/index.tsx
```diff
@@ -231,47 +166,148 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
-            onCancel={handleClose}
-            variant={hasFinished && hasError ? 'warning' : 'primary'}
-            // eslint-disable-next-line no-nested-ternary
-            iconName={hasFinished ? (hasError ? 'warning' : 'check') : undefined}
-        >
-            {getStep()}
-        </NewModal>
+                    </>
+                }
+                onBackClick={
```
Lemonexe: I would extract this to a function, just for better code readability (shorten the JSX).

packages/suite/src/views/recovery/index.tsx
```diff
@@ -231,47 +166,148 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
+                    );
+                }
+        }
+    };
+
+    const getIconName = () => {
+        switch (recovery.status) {
+            case 'finished':
+                return hasError ? 'warning' : 'check';
+        }
```
Lemonexe, line 237: is switchCase really necessary when you have only one case? I'd simplify it with an `if`.
And `return undefined` at the end, so it is explicitly apparent what is returnable

packages/suite/src/views/recovery/index.tsx
```diff
@@ -231,47 +166,148 @@ export const Recovery = ({ onCancel }: ForegroundAppProps) => {
...
+    };
+
+    const getVariant = () => {
+        switch (recovery.status) {
+            case 'in-progress':
+                return 'info';
+            default:
+                return hasError ? 'warning' : 'primary';
+        }
```
Lemonexe, line 245: same here, I'd simplify it to if with early return and default return

packages/suite/src/views/recovery/index.tsx
seibei-iguchi: I tried but it didn't work anyway. And it still felt better this way, as the flow is not interrupted (the heading remains the same)

packages/suite/src/views/recovery/index.tsx
seibei-iguchi: Yes, onboarding will also be refactored:)

packages/suite/src/views/recovery/index.tsx
Lemonexe: Oh right. I did not realize, but heading and bottom content should have continuity. Then no objection :slightly_smiling_face:

packages/suite/src/components/suite/WordInput.tsx
```diff
@@ -1,11 +1,9 @@
-import { Children, ReactElement, memo, useEffect, useRef } from 'react';
+import { Children, ReactElement, memo, useEffect, useRef, useState } from 'react';
 import { MenuListProps, SelectInstance, createFilter } from 'react-select';
 import { useKeyPress } from 'react-use';
 import { FixedSizeList as List } from 'react-window';
```
jvaclavik: Do we need this? Looks like we use `react-window` just because of this. And we have only 2048 items here AFAIK. Couldn't we delete it? 🤔

packages/suite/src/components/suite/WordInput.tsx
```diff
@@ -53,47 +31,51 @@ const MenuList = (props: MenuListProps<Option, boolean>) => {
...
+            isSearchable
+            isClearable={false}
+            isMenuOpen={!isLoading}
+            isLoading={isLoading}
+            noOptionsMessage={({ inputValue }: { inputValue: string }) =>
+                translationString('TR_WORD_DOES_NOT_EXIST', { word: inputValue })
+            }
+            onChange={async (item: Option, ref?: SelectInstance<Option, boolean> | null) => {
+                setIsLoading(true);
+                await resolveAfter(1000);
```
Lemonexe: What is the reason to wait for 1000ms? 
It's not clear from code, so if it's really necessary, comment would be beneficial

packages/suite/src/components/suite/WordInput.tsx
```diff
@@ -53,47 +31,51 @@ const MenuList = (props: MenuListProps<Option, boolean>) => {
...
+                TrezorConnect.uiResponse({ type: UI.RECEIVE_WORD, payload: item.value });
+                ref?.clearValue();
+                setIsLoading(false);
+                setTimeout(() => {
+                    ref?.focusInput();
+                });
```
Lemonexe: Multiple questions about this:
- sure it's necessary at all? I tried it locally without it, and seems to me that after the 1000 ms, the `autofocus` works correctly. Or are there some edge cases?
- if so, then are you sure it needs to be wrapped in instant `setTimeout`?
- if it needs to be focused manually, then a comment would be appreciated 
```suggestion
```

packages/suite/src/components/suite/WordInput.tsx
Lemonexe: OK I see that it was here before, you only increased it from 600 to 1000.. And I see locally, that the select breaks horribly without the timer :scream: So yeah, it's sadly necessary :see_no_evil:  
But using arbitrary timeouts is IMO a nasty code smell. It'd be ideal to find a a way to await device response from Connect instead of relying on timer.

No need to do it in this PR, but would you please create an issue, and link it with a TODO comment here?

seibei-iguchi: There was a 600 ms delay before. Not sure, it just felt natural that the action is not immediate:)

packages/suite/src/components/suite/WordInput.tsx
seibei-iguchi: It didn't work without the setTimeout. The autofocus didn't work for me either, apart from the first word:)

packages/suite/src/components/suite/WordInput.tsx
seibei-iguchi: Not sure, TBH. I can try.

packages/suite/src/components/suite/WordInputAdvanced.tsx
```diff
@@ -107,77 +92,59 @@ export const WordInputAdvanced = ({ count }: WordInputAdvancedProps) => {
...
+                                <PinButton data-value="5" onClick={() => onSubmit('4')} />
+                                <PinButton data-value="6" onClick={() => onSubmit('6')} />
+                                <PinButton
+                                    data-value="2"
+                                    onClick={() => onSubmit('1')}
+                                    data-testid="@recovery/word-input-advanced/1"
+                                />
+                                <PinButton data-value="3" onClick={() => onSubmit('3')} />
+                            </>
+                        )}
```
Lemonexe, line 150: as discussed personally, this is really obscure code.
Ofc no need to refactor it now, but I'd at least
- place TODO to investigate if we need `data-value` at all
- disable hotkeys definitely when `count === 6`, because then all hotkeys send something, but 3 of the 9 hotkeys send payload to device that is likely invalid (I have no idea what happens then?)
- consider removing the hotkeys, but idk. It's only T1B1, so no need for such first-class support, but then again, for PIN or 9words input it works well, and it doesn't hurt us to just keep that code here

packages/suite/src/views/onboarding/UnexpectedState/index.tsx
```diff
@@ -60,33 +65,35 @@ const UnexpectedState = ({ children }: UnexpectedStateProps) => {
...
     if (!activeStep) {
         return null;
     }
 
-    const pinComponent = getPinComponent();
-    if (pinComponent) {
+    // After the PIN is set it may happen that it takes too long for an user to finish the onboarding process.
+    // Then the device will get auto locked and requests to show a PIN matrix next before changing its setting.
+    // (which could happen on Final step where we set device name and homescreen)
+    if ((!device?.features || activeStepId === 'set-pin') && showPinMatrix) {
```
Lemonexe: idk how to reproduce the so called Unexpected States, but moving the condition here from getPinComponent is much nicer :+1: 
:point_right:  But I believe you forgot to negate the `(!device?.features || activeStepId === 'set-pin')`
(if either is truth, then it should be null -> don't render this).

Though I'd invert that negation to simplify it:
```suggestion
    if (device?.features && activeStepId !== 'set-pin' && showPinMatrix) {
```

Lemonexe (commented):
Finished review now. That was quite a ride :smile: I tried to be thorough, hope I didn't miss something given the size of the PR.

I wrote many comments, but most of them are nits or discussion of preexisting issues... Overall I have to say I love this! :raised_hands: It's really huge improvement in UI and code structure as well.

:warning: but the failing test is related to T1B1 advanced recovery, so please fix that, then I'll give :heavy_check_mark:

packages/suite/src/components/suite/WordInput.tsx
seibei-iguchi: I removed it. It's slightly laggy, but I think it feels ok. What do you think, will you try it yourself?

packages/suite/src/components/suite/WordInput.tsx
```diff
@@ -14,33 +12,11 @@ const options = bip39.map(item => ({ label: item, value: item }));
...
-            width="100%"
-        >
-            {({ index, style }) => <div style={style}>{children[index]}</div>}
-        </List>
-    );
+const styles: StylesConfig<Option, boolean> = {
+    menuList: base => ({
+        ...base,
+        maxHeight: '180px',
+    }),
```
Lemonexe, line 19: great simplifaction :+1: 
I didn't even notice before what we had here. Kudos for identifying a useless package!
Idk if it keeps 1:1 functionality, but IMO it doesn't have to. This works well enough, with less complexity.

Lemonexe (approved):
LGTM codewise and works well.
:warning: There is still the E2E test. After fixing it, :ship: :shipit: :rocket:

bosomt:
**QA OK**

checked:
- enter PIN 
- reenter PIN
- Check wallet backup
- [Learn more](https://trezor.io/learn/a/pin-protection-on-trezor-devices) link
- PIN mismatch! 
- Enter Wipe Code
- https://trezor.io/learn/a/advanced-recovery-on-model-one link
- advanced recovery
- recovery

**Info:**
 - Suite version: web 25.4.0 (e99e6ddcbb061d9c413e15a96c4dfa32e0c5bbd4)
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T1B1 1.11.1 regular (revision 85a26d2c9593bcdf858c2d718d79951ca927a0c3)
 - Transport: WebUsbTransport

PR #17908, opened 2025-03-26
connect: sign and verify methods - add message lenght connect-side validation

### LINKED ISSUE #17893: Bytes overflow in latest firmware version
**Describe the bug**
Signing a quite large message is resulting in a "bytes overflow" error.

Image

**Firmware version and revision**
v1.13.0 (Trezor One)

**Desktop/smartphone setup (please complete the following information):**
 - Wallet software: Rabby
 - OS: Linux
 - Browser: Brave
 - Version 1.76.81

**To Reproduce**
1. Go to https://www.atomiclore.io/nftProfile/base/0xaf2430e1d91c7e22bb7d84588bccf8bc8301a09c/154
2. Click on `Sign`
3. Select `Comment` in the Provenance Templates
4. Paste below comment
```
blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah
``` 
5. Click `Continue` then `Sign`
6. See error

**Expected behavior**
I am able to sign any message regardless of its size.

**Additional context**
I was able to sign the message above (or at least a message of equivalent character count) just fine from a non-Trezor account via Rabby.

mroz22:
image

resolve #17893

### DIFF
--- a/packages/connect/e2e/__fixtures__/signMessage.ts
+++ b/packages/connect/e2e/__fixtures__/signMessage.ts
@@ -244,5 +244,21 @@ export default {
             },
             legacyResults,
         },
+        {
+            description: 'Message over maximum limit allowed by T1B1',
+            skip: ['2'],
+            params: {
+                coin: 'Bitcoin',
+                path: "m/84'/0'/0'/0/0",
+                message: 'a'.repeat(1024 + 1),
+            },
+
+            legacyResults: [
+                {
+                    rules: ['1'],
+                    success: false,
+                },
+            ],
+        },
     ],
 } satisfies TestCase;
--- a/packages/connect/src/api/ethereum/api/ethereumSignMessage.ts
+++ b/packages/connect/src/api/ethereum/api/ethereumSignMessage.ts
@@ -6,6 +6,7 @@ import { Assert } from '@trezor/schema-utils';
 import type { PROTO } from '../../../constants';
 import { AbstractMethod } from '../../../core/AbstractMethod';
 import { getEthereumNetwork } from '../../../data/coinInfo';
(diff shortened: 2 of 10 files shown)


martykan:
If we are just checking hex strings the `getByteSizeOfString` utility is not really necessary right?

marekrjpolak:
> If we are just checking hex strings the getByteSizeOfString utility is not really necessary right?

Agree with @martykan. The `getByteSizeOfString` util would be helpful for strings we don't have control over. Here we're :100: sure that they're in hex.

Also, in the future we may have a branded type for hex strings and allow only them as `validateModelOneMessageSize` param?

Also a nit: I'd slightly prefer to name it `assertModelOne...` as it's basically what it does, and we may even make it a real assertion (well, probably not here as it doesn't seem necessary).

mroz22:
yes, you are right, I realized only later that at the point where it makes sense to run validation I am already converted to hex everywhere

PR #17915, opened 2025-03-26
Optimization of loading and updating fees

### LINKED ISSUE #15087: Optimise fee fetching and updating

### LINKED ISSUE #17772: When I refresh send form I get default hardcoded fees
**Describe the bug**

Image

When I leave form and enter again, it is ok

Image

izmy:
## Description

This update optimizes how fees are loaded and updated in the application.

1. Optimize fee updates to reduce unnecessary requests

2. Fix: Update default fees in the send form after fetching

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15087
Resolve https://github.com/trezor/trezor-suite/issues/17772

### DIFF
--- a/packages/connect/src/api/common/MiscFees.ts
+++ b/packages/connect/src/api/common/MiscFees.ts
@@ -20,9 +20,9 @@ export class MiscFeeLevels {
         this.levels = coinInfo.defaultFees;
     }
 
-    async load(blockchain: Blockchain) {
+    async load(blockchain: Blockchain, request: Parameters<typeof blockchain.estimateFee>[0]) {
         try {
-            const [response] = await blockchain.estimateFee({ blocks: [1] });
+            const [response] = await blockchain.estimateFee(request);
 
             // validate `feePerUnit` from the backend
             // should be lower than `coinInfo.maxFee` and higher than `coinInfo.minFee`
--- a/packages/suite/src/components/wallet/Fees/Fees.tsx
+++ b/packages/suite/src/components/wallet/Fees/Fees.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
 import {
     Control,
     FieldErrors,
@@ -12,6 +12,7 @@ import { useTheme } from 'styled-components';
 
 import { TranslationKey } from '@suite-common/intl-types';
 import { NetworkSymbol, NetworkType } from '@suite-common/wallet-config';
+import { updateFeeInfoThunk } from '@suite-common/wallet-core';
 import {
     FeeInfo,
     FormState,
@@ -25,6 +26,7 @@ import { spacings } from '@trezor/theme';
 import { HELP_CENTER_TRANSACTION_FEES_URL } from '@trezor/urls';
 
 import { Translation } from 'src/components/suite';
+import { useDispatch } from 'src/hooks/suite';
 import { Account } from 'src/types/wallet';
 
 import { CustomFee } from './CustomFee/CustomFee';
@@ -141,6 +143,7 @@ export const Fees = <TFieldValues extends FormState>({
     // Type assertion allowing to make the component reusable, see https://stackoverflow.com/a/73624072.
     const { getValues, register, setValue } = props as unknown as UseFormReturn<FormState>;
     const theme = useTheme();
+    const dispatch = useDispatch();
 
     const selectedOption = getValues('selectedFee') || 'normal';
     const isCustomFee = selectedOption === 'custom';
@@ -153,6 +156,10 @@ export const Fees = <TFieldValues extends FormState>({
     const feeOptions = buildFeeOptions(feeInfo.levels, networkType, symbol, composedLevels);
 
     const supportsCustomFee = networkType !== 'solana';
+
+    useEffect(() => {
+        dispatch(updateFeeInfoThunk({ networkSymbol: symbol }));
+    }, [dispatch, symbol]);
 
     const feeLabelId = useMemo(() => {
         switch (networkType) {
--- a/packages/suite/src/hooks/wallet/__fixtures__/useSendForm.ts
+++ b/packages/suite/src/hooks/wallet/__fixtures__/useSendForm.ts
@@ -844,6 +844,7 @@ export const setMax = [
             },
         },
         connect: [
+            undefined, // updateFeeInfoThunk
             {
                 success: true,
                 payload: [
@@ -874,6 +875,7 @@ export const setMax = [
     {
         description: 'setMax sequence: compose not final (without address) then add address',
         connect: [
+            undefined, // updateFeeInfoThunk
             {
                 success: true,
                 payload: [
@@ -945,6 +947,7 @@ export const setMax = [
         description:
             'setMax sequence: compose final with address, disable setMax, add second output',
         connect: [
+            undefined, // updateFeeInfoThunk
             {
                 success: true,
                 payload: [
@@ -1169,7 +1172,7 @@ export const setMax = [
             selectedAccount: ETH_ACCOUNT,
         },
         finalResult: {
-            estimateFeeCalls: 1,
+            estimateFeeCalls: 2,
             composedLevels: {
                 normal: {
                     type: 'final',
@@ -1201,7 +1204,7 @@ export const setMax = [
             selectedAccount: ETH_ACCOUNT,
         },
         finalResult: {
-            estimateFeeCalls: 1,
+            estimateFeeCalls: 2,
             composedLevels: {
                 normal: {
                     type: 'final',
@@ -1269,6 +1272,7 @@ export const setMax = [
             selectedAccount: SOL_ACCOUNT,
         },
         connect: [
+            undefined, // updateFeeInfoThunk
             undefined,
             {
                 success: true,
@@ -1282,7 +1286,7 @@ export const setMax = [
         ],
         finalResult: {
             solanaComposeTransactionCalls: 1,
-            estimateFeeCalls: 1,
+            estimateFeeCalls: 2,
             composedLevels: {
                 normal: {
                     type: 'final',
@@ -1367,6 +1371,7 @@ export const signAndPush = [
             selectedAccount: ETH_ACCOUNT,
         },
         connect: [
+            undefined, // updateFeeInfoThunk
             undefined, // estimateFee
             {
                 success: true,
@@ -1424,6 +1429,7 @@ export const signAndPush = [
             selectedAccount: XRP_ACCOUNT,
         },
         connect: [
+            undefined, // updateFeeInfoThunk
             undefined, // getAccountInfo address check
             {
                 success: true,
@@ -1491,6 +1497,7 @@ export const signAndPush = [
             },
         },
         connect: [
+            undefined, // updateFeeInfoThunk
             getComposeResponse({
                 payload: [
                     {
@@ -1610,6 +1617,7 @@ export const signAndPush = [
             },
         },
         connect: [
+            undefined, // updateFeeInfoThunk
             getComposeResponse(),
             {
                 success: false,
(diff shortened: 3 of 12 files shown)


suite-common/wallet-core/src/fees/feesThunks.ts
```diff
@@ -78,6 +78,10 @@ export const updateFeeInfoThunk = createThunk(
         { networkSymbol, forceUpdate }: { networkSymbol: string; forceUpdate?: boolean },
         { dispatch, getState, extra },
     ) => {
+        const selectedAccount = extra.selectors.selectSelectedAccount(getState());
+
+        if (networkSymbol.toLowerCase() !== selectedAccount.network?.symbol) return;
+
```
tomasklim: is this needed?

suite-common/wallet-core/src/fees/feesThunks.ts
```diff
@@ -78,6 +78,10 @@ export const updateFeeInfoThunk = createThunk(
...
         const network = getNetworkOptional(networkSymbol.toLowerCase());
         if (!network) return;
         const blockchainInfo = selectNetworkBlockchainInfo(getState(), network.symbol);
```
tomasklim, line 80: the condition logic in this method 
```
        if (
            feeInfo &&
            feeInfo.blockHeight > 0 &&
            blockchainInfo.blockHeight - feeInfo.blockHeight < 10 &&
            !forceUpdate
        ) {
            return;
        }
```
has to be dropped

packages/suite/src/hooks/wallet/useSendForm.ts
```diff
@@ -258,6 +258,19 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
...
     }, [getValues, composedLevels, dispatch, resetContext, props.selectedAccount.account]);
 
+    // replace default feeInfo with data loaded from the server
+    useEffect(() => {
+        const newFeeInfo = getFeeInfo({
+            networkType: state.account.networkType,
+            feeInfo: props.fees[state.account.symbol],
+        });
+
+        if (JSON.stringify(newFeeInfo) !== JSON.stringify(state.feeInfo)) {
```
tomasklim: compare `blockHeight`

tomasklim (changes requested):

tomasklim:
When you finish it, we have to make sure it does not break mobile / mobile will not spam fee fetching

suite-common/wallet-core/src/fees/feesThunks.ts
```diff
@@ -78,6 +78,10 @@ export const updateFeeInfoThunk = createThunk(
         { networkSymbol, forceUpdate }: { networkSymbol: string; forceUpdate?: boolean },
```
tomasklim: I think we will be able to get rid of `forceUpdate` but not 100% sure

suite-common/wallet-core/src/fees/feesThunks.ts
izmy: yes, you're right... I deleted it

packages/suite/src/hooks/wallet/useSendForm.ts
izmy: sure... I didn't think of that and it's definitely better than the ugly `JSON.stringify` 😄

suite-common/wallet-core/src/fees/feesThunks.ts
izmy: yes, `forceUpdate` is now useless... I deleted it

packages/suite/src/middlewares/suite/suiteMiddleware.ts
```diff
@@ -121,7 +121,7 @@ const suite =
                 const { enabledNetworks } = api.getState().wallet.settings;
```
tomasklim: I think you can remove the whole `case SUITE.SET_DEBUG_MODE` case

packages/suite/src/hooks/wallet/useSendForm.ts
```diff
@@ -258,6 +258,19 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
...
+        const newFeeInfo = getFeeInfo({
+            networkType: state.account.networkType,
+            feeInfo: props.fees[state.account.symbol],
+        });
+
+        if (newFeeInfo.blockHeight !== state.feeInfo.blockHeight) {
+            setState((prev: UseSendFormState) => ({ ...prev, feeInfo: newFeeInfo }));
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [props.fees]);
```
TomasBoda: nitpick (the `feeInfo` vs `newFeeInfo` is up to you but I think the explicit type of `prev` is not necessary)
```suggestion
    useEffect(() => {
        const feeInfo = getFeeInfo({
            networkType: state.account.networkType,
            feeInfo: props.fees[state.account.symbol],
        });

        if (feeInfo.blockHeight !== state.feeInfo.blockHeight) {
            setState(prev => ({ ...prev, feeInfo }));
        }
        // eslint-disable-next-line react-hooks/exhaustive-deps
    }, [props.fees]);
```

packages/suite/src/components/wallet/Fees/Fees.tsx
```diff
@@ -7,9 +7,11 @@ import {
     UseFormReturn,
     UseFormSetValue,
 } from 'react-hook-form';
+import { useDispatch } from 'react-redux';
```
TomasBoda: I think we have custom `useDispatch` in `src/hooks/suite` that is used across the app.

packages/suite/src/hooks/wallet/useSendForm.ts
izmy: changed

packages/suite/src/components/wallet/Fees/Fees.tsx
izmy: good catch, thanks

packages/suite/src/hooks/wallet/__fixtures__/useSendForm.ts
```diff
@@ -1169,7 +1172,7 @@ export const setMax = [
             selectedAccount: ETH_ACCOUNT,
         },
         finalResult: {
-            estimateFeeCalls: 1,
+            estimateFeeCalls: 2,
```
tomasklim, line 1175: why is it 2 now? maybe it is the same problem as that it fetches fee twice when the discovery starts

Screenshot 2025-03-29 at 10 14 19

I am thinking if we even need to fetch the fee before user enters send form, it would solve this 😀

packages/connect/src/api/common/MiscFees.ts
```diff
@@ -20,9 +20,9 @@ export class MiscFeeLevels {
         this.levels = coinInfo.defaultFees;
     }
 
-    async load(blockchain: Blockchain) {
+    async load(blockchain: Blockchain, request: Parameters<typeof blockchain.estimateFee>[0]) {
```
tomasklim, line 23: For Solana, in some cases it does not work because of wrong blockhash 🤯 it is probably not caused by this PR maybe it just shows this problem

Screenshot 2025-03-29 at 10 20 56

Screenshot 2025-03-29 at 10 25 56

Probable replication: fill send form. Wait for 1 minute, enter send form again

So I will create an issue

tomasklim, line 23: For Cardano it does not work either, but there is problem also on production 😀 
Screenshot 2025-03-29 at 10 57 09
Screenshot 2025-03-29 at 10 57 01

So I will create an issue

packages/suite/src/hooks/wallet/useSendForm.ts
```diff
@@ -258,6 +258,19 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
...
+    // replace default feeInfo with data loaded from the server
+    useEffect(() => {
+        const feeInfo = getFeeInfo({
+            networkType: state.account.networkType,
+            feeInfo: props.fees[state.account.symbol],
+        });
+
+        if (feeInfo.blockHeight - state.feeInfo.blockHeight > 0) {
+            setState(prev => ({ ...prev, feeInfo }));
+        }
```
tomasklim, line 272: Maybe: this is not great for bitcoin as fee can change within the same block, we should add note to the issue with bitcoin cache fee that this can be also problem

packages/connect/src/api/common/MiscFees.ts
tomasklim: #17976

tomasklim: #17975

packages/connect/src/api/bitcoin/BitcoinFees.ts
```diff
@@ -106,7 +106,7 @@ export class BitcoinFeeLevels {
         blocks.push(...fillGap(oneDayBlocks, oneDayBlocks / 2, oneDayBlocks * 6));
 
         try {
-            const response = await blockchain.estimateFee({ blocks });
+            const response = await blockchain.estimateFee({ ...request, blocks });
```
tomasklim: can we omit changes for Bitcoin, for now? I created an issue, let's leave it for someone who will understand it better https://github.com/trezor/trezor-suite/issues/17977

packages/suite/src/hooks/wallet/__fixtures__/useSendForm.ts
izmy: It is correct behavior and is not related to the discovery start. The first call is just useEffect after rendering the component `<Fees />`, the second call is after filling sendForm with correct values (it calls `composeEthereumTransactionFeeLevelsThunk`).

tomasklim: refresh app on dashboard

izmy: it was happening because in the action `TREZOR_CONNECT_BLOCKCHAIN_ACTIONS.BLOCK` was dispatch for `updateFeeInfoThunk`, so I removed it

packages/suite/src/hooks/wallet/useSendForm.ts
izmy: sure, I added a comment

packages/connect/src/api/bitcoin/BitcoinFees.ts
izmy: reverted

packages/suite/src/hooks/wallet/__fixtures__/useSendForm.ts
tomasklim: Ah I see, `TREZOR_CONNECT_BLOCKCHAIN_ACTIONS.CONNECT` is fired twice

tomasklim (approved):

mroz22:
please name your commits better next time! 2fc286bbdf38c411ef134de5817a88733c00803e has only changes to packages/connect yet it is named "fix(suite)". We auto pre-generate connect changelogs and a commit message like this doesn't helps us.

mroz22:
and it sounds like a feature to me when scoped to connect

PR #17903, opened 2025-03-25
Clean `connectPromise` when all backends down

marekrjpolak:
## Description

WIP

PR #17916, opened 2025-03-26
fix(discovery): recover discovery when connection is re-established

### LINKED ISSUE #17524: Interrupted discovery never finishes
**Describe the bug**
If you go offline during the discovery, it never finishes even if you go back online and even if you replug Trezor.

**Info:**

-   Mobile Suite Version: 25.2.1
-   OS: Android

**How to reproduce**

1. Connect Trezor to the app
1. Before discovery is finalized, go offline (flight mode)
1. Turn the internet back again
1. See error

**Expected behavior**
The discovery should continue if you go back online.

**Additional context**
The result for the user is similar to [this issue](https://github.com/trezor/trezor-suite/issues/13854), but it's a completely different problem

BrantalikP:
## Description

- Resuming Discovery when the internet connection is re-established

note: The fix is build upon https://github.com/trezor/trezor-suite/pull/17903 by @marekrjpolak 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17524

### Before
https://github.com/user-attachments/assets/e84355d3-fc28-4f1b-bfd7-32ba78d38135

### After
https://github.com/user-attachments/assets/046a15b2-b58b-437e-9aeb-a700239b1f49

### DIFF
--- a/suite-native/discovery/src/discoveryThunks.ts
+++ b/suite-native/discovery/src/discoveryThunks.ts
@@ -289,10 +289,10 @@ const discoverAccountsByDescriptorThunk = createThunk(
         },
         { dispatch, getState },
     ) => {
-        let isFinalRound = false;
+        let isFinished = false;
 
         if (A.isEmpty(descriptorsBundle)) {
-            isFinalRound = true;
+            isFinished = true;
         }
 
         const device = selectSelectedDevice(getState());
@@ -315,7 +315,7 @@ const discoverAccountsByDescriptorThunk = createThunk(
 
             if (success && !isAccountAlreadyDiscovered) {
                 if (accountInfo.empty) {
-                    isFinalRound = true;
+                    isFinished = true;
                 }
 
                 const isVisible =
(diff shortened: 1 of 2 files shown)


marekrjpolak (approved):
Approve for Connect part. I haven't tested the native discovery part, but the change looks reasonable.

suite-native/discovery/src/discoveryThunks.ts
```diff
@@ -525,9 +527,11 @@ const discoverNetworkBatchThunk = createThunk(
                     deviceState,
                     network,
                     accountType,
-                    round: round + 1,
+                    // Retry current round if discoverAccountsByDescriptorThunk fails
+                    round: error ? round : round + 1,
                 }),
             );
+            console.warn(error);
```
PeKne, line 534: Does this have any other purpose than debugging? If not, please remove it

PeKne (approved):
Looks good :ok_hand:

suite-native/discovery/src/discoveryThunks.ts
BrantalikP: Its intentional👌

matejkriz (approved):
Seems to works nice 👍

PR #17922, opened 2025-03-26
fix(suite): make the sourcemaps correct in vite dev server

### LINKED ISSUE #17919: Vite sourcemaps doesn't work properly
Since the inception of thte Vite dev server the sourcemaps aren't accurate. The breakpoints and debugger is stopped at the wrong line. It seems that the real position of the breakpoint is like 40 lines under the correct position.

Webpack's dev server works properly.

Likey reasons why it doesn't work:
- there is manualy injection of the code that sets up the globals in vite, reason why is that was that adding polyfill plugin to vite caused failure of the Android build
```ts
const injectBufferPolyfillPlugin = (): Plugin => {
    const bufferPolyfillCode = `
// Ensure Buffer is available globally
import { Buffer as ImportedBuffer } from 'buffer';


// Define Buffer in all possible global scopes
if (typeof window !== 'undefined') {
    window.Buffer = window.Buffer || ImportedBuffer;
};

if (typeof global !== 'undefined') {
    global.Buffer = global.Buffer || ImportedBuffer;
};

if (typeof globalThis !== 'undefined') {
    globalThis.Buffer = globalThis.Buffer || ImportedBuffer;
};

// Make sure global is defined
if (typeof window !== 'undefined' && typeof global === 'undefined') {
    window.global = window;
};

// Make sure globalThis is defined
if (typeof window !== 'undefined' && typeof globalThis === 'undefined') {
    window.globalThis = window;
};
`;
```

- other may be `workerPlugin` that changes of how the service `Worker` is loaded
```ts
const workerPlugin = (): Plugin => ({
    name: 'worker-loader',
    transform(_code, id) {
        if (/\/workers\/[^/]+\/index\.ts$/.test(id)) {
            // Return a virtual module that creates a web worker
            return {
                code: `
                    const worker = () => new Worker(new URL('${id}', import.meta.url), { type: 'module' });
                    export default worker;
                `,
                map: null,
            };
        }
    },
});

```

vojtatranta:
## Description

The source maps reflected code withtout injected source that was needed for proper working of the node polyfils

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17919

### DIFF
--- a/packages/suite-build/vite.config.ts
+++ b/packages/suite-build/vite.config.ts
@@ -51,14 +51,18 @@ const workerPlugin = (): Plugin => ({
     transform(_code, id) {
         if (/\/workers\/[^/]+\/index\.ts$/.test(id)) {
             // Return a virtual module that creates a web worker
+
             return {
                 code: `
                     const worker = () => new Worker(new URL('${id}', import.meta.url), { type: 'module' });
                     export default worker;
                 `,
-                map: null,
+                // Use an empty source map to preserve the original file's mapping
+                map: { mappings: '' },
             };
         }
+
+        return null;
     },
 });
 
@@ -107,12 +111,14 @@ const createWorkspaceAliases = () => {
 
 const commitId = execSync('git rev-parse HEAD').toString().trim();
 
-// Plugin to inject Buffer polyfill code at the beginning of every module
-const injectBufferPolyfillPlugin = (): Plugin => {
-    const bufferPolyfillCode = `
+// Plugin to provide Buffer polyfill via a virtual module
+const bufferPolyfillPlugin = (): Plugin => {
+    const virtualModuleId = 'virtual:buffer-polyfill';
+    const resolvedVirtualModuleId = '\0' + virtualModuleId;
+
+    const polyfillCode = `
 // Ensure Buffer is available globally
 import { Buffer as ImportedBuffer } from 'buffer';
-
 
 // Define Buffer in all possible global scopes
 if (typeof window !== 'undefined') {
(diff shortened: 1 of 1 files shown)


vojtatranta:
I tested that the break points in the Dev tools are placed correctly and code is stopped where it supposed to be.

Lemonexe (approved):
Looks OK, I did not try breakpoint debugging, but checked that traces of console errors have _less_ offset line numbers.

~Seems to me they are still inaccurate, but now it's just off by 2 lines instead of 25 lines. (lol I hadn't even known about this issue :smile: )~
EDIT: at second look it seems they are correct now :+1: 
EDIT: tried also debugger and works correctly – browser indicates the debugger at the appropriate line.

packages/suite-build/vite.config.ts
```diff
@@ -51,14 +51,18 @@ const workerPlugin = (): Plugin => ({
...
             // Return a virtual module that creates a web worker
+
             return {
                 code: `
                     const worker = () => new Worker(new URL('${id}', import.meta.url), { type: 'module' });
                     export default worker;
                 `,
-                map: null,
+                // Use an empty source map to preserve the original file's mapping
+                map: { mappings: '' },
```
vojtatranta, line 61: not sure if it works, but I'll rather keep it here :D

Lemonexe: Answer from AI:

> Use `map: null` if your transformation does not alter the source code or you don’t want any source map processing.
> Use `map: { mappings: '' }` if you need to return a valid but empty source map, potentially to satisfy Rollup’s expectations for source maps...  ...means that no transformations are mapped, but a source map technically exists.

So it seems that what you return in `map` means only the _additional_ mapping for the _transformation_.
Both of them effectively mean: keep the original source map even for the transformed code..

PR #17925, opened 2025-03-26
feat: refactorig of the Wellcome screen to handle Bluetoth devices

### LINKED ISSUE #17892: Welcome screen rework to handle bluetoth well
https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=7242-38262&t=Sr5E29pgEg0hBapE-1

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite/issues/17892

Cable:
image
image

BT: 
image
image

Error:
image
image

No BT enable:
image
image

Tip with action button
image

### DIFF
--- a/packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
+++ b/packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
@@ -1,160 +1,144 @@
-import { ReactNode, useMemo } from 'react';
+import { ReactNode, useState } from 'react';
 
-import styled from 'styled-components';
+import {
+    Card,
+    Collapsible,
+    Column,
+    ElevationContext,
+    ElevationDown,
+    ElevationUp,
+    IconName,
+    Row,
+    SelectBar,
+    Text,
+} from '@trezor/components';
+import { spacings } from '@trezor/theme';
 
-import { Button, CollapsibleBox, useElevation, variables } from '@trezor/components';
-import TrezorConnect from '@trezor/connect';
-import type TrezorConnectWeb from '@trezor/connect-web';
-import { isAndroid } from '@trezor/env-utils';
-import { Elevation, mapElevationToBorder } from '@trezor/theme';
-import { TREZOR_SUPPORT_DEVICE_URL } from '@trezor/urls';
+import { TroubleshootingTipsFooter } from './TroubleshootingTipsFooter';
+import { TroubleshootingTipsListCard } from './TroubleshootingTipsListCard';
+import { TroubleshootingTipsToggle } from './TroubleshootingTipsToggle';
 
-import { Translation, TrezorLink } from 'src/components/suite';
-
-const ItemLabel = styled.span`
-    color: ${({ theme }) => theme.legacy.TYPE_DARK_GREY};
-    font-size: ${variables.FONT_SIZE.SMALL};
-    font-weight: ${variables.FONT_WEIGHT.MEDIUM};
-`;
-
-const ItemDescription = styled.span`
-    color: ${({ theme }) => theme.legacy.TYPE_LIGHT_GREY};
-    font-size: ${variables.FONT_SIZE.TINY};
-    font-weight: ${variables.FONT_WEIGHT.MEDIUM};
-    margin-top: 2px;
-`;
-
-const Bullet = styled.span`
-    margin-right: 8px;
-    font-size: ${variables.FONT_SIZE.NORMAL};
-    color: ${({ theme }) => theme.legacy.TYPE_DARK_GREY};
-`;
-
-const Items = styled.div`
-    display: flex;
-    flex-direction: column;
-    margin-bottom: 20px;
-`;
-
-const Item = styled.div`
-    display: flex;
-
-    & + & {
-        margin-top: 16px;
-    }
-`;
-
-const ItemContent = styled.div`
-    display: flex;
-    flex-direction: column;
-    width: 100%;
-`;
-
-const ItemAction = styled.div`
-    display: flex;
-    flex: 1;
-    justify-content: flex-end;
-    padding-left: 24px;
-`;
-
-const ContactSupport = styled.div<{ $elevation: Elevation }>`
-    display: flex;
-    justify-content: space-between;
-    margin: 24px -16px 0;
-    padding: 20px 20px 0;
-    border-top: 1px solid ${mapElevationToBorder};
-    align-items: center;
-`;
-
-const FooterText = styled.span`
-    color: ${({ theme }) => theme.legacy.TYPE_LIGHT_GREY};
-    font-size: ${variables.FONT_SIZE.TINY};
-    font-weight: ${variables.FONT_WEIGHT.MEDIUM};
-`;
-
-// eslint-disable-next-line local-rules/no-override-ds-component
-const StyledButton = styled(Button)`
-    margin: 0 20px 20px;
-`;
-
-interface Item {
+export type TroubleshootingTipsItem = {
     key: string;
     heading?: ReactNode;
     description?: ReactNode;
     hide?: boolean;
     noBullet?: boolean;
-    action?: ReactNode;
-}
+    icon?: IconName;
+};
 
-interface TroubleshootingTipsProps {
+type SectionDefinition = { label: ReactNode; items: TroubleshootingTipsItem[] };
+
+type TroubleshootingTipsWithSectionsProps<K extends string, T extends K> = {
     label: ReactNode;
     cta?: ReactNode;
-    items: Item[];
-    offerWebUsb?: boolean;
     opened?: boolean;
     'data-testid'?: string;
-}
+    items: Record<K, SectionDefinition>;
+    sectionLabel?: ReactNode;
+    defaultSection: T;
+    toggleText?: ReactNode;
+};
 
-export const TroubleshootingTips = ({
+export const TroubleshootingTipsWithSections = <K extends string, T extends K>({
     label,
     items,
     cta,
-    offerWebUsb,
     opened,
+    sectionLabel,
+    defaultSection,
+    toggleText,
     'data-testid': dataTest,
-}: TroubleshootingTipsProps) => {
-    const { elevation } = useElevation();
+}: TroubleshootingTipsWithSectionsProps<K, T>) => {
+    const [isOpened, setIsOpened] = useState(opened === true);
+    const [selectedSection, setSelectedSection] = useState<K>(defaultSection);
 
-    const memoizedItems = useMemo(
-        () =>
-            items
-                .filter(item => !item.hide)
-                .map(item => (
-                    <Item key={item.key}>
-                        {!item.noBullet && <Bullet>&bull;</Bullet>}
+    const hasMultipleSections = Object.keys(items).length > 1;
 
-                        <ItemContent>
-                            <ItemLabel>{item.heading}</ItemLabel>
-                            <ItemDescription>{item.description}</ItemDescription>
-                        </ItemContent>
+    const headingRow = (
+        <Row
+            justifyContent="space-between"
+            alignItems="center"
+            margin={{ horizontal: spacings.sm }}
+        >
+            <Text typographyStyle="body">{hasMultipleSections ? sectionLabel : label}</Text>
 
-                        {item.action && <ItemAction>{item.action}</ItemAction>}
-                    </Item>
-                )),
-        [items],
+            {hasMultipleSections ? (
+                <Row>
+                    <SelectBar<K>
+                        onChange={setSelectedSection}
+                        options={Object.entries(items).map(([k, v]) => ({
+                            label: (v as SectionDefinition).label,
+                            value: k as K,
+                        }))}
+                        selectedOption={selectedSection}
+                        size="small"
+                    />
+                </Row>
+            ) : undefined}
+        </Row>
     );
 
     return (
-        <CollapsibleBox
-            paddingType="large"
-            heading={cta}
-            toggleLabel={label}
-            defaultIsOpen={opened}
-            data-testid={dataTest || '@onboarding/expand-troubleshooting-tips'}
-        >
-            {items.length > 0 && <Items>{memoizedItems}</Items>}
-
-            {offerWebUsb && !isAndroid() && (
-                <StyledButton
-                    variant="tertiary"
-                    data-testid="@onboarding/try-bridge-button"
-                    onClick={() => (TrezorConnect as typeof TrezorConnectWeb).disableWebUSB()}
-                >
-                    <Translation id="TR_DISABLE_WEBUSB_TRY_BRIDGE" />
-                </StyledButton>
+        <Column gap={spacings.xxxxl} alignItems="center">
+            {cta && (
+                <Card width="auto">
+                    <Row gap={spacings.md}>
+                        {label}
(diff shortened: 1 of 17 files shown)


packages/suite/src/components/suite/troubleshooting/tips/index.tsx
```diff
@@ -9,29 +7,23 @@ import { Translation } from 'src/components/suite/Translation';
 
 import { BridgeStatus, BridgeToggle, SuiteDesktopTip, Wrapper } from './BridgeTip';
 import { UdevDescription } from './UdevDescription';
+import { TroubleshootingTipsItem } from '../TroubleshootingTips';
 
-export type TipItem = {
```
peter-sanderson, line 13: this type was not same as the API of the component and some props were missing so I unified it and put it into component code

packages/suite/src/support/messages.ts
```diff
@@ -9686,9 +9684,49 @@ export default defineMessages({
...
+        id: 'TR_OR',
+        defaultMessage: 'OR',
+    },
+    TR_PAIR_NEW_BLUETOOTH_DEVICE: {
+        id: 'TR_PAIR_NEW_BLUETOOTH_DEVICE',
+        defaultMessage: 'Pair over Bluetooth',
+    },
+    TR_BLUETOOTH_TIP1_HEADER: {
+        id: 'TR_BLUETOOTH_TIP1_HEADER',
+        defaultMessage: 'Your Trezor Safe 7 is within 7km from computer',
```
peter-sanderson: - [ ] `// Todo:` Get some sane copy before we merge this. We dont want to get this translated I guess.

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
```diff
@@ -1,160 +1,147 @@
...
+                    <Collapsible.Toggle onClick={() => setIsOpened(!isOpened)}>
+                        <Row justifyContent="center" flex="1" margin={{ bottom: spacings.xs }}>
+                            <TroubleshootingTipsToggle isOpen={isOpened === true}>
+                                {toggleText}
+                            </TroubleshootingTipsToggle>
+                        </Row>
+                    </Collapsible.Toggle>
+                    <Collapsible.Content>
+                        <ElevationContext baseElevation={-1}>
+                            <ElevationDown>
```
jvaclavik, line 107: why do you need both of these?

packages/suite/src/support/messages.ts
```diff
@@ -4707,12 +4707,6 @@ export default defineMessages({
         id: 'TR_HEX_FORMAT',
         defaultMessage: 'Hex format',
     },
-    TR_DISABLE_WEBUSB_TRY_BRIDGE: {
```
peter-sanderson, line 4710: This was dead code. @mroz22 acked its removal

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
```diff
@@ -1,160 +1,147 @@
...
+};
+
+type SectionDefinition = { label: ReactNode; items: TroubleshootingTipsItem[] };
 
-interface TroubleshootingTipsProps {
+type TroubleshootingTipsWithSectionsProps<K extends string, T extends K> = {
     label: ReactNode;
     cta?: ReactNode;
-    items: Item[];
-    offerWebUsb?: boolean;
```
peter-sanderson, line 93: This was dead code. @mroz22 acked its removal

packages/suite/src/components/suite/troubleshooting/TroubleshootingTipsItemComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <Icon name={item.icon ?? 'dotOutlineFilled'} size={iconSizes.large} />
+
+        <Column gap={spacings.xs}>
+            <Text typographyStyle="body">{item.heading}</Text>
+            <Text typographyStyle="hint" variant="tertiary">
+                {item.description}
+            </Text>
+        </Column>
+
+        {item.action && <ItemAction>{item.action}</ItemAction>}
```
peter-sanderson: - [x] `// Todo:` test this action sitation

packages/suite/src/components/suite/troubleshooting/TroubleshootingTipsListCard.tsx
```diff
@@ -0,0 +1,27 @@
...
+type TroubleshootingTipsListCardProps = {
+    items: TroubleshootingTipsItem[];
+};
+
+export const TroubleshootingTipsListCard = ({ items }: TroubleshootingTipsListCardProps) => {
+    const memoizedItems = useMemo(
+        () =>
+            items
+                .filter(item => !item.hide)
+                .map(item => <TroubleshootingTipsItemComponent item={item} key={item.key} />),
```
jvaclavik: I suggest using `List` component from our DS (https://dev.suite.sldev.cz/components/develop/?path=/story/list--list)

packages/suite/src/support/messages.ts
```diff
@@ -9686,9 +9684,49 @@ export default defineMessages({
...
+        id: 'TR_BLUETOOTH_TIP3_HEADER',
+        defaultMessage: 'Your Trezor Safe 7 is in pairing mode',
+    },
+    TR_BLUETOOTH_TIP3_TEXT: {
+        id: 'TR_BLUETOOTH_TIP3_TEXT',
+        defaultMessage: 'Idk, need more info, this is just a template',
+    },
+    TR_CONNECTION_TYPE: {
+        id: 'TR_CONNECTION_TYPE',
+        defaultMessage: 'Conenction type',
```
jvaclavik: typo

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
```diff
@@ -1,160 +1,145 @@
...
+    const [selectedSection, setSelectedSection] = useState<K>(defaultSection ?? null);
+
+    const headingRow = (
+        <Row
+            justifyContent="space-between"
+            alignItems="center"
+            margin={{ horizontal: spacings.sm }}
+        >
+            <Text typographyStyle="body">
+                {Object.keys(items).length > 1 ? sectionLabel : label}
```
jvaclavik: What about extracting `Object.keys(items).length > 1` to some const? You use it twice and it could improve readability

packages/suite/src/components/suite/troubleshooting/TroubleshootingTipsFooter.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <Row justifyContent="space-between" alignItems="center" margin={{ horizontal: spacings.sm }}>
+        <Text typographyStyle="body">
+            <Translation id="TR_ONBOARDING_TROUBLESHOOTING_FAILED" />
+        </Text>
+
+        <TrezorLink variant="nostyle" href={TREZOR_SUPPORT_DEVICE_URL}>
+            <Button variant="tertiary" size="small">
+                <Translation id="TR_CONTACT_SUPPORT" />
+            </Button>
+        </TrezorLink>
```
jvaclavik: Please use `Button` with `href` prop + `useExternalLink` hook. It's not a good pattern to wrap `button` with `a`

packages/suite/src/components/suite/troubleshooting/TroubleshootingTipsItemComponent.tsx
```diff
@@ -0,0 +1,32 @@
+import styled from 'styled-components';
+
+import { Column, Icon, Row, Text, iconSizes } from '@trezor/components';
+import { spacings } from '@trezor/theme';
+
+import { TroubleshootingTipsItem } from './TroubleshootingTips';
+
+const ItemAction = styled.div`
```
jvaclavik: Would it be possible to use `Row` here?

jvaclavik (commented):
I think we should to polish the design, but it doesn't have to part of this PR. I already pinged Michael and Ondra.

packages/suite/src/components/suite/troubleshooting/TroubleshootingTipsListCard.tsx
jvaclavik: Maybe `BulletList` would be even better: https://dev.suite.sldev.cz/components/develop/?path=/story/bulletlist--bullet-list

peter-sanderson: [fixed](https://github.com/trezor/trezor-suite/pull/17925/commits/d12eb4c81dba04c8bd77a4cdedfd567c303e28cc)

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
peter-sanderson: If you want to have Card with `-1` elevation background, you cannot place it into `<ElevationContext baseElevation={-1}>` directly. Because from `useElevation` you already get `+1` and the card would have backgroundElevation0.

So you need to place it into `-2`, but we dont have it. So you would have to do `<ElevationContext baseElevation={3}>` (as we are rotating the elevations when we reach the start or the end). 

But I thinks doing so (using `3`) is even more confusing then doing `-1` and then `ElevationDown`. Internally it will be `3` as `prev(-1) = 3` anyway but at least it seems to me more readable and understandable.

packages/suite/src/support/messages.ts
peter-sanderson: fixed

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
peter-sanderson: fixed

packages/suite/src/components/suite/troubleshooting/TroubleshootingTipsFooter.tsx
peter-sanderson: fixed

packages/suite/src/components/suite/troubleshooting/TroubleshootingTipsItemComponent.tsx
peter-sanderson: Fixed: 
image

packages/suite/src/components/suite/troubleshooting/TroubleshootingTipsItemComponent.tsx
peter-sanderson: fixed: 
image

jvaclavik: I think settings button should be tertiary, not primary

jvaclavik (approved):

peter-sanderson:
@jvaclavik wanted tertiary button for action => 
image

rabbitholiness:
@peter-sanderson updated the wording here: https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=7300-16820&t=Ahr3YxFt6aahxHXy-1

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
```diff
@@ -1,160 +1,144 @@
...
     noBullet?: boolean;
-    action?: ReactNode;
-}
+    icon?: IconName;
+};
+
+type SectionDefinition = { label: ReactNode; items: TroubleshootingTipsItem[] };
 
-interface TroubleshootingTipsProps {
+type TroubleshootingTipsWithSectionsProps<K extends string, T extends K> = {
```
Lemonexe, line 32: Little late to the party, but the TS is beautiful :ok_hand:

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
```diff
@@ -1,160 +1,144 @@
...
+};
 
-export const TroubleshootingTips = ({
+export const TroubleshootingTipsWithSections = <K extends string, T extends K>({
     label,
     items,
     cta,
-    offerWebUsb,
     opened,
+    sectionLabel,
```
Lemonexe, line 48: Took me quite a while to figure out what is the distinction between `label`, `sectionLabel` and `items[K].label`, plus there is `toggleText`...
It's not quite clear from the code. If I understand correctly:

- `items[K].label` is used only for the Select Options, which visually acts as the label for the current section
  - that makes sense!
- `label` is used for CTA label
- `label` is also used for heading row, but **only** if there is one section
   - if multiple sections, `sectionLabel` is used for heading row instead (??)

I think it'd be better to refactor it to make it more explicit and predictable:
- rename `label` → `ctaLabel` and use it only for CTA label
- always use `sectionLabel`  for heading row
if they're same you'll have to pass two props with the same content, but I'd strongly prefer this explicitness

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
```diff
@@ -1,160 +1,144 @@
...
-            </ContactSupport>
-        </CollapsibleBox>
+            <Collapsible
+                isOpen={isOpened}
+                data-testid={dataTest || '@onboarding/expand-troubleshooting-tips'}
+            >
+                <Column gap={spacings.md}>
+                    <Collapsible.Toggle onClick={() => setIsOpened(!isOpened)}>
+                        <Row justifyContent="center" flex="1" margin={{ bottom: spacings.xs }}>
+                            <TroubleshootingTipsToggle isOpen={isOpened === true}>
```
Lemonexe, line 100: isOpened is only boolean, so it can be simplified to `isOpen={isOpened}`

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
```diff
@@ -1,160 +1,144 @@
...
+    'data-testid'?: string;
+    items: TroubleshootingTipsItem[];
+    toggleText?: ReactNode;
+};
+
+export const TroubleshootingTips = ({ items, ...props }: TroubleshootingTipsProps) => (
+    <TroubleshootingTipsWithSections
+        {...props}
+        items={{ '': { items, label: '' } }}
+        defaultSection=""
```
Lemonexe, line 142: Using empty strings doesn't look very nice to me..
I could have realized it sooner, but `defaultSection` can actually be omitted.
You could make it optional and use the first key of items.

:thought_balloon: Or you could even remove the prop and make always the first item selected, but who knows, maybe it will be needed to dynamically change what's default cable vs. bluetooth based on some parameters, so let's keep it..

Lemonexe (commented):
Few ex post comments that could be addressed in subsequent PR :pray:

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.4.1 (d6bedb2912b4d9a870d6a99dd65327c90701defd)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.4.1 Chrome/134.0.6998.44 Electron/35.0.1 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: 

PR #17951, opened 2025-03-28
feat(suite-native): Mobile Trade: improve list item animations

jbazant:
## Description

Changes animations in mobile trade section lists. Adds entering and leaving animations to country picker, where animations were missing.

## Related Issue

Resolve


vytick (approved):
👌

PR #17825, opened 2025-03-21
chore(suite-native): add node-gyp to main dev dependencies to satisfy sharp build

matejkriz:
## Description

iOS builds started failing after dependencies update; this seems to satisfy the build process of sharp and shouldn’t cause any issues, as we already have this version of node-gyp specified in resolutions anyway.

Example of failing build on develop: https://expo.dev/accounts/trezorcompany/projects/trezor-suite-develop/builds/7c5cca20-8b58-42bc-8b24-8964a9d98b05

Adhoc build with this change applied: https://expo.dev/accounts/trezorcompany-develop/projects/trezor-suite-preview/builds/5b348c4d-e0b9-4713-85e4-dce9b39ee483

PR #17954, opened 2025-03-28
Dependencies cleanup

komret:
- Removing dependencies that are unused according to `npx depcheck`.
- Moving dependencies from global scope to packages where they are used.
- Update `semver` to unify versions.

Before:
Screenshot 2025-03-28 at 10 39 47

After:
Screenshot 2025-03-28 at 10 38 45


package.json
```diff
@@ -100,7 +100,6 @@
         "@types/node": "22.13.10",
         "@types/react": "18.2.55",
         "bn.js": "5.2.1",
-        "node-gyp": "10.2.0",
```
Lemonexe: we had this explicitly for a reason: `node-gyp` is implicitly required by `yarn`. At the same time, it used to resolved as a subdependency by other deps to version 9. But v9 was not sufficient and `bcrypto` build always failed → explicitly installed v10.
But I see in `yarn.lock`, that `node-gyp` is now resolved as subdep to v11 without our explicit installation, which LGTM.

I tested it:
`rm -r node_modules/bcrypto/ && yarn` :heavy_check_mark:

package.json
```diff
@@ -116,36 +115,24 @@
         "@babel/preset-react": "^7.24.7",
         "@babel/preset-typescript": "^7.24.7",
         "@babel/runtime": "^7.23.9",
-        "@suite-common/wallet-types": "workspace:*",
```
Lemonexe, line 119: I have no idea how this would even work in top-scope package.json, and to what effect? :laughing: 
Good riddance :+1:

Lemonexe:
All of those packages are owned by `npm`, and that hasn't change. So I'd trust that a new collaborator on npm is still part of the same organization..
    @SocketSecurity ignore npm/unique-slug@5.0.0
    @SocketSecurity ignore npm/@npmcli/fs@4.0.0
    @SocketSecurity ignore npm/unique-filename@4.0.0

komret:
Reverted the change on `node-gyp` due to https://github.com/trezor/trezor-suite/pull/17825.

Lemonexe (approved):
LGTM. Now that node-gyp is explicitly kept at `10.2.0`, there are now no new packages added to `yarn.lock`, changes are now minimal :+1: 

We can see in future if we can bump it to 11.

PR #17955, opened 2025-03-28
fix: PinInput support for disabled + add missing export + autofocus

peter-sanderson:
- Add `disabled` and `autoFocus` prop to it
- fix missing export

### DIFF
--- a/packages/components/src/components/PinInput/PinInput.stories.tsx
+++ b/packages/components/src/components/PinInput/PinInput.stories.tsx
@@ -10,14 +10,14 @@ const meta: Meta = {
 } as Meta;
 export default meta;
 
-const Component = ({ length }: { length: number }) => {
+const Component = ({ ...props }: Omit<PinInputProps, 'onChange' | 'onComplete'>) => {
     const [code, setCode] = useState('');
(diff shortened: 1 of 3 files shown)


packages/components/src/components/PinInput/PinInput.stories.tsx
```diff
@@ -10,14 +10,14 @@ const meta: Meta = {
...
-const Component = ({ length }: { length: number }) => {
+const Component = ({ ...props }: Omit<PinInputProps, 'onChange' | 'onComplete'>) => {
     const [code, setCode] = useState('');
     const [isComplete, setIsComplete] = useState(false);
 
     return (
         <div>
             <PinInputComponent
-                length={length}
+                {...props}
```
jvaclavik, line 20: I would prefer not using spread, but OK

jvaclavik (approved):

packages/components/src/components/PinInput/PinInput.stories.tsx
peter-sanderson: It is in case where components have same API (same props type with omit) so it seems to me ok, to use spread here as I am literally passing the same stuff.

PR #17958, opened 2025-03-28
fix(message-system): remove unsued gitlab key

karliatto:
## Description
We do not use GitLab anymore so it seams that `JWS_PRIVATE_KEY_FILE` is not necessary anymore. @matejkriz  ???


tomasklim (approved):

vdovhanych (approved):
I would keep it, if you have to deploy it locally for some reason we have the key stored as file for backup so we would save time putting it to env but whatever in the end :D

PR #16937, opened 2025-02-11
feat(suite): update passphrase flow

jvaclavik:
## The problem

There are two entry points for showing modal:

- `PassphraseModal` (called by Trezor hardware device)
- `SwitchDevice` (shown by us in Suite)

Unfortunately part of the flow is handled in `PassphraseModal` and part of it in `SwitchDevice`.

## Diagram

image

## What happened in this PR
- update flow, before there was no switch between creating new or existing one
- polish Passphrase modal - divide to multiple external components
- disable PassphraseWalletConfirmationSteps (1/2/3) for passing through the flows and introducing `PassphraseModalContext` for keeping the current state. It allows custom order without specific context
- there are similar states for existing and non-existing flow, but they have different handling of back/next/cancel events

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15794

PR #17966, opened 2025-03-28
Passphrase flow revamp

### LINKED ISSUE #12468: Not showing "duplicite passphrase modal" while creating new passphrase wallet (view-only)
Screenshot 2024-05-17 at 16 08 26

vojtatranta:
- [x] **Passphrase on device** !!!
- [x] **Going back - restore the device state**
- [ ] cancelling of flow (button top right on some modals)
- [x] When device itselfs requests the passphrase (startup)
- [x] **  // "view-only" is active, device is reconnected and you fired an action that needs passphrase (e.g. add coin, show receive address) **
- [x] alt+P flow
- [x] enter press int he passprase enter / confirm throws error in React
- [x] add standard wllate button not visible
- [x] start over - problem, cannot submit
- [x] cardano confirm passphrase on address receive
- [x]  Added redirect to `/accounts = wallet-index` when flow is finished
- [x]  Added redirect to `/ = suite-index` when passprhase flow is initialized
- [x] QA talk about tests needing redirects
- [ ] check that when a device is locked and you want to add a passphrase there, the flow works properly
- [ ] if passphrase flow is on and you disconnect the device - what should happen? 
- [ ] device not ready error handling (when you start up an outdated version of fw)
- [ ] firmware hash check fails (on the older FW) `mware hash check failed` during accessing a passphrase wallet
- [x] Check the UI tests - why do they suddenly started to fail
- [ ] Two device - one disconnected, remembered wallet, other connected no wallet - add a new passphrase wallet says "device nto ready"
- [x] confirm passphrase back - throws error, stays on the device
- [ ] add E2E test case - a remember wallet + a new coonected device with no wallets
- [x] **TO TEST: did not receive auth state in _iniCOnfirmation **after cancelling confirmaiton passphrase, you cannot then add a passphrase wallet again.
- [ ] ‼ **cannot add multiple existing wallets** the state stucks `authConfirm: true` for some reason and so the device selector is still unusable
- [ ] **large problem with the device what is to be used for the passphrase flow** - solved by on-demand discovery. Without it, it happens taht we are stucked back to "auth failed" state. Which is complained about here https://github.com/trezor/trezor-suite/issues/18149
### A lot of changes (+4000 lines)

Reasons:
- remade the passphrase flow again outside of current middelware + thunks structure in to a `passphrsaeFlowManager` that handles it all transparently (back, forward, cancelling, deletion of the invalid state)
- added thorough tests for the passphrase flow module (actions, reducer, manager)
- updated the E2E tests to reflect a new flow
- inabiliy to simply hook to the discovery process and just await a discovery state of some particular device
- overall new architecture

The goal is to remove the passphrase flow from the previous thunks and also move the adding of the wallet to the "new architecture".

### Architecture changes

Create a new "category" of objects:
- services
   -  with dependencies: manager (`passphraseFlowManager`)
   -  withtout dependencies: `trezorConnectServices`, `discoveryHook`

Rule is that `manager` shouldn't have any state at all. Services can have states (typically some event listeners (if I was accurate I would creat a `timer` services to replace global `setTimeout`).

### Pros
- no need to `dispatch()` anything, calling directly `passphraseFlowManager.startFlow()`
- direct results can be returned/awaited , not via `await dispatch` =  `const result = await trezorConnectService.invoke(connect => connect.getDeviceState())`
- thunk action results can be reused just with `dispatch` = `dispatch(authorizeResultThunkFulfilled({}))`
- you can still call thunks if you want
- highly testable - all the dependencies are passed, testing is very easy - set up state, set up mocks , visible in the code

### Cons
- aren't available from thunks, yet, but it is possible to do
- dunno, any?

These are usable in react components via `const services = useServices()`. These already have `store` as their depeendency (if specified in the `iniStore()` argument)

### Added hacky listener to hook into the discovery

I added the `discoveryHook` service as a hack to "get into the" process of the discovery. I can't really tell what tirggers the discovey + it requires a specific selected device to be mantained. There is not yet a wait to just "run discovery" on any device and await it / listen to the progress, without looking into the store / middleware. Which is wrong / doesn't scale, etc.
So I created this hacky hook which you can await if you are sure that the discovery is somehow triggered.

The new passphrase flow chart:

Next flow
416615188-e396decd-efcc-452f-a668-3e4732139f7d

## The charts of the call flows:

[The updated one](https://mermaid.live/edit#pako:eNq9WFtv4jgU_itWXspIgNrSLp08jNShdAapF0RgR1pVqtzYFG9TOxs7ZRjU_77HjpM4JFDaWS0PAZJzP9-5OGsvFIR6vifpPynlIb1g-DHBz3ccwQeHSiRoJmmS_Y9xoljIYswVmo0Qlvo6EM-x4JQrWScaYynjRYIlvRYER5pj41ad5xpz_EiTKu1lJJb2QZ3jPFRMcFnnsA_qHBNK0rBJh31Q5_iBo4gqx7bKjTr9QHBOQ6Uppwn9JRJ7o055QV9YSEtCe6OBkMlQvNBkpWnLP-NEhFSCkxmHzlbny5fZyEeDiIVP6M4rXbRW33mIcRQsmQoXFXWzEbBan3wkFWjXUWkRQ9NGayaHP5lUjD_6aI4jSV8_ZZzZ1bKCEBt6HxEmYwx6WkZaNdpWbpeRJtFtMJIphqNAYUV9dMCF6lBN03mgcIkTwCewy4PXT9YKqxX020T6aHweBOPvk_NgeH95dfujG0zPJ1ONbCDMmCxpx0ZtFhNQp52HqxI71Lqem8htYNsH0ZwAXB7gDkGCQzpzAuNT1yjJBGzwWmOChVg6j7L0fQUzxrkVKMzrr2JOHQUAQAhuSiH5YIl2BRWuNGU_FHzOkueKMpuwLTkvWFWCuQQW42QBHieQYC1wlNE4-LQHfugGejLpO6W-AxbD6T1AYzr8IDRqut0QvSO3I3kjlCkE7WJzcnFUdA0iqOQHCi3wC3XQBZkVijqOVINb6V2-JtYdZNOUc0KYllBNuv5U-EGe7W1-tdV1cRxHq4AqXdOytc5xkEp6X5qq0ZIWbUR_LDvIzZz00ZDjh-hN_zLqjmtPkIbaM5RQCWGUtK6kHo3tPJtu6yQW8AQosDkLsUGaBVwXEzIVWKrWGqlVrHuYtOHo6OAwSg5Q0bygV7hZ3qktTChgMPN4xAGQMLov02jOgIeANmIjx-mySlSq2ylf0giik7FOFyl_2k9mrfkMdVk4idvVLAUP0odnploledthveUX9eajSwGnajHImpWJPmLSYKpMXL3-NrtcSdF6wVGqR11dcw5SCsOpUasZW3uplcbTBq15kezQXoXJlgRmcnIcpqZxVeyhZUEuGCGUtyGYS7ySpb5aU95S5ymb2FpprTOYz0bdyXAwHP05vC8brXZqFQlMck_qhR6Ac26ZQ48lznqSXaFBUqR3n1xAG-X804XJPXsEe-HHglp2ky2RsF9ZqubQWneV2hY_H6mtCHeujTGsy3KrRxO9V8OotXY4A9_p5Sm0glDPFMd1QnWDo-S_aT-2--gwdOaYQZM4aCOaJCIp0vymijyG1GkLE_q3sbK1NsJARelCh-R-gS5SxdT7FodGmY2S9l8bmmU6Qj-8NOyzODRq39U5GhaGAjbOQcq0phuomvzZzvF44YCymHcIjgULc4-FAQxC8GtEPjI4S0H0NxAGU60YazkgjMwPYqlc2fASMz2LOyQ_S_0upnbL_r-wtdOKvTBWHC6voFtDn93YQjcmUNmPbTIz3muYKhFd4oTCOQ62gJDGSqKmFHfn-eZSO1EVltR3n_yJXU9I2Zlr3aag3bJG5Y83pDgjXp9FCf3po8MMfSnAYXRxBTONU7UUyZPc2IFclU1OmNPwhg9vmltlaq273W4R7tKuyezmZnTzbYdB2-fbeRiKlKsRn4uy3B5SDplsoydKY9sRQNsTiy8Zx9GE6plu1vrhc6xWJa5e35yKOFMHwZ0LNBeJVeXaHQkRo0t4RHG4cBgqpFv627kr3nSjvCnoLQB2BSWSVePBo8wYrDJfZzeQ6vvx5PbbZBgEiL4UpbAzXdbYYhxnaLNGbSI2FzJSFNCGyzy00QuTDM4_2w4KTkd27K4mc-M0U-4eX00UTYVDX6YENlPTxaEko1Wzj9vRcwn-paCphI4DmY0XRtuAEVG9_ciMZ2d2c20NJ7U9kpN7vK3482oa3F6Pr4bT4cVb82bOOJML81przzGScWy8EnvnmLgc3YyC72_NCNjPqXK3S70BlyODQ7Kd_cFiwmx0e0QSqih-K4rB9HY8Hl7YndN-DeAU4oLaa3vPFI5TjHi-t9a37zzY4Z_pnefDT0LnOI3UnXfHX4EUurMIVjz0fH3Ya3uJSB8Xnp-9OvSyE499m13cjTH_S4jnnIUSBh3gOnsDbl6EGxLPX3s_Pf-s3z09PTw96fU_nx2dnfXb3srzO3_0-t2j0_5Zr9f74_TksH_y2vZ-GZnH3cPj49N-r3fcOznsfT47aXuPifbF2mfeBQ50RXr-0dFh__VfzaEJ6A) (notice: just middleware to trigger discovery)

[Previous previous flow chart](https://mermaid.live/edit#pako:eNqlWG1v2zYQ_iusPrQJ4AjJ4iSIsBZwY7cz1iSeZW_AlqFgpLNNRCY1kXLqpvnvO5J6tWTHbfMhoKTj8V6ee-7oJycQITieI-G_FHgAfUbnCV3ecYJ_NFAiIVMJiX2OaaJYwGLKFZkOCZX6_5VYxoIDV7IpNIYw_aLl7MJHddCUmixS_iBLMfvclLsSnEOgtOAkga8iyV40JfuwYgGUgtmLFkEmA7GCZK1ly4dRIgKQaILdoSNw9O7ddOiRq4gFD-TOGVEp40VCJZC_aBSBunMI48R_ZCpY1I6bDnGrdckjIZMxRYkDGoZ2n_ly8PRoH9YxeJlCvXZHPd8f_Tbu-YMOCY3S58NDq9f-t4orJwQJUAXWgCGXimJSszOsgg5JJQyWsVqXPnhkRiOJujeUmnxUrLYaeoFigku37aj8lA0zaVQkJRQg-RtFFnQFmIsijHEiFBjNdkvNkCzTXj3xLo3jaO2DUozPZc3Bz3HFOZWkhW_6L9uOeq1NHhlweh-9aI6VPqra46eBRgpJQGIVSGgeUslNuzDwsDWju4OfxiEGv0zhNdZxGYEFC0PgueeHW7BSqJYQoanWvRpY9tibQMgS3N2bKUgsdH2jDkKragtgdTX5C_FYDTrjcarIUoQ0alTegKP-ivCu6hK8DIyf3i-ZOnha0SjFyJQKbnl_Z0X9fPw7iPtHupaNk17CdcrGGUQOnpThhOnQHQ-uBsM_B5-rnBDTdSRomIO7CWwf0VWNsBIZj1RdvkGoE818uYIOyfdPFkwiith8DgkuFpBtJzRVC5Gwr1SHg8wi8VjVmFVKxcGxbjBS5duRLRRsGq3TbBz9YzrwJ5_fTyeT2xsCK2wulWxrTGTQuXN-ozyMcqOQg5vYyT3BY2YsWRK0thmBlsruVwwtKpYgvy_MOxb4WMro-zCsu1HBYz2pc8gqzK-pbE9EriIPM1SKkwTYcrHMQBJpGWWWRtG6Bq0Sv20aXNwwY1iqoe7xBcdtGlFWgO2L14jqCB5poisVqzGAWMkXDqhqNvra6jXrJPk5GQX9ep-8C8uIeZmtbgZ616TG3UhGp9ylV_sW94YFjcONWu1pBiOPvLJf3WY3NbKMh4AHHJsHbWOKDhf6ffPCHfY_DYwAB_UoEh2TfOWvl_cikrt8aNK4wtFmI4p7-l_fat13XbdI_U43xtObm-HNxxZbdd-_oklIuSg8I1iyhGseStjK0Ids6fgN57B8eivKIt2oM5X9UkOta5mMFZbsQ7iBVfgR1Ci9xxHvd1gfHJIZTo5AcZ4rTSWajndNEjnP5V7HRh95gLXcOUiMSrn9hokiGC8EcjPTtnG9lGq6PdS1NG8bXtrS9z7VXF0t7tqR31-0ra5sFu29OdVHfvKIXbsR8Lla_CCkX4KSxmkQiJSrIZ-Jqk0Vc8zyASDOeMtOataiBxZ_YJxGY9AsV_nSNrdvJSFredXw7Vil1l4k9ZkwoLdGVjdHQsTkQ14P5Yaa6BZo96rqTQNVCeXSsjmOShKvheu2mspBhDPB--kNcuXn0fj243jg-9WhYDfuFmZAwAvdHKtKVsBnNHTaEMMULHOs_BNnO_81n5agKGKO2vtCWINHGayc-W4EgSTBoGGcmjW9E9pZhGX9rpVFsh3oRTUNjf3aC1vKJRq9wgzzZcUkQzc8cqCF3UxQXzzJ27dvyRsukiWN3pDXr4021_Q08-n4kHz7Rl7l2lzQ4GtEox6RH2KlPPyb_bdR2oX3bvmWHJETIzhDMmvJFuCdlwy-L0O9MNRDtNWYF0KFdwtC3GDGFgavM0XrQKhRZGmzGPrClqFvT176gBhK8ZgGKdWYyPwU8HOEs41uItB3EGlP2skZuaktWanXWAUTxFxV0n1LLLHs1ysVHNTAkeevfub2CTLLzwvtaFvDubq9Hn0aTAb9envVCM0AYKhkD0OQTuMfNMKf3I5Gg76RMceVqyu83252fqfjLAEpgoWO5zzp13cOXhGXeBHzcBnCjKaRunPu-DOKYqaEv-aB45mO5iQinS8cz8LNyQnA_PyYi8SU_y1E8Qghw1ZxbX-xND9cGhHHe3K-ON7R6dmF2z09Pz0-uzjtnhxfHF90nDW-Pz-_dLtn5ycn3e7ZCX44ee44X43aX9zLC3x9ctk97p5e4veOM0-0N5mF6COg51ihKNt9_h8RYEDa) 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/12468#issuecomment-2775298696

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=vt-passphrase-flow-update&lookback=7)


vojtatranta:
based on: https://github.com/trezor/trezor-suite/pull/16937

packages/suite/src/actions/wallet/passphraseFlowManager.ts
```diff
@@ -0,0 +1,775 @@
...
+        if (!selectedDevice) {
+            console.warn('No selected device in passphraseFlowManager#confirmPassphrase');
+
+            return;
+        }
+
+        const alwaysOnDevice = Boolean(passphraseOnDevice);
+        const hidden = Boolean(passphraseOnDevice || !!confirmedPassphrase);
+
+        if (
```
vojtatranta: this is pretty bad, that's due to completely weird `PassphraseTypeCard` that `const enterPressed = useKeyPress('Enter');`

This bizzare thing logically cauess re-submitting, coz there's a `useEffect` on that that suese a callback, that chagnes with the device instance.

Witohout this check, the infinite loop is called coz it is being called over and over. That press hook relies on unmout, which is bizarely wrong.

packages/suite-desktop-core/e2e/tests/passphrase/passphrase-cancel.test.ts
```diff
@@ -8,9 +9,14 @@ test.describe('Passphrase cancel', { tag: ['@group=passphrase'] }, () => {
     });
 
     test('possible to cancel passphrase', async ({ page, devicePrompt, dashboardPage }) => {
+        await createWaitPromise();
```
vojtatranta: god forbid, but I was desperate.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -615,6 +619,10 @@ export const startDiscoveryThunk = createThunk(
                 // try to generate device metadata master key
                 await dispatch(initMetadata(false));
             }
+
+            // NOTE: here we know what's the status of the wallet auth confirmed or not.
+            extra.services.discoveryHook.triggerDiscoveryAuthHooks(deviceState, device);
```
vojtatranta, line 624: This is ugly as hell , but it is just a hotfix. We should be able to run the discovery on-demand regardless of which device is selected.

packages/suite/src/support/trezorConnectService.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This service allows you to call TrezorConnect methods and provides a way to mock
+ * those calls in tests.
+ */
+export const createTrezorConnectService = (trezorImplementation: typeof TrezorConnect) => {
+    /**
+     * Call a TrezorConnect method
+     * @param fn Function to call with TrezorConnect as the argument
+     * @returns The result of the function call
+     */
+    const invoke = <R>(fn: (trezorConnect: typeof TrezorConnect) => R): R =>
```
vojtatranta, line 31: Why? To not call the `TrezorConnect` statically! (we should avoid to prevent us from mamking that bizzare build steps replacement of implmentations btw)

suite-common/redux-utils/src/createThunk.ts
```diff
@@ -16,3 +17,25 @@ export const createThunk = <TParams = void, TPayload = void, TThunkAPI = void>(
...
     );
+
+/**
+ * Utility to extract the action type and the parameters from the standard createThunk result.
+ * The result of this utility can be used as a pure argument of dispatch() call.
+ * @param {(...args: any[]) => any} action - the action returned by createThunk
+ * @returns {(...args: T[]) => ThunkAction<T, any, unknown, any> & { type: string }} - an action that can be used as a pure dispatch() call
+ */
+
+export function createAsyncThunkResultActionCreator<T extends Record<string, any>>(
```
vojtatranta: enables reusing of the `createThunk` success or result action type

suite-common/test-utils/src/configureMockStore.ts
```diff
@@ -66,6 +66,18 @@ export function configureMockStore<S = any, A extends AnyAction = AnyAction>({
         return next(action);
     });
 
+    const dispatchListeners: ((action: AnyAction) => void)[] = [];
+
+    const onDispatch = (listener: (action: AnyAction) => void) => {
```
vojtatranta, line 71: This is better that the action log, coz I want to do something in sync when this is triggered. Maybe even finish the test or a make a change in state of the store.

packages/suite-desktop-core/e2e/support/common.ts
```diff
@@ -125,3 +125,6 @@ export const calculatePercentageOfBalance = (params: PercentageOfBalanceParams)
 
     return localizeNumber(fraction, 'en', 0, maxDecimals);
 };
+
+export const createWaitPromise = (timeout: number = 1000) =>
```
vojtatranta: I'd be glad for an advice, coz without it , last time i Checked, the test are randomly failing even though they see the element they should click.

vojtatranta:
@coderabbitai full review

vojtatranta:
@coderabbitai summary

packages/suite/src/reducers/services.ts
```diff
@@ -0,0 +1,19 @@
+import { createContext, useContext } from 'react';
+
+import type { PassphraseFlowManager } from 'src/actions/wallet/passphraseFlowManager';
+
+export type ServiceFactory<T> = (deps: any) => T;
```
peter-sanderson: This seems to be unused

packages/suite/src/reducers/store.ts
```diff
@@ -104,32 +106,53 @@ type RootReducerShape = typeof rootReducer;
...
         middleware: getDefaultMiddleware =>
             getDefaultMiddleware({
-                thunk: { extraArgument: extraDependencies },
+                thunk: { extraArgument: { ...extraDependencies } },
             }).concat(getCustomMiddleware()),
         devTools,
     } as const);
+
+    // NOTE: creates object { getState, dispatch, ...services[number][key] }
+    const services = (options.serviceFactories ?? []).reduce<{
```
peter-sanderson: Instead of need for this logic I suggest do an explicit DIC composition. This would have couple of advantages:

1. Only required dependencies will be passed to the services
2. Strict types without compormise
3. Explicitly inferred `Servicies` types for the result of the DIC that can be used for `useServices` hook


WIP: https://github.com/trezor/trezor-suite/commit/14945cd8c3b9c2333aa925160df3b6f3586ec7c2#diff-f6cbd9448cc861419263766f38aab404421e1d4aea98af18f7e09f78f9722cbdR141

packages/suite-desktop-ui/src/MainDesktop.tsx
```diff
@@ -89,7 +91,15 @@ export const init = async (container: HTMLElement) => {
 
     const preloadAction = await preloadStore();
     const { statePatch } = await desktopApi.handshake();
-    const store = initStore(preloadAction, statePatch);
+    const { store, services } = initStore({
+        preloadStoreAction: preloadAction,
+        statePatch,
+        serviceFactories: [
```
peter-sanderson: This definition is duplicated here and in the `MainWeb`. I think is can (and) shell be just in the one place. (Something like [this](https://github.com/trezor/trezor-suite/commit/14945cd8c3b9c2333aa925160df3b6f3586ec7c2#diff-122f9a65147ef188794bd6885401bfe598b27cb58ff85d61648c84811dd05dafR8-R15))

peter-sanderson (commented):
First part of the review (from many, I suppose)

packages/suite-desktop-ui/src/MainDesktop.tsx
vojtatranta: yeah, but obivously, it won't be always the same for different platform (that's the whole point of modularization).

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,787 @@
...
+        case 'not-exist-awaiting-discovery':
+            return 'not-exist-enter-passphrase';
+        case 'exists-empty-wallet':
+            return 'exists-enter-passphrase';
+        case 'exists-best-practices':
+            return 'exists-empty-wallet';
+        default:
+            return null;
+    }
+};
```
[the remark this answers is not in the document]
vojtatranta: you are correft!

packages/suite/src/support/discoveryHook.ts
```diff
@@ -0,0 +1,183 @@
...
+
+                if (notTheOne) {
+                    return true;
+                }
+
+                descriptor.callbacks.reject(new TimeoutError());
+
+                return false;
+            });
+        }, hookTimeout);
```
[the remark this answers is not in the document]
vojtatranta: you are right about this one pal

packages/suite/src/reducers/services.ts
vojtatranta: fxd by your propopsal

packages/suite/src/reducers/store.ts
vojtatranta: yeah picked that

packages/suite-desktop-ui/src/MainDesktop.tsx
peter-sanderson: In general yes: I was thinking about this, and if we go this way, I assume we would have to do something like having a multiple DI containers (configuration for it) at some point. So not just services, that depends on one another, but also DI configs from different modules depends on one another.

In this case, I am not sure: the way how we separate web from desktop is a bit weird and I am not sure if its a good example of modularization.

vojtatranta: Well, in correct sceanrion,m you have a different implemenation for TrezorConnect here defined for web and desktop a classic example.

For this case, it doesn't matter, but whole point of this is to have other configs for different platforms (or for tests fro that matter)

packages/suite-web/e2e/downloads/Bitcoin_10_20240829T072722.csv
```diff
@@ -0,0 +1,19 @@
+Timestamp;Date;Time;Type;Transaction ID;Fee;Fee unit;Address;Label;Amount;Amount unit;Fiat (USD);Other
```
mroz22: this is some leftover?

vojtatranta: I'll ask @jvaclavik what's that I am really not sure where this came from.  THanks

jvaclavik: I have no idea

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,876 @@
...
+
+            const freshDevice = getSelectedDevice(device, selectDevices(getState()));
+
+            if (!freshDevice) {
+                throw new Error('device-not-selected');
+            }
+
+            // NOTE: triggers the discovery
+            dispatch(
+                authorizeDeviceThunkFulfilled({
```
vojtatranta: @martykan says:
Remove this in favor of direct dispatch of the discovery. 

Then, remove the helper that creates this type.

packages/suite-desktop-core/e2e/tests/passphrase/passphrase-cancel.test.ts
```diff
@@ -9,6 +9,8 @@ test.describe('Passphrase cancel', { tag: ['@group=passphrase'] }, () => {
     test('possible to cancel passphrase', async ({ page, devicePrompt, dashboardPage }) => {
         await dashboardPage.deviceSwitchingOpenButton.click();
         await dashboardPage.addHiddenWalletButton.click();
+        await page.getByTestId('@switch-device/add-existing-hidden-wallet-button').click();
+        await dashboardPage.passphraseInput.fill('abc');
```
martykan, line 13: This line is duplicated, intentional?

packages/suite-desktop-core/e2e/tests/passphrase/passphrase-cancel.test.ts
```diff
@@ -9,6 +9,8 @@ test.describe('Passphrase cancel', { tag: ['@group=passphrase'] }, () => {
     test('possible to cancel passphrase', async ({ page, devicePrompt, dashboardPage }) => {
         await dashboardPage.deviceSwitchingOpenButton.click();
         await dashboardPage.addHiddenWalletButton.click();
+        await page.getByTestId('@switch-device/add-existing-hidden-wallet-button').click();
```
martykan, line 12: Ideally this should be also accessed thorugh `dashboardPage`, it's the new convention

packages/suite/src/reducers/services.ts
```diff
@@ -0,0 +1,34 @@
+import { createContext, useContext } from 'react';
+
+import { ExtraDependencies } from '@suite-common/redux-utils/libDev/src';
+
+import { createPassphraseFlowManager } from '../actions/wallet/createPassphraseFlowManager';
+import { Store } from '../types/suite';
+
+export const dependencyInjectionContainer = (
+    store2: Store,
```
martykan: why store2 ?

packages/suite/src/reducers/services.ts
martykan, line 1: Should this be under reducers? I'd probably make a services folder and move it there, with `createPassphraseFlowManager`, `trezorConnectService` etc. as well. 

Also going forward we need to figure out how services can be used in Suite Common or Suite Native packages.

packages/suite/src/reducers/store.ts
```diff
@@ -104,32 +105,38 @@ type RootReducerShape = typeof rootReducer;
...
 
-export const initStore = (
-    preloadStoreAction?: PreloadStoreAction,
-    statePatch?: Record<string, any>,
-) => {
-    // get initial state by calling STORAGE.LOAD action with optional payload
-    // payload will be processed in each reducer explicitly
-    const preloadedState = preloadStoreAction
-        ? rootReducer(undefined, preloadStoreAction)
+export const initStore = <T extends Record<string, any>>(options: {
```
martykan, line 109: I don't quite get this, why are the services handled here, when it seems to be completely independent from what the rest of this function is doing? Why not make a new function? 

And also shouldn't services be passed down to extraDependencies and not the other way around? What if we need to create a service that needs to be used through extraDependencies but also depends on other services?

packages/suite/src/types/suite/index.ts
```diff
@@ -110,7 +110,7 @@ export interface Dispatch extends ThunkDispatch<AppState, any, Action> {
 
 export type GetState = () => AppState;
 
-export type Store = ReduxStore<AppState, Action>;
+export type Store = ReduxStore<AppState, Action | UnknownAction>;
```
martykan, line 113: Doesn't this weaken our Redux types a lot?

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseWalletIsEmpty.tsx
```diff
@@ -1,32 +1,39 @@
...
 import { spacings } from '@trezor/theme';
 import { HELP_CENTER_PASSPHRASE_URL } from '@trezor/urls';
 
-import { onCancel as onCancelModal } from 'src/actions/suite/modalActions';
-import { goto } from 'src/actions/suite/routerActions';
-import { Translation } from 'src/components/suite/Translation';
-import { useNetworkSupport } from 'src/hooks/settings/useNetworkSupport';
-import { useDispatch, useSelector } from 'src/hooks/suite';
-import { selectEnabledNetworks } from 'src/reducers/wallet/settingsReducer';
+import { onCancel as onCancelModal } from '../../../../../actions/suite/modalActions';
```
martykan, line 7: Pls use absolute imports

packages/suite/src/components/suite/layouts/SuiteLayout/PassphraseFlow.tsx
martykan, line 1: Can't we just keep this in ModalSwitcher, in one place?

packages/suite/src/reducers/services.ts
vojtatranta: @peter-sanderson why? :)

packages/suite/src/types/suite/index.ts
vojtatranta: Didn't notice. This fixes actually the typing tho. But will check. Good point.

packages/suite/src/reducers/store.ts
vojtatranta: Yeah, this idiotic extra dependencies parameter. Years ago, i tried to tell the maintainers that it sucks, coz you don't have access to store from there.  They said that I need to make my.own middleware for it instead of using official redux thunk. 

You are right that it should be done like so. Since we use the official redux thunk where extra deps don't have access to the store, I'd say that either write own middleware (not a fan) or declare thst eversthjng in exrra dependencies has to be without any other dependency.

packages/suite-desktop-core/e2e/tests/passphrase/passphrase-cancel.test.ts
vojtatranta: dunno. I think John made these changes.

vojtatranta: Will check

packages/suite/src/reducers/store.ts
martykan: We could structure it like this - assuming we have some more capable DI container implementation. 
1. Start with a minimal extraDependencies object without the DI container
2. Pass the minimal extraDependencies into the DI container to construct all services. Services don't ever need to access other services from extraDependencies, since they have DI.
3. Then inject the resulting services back into a final extraDependencies object for use in Redux. Or possibly just pass the DI container itself.

packages/suite/src/reducers/services.ts
peter-sanderson: No idea, probably typo/leftover from some other situation where it was shadowed and I moved that later and forgot to rename back

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+            return 'not-exist-enter-passphrase';
+        case 'exists-enter-passphrase':
+            return 'initial';
+        case 'not-exist-awaiting-discovery':
+            return 'not-exist-enter-passphrase';
+        case 'exists-empty-wallet':
+            return 'exists-enter-passphrase';
+        case 'exists-best-practices':
+            return 'exists-empty-wallet';
+        default:
```
peter-sanderson:
```suggestion
        case 'exists-awaiting-discovery':
        case 'exists-confirm-passphrase':
        case 'exists-passphrase-mismatch-warning':
        case 'not-exist-passphrase-mismatch-warning':
        case 'initial':
        case 'passphrase-duplicate':
            return null;
        default: {
            const _unhandledCase: never = state;
            throw new Error(`Unhandled state: ${_unhandledCase}`);
        }
```

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+        const flow = selectPassphraseFlow(getState());
+        if (!flow) {
+            console.warn('Now passphrase flow to restart, nothing will be done');
+
+            return;
+        }
+
+        const prevState = flow.state;
+
+        let mantainLoadingForState = false;
```
peter-sanderson, line 519: typo `mantainLoadingForState` => `maintainLoadingForState`

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+            trezorConnectService.invoke(trezorConnect => trezorConnect.cancel(cancelReason));
+        }
+
+        if (previousState) {
+            transactState(device, previousState, { cancelReason });
+
+            return;
+        }
+
+        _resetFlow();
```
peter-sanderson, line 583: Here, when `previousState` is `null` technically there is transition to the `'initial'` state. 

I would like to see unification of the code, where there is no special handling for the `null` state. => `determineBackState` woudl return `'initial'` in that situation. 

Is there a reason (that I missed)  why not do this?

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
peter-sanderson: fix in this commit: https://github.com/trezor/trezor-suite/commit/286146094e1c59880cc770ff60760832d8487802

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
peter-sanderson: Suggestion fix in this commit: https://github.com/trezor/trezor-suite/commit/286146094e1c59880cc770ff60760832d8487802

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+};
+
+const STATE_CANCEL_MAP: { [K in PassphraseWalletState]?: string } = {
+    'not-exist-enter-passphrase': 'enter-passphrase-back',
+    'not-exist-confirm-passphrase': 'confirm-passphrase-back',
+    'exists-enter-passphrase': 'enter-passphrase-back',
+    'exists-confirm-passphrase': 'confirm-passphrase-back',
+};
+
+const matchState = <A extends PassphraseWalletState, B extends PassphraseWalletState>(
```
peter-sanderson, line 72: Here https://github.com/trezor/trezor-suite/commit/b150c346ce8f3ad6e4c9cbf02539f792c3ea9486

I propose to use this more explicit naming. It is used only locally a this naming makes IMO much cleared what is happening

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+        switch (createStateMatcher(prevState, state)) {
+            case matchState('not-exist-enter-passphrase', 'not-exist-best-practices'):
+            case matchState('exists-empty-wallet', 'exists-enter-passphrase'):
+                trezorConnectService.invoke(trezorConnect =>
+                    trezorConnect.cancel(options.cancelReason ?? 'enter-passphrase-back'),
+                );
+                break;
+        }
+
+        switch (state) {
```
peter-sanderson, line 530: I am thinking: would it be beneficial to unify those 2 switches? 

I mean something like: 
```ts
switch (createStateTransitionMatcher(prevState, state)) {
    case matchStateTransition('not-exist-enter-passphrase', 'not-exist-best-practices'):
    case matchStateTransition('exists-empty-wallet', 'exists-enter-passphrase'):
        trezorConnectService.invoke(trezorConnect =>
            trezorConnect.cancel(options.cancelReason ?? 'enter-passphrase-back'),
        );
        break;
                
        case matchAnyTransitionTo('not-exist-enter-passphrase'):
        case matchAnyTransitionTo('exists-enter-passphrase'):
           ...
           ...
}
```

Currently the transition code is fragmented into two places

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+                    keepDevice: stateInStates(prevState, [
+                        'exists-passphrase-mismatch-warning',
+                        'not-exist-best-practices',
+                        'initial',
+                    ]),
+                });
+                break;
+
+            case 'exists-confirm-passphrase':
+                mantainLoadingForState = true;
```
peter-sanderson, line 546: Also I would like to see every state-transition handler separated into its own method. And the `maintainLoadingForState` which is currently mutable IMO shall be discared.

I would simply repeat the:
```ts
dispatch(
            setPassphraseFlowState({
                state,
                maintainLoadingForState: true |false,
            }),
        );
```

At the end of every of the state-transition methods, or I would let those methods to return the `state` and `maintainLoadingForState`, which would be then passed to `dispatch(setPassphraseFlowState(...))`

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+        if (!flow) return;
+
+        const { state } = flow;
+
+        const previousState = determineBackState(state);
+
+        const cancelReason = STATE_CANCEL_MAP[state];
+
+        if (cancelReason) {
+            trezorConnectService.invoke(trezorConnect => trezorConnect.cancel(cancelReason));
```
peter-sanderson, line 574: just thinking out lout: but can this be part of the handles for the state-transition?

So the state transition that goes back, would automatically trigger this. This may also remove the need for `STATE_CANCEL_MAP` as it would be places in every distinct transition method directly

packages/suite/src/reducers/wallet/passphraseFlowReducer.ts
```diff
@@ -0,0 +1,68 @@
...
+    PassphraseWalletState,
+    finishPassphraseFlow,
+    resetPassphraseFlow,
+    setPassphraseFlowState,
+    setPassphraseFlowTransactionLoading,
+    startPassphraseFlow,
+} from '../../actions/wallet/passphraseFlowActions';
+
+export type PassphraseFlowState = {
+    state: PassphraseWalletState;
```
peter-sanderson, line 12: maybe heretical suggestion, but I would name this `screen` or `step`. State is the whole `PassphraseFlowState` and the `PassphraseWalletState` represents the `step`/`screen` which is user presented with.

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+        toBestPractices,
+        finishFlow,
+        submitPassphrase,
+        confirmPassphrase,
+        confirmBestPractices,
+        transactState,
+    };
+};
+
+export type PassphraseFlowManager = ReturnType<typeof createPassphraseFlowManager>;
```
peter-sanderson, line 896: This type seems to be unused. As the Manager shall never be re-used as dependency, explicit type is probably not needed.

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+                            _resetFlow();
+                            break;
+                        case 'enter-passphrase-back':
+                            // NOTE: this should be somehow handled
+                            // like, we should open the device switch
+                            // or delete the device
+                            break;
+                        case 'passphrase-duplicate':
+                            transactState(baseDevice, 'passphrase-duplicate');
+                            break;
```
peter-sanderson, line 437:
```suggestion
                            case 'device-not-ready':
                            // Do nothing ???
                            break;

                        default: {
                            const _exhaustiveCheck: never = response.error; // Poor-man's `switch-exhaustiveness-check`
                            throw new Error(`Unhandled response.error: ${_exhaustiveCheck}`);
                        }
```

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
```diff
@@ -0,0 +1,887 @@
...
+                    const deviceToSelect = [baseDevice, ...devices].find(device =>
+                        Boolean(device.state),
+                    );
+
+                    dispatch(deviceActions.selectDevice(deviceToSelect ?? baseDevice));
+
+                    switch (response.error) {
+                        case 'auth-failed':
+                        case 'enter-passphrase-cancel':
+                            _resetFlow();
```
peter-sanderson, line 428: I would assume that this shall be replaced with `transactState(baseDevice, 'init');` in the sense of a change I described here: https://github.com/trezor/trezor-suite/pull/17966/files?w=1#r2073171679

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
peter-sanderson, line 1: This file is really big, I would prefer to have it separated (for example, every state/screen/step transition in separate file

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
vojtatranta: No, it is not technically transition to the initial state. Initial state is when the there is some passphrase flow already created.

Null is intentional situation when there is no passphrase flow at all.

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
vojtatranta: No, keep the related things together. Meaningful separation would be creation of the wallet and passphrase flow. Certinly not by screen or step or state which are tightly coupled.

packages/suite/src/reducers/store.ts
vojtatranta: I dont understand this line:
> Pass the minimal extraDependencies into the DI container to construct all services. Services don't ever need to access other services from extraDependencies, since they have DI.

What do you mean by `Services don't ever need to access other services from extraDependencies, since they have DI.` Are you saying that a Services A that depends on B and C does not access other dependency D that's inside `extraDependencies`? 

So like just pick what you need from `extraDependencies` and pass it as dependency to the service?

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
vojtatranta: This is better, I think
```ts
switch (type) {
            case DEVICE.CONNECT:
                break;
            case DEVICE.CONNECT_UNACQUIRED:
                dispatch(deviceActions.connectUnacquiredDevice({ device, settings }));
                break;
            default:
                ((_: never) => {})(type);
        }
```

packages/suite/src/reducers/wallet/passphraseFlowReducer.ts
vojtatranta: 👿 heretic! :D we'll see on wednesday.

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
peter-sanderson: Whatever what gets the job one :)

peter-sanderson: If this is better => would be great to send PR o replace it everywhere. Its less lines of code so I think its good :D

peter-sanderson: maybe we can have some named util for this case that 

`export const exhaustive = (_: never) => {};`

peter-sanderson: I really like it, I prepare PR for it directly.

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
peter-sanderson: Whatever suits you and gets the job of making this file smaller and thus easier to navigate better

packages/suite/src/actions/wallet/createPassphraseFlowManager.ts
peter-sanderson: Discussed on Slack => will be unified to one state transition function

packages/suite/src/components/suite/layouts/SuiteLayout/AppShortcuts.tsx
```diff
@@ -19,15 +20,14 @@ export const useAppShortcuts = () => {
     useEvent('keydown', e => {
         const { altKey, metaKey } = e;
         const isDeviceSelected = selectedDevice !== undefined;
-
         // press ALT + P to show PassphraseModal
         if (
             selectedDevice?.connected &&
-            altKey &&
+            (altKey || metaKey) &&
```
peter-sanderson, line 26: Is this a separate fix? Maybe lets create separate commit/PR for it. I know its small but can be delivered immediately. This PR will take a lot of time to merge.

packages/suite/src/components/suite/layouts/SuiteLayout/PassphraseFlow.tsx
```diff
@@ -0,0 +1,77 @@
...
+const ForegroundAppModal = ({ app, cancelable }: ForegroundAppModalProps) => {
+    const dispatch = useDispatch();
+
+    const onCancel = () => dispatch(closeModalApp());
+
+    if (app === 'switch-device')
+        return <SwitchDevice cancelable={cancelable} onCancel={onCancel} />;
+};
+
+export const PassphraseFlow = () => {
```
peter-sanderson, line 61: @jvaclavik this is something temporal? Or what is the plan with our little-big modal madness?

packages/suite/src/components/suite/layouts/SuiteLayout/AppShortcuts.tsx
```diff
@@ -4,12 +4,13 @@ import { selectSelectedDevice } from '@suite-common/wallet-core';
 import { KEYBOARD_CODE } from '@trezor/components';
 
 import { closeModalApp, goto } from 'src/actions/suite/routerActions';
-import { addWalletThunk } from 'src/actions/wallet/addWalletThunk';
 import { useDiscovery, useDispatch, useSelector } from 'src/hooks/suite';
+import { useServices } from 'src/reducers/services';
 
-export const useAppShortcuts = () => {
+export const AppShortcuts = () => {
```
peter-sanderson, line 10: Separate PR?

packages/suite/src/components/suite/layouts/SuiteLayout/PassphraseFlow.tsx
peter-sanderson: Maybe this good way to go, but I think we have to remove this dependency on global modal state, and use local state for rendering of the modals. => passphraseFlow has its own state and in that own state, its own representations for rendering of the modal.

packages/suite/src/components/suite/modals/ModalSwitcher/ForegroundAppModal.tsx
```diff
@@ -29,7 +28,7 @@ const getForegroundApp = (app: ForegroundAppRoute['app']) => {
         'bridge-requested': BridgeRequested,
         'bridge-deprecated': BridgeDeprecated,
         udev: UdevRules,
-        'switch-device': SwitchDevice,
+        'switch-device': null,
```
peter-sanderson, line 31: Add comment explaining that it is moved to `PassphraseFlow.tsx` 

... but we shall refactor this anyway, lets taslk about this @jvaclavik

packages/suite/src/components/suite/layouts/SuiteLayout/PassphraseFlow.tsx
jvaclavik: I hope it's temporal, it's definitely not good. Just an effort to have related stuff together. But it's definitely not nice.

vojtatranta:
## Discovery - based model of the passphrase flow.

### Discover metadata
- the device / wallet on which the passphrase is set 
  -  determination if it possible to delete this device / wallet on the process failure (probably doable with mantain the device in the scope and adding to the store at the end of the process if needed)
- the flag present in the discovery highligting that it is a passphrase flow
- the flag representing flow branch - new wallet / existing wallet
- figuring out, where to store the state of the flow - as per the chart, there [are multiple "explanatorary - just UI" states](https://private-user-images.githubusercontent.com/4154045/432766241-872ccdae-5367-49f0-b7b7-3426a99af17d.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDY0NTQ2NjksIm5iZiI6MTc0NjQ1NDM2OSwicGF0aCI6Ii80MTU0MDQ1LzQzMjc2NjI0MS04NzJjY2RhZS01MzY3LTQ5ZjAtYjdiNy0zNDI2YTk5YWYxN2QucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDUwNSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA1MDVUMTQxMjQ5WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDAxYmMxNzVkZWFlZmY0MmMxNGYzZWEzNDJhZWRlNmRlNWVlMjQ4MWI3YTNjM2E3OGUyNjU2YzhmMzQ2NjNjNyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.MAgzbpRXzomU4m6xMggH-JCqq8yUkgmoDu3J0NfRPGw) to be displayed to the user
- additional events / metadata flagging to the discovery to indicate "loading state" of the forms / transition states from the forms eg. device is entering to the "enter passphrase" mode 
- determine the "go back" states , that's likely hardly possible without imperative switching of the state. eg `passphrase duplicity -> back to -> enter passphrase`

packages/suite/src/components/suite/layouts/SuiteLayout/PassphraseFlow.tsx
```diff
@@ -0,0 +1,77 @@
...
+    cancelable: boolean;
+};
+
+const ForegroundAppModal = ({ app, cancelable }: ForegroundAppModalProps) => {
+    const dispatch = useDispatch();
+
+    const onCancel = () => dispatch(closeModalApp());
+
+    if (app === 'switch-device')
+        return <SwitchDevice cancelable={cancelable} onCancel={onCancel} />;
```
peter-sanderson, line 58: I believe this can (and shall) be handled by the state machine

peter-sanderson: An we can completely remove the `'switch-device'` Foreground App

packages/suite/src/hooks/suite/usePreferredModal.ts
```diff
@@ -47,18 +49,35 @@ export const usePreferredModal = () => {
...
     if (route && isForegroundApp(route) && hasPriority(route)) {
         return getForegroundAppAction(route, params);
     }
 
     if (modal.context !== MODAL.CONTEXT_NONE) {
+        // NOTE: in case when passphrase flow is active, we handle the device passphrase request
+        // within the passphrase flow
+        if ('windowType' in modal && modal.windowType === UI.REQUEST_PASSPHRASE) {
+            return {
+                type: !passphraseFlow ? 'device-request-passphrase' : 'passphrase-flow',
```
peter-sanderson, line 63: I think this shall not be handled here and all logic can be simplified in `PassphraseFlow`

packages/suite/src/components/suite/layouts/SuiteLayout/PassphraseFlow.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const PassphraseFlow = () => {
+    const { type, payload } = usePreferredModal();
+
+    switch (type) {
+        case 'foreground-app':
+            return <ForegroundAppModal {...payload} />;
+        case 'redux-modal':
+            return <ReduxModal {...payload} />;
+        case 'passphrase-flow':
```
peter-sanderson, line 69: Instead of relaying logic to `usePreferedModal` I would do all the `if`s here

Something like:

```ts
    const { type, payload } = usePreferredModal();
    const modal = useSelector(state => state.modal);
    const passphraseFlow = useSelector(selectPassphraseFlow);

    if (passphraseFlow !== null) {
        return <PassphraseModal />;
    }

    // NOTE: in case when passphrase flow is active, we handle the device passphrase request
    // within the passphrase flow
    if (
        modal.context !== MODAL.CONTEXT_NONE &&
        'windowType' in modal &&
        modal.windowType === UI.REQUEST_PASSPHRASE
    ) {
        return <ConfirmPassphraseBeforeAction />;
    }
```

packages/suite/src/hooks/suite/usePreferredModal.ts
peter-sanderson: See: https://github.com/trezor/trezor-suite/pull/17966/files#r2074942929

packages/suite/src/components/suite/layouts/SuiteLayout/PassphraseFlow.tsx
```diff
@@ -0,0 +1,77 @@
...
+        AppState['modal']['context'],
+        typeof MODAL.CONTEXT_NONE
+    >,
+> = Extract<AppState['modal'], { context: T }>;
+
+/** Modals initiated by redux state.modal */
+export const ReduxModal = (modal: ReduxModalProps) => {
+    switch (modal.context) {
+        case MODAL.CONTEXT_DEVICE: // Modals requested by Device from `trezor-connect`
+            return <DeviceContextModal {...modal} />;
```
peter-sanderson, line 41: I would just leave this where it was and ... (╯°□°)╯︵ ┻━┻

Moving it here would just bring more confusion. We talked with @jvaclavik and we dont concluded any master-plan from here, so I would not event start any change to this without the grand master-plan.

vojtatranta: :-D welcome to my life  https://www.youtube.com/watch?v=r0U0AlLVqpk

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/ConfirmPassphraseBeforeAction.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    const intl = useIntl();
+
+    const onEnterPassphraseDialogCancel = () =>
+        TrezorConnect.cancel(intl.formatMessage(messages.TR_CANCELLED));
+
+    const onSubmit = useCallback(
+        (value: string, passphraseOnDevice?: boolean) => {
+            dispatch(
+                onPassphraseSubmit({
```
peter-sanderson, line 35: nit/optional: `onPassphraseSubmit` => `onPassphraseSubmitThunk`

vojtatranta: No to si teda Bohdan dovolil dost! :D

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseModal.tsx
```diff
@@ -1,179 +1,112 @@
...
-        </SwitchDeviceModal>
+    const deviceOffer = !!(
+        device.features &&
+        device.features.capabilities &&
+        device.features.capabilities.includes('Capability_PassphraseEntry')
     );
 
-    // passphrase needs to be confirmed because wallet is empty
-    if (authConfirmation) {
+    switch (passphraseState?.state) {
```
peter-sanderson, line 68: This `switch` shall be `exhaustive`.

I think this shall be before the switch:
```ts
if (passphraseState === null) {
        return null;
}
```

And there shall be handling for this shall be handled in the switch:
```ts
        case 'exists-enter-passphrase':
        case 'exists-confirm-passphrase':
            return isExisting ? (
                <PassphraseWalletExistsFlow
                    device={device}
                    deviceOffer={deviceOffer}
                    authConfirmation={authConfirmation}
                    onSubmit={onSubmit}
                />
            ) : (
                <PassphraseWalletIsNotExistFlow
                    device={device}
                    deviceOffer={deviceOffer}
                    onSubmit={onSubmit}
                />
            );

        default: {
            const t: never = passphraseState?.state; // Todo: use `exhaustive` utils
            console.log(t);
        }
```

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseModal.tsx
```diff
@@ -1,179 +1,112 @@
...
-    };
+    const authConfirmation =
+        useSelector(selectIsDiscoveryAuthConfirmationRequired) || device?.authConfirm;
 
     const onSubmit = useCallback(
         (value: string, passphraseOnDevice?: boolean) => {
-            dispatch(onPassphraseSubmit({ value, passphraseOnDevice: !!passphraseOnDevice }));
+            if (!device) return;
+
+            if (authConfirmation) {
```
peter-sanderson, line 46: Is there reason why this logic is in the component and not in the `passphraseFlowManager`?

It seems that both `confirmPassphrase` and `submitPassphrase` are called only here. So I would say, that it can be just unified into one method on the `passphraseFlowManager` and called from here atomically

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseWalletConfirmation.tsx
```diff
@@ -1,97 +1,87 @@
...
-import { PassphraseWalletConfirmationStep2 } from './PassphraseWalletConfirmationStep2';
-import { PassphraseWalletConfirmationStep3 } from './PassphraseWalletConfirmationStep3';
-import { ContentType } from './types';
+import { useSelector } from '../../../../../hooks/suite';
+import { OpenGuideFromTooltip } from '../../../../guide';
+import { Translation } from '../../../Translation';
 
 type PassphraseWalletConfirmationContentProps = {
-    onSubmit: (value: string, passphraseOnDevice?: boolean) => void;
     onDeviceOffer: boolean;
```
peter-sanderson, line 15: It is super-confusion that this is named `onXyz` but it is a boolean

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseModal.tsx
```diff
@@ -1,179 +1,112 @@
...
-                            <OpenGuideFromTooltip
-                                data-testid="@tooltip/guideAnchor"
-                                id="/1_initialize-and-secure-your-trezor/6_passphrase.md"
-                            />
-                        }
-                    />
-                </Column>
-            </CardWithDevice>
-        </SwitchDeviceModal>
+    const deviceOffer = !!(
```
peter-sanderson, line 62: This is weird, I think this can be turned into selector and used where it is relevant and **not** to be prop-drilled down the component tree.

I think at the end it is always used in the `PassphraseTypeCard` so I think it can be moved there as `device` is there anyway. Or better, create a selector for it.

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseWalletConfirmation.tsx
peter-sanderson: But see https://github.com/trezor/trezor-suite/pull/17966/files#r2075020403 I think we can get rid of it complately

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseWalletConfirmation.tsx
```diff
@@ -1,97 +1,87 @@
...
     onSubmit: (value: string, passphraseOnDevice?: boolean) => void;
     onDeviceOffer: boolean;
     onCancel: () => void;
-    onRetry: () => void;
+    onBack: () => void;
     device: TrezorDevice;
+    deviceLoading?: boolean;
 };
 
 export const PassphraseWalletConfirmation = ({
```
peter-sanderson, line 65: `[to consider]`: I would consider naming all components that are modal with `Modal` suffix
- `PassphraseWalletConfirmation` -> `PassphraseWalletConfirmation`
- `PassphraseWalletBestPractices` -> `PassphraseWalletBestPracticesModal`
- `DiscoveryLoader` -> `DiscoveryLoaderModal`
- `PassphraseWalletIsNotExistFlow` -> ...
- `PassphraseWalletConfirmation` -> ...
- `PassphraseWalletExistsFlow` -> ...

Maybe it feels redundant, but to me, it seems a bit more clear what is actually just component and what is rendered as overlay modal. I know .. everything is modal in the PassphraseFlow, but it is not clear on all levels in the code.

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseWalletExistsFlow.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const passphraseState = useSelector(selectPassphraseFlowState);
+    const passphraseFlow = useSelector(selectPassphraseFlow);
+    const isExisting = useSelector(selectIsExistingWallet);
+
+    const onConfirmPassphraseDialogCancel = () => {
+        passphraseFlowManager.cancelFlow({ reason: 'auth-confirm-cancel' });
+    };
+
+    if (isExisting) {
+        if (passphraseState === 'exists-empty-wallet') {
```
peter-sanderson, line 38: this kinda feels wrong, I would rather see separate modal component for every state, exhaustively switched in the `PassphraseModal`

Here is some "second-level" switching for 3 cases but it seems to me as not necessary. Those 3 components can be (nearly) directly used in the  `PassphraseModal`

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/PassphraseWalletIsNotExistFlow.tsx
```diff
@@ -0,0 +1,61 @@
...
+}: PassphraseWalletIsNotExistFlowProps) => {
+    const { passphraseFlowManager } = useServices();
+    const passphraseState = useSelector(selectPassphraseFlowState);
+    const passphraseFlow = useSelector(selectPassphraseFlow);
+
+    const onConfirmPassphraseDialogCancel = () => {
+        passphraseFlowManager.cancelFlow({ reason: 'auth-confirm-cancel' });
+    };
+
+    if (passphraseState === 'not-exist-confirm-passphrase') {
```
peter-sanderson, line 32: Same situation as here: https://github.com/trezor/trezor-suite/pull/17966/files#r2075055284

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/UserContextModal.tsx
```diff
@@ -212,8 +206,6 @@ export const UserContextModal = ({ payload }: ReduxModalProps<typeof MODAL.CONTE
             );
         case 'unhide-token':
             return <UnhideTokenModal onCancel={onCancel} address={payload.address} />;
-        case 'passphrase-mismatch-warning':
```
peter-sanderson, line 216: when I search for this, I find that this can happen (see `suite-common/wallet-core/src/device/deviceThunks.ts:454`)

But no code actually handles this anywhere. So I think the code in `deviceThunks`  is kinda dead code somehow. 

image

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/UserContextModal.tsx
```diff
@@ -102,10 +100,6 @@ export const UserContextModal = ({ payload }: ReduxModalProps<typeof MODAL.CONTE
             return <QrScannerModal decision={payload.decision} onCancel={onCancel} />;
         case 'transaction-detail':
             return <TxDetailModal {...payload} onCancel={onCancel} />;
-        case 'passphrase-duplicate':
```
peter-sanderson, line 106: I think this is also lefover somwhat similar to https://github.com/trezor/trezor-suite/pull/17966/files#r2075068129

This handled by the PassphraseFlow I would say and it is triggered in the `suite-common/wallet-core/src/device/deviceThunks.ts:358` .. but I think it does nothing now? 

Or am I missing something?

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/UserContextModal.tsx
```diff
@@ -212,8 +206,6 @@ export const UserContextModal = ({ payload }: ReduxModalProps<typeof MODAL.CONTE
             );
         case 'unhide-token':
             return <UnhideTokenModal onCancel={onCancel} address={payload.address} />;
-        case 'passphrase-mismatch-warning':
-            return <PassphraseMismatchModal onCancel={onCancel} />;
         case 'connect-popup':
             return <ConnectPermissionsModal />;
         case 'walletconnect-proposal':
```
peter-sanderson, line 212: Also this switch shall be exhaustive.

peter-sanderson: And if we decide not handle something here, there shall be empty case for it anyway with a **very** good comment explaining the situation, and leading to the code where it is handled.

vojtatranta:
closing for not, will pick stuff if needed,

vojtatranta:
in favor of https://github.com/trezor/trezor-suite/pull/18724

PR #17967, opened 2025-03-28
fix: Troubleshooting tips in relation with WebUSB

### LINKED ISSUE #17964: Fix WebUSB Toubleshooting Tips (repeaded info, ...)
Image

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite/issues/17964

image


Lemonexe (approved):
LGTM. Tested web with webusb and desktop witj BT on | off.

:thought_balloon:  The component is getting quite too smart. Maybe it'd be nicer to remake it with slot component props instead of all the conditions. But that's just thinking out loud; let's not churn code refactoring it over and over again :laughing:

PR #17971, opened 2025-03-28
Http receiver hardening CR suggestion

Lemonexe:
## Description

CR suggestion for #17589

### DIFF
--- a/packages/node-utils/src/http.ts
+++ b/packages/node-utils/src/http.ts
@@ -9,6 +9,7 @@ import { Log, TypedEmitter, arrayPartition } from '@trezor/utils';
 import { getFreePort } from './getFreePort';
 
 type Request = RequiredKey<http.IncomingMessage, 'url'>;
+const isRequest = (request: http.IncomingMessage): request is Request => request.url !== undefined;
 type EventMap = { [event: string]: any };
 
 export type RequestWithParams<B = unknown, P = unknown> = Request & {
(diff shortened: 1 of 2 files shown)


packages/node-utils/src/http.ts
```diff
@@ -295,7 +296,7 @@ export class HttpServer<T extends EventMap> extends TypedEmitter<T & BaseEvents>
      */
     private onRequest = (request: http.IncomingMessage, response: http.ServerResponse) => {
         // mostly ts stuff. request should always have url defined.
-        if (!request.url) {
+        if (!isRequest(request)) {
```
Lemonexe, line 299: this type guard is useful so that later we can assign `requestWithParams: RequestWithParams` without `as`.
The previous code was runtime safe, but TS does not narrow down type when checking object properties
`if (!request.url) return` so it would scream that `url` can still be `undefined`...

packages/node-utils/src/http.ts
```diff
@@ -359,27 +363,22 @@ export class HttpServer<T extends EventMap> extends TypedEmitter<T & BaseEvents>
...
+                }
 
-            return response.end();
-        }
+                acc[param.replace(':', '')] = sanitized;
+
+                return acc;
+            }, {}),
+        };
+        if (response.statusCode === 403) return response.end();
```
Lemonexe, line 381: I'd like to avoid using throw as goto...

packages/node-utils/src/http.ts
```diff
@@ -359,23 +368,24 @@ export class HttpServer<T extends EventMap> extends TypedEmitter<T & BaseEvents>
...
-                    return acc;
-                },
-                {} as Record<string, string>,
-            );
-        } catch {
+                return acc;
+            },
+            {},
+        );
+        if (isSegmentInvalid) {
```
Lemonexe, line 388: I'd like to avoid using throw as goto

mroz22 (approved):
lgtm, thanks 👍 pushed one more unit test for this change

PR #17972, opened 2025-03-28
chore(suite): interface cleanup for TroubleshootingTips

Lemonexe:
## Description

Ex post CR suggestions for #17925

Mainly refactor props, IMO more intuitive now and without too many diffs.

And small nits

## Screenshots:
Proof that I did not mess up the order of the props:
**BT on**
bt
**BT off**
without bt
**What we've got here is failure to communicate**
failed comm

@coderabbitai ignore

### DIFF
--- a/packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
+++ b/packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
@@ -81,7 +81,8 @@ export const DeviceConnect = ({ setIsBluetoothConnectOpen }: DeviceConnectProps)
 
         return (
             <TroubleshootingTipsWithSections
-                label={<Translation id="TR_TREZOR_SAFE_7" />}
+                label={<Translation id="TR_CONNECTION_TYPE" />}
+                ctaLabel={<Translation id="TR_TREZOR_SAFE_7" />}
                 items={{
                     cable: { items: cableItem, label: <Translation id="TR_CABLE" /> },
                     bluetooth: { items: bluetoothItems, label: <Translation id="TR_BLUETOOTH" /> },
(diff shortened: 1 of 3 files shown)


packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
```diff
@@ -81,15 +81,15 @@ export const DeviceConnect = ({ setIsBluetoothConnectOpen }: DeviceConnectProps)
 
         return (
             <TroubleshootingTipsWithSections
-                label={<Translation id="TR_TREZOR_SAFE_7" />}
+                label={<Translation id="TR_CONNECTION_TYPE" />}
+                ctaLabel={<Translation id="TR_TREZOR_SAFE_7" />}
```
Lemonexe, line 85: before → after:
- `sectionLabel` → `label`
- `label` → `ctaLabel`

`label` is now the same thing for `TroubleshootingTipsWithSections` and for `TroubleshootingTips`, which was the most confusing thing for me personally..
Visually it is the most prominent label, as it headlines the table, so it makes sense that the name reflects that this is **_the_** label.

`ctaLabel` now actually behaves analogically to `toggleText` – it is just custom override, and if not supplied, a default will be used (`label` for cta, constant for toggle).

Here is currently the only usage of the `ctaLabel` override...

packages/suite/src/components/suite/troubleshooting/TroubleshootingTips.tsx
```diff
@@ -32,26 +32,27 @@ type SectionDefinition = { label: ReactNode; items: TroubleshootingTipsItem[] };
...
+    initiallyIsOpen,
     defaultSection,
     toggleText,
     'data-testid': dataTest,
 }: TroubleshootingTipsWithSectionsProps<K, T>) => {
-    const [isOpened, setIsOpened] = useState(opened === true);
-    const [selectedSection, setSelectedSection] = useState<K>(defaultSection);
+    const [isOpened, setIsOpened] = useState(initiallyIsOpen === true);
+    const firstSectionKey = Object.keys(items)[0] as K;
+    const [selectedSection, setSelectedSection] = useState<K>(defaultSection ?? firstSectionKey);
```
peter-sanderson, line 55: :clap:

peter-sanderson (approved):

PR #17974, opened 2025-03-28
chore(message-system): introduce t3w1 type to message system schema

MiroslavProchazka:
Introducing T3W1 type to the message system schema so we can target this model via message system in the future even on older suite versions


tomasklim (approved):
You can target it even without having it here

MiroslavProchazka:
It didnt pass the validation script, so I added it
image

PR #18000, opened 2025-03-31
chore(suite-native): android e2e flakiness

PeKne:
## Description
- bumps detox version to latest 20.36.3
- fixes `scrollUntilVisible` utility function for android

### DIFF
--- a/suite-native/app/e2e/utils.ts
+++ b/suite-native/app/e2e/utils.ts
@@ -79,18 +79,21 @@ export const restartApp = async () => {
 };
 
 export const scrollUntilVisible = async (
-    element: Detox.IndexableNativeElement,
+    target: Detox.IndexableNativeElement,
     scrollViewTestId: string = '@screen/mainScrollView',
 ) => {
     try {
         // Try to confirm that the element is visible without scrolling.
-        await detoxExpect(element).toBeVisible();
+        await detoxExpect(target).toBeVisible();
     } catch {
         // If the element is not visible, then use the scroll to find it.
-        await waitFor(element)
+        await waitFor(target)
             .toBeVisible()
             .whileElement(by.id(scrollViewTestId))
             .scroll(300, 'down');
+
+        // add extra scroll in case that the element is still not fully visible.
+        await element(by.id(scrollViewTestId)).scroll(150, 'down');
     }
 };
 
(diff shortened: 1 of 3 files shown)


suite-native/app/e2e/utils.ts
```diff
@@ -79,18 +79,21 @@ export const restartApp = async () => {
...
     } catch {
         // If the element is not visible, then use the scroll to find it.
-        await waitFor(element)
+        await waitFor(target)
             .toBeVisible()
             .whileElement(by.id(scrollViewTestId))
             .scroll(300, 'down');
+
+        // add extra scroll in case that the element is still not fully visible.
+        await element(by.id(scrollViewTestId)).scroll(150, 'down');
```
jbazant, line 96: **question:** I believe the comment here is misleading. The scroll adds another 150 points of scroll to 300 points from previous step in case the element was scrolled to be visible in previous command (line 90). In case previous scroll was not enough it will fail and never reaches here. Is this intended? How it actually helps?

PeKne: Sometimes, the element was visible only partially (e.g., peaking just a few pixels behind the bottom edge of the screen), and in that case, follow-up actions (e.g.,  `tap()`) fails. 

This fixes that, because now there is at least 150px visible of the target element and it should be enough for all our use cases.

PeKne: This fixes all the test fails ending with this error: https://github.com/trezor/trezor-suite/actions/runs/14171498734/job/39696482175?pr=17981.

Screenshot 2025-04-01 at 9 26 27 AM

matejkriz: It can also happen, that the element is visible without scrolling, so catch block is not executed, but it's again not visible fully, right? 

What about to add something like this:
```
const ensureFullyVisible = async () => {
        const elementFrame = await target.getAttributes();
        const scrollViewFrame = await scrollView.getAttributes();
        
        const scrollAmount = (elementFrame.y + elementFrame.height) - (scrollViewFrame.y + scrollViewFrame.height);
        if (scrollAmount > 0) {
            await scrollView.scroll(scrollAmount, 'down');
        }
    };
```

And call it from `finally` block?

matejkriz (approved):
If my suggestion is not working, I'm fine with the proposed solution.

PR #18004, opened 2025-03-31
Don't expect BIP-380 from older fw

marekrjpolak:
## Description

Recently we started to return BIP-380 descriptors from `GetPublicKey`, which was [added in 2.7.0 fw](https://github.com/trezor/trezor-firmware/blob/main/core/CHANGELOG.md#added-6), so we have to conditionally omit it from fixtures when testing older fw.

I'm not sure whether T1 supports it, but we're not testing it anyway. :smirk:


mroz22 (commented):
ah, sure, I forgot about the fact we are testing version 2 legacy firmwres. I only checked for model one and realized I am relatively save to ignore it.

mroz22 (approved):

PR #18005, opened 2025-03-31
chore(connect-explorer, connect-explorer-theme): upgrade next

komret:
Resolves https://github.com/trezor/trezor-suite/security/dependabot/318


martykan (approved):
Works, LGTM

PR #18007, opened 2025-03-31
feat(suite): hide maximum fee when data is missing

### LINKED ISSUE #17997: Show "maximum fee" in advanced fee settings only when data is available
Requirements
- show "maximum fee" line in advance fee section only if we know amount (cached it the same way as gas limit is cached)

Image

izmy:
## Description

Show the "maximum fee" line in the advance fee section only if the amount exists.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17997

## Screenshots:

image

image


tomasklim (changes requested):
Please use cache I use for gas limit in standard. That it does not flash on change
Screenshot 2025-03-31 at 22 13 33

tomasklim (approved):
lgtm

PR #18010, opened 2025-03-31
Chore/upgrade express

komret:
Upgrade Express to v5.

I ran the codemod from [migration guide ](https://expressjs.com/en/guide/migrating-5.html)but no changes were made.

I tested `auth-server` locally according to [this guide](https://github.com/trezor/trezor-suite/blob/develop/packages/auth-server/README.md).

### DIFF
--- a/packages/auth-server/package.json
+++ b/packages/auth-server/package.json
@@ -14,7 +14,7 @@
     },
     "dependencies": {
         "cors": "^2.8.5",
-        "express": "^4.21.2"
+        "express": "^5.1.0"
     },
     "devDependencies": {
         "@types/cors": "^2.8.17"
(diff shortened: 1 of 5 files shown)


Lemonexe (approved):
LGTM. I tested that the server runs & responds, but did not manage to jump through all the hoops in google could to successfully authenticate myself.

Lemonexe (approved):
Managed to get local google oauth up & running :rocket: 
I also reproduced the bug where it sends you twice through the flow when using testing app. And I also reproduced it on `develop` so it's preexisting..

:point_right: I'll update the docs to make it clearer.

komret:
@SocketSecurity ignore npm/type-is@2.0.1

PR #18013, opened 2025-03-31
feat: add TroubleShooting tip for removal of the Trezor from Bluettoth OS Settings

### LINKED ISSUE #17829: Device not removed from OS prevents new on to be paired
When bonds are erased, but the device is NOT manually removed from OS settings → then, when device is put into pairing mode again, it is NOT visible until its removed from the OS settings.

Image

AFAIK this cannot be solved and must adjust UI for it so user is properly informed and instructed what to do.

peter-sanderson:
Resolves: https://github.com/trezor/trezor-suite/issues/17829

image
image

### DIFF
--- a/packages/suite/src/components/suite/notifications/NotificationRenderer/NotificationRenderer.tsx
+++ b/packages/suite/src/components/suite/notifications/NotificationRenderer/NotificationRenderer.tsx
@@ -119,6 +119,12 @@ export const NotificationRenderer = ({
             return success(render, notification, 'TOAST_VERIFY_MESSAGE_SUCCESS');
         case 'error':
             return error(render, notification, 'TOAST_GENERIC_ERROR');
+        case 'cannot-open-bluetooth-settings-error':
+            return error(
+                render,
+                notification,
+                'TR_BLUETOOTH_CANNOT_OPEN_BLUETOOTH_SETTINGS_REMOVE_DEVICE',
+            );
         case 'clear-storage':
             return success(render, notification, 'TR_STORAGE_CLEARED');
         case 'firmware-check-authenticity-success':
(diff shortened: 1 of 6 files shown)


packages/suite/src/components/suite/notifications/NotificationRenderer/NotificationRenderer.tsx
```diff
@@ -119,6 +119,8 @@ export const NotificationRenderer = ({
             return success(render, notification, 'TOAST_VERIFY_MESSAGE_SUCCESS');
         case 'error':
             return error(render, notification, 'TOAST_GENERIC_ERROR');
+        case 'translated-error':
+            return error(render, notification, notification.error);
```
Lemonexe: The new `translated-error` goes against the convention of this `NotificationRenderer`.
`NotificationRenderer` has taxatively defined all possible notifications as single source of truth, and knows what to render accordingly, but here you are adding an escape clause – a generic configurable notification to bypass that SSoT. 

Personally, I wholeheartedly dislike the pattern of giving a component "an Id of what to render", and a very smart component that renders by Id. Ideally I'd refactor it entirely in the spirit of translated error, with more props for severity and optionally custom jsx content, and scrap all this switch case, so I love your approach!

:arrow_right:  But for the time being, let's stick to consistency please :pray: 
```suggestion
        case 'cannot-open-bluetooth-settings':
            return error(render, notification, 'TR_BLUETOOTH_CANNOT_OPEN_BLUETOOTH_SETTINGS_REMOVE_DEVICE');
```

Lemonexe (approved):
LGTM. Tested locally rebased to `fix-web-usd-troubleshooting-tips` and `develop` for both success and failure when opening the desktop environment BT settings.

But first, please address:
- :speech_balloon: my comment above
- :rabbit2: , he's right you know :smile:

packages/suite/src/components/suite/notifications/NotificationRenderer/NotificationRenderer.tsx
peter-sanderson: we are already breaking this pattern here: 
image

Lemonexe: Yes, I did notice, but only partially. That one is still a case for specific error subject (only tor toggle), not a generic one

peter-sanderson: ill change it

PR #18017, opened 2025-04-01
fix(ci): Added missing permissions to all e2e jobs

HajekOndrej:
## Description

As discussed with @vdovhanych I have added permissions to all E22 jobs. They were using the default ones and we think it's better to define them explicitly.

## Related Issue

Resolve


vdovhanych (approved):
lgtm

Issue #16985, opened 2025-02-13
Fix macOS version detection


komret:
Browsers (including Electron) report all Mac OS versions from 10.15.7 onwards as 10.15.7 to prevent fingerprinting. This makes our `getOsVersion` util unusable for Mac OS - for example, it cannot be reliably used in messsage system, see https://github.com/trezor/trezor-suite/pull/16741. However, there is a way to circumvent it in desktop apps. It was partially replaced [here](https://github.com/trezor/trezor-suite/pull/15789), but it should be done everywhere in the app.

Lemonexe:
I was curious if there is any alternative at all in browser.
The `desktopOsVersion` is never available on Suite Web.
On chromium based browsers, we can use
```
window.navigator.userAgentData.getHighEntropyValues(['platform', 'platformVersion']).then(console.log)
```
But not available on Firefox or Safari..


But then again, Desktop is all we need..

tomasklim:
Duplicate of #8674

tomasklim:
> But not available on Firefox or Safari..

`Firefox` is depracated
`Safari` is not supported

Issue #15770, opened 2024-12-04
Data Analytics: New desktop OS tracking


kolomaznikt:
We would like to add new parameters to event [suite-ready](https://www.notion.so/satoshilabs/suite-ready-1e5cbba0a1a545edb437185c37ddafdc?pvs=4) to correctly track OS name and version for desktop, especially for mac. The new method is described [here](https://www.notion.so/satoshilabs/New-operating-system-tracking-for-desktop-152dc526060680eaaeb3d89162d69a4e)  in Notion. There should be 3 new parameters:

- `desktopOsVersion` (from release)
- `desktopOsName` (from platform)
- `desktopOsArchitecture` (from architecture)

bosomt:
**QA OK**

```
desktopOsVersion=24.1.0&desktopOsName=darwin&desktopOsArchitecture=arm64
```

in sent https://data.trezor.io/suite/log/desktop/


**Info:**
 - Suite version: desktop 25.1.0 (0b116799334cb4a05fcb59cc480deb7c4b9a63e9)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.1.0 Chrome/128.0.6613.186 Electron/32.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.7 regular (revision e196413bb745443610211c325623ba7cee7309e9)
 - Transport: BridgeTransport 2.0.33

Issue #8674, opened 2023-06-12
macOS/Windows version is incorrectly reported which breaks analytics and message system


tomasklim:
**Describe the bug**

- our current v1 version of `ua-parser-js` detects computer data from user agent which does not reflect macOS version correctly

- update this lib to v2  https://docs.uaparser.dev/intro/whats-new.html

https://www.notion.so/satoshilabs/e19845789ccb47a0baf36d7f8463f196?v=2183c8228ebb4df38f358d1b2744e1b8&p=f1129e10da8d418da11186761754ad36&pm=s

**Info:**

-   Suite Version: 23.6.0

**How to reproduce**

1. Check analytics

**Expected behavior**
`macOS` version is reported correctly

**Screenshots**
![Screenshot 2023-06-12 at 15 15 54](https://github.com/trezor/trezor-suite/assets/33235762/9649c236-fba7-46c1-8f32-71a253b14a69)

tomasklim:
Add note that it was fixed here https://www.notion.so/satoshilabs/Changelog-Suite-1551ab666b1943f080ff56ffc6896d12

sime:
Inspired by https://github.com/electron/electron/issues/3012#issuecomment-643794970 for desktop suite, we can invoke `os.version()` and deduce the macOS version based on the kernel version.

```
> os.version()
'Darwin Kernel Version 22.6.0: Mon Feb 19 19:43:41 PST 2024; root:xnu-8796.141.3.704.6~1/RELEASE_ARM64_T8103'
```

Where **xnu-8796** is the darwin version for [macOS 13.7 (Ventura)](https://en.wikipedia.org/wiki/MacOS_Ventura#Release_history)

tomasklim:
@komret https://docs.uaparser.dev/guides/how-to-detect-macos-10157-using-javascript.html please check

Lemonexe:
@tomasklim I believe this is solved in https://github.com/trezor/trezor-suite/pull/15789. I did not know about this issue..
The article you linked says it's impossible to detect from userAgent alone, that's why I've added another field, obtained from node api. And it is only available on desktop, not on web 😞

Also, it cannot be used in message system until someone adds conditions for `desktopOsVersion` to `messageSystemUtils`

tomasklim:
```
    uap.getOS()
        .withClientHints()
        .then(os => {
            console.log('Is desktop', isDesktop(), os);
        });
```

desktop:
[...]

Issue #6877, opened 2022-11-10
Trezor Suite log and open issue on Github are showing wrong Windows version


STew790:
**Describe the bug**
Log of the Suite in settings and "Info" part of creating a new issue on Github through Suite shows "Windows 10" even though I have Windows 11.

**Steps to reproduce:**
1. Start Suite on Windows 11 machine
2. Go to settings
3. Open Suite log and find "osVersion" line

1. Go to Dev Settings
2. Click on "Open issue"
3. In the "Info" part in the opened issue, it states "Windows NT 10.0"

**Info:**
 - Suite version: desktop 22.11.1 (73ccf723353e0575a64f58b2f5775d20f072311f)
 - Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/22.11.1 Chrome/100.0.4896.160 Electron/18.3.5 Safari/537.36
 - OS: Win32
 - Screen: 1920x1080
 - Device: 
 - Transport: bridge 2.0.31

**Expected result:**
Log and Github issue created through Suite should reflect what Windows version I have

**Actual result:**
Log and Github issue created through Suite shows wrong Windows version

**Screenshots:**
![image](https://user-images.githubusercontent.com/48477147/201141450-b43607f6-fd6e-4e1c-92d8-bf9f105d24b3.png)

evgenysl:
Confirmed same behaviour on 25.3.2

bosomt:
**QA OK**


- tested on latest develop Win 11 is reported in user log
- open issue is still Win32 /same like macARM is still reported as macINTEL /
- we have separate issue for github issue creation https://github.com/trezor/trezor-suite/issues/5880

PR #18023, opened 2025-04-01
Feat: use UAParser2 for OS version

Lemonexe:
## Description

:eyes: you can review commit by commit if you wish :slightly_smiling_face: 

- use [new UAParser2 API](https://docs.uaparser.dev/guides/how-to-detect-macos-10157-using-javascript.html) to get system information in browser, without need for nodeJS
  - detection of new macOS or Win11 versions is only available on Chromium-based browsers. But we do not support Mozilla or Safari so that's ok :heavy_check_mark: 
    - BTW it's because it uses [getHighEntropyValues under the hood](https://github.com/faisalman/ua-parser-js/blob/598c51c69cb6533ab3ab85085f32727ddb416fd3/src/main/ua-parser.mjs#L1046-L1057)
  - this also enables us to detect Win11 vs Win10, which even nodeJS couldn't distinguish properly, see below! 
- remove `getSystemInformation` IPC api and `system-information` module from electron-main (unneeded)

## Related Issue

Resolves (again) #16985
Effectively reverts #15770
Necessary for #6877 (next PR shall implement it in App Logs)
Necessary for #8674 (next PR shall implement it in message system)

More info [in notion](https://www.notion.so/satoshilabs/e19845789ccb47a0baf36d7f8463f196?v=2183c8228ebb4df38f358d1b2744e1b8&p=f1129e10da8d418da11186761754ad36&pm=s)

## Notes

From [testing branch](https://github.com/trezor/trezor-suite/tree/feat/use-uaparser2-for-osVersion-TESTING) I printed `SYSTEMINFO` in console.
You can verify that the `desktopOs` values from nodeJS are fully replaceable by UAParser. They are not 1:1, but equivalent.

**macOS** _(they have two different versioning sequences, [see wiki page](https://en.wikipedia.org/wiki/MacOS_version_history#Releases), the darwin 24 reported by desktopOsVersion corresponds to macOS 15 reported by osVersion. Either of them is useful for us, I believe it doesn't matter. Actually, the latter is IMO more understandable. Hardly anyone knows what is "darwin version", it's the macOS version that you see in UI)_
```
desktopOsVersion: “24.3.0",
desktopOsName: “darwin”,
desktopOsArchitecture: “arm64",
osCpuArch: “arm64",
osName: “macos”,
osVersion: “15.3.2"
```

**Win10**
```
desktopOsArchitecture: "x64"
desktopOsName: "win32"
desktopOsVersion: "10.0.19045"
osCpuArch: "amd64"
osName: "windows"
osVersion: "10"
```

**Win11** _(now we can easily tell 10 vs 11. We could determine it from build number 26100, but it's more work)_
```
desktopOsArchitecture: "x64"
desktopOsName: "win32"
desktopOsVersion: "10.0.26100"
osCpuArch: "amd64"
osName: "windows"
osVersion: "11"
```

**Linux**
```
desktopOsArchitecture: "x64"
desktopOsName: "linux"
desktopOsVersion: "6.8.0-56-generic"
osCpuArch: "amd64"
osName: "linux"
osVersion: "6.8.0"
```


packages/suite/src/utils/suite/analytics.ts
```diff
@@ -51,7 +39,8 @@ export const redactRouterUrl = (url: string) => url.replace(/coinjoin/g, 'taproo
 export const getSuiteReadyPayload = async (
     state: AppState,
 ): Promise<SuiteAnalyticsEventSuiteReady['payload']> => {
-    const systemInformation = await getOptionalSystemInformation();
+    const osVersion = await getOsVersion();
+    const osCpuArch = await getCpuArch();
```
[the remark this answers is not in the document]
Lemonexe: No, we are fine with `''` in analytics for Firefox or Safari, they are not officially supported.

packages/env-utils/src/envUtils.native.ts
```diff
@@ -24,7 +24,11 @@ const getBrowserVersion = () => '';
 
 const getDeviceType = () => '';
 
-const getOsVersion = () => `${Platform.Version}`;
+// no need for async fn on native; only for the sake of consistency with Web/Desktop
+const getOsVersion = () => Promise.resolve(`${Platform.Version}`);
+const getDeprecatedOsVersion = () => `${Platform.Version}`;
```
Lemonexe, line 29: `getOsVersion` is not directly used in native at all.
It is used indirectly in message system, which currently uses `getDeprecatedOsVersion` so it will be compatible :heavy_check_mark: Refactoring of message system will be done in next PR, and that will affect mobile too..

packages/suite/src/utils/suite/analytics.ts
Lemonexe: Nit: could be `Promise.all`, but the `getHighEntropyValues` are almost instant so I think it's ok to be lazy and await them sequentially.

komret: I'd prefer `Promise.all` anyway, it fits here better.

komret (approved):
I appreciate the separation into commits and the comments 🎖️

PR #18026, opened 2025-04-01
chore(suite-desktop): make node-bridge rollout conditions equal for a…

mroz22:
…ll platforms (1%)

what next? do you want to go higher then 1%? going from 0 to 1 is quite a big change for windows 🤔 

cc @MiroslavProchazka @Hannsek


Hannsek:
> what next? do you want to go higher then 1%? going from 0 to 1 is quite a big change for windows 🤔

We'll see after the release, no?

MiroslavProchazka:
Let’s go to 1% on all platforms on 25.4. Then, I would wait and go for 10%. But let’s see what this change will bring

martykan (approved):

PR #18029, opened 2025-04-01
fix(suite-native): ios sharp build

PeKne:
Attempt to fix CI iOS build failing due to sharp lib build. Solution inspired by: https://github.com/lovell/sharp/issues/4253.


matejkriz (approved):
Let's give it a try. Comment in code would be nice, but we can always check the PR that introduced this env var, it's a nitpick.

PeKne:
makes sense, comment added.

PR #18058, opened 2025-04-01
chore(connect-examples): remove electron-renderer-with-assets example

mroz22:
I was about to update the `wait-on` dependency and realized that the `electron-renderer-with-assets` connect example does not work anymore. 

What do you guys think? Is it worth fixing? My intuition is no because we want to streamline the way how people use connect and the way documented by this example is not something that we really want to promote. cc @Hannsek, @martykan


martykan:
Let's kill it 👍 
It's also mentioned [connect-examples README](https://github.com/trezor/trezor-suite/blob/develop/packages/connect-examples/README.md) so should be removed there

martykan (approved):

PR #18076, opened 2025-04-02
chore(message-system): temporarily disable other error on hash check

MiroslavProchazka:
We need to tackle better the UI so users will not panic - cc @hynek-jina & @rabbitholiness

### DIFF
--- a/suite-common/message-system/config/config.v1.json
+++ b/suite-common/message-system/config/config.v1.json
@@ -1,7 +1,7 @@
 {
     "version": 1,
-    "timestamp": "2025-03-25T00:00:00+00:00",
-    "sequence": 86,
+    "timestamp": "2025-04-02T00:00:00+00:00",
+    "sequence": 87,
     "actions": [
         {
             "conditions": [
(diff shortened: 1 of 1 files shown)


Lemonexe (approved):
Tested locally that it
- hides other error UI
- does not hide hash-mismatch UI
:heavy_check_mark:

hynek-jina:
Resolved in Figma - Giving some time to copy review...

suite-common/message-system/config/config.v1.json
```diff
@@ -1528,6 +1528,45 @@
                 ]
             }
         },
+        {
+            "conditions": [
+                {
+                    "environment": {
+                        "desktop": "*",
+                        "mobile": "!",
+                        "web": "*"
```
komret: Let's disable for 25.3 only as it will be fixed in 25.4.

komret (changes requested):

suite-common/message-system/config/config.v1.json
komret: I'll do it.

komret (approved):

komret (approved):

PR #18086, opened 2025-04-02
fix(suite): fix warnings in dev console when fiat rates load fails

vojtatranta:
## Related Issue

Resolve

### DIFF
--- a/suite-common/wallet-core/src/fiat-rates/fiatRatesReducer.ts
+++ b/suite-common/wallet-core/src/fiat-rates/fiatRatesReducer.ts
@@ -71,7 +71,9 @@ export const prepareFiatRatesReducer = createReducerWithExtraDeps(
                         state[rateType][fiatRateKey] = {
                             ...currentRate,
                             isLoading: false,
-                            error: result.reason || `Failed to update ${ticker.symbol} fiat rate.`,
+                            error:
+                                String(result.reason) ||
+                                `Failed to update ${ticker.symbol} fiat rate.`,
                         };
 
                         continue;
--- a/suite-common/wallet-core/src/fiat-rates/fiatRatesThunks.ts
+++ b/suite-common/wallet-core/src/fiat-rates/fiatRatesThunks.ts
@@ -89,11 +89,6 @@ export const updateTxsFiatRatesThunk = createThunk(
     },
 );
 
-const fetchFn: Record<RateTypeWithoutHistoric, typeof fetchCurrentFiatRates> = {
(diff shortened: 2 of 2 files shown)


suite-common/wallet-core/src/fiat-rates/fiatRatesReducer.ts
```diff
@@ -71,7 +71,7 @@ export const prepareFiatRatesReducer = createReducerWithExtraDeps(
                         state[rateType][fiatRateKey] = {
                             ...currentRate,
                             isLoading: false,
-                            error: result.reason || `Failed to update ${ticker.symbol} fiat rate.`,
+                            error: `${result.reason || `Failed to update ${ticker.symbol} fiat rate.`}`,
```
Lemonexe: Nit: no need to wrap the entire thing, when the fallback is already string
```suggestion
                            error: String(result.reason) || `Failed to update ${ticker.symbol} fiat rate.`,
```

suite-common/wallet-core/src/fiat-rates/fiatRatesThunks.ts
```diff
@@ -130,20 +125,43 @@ export const updateFiatRatesThunk = createThunk(
...
+                            isElectrumBackend,
+                        });
+                    case 'lastWeek':
+                        return fetchLastWeekFiatRates({
+                            ticker,
+                            localCurrency,
+                            isElectrumBackend,
+                        });
+                    default:
+                        ((_: never) => {})(rateType);
```
Lemonexe, line 143: nice :+1:

suite-common/wallet-core/src/fiat-rates/fiatRatesThunks.ts
```diff
@@ -89,11 +89,6 @@ export const updateTxsFiatRatesThunk = createThunk(
     },
 );
 
-const fetchFn: Record<RateTypeWithoutHistoric, typeof fetchCurrentFiatRates> = {
-    current: fetchCurrentFiatRates,
-    lastWeek: fetchLastWeekFiatRates,
-};
```
Lemonexe, line 95: good riddance, I think the code was just more confusing with this map than explicitly :+1:

Lemonexe (approved):
Good catch, nice cleanup, seems to work :rocket: 
Thanks for fixing it!

Lemonexe:
Merging, failing test is confirmed to be broken [(slack)](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1743572836379729)

PR #18094, opened 2025-04-02
Connect popup desktop nitpicks

mroz22:


suite-common/connect-popup/src/connectPopupThunks.ts
```diff
@@ -57,11 +57,11 @@ export const connectPopupCallThunkInner = createThunk<
                 dispatch(
                     connectPopupActions.initiateCall({
                         state: 'call-error',
-                        callError: ERRORS.TypedError('Method_NotAllowed'),
+                        callError: TypedError('Method_NotAllowed'),
                     }),
                 );
 
-                return;
+                throw TypedError('Method_NotAllowed');
```
mroz22, line 64: we need to return something I think

suite-common/connect-popup/src/connectPopupThunks.ts
```diff
@@ -57,11 +57,11 @@ export const connectPopupCallThunkInner = createThunk<
                 dispatch(
                     connectPopupActions.initiateCall({
                         state: 'call-error',
```
mroz22, line 59: this does not have any UI yet, or does it? something similar to https://github.com/trezor/trezor-suite/blob/develop/suite-native/module-connect-popup/src/screens/ConnectPopupScreen.tsx#L99

martykan: Not yet on desktop. Also the `push_tx` restriction is only needed on mobile due to the callback

martykan (approved):

PR #18096, opened 2025-04-02
Chore/run depcheck in root

komret:
Follow-up to https://github.com/trezor/trezor-suite/pull/17954. We skipped `depcheck` for root until now. Enabling for root and adding some of the newly reported dependencies to ignore list.


Lemonexe (approved):
Very nice!

PR #18098, opened 2025-04-02
refactor(trading): move send transaction

### LINKED ISSUE #17733: Duplicate swap context functions in suite-common and add more tests

adderpositive:
## Description
- created copy of send transaction

## Related Issue
Resolve https://github.com/trezor/trezor-suite/issues/17733

### DIFF
--- a/suite-common/trading/src/thunks/exchange/sendTransactionThunk.ts
+++ b/suite-common/trading/src/thunks/exchange/sendTransactionThunk.ts
@@ -1,0 +1,123 @@
+import { createThunk } from '@suite-common/redux-utils';
+import { amountToSmallestUnit } from '@suite-common/wallet-utils';
+
+import { exchangeThunks, tradingThunks } from '../';
+import { SendDexTransactionThunkProps } from './sendDexTransactionThunk';
+import { TRADING_EXCHANGE_THUNK_PREFIX } from '../../constants';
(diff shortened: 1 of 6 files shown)


suite-common/trading/src/thunks/exchange/sendTransactionThunk.ts
```diff
@@ -0,0 +1,123 @@
...
+            tradingThunks.recomposeAndSignTxThunk({
+                account,
+                address: sendAddress,
+                amount: sendStringAmount,
+                destinationTag: sendPaymentExtraId,
+                signAndPushSendFormTransaction,
+            }),
+        );
+
+        if (!payload || (payload && 'error' in payload) || !payload.success) {
```
TomasBoda: nitpick:
```suggestion
        if (!payload || 'error' in payload || !payload.success) {
```

adderpositive: Updated here https://github.com/trezor/trezor-suite/pull/18098/commits/c3049859767e62634fcd713528021d469442a9f8

izmy (approved):

LukasRada (changes requested):
Please consider to extend assert section in tests with another checks. For example check that in `should return error` is called `recomposeAndSignTxThunk` zero times.

LukasRada (approved):

PR #18107, opened 2025-04-03
Chore: remove dead code in CommonDB

### LINKED ISSUE #18106: Simplify CommonDB
Remove obsolete code in 

Discovered in #18047 while bumping an outdated library – but instead, the whole code around it will be deleted. 

## QA instructions

Test suite storage both Web & Desktop
- migration should work
- clearing storage should work
- persisting anything in storage should work (change something in settings, activate coins etc.)

**Optional:** test on Firefox in anonymous mode – some of the dead code was a workaround specifically for that, and I confirmed it's not necessary anymore. But we don't officially support it, so whatever 🤷

Lemonexe:
## Description

- remove broadcast message events for Storage operations
  - they were never listened to anywhere in the repo!
    - would expect it to communicate with some iframe, web workers, or in-between Suite Web tabs
  - unclear what purpose this feature had at all? Orig PR didn't say. Maybe just code copied from internet, idk.  
-  remove now-obsolete IDB workaround specifically for Firefox in Incognito mode

## Related Issue

Resolve #18106


packages/suite-storage/package.json
```diff
@@ -17,10 +17,8 @@
     "react-native": "./src/native/index.ts",
     "browser": "./src/web/index.ts",
     "dependencies": {
-        "@trezor/env-utils": "workspace:*",
         "@trezor/eslint": "workspace:*",
         "@trezor/utils": "workspace:*",
-        "broadcast-channel": "^7.0.0",
```
Lemonexe, line 23: - [This lib](https://github.com/pubkey/broadcast-channel) is what led me to find this obsolete code at all (I was about to bump it)
- It' super weird because `BroadcastChannel` is natively offerred by browsers, and we rely on the native implementation at multiple places!
- _only here_ we use this polyfill lib for old browsers – meaning [Chrome & Firefox < 2016 & Safari < 2022](https://caniuse.com/broadcastchannel), lol

packages/suite-storage/src/web/index.ts
```diff
@@ -67,39 +62,14 @@ class CommonDB<TDBStructure> {
...
-        // this solution calls callback function from within onerror/onsuccess event handlers.
-        // For other browsers checking the window.indexedDB should be enough.
-        new Promise<boolean>(resolve => {
-            if (isFirefox()) {
-                const r = indexedDB.open('test');
-                r.onerror = () => resolve(false);
-                r.onsuccess = () => {
-                    indexedDB.deleteDatabase('test');
-                    resolve(true);
-                };
```
Lemonexe, line 89: This workaround is not needed anymore. The [bug linked in comment](https://bugzilla.mozilla.org/show_bug.cgi?id=781982)  was marked resolved 5 yrs ago.
I also tested it myself in Firefox Private window at any `https` page, **not** `about:blank` or `file:///` etc.:
```
const r = indexedDB.open('test');
r.onerror = () => console.log('NOT SUPPORTED');
r.onsuccess = () => {indexedDB.deleteDatabase('test'); console.log('WORKS'); };
```

packages/suite-storage/src/web/index.ts
```diff
@@ -109,22 +79,11 @@ class CommonDB<TDBStructure> {
...
-    notify = (store: StoreNames<TDBStructure>, keys: any[]) => {
-        // sends the message containing store, keys which were updated to other tabs/windows
-        const message = { store, keys };
-        this.broadcastChannel.postMessage(message);
-    };
-
-    onChange = (handler: (event: StorageMessageEvent<TDBStructure>) => any) => {
-        // listens to the channel. On receiving a message triggers the handler func
-        this.broadcastChannel.onmessage = handler;
+        return Promise.resolve(isSupported && !this.blocking && !this.blocked);
```
Lemonexe, line 86: At first I thought that removing the FF workaround can reduce this whole chain of function from async to sync.
Yes it can, but it's not practical!
In [storageActions](https://github.com/trezor/trezor-suite/blob/bc3e808263a13c72c7fe2a5d3d3f5d44709dddc1/packages/suite/src/actions/suite/storageActions.ts) there are 37 occurrences of `await db.isAccessible()`. These functions have to consistently return Promise, but without `await` you'd have to remove `async`, and it would be a lot of boilerplate code to `Promise.resolve()` everything to ensure they always return a Promise :facepalm:

packages/suite-storage/src/web/index.ts
```diff
@@ -245,11 +203,10 @@ class CommonDB<TDBStructure> {
                     });
                 })
                 .concat(tx.done),
-        ).then(_ => {
-            this.notify(store, keys);
-        });
+        );
+        const aggregatedPromise: Promise<void> = promisesOfItems.then(() => {});
```
Lemonexe, line 207: this fn is expected to return `Promise<void>`, not `Promise<void[]>` (would fail downstream in suite).
Naming the function makes this clear without code commentary

Lemonexe:
Failing test is confirmed to be broken [(slack)](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1743572836379729)

marekrjpolak (approved):
Haven't tested, looks great tho. By the way, the trade deficit of `+17 -158` lines is enormous so I'm gonna impose some tariffs on you.

image

PR #18126, opened 2025-04-03
Feat: `transport-bluetooth` rust websocket server

szymonlesisz:
## Description

cherry-picking initial part of `trezor-bluetooth` websocket server aka. bridge for bluetooth

the server will be used in `suite-desktop` as a bluetooth transport.
server has no method implementation yet, returns `{"error":"NOOP"}` for the requests, bluetooth logic will follow in upcoming pull requests.

im picking this part to get the early review of the architecture, naming and binary builds

### DIFF
--- a/packages/transport-bluetooth/src/server/connection_handler.rs
+++ b/packages/transport-bluetooth/src/server/connection_handler.rs
@@ -1,0 +1,168 @@
+use futures::{SinkExt, StreamExt};
+use http_body_util::Full;
+use hyper::{
+    body::{Bytes, Incoming},
+    Response as HyperResponse,
+};
+use hyper_tungstenite::{tungstenite::Message, HyperWebsocket};
+use hyper_util::rt::TokioIo;
+use log::info;
+use std::sync::Arc;
+use tokio::net::TcpListener;
+use tokio::sync::broadcast;
+use tokio::sync::Mutex;
+
+use crate::server::handle_message;
+use crate::server::types::{AbortProcess, ChannelMessage};
+use crate::server::utils;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ServerError {
+    #[error("WebSocket error: {0}")]
+    WebSocketError(#[from] hyper_tungstenite::tungstenite::Error),
+
+    #[error("HTTP error: {0}")]
+    HttpError(#[from] hyper::Error),
+
+    #[error("IO error: {0}")]
+    IoError(#[from] std::io::Error),
+}
+
+async fn handle_ws_connection(peer: String, websocket: HyperWebsocket) -> Result<(), ServerError> {
+    info!("WebSocket connection: {peer}");
+
+    let websocket = websocket.await?;
+    let (ws_write, mut ws_read) = websocket.split();
+    let (sender, mut receiver) = broadcast::channel::<ChannelMessage>(32);
+
+    // create websocket stream mutex to be shared between two threads
+    let ws_write = Arc::new(Mutex::new(ws_write));
+
+    // start thread and listen for ChannelMessages emitted by current connection processes
+    let ws_write_event = ws_write.clone();
+    let channel_message_listener = tokio::spawn(async move {
+        while let Ok(event) = receiver.recv().await {
+            match event {
+                ChannelMessage::Notification(event) => {
+                    info!("Sending notification to peer {peer}");
+                    let response = match serde_json::to_string(&event) {
+                        Ok(json) => json,
+                        Err(err) => {
+                            info!("Error serialize notification {err:?}");
+                            return;
+                        }
+                    };
+
+                    let mut ws = ws_write_event.lock().await;
+                    if let Err(err) = ws.send(response.into()).await {
+                        info!("Error sending notification {err:?}");
+                    };
+                }
+                _ => {
+                    info!("ChannelMessage listener {event:?}");
+                }
+            }
+        }
+    });
+
+    // in current thread keep listening for incoming websocket messages
+    let ws_write_method = ws_write.clone();
+    while let Some(msg) = ws_read.next().await {
+        // TODO: panic thrown here when computer suspended (macos?)
+        // fallback to empty json. it should fail in handle_message serialization
+        let request = msg.unwrap_or(Message::text("{}"));
+        let ws_write_method = Arc::clone(&ws_write_method);
+        let sender = sender.clone();
+        // spawn new thread and process request
+        tokio::spawn(async move {
+            let response = handle_message(request.clone(), sender).await;
+            match response {
+                Some(response) => {
+                    let mut ws = ws_write_method.lock().await;
+                    if let Err(err) = ws.send(response).await {
+                        info!("Error response to ws {err:?}");
+                    };
+                }
+                None => {
+                    info!("No response request {request:?}");
+                }
+            }
+        });
+    }
+
+    // peer disconnected
+    channel_message_listener.abort();
+
+    if let Err(err) = sender.send(ChannelMessage::Abort(AbortProcess::ClientDisconnected)) {
+        info!("AbortProcess not sent {err}");
+    }
+
+    info!("Closing websocket connection");
+
+    Ok(())
+}
+
+async fn handle_http_request(
+    peer: String,
+    req: hyper::Request<Incoming>,
+) -> Result<HyperResponse<Full<Bytes>>, ServerError> {
+    // TODO: cors check like trezord-go and node-bridge
+    // let = req.headers().get("origin");
+
+    if hyper_tungstenite::is_upgrade_request(&req) {
+        let (response, websocket) = match hyper_tungstenite::upgrade(req, None) {
+            Ok(r) => r,
+            Err(e) => {
+                info!("Failed to upgrade WebSocket: {e:?}");
+                return Ok(HyperResponse::new(Full::<Bytes>::from("Upgrade failed")));
+            }
+        };
+        tokio::spawn(async move {
+            if let Err(err) = handle_ws_connection(peer.clone(), websocket).await {
+                log::error!("WebSocket connection error for peer {}: {:?}", peer, err);
+
+                match err {
+                    ServerError::WebSocketError(ws_err) => {
+                        log::debug!("WebSocket protocol error: {}", ws_err);
+                    }
+                    ServerError::IoError(io_err) => {
+                        log::warn!("IO error in WebSocket connection: {}", io_err);
+                    }
+                    ServerError::HttpError(http_err) => {
+                        log::warn!("HTTP error in WebSocket connection: {}", http_err);
+                    }
+                }
+            }
+        });
+        Ok(response)
+    } else {
+        // TODO: serve index.html file
+        Ok(HyperResponse::new(Full::<Bytes>::from("OK")))
+    }
+}
+
+pub async fn start_server(address: &str) -> Result<(), ServerError> {
+    let tcp_listener = TcpListener::bind(&address).await.expect("Failed to bind");
+    info!("Version: {} Listening on: {}", utils::APP_VERSION, address);
+
+    let mut http = hyper::server::conn::http1::Builder::new();
+    http.keep_alive(true);
+
+    loop {
+        let (stream, _) = tcp_listener.accept().await?;
+        let peer = stream.peer_addr().expect("Missing peer address");
+
+        let service =
+            hyper::service::service_fn(move |req| handle_http_request(peer.to_string(), req));
+
+        let connection = http
+            .serve_connection(TokioIo::new(stream), service)
+            .with_upgrades();
+
+        tokio::spawn(async move {
+            if let Err(err) = connection.await {
+                info!("HTTP connection error {err:?}");
+            }
+        });
+    }
+}
--- a/packages/transport-bluetooth/src/server/message_handler.rs
+++ b/packages/transport-bluetooth/src/server/message_handler.rs
@@ -1,0 +1,93 @@
+use hyper_tungstenite::tungstenite::Message;
+use log::info;
+use tokio::sync::broadcast::Sender;
+
+use crate::server::methods;
+use crate::server::types::{ChannelMessage, WsError, WsRequest, WsRequestMethod, WsResponse};
+
+pub async fn handle_message(message: Message, sender: Sender<ChannelMessage>) -> Option<Message> {
+    let msg: String = match message {
+        Message::Text(msg) => Some(msg.to_string()),
+        Message::Binary(msg) => {
+            let msg_str = msg.escape_ascii().to_string();
+            if msg_str == "PING" {
+                Some(msg_str.to_string())
+            } else {
+                None
+            }
+        }
+        Message::Close(msg) => {
+            if let Some(msg) = &msg {
+                info!("Message::Close code {}, {}", msg.code, msg.reason);
+            } else {
+                info!("Message::Close without message");
+            }
+            None
+        }
+        msg => {
+            info!("unknown Message {msg:?}");
+            None
+        }
+    }?;
+
+    if msg.to_string() == "PING" {
+        return Some(Message::text("PONG"));
+    }
+
+    info!("WsRequest: {:?}", msg);
+    let request = match serde_json::from_str::<WsRequest>(&msg) {
+        Ok(req) => req,
+        Err(err) => {
+            let json_error = serde_json::json!({
+                "id": "-1", // TODO: try to read "id" from malformed json? this could not be exact WsRequest but it can have "id"
+                "error": format!("WsRequest serialization error: {err:?}"),
+            });
+            return Some(Message::text(json_error.to_string()));
+        }
+    };
+    info!("Method: {request:?}");
+
+    let payload = match request.method.clone() {
+        WsRequestMethod::GetInfo => methods::get_info().await,
+        _ => methods::noop().await, // TODO: implement methods
+    };
+
+    match payload {
+        Ok(payload) => {
+            info!("WsResponse {payload:?}");
+
+            let response = WsResponse {
+                id: request.id.to_string(),
+                payload,
+            };
+
+            match serde_json::to_string(&response) {
+                Ok(json) => Some(Message::text(json)),
+                Err(err) => {
+                    let err_json = serde_json::json!({
+                        "id": request.id,
+                        "error": format!("WsResponse serialization error: {err:?}"),
+                    });
+                    Some(Message::text(err_json.to_string()))
+                }
+            }
+        }
+        Err(error) => {
+            info!("WsError {error:?}");
+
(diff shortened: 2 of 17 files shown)


packages/transport-bluetooth/src/server/connection_handler.rs
```diff
@@ -0,0 +1,136 @@
...
+    if hyper_tungstenite::is_upgrade_request(&req) {
+        let (response, websocket) = match hyper_tungstenite::upgrade(req, None) {
+            Ok(r) => r,
+            Err(e) => {
+                info!("Failed to upgrade WebSocket: {e:?}");
+                return Ok(HyperResponse::new(Full::<Bytes>::from("Upgrade failed")));
+            }
+        };
+        tokio::spawn(async move { handle_ws_connection(peer, websocket).await });
+        Ok(response)
```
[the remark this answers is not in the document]
szymonlesisz: yes, it is commented as TODO in line 93

packages/transport-bluetooth/src/server/connection_handler.rs
```diff
@@ -0,0 +1,136 @@
...
+                }
+            }
+        }
+    });
+
+    // in current thread keep listening for incoming websocket messages
+    let ws_write_method = ws_write.clone();
+    while let Some(msg) = ws_read.next().await {
+        // TODO: panic thrown here when computer suspended?
+        let request = msg.unwrap_or(Message::text("{}"));
```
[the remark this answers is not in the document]
szymonlesisz: the message is not empty, it's a empty json string which it should fail at handle_message serialization

packages/transport-bluetooth/src/server/methods/get_info.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    match result {
+        Ok(info) => Ok(format!("{:?}", info)),
+        Err(error) => Err(error.into()),
+    }
+}
+
+#[cfg(target_os = "windows")]
+pub async fn get_adapter_info() -> Result<String, Box<dyn std::error::Error>> {
+    Ok("TODO: get_adapter_info windows.".to_string())
```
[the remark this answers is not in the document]
szymonlesisz: its ok i will add this logic later using windows crate

packages/transport-bluetooth/src/server/message_handler.rs
```diff
@@ -0,0 +1,88 @@
...
+        Message::Binary(msg) => {
+            let msg_str = msg.escape_ascii().to_string();
+            if msg_str == "PING" {
+                Some(msg_str.to_string())
+            } else {
+                None
+            }
+        }
+        Message::Ping(_msg) => None,
+        Message::Close(msg) => {
```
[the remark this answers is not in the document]
szymonlesisz: i think Tungstenite sends pong frame automatically,
anyway i've removed this case, it should never happen so it will fall down to unknown Message case

packages/transport-bluetooth/src/server/types.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(serde::Serialize, Clone, Debug)]
+pub enum ChannelMessage {
+    Abort(AbortProcess),
+    Notification(NotificationEvent),
+}
+
+#[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
+#[serde(tag = "method", content = "params", rename_all = "snake_case")]
+pub enum WsRequestMethod {
+    GetInfo(),
```
obrusvit: No need to use `()` if the enum variant does not contain any inner fields.
```
pub enum WsRequestMethod {
GetInfo,
StartScan,
StopScan,
}
```

packages/transport-bluetooth/src/server/message_handler.rs
```diff
@@ -0,0 +1,87 @@
...
+        msg => {
+            info!("unknown Message {msg:?}");
+            None
+        }
+    };
+
+    if msg.is_none() {
+        return None;
+    }
+    let msg = msg.unwrap();
```
obrusvit: Generally, plain `unwrap()` panics if the underlying variant is `None` so it is a code smell.

This particular case is fine because `msg`  is checked for `is_none()` above, but consider using these options:
1. `?` operator which works with `Option` and `Result`
```rust
let msg = msg?;  // from Option<T> to T and returns None from the function if None
```
2. `unwrap_or` so that you have some default case
3. using `let` or `if let` expression for safe unpacking
```rust
let Some(msg) = msg else {
    return None;
};
```


I'd recommend option 1 for now. It can be chained, e.g.
```rust
let p = msg?.find("P")?;  // returns None if either `msg` is `None` or "P" is not found within msg
```

packages/transport-bluetooth/src/server/message_handler.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    if msg.clone().to_string() == "PING" {
```
obrusvit: Unnecessary `clone().to_string()`

packages/transport-bluetooth/src/server/message_handler.rs
```diff
@@ -0,0 +1,87 @@
...
+        return Some(Message::text("PONG".to_string()));
```
obrusvit: Unnecessary `to_string()`

packages/transport-bluetooth/src/server/message_handler.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    info!("WsRequest: {:?}", msg.clone());
```
obrusvit: Unnecessary `clone()`

packages/transport-bluetooth/src/server/message_handler.rs
```diff
@@ -0,0 +1,87 @@
...
+    let json = serde_json::from_str::<WsRequest>(&msg.to_string());
```
obrusvit: unnecessary costly `to_string()`.

If a function takes `&str` (string slice), then passing just a reference to the `String`, i.e. `&msg` is fine.

packages/transport-bluetooth/src/server/message_handler.rs
```diff
@@ -0,0 +1,87 @@
...
+    if let Err(err) = json {
+        let json_error = serde_json::json!({
+            "id": "-1",
+            "error": format!("WsRequest serialization error: {err:?}"),
+        });
+
+        return Some(Message::text(json_error.to_string()));
+    }
+
+    let request = json.unwrap();
```
obrusvit: to prevent `unwrap()`, you can write this:

```rust
    let request = match serde_json::from_str::<WsRequest>(&msg) {
        Ok(req) => req,
        Err(err) => {
            let json_error = serde_json::json!({
                "id": "-1",
                "error": format!("WsRequest serialization error: {err:?}"),
            });
            return Some(Message::text(json_error.to_string()));
        }
    };

```

packages/transport-bluetooth/src/server/message_handler.rs
```diff
@@ -0,0 +1,87 @@
...
+    info!("Method: {request:?}");
+
+    let payload = match request.method.clone() {
+        WsRequestMethod::GetInfo() => methods::get_info().await,
+        _ => methods::noop().await, // TODO: implement methods
+    };
+
+    match payload {
```
obrusvit, line 55: This is more idiomatic version which does not have `unwrap()`.

Also for the `Err` case, I added default case. Even though, that one was relatively safe because `WsError` is a simple structure.

```rust
    match payload {
        Ok(payload) => {
            info!("WsResponse {payload:?}");
            let response = WsResponse {
                id: request.id.to_string(),
                payload,
            };

            match serde_json::to_string(&response) {
                Ok(json) => Some(Message::text(json)),
                Err(err) => {
                    let err_json = serde_json::json!({
                        "id": request.id,
                        "error": format!("WsResponse serialization error: {err:?}"),
                    });
                    Some(Message::text(err_json.to_string()))
                }
            }
        }
        Err(error) => {
            info!("WsError {error:?}");

            let ws_error = WsError {
                id: request.id.to_string(),
                error: error.to_string(),
            };
            // This unwrap is relatively safe as WsError is a simple structure
            let json = serde_json::to_string(&ws_error).unwrap_or_else(|_| {
                format!(
                    "{{\"id\":\"{}\",\"error\":\"Failed to serialize error\"}}",
                    request.id
                )
            });

            Some(Message::text(json))
        }
    }
```

packages/transport-bluetooth/src/server/connection_handler.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+    });
+
+    // in current thread keep listening for incoming websocket messages
+    let ws_write_method = ws_write.clone();
+    while let Some(msg) = ws_read.next().await {
+        // TODO: panic thrown here when computer suspended (macos?)
+        // fallback to empty json. it should fail in handle_message serialization
+        let request = msg.unwrap_or(Message::text("{}"));
+        let ws_write_method = ws_write_method.clone();
```
obrusvit: You should just use counted reference to the same object
```rust
        let ws_write_ref = Arc::clone(&ws_write_method);
```

packages/transport-bluetooth/src/server/connection_handler.rs
```diff
@@ -0,0 +1,143 @@
...
+        let sender = sender.clone();
+        // spawn new thread and process request
+        tokio::spawn(async move {
+            let response = handle_message(request.clone(), sender.clone()).await;
```
obrusvit: No need to `sender.clone()` here I think.

packages/transport-bluetooth/src/server/connection_handler.rs
```diff
@@ -0,0 +1,143 @@
...
+                    info!("No response request {request:?}");
+                }
+            }
+        });
+    }
+
+    // peer disconnected
+    channel_message_listener.abort();
+
+    if let Err(err) = sender.send(ChannelMessage::Abort(AbortProcess::ClientDisconnected)) {
```
obrusvit, line 96: The `Error` could be propagated maybe? Instead of just logged.

packages/transport-bluetooth/src/server/connection_handler.rs
```diff
@@ -0,0 +1,143 @@
...
+        // spawn new thread and process request
+        tokio::spawn(async move {
+            let response = handle_message(request.clone(), sender.clone()).await;
+            match response {
+                Some(response) => {
+                    let mut ws = ws_write_method.lock().await;
+                    if let Err(err) = ws.send(response).await {
+                        info!("Error response to ws {err:?}");
+                    };
+                    drop(ws);
```
obrusvit: I believe this `drop` is unnecessary as variable is dropped automatically at the end of the scope.

packages/transport-bluetooth/src/server/connection_handler.rs
```diff
@@ -0,0 +1,143 @@
...
+use std::sync::Arc;
+use tokio::net::TcpListener;
+use tokio::sync::broadcast;
+use tokio::sync::Mutex;
+
+use crate::server::handle_message;
+use crate::server::types::{AbortProcess, ChannelMessage};
+use crate::server::utils;
+
+type Error = Box<dyn std::error::Error + Send + Sync + 'static>;
```
obrusvit: Using this `Box`ed generic error is convenient but handling of those errors might get messy. 

If you think the possible error types through and use custom `enum` for the error type with conversions from the underlying errors using `thiserror` crate, then we might get less surprises. Something like this:

```rust
#[derive(Debug, thiserror::Error)]
pub enum ServerError {
    #[error("WebSocket error: {0}")]
    WebSocketError(#[from] tungstenite::Error),
    
    #[error("HTTP error: {0}")]
    HttpError(#[from] hyper::Error),
    
}

```

obrusvit (commented):
The WebSocket server is a smart idea, I left rust-specific comments.

Generally, I'd advise to be more conscious about using `clone()`, `to_string()`, `unwrap()`. Second big thing is the error handling.

packages/transport-bluetooth/src/server/message_handler.rs
szymonlesisz: isnt it covered by the block in line 33  `if msg.is_none()` ? if it's not `none` then i can safely unwrap it, no?

packages/transport-bluetooth/src/server/connection_handler.rs
szymonlesisz: there is no peer to propagate it to, it was disconnected (comment above)

packages/transport-bluetooth/src/server/connection_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/connection_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/connection_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/message_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/message_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/message_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/message_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/message_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/message_handler.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/types.rs
szymonlesisz: [f1cd2b5](https://github.com/trezor/trezor-suite/pull/18126/commits/f1cd2b5e36d8b4300b326b729f8e31fb1ec3ce04)

packages/transport-bluetooth/src/server/connection_handler.rs
szymonlesisz: [5a01160](https://github.com/trezor/trezor-suite/pull/18126/commits/5a011607fa7cd7b947932ab3e1810742348e07cf)

i've also added IoError case to satisfy error from 
```
tcp_listener.accept().await?
'``

File: packages/transport-bluetooth/src/server/message_handler.rs (hunk 6 above)
Comment (obrusvit) [reply packages/transport-bluetooth/src/server/message_handler.rs]: Yes, it is safe to `unwrap()` because you checked for `is_none()` but it is better to do it in one expression. 

Your code:
```rust
    if msg.is_none() {
        return None;
    }
    let msg = msg.unwrap();
```
is identical to this:
```rust
   msg = msg?;
``` 

In general, whenever you see plain `unwrap()`, you must be very sure that the value is there because `None.unwrap()` panics and crashes the whole program. That's why it's better to use safer alternative, e.g. `None.unwrap_or(defaul_msg)`

packages/transport-bluetooth/src/server/connection_handler.rs
obrusvit: The benefit is that while handling the `Result`, you can now react to each Error type separatedly.

You can maybe expand the usage of `handle_ws_connection` within the function `handle_http_request` like this:
```rust
tokio::spawn(async move {
    if let Err(err) = handle_ws_connection(peer.clone(), websocket).await {
        log::error!("WebSocket connection error for peer {}: {:?}", peer, err);
        
        match err {
            ServerError::WebSocketError(ws_err) => {
                log::debug!("WebSocket protocol error: {}", ws_err);
            },
            ServerError::IoError(io_err) => {
                log::warn!("IO error in WebSocket connection: {}", io_err);
            },
            ServerError::HttpError(http_err) => {
                log::warn!("HTTP error in WebSocket connection: {}", http_err);
            }
        
    }
});
```

packages/transport-bluetooth/src/server/types.rs
obrusvit: I cannot resolve these conversations so feel free to close them.

packages/transport-bluetooth/src/server/message_handler.rs
szymonlesisz: so maybe it coud be even shortened to this, no?
```
let msg: String = match message {
     .....
}?;
```

szymonlesisz: like this [27c82ab](https://github.com/trezor/trezor-suite/pull/18126/commits/27c82ab1914be1edc70e76f31d308f0342d9bdc4)

packages/transport-bluetooth/src/server/connection_handler.rs
szymonlesisz: cool, thanks, done in [27c82ab](https://github.com/trezor/trezor-suite/pull/18126/commits/27c82ab1914be1edc70e76f31d308f0342d9bdc4)

mroz22 (approved):

PR #18131, opened 2025-04-03
Polish colors after palette update

matejkriz:
## Description

Only a few subtle fixes in light mode done together with @shenkys. We haven't checked dark mode yet.

- [fix elevations on border and background colors](https://github.com/trezor/trezor-suite/commit/2d3fea9018d71b2ed55a572e972e4cc82ef033fa)
- [fix background color of FW update progress circle](https://github.com/trezor/trezor-suite/commit/c2177f6f37fa4e98fd63a8ec0374f79aff7e41de)

## Screenshots:
image
image

image

### DIFF
--- a/suite-native/atoms/src/IconListItem.tsx
+++ b/suite-native/atoms/src/IconListItem.tsx
@@ -19,7 +19,7 @@ type IconColors = {
 const iconColorsMap = {
     default: {
         iconColor: 'iconDefault',
-        iconBorderColor: 'borderElevation1',
+        iconBorderColor: 'borderElevation0',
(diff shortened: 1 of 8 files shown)


PeKne (approved):

matejkriz:
@STew790 This is so subtle changes, that I guess you can move it directly to approved... But on the other hand, you can help testing the colours globally and point out any places where it looks really broken from your point of view (invisible elements, low contrast...)

STew790:
**QA OK**
Approving but will definitely monitor this with other QA and create separate issues for what we find.

PR #18133, opened 2025-04-03
refactor(trading): exchange load info thunk

### LINKED ISSUE #17733: Duplicate swap context functions in suite-common and add more tests

adderpositive:
## Description
- created thunk for loading exchange info

## Related Issue
Resolve https://github.com/trezor/trezor-suite/issues/17733

### DIFF
--- a/suite-common/trading/src/thunks/index.ts
+++ b/suite-common/trading/src/thunks/index.ts
@@ -1,12 +1,13 @@
 import { confirmTradeThunk as confirmBuyTradeThunk } from './buy/confirmTradeThunk';
 import { handleRequestThunk as handleBuyRequestThunk } from './buy/handleRequestThunk';
-import { loadInfoThunk } from './buy/loadInfoThunk';
+import { loadInfoThunk as loadBuyInfoThunk } from './buy/loadInfoThunk';
 import { selectQuoteThunk as selectBuyQuoteThunk } from './buy/selectQuoteThunk';
 import { loadInitialDataThunk } from './common/loadInitialDataThunk';
 import { recomposeAndSignTxThunk } from './common/recomposeAndSignTxThunk';
 import { verifyAddressThunk } from './common/verifyAddressThunk';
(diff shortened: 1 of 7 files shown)


suite-common/trading/src/thunks/index.ts
```diff
@@ -1,12 +1,13 @@
 import { confirmTradeThunk as confirmBuyTradeThunk } from './buy/confirmTradeThunk';
 import { handleRequestThunk as handleBuyRequestThunk } from './buy/handleRequestThunk';
-import { loadInfoThunk } from './buy/loadInfoThunk';
+import { loadInfoThunk as loadBuyInfoThunk } from './buy/loadInfoThunk';
```
izmy, line 3: Consider keeping the original import names for consistency. Using aliases (`loadBuyInfoThunk`, `loadExchangeInfoThunk`) makes it harder to track the original function names and creates an inconsistency with `buyThunks` and `exchangeThunks`, where `loadInfoThunk` is still used.

adderpositive: I get your point. I assumed `buyThunks` and `exchangeThunks` should be used in the following code; this is only the file used to export them. 

Do you think I should reverse it a little bit? 

Thunk name would be `loadBuyInfoThunk` and then export

```ts
import { loadBuyInfoThunk } from './buy/loadInfoThunk';

export const buyThunks = {
  loadInfoThunk: loadBuyInfoThunk
}
```

This is what I don't want - using more `buy` prefixes than necessary.
```ts
export const buyThunks = {
  loadBuyInfoThunk
}
```

TomasBoda (approved):
LGTM

suite-common/trading/src/thunks/index.ts
izmy: Yes, renaming it this way seems better to me

adderpositive: I have created follow-up issue to this - https://github.com/trezor/trezor-suite/issues/18212

LukasRada (approved):

izmy (approved):

PR #18138, opened 2025-04-04
chore(connect-ui): implement new async getOsVersion in analytics

Lemonexe:
## Description

Followup to https://github.com/trezor/trezor-suite/pull/18023:
use the new async `getOsVersion` in connect-ui analytics, so that we can distinguish macOS > catalina (currently targetted as if it was catalina), and Windows 11 (currently targetted as if it was 10).

@coderabbitai ignore

### DIFF
--- a/packages/connect-ui/src/utils/analytics.ts
+++ b/packages/connect-ui/src/utils/analytics.ts
@@ -3,8 +3,8 @@ import { storage } from '@trezor/connect-common';
 import {
     getBrowserName,
     getBrowserVersion,
-    getDeprecatedOsVersion,
     getOsName,
+    getOsVersion,
     getPlatformLanguages,
     getScreenHeight,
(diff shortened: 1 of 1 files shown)


packages/connect-ui/src/utils/analytics.ts
```diff
@@ -39,7 +39,7 @@ const onDisable = () => {
     );
 };
 
-export const initAnalytics = () => {
+export const initAnalytics = async () => {
```
Lemonexe, line 42: this fn is called as side-effect [here](https://github.com/trezor/trezor-suite/blob/a598bc7dfb6345978e9b9dd9aacd556a591455d3/packages/connect-ui/src/index.tsx#L76), nothing seems to depend on it running sync, so LGTM

mroz22 (approved):
ok

PR #18155, opened 2025-04-04
chore(suite-native): Bump RN to 0.76.9, @sentry/react-native to 6.10

jbazant:
Resolves problems of iOS compilation for Xcode 16.3
## Description

See:
- https://expo.dev/changelog/xcode-16-3-patches
- https://github.com/expo/expo/issues/35807#issuecomment-2770621907
- https://github.com/expo/expo/pull/35811
- https://github.com/getsentry/sentry-react-native/issues/4715

## Related Issue

Resolve


BrantalikP (changes requested):
Have not run into any issues while trying both iOS and Android - good job! 🚀

Regarding the TS issues

1.  It helps to explicitly type the return value in https://github.com/trezor/trezor-suite/blob/2490c03cf4714500158ccf32f56d80f316312d43/packages/react-native-usb/src/index.ts#L113 with `EventSubscription` from `import { EventSubscription } from 'expo-modules-core';`
at

 ```
 export function onDeviceConnected(listener: (event: OnConnectEvent) => void): EventSubscription
 
 ```
and 

 ```
 export function onDeviceDisconnect(listener: (event: OnConnectEvent) => void): EventSubscription 
 ```
 
 2. You might also need to update [@sentry/core ](https://github.com/trezor/trezor-suite/blob/2490c03cf4714500158ccf32f56d80f316312d43/suite-native/app/package.json#L34) to "8.54.0" 
 
 Hope it helps!

jbazant:
@BrantalikP Missed that one, thank you for pointing it out! Bumped `@sentry/core` to 8.55.0 everywhere, seems it solved all type issues.

Edit: Are results cached? Locally it works. No idea whats the problem here in pipeline.

BrantalikP:
> @BrantalikP Missed that one, thank you for pointing it out! Bumped `@sentry/core` to 8.55.0 everywhere, seems it solved all type issues.
> 
> Edit: Are results cached? Locally it works. No idea whats the problem here in pipeline.

Not sure, but what if you run `yarn type-check:force` from root? I still get the issue locally. 


```
    ✖  nx run @trezor/react-native-usb:type-check
       src/index.ts(113,17): error TS2742: The inferred type of 'onDeviceConnected' cannot be named without a reference to 'expo/node_modules/expo-modules-core'. This is likely not portable. A type annotation is necessary.
       src/index.ts(138,17): error TS2742: The inferred type of 'onDeviceDisconnect' cannot be named without a reference to 'expo/node_modules/expo-modules-core'. This is likely not portable. A type annotation is necessary.
       
```
       
in develop its ok 🤔

BrantalikP (approved):
:shipit:

packages/react-native-usb/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import { EventSubscription } from 'expo-modules-core/src/ts-declarations/EventEmitter';
```
BrantalikP: nit: could be just `import { EventSubscription } from 'expo-modules-core';`

jbazant: And that is what happen when you let IDEA on it's own. 🤦 

Fixed.

matejkriz:
@SocketSecurity ignore-all

matejkriz:
Fixed by `yarn run deps && yarn run native:prebuild:clean`
~~I'm trying `yarn ios` after `yarn install` and `yarn p` and getting following error:~~

```
⚠️  Something went wrong running `pod install` in the `ios` directory.
Command `pod install` failed.
└─ Cause: Invalid `Podfile` file: cannot load such file -- ./scripts/setup.
...
>  require File.join(File.dirname(`node --print "require.resolve('react-native-permissions/package.json')"`), "scripts/setup")
```

matejkriz (approved):
I'm finally able to run it on iOS again 🎉

PR #18158, opened 2025-04-04
chore(suite-native): changed btc to sats conversion

vytick:
Correct method for conversion btc to sats in buy form


jbazant (approved):
LGTM

PR #18169, opened 2025-04-07
Connect in desktop improvements

mroz22:
## Related Issue

Resolve

### DIFF
--- a/packages/connect/src/data/connectSettings.ts
+++ b/packages/connect/src/data/connectSettings.ts
@@ -36,6 +36,9 @@ const parseManifest = (manifest?: Manifest) => {
     if (!manifest) return;
     if (typeof manifest.email !== 'string') return;
     if (typeof manifest.appUrl !== 'string') return;
+    // todo [connect10]: appName should become required
+    if (typeof manifest.appName !== 'undefined' && typeof manifest.appName !== 'string') return;
+    if (typeof manifest.appIcon !== 'undefined' && typeof manifest.appIcon !== 'string') return;
 
     return {
         email: manifest.email,
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConnectPopupModal.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConnectPopupModal.tsx
@@ -17,18 +17,26 @@ export const ConnectPopupModal = () => {
 
     const { methodInfo, source } = popupCall;
     const { methodTitle, confirmLabel, permissionTypes } = methodInfo;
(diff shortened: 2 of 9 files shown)


packages/connect/src/data/connectSettings.ts
```diff
@@ -36,6 +36,9 @@ const parseManifest = (manifest?: Manifest) => {
     if (!manifest) return;
     if (typeof manifest.email !== 'string') return;
     if (typeof manifest.appUrl !== 'string') return;
+    // todo [connect10]: appName should become required
+    if (typeof manifest.appName !== 'undefined' && typeof manifest.appUrl !== 'string') return;
```
martykan: Bit strange condition, it will never happen since `typeof manifest.appUrl !== 'string'` is handled above

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConnectPopupModal.tsx
```diff
@@ -17,18 +17,26 @@ export const ConnectPopupModal = () => {
...
     const { methodInfo, source } = popupCall;
     const { methodTitle, confirmLabel, permissionTypes } = methodInfo;
-    const { processName, origin, manifest } = source;
+
+    const rememberPayload = {
+        origin,
+        types: permissionTypes,
+        ...(source.__walletConnect
+            ? {
+                  __walletConnect: true as const,
```
martykan: Why use underscores here?

mroz22: to signal that this is not some kind of real value but a mere branding field, I thought there was some kind of convention like this? 🤷

packages/connect/src/data/connectSettings.ts
mroz22: its a bug, should be 
```
if (typeof manifest.appName !== 'undefined' && typeof manifest.appName !== 'string') return;
```

mroz22: 391b09103212951e50269373803d6abd39c3214a

packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ConnectPopupModal.tsx
martykan: I would say it is useful as a real value

mroz22: ok a04e067015ba719343871062c1feded522692234

martykan (approved):

Issue #17959, opened 2025-03-28
THP: Multi-device


peter-sanderson:
- [ ] Welcome screen (2+ devices that requires THP pairing)
- [ ] Wallet switcher (1 device connected, 1+ devices that requires THP pairing)

rabbitholiness:
What do we know about the devices? Model? Name?

peter-sanderson:
> What do we know about the devices? Model? Name?

@rabbitholiness depends, but we shall always know at least the color-variant and the model

rabbitholiness:
Here's the [link to Figma](https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Device-Connection?node-id=7788-68241&t=NVsuq0ZcPLhWNpr3-1)

peter-sanderson:
https://github.com/trezor/trezor-suite/commit/6f335cf0f9d96b7fdf9b023d6da4e900b0120484

bosomt:
**QA OK**

- welcome screen works as expected
- device switcher is present
- im able to onboard device 
- install of FW is not allowed when 2 devices connected as expected
- im able to create THP connection for both devices
- tested on T2T1 + T3W1 with THP FW



**Info:**
 - Suite version: desktop 25.6.0 (981bf56dfcf63d14b8001e695389d3e31951d28c)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.6.0 Chrome/136.0.7103.49 Electron/36.2.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: 
 - Transport: BridgeTransport 3.1.0-bundled.25.6.0

PR #18178, opened 2025-04-07
feat: refactoring of the switch device logic, strictier types

peter-sanderson:
Part of https://github.com/trezor/trezor-suite/issues/17959

- adds some stricter types
- enabled direct CTA into onboarding

image
image

### DIFF
--- a/suite-common/suite-utils/src/device.ts
+++ b/suite-common/suite-utils/src/device.ts
@@ -85,8 +85,6 @@ export const isDevicePerceivedAsNew = (device: TrezorDevice | null | undefined) 
 
 export const deviceNeedsAttention = (deviceStatus: ConnectedDeviceStatus) => {
     switch (deviceStatus) {
-        // case 'firmware-recommended':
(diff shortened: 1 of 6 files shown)


suite-common/suite-utils/src/device.ts
```diff
@@ -115,58 +123,6 @@ export const shouldDisplayInitialWarningIcon = (deviceStatus: ConnectedDeviceSta
     }
 };
 
-export const getDeviceStatusWarningVariant = (
```
peter-sanderson, line 118: This is used only once at one place only.

suite-common/suite-utils/src/device.ts
```diff
@@ -96,8 +94,18 @@ export const deviceNeedsAttention = (deviceStatus: ConnectedDeviceStatus) => {
         case 'firmware-required':
         case 'unreadable':
             return true;
-        default:
```
Lemonexe, line 99: :clap:

Lemonexe (approved):
LGTM!

PR #18179, opened 2025-04-07
refactor(trading): watch trade

### LINKED ISSUE #18180: Move watch trade

adderpositive:
## Description

- moved buy watch trade

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18180

### DIFF
--- a/suite-common/trading/src/thunks/common/__tests__/watchTradeThunk.test.ts
+++ b/suite-common/trading/src/thunks/common/__tests__/watchTradeThunk.test.ts
@@ -1,0 +1,247 @@
+import { combineReducers } from '@reduxjs/toolkit';
+
+import { configureMockStore, extraDependenciesMock } from '@suite-common/test-utils';
+import { Account } from '@suite-common/wallet-types';
+
+import { accountBtc } from '../../../__fixtures__/utils';
+import { invityAPI } from '../../../invityAPI';
+import {
+    TradingState,
+    initialState,
+    prepareTradingReducer,
+} from '../../../reducers/tradingReducer';
+import {
+    TradingTransactionBuy,
+    TradingTransactionExchange,
+    TradingTransactionSell,
+} from '../../../types';
+import { watchTradeThunk } from '../watchTradeThunk';
+
+describe('watchTradeThunk', () => {
+    jest.mock('../../../invityAPI');
+
+    const tradingReducer = prepareTradingReducer(extraDependenciesMock);
+    const account = accountBtc as Account;
+    const accountData = {
+        symbol: account.symbol,
+        descriptor: account.descriptor,
+        accountIndex: account.index,
+        accountType: account.accountType,
+    };
+    const refreshCount = 1;
+
+    const getStore = (updatedState: Partial<TradingState>) =>
+        configureMockStore({
+            extra: {},
+            reducer: combineReducers({
+                wallet: combineReducers({
+                    tradingNew: tradingReducer,
+                }),
+            }),
+            preloadedState: {
+                wallet: {
+                    tradingNew: {
+                        ...initialState,
+                        ...updatedState,
+                    },
+                },
+            },
+        });
+
+    const date = new Date('2025-04-09');
+    const dateISO = date.toISOString();
+
+    beforeAll(() => {
+        jest.useFakeTimers();
+        jest.setSystemTime(date);
+    });
+
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+
+    it('should remain in the same status when there is not any response', async () => {
+        const trade = {
+            date: dateISO,
+            key: 'tradeKey',
+            tradeType: 'buy',
+            data: {
+                status: 'LOGIN_REQUEST',
+                paymentId: 'tradeKey',
+            },
+            account: accountData,
+        } as TradingTransactionBuy;
+
+        const store = getStore({
+            trades: [trade],
+        });
+
+        invityAPI.watchTrade = () => Promise.resolve(undefined as any);
+
+        await store.dispatch(
+            watchTradeThunk({
+                account,
+                trade,
+                refreshCount,
+            }),
+        );
+
+        const actions = store.getActions();
+        const saveTradeAction = actions.find(action => action.type === '@trading/saveTrade');
+
+        expect(saveTradeAction).toBeUndefined();
+    });
+
+    it('should remain in the status when there is same the status in response', async () => {
+        const trade = {
+            date: dateISO,
+            key: 'tradeKey',
+            tradeType: 'buy',
+            data: {
+                status: 'LOGIN_REQUEST',
+                paymentId: 'tradeKey',
+            },
+            account: accountData,
+        } as TradingTransactionBuy;
+
+        const store = getStore({
+            trades: [trade],
+        });
+
+        invityAPI.watchTrade = () =>
+            Promise.resolve({
+                status: 'LOGIN_REQUEST',
+            } as any);
+
+        await store.dispatch(
+            watchTradeThunk({
+                account,
+                trade,
+                refreshCount,
+            }),
+        );
+
+        const actions = store.getActions();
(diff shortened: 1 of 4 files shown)


suite-common/trading/src/thunks/common/__tests__/watchTradeThunk.test.ts
```diff
@@ -0,0 +1,244 @@
...
+    TradingTransactionExchange,
+    TradingTransactionSell,
+} from '../../../types';
+import { watchTradeThunk } from '../watchTradeThunk';
+
+describe('watchTradeThunk', () => {
+    jest.mock('../../../invityAPI');
+
+    invityAPI.setInvityServersEnvironment = () => {};
+    invityAPI.createInvityAPIKey = () => {};
```
jbazant: **issue:** `invityAPI` is already mocked (line 21), this is unnecessary.

suite-common/trading/src/thunks/common/__tests__/watchTradeThunk.test.ts
```diff
@@ -0,0 +1,244 @@
...
+                    },
+                },
+            },
+        });
+
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+
+    it('status should stay same when there is not any response', async () => {
```
jbazant: **nitpick (non-blocking):** Wording

suite-common/trading/src/thunks/common/__tests__/watchTradeThunk.test.ts
```diff
@@ -0,0 +1,244 @@
...
+    const account = accountBtc as Account;
+    const accountData = {
+        symbol: account.symbol,
+        descriptor: account.descriptor,
+        accountIndex: account.index,
+        accountType: account.accountType,
+    };
+    const refreshCount = 1;
+    const dateString = new Date().toISOString();
+    jest.spyOn(Date.prototype, 'toISOString').mockImplementation(() => dateString);
```
jbazant: **issue:** This can be achieved in a bit more conventional manner. See [Timer mocks](https://jestjs.io/docs/timer-mocks)

```ts
    const dateString = 'THE MOST IMPORTANT DATE BY YOUR CHOICE';

    beforeAll(() => {
        jest.useFakeTimers();
        jest.setSystemTime(new Date(dateString));
    });

    afterAll(() => {
        jest.useRealTimers();
    });

File: suite-common/trading/src/thunks/common/watchTradeThunk.ts [hunk 4]
```diff
@@ -0,0 +1,137 @@
...
+
+        const accountData = {
+            descriptor: account.descriptor,
+            symbol: account.symbol,
+            accountType: account.accountType,
+            accountIndex: account.index,
+        };
+        const date = new Date().toISOString();
+
+        if (trade.tradeType === 'sell') return;
```
jbazant: **suggestion (non-blocking):** Can we use exhaustive switch here instead?

```ts
        switch (trade.tradeType) {
            case 'sell':
            case 'exhange':
              return;

            case 'buy':
                // ...
                return;

            default:
                throw new UnreachableCaseError(trade.tradeType, 'Unexpected trade type');
        }

**Review by jbazant — COMMENTED** [review]
LGTM. I have a hard time to understand what is this good for as there is no proper assignment.

There are some code quality comments, nothing critical. But I would prefer if anyone more experienced with desktop and it's flow will at this PR as well.

File: suite-common/trading/src/thunks/common/watchTradeThunk.ts [hunk 5]
```diff
@@ -0,0 +1,137 @@
...
+
+const watchTradeData = async <T extends TradingType>({
+    trade,
+    refreshCount,
+}: WatchTradeDataProps): Promise<WatchTradeDataResultProps<T> | undefined> => {
+    const response = await invityAPI.watchTrade<T>(trade.data, trade.tradeType, refreshCount);
+
+    if (!response) return;
+
+    const isStatusSame = response.status && response.status === trade.data.status;
```
vytick: The condition might be simplified to this, maybe?
```suggestion
if (response?.status === trade.data.status) {
	return
}
```

suite-common/trading/src/thunks/common/watchTradeThunk.ts
```diff
@@ -0,0 +1,137 @@
...
+            descriptor: account.descriptor,
+            symbol: account.symbol,
+            accountType: account.accountType,
+            accountIndex: account.index,
+        };
+        const date = new Date().toISOString();
+
+        if (trade.tradeType === 'sell') return;
+        /*  TODO: trading - sell watchTrade
+            const data = await watchTradeData<typeof trade.tradeType>({
```
vytick: Is having this whole code commented out here valuable? I would suggest to keep the TODO notice about the functionality being handled once this is moved to common.

it makes it harder to read now

suite-common/trading/src/thunks/common/watchTradeThunk.ts
```diff
@@ -0,0 +1,93 @@
...
+                        account: accountData,
+                        data: data.tradeData,
+                    }),
+                );
+
+                return;
+            }
+            case 'sell':
+            case 'exchange':
+                break;
```
vytick, line 89: Much better readability, thank you 🙏 . Can you, please, add comment that sell and exchange should be done later?

suite-common/trading/src/thunks/common/__tests__/watchTradeThunk.test.ts
```diff
@@ -0,0 +1,279 @@
...
+            watchTradeThunk({
+                account,
+                trade,
+                refreshCount,
+            }),
+        );
+
+        if ('error' in watchTrade && watchTrade.error instanceof UnreachableCaseError) {
+            expect(watchTrade.error.message).toBe('Unexpected trade type: ["gift"]');
+        }
```
jbazant: **question:** Not sure whether I get it correctly, but to me it seems that this condition is never truthy, because `watchTrade.error` is not an instance of `UnreachableCaseError`.

Also I would consider this as a useless test as `UnreachableCaseError` should throw TS error as soon as it becomes reachable (someone extends tradeType by new value).

And also I would use something like this to test it.
```ts
        const { error } = watchTrade as any;
        expect(error).toBeInstanceOf(UnreachableCaseError);
        expect(error.message).toBe('Unexpected trade type: ["gift"]');
```

jbazant (approved):
LGTM! Just one question.

vytick (approved):

suite-common/trading/src/thunks/common/__tests__/watchTradeThunk.test.ts
adderpositive: Yes, you are right. It is not working as I expected, it is also a useless test. 🫣

adderpositive: Deleted and added `/* istanbul ignore next */` to skip coverage for this specific case. https://github.com/trezor/trezor-suite/pull/18179/commits/5b01f457dad69cffbfff21f93884e38770ab29d4

suite-common/trading/src/thunks/common/__tests__/watchTradeThunk.test.ts
```diff
@@ -0,0 +1,247 @@
...
+            }),
+        );
+
+        const actions = store.getActions();
+        const saveTradeAction = actions.find(action => action.type === '@trading/saveTrade');
+
+        expect(saveTradeAction).toBeUndefined();
+    });
+
+    it('status should stay same when status is the same', async () => {
```
LukasRada: Please consider to rename it to something like this `should remain in the same status when there is same status in response` or better :)

LukasRada (changes requested):

LukasRada (approved):

vytick (approved):
🚀

PR #18189, opened 2025-04-07
fix(suite-native): fix Requiring unknown module undefined after wrong yarn.lock merge

matejkriz:
## Description

After "fixing" the issue by https://github.com/trezor/trezor-suite/pull/18119 it was broken again by https://github.com/trezor/trezor-suite/pull/18058. My theory is, that @mroz22 didn't rebase properly the yarn.lock but of course it's not his fault but the problem is the fragile dependency tree. I don't know how to prevent this issue properly. Maybe ban dedupe?

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -15392,14 +15392,15 @@ __metadata:
   languageName: node
   linkType: hard
 
-"asn1.js@npm:^4.10.1":
-  version: 4.10.1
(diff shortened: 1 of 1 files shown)


yanascz (approved):

mroz22:
> didn't rebase properly the yarn.lock

how do you actually rebase yarn.lock guys?

Issue #11676, opened 2024-03-21
Force auto-update option


MiroslavProchazka:
Investigate if we can force Suite to auto-update even when updaterId is not on the current StagingPercentage=xy pool. 

You can force the auto-update via adjusting the _~/Library/Application\ Support/@trezor/suite-desktop/.updaterId_ file from 
`xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` to `xxxxxxxx-xxxx-xxxx-xxxx-00000000`

The idea is to add button "Force update" in the settings that will adjust this file and force Suite to auto-update to the latest release on Github. 

The file will be probably hidden in different folders throughout operating systems, so we would need to adjust the prompt accordingly. 

I would recommend putting it behind Debug mode for the initial phase as well.

matejkriz:
check this for a reference how it works https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/AppUpdater.ts#L350

MiroslavProchazka:
Please ensure that a condition is included to prevent offering the update if the StagingPercentage is reduced to 0 — for example, in cases where an issue occurs during the release and we had to to limit the rollout.

komret:
I couldn't come up with a reasonable way how to implement this, everything is hacky. Consider contributing to [electron-builder](https://github.com/electron-userland/electron-builder) to enable support.

MiroslavProchazka:
Ok, closing for now.

Lemonexe:
@komret  Took a look at electron-builder codebase, as their documentation is not very detailed (only covers setting the percentage, not overriding it). At quick glance it doesn't seem too difficult to add it:
- [this is the public method](https://github.com/electron-userland/electron-builder/blob/065c6a456e34e7f8c13cba483d433502b9325168/packages/electron-updater/src/AppUpdater.ts#L301) which we call [from our auto-updater module](https://github.com/trezor/trezor-suite/blob/c2aec5c0f458c469a8c4efeb83942416e5183e01/packages/suite-desktop-core/src/modules/auto-updater.ts#L211)
  - this method could add option `{force?: boolean}`
- pass it to `doCheckForUpdates`
- pass it to `isUpdateAvailable`
- and [here](https://github.com/electron-userland/electron-builder/blob/065c6a456e34e7f8c13cba483d433502b9325168/packages/electron-updater/src/AppUpdater.ts#L419) add `if (forced) return true  `
[...]

PR #18190, opened 2025-04-07
chore(suite-desktop): desktop update TS cleanup + docs

Lemonexe:
## Description

- :broom: cleanup in `DesktopApi` TS and fix a _future bug_ caused by it
- :books: describe how to easily develop app update locally

## Related Issue

Side-effect of work on #11676

@coderabbitai ignore

### DIFF
--- a/packages/suite-desktop-api/src/__tests__/api.test.ts
+++ b/packages/suite-desktop-api/src/__tests__/api.test.ts
@@ -89,14 +89,10 @@ describe('DesktopApi', () => {
 
         it('DesktopApi.checkForUpdates', () => {
             const spy = jest.spyOn(ipcRenderer, 'send');
-            api.checkForUpdates();
-            expect(spy).toHaveBeenCalledWith('update/check', undefined);
-            api.checkForUpdates(true);
-            expect(spy).toHaveBeenCalledWith('update/check', true);
-
-            // @ts-expect-error invalid param
-            api.checkForUpdates('string');
-            expect(spy).toHaveBeenCalledTimes(2); // invalid param not processed
+            api.checkForUpdates({ isManual: false });
+            expect(spy).toHaveBeenCalledWith('update/check', { isManual: false });
+            api.checkForUpdates({ isManual: true });
+            expect(spy).toHaveBeenCalledWith('update/check', { isManual: true });
         });
 
         it('DesktopApi.downloadUpdate', () => {
--- a/packages/suite-desktop-api/src/api.ts
+++ b/packages/suite-desktop-api/src/api.ts
@@ -35,7 +35,7 @@ export interface MainChannels {
     'update/set-automatic-update-enabled': boolean;
     'update/set-auto-install-on-app-quit': void;
     'update/cancel': void;
-    'update/check': boolean | undefined;
+    'update/check': { isManual: boolean };
     'update/download': void;
     'update/install': void;
     'logger/config': LoggerConfig;
--- a/packages/suite-desktop-api/src/factory.ts
+++ b/packages/suite-desktop-api/src/factory.ts
@@ -51,9 +51,9 @@ export const factory = <R extends StrictIpcRenderer<any, IpcRendererEvent>>(
         getAppAutoStartIsEnabled: () => ipcRenderer.invoke('app/auto-start/is-enabled'),
 
         // Auto-updater
-        checkForUpdates: isManual => {
-            if (validation.isPrimitive(['boolean', true], isManual))
-                ipcRenderer.send('update/check', isManual);
+        checkForUpdates: ({ isManual }) => {
+            if (validation.isPrimitive('boolean', isManual))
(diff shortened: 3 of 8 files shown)


packages/suite-desktop-ui/src/support/DesktopUpdater.tsx
```diff
@@ -54,11 +54,11 @@ export const DesktopUpdater = ({ children }: DesktopUpdaterProps) => {
         desktopApi.on('update/error', params => dispatch(error(params)));
 
         // Initial check for updates
-        desktopApi.checkForUpdates();
+        desktopApi.checkForUpdates(false);
         // Check for updates every hour
         const checkForUpdatesInterval = setInterval(
             () => {
-                desktopApi.checkForUpdates();
+                desktopApi.checkForUpdates(false);
```
Lemonexe: because of the default `true` in `packages/suite-desktop-api/src/factory.ts`, the automatic update was actually labeled as manual.

This is does not _yet_ cause any problems in develop, but it does in my WIP PR where I use the `isManualCheck` param.

packages/suite-desktop-core/src/modules/auto-updater.ts
```diff
@@ -63,6 +63,8 @@ export const init: ModuleInit = ({ mainWindowProxy, store }) => {
 
     // Prevent downloading an update unless user explicitly asks for it.
     autoUpdater.autoDownload = false;
+    // You may turn this on for dev purposes, see docs/releases/desktop_updates.md
+    autoUpdater.forceDevUpdateConfig = false;
```
Lemonexe, line 67: `false` is [ofc default](https://github.com/electron-userland/electron-builder/blob/106640dd42a3db08bfbe3a3a32fe333e93ba5c10/packages/electron-updater/src/AppUpdater.ts#L116-L118), I'm adding it here for easier orientation when you want to set it `true`

packages/suite-desktop-ui/src/support/DesktopUpdater.tsx
peter-sanderson: nit but maybe not nit: If possible lets change signature of the method to named props: `{ isManual: ... }` and lets make `isManual` required.

packages/suite-desktop-api/src/factory.ts
```diff
@@ -52,7 +52,7 @@ export const factory = <R extends StrictIpcRenderer<any, IpcRendererEvent>>(
 
         // Auto-updater
         checkForUpdates: isManual => {
-            if (validation.isPrimitive(['boolean', true], isManual))
+            if (validation.isPrimitive('boolean', isManual))
```
peter-sanderson, line 55: This syntax is WTF :D 
(I mean the array where second arg means optionality)

packages/suite-desktop-api/src/api.ts
```diff
@@ -35,7 +35,7 @@ export interface MainChannels {
     'update/set-automatic-update-enabled': boolean;
     'update/set-auto-install-on-app-quit': void;
     'update/cancel': void;
-    'update/check': boolean | undefined;
+    'update/check': boolean;
```
peter-sanderson: :clap:

packages/suite-desktop-api/src/__tests__/api.test.ts
```diff
@@ -89,13 +89,15 @@ describe('DesktopApi', () => {
...
-            expect(spy).toHaveBeenCalledWith('update/check', undefined);
+            api.checkForUpdates(false);
+            expect(spy).toHaveBeenCalledWith('update/check', false);
             api.checkForUpdates(true);
             expect(spy).toHaveBeenCalledWith('update/check', true);
 
             // @ts-expect-error invalid param
             api.checkForUpdates('string');
+            // @ts-expect-error invalid param
+            api.checkForUpdates(undefined);
```
peter-sanderson: not needed I think we dont need to test ts

peter-sanderson (approved):

packages/suite-desktop-core/src/modules/auto-updater.ts
```diff
@@ -209,8 +209,8 @@ export const init: ModuleInit = ({ mainWindowProxy, store }) => {
         );
     });
 
-    ipcMain.on('update/check', (_, isManual) => {
-        if (isManual) {
+    ipcMain.on('update/check', (_, { isManual }) => {
+        if (isManual === true) {
```
Lemonexe, line 213: @peter-sanderson absolutely agree with you that explicit vaulue checks are a must. 

if I didn't remember this condition was here, just because I worked here an hour ago, there would have been a new future bug (object always truthy)

peter-sanderson (approved):

Lemonexe:
Failing test is web only, [confirmed to be broken and fix is on its way](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1744096117384669?thread_ts=1743572836.379729&cid=CKYFBT2C9)

PR #18202, opened 2025-04-08
feat(suite): select account modal

martykan:
## Description

Suite modal for Connect's `UI.SELECT_ACCOUNT`. The current flow is unchanged, this is basicailly a 1:1 compatible replacement. One bonus is that account labeling should work.

## Screenshots:

Screenshot 2025-04-09 at 09 35 03

### DIFF
--- a/packages/suite/src/actions/suite/modalActions.ts
+++ b/packages/suite/src/actions/suite/modalActions.ts
@@ -40,6 +40,18 @@ export const onReceiveConfirmation = (confirmation: boolean) => (dispatch: Dispa
 
     dispatch(onCancel());
 };
+export const onReceiveAccount = (accountIndex: number | null) => (dispatch: Dispatch) => {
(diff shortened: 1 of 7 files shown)


packages/suite/src/actions/suite/modalActions.ts
```diff
@@ -40,6 +40,18 @@ export const onReceiveConfirmation = (confirmation: boolean) => (dispatch: Dispa
 
     dispatch(onCancel());
 };
+export const onReceiveAccount = (accountIndex: number | null) => (dispatch: Dispatch) => {
+    if (accountIndex === null) {
+        TrezorConnect.cancel();
```
mroz22, line 45: this might cause problmes with bridge 2.0.27 but I'd say we may ignore it since it is going to be deprecated soon

mroz22 (approved):

PR #18077, opened 2025-04-02
test(suite-desktop): add connect websocket tests

mroz22:
this is just a wip kickoff to be discussed with @Vere-Grey and @HajekOndrej

PR #18214, opened 2025-04-08
feat(transport-bluetooth): add TrezorBluetooth client

szymonlesisz:
## Description
cherry-picked from `feat/rust-bluetooth`

WebsocketClient for [trezor-bluetooth server](https://github.com/trezor/trezor-suite/pull/18126)

will be used by `BluetoothIpc` and `BluetoothApi` transport

### DIFF
--- a/packages/transport-bluetooth/src/client/trezor-bluetooth.ts
+++ b/packages/transport-bluetooth/src/client/trezor-bluetooth.ts
@@ -1,0 +1,69 @@
+import { WebsocketClient } from '@trezor/websocket-client';
+
+import {
+    BluetoothDevice,
+    BluetoothInfo,
+    Logger,
+    NotificationEvent,
+    TrezorBluetoothSettings,
+} from './types';
+
+// Client for trezor-bluetooth rust websocket server
+export class TrezorBluetooth extends WebsocketClient<NotificationEvent> {
+    readonly settings: TrezorBluetoothSettings;
+    readonly logger: Logger;
+
+    constructor(settings: TrezorBluetoothSettings) {
+        super({
+            url: settings.url,
+        });
+        this.settings = Object.freeze(settings);
+        this.logger = settings.logger || {
+            info: () => {},
+            debug: () => {},
+            log: () => {},
+            warn: () => {},
+            error: () => {},
+        };
+    }
+
+    createWebsocket() {
+        return this.initWebsocket({
+            url: this.settings.url,
+        });
+    }
+
+    ping() {
+        return Promise.resolve(this.sendRawMessage('PING'));
+    }
+
+    send(method: 'get_info', adapter?: boolean): Promise<BluetoothInfo>;
+    send(method: 'enumerate'): Promise<BluetoothDevice[]>;
+    send(method: 'start_scan'): Promise<BluetoothDevice[]>;
+    send(method: 'stop_scan'): Promise<boolean>;
+    send(method: 'connect_device', id: string): Promise<boolean>; // TODO: timeout
+    send(method: 'disconnect_device', id: string): Promise<boolean>;
+    send(method: 'forget_device', id: string): Promise<boolean>;
+    send(method: 'open_device', id: string): Promise<boolean>;
+    send(method: 'close_device', id: string): Promise<boolean>;
+    send(method: 'read', id: string): Promise<boolean>;
+    send(method: 'write', args: [string, number[]]): Promise<boolean>;
+    public send(method: string, args?: any) {
+        return this.sendMessage({ method, params: args || [] });
+    }
+
+    protected onMessage(message: string | Buffer) {
+        super.onMessage(message, data => {
+            if (data.event) {
+                this.emit(data.event, data.payload);
+
+                return;
+            }
+            if (data.error) {
+                throw new Error(data.error);
+            }
+
+            return data.payload;
+        });
+    }
+}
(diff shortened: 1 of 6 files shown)


packages/transport-bluetooth/src/client/trezor-bluetooth.ts
```diff
@@ -0,0 +1,75 @@
...
+            warn: () => {},
+            error: () => {},
+        };
+    }
+
+    createWebsocket() {
+        return this.initWebsocket({
+            url: this.settings.url,
+            headers: {
+                Origin: 'https://node.trezor.io',
```
mroz22: fyi - i have this commit in another branch https://github.com/trezor/trezor-suite/pull/18077/commits/dfdab84b21e7d336ca7190c462f8e6f5277c2af0 
what do you think?

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
```diff
@@ -0,0 +1,75 @@
...
+                'User-Agent': 'Trezor Suite',
+            },
+            timeout: this.settings.timeout || 60 * 1000,
```
mroz22: line 19 and 38 look suspiciously same. are they both needed?

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
```diff
@@ -0,0 +1,75 @@
...
+        });
+    }
+
+    ping() {
+        return Promise.resolve(this.sendRawMessage('PING'));
+    }
+
+    send(method: 'get_info', adapter?: boolean): Promise<BluetoothInfo>;
+    send(method: 'enumerate'): Promise<BluetoothDevice[]>;
+    send(method: 'start_scan'): Promise<BluetoothDevice[]>;
```
mroz22, line 42: I wonder what is the difference between enumerate and start_scan. they have the same signature

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
szymonlesisz: cool :+1:  then i will remove it from here

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
szymonlesisz: `enumerate` will just return known BT adapter state, BT adapter doesnt send scan requests, adapter is not subscribed to scan events. basically without scanning you will receive only paired devices even if they not around (works on linux and windows, macos returns nothing)

`start_scan` will start long term scanning process in the background which eventually should emit some scan events (device discovered, connected, disconnected etc...) until stopped by `stop_scan`
method returns `enumerate` but after(!) scanning, so the device list could be extended with scan results

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
szymonlesisz: well yes, both are needed `WebsocketClient` class naming is confusing.
`timeout` in WebsocketClient constructor (line 19) is used in `createDeferredManager` while `timeout` in `initWebsocket` (here) is used in `new WebSocket`

szymonlesisz: but maybe i will removed it from here all together with the origin, i think i was using it only for testing purposes

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
mroz22: #18223

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
szymonlesisz: removed [22ad13b](https://github.com/trezor/trezor-suite/pull/18214/commits/22ad13bb3846ad7fd0db2f7cce5db33c45dd8d2a) all unnecessary options, we can rely on defaults for now and adjust it later if needed

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
```diff
@@ -0,0 +1,69 @@
...
+            warn: () => {},
+            error: () => {},
+        };
+    }
+
+    createWebsocket() {
+        return this.initWebsocket({
+            url: this.settings.url,
+        });
+    }
```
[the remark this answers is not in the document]
szymonlesisz: i will rely on default as i mentioned here https://github.com/trezor/trezor-suite/pull/18214#discussion_r2034939301

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
```diff
@@ -0,0 +1,69 @@
...
+
+                return;
+            }
+            if (data.error) {
+                throw new Error(data.error);
+            }
+
+            return data.payload;
+        });
+    }
```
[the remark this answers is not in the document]
szymonlesisz: no need for that, error will be catched by upper layer (client implementation)

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
mroz22: already merged

packages/transport-bluetooth/src/client/trezor-bluetooth.ts
mroz22: ok, thanks for explanation

mroz22 (approved):
no objections

PR #18215, opened 2025-04-08
feat(suite-native): display date and time based on localization

### LINKED ISSUE #17747: Display dates according the device location settings
As a user, I would like to see the dates based on my device location settings

**Problem**
- As of Today, the dates that we show in the transaction list is currently hard-coded as `DD/MM/YYYY`. Some [users have complained](https://satoshilabs.slack.com/archives/C03FL9TCURZ/p1742213409630779) about this date format, as for example the standard in the US is `MM/DD/YYYY`

**Proposed solution**
- Display the date format according to the device location settings of the User's mobile device

BrantalikP:
## Description

remove HC Date/time formatting

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17747

## Screenshots:
Screenshot_2025-04-08-16-30-24-572_io trezor suite debug
Screenshot_2025-04-08-16-30-09-613_io trezor suite debug
Screenshot_2025-04-08-16-29-06-710_io trezor suite debug

### DIFF
--- a/suite-common/formatters/src/formatters/prepareDateFormatter.ts
+++ b/suite-common/formatters/src/formatters/prepareDateFormatter.ts
@@ -1,13 +1,14 @@
-import { format } from 'date-fns';
-import enUS from 'date-fns/locale/en-US';
-
 import { makeFormatter } from '../makeFormatter';
 import { FormatterConfig } from '../types';
 
-export const DATE_FORMAT = 'dd/MM/yyyy';
+export const dateFormatterOptions: Intl.DateTimeFormatOptions = {
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+};
 
 export const prepareDateFormatter = (_config: FormatterConfig) =>
     makeFormatter<Date | number, string>(
-        value => format(value, DATE_FORMAT, { locale: enUS }),
+        value => new Intl.DateTimeFormat(undefined, dateFormatterOptions).format(value),
         'DateFormatter',
     );
--- a/suite-common/formatters/src/formatters/prepareDateTimeFormatter.ts
+++ b/suite-common/formatters/src/formatters/prepareDateTimeFormatter.ts
@@ -1,16 +1,17 @@
-import { format } from 'date-fns';
-import enUS from 'date-fns/locale/en-US';
-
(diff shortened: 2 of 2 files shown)


PeKne:
Is it OK that `useFormattersConfig` hardcodes the `locale` to  `"en:"` (https://github.com/trezor/trezor-suite/blob/7819c5a8687ee56bf4ccee446f07538d30f4628b/suite-native/app/src/hooks/useFormattersConfig.ts#L17). Maybe it could be removed now? But not sure about it, it might mess up the other formatters.

suite-common/formatters/src/formatters/prepareDateFormatter.ts
```diff
@@ -1,13 +1,14 @@
...
 import { FormatterConfig } from '../types';
 
-export const DATE_FORMAT = 'dd/MM/yyyy';
+export const dateFormatterOptions: Intl.DateTimeFormatOptions = {
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+};
 
 export const prepareDateFormatter = (_config: FormatterConfig) =>
```
PeKne, line 10: You are changing code that is in `suite-common` that means it is shared with desktop. Did you check if the desktop behavior is untouched?

BrantalikP: Ive checked. its used only by the app, What I remember Web peeps are already supporting the time formatting based on the localization or at least based on the selected language

BrantalikP:
> Is it OK that `useFormattersConfig` hardcodes the `locale` to `"en:"` (
> 
> https://github.com/trezor/trezor-suite/blob/7819c5a8687ee56bf4ccee446f07538d30f4628b/suite-native/app/src/hooks/useFormattersConfig.ts#L17
> ). Maybe it could be removed now? But not sure about it, it might mess up the other formatters.

It does not affect the formatting functions at all. Actually, it’s interesting because even previously both of them had a hardcoded `local=enUS `value rather than using it from the config. https://github.com/trezor/trezor-suite/pull/18215/files#diff-9f8fe5e96489cb3c41c0040e4115f62b16351d3d5811994edc51cc4b5616f1c5L15

PeKne (approved):

PR #18220, opened 2025-04-08
Feat(suite): hide coins step for BTC-only onboarding

### LINKED ISSUE #18205: Remove redundant coin selection screen in Bitcoin-only onboarding
Fourth step of onboarding offers coin selection. It's not useful for bitcoin-only users. Testnet and Tor settings is not necessary at this prominent place.

Image

## QA instructions

There was quite some refactoring in the end, so please test:

Test onboarding with a wiped device with universal firmware:
- T1B1
- T2T1
- T2B1 or T3B1 or T3T1

Test onboarding with a wiped device with btc-only firmware:
- T1B1
- T2T1
- T2B1 or T3B1 or T3T1

Test onboarding with any fresh device:
- with installing universal firmware
- with installing btc-only firmware

Lemonexe:
## Description

:eyes: CR commit by commit is recommended

- hide coins step when onboarding a device with BTC-only firmware
- and do a lot of refactoring that I didn't plan, but I evaluated it necessary to make it _clean_  
  - move `progressBarSteps` into config
  - refactor `isStepUsed`
    - currently takes getState fn as param, which is suitable when called in a thunk, but very unoptimal in a Component (would need `useSelector(s=>s)`) 
    - cleanup in the test
  - ~refactor the progressStep groups, which used index, now it will use the existing keys~ not relevant in the end
  - refactor `progressBarSteps` & `steps` into hierarchical `stepCategories`

## Related Issue

Resolve #18205

## Screenshots:
Onboarding BTC-only T2T1 **after**
[onboarding btc.webm](https://github.com/user-attachments/assets/b6e08961-43dc-489b-85c8-49ebd3e36637)

Onboarding universal T1B1 **unchanged**  
[onboarding uni T1B1.webm](https://github.com/user-attachments/assets/27973cdb-d17c-4f43-89b8-c9d5423d3714)

Onboarding universal T3T1 **unchanged**
[onboarding uni T3T1.webm](https://github.com/user-attachments/assets/00641431-3bf1-49c6-9866-b8f11f3dcad7)

[unexpected state](https://github.com/user-attachments/assets/bf1a7f3d-90a7-4864-b7c2-e6b09cc81587) **unchanged**


packages/suite/src/components/onboarding/OnboardingProgressBar.tsx
```diff
@@ -88,36 +92,33 @@ const Divider = styled.div`
...
     const theme = useTheme();
 
+    const progressBarStepsInPath = useOnboardingProgressBarStepsInPath();
+    const lastStepNumber = progressBarStepsInPath.length - 1;
+    const indexOfActiveStep = progressBarStepsInPath.findIndex(({ key }) => key === activeStep);
+
     return (
         <ProgressBarWrapper className={className}>
-            {steps.map((step, index) => {
-                const stepCompleted = (activeStep ?? 0) > index;
```
Lemonexe, line 109: I was just thinking about the `?? 0` fallback.
0 can never be > index, because min index is 0.
So if `findIndex` returns -1, the behavior is unchanged.

packages/suite/src/hooks/suite/useOnboardingProgressBarStepsInPath.tsx
```diff
@@ -0,0 +1,17 @@
...
+    const isStepUsedContext = useSelector(selectIsStepUsedContext);
+
+    return useMemo(() => {
+        const stepsInPath = steps.filter(step => isStepUsed(step, isStepUsedContext));
+
+        return progressBarSteps.filter(progressBarStep =>
+            stepsInPath.some(({ stepGroup }) => stepGroup === progressBarStep.key),
+        );
+    }, [isStepUsedContext]);
+};
```
Lemonexe: I wanted to avoid duplicating the logic that
- coins step is hidden when isBitcoinOnly
- coins `progressBarStep` is hidden when isBitcoinOnly
So instead of hacking the `coins` `progressBarStep` to hide in that condition, this does it generally.

packages/suite/src/types/onboarding/index.ts
```diff
@@ -9,12 +10,20 @@ type ModelWithFirmwareVersion = {
     minFwVersion: FirmwareVersionString;
 };
 
+export type ProgressBarStepKey = 'device' | 'wallet' | 'pin' | 'coins' | 'final';
+
+export type ProgressBarStep = {
+    key: ProgressBarStepKey;
```
Lemonexe: this is not ideal, because with `Array<ProgressBarStep>` type, there is no way to have TS enforce that _all_ `ProgressBarStepKey`s are used in the array. Theoretically you can drop a progressBarStep and have a key hanging here unused, but this is not modified very often, so I think it's fine.

The proper solution would be to have `ProgressBarStepKey` inferred from `progressBarSteps`, which then only `satisfies` a broader type with `key: string`, but that would mean :recycle: import :hear_no_evil: 

But :recycle: imports are fine when it's just TS, so idk, should I do it?

packages/suite/src/components/onboarding/OnboardingProgressBar.tsx
```diff
@@ -88,36 +92,33 @@ const Divider = styled.div`
...
-                const stepActive = index === activeStep;
+            {progressBarStepsInPath.map(({ key, labelTranslationId }, index) => {
+                // if active step was not found (-1), no step will be considered completed, but that should not happen, see the types
```
Lemonexe, line 109: I was just thinking about the `?? 0` fallback that was there.
0 can never be > index, because min index is 0.
So if `findIndex` returns -1, the behavior is unchanged.

packages/suite/src/components/onboarding/OnboardingProgressBar.tsx
```diff
@@ -88,36 +92,38 @@ const Divider = styled.div`
...
+        ({ key }) => key === activeStepGroup,
+    );
+
     return (
         <ProgressBarWrapper className={className}>
-            {steps.map((step, index) => {
-                const stepCompleted = (activeStep ?? 0) > index;
-                const stepActive = index === activeStep;
+            {progressBarStepsInPath.map(({ key, labelTranslationId }, index) => {
+                // if active step was not found (-1) because activeStepGroup is undefined, no step will be considered completed
```
Lemonexe, line 136: I was just thinking about the `?? 0` fallback that was there.
0 can never be > index, because min index is 0.
So if `findIndex` returns -1 because `activeStep` is `undefined`, the behavior is unchanged :heavy_check_mark:

packages/suite/src/types/onboarding/index.ts
```diff
@@ -23,6 +23,7 @@ export type Step = {
     prerequisites?: (PrerequisiteType | 'device-different')[];
     path?: AnyPath[];
     supportedModels?: (DeviceModelInternal | ModelWithFirmwareVersion)[];
+    hideForBitcoinOnly?: boolean;
```
nexus-labs-admin, line 26: negative term, better use something like `isVisibleForBitcoinOnly`

packages/suite/src/config/onboarding/steps.ts
```diff
@@ -121,6 +121,7 @@ export const steps: Step[] = [
         id: STEP.ID_COINS_STEP,
         stepGroup: 'coins',
         prerequisites: [...commonPrerequisites, ...afterInitializePrerequisites],
+        hideForBitcoinOnly: true,
```
peter-sanderson: negative if -> something like `isVisibleForBitcoinOnly`

peter-sanderson: Or `bitcoinOnly | universal | both`?

packages/suite/src/config/onboarding/steps.ts
```diff
@@ -19,7 +19,29 @@ const afterInitializePrerequisites: Step['prerequisites'] = [
     'device-different',
 ];
 
-const steps: Step[] = [
+export const progressBarSteps: ProgressBarStep[] = [
```
peter-sanderson: I think steps can be sub-array config of the progressBarSteps -> you then dont need the mapping over keys

packages/suite/src/components/onboarding/OnboardingLayout.tsx
```diff
@@ -151,7 +151,9 @@ export const OnboardingLayout = ({ children }: OnboardingLayoutProps) => {
                                     </LogoHeaderRow>
 
                                     <ProgressBarRow>
-                                        <OnboardingProgressBar activeStep={activeStep.stepGroup} />
+                                        <OnboardingProgressBar
+                                            activeStepGroup={activeStep.stepGroup}
```
peter-sanderson: stepGroup => progressBarStep

packages/suite/src/utils/onboarding/steps.ts
```diff
@@ -4,13 +4,23 @@ import { versionUtils } from '@trezor/utils';
 
 import { ID_AUTHENTICATE_DEVICE_STEP } from 'src/constants/onboarding/steps';
 import { AnyPath, AnyStepId, Step } from 'src/types/onboarding';
-import { GetState } from 'src/types/suite';
+import { AppState } from 'src/types/suite';
 
-export const isStepUsed = (step: Step, getState: GetState): boolean => {
-    const state = getState();
-    const device = selectSelectedDevice(state);
+export const selectIsStepUsedContext = (state: AppState) => ({
```
peter-sanderson: it uses whole state anyway, if you want to optimize, memo is needed

packages/suite/src/components/onboarding/OnboardingLayout.tsx
Lemonexe: Haha, not even needed in the end!

packages/suite/src/config/onboarding/steps.ts
Lemonexe: Included in a81134c6b4b9ff29e918002926dea856e30d720c

But I decided to keep consistency with `supportedModels`. They are optional: if not specified, all are allowed.
If specified, there must be full specification what is allowed and what not.
So in the end only the coins step has 
`supportedFirmwareTypes: ['universal']`

packages/suite/src/config/onboarding/steps.ts
Lemonexe: Done in a81134c6b4b9ff29e918002926dea856e30d720c :muscle:

packages/suite/src/utils/onboarding/steps.ts
Lemonexe: Renamed to Props as discussed.
And I removed the selector (not that useful), instead defined Props as a specified structure, instead of inferred ReturnType.

packages/suite/src/components/onboarding/OnboardingProgressBar.tsx
```diff
@@ -1,10 +1,43 @@
...
+                isStepCategoryUsed(stepCategory, {
+                    device,
+                    onboardingPath,
+                    isDeviceAuthenticityCheckEnabled,
+                    isUnlockedBootloaderAllowed,
+                }),
+            ),
+        [device, onboardingPath, isDeviceAuthenticityCheckEnabled, isUnlockedBootloaderAllowed],
+    );
+};
```
Lemonexe, line 39: I moved this here instead of the separate file I had previously. It's not really relevant anywhere else.

packages/suite/src/hooks/suite/useOnboarding.ts
```diff
@@ -6,10 +8,13 @@ import * as routerActions from 'src/actions/suite/routerActions';
...
 import { useActions, useDispatch, useSelector } from 'src/hooks/suite';
 
+import { parseStepId } from '../../utils/onboarding/steps';
+
 export const useOnboarding = () => {
     const dispatch = useDispatch();
 
-    const { onboarding, modal } = useSelector(({ onboarding, modal }) => ({ onboarding, modal }));
+    const onboarding = useSelector(state => state.onboarding);
+    const modal = useSelector(state => state.modal);
```
Lemonexe, line 17: :warning: this was inefficient! Creating new reference everytim

packages/suite/src/utils/onboarding/__tests__/steps.test.ts
```diff
@@ -58,79 +56,55 @@ describe('steps', () => {
 
     describe('isStepUsed', () => {
```
peter-sanderson, line 63: I would say newly added `if` for `supportedFirmwareTypes` is not covered in any test

peter-sanderson (approved):

packages/suite/src/utils/onboarding/__tests__/steps.test.ts
Lemonexe: Done in b7fe42fe5ba35666175664b34c6040dce36716ed

Lemonexe:
[Failing Web E2E test confirmed to be broken](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1744096117384669?thread_ts=1743572836.379729&cid=CKYFBT2C9)

PR #18227, opened 2025-04-09
fix(transport): readMessageBuffer.cancleRead

szymonlesisz:
## Description

pending transport.read will never be resolved if device is closed

e2e testing script with emulator:
- create ./packages/transport/test.ts
- run `tsx  ./packages/transport/test.ts`

```
import * as messages from '@trezor/protobuf/messages.json';

import { UdpTransport } from '../transport/src';

const run = async () => {
    const debugTransport = new UdpTransport({ id: 'udp', messages, logger: console });

    await debugTransport.init();
    debugTransport.listen();

    const enumerate = await debugTransport.enumerate();
    const device = enumerate.payload[0];

    let session = await debugTransport.acquire({ input: device }).then(p => p.payload);
    let readPromise = debugTransport.receive({ session });
    await debugTransport.release({ path: device.path, session });
    let result = await readPromise;
    console.warn('result1', result);

    session = await debugTransport.acquire({ input: device }).then(p => p.payload);
    readPromise = debugTransport.receive({ session });
    console.warn('waiting for disconnect');

    result = await readPromise;
    console.warn('result2', result);
};

run();
```

### DIFF
--- a/packages/transport/src/utils/readMessageBuffer.ts
+++ b/packages/transport/src/utils/readMessageBuffer.ts
@@ -1,10 +1,13 @@
 import { Deferred, createDeferred } from '@trezor/utils';
 
 import { error, success } from './result';
-import { ABORTED_BY_SIGNAL } from '../errors';
+import { ABORTED_BY_SIGNAL, INTERFACE_DATA_TRANSFER } from '../errors';
 import { ResultWithTypedError } from '../types';
 
-type ReadResult = ResultWithTypedError<Buffer, typeof ABORTED_BY_SIGNAL>;
+type ReadResult = ResultWithTypedError<
+    Buffer,
+    typeof ABORTED_BY_SIGNAL | typeof INTERFACE_DATA_TRANSFER
+>;
 
 export const readMessageBuffer = () => {
     const readBuffer: Record<string, Buffer[]> = {};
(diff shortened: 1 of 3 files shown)


packages/transport/src/utils/readMessageBuffer.ts
```diff
@@ -48,8 +48,18 @@ export const readMessageBuffer = () => {
         });
     };
 
+    const close = (path: string) => {
+        if (readRequests[path]) {
+            // @ts-expect-error is this a new expected error?
+            readRequests[path].resolve(error({ error: UNEXPECTED_ERROR, message: 'Closed' }));
```
szymonlesisz: cc @mroz22 
or should i use existing like INTERFACE_DATA_TRANSFER or DEVICE_NOT_FOUND?

mroz22: maybe it should be `INTERFACE_UNABLE_TO_OPEN_DEVICE`. this is what you would get over USB when you fail to open device before reading.

mroz22 (approved):
just one note https://github.com/trezor/trezor-suite/pull/18227#discussion_r2037075988

packages/transport/src/utils/readMessageBuffer.ts
szymonlesisz: well, not quite. we are already in opened state and reading, but closeDevice (and now im starting to realize that also disconnect and interface read errors) should interrupt pending read promise

szymonlesisz: INTERFACE_DATA_TRANSFER it is, resolved irl

szymonlesisz:
last minute forcepush. renamed commit and UDP find disconnected device removed `if (known.length < this.devices.length)` condition

PR #18224, opened 2025-04-09
Feat: suite desktop BluetoothProcess

szymonlesisz:
## Description

Cherry picked from `feat/rust-bluetooth` branch

- added `BluetoothProcess` to suite-desktop. The process will start [websocket server](https://github.com/trezor/trezor-suite/pull/18126)
- picked some components and actions to call `bluetoothIpc.init` and `bluetoothIpc.dispose` - which should start or stop the process. cc @peter-sanderson

~~I've found a bug in the `BaseProcess` and i think its there since forever.
When BaseProcess fails to run the binary (this exact case - binary is not here yet) then stopping this process is not possible.
as a workaround i've used code below, but i'm not sure if this is the right solution since error event doesnt exactly mean that the process is not running (or does it? i could be wrong)~~
addressed: https://github.com/trezor/trezor-suite/pull/18234

PR #18234, opened 2025-04-09
Improved `BaseProcess` error handling

marekrjpolak:
## Description

This PR should fix two bugs:

- When process couldn't be spawned at all, no `exit` event was emitted, therefore it couldn't be cleaned up and/or autorestarted. Fixed by subscribing to `close` event instead.
- When process was stopped after autorestart had already been planned but before it was executed, it was autorestarted anyway. Fixed by checking `this.stopped` again just before restarting.


szymonlesisz (approved):
:1st_place_medal: tested with https://github.com/trezor/trezor-suite/pull/18224 and works as expected

PR #18241, opened 2025-04-09
Data Analytics: add value closed to the provider parameter of the event settings/general/labeling-provider

### LINKED ISSUE #15717: Data Analytics: Labeling modal closed
We would like to add new value (`closed`) to parameter `provider` of event [settings/general/labeling-provider](https://www.notion.so/satoshilabs/settings-general-labeling-provider-6880d295bd084c9d8e1e91199c201bd4?pvs=4) triggered when modal is closed without selecting provider.

izmy:
## Description

The settings/general/labeling-provider event now includes a new provider value: "closed". This value is triggered when the modal is closed without selecting a provider, improving tracking of user interactions.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/15717

### DIFF
--- a/packages/suite-analytics/src/types/events.ts
+++ b/packages/suite-analytics/src/types/events.ts
@@ -454,6 +454,7 @@ export type SuiteAnalyticsEvent =
                   | 'fileSystem'
                   | 'missing-provider'
                   | 'inMemoryTest'
+                  | 'closed'
                   | ''; // Todo: 'sdCard' not implemented yet
           };
       }
--- a/packages/suite/src/actions/suite/metadataLabelingActions.ts
+++ b/packages/suite/src/actions/suite/metadataLabelingActions.ts
@@ -4,6 +4,7 @@ import {
     selectSelectedDevice,
 } from '@suite-common/wallet-core';
 import TrezorConnect, { StaticSessionId } from '@trezor/connect';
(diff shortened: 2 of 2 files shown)


packages/suite-analytics/src/types/events.ts
```diff
@@ -454,6 +454,7 @@ export type SuiteAnalyticsEvent =
                   | 'fileSystem'
                   | 'missing-provider'
                   | 'inMemoryTest'
+                  | 'closed'
```
tomasklim, line 457: what is the difference between `closed` and `missing-provider`. I would report `missing-provider`

tomasklim (changes requested):

packages/suite-analytics/src/types/events.ts
tomasklim: please also update it here https://www.notion.so/satoshilabs/settings-general-labeling-provider-6880d295bd084c9d8e1e91199c201bd4?pvs=4, there is mistake `no-provider` instead of `missing-provider`

you can provide description when it is tracked

izmy: > what is the difference between `closed` and `missing-provider`. I would report `missing-provider`

well, I think that's a question for @kolomaznikt

tomasklim (approved):

PR #18246, opened 2025-04-09
fix(wallet-utils): do not consider pending receive tx for nonces

tomasklim:
## Description

- it took nonce from pending receive txs

## Related Issue

Resolve


izmy (approved):

Ondra-Zik:
**QA OK**

Had a pending incoming transaction with nonce 119. A new outgoing transaction had nonce 250.

Screenshot From 2025-04-16 13-39-15
Screenshot From 2025-04-16 13-38-52

tomasklim:
@Ondra-Zik-SL please tested it the otherway around. You have receiving tx with higher nonce than there is current one on your account. Your case is probably ok also on production right now.

PR #18249, opened 2025-04-09
Disable sending automatic analytics events to Sentry

### LINKED ISSUE #17116: Disable session tracking in Sentry
autoSessionTracking will be removed in the next Sentry version https://docs.sentry.io/platforms/javascript/migration/v8-to-v9/#removed-apis that is why I didn't implement it right away, but we don't want to send any analytics data to Sentry before analytics consent.

### LINKED ISSUE #17846: Set Sentry user by instanceID
On the Desktop, we set instanceId as Sentry user so we can connect analytics data with reported errors to get more info about what happened so we know how to reproduce the issue.

See desktop implementation https://github.com/trezor/trezor-suite/blob/b02953e2e73d7e41404021af6ee3a4046eff3950/packages/suite/src/actions/suite/analyticsActions.ts#L64

Compared to mobile https://github.com/trezor/trezor-suite/blob/b02953e2e73d7e41404021af6ee3a4046eff3950/suite-native/analytics/src/analyticsThunks.ts#L40

matejkriz:
## Description

Disable automatic session tracking in Sentry to prevent sending analytics data from users without analytics consent. We will lose Release Health statistics and automatic tracing, but we don't use them anyway. We should reintroduce custom tracing later once needed.

Introduce Sentry wrapper package for suite-native, so we have some place for Sentry related utils for suite-native and we can enhance Sentry calls globally in the future.

Filter Sentry events by `allowReportTag` same as on desktop and allow to report only redacted events if info about the user confirmation of analytics is not yet loaded or during initial launch before we even asked the user to allow analytics.

Set Sentry User Id same as analytics `instanceId` (again alignment with the desktop implementation). It gives us a way to connect the error event with analytics events.

Stop removing `console.error` from the release bundle to keep it reported to Sentry.

How do I verify that we are no longer sending any analytics events to Sentry? I tested in the simulator and inspected network traffic using the Little Snitch app to see that there is no request to Sentry.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17116
Resolve https://github.com/trezor/trezor-suite/issues/17846


packages/suite/src/utils/suite/sentry.ts
```diff
@@ -24,10 +24,6 @@ export const setSentryUser = (instanceId: string) => {
     Sentry.setUser({ id: instanceId });
 };
 
-export const unsetSentryUser = () => {
```
matejkriz, line 27: never used...

suite-common/analytics/src/analyticsReducer.ts
```diff
@@ -61,11 +61,8 @@ export const selectIsAnalyticsEnabled = (state: AnalyticsRootState): boolean =>
 };
 
 export const selectHasUserAllowedTracking = (state: AnalyticsRootState): boolean | undefined => {
-    // If the user has not yet confirmed analytics, return undefined.
-    // Otherwise, return true or false based on the 'confirmed' and 'enabled' flags.
-    if (!state.analytics.confirmed) {
-        return undefined;
-    }
+    const isAnalyticsConfirmed = selectIsAnalyticsConfirmed(state);
```
matejkriz: I believe the logic is unchanged.

vytick: You can make a test for this selector and see for sure 😇 

Also `selectHasUserAllowedTracking` and `selectIsAnalyticsEnabled` look very similar, is there a real difference? On first sight the difference is one being undefined, but in the referenced code it is then used as a boolean anyway 🤔

matejkriz: I'll move it to separate PR, it's indeed overcomplicated.

matejkriz: removed from here, forcepushed

matejkriz: @vytick Here is the refactoring PR based on this PR so I don't have to resolve conflicts later. https://github.com/trezor/trezor-suite/pull/18300

vytick (approved):
Looks good and seems to work correctly (tested and found debug error in sentry when enabled). Nice job 👌

PR #18260, opened 2025-04-10
Chore/optimize regex

komret:
An attempt to optimize regex. Suite is not really vulnerable to ReDoS attacks, but I want to silence the alerts.

Capped the capture group at 4 characters length so that it is not so greedy, CSS doesn't have any longer unit. [source](https://www.w3schools.com/cssref/css_units.php)

Resolves:
https://github.com/trezor/trezor-suite/security/code-scanning/157

### DIFF
--- a/packages/styles/src/utils.ts
+++ b/packages/styles/src/utils.ts
@@ -4,7 +4,7 @@ import { CSSColor } from '@trezor/theme';
 
 export const getValueAndUnit = (valueAndUnit: string): [value: number, unit: string] => {
     const value = parseFloat(valueAndUnit);
-    const [unit] = valueAndUnit.match(/([a-zA-Z]*|%)$/) ?? [''];
+    const [unit] = valueAndUnit.match(/([a-zA-Z]{1,4}|%)$/) ?? ['']; // CSS units shouldn't have more than 4 characters.
 
     return [value, unit];
 };


packages/styles/src/utils.ts
```diff
@@ -4,7 +4,7 @@ import { CSSColor } from '@trezor/theme';
 
 export const getValueAndUnit = (valueAndUnit: string): [value: number, unit: string] => {
     const value = parseFloat(valueAndUnit);
-    const [unit] = valueAndUnit.match(/([a-zA-Z]*|%)$/) ?? [''];
+    const [unit] = valueAndUnit.match(/([a-zA-Z]+|%)$/) ?? [''];
```
komret, line 7: This should behave the same as the previous version because there is the fallback to `''`.

packages/utils/src/extractUrlsFromText.ts
```diff
@@ -1,5 +1,5 @@
 const URL_REGEX =
-    /\b(?:https?:\/\/|www\.)[a-zA-Z0-9-._~:/?#[\]@!$&'()*+,;=%]+\b|(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(?=\b|\s|$|\])/gi;
```
komret, line 2: The nested `+` signs could be an issue, according to ChatGPT. Not sure about this.

Lemonexe (approved):

PR #18275, opened 2025-04-10
chore(suite): Update deps - react-qr-reader

jvaclavik:
## Related Issue

Resolve


jvaclavik:
we dont' want to update to beta, devs also probably left the repo

PR #18286, opened 2025-04-10
chore(suite-native): Mobile Trade: Memoize BuyForm

jbazant:
## Description
Less re-renders caused by timer ticks.

## Related Issue

Resolve


vytick (approved):
nice job 👌

PR #18287, opened 2025-04-10
feat(suite-native): Device onboarding: Recovery

### LINKED ISSUE #18186: Onboarding - Recovery V1
The device onboarding will be released in May. the first version will not support wallet backup recovery. Implement [this desing](https://www.figma.com/design/UdQpTfz2RhsoZzqLp9L8fU/Onboarding?node-id=4533-10110&t=ANQEc6vWUtEuPRxk-4) that communicates it to the user. 

In case you have any design specific question, contact @shenkys directly.

BrantalikP:
## Description

Add Recovery Screen

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18186


suite-native/atoms/src/IconListItem.tsx
```diff
@@ -32,6 +32,11 @@ const iconColorsMap = {
         iconBorderColor: 'backgroundAlertYellowSubtleOnElevation0',
         iconBackgroundColor: 'backgroundAlertYellowSubtleOnElevation1',
     },
+    primary: {
```
matejkriz, line 35: Not sure if we should introduce a new variant that is used only in one place, OrderedListIcon can be used directly, but in this case I agree that it's a good idea to add the primary variant since it's more common pattern in our UI, that las item in the list has this variant. So let's keep it 👌

suite-native/module-device-onboarding/src/screens/RecoveryScreen.tsx
```diff
@@ -0,0 +1,80 @@
...
+    DeviceOnboardingStackParamList,
+    DeviceOnboardingStackRoutes,
+    HomeStackRoutes,
+    RootStackParamList,
+    RootStackRoutes,
+    Screen,
+    ScreenHeader,
+    StackToStackCompositeNavigationProps,
+} from '@suite-native/navigation';
+import { SUITE_WEB_URL } from '@trezor/urls';
```
matejkriz, line 17: We shouldn’t import from packages, but I see that @trezor/urls is already imported in multiple places, including in suite-common, and since it has no dependencies, it’s not really an issue. We should move it to the suite-common folder, though, but that’s outside the scope of this PR, so let’s leave it for a separate refactor.

matejkriz (approved):
Looks good and works as expected. The user flow is not ideal (especially how to get back to Suite Lite after recovery and broken layout of web Suite on initial screen) but that is expected and only temporary.

My comments are just comment, no action needed now so feel free to merge after conflicts resolution 👍

Issue #17928, opened 2025-03-26
Mobile Trade: get trade for selected quote and present webview


vytick:
When there is a selected quote:
- make the continue button visible, make it hidden otherwise. Make it visible even if there is no account selected to receive the asset
- when there is no account selected open account selection on continue button tapped
- when there is account selected, present legal consent screen and when this one is present, request trade for the quote
- when the trade response is received, get tradeForm data, using `tradeFormUtils` extract source and present `TradingWebviewScreen`

Ondra-Zik:
**QA OK**

Described functions are operating

version: 316e68d2c8745164a5eb3d7fb4aafd11e9bd2a29

PR #18288, opened 2025-04-10
Fix mobile Buy trade flow

vytick:
- During optimisations I messed up buy flow a bit. This fixes it
- Trading url scheme needs to be different than scheme registered for deeplinking

Followup to #17928


jbazant (approved):
LGTM!

PR #18290, opened 2025-04-11
fix: deeplinks into debug settings

peter-sanderson:
On Linux: `xdg-open trezorsuite://anchor-@debug-settings/github-issue`

image

### DIFF
--- a/packages/suite/src/constants/suite/anchors.ts
+++ b/packages/suite/src/constants/suite/anchors.ts
@@ -1,10 +1,6 @@
 import { Route } from '@suite-common/suite-types';
 
-export type AnchorSettingSection =
-    | 'general-settings'
-    | 'device-settings'
-    | 'coin-settings'
-    | 'debug-settings';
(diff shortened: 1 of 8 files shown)


packages/suite/src/constants/suite/anchors.ts
```diff
@@ -64,7 +64,7 @@ export const mapAnchorToRoute: Record<AnchorSettingSection, Route['name']> = {
     'general-settings': 'settings-index',
     'device-settings': 'settings-device',
     'coin-settings': 'settings-coins',
-    'debug-settings': 'suite-start',
+    'debug-settings': 'settings-debug',
```
komret: Do we have a use case for this? I suppose it doesn't work until you have debug enabled and we should not send customers to debug settings anyway? So I'd rather delete it to prevent confusion.

komret (changes requested):

Lemonexe (approved):
Works :tada:

komret (approved):
Thanks for the cleanup!

PR #18299, opened 2025-04-11
feat(suite-desktop): add AssetsFilterPlugin webpack plugin

szymonlesisz:
## Description

I want to exclude [bluetooth assets/binaries](https://github.com/trezor/trezor-suite/pull/18126) from the compilation and [include them conditionally ](https://github.com/trezor/trezor-suite/pull/18196)

`suite-data/files/bin` directory is copied as a whole so i need to filter out files using regex

once binary is in develop branch you should see:
```
yarn workspace @trezor/suite-desktop dev
3% setup watch run webpack-plugin-serve⬡ wps: Server Listening on: http://localhost:8000

92% sealing asset processing AssetsFilterPlugin files removed from compilation: static/bin/bluetooth/mac-x64/trezor-bluetooth, static/bin/bluetooth/mac-arm64/trezor-bluetooth, static/bin/bluetooth/linux-arm64/trezor-bluetooth, static/bin/bluetooth/linux-x64/trezor-bluetooth, static/bin/bluetooth/win-x64/trezor-bluetooth.exe, static/bin/bluetooth/win-arm64/trezor-bluetooth.exe
98% emitting after emit NixosInterpreterPluginNixosInterpreterPlugin not used
98% emitting after emit ShellSpawnPluginassets by chunk 14.5 MiB (id hint: vendors) 20 assets
```

until then it should be:
```
92% sealing asset processing AssetsFilterPlugin no files removed
```

while doing so i've noticed that there is 4500+ assets in the compilation where more than 80% are svg and png files.
`AssetsFilterPlugin` could be good place to manage them and optimize the build

### DIFF
--- a/packages/suite-build/configs/desktop.webpack.config.ts
+++ b/packages/suite-build/configs/desktop.webpack.config.ts
@@ -5,6 +5,7 @@ import webpack from 'webpack';
 
 import { FLAGS } from '@suite-common/suite-config';
 
+import { AssetsFilterPlugin } from '../plugins/assets-filter-plugin';
 import { NixosInterpreterPlugin } from '../plugins/nixos-interpreter-plugin';
 import ShellSpawnPlugin from '../plugins/shell-spawn-plugin';
 import { assetPrefix, isCodesignBuild, isDev, launchElectron } from '../utils/env';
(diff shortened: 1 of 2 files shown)


mroz22 (approved):
looks nice. maybe @komret would like to take a look as well

komret (approved):

PR #18302, opened 2025-04-11
fix(suite-native): Alert Sheet fix

### LINKED ISSUE #17838: AlertSheet component behaves incorrectly on low-end Android devices
**Describe the bug**
On slower or older Android devices, the `AlertSheet` component does not animate correctly. It either start at wrong position or it is triggered multiple times

**Info:**

-   Mobile Suite Version: 25.3.2/85538688bc
-   OS: Galaxy S8+, Android 9

**How to reproduce**

1. Use a low-end Android device (older or slower model).
2. Trigger the Alert Sheet Modal
3. Observe

**Expected behavior**
The AlertSheet should slide in from the bottom of the screen with no glitches

**Screenshots**
https://github.com/user-attachments/assets/2a75cbfe-57a9-4698-8fa4-aa9050c40431

BrantalikP:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17838

### DIFF
--- a/suite-native/alerts/src/components/AlertSheet.tsx
+++ b/suite-native/alerts/src/components/AlertSheet.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from 'react';
-import { Modal, Pressable } from 'react-native';
+import { Modal, Pressable, StyleSheet } from 'react-native';
 import Animated from 'react-native-reanimated';
 
 import {
@@ -11,6 +11,7 @@ import {
     VStack,
     useBottomSheetAnimation,
 } from '@suite-native/atoms';
+import { getScreenHeight, getScreenWidth } from '@trezor/env-utils';
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 
 import { Alert } from '../alertsAtoms';
@@ -20,6 +21,9 @@ import { useShakeAnimation } from '../useShakeAnimation';
 type AlertSheetProps = {
     alert: Alert;
 };
+
+const SCREEN_WIDTH = getScreenWidth();
+const SCREEN_HEIGHT = getScreenHeight();
 
 const alertSheetContainerStyle = prepareNativeStyle(utils => ({
     justifyContent: 'center',
@@ -43,8 +47,12 @@ const shakeTriggerStyle = prepareNativeStyle(_ => ({
 }));
 
 const sheetOverlayStyle = prepareNativeStyle(_ => ({
-    flex: 1,
+    width: SCREEN_WIDTH,
+    height: SCREEN_HEIGHT,
+    ...StyleSheet.absoluteFillObject,
 }));
+
+const AnimatedPressable = Animated.createAnimatedComponent(Pressable);
 
 export const AlertSheet = ({ alert }: AlertSheetProps) => {
     const { hideAlert } = useAlert();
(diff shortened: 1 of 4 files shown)


suite-native/module-home/src/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -45,6 +49,23 @@ export const HomeScreen = () => {
             refreshControl={refreshControl}
             noHorizontalPadding
         >
```
BrantalikP: Note: This file has been changed solely for testing purposes, once the fix is validated, it will be removed via `git ammend`

suite-native/alerts/src/components/AlertSheet.tsx
```diff
@@ -46,6 +46,7 @@ const sheetOverlayStyle = prepareNativeStyle(_ => ({
     flex: 1,
 }));
 
+const ReanimatedPressable = Animated.createAnimatedComponent(Pressable);
```
yanascz, line 49: Could we please separate by a blank line from the following const?

yanascz: This is already used on four different places. Would it make sense to extract a component called `AnimatedPressable` into `atoms`?

BrantalikP: Thank you for the comment! 

IMO, it should not be in `Atoms`, although I understand the argument that we could benefit from making it reusable (following DRY). However, based on what I’ve observed and from an architectural point of view `Atoms` should solely serve as **components/primitives** coupled to our **Design System** (or at least related to the app’s UI in some way), either connected to some `business logig`(e.g `Skeleton`) or the `theme` itself (e.g Box).
`ReanimatedPressable` doesn’t follow either of these. It serves only as a consumer of the `animated style`. It’s similar to re-exporting `Animated.View` just to avoid copy-pasting. in this case, I think it’s completely fine to duplicate and adjust the component to meet the specific needs of the component. Happy to discuss this further if you’d like  :)

suite-native/module-home/src/screens/HomeScreen/HomeScreen.tsx
BrantalikP: removed by https://github.com/trezor/trezor-suite/pull/18302/commits/df36195c890072467ad77c1e8420984750b05d65

suite-native/alerts/src/components/AlertSheet.tsx
yanascz: Well, we've already got `AnimatedHStack`, `AnimatedBox`, `AnimatedCard`, ... But it's true, they're tied to a corresponding design system component. So let's leave it here. 👍 

Could we at least stay consistent with the naming and call it `AnimatedPressable`?

suite-native/alerts/src/components/AlertSheet.tsx
```diff
@@ -21,6 +22,9 @@ type AlertSheetProps = {
     alert: Alert;
 };
 
+const screenWidth = getScreenWidth();
+const screenHeight = getScreenHeight();
```
yanascz: I'd go with the same naming as in other files:
```suggestion
const SCREEN_WIDTH = getScreenWidth();
const SCREEN_HEIGHT = getScreenHeight();
```

yanascz (approved):
Works like a charm! 🪄

PR #18330, opened 2025-04-11
Data Analytics: device-connect - add isBitcoinOnlyDevice to the event

### LINKED ISSUE #9752: Report btc unit device to analytics
Report btc unit device to analytics in related events so that we can measure installing regular fw on those devices.

It can also give us an estimate how many btc unit devices are actually used and how many of them is just for swagging on the shelf

izmy:
## Description

Data Analytics: `device-connect` now has new parameter `isBitcoinOnlyDevice`

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/9752

### DIFF
--- a/packages/suite-analytics/src/types/events.ts
+++ b/packages/suite-analytics/src/types/events.ts
@@ -98,6 +98,7 @@ export type SuiteAnalyticsEvent =
               totalInstances?: number | null;
               backup_type?: string;
               isBitcoinOnly?: boolean;
+              isBitcoinOnlyDevice?: boolean;
               totalDevices?: number;
               language?: string | null;
               model?: string;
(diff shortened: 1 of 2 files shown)


tomasklim (changes requested):
Please use util `isBitcoinOnlyDevice` 😀 put it under `isBitcoinOnly` and name it `isBitcoinOnlyDevice`

packages/suite-analytics/src/types/events.ts
```diff
@@ -97,13 +97,14 @@ export type SuiteAnalyticsEvent =
               passphrase_protection?: boolean | null;
               totalInstances?: number | null;
               backup_type?: string;
-              isBitcoinOnly?: boolean;
```
matejkriz, line 100: You can add `isBitcoinOnlyDevice` but `isBitcoinOnly` has to remain here too...

izmy: sure... at first, I didn't understand the issue

tomasklim (approved):

PR #18333, opened 2025-04-12
refactor(e2e): Move discoverShouldFInish on page Fixture

Vere-Grey:
As discussed, I am introducing a way how to share some minimal functionality across whole tests and prevent necessity of complicated importing between page objects

This PR introduces enhancing of the basic `page` object.  It allows us to expand it by methods and properties.
We should keep those to the minimal and most important things that we need to access in all test and situations.
The first one we have is `discoveryShouldFinish`

I have following [PR](https://github.com/trezor/trezor-suite/pull/18335) ready that utilizes that to reduce a lot of common steps for changing which coins are neabled

The use is `await page.discoveryShouldFinish();` and can be run anywhere you have page object.

### DIFF
--- a/packages/suite-desktop-core/e2e/support/pageObjects/onboarding/onboardingPage.ts
+++ b/packages/suite-desktop-core/e2e/support/pageObjects/onboarding/onboardingPage.ts
@@ -98,7 +98,7 @@ export class OnboardingPage {
     }
 
     @step()
-    async completeOnboarding({ enableViewOnly = false } = {}) {
+    async completeOnboarding(options?: { enableViewOnly?: boolean }) {
         await this.disableNecessaryFirmwareChecks();
         await this.optionallyDismissFwHashCheckError();
         await this.analyticsSection.continueButton.click();
@@ -106,12 +106,13 @@ export class OnboardingPage {
         if (this.isModelWithSecureElement()) {
             await this.passThroughAuthenticityCheck();
         }
-        if (enableViewOnly) {
+        if (options?.enableViewOnly) {
             await this.onboardingViewOnlyEnableButton.click();
         } else {
             await this.onboardingViewOnlySkipButton.click();
         }
         await this.viewOnlyTooltipGotItButton.click();
+        await this.page.discoveryShouldFinish();
     }
 
     @step()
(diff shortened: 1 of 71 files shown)


packages/suite-desktop-core/e2e/support/pageObjects/onboarding/onboardingPage.ts
Vere-Grey, line 1: minor refactoring

packages/suite-desktop-core/e2e/support/testExtends/enhancePage.ts
```diff
@@ -0,0 +1,29 @@
...
+        discoveryShouldFinish(): Promise<void>;
+    }
+}
+
+// This function enhances the Page object with additional properties and methods
+// These properties and methods have general use across the whole test suite.
+// It is not specific to any particular test or feature.
+export const enhancePage = (page: Page): Page => {
+    // Locators
+    page.discoveryBar = page.locator('[data-test="@wallet/discovery-progress-bar"]');
```
Vere-Grey: not sure we can somehow enforce private here. Maybe the locator could be a const in the method?

packages/suite-desktop-core/e2e/support/pageObjects/onboarding/onboardingPage.ts
```diff
@@ -98,20 +98,21 @@ export class OnboardingPage {
...
             await this.passThroughAuthenticityCheck();
         }
-        if (enableViewOnly) {
+        if (options?.enableViewOnly) {
             await this.onboardingViewOnlyEnableButton.click();
         } else {
             await this.onboardingViewOnlySkipButton.click();
         }
         await this.viewOnlyTooltipGotItButton.click();
+        await this.page.discoveryShouldFinish();
```
Vere-Grey, line 115: I don't know about situation where we wouldn't want to wait for the discovery after a onboarding. It is always the safe thing to do.

packages/suite-desktop-core/e2e/support/testExtends/enhancePage.ts
Vere-Grey: moved to method const

HajekOndrej (approved):

PR #18340, opened 2025-04-14
fix(suite-native): fix token send

PeKne:
## Description
Fixes send review of a token transactions. The bug was introduced by https://github.com/trezor/trezor-suite/pull/16905 where the `enhancePrecomposedTransactionThunk` was changed. Initially, the `enhancedPrecomposedTransaction` object was duplicated (by the spread operator) but this PR removed the spreading. For this reason `isTokenKnown` prop change failed on the unwritable object and even if token definitions knew the token, the app treated it as an unknown token, and the flow got stuck.

The bug was located in the `suite-common` parts so it is very likely that desktop suffered by this bug as well.


peter-sanderson (approved):
:clap: good job finding it!

izmy (approved):

STew790:
**QA OK**
I wasn't able to replicate it anymore 🙏 

**Info**
25.4.1 1e1a7dc

PR #18354, opened 2025-04-14
17999 flicking of token fiat value due loadaing

### LINKED ISSUE #17999: ETC tokens list animation during discovery running.
**Describe the bug**
ETC tokens list animation during discovery running.
See video, i was not able to reproduce it on other coins / havent tried all of them/

**Steps to reproduce:**
1. Load testing passphrase
2. During discovery open ETC/Tokens section 
3. c

**Info:**
 - Suite version: web 25.4.0 (73c7fede7e619e1ad81d7c4e808d6b1e3b1c42b8)
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T2T1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)
 - Transport: WebUsbTransport

**Screenshots:**

https://github.com/user-attachments/assets/e92ad74b-dccd-4c51-84be-90d0c9c0d253

vojtatranta:
## Related Issue

Resolve  https://github.com/trezor/trezor-suite/issues/17999

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=17999-flicking-of-token-fiat-value-due-loadaing&lookback=7)

### DIFF
--- a/suite-common/fiat-services/src/rates.ts
+++ b/suite-common/fiat-services/src/rates.ts
@@ -63,6 +63,20 @@ export const fetchCurrentFiatRates = ({
                         { timeout: CONNECT_FETCH_TIMEOUT },
                     );
 
+                    if (!success && payload.error === 'No tickers found!') {
+                        const fallbackCoinGeckoResponse =
+                            await coingeckoService.fetchCurrentFiatRates(ticker);
+
+                        if (!fallbackCoinGeckoResponse) {
+                            return null;
+                        }
+
+                        return {
+                            rate: fallbackCoinGeckoResponse?.rates?.[localCurrency],
+                            lastTickerTimestamp: fallbackCoinGeckoResponse?.ts as Timestamp,
+                        };
+                    }
+
                     if (!success) return null;
 
                     const rate = payload.rates?.[localCurrency];


suite-common/wallet-config/src/networksConfig.ts
```diff
@@ -312,6 +312,10 @@ export const networks = {
         accountTypes: {},
         coingeckoId: 'ethereum-classic',
         tradeCryptoId: 'ethereum-classic',
+        excludedBlockbookTokens: [
```
vojtatranta: either this: we list out those tokens that aren't available through block book, or we will do a generic fallback

suite-common/fiat-services/src/rates.ts
```diff
@@ -63,6 +67,20 @@ export const fetchCurrentFiatRates = ({
                         { timeout: CONNECT_FETCH_TIMEOUT },
                     );
 
+                    if (!success && payload.error === 'No tickers found!') {
```
vojtatranta, line 66: The error here from connect is pretty bad, jsut a sentence, no error code.

So either this, or [this solution ](https://github.com/trezor/trezor-suite/pull/18354/files#r2042205856)

Or we will make up something even better?

marekrjpolak: `this solution` link points to the same thing.

Yes, the error is just a string as it comes directly from Blockbook, so if you really want to ignore other possible errors, this is the way.

Also, this fallback is needed only for current fiat rates and not last week fiat rates or specific timestamp rates? Why not?

By the way, @tomasklim why have we decided to remove fiat rate fallbacks so now it seems we have to readd (some of) them? I believe it had worked that when Connect failed to return fiat rates, we tried request them directly from Blockbook and if it failed as well, we finally tried coingecko.

marekrjpolak (approved):

suite-common/fiat-services/src/rates.ts
tomasklim: > Also, this fallback is needed only for current fiat rates and not last week fiat rates or specific timestamp rates? Why not?

I would say it is needed for all of them, the question is, if we even need fallback for rates which should be on blockbook but for some reason they are not there.

tomasklim: > By the way, @tomasklim why have we decided to remove fiat rate fallbacks so now it seems we have to readd (some of) them? I believe it had worked that when Connect failed to return fiat rates, we tried request them directly from Blockbook and if it failed as well, we finally tried coingecko.

Until now I though it works like that but it does not. However, is it even needed? When it is not available on blockbook (which takes it from coingecko) it is probably not available on coingecko at all (for blockbook based networks)

vojtatranta: Well, I know of 2 coins that are availabeo n coingecko and not available at blockbook.

bosomt:
**QA OK**

https://github.com/trezor/trezor-suite/issues/17999#issuecomment-2820722954

PR #18361, opened 2025-04-14
fix(suite): Show convert to lowercase button for bch CashAddress

### LINKED ISSUE #5151: Add convert to lowercase button for BCH send dialogue
**Describe the bug**
Found this issue during new firmware change in QR code rendering
` Do not convert bech32 addresses to upper-case for QR code #1751 `
QR code generated from Trezor is not accepted by Trezor Suite

**Steps to reproduce:**
1. generate BCH Receive address
2. Scan QR code
3. Use it as receive address in Suite

**Info:**
 - Suite version: web 22.3.2 (563be1e28ad75ddd4e38c2002e6032b0743511d2)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:98.0) Gecko/20100101 Firefox/98.0
 - OS: MacIntel
 - Screen: 1680x1050
 - Device: model T 2.4.3 regular

**Screenshots:**
Screenshot 2022-03-18 at 13 52 42
Screenshot 2022-03-18 at 13 53 20
Screenshot 2022-03-18 at 13 54 33
Screenshot 2022-03-18 at 13 54 38

komret:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/5151

### DIFF
--- a/packages/suite/src/views/wallet/send/Outputs/Address.tsx
+++ b/packages/suite/src/views/wallet/send/Outputs/Address.tsx
@@ -13,6 +13,7 @@ import {
     isAddressDeprecated,
     isAddressValid,
     isBech32AddressUppercase,
+    isCashAddressUppercase,
     isTaprootAddress,
 } from '@suite-common/wallet-utils';
 import { Button, Icon, IconButton, Input, Link, Row } from '@trezor/components';
@@ -210,7 +211,8 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                 }
 
                 return {};
-            case 'uppercase':
+            case 'cashaddr_uppercase':
+            case 'bech32_uppercase':
                 return {
                     buttonProps: {
                         onClick: () =>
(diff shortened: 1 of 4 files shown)


packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -240,6 +242,11 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                     return translationString('TR_UNSUPPORTED_ADDRESS_FORMAT');
                 }
             },
+            cashaddr_uppercase: (value: string) => {
```
komret: I'd keep both of these under the `uppercase` case.

suite-common/wallet-utils/src/validationUtils.ts
```diff
@@ -62,6 +62,9 @@ export const isTaprootAddress = (address: string, symbol: Account['symbol']) =>
     );
 };
 
+export const isCashAddressUppercase = (address: string) =>
+    /^(bitcoincash|bchtest|bchreg):/.test(address.toLowerCase()) && /[A-Z]/.test(address);
```
komret: No need to check for `bchtest` and `bchreg` as they are not supported in Suite and never will be. Please remove also `tltc1` and `tvtc1` from `isBech32AddressUppercase` and shorten unit tests accordingly.

komret (commented):
Now I'm reviewing my own PR but whatever :)

Good job, it works, just adding some nitpicks.

There are clearly more issues with BCH, though. Let's tackle the prefix here, as that is also mentioned in the issue.

If an otherwise valid BCH address without the "bitcoincash:" prefix is entered, input validation should fail. There should be a button to add the prefix and the same error message ("RECIPIENT_IS_NOT_VALID").

komret:
Closing in favour of https://github.com/trezor/trezor-suite/pull/18519

PR #18364, opened 2025-04-14
Feat(suite-native): mobile entropy check

### LINKED ISSUE #18070: Entropy check in mobile Suite
Implement entropy check for mobile Suite. I don't think we need any special designs or copy becvause designs should reuse what we've done for revision check and copy should be the same as on desktop.

## Dev details
There is a new `createAndBackupWalletThunk` added in PR #18362. Improve it so it uses entropy check same as desktop does.

Lemonexe:
## Description

Implement entropy check on mobile:
- implement turning it off via message system 
- execute it in `createAndBackupWalletThunk`
- if failure:
  - report to Sentry
  - display :ghost: modal that is not dismissible, but disappears when disconnecting the device
- cleanup in react components for :ghost: modals

## Related Issue

Resolve #18070 (no design yet, so I repurposed FW revision check UI)

## Screenshots:

1. [Simualted failure with T2B1](https://github.com/user-attachments/assets/833d8f20-16a5-41bd-81e2-2886eeee5690)
_Btw in the testing code I also skipped tutorial screen to last page to save time..
Also, don't mind the tenv graphical glitch at the end xD_
2. [Wallet creation success with T3T1](https://github.com/user-attachments/assets/6fe687dd-6e2d-424d-89de-2a9497a5fa09)
_Similarly it passes with simulated failure && local message system that turns off_ `security.entropyCheck.mobile`
3. [Revision check modal flow is unchanged](https://github.com/user-attachments/assets/54a4aced-e6ee-4328-87a4-6c5d30117d0b)

Just the Entropy :ghost: modal itself:
entropy ghost

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/mobile-entropy-check&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/mobile-entropy-check&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/mobile-entropy-check&lookback=7)

### DIFF
--- a/suite-native/device/src/deviceThunks.ts
+++ b/suite-native/device/src/deviceThunks.ts
@@ -1,8 +1,9 @@
 import { Feature, selectIsFeatureEnabled } from '@suite-common/message-system';
 import { createThunk } from '@suite-common/redux-utils';
 import { deviceActions, selectDevicePath, selectSelectedDevice } from '@suite-common/wallet-core';
-import { WalletBackupType } from '@suite-native/device';
+import { WalletBackupType, reportCheckFail } from '@suite-native/device';
 import TrezorConnect, { PROTO } from '@trezor/connect';
+import { getFirmwareVersion } from '@trezor/device-utils';
 
 const NATIVE_DEVICE_MODULE_PREFIX = 'nativeDevice';
 
@@ -53,7 +54,11 @@ const getResetDeviceConfig = (walletBackupType: WalletBackupType): PROTO.ResetDe
 
 export const createAndBackupWalletThunk = createThunk(
     `${NATIVE_DEVICE_MODULE_PREFIX}/resetDevice`,
-    async ({ walletBackupType }: { walletBackupType: WalletBackupType }, { getState }) => {
+    async (
+        { walletBackupType }: { walletBackupType: WalletBackupType },
+        { getState, dispatch },
+    ) => {
+        const device = selectSelectedDevice(getState());
         const devicePath = selectDevicePath(getState());
         const isEntropyCheckEnabled = selectIsFeatureEnabled(
             getState(),
@@ -61,16 +66,32 @@ export const createAndBackupWalletThunk = createThunk(
             true,
         );
 
-        if (!devicePath) {
+        if (!device || !device.features || !devicePath) {
             throw new Error('Device not found');
         }
 
-        return await TrezorConnect.resetDevice({
+        const result = await TrezorConnect.resetDevice({
             device: { path: devicePath },
             skip_backup: false,
             ...getResetDeviceConfig(walletBackupType),
             //Entropy check can be toggled via message system config so it should be always last to avoid unintentional disabling.
             entropy_check: isEntropyCheckEnabled,
         });
+
+        if (!result.success && result.payload.code === 'Failure_EntropyCheck') {
+            const contextData = {
+                model: device?.features?.internal_model,
+                revision: device?.features?.revision,
+                version: getFirmwareVersion(device),
+                vendor: device?.features?.fw_vendor,
+            };
+            reportCheckFail('Entropy', contextData, result.payload.error);
+            // TODO: temporary exception to avoid false positives, see https://github.com/trezor/trezor-suite-private/issues/135
+            if (result.payload.error !== 'device disconnected during action') {
+                dispatch(deviceActions.setEntropyCheckFail(device.id));
+            }
+        }
+
+        return result;
     },
 );
--- a/suite-native/device/src/deviceThunks.ts
+++ b/suite-native/device/src/deviceThunks.ts
@@ -1,3 +1,4 @@
+import { Feature, selectIsFeatureEnabled } from '@suite-common/message-system';
 import { createThunk } from '@suite-common/redux-utils';
 import { deviceActions, selectDevicePath, selectSelectedDevice } from '@suite-common/wallet-core';
 import { WalletBackupType } from '@suite-native/device';
@@ -54,14 +55,15 @@ export const createAndBackupWalletThunk = createThunk(
     `${NATIVE_DEVICE_MODULE_PREFIX}/resetDevice`,
     async ({ walletBackupType }: { walletBackupType: WalletBackupType }, { getState }) => {
         const devicePath = selectDevicePath(getState());
+        const isEntropyCheckEnabled = selectIsFeatureEnabled(
+            getState(),
+            Feature.entropyCheckMobile,
+            true,
+        );
 
         if (!devicePath) {
             throw new Error('Device not found');
         }
-
-        // TODO: implement entropy check for mobile
-        // https://github.com/trezor/trezor-suite/issues/18355
-        const isEntropyCheckEnabled = false;
 
         return await TrezorConnect.resetDevice({
             device: { path: devicePath },
--- a/suite-native/device/src/hooks/useHandleDeviceConnection.ts
+++ b/suite-native/device/src/hooks/useHandleDeviceConnection.ts
@@ -39,6 +39,7 @@ import {
 import {
     selectHasFirmwareAuthenticityCheckHardFailed,
     selectIsDeviceSetupSupported,
+    selectIsEntropyCheckEnabledAndFailed,
 } from '../selectors';
 
 type NavigationProp = StackToStackCompositeNavigationProps<
@@ -70,6 +71,7 @@ export const useHandleDeviceConnection = () => {
     const isFirmwareAuthenticityCheckDismissed = useSelector(
         selectIsFirmwareAuthenticityCheckDismissed,
     );
+    const shouldDisplayEntropyCheckError = useSelector(selectIsEntropyCheckEnabledAndFailed);
 
     const { isBiometricsOverlayVisible } = useIsBiometricsOverlayVisible();
     const isOnboardingDeviceDisconnectedAlertDisplayed = useAtomValue(
@@ -103,10 +105,16 @@ export const useHandleDeviceConnection = () => {
     const isOnPinMatrixBlacklistedRoute = pinMatrixBlacklistedScreens.includes(
         lastRoute as RootStackRoutes,
     );
+
+    const shouldDisplayFirmwareAuthenticityError =
+        hasFirmwareAuthenticityCheckHardFailed && !isFirmwareAuthenticityCheckDismissed;
+
+    // any failing check should navigate to the DeviceCompromisedModal
     const shouldNavigateToDeviceCompromisedModal =
-        hasFirmwareAuthenticityCheckHardFailed &&
-        !isFirmwareAuthenticityCheckDismissed &&
-        isOnboardingFinished;
+        shouldDisplayEntropyCheckError || shouldDisplayFirmwareAuthenticityError;
+    // but the DeviceCompromisedModal shall not be persistent for Entropy check, because you cannot exit the modal via normal means
+    const shouldKeepDeviceCompromisedModal =
+        isDeviceCompromisedModalFocused && !shouldDisplayEntropyCheckError;
 
     // When is an uninitialized device model that supports device setup, navigate to device onboarding.
     useEffect(() => {
@@ -120,7 +128,8 @@ export const useHandleDeviceConnection = () => {
             !isOnboardingDeviceDisconnectedAlertDisplayed &&
             !isFirmwareInstallationRunning &&
             (!isDeviceOnboardingStackFocused || isDeviceOnboardingConnectAndUnlockScreenFocused) &&
-            !wasDeviceOnboardingCancelled
+            !wasDeviceOnboardingCancelled &&
+            !shouldNavigateToDeviceCompromisedModal
         ) {
             navigation.navigate(RootStackRoutes.DeviceOnboardingStack, {
                 screen: DeviceOnboardingStackRoutes.UninitializedDeviceLanding,
@@ -140,6 +149,7 @@ export const useHandleDeviceConnection = () => {
         isOnboardingDeviceDisconnectedAlertDisplayed,
         isDeviceOnboardingConnectAndUnlockScreenFocused,
         wasDeviceOnboardingCancelled,
+        shouldNavigateToDeviceCompromisedModal,
     ]);
 
     // At the moment when unauthorized physical device is selected,
@@ -200,7 +210,7 @@ export const useHandleDeviceConnection = () => {
             // TODO: this hook is getting very complex, and it's hard to understand the logic when it navigates there and back again.
             //  Ideally there'd be a single source of truth, a function returning "where we should be as per current state"
             //  rather than multiple useEffects with imperative instructions "go there when X changes"
-            if (isDeviceCompromisedModalFocused || isSuspiciousDeviceScreenFocused) {
+            if (shouldKeepDeviceCompromisedModal || isSuspiciousDeviceScreenFocused) {
                 return;
             }
 
@@ -225,7 +235,7 @@ export const useHandleDeviceConnection = () => {
         navigation,
         shouldBlockSendReviewRedirect,
         isFirmwareInstallationRunning,
-        isDeviceCompromisedModalFocused,
+        shouldKeepDeviceCompromisedModal,
         isSuspiciousDeviceScreenFocused,
         isOnboardingStackFocused,
         isDeviceOnboardingStackFocused,
--- a/suite-native/device/src/hooks/useReportDeviceCompromised.ts
+++ b/suite-native/device/src/hooks/useReportDeviceCompromised.ts
@@ -8,9 +8,9 @@ import { getFirmwareVersion } from '@trezor/device-utils';
 
 import { revisionCheckErrorScenarios } from '../config/firmware';
 
-const reportCheckFail = (
-    checkType: 'Firmware revision',
-    contextData: any,
+export const reportCheckFail = (
+    checkType: 'Entropy' | 'Firmware revision',
+    contextData: Record<string, any>,
     errorPayload?: unknown,
 ) => {
     const payloadLabel = `${checkType} check failed!`;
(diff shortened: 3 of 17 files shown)


suite-native/device/src/hooks/useReportDeviceCompromised.ts
```diff
@@ -9,9 +9,9 @@ import { getFirmwareVersion } from '@trezor/device-utils';
 
 import { revisionCheckErrorScenarios } from '../config/firmware';
 
-const reportCheckFail = (
-    checkType: 'Firmware revision',
-    contextData: any,
+export const reportCheckFail = (
+    checkType: 'Entropy' | 'Firmware revision',
+    contextData: Record<string, any>,
```
Lemonexe, line 13: TODO unify between native and desktop in #18371 
At first I wanted to do it next commit, but it's quite a rabbit hole of tech debt. There is a lot of duplication between Suite | Suite Lite, which must be unified in order for this to make sense.

suite-native/device/src/deviceThunks.ts
```diff
@@ -67,11 +73,27 @@ export const createAndBackupWalletThunk = createThunk(
...
+                vendor: device?.features?.fw_vendor,
+            };
+            reportCheckFail('Entropy', contextData, result.payload.error);
+            // TODO: temporary exception to avoid false positives, see also /packages/suite/src/actions/settings/deviceSettingsActions.ts
+            if (result.payload.error !== 'device disconnected during action') {
+                dispatch(deviceActions.setEntropyCheckFail(device.id));
+            }
+        }
+
+        return result;
```
Lemonexe, line 95: The whole thunk is different for desktop and native. Hmm, @PeKne Could the `createAndBackupWalletThunk` be unified with `resetDevice` thunk in desktop and live in suite-common?

Not necessary, but at least I should probably extract the common part that processes `result` (the entropy check part). This logic is shared entirely, but I should unify  [reportCheckFail first](https://github.com/trezor/trezor-suite/pull/18364/files#r2042527980)

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -97,10 +99,16 @@ export const useHandleDeviceConnection = () => {
...
+
+    // any failing check should navigate to the DeviceCompromisedModal
     const shouldNavigateToDeviceCompromisedModal =
-        hasFirmwareAuthenticityCheckHardFailed &&
-        !isFirmwareAuthenticityCheckDismissed &&
+        (shouldDisplayEntropyCheckError || shouldDisplayFirmwareAuthenticityError) &&
         isOnboardingFinished;
+    // but the DeviceCompromisedModal shall not be persistent for Entropy check, because you cannot exit the modal via normal means
+    const shouldKeepDeviceCompromisedModal =
+        isDeviceCompromisedModalFocused && !shouldDisplayEntropyCheckError;
```
Lemonexe, line 117: On mobile, FW **revision** check :ghost: modal persists after disconnecting device, which creates slightly more complexity, because that persistence does not make sense for Entropy check:
- entropy check can only fail during _fresh wallet creation_ (not even restoring backup)
- so it does not make sense to be able to close the modal
  - wallet is fresh, so there are no funds that the user should want to transfer away in emergency. So there is just no point in interacting neither with the wallet, nor with the device itself
- the :ghost: modal on desktop is **not dismissible** but you can make it go away by disconnecting device
  - then you may connect a good device and Suite works fine..
  - but it remembers the bad device, so you get :ghost: modal when you connect it again
- here I made it work like desktop
  - if the modal was non-dismissible && persistent on device disconnection, it'd be a UX deadend

suite-native/device/src/selectors.ts
```diff
@@ -199,6 +200,26 @@ export const selectHasFirmwareAuthenticityCheckHardFailed = (state: FwAuthentici
...
 
+export const selectIsEntropyCheckEnabled = (state: FwAuthenticityCheckState) => {
+    const isDeviceConnectEnabled = selectIsFeatureFlagEnabled(
+        state,
+        FeatureFlag.IsDeviceConnectEnabled,
+    );
+    const isLocalFeatureFlagEnabled = selectIsFeatureFlagEnabled(
+        state,
+        FeatureFlag.IsEntropyCheckEnabled,
+    );
```
Lemonexe, line 211: I implemented local feature flag for dev purposes.
At desktop you can similarly turn off entropy check _only in Debug settings_.
But maybe it's not necessary – if you want to play around with a fake device, you can modify codebase :shrug: 
@matejkriz should I have it here or remove?

suite-native/device/src/deviceThunks.ts
PeKne: Theoretically, we could unify it, but I'm not sure if it makes sense in this case. On the desktop there are some actions dispatched that are unknown for mobile `dispatch(notificationsActions.addToast)` and `dispatch(deviceActions.setEntropyCheckFail(device.id))` (this one will be migrated to common by you?). 

Another thing is that on desktop you always skip the wallet backup at this step, and then allow user to create backup with `TrezorConnect.backupDevice`. On mobile the backup is always required already during the `TrezorConnect.resetDevice` call, but this could be parametrized as well. 

I agree that we should share as much code as possible, but I would also like to avoid situations where we will have very complicated parametrized thunks that are prone to errors. But if it would make the entropy check easier and more clear, I agree with you and we should unify. What do you think?

Lemonexe: Right, I was just asking. If there are UX different flows on Desktop/Mobile as you say, than the abstraction for `createAndBackupWalletThunk & resetDevice` would be overly complicated.

The entropy check code is duplicated after my PR, so I intend to extract it as a separate thunk :slightly_smiling_face:
But there are requirements for that first, so [I made an issue](https://github.com/trezor/trezor-suite/issues/18371) to not bloat this PR.

suite-native/module-authenticity-checks/src/screens/EntropyCheckFailModalScreen.tsx
```diff
@@ -0,0 +1,9 @@
+import { HELP_CENTER_ENTROPY_CHECK_URL } from '@trezor/urls';
+
+import { DeviceCompromisedModalContent } from '../components/DeviceCompromisedModalContent';
+
+const supportUrlWithChat = `${HELP_CENTER_ENTROPY_CHECK_URL}#open-chat`;
```
Lemonexe: the same screen will probably be kept for mobile & desktop.
@hynek-jina believes that we should eventually unify it for FW revision check too (that it was excessive to create two different pages). Let's see :shrug: 

This URL may be final, or a perfectly good placeholder at the very least.

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -103,10 +105,17 @@ export const useHandleDeviceConnection = () => {
...
+
+    const shouldDisplayFirmwareAuthenticityError =
+        hasFirmwareAuthenticityCheckHardFailed && !isFirmwareAuthenticityCheckDismissed;
+
+    // any failing check should navigate to the DeviceCompromisedModal
     const shouldNavigateToDeviceCompromisedModal =
-        hasFirmwareAuthenticityCheckHardFailed &&
-        !isFirmwareAuthenticityCheckDismissed &&
+        (shouldDisplayEntropyCheckError || shouldDisplayFirmwareAuthenticityError) &&
         isOnboardingFinished;
```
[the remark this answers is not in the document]
Lemonexe: Irrelevant I think: entropy check may not spontaneously happen during Suite Lite onboarding (`isOnboardingFinished`), it only happens during Device onboarding, which is completely different area.

suite-native/module-authenticity-checks/src/screens/EntropyCheckFailModalScreen.tsx
```diff
@@ -0,0 +1,9 @@
...
+
+export const EntropyCheckFailModalScreen = () => (
```
PeKne: Could you please move this out of the `screens` directory are rename it to something like: `EntropyCheckFailModalScreen`? In this folder and named with sufix `...Screen` should be only the top level components that you can navigate to. 

Please do the same for the `FirmwareAuthenticityCheckFailModalScreen` as well.

suite-native/device/src/selectors.ts
```diff
@@ -199,6 +200,22 @@ export const selectHasFirmwareAuthenticityCheckHardFailed = (state: FwAuthentici
     return isRevisionHardError;
 };
 
+export const selectIsEntropyCheckEnabled = (state: FwAuthenticityCheckState) => {
+    const isDeviceConnectEnabled = selectIsFeatureFlagEnabled(
```
PeKne: Is check for `isDeviceConnectEnabled` needed? How could user get to some device check error, if he is using device where connection is not possible?

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -103,10 +105,17 @@ export const useHandleDeviceConnection = () => {
...
     );
+
+    const shouldDisplayFirmwareAuthenticityError =
+        hasFirmwareAuthenticityCheckHardFailed && !isFirmwareAuthenticityCheckDismissed;
+
+    // any failing check should navigate to the DeviceCompromisedModal
     const shouldNavigateToDeviceCompromisedModal =
-        hasFirmwareAuthenticityCheckHardFailed &&
-        !isFirmwareAuthenticityCheckDismissed &&
+        (shouldDisplayEntropyCheckError || shouldDisplayFirmwareAuthenticityError) &&
```
PeKne: nit: the `&& isOnboardingFinished` can be removed. It is mentioned in the top of the useEffects in the if statement next to the `shouldNavigateToDeviceCompromisedModal`.

suite-native/device/src/deviceThunks.ts
```diff
@@ -52,23 +57,40 @@ const getResetDeviceConfig = (walletBackupType: WalletBackupType): PROTO.ResetDe
...
+
+        if (!result.success && result.payload.code === 'Failure_EntropyCheck') {
+            const contextData = {
+                model: device?.features?.internal_model,
+                revision: device?.features?.revision,
+                version: getFirmwareVersion(device),
+                vendor: device?.features?.fw_vendor,
+            };
+            reportCheckFail('Entropy', contextData, result.payload.error);
+            // TODO: temporary exception to avoid false positives, see also /packages/suite/src/actions/settings/deviceSettingsActions.ts
```
matejkriz: There is nothing more permanent than a temporary exception, right? 😆  

Why do you see this as temporary? We can't report disconnected device as hard failure, right? We should add some retry option?

Lemonexe: Oh yeah good point. I'll discuss with @komret , who implemented it on Desktop, but had to exclude this case because of a lot of false positves.
We should link issue in code instead of just "TODO"

komret: Issue here: https://github.com/trezor/trezor-suite-private/issues/135

suite-native/device/src/deviceThunks.ts
```diff
@@ -52,23 +57,40 @@ const getResetDeviceConfig = (walletBackupType: WalletBackupType): PROTO.ResetDe
 
 export const createAndBackupWalletThunk = createThunk(
```
PeKne, line 55: Please add handling of failed entropy check to the `WalletCreationScreen` where is this thunk called. In case that it fails it is reruned automatically so user can try again. But it should not be the case for failed entropy check.

suite-native/module-authenticity-checks/src/screens/EntropyCheckFailModalScreen.tsx
```diff
@@ -0,0 +1,9 @@
...
+    <DeviceCompromisedModalContent contactSupportUrl={supportUrlWithChat} />
```
PeKne: When the screenHeader is not provided, there is missing top padding. The content of the screen is too high.

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
```diff
@@ -1,98 +1,19 @@
...
-    RootStackRoutes.AppTabs,
-    RootStackParamList
->;
+import { EntropyCheckFailModalScreen } from './EntropyCheckFailModalScreen';
+import { FirmwareAuthenticityCheckFailModalScreen } from './FirmwareAuthenticityCheckFailModalScreen';
 
+/**
+ * The very similar modal can be displayed for entropy check failure or FW authenticity check failure
+ */
 export const DeviceCompromisedModalScreen = () => {
```
PeKne, line 11: Could you please add a handling of back button press? Because now if you get to the entropy check failed screen and press back button on Android, you can get back to the previous screen. I think that it should not be allowed.

You can do it the same way is implemented here: https://github.com/trezor/trezor-suite/blob/64cc41292a9f767c7fdc8c0b006432d20c04b9e2/suite-native/module-device-onboarding/src/components/DeviceOnboardingScreenWithExitButton.tsx#L28-L39

suite-native/module-authenticity-checks/src/screens/EntropyCheckFailModalScreen.tsx
PeKne: Screenshot_2025-04-24-13-37-01-66_f0e0ef16e89dc3099c9d2ea44a0bcfa0

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
PeKne: current behavior: 
https://github.com/user-attachments/assets/a1a256f2-4ec9-400d-b220-e6f78da9b0ab

PeKne (changes requested):
I have some requests, but the core works as expected. Tested on physical Android.

suite-native/module-authenticity-checks/src/screens/EntropyCheckFailModalScreen.tsx
Lemonexe: Good call. I thought it is ok like that, but yeah I agree

suite-native/device/src/deviceThunks.ts
Lemonexe: Addressed in fe8ebe1b

suite-native/module-authenticity-checks/src/screens/EntropyCheckFailModalScreen.tsx
Lemonexe: d30e69ed

suite-native/device/src/selectors.ts
Lemonexe: Yeah, you're right. Removed in c25f9f9812

suite-native/device/src/hooks/useHandleDeviceConnection.ts
Lemonexe: Ah yes, I see that condition was added later. Addressed in 72d8fd5d

suite-native/device/src/deviceThunks.ts
Lemonexe: Oh, so that's why! I was surprised that Trezor initiates wallet creation again, but then forgot to investigate that while doing other things.
Addressed that in 3c08606b45

suite-native/module-authenticity-checks/src/screens/EntropyCheckFailModalScreen.tsx
Lemonexe: c1a1ff6a6d

suite-native/module-authenticity-checks/src/screens/DeviceCompromisedModalScreen.tsx
Lemonexe: Good catch. Fixed in d04b0c53c9
I don't know if this behavior will be final, maybe the design team will specifys ome Close behavior (I could imagine disconnecting the device, that'd work), but if it stays, this change patches a hole.

suite-native/navigation/src/components/ScreenHeader.tsx
```diff
@@ -49,7 +49,9 @@ export const ScreenHeader = ({
     return (
         <Box style={applyStyle(headerStyle)}>
             <Box style={applyStyle(iconWrapperStyle)} testID="@screen/sub-header/icon-left">
-                {leftIcon || (
+                {leftIcon !== undefined ? (
+                    leftIcon
+                ) : (
```
Lemonexe, line 54: this enables explicitly disabling the icon

suite-native/device/src/selectors.ts
```diff
@@ -199,6 +200,12 @@ export const selectHasFirmwareAuthenticityCheckHardFailed = (state: FwAuthentici
     return isRevisionHardError;
 };
 
+export const selectIsEntropyCheckEnabled = (state: FwAuthenticityCheckState) =>
```
PeKne: nit: this selector is probably not needed anymore. `selectIsFeatureEnabled(state, Feature.entropyCheckMobile, true)` could be called directly in the `selectIsEntropyCheckEnabledAndFailed`.

PeKne (approved):
Thanks for reacting to my comments. :shipit:

suite-native/device/src/selectors.ts
Lemonexe: Yes, I've thought about that, but I wanted to leave a single point, as done on desktop. But now looking again, the selector `selectIsEntropyCheckEnabled` on desktop does something different, and the message system is selected outside of that, so I'm gonna remove it as you suggest to unify it.

PR #18197, opened 2025-04-08
refactor(suite): remove deprecated icons

seibei-iguchi:

PR #18368, opened 2025-04-15
ci: remove release coins icons workflow

mroz22:
I guess this one should either get removed or updated as it is failing in nighty CI https://github.com/trezor/trezor-suite/actions/runs/14458423648/job/40546388535 probably after this one #18197 getting merged.

cc @seibei-iguchi  also maybe @MiroslavProchazka knows something about it 🤷


komret (approved):

vdovhanych:
why do you remove this when there is a pr that is fixing this workflow? 
https://github.com/trezor/trezor-suite/pull/18473

PR #18386, opened 2025-04-15
Modals, afuera!

### LINKED ISSUE #17422: Finish modals refactoring

seibei-iguchi:
Afuera!


packages/components/src/components/Box/Box.tsx
```diff
@@ -38,36 +38,56 @@ type AllowedFrameProps = Pick<FrameProps, (typeof allowedBoxFrameProps)[number]>
...
 `;
 
+type BorderWidth =
+    | {
+          top?: BorderWidths;
+          bottom?: BorderWidths;
+          left?: BorderWidths;
+          right?: BorderWidths;
+          horizontal?: BorderWidths;
+          vertical?: BorderWidths;
```
peter-sanderson, line 83: can we do something like:
```ts
{
          top?: BorderWidths;
          bottom?: BorderWidths;
          left?: BorderWidths;
          right?: BorderWidths;
} | {
         horizontal?: BorderWidths;
          vertical?: BorderWidths;
}
```

seibei-iguchi: That's a step back.. we wanna do something like { horizontal: foo, top: bar }

peter-sanderson: Ah ok, that would be super-complicated to create some smart type.

packages/suite/src/components/guide/GuideRouter.tsx
```diff
@@ -17,98 +17,78 @@ import {
...
+            return <GuideArticle />;
+        case 'GUIDE_CATEGORY':
+            return <GuideCategory />;
+        case 'SUPPORT_FEEDBACK_SELECTION':
+            return <SupportFeedbackSelection />;
+        case 'FEEDBACK_BUG':
+            return <Feedback type="BUG" />;
+        case 'FEEDBACK_SUGGESTION':
+            return <Feedback type="SUGGESTION" />;
+        default:
```
peter-sanderson: Lets not fallback to default, but do a:
```ts
default: {
            const _unhandledCase: never = activeView;
            throw new Error(`Unhandled activeView type: ${_unhandledCase}`);
        }
```

peter-sanderson (approved):
`Modal` is dead, long live the `Modal`. :grin: Good job!

packages/suite/src/views/suite/bridge/index.tsx
```diff
@@ -57,52 +32,49 @@ export const BridgeUnavailable = () => {
...
+                </H3>
+                <Paragraph variant="tertiary">
+                    {hasTransport ? description : <Translation id="TR_BRIDGE_NEEDED_DESCRIPTION" />}
+                </Paragraph>
+                <Paragraph variant="tertiary">
+                    Alternatively you may{' '}
+                    <Link variant="underline" href={target}>
+                        download
+                    </Link>{' '}
+                    a standalone Trezor Bridge binary.
```
jvaclavik, line 75: I know it was here before, but shouldn't we add a translation here?

jvaclavik (approved):
Well done Adam!!! I'm glad we finally can get rid of old modals. Good job 💪 PR LGTM

Issue #18406, opened 2025-04-16
Update fee rate regularly while being in send form


hynek-jina:
User can remain in send form for a while and fee rate become outdated - which results in suboptimal fee rate used.
- Check fee rate regularly every minute (or two) while being in send form
  - EDIT: 60s interval was implemented
- ~~Offer option to~~ Refresh the fee rates when Suite gets newer data

[Figma design here](https://www.figma.com/design/qx5aOavcU7I2MW2PtnQtfz/Small-Things--DSK-?node-id=4048-978&t=MFI4WubrD2PkITeY-1) – shall be done consistently with mobile where it's already implemented

---

![Image](https://github.com/user-attachments/assets/2f5578b9-a80a-4557-bce3-11661696d718)

bosomt:
**QA OK**

i see that fee is changing from time to tome when you keep send BTC page open and request updateFee is sent every minute

<img width="527" alt="Image" src="https://github.com/user-attachments/assets/ecdbbbaa-1948-4559-a52c-f5650a0f3261" />

**Info:**
 - Suite version: web 25.6.0 (a547ecb0691f2a7df78d1bf0bb6b97f4615de4e1)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2T1 2.8.10 regular (revision 2a65d18200580005dc419b9569ed97fae440806a)
 - Transport: BridgeTransport 3.0.0-bundled.25.6.0

Lemonexe:
Reopened because #19242 was reverted.

bosomt:
**QA OK**

- remained in Send form, noticed that fees are changing ✅ 
- fees are not changing when sending all ✅ 

**Info:**
 - Suite version: web 25.7.0 (2b58a118407d9e8a2e79b4d5f463f81b96b0fd1b)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3T1 2.8.10 bitcoin-only (revision 2a65d18200580005dc419b9569ed97fae440806a)
 - Transport: WebUsbTransport

PR #18408, opened 2025-04-16
feat(connect): remove estimateFee caching

### LINKED ISSUE #17977: Bitcoin: fee can be updated only every 20 minutes
**Describe the bug**

Check `packages/connect/src/backend/Blockchain.ts`, `estimateFee` method, there is `useCache`

In case you want to update fee in send form while network is on same block, please also check hook in `useSendForm` `// replace default feeInfo with data loaded from the server`

Lemonexe:
## Description

:eyes: CR commit by commit recommended..

- Remove caching behavior for `Blockchain.estimateFee`, so that Suite can update fees at will.
- But the caching was in place [because of compose transaction for bitcoin-like coins](https://github.com/trezor/trezor-suite/pull/18408/files#r2054536369)
  - Because composeTransaction calls blockbook needlessly, the caching was a "fix" to solve the symptoms
  - so let's rather fix the cause and stop updating fees when they are there already
- No breaking change in Connect: `TrezorConnect.composeTransaction` will call backend if fees are not fetched yet 
- But in order to do what I just described, I needed to organize FeeLevels classes into singletons, because currently, `BitcoinFeeLevels` is recreated everytime you call `TrezorConnect.composeTransaction`. In other words, anytime you change anything in Send Form :see_no_evil:

## Related Issue

Resolve  #17977

Followup #18406 to update the fees in Suite in regular intervals (currently only updated when opening Send Form)

## QA

You may verify in websocket monitor that:
- Navigating from Send form & reopening it updates the fees
- Working in the Send form does not update the fees

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/tweak-updating-tx-fees&lookback=7)


tomasklim (approved):

mroz22:
@szymonlesisz do you remember any hard requirements why we had this?

packages/connect/src/backend/__tests__/Blockchain.test.ts
Lemonexe, line 1: This test emphasized the caching behavior. Without it, the test would be very trivial.

Lemonexe:
Closing, discovered that the caching is needed, so I will remake it and explain better.

packages/connect/src/api/bitcoin/TransactionComposer.ts
```diff
@@ -72,8 +73,6 @@ export class TransactionComposer {
     async init(blockchain: Blockchain) {
         const { blockHeight } = await blockchain.getNetworkInfo();
         this.blockHeight = blockHeight;
-
-        await this.feeLevels.load(blockchain);
```
Lemonexe, line 76: Because TransactionComposer.init is called [on every precompose & selectFee](https://github.com/trezor/trezor-suite/blob/f9998f3ccfd9390ad7dcbd909a5b9519593716af/packages/connect/src/api/composeTransaction.ts#L180), this line meant that everytime you change anything in Send Form, fees get refreshed from blockbook. That spams blockbook with unnecessary requests.

martykan:
> This means a small breaking change for Connect: you have to call TrezorConnect.blockchainEstimateFee yourself, before you proceed to TrezorConnect.composeTransaction, because the latter does not load current fee rates, so it would just default to the hardcoded constants

I really don't like this idea, Connect methods shouldn't have global shared state like that. 
If we want composeTransaction use already loaded fees, let's pass them as an argument.

packages/connect/src/api/bitcoin/TransactionComposer.ts
```diff
@@ -47,7 +48,7 @@ export class TransactionComposer {
         this.blockHeight = 0;
         this.baseFee = options.baseFee || 0;
         this.sortingStrategy = options.sortingStrategy;
-        this.feeLevels = new BitcoinFeeLevels(options.coinInfo);
+        this.feeLevels = getOrInitFeeLevels(options.coinInfo) as BitcoinFeeLevels;
```
Lemonexe: Because a new instance was created for each `TrezorConnect.composeTransaction`, the fees loaded from previously were lost

packages/connect/src/api/bitcoin/TransactionComposer.ts
```diff
@@ -73,7 +74,9 @@ export class TransactionComposer {
         const { blockHeight } = await blockchain.getNetworkInfo();
         this.blockHeight = blockHeight;
 
-        await this.feeLevels.load(blockchain);
+        if (!this.feeLevels.wasFetchedSuccessfully) {
+            await this.feeLevels.load(blockchain);
+        }
```
Lemonexe, line 79: Because TransactionComposer.init is called [on every precompose & selectFee](https://github.com/trezor/trezor-suite/blob/f9998f3ccfd9390ad7dcbd909a5b9519593716af/packages/connect/src/api/composeTransaction.ts#L180), and the `BitcoinFeeLevel` was recreated with each call,  this line meant that everytime you change anything in Send Form, fees would get refreshed from blockbook. That spams blockbook with unnecessary requests.
Now the old behavior is kept: it will be fetched if feeLevels were not loaded (the instance from `getOrInitFeeLevels` is fresh), otherwise reuse the already fetched fees on the shared instance :slightly_smiling_face:

Lemonexe: Note that Suite calls the `TrezorConnect.blockchainEstimateFee` via `updateFeeInfoThunk` when opening Send Form – so that first `TrezorConnect.composeTransaction` call will have the fees available already.

Lemonexe:
> I really don't like this idea, Connect methods shouldn't have global shared state like that. If we want composeTransaction use already loaded fees, let's pass them as an argument.

Good call. Patched that in f7d8193884db1473b0f5cf2c6f379e46ccd18251. Old behavior is kept, [see comment](https://github.com/trezor/trezor-suite/pull/18408#discussion_r2055459142)

Lemonexe:
[connect popup e2e is broken](https://satoshilabs.slack.com/archives/G019WLX2P7B/p1745395827587149)

packages/connect/src/backend/fees/index.ts
```diff
@@ -0,0 +1,29 @@
+import { BitcoinFeeLevels } from './BitcoinFeeLevels';
+import { EthereumFeeLevels } from './EthereumFeeLevels';
+import { MiscFeeLevels } from './MiscFeeLevels';
+import type { CoinInfo } from '../../types';
+
+const instancesPerCoin: { [shortcut: CoinInfo['shortcut']]: MiscFeeLevels } = {};
+
+const feeLevelsPerTypeFactory = (coinInfo: CoinInfo): MiscFeeLevels => {
```
peter-sanderson, line 8: something like:

```ts
const feeLevelsPerTypeFactory = (coinInfo: CoinInfo): MiscFeeLevels => {
    const { type } = coinInfo;

    switch (type) {
        case 'bitcoin':
            return new BitcoinFeeLevels(coinInfo);
        case 'ethereum':
            return new EthereumFeeLevels(coinInfo);
        case 'misc':
            return new MiscFeeLevels(coinInfo);
        case 'nem':
            return new MiscFeeLevels(coinInfo);
        default: {
            const _unhandledCase: never = type;

            throw new Error(`Unhandled type: ${_unhandledCase}`);
        }
    }
};
```

Lemonexe: Normally I'd agree, but given that one of the types is `'misc'`, where most of the other coins are lumped in, I don't think it's beneficial. But ok, why not :shrug:

martykan (approved):
LGTM

packages/connect/src/api/bitcoin/TransactionComposer.ts
peter-sanderson: I dont like this, -> discussed in Slack

Lemonexe: I created a new function `getOrInitBitcoinFeeLevels` to at least centralize the type casting. I wanted to fix this in `packages/connect/src/backend/fees/index.ts` by 
 separating the instances to btcInstances, ethInstances and miscInstances, and then return from either of those array according to `coinInfo.type`. That would work normally with POJOs or unrelated classes, but not when `BitcoinFeeLevels` and `EthereumFeeLevels` extend `MiscFeeLevels`.
TS then infers that such function always returns `MiscFeeLevels` instead of distinguishing them.

PR #18421, opened 2025-04-17
chore(suite-native): Mobile Trade: refactor BuyForm to use quote directly

jbazant:
## Description
Mobile trade buy form now uses single value of `quote` instead of `paumentMethod` and `provider`.

## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=mobile-trade-refactor-form&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=mobile-trade-refactor-form&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=mobile-trade-refactor-form&lookback=7)

PR #18425, opened 2025-04-17
Mobile trade fixes

### LINKED ISSUE #18426: Fix useNetworkChangeEffect hook in useTradingBuyForm
## Issue
  
There's a TODO comment in `suite-native/module-trading/src/hooks/useTradingBuyForm.ts` that needs to be addressed.

The `useNetworkChangeEffect` hook doesn't use React's `useEffect` pattern unlike all other effect hooks in the same file. This can cause:
- Code execution on every render instead of only when dependencies change
- Potential performance issues
- Inconsistent behavior compared to other hooks

## Proposed Solution

Refactor the hook to use the `useEffect` pattern:

```typescript
const useNetworkChangeEffect = (form: TradingBuyForm) => {
    const prevNetworkId = useRef<string | undefined>(undefined);

    useEffect(() => {
        const asset = form.watch('asset');
        
        if (asset?.networkId !== prevNetworkId.current) {
            prevNetworkId.current = asset?.networkId;
            dispatch(setBuySelectedReceiveAccount({ selectedReceiveAccount: undefined }));
        }
    }, [form, dispatch]);
};
```

Or consider using form.watch() within useEffect as is done in other hooks in the file.

## Related PR
This issue was identified during PR review: https://github.com/trezor/trezor-suite/pull/18421#discussion_r2048473599

jbazant:
## Description

Various mobile trade fixes.

## Related Issue

Resolve [#18426](https://github.com/trezor/trezor-suite/issues/18426)

### DIFF
--- a/suite-native/module-trading/src/hooks/useTradingBuyForm.ts
+++ b/suite-native/module-trading/src/hooks/useTradingBuyForm.ts
@@ -27,18 +27,6 @@ import { TradingBuyForm, TradingBuyFormValues } from '../types';
 import { buyFormValidationSchema } from '../utils/buyFormValidationSchema';
 import { getSelectedSymbolFromBuyForm } from '../utils/tradeableAssetUtils';
 
-const useNetworkChangeEffect = (form: TradingBuyForm) => {
-    const dispatch = useDispatch();
-    const prevNetworkId = useRef<string | undefined>(undefined);
-
-    const asset = form.watch('asset');
-
-    if (asset?.networkId !== prevNetworkId.current) {
-        prevNetworkId.current = asset?.networkId;
-        dispatch(setBuySelectedReceiveAccount({ selectedReceiveAccount: undefined }));
-    }
-};
-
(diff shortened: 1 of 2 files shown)


suite-native/module-trading/src/hooks/useTradingBuyForm.ts
```diff
@@ -73,10 +64,18 @@ const useAmountAndCurrencyFieldsChangeEffect = (form: TradingBuyForm) => {
                     form.trigger(['cryptoValue', 'fiatValue']);
                     break;
 
-                case 'asset':
+                case 'asset': {
                     form.setValue('cryptoValue', undefined);
                     form.trigger('cryptoValue');
```
vytick, line 69: user could have originally inserted `cryptoValue`, what would you think about clearing even `fiatValue` in this case?

vytick: resolving as this will be resolved in #18421

vytick (approved):
🚀

PR #11523, opened 2024-03-08
chore(ci): speedup connect tests + explorer build

Nodonisko:
#11526 needs to be merged first to solve type-check errors.

[Build] connect / build-connect-explorer - 7 min ➡️ **3 min**
[Test] connect / build (push) - 4 min ➡️ **1 min**
[Test] connect / web / node - **reduced about ~3 minutes** from each flow run time
Sum of total run time - 3h 17m ➡️ **2h 22m (~1 hour reduction)**

## Related Issue

Resolve

PR #18432, opened 2025-04-17
Remove deprecated `node-libs-browser`

marekrjpolak:
## Description

When I tried to resolve `punycode` deprecation warnings, I've found out that we use six-years-unupdated lib `node-libs-browser` with many deprecated dependencies as well. Given that it seems [not doing anything important](https://github.com/webpack/node-libs-browser/blob/master/index.js), I've tried to replace it with its dependencies themselves, but only the used ones.

## What to discuss

- `process` was allegedly used as a polyfill (https://github.com/trezor/trezor-suite/pull/11523/files#r1519412045) so I've added it as a direct dependency
- even if `url` transitive dependency is not directly used by us, it's used in `micro-ftch` which is a dependency of `@ethereumjs/util` so it must be either a) added as an unused dependency of connect package and ignored in `.depcheckrc` (which is basically status quo), or b) added as an unresolved fallback into `base.webpack.config.ts` which I did, but it may break eth logic in some cases? Dunno.
- which team should be responsible for new dependencies (previously transitive dependencies of `node-libs-browser`)? For now I've just left them to trade as they were responsible for `node-libs-browser` anyway.

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/node-libs-browser&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/node-libs-browser&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/node-libs-browser&lookback=7)

### DIFF
--- a/suite-native/app/metro.config.js
+++ b/suite-native/app/metro.config.js
@@ -5,7 +5,6 @@ const { withRozeniteReduxDevTools } = require('@rozenite/redux-devtools-plugin/m
 const { getSentryExpoConfig } = require('@sentry/react-native/metro');
 const { withStorybook } = require('@storybook/react-native/metro/withStorybook');
 const { mergeConfig } = require('metro-config');
-const nodejs = require('node-libs-browser');
 
 const { metroSecureResolver } = require('@trezor/bundler-security/src/metroSecureResolver');
 
@@ -36,12 +35,12 @@ const config = {
         blockList: [/libDev/],
         extraNodeModules: {
             // modules needed for trezor-connect
-            crypto: nodejs.crypto,
-            stream: nodejs.stream,
-            https: nodejs.https,
-            http: nodejs.http,
-            zlib: nodejs.zlib,
-            vm: nodejs.vm,
+            crypto: require.resolve('crypto-browserify'),
+            stream: require.resolve('stream-browserify'),
+            https: require.resolve('https-browserify'),
+            http: require.resolve('stream-http'),
+            zlib: require.resolve('browserify-zlib'),
+            vm: require.resolve('vm-browserify'),
         },
         sourceExts,
         resolveRequest: (context, moduleName, platform) => {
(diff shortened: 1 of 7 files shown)


suite-native/app/metro.config.js
```diff
@@ -37,12 +36,12 @@ const config = {
...
-            https: nodejs.https,
-            http: nodejs.http,
-            zlib: nodejs.zlib,
-            vm: nodejs.vm,
+            crypto: require.resolve('crypto-browserify'),
+            stream: require.resolve('stream-browserify'),
+            https: require.resolve('https-browserify'),
+            http: require.resolve('stream-http'),
+            zlib: require.resolve('browserify-zlib'),
+            vm: require.resolve('vm-browserify'),
```
Lemonexe, line 43: Are all of them really needed for connect?

Lemonexe (approved):
Very nice, lot of dependencies removed from `yarn.lock` 🎉 
At quick glance, Web & Desktop & Native are all working :heavy_check_mark: 

I don't get this, though:

> `process` was allegedly used as a polyfill (https://github.com/trezor/trezor-suite/pull/11523/files#r1519412045) so I've added it as a direct dependency

The conversation or the code in question does not mention `process`?

marekrjpolak:
> The conversation or the code in question does not mention process?

It's this one. And I'm not sure why it's needed, but it is.
image

suite-native/app/metro.config.js
marekrjpolak: Honestly, I don't know... 🫠

Lemonexe: Well this PR maintains current status quo, and the app works, so I approved.
@matejkriz care to take a look? Maybe not everything is needed..

PR #18435, opened 2025-04-17
feat: firmware release config new package

karliatto:
## Description
This PR creates a new message that contains the latest releases for all models so it could be used by Suite to fetch those binaries and offer them for update.

* b0b939bba55cfb58bb1535d57fca2f5e0679ceee - Creates new package `connect-message-releases`
* ec74adc91f9da3f558f39a5b627afbfc728d2d79 - Fetch releases for now only local but once they are releases it will work for remote as well.
* 83b4e5a3be38fecf076871efafc86a9104695e4a - adding the new dev keys
* 78b7430e762b7b5a490769add89271ca024dbc13 - update getJWSPublicKey
* 929392f232f21dc4501cd3317f772eb57c5d3be5 - Scripts that generate the message from releases.json and sign the new created releases message and also uses JSON schema to validate the proper structure of the new created file. The idea is that this file could have some parts specially conditions that could be modified manually.
* 7994e7fd7d3d0d73a2626d8b05fcfeb68fd3b4b4 tests
* 0694939bf5dab66f13637b2e5d561887c777ea2b - new misc that test that the scripts are working properly


.github/workflows/release-message-release.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  release-message-release:
+    if: ${{ github.repository == 'trezor/trezor-suite' && (github.ref == 'refs/heads/release-message-release-production' || github.ref == 'refs/heads/release-message-release-develop') }}
+    environment: ${{ github.ref == 'refs/heads/release-message-release-production' && 'production-message-release' || 'develop-message-release' }}
+    runs-on: ubuntu-latest
+    env:
+      RELEASE_ENV: ${{ github.ref == 'refs/heads/release-message-release-production' && 'production' || 'develop' }}
+      AWS_REGION: "eu-central-1"
+      AWS_CLOUDFRONT_ID: E1ERY5K2OTKKI1
+      ROLE_TO_ASSUME: ${{ github.ref == 'refs/heads/release-message-release-production' && 'arn:aws:iam::538326561891:role/gh_actions_suite_production_message_release' || 'arn:aws:iam::538326561891:role/gh_actions_suite_develop_message_release' }}
```
karliatto: @vdovhanych I assume data in this part and probably other parts of the file will have to be different.
For now just putting this as a placeholder.

.github/workflows/release-firmware-release-config.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  release-message-release:
```
vdovhanych: this should be renamed

.github/workflows/release-firmware-release-config.yml
```diff
@@ -0,0 +1,46 @@
...
+    if: ${{ github.repository == 'trezor/trezor-suite' && (github.ref == 'refs/heads/release-firmware-release-config-production' || github.ref == 'refs/heads/release-firmware-release-config-develop') }}
+    environment: ${{ github.ref == 'refs/heads/release-firmware-release-config-production' && 'production-firmware-release-config' || 'develop-firmware-release-config' }}
```
vdovhanych: i need to first create these envs

.github/workflows/release-firmware-release-config.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-latest
+    env:
+      RELEASE_ENV: ${{ github.ref == 'refs/heads/release-firmware-release-config-production' && 'production' || 'develop' }}
+      AWS_REGION: "eu-central-1"
+      AWS_CLOUDFRONT_ID: E1ERY5K2OTKKI1
+      ROLE_TO_ASSUME: ${{ github.ref == 'refs/heads/release-firmware-release-config-production' && 'arn:aws:iam::538326561891:role/gh_actions_suite_production_firmware_release_config' || 'arn:aws:iam::538326561891:role/gh_actions_suite_develop_firmware_release_config' }}
```
vdovhanych: We can keep the names like this, I'll create the roles on AWS tomorrow with this name.

.github/workflows/release-firmware-release-config.yml
```diff
@@ -0,0 +1,46 @@
...
+          IS_CODESIGN_BUILD: ${{ env.RELEASE_ENV == 'production' && 'true' || 'false' }}
+          JWS_PRIVATE_KEY_ENV: ${{ secrets.JWS_PRIVATE_KEY_ENV }}
+        run: |
+          yarn install
+          yarn validate-release-config
+          yarn sign-release-config
+
+      - name: Upload ${{ env.RELEASE_ENV }}  firmware-release-config file
+        run: |
+          aws s3 cp packages/firmware-release-config/files/releases.v1.jws s3://data.trezor.io/releases/${{ env.RELEASE_ENV == 'production' && 'stable' || 'develop'}}/releases.v1.jws --cache-control no-cache
```
vdovhanych: This should probably be a different path. I'm not sure we want to have this in the /releases/ folder on datat.trezor.io, it should be either in the /firmware/releases/folder, but then again, we already have files there, or part of the suite path, so /suite/releases/firmware/file.jws or whatever it's named.

.github/workflows/release-firmware-release-config.yml
karliatto: Thanks! done https://github.com/trezor/trezor-suite/pull/18435/files#diff-45639c8f4273030d4c3ce8b0a6245bb2e38ed51f8ca515f30bb12594d43e61e7R15

.github/workflows/release-firmware-release-config.yml
karliatto: Ok, done https://github.com/trezor/trezor-suite/pull/18435/files#diff-45639c8f4273030d4c3ce8b0a6245bb2e38ed51f8ca515f30bb12594d43e61e7R45

It will be `data.trezor.io/suite/releases/firmware/stable/releases.v1.jws` and `data.trezor.io/suite/releases/firmware/develop/releases.v1.jws`.

packages/firmware-release-config/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "version": "1.0.0",
+    "private": true,
+    "license": "See LICENSE.md in repo root",
+    "sideEffects": false,
+    "main": "src/index",
+    "scripts": {
+        "build:lib": "yarn validate-release-config && yarn sign-release-config",
+        "depcheck": "yarn g:depcheck",
+        "lint:js": "yarn g:eslint '**/*.{ts,tsx,js}'",
+        "test:unit": "yarn g:jest",
```
mroz22, line 12: gives me. how come it doesn't fail in the CI 🤔 
image

packages/firmware-release-config/package.json
```diff
@@ -0,0 +1,29 @@
...
+        "type-check": "yarn g:tsc --build",
+        "generate-release-config": "yarn tsx ./scripts/generate-release-config.ts",
+        "sign-release-config": "yarn tsx ./scripts/sign-release-config.ts",
+        "validate-release-config": "yarn tsx ./scripts/validate-release-config.ts"
```
mroz22, line 16: it could be useful to run some of these utility scripts (if it makes sense) in [the nightly misc job](https://github.com/trezor/trezor-suite/blob/chore/connect-message-release/.github/workflows/test-misc.yml#L4) to help us prevent cases when lets say a lib update breaks things and we won't notice until release

packages/firmware-release-config/package.json
```diff
@@ -0,0 +1,29 @@
...
+    },
+    "dependencies": {
+        "@trezor/device-utils": "workspace:^",
+        "@trezor/env-utils": "workspace:^",
+        "fs-extra": "^11.3.0",
+        "jws": "^4.0.0",
+        "tsx": "^4.19.3"
+    },
+    "devDependencies": {
+        "@types/fs-extra": "^11",
```
mroz22: other packages in the monorepo have this version pinned as follows 
```
"@types/fs-extra": "^11.0.4",
```
but it probably does not matter

packages/firmware-release-config/package.json
karliatto: The CI uses nx that runs `build:lib` for this package before.
It is configure here: https://github.com/trezor/trezor-suite/pull/18435/files#diff-15552e50b1b7c05b05b7ffe08ee47b5ed62b8d2039229c58972a42fd22a7381cR28

packages/firmware-release-config/releases/releases.v1.json
```diff
@@ -0,0 +1,221 @@
...
+                },
+                "release": {
+                    "required": false,
+                    "version": [1, 13, 0],
+                    "min_bootloader_version": [1, 12, 0],
+                    "min_firmware_version": [1, 12, 0],
+                    "bootloader_version": [1, 12, 1],
+                    "translations": [],
+                    "firmware_revision": "592590cf66a9b62dfeee7e4d2afb6e01005e5b2c"
+                }
```
mroz22: how about changelog?

packages/firmware-release-config/schema/releases.schema.v1.json
```diff
@@ -0,0 +1,144 @@
...
+                                    }
+                                },
+                                "firmware_revision": {
+                                    "type": "string"
+                                },
+                                "changelog": {
+                                    "type": "string"
+                                }
+                            },
+                            "required": [
```
mroz22, line 101: the aforomention changelog - it appears to be required in types https://github.com/trezor/trezor-suite/blob/chore/connect-message-release/packages/connect/src/types/firmware.ts#L16, the same applies for `fingerprint`. are they needed?

packages/firmware-release-config/src/types.ts
```diff
@@ -0,0 +1,62 @@
...
+export interface ReleaseInfo {
+    required: boolean;
+    version: VersionArray;
+    bootloader_version: VersionArray;
+    min_firmware_version: VersionArray;
+    min_bootloader_version: VersionArray;
+    translations: string[];
+    firmware_revision: string;
+    changelog: string;
+}
```
mroz22: sure thing we don't want to import anything from connect, but I can see some duplicities in types.

karliatto:
@vdovhanych I created new PR https://github.com/trezor/trezor-suite/pull/18575 to separate the code for the workflow from the rest of the code. Making PRs smaller and easier to review.

packages/firmware-release-config/releases/releases.v1.json
karliatto: Yes, we need it adding it, it is displayed in suite when offering update https://github.com/trezor/trezor-suite/pull/18435/commits/367971335396b9ccbee933af4521e11194d4d2a0

packages/firmware-release-config/schema/releases.schema.v1.json
karliatto: yes adding it here https://github.com/trezor/trezor-suite/pull/18435/commits/367971335396b9ccbee933af4521e11194d4d2a0

packages/firmware-release-config/package.json
karliatto: Makes sense, adding that test https://github.com/trezor/trezor-suite/pull/18435/commits/edf65c2b1e8312f3832a364a039155eb7b15ee8f

packages/firmware-release-config/package.json
karliatto: Good catch! Changed it https://github.com/trezor/trezor-suite/pull/18435/commits/363c3947f3319e0e00df6ed5477854f5a99696bd#diff-cd5947ee61c2e487be7b565603f4b34fd2fedcfe63ff46b9beaa140ef16a7aadR26

packages/firmware-release-config/src/types.ts
karliatto: What about having that sort of types in connect-common so it can be imported by `connect` and `firmware-release-config` packages and avoid circular importing dependencies?

karliatto:
@mroz22 I rebased this branch on develop to keep it updated.

packages/firmware-release-config/src/types.ts
mroz22: connect-common doesn't sound too bad for this.

packages/firmware-release-config/jest.config.js
```diff
@@ -0,0 +1,5 @@
+const { ...baseConfig } = require('../../jest.config.base');
+
+module.exports = {
+    ...baseConfig,
+};
```
mroz22, line 5: nitcpick, when I am reading the PR by commits, this commit does not contain any test - only this config. The test is probably included in some other commit

mroz22 (approved):
looks good.  maybe I would suggest getting rid of duplicity types definitions before merging.

karliatto:
> looks good. maybe I would suggest getting rid of duplicity types definitions before merging.

I did it under `device-utils` package since some related types are already there so it made sense to me.

PR #18478, opened 2025-04-22
18253 shaking of device

### LINKED ISSUE #18253: Shift focus to the newly connected device
We’ve learned that the way we try to draw users’ attention to the wallet switcher is too subtle and often goes unnoticed—the wallet switcher only gives a slight shake.

At the same time, it’s been overused in situations where it’s not really necessary, which further diminishes its impact (especially when it appears during moments where users think, “meh, nothing important happened”).

Image

Please apply the following changes:
	1.	**Remove the wallet switcher shaker** from all current cases.
	2.	I**nstead, open the wallet switcher with a smooth animation** when a new device is connected. This is the one case where users have clearly expressed the need to interact with the device (e.g. when connecting it and entering the PIN).

vojtatranta:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18253

Comments about the issue here: https://github.com/trezor/trezor-suite/issues/18253#issuecomment-2821217915

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=18253-shaking-of-device&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=18253-shaking-of-device&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=18253-shaking-of-device&lookback=7)

### DIFF
--- a/packages/suite/src/actions/suite/initAction.ts
+++ b/packages/suite/src/actions/suite/initAction.ts
@@ -7,9 +7,11 @@ import {
     initDevices,
     periodicCheckStakeDataThunk,
     periodicFetchFiatRatesThunk,
+    selectDevices,
     updateMissingTxFiatRatesThunk,
 } from '@suite-common/wallet-core';
 import * as walletConnectActions from '@suite-common/walletconnect';
+import { DEVICE } from '@trezor/connect';
 import { isDesktop } from '@trezor/env-utils';
 import { desktopApi } from '@trezor/suite-desktop-api';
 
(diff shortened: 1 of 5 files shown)


suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -18,157 +19,163 @@ const CONNECT_INIT_MODULE = '@common/connect-init';
...
+
+    // set event listeners and dispatch as
+    TrezorConnect.on(DEVICE_EVENT, ({ event: _, ...eventData }) => {
+        if (eventData.type === DEVICE.CONNECT || eventData.type === DEVICE.CONNECT_UNACQUIRED) {
+            // This special case here allows us to "inject" extra data into action's payload
+            // and change the type of the action (in this case DeviceEvent type !== Redux Action type)
+            const connectedDevices = selectDevices(getState());
+            dispatch(deviceConnectThunks({ type: eventData.type, device: eventData.payload }));
+
+            connectInitHooks?.[eventData.type]?.(eventData.payload, connectedDevices);
```
vojtatranta, line 50: this is not great, but I had to pass the hooks here and previous state there , because this module is reused in the native app and hence I can't call things like `goto` here as this is not possible on the mobile.

suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -18,157 +20,163 @@ const CONNECT_INIT_MODULE = '@common/connect-init';
...
-
-        TrezorConnect.on(TRANSPORT_EVENT, ({ event: _, ...action }) => {
-            // dispatch event as action
-            dispatch(action);
-        });
-
-        TrezorConnect.on(BLOCKCHAIN_EVENT, ({ event: _, ...action }) => {
+export const connectInitThunk = createThunk<
+    void,
+    | {
```
vojtatranta, line 25: Unfortunatell,y I didn't see a better way how to "hook" into this process here. I really, really wanted to avoid using middleware. IMHO this is better than making the middleware hell worse.

In the ideal world, this would be modular.

packages/suite/src/actions/suite/initAction.ts
```diff
@@ -74,7 +76,31 @@ export const init = () => async (dispatch: Dispatch, getState: GetState) => {
...
+            trezorConnectActions.connectInitThunk({
+                [DEVICE.CONNECT]: (_device, prevConnectedDevices) => {
+                    const previouslyConnectedDevices = prevConnectedDevices.filter(
+                        device => device.connected,
+                    );
+                    const currentConnectedDevices = selectDevices(getState()).filter(
+                        device => device.connected,
+                    );
+
+                    if (
```
vojtatranta, line 89: It would be cool if I could say that I don't want have the `device connected` triggered "on initial" run. when the connect is initialized for the first time.

Unfortunatel,y there isn't yet such thing.

vojtatranta: Such thing would an internal state of the connect, it would be cool to have a method for it.

jvaclavik (approved):

PR #18497, opened 2025-04-22
Fix/import transaction

### LINKED ISSUE #18495: TypeError: Cannot read properties of undefined (reading 'toLowerCase')
Sentry Issue: [TREZOR-SUITE-11WF](https://satoshilabs.sentry.io/issues/6226750394/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
  at currencies (@trezor/suite-build/../suite/src/hooks/wallet/useSendFormImport.ts:44:57)
  at <object>.importTransaction (@trezor/suite-build/../suite/src/hooks/wallet/useSendFormImport.ts:44:35)
  at <object>.loadTransaction (@trezor/suite-build/../suite/src/hooks/wallet/useSendForm.ts:234:15)
```

komret:
## Description

Commits 1 & 2: render CSV example as `pre` so that when it is copied, it retains line-breaks
Commit 3: add toast to notify the user that the CSV input could not be parsed
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18495

#### Before:
Example was copied as one line and couldn't be parsed - confusing:

https://github.com/user-attachments/assets/f0c6adfa-7348-4f63-acfd-69bf99b11782

Failed to parse, yet no error shown - confusing and also created an unhandled error that got to Sentry:

https://github.com/user-attachments/assets/82531af8-a099-438a-ba68-da6b4c331236

#### After:

https://github.com/user-attachments/assets/f5ea9e46-a91c-4155-81fa-362fc3b278c1

https://github.com/user-attachments/assets/98734672-67d6-4008-b6a7-d3eacddeecf5

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/import-transaction&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/import-transaction&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/import-transaction&lookback=7)

### DIFF
--- a/packages/components/src/components/typography/Paragraph/Paragraph.tsx
+++ b/packages/components/src/components/typography/Paragraph/Paragraph.tsx
@@ -1,8 +1,10 @@
 import React from 'react';
 
+import { Without } from '@trezor/type-utils';
+
 import { Text, TextProps } from '../Text/Text';
 
-export const Paragraph = ({ children, ...rest }: TextProps) => (
+export const Paragraph = ({ children, ...rest }: Without<TextProps, 'as'>) => (
     <Text {...rest} as="p">
         {children}
     </Text>
--- a/packages/suite/src/hooks/wallet/useSendFormImport.ts
+++ b/packages/suite/src/hooks/wallet/useSendFormImport.ts
@@ -1,6 +1,7 @@
 import { useEffect, useState } from 'react';
 
 import { FiatCurrencyCode, fiatCurrencies } from '@suite-common/suite-config';
+import { notificationsActions } from '@suite-common/toast-notifications';
 import { DEFAULT_PAYMENT } from '@suite-common/wallet-constants';
 import { updateFiatRatesThunk } from '@suite-common/wallet-core';
 import { FiatRates, FiatRatesResult, Output, Rate, Timestamp } from '@suite-common/wallet-types';
@@ -41,6 +42,13 @@ export const useSendFormImport = ({
         if (!result || result.length < 1) return; // cancelled
 
         const rates: { currency: string; rate?: number }[] = [];
+
+        if (result.some(it => !Object.prototype.hasOwnProperty.call(it, 'currency'))) {
+            dispatch(notificationsActions.addToast({ type: 'could-not-parse-csv' }));
+
+            return;
+        }
+
         const currencies = result.map(it => it.currency.toLowerCase());
         const uniqueCurrencies = [...new Set(currencies)];
 
@@ -140,9 +148,7 @@ export const useSendFormImport = ({
                 }
 
                 if (!output.amount || !output.fiat) {
-                    // TODO: display Toast notification with invalid currency error?
-                    // what if there will be multiple errors? Toast spamming...
-                    console.warn('import error', itemCurrency, output);
+                    dispatch(notificationsActions.addToastOnce({ type: 'could-not-parse-csv' }));
                 }
             }
 
--- a/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ImportTransactionModal/ImportTransactionModal.tsx
+++ b/packages/suite/src/components/suite/modals/ReduxModal/UserContextModal/ImportTransactionModal/ImportTransactionModal.tsx
@@ -4,7 +4,7 @@ import { ExtendedMessageDescriptor } from '@suite-common/intl-types';
 import { UserContextPayload } from '@suite-common/suite-types';
 import { networksCollection } from '@suite-common/wallet-config';
 import { parseCSV } from '@suite-common/wallet-utils';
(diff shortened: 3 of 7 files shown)


packages/components/src/components/typography/Paragraph/Paragraph.tsx
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 
 import { Text, TextProps } from '../Text/Text';
 
-export const Paragraph = ({ children, ...rest }: TextProps) => (
-    <Text {...rest} as="p">
+export const Paragraph = ({ children, ...rest }: Omit<TextProps, 'as'>) => (
+    <Text {...(rest as TextProps) /* type-casting to prevent weird TS error */} as="p">
```
komret: Any idea what is happening here?

packages/suite/src/hooks/wallet/useSendFormImport.ts
```diff
@@ -140,9 +149,7 @@ export const useSendFormImport = ({
                 }
 
                 if (!output.amount || !output.fiat) {
-                    // TODO: display Toast notification with invalid currency error?
```
komret, line 143: A five-year old todo :D

packages/suite/src/hooks/wallet/useSendFormImport.ts
```diff
@@ -41,7 +42,15 @@ export const useSendFormImport = ({
...
-        const currencies = result.map(it => it.currency.toLowerCase());
+
+        let currencies;
+        try {
+            currencies = result.map(it => it.currency.toLowerCase());
+        } catch {
+            dispatch(notificationsActions.addToast({ type: 'could-not-parse-csv' }));
+
+            return;
+        }
```
Lemonexe: `let` must have type declaration, otherwise it implicitly types as `any` But that's besides the point..
The try catch seems very weird. How could .map + .toLowerCase() throw error?
I see that at this point, `result` is typed as `{ [p: string]: string }[]`
So it may contain any properties →  `it.currency` may be undefined, and that's the only way where I would see this code throwing (can't access undefined.toLowerCase).
So let's look at that explicitly, instead of try catch:
```suggestion

        if (result.some(it => !Object.prototype.hasOwnProperty.call(it, 'currency'))) {
            dispatch(notificationsActions.addToast({ type: 'could-not-parse-csv' }));

            return;
        }
        const currencies = result.map(it => it.currency.toLowerCase());
```

packages/components/src/components/typography/Paragraph/Paragraph.tsx
Lemonexe: Yes, use `Without` instead of `Omit` :slightly_smiling_face: 
This is a known limitation, and AFAIK we have Without [exactly for that purpose](https://github.com/trezor/trezor-suite/blob/fbe8426fb8ac16c4b951367a9802708b05b81491/packages/type-utils/src/utils.ts#L35)

komret: image

I added a small tweak in ecda1ce01c742e42d555fdb6ceca625338bb3ce1 for `Without` to only allow known properties of `T`.

packages/suite/src/hooks/wallet/useSendFormImport.ts
komret: I didn't want to iterate twice but I have to agree that being explicit is better. 2c1ca42618cd3b0b5fcf3a01960c5824ab50dd37

packages/components/src/components/typography/Paragraph/Paragraph.tsx
Lemonexe: thanks for noticing that. Seems that previously, you could have used `Without<TextProps, 'asdf nonsense'>` without any warning, which is now not possible :+1: :+1:

Lemonexe (approved):
Codewise LGTM.
Works correctly with the example on BTC mainnet :heavy_check_mark:

bosomt:
**QA OK**

**Info:**
 - Suite version: desktop 25.5.0 (6faf4d3002d80e8453a19b108a08dc8027e508f8)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.5.0 Chrome/134.0.6998.178 Electron/35.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T2T1 2.8.9 regular (revision fad9682201cf9289bba2adb66e6e07ed1cf78936)

PR #18511, opened 2025-04-23
fix(device-manager): fix DeviceItem key to prevent two children with the same key

matejkriz:
## Description

Prevent `two children with the same key, ""` on device switcher open (with multiple devices remembered).

## Screenshots:
image

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/device-list-key&lookback=7)


BrantalikP (approved):
:shipit:

PR #18541, opened 2025-04-24
chore(suite): ignore non-serializable fields in serializableCheck

komret:
## Description

Removes annoying warnings. Although some fix akin to https://github.com/trezor/trezor-suite/pull/17894 would be preferable. Any ideas?

## Screenshots:

https://github.com/user-attachments/assets/5c60c424-8dd0-41c3-8ad7-f3ea0120849d

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/ignore-serializable-check&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/ignore-serializable-check&lookback=7)

### DIFF
--- a/packages/suite/src/reducers/store.ts
+++ b/packages/suite/src/reducers/store.ts
@@ -36,6 +36,7 @@ import type { PreloadStoreAction } from 'src/support/suite/preloadStore';
 import { desktopReducer } from './desktop';
 import { extraDependencies } from '../support/extraDependencies';
 import { BluetoothDevice } from '@trezor/transport-bluetooth';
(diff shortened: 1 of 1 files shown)


Lemonexe (approved):
This is just ignoring the issue, but approve, because the issue itself is just "redux best practices". However if we don't try to store these values in Storage it works all right.

Still, ideal solution is to write a type that would enforce only serializable values.
If we want to enforce it in DX, it has to be TS, not a warning you randomly get in console.

I'll try to play around with it later if I have time

PR #18547, opened 2025-04-24
fix(suite-desktop): url decoding before sanitization

### LINKED ISSUE #18549: Trading - Can not complete trade on Desktop
## Describe the bug
Via Simplex trade can't be completed at the moment.

## Info 

It occurs in TS version 25.4.2, but it doesn’t happen in version 25.3.3. We've tested with BTC and ETH. Video for reference

## Screenshots

https://github.com/user-attachments/assets/d8d7e886-48ef-4902-a10f-6c0e7e63ee41

adderpositive:
## Description
- fixed bug when decoded URL was rejected with `403` 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18549

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/url-decoding-in-http&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/url-decoding-in-http&lookback=7)

### DIFF
--- a/packages/node-utils/src/http.ts
+++ b/packages/node-utils/src/http.ts
@@ -296,6 +296,14 @@ export class HttpServer<T extends EventMap> extends TypedEmitter<T & BaseEvents>
         }
     }
 
+    private getSafeDecodedURI(param: string) {
+        try {
+            return decodeURIComponent(param);
+        } catch (e) {
+            this.logger.error(`Failed to decode param: ${e.message}`);
+        }
+    }
+
     /**
      * pathname could be /a/b/c/d
      * return route with highest number of matching segments
@@ -348,8 +356,9 @@ export class HttpServer<T extends EventMap> extends TypedEmitter<T & BaseEvents>
                     let isParamInvalid = false;
 
                     query[key] = allParamsOfSameKey.map(singleParam => {
-                        const sanitized = sanitizeUrl(singleParam);
-                        if (sanitized !== singleParam) isParamInvalid = true;
+                        const decoded = this.getSafeDecodedURI(singleParam);
+                        const sanitized = sanitizeUrl(decoded);
+                        if (sanitized !== decoded) isParamInvalid = true;
 
                         return sanitized;
                     });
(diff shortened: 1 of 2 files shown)


packages/node-utils/src/http.ts
```diff
@@ -405,9 +406,13 @@ export class HttpServer<T extends EventMap> extends TypedEmitter<T & BaseEvents>
...
-                if (!paramsSegments[index]) return acc;
-                const sanitized = sanitizeUrl(paramsSegments[index]);
-                if (sanitized !== paramsSegments[index]) {
+                const paramsSegment = paramsSegments[index];
+                if (!paramsSegment) return acc;
+
+                const decoded = decodeURIComponent(paramsSegment);
+                const sanitized = sanitizeUrl(decoded);
+
+                if (sanitized !== decoded) {
```
mroz22: I'd say that it params should never contain URL encoded strings. At least I have never seen a rest API doing this. 🤔

packages/node-utils/src/tests/http.test.ts
```diff
@@ -365,12 +365,15 @@ describe('HttpServer', () => {
...
+        const res = await fetch(
+            `http://${address.address}:${address.port}/foo?a=1&a=2&b=3&c=%2Fredirect%2Fdetail`,
+            {
+                method: 'POST',
+                body: 'foo',
+            },
+        );
         expect(res.status).toEqual(200);
-        expect(await res.text()).toEqual('?a=1&a=2&b=3');
+        expect(await res.text()).toEqual('?a=1&a=2&b=3&c=%2Fredirect%2Fdetail');
```
mroz22: I added a test like this 
```
  test('query string as url encoded strings', async () => {
        const handler = jest.fn((request, response) => {
            const { search } = url.parse(request.url, true);
            response.end(search);
        });
        server.post('/foo', [parseBodyText, handler]);
        await server.start();
        const address = server.getServerAddress();
        expect(address).toBeDefined();
        const res = await fetch(
            `http://${address.address}:${address.port}/foo?c=%2Fredirect%2Fdetail`,
            {
                method: 'POST',
                body: 'foo',
            },
        );
        expect(res.status).toEqual(200);
        expect(await res.text()).toEqual('?c=%2Fredirect%2Fdetail');
    });
```

and it passed even on develop. I'd prefer adding a test that would fail on develop. Also maybe lets please add a new test case instead of modifying the already existing one

packages/node-utils/src/http.ts
adderpositive: It happens in the endpoint `/buy-post` called in `submitRequestForm`. We use `encodedURIcomponent` for params on the backend side of the trading API.

adderpositive: True!

martykan (approved):

PR #18564, opened 2025-04-25
feat: local firmwares in connect from suite desktop

karliatto:
## Description

This PR creates a new Connect setting called `localFirmwares` that is taken when connect initializes so Connect knows the available firmware that can be used from disk. This will be only used by Trezor Suite desktop and ignored by web.

It also creates the event 'ui-firmware_downloaded' that will be emitted by connect when new FW is downloaded (this part will come in next PR). Suite web will ignore the event but Suite Desktop will listen to it and store that fresh downloaded FW in the Suite config directory to be used in future.

There is the event that connect will listen to `ui-receive_firmware` that will be sent by Suite Desktop when the new downloaded FW is successfully store so Connect knows for sure it can be used during same session.

### DIFF
--- a/packages/connect/src/data/DataManager.ts
+++ b/packages/connect/src/data/DataManager.ts
@@ -7,7 +7,7 @@ import messages from '@trezor/protobuf/messages.json';
 
 import { parseCoinsJson } from './coinInfo';
 import { parseFirmwareReleases } from './firmwareInfo';
-import { ConnectSettings } from '../types';
+import type { ConnectSettings, LocalFirmwares } from '../types/settings';
 import { firmwareAssets } from '../utils/assetUtils'; // Adjust the path as necessary
 
 type AssetCollection = { [key: string]: Record<string, any> };
@@ -17,6 +17,7 @@ export class DataManager {
 
     private static settings: ConnectSettings;
     private static messages: Record<string, any> = messages;
+    private static localFirmwares: LocalFirmwares = { firmwareDir: '', firmwareList: [] };
 
     static load(settings: ConnectSettings, withAssets = true) {
         this.settings = settings;
@@ -65,5 +66,12 @@ export class DataManager {
         }
 
         return this.settings;
+    }
+
+    static setLocalFirmwares(firmwares: LocalFirmwares): void {
+        this.localFirmwares = firmwares;
+    }
+    static getLocalFirmwares(): LocalFirmwares {
+        return this.localFirmwares;
     }
 }
--- a/packages/connect/src/types/firmware.ts
+++ b/packages/connect/src/types/firmware.ts
@@ -42,3 +42,9 @@ export type ReleaseInfo = {
     intermediaryVersion?: IntermediaryVersion;
     translations?: string[];
 };
+
+export type BinaryInfo = {
+    binary: ArrayBuffer;
+    binaryVersion: VersionArray;
+    releaseVersion: undefined;
+};
--- a/packages/suite-desktop-core/src/libs/user-data.ts
+++ b/packages/suite-desktop-core/src/libs/user-data.ts
@@ -31,13 +31,24 @@ export const initUserData = () => {
     }
 };
 
-export const save = async (
-    directory: string,
-    name: string,
-    content: string,
-): Promise<InvokeResult> => {
+export const save: {
+    (directory: string, name: string, content: string, encoding: 'utf-8'): Promise<InvokeResult>;
+    (
+        directory: string,
+        name: string,
+        content: ArrayBuffer,
+        encoding: 'binary',
+    ): Promise<InvokeResult>;
+} = async (directory, name, content, encoding) => {
     const dir = path.join(app.getPath('userData'), directory);
     const file = path.join(dir, name);
+
+    let data: string | Buffer;
+    if (encoding === 'binary') {
+        data = Buffer.from(content as ArrayBuffer);
+    } else {
+        data = content as string;
+    }
 
     try {
         try {
@@ -46,7 +57,7 @@ export const save = async (
             await fs.promises.mkdir(dir);
         }
 
-        await fs.promises.writeFile(file, content, 'utf-8');
+        await fs.promises.writeFile(file, data, encoding);
 
         return { success: true };
     } catch (error) {
--- a/packages/suite-desktop-core/src/modules/firmware.ts
+++ b/packages/suite-desktop-core/src/modules/firmware.ts
@@ -1,0 +1,91 @@
+import { FirmwareType } from '@trezor/connect';
+import { buildFirmwareFileName } from '@trezor/connect/src/utils/firmwareUtils';
+import { DeviceModelInternal, VersionArray } from '@trezor/device-utils';
+
+import { readDir, save } from '../libs/user-data';
+import { app } from '../typed-electron';
+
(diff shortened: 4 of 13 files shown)


packages/suite-desktop-core/src/modules/firmware.ts
```diff
@@ -0,0 +1,91 @@
...
+    return {
+        success: true,
+        payload: {
+            firmwareDir: `${userDataDir}${FIRMWARE_DIR}/`,
+            firmwareList: payload,
+        },
+    };
+};
+
+export const init: ModuleInit = ({ mainThreadEmitter }) => {
```
marekrjpolak, line 31: The whole `firmware` module is currently unused, right? Only `getStoredFirmwares` util is imported somewhere. Therefore, `saveBuffer` util is in my opinion unused as well.

packages/suite-desktop-core/src/libs/user-data.ts
```diff
@@ -56,6 +56,29 @@ export const save = async (
...
 
+export const saveBuffer = async (directory: string, name: string, content: ArrayBuffer) => {
+    const dir = path.join(app.getPath('userData'), directory);
+    const file = path.join(dir, name);
+
+    const buffer = Buffer.from(content);
+
+    try {
+        try {
+            await fs.promises.access(dir, fs.constants.R_OK);
```
marekrjpolak, line 55: As `saveBuffer` is practically the same thing as `save`, I'd propose something like this (plus necessary changes on caller's side):

```typescript
export const save: {
    (directory: string, name: string, content: string, encoding: 'utf-8'): Promise<InvokeResult>;
    (directory: string, name: string, content: Buffer, encoding: 'binary'): Promise<InvokeResult>;
} = async (directory, name, content, encoding) => {
    const dir = path.join(app.getPath('userData'), directory);
    const file = path.join(dir, name);

    try {
        try {
            await fs.promises.access(dir, fs.constants.R_OK);
        } catch {
            await fs.promises.mkdir(dir);
        }

        await fs.promises.writeFile(file, content, encoding);

        return { success: true };
    } catch (error) {
        global.logger.error('user-data', `Save failed: ${error.message}`);

        return { success: false, error: error.message, code: error.code };
    }
};
```

packages/suite-desktop-core/src/modules/trezor-connect.ts
```diff
@@ -99,10 +105,23 @@ export const initBackground: ModuleInitBackground = ({ mainThreadEmitter, store
     return { onLoad, onQuit };
 };
 
-export const init: ModuleInit = () => {
+export const init: ModuleInit = ({ mainThreadEmitter }) => {
     const onLoad = () => {
         // reset previous instance, possible left over after renderer refresh (F5)
         TrezorConnect.dispose();
+        mainThreadEmitter.on('module/firmware/list', (event: LocalFirmwares) => {
```
marekrjpolak: The `onLoad` is called on every desktop app reload (e.g. on Ctrl+R), so I'd rather unsubscribe previous listener from `mainThreadEmitter`, or even better, subscribe it right above `onLoad` in `init` which is executed only once?

packages/connect/src/data/DataManager.ts
```diff
@@ -7,11 +7,13 @@ import messages from '@trezor/protobuf/messages.json';
 
 import { parseCoinsJson } from './coinInfo';
 import { parseFirmwareReleases } from './firmwareInfo';
-import { ConnectSettings } from '../types';
+import type { ConnectSettings, LocalFirmwares } from '../types/settings';
 import { firmwareAssets } from '../utils/assetUtils'; // Adjust the path as necessary
 
 type AssetCollection = { [key: string]: Record<string, any> };
 
+let localFirmwares: LocalFirmwares = { firmwareDir: '', firmwareList: [] };
```
marekrjpolak: Does this have to be outside of `DataManager` class, unlike other data fields?

marekrjpolak (changes requested):

packages/suite-desktop-core/src/modules/firmware.ts
karliatto: Yes, it will be used in next PR where I am adding the FW downloaded and event `FIRMWARE_DOWNLOADED` will be emitted by connect. Just separating it here so PRs are smaller to review.

packages/suite-desktop-core/src/modules/trezor-connect.ts
karliatto: Ok, that makes sense move it to right before `onLoad`.

I also moved the subscription `TrezorConnect.on(DEVICE_EVENT, event =>` to `onLoad` in `init` right after `TrezorConnect.dispose();` it is not part of the scope of this PR but it was not working where it was since `dispose()` unsubscribe all the events.

https://github.com/trezor/trezor-suite/pull/18564/files#diff-6d770fd4f2933ede123857d15a676c2472d44d9bb5d820a2082803913d33fc75R102

packages/connect/src/data/DataManager.ts
karliatto: Right, that makes sense, I created it as `private static`  https://github.com/trezor/trezor-suite/pull/18564/files#diff-54a3dc9015ab9933751016bc35e7987f834b59b24a8d0cc05ef51d1ddabdd1c0R20

packages/suite-desktop-core/src/libs/user-data.ts
karliatto: Done https://github.com/trezor/trezor-suite/pull/18564/files#diff-556493a289f966edbd6adff4a8220c15841af2e13dd306f15aca2a095ecfa722R34

karliatto:
@marekrjpolak I had to rebase it with latest develop.

packages/suite-desktop-core/src/modules/trezor-connect.ts
marekrjpolak: Hmm, that may be a bug. But before that, it was in `initBackground`'s `onLoad`, now it's in `init`'s `onLoad`. I'm not familiar with this relatively new architecture, but it could mean that some background-only functionality will stop working (tray module?). WDYT @martykan?

karliatto: Ok, I reverted that change here https://github.com/trezor/trezor-suite/pull/18564/commits/b64648aeb29541950e533f56cdf338d657e89d61 and created new PR only with that change so we can talk about it there https://github.com/trezor/trezor-suite/pull/18716 since it is not the scope of this PR.

packages/connect/src/types/firmware.ts
```diff
@@ -42,3 +42,9 @@ export type ReleaseInfo = {
     intermediaryVersion?: IntermediaryVersion;
     translations?: string[];
 };
+
+export type BinaryInfo = {
+    binary: ArrayBuffer;
+    binaryVersion: VersionArray;
+    releaseVersion: undefined;
```
marekrjpolak, line 49: Will this always-undefined field have a meaning in following PRs?

marekrjpolak (approved):
Apart from one nitpicky comment (which I forgot to mention last time) seems fine. :+1: Anyway, it doesn't really do anything yet so I may have done some wrong assumptions which I'm willing to update in the followup.

PR #18584, opened 2025-04-25
Better device run interruption

marekrjpolak:
## Description

1) Improved device run overriding - instead of `getOverridePromise`/`setOverridePromise` and awaiting device override promise, `device.interrupt` method was implemented which aborts current run **and** awaits its end, including final cleaning.

2) First run improved - device's `firstRunPromise` and `loaded` replaced by regular run promise and its `'acquiredness'` state.

3) Final reload - final reload was explicitly omitted from `GetFeatures` method and from initial device run (the one without `fn` as it's been done just before that).

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/method-overriding&lookback=7)


mroz22 (approved):
looks ok ✅ and I tested it briefly too ✅

PR #18614, opened 2025-04-28
chore(connect): ethereumSignMessage - catch errors in payload to prec…

mroz22:
just for consistency with other implementations for example: https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/api/solana/api/solanaSignTransaction.ts#L221

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=catch-payload-to-precomposed&lookback=7)


martykan (approved):
I think I didn't expect this could throw, but anything can happen of course

mroz22:
> I think I didn't expect this could throw, but anything can happen of course

that was my understanding why it was "missing" too 👍

PR #18629, opened 2025-04-29
docs(suite-native): elaborate on connecting physical android

Lemonexe:
## Description

Elaborate on information about `adb` over USB | Wi-Fi with physical Android.

@coderabbitai ignore

### DIFF
--- a/suite-native/app/README.md
+++ b/suite-native/app/README.md
@@ -20,7 +20,8 @@ Generally it's recommended to follow official [React Native environment setup](h
 ## Running app on Android
 
 1. Connect device or run emulator.
-    - For a physical device, it's recommended to use [adb over wifi](https://developer.android.com/studio/command-line/adb#connect-to-a-device-over-wi-fi-android-11+) because you will have free up a USB port to connect Trezor device.
+    - For a physical device, [refer to instructions](https://developer.android.com/studio/run/device).
+        - It's recommended to use [adb over wifi](https://developer.android.com/studio/command-line/adb#connect-to-a-device-over-wi-fi-android-11+) because you will have free up a USB port to connect Trezor device.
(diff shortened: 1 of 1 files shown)


suite-native/app/README.md
```diff
@@ -20,7 +20,9 @@ Generally it's recommended to follow official [React Native environment setup](h
 ## Running app on Android
 
 1. Connect device or run emulator.
-    - For a physical device, it's recommended to use [adb over wifi](https://developer.android.com/studio/command-line/adb#connect-to-a-device-over-wi-fi-android-11+) because you will have free up a USB port to connect Trezor device.
+    - For a physical device, [refer to instructions](https://developer.android.com/studio/run/device).
+        - It's recommended to use [adb over wifi](https://developer.android.com/studio/command-line/adb#connect-to-a-device-over-wi-fi-android-11+) because you will have free up a USB port to connect Trezor device.
+        - Though you may have to use USB anytime you need to (re)install the .apk, depending on your device manufacturer
```
PeKne: D you have problem with installing `.apk` over the wifi? Never happened to me 🤔

PeKne (approved):

suite-native/app/README.md
Lemonexe: Maybe it depends on phone manfucaturer?
But anyway I'll try again, because  I changed some more settings in the meantime..

matejkriz: I don't believe it depends on the manufacturer of the phone. ADB over wifi should work, but it's totally fine to not use it and install the app via USB if you don't need to see native logs and have connected physical Trezor at the same time.

Lemonexe: You're right, I tried again to install .apk over wifi and now I have no problems, so I removed that line.

PR #18636, opened 2025-04-29
Mobile trade BTC only device

### LINKED ISSUE #18080: Mobile Trade: handle BTC only device
Adjust the trade flow behavior in case user is using "BTC only" device in the same way as it is handled on desktop. 

Requirements:
1) Buy
- buy flow should be enabled for BTC only device
- user can only select Bitcoin in the coin selection dialog. 
- all other coins should be either hidden from the input or the whole input should be inactive (user wont be able to open the dropdown at all)

2) Swap
- will be solved in  #18616 

3) Sell
- will be hidden in #18102

jbazant:
## Related Issue

Resolve [#18080](https://github.com/trezor/trezor-suite/issues/18080)

### DIFF
--- a/suite-native/module-trading/src/components/buy/TradeableAssetPicker.tsx
+++ b/suite-native/module-trading/src/components/buy/TradeableAssetPicker.tsx
@@ -1,20 +1,50 @@
+import { useEffect, useMemo } from 'react';
+import { useSelector } from 'react-redux';
+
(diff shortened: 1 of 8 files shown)


jbazant:
@coderabbitai review

vytick (approved):
works nicely. Smooth job 🚀

PR #18679, opened 2025-05-01
Connect Explorer backupDevice params, recoveryDevice method

martykan:
## Description

Add missing backupDevice params, add recoveryDevice method


mroz22 (approved):
did not test but it looks legit

PR #18682, opened 2025-05-01
Chore: bump CI related deps

Lemonexe:
## Description

Followup to #18323: update CI-related dependencies, some of them newly owned by @trezor/suite-foundation.
Not those used in runtime, I'm gonna update those after May freeze :snowflake: 

**major version:**
- `nx` (a ton of changes, but we only use a tiny fraction of what nx offers, and I didn't notice anything relevant to us)
- `glob` (breaking change is only node 20 → 22 ✅)
- `sort-package-json`
- `fake-indexeddb` (drops a polyfill that we now have to include)
  
**minor version:**
- `fs-extra`
- `prettier`
- `@eslint/js`
- `eslint`
- `eslint-plugin-mdx`
- `typescript-eslint`

**patch version:**
- `eslint-plugin-react`

**not updated:**
- `chalk` (v5 is ESM only, and some scripts that use it are still CJS)

Bump deps that are related to CI and not runtime.

Only glob is used in build (webpack electron-main bundle), so I checked that part of build.

## QA

:eye: Besides CI checks, I have tested locally:
```
yarn workspace @trezor/suite-desktop build:app  # glob
yarn generate-package @suite-common/something-new  # fs-extra, sort-package-json
yarn workspace @suite-common/message-system sign-config  # fs-extra
yarn workspace @trezor/suite-data build:lib  # fs-extra
```

---

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/bump-ci-deps&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/bump-ci-deps&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/bump-ci-deps&lookback=7)


Lemonexe:
@SocketSecurity ignore npm/@tybys/wasm-util@0.9.0 uses fetch to load async WASM modules and seems legit
@SocketSecurity ignore npm/@emnapi/core@1.4.3 uses fetch & eval to load WASM modules, but to be honest idk what it does.. But it seems we don't use the part of NX that uses the WASM in browser, so both of these packages are probably not relevant.
@SocketSecurity ignore npm/nx@20.8.1 uses fetch for myriad of purposes - the only case relevant for us is that it fetches remote cache..

suite-common/test-utils/src/globalOverrides.ts
Lemonexe, line 1: Found out that side effects like thsi absolutely **must** be in its separate file.
I spent a while time debugging Suite Native app crashing during runtime because of infinite loop:
```
RangeError: Maximum call stack size exceeded
```

I didn't think this would be the cause, because `test-utils` are only used in tests.. The file `mocks.ts` is indirectly included in suite-native tests, but that shouldn't leak into runtime code!

It seems to me, that precisely that is happening. Just because `mocks.ts` is included in the suite-native packages, even though only in tests, its side effects are performed and leak into runtime code even though never referenced. Idk it's weird AF. Isolating the browser mocks into special file solves this :see_no_evil:

komret (approved):

PR #18695, opened 2025-05-02
feat(suite-native): prepare github release draft together with APK build

### LINKED ISSUE #18480: Preprare github release draft on CI during release preparation
We used to have it implemented but then we deleted that code. Let's get it back and update for current needs. See history for a reference:
https://github.com/trezor/trezor-suite/blob/4bcc9d02c7d8fa6964ac0590b8e5f4e0d4f586cc/suite-native/app/eas-post-success.sh

We need to prepare the draft release and also upload the APK as an asset there, but that has to be confirmed (not sure if we will be uploading binaries since the first release or only providing a link to data.trezor.io).

matejkriz:
## Description

- Prepare github release draft together with APK build.

- Changing the S3 upload path to folders by version name 
`https://data.trezor.io/suite/releases/mobile/Trezor-Suite-{version}.apk` -> `https://data.trezor.io/suite/releases/mobile/v{version}/Trezor-Suite-{version}.apk`

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18480

### DIFF
--- a/.github/workflows/release-suite-native-production.yml
+++ b/.github/workflows/release-suite-native-production.yml
@@ -125,7 +125,29 @@ jobs:
           BUILD_URL=$(eas build:view "$BUILD_ID" --json | jq -r '.artifacts.buildUrl')
           echo "build_url=$BUILD_URL" >> $GITHUB_OUTPUT
 
-      - name: Upload APK to S3
+      - name: Download APK
         run: |
           wget -O "Trezor-Suite-${{ steps.get_version.outputs.app_version }}.apk" "${{ steps.eas_build.outputs.build_url }}"
-          aws s3 cp "./Trezor-Suite-${{ steps.get_version.outputs.app_version }}.apk" s3://data.trezor.io/suite/releases/mobile/
+
+      - name: Upload APK to S3
+        run: |
+          aws s3 cp "./Trezor-Suite-${{ steps.get_version.outputs.app_version }}.apk" s3://data.trezor.io/suite/releases/mobile/v${{ steps.get_version.outputs.app_version }}/
+
+      - name: Generate GitHub App token
+        id: trezor-bot-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.TREZOR_BOT_APP_ID }}
+          private-key: ${{ secrets.TREZOR_BOT_PRIVATE_KEY }}
+
+      - name: Prepare release notes
+        run: |
+          echo "https://data.trezor.io/suite/releases/mobile/v${{ steps.get_version.outputs.app_version }}" > release-notes.md
+
+      - name: Create GitHub draft release
+        env:
+          GITHUB_TOKEN: ${{ steps.trezor-bot-token.outputs.token }}
+        run: |
+          gh config set prompt disabled
+          TAG="v${{ steps.get_version.outputs.app_version }}@mobile"
+          gh release create --repo trezor/trezor-suite --draft ${TAG}  --title ${TAG} --latest=false --target ${{ github.ref_name }} --notes-file release-notes.md "./Trezor-Suite-${{ steps.get_version.outputs.app_version }}.apk"


.github/workflows/release-suite-native-production.yml
```diff
@@ -125,7 +125,39 @@ jobs:
           BUILD_URL=$(eas build:view "$BUILD_ID" --json | jq -r '.artifacts.buildUrl')
           echo "build_url=$BUILD_URL" >> $GITHUB_OUTPUT
 
-      - name: Upload APK to S3
+      - name: Download APK
         run: |
           wget -O "Trezor-Suite-${{ steps.get_version.outputs.app_version }}.apk" "${{ steps.eas_build.outputs.build_url }}"
-          aws s3 cp "./Trezor-Suite-${{ steps.get_version.outputs.app_version }}.apk" s3://data.trezor.io/suite/releases/mobile/
+
+      - name: Upload APK as workflow artifact
```
matejkriz: This is probably an extra step, but can be useful to see APK as artifact in the github?

.github/workflows/release-suite-native-production.yml
```diff
@@ -125,7 +125,39 @@ jobs:
...
+        with:
+          app-id: ${{ secrets.TREZOR_BOT_APP_ID }}
+          private-key: ${{ secrets.TREZOR_BOT_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: |
+            trezor/trezor-suite
+
+      - name: Prepare release notes
+        run: |
+          echo "https://data.trezor.io/suite/releases/mobile/v${{ steps.get_version.outputs.app_version }}/Trezor-Suite-${{ steps.get_version.outputs.app_version }}.apk" > release-notes.md
```
matejkriz: Not final text, I guess... We can add there any copy.

.github/workflows/release-suite-native-production.yml
vdovhanych: We're adding it to the release in this exact job. Why have it duplicated? It isn't necessary.

.github/workflows/release-suite-native-production.yml
```diff
@@ -125,7 +125,39 @@ jobs:
...
+          aws s3 cp "./Trezor-Suite-${{ steps.get_version.outputs.app_version }}.apk" s3://data.trezor.io/suite/releases/mobile/v${{ steps.get_version.outputs.app_version }}/
+
+      - name: Generate GitHub App token
+        id: trezor-bot-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.TREZOR_BOT_APP_ID }}
+          private-key: ${{ secrets.TREZOR_BOT_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: |
```
vdovhanych: This is only needed if you're generating token for a different repository, if it runs in the one you need access to yo don't need it here.

.github/workflows/release-suite-native-production.yml
vdovhanych: Why are we adding a URL to a file directly? I think having just a link to the data.treozor.io/suite/releases/mobile/ is sufficient. I thought that was the reason I'm adding the indexes there.

.github/workflows/release-suite-native-production.yml
matejkriz: removed 52377945ad7d5e59bd2dda22035765a9ef0510b0

.github/workflows/release-suite-native-production.yml
matejkriz: thanks, removed 52377945ad7d5e59bd2dda22035765a9ef0510b0

.github/workflows/release-suite-native-production.yml
matejkriz: Changed to link to version folder. I don't see a reason to have a fixed link to all releases in every release. 

But by changing a copy, I meant some basic text, something like "Trezor Suite Lite for Android is now available from: " and the link on the next line.

vdovhanych (approved):

MiroslavProchazka:
@coderabbitai review

MiroslavProchazka (approved):
Squash the fixup commit please, otherwise looks good

PR #18715, opened 2025-05-05
feat: change Tor enable model to user expression instead of Tor status

### LINKED ISSUE #18683: Disabling Tor should only happen by user action, not automatically
We have few problems in current implementation:
- Tor will shut down if it doesn't boot properly or if the user shuts down the Suite during loading
- When tor is loading, user can't turn it off in settings

I propose to not disable tor toggle during it's booting. Rather use the toggle as a users expression, not connection status. So user can anytime change his will to turn the Tor on and off. If user turns the tor on, Suite start communication only through tor.. 
_→ Some functions might be unavailable for a while similarly like when user change the backend._

karliatto:
## Description

* b5d41d6b92665f709482b97130662bf2c43dbb28 - adding `retryForever` so it makes the option to enable Tor and it will try to restart untill it successfully works. This makes Tor enabling just a user decision so in order for Tor to stop trying to connect the user has to disable it.
* aeab6d9c2db9aab489383698b501567df58f4ffc- This makes the changes in setting section to make the Tor enable toggle to work as user decision instead of Tor status.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18683

### DIFF
--- a/packages/suite-desktop-core/src/modules/tor.ts
+++ b/packages/suite-desktop-core/src/modules/tor.ts
@@ -76,21 +76,28 @@ const load = async ({ mainWindowProxy, store, mainThreadEmitter }: Dependencies)
               }
             : { proxy: '' };
     };
-    const handleTorProcessStatus = (status: TorProcessStatus) => {
-        const { useExternalTor, running } = store.getTorSettings();
+    const handleTorProcessStatus = (status: TorProcessStatus, shouldEnableTor: boolean) => {
+        const { useExternalTor } = store.getTorSettings();
         let type: TorStatus;
 
-        if (!status.process) {
-            type = TorStatus.Disabled;
-        } else if (status.isBootstrapping) {
-            type = TorStatus.Enabling;
-        } else if (status.service && !useExternalTor) {
-            type = TorStatus.Enabled;
-        } else if (useExternalTor && running) {
-            type = TorStatus.Enabled;
-        } else {
+        // 1. Check for Disabled state:
+        if (shouldEnableTor === false) {
             type = TorStatus.Disabled;
         }
+        // 2. Check for Enabled state (only if not Disabled):
+        //    It's Enabled if:
+        //    a) We are using External Tor and user is responsible for running the service.
+        //    b) We are using Internal Tor and the service is confirmed ready (`status.service`).
+        else if (useExternalTor || status.service) {
+            // Since we passed the first `if`, `shouldEnableTor` must be true here.
+            type = TorStatus.Enabled;
+        }
+        // 3. Otherwise, it must be Enabling:
+        //    If it's supposed to be shouldEnableTor (`shouldEnableTor` is true) but isn't fully Enabled yet.
+        else {
+            type = TorStatus.Enabling;
+        }
+
         mainThreadEmitter.emit('module/tor-status-update', type);
         mainWindowProxy.getInstance()?.webContents.send('tor/status', {
             type,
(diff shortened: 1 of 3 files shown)


packages/suite/src/views/settings/SettingsGeneral/Tor.tsx
```diff
@@ -57,11 +58,7 @@ export const Tor = () => {
     return (
         <SettingsSectionItem anchorId={SettingsAnchor.Tor}>
             <TextColumn
-                title={
-                    <LoadingContent isLoading={isTorLoading} isSuccessful={!hasTorError}>
```
martykan, line 61: Why not keep this? It would make sense to show a loading indicator here in my opinion

martykan (approved):

packages/suite/src/views/settings/SettingsGeneral/Tor.tsx
karliatto: That was they way it was working before, so it was loading and when Tor was ready it was marked as enabled but if the process of starting Tor and establishing a successful Tor circuit was failing for some reason (for instance timeout) the loading was completed and Tor was marked as disabled. 

With this PR the concept is changing to have this toggle to work just as user expression to use Tor, so it Tor fails it starts again until user manually disable it. This is in order to avoid disable Tor when user actually wants to use it.

There is plan to add more information about the Tor status in different ways in the UI.

martykan: Yes I see the change, but I still think it's useful to keep this so the user can clearly see it's loading. 
When I come to the settings and turn Tor on, it's more obvious to see that something is happening right there, than looking at the small Tor status icon in the bottom left.

karliatto: Yes, I know what you mean and I agree, I would add there in the setting toggle below or something more information about the Tor status, and why it is taking more time and so on. But it looks we will address that in future improvement.

martykan: Ok got it

packages/suite-desktop-core/src/modules/tor.ts
```diff
@@ -143,32 +150,39 @@ const load = async ({ mainWindowProxy, store, mainThreadEmitter }: Dependencies)
         }, interval);
     };
 
-    const setupTor = async (shouldEnableTor: boolean) => {
-        const { useExternalTor, externalPort } = store.getTorSettings();
+    const setupTor = async (shouldEnableTor: boolean, retryForever = true) => {
```
marekrjpolak: It seems that `retryForever` param is always true and therefore useless.

packages/suite-desktop-core/src/modules/tor.ts
```diff
@@ -76,21 +76,28 @@ const load = async ({ mainWindowProxy, store, mainThreadEmitter }: Dependencies)
               }
             : { proxy: '' };
     };
-    const handleTorProcessStatus = (status: TorProcessStatus) => {
+    const handleTorProcessStatus = (status: TorProcessStatus, shouldEnableTor?: boolean) => {
```
marekrjpolak: Is this optional `shouldEnableTor` param necessary? It seems that it's passed only from `tor/toggle` listener, which is also practically the only place setting the `running` flag in store.

So unless `tor/toggle` can be called twice in parallel (can it?), there will always be the same value in `running` and `shouldEnableTor` and either one can be omitted (at least I think so).

packages/suite-desktop-core/src/modules/tor.ts
karliatto: Yes, I am aware of it, but I wanted to keep it in case we want to have some way to make it fail in some situations like when starting the app there is a flow when it fails. But after thinking it a bit I removed it and add a comment.

packages/suite-desktop-core/src/modules/tor.ts
karliatto: The issue here is that it looks like the store updates is not blocking and is not awaiting so in some cases it is not properly updated to the latest value. But in order to make it nicer I am now using always `shouldEnableTor` but in one case that is also coming from the store value but is obtained outside this function.

marekrjpolak: Great, thanks! :heart:

PR #18716, opened 2025-05-05
fix(suite-desktop-core): connect event was unsubscribed

karliatto:
## Description

I am not really familiar with `initBackground` but it looks to me that subscription to `DEVICE_EVENT` is not working since after there is `TrezorConnect.dispose();` that unsubscribe those events. WDYT @martykan ?

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/is-trezor-event-working&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/is-trezor-event-working&lookback=7)

### DIFF
--- a/packages/suite-desktop-core/src/modules/trezor-connect.ts
+++ b/packages/suite-desktop-core/src/modules/trezor-connect.ts
@@ -1,6 +1,6 @@
 import { ipcMain } from 'electron';
 
-import TrezorConnect, { DEVICE_EVENT } from '@trezor/connect';
+import TrezorConnect from '@trezor/connect';
 import { IpcProxyHandlerOptions, createIpcProxyHandler } from '@trezor/ipc-proxy';
 import { parseElectrumUrl } from '@trezor/utils';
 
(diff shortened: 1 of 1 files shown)


martykan:
Yes, you're right about this. This feature is not used much, I added it to experiment what we can do with the tray icon, but we will skip on showing the devices for now, because we still don't init Connect in background

karliatto:
> Yes, you're right about this. This feature is not used much, I added it to experiment what we can do with the tray icon, but we will skip on showing the devices for now, because we still don't init Connect in background

So then we leave as it is?

martykan:
Well with this fix the code at least makes more sense, but maybe I'd just comment it out for now

karliatto:
> Well with this fix the code at least makes more sense, but maybe I'd just comment it out for now

Something like this github.com/trezor/trezor-suite/pull/18716/files#diff-6d770fd4f2933ede123857d15a676c2472d44d9bb5d820a2082803913d33fc75R107  ?

martykan (approved):

martykan (approved):

PR #18724, opened 2025-05-05
Refactor: complete rewrite of discovery

### LINKED ISSUE #12468: Not showing "duplicite passphrase modal" while creating new passphrase wallet (view-only)

### LINKED ISSUE #18149: Improve Passphrase cancel flows
Canceling the passphrase entry flow in the step of on device confirmation/entry currently leads users to a confusing “Authorization error” screen—something that would only make sense if an actual technical failure occurred.

If a user cancels the flow, we should return them to the previous state as if nothing happened—after all, that’s what “cancel” is supposed to mean.

Image

If the user cancels the passphrase flow during the on-device confirmation step—either directly on the device or via Suite’s “Confirm on Trezor” modal—we should not redirect them to the Authorization error page.

Instead, the on-device confirmation should simply be aborted, and the user should be returned to the input field where they initiated the confirmation. This input should now be empty, as the passphrase should not be retained in memory.

Re-entering the passphrase makes sense, since the user explicitly canceled the confirmation of the previously entered value, making it no longer relevant.

Image

Image

In cases where a passphrase input is automatically requested after reconnecting a remembered (view-only) wallet, the cancellation flow already works well.

Only the notification copy needs improvement—but let’s handle that in a separate issue, as we have several suboptimal notifications to revisit.

Image

### LINKED ISSUE #18391: Move passphrase device prompt from modal to wallet switcher
Please move the device prompt from modal directly to the wallet switcher in passphrase related user flows. Currently the interface is "jumping" between the wallet switcher and modals, damaging continuity, consistency and smoothness of the UI.

Figma design: https://www.figma.com/design/Y2E8dcdNYrSMGIVRjYFXKp/-Suite--Components-1.0?m=auto&node-id=4539-41621&t=yEult2ks4sEHoeCJ-1
_(please mind only the confirmation part, the rest is more mature n implementation than here in design)_

Prompts to cover:
- When entering passphrase in Suite, user is prompted to confirm the pp on device
- When entering passphrase on device
- Copywriting needs to match each case above as it does today (confirm vs. enter)

### LINKED ISSUE #19061: Enabling ADA right after first time app start shows "Your wallet is empty" for few seconds
**Describe the bug**
When I enable ADA as one of the coins during the first time I open the app coin enabling, the app shows after confirming the selection a screen with "Your wallet is empty" for a few seconds. After that, the discovery starts normally.

**Info:**

- Mobile Suite Version: 25.4.1
- OS: name, version

**How to reproduce**

1. Wipe the app
2. Connect the device in the welcome screens
3. In coin enabling select ADA
4. Confirm the selection
5. Observe the "Your wallet is empty" screen appearing for a few seconds

**Expected behavior**
Discovery starts right away after coin enabling.

**Screenshots**

Image

mroz22:
Motivated by #17966 to refactor passphrase flows – instead, **refactor the entire discovery**, moving the control from redux to connect, and refactor passhprase flows along with it (heavily interlinked).

## Description
**Previously** Suite & Suite Lite would send discovery request bundles back and forth with Connect.
**Now** Suite uses the new Connect API (#18973) where Connect has full control over the gradual discovery of accounts and Suite & Suite Lite listen to the result progress events.
Also needed #18798 (unify wallet settings for Suite & Suite Lite).

Selection of notable changes:
- Reimplement discovery, turning networks on/off
  - Reimplement adding account types
  - Reimplement metadata initiation
  - Handle discovery failures
- Refactor passphrase flows in Suite & Suite Lite
- Reimplement pin flows in Suite Lite
- Delete `discovery` store; now only `accounts` bear info that is worth persisting for remembered wallets
- Remove and cleanup lots of obsolete code

## Related Issue

resolves #18149 (the original intent)
resolves #18391
probably resolves #19061
probably resolves #12468

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=discovery&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=discovery&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=discovery&lookback=7)


vojtatranta:
UX problems:
- when hitting cancel on device / in the Device UI, we couldn't get user "back" to the passphrase flow, unless we specify some "where to go after cancel" before, but that's in the discovery unrealted to the "passphrase" flow, it would have to be "registered" before

vojtatranta:
As of `2fb27ab788544fff52c926e40ff461dacf19f85e` - `fix turning on passphrase flow from add wallet button` we have:
- the flows are somewhat working, with back and cancel transitions
- you can add passphrase wallet existing / not existing state


what doesnt work:
- there isn't initial "discovery" I am not sure what should cause the "initial" disdcovery so there is a device with a proper `state` property
- this causes that there isn't any device to be used, so the device switcher doesnt work
- you cannot add multiple passphrase wallets, the previous one is always replaced
- the "duplicity" check doesn't work properly, coz the devices aren't in the state
- there is a problem with the modals clashing - especially with that 'confirm passphrase' modal triggered by device, it needs thinking how to fix it properly https://github.com/trezor/trezor-suite/pull/18724/files#r2077659056
- **flashing out of the confirm passphrase modal** -  (repro: go to the passphrase flow to enter pasphrse form and hit cancel on top right), reason: the discvoery is cancelled syncrhnoosly, but the event to close UI action from trezor comes back later
  -  solution 1. add "passphrase forms" to the `<ConfirmPassphraseBeforeAction />`
  - solution 2. Marek - when intercepting a UI event from trezor, we could not display the modal, based on the state of discovery
  - solution 3. add some "context" to the trezor events which we could use to "switch" whether to display the DEVICE UI modal based on the current state

So the "next passphrase wallet" overwites the previous one

packages/suite/src/components/suite/layouts/SuiteLayout/PassphraseFlow.tsx
```diff
@@ -0,0 +1,77 @@
...
+const ForegroundAppModal = ({ app, cancelable }: ForegroundAppModalProps) => {
+    const dispatch = useDispatch();
+
+    const onCancel = () => dispatch(closeModalApp());
+
+    if (app === 'switch-device')
+        return <SwitchDevice cancelable={cancelable} onCancel={onCancel} />;
+};
+
+export const PassphraseFlow = () => {
```
vojtatranta, line 62: This thing sucks like hell btw

packages/suite/src/hooks/suite/usePreferredModal.ts
```diff
@@ -43,22 +44,55 @@ const getForegroundAppAction = (route: ForegroundAppRoute, params: Partial<Modal
     }) as const;
 
 export const usePreferredModal = () => {
-    const { getDiscoveryStatus } = useDiscovery();
+    const { getDiscoveryStatus, discovery } = useDiscovery();
```
vojtatranta: this `usePreferredModal` is absolutely horrible btw

packages/suite/src/hooks/suite/usePreferredModal.ts
```diff
@@ -43,22 +44,55 @@ const getForegroundAppAction = (route: ForegroundAppRoute, params: Partial<Modal
...
 
     if (route && isForegroundApp(route) && hasPriority(route)) {
         return getForegroundAppAction(route, params);
     }
 
     if (modal.context !== MODAL.CONTEXT_NONE) {
+        // NOTE: in case when passphrase flow is active, we handle the device passphrase request
+        // within the passphrase flow
+        if (
+            'windowType' in modal &&
```
vojtatranta, line 67: I'll try to remove the need for this

vojtatranta: yeah, so the whole thing is kinda problematic.

This is used for the "global' confirmation of the passprhase eg. when you didn't yet enter passphrase to a remember wallet but you want to revel a a public key.

But for the "passphrase flow" this input is handled within a different modal.

So it is mess. 

Also, it happens, that when you remove the discovery from the redux store (eg. on cancel) it takes a while to remove the UI.REQUEST_PASSPHRSE from the redux state coz it is async, so there is flashing of that modal in the background too.

This sucks. Don't know how to fix it properly.

vojtatranta:
Current state:
- no "initial" discovery - there aren't any device when the app is first loaded

vojtatranta:
I tried to run the discovery when app starts / when onboarding is finished and we go to the suite-index.

It seemed that multiple discoveries were run over-each other, causing error `device call in progress` or something.

I think we shouldn't start a new discovery (that is index by device.path, hence nothing unique) only in states when the discovery is stopped / failed / cancelled/. completed.

Will add provisional code for this cc @mroz22

vojtatranta:
Okay, so right now I am trying to run the E2E tests for the passphrase flow and it throws error in the discovery saying that there is some other device call in progress, so this fails the discovery. 

Wondering what to do. The state before:
- discovery was run based on the result of the `authorizeDevice ` call, and it was matched in the middleware.
- this thunk deosn't seem to be called now / is called but doesn't trigger the discovery through middleware anymore

vojtatranta:
The authenticate device vs. discovery clash.`

- I tried to run discovery on when device is connected (that's likely incorrect , coz the device doesn't have yet state)
- this is also triggered when device is not yet initialized
- when device is in onboarding, there is authenticity check
- this check is run during when the discovery started on "device coonnected" was triggered
- this throws error coz the authenticity check is run during th discovery -> method call in progress error

I guess this whole thing is wrong somehow and the time when the suite is started.

I really can't tell where it should be, likely at the point of starting the suite.

packages/suite/src/views/onboarding/steps/Final.tsx
```diff
@@ -219,6 +221,11 @@ export const FinalStep = () => {
                             });
 
                             goToSuite(true);
+                            dispatch(
```
vojtatranta: this is likely wrong, although needed

suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -648,6 +645,7 @@ export const deviceConnectThunks = createThunk<void, DeviceConnectThunksParams,
         switch (type) {
             case DEVICE.CONNECT:
                 dispatch(deviceActions.connectDevice({ device, settings }));
+                dispatch(startDiscoveryThunk({ device }));
```
vojtatranta: When device is connected at some point, discovery should start, but this clashes with the authenticity check as explained in the main thread.

vojtatranta:
as of `15b60ec96abeaf4a77c2393cec9f1bd44db2f88a` = `add awaiting for the UI action to display device confirmation due to device 'ready' state`, the Passphrase flow e2e test is **PASSING**.

What'a re the bugs I noticed:
- the start of the discovery has be properly timed also restarts, etc. (due to onboarding authenticitiy check clash, etc.)
- there will surely be bugs in the when the discovery is started
- there isn't discovery suggested when activating coints
- the discovery progoress behaves weirdly
- in some cases, when app is started, there addded more and more wallets to a same device
- modal clashing - the async event of close UI is slower than the sync action of ending the discovery, resulting in this https://github.com/trezor/trezor-suite/pull/18724#issuecomment-2858604634

mroz22:
squashed and rebased onto #18798

vojtatranta:
Here's an ilustration of the race condition between closing down the trezor device UI, and turning of "passphrase flow discovery".


https://github.com/user-attachments/assets/90271b9c-fb81-49b5-ab1b-05852f29eb27

vojtatranta:
Here are required changes for the passphrase flow https://github.com/trezor/trezor-suite/issues/18149

- it requires jumping back to the discovery states
- that's fine to do, although there will have to be a larger logic to determine to which phase of the discovery to go

vojtatranta:
Adding passphrsae flow chart for clarity https://private-user-images.githubusercontent.com/4154045/432766241-872ccdae-5367-49f0-b7b7-3426a99af17d.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDY2MTA4OTgsIm5iZiI6MTc0NjYxMDU5OCwicGF0aCI6Ii80MTU0MDQ1LzQzMjc2NjI0MS04NzJjY2RhZS01MzY3LTQ5ZjAtYjdiNy0zNDI2YTk5YWYxN2QucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDUwNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA1MDdUMDkzNjM4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MWU0ZmNmMDQyMGM3MjVmYWI3N2NmMzExNzMxYjM3YTAwNmMxNzExYTcyNmJkYmU3MzE1MzdjYWU0NGVkMTMwMSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.KPWYQpe5WgC7OelOvplctQlyViya553paDNYfkNdd28

vojtatranta:
Closes https://github.com/trezor/trezor-suite/issues/18391

vojtatranta:
Aha, so here's another task in the pipeline to move the confirmation on the device within the passphrase flow from the global modal up to the left top "Wallet switcher" https://github.com/trezor/trezor-suite/issues/18391

vojtatranta:
As of `[feat(suite-native): make the passphrase flow in native declarative](https://github.com/trezor/trezor-suite/pull/18724/commits/3c22383d9605dcd1f03ed2ea1000d081314e3766)`

The adding of the wallet works nice ! Also like password mismatch can be cancelled or started over, stuff thats' wrong:
- if you confirm a passphrase with matching one, the wallet isn't selected on the main page
- there is unsolved myster with `useNavigateToInitialScreen` I think I should use it instead of "completion by redirect"
- the go back button and cancelling should be checked if it works correctly

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -700,7 +700,19 @@ export const prepareDeviceReducer = createReducerWithExtraDeps(initialState, (bu
...
+    const deviceInstance =
+        state.device?.selectedDevice?.instance ?? state.device?.selectedDevice?.walletNumber;
+
+    return state.device?.devices.find(
+        device =>
+            device.id === deviceId &&
+            device.path === devicePath &&
+            (device.instance === deviceInstance || device.walletNumber === deviceInstance),
+    );
+};
```
mroz22: @vojtatranta why this change? I agree that device.selectedDevice should only be a pointer to device.devices. But it should be refactored everywhere and in a single blow. This indicated that there is a problem with updating selected device and you are trying to solve it this way?

vojtatranta: Yeah, we could drop it and use it just for native.

The problem in native is that the devices array is updated with things like `state` but selected device isn't.`

So either use selector or middleware to update it.

vojtatranta: I didn't notice that it is in common, thanks

vojtatranta: I put it back

mroz22: it is logic around this that you want to check http://github.com/trezor/trezor-suite/blob/develop/suite-common/wallet-core/src/device/deviceThunks.ts#L227 

and it was introduced by me in this branch that stat is not updated on purpose. there is a special action that does it only when we are sure that we do want to apply state update 
https://github.com/trezor/trezor-suite/blob/discovery/suite-common/wallet-core/src/device/deviceReducer.ts#L269

vojtatranta: well, it is suppose to work on mobile, I guess, right?

mroz22: yes. this should be shared logic already. if state does not get updated correctly, this thunk isn't called when it should https://github.com/trezor/trezor-suite/blob/discovery/suite-common/wallet-core/src/discovery/discoveryThunks.ts#L67

vojtatranta: key, checking

suite-native/discovery/src/discoverySelectors.ts
Lemonexe, line 1: Note: original file deleted, and instead `discoveryConfigSlice` was renamed to `discoverySelectors` because the only thing that remained from both files, was a few selectors in the config slice

packages/suite-desktop-core/e2e/tests/wallet/add-account-types.test.ts
```diff
@@ -128,7 +128,6 @@ test.describe('Account types suite', { tag: ['@group=wallet'] }, () => {
                     await expect(page.getByTestId('@modal')).toBeVisible();
                     await page.getByTestId(`@settings/wallet/network/${coin.symbol}`).click();
                     await page.getByTestId('@add-account').click();
-                    await page.discoveryShouldFinish();
```
Lemonexe, line 131: When you have `account #1` and add `account #2` then discovery is no longer needed :slightly_smiling_face:  
This was relevant for the old discovery flow.
After the refactoring, the next empty account is always pre-discovered, but not visible. The UI only makes it visible.

mroz22: yep, it can probably be removed. the only question is how we should handle the edgecase of adding another account: 
1. I have added an account by making it visible. 
2. now I don't have any other empty account. how to refill? shouldn't now another empty account be added? that stays in buffer and can't be made visible until the previously added account becomes non empty?

Lemonexe (approved):
Looks LGTM to me, :shipit:

suite-native/discovery/src/discoveryConfigSlice.ts
```diff
@@ -1,140 +0,0 @@
-import { A, pipe } from '@mobily/ts-belt';
```
PeKne, line 1: please delete this file, it has no content now.

suite-native/firmware/src/components/ConfirmFirmwareUpdateScreenFooter.tsx
```diff
@@ -28,10 +28,11 @@ export const ConfirmFirmwareUpdateScreenFooter = ({
...
     const { translate } = useTranslate();
 
-    const deviceState = useSelector(selectDeviceState);
+    const devicePath = useSelector(selectDevicePath);
 
-    const isDiscoveryRunning = useSelector((state: DiscoveryRootState & DeviceRootState) =>
-        selectIsDiscoveryActiveByDeviceState(state, deviceState),
+    const isDiscoveryRunning = useSelector(
+        (state: DiscoveryRootState & DeviceRootState) =>
+            selectDiscoveryByDevicePath(state, devicePath)?.status === 'progress',
```
PeKne, line 35: nit: this should be a new selector

suite-native/module-authorize-device/src/components/connect/PinFormControlButtons.tsx
```diff
@@ -37,7 +37,7 @@ export const PinFormControlButtons = ({ onSuccess }: PinFormControlButtonsProps)
     const [containerHeight, setContainerHeight] = useState(0);
     const animatedHeight = useSharedValue(0);
 
-    const dispatch = useDispatch();
+    // const dispatch = useDispatch();
```
PeKne, line 40: nit: remove

suite-native/module-authorize-device/src/components/connect/PinFormControlButtons.tsx
```diff
@@ -78,9 +78,9 @@ export const PinFormControlButtons = ({ onSuccess }: PinFormControlButtonsProps)
             onPressPrimaryButton: () => {
```
PeKne, line 78: The button does nothing currently. Does the button still have any meaning?

suite-native/module-home/src/screens/HomeScreen/components/FirmwareUpdateAlert.tsx
```diff
@@ -59,9 +59,10 @@ export const FirmwareUpdateAlert = () => {
     const deviceId = useSelector(selectDeviceId);
     const isConnected = useSelector(selectIsDeviceConnected);
     const isDeviceBackedUp = useSelector(selectIsDeviceBackedUp);
-    const deviceState = useSelector(selectDeviceState);
-    const isDiscoveryRunning = useSelector((state: DiscoveryRootState & DeviceRootState) =>
-        selectIsDiscoveryActiveByDeviceState(state, deviceState),
+    const devicePath = useSelector(selectDevicePath);
+    const isDiscoveryRunning = useSelector(
```
PeKne, line 63: Could not be `selectHasRunningDiscovery` used? If not create different more specific selector.

suite-native/coin-enabling/src/coinEnablingSelectors.ts
```diff
@@ -1,31 +1,37 @@
...
 ) => {
     const device = selectSelectedDevice(state);
     const isDeviceUnlocked = selectIsDeviceUnlocked(state);
     const isPortfolioTrackerDevice = selectIsPortfolioTrackerDevice(state);
     const isCoinEnablingInitFinished = selectIsCoinEnablingInitFinished(state);
-    const isDeviceConnectedAndAuthorized = selectIsDeviceConnectedAndAuthorized(state);
+    const discovery = selectDiscoveryForSelectedDevice(state);
+
+    // NOTE: in this state, discovery threw error that no account is selected
+    const isAccountsDiscoveryEmpty =
```
PeKne, line 23: nit: could this be some new selector?

suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -65,6 +65,13 @@ export const connectInitThunk = createThunk<
                 'Hey, it looks like you called a TrezorConnect method without providing device property.',
             );
         }
+
+        if (
+            action.type === 'ui-close_window'
+            // && getState().wallet.discovery[getState().device?.selectedDevice?.path]?.status ==='progress'
+        ) {
+            // return;
```
matejkriz, line 73: leftover, righ? followup?

suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
```diff
@@ -104,14 +102,14 @@ export const FirmwareInstallationScreenContent = ({
         };
     }, [dispatch, isTemporaryRememeberAllowed, resetReducer, setIsFirmwareInstallationRunning]);
 
-    const handleFirmwareUpdateFinished = useCallback(async () => {
-        await requestPrioritizedDeviceAccess({
-            deviceCallback: () => dispatch(authorizeDeviceThunk()),
-        });
+    const handleFirmwareUpdateFinished = useCallback(() => {
+        console.warn(
```
PeKne, line 106: leftover console.warn

suite-native/module-authorize-device/src/components/connect/PinFormControlButtons.tsx
```diff
@@ -7,12 +7,12 @@ import Animated, {
...
 } from 'react-native-reanimated';
-import { useDispatch, useSelector } from 'react-redux';
+import { useSelector } from 'react-redux';
 
-import { authorizeDeviceThunk, selectDeviceAuthFailed } from '@suite-common/wallet-core';
+import { selectDeviceAuthFailed } from '@suite-common/wallet-core';
 import { useAlert } from '@suite-native/alerts';
 import { Box, Button, HStack, IconButton } from '@suite-native/atoms';
-import { requestPrioritizedDeviceAccess } from '@suite-native/device-mutex';
+// import { requestPrioritizedDeviceAccess } from '@suite-native/device-mutex';
```
PeKne, line 15: remove if not needed

suite-native/module-authorize-device/src/screens/connect/ConnectAndUnlockDeviceScreen.tsx
```diff
@@ -45,12 +43,13 @@ export const ConnectAndUnlockDeviceScreen = ({
             // When selected device become connected, we need to navigate out of this screen.
             navigateBack();
         } else {
```
PeKne, line 45: is this else still needed?

PeKne:
I was able to get to a state when there is the standard wallet twice in the wallet switcher. Could not reproduce that later 🤷 :/. It happened to me right after fresh install of the app. 



EDIT from @matejkriz: I reproduced it in 2 out of 2 tries.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -1,801 +1,926 @@
...
-        }
-        // calculate progress
-        const progress = dispatch(
-            calculateProgress({
-                ...discovery,
-                failed,
-            }),
-        );
+    return (event: ProgressEvent) => {
+        console.warn('bundle progress handler', event);
```
matejkriz, line 206: Can we please remove this console.warn? It can be in a fast follow-up...

PeKne:
https://github.com/user-attachments/assets/edfdec26-435f-434b-aa2d-8e7f057a3bf1

If I change enabled networks in the settings, the discovery correctly triggers for the selected device. But If I go to some other device via the device switcher, the discovery is not triggered and the whole device looks empty even though it is not.

PeKne:
https://github.com/user-attachments/assets/de781242-54c0-4114-8c12-2469a88c5703

Firmware update is broken completely for mobile :/. The state of the screen is changing unexpectedly at the beginning. Then it stops and if I continue on the Trezor device, the app seems stuck and do nothing anymore.

matejkriz:
When I disconnected the remembered device, I got: "Your wallet is empty". It happened to me in 2 out of 2 attempts.
image

suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
Lemonexe: Yeah there a lot of them, good point!

In `chore: discovery code cleanup` I went through them systematically.
-  I deleted useless debugging consoles
  - :thought_balloon:  It's not a really good practice to use `console.warn` for debug, as eslint doesn't catch it :confused: 
 - I kept only the ones that handle unexpected error flows and escalated them to `console.error`, so we will know in Sentry if the unexpected happens

suite-native/discovery/src/discoveryConfigSlice.ts
Lemonexe: Oh thx, I thought I already did, but messed up git.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
Lemonexe: https://github.com/trezor/trezor-suite/pull/18724#discussion_r2123331129

Lemonexe:
ok idk what's up with the android test, [but the very same commit passed here](https://github.com/trezor/trezor-suite/actions/runs/15417839400), so I am merging it

PR #18736, opened 2025-05-05
fix(suite): keep `reconnectRequested` flag after wipeDevice

szymonlesisz:
## Description

looks like this bug is been here since forever

After successful  wipeDevice `dispatch(deviceActions.requestDeviceReconnect());` is called, 
it sets `reconnectRequested` flag only for selectedDevice,  state.devices array is not updated therefore next trigger of `updateSelectedDevice`  action will remove that flag (see screenshot console logs)

im not sure if the comment in [deviceSettingsActions](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/actions/settings/deviceSettingsActions.ts#L136-L142) is still valid since DeviceDisconnectRequired was never shown. tagging @mroz22 here for further investigation :)
and @komret  since it also changes the behavior in suite

Screenshot from 2025-05-05 16-40-44

this fix will render a proper view: `DeviceDisconnectRequired` component in which i've opened TroubleshootingTips by default it looks confusing with closed
Screenshot from 2025-05-05 17-30-47

 

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/suite-wipe-device&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/suite-wipe-device&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/suite-wipe-device&lookback=7)

### DIFF
--- a/packages/suite/src/components/suite/PrerequisitesGuide/DeviceDisconnectRequired.tsx
+++ b/packages/suite/src/components/suite/PrerequisitesGuide/DeviceDisconnectRequired.tsx
@@ -2,6 +2,7 @@ import { Translation, TroubleshootingTips } from 'src/components/suite';
 
 export const DeviceDisconnectRequired = () => (
     <TroubleshootingTips
+        initiallyIsOpen={true}
         label={<Translation id="TR_DISCONNECT_YOUR_DEVICE" />}
         items={[
             {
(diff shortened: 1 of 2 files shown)


mroz22 (approved):
not sure if the webusb case is still valid. unfortunately no time to investigate now :(

komret (approved):

PR #18780, opened 2025-05-07
chore: extract WipeDevice modal form UserContextModal, its not needed to have it global

peter-sanderson:
This is meant as an example how I plan to get rid of the `UserContextModal` completely. For situations, where the modal does NOT need to be global, this is the way.

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=refactor-wipe-modal&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=refactor-wipe-modal&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=refactor-wipe-modal&lookback=7)


suite-common/suite-types/src/modal.ts
```diff
@@ -98,18 +98,12 @@ export type UserContextPayload =
...
-          type: 'wipe-device';
-      }
     | {
           type: 'device-authenticity-check-opt-out';
       }
     | {
           type: 'firmware-authenticity-checks-opt-out';
       }
-    | {
-          type: 'disconnect-device';
```
peter-sanderson, line 111: This was unused, probably some forgotten leftover.

vojtatranta:
@seibei-iguchi may know, wherea are the modals that are local-only.

komret (dismissed):
If we're going to introduce a new pattern like this, I'd like to have some unified API for it, e.g. a `useModal` hook. I don't mind this approach, but it is not consistent and may appear rather random to someone looking at this code in future - as if some developer didn't know `UserContextModal` existed.

packages/components/src/components/Modal/useModal.tsx
```diff
@@ -0,0 +1,17 @@
+import { useState } from 'react';
+
+import { ModalProps } from './Modal';
+
+export const useModal = <T extends ModalProps>(ModalComponent: React.ComponentType<T>) => {
```
vojtatranta, line 5: I think it could be like:
`<C extends React.ComponentType<any>>(ModalComponent: C)`

packages/components/src/components/Modal/useModal.tsx
```diff
@@ -0,0 +1,17 @@
...
+    const [isModalOpen, setIsModalOpen] = useState(false);
+
+    const openModal = () => setIsModalOpen(true);
+
+    return {
+        openModal,
+        Modal: (props: Omit<T, 'onCancel'>) =>
```
vojtatranta, line 12: `Modal: (props: Omit<C['props', 'onCancel'>) =>` in order to mantain  different props than the specified.

But I am not sure. Point is to not enforce just one set of props.

packages/components/src/components/Modal/useModal.tsx
Lemonexe: But the current solution relies on `ModalComponent onCancel`, which is declared in `ModalProps`. By allowing any component props, this would either break TS or be type unsafe, right?

packages/components/src/components/Modal/useModal.tsx
```diff
@@ -0,0 +1,17 @@
...
+            isModalOpen ? (
+                <ModalComponent {...(props as T)} onCancel={() => setIsModalOpen(false)} />
+            ) : null,
+    };
+};
```
Lemonexe, line 17: I am personally not convinced that this helper saves time.
It removed boilerplate that is IMHO clearly understandable, no lines of code were saved and instead there is a more complex abstraction.

> but it is not consistent and may appear rather random to someone looking at this code in future

This is a good argument though.

Lemonexe: :thought_balloon:  But what if, theoretically `WipeDevice.tsx` needs to do some custom behavior in `onCancel`, not just set state false? You could solve it with composition: do not omit onCancel from props, and instead:
`onCancel={() => { setIsModalOpen(false); onCancel?.()  }}`

Maybe that won't be ever necessary.. But maybe "improvements" like this will pop up and increase the complexity of this abstraction..

Lemonexe (approved):
I'd prefer it without the last two commits ([comments](https://github.com/trezor/trezor-suite/pull/18780#discussion_r2081818107)), but I'm not against it :shrug: 
Works as intended and I very much like the pattern of moving Modal state to local where it's practical :heart_hands: 

The flow is not 1:1 (WipeDeviceModal is now visible in backdrop of ConfirmOnDevice, while before it was replaced), but I believe that to be an improvement – the visual "verticality" of modals shows the history where you came from one modal to another – describes the meaning of ConfirmOnDevice  :+1:

packages/components/src/components/Modal/useModal.tsx
peter-sanderson: I agree with @Lemonexe

packages/components/src/components/Modal/useModal.tsx
peter-sanderson: fixed

vojtatranta: Call me crazy, but I'd expect this to be working well:
```ts
export const useModal = <T extends ModalProps>(ModalComponent: React.ComponentType<T>) => {
    const [isModalOpen, setIsModalOpen] = useState(false);

    const openModal = () => setIsModalOpen(true);

    return {
        openModal,
        Modal: (props: Omit<T, 'onCancel'>) =>
            isModalOpen ? (
                <ModalComponent {...(props as T)} onCancel={() => setIsModalOpen(false)} />
            ) : null,
    };
};

function TestModal({ myProp }: { myProp: string }) {
    return (
        <Modal
            variant="primary"
            heading={'Modal heading' + myProp}
            description="Modal description"
            onCancel={() => {}}
        />
    );
}

function MyTestModal() {
    const { openModal, Modal: RenderModal } = useModal(TestModal);

    return (
        <>
            <button onClick={openModal}>Open modal</button>
            <RenderModal />
        </>
    );
}
```

But it throws Ts error at ` const { openModal, Modal: RenderModal } = useModal(TestModal);`

Which is strange

peter-sanderson:
in favor of https://github.com/trezor/trezor-suite/pull/18880

PR #18788, opened 2025-05-07
fix(suite-native): crash related to RefreshControl

martykan:
## Description

Fix crash on iOS due to `'NSInternalInconsistencyException', reason: 'we should only have exactly one subview'`

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/devicemanager-scrollview-one-child&lookback=7)

PR #18798, opened 2025-05-08
chore: unify suite and native wallet settings

martykan:
## Description

The goal of this PR is to move Suite's WalletSettings reducer into Suite Common so it can be shared between all platforms. 
It replaces duplicate functionality present in AppSettings and Discovery reducers in Native. 
This is part of the Discovery in Connect initiative, so we can eventually unify Discovery as well. 

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/unify-suite-wallet-settings&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/unify-suite-wallet-settings&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/unify-suite-wallet-settings&lookback=7)


suite-common/wallet-core/src/settings/walletSettingsActions.ts
```diff
@@ -27,53 +26,66 @@ export const changeNetworks = createAction(
...
-        const isAlreadyHidden = enabledNetworks.find(enabledSymbol => enabledSymbol === symbol);
-        if (!shouldBeVisible) {
-            enabledNetworks = enabledNetworks.filter(enabledSymbol => enabledSymbol !== symbol);
-        } else if (!isAlreadyHidden) {
-            enabledNetworks = [...enabledNetworks, symbol];
-        }
-        dispatch(changeNetworks(enabledNetworks));
+        /*if (!getState().suite.flags.discreetModeCompleted) {
+            dispatch(suiteActions.setFlag('discreetModeCompleted', true));
+        }*/
```
Lemonexe: should be uncommented and wrapped with if. 
Though idk if this flag is useful at all

suite-common/wallet-core/src/settings/walletSettingsActions.ts
```diff
@@ -27,53 +26,66 @@ export const changeNetworks = createAction(
...
-    };
+    },
+);
+
+export const changeCoinVisibility = createThunk<
+    void,
+    {
+        symbol: NetworkSymbol;
+        shouldBeVisible: boolean;
+    },
```
Lemonexe, line 62: thanks! Props nicer than positional arguments :+1:

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export type WalletSettingsRootState = {
+    wallet: {
+        settings: State;
+    };
+};
+
+const initialState: State = {
+    localCurrency: 'usd',
+    discreetMode: false,
```
PeKne, line 20: Mobile uses Jotai atom for discreet mode [see this](https://github.com/trezor/trezor-suite/blob/1216f23bb12368a71160f5ba8bb97deae99648d6/suite-native/atoms/src/DiscreetText/useDiscreetMode.ts#L5-L6). I suggest to remove it and use this redux value instead. Could you change it? Or I can help with that. 

I would not bother with preparing a complicated migration between jotai and redux. If the value is reset by this commit, it is very easy for the users to turn to discreet mode again.

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
```diff
@@ -0,0 +1,110 @@
...
+    'hideSuspiciousTransactions',
+    'bitcoinAmountUnit',
+];
+
+export const prepareWalletSettingsReducer = createReducerWithExtraDeps(
+    initialState,
+    (builder, extra) => {
+        builder.addCase(extra.actionTypes.storageLoad, extra.reducers.storageLoadWalletSettings);
+        builder.addCase(walletSettingsActions.setLocalCurrency.type, (state, action) => {
+            if (walletSettingsActions.setLocalCurrency.match(action)) {
```
PeKne: Are not these `if`s redundant? the case should never run if it won't be matching, right?

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
```diff
@@ -0,0 +1,110 @@
...
+    'discreetMode',
+    'enabledNetworks',
+    'hideSuspiciousTransactions',
+    'bitcoinAmountUnit',
+];
+
+export const prepareWalletSettingsReducer = createReducerWithExtraDeps(
+    initialState,
+    (builder, extra) => {
+        builder.addCase(extra.actionTypes.storageLoad, extra.reducers.storageLoadWalletSettings);
```
Lemonexe, line 39: I see that `extra.reducers.storageLoadWalletSettings` is only defined in desktop, but shouldn't it be in native too?

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
```diff
@@ -0,0 +1,110 @@
...
+            if (walletSettingsActions.setLocalCurrency.match(action)) {
+                const { localCurrency } = action.payload;
+                state.localCurrency = localCurrency;
+            }
+        });
+        builder.addCase(
+            WALLET_SETTINGS.SET_HIDE_BALANCE,
+            (
+                state,
+                action: { type: typeof WALLET_SETTINGS.SET_HIDE_BALANCE; toggled: boolean },
```
Lemonexe: nit: this'd read easier if the type was extracted.

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
```diff
@@ -0,0 +1,110 @@
...
+            }
+        });
+        builder.addCase(
+            WALLET_SETTINGS.SET_BITCOIN_AMOUNT_UNITS,
+            (
+                state,
+                action: {
+                    type: typeof WALLET_SETTINGS.SET_BITCOIN_AMOUNT_UNITS;
+                    payload: PROTO.AmountUnit;
+                },
```
Lemonexe: same nit: this'd read easier if the type was extracted.

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -920,10 +920,7 @@ export const selectDeviceById = createMemoizedSelector(
     (devices, deviceId) => devices.find(device => device.id === deviceId),
 );
 
-export const selectDeviceAuthenticity = createMemoizedSelector(
-    [state => state.device.deviceAuthenticity],
-    deviceAuthenticity => deviceAuthenticity,
-);
+export const selectDeviceAuthenticity = (state: DeviceRootState) => state.device.deviceAuthenticity;
```
Lemonexe: makes sense, shallow reference comparison done by `useSelector` should be enough here :+1:

Lemonexe (approved):
For **Suite Web/Desktop**: I reviewed carefully and LGTM, except one tiny leftover in [comment](https://github.com/trezor/trezor-suite/pull/18798#discussion_r2090670637). The settings work correctly at quick glance, incl. persistence.

Someone else should review the mobile part.

suite-common/wallet-core/src/settings/walletSettingsActions.ts
martykan: Seems like it's actually never used

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
martykan: No, Native uses `redux-persist` for storage

suite-common/wallet-core/src/device/deviceReducer.ts
martykan: I was getting a warning on native that the memoization is useless in this case

suite-native/device-manager/src/components/DeviceManagerContent.tsx
```diff
@@ -98,27 +98,31 @@ export const DeviceManagerContent = () => {
...
-                            style={applyStyle(deviceButtonsStyle)}
-                        >
-                            <DeviceSettingsButton
-                                showAsFullWidth={!isAddHiddenWalletButtonVisible}
-                            />
-                            {isAddHiddenWalletButtonVisible && <AddHiddenWalletButton />}
-                        </Stack>
-                    </VStack>
-                )}
+                <VStack spacing="sp24">
```
PeKne, line 101: Is this related? Is it some UI discrepancy that you found? :D

suite-native/settings/src/settingsSlice.ts
```diff
@@ -1,59 +1,48 @@
...
 
 export type SettingsSliceRootState = {
     appSettings: AppSettingsState;
 };
 
 export const appSettingsInitialState: AppSettingsState = {
-    fiatCurrencyCode: 'usd',
-    bitcoinUnits: PROTO.AmountUnit.BITCOIN,
     isOnboardingFinished: false,
+    isCoinEnablingInitFinished: false,
```
PeKne, line 21: What are the benefits of moving these two values from the `discoveryConfigSlice`? You plan to get rid of `discoveryConfigSlice` completely in some followup? If so, then it make sense and I agree.

suite-native/discovery/src/discoveryConfigSlice.ts
```diff
@@ -147,39 +103,12 @@ export const selectSupportedNetworkSymbols = createMemoizedSelector(
...
-export const selectEnabledDiscoveryNetworkSymbols = (state: DiscoveryConfigSliceRootState) =>
-    state.discoveryConfig.enabledDiscoveryNetworkSymbols;
-
 // this includes only networks supported by current device
 export const selectDeviceEnabledDiscoveryNetworkSymbols = createMemoizedSelector(
-    [selectDiscoveryNetworkSymbols, selectEnabledDiscoveryNetworkSymbols],
+    [selectDiscoveryNetworkSymbols, selectEnabledNetworks],
     (networkSymbols, enabledSymbols) =>
         returnStableArrayIfEmpty(networkSymbols.filter(s => enabledSymbols.includes(s))),
 );
```
PeKne, line 116: Was it discussed with someone that this selector won't be used anymore (could you confirm @vytick :pray:)? Usage of this selecotor was delete from `initTokenDefinitionsThunk`. If so, please delete `selectNetworkSymbolsOfAccountsWithTokensAllowed` because it is not used anywhere else after this change.

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
```diff
@@ -0,0 +1,110 @@
...
+        settings: State;
+    };
+};
+
+const initialState: State = {
+    localCurrency: 'usd',
+    discreetMode: false,
+    // Suite Lite did not have BTC enabled by default
+    enabledNetworks: isNative() ? [] : ['btc'],
+    hideSuspiciousTransactions: false,
```
PeKne, line 23: This state mobile do not need at all. But it is not a big issue.

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
martykan: Yeah probably bad copy pasting

suite-native/device-manager/src/components/DeviceManagerContent.tsx
martykan: I was getting a crash "ScrollView can only have one child"
I eventually found out it's [due to something else](https://github.com/trezor/trezor-suite/pull/18788), but others told me they can't reproduce this error

suite-native/settings/src/settingsSlice.ts
martykan: Yes eventually we plan to kill it

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
martykan: Ok so actually the reason is that for some reason without it the type is not narrowed properly

suite-common/wallet-core/src/settings/walletSettingsActions.ts
martykan: Moved to middleware

suite-common/wallet-core/src/settings/walletSettingsReducer.ts
martykan: Done in https://github.com/trezor/trezor-suite/pull/18798/commits/d09cd325edaa3d01bd059c5f413446a6c46ec1a6

suite-native/discovery/src/discoveryConfigSlice.ts
vytick: `selectTokenDefinitionsEnabledNetworks` is needed so token definitions are downloaded for portfolio tracker. If we only use discovery symbols we might not have token definitions for some imported accounts. 

Example:

User has enabled `BTC` for discovery. In portfolio account he imports `ETH` account. But he has not enabled eth account for discovery meaning that `ETH` will not be considered in `initTokenDefinitionsThunk` and no token definitions will be downloaded and this also means that no fiat rates will be fetched for tokens. Using `selectNetworkSymbolsOfAccountsWithTokensAllowed ` to get network symbols of all imported accounts  (including portfolio tracker ones)  and checking which network has enabled tokens in mobile app, fixes this.

TLDR: no `selectTokenDefinitionsEnabledNetworks` => possibly no tokens and fiat rates for imported accounts

vytick: There is also another edgecase that is being resolved by use of this selector:

`ADA` (and maybe also `ETC`) tokens are NOT shown in the mobile app by decision. But if we only use discovery enabled symbols and filter out Cardano, it will show them on some places of the app inconsistently. Account import will show them including fiat rate through the process of importing and on some places it will be counted into account balance, but not on other places

TLDR2: not using `isCoinWithTokens` filter makes inconsistency with `ADA`

martykan: Got it. Will bring it back

PeKne: Let me know when it is ready for final review check

martykan: Token defs selector [re-added here](https://github.com/trezor/trezor-suite/pull/18798/commits/5e241d475bd130f778a91dc41e7104a9f17cfb7b)

suite-common/token-definitions/src/tokenDefinitionsThunks.ts
```diff
@@ -3,6 +3,7 @@ import { decode, verify } from 'jws';
 
 import { createThunk } from '@suite-common/redux-utils';
 import { NetworkSymbol, getCoingeckoId } from '@suite-common/wallet-config';
+import { selectTokenDefinitionsEnabledNetworks } from '@suite-native/discovery';
```
PeKne: Direct import from @suite-native to @suite-common is not possible. Fix this please.

martykan: Added back to extra https://github.com/trezor/trezor-suite/pull/18798/commits/87bd9050ef24feca2f4c28f3bfc83490ea3316e2

PeKne (approved):

PR #18810, opened 2025-05-09
fix(suite-native): Data Analytics - Coin Discovery

### LINKED ISSUE #16956: Data Analytics: Coin Discovery fix and update
At the moment, [Coin Discovery](https://www.notion.so/satoshilabs/coin_discovery-c2f2fd572d064470a5bb19e28faf4464?pvs=4) event does not have a meaningful payload. It reports even unused empty accounts, and hidden accounts, meaning in case anyone enables btc, we discover at least 8 (2 for each of 4 different account types). Instead we should consider only accounts with some transaction history and count them once. Additionally, number of accounts with non-zero balance would be useful. So we would have:

- (old) parameter `numberOfAccounts` : now with only accounts with some transaction history counted once
- (new) parameter `numberOfNon-zeroAccounts` : with accounts with non-zero balance

BrantalikP:
## Description

- Changed: **numberOfAccounts** now counts only accounts with transaction history
- Added: **numberOfNonZeroAccounts** – counts accounts (with transaction history) that have a non-zero balance.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16956

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/analytics-coin-discovery&lookback=7)

### DIFF
--- a/suite-native/discovery/src/discoverySelectors.ts
+++ b/suite-native/discovery/src/discoverySelectors.ts
@@ -80,7 +80,13 @@ export const selectDiscoveryAccountsAnalytics = (
         selectDeviceAccounts(state),
         A.groupBy(account => account.symbol),
         D.mapWithKey((symbol, accounts) => {
-            const numberOfAccounts = accounts?.length ?? 0;
+            const accountsWithTransactionHistory = accounts?.filter(acc => !acc.empty) || [];
+
+            const numberOfAccounts = accountsWithTransactionHistory?.length;
+
+            const numberOfNonZeroAccounts = accountsWithTransactionHistory?.filter(
+                acc => parseFloat(acc.balance) > 0,
+            ).length;
 
             const validTokens = selectValidTokensByDeviceStateAndNetworkSymbol(
                 state,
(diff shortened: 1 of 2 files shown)


suite-native/discovery/src/discoverySelectors.ts
```diff
@@ -80,7 +80,13 @@ export const selectDiscoveryAccountsAnalytics = (
         selectDeviceAccounts(state),
         A.groupBy(account => account.symbol),
         D.mapWithKey((symbol, accounts) => {
-            const numberOfAccounts = accounts?.length ?? 0;
+            const accountsWithTransactionHistory = accounts?.filter(acc => acc.history?.total > 0);
```
PeKne: This check is tricky, because it will not work for XRP accounts (these will have `history.total` always equal to 0), see this line: https://github.com/trezor/trezor-suite/blob/385740d4c91f6c1e235a559a179f3414218a90de/packages/blockchain-link-types/src/common.ts#L236-L237. 

i am afraid you will need to check the transaction (but not sure if they will be be contained in this data already, you have to test it). See this selector for inspiration: https://github.com/trezor/trezor-suite/blob/385740d4c91f6c1e235a559a179f3414218a90de/suite-native/transactions/src/selectors.ts#L167-L178

PeKne (changes requested):

PeKne:
Please add these changes to the [analytics changelog](https://www.notion.so/satoshilabs/Mobile-app-data-analytics-changelog-946307159a294bcdb90b94e90eefd292) so the data team has up to date info. 

Also don't forget to mention the changes of the coin_discovery event changes here: https://www.notion.so/satoshilabs/coin_discovery-c2f2fd572d064470a5bb19e28faf4464?pvs=4.

Thanks!

PeKne (approved):

PR #18814, opened 2025-05-09
fix(suite): can't change device name from onboarding

### LINKED ISSUE #14981: Onboarding: Rename wallet step improvements
At the end of the onboarding, when renaming the device:

- After entering the edit mode (by clicking the [Edit name] button),
     - Rename the [Edit name] button to [Save name]
     - Instead of the default name being a placeholder, please make it pre-filled input content and all of it selected and in focus
          - That means when user starts typing right away, the content will be replaced by theirs
          - If they click into the content or use arrow keys, they will be able to edit the content
     - **Save the new name even when clicking the [Access Suite –>] button**
     - Add 'esc' keyboard button functionality to close the input without saving

Please mind feasibility. If it would be effort-heavy, let's at least do the bold one ☝️ and discuss the rest.

Current implementation:
Image

viktordanko:
## Description

Improves changing of the name of the device while onboarding.

- whenever user click on `Access suite` button, current name is saved
- add option to exit rename mode, when pressing `Esc` key

## Related Issue

Resolves:  #14981 

## Screenshots:
onboarding


packages/suite/src/views/onboarding/steps/Final.tsx
```diff
@@ -120,16 +124,55 @@ export const FinalStep = () => {
...
+
+        try {
+            await onSubmit();
+            goToSuite();
+        } catch (error) {
+            console.error(error);
+            dispatch(
+                notificationsActions.addToast({
+                    type: 'error',
+                    // TODO: replace with `translationString('TR_DEVICE_SETTINGS_DEVICE_EDIT_LABEL_ERROR')`,
```
viktordanko: What's the workflow for generating new messages for `L10N`, pls?
cc @komret

packages/suite/src/hooks/suite/useChangeDeviceLabel.ts
```diff
@@ -0,0 +1,55 @@
...
+export const useChangeDeviceLabel = () => {
+    const { translationString } = useTranslation();
+    const deviceLabel = useSelector(selectDeviceLabel);
+    const deviceName = useSelector(selectDeviceName);
+    const dispatch = useDispatch();
+
+    const form = useForm({
+        resolver: yupResolver(changeDeviceLabelSchema(translationString)),
+        defaultValues: {
+            deviceLabel: deviceLabel ?? deviceName,
```
viktordanko: This was a bit confusing to me, since I'm missing the context.
I assume `label !== name`, but don't know what has higher prio

packages/suite/src/views/onboarding/steps/Final.tsx
```diff
@@ -197,29 +240,19 @@ export const FinalStep = () => {
                     )}
                     {state === 'rename' && (
                         <SetupActions>
-                            <ChangeDeviceLabel onClick={onClick} isDeviceLocked={isDeviceLocked} />
+                            <FormProvider {...form}>
+                                <ChangeDeviceLabelForm
+                                    isDeviceLocked={isDeviceLocked}
+                                    onClick={handleClick}
```
komret: This takes me to Suite settings after I click "Change name", but I would expect to stay on the Final screen so that I can set homescreen.

komret (commented):
Thanks for the contribution! Can you please separate the changes into more commits, one commit per feature? The ESC functionality is undoubtedly a nice improvement, removing the placeholder is OK, but I am not so keen on changing the device name after clicking the "Go to Suite" button as it requires confirmation on Trezor. Unfortunately, the issue is quite old and I am not sure as to its validity. Personally, I'd rather remove this whole screen from onboarding. @Hermez-cz, WDYT?

packages/suite/src/hooks/suite/useChangeDeviceLabel.ts
komret: This is correct. Label is set by user, while name is defined by the device. Perhapst you can use `selectSelectedDeviceLabelOrName` selector here.

packages/suite/src/views/onboarding/steps/Final.tsx
komret: A new message should be added to `messages.ts`. But I wouldn't handle errors here, we never had an issue here, probably not worth adding new translation strings.

packages/suite/src/hooks/suite/useChangeDeviceLabel.ts
Lemonexe, line 1: Very nice that you've rewritten this form to `react-hook-form`, incl. `yup` instead of homecooked validation, so there will be better codestyle consistency with other forms :tada:

packages/suite/src/views/onboarding/steps/Final.tsx
```diff
@@ -120,16 +124,55 @@ export const FinalStep = () => {
...
+        };
+        delete payload.startTime;
+
+        analytics.report({
+            type: EventType.DeviceSetupCompleted,
+            payload,
+        });
+
+        try {
+            await onSubmit();
```
Lemonexe: this works well when changing a name, but needs to be conditionally called only if formstate.value is different from initial value. Now the device prompts me to change name from Hello to Hello even when I don't interact with the form: 
[Screencast from 2025-05-14 15-32-37.webm](https://github.com/user-attachments/assets/f8ce20c8-4a0d-4aff-9e40-988e5898500e)

packages/suite/src/views/onboarding/steps/Final.tsx
Lemonexe: In any case, the error while changing name effectively blocks you from proceeding to Suite.

packages/suite/src/views/onboarding/steps/Final.tsx
Lemonexe: Yes, the function would be best split into two.
`ChangeDeviceLabelForm onClick`: submit form, catch error to display notification (but don't go to Suite) 
`EnterSuiteButton onClick`: do this :arrow_up: if values have changed, and go to Suite regardless of success.

packages/suite/src/hooks/suite/useChangeDeviceLabel.ts
```diff
@@ -0,0 +1,55 @@
...
+        deviceLabel: yup
+            .string()
+            .max(
+                MAX_LABEL_LENGTH,
+                t('TR_LABEL_ERROR_LENGTH', {
+                    length: MAX_LABEL_LENGTH,
+                }),
+            )
+            .test({
+                test: value => isAscii(value),
```
Lemonexe: small nit to simplify:
```suggestion
                test: isAscii,
```

packages/suite/src/views/settings/SettingsDevice/DeviceLabel.tsx
```diff
@@ -1,23 +1,40 @@
...
     isDeviceLocked: boolean;
-}
+};
+
+export const DeviceLabel = ({ isDeviceLocked }: DeviceLabelProps) => {
+    const { form, onSubmit } = useChangeDeviceLabel();
+
+    const handleClick = async () => {
+        await onSubmit();
+    };
```
Lemonexe: I believe you can delete this and pass `onSubmit` directly to `ChangeDeviceLabelForm`, right?

Lemonexe (commented):
Thank you for your contribution :raised_hands: 
Overall very nice work, the user flow specified by issue is works well :heavy_check_mark: 

But in order for this PR to be mergable, other user flows should be fixed, particularly [this](https://github.com/trezor/trezor-suite/pull/18814#discussion_r2084545728) and [this](https://github.com/trezor/trezor-suite/pull/18814#discussion_r2088964300) comments.

Lemonexe:
Oh and @Hermez-cz confirmed that the issue is indeed relevant.
FWIW to me it definitely makes sense, as I've found this very confusing in past when I first interacted with it.

packages/suite/src/hooks/suite/useChangeDeviceLabel.ts
viktordanko: Nice, that should do the trick. Thanks 🙏

packages/suite/src/views/onboarding/steps/Final.tsx
viktordanko: Good catch, thanks!

packages/suite/src/views/onboarding/steps/Final.tsx
viktordanko: Okay, np. Gonna remove it

packages/suite/src/views/onboarding/steps/Final.tsx
viktordanko: That's right, thx!

packages/suite/src/views/settings/SettingsDevice/DeviceLabel.tsx
viktordanko: Absolutely :]

viktordanko:
Thanks for your valid inputs 🙇 
I overlooked the conditional submitting of the form at first. Now, it should be all fine.
Here's separate PR for resetting the state when pressing `Esc` button - https://github.com/trezor/trezor-suite/pull/18954

Lemonexe (approved):
Perfect, all user flows now work correctly, LGTM :rocket: 
I'm gonna force push to resolve conflicts with `develop` and also because merge commits are not allowed in the repo.

Like with the other PRs, I'm gonna run CI checks in another PR #18968 and then merge this one.

EDIT: CI passed except for tests temporarily confirmed broken ([slack](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1747318218774009))

bosomt:
**QA OK**

- escape cancels edit
- after i change name and select go to TS im asked to change device name upon entering to TS

**Info:**
 - Suite version: web 25.6.0 (3a8214d7953b03f3648e9bf592d1ce7b16e8cd99)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1470x956
 - Device: Trezor T3T1 2.8.8 regular (revision bbcb920213b0ab847a2fedf825eac70652e33279)
 - Transport: WebUsbTransport

tomasklim:
@bosomt renamed `Change name` to `Save name` in crowdin

@viktordanko you can edit `messages.ts` and `en.json` next time just to have it correct in the screenshot without syncing crowdin

PR #18821, opened 2025-05-09
feat: firmware update using firmware release config

karliatto:
## Description
This Pull Request completes the changes to allow to use the new `firmware-release-config`  for firmware update in suite. This makes possible to have a new firmware update without releasing a new Suite.

* Add new firmware release config signed file that will be stored bundled into the app and fetched online with information about latest available FW and some conditions that Suite has to meet in order for that new FW to be offered.
* When new FW is available Suite will download it and in the case of Desktop app it will store it for later usage.
* In order to achieve these things new format of the old `releases.json` that was existing for each device model has been changed to be separated files each for each firmware by doing that we are separating Universal from Bitcoin Only firmware and it allows us to release one of them without releasing both.
* Many of the changes are in one commit that is changing the structure of the previously mentioned `releases.json` in commit with name `chore(connect-common): releases JSON to new format`. This commit is still subject of potential changes in the structure of the JSON files

## What to test

- test all the flows (from factory reset, from normal from bitcoin only to universal from universal to bitcoin only with all devices model
- test with and without internet
- test in web and desktop (for now only desktop since the testing server needs to be updated to provide header for CORS)
- test if downloading and storing and reusing FW works in desktop

### On-boarding flow
1. Reset device to factory so it can be used to install unsigned binaries
2. Open the provided testing version of suite and go throw on-boarding installing firmware (online and offline)
3. Firmware installation should complete properly and should be possible to complete on-boarding.

### Firmware update flow
1. Prepare device with older unsigned firmware installing it manually with `trezorctl` or Trezor Suite custom FW installation. 
2. It will  require to do on-boarding in this device skipping the firmware update during on-boarding
3. Once on-boarding is completed firmware update should be offered, click on it and complete the FW update process

### Update in bootloader mode
1. Prepare device with older unsigned firmware installing it manually with `trezorctl` or Trezor Suite custom FW installation. 
2. With device in bootloader mode start Suite and go to Settings > Device  > Firmware > Version updata
3. It should be possible to update FW to latest from bootloader.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/firmware-update&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/firmware-update&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/firmware-update&lookback=7)


packages/connect/src/types/device.ts
```diff
@@ -90,7 +90,7 @@ export type KnownDevice = BaseDevice & {
     label: string;
     error?: typeof undefined;
     firmware: DeviceFirmwareStatus;
-    firmwareRelease?: ReleaseInfo | null;
+    firmwareReleaseConfig?: FirmwareReleaseConfigInfo;
```
mroz22: this is potentially breaking but I don't think 3rd parties are supposed to use it

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/QuickActions/Update/useUpdateStatus.ts
```diff
@@ -70,15 +73,23 @@ export const useUpdateStatus = (): UpdateStatusData => {
...
     const isSuiteUpdateInProgress = [UpdateState.Downloading, UpdateState.Checking].includes(
         desktopUpdate.state,
     );
 
+    const { releaseConditions: { environment, shouldBeOffered } = {} } =
+        device?.firmwareReleaseConfig || {};
+
+    const isValidSuiteVersion =
+        !isDeviceDisconnected &&
+        versionUtils.isNewerOrEqual(getSuiteVersion(), environment?.min_suite_version);
```
mroz22: isValidSuiteVersion variable name based on isDeviceDisconnected  does not sounds logic. nitpick

mroz22 (commented):
I am reading it, I promise :D

mroz22 (commented):
btw to make it work, I need to run 
```
yarn sign-release-config
```
is it documented somewhere? can we somehow make this implicit/automatic?

mroz22:
i tried to run it locally and got this 
image

karliatto:
> btw to make it work, I need to run
> 
> ```
> yarn sign-release-config
> ```
> 
> is it documented somewhere? can we somehow make this implicit/automatic?

You should only run `yarn build:libs` or `yarn workspace @trezor/firmware-release-config build:lib` it works the same way as the message system. But yes, I will add documentation to the whole thing and process once is set into stone.

packages/suite/src/components/suite/layouts/SuiteLayout/Sidebar/QuickActions/Update/useUpdateStatus.ts
karliatto: The reason of this is that when device is not connected `environment?.min_suite_version` is undefined and when you start the process of flashing firmware since it reboots from Firmwere mode to Bootloader mode there is a moment when device is "disconnected" and therefore that fails.

I could add a comment there, or maybe find other solution.

packages/connect/src/types/device.ts
karliatto: Yes, here I preferred doing a clear change by changing the name of the attribute so we make clear it is something different than the previous one before.  But yes it is a breaking change so we could properly state in changelog and make a minor release next time.

mroz22:
Unfortunately, yarn build:libs still fails with many errors.

karliatto:
> Unfortunately, yarn build:libs still fails with many errors.

I just cloned a new trezor-suite in order to avoid potential file cached in my working directory and run `yarn && yarn build:libs && yarn suite:dev:desktop` and it was working ok for me. 

What errors do you get in terminal? Can you share those?

karliatto:
> Unfortunately, yarn build:libs still fails with many errors.

Ok, the issue was with `suite:web` since I was using a shared file with the scripts for `firmware-release-config` package that was using some nodejs utils that were of course not loading in web. 


I added new commit `wip: fix the tests` just for now until I fix the tests to avoid having typescript errors regarding those.

packages/connect/src/types/device.ts
mroz22: I think this is ok.

packages/firmware-release-config/src/index.ts
```diff
@@ -12,7 +11,7 @@ import { jws as releasesJwsLocal } from '../files/releases.v1';
 // TODO: WIP: for now we are foring local since it was not deployed yet.
 const FORCE_LOCAL_JWS = true;
 
-export const getFirmwareReleaseConfig = async () => {
+export const getReleaseJWS = async () => {
```
mroz22: this is exported but only used locally

packages/firmware-release-config/src/index.ts
```diff
@@ -54,8 +53,8 @@ export const getFirmwareReleaseConfig = async () => {
     }
 };
 
-export const getReleasesMessage = async () => {
-    const { releasesJws } = await getFirmwareReleaseConfig();
+export const getFirmwareReleaseConfig = async () => {
```
mroz22: this interanlly uses "isCodesignBuild" utility and is called from inside connect (parseFirmwareReleaseConfig). Are ou sure that this env variable is set correctly for all thinkable environments where connect might be used?

packages/firmware-release-config/package.json
```diff
@@ -5,15 +5,24 @@
...
+    ],
     "scripts": {
-        "build:lib": "yarn validate-release-config && yarn sign-release-config",
+        "build:lib": "yarn sign-release-config && yarn g:rimraf -rf lib && yarn g:tsc --build tsconfig.lib.json && ../../scripts/replace-imports.sh ./lib",
         "depcheck": "yarn g:depcheck",
         "lint:js": "yarn g:eslint '**/*.{ts,tsx,js}'",
         "test:unit": "yarn g:jest",
         "type-check": "yarn g:tsc --build",
-        "generate-release-config": "yarn tsx ./scripts/generate-release-config.ts",
+        "generate-release-config": "yarn validate-release-confi yarn tsx ./scripts/generate-release-config.ts",
```
mroz22: typo? `validate-release-confi` and missing &

packages/connect/src/device/Device.ts
```diff
@@ -770,64 +775,70 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
+        const firmwareLocation = getFirmwareLocation({
+            firmwareVersion,
+            internalModel: this.features.internal_model,
+            firmwareType: this.firmwareType ? this.firmwareType : FirmwareType.Regular,
+        });
 
-            return createFailResult('check-unsupported');
-        }
+        const enabled = DataManager.getSettings('enableFirmwareHashCheck');
+        if (!enabled || !firmwareLocation) return createFailResult('check-skipped');
```
mroz22: per typescript, firmwareLocation can't be undefined here.

packages/connect/src/device/Device.ts
mroz22, line 1: changes in this file might interest @Lemonexe

packages/connect/src/types/firmware.ts
```diff
@@ -9,25 +12,20 @@ export type FirmwareRange = Record<
...
+    binaryVersion: VersionArray;
+    releaseVersion?: VersionArray;
+};
 
-export type ReleaseInfo = {
-    changelog: FirmwareRelease[] | null;
-    release: FirmwareRelease;
+export type FirmwareReleaseConfigInfo = {
+    firmwareType: FirmwareType;
+    // TODO: maybe this is not necessary anymore since we have bitcoin-only FW for all models.
```
mroz22: well, if it is always going to be set to true, then it might be removed 🤔 🤷

packages/suite/src/components/firmware/FirmwareInitialStandalone.tsx
```diff
@@ -61,7 +61,7 @@ export const FirmwareInitialStandalone = ({
     }
 
     // Bitcoin-only firmware is only available on T2T1 from v2.0.8 - older devices must first upgrade to 2.1.1 which does not have a Bitcoin-only variant
-    const isBitcoinOnlyAvailable = !!device.firmwareRelease?.release.url_bitcoinonly;
+    const isBitcoinOnlyAvailable = !!device.firmwareReleaseConfig?.isBitcoinOnlyAvailable;
```
mroz22: well, so maybe `isBitcoinOnlyAvailable` is still needed - there was this concept of 'incremental update' for older model Ts. it was not possible to go directly to the latest one. but maybe this concept has been obsoleted by some intermediary fw? who knows

submodules/trezor-common
mroz22, line 1: probably no need to bump common in this PR? right?

packages/connect/src/api/firmware/getBinary.ts
```diff
@@ -1,26 +1,22 @@
-import type { FirmwareRelease } from '@trezor/device-utils';
-
 import { httpRequest } from '../../utils/assets';
 
 const ALL_SLASHES_AT_THE_END_REGEX = /\/+$/;
```
mroz22: this regex can be found also here https://github.com/trezor/trezor-suite/blob/feat/firmware-update/packages/suite-desktop-core/src/libs/update-checker.ts#L18
but I am not sure if there is a meaningful way to share it

packages/connect/src/api/firmware/getBinary.ts
```diff
@@ -1,26 +1,22 @@
...
+interface GetBinaryParams {
     baseUrl: string;
-    btcOnly?: boolean;
-    release: FirmwareRelease;
+    firmwareName: string;
 }
 
-export const getBinary = ({ baseUrl, btcOnly, release }: GetBinaryProps) => {
-    const fwUrl = release[btcOnly ? 'url_bitcoinonly' : 'url'];
+export const getBinary = ({ baseUrl, firmwareName }: GetBinaryParams) => {
```
mroz22: this sanity check https://github.com/trezor/trezor-suite/blob/feat/firmware-update/packages/connect/src/core/onCallFirmwareUpdate.ts#L249 
is now duplicated in the code base. it might make sense to move it only here.

mroz22 (commented):
another batch of rather lower hanging comments

packages/connect/src/device/Device.ts
```diff
@@ -770,64 +775,70 @@ export class Device extends TypedEmitter<DeviceEvents> {
...
+                if (deviceResponse.message.hash !== expectedHash) {
+                    return createFailResult('hash-mismatch');
+                }
+
+                return { success: true };
+            } catch (errorPayload) {
+                return createFailResult('other-error', serializeError(errorPayload));
+            }
+        } catch (error) {
+            return createFailResult('other-error', serializeError(error));
```
Lemonexe: I am against wrapping more code in a try cach that can result in `other-error`.

Only code that throws because of aberrant _device_ behavior should throw `other-error`.
Can `getBinaryOptional` throw? If it does, it's probably not caused by device, right?
:point_right:  If we are sure that the binary should exist (it's a valid version), but it doesn't download, then IMO we should `createFailResult('check-unsupported')`, as we do currently.
But maybe I don't understand fully...

:thought_balloon:  it's not ideal that hash check relies on internet to check the device, but since we can't force users to get internet connection, we have to skip the check then.

packages/firmware-release-config/src/index.ts
karliatto: It should be available in suite-web and suite-desktop both environments where it is expected to be used.

It is also available in iframe (https://github.com/trezor/trezor-suite/blob/develop/packages/connect-iframe/webpack/base.webpack.config.ts#L112) but I do not think it is required there.

packages/connect/src/types/firmware.ts
karliatto: Yes, right now it always true, since all the models have a bitcoin-only version.

packages/suite/src/components/firmware/FirmwareInitialStandalone.tsx
karliatto: Interesting, I will get deeper into this scenario. There is no intermediary FW for model Ts so this might be an issue.

packages/connect/src/device/Device.ts
karliatto: Sure, makes sense. I added changes to get rid of that try...catch https://github.com/trezor/trezor-suite/pull/18821/commits/7e855a60219534d0a0424774e1a18c02d6a4c577#diff-ca03cc83fb0e3d98908b763de7901b240e15b59ba046942b514f98801f3a8afaL803

karliatto: > it's not ideal that hash check relies on internet to check the device 

We do not rely only on internet. With changes in this PR we are storing the binaries for future check, for now only if the user flashes the firmware from that machine.

packages/connect/src/api/firmware/getBinary.ts
karliatto: Do you mean the sanity check for `suspiciously small binary` ?

packages/connect/src/device/Device.ts
Lemonexe: Looks good now :+1: Pls let me know when the PR is ready for review, I'll be glad to test it :slightly_smiling_face: 

Yes, I wrote it badly. Only on Web we (currently) rely on downloading binaries from internet; sure, the user probably _is_ connected if they loaded Suite Web, but that fetch can fail (and does). On desktop, it fetches them currently locally (and I see that has not changed), and if they are not available it will skip the check (meaning: old fw version).

packages/connect/src/api/firmware/getBinary.ts
karliatto: What about something like this https://github.com/trezor/trezor-suite/pull/19109 ?

submodules/trezor-common
karliatto: Yes, my mistake. Removed.

packages/firmware-release-config/package.json
karliatto: Fixed https://github.com/trezor/trezor-suite/pull/18821/commits/681adbcf2fb191d1d66058d6dccf2623cea527ba

packages/firmware-release-config/src/index.ts
karliatto: `export` removed https://github.com/trezor/trezor-suite/pull/18821/commits/681adbcf2fb191d1d66058d6dccf2623cea527ba

packages/connect/src/api/firmware/getBinary.ts
mroz22: yes

packages/connect/src/device/Device.ts
karliatto: @Lemonexe I think it is good moment to review it. 
I still have some react-native tests failing that probably I need to update but I am working on it in the mean time.

packages/firmware-release-config/src/types.ts
```diff
@@ -32,10 +32,9 @@ export interface ReleasesConfig {
     T3W1: ConditionalRelease[];
 }
 export interface IntermediaryReleaseConfig {
-    if_version_less_than: string;
+    min_firmware_version: VersionArray;
```
mroz22: btw, can't the above type (ReleseCOnfig) be defined this way?

```
export type ReleasesConfig = Record<DeviceModelInternal, ConditionalRelease[]>;
```
so that when a future model comes, you don't need to add it to the definition manually?

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -18,9 +29,25 @@ const releases = Object.values(DeviceModelInternal).reduce(
...
+    FirmwareRelease
+>;
+
+export const parseFirmwareReleases = (
+    modelReleases: FirmwareRelease[],
+    deviceModel: DeviceModelInternal,
+) => {
+    if (!Array.isArray(modelReleases)) {
+        return;
+    }
```
mroz22: per typescript - this should never happen

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -18,9 +29,25 @@ const releases = Object.values(DeviceModelInternal).reduce(
...
+    const [latestRelease] = modelReleases;
```
mroz22: [latestRelease] would be undefined for Model.UNKNOWN. It probably does not matter?

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -44,220 +71,228 @@ export const getOnlineReleases = async (internalModel: DeviceModelInternal) => {
...
+        releases,
+        intermediaryReleases: firmwareIntermediaryReleasesConfig,
+    };
 };
 
-const isNewer = (release: FirmwareRelease, features: StrictFeatures) => {
-    if (features.major_version === 1 && features.bootloader_mode) {
-        return null;
-    }
+const getCurrentFirmwareVersion = (features: Features): VersionArray => {
```
mroz22: I think we already had this discussion but some older model Ones can't tell firmware version in the bootloader. 

Also maybe some comments to this fn would be nice. Code looks cryptic for people not familiar about this topic.

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -44,220 +71,228 @@ export const getOnlineReleases = async (internalModel: DeviceModelInternal) => {
...
+    const version = [major_version, minor_version, patch_version] as VersionArray;
+
+    return bootloader_mode && firmwareVersionInBootloaderMode
+        ? firmwareVersionInBootloaderMode
+        : version;
 };
 
-const isRequired = (changelog: ReturnType<typeof getChangelog>) => {
-    if (!changelog || !changelog.length) return null;
+const getCurrentBootloaderVersion = (features: Features): VersionArray => {
```
mroz22: personally, I don't like very much creating a method with a very specific name `getCurrentBootloaderVersion` that gives you current bootloader version only if called under specific circumstances (features.bootloader_mode === true)

At the first glance I thought that this would give me bootloader version of any connected device and went to create a screenshot to prove that this is not true only later to notice that it was not meant to be used like this.

image

mroz22: if it is meant to be specif I would call it `getCurrentBootloaderVersionForDeviceInBootloaderMode`

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -44,220 +71,228 @@ export const getOnlineReleases = async (internalModel: DeviceModelInternal) => {
...
+        intermediary: requiresIntermediary ? intermediary : undefined,
+        isRequired: required,
+        isNewer: isNewerResult,
+        translations: release.translations || [],
+    };
+};
 
-    const release = releasesParsed[0];
-    const prevRelease = releasesParsed[1];
+export const getFirmwareRelaseConfig = (features: Features, firmwareType: FirmwareType) => {
```
mroz22: type ...RelAse...

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -44,220 +71,228 @@ export const getOnlineReleases = async (internalModel: DeviceModelInternal) => {
...
-    }
+export const getReleaseConfig = (features: Features, firmwareType: FirmwareType) => {
+    const { internal_model } = features;
 
-    const bootloaderVersion = getT1BootloaderVersion(releases2, features);
-
-    if (versionUtils.isNewerOrEqual(bootloaderVersion, [1, 12, 0])) {
-        return 3;
+    if (internal_model === DeviceModelInternal.UNKNOWN) {
+        throw new Error('Missing Firmware release config for UNKNOWN device');
```
mroz22: What now happens if you connect an unknown device? It should somehow work - not fw update of course but connect should try to read the device and emit device event if possible. This was the purpose of unknown device. I did not check the entire call chain, maybe it is caught somewhere up

mroz22 (commented):
couple of notes. I continue reading and testing

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -44,220 +71,228 @@ export const getOnlineReleases = async (internalModel: DeviceModelInternal) => {
...
 
-    const firmwareVersion = [major_version, minor_version, patch_version];
+    const deviceMessageRelease = firmwareReleasesConfig[internal_model];
 
-    if (!versionUtils.isVersionArray(firmwareVersion)) {
-        return [];
+    return deviceMessageRelease.some(item => item.firmware_type === FirmwareType.BitcoinOnly);
+};
+
+export const isValidConditionalRelase = (
```
mroz22: typo

mroz22: plus it is exported but only used locally

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -44,220 +71,228 @@ export const getOnlineReleases = async (internalModel: DeviceModelInternal) => {
...
+    ) {
+        // Some old version of T1B1 doe not report FW version in bootloader mode.
+        const bootloaderVersion = getCurrentBootloaderVersion(features);
+        isNewerResult = versionUtils.isNewer(release.bootloader_version, bootloaderVersion);
+        requiresIntermediary = versionUtils.isNewer(min_bootloader_version, bootloaderVersion);
+    } else {
+        const firmwareVersion = getCurrentFirmwareVersion(features);
+        if (!versionUtils.isVersionArray(firmwareVersion)) {
+            throw new Error('Firmware version is not version array.');
+        }
```
mroz22, line 465: `getCurrentFirmwareVersion` is called 2 times in this file, but sanity check is carried out only once, here. Is it intended? maybe the sanity check chould be moved into `getCurrentFirmwareVersion`

packages/firmware-release-config/src/types.ts
karliatto: Sure, I changed it.

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -44,220 +71,228 @@ export const getOnlineReleases = async (internalModel: DeviceModelInternal) => {
...
-    if (major_version === 2 && bootloader_mode) {
-        const fwVersion = [fw_major, fw_minor, fw_patch];
-        if (versionUtils.isVersionArray(fwVersion)) {
-            // in bootloader, T2T1 or newer devices know their firmware, so we still may filter "by firmware".
-            return filterSafeListByFirmware(releases, fwVersion);
+    let valueToCompare: number;
+
+    if (deviceId === null) {
+        // If deviceId is null, use the original random behavior
+        valueToCompare = Math.random() * 100;
```
mroz22: This is primarily for newly unpacked devices during onboarding. We have about 3 choices here:

1. They always get update - ignore rollout
2. a] They take rollout into account based on random. This is the current approach. Are we ok with the fact that reloading the app could be used to skip rollout and force update?
2. b] Try to come up with something deterministic. But I can't see any good way to do it. 
3. Say that users going through onboarding never get the latest update (as long as rollout period for the latest fw is active). 

Was this considered product-wise? @MiroslavProchazka

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -44,220 +71,228 @@ export const getOnlineReleases = async (internalModel: DeviceModelInternal) => {
...
+    } else {
+        // If deviceId is provided, use the deterministic approach
+        let hash = 0;
+        for (let i = 0; i < deviceId.length; i++) {
+            const char = deviceId.charCodeAt(i);
+            hash = (hash << 5) - hash + char;
+            // Convert to 32bit integer
+            hash |= 0;
```
mroz22: lets please move it to packages/utils. a generic function that converts string to int or maybe directly to a float between 0 and 1. couple of unit tests would be nice to have

packages/connect/src/data/firmwareInfo.ts
karliatto: Removed https://github.com/trezor/trezor-suite/pull/18821/commits/e4dbafe33c4a43d37f5792e28afb7d999b480391

packages/connect/src/data/firmwareInfo.ts
karliatto: Yes, I have been thinking about it and think it is right that it is undefined when Model.UNKNOWN but I added that to the types so it is expected to be sometimes undefined https://github.com/trezor/trezor-suite/pull/18821/commits/e4dbafe33c4a43d37f5792e28afb7d999b480391

packages/connect/src/data/firmwareInfo.ts
karliatto: Sure, I am adding some comment https://github.com/trezor/trezor-suite/pull/18821/commits/e4dbafe33c4a43d37f5792e28afb7d999b480391#diff-c30ddd912ad02dc9dace968df9d44aebc836ee02b76e3da843475b5478116d18R90

But also thinking about the function again, the way it was done was very confusing for me so I was trying to make it better for understanding. Now I am looking into it in a different way.

packages/connect/src/data/firmwareInfo.ts
karliatto: You are right, but thinking about it. My goal separating into functions was to make it more clear because before there was a smart function that was returning you version but was not clear if bootloader or firmware version.

I am thinking to get back to the concept of using a smarter function that will handled all the functions. Something maybe called `getCurrentDeviceVersions` that will return an object:

```
{
  firmwareVersion: ...
  bootloaderVersion: ...
}
```
And depending of if it is run in bootloader/firmware mode it will return you the version or undefined or whatever it is, and well commented.

WDYT?

packages/connect/src/data/firmwareInfo.ts
karliatto: Fixed https://github.com/trezor/trezor-suite/pull/18821/commits/e4dbafe33c4a43d37f5792e28afb7d999b480391#diff-c30ddd912ad02dc9dace968df9d44aebc836ee02b76e3da843475b5478116d18R90

packages/connect/src/data/firmwareInfo.ts
karliatto: That makes sense, so instead of throwing error what about https://github.com/trezor/trezor-suite/pull/18821/commits/1c9c1840679e05f7d2080f5719940bded3cf95db ?

packages/connect/src/data/firmwareInfo.ts
karliatto: Thanks, https://github.com/trezor/trezor-suite/pull/18821/commits/27b8e9c77d7fbe882181f4629e975f1e152f9fb2

packages/connect/src/data/firmwareInfo.ts
karliatto: Sure. I added it here 88dc02e14418617dd8491cabd32606b2a1380fe5  but I am going to change `getCurrentFirmwareVersion` as per comment above but anyway I am adding it there already.

packages/connect/src/data/firmwareInfo.ts
karliatto: Yes, that make sense, what about e7619774145e645b0f2069441e4783e788555938 ?

And applied it here 89112a08b3672bb7b4ee083b5771eee21b0bf0ba

packages/connect/src/data/firmwareInfo.ts
MiroslavProchazka: Hey, 
we definitely wants to have the option 1. - so during onboarding they will always receive the latest deployed, even if rollout is for 1%

Btw. in that case we need some "rollback" measure, for example if I set up StagingPercentage to 0, no one even during onboarding will receive the update.

MiroslavProchazka: cc @karliatto

karliatto: Ok, that makes sense, what about https://github.com/trezor/trezor-suite/pull/18821/commits/0c9c5f0f0807db5de1d0362588c724ec5f1f47d5 ? But there is one thing I still want to test.

karliatto: I am thinking it a bit more.
When device is fresh we need to install FW for device to work. So in case rollout probability 0 in remote we want to use the firmware-release-config that is bundled into the app which should always have rollout probability 100.

packages/connect/src/data/firmwareInfo.ts
karliatto: @mroz22 what about https://github.com/trezor/trezor-suite/pull/18821/commits/dbe8cff129803fd3f92164215064c32ea3019d4d  ?

I am consider if I should consider the situation `Some old version of T1B1 do not report FW version in bootloader mode` in the typescript types and check for that in the function, or is it enough to ignore it there and just check for it in https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/data/firmwareInfo.ts#L244  since it only happens in very old versions of T1.

packages/utils/src/index.ts
mroz22, line 1: cool, now I would suggest separating this commit and review it in isolation 🙏

mroz22: also there is a similar one, might be worth checking https://github.com/trezor/trezor-suite/blob/feat/firmware-update/suite-common/message-system/src/experimentUtils.ts#L21

mroz22: I also got some ideas for improvemnets https://github.com/trezor/trezor-suite/pull/19514

karliatto: I separated it here https://github.com/trezor/trezor-suite/pull/19539 we can combine it with your improvement as well.

packages/connect-common/files/firmware/release/releases.v1.jws
mroz22, line 1: is this file expected to be version controlled?

karliatto: Yes, and I think there is no check on that. I will add it.
Thank you for your comment.

karliatto: The version in the file is hard-coded and we fetch it so no way/reason to verify it but the version itself in the file is now checked  https://github.com/trezor/trezor-suite/pull/18821/files#diff-7b2bfbaebbf12e1ccf4aa7e84effcc7e4c2102fc96afcceab3a4b9b8cf388618

I also added check for timestamp remote.

packages/connect-common/files/firmware/t1b1/bitcoinonly/t1b1-1.11.2-bitcoinonly.json
```diff
@@ -0,0 +1,12 @@
+{
+    "required": false,
+    "version": [1, 11, 2],
+    "bootloader_version": [1, 11, 0],
+    "min_bridge_version": [2, 0, 25],
+    "min_firmware_version": [1, 6, 2],
+    "min_bootloader_version": [1, 5, 0],
+    "url": "/suite/firmware/t1b1/universal/t1b1-1.11.2-bitcoinonly.bin",
```
mroz22: soo, there is `universal` subfolder and `unsinged` subfolder?

karliatto: They are different things.

In testing environments we have `unsigned` and `signed` subfolders depending on the environment.

In production this PR is like the structure is going to look like: https://github.com/trezor/data/pull/140/files. So there will only be `universal` and `bitcoinonly` inside each device model. The reason to do that is to organize better the files, since having all the new files in one directory could be confusing. But we keep the location of all the already existing files for backward compatibility.

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -275,34 +478,76 @@
...
+        : buildLocalFirmwareFileName(firmwareType, deviceModel, firmwareVersion);
+
+    const versionString = intermediaryVersion
+        ? String(intermediaryVersion)
+        : firmwareVersion.join('.');
+
+    const bundledBaseUrl = removeTrailingSlashes(DataManager.getSettings('binFilesBaseUrl'));
+    // Here we care just to know if the binaries are bundled, in order to use them locally instead of fetching them
+    // if they are in default remote we ignore it.
+    const isRealBundled = !bundledBaseUrl.includes('data.trezor.io');
```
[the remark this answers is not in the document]
karliatto: It does not apply to this use case.

mroz22 (approved):

PR #18833, opened 2025-05-10
fix(suite-native): Device setting: fixing various issues related to navigation behaviour

### LINKED ISSUE #16510: Unable to connect device after failed fw update
**Describe the bug**
If I fail the fw update and go back to dashboard and try to reconnect the device (or connect any other device) I get stuck on the screen provided on the screenshot. This device have to have the "view only" disabled.

**Info:**

-   Mobile Suite Version: 25.1.2 baf7c4f
-   OS: Android

**How to reproduce**

1. connect device
2. try to do fw update
3. disconnect the device during the update
4. you get the fail screen
5. go back to dashboard
6. connect any device that is not remembered in your view only (even the one you just disconnected)
7. it fails on the screen below

**Screenshots**

Image

BrantalikP:
## Description

- Fix back navigation on the PIN screen
- Fix back navigation behavior when going back from the firmware update screen while the device is disconnected
- Forget temporarily remembered device if leaving fw screen without connected device
- Reset firmwareRunning state on screen unmount
- Fix cancel action for T1 on pin Screen

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/16510

## Screenshots:

FW issue:

https://github.com/user-attachments/assets/e44024b5-6019-4634-8a50-7ab1929b6242

FW: temporary remembered device issue

https://github.com/user-attachments/assets/474417b4-de12-44c8-a5c7-cf626fb56b85

Pin issue

https://github.com/user-attachments/assets/e0d6a534-7b8e-45c5-9ac3-06d91f1fa640

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/unable-to-connect-after-failed-fw-update&lookback=7)

### DIFF
--- a/suite-native/device/src/hooks/usePinAction.tsx
+++ b/suite-native/device/src/hooks/usePinAction.tsx
@@ -110,15 +110,21 @@ export const usePinAction = ({ type, onSuccess }: PinActionProps) => {
             onSuccess();
         } else {
             const errorCode = payload.code;
-            if (errorCode === 'Failure_ActionCancelled' || errorCode === 'Failure_PinCancelled') {
+            if (
+                errorCode === 'Failure_ActionCancelled' ||
+                errorCode === 'Failure_PinCancelled' ||
+                errorCode === 'Method_Interrupted'
+            ) {
                 showError(canceledMessageKey, handlePinAction);
             } else if (errorCode === 'Failure_PinInvalid') {
                 showError('moduleDeviceSettings.pinProtection.errors.pinInvalid', handlePinAction);
             } else if (errorCode === 'Failure_PinMismatch') {
                 showError('moduleDeviceSettings.pinProtection.errors.pinMismatch', handlePinAction);
+            } else {
+                navigation.goBack();
             }
         }
-    }, [device, dispatch, onSuccess, showError, showSuccess, type]);
+    }, [device, dispatch, navigation, onSuccess, showError, showSuccess, type]);
 
     useEffect(() => {
         if (isDeviceConnected) handlePinAction();
--- a/suite-native/module-device-settings/src/hooks/useDeviceConnectionGuard.ts
+++ b/suite-native/module-device-settings/src/hooks/useDeviceConnectionGuard.ts
@@ -1,7 +1,7 @@
-import { useCallback, useEffect } from 'react';
+import { useCallback } from 'react';
 import { useSelector } from 'react-redux';
 
-import { useNavigation } from '@react-navigation/native';
+import { useFocusEffect, useNavigation } from '@react-navigation/native';
 
 import { selectIsDeviceConnected } from '@suite-common/wallet-core';
 import {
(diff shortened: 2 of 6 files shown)


suite-native/module-device-settings/src/hooks/useDeviceConnectionGuard.ts
```diff
@@ -36,10 +36,9 @@ export const useDeviceConnectionGuard = () => {
         });
     }, [navigation]);
 
-    useEffect(() => {
+    useFocusEffect(() => {
         if (!isDeviceConnected) return redirectToConnectAndUnlockScreen();
```
yanascz, line 40: Could we change to this?
```suggestion
        if (isDeviceConnected) {
            redirectToConnectAndUnlockScreen();
        }
```

matejkriz: This needs to be wrapped in `useCallback`, otherwise it would re-run on each re-render, right?

BrantalikP: oh, ok I did not know you must do that! Thanks! https://github.com/trezor/trezor-suite/pull/18833/commits/eefc580074247931316e18b567fcb19a7cd00a1d

suite-native/device/src/hooks/usePinAction.tsx
```diff
@@ -110,15 +110,21 @@ export const usePinAction = ({ type, onSuccess }: PinActionProps) => {
             onSuccess();
         } else {
             const errorCode = payload.code;
-            if (errorCode === 'Failure_ActionCancelled' || errorCode === 'Failure_PinCancelled') {
+            if (
+                errorCode === 'Failure_ActionCancelled' ||
+                errorCode === 'Failure_PinCancelled' ||
+                errorCode === 'Method_Interrupted'
+            ) {
                 showError(canceledMessageKey, handlePinAction);
```
matejkriz, line 118: Nitpick: canceled message says "has been canceled on your Trezor" but it doesn't have to be true (it appears if you cancel from the app too...). I would just remove the part "on your Trezor" from all the cancelled messages, but we can do it in next PR together with handling GO_BACK navigation on PIN screens.

BrantalikP: Yeah, this also could actually be an enhancement from @trezor/suite-connect  to be able to differentiate between a cancel action from the app and from the Trezor. 🤔?

martykan: `Method_Interrupted` or `Method_Cancel` are the ones coming from the app

matejkriz: I believe we can handle it but I don't believe it's worth it here...

BrantalikP: When I tried to debug it it showed the same message, regardless the origin.
So maybe the `pin cancel ` action is throwing a wrong msg?
 btw: https://github.com/trezor/trezor-suite/pull/18833/commits/ceec45ed9bc113ea1be1b0d2293700142423a2fe

matejkriz (approved):
I wasn't able to break it 😮 👏 Next time it would be great to have those commits separated in multiple PRs so we can merge them sooner 🙏

suite-native/device/src/hooks/usePinAction.tsx
BrantalikP: only T1 trhows `Method_Interrupted` thats why we needed the "fix"

STew790:
**QA OK**
Issues have been fixed and I also wasn't able to fix it. The linked issue was commented separately.

**Info**
25.6.2 02d301e

Issue #9925, opened 2023-11-13
Navigating from settings to welcome page not reflected in the URL


komret:
**Describe the bug**
The route change is not propagated to the URL. When user continues to Suite, they are navigated to settings instead of dashboard.

**How to reproduce**

1. Reset the app in settings.
2. Go to settings.
3. Close settings.
4. Continue to Suite.

**Expected behavior**
URL changes back to `/`.  See dashboard after accessing Suite.

**Screenshots**

https://github.com/trezor/trezor-suite/assets/42465546/5cb878ab-0c7c-4acf-87e7-8b4cb85b0fcd

tomasklim:
I guess bug like this was fixed here https://github.com/trezor/trezor-suite/pull/5899/files#diff-d4bb5d665d9ff799cf8338596feec3b629110ff2d4ac3e865e2c6f2b512d9ee7 so probably reintroduced

vojtatranta:
Question: Can anyone think of a case, where such behavior could make sense? That eg. even if we leave the settings, we want them be kept in the background to be used? My ideas:
- some overlays being displayed based on the user states
- some onboardings etc

If it was reintroduced, maybe there is a reason of why this behavior can be valid in some cases, and this is just some unintended consequence.

vojtatranta:
These are the typical foreground apps, for these, the url is locked to be the same. (eg. going from `suite-start` to `settings-index` causes the url to change to `/settings` but going back from settings to `suite-start` (welcome screen) will keep the `settings` in place.

My idea: The likely reason, of why this could be is when a foreground app was "started" from a particular URL and you could navigate within the foreground app. Then, after the foreground app interaction is done, you wanted to "stay" on the original url that triggered it. BUt llike, you'd only have this url in the window, so, why would you want that? 

The solution for this case should have been like `foregroundAppInitialUrl` and not this. So I am not sure.

```
  {
        name: 'suite-start',
        pattern: '/start',
        app: 'start',
        isFullscreenApp: true,
        isForegroundApp: true,
    },
    {
        name: 'suite-index',
        pattern: '/',
        app: 'dashboard',
        exact: true,
    },
    {
        name: 'suite-version',
        pattern: '/version',
        app: 'version',
        isForegroundApp: true,
        params: modalAppParams,
    },
    {
        name: 'suite-bridge-requested',
        pattern: '/bridge-requested',
        app: 'bridge-requested',
        isForegroundApp: true,
        params: modalAppParams,
    },
    {
[...]

PR #18872, opened 2025-05-12
fix: clear the URL when going back to the suite-start route to preven…

### LINKED ISSUE #13417: Fix SUite returning to settings instead of Dashboard

vojtatranta:
…t stale-state in the url

## Description

The problem is taht `suite-start` is a "foreground app" route. Meaning, the URL / history isn't change when navigating to it. It makes sense, coz we want it to have `/` url but internally, it should have some distinguished url, especially for the web version.

But the problem is that on this `suite-start` page, `settings` are availabe as route. And these require the url to be change to properly working.

So likely, only possible solution would be to have a parameter for this particular case that would say that "hey, for this route I want to reset the URL".

I solved by a static param on the `suite-start` route.

Better ideas are appreaciated.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/13417

Also closes the issue with comment here: https://github.com/trezor/trezor-suite/issues/9925#issuecomment-2872267114

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=13417-prevent-going-back-settings-on-welcome&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=13417-prevent-going-back-settings-on-welcome&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=13417-prevent-going-back-settings-on-welcome&lookback=7)

### DIFF
--- a/packages/suite/src/actions/suite/routerActions.ts
+++ b/packages/suite/src/actions/suite/routerActions.ts
@@ -137,6 +137,13 @@ export const goto =
         if (route?.isForegroundApp) {
             dispatch(suiteActions.lockRouter(true));
 
+            // NOTE: this is useful eg. on welcome screen / logged out screen
+            // where we want to have suite-start router clearing the URL to ensure
+            // that there isn't a state stuck
+            if (route?.clearUrl) {
+                history.push(urlBase);
+            }
+
             return;
         }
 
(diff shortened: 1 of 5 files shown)


peter-sanderson (approved):
One day, I will nuke the foreground app pattern. One day...

vojtatranta:
> One day, I will nuke the foreground app pattern. One day...

hm, it should be just renamed to "modal" imho :D and it shoudn't be imperative in my mind

PR #18896, opened 2025-05-13
ci: include tags from private in FW releases check

Lemonexe:

### DIFF
--- a/packages/connect-common/scripts/check-firmware-revisions.sh
+++ b/packages/connect-common/scripts/check-firmware-revisions.sh
@@ -36,17 +36,21 @@ git fetch origin
 git checkout "$BRANCH"
 git reset "origin/$BRANCH" --hard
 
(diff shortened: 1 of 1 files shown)


packages/connect-common/scripts/check-firmware-revisions.sh
```diff
@@ -36,17 +36,21 @@ git fetch origin
 git checkout "$BRANCH"
 git reset "origin/$BRANCH" --hard
 
+# add private repository if it does not exist, or suppress error output (idempotent)
+git remote add private git@github.com:trezor/trezor-firmware-private.git >/dev/null 2>&1 || true
+git fetch --tags private
```
Lemonexe, line 41: :bulb:  it is enough to fetch tags, no need to be checked out on the branch.
Because `git tag --points-at` reads all locally present tags, not just those that point to commits on current branch.

PR #18819, opened 2025-05-09
Bump foundation deps 2025.05

Lemonexe:
## Description

Update most Foundation-related dependencies.

**major version:**
- `electron` (stack changes)
- `globals` (only for eslint)
- `minimatch` (node 20 requirement)
  
**minor version:**
- `bignumber.js` (TS-only changes)
- `@eslint/js`
- `eslint`
- `eslint`
- `typescript-eslint`

**patch version:**
- `electron-builder`
- `electron-updater`
- `idb`
- `react-hook-form`
- `bn.js`

**not updated:**
- `tiny-secp256k1` TODO in [#12261](https://github.com/trezor/trezor-suite/pull/12261)
- `electron-store + chalk` TODO in [#14482](https://github.com/trezor/trezor-suite/issues/14482)
- `react-router + react-router-dom + history` TODO in [#4649](https://github.com/trezor/trezor-suite/issues/4649)
- `nx` TODO in [#18812](https://github.com/trezor/trezor-suite/issues/18812)

:information_source: For reference, last bump foundation deps PR was #18087 + #18682 + #18107 + #18704 + #18680
:eye: I skimmed through all code changes _except_ eslint-related packages and `electron` ; found nothing suspicious ✅ 

## Related Issue

Resolve #18774

## Dev QA
:eye: Besides CI checks, I have tested locally:
- suide dev web
  - app builds & runs
  - resetting app storage works (`idb`)
- suite dev desktop
  - app builds & runs
  - tor works
  - resetting app storage works (`idb`)
  - send form & sign/verify form works
- suite native dev on android emu
  - trading form works
- `yarn refs` && `yarn verify-project-references` work correctly (`minimatch`)
- `yarn generate-package` works correctly (`sort-package-json`)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=bump-foundation-deps-2025.05&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=bump-foundation-deps-2025.05&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=bump-foundation-deps-2025.05&lookback=7)

Issue #9007, opened 2023-07-25
UI: Components - Tabs


Hermez-cz:
**Design:** [Figma](https://www.figma.com/file/Y2E8dcdNYrSMGIVRjYFXKp/%5Bs2.0%5D-Components?type=design&node-id=1810-56915&mode=design&t=OiKb67lgDy7MH3qz-4)

**Spec:** 
* overline tabs to be used in global send/receive and notifiations.

dahaca:
Is postponed till later – not needed for the first iteration. 

TODO:

- [ ] Move currently used tabs to `@trezor/components`
- [ ] Add modal tabs – a modal with built in tabs and no heading, like in global send/receive. Or maybe an entirely new comp that can be rendered as a modal or a dropdown?

Hermez-cz:
Needs to wait for [Global Send/Receive](https://www.notion.so/satoshilabs/Global-Send-Receive-Trade-1cec5b7b56ce4f38bf12be4595afa535?pvs=4), lowering prio.

PR #18906, opened 2025-05-13
fix(suite-desktop): macOS ARM build adhoc signing

Lemonexe:
## Description

Followup to #18819, which broke CI dev desktop build for macOS :apple: 

With [electron-builder 26.0.13](https://github.com/electron-userland/electron-builder/releases/tag/v26.0.13) a change was done in [#9007](https://github.com/electron-userland/electron-builder/pull/9007):
- if `identity` ENV is missing (as in dev builds) ARM builds are now adhoc signed by default
  - [docs here](https://github.com/electron-userland/electron-builder/blob/master/pages/code-signing-mac.md)
- [that crashes in CI](https://github.com/trezor/trezor-suite/actions/runs/14991949805/job/42117150993)
- so explicitly skip signing when not codesign build 

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/macos-arm-adhoc-signing&lookback=7)

### DIFF
--- a/packages/suite-desktop/electron-builder-config.js
+++ b/packages/suite-desktop/electron-builder-config.js
@@ -81,6 +81,9 @@ module.exports = {
         oneClick: false,
     },
     mac: {
+        // identity: undefined means default behavior (autodetect the ENV variables necessary for signing and perform it)
+        // identity: null means disable any signing – necessary for dev builds (default behavior is to adhoc sign ARM builds, which would fail in CI)
+        identity: isCodesignBuild ? undefined : null,
         files: ['entitlements.mac.inherit.plist'],
         extraResources: [
             {


vojtatranta:
I think there is a problem:
- for some reason, there isn't an icon of the Trezor Suite app, but it was in the previous version of the electron-builder
- now, x64 build WON'T be started on Mac but it WAS possible to be started on the previous version of the electron builder, but it was likely because I made a build with `identity: '-'`  which triggered ad-hoc signing

vojtatranta:
And local build x64 won't start at all  with error: https://pastebin.com/u6GgqPSU

PR #18900, opened 2025-05-13
chore(suite-native): remove device onboarding feature flag

PeKne:
## Description
- removes Feature flag for device onboarding
- enables message system to toggle device onboarding flow on mobile

## Related Issue

Resolve #16295 

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/native/remove-device-onboarding-ff&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/native/remove-device-onboarding-ff&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/remove-device-onboarding-ff&lookback=7)

PR #18915, opened 2025-05-13
feat(suite-native): Mobile Trade: Allow buy by default

jbazant:
## Description

Allow mobile trade buy. 🎉 

## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=mobile-trade-allow-buy-ff&lookback=7)

### DIFF
--- a/suite-native/app/src/navigation/__tests__/AppTabNavigator.comp.test.tsx
+++ b/suite-native/app/src/navigation/__tests__/AppTabNavigator.comp.test.tsx
@@ -1,5 +1,10 @@
 import { FeatureFlag, featureFlagsInitialState } from '@suite-native/feature-flags';
-import { PreloadedState, fireEvent, renderWithStoreProviderAsync } from '@suite-native/test-utils';
+import {
+    PreloadedState,
+    act,
+    fireEvent,
+    renderWithStoreProviderAsync,
+} from '@suite-native/test-utils';
 
 import { AppTabNavigator } from '../AppTabNavigator';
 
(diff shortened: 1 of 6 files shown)


suite-native/feature-flags/src/featureFlagsSlice.ts
```diff
@@ -38,7 +38,8 @@ export const featureFlagsInitialState: FeatureFlagsState = {
         process.env.EXPO_PUBLIC_FF_IS_MODEL_T_DEVICE_ONBOARDING_ENABLED === 'true',
     [FeatureFlag.IsWalletConnectEnabled]:
         process.env.EXPO_PUBLIC_FF_IS_WALLET_CONNECT_ENABLED === 'true',
-    [FeatureFlag.IsTradingBuyEnabled]: process.env.EXPO_PUBLIC_FF_IS_TRADING_BUY_ENABLED === 'true',
+    [FeatureFlag.IsTradingBuyEnabled]:
+        process.env.EXPO_PUBLIC_FF_IS_TRADING_BUY_ENABLED !== 'false',
```
PeKne, line 42: This do not seems to be correct. Now if you would set `EXPO_PUBLIC_FF_IS_TRADING_BUY_ENABLED=true`, the feature will be turned off, because 'true' is not equal to 'false'.

PeKne:
What about adding the buy feature to the message system config so it can be turned of in case that there is some critical bug found? We decided to do so for device onboarding (https://github.com/trezor/trezor-suite/pull/18900) since it is a quite a big feature as well. Just my two sats.

vytick (changes requested):
I think we should just change 
`selectIsFeatureEnabled(state, Feature.trading.buy, false);`
to
`selectIsFeatureEnabled(state, Feature.trading.buy, true);`

and leave local feature flag as is to keep possibility to enable it locally in case we disable it remotely. Sorrz for not being clear when we discussed this on slack. 🙏

vytick (approved):
works nicely 👌 
Nice improvement on tests 👍

vytick:
@trezor/qa Trade Buy should now be enabled by default for mobile app

Ondra-Zik:
**QA OK**

Trade is now visible without the flag.

Build: c274554

PR #18976, opened 2025-05-15
fix(suite-native): better handling of cancellation of device auth check on the device

### LINKED ISSUE #18980: Onboarding - reconnecting when the authenticity check is going on results in immediate authenticity check
**Describe the bug**
If I disconnect the device when I am prompted to do the DAC in the onboarding, I can see the authenticity check on the device right when I connect it to the phone instead of when being prompted to do in the onboarding.

**Info:**

- Mobile Suite Version: 25.5.1 https://github.com/trezor/trezor-suite/commit/a25f124ef0cf72da97038d0fcd3d697e5484a485
- OS: name, version

**How to reproduce**

1. Connect device with FW but no seed
2. Start the onboarding
3. Click on "Lets get started"
4. Disconnect the device when being asked to do auth check
5. Connect it again
6. On device I can see the device auth check but on phone there is the "Continue with onboarding" button

**Expected behavior**
I am prompted to do the DAC after reconnecting.

matejkriz:
## Description

The previous implementation restarted the device auth check immediately after cancellation, which resulted in a strange UX (the user cancelled the action, and it was immediately initiated again).
Now, it is only triggered after the user clicks the “Continue setup” button. See the video.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18980, https://github.com/trezor/trezor-suite/issues/18984

## Screenshots:

https://github.com/user-attachments/assets/ede2947e-7955-40ba-8c87-475c43465e39

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/device-auth-check-in-onboarding-cancelation&lookback=7)

### DIFF
--- a/suite-native/module-device-onboarding/src/components/DeviceOnboardingScreenWithExitButton.tsx
+++ b/suite-native/module-device-onboarding/src/components/DeviceOnboardingScreenWithExitButton.tsx
@@ -21,9 +21,15 @@ type NavigationProps = StackToStackCompositeNavigationProps<
     RootStackParamList
 >;
 
-const DeviceOnboardingExitButtonScreenHeader = () => {
+type DeviceOnboardingExitButtonScreenHeaderProps = {
+    onAlertContinueButtonPress?: () => void;
+};
+
+const DeviceOnboardingExitButtonScreenHeader = ({
+    onAlertContinueButtonPress,
+}: DeviceOnboardingExitButtonScreenHeaderProps) => {
     const navigation = useNavigation<NavigationProps>();
-    const { handleExitButtonPress } = useExitAlert();
+    const { handleExitButtonPress } = useExitAlert(onAlertContinueButtonPress);
 
     useEffect(() => {
         // Override default navigation GO_BACK action to align it with the exit button behavior.
@@ -40,8 +46,19 @@ const DeviceOnboardingExitButtonScreenHeader = () => {
     return <ScreenHeader closeActionType="close" closeAction={handleExitButtonPress} />;
 };
 
-export const DeviceOnboardingScreenWithExitButton = ({ children, ...screenProps }: ScreenProps) => (
-    <Screen header={<DeviceOnboardingExitButtonScreenHeader />} {...screenProps}>
+export const DeviceOnboardingScreenWithExitButton = ({
+    children,
+    onAlertContinueButtonPress,
+    ...screenProps
+}: ScreenProps & DeviceOnboardingExitButtonScreenHeaderProps) => (
+    <Screen
+        header={
+            <DeviceOnboardingExitButtonScreenHeader
+                onAlertContinueButtonPress={onAlertContinueButtonPress}
+            />
+        }
+        {...screenProps}
+    >
         <Box flex={1} marginTop="sp16">
             {children}
         </Box>
--- a/suite-native/module-device-onboarding/src/hooks/useExitAlert.ts
+++ b/suite-native/module-device-onboarding/src/hooks/useExitAlert.ts
@@ -24,7 +24,7 @@ type NavigationProps = StackToStackCompositeNavigationProps<
     DeviceOnboardingStackRoutes,
     RootStackParamList
 >;
-export const useExitAlert = () => {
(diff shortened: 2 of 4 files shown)


suite-native/module-device-onboarding/src/hooks/useExitAlert.ts
```diff
@@ -24,7 +24,7 @@ type NavigationProps = StackToStackCompositeNavigationProps<
     DeviceOnboardingStackRoutes,
     RootStackParamList
 >;
-export const useExitAlert = () => {
+export const useExitAlert = (handleContinueButtonPress?: () => void) => {
```
PeKne, line 27: nit:  I would prefer `on` prefix since the whole repo follows this convention. (https://medium.com/@master_43681/understanding-and-proper-usage-of-on-and-handle-prefixes-in-react-function-naming-119f0beea400)
```suggestion
export const useExitAlert = (onContinueButtonPress?: () => void) => {
```

suite-native/module-device-onboarding/src/components/DeviceOnboardingScreenWithExitButton.tsx
```diff
@@ -21,9 +21,15 @@ type NavigationProps = StackToStackCompositeNavigationProps<
     RootStackParamList
 >;
 
-const DeviceOnboardingExitButtonScreenHeader = () => {
+type DeviceOnboardingExitButtonScreenHeaderProps = {
+    onContinueButtonPress?: () => void;
```
PeKne: nit:
```suggestion
    onAlertContinueButtonPress?: () => void;
```
since the component is not containing only the `Alert`, more descriptive naming seems useful to me.

suite-native/module-device-onboarding/src/screens/DeviceAuthenticityScreen.tsx
```diff
@@ -14,37 +14,36 @@ import { DeviceOnboardingScreenWithExitButton } from '../components/DeviceOnboar
...
-            setIsCheckActive(false);
-        }
-    }, [hasSoftError]);
-
-    useEffect(() => {
-        if (!isCheckActive) {
-            setIsCheckActive(true);
+    // Called automatically on first focus
+    const startCheckDeviceAuthenticityOnFocus = useCallback(() => {
+        if (!hasCheckedOnFocusRef.current) {
```
PeKne: What about using useEffect with `[]` empty dependencies array instead of `useFocusEffect`? Then this function will be triggered only once during the initial render and the `ref` won't be needed. https://infinitejs.com/posts/mastering-useeffect-avoid-dependencies

PeKne (commented):
have one question and two naming nits (feel free to resolve those)

suite-native/module-device-onboarding/src/hooks/useExitAlert.ts
matejkriz: But inside useExitAlert, it is a handler, it's not a named property of the hook, so I believe it follows the convention you've linked.

suite-native/module-device-onboarding/src/screens/DeviceAuthenticityScreen.tsx
matejkriz: It looks like I'm overcomplicating this, I'll think about it twice.

suite-native/module-device-onboarding/src/components/DeviceOnboardingScreenWithExitButton.tsx
matejkriz: 27c740d902bfe3dc997f1e3b08838144de2b32db

suite-native/module-device-onboarding/src/screens/DeviceAuthenticityScreen.tsx
matejkriz: rebased because I rewrite the original commit almost completely, please check the current state after rebase 🙏

PeKne (approved):

PR #18996, opened 2025-05-16
fix(blockchain-link): do not set `User-Agent` for the Stellar SDK in web environment.

### LINKED ISSUE #18862: XLM: User-agent error on web, cannot be set
**Describe the bug**
We cannot change user agent in browsers. That's known. However, it should not put error into console.

Image

overcat:
## Description

When the current environment is detected as web, do not set User-Agent.

## Related Issue

Resolve #18862

## Screenshots:

N/A

### DIFF
--- a/packages/blockchain-link/src/workers/stellar/index.ts
+++ b/packages/blockchain-link/src/workers/stellar/index.ts
@@ -5,7 +5,7 @@ import { MESSAGES, RESPONSES } from '@trezor/blockchain-link-types/src/constants
 import { CustomError } from '@trezor/blockchain-link-types/src/constants/errors';
 import type * as MessageTypes from '@trezor/blockchain-link-types/src/messages';
 import * as utils from '@trezor/blockchain-link-utils/src/stellar';
-import { getSuiteVersion } from '@trezor/env-utils';
+import { getSuiteVersion, isDesktop, isNative } from '@trezor/env-utils';
 import { BigNumber } from '@trezor/utils/src/bigNumber';
 
 import { BaseWorker, CONTEXT, ContextType } from '../baseWorker';
@@ -291,7 +291,9 @@ class StellarWorker extends BaseWorker<Horizon.Server> {
     async tryConnect(url: string): Promise<Horizon.Server> {
         const api = new Horizon.Server(url, {
             headers: {
-                'User-Agent': `Trezor Suite ${getSuiteVersion()}`,
+                ...(isDesktop() || isNative()
+                    ? { 'User-Agent': `Trezor Suite ${getSuiteVersion()}` }
+                    : {}),
             },
         });
 


packages/blockchain-link/src/workers/stellar/index.ts
```diff
@@ -291,7 +291,9 @@ class StellarWorker extends BaseWorker<Horizon.Server> {
     async tryConnect(url: string): Promise<Horizon.Server> {
         const api = new Horizon.Server(url, {
             headers: {
-                'User-Agent': `Trezor Suite ${getSuiteVersion()}`,
+                ...(isDesktop() || isNative()
```
overcat, line 294: It's strange that `isDesktop()` returns `true` on the desktop, `isNative()` returns `true` on Android, but `isWeb()` returns `false` here, in the Chrome browser.

tomasklim: How come other libs (Solana, ripple,...) do not have this console error issue?

overcat: I speculate this is related to the different HTTP libraries they use; the Solana SDK uses [fetch](https://github.com/anza-xyz/kit/blob/66808064e8251c452a4c339791b2985bf488e514/packages/rpc-transport-http/src/http-transport.ts#L69), while the Stellar SDK uses axios.
image

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
    <script src="https://cdn.jsdelivr.net/npm/axios/dist/axios.min.js"></script>

    <script>
        console.log("with fetch")
        fetch('https://horizon.stellar.org/fee_stats', {
            method: 'GET',
            headers: {
                'Content-Type': 'application/json',
                'User-Agent': 'MyApp/1.0.0',
            }
        }).then(response => {
            if (!response.ok) {
                throw new Error('Network response was not ok ' + response.statusText);
            }
            return response.json();
        }).then(data => {
            console.log('Data fetched:', data);
        }).catch(error => {
            console.error('Error fetching data:', error);
        });

        console.log("with axios")
        axios.get('https://horizon.stellar.org/fee_stats', {
            headers: {
                'Content-Type': 'application/json',
                'User-Agent': 'MyApp/1.0.0',
            }
        }).then(response => {
            console.log('Data fetched with axios:', response.data);
        }).catch(error => {
            console.error('Error fetching data with axios:', error);
        });
    </script>
</head>
<body>
    Check the console for output.
</body>
</html>
```

overcat:
https://github.com/axios/axios/issues/1231

packages/blockchain-link/src/workers/stellar/index.ts
tomasklim: @overcat can we use @stellar/sdk without axios? https://www.npmjs.com/package/@stellar/stellar-sdk `Build without Axios`

overcat: The SDK documentation introduces [a method](https://github.com/stellar/js-stellar-sdk?tab=readme-ov-file#usage-with-cloudflare-workers), but I find it quite strange, and it requires the introduction of an old version of Axios. 

As for building directly `without Axios`, I suspect this is not feasible; without Axios, it cannot make HTTP requests. I'll test this tomorrow.

overcat: Hi @tomasklim, After reading [the source code](https://github.com/stellar/js-stellar-sdk/blob/master/src/http-client/index.ts#L14-L24), I found that if axios is not used, it will use fetch, so I made the corresponding modifications.

overcat: Please hold on, the SDK's fetch implementation seems to have a bug :-(

https://github.com/stellar/js-stellar-sdk/issues/1175

packages/blockchain-link/src/workers/stellar/index.ts
```diff
@@ -1,4 +1,6 @@
-import { Horizon, Networks, Transaction as StellarTransaction } from '@stellar/stellar-sdk';
+import { Networks, Transaction as StellarTransaction } from '@stellar/stellar-sdk';
+// See https://github.com/trezor/trezor-suite/pull/18996#discussion_r2107284968
+import { Horizon } from '@stellar/stellar-sdk/no-axios';
```
tomasklim: Great job, would be nice to use this everywhere you import horizon, even in tests

tomasklim:
@overcat Hello, please use the no-axios Horizon everywhere and then squash it please

overcat:
Hi @tomasklim, as I mentioned earlier, there is a bug in the Stellar SDK that prevents us from using no-axios. I have submitted [a fix](https://github.com/stellar/js-stellar-sdk/pull/1176), and I expect a new version of the Stellar SDK will be released soon.

tomasklim:
@overcat aha, we have freeze tomorrow, so maybe can we go with `...(isDesktop() || isNative()` and do it correctly in the next release? 🙏

overcat:
Hi @tomasklim, I rolled it back; let's improve it in the next release.

tomasklim (approved):

PR #19002, opened 2025-05-16
test(firmware-release-config): change test file name

karliatto:
## Description
Change file name from `messageReleases` to `firmwareReleaseConfig` since the name of the package was changed.


submodules/trezor-common
mroz22, line 1: this change isn't intended I guess?

karliatto: Yes, that change was not intended, thanks. I removed and rebased it.

mroz22 (approved):

PR #19059, opened 2025-05-21
fix(suite-native): reanimated shadow warning

PeKne:
## Description
- fixes reanimated warning complaining about adding style props conditionally to the `AnimatedCard`  component that is evaluated in the worklet

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/reanimated-shadow-warning&lookback=7)

### DIFF
--- a/suite-native/atoms/src/Card/Card.tsx
+++ b/suite-native/atoms/src/Card/Card.tsx
@@ -35,7 +35,7 @@ const cardInnerContainerStyle = prepareNativeStyle<{
 }>((utils, { alertPosition, noPadding, borderColor, noShadow }) => ({
     backgroundColor: utils.colors.backgroundSurfaceElevation1,
     borderRadius: utils.borders.radii.r16,
-    padding: noPadding ? 0 : utils.spacings.sp16,
+    padding: utils.spacings.sp16,
 
     extend: [
         {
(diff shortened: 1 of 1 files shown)


suite-native/atoms/src/Card/Card.tsx
```diff
@@ -36,6 +36,7 @@ const cardInnerContainerStyle = prepareNativeStyle<{
     backgroundColor: utils.colors.backgroundSurfaceElevation1,
     borderRadius: utils.borders.radii.r16,
     padding: noPadding ? 0 : utils.spacings.sp16,
+    ...(noShadow ? utils.boxShadows.none : utils.boxShadows.small),
```
juriczech:
```suggestion
        {
            condition: !noShadow,
            style: { ...utils.boxShadows.small },
        },
```

Alternatively, you can leave it within the extend array with condition as it was and spread it there. It should work. But it's just a nit to keep it consistent, either way is fine

juriczech (approved):

suite-native/atoms/src/Card/Card.tsx
PeKne: True, I will put it back to the `extend` array. I will also move the `noPadding` condition there as well to make it consistent. FORCE PUSHED

PR #19073, opened 2025-05-21
Feat: protocol-thp encode/decode

szymonlesisz:
## Description

add THP encoding/decoding logic

- only one message exchange is supported `ThpCreateChannelRequest` > `ThpCreateChannelResponse`
- encoding/decoding `ThpAck` and `ThpError`
- crc32 module: create crc32 for encoded message + validate crc32 for decoded message
- cherry-picked without cryptography module. cryptography will be added in the next commit (see TODO: add-crypto)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/protocol-thp-encode&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/protocol-thp-encode&lookback=7)


packages/protocol/tests/protocol-thp/crc32.test.ts
```diff
@@ -0,0 +1,127 @@
...
+});
+
+/* eslint-disable jest/no-commented-out-tests */
+
+// test('another simple one', function (t) {
+//     var input = Buffer.from('IEND');
+//     var expected = Buffer.from([0xae, 0x42, 0x60, 0x82]);
+//     t.same(crc32(input), expected);
+//     t.end();
+// });
```
mroz22: these commented out tests caught my attention. do we need them?

packages/protocol/src/protocol-thp/ThpState.ts
```diff
@@ -15,9 +29,133 @@ export class ThpState {
...
+        }
+    }
+
+    updateNonce(type: 'send' | 'recv', nonce?: number) {
+        const calc = (curr: number) => {
+            if (typeof nonce === 'number') {
+                return nonce;
+            }
+
+            return curr > 0 ? 0 : 1;
```
marekrjpolak: I'm confused. Nonce can be any number (unlike `syncBit`) but still flips between 0 and 1 by default?

packages/protocol/src/protocol-thp/ThpState.ts
```diff
@@ -15,9 +29,133 @@ export class ThpState {
...
+                throw new Error('ThpState.deserialize invalid state');
+            }
+        });
+
+        this.setChannel(Buffer.from(json.channel, 'hex'));
+        this.setExpectedResponses(json.expectedResponses);
+        this.updateSyncBit('send', json.sendBit);
+        this.updateSyncBit('recv', json.recvBit);
+        this.updateNonce('send', json.sendNonce);
+        this.updateNonce('recv', json.recvNonce);
```
marekrjpolak: If these updates with explicitly passed values are used only from inside `ThpState` (which I'm not sure about, but it's the current state), wouldn't it be cleaner to assign them directly to corresponding fields and remove the optional params from update methods, so they can't be set from outside?

packages/protocol/src/protocol-thp/encode.ts
```diff
@@ -0,0 +1,163 @@
...
+) => {
+    if (!thpState) {
+        throw new Error('ThpState missing');
+    }
+
+    const length = Buffer.alloc(2);
+    length.writeUInt16BE(1 + 2 + data.length + TAG_LENGTH + CRC_LENGTH); // 1 session_id + 2 messageType + protobuf len + 16 tag + 4 crc
+
+    const magic = addSequenceBit(
+        THP_CONTROL_BYTE_ENCRYPTED || THP_CONTROL_BYTE_DECRYPTED,
```
marekrjpolak: What does this mean? It will always be just the first one as it's not bitwise OR, won't it?

packages/protocol/src/protocol-thp/decode.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    if (magic === THP_CREATE_CHANNEL_REQUEST) {
+        return [THP_CREATE_CHANNEL_RESPONSE];
+    }
+
+    return [];
+};
+
+// check if
+export const isExpectedResponse = (bytes: Buffer, state: ThpState) => {
```
marekrjpolak: Returns boolean, buffer or undefined.

packages/protocol/tests/protocol-thp/crc32.test.ts
szymonlesisz: these are the fixtures from the [upstream repo](https://github.com/brianloveswords/buffer-crc32/blob/master/tests/crc.test.js) not all were useful for us as we dont have all the features (crc32.signed, crc32.unsigned, crc32 append mode)

packages/protocol/src/protocol-thp/encode.ts
szymonlesisz: `THP_CONTROL_BYTE_DECRYPTED` is never really used. it was meant to be sent in bootloader but never gets there.

removed from here

packages/protocol/src/protocol-thp/decode.ts
szymonlesisz: moved to `utils.ts` and fixed

packages/protocol/src/protocol-thp/ThpState.ts
szymonlesisz: totally true, fixed

packages/protocol/src/protocol-thp/ThpState.ts
szymonlesisz: thanks for noticing, fixed as discussed IRL

marekrjpolak (approved):
LGTM

szymonlesisz:
Lastminute change as promised, 
i've rebased `bt-thp` branch over this and found some inconsistency in naming and missing utility

PR #19081, opened 2025-05-21
fix(suite-native): fix styles of Card with top Alert

### LINKED ISSUE #19068: Confirmation notification after confirming address has weird alignment
**Describe the bug**
The green confirmation notification in address verification is weirdly aligned with the background.

**Info:**

- Mobile Suite Version: 25.5.1

**How to reproduce**

1. Connect device
2. Navigate to any address
3. Confirm it on device
4. Observe the alignment issue

**Screenshots**

Image

matejkriz:
## Description

Fix the styles of Card if there is an Alert in top position. 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19068

## Screenshots:
image

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/card-styles-with-alert&lookback=7)

### DIFF
--- a/suite-native/atoms/src/Card/Card.tsx
+++ b/suite-native/atoms/src/Card/Card.tsx
@@ -9,6 +9,9 @@ import { Color } from '@trezor/theme';
 
 import { InlineAlertBox, InlineAlertBoxProps } from '../InlineAlertBox/InlineAlertBox';
 
+const CARD_CONTAINER_TEST_ID = '@atom/card/container';
+const ALERT_TEST_ID = '@atom/card/alert/';
+
 type AlertPosition = 'top' | 'bottom';
 
 export type CardProps = {
@@ -103,7 +106,7 @@ export const Card = React.forwardRef<View, CardProps>(
             children,
             style,
             alertProps,
-            alertPosition,
+            alertPosition: alertPositionProp,
             borderColor,
             noPadding = false,
             noShadow = false,
@@ -112,10 +115,18 @@ export const Card = React.forwardRef<View, CardProps>(
     ) => {
         const { applyStyle } = useNativeStyles();
 
+        const isAlertDisplayed = !!alertProps;
+        const alertPosition = isAlertDisplayed ? (alertPositionProp ?? 'top') : undefined;
+
         return (
             <View style={applyStyle(cardOuterContainerStyle, { flex: style?.flex })}>
-                {!!alertProps && alertPosition !== 'bottom' && (
-                    <View style={applyStyle(alertBoxWrapperStyle, { alertPosition })}>
+                {isAlertDisplayed && alertPosition === 'top' && (
+                    <View
+                        style={applyStyle(alertBoxWrapperStyle, {
+                            alertPosition,
+                        })}
+                        testID={ALERT_TEST_ID + 'top'}
+                    >
                         <InlineAlertBox {...alertProps} />
                     </View>
                 )}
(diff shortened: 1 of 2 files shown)


suite-native/atoms/src/Card/Card.tsx
```diff
@@ -112,10 +115,18 @@ export const Card = React.forwardRef<View, CardProps>(
     ) => {
         const { applyStyle } = useNativeStyles();
 
+        const isAlertDisplayed = !!alertProps;
```
PeKne, line 118: It would make sense to use `RequireAllOrNone` type helper from `type-fest` for combination  of `alertPosition` and `alertProps`. Then the TypeScript will not allow using `alertPosition` without alertProps defined. We have this helper already used in our codebase for inspiration.

suite-native/atoms/src/Card/Card.tsx
```diff
@@ -112,10 +115,18 @@ export const Card = React.forwardRef<View, CardProps>(
...
+        const alertPosition = isAlertDisplayed ? (alertPositionProp ?? 'top') : undefined;
+
         return (
             <View style={applyStyle(cardOuterContainerStyle, { flex: style?.flex })}>
-                {!!alertProps && alertPosition !== 'bottom' && (
-                    <View style={applyStyle(alertBoxWrapperStyle, { alertPosition })}>
+                {isAlertDisplayed && alertPosition === 'top' && (
```
PeKne, line 123: Then this could be simplified as 
```suggestion
                {alertPosition === 'top' && (
```

suite-native/atoms/src/Card/Card.tsx
```diff
@@ -130,13 +141,19 @@ export const Card = React.forwardRef<View, CardProps>(
...
                     ]}
+                    testID={CARD_CONTAINER_TEST_ID}
                     // Ref must be here otherwise the animation will not work
                     ref={ref}
                 >
                     {children}
                 </View>
-                {!!alertProps && alertPosition === 'bottom' && (
-                    <View style={applyStyle(alertBoxWrapperStyle, { alertPosition })}>
+                {isAlertDisplayed && alertPosition === 'bottom' && (
```
PeKne, line 150: as well this

suite-native/atoms/src/Card/__tests__/Card.comp.test.tsx
```diff
@@ -20,10 +20,61 @@ describe('Card', () => {
         expect(getByText('hello')).toBeTruthy();
     });
 
-    it('should render children and alert, when alert props are specified', () => {
-        const { getByText } = renderComponent({ alertProps: { title: 'alert', variant: 'info' } });
+    it('should render only children if alertProps are not provided, even if alertPosition is set', () => {
```
PeKne, line 23: and this test would not be needed.

PeKne (commented):
I have small suggestion to improve the props handling. Otherwise it looks good.

suite-native/atoms/src/Card/Card.tsx
matejkriz: I thought that we currently have only one use of bottom alert and all the others are on top, so I wanted to keep the top as the default. But now I'm thinking more about removing the alert from inside the Card component completely or prepare "CardWithTopAlert" and "CardWithBottomAlert" 🤔

PeKne (approved):

suite-native/atoms/src/Card/Card.tsx
PeKne: All right, It makes sense to me to prepare specialized components as `CardWithTopAlert` and do not overcomplicted generic `Card`. So I am approving this and we can do refactor later after this is cherry-picked. 👍

matejkriz: followup issue here https://github.com/trezor/trezor-suite/issues/19123

PR #19082, opened 2025-05-21
Extract device from wallet-core & unify it with firmware

### LINKED ISSUE #19072: Fix circular dependencies
Currently we have a lot of circular dependencies. I tried this, only in wallet-core as example:
```
npx madge --circular --extensions ts,tsx suite-common/wallet-core/src
```
and got ~90 circular graphs 😱  (but that number is heavily overstated, lot of them are subgraphs).

## Todo
- solve them one by one in each package
- when all are solved (or ignored with good reason), then estabilish a CI check
  - 💭 idk if `madge` is the best tool for this, we could go with something else for that
- 💭 it'd be best to start with the most upstream packages (e.g. connect), because downstream packages (e.g. suite) make a lot of noise including the cycles found upstream

## Why?

Webpack and TS are tolerant to ♻ on _files_, they will only throw when there are ♻ references on _variables_ (except function recursion).
So the app works fine... until it doesn't! When too much ♻, you get hard-to-debug errors: [what do you make of this](https://github.com/trezor/trezor-suite/actions/runs/15160019252/job/42623908071#step:5:557)? It was a ♻ which worked in the webpack app, but crashed in jest with an esoteric looking error (an import was imported as `undefined`).

peter-sanderson:
resolves https://github.com/trezor/trezor-suite/issues/19072


peter-sanderson:
failed attempt

PR #19107, opened 2025-05-22
refactor(suite): update layout size handling

seibei-iguchi:
🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=refactor/layout-size-change&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=refactor/layout-size-change&lookback=7)


seibei-iguchi:
@coderabbitai summary

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -215,187 +269,174 @@ export const ResizableBox = ({
...
     onWidthResizeEnd,
     onHeightResizeEnd,
     onWidthResizeMove,
     onHeightResizeMove,
     disabledWidthInterval,
     disabledHeightInterval,
+    ...rest
 }: ResizableBoxProps) => {
     const resizableBoxRef = useRef<HTMLDivElement>(null);
+    const rafRef = useRef<number | null>(null);
```
jvaclavik, line 282: raf? royal air force? :D

packages/suite/src/views/wallet/send/index.tsx
```diff
@@ -3,7 +3,7 @@ import { ReactNode } from 'react';
 import styled from 'styled-components';
 
 import { Banner, Column } from '@trezor/components';
-import { breakpointMediaQueries } from '@trezor/styles';
+import { SCREEN_QUERY } from '@trezor/components/src/config/variables';
```
jvaclavik, line 6: nit: I suggest to import just the package:

```tsx
import { variables } from '@trezor/components';

...

${variables.SCREEN_QUERY.ABOVE_DESKTOP} {
```

or export it explicitly in `@trezor/components`

jvaclavik: Also thinking about refactoring naming and structure without keeping in mind our previous state. 
It would be really nice have everything on the same place and not import it from 3 different packages with different naming conventions:
- `SCREEN_QUERY` from `@trezor/components` 
- `useLayoutSize` from `@trezor/hooks`
- `breakpoints` from `@trezor/theme`

Not sure if we are able to unify all of these or useLayoutSize must be somewhere else. In ideal world I would suggest to do something like this:

```tsx
import { media } from '@trezor/???'; 

// media object
{
  query: {
    mobile: `@media ...`,
    aboveMobile: `@media ...`,
    ...
  }
  breakpoints: {
    mobile: 576,
    table: 768,
    ...
  }
  useQuery: fn
}
```

WDYT?

packages/components/src/components/ResizableBox/ResizableBox.tsx
```diff
@@ -179,30 +196,67 @@ const getMaxResult = (max: number | undefined, result: number) =>
...
+                ...state,
+                isHovering: false,
+                direction: state.isResizing ? state.direction : null,
+            };
+        default:
+            return state;
+    }
 };
 
 export const ResizableBox = ({
```
jvaclavik, line 262: Good job here, this component needed some love. Reducer and 100% values look much better than before

packages/suite/src/hooks/guide/useGuide.ts
```diff
@@ -11,12 +9,10 @@ export const useGuide = () => {
...
 
-    const { layoutSize } = useLayoutSize();
+    const { isBelowLaptop } = useLayoutSize();
 
-    const isGuideOnTop = useMemo(
-        () => ['NORMAL', 'SMALL', 'TINY'].includes(layoutSize),
-        [layoutSize],
-    );
+    // The guide should be on top for smaller screens (below laptop size)
+    const isGuideOnTop = isBelowLaptop;
```
jvaclavik, line 15: nit: I suggest `isGuideOverlayContent`

Why: I had to test it in the app because I though it does overlay content all the time. This seems to me more descriptive what's actually hapening. Up to you.

jvaclavik (approved):
I like it very much! 👌

Just thinking if we can unify breakpoints, screen_query and useLayout to similar style and package. See my comment below

packages/suite/src/views/wallet/send/index.tsx
seibei-iguchi: I don't want to touch SCREEN_QUERY as it's deprecated:) breakpoints are in theme to follow suit with other similar constants. useLayoutSize must be in hooks, because it depends on Redux.

packages/components/src/components/ResizableBox/ResizableBox.tsx
seibei-iguchi: requestAnimationFrame:)

PR #19109, opened 2025-05-22
Feat/remove trailing slashes utils

karliatto:
## Description
New utility function `removeTrailingSlashes` to avoid duplicated regex.

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/removeTrailingSlashes-utils&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/removeTrailingSlashes-utils&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/removeTrailingSlashes-utils&lookback=7)

### DIFF
--- a/packages/connect/src/api/firmware/getBinary.ts
+++ b/packages/connect/src/api/firmware/getBinary.ts
@@ -1,8 +1,7 @@
 import type { FirmwareRelease } from '@trezor/device-utils';
+import { removeTrailingSlashes } from '@trezor/utils';
 
 import { httpRequest } from '../../utils/assets';
-
-const ALL_SLASHES_AT_THE_END_REGEX = /\/+$/;
 
 interface GetBinaryProps {
     baseUrl: string;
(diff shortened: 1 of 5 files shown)


marekrjpolak (approved):
This is perfectly fine, I'm just thinking about the CodeQL warning, which is somehow correct - on my testing string of hundred thousands of slashes and a single character, it runs approx. 5 seconds.

- One way is to revert the string, remove leading slashes and then revert it again.
- Another way is to replace a single non-slash character followed by slashes to the end by that single character alone (`replace(/([^/])\/+$/, '$1')`), which unfortunately doesn't work for slash-only strings.
- Or we can ignore the warning as we have control over the inputs and know that there won't be anything harmful.

karliatto:
> This is perfectly fine, I'm just thinking about the CodeQL warning, which is somehow correct - on my testing string of hundred thousands of slashes and a single character, it runs approx. 5 seconds.
> 
>     * One way is to revert the string, remove leading slashes and then revert it again.
> 
>     * Another way is to replace a single non-slash character followed by slashes to the end by that single character alone (`replace(/([^/])\/+$/, '$1')`), which unfortunately doesn't work for slash-only strings.
> 
>     * Or we can ignore the warning as we have control over the inputs and know that there won't be anything harmful.

My initial approach was `ignore the warning as we have control over the inputs` but since you commented on it, let's address it. I took the backward loop approach.

https://github.com/trezor/trezor-suite/pull/19109/commits/e31300eed166bf7212d349f7fd16b6128649e9f1#diff-09e5fb44e20ae56faec6bdca781c70284768e3667a26d02bba61a794de57e4b2R7

marekrjpolak:
>  I took the backward loop approach

I believe this is the best approach possible :+1:

PR #19138, opened 2025-05-23
test(suite-native): reassure performance tests init

PeKne:
## Description
- adds support for Reassure performance tests ([see more here](https://oss.callstack.com/reassure/))
  - this testing framework lets us test the number of renders and the render time of specific components
  - the result will be compared with the baseline branch (`develop`) and shout if there will be a significant performance drawback found
- added first tests for `@suite-natieve/assests` components and selectors that are suboptimal. After merging this PR I would like to improve the implementation and see the improvement based on these tests
- The `suite-native/assets/Assets.tsx` test simulates a running discovery by calling a `accountActions.createAccount()` action for various accounts.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=test/native/reassure-performace-tests&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=test/native/reassure-performace-tests&lookback=7)


PeKne:
@SocketSecurity ignore npm/@gitbeaker/rest@38.12.1
@SocketSecurity ignore npm/danger@13.0.5
@SocketSecurity ignore npm/memfs-or-file-map-to-github-branch@1.3.0
@SocketSecurity ignore npm/readline-sync@1.4.10


Can be ignored because danger is used only as a devDependency.

suite-native/app/scripts/reassure-tests.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+set -e 
+
+# TODO: change to develop after merger
+BASELINE_BRANCH=test/native/reassure-baseline-branch
```
PeKne, line 5: This needs to be changed to `develop` branch with a follow-up right after merging this PR. It can not be develop` right away because it does not contain the reassure library yet.

suite-native/assets/src/__tests__/perf/Assets.perf.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { connectedDeviceBeforeDiscoveryState } from './fixtures/connectedDeviceBeforeDiscoveryState';
+import { createAccountActions } from './fixtures/mockedDiscoveryFixtures';
+import { Assets } from '../../components/Assets';
+
+// The test should not rely on the fiat services, so we mock it to avoid the delay.
+jest.mock('@suite-common/fiat-services', () => ({
+    fetchCurrentFiatRates: jest.fn().mockResolvedValue({}),
+}));
+
+test('@suite-native/assets - Assets.tsx', async () => {
```
jbazant: **nitpick (non-blocking):** Here is used `test` and in next file (`assetsSelector.perf.ts`) is used `it` instead. WDYT about unifying it?‏

suite-native/test-utils/src/renderWithStore.tsx
```diff
@@ -42,6 +43,31 @@ export const renderWithStoreProviderAsync = async <Props,>(
     return ret;
 };
 
+export const measureRendersWithStoreProviderAsync = async <Props,>(
```
jbazant: **question:** Does it make sense to create `measureHookRendersWithStoreProviderAsync` as well? If yes, does it makes sense to you to move them to separate file?‏

suite-native/test-utils/src/renderWithStore.tsx
```diff
@@ -42,6 +43,31 @@ export const renderWithStoreProviderAsync = async <Props,>(
...
+        preloadedState,
+        wrapper: Wrapper,
+        store,
+        ...options
+    }: Reassure.MeasureRendersOptions & {
+        preloadedState?: PreloadedState;
+        store?: TestStore;
+    } = {},
+) => {
+    const injectedStore = store ?? (await initStore(preloadedState));
```
jbazant: **thought (non-blocking):** I believe we do not need this here. If we need initialized store, mark it as required prop and just pass it down when calling `measureRendersWithStoreProviderAsync`. 

This way it just more or less duplicates logic of `StoreProviderForTests`.‏

suite-native/test-utils/src/StoreProviderForTests.tsx
```diff
@@ -34,10 +34,10 @@ to pass `preloadedState` of `injectedStore` as a prop and without the `Persistor
  */
 export const StoreProviderForTests = ({
     children,
-    injectedStore,
+    injectedStore = null,
     preloadedState,
 }: ReduxProviderProps) => {
-    const [store, setStore] = useState<TestStore | null>(null);
+    const [store, setStore] = useState<TestStore | null>(injectedStore);
```
jbazant, line 40: **praise:** -1 re-render. Nice.‏

jbazant (approved):
LGTM

suite-native/assets/src/__tests__/perf/Assets.perf.tsx
PeKne: fixed 17e44f7ef2

suite-native/test-utils/src/renderWithStore.tsx
PeKne: reassure does not have `hook` equivalent of this function. But good point with moving it to other file: 17e44f7ef2

suite-native/test-utils/src/renderWithStore.tsx
PeKne: good point: 17e44f7ef2

Issue #18706, opened 2025-05-02
XLM: align handling in Trezor Suite with the Trezor device when activating a new account


Ondra-Zik:
**Describe the bug**
When you send Lumens to an empty address and want to activate it, the sending flow is completely different between the device and Trezor Suite.

**Info:**
 - Suite version: web 25.5.0 (bb28264b87ee6e45934a58c3e0947fa4748cb87c) 
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 2560x1440
 - Device: Trezor T3T1 2.8.10 regular (revision 41b786353e09390b64b107e64e1cfe3020514ad5)
 - Transport: WebUsbTransport

**Screenshots:**

![Image](https://github.com/user-attachments/assets/bb2d8a04-b39c-4231-8671-cf2e44a39ae9)

tomasklim:
Please check Trezor Model 1 and others

overcat:
`Account` and `Initial Balance` appear on the same confirmation page. I think we should modify it on the firmware side to split it into two pages.

Ondra-Zik:
**QA OK** with comments

The flow is better now, but still not totally ideal. I assume you'll be fixing it on the FW side, right @overcat ?

**Info:**
 - Suite version: web 25.6.0 (26069e28b54c7702b3e3b061604bf87324a2cd7b)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 3440x1440
 - Device: Trezor T3T1 2.8.10 regular (revision 2a65d18200580005dc419b9569ed97fae440806a)
 - Transport: WebUsbTransport

overcat:
Hi @Ondra-Zik-SL, yes, there will still be a slight inconsistency with the Create Account operation and the T1B1 device; we will improve it in the firmware.

PR #19201, opened 2025-05-29
fix(suite, native): fix the issue of inconsistency between the review flow and hardware display.

### LINKED ISSUE #19014: XLM: Align Trezor Suite with T1B1
**Describe the bug**  
Currently, Trezor Suite displays a generic "Follow instructions" message instead of the proper send flow, which only appears after confirming on the Trezor device.

**Info:**  
- Suite version: web 25.6.0 (f57a9e24a5c77183ff4c892c05c932772931d69a)  
- Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36  
- OS: Linux x86_64  
- Screen: 2880x1800  
- Device: Trezor T1B1 1.13.1 regular (revision 2a65d18200580005dc419b9569ed97fae440806a)  
- Transport: WebUsbTransport  

**Screenshots:**  
Image

overcat:
## Description

The review interface of Stellar transactions has discrepancies with the hardware display (Safe 3 / Safe 5 and Model T), and this PR fixes that issue.

For transactions related to creating new accounts, there will still be a slight deviation, but I do not intend to fix it in the suite; instead, I will modify the firmware. See https://github.com/trezor/trezor-suite/issues/18706#issuecomment-2911030570

For Model One, its flow is quite different from the other three devices, and I will modify it in another PR.

## Related Issue
#19014 and #18706

## Screenshots:
desktop
native


vytick (commented):
Mobile signing flow doesnt really follow what is on trezor screen. 

For example there is no confirmation of memo compared to what is there for ripple memo and it starts showing confirmed data only after a few steps are confirmed on trezor.

overcat:
> Mobile signing flow doesnt really follow what is on trezor screen.
> 
> For example there is no confirmation of memo compared to what is there for ripple memo and it starts showing confirmed data only after a few steps are confirmed on trezor.

Hi @vytick, which wallet device are you using?

vytick:
Trezor Safe 5 (I hope that I havent missed latest updates on the code)

overcat:
Hi @vytick, maybe you can try running `yarn build:libs` first? I just tested it on Safe 3 and Model T, and it worked as expected.

In addition, I did not design a separate page for memo; it only appears in the review entries and does not have any special design compared to other entries.

vytick:
I cleaned everything and tried once more and it works since then 🤷‍♂️. So there was probably something bad at my machine after I checked out this branch.

tomasklim (approved):
Desktop approved, checked on model T

tomasklim:
> ... and I will modify it in another PR.

I think we can also get rid of `Signing on testnet network` 
+ ignoring timebounds in case they are not set

Let's leave it in suite until it is changed in fw

overcat:
> I think we can also get rid of Signing on testnet network

I think it still makes sense to keep it, but we could consider merging it with other content and placing it on the last page.

suite-native/module-send/src/screens/SendOutputsReviewScreen.tsx
```diff
@@ -35,6 +38,27 @@ export const SendOutputsReviewScreen = ({
...
+        (state: AccountsRootState & DeviceRootState & SendRootState) =>
+            selectIsTransactionReviewInProgress(state, accountKey, tokenContract),
+    );
+
+    const handleOnDeviceTransactionReview = useHandleOnDeviceTransactionReview();
+
+    useFocusEffect(
+        useCallback(() => {
+            if (!isTransactionReviewInProgress && !hasReviewAlreadyStarted) {
+                handleOnDeviceTransactionReview();
```
PeKne: What about calling the `handleOnDeviceTransactionReview()` already at line 136 of the `SendAddressReviewScreen` file inside of the `if (networkType === 'stellar')` code block?

I would like it more because this file is now more complicated than it needs to be by stellar-specific code even though the keyword stellar is not mentioned anywhere in this file. Might bring some confusion during later development. If you move the call there, you can remove all the changes of this file added in this PR.

suite-native/module-send/src/components/SendFeesForm.tsx
```diff
@@ -132,6 +132,13 @@ export const SendFeesForm = ({ accountKey, tokenContract }: SendFormProps) => {
                 tokenContract,
                 transaction: selectedFeeLevelTransaction,
             });
+        } else if (networkType === 'stellar') {
+            // The first review entry of Stellar is neither a destination address nor a destination tag.
```
PeKne, line 143:
```suggestion
            // The first review entry of Stellar is neither a destination address nor a destination tag.
            handleOnDeviceTransactionReview();
```

overcat: Hi @PeKne, I think we can't simply use this method because `handleOnDeviceTransactionReview` internally uses [`const { transaction } = route.params;`](https://github.com/trezor/trezor-suite/blob/6093a076fbff8b693ce596464cad0d2bf3ea4ee7/suite-native/module-send/src/hooks/useHandleOnDeviceTransactionReview.tsx#L42
), and this tx is generated on [line L101](https://github.com/trezor/trezor-suite/blob/6093a076fbff8b693ce596464cad0d2bf3ea4ee7/suite-native/module-send/src/components/SendFeesForm.tsx#L101
), but we haven't placed it into `route.params` yet.

suite-native/module-send/src/screens/SendOutputsReviewScreen.tsx
overcat: In fact, it is not only applicable to Stellar but also to all transactions that do not use the destination address/memo as the first review item. However, among the chains we currently support, only Stellar meets this condition, so it appears to be specific to Stellar.

suite-native/module-send/src/components/SendFeesForm.tsx
PeKne: Then you would have it pass it as a prop to the `usHandleOnDeviceTransactionReview` hook instead of reading it directly from the navigation. But if you don't like it I propose another solution:

1. You keep all the code that was added to `SendOutputsReviewScreen` there, but you will extract it to some Stellar-specific hook. e.g. `useHandleStellarOnDeviceTransactionReview`.
2. this hook triggers strictly for the Stellar network only

What I am trying to achieve is separation of concern. In the desktop suite, there is everything happening in one single complicated hook and there is always running a lot of network-specific code that would not have to be executed most of the time. It is very non-transparent and bug-prone. We would like to avoid this in mobile (native) code.

What do you think?

overcat: Hi @PeKne, thanks for the hit! See 6c1734f91e57fd267169c21563abc2b4f2007784

PeKne (approved):

overcat:
Hi, if you need me to rebase and squash, please let me know.

tomasklim:
@overcat yes please

suite-common/wallet-core/src/send/sendFormSelectors.ts
```diff
@@ -50,13 +50,15 @@ export const selectSendFormReviewButtonRequestsCount = (
 
     const isCardano = networkType === 'cardano';
     const isEthereum = networkType === 'ethereum';
+    const isStellar = networkType === 'stellar';
```
overcat, line 53: Hi @tomasklim, finished. 

P.S. This new change is due to the relocation of the relevant code, which is consistent with the previous changes.

PR #19204, opened 2025-05-29
chore(suite, suite-build): serve guide markdown as static assets and load via fetch

### LINKED ISSUE #18075: In vite dev server, guide content cannot be displayed
There is a markdown file generated and it is not properly loaded to the guide when using vite dev server

matusbalascak:
## Description

Guide files are now loaded with `fetch()` instead of a dynamic `import()`, which makes it work across different build tools.

The only build config tweak is to mark the entire `guide/` directory as static so the Markdown files are included. This doesn’t increase the bundle size; in fact, it’s now slightly smaller.

An alternative is `import.meta.glob` (with runtime checks to distinguish Vite from Webpack), but that adds tool-specific complexity unrelated to the feature itself.

**suite-build**

- Updated `CopyWebpackPlugin` to copy the `suite-data/files/guide` directory into `build/static/guide` (instead of only `guide/assets`).

**suite**

- Switched `useGuideLoadArticle` to load guide files via `fetch('/static/guide/${language}${id}')` instead of dynamic `import()`.

## Related Issue

Resolve [#18075](https://github.com/trezor/trezor-suite/issues/18075)

### DIFF
--- a/packages/suite-build/configs/desktop.webpack.config.ts
+++ b/packages/suite-build/configs/desktop.webpack.config.ts
@@ -31,7 +31,7 @@ const config: webpack.Configuration = {
     },
     plugins: [
         new CopyWebpackPlugin({
-            patterns: ['bin', 'fonts', 'images', 'videos', 'guide/assets']
+            patterns: ['bin', 'fonts', 'images', 'videos', 'guide']
                 .map(dir => ({
                     from: path.join(__dirname, '..', '..', 'suite-data', 'files', dir),
                     to: path.join(baseDir, 'build', 'static', dir),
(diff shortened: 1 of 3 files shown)


vojtatranta:
Hi @matusbalascak .
I deployed this PR here https://dev.suite.sldev.cz/suite-web/test-deploymatusbalascak-chore-fetch-guide-files/web/start.
Unfortunately, the Guide content isn't loaded there.

So somehow, this breaks the build. Standard develop deployment works https://dev.suite.sldev.cz/suite-web/develop/web/start

I am also testing the desktop build.

Vite and Webpack dev servers works well and Electron dev also runs well.

vojtatranta:
Yeah, so the both production builds are broken. Web and Desktop. Guide contents won't load.

matusbalascak:
Thanks for checking it @vojtatranta! 

I've created a new PR for this so I'm not working with a forked version. You can find it [here](https://github.com/trezor/trezor-suite/pull/19654) and I'll close this one.

It also includes the fix for what you described. I wrapped the url that's being fetched with `resolveStaticPath` and it seems to have fixed the issue as now it has a proper prefix where needed.

PR #19242, opened 2025-05-31
Feat: update fee rate options regularly

### LINKED ISSUE #18406: Update fee rate regularly while being in send form

Lemonexe:
## Description

Update fee rates regularly at all places where fee selection is available.

- periodicaly refetch fees every 60s
  - it works everywhere where the Fees levels options are displayed (not just send form)
- display skeletons when loading [as per figma](https://www.figma.com/design/qx5aOavcU7I2MW2PtnQtfz/Small-Things--DSK-?node-id=4048-978&t=MFI4WubrD2PkITeY-1)
- a little cleanup in feeThunks
- new `isApproxEqual` util for bignumbers

what I did different than specified in Figma:
- disabled the Review & send button consistently with the loaders around it
  - otherwise it would flick on/off, but with inconsistent timing (only as the draft is recomposed)
- the fetching is very quick, so if we displayed loaders only during the loading itself, they just blink on/off quickly.
  - that looks disruptive and it's not clear what is happening, just that _something blinked on the screen_
  - the point of the loaders is to visually signify to user that the fees are changing, so I extend the loader to at least one second. IMHO it works very well, see videos below :slightly_smiling_face:
- it also works in: RBF, Stake, Unstake, Trading–sell
  - the UX details are not that polished as in Send Form (no figma designs). But it works and looks good :heavy_check_mark:  

## Related Issue

Resolve #18406

## Screenshots:

Video of the new behavior, when high fee value is selected and it's just changing: 

[demo.webm](https://github.com/user-attachments/assets/a0354d18-3949-464d-b0ff-0af6590d2573)

Playing around with various scenarios (it does not lose my progress, it does not reset my fee level choice, dust prevention works etc.), with update time lowered to 10 seconds

[playing around.webm](https://github.com/user-attachments/assets/ecd01cef-9a68-41eb-a82c-796b14fc5b40)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/send-form-update-fees-regularly&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/send-form-update-fees-regularly&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/send-form-update-fees-regularly&lookback=7)


suite-common/wallet-core/src/fees/feesConstants.ts
```diff
@@ -0,0 +1 @@
+export const FEES_UPDATE_INTERVAL = 60_000; // [ms] interval to refetch estimated fees from blockbook
```
tomasklim:
```suggestion
export const FEES_UPDATE_INTERVAL = 60_000; // [ms] interval to refetch estimated fees from backend
```

tomasklim: we use also other backends, not just blockbook

packages/suite/src/components/wallet/Fees/useRefetchFees.ts
```diff
@@ -0,0 +1,44 @@
...
+    };
+
+    // Initial fetch when component mounts
+    useEffect(() => {
+        refetchFees(0);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    // Refetch fees periodically
+    useInterval(() => refetchFees(MIN_LOADER_LIFETIME), FEES_UPDATE_INTERVAL);
```
tomasklim: is it good idea to refetch it in case user chooses `Send max`? the value can change under his hands

Lemonexe: Yes, it will change under your hands, and as I understand that's a desired scenario.
When you send max, you usually don't care about specific exact output, but just to spend the full input, no?

packages/suite/src/components/wallet/Fees/DustPreventionNotice.tsx
```diff
@@ -15,10 +16,11 @@ export const DustPreventionNotice = ({
     baseFee,
     feeUnits,
 }: DustPreventionNoticeProps) => {
+    const relTol = 1e-3;
     const isComposedFeeRateDifferent =
         composedFeePerByte !== undefined &&
-        composedFeePerByte !== '' &&
-        chosenFeePerByte !== composedFeePerByte;
+        chosenFeePerByte !== undefined &&
+        !isApproxEqual(composedFeePerByte, chosenFeePerByte, relTol);
```
Lemonexe: I found that the hard way that recomposing the transaction with unchanged feeLevels may produce different string representation of the same number: `'5.1'` vs `'5.10'`
and the `DustPreventionNotice` would display when it shouldn't.

packages/suite/src/hooks/wallet/useSendForm.ts
```diff
@@ -348,7 +349,14 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
...
         draft.current = undefined;
-    }, [draft, composeDraft]);
+        // composeDraft is excluded because its reference changes with each feeInfo update.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [draft]);
+
+    // update composedLevels when feeInfo changes
+    useEffect(() => {
+        composeDraft(getValues());
+    }, [composeDraft, getValues]);
```
Lemonexe, line 359: basically I extracted partial functionality from the hook above: when fees change, just recompose levels but do not reset the whole form..

Lemonexe: :thought_balloon:  I am not happy that Send Form relies so much on useEffects.

The lifecycle is really unpredictable – there was original assumption that the two useEffects above would run only once when mounted, but now `composeDraft` fn legitimately changes reference when fees are updated.. 

I'm not the first one to have this problem, maybe someone will rewrite it to OOP one day :sweat_smile:

packages/utils/src/isApproxEqual.ts
```diff
@@ -0,0 +1,24 @@
...
+    value2 = new BigNumber(value2);
+    relTol = new BigNumber(relTol);
+
+    // Cannot calculate relative difference if one value is zero, but then they must be equal
+    if (value1.eq(0)) return value1.eq(value2);
+
+    const relativeDifference = value2.minus(value1).abs().dividedBy(value1);
+
+    return relativeDifference.lte(relTol);
+};
```
Lemonexe, line 24: This util may seem like overkill, but it's methodically correct to do it when comparing floats.
There could be a lot of places in code where this should be used instead of `===`.
Also I had fun writing it :heart: 

Btw `bignumber.js` does not have this, not even simpler version with absolute tolerance.

suite-common/wallet-core/src/fees/feesThunks.ts
Lemonexe, line 1: in this file I only moved code around and inverted conditions to early returns, so that it is better readable.
No functional changes here.

suite-common/wallet-core/src/send/sendFormThunks.ts
```diff
@@ -606,3 +608,22 @@ export const enhancePrecomposedTransactionThunk = createThunk<
...
+
+        dispatch(sendFormActions.setAreFeesLoading(true));
+        const FEE_UPDATE_DELAY = 1000; // [ms]
+        await Promise.all([
+            dispatch(updateFeeInfoThunk({ networkSymbol })),
+            new Promise(resolve => setTimeout(resolve, FEE_UPDATE_DELAY)),
+        ]);
+        dispatch(sendFormActions.setAreFeesLoading(false));
+    },
+);
```
Lemonexe, line 629: why did I put this into `sendFormReducer`?
Initially I started in `feesReducer`. But you can see that its state is just `Record<NetworkSymbol, FeeInfo>` so there would be no place to easily add it. I'd either have to append a veery hacky `& {commonState: {} }`, or refactor it completely `state` to `{ feesPerNetwork: state, ...otherThings }`, and that would be a very extensive refactoring :cry: 

But I belive that this actually makes some sense. The loading state is just an visual indication for send form that serves no relation to how the fees are actually fetched (they are already fetched and processed in reducer even when the artificial delay is still ticking). This "loading state" is just for send form..

Lemonexe:
the failing desktop E2E [is confirmed broken](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1748802827002579)
the failing android E2E seems broken as well ([failed also here](https://app.currents.dev/run/9f046eabddc75ed1/specs?filter=FAILED&filter=FAILING&filter=CANCELLED&filter=TIMED_OUT) which is unrelated)

packages/suite/src/components/wallet/Fees/DustPreventionNotice.tsx
```diff
@@ -15,10 +16,11 @@ export const DustPreventionNotice = ({
     baseFee,
     feeUnits,
 }: DustPreventionNoticeProps) => {
+    const relTol = 1e-3;
```
peter-sanderson: :pray: `relativeTolerance` I had to google it.

packages/utils/tests/isApproxEqual.test.ts
```diff
@@ -0,0 +1,32 @@
+import { BigNumber } from '../src/bigNumber';
+import { isApproxEqual } from '../src/isApproxEqual';
+
+describe(isApproxEqual.name, () => {
+    it('always returns true for equal values', () => {
+        expect(isApproxEqual(100, 100, 1e-3)).toBe(true);
```
peter-sanderson: I would prefer `0.001` over scientific format. :pray: I understand that when you want to express some really small number it is useful, but for `0.001` i would prefer conventional way so the common folk can easily understand.

suite-common/wallet-core/src/fees/feesConstants.ts
```diff
@@ -0,0 +1 @@
+export const FEES_UPDATE_INTERVAL = 60_000; // [ms] interval to refetch estimated fees from backend
```
peter-sanderson: nit: `FEES_UPDATE_INTERVAL_MILLIS`

tomasklim:
second video, 18 sec, fee rate adjusted appears there for few ms and then disappears again. We should probably hide it during loading
Screenshot 2025-06-02 at 9 53 22

packages/suite/src/hooks/wallet/useRefetchFees.ts
```diff
@@ -0,0 +1,27 @@
...
+export const useRefetchFees = ({ networkSymbol }: UseRefetchFeesProps) => {
+    const dispatch = useDispatch();
+
+    // Initial fetch only when component mounts
+    useEffect(() => {
+        dispatch(updateFeeInfoThunk({ networkSymbol }));
+    }, [dispatch, networkSymbol]);
+
+    // Refetch fees periodically incl. loading behavior
+    useInterval(() => {
```
peter-sanderson, line 24: This is cool! I didn't know about this util. Its handy so we dont need to handle the unount every time.

packages/suite/src/hooks/wallet/useSendForm.ts
```diff
@@ -335,7 +334,9 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
             }
         };
         loadDraftValues();
-    }, [dispatch, getLoadedValues, reset, composeDraft]);
+        // composeDraft is excluded because its reference changes with each feeInfo update.
```
peter-sanderson, line 337: I love the sendForm. 
image

suite-common/wallet-core/src/send/sendFormActions.ts
```diff
@@ -50,6 +50,11 @@ const sendRaw = createAction(`${SEND_MODULE_PREFIX}/sendRaw`, (payload: boolean)
 
 export const dispose = createAction(`${SEND_MODULE_PREFIX}/dispose`);
 
+const setAreFeesLoading = createAction(
+    `${SEND_MODULE_PREFIX}/setAreFeesLoading`,
+    (payload: boolean) => ({ payload }),
```
peter-sanderson: I prefer wrapping the payload to a {}. So we use named parameter everywhere. Here it is a nit, but I really think in general this is a good practice. It makes refactoring easier (adding new parameter)

suite-common/wallet-core/src/send/sendFormThunks.ts
```diff
@@ -606,3 +608,22 @@ export const enhancePrecomposedTransactionThunk = createThunk<
...
+ * Delay updateFeeInfoThunk with an arbitrary timeout, because backend request is usually very quick.
+ * This can be used to display loader for a bit longer, to visually draw users attention to the fees which are changing.
+ */
+export const delayedUpdateFeeInfoThunk = createThunk(
+    `${SEND_MODULE_PREFIX}/delayedUpdateFeeInfoThunk`,
+    async ({ networkSymbol }: { networkSymbol: NetworkSymbol }, { dispatch, getState }) => {
+        if (selectAreFeesLoading(getState())) return;
+
+        dispatch(sendFormActions.setAreFeesLoading(true));
+        const FEE_UPDATE_DELAY = 1000; // [ms]
```
peter-sanderson: - can be outside of the function
- `FEE_UPDATE_DELAY_MILLIS`

suite-common/wallet-core/src/send/sendFormThunks.ts
```diff
@@ -606,3 +608,22 @@ export const enhancePrecomposedTransactionThunk = createThunk<
...
+        await Promise.all([
+            dispatch(updateFeeInfoThunk({ networkSymbol })),
+            new Promise(resolve => setTimeout(resolve, FEE_UPDATE_DELAY)),
```
peter-sanderson: maybe we can have util for this, there is 56 usages of this pattern

peter-sanderson (approved):

packages/utils/tests/isApproxEqual.test.ts
Lemonexe: ah, old habits dies hard :laughing: 
I'm used to writing mathematical code and find e-3 easier to read than counting zeroes, but I understand that most people aren't used to it and have to make extra effort to "decode" it.

suite-common/wallet-core/src/send/sendFormThunks.ts
Lemonexe: maybe.. It could save some time, but it sounds a little bit like `isOdd` to me – a pattern that's so short and trivial that everyone writes it by hand; using an external function for it wastes my time, cause I have to import it, and when reading it I still click through it to confirm that it does what I assume it does..
But idk, maybe it's good for this. You can do a PR for that, I don't think it's worth in _this PR_ :slightly_smiling_face:

packages/suite/src/hooks/wallet/useRefetchFees.ts
Lemonexe: Yeah, I started first writing it by hand, and realizing the complexity, but I'm a lazy programmer so I looked for easier path :smile:  

Otherwise I would have to create `useEffect` with eslint-ignored exhaustive deps + relying on `useRef` for the timeoutId.

packages/suite/src/components/wallet/Fees/DustPreventionNotice.tsx
Lemonexe: f9dca71d

packages/utils/tests/isApproxEqual.test.ts
Lemonexe: f9dca71d

suite-common/wallet-core/src/fees/feesConstants.ts
Lemonexe: c189360c
I wrote `_MILLISECONDS` to make it clear.

suite-common/wallet-core/src/send/sendFormActions.ts
Lemonexe: c189360c

suite-common/wallet-core/src/send/sendFormThunks.ts
Lemonexe: c189360c
I wrote `_MILLISECONDS` to make it clear.

Lemonexe:
> second video, 18 sec, fee rate adjusted appears there for few ms and then disappears again. We should probably hide it during loading

thanks, I noted I should fix it but then forgot to do it. Fixed in 384543dc

suite-common/wallet-core/src/fees/feesThunks.ts
```diff
@@ -71,84 +75,84 @@ export const preloadFeeInfoThunk = createThunk(
...
-                ...newFeeInfo,
-            };
+        const newFeeInfo = await getNewFeeInfo({ network, device });
+        if (newFeeInfo === undefined) return;
 
-            dispatch(feesActions.updateFee(partial));
-        }
+        const partialFees: Partial<NetworksFees> = {};
+        partialFees[network.symbol] = {
+            blockHeight: blockchainInfo.blockHeight,
```
tomasklim, line 152: is `blockHeight` required in fees? AFAIK some blocks for some networks are skipped (like that it is refreshed just once per minute etc.)

Lemonexe: Not sure tbh. In _this file_ I just shoveled code around.
At quick glance I see it is used [at least here](https://github.com/trezor/trezor-suite/blob/6e0122ec86d465952d22b068870f2365b65369cd/packages/suite/src/hooks/wallet/trading/form/common/useTradingComposeTransaction.ts#L102)

PR #19244, opened 2025-05-31
feat(suite): prevent sending funds to ATA address

### LINKED ISSUE #19145: Warn user when sending funds to associated account
**Describe the bug**

- Currently the user can send funds to ATA address without any warning.
- We can reduce the risk by giving users a warning when they are trying to send funds to ATA address, instead of wallet address
- Simple alert box stating what is happening should do the trick.

Image

Inspire by Ethereum implementation
Image

viktordanko:
## Description

When user tries to send funds to ATA address, he will be prompted by the warning and we prevent the transaction from happening. This reduces the risk of sending funds to the wrong address.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19145

## Screenshots:
image

## TODO

- [ ] show `Learn more` button with appropriate link
- [ ] improve types, so we don't need to cast [here](https://github.com/trezor/trezor-suite/compare/develop...viktordanko:trezor-suite:feat/prevent-sending-funds-to-ATA-address?expand=1#diff-66f3f4da04de541a8ab04117627c64d43abb6e28e03534d82c4d3ae6feccb6cfR396)
- [ ] add tests


packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -285,6 +290,23 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
...
+                if (networkType === 'solana') {
+                    const { payload, success } = await TrezorConnect.getAccountInfo({
+                        descriptor: value,
+                        coin: symbol,
+                        details: 'txs',
+                    });
+
+                    if (!success) {
+                        return translationString('RECIPIENT_IS_NOT_VALID');
+                    }
```
tomasklim: User can be offline, not sure if good message. For ETH we use `TR_ETH_ADDRESS_CANT_VERIFY_HISTORY`

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -285,6 +290,23 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                     return translationString('RECIPIENT_IS_NOT_VALID');
                 }
             },
+            sol_account_type: async (value: string) => {
```
tomasklim, line 305: you are missing `I understand` button which allows user to continue. We don't block users from using their funds

packages/suite/src/components/wallet/InputError.tsx
```diff
@@ -18,7 +18,7 @@ export const InputError = ({ buttonProps, learnMoreUrl, message }: InputErrorPro
     <Row gap={spacings.xs} justifyContent="space-between" flex="1">
         <Row gap={spacings.xs}>
             <Paragraph>{message}</Paragraph>
-            {learnMoreUrl && <LearnMoreButton url={learnMoreUrl} />}
+            {learnMoreUrl && <LearnMoreButton url={learnMoreUrl} textWrap={false} />}
```
tomasklim, line 21: Now it is not unified
Screenshot 2025-06-01 at 13 46 44

packages/blockchain-link-types/src/common.ts
```diff
@@ -271,6 +274,7 @@ export interface AccountInfo {
         rent?: number; // The rent required for the account to opened
         solStakingAccounts?: SolanaStakingAccount[]; // Solana staking accounts
         solEpoch?: number; // Solana current epoch
+        accountInfo?: ParsedAccountData;
```
tomasklim: sorry for wrong hint, it can be read from `owner` field, no need to introduce this

nok
Screenshot 2025-06-01 at 13 48 38
ok
Screenshot 2025-06-01 at 13 49 24
ok
Screenshot 2025-06-01 at 13 51 34

warning should be visible in case owner is not `undefined` or `11111111111111111111111111111111`

packages/suite/src/views/wallet/send/Outputs/Address.tsx
```diff
@@ -236,6 +236,11 @@ export const Address = ({ output, outputId, outputsCount }: AddressProps) => {
                         text: translationString('TR_ADD_BITCOINCASH_PREFIX'),
                     },
                 };
+            case 'sol_account_type':
+                return {
+                    // TODO: add link to packages/urls/src/urls.ts when article is ready
+                    // learnMoreUrl: 'https://trezor.io',
```
tomasklim: return at least some placeholder so that it works and can be validated

I think something like

```
export const HELP_CENTER_SOLANA_HELP_URL =
    'https://trezor.io/support/a/where-is-my-solana';
```

could work

tomasklim (changes requested):
Hello Viktor, thank you for your contribution. Few comments

PS: please always use latest develop branch, your is almost 1 month old

packages/blockchain-link-types/src/common.ts
viktordanko: Thanks for clarification, that was my initial thought

packages/suite/src/components/wallet/InputError.tsx
viktordanko: Gonna prevent wrapping also the second one

packages/suite/src/views/wallet/send/Outputs/Address.tsx
viktordanko: Kk

packages/suite/src/views/wallet/send/Outputs/Address.tsx
viktordanko: Kk, will do

packages/suite/src/views/wallet/send/Outputs/Address.tsx
viktordanko: Good point, thanks

viktordanko:
Thanks for the review, @tomasklim. 
I'm gonna process the feedback in the afternoon

packages/suite/src/views/wallet/send/Outputs/Address.tsx
viktordanko: Would it be better to unify the names for ETH-like message and use it also here or should I create a new one?

tomasklim: unify

tomasklim: `TR_ADDRESS_CANT_VERIFY_HISTORY`

PR #19292, opened 2025-06-03
chore(connect): log error when trying to call device which was not …

mroz22:
…Initialized

There are occasional [occurrences](https://satoshilabs.sentry.io/issues/6620590564/events/?environment=production&project=5193825&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20entropy&referrer=issue-stream&stream_index=0) of `Invalid session` error in Sentry. [Most likely](https://github.dev/trezor/trezor-firmware/blob/0081788d48af58daad600054ef65b1b53bf79c16/legacy/firmware/config.c#L581) because we send a call which requires calling Initialize first without calling it. 

Here I am adding console.error, which should (is it right? @Lemonexe) be propagated to Sentry to get some more intel on this. Later we might turn console.error to hard error if we are sure there are no false positives. 

cc @marekrjpolak for post-merge review some day later

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=meditation-over-entropy-check-false-positives&lookback=7)

### DIFF
--- a/packages/connect/src/device/DeviceCurrentSession.ts
+++ b/packages/connect/src/device/DeviceCurrentSession.ts
@@ -21,6 +21,19 @@ const blacklist: Record<string, string[] | true> = {
     Features: true,
 };
 
+const allowedCallsBeforeInitialize: Messages.MessageKey[] = [
+    // Preventive Cancel
+    'Cancel',
+    // At the beginning of each call either Initialize or GetFeatures is called.
+    'Initialize',
+    'GetFeatures',
+    // After one of the two above there might be optionally some of these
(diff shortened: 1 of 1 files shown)


Lemonexe (approved):
Thanks a lot :pray: 
Tested suite desktop & web, device connects normally.

Yes, console.errors are picked up in Suite **Web**. Now thinking about it, in Site Desktop I don't think it will work automatically, but the sample from Web should be enough to tell if this happens at all or not.

PR #19298, opened 2025-06-03
fix(connect): refactor check firmware revision download

karliatto:
## Description
Having firmware release related code together makes it more clear to understand what is happening.  And adding some missing types.


mroz22 (approved):
looks ok

PR #19309, opened 2025-06-03
fix(transport): make sure that chunk size is even with chunkSize

szymonlesisz:
## Description

make sure that chunk is equal with transport medium chunksize. same as in `UsbApi`

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/transport-chunk-size&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/transport-chunk-size&lookback=7)

### DIFF
--- a/packages/transport-bluetooth/src/client/bluetooth-api.ts
+++ b/packages/transport-bluetooth/src/client/bluetooth-api.ts
@@ -90,8 +90,11 @@ export class BluetoothApi extends AbstractApi {
     }
 
     write(path: string, buffer: Buffer) {
+        const chunk = Buffer.alloc(this.chunkSize);
+        buffer.copy(chunk);
+
         return this.api
-            .send('write', [path, Array.from(buffer)])
+            .send('write', [path, Array.from(chunk)])
             .then(() => this.success(undefined))
             .catch(e => this.error({ error: ERRORS.INTERFACE_DATA_TRANSFER, message: e.message }));
     }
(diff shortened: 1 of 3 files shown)


packages/transport-bluetooth/src/client/bluetooth-api.ts
mroz22, line 1: at the first glance, why is this better than checking `buffer` (incoming param) byteLenght and comparing it with `this.chunkSize`?

szymonlesisz: it's not about check, rather code unification

same as we do in `api/usb.ts`

```
const newArray = new Uint8Array(this.chunkSize);
newArray.set(new Uint8Array(buffer));
```

if chunkSize is 64 and message(chunk) is smaller then you want to fill the rest of the chunk with zeroes so 123456 becomes 12345600000000000....00

this will be requirement for THP firmware

mroz22: understood, thanks

mroz22 (approved):

packages/transport-bluetooth/src/client/bluetooth-api.ts
szymonlesisz: well, i've found an exception in udp and PINGPING :(

https://github.com/trezor/trezor-suite/pull/19309/commits/fa8c45c451f56a07039dc8adaef44b8ae715a2ea

szymonlesisz:
also cc @yanascz as im adding the same for native BlutoothApi https://github.com/trezor/trezor-suite/pull/19309/commits/27008fdf9a0b34ec74a9e187de41b8f662b1c088

yanascz (approved):
Native part looks good. 👍

PR #19335, opened 2025-06-04
attempt to fixed way-long loading when adding a non-empty passph…

vojtatranta:
…rase wallet

## Description

Bug: When the discovery is in state when it is known that the passphrase wallet ISNT empty, so that wallet can be opened nad user can interact with the app while the accounts are being loaded, the app is still in "loading state" with a discovery loading modal on

The solution:
- we don't want to wait for the discovery to completely finish to "hide" the discovery loader in the passphrase flow context
- create a "hook" to the discovery progress handler, which will trigger only once to select the added wallet
- add a flag to the discovery to be able to trigger said hook and to if-out the "passphrase flow" modals, so their are hidden

## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix-stucked-progress-passphrase&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix-stucked-progress-passphrase&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix-stucked-progress-passphrase&lookback=7)

### DIFF
--- a/packages/suite/src/hooks/suite/usePreferredModal.ts
+++ b/packages/suite/src/hooks/suite/usePreferredModal.ts
@@ -1,11 +1,11 @@
 import { Route } from '@suite-common/suite-types';
+import { selectHasRunningDiscovery } from '@suite-common/wallet-core';
 import { UI } from '@trezor/connect';
 
 import { MODAL } from 'src/actions/suite/constants';
 import { useDiscovery, useSelector } from 'src/hooks/suite';
 import type { ForegroundAppRoute } from 'src/types/suite';
 import { ModalAppParams } from 'src/utils/suite/router';
-import { selectDiscoveryOverallStatus } from 'src/utils/wallet/selectDiscoveryOverallStatus';
 
 const isForegroundApp = (route: Route): route is ForegroundAppRoute =>
     !route.isFullscreenApp && !!route.isForegroundApp;
@@ -46,15 +46,17 @@ const getForegroundAppAction = (route: ForegroundAppRoute, params: Partial<Modal
 
 export const usePreferredModal = () => {
     const { discovery: discoveryForSelectedDevice } = useDiscovery();
-    const discoveryStatus = useSelector(selectDiscoveryOverallStatus);
     const route = useSelector(state => state.router.route);
     const params = useSelector(state => state.router.params as Partial<ModalAppParams>);
     const modal = useSelector(state => state.modal);
+    const discoveryInProgress = useSelector(selectHasRunningDiscovery);
     const isPassphraseFlow =
         Boolean(discoveryForSelectedDevice?.isAddingHiddenWallet) &&
-        discoveryForSelectedDevice?.status !== 'cancelled' &&
-        discoveryForSelectedDevice?.status !== 'complete' &&
-        discoveryForSelectedDevice?.status !== 'failed';
+        discoveryInProgress &&
+        !(
+            discoveryForSelectedDevice?.status === 'progress' &&
+            discoveryForSelectedDevice.hasLoadedAnyNonEmptyAccount
+        );
 
     if (route && isForegroundApp(route) && hasPriority(route)) {
         return getForegroundAppAction(route, params);
@@ -90,15 +92,6 @@ export const usePreferredModal = () => {
     if (isPassphraseFlow && discoveryForSelectedDevice) {
         return {
             type: 'passphrase-flow',
-        } as const;
-    }
-
-    // account discovery in progress and didn't find any used account yet.
-    // display Loader wrapped in modal above requested route to keep "modal" flow continuity.
-    // or display "Action modal" (like: pin/passphrase request)
-    if (discoveryStatus?.type === 'auth-confirm') {
-        return {
-            type: 'discovery-loading',
         } as const;
     }
 
--- a/suite-common/wallet-core/src/discovery/discoveryReducer.ts
+++ b/suite-common/wallet-core/src/discovery/discoveryReducer.ts
@@ -16,9 +16,17 @@ const update = (draft: Discovery, payload: { status: DiscoveryStatus; path: Devi
     if (!draft[payload.path]) {
         return;
     }
+
+    const currentStatus = draft[payload.path];
+    const hasLoadedAnyNonEmptyAccount: true | undefined =
+        (currentStatus.status === 'progress' && currentStatus.hasLoadedAnyNonEmptyAccount) ||
+        (payload.status.status === 'progress' && payload.status.hasLoadedAnyNonEmptyAccount) ||
+        undefined;
+
     draft[payload.path] = {
-        ...draft[payload.path],
+        ...currentStatus,
         ...payload.status,
+        ...{ hasLoadedAnyNonEmptyAccount },
     };
 };
 
--- a/suite-common/wallet-core/src/discovery/discoveryThunks.ts
+++ b/suite-common/wallet-core/src/discovery/discoveryThunks.ts
@@ -97,6 +97,18 @@ const applyDeviceStatesThunk = createThunk(
             const devices = selectDevices(getState());
             const devicesByPath = devices.filter(d => d.path === devicePath);
 
+            const currentDeviceByStaticSessionId = newDeviceState.staticSessionId
+                ? selectDeviceByStaticSessionId(getState(), newDeviceState.staticSessionId)
+                : null;
+
+            if (currentDeviceByStaticSessionId && isAddingHiddenWallet) {
+                console.warn(
+                    'applyDeviceStatesThunk: applying state to a device with static session id',
+                );
+
+                return;
+            }
+
             // sanity check that there is no 2 devices sharing the same path. this shouldn't happen, the only way that comes to my mind
             // is when you would create a copy of device and store it in redux before authorizing it (this is actually the old way of doing things)
             // todo: this sanity check could be moved somewhere higher.
@@ -455,9 +467,6 @@ export const runDiscoveryThunk = createThunk(
                 return;
             }
 
-            // NOTE: this should be true anytime the applyDeviceStatesThunk() is called
-            let deviceStateApplied = false;
-
             const onBundleProgress = createOnBundleProgressHandler(
                 device.path,
                 deviceStateResponse.payload._state.staticSessionId,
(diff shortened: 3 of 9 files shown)


suite-common/wallet-core/src/discovery/discoveryReducer.ts
```diff
@@ -16,9 +16,17 @@ const update = (draft: Discovery, payload: { status: DiscoveryStatus; path: Devi
     if (!draft[payload.path]) {
         return;
     }
+    const currentStatus = draft[payload.path];
     draft[payload.path] = {
-        ...draft[payload.path],
+        ...currentStatus,
         ...payload.status,
+        ...{
+            hasLoadedAnyNonEmptyAccount:
```
vojtatranta: no: this is set only once on the discovery

suite-common/wallet-types/src/discovery.ts
```diff
@@ -45,6 +45,7 @@ export type DiscoveryStatus = CommonDiscoveryStatus &
               status: 'progress';
               total: BundleProgress<any>['payload']['total'];
               progress: BundleProgress<any>['payload']['progress'];
+              hasLoadedAnyNonEmptyAccount?: boolean; // NOTE: used to indicate the the disocovery started loading actual accounts
```
Lemonexe, line 48: isn't this information duplicated? You can select something like this, right?
```
accounts.some(a => a.staticSessionId === staticSessionId && !a.empty )
```

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -206,17 +209,31 @@ const createOnBundleProgressHandler = (
             return;
         }
 
+        const hasLoadedAnyNonEmptyAccount =
+            'balance' in event.response &&
+            !isNaN(Number(event.response.balance)) &&
```
Lemonexe, line 209: superfluous because the next line will always return false on `NaN > 0`

Lemonexe (commented):
I don't think this is an optimal solution; I believe same should be achieved much easier with `accounts`, see comment. Maybe I'm missing something?

packages/suite/src/hooks/suite/usePreferredModal.ts
```diff
@@ -46,15 +46,17 @@ const getForegroundAppAction = (route: ForegroundAppRoute, params: Partial<Modal
...
+    const discoveryInProgress = useSelector(selectHasRunningDiscovery);
     const isPassphraseFlow =
         Boolean(discoveryForSelectedDevice?.isAddingHiddenWallet) &&
-        discoveryForSelectedDevice?.status !== 'cancelled' &&
-        discoveryForSelectedDevice?.status !== 'complete' &&
-        discoveryForSelectedDevice?.status !== 'failed';
+        discoveryInProgress &&
+        !(
+            discoveryForSelectedDevice?.status === 'progress' &&
+            discoveryForSelectedDevice.hasLoadedAnyNonEmptyAccount
```
Lemonexe, line 58: yeah this is what I would replace with 

`const hasLoadedAnyNonEmptyAccount = useSelector(selectAccountsByDeviceState).some(a => !a.isEmpty)`

vojtatranta: yeah, you still need that flag for imperative call for the selection of the device.

vojtatranta: yeah good idae

suite-common/wallet-core/src/discovery/discoveryThunks.ts
vojtatranta: I like have it transaprent

suite-common/wallet-types/src/discovery.ts
vojtatranta: yeah, likely, thakns

packages/suite/src/hooks/suite/usePreferredModal.ts
vojtatranta: Yeah, so it wont work :-D.

The problem: I don't have the device. The device and it's state is being just loaded in the discovery itself. And deivce is added at the end of the process. 

So, I can't select the device here to get its static session id. I could store this devide session ID at the discovery.

Lemonexe: Right, I did not think of that. This is not the usual discovery, I'll think about it once more if this code can be simplified :thinking:

vojtatranta: we can add that static session id thre, altough I am not sure if this is actually correct thing. But likely is though. YOu can add just one deivce like this

vojtatranta: Naturally, I think that it is a right thing to have a flag like this in the discovery

suite-common/wallet-core/src/discovery/discoveryReducer.ts
Lemonexe: Ok. I would just rewrite it like below.
Type narrowing by status is IMO cleaner than "blindly" checking if the property is there.
_Also_, the type prescribes that `hasLoadedAnyNonEmptyAccount` is present only when `progress`, not afterwards. But the ...spreading is a hack that bends TS to include the property on later statuses, which should not have it anymore.
Now this is not critique at your code, because I just realized that the `update` function does it _`already`_ – it includes `total, progress` at the later statuses too :see_no_evil:
But this means we'll not introduce more of it..
```
    const currentStatus = draft[payload.path];
    const hasLoadedAnyNonEmptyAccount: true | undefined =
        (currentStatus.status === 'progress' && currentStatus.hasLoadedAnyNonEmptyAccount) ||
        (payload.status.status === 'progress' && payload.status.hasLoadedAnyNonEmptyAccount) ||
        undefined;

    draft[payload.path] = {
        ...currentStatus,
        ...payload.status,
        ...{ hasLoadedAnyNonEmptyAccount },
    };
```

Lemonexe (approved):
Well done :+1:
I wondered if we can avoid the newly added complexity of `progressHooks` but did not find anything better.
I tested suite web rebased on develop with the change I proposed and works perfectly, passphrase is much faster now :+1:

Btw will you do the same for mobile passphrase flow?

vojtatranta:
> Well done 👍 I wondered if we can avoid the newly added complexity of `progressHooks` but did not find anything better. I tested suite web rebased on develop with the change I proposed and works perfectly, passphrase is much faster now 👍
> 
> Btw will you do the same for mobile passphrase flow?

I woould rather not touch mobile passphrase :D Can't tell what's going on there

vojtatranta:
> Well done 👍 I wondered if we can avoid the newly added complexity of `progressHooks` but did not find anything better. I tested suite web rebased on develop with the change I proposed and works perfectly, passphrase is much faster now 👍
> 
> Btw will you do the same for mobile passphrase flow?

Yeah, it is broken also on the mobile as I checked now, so, I will fix there too.

vojtatranta:
Hm, this fix won't be enough, the account can be duplicated I gues...

vojtatranta:
We can't do it like this I guess, or there has to be check for the duplicity first before discovering accounts...

vojtatranta:
Yeah, so several problems here.

If I flag the discovery here, I do hide the loading, but if the device is duplicated, it is messed up.

So I am thinking about moving the "duplicity" check before discovering the acounts.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -416,11 +433,52 @@ export const runDiscoveryThunk = createThunk(
             assertDeviceIsAcquired(device);
 
             assertStaticSessionId(deviceStateResponse.payload._state);
+
+            const duplicate = selectDevices(getState())
```
vojtatranta, line 447: this major change: I moved this duplicity check before we even try to load accounts. 

I am thinking if it won't break anything and it shouldn't. Coz the wallet empty check is below and that's find. And we only care about duplicity during adding a passphrase wallet.

Reason of this:
I need to check for the duplicity BEFORE I may ever add the wallet to the state. We wanteed to add the wallet to the state ass soon as just 1 non-empty account is found, so that user doesnt tneed to wait for the app to be interactible



@Lemonexe @mroz22

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -416,11 +433,52 @@ export const runDiscoveryThunk = createThunk(
...
+                        },
+                        device.path,
+                    ),
+                );
+
+                return;
+            }
+
+            // NOTE: this should be true anytime the applyDeviceStatesThunk() is called
+            let deviceStateApplied = false;
```
vojtatranta: Any idea how to improve this? Maybe I could use redux store to determine the duplicity but the code tis more complicated than this...

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -97,6 +97,18 @@ const applyDeviceStatesThunk = createThunk(
...
+                ? selectDeviceByStaticSessionId(getState(), newDeviceState.staticSessionId)
+                : null;
+
+            if (currentDeviceByStaticSessionId && isAddingHiddenWallet) {
+                console.warn(
+                    'applyDeviceStatesThunk: applying state to a device with static session id',
+                );
+
+                return;
+            }
```
Lemonexe, line 110: does this need to be here? Is this something that happens and you need diagnostics? :thinking:

Lemonexe (approved):
Works well, just the one question in comment above.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
vojtatranta: I really , really dont like when there is random 'return' without explanation

Lemonexe: I meant the whole if. But yeah, reading it again, it makes sense!

bosomt:
**QA OK**

works as expected, wallet is available immediately as discovery finds that is not empty

**Info:**
 - Suite version: desktop 25.7.0 (d4c6324599ef00707d8f2662d0143df6bea50c5c)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.7.0 Chrome/134.0.6998.178 Electron/35.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982

PR #19363, opened 2025-06-05
chore(connect): separate Device.handshake and Device.run

szymonlesisz:
## Description

sepatare handshake (run with no `fn`) and run

back to draft, it will not be so easy as i thought


szymonlesisz:
outdated branch, will get back to it some other time

PR #19364, opened 2025-06-05
feat: add BT UI :blue_square: :tooth: 

peter-sanderson:
Bluetooth Suite components for BT implementation.

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=bluetooth_components&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=bluetooth_components&lookback=7)


packages/suite/src/components/suite/PrerequisitesGuide/PrerequisitesGuide.tsx
```diff
@@ -50,11 +52,11 @@ const BottomAnimatedContainer = styled(motion.div)`
     display: flex;
 `;
 
-const Bluetooth = () => (
+const Bluetooth = ({ children }: any) => (
```
Lemonexe: Rename & type pls.
(before it wasn't clear that it is gonna be a style wrapper, but now it is)
```suggestion
const BluetoothWrapper = ({ children }: PropsWithChildren) => (
```

packages/suite/src/components/suite/bluetooth/BluetoothDialogCard.tsx
```diff
@@ -0,0 +1,51 @@
...
+    children,
+    footer,
+    cardHeader,
+    floatingHeader,
+    headerOnClose,
+}: BluetoothDialogCardProps) => {
+    const content = (
+        <Column
+            gap={spacings.md}
+            margin={{ vertical: spacings.xxs, horizontal: spacings.xxs }}
```
Lemonexe: Nit:
```suggestion
            margin={spacings.xxs}
```

packages/suite/src/components/suite/bluetooth/BluetoothDialogCard.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <Column gap={spacings.xxxxl} flex="1" alignItems="stretch">
+            {floatingHeader && (
+                <Text typographyStyle="titleMedium" align="center">
+                    {floatingHeader}
+                </Text>
+            )}
+            <Column gap={spacings.sm} flex="1" alignItems="stretch">
+                <Card paddingType="none">
+                    {/* Hardcoded +2 in Elevation, intentionally, its is a design decision */}
+                    {/*{uiMode === 'spatial' ? <ElevationUp>{content}</ElevationUp> : content}*/}
```
Lemonexe: forgotten code; please check if it needs to be only for `spatial` or for both variants

packages/suite/src/components/suite/bluetooth/BluetoothDialogHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+type BluetoothScanHeaderProps = {
+    children?: ReactNode;
+    onClose?: () => void;
+};
+
+export const BluetoothDialogHeader = ({ children, onClose }: BluetoothScanHeaderProps) => (
+    <Row
+        justifyContent="space-between"
+        margin={{ top: spacings.sm, left: spacings.lg, right: spacings.lg }}
```
Lemonexe: Nit
```suggestion
        margin={{ top: spacings.sm, horizontal: spacings.lg }}
```

packages/suite/src/components/suite/bluetooth/BluetoothDialogCard.tsx
Lemonexe, line 1: :thought_balloon:  note on the `BluetoothDialogCard + BluetoothDialogHeader` as whole: these components are very general, and there is actually nothing strictly BT related.
If it wasn't for the one <Translation cancel>, it could be moved to `product-components` (not components, since this is an arbitrary composition of basic blocks)
But probably not worth it..

packages/suite/src/components/suite/bluetooth/NotTrezorYouAreLookingFor.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export const NotTrezorYouAreLookingFor = ({ onReScanClick }: NotTrezorYouAreLookingForProps) => {
+    const [showTips, setShowTips] = useState(false);
+
+    return (
+        <Collapsible isOpen={showTips}>
+            <Collapsible.Toggle>
+                <Row justifyContent="center" flex="1" margin={{ bottom: spacings.xs }}>
+                    <Row onClick={() => setShowTips(!showTips)} gap={spacings.xs}>
```
Lemonexe:
```suggestion
                    <Row onClick={() => setShowTips(prev => !prev)} gap={spacings.xs}>
```

packages/suite/src/components/suite/bluetooth/NotTrezorYouAreLookingFor.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Collapsible, Flex, Icon, Link, Row, Text } from '@trezor/components';
+import { negativeSpacings, spacings } from '@trezor/theme';
+
+import { BluetoothTips } from './BluetoothTips';
+import { Translation } from '../Translation';
+
+const Toggle = styled.div`
+    transition: opacity 0.15s;
+    cursor: pointer;
+`;
```
Lemonexe: This should probably be a design system component :thinking: 
It's a common pattern.
Or at least a suite component.

packages/suite/src/components/suite/bluetooth/types.ts
```diff
@@ -0,0 +1,12 @@
+import type { DeviceConnectionStatus } from '@trezor/transport-bluetooth';
+
+export type FakeScanStatus = 'running' | 'done';
```
Lemonexe: unused

packages/suite/src/components/suite/bluetooth/types.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type DeviceBluetoothStatus =
+    | DeviceConnectionStatus
+    | {
+          id: string;
+          type: 'found' | 'error';
+      };
+
+export type DeviceBluetoothStatusType = DeviceBluetoothStatus['type'];
```
Lemonexe: unused, but this can stay, at least here it is clear what it is and what it could be useful for :smile:

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    useEffect(() => {
+        dispatch(bluetoothStartScanningThunk());
+
+        return () => {
+            dispatch(bluetoothStopScanningThunk());
+        };
+    }, [dispatch]);
+
+    const clearScamTimer = useCallback(() => {
```
Lemonexe: this looks like a scam :smile:

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -0,0 +1,199 @@
...
+    const [waitingForFirstUpdate, setWaitingForFirstUpdate] = useState<boolean>(true);
+    const [selectedDeviceId, setSelectedDeviceId] = useState<string | null>(null);
+    const [scannerTimerId, setScannerTimerId] = useState<TimerId | null>(null);
+
+    const bluetoothAdapterStatus = useSelector(selectAdapterStatus);
+    const allDevices = useSelector(selectAllDevices);
+    const knownDevices = useSelector(selectKnownDevices);
+    const nearbyDevices = useSelector(selectNearbyDevices);
+
+    const lasUpdatedBoundaryTimestamp =
```
Lemonexe: typo
```suggestion
    const lastUpdatedBoundaryTimestamp =
```

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -0,0 +1,199 @@
...
+            it.lastUpdatedTimestamp < lasUpdatedBoundaryTimestamp;
+
+        if (isDeviceUnresponsiveForTooLong) {
+            // If the device is connected or paired (it may have been paired in the OS system directly)
+            // => do not filter it based isDeviceUnresponsiveForTooLong
+
+            return (
+                it.connected ||
+                it.paired ||
+                knownDevices.find(knownDevice => knownDevice.id === it.id) !== undefined
```
Lemonexe: My favourite nit suggestion
```suggestion
                knownDevices.some(knownDevice => knownDevice.id === it.id)
```

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -0,0 +1,199 @@
...
+    }, [dispatch]);
+
+    const clearScamTimer = useCallback(() => {
+        if (scannerTimerId !== null) {
+            clearTimeout(scannerTimerId);
+        }
+    }, [scannerTimerId]);
+
+    useEffect(() => {
+        // Intentionally no `clearScamTimer`, this is first run and if we use this we would create infinite re-render
```
Lemonexe: well yes, I understand that including `return clearScamTimer` would transitively bring `scannerTimerId` as dependency and loop it. It's always the same problem when working with timers in React, and a
[common workaround](https://stackoverflow.com/questions/53090432/react-hooks-right-way-to-clear-timeouts-and-intervals) that's IMO a valid case for useRef, see headline "If you need to clear timeouts or intervals in another component". 

Cleaning up setTimeout is not as vital as setInterval, but still a good practice..

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -0,0 +1,199 @@
...
+            dispatch(bluetoothActions.scanStatusAction({ status: 'idle' }));
+        }
+    }, [devices, dispatch, clearScamTimer]);
+
+    // When we have some knownDevices, it may happen that we get empty nearbyDevices first and then,
+    // in split second we get the update and there are some nearbyDevices present.
+    // Device in knownDevices but not in the nearbyDevices is recognized as to-be-deleted,
+    // and ui flickers. This small loading is here to hide that.
+    useEffect(() => {
+        setTimeout(() => setWaitingForFirstUpdate(false), 100);
```
Lemonexe: Nit. Yeah I realize chance of unmounting component so soon is negligible, but still a good practice IMO
```suggestion
        const timeoutHandle = setTimeout(() => setWaitingForFirstUpdate(false), 100);
        
        return () => clearTimeout(timeoutHandle); 
```

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -0,0 +1,199 @@
...
+import { bluetoothStartScanningThunk } from '../../../actions/bluetooth/bluetoothStartScanningThunk';
+import { bluetoothStopScanningThunk } from '../../../actions/bluetooth/bluetoothStopScanningThunk';
+import { closeModalApp } from '../../../actions/suite/routerActions';
+import { useDispatch, useSelector } from '../../../hooks/suite';
+
+const SCAN_TIMEOUT = 30_000;
+const UNPAIRED_DEVICES_LAST_UPDATED_LIMIT_SECONDS = 30;
+
+type BluetoothConnectProps = {
+    uiMode: 'spatial' | 'card';
```
Lemonexe: nit: extract the type instead of repeating it in each Component.
I considered a context instead of propdrilling, but that's an overkill.

packages/suite/src/components/suite/bluetooth/BluetoothDebugInfo.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { DesktopBluetoothDevice } from '../../../actions/bluetooth/DesktopBluetoothDevice';
+import { useSelector } from '../../../hooks/suite';
+
+const TimeAgo = ({ timestamp }: { timestamp: number }) => {
+    const [secAgo, setSecAgo] = useState(0);
+
+    useEffect(() => {
+        setSecAgo(Math.floor(Date.now() / 1000 - timestamp));
+        const interval = setInterval(() => setSecAgo(t => t + 1), 1000);
```
Lemonexe, line 14: hmm, now I'm just overnerding it, **feel free to ignore**.

Just a thought exercise how to avoid using useEffect to set `data2` when `data1` changes, instead find a way to write transformation `data1 → data2` 

This is how the transformation could work here:
- Split `secAgo` into two variables `mountedTimestamp + timerSinceMounted`
- `mountedTimestamp` would be a constant `Math.floor(Date.now() / 1000 - timestamp`
- `timerSinceMounted` would be `useState(0)`  and increment in `useEffect` interval :slightly_smiling_face:

Lemonexe (commented):
CR in progress, dumping review comments.
Most of them nits, but consider the refactoring to move the generic components..

packages/suite/src/components/suite/bluetooth/BluetoothDeviceList.tsx
```diff
@@ -0,0 +1,38 @@
...
+        </Column>
+        <SkeletonRectangle width="86px" height="36px" animate />
+    </Row>
+);
+
+type BluetoothDeviceListProps = {
+    deviceList: DesktopBluetoothDevice[];
+    onConnect: (deviceId: string) => Promise<void>;
+    isScanning: boolean;
+    isDisabled: boolean;
```
Lemonexe: unused, remove

packages/suite/src/components/suite/bluetooth/BluetoothDeviceListItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const buttonLabel = labelMap[isConnectingDevice ? 'connecting' : device.connectionStatus.type];
+
+    const isGhostDevice = isKnownDevice && !isNearbyDevice;
+
+    return (
+        <>
+            <Column gap={spacings.xs}>
+                <Row gap={spacings.md} alignItems="center">
+                    <BluetoothDeviceComponent device={device} flex="1" />
+                    {buttonLabel !== null ? (
```
Lemonexe: I think this whole ternary would be good to extract into subcomponent. There is a lot of logic in `BluetoothDeviceListItem` render function that is related just to the buttons, so it'd be easier to read.

Also currently the flow is chaotic: there is a map of actions, but another ternary with `isGhostDevice` that bypasses it → early return for `isGhostDevice` would be clearer.

packages/suite/src/components/suite/bluetooth/BluetoothDeviceListItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+import { BluetoothDeviceComponent } from './BluetoothDeviceComponent';
+import { DesktopBluetoothDevice } from '../../../actions/bluetooth/DesktopBluetoothDevice';
+import { useDispatch, useSelector } from '../../../hooks/suite';
+import { Translation, TranslationKey } from '../Translation';
+
+const labelMap: Record<DeviceBluetoothConnectionStatusType, TranslationKey | null> = {
+    disconnected: 'TR_BLUETOOTH_CONNECT',
+    connecting: 'TR_BLUETOOTH_CONNECTING',
+    connected: null, // Do not offer disconnect, it is confusing to the user as BT device auto-connects anyway.
```
Lemonexe, line 21: Yes, this makes sense!
I remember that when playing around with it, there was some disconnect button somewhere that I clicked and it did precisely that – the device autoconnected anyway, and it did confuse me. Sry don't remember where it was. For consistency, it should probably be hidden everywhere when `'connected'` :thinking:

packages/suite/src/components/suite/bluetooth/BluetoothDeviceListItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+        connected: undefined,
+        connecting: undefined,
+        disconnected: () => onConnect(device.id),
+        paired: undefined,
+        pairing: undefined,
+    };
+
+    const handleOnclick = onClickMap[device.connectionStatus.type];
+
+    const handleOnClick = async () => {
```
Lemonexe: two variables that differ only in casing :-1: 
I would rename the first one (the map) to `onClickAction` for example

packages/suite/src/components/suite/bluetooth/BluetoothDeviceListItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+                <Row gap={spacings.md} alignItems="center">
+                    <BluetoothDeviceComponent device={device} flex="1" />
+                    {buttonLabel !== null ? (
+                        <Row gap={spacings.xs}>
+                            {isGhostDevice ? (
+                                <Button
+                                    variant="warning"
+                                    size="small"
+                                    margin={{ vertical: spacings.xxs }}
+                                    isDisabled={isDisabled || handleOnclick === undefined}
```
Lemonexe: another confusing part: why `handleOnclick === undefined` for the _remove ghost button?_
That condition belongs to the other button, right?

packages/suite/src/components/suite/bluetooth/BluetoothDeviceListItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+                                    onClick={handleDelete}
+                                >
+                                    <Translation id="TR_REMOVE" />
+                                </Button>
+                            ) : (
+                                <Button
+                                    variant="primary"
+                                    size="small"
+                                    margin={{ vertical: spacings.xxs }}
+                                    isDisabled={isDisabled || handleOnclick === undefined}
```
Lemonexe: and here the `isDisabled` is superfluous, because `handleOnclick === undefined` in the same cases, making me conclude that isDisabled is only for the remove ghost button..?
So maybe even further separating the two buttons into subcomponents would help, but not necessary.

packages/suite/src/components/suite/bluetooth/BluetoothDialogCard.tsx
peter-sanderson: Design of the BT flow will (most) probably change. There are some _movements_ in the Engagement team :)

packages/suite/src/components/suite/bluetooth/BluetoothDeviceListItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+        setIsLoading(true);
+        await handleOnclick?.();
+        setIsLoading(false);
+    };
+
+    const handleDelete = () => {
+        dispatch(bluetoothActions.removeKnownDeviceAction({ id: device.id }));
+    };
+
+    const buttonLabel = labelMap[isConnectingDevice ? 'connecting' : device.connectionStatus.type];
```
Lemonexe: is `isConnectingDevice` necessary at all? 
'connecting' is one of the device status types. If it is not `'connecting'`, maybe it should be and there is some incomplete handling in the redux?

packages/suite/src/components/suite/bluetooth/NotTrezorYouAreLookingFor.tsx
peter-sanderson: I will ask sombody from engagement

packages/suite/src/components/suite/bluetooth/BluetoothDeviceListItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    const isGhostDevice = isKnownDevice && !isNearbyDevice;
```
Lemonexe, line 147: this is ok, but if the same logic will be implemented elsewhere, you may consider creating a memoized selector for this (it depends only on other selectors and a `deviceId` param) → make React components as lightweight as possible, but do as per your judgement :shrug:

packages/suite/src/components/suite/bluetooth/types.ts
peter-sanderson: seems to me that whole file is some leftover

packages/suite/src/actions/bluetooth/bluetoothDisconnectDeviceThunk.ts
```diff
@@ -0,0 +1,37 @@
...
+    async ({ id }, { fulfillWithValue, dispatch }) => {
+        const result = await bluetoothIpc.disconnectDevice(id);
+
+        if (!result.success) {
+            dispatch(
+                bluetoothActions.updateDeviceConnectionStatus({
+                    deviceId: id,
+                    connectionStatus: { type: 'disconnected' },
+                }),
+            );
```
Lemonexe, line 25: shouldn't this be done in success too?

Lemonexe (commented):
Wow, that's a lot of work over the months!
Overall looks good, and works :tada:  (some glitches, but well enough, that can be solved later).
I have a lot of comments concerning codestyle, some questions, but nothing really major.

packages/suite/src/components/suite/bluetooth/BluetoothDialogCard.tsx
Lemonexe: ok, if it is subject to change then no need for perfectionism.

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
peter-sanderson: Good point, fixed

packages/suite/src/components/suite/bluetooth/BluetoothDeviceListItem.tsx
peter-sanderson: There are different types of "connecting" `connectingDevicesIds` are those devices that Suite initiated as connecting. Other connecting statuses are usualy data from underlying BT code/OS

peter-sanderson: image

peter-sanderson: I renamed the prop to `isSuiteTryingToConnectToDevice` in the component to make it more clear

packages/suite/src/actions/bluetooth/bluetoothDisconnectDeviceThunk.ts
peter-sanderson: Suite in not autoritative here with state. It takes whatever it gets from BT transport code (os). After successfull diconenction, there shall be an update from BT protocol, tajt device is disconnected.

This here, is actually a hack, where we optimistically disconnect the device, as failure to disconnect the device means, that device is not really working anyway.

packages/suite/src/components/suite/bluetooth/NotTrezorYouAreLookingFor.tsx
peter-sanderson: => https://github.com/trezor/trezor-suite/pull/19403

Done

packages/suite/src/components/suite/bluetooth/BluetoothConnect.tsx
```diff
@@ -0,0 +1,201 @@
...
+        return () => {
+            dispatch(bluetoothStopScanningThunk());
+        };
+    }, [dispatch]);
+
+    const clearScanTimer = useCallback(() => {
+        if (scannerTimerId.current !== null) {
+            clearTimeout(scannerTimerId.current);
+        }
+    }, [scannerTimerId]);
```
Lemonexe: One last thing: remove `scannerTimerId` from dependencies.
As far as I understand it doesn't harm anything, just useless (the ref is the same pointer, so no rerenders when `.current` changes)

Lemonexe (approved):
Thx for implementing all the review, LGTM :+1:

packages/suite/src/components/suite/bluetooth/BluetoothDebugInfo.tsx
vojtatranta: My bigger worry is that this won't work at all. It will drift away after like 2 minutes to complete time mess.

`useInterval` cannot perform fixed `1000s` loops to be called. (I trid make JS clocks and it is complete mess`.

It should reather be done through `requestAnimationFrame` and taking current time and with module adding some time.

But if, of course, it d oesn't need to be precies or the interval is remounted often, it won't be such problem.

peter-sanderson: Please note: It is just a debug info and only relevant for ~30sec when it gets filtered out as dead/disconnected device

Lemonexe: Yes, @vojtatranta I thought of the exact same thing, but considering it's only debug info I don't think it's important.

PR #19369, opened 2025-06-05
fix(suite-native): remove bluetoothIpc dep from common

juriczech:
## Description
- having `@trezor/transport-bluetooth` in suite-common was a mistake since it's purely desktop implementation of bluetooth transport. This PR will remove the dependency and move parts of the code that are desktop specific to suite package while keeping calls that could be useful for both platforms in suite-common.

## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/native/remove-bluetoothipc-dep&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/native/remove-bluetoothipc-dep&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/remove-bluetoothipc-dep&lookback=7)

### DIFF
--- a/suite-common/wallet-core/src/device/deviceThunks.ts
+++ b/suite-common/wallet-core/src/device/deviceThunks.ts
@@ -27,7 +27,6 @@ import TrezorConnect, {
     UI,
 } from '@trezor/connect';
 import { getEnvironment } from '@trezor/env-utils';
-import { bluetoothIpc } from '@trezor/transport-bluetooth';
 import { exhaustive } from '@trezor/type-utils';
 import { isChanged } from '@trezor/utils';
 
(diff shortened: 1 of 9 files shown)


packages/suite/src/middlewares/suite/suiteMiddleware.ts
```diff
@@ -82,6 +85,10 @@ const suite =
             }
         }
 
+        if (bluetoothActions.removeKnownDeviceAction.match(action)) {
```
peter-sanderson: Using middlewares for this is something I believe we shall avoid and gradually remove it from Suite completely (actually we shall remove all middlewares)

suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -553,15 +552,6 @@ export const wipeDeviceThunk = createThunk(
                 dispatch(
                     bluetoothActions.removeKnownDeviceAction({ id: device.bluetoothProps.id }),
                 );
-
-                const resultForget = await bluetoothIpc.forgetDevice(device.bluetoothProps.id);
```
peter-sanderson, line 557: I think we shall call some `extra` dependency here instead. So we can have platform-specific code injected here.

peter-sanderson: => `extra.thunks.forgetBluetoothDevice(...);` here

peter-sanderson (approved):

PR #19341, opened 2025-06-04
feat: THP common update

peter-sanderson:
🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=suite_thp2&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=suite_thp2&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=suite_thp2&lookback=7)

PR #19373, opened 2025-06-05
THP UI :tada: :closed_lock_with_key: 

peter-sanderson:
needs to be merged after:
1. https://github.com/trezor/trezor-suite/pull/19341
2. https://github.com/trezor/trezor-suite/pull/19364

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=final_thp_and_bt_impl&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=final_thp_and_bt_impl&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=final_thp_and_bt_impl&lookback=7)


suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -500,9 +512,11 @@ export const deviceConnectThunks = createThunk<void, DeviceConnectThunksParams,
         switch (type) {
             case DEVICE.CONNECT:
                 dispatch(deviceActions.connectDevice({ device, settings }));
+                dispatch(connectThpDeviceThunk({ device }));
                 break;
             case DEVICE.CONNECT_UNACQUIRED:
                 dispatch(deviceActions.connectUnacquiredDevice({ device, settings }));
+                dispatch(autoInitThpAfterDeviceConnectionThunk({ device }));
```
peter-sanderson, line 519: @Lemonexe This is the problematic code, that has the `Device` vs. `TrezorDevice` issue.

packages/suite/src/actions/thp/removeThpAutoconnectThunk.ts
```diff
@@ -0,0 +1,60 @@
...
+import { thpActions } from '@suite-common/thp';
+import { notificationsActions } from '@suite-common/toast-notifications';
+import { selectSelectedDevice } from '@suite-common/wallet-core';
+// TODO thp-post-fixes
+// import TrezorConnect from '@trezor/connect';
+
+import { THP_PREFIX } from './thpActions';
+
+// TODO thp-post-fixes
+type Fn = (args: { credentials: any[] }) => any; // typeof TrezorConnect.thpRemoveCredentials;
```
peter-sanderson, line 11: This (or something like this) needs to be here until those methods became available in the Connect

suite-common/wallet-core/src/device/deviceThunks.ts
Lemonexe: Side note: uh-oh, circular dependency between `wallet-core/deviceThunks` and the `thp` thunk. To me that means that `autoInitThpAfterDeviceConnectionThunk` belongs to `deviceThunks`, despite being THP-related.

I see the issue. Here we are getting raw `Device` from connect ([see connect init thunks](https://github.com/trezor/trezor-suite/blob/cdd73c596f3dafe967075c335b3aa7ec6ceb9dae/suite-common/connect-init/src/connectInitThunks.ts#L47-L52))
The extra `ExtendedDevice` params that make it a `TrezorDevice` are defined by deviceReducer on `deviceActions.connectDevice | deviceActions.connectUnacquiredDevice`.
That means you should let device reducer do its thing and reselect the device (which now actually **is** a `TrezorDevice`) and call the thp thunk on it.
It's a bit complex but correct without bending TS.

EDIT:  `device.path` is the correct property by which you should reselect it, i.e.
```
extendedDevice = getState().devices.find(d => d.path === device.path)
```
`id` may not be known if it's unacquaried, state session ids are per wallet not device, but `path` is always unique for physical connected device.

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/DeviceContextModal.tsx
```diff
@@ -52,6 +52,11 @@ export const DeviceContextModal = ({
         // T1B1 firmware
         case UI.INVALID_PIN:
             return <PinInvalidModal device={device} />;
+        // Passphrase on host
+        // some leftower on rebase:
+        // case UI.REQUEST_PASSPHRASE:
+        //     return <PassphraseModal device={device} />;
```
Lemonexe: Yes it now lives in `PassphraseFlow.tsx` → `PassphraseModal.tsx`.
Pls confirm if current code after rebase works for passphrase under THP.
If there is a new case for THP that must be handled, it should be added to `PassphraseModal`.

packages/suite/src/components/suite/ConnectDevicePrompt.tsx
```diff
@@ -93,34 +96,53 @@ const getWarningMessage = ({
...
+        'device-disconnect-required': defaultKey,
+        'device-disconnected': defaultKey,
+        'device-initialize': defaultKey,
+        'device-recovery-mode': defaultKey,
+        'device-seedless': defaultKey,
+        'device-unknown': defaultKey,
+        'device-unreadable': defaultKey,
+        'firmware-missing': defaultKey,
+        'firmware-required': defaultKey,
+        'multi-share-backup-in-progress': defaultKey,
```
Lemonexe, line 142: very exhaustive, nice :+1:

packages/suite/src/components/thp/ThpConnectionModal.tsx
```diff
@@ -0,0 +1,25 @@
+import { TrezorDevice } from '@suite-common/suite-types';
+import { thpActions } from '@suite-common/thp';
+
+import { useDispatch } from '../../hooks/suite';
+import { ConfirmActionModal } from '../suite/modals/ReduxModal/DeviceContextModal/ConfirmActionModal';
+
+type PinInvalidModalProps = {
```
Lemonexe: rename

packages/suite/src/components/thp/ThpAutoconnectionModal.tsx
```diff
@@ -0,0 +1,25 @@
+import { TrezorDevice } from '@suite-common/suite-types';
+import { thpActions } from '@suite-common/thp';
+
+import { useDispatch } from '../../hooks/suite';
+import { ConfirmActionModal } from '../suite/modals/ReduxModal/DeviceContextModal/ConfirmActionModal';
+
+type PinInvalidModalProps = {
```
Lemonexe: rename

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/ConfirmActionModal.tsx
```diff
@@ -19,19 +22,25 @@ const ImageWrapper = styled.div`
 
 interface ConfirmActionProps {
     device: TrezorDevice;
+    children?: ReactNode;
```
Lemonexe: looks like `children` are never used; can you remove?

Lemonexe (commented):
CR WIP, just leaving a few comments

packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/DeviceContextModal.tsx
peter-sanderson: Good catch, I will remove this, PW flow seems to work

suite-common/wallet-core/src/device/deviceThunks.ts
peter-sanderson: > To me that means that autoInitThpAfterDeviceConnectionThunk belongs to deviceThunks, despite being THP-related.

:point_up: yeah, probably

peter-sanderson: > should let device reducer do its thing and reselect the device

This sounds like middleware magic, and I really want to avoid it. Lets call later and lets try to entangle this.

packages/suite/src/views/settings/SettingsDebug/ResetThpCredentials.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+        if (result?.success) {
+            // This is a bit of a hack, to force use to reconnect the device. Device still has
+            // the session, but Suite discarded all THP credentials.
+            dispatch(deviceActions.deviceDisconnect(device));
+
+            dispatch(notificationsActions.addToast({ type: 'thp-credentials-reset' }));
+        }
+
+        setTimeout(() => setIsLoading(false), 300);
```
Lemonexe, line 34: I know it's just debug code, but still, why?

packages/suite/src/actions/thp/startThpAutoconnectThunk.ts
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        // TODO thp-post-fixes
+        // const response = await TrezorConnect.thpGetCredentials({ device });
+        const response: any = await Promise.resolve({
+            success: false,
+            payload: { error: 'Implement TrezorConnect.thpGetCredentials' },
+        });
+
+        console.log('response', response);
```
Lemonexe: remove

packages/suite/src/components/firmware/ReconnectDevicePrompt.tsx
```diff
@@ -67,9 +67,10 @@ export const ReconnectDevicePrompt = ({ onClose, onSuccess }: ReconnectDevicePro
     const isWebUsbTransport = useSelector(selectHasTransportOfType('WebUsbTransport'));
     const { showManualReconnectPrompt, status, uiEvent } = useFirmwareInstallation();
     const { device } = useDevice();
+    const eventDevice = uiEvent && 'device' in uiEvent.payload ? uiEvent.payload.device : undefined;
```
Lemonexe: I'd prefer type narrowing to type which is known to have `payload.device` rather than blindly checking if it exists.
```suggestion
    const eventDevice = uiEvent && uiEvent.type === DEVICE.BUTTON ? uiEvent.payload.device : undefined;
```

packages/suite/src/components/suite/layouts/SuiteLayout/DeviceSelector/SidebarDeviceStatus.tsx
```diff
@@ -31,6 +31,7 @@ export const SidebarDeviceStatus = () => {
 
     const handleRefreshClick: MouseEventHandler = e => {
         e.stopPropagation();
+        console.log('SidebarDeviceStatus: handleRefreshClick', deviceNeedsRefresh, selectedDevice);
```
Lemonexe: remove?

packages/suite/src/components/thp/ThpPairingCodeEntry.tsx
```diff
@@ -0,0 +1,58 @@
...
+    | { disabled: true; lastCode: string };
+
+export const ThpPairingCodeEntry = ({ disabled, lastCode }: ThpPairingPinEntryProps) => {
+    const [isLoading, setLoading] = useState(false);
+
+    const dispatch = useDispatch();
+
+    const onCodeEntry = (tag: string) => {
+        setLoading(true);
+        dispatch(thpActions.setLastThpCode({ code: tag }));
```
Lemonexe, line 26: At first I was surprised about the naming – shouldn't it be `setThpCode` , and the state variable `currentThpCode`?
But if I understand correctly, the code is only used onetime during connection, and then you receive credentials. So this is the "last thp code that was used to get credentials"? Makes sense then.

packages/suite/src/components/thp/ThpPairingCodeEntry.tsx
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    return (
+        <Row
+            gap={spacings.xl}
+            margin={{
+                right: isLoading
+                    ? // This is a bit hack, but I think it is better to make it explicitly
+                      // bound to SPINNER_SIZE+gap so it is clear why it is shifted (to prevent jumping)
+                      (-(SPINNER_SIZE + spacings.xl) as unknown as SpacingValues)
```
Lemonexe: 1. no need for unknown, number can be cast as SpacingValues.
2. But anyway, in which component is it needed to prevent jumping? In `ThpPairingCodeEntry` it is ok even without it.
3. Also, why don't you rather display isLoading ? spinner : placeholder with `SPINNER_SIZE` instead of negative margin? Negative margin will work only if there is an element next to it that has margin left which can be subtracted.. if there isn't, there is nothing to subtract.

packages/suite/src/views/firmware/FirmwareModal.tsx
```diff
@@ -72,6 +80,39 @@ export const FirmwareModal = ({
     const trezorCancel = () => TrezorConnect.cancel(intl.formatMessage(messages.TR_CANCELLED));
 
     const getContent = () => {
+        console.log('_____FirmwareModal::status', status);
+        console.log('_____FirmwareModal::thpStep', thpStep);
```
Lemonexe: remove. I wonder why doesn't eslint pick it up? :thinking:

packages/suite/src/views/firmware/FirmwareModal.tsx
```diff
@@ -72,6 +80,39 @@ export const FirmwareModal = ({
...
+                case 'BeforeConnectionInfo':
+                    return <StepThpStart modalHeading={heading} />;
+                case 'ConfirmConnectionBeforePairing':
+                    return device !== undefined ? (
+                        <StepThpPairingRequest modalHeading={heading} />
+                    ) : null;
+                case 'ConfirmOnlyConnection':
+                    return device !== undefined ? (
+                        <StepThpPairingRequest modalHeading={heading} />
+                    ) : null;
```
Lemonexe: these cases can be joined, they have the same code

packages/suite/src/views/onboarding/steps/ThpPairingFailedStep.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { startThpSessionThunk } from '../../../actions/thp/startThpSessionThunk';
+import { ThpPairingFailedForFirmwareInstallation } from '../../../components/thp/ThpPairingFailedForFirmwareInstallation';
+import { useDispatch, useSelector } from '../../../hooks/suite';
+import { selectIsActionAbortable } from '../../../reducers/suite/suiteReducer';
+
+export const ThpPairingFailedStep = () => {
+    const dispatch = useDispatch();
+    const [isLoading, setIsLoading] = useState(false);
+    const isActionAbortable = useSelector(selectIsActionAbortable);
```
Lemonexe, line 17: I didn't even know there was such a thing..
But isn't it always true for thp?
The selector asks if bridge version > 2.0.31 but thp requires bridge > 3

Same applies for other occurrences in THP-only flows, I saw it elsewhere before..

packages/urls/src/urls.ts
```diff
@@ -137,6 +137,7 @@ export const HELP_CENTER_CANCEL_TRANSACTION: Url =
 // TODO: update this link when the article is ready
 export const HELP_CENTER_SOL_SEND: Url =
     'https://trezor.io/learn/a/solana-sol-on-trezor-safe-5-trezor-safe-3-and-trezor-model-t';
+export const HELP_TREZOR_HOST_PROTOCOL: Url = 'https://trezor.io/'; // Todo: add the proper link once we have the article
```
Lemonexe: I don't see it used anywhere, so the placeholder is not needed.

suite-common/wallet-core/src/device/deviceActions.ts
```diff
@@ -122,6 +122,18 @@ const setThpCredentials = createAction(
...
+    `${DEVICE_MODULE_PREFIX}/setDeviceThpAutoConnect`,
+    (payload: { thpAutoConnect: boolean }) => ({ payload }),
+);
+
+const resetThpCredentials = createAction(
+    `${DEVICE_MODULE_PREFIX}/resetThpCredentials`,
+    ({ credentials }: { credentials: ThpSettings['knownCredentials'] }) => ({
+        payload: { credentials },
+    }),
+);
```
Lemonexe: these two are not referenced anywhere but in a middleware, nothing dispatches them, so maybe a leftover?

packages/suite/src/views/onboarding/steps/FirmwareStep.tsx
```diff
@@ -15,18 +19,29 @@ import { useOnboarding, useSelector } from 'src/hooks/suite';
...
 
     const install = () => firmwareUpdate({ firmwareType: targetType });
-    const goToNextStepAndResetReducer = () => {
+    const goToNextStepAndResetReducer = useCallback(() => {
         goToNextStep();
         resetReducer();
-    };
+    }, [goToNextStep, resetReducer]);
+
+    useEffect(() => {
```
Lemonexe, line 40: I see the comment in `suite_thp` branch. I hope you're merge that eventually too, because this would not be clear at all without it

packages/suite/src/views/onboarding/steps/FirmwareStep.tsx
```diff
@@ -107,17 +122,48 @@ export const FirmwareStep = () => {
...
     }
 
+    if (thpStep !== null) {
+        switch (thpStep) {
+            case 'BeforeConnectionInfo':
+                return device !== undefined ? <ThpPairingStartStep /> : null;
+            case 'ConfirmOnlyConnection':
+                return device !== undefined ? <ThpPairingConfirmStep device={device} /> : null;
+            case 'ConfirmConnectionBeforePairing':
+                return device !== undefined ? <ThpPairingConfirmStep device={device} /> : null;
```
Lemonexe: cases return the same code, you can merge them

packages/suite/src/views/onboarding/steps/FirmwareStep.tsx
```diff
@@ -107,17 +122,48 @@ export const FirmwareStep = () => {
...
+
+            // Auto-connect not relevant for Onboarding Firmware Installation.
+            // 1) We don't want to ask user for autoconnect during FW installation.
+            // 2) It shall never happen anyway, onboarding is the 1st connection.
+            case 'AutoconnectInfo':
+            case 'Autoconnect':
+                return null;
+
+            case 'CodeInvalid':
+                return device !== undefined ? <ThpPairingFailedStep /> : null;
```
Lemonexe: can you move the `device !== undefined` condition to the `if (thpStep !== null) {`, so you don't have to repeat it so much?

packages/suite/src/components/suite/PrerequisitesGuide/DeviceTrezorHostProtocolPair.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useDevice, useDispatch } from 'src/hooks/suite';
+
+export const DeviceTrezorHostProtocolPair = () => {
+    const { isLocked, device } = useDevice();
+    const dispatch = useDispatch();
+
+    const isDeviceLocked = isLocked();
+
+    const handleClick: MouseEventHandler = e => {
+        e.stopPropagation();
```
Lemonexe: why? Does the Button have any default actions? Is there any onClick on a whole wrapper component?

suite-common/firmware/src/hooks/useFirmwareInstallation.ts
```diff
@@ -76,46 +76,57 @@ export const useFirmwareInstallation = (
         firmware.uiEvent.payload.method === 'manual';
 
     const showReconnectPrompt =
+        // For some magic reason, the `ButtonRequest_Other` is present in FW after THP pairing;
+        // This causes the UI to show the reconnect prompt, despite we are already done.
```
Lemonexe, line 80: that should probably be fixed in FW, but this condition makes sense to stay here anyway so ok

Lemonexe (commented):
Many comments, but nothing major

suite-common/wallet-core/src/device/deviceThunks.ts
Lemonexe: oh sry, I overlooked the reply. It doesn't have to do anything with  middleware, just a normal flow action → reducer → selector, discussed in person.

peter-sanderson: We solved this on the call will be adressed in later PR

packages/suite/src/views/settings/SettingsDebug/ResetThpCredentials.tsx
peter-sanderson: Just to fake the UI so it looks that something happened.

packages/suite/src/components/thp/ThpPairingCodeEntry.tsx
peter-sanderson: PIN is "centered" when spinner is show, it would "jump" -> this forces the spinner to show on the right but keeps the input in the center

packages/suite/src/views/onboarding/steps/ThpPairingFailedStep.tsx
peter-sanderson: I think I simply kept it from other places where `OnboardingStepBox` is used

packages/suite/src/views/onboarding/steps/FirmwareStep.tsx
peter-sanderson: yes, next PR

packages/suite/src/components/suite/PrerequisitesGuide/DeviceTrezorHostProtocolPair.tsx
peter-sanderson: I am not sure, I removed it and we will see

packages/suite/src/views/settings/SettingsDebug/ResetThpCredentials.tsx
Lemonexe: ok..

packages/suite/src/views/onboarding/steps/ThpPairingFailedStep.tsx
Lemonexe: That does not answer my comment :smile: 
For THP it seems useless as `bridge version > 2.0.31` is always true.

packages/suite/src/components/thp/ThpPairingCodeEntry.tsx
Lemonexe: That explains it, just at some places it is **not** centered, and that confused me because it seemed useless.
I do not have THP device with me to verify atm, but think it was this one:
`packages/suite/src/components/suite/modals/ReduxModal/DeviceContextModal/ThpPairingPinEntryModal.tsx`

Definitely it'd help to explain the comment "what kind of margin are we cutting from", that it's the margin: auto created by centering from a parent component.
Basically, this component relies on styles defined by parent, and that should be explained..

Lemonexe (approved):
LGTM, no blockers for merging it :rocket: 
I would like this addressed ([one](https://github.com/trezor/trezor-suite/pull/19373#discussion_r2132010585), [two](https://github.com/trezor/trezor-suite/pull/19373#discussion_r2140194341), [three](https://github.com/trezor/trezor-suite/pull/19373#discussion_r2142891890)) but can be done in the followup PR, no problem :slightly_smiling_face:

PR #19377, opened 2025-06-05
Fix Select scroll to selected option

### LINKED ISSUE #18749: Select - scroll to selected item
## Description

After selecting specific option from `Select` component, scroll to the selected option/option's group header.

TomasBoda:
## Description

Important changes:
- network and token info for buy/sell/swap are loaded independently
- while the options are being loaded, select is disabled and loader is displayed
- selected option is persisted and scrolled to upon opening the select dropdown

## Related Issue

Resolve #18749 

## Screenshots:

https://github.com/user-attachments/assets/b03e089e-500f-4bcf-a19e-bde96994b7ba

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/select&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/select&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/select&lookback=7)

### DIFF
--- a/packages/components/src/components/form/Select/Select.tsx
+++ b/packages/components/src/components/form/Select/Select.tsx
@@ -1,18 +1,5 @@
-import {
-    ComponentProps,
-    ReactNode,
-    useCallback,
-    useEffect,
-    useMemo,
-    useRef,
-    useState,
-} from 'react';
-import ReactSelect, {
-    Props as ReactSelectProps,
-    SelectInstance,
-    StylesConfig,
-    components as reactSelectComponents,
-} from 'react-select';
+import { ReactNode, useCallback, useMemo, useRef, useState } from 'react';
+import ReactSelect, { Props as ReactSelectProps, SelectInstance, StylesConfig } from 'react-select';
 
 import styled, { CSSObject, DefaultTheme, css, useTheme } from 'styled-components';
 
@@ -26,6 +13,11 @@ import {
     zIndices,
 } from '@trezor/theme';
 
+import { FrameProps } from '../../../utils/frameProps';
+import { TransientProps } from '../../../utils/transientProps';
+import { useElevation } from '../../ElevationContext/ElevationContext';
+import { DROPDOWN_MENU, menuStyle } from '../../Menu/menuStyle';
+import { Spinner } from '../../loaders/Spinner/Spinner';
 import {
     FormCell,
     FormCellProps,
@@ -43,11 +35,6 @@ import {
 } from './customComponents';
 import { useDetectPortalTarget } from './useDetectPortalTarget';
 import { useOnKeyDown } from './useOnKeyDown';
-import { FrameProps } from '../../../utils/frameProps';
-import { TransientProps } from '../../../utils/transientProps';
-import { useElevation } from '../../ElevationContext/ElevationContext';
-import { DROPDOWN_MENU, menuStyle } from '../../Menu/menuStyle';
-import { Spinner } from '../../loaders/Spinner/Spinner';
 
 const reactSelectClassNamePrefix = 'react-select';
 
@@ -303,25 +290,6 @@ export type SelectProps = KeyPressScrollProps &
         'data-testid'?: string;
     };
 
-type MenuProps = ComponentProps<typeof reactSelectComponents.Menu> & {
-    selectRef: React.RefObject<SelectInstance<Option, boolean>>;
-    scrollIndex?: number;
-};
-
-const Menu = ({ selectRef, scrollIndex, ...rest }: MenuProps) => {
-    useEffect(() => {
-        if (scrollIndex) {
-            const menuList = selectRef.current?.menuListRef;
-            const option = menuList?.children[scrollIndex];
-            option?.scrollIntoView({ behavior: 'instant' });
-        }
-    }, [selectRef, scrollIndex]);
-
-    return <reactSelectComponents.Menu {...rest} />;
-};
-
-const OPTION_INDEX_NAME = 'optionIndex';
-
 export const Select = ({
     isClean = false,
     label,
@@ -346,34 +314,14 @@ export const Select = ({
     const formCellProps = pickFormCellProps(rest);
     const { isDisabled } = formCellProps;
     const isRenderedInModal = menuPortalTarget !== null;
-    const [scrollIndex, setScrollIndex] = useState<number>(0);
 
-    const isGrouped =
-        Array.isArray(rest.options) && rest.options.some(opt => 'label' in opt && 'options' in opt);
-
-    const indexedOptions = useMemo(() => {
-        if (isGrouped) {
-            return rest.options?.map((group, index) => ({
-                ...group,
-                options: group.options.map((option: Option) => ({
-                    ...option,
-                    [OPTION_INDEX_NAME]: index,
-                })),
-            }));
-        }
-
-        return rest.options?.map((option, index) => ({ ...option, [OPTION_INDEX_NAME]: index }));
-    }, [rest.options, isGrouped]);
+    const [selectedOption, setSelectedOption] = useState<Option | undefined>(rest.value);
 
     const handleOnChange = useCallback<Required<ReactSelectProps<Option>>['onChange']>(
         (value, { action }) => {
             if (value) {
-                const optionIndex = value[OPTION_INDEX_NAME] as number;
-                const option = { ...value };
-                delete option[OPTION_INDEX_NAME];
-
-                onChange?.(option as Option, selectRef.current);
-                setScrollIndex(optionIndex);
+                onChange?.(value, selectRef.current);
+                setSelectedOption(value);
 
                 if (!isMenuOpen && action === 'select-option') {
                     selectRef.current?.blur();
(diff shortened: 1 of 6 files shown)


packages/suite/src/views/wallet/trading/common/TradingForm/TradingFormInput/TradingFormInputAccount.tsx
```diff
@@ -37,6 +38,11 @@ export const TradingFormInputAccount = <
     } = useTradingFormContext<TradingTradeSellExchangeType>();
     const optionGroups = useTradingBuildAccountGroups(type);
 
+    const exchangeInfo = useSelector(state => state.wallet.tradingNew.exchange.exchangeInfo);
+    const sellInfo = useSelector(state => state.wallet.tradingNew.sell.sellInfo);
```
tomasklim: there should be `isLoading` in `tradingNew`. Can it be used or not?

tomasklim: because in case fetch fails, it will be stuck in loading state

suite-common/trading/src/thunks/common/__tests__/loadInitialDataThunk.test.ts
```diff
@@ -213,51 +181,6 @@ describe('loadInitialDataThunk', () => {
         await testUpdatedInfoData('outdated');
     });
 
-    it('should keep same version of data without update', async () => {
```
tomasklim: are there test covering the logic after changes?

suite-common/trading/src/thunks/common/loadInitialDataThunk.ts
```diff
@@ -42,7 +42,7 @@ export const loadInitialDataThunk = createThunk(
 
         dispatch(tradingActions.setTradingActiveSection(activeSection));
 
-        if (account && !isLoading && (isDifferentAccount || areDataOutdated)) {
+        if (account && !isLoading) {
```
adderpositive: `invityAPI.getInfo()` will be called every time, regardless of whether the data is outdated or is for a different account.

suite-common/trading/src/thunks/common/loadInitialDataThunk.ts
```diff
@@ -58,19 +58,22 @@ export const loadInitialDataThunk = createThunk(
                 dispatch(tradingActions.saveInfo(info));
             }
 
-            if (isDifferentAccount || !buyInfo) {
+            if (activeSection === 'buy' && (isDifferentAccount || !buyInfo || areDataOutdated)) {
```
adderpositive: `activeSection === 'buy'` should not be here. I got your point, but it will not work for another part of the suite, when we need to get data regardless of the `activeSection`, for example, `TokenList` page - `Swap` will be unavailable unless you visit Swap in the trading before.

image

packages/components/src/components/form/Select/customComponents.tsx
```diff
@@ -22,23 +25,37 @@ export const Control = ({ 'data-testid': dataTest, ...controlProps }: ControlCom
...
-                }`,
-            } as OptionProps<OptionType, boolean>['innerProps']
+export const Option = ({
+    selectedOption,
+    'data-testid': dataTest,
+    ...props
+}: OptionComponentProps) => {
+    const ref = useRef<HTMLDivElement>();
+
+    useEffect(() => {
```
adderpositive: `useDeepCompareEffect` should solve the problem with creating a new `deepEqual` function.

packages/suite/src/views/wallet/trading/common/TradingForm/TradingFormInput/TradingFormInputAccount.tsx
adderpositive: For `exchangeInfo`, `sellInfo`, and `buyInfo` are prepared selectors, which transform values correctly.

`selectTradingExchangeInfo`
`selectTradingSellInfo`
`selectTradingBuyInfo`

Please use them if they are necessary.

adderpositive: For loading info from the API (common info, buyInfo, sellInfo, exchangeInfo) serves `isLoading` at the top of the trading root. 
You can use the `selectTradingLoadingAndTimestamp` selector for that.

There is also `isLoading` in the sub-reducers (buy, sell, exchange) used for loading quotes.

packages/components/src/components/form/Select/Select.tsx
```diff
@@ -346,34 +314,14 @@ export const Select = ({
...
-                options: group.options.map((option: Option) => ({
-                    ...option,
-                    [OPTION_INDEX_NAME]: index,
-                })),
-            }));
-        }
 
-        return rest.options?.map((option, index) => ({ ...option, [OPTION_INDEX_NAME]: index }));
-    }, [rest.options, isGrouped]);
+    const [selectedOption, setSelectedOption] = useState<any>(rest.value);
```
adderpositive: `any`? Can we use `Option | undefined` or something like that?

suite-common/trading/src/thunks/common/__tests__/loadInitialDataThunk.test.ts
TomasBoda: added tests for the separate loading of buy/sell/exchange sections

packages/suite/src/views/wallet/trading/common/TradingForm/TradingFormInput/TradingFormInputAccount.tsx
TomasBoda: yes it can, changed it

adderpositive (changes requested):
1. UX - It would be nice to also see under what account it belongs to. What do you think? But I am not sure if there will be, for example, 20 coins under one account…
image

2. Sometimes the scrolling is jumping, check the video.
https://github.com/user-attachments/assets/83b2e468-3410-400c-9bc9-350f883203be

3. I don't understand changes in trading. 🤷‍♂️ Why the changes?

suite-common/trading/src/thunks/common/loadInitialDataThunk.ts
TomasBoda: you're right, thanks for the catch. It doesn't make sense to load `activeSection` only then, I reverted it back to as it was before

packages/components/src/components/form/Select/customComponents.tsx
TomasBoda: It doesn't work, it scrolls to the very bottom of the dropdown everytime for some reason. I kept the `deepEqual`

packages/suite/src/views/wallet/trading/common/TradingForm/TradingFormInput/TradingFormInputAccount.tsx
TomasBoda: I am now using the `isLoading` state for this

packages/components/src/components/form/Select/Select.tsx
TomasBoda: done

TomasBoda:
> 1. UX - It would be nice to also see under what account it belongs to. What do you think? But I am not sure if there will be, for example, 20 coins under one account…
>    image
> 2. Sometimes the scrolling is jumping, check the video.
>    https://github.com/user-attachments/assets/83b2e468-3410-400c-9bc9-350f883203be
> 3. I don't understand changes in trading. 🤷‍♂️ Why the changes?

1. This was my intention from the start, but as token info is loading for a few tens of seconds after the select is ready, it refreshes the options and makes it difficult to handle scrolling in the select. The current solution is the only one that works at least somehow consistently
2. I fixed the jumping to some extent
3. The changes were intended to load the sections separately, but as you pointed out, it causes problems, so I reverted it for the time being

suite-common/trading/src/thunks/common/__tests__/loadInitialDataThunk.test.ts
TomasBoda: I reverted the separate loading of sections, because it caused issues in other parts of Suite. Now the select works even though they are loaded all at once, so the tests stay the same as they were

packages/components/src/components/form/Select/customComponents.tsx
adderpositive: Maybe `useDeepCompareEffect` is for a bit different use case. Sorry for misleading.

packages/components/src/components/form/Select/customComponents.tsx
```diff
@@ -22,23 +25,38 @@ export const Control = ({ 'data-testid': dataTest, ...controlProps }: ControlCom
...
```
adderpositive: Replaced with `useDeepCompareEffect`, it should prevent jumping in the select list

adderpositive (changes requested):
Thank you for the updates. Select list still jumps a bit.

adderpositive (approved):

PR #19396, opened 2025-06-06
feat(suite)/coins filter improvement

### LINKED ISSUE #19185: Coins filter improvements stage 2
### 1. Show active indicator in filter icon

If a filter is active (i.e. a coin/network is selected), show the active indicator in the filter icon. Remove the indicator when the filter is either manually or automatically disabled. Design: [Figma](https://www.figma.com/design/Y2E8dcdNYrSMGIVRjYFXKp/-Suite--Components-1.0?m=auto&node-id=7565-27290&t=5VtNqDkT1Oet7I2K-1)

### 2. Reset filter on network change

If the filter is active and the user navigates to a different network (e.g. via a dashboard tile or BTC pay link), automatically reset the filter.

Image

### 3. Bug: List view still applies filter

When the dashboard is set to list view (as opposed to tiles), the filter is still applied after selecting a coin/network from the list. This should be disabled—same as it already is for the tile view.

https://github.com/user-attachments/assets/7ad04131-2493-4723-ae69-f19a27c78e43

OriginalEveres:
Added active state icon that indicates active coin filter, resets coin filter if user changes network that is not within the filter. Removed reset coin filter after selecting network in AssetRow (so it matches the selection in AssetCard).

## Related Issue

Resolves https://github.com/trezor/trezor-suite/issues/19185

### Reset filter on network change
Before
447884447-9373f643-b352-40ec-b52c-fccee7cead9c
Now
output
### Bug: List view still applies filter
Before
output
output

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/coins-filter-improvement&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/coins-filter-improvement&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/coins-filter-improvement&lookback=7)


suite-common/icons/src/icons.ts
```diff
@@ -1186,6 +1186,7 @@ export const icons = {
     funnel: require('../assets/funnel.svg'),
     funnelFilled: require('../assets/funnelFilled.svg'),
     funnelSimple: require('../assets/funnelSimple.svg'),
+    funnelSimpleActive: require('../assets/funnelSimpleActive.svg'),
```
OriginalEveres: I have added the active state of the icon as a new icon since the green indicator seems not to be used anywhere else - so I did not see a reason to create/edit current  Icon component with indicator support

peter-sanderson: cc @jvaclavik :point_up: I am not sure how "legal" :grin: it is to add and icon. :pray: lets wait for Jan comment on this. I think there is some process for new icons.

peter-sanderson: Anyway, is the only difference that the icon just adds an :green_circle: into it?

peter-sanderson: I would maybe consider simply position-absolute the :green_circle: into an existing icon. WDYT @jvaclavik ?

jvaclavik: Sorry, it's not legal to add a new icon 😅  please use icon + element with position absolute. It's not ideal solution, but we don't have a better one right now.

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/CoinsFilter.tsx
```diff
@@ -47,6 +50,22 @@ export const CoinsFilter = () => {
     const { coinFilter, setCoinFilter } = useAccountSearch();
     const availableNetworksSymbols = useAvailableNetworkSymbols();
 
+    const selectedAccount = useSelector(state => state.wallet.selectedAccount);
+
+    useEffect(() => {
```
peter-sanderson: I am really not a fan of using the effects. Is there a way ho to do this in some thunk instead. 

I can imagine, that selecting an account is some action/thunk and this filter-reset (I believe) belongs there.

suite-common/icons/src/icons.ts
jvaclavik: You also might use `ComponentWithSubIcon` if we have icon for circle, but it's more complicated I think.

jvaclavik: In general we can't use icons with multiple colors in one icon and also we don't use stroke in SVG icon files. It's strictly limited to predefined set of icons, because we had a lot of issues with them when we had icon anarchy before :D

OriginalEveres: understandable 😁 I'll make the indicator as a separate element

packages/suite/src/components/wallet/WalletLayout/AccountsMenu/CoinsFilter.tsx
OriginalEveres: fyi I have moved it to `selectedAccountActions.ts`

peter-sanderson (approved):

peter-sanderson (approved):

PR #19404, opened 2025-06-06
refactor(connect): split Device related code to workflow modules

szymonlesisz:
## Description

We need to start untightening this knot :) ~100 lines moved to a separate file

after irl discussion with @marekrjpolak  im creating `device/workflow` modules where we should move all the async code related to method (iframe-call) workflow /lifecycle like `initDevice`, `hasUnexpectedMode`, `checkPermissions`, `firmwareException` etc.

ideally each workflow should have access to some top level context and use abort signal to interrupt async actions.

i started to mark those places with `add-abort-signal` comment next to that line

im starting from the middle with `validateState` because this part will be modified with upcomming THP implementation

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/connect-workflow&lookback=7)


marekrjpolak (approved):
It's cleaning both Device and core/index at once! :heart_hands:

Issue #18392, opened 2025-04-15
Improve primary passphrase entry abandon flow


Hermez-cz:
Please improve the flow where user has the passphrase wallet set as primary to load, connects device, passphrase inputs shows up and the user chooses to close it.

Currently, the input is closed and wildly reappears. Ideally it should take the user one step higher in the hierarchy and let them choose between opening a Standard or Passphrase wallet. When they choose to close even that, it should ideally be respected and Suite should stay open "empty" without any wallet loading. Perhaps the user wanted to go to settings first and will revisit wallet switcher later.

![Image](https://github.com/user-attachments/assets/c5284d41-5e19-492f-92e0-4d5865a92be9)

![Image](https://github.com/user-attachments/assets/a06d5662-8004-43cc-81d9-25ca55a37e65)

Issue #15411, opened 2024-11-15
Wallet selector reopening repeatedly when not entering passphrase and only closing it 


bosomt:
**Describe the bug**
Wallet selector reopening repeatedly when not entering passphrase and only closing it 
See video.

**Steps to reproduce:**
1. Connect device wit Suite in 'passphrase' mode
2. Do not enter passphrase or de not select empty passphrase
3. Selector. is opening over and over again

**Info:**
 - Suite version: desktop 24.11.1 (7019378002e060194a5a877e77b398b8eae86390)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/24.11.1 Chrome/128.0.6613.162 Electron/32.1.2 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T2B1 2.8.2 regular (revision 9e18e72a2c7896cbcce1df4c2d908eab0f01564f)
 - Transport: BridgeTransport 2.0.33

**Screenshots:**

https://github.com/user-attachments/assets/d2232513-7f7a-40d3-98a3-9a0c07d209c5

vojtatranta:
The correct behavior, in my opinion (correct me if I am wrong) is that this passphrase modal sticks open until some passphrase is entered. As "passphrase" mode is meant that the default wallet is a passphrase wallet not a standard one. 

And that is working as off `26f46e39451b8331ed57f7c14136e3ac43fbbacc`

honzauher:
**QA OK**

Working as expected

**Info:**
 - Suite version: web 25.8.0 (89ea587116f51adea4ddb41e216e464df0cf4a39)
 - Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36
 - OS: Win32
 - Screen: 1920x1080
 - Device: Trezor T3T1 2.8.10 universal (revision 2a65d18200580005dc419b9569ed97fae440806a)
 - Transport: WebUsbTransport

Issue #19426, opened 2025-06-09
wallet default is not respected


mroz22:
this settings appears to do nothing

<img width="777" alt="Image" src="https://github.com/user-attachments/assets/82aa9d49-a2a0-4a68-a218-d1e3eca183e5" />

this does not make sense in device reducer and should be moved to discovery thunks https://github.com/trezor/trezor-suite/blob/develop/suite-common/wallet-core/src/device/deviceReducer.ts#L97

bosomt:
**QA OK**

**Info:**
 - Suite version: web 25.7.0 (be7ef183703a164f4036fe0f2d590a82a2662091)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3B1 2.9.0 regular (revision 5e4c97f98d2e139aa82135a0ee43003816f62c6c)
 - Transport: WebUsbTransport

PR #19427, opened 2025-06-09
fix(suite): take wallet type loading settings into account

mroz22:
I'm contemplating #19426, but I don't know if this is a good solution from a UX perspective. I guess @vojtatranta should take over this 

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=discovery-wallet-loading-settings&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=discovery-wallet-loading-settings&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=discovery-wallet-loading-settings&lookback=7)

### DIFF
--- a/packages/suite/src/utils/suite/router.ts
+++ b/packages/suite/src/utils/suite/router.ts
@@ -68,15 +68,38 @@ const validateWalletParams = (url: string): CommonWalletParams => {
     };
 };
 
-const validateModalAppParams = (url: string) => {
+const parseParamValue = <T>(value: string, defaultValue?: T) => {
+    if (value === 'true') return true;
+    if (value === 'false') return false;
+
+    return value ?? defaultValue;
+};
+
+export type ModalAppParams = {
+    cancelable: boolean;
+    isAddingHiddenWalletWithRespectToSettings?: boolean;
+    variant?: string;
+};
+
+const validateModalAppParams = (url: string, params?: Route['params']): ModalAppParams | null => {
     const [, hash] = stripPrefixedURL(url).split('#');
-    if (!hash) return;
-    const [cancelable] = hash.split('/').filter(p => p.length > 0);
-    if (cancelable !== 'true') return;
+    if (!hash) return null;
+    const splitted = hash.split('/').filter(p => p.length > 0);
+    if (splitted.length === 0) return null;
+
+    const defaults: ModalAppParams = { cancelable: true };
+
+    if (params === undefined) return defaults;
 
     return {
-        cancelable: true,
-    };
+        ...defaults,
+        ...Object.fromEntries(
+            params.map((param, index) => [
+                param,
+                parseParamValue(splitted[index], defaults[param as keyof ModalAppParams]),
+            ]),
+        ),
+    } as ModalAppParams;
 };
 
 // Used in routerReducer
@@ -102,7 +125,7 @@ export const getAppWithParams = (url: string): RouterAppWithParams => {
     if (route.params) {
         return {
             app: route.app,
-            params: validateModalAppParams(url),
+            params: validateModalAppParams(url, route.params),
             route,
         } as RouterAppWithParams;
     }
@@ -115,8 +138,9 @@ export const getAppWithParams = (url: string): RouterAppWithParams => {
 };
 
 export type WalletParams = CommonWalletParams;
-export type ModalAppParams = NonNullable<ReturnType<typeof validateModalAppParams>>;
-export type RouteParams = WalletParams | ModalAppParams;
+export type RouteParams = (WalletParams | ModalAppParams) & {
+    isAddingHiddenWalletWithRespectToSettings?: boolean;
+};
 
 export const getRoute = (name: Route['name'], params?: RouteParams) => {
     const route = findRouteByName(name);
--- a/suite-common/wallet-core/src/discovery/discoveryThunks.ts
+++ b/suite-common/wallet-core/src/discovery/discoveryThunks.ts
@@ -668,16 +668,17 @@ export const startDiscoveryThunk = createThunk(
         {
             device,
             isAddingHiddenWallet,
+            isAddingHiddenWalletWithRespectToSettings,
             isAddingExistingWallet,
         }: {
             device?: TrezorDevice;
             isAddingHiddenWallet?: boolean;
+            isAddingHiddenWalletWithRespectToSettings?: boolean;
             isAddingExistingWallet?: boolean;
         },
         { dispatch, getState },
     ): void => {
         const selectedDevice = selectSelectedDevice(getState());
-
         const actualDevice = device ?? selectedDevice;
 
         if (!actualDevice) {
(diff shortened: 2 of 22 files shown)


vojtatranta:
Taking over

Hermez-cz:
Quick spec:

- Setting has to be respected when no wallet is loaded (= from the welcome screen or after onboarding; there's different logic already in place when some wallet is already loaded)
- When the user reaches the point where a wallet should be loaded, then if this is set to Passphrase", wallet switcher is opened and Passphrase entry is offered ([New passphrase] / [Open previously used] buttons)
- If the user hits the back button, the [Standard wallet] / [Passphrase wallet] choice is shown
image
- The (^) close wallet switcher button should not be shown in this case, because there would be nothing to show or load.
    - That means this should be fixed: https://github.com/trezor/trezor-suite/issues/15411
    - And this resolved: https://github.com/trezor/trezor-suite/issues/18392

packages/suite/src/utils/suite/router.ts
```diff
@@ -116,7 +123,9 @@ export const getAppWithParams = (url: string): RouterAppWithParams => {
 
 export type WalletParams = CommonWalletParams;
 export type ModalAppParams = NonNullable<ReturnType<typeof validateModalAppParams>>;
```
Lemonexe: this does not make sense anymore, after refactoring `validateModalAppParams` it returns `{[p: string]: any}` (return type of Object.fromEntries)

Please create an explicit declaration for `ModalAppParams`, because this completley taints the type declaration..

packages/suite/src/utils/suite/router.ts
```diff
@@ -68,15 +68,22 @@ const validateWalletParams = (url: string): CommonWalletParams => {
...
+    return value;
+};
+
+const validateModalAppParams = (url: string, params?: Route['params']) => {
     const [, hash] = stripPrefixedURL(url).split('#');
     if (!hash) return;
-    const [cancelable] = hash.split('/').filter(p => p.length > 0);
-    if (cancelable !== 'true') return;
+    const splitted = hash.split('/').filter(p => p.length > 0);
+    if (splitted.length === 0) return;
```
Lemonexe: L85 can access out of bounds index from `splitted`, returning undefined, and idk how that will behave..
This will handle it gracefully I think
```suggestion
    if (params === undefined) return;
    if (splitted.length < params?.length) return;
```

packages/suite/src/utils/suite/router.ts
```diff
@@ -68,15 +68,22 @@ const validateWalletParams = (url: string): CommonWalletParams => {
...
 
-    return {
-        cancelable: true,
-    };
+    return Object.fromEntries(
+        params?.map((param, index) => [param, parseParamValue(splitted[index])]) ?? [],
```
Lemonexe: following comment :point_up:  you can remove `?.` and `?? []`

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -697,23 +698,44 @@ export const startDiscoveryThunk = createThunk(
...
             dispatch(runDiscoveryThunk(actualDevice));
         }
     },
 );
 
+export const startInitialDiscovery = createThunk(
+    `${DISCOVERY_MODULE_PREFIX}/startInitial`,
+    ({ device }: { device: TrezorDevice }, { dispatch, getState, extra }) => {
+        const isAddingHiddenWalletWithRespectToSettings =
+            extra.selectors.selectSuiteSettings(getState()).defaultWalletLoading === 'passphrase';
```
Lemonexe, line 728: I don't like using extra when it can be avoided, and this can be easily replaced by parametriation.
This new thunk is called only in `discoveryMiddleware`, nowhere else, so let's select `.defaultWalletLoading === 'passphrase'` there and pass it.
After all, what if you need to implement the same in mobile later? You'd have to create `selectSuiteSettings` extra for mobile (currently a suite-only reducer).

suite-common/wallet-core/src/discovery/passphraseUtils.ts
```diff
@@ -17,6 +17,11 @@ export const determinePassphraseFlowState = (
         (!modalState.modalContextDevice || modalState.context === modalState.modalContextDevice) &&
         modalState.windowType === UI.REQUEST_PASSPHRASE
     );
+
+    const isAddingHiddenWalletWithRespectToSettings = Boolean(
+        discovery.isAddingHiddenWalletWithRespectToSettings,
+    );
```
Lemonexe: I'd prefer explicit. It's really a nit, but the principle I follow is: handling falsy as false makes sense, but future refactorings can make unwanted truthy values silently handled as true (while with === TS will give error)
```suggestion
    const isAddingHiddenWalletWithRespectToSettings =
        discovery.isAddingHiddenWalletWithRespectToSettings === true;
```

Lemonexe (changes requested):
generally looks good, some comments, mostly codestyle, but bug:

SwitchDevice modal leaves you stuck in uncancelable state, can't close this :grimacing:
(no passphrase setting, just normally loaded dashboard page with connected device)
image


Though the default passphrase flow itself, that works well!

vojtatranta:
> generally looks good, some comments, mostly codestyle, but bug:
> 
> SwitchDevice modal leaves you stuck in uncancelable state, can't close this 😬 (no passphrase setting, just normally loaded dashboard page with connected device) image
> 
> Though the default passphrase flow itself, that works well!

yeah, I jsut fixed it

suite-common/wallet-core/src/discovery/discoveryThunks.ts
vojtatranta: ok

packages/suite/src/utils/suite/router.ts
vojtatranta: cancelable can be there

suite-common/wallet-core/src/discovery/discoveryThunks.ts
vojtatranta: yeah, I had it there and removed it, coz I didn't want to polute horrible middleware. I intentiaonlyl called "initial discovery" there to have it explicit. So I'll leave it like so. I though you want me to move this prop to the thunk. But I have it there already.

packages/suite/src/utils/suite/router.ts
vojtatranta: done

Hermez-cz (approved):

Lemonexe (approved):
Ok, it works properly now! :+1:
I did some minor CR changes in fixup f0cb1b688ca486ddc280c51b92e438a0774c2307
and will merge it

bosomt:
**QA OK**

https://github.com/trezor/trezor-suite/issues/19426#issuecomment-3016461771

**Info:**
 - Suite version: web 25.7.0 (be7ef183703a164f4036fe0f2d590a82a2662091)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3B1 2.9.0 regular (revision 5e4c97f98d2e139aa82135a0ee43003816f62c6c)
 - Transport: WebUsbTransport

PR #19406, opened 2025-06-06
refactor(suite-native): run discovery from middleware

PeKne:
## Description
- removes most of the discovery triggers from the components code
- discovery i now triggered from DiscoveryMiddleware based on incoming redux actions
- 💡 There are still some discovery thunks called from the passphrase wallet adding flow, it might be revisited later to make this refactor as transparent as possible

## Related Issue

Resolve #19351 
Resolve #19350 

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=refactor/native/discovery-start-from-middleware&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=refactor/native/discovery-start-from-middleware&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=refactor/native/discovery-start-from-middleware&lookback=7)

PR #19452, opened 2025-06-09
Fix: add standard wallet

### LINKED ISSUE #19408: Discovery refactor - can't discover Standard if only Passhprase wallet was rememberd
**Steps to reproduce**

1. Only passphrase wallet in view-only, disable view-only for Standard wallet 
1. Restart the app
1. Connect trezor
1. Try to add Standard wallet

End on this never ending loading screen:

Image

Lemonexe:
## Description

Fix several errors that arise when adding a standard wallet on mobile (see issue incl. comments).

Note: you never have to add the standard wallet when connecting a non-remembered device,  but when you remember only a passhprase, and connect device, the passphrase gets selected & discovered, so you may add a standard one.

## Related Issue

Resolve #19408

## Screenshots:

See video, I start on a passhprase wallet w/ viewonly, disable viewonly for standard, **then I reload the app** to forget the standard and readd it. Switching back and forth works. 
[Screencast from 2025-06-09 18-04-44.webm](https://github.com/user-attachments/assets/88f73fc2-882d-4ede-9444-7795aac4b3f4)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/add-standard-wallet&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/add-standard-wallet&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/add-standard-wallet&lookback=7)

### DIFF
--- a/suite-common/wallet-core/src/device/deviceReducer.ts
+++ b/suite-common/wallet-core/src/device/deviceReducer.ts
@@ -306,14 +306,19 @@ const setDeviceState = (
     if (!device.features) return;
 
     // find devices with the same "device_id"
-    const affectedDevice = draft.devices.filter(
-        d =>
-            d.features &&
-            ((d.connected &&
-                (d.id === device.id || (d.path.length > 0 && d.path === device.path))) ||
-                // update "disconnected" remembered devices if in bootloader mode
-                (d.mode === 'bootloader' && d.remember && d.id === device.id)),
-    ) as AcquiredDevice[];
+    const affectedDevice = draft.devices.filter(d => {
+        if (!d.features) return false;
+
+        const isConnectedDeviceMatch =
+            d.connected &&
+            d.instance === device.instance &&
+            (d.id === device.id || (d.path.length > 0 && d.path === device.path));
+
+        // update "disconnected" remembered devices if in bootloader mode
+        const isRememberedDeviceMatch = d.mode === 'bootloader' && d.remember && d.id === device.id;
+
+        return isConnectedDeviceMatch || isRememberedDeviceMatch;
+    });
 
     if (affectedDevice.length > 1) {
         console.error('there must be only one device with the same id and without state');
@@ -323,7 +328,6 @@ const setDeviceState = (
 
     affectedDevice[0].state = state;
     affectedDevice[0].useEmptyPassphrase = useEmptyPassphrase;
-    // affectedDevice[0].instance = Number.parseInt(state.staticSessionId?.split(':')[1]!);
     affectedDevice[0].walletNumber = deviceUtils.getNewWalletNumber(draft.devices, device);
 };
 
--- a/suite-common/wallet-core/src/discovery/discoveryThunks.ts
+++ b/suite-common/wallet-core/src/discovery/discoveryThunks.ts
@@ -107,6 +107,7 @@ const applyDeviceStatesThunk = createThunk(
 
             assertDeviceIsAcquired(device);
             assertStaticSessionId(newDeviceState);
+            const { staticSessionId } = newDeviceState;
 
             const physicalDevices = selectPhysicalDevices(getState());
             const devicesWithoutState = physicalDevices.filter(d => !d.state?.staticSessionId);
@@ -133,16 +134,12 @@ const applyDeviceStatesThunk = createThunk(
                     }),
                 );
 
-                // todo: there is probably more efficient way to select device after it was created
-                const newlyAddedDevice = selectDeviceByStaticSessionId(
-                    getState(),
-                    newDeviceState.staticSessionId,
-                );
-                if (!newlyAddedDevice) return;
+                // select the device after deviceReducer updates it (it's a new object reference)
+                const newlyAddedDevice = selectDeviceByStaticSessionId(getState(), staticSessionId);
+                if (newlyAddedDevice === undefined) return;
                 dispatch(selectDeviceThunk({ device: newlyAddedDevice }));
             }
 
-            const { staticSessionId } = newDeviceState;
             await dispatch(initNewDeviceStateMetadataThunk(staticSessionId));
         } catch (error) {
             console.error('applyDeviceStatesThunk error', error);
--- a/suite-native/device-manager/src/components/WalletList.tsx
+++ b/suite-native/device-manager/src/components/WalletList.tsx
@@ -4,16 +4,17 @@ import { A } from '@mobily/ts-belt';
 
 import { TrezorDevice } from '@suite-common/suite-types';
 import {
-    createDeviceInstanceThunk,
     selectDeviceInstances,
     selectIsPortfolioTrackerDevice,
     selectSelectedDevice,
+    startDiscoveryThunk,
 } from '@suite-common/wallet-core';
 import { VStack } from '@suite-native/atoms';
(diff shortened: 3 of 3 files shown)


suite-native/discovery/src/discoveryMiddleware.ts
```diff
@@ -64,12 +65,14 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
...
             const device = selectSelectedDevice(getState());
+            const hasRunningDiscovery = selectHasRunningDiscovery(getState());
             if (
                 isCoinEnablingInitFinished &&
                 isDeviceFirmwareVersionSupported &&
                 device &&
                 device.connected &&
-                isDeviceAcquired(device)
+                isDeviceAcquired(device) &&
+                !hasRunningDiscovery
```
Lemonexe: just a small tweak to avoid starting the thunk needlessly, although it would just early return, so it's not important

suite-native/device-manager/src/components/WalletList.tsx
```diff
@@ -26,23 +27,23 @@ export const WalletList = ({ onSelectDevice }: WalletListProps) => {
...
+    const handlePlaceholderPress = () => {
+        if (!selectedDevice) return;
+        setIsDeviceManagerVisible(false);
+        dispatch(
+            startDiscoveryThunk({
+                device: selectedDevice,
+                isAddingHiddenWallet: false,
+                isAddingExistingWallet: false,
+            }),
+        );
```
Lemonexe, line 49: behavior partially copied from Suite [AddWalletButton.](https://github.com/trezor/trezor-suite/blob/fd745c6bca0e7bb391031f9c943e9bdc465efff2/packages/suite/src/views/suite/SwitchDevice/DeviceItem/AddWalletButton.tsx#L40-L56)

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -136,17 +136,14 @@ const applyDeviceStatesThunk = createThunk(
                         },
                     }),
                 );
-
-                // todo: there is probably more efficient way to select device after it was created
```
Lemonexe, line 136: nah, I don't think there is. It's the job of `deviceReducer` to update the device object with new `state`, and these are just actions, not thunks, so they can't be "smart" and return the updated object. We need to let reducer do its thing and reselect the new device from state.

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -136,17 +136,14 @@ const applyDeviceStatesThunk = createThunk(
...
-                );
-                if (!newlyAddedDevice) return;
-                dispatch(selectDeviceThunk({ device: newlyAddedDevice }));
             }
 
+            // select the device after deviceReducer updates it (it's a new object reference)
             const { staticSessionId } = newDeviceState;
+            const newlyAddedDevice = selectDeviceByStaticSessionId(getState(), staticSessionId);
+            if (newlyAddedDevice === undefined) return;
+            dispatch(selectDeviceThunk({ device: newlyAddedDevice }));
```
Lemonexe, line 145: note: just code moved from `else`, so it executes for both `if` & `else`

suite-common/wallet-core/src/device/deviceReducer.ts
```diff
@@ -313,14 +313,19 @@ const setDeviceState = (
...
+        const isConnectedDeviceMatch =
+            d.connected &&
+            d.instance === device.instance &&
+            (d.id === device.id || (d.path.length > 0 && d.path === device.path));
+
+        // update "disconnected" remembered devices if in bootloader mode
+        const isRememberedDeviceMatch = d.mode === 'bootloader' && d.remember && d.id === device.id;
+
+        return isConnectedDeviceMatch || isRememberedDeviceMatch;
+    });
```
Lemonexe, line 321: Besides reorganizing to make the big logical statement more readable, I only added the instance `condition` to fix the bug.
The bug is: both standard & passphrase are in `affectedDevice`, and `if (affectedDevice.length > 1)` errors out.

suite-native/device-manager/src/components/WalletList.tsx
PeKne: Is this still necessary? Now is discovery executed from middleware whenever is the `selectedDevice`  Changed. Do you have idea why it does not work for this case?

suite-native/discovery/src/discoveryMiddleware.ts
Lemonexe: Oh no! I'll revert it, I see big bad.

suite-native/device-manager/src/components/WalletList.tsx
Lemonexe: It unfortuantely does not work because when selected device already has staticSessionId, the discoveryMiddleware just exectures restartDiscoveryThunk, which means: for current passphrase.

We need to explicitly start new discovery with empty passhprase.
What I **don't understand** is why the `onSelectDevice(selectedDevice);` line works. It must be there, otherwise you just create wallet while staying on the passphrase. But `selectedDevice` _was the passphrase device_

PeKne (approved):

suite-native/device-manager/src/components/WalletList.tsx
vojtatranta: We shoudn't really rely on the middleware. If you want to do something, just call the thunk for it whenever possible.

Lemonexe: @vojtatranta then take a look at https://github.com/trezor/trezor-suite/pull/19406, which is alredy merged. If you have a better idea that covers all the cases, then it'd be great.

vojtatranta: Yeah, I see it. I am late in the game.

I think we could make some ESLint rule which would prohibit adding anymore code to the middlewares.

matejkriz: In suite-native, we’re very cautious about adding things to middlewares, but in this case, it seems to me that it clarifies the logic enough to be justifiable.

matejkriz:
EDIT: fixed in https://github.com/trezor/trezor-suite/pull/19452/commits/04e40a1a5dcbcacda30e5afeb45e880a4ef16fb3

It starts the discovery of Standard wallet even if the trezor is not connected at the moment. And even if the trezor is connected, user is not redirected in the Standard wallet right away, stays on the Passhprase wallet for a moment before and switched to Standard later. 

https://github.com/user-attachments/assets/0d2ec530-4f16-410b-b07b-ba803401bf3a

Lemonexe:
> It starts the discovery of Standard wallet even if the trezor is not connected at the moment. And even if the trezor is connected, user is not redirected in the Standard wallet right away, stays on the Passhprase wallet for a moment before and switched to Standard later.

Yes, you're right! In the meantime I've also discovered that and I have fix locally, but need to finish things up.

Lemonexe:
The Android E2E [confirmed to be broken see Slack](https://satoshilabs.slack.com/archives/C03H2Q5MX51/p1749476043443559)

PR #19481, opened 2025-06-10
fix(suite-native): make sure that passphrase form is opened when user…

vojtatranta:
… requests entering the adding of passphrase wallet again after passphrase mismatch

## Description

Problem: the user was stuck on hte previosu wallet after he clicked "start over". 

Solution: the reason for the error was that the preivous discvoery was properly ended and the redirect didnt happen (coz it is in the add wallet button). So manual navigation to the passphrase form was made to open the proper screen when user requested it.

## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=native-fix-passphrase-mismatch-retry&lookback=7)

### DIFF
--- a/suite-native/module-authorize-device/src/components/passphrase/PassphraseMismatchAlert.tsx
+++ b/suite-native/module-authorize-device/src/components/passphrase/PassphraseMismatchAlert.tsx
@@ -1,6 +1,8 @@
 import { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
+import { useNavigation } from '@react-navigation/native';
+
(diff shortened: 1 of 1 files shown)


matejkriz (approved):
It works as expected 👍  The fact that screen that looks like there is a running discovery for standard wallet before redirecting again to passphrase form is not great, but not terrible and I think it was like that even before 👌 

https://github.com/user-attachments/assets/17c3020e-25e7-42c0-90e1-ad06a25791e1

PR #19489, opened 2025-06-10
Feat/connect handshake with THP device

szymonlesisz:
## Description
Something to discuss with @marekrjpolak related to https://github.com/trezor/trezor-suite/pull/19048

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/connect-thp-handshake&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/connect-thp-handshake&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/connect-thp-handshake&lookback=7)

PR #19490, opened 2025-06-10
feat(connect): add thpCall module

szymonlesisz:
## Description

cherry picked module for communication on `THP` layer.

this is basically the same as regular `TypedCall` (same-same but different) using `THP` protobuf messages which are not a part of `@trezor/protobuf` package (TypedCall message assertion wont work)

`thpCall` will be used only in THP [handshake and pairing process](https://github.com/trezor/trezor-suite/pull/19489)

im also changing `DeviceCurrentSession` **call** and **send** argument types to match actual `@trezor/transport` call and send.
those 2 functions should not be strongly typed (associated only with @trezor/protobuf like .typedCall)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/connect-thpcall&lookback=7)

### DIFF
--- a/packages/connect/src/device/thp/thpCall.ts
+++ b/packages/connect/src/device/thp/thpCall.ts
@@ -1,0 +1,83 @@
+import { thp as protocolThp } from '@trezor/protocol';
+
+import { ERRORS } from '../../constants';
+import { DEVICE } from '../../events';
+import type { Device } from '../Device';
+
+type ThpTypedCall = {
+    ThpCreateChannelRequest: 'ThpCreateChannelResponse';
+    ThpHandshakeInitRequest: 'ThpHandshakeInitResponse';
+    ThpHandshakeCompletionRequest: 'ThpHandshakeCompletionResponse';
+    ThpPairingRequest: 'ThpPairingRequestApproved';
+    ThpSelectMethod: [
+        'ThpCodeEntryCommitment',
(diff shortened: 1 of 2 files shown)


mroz22:
looks like brain food for @marekrjpolak

packages/connect/src/device/thp/thpCall.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    if (result.payload.type === 'Failure') {
+        throw ERRORS.serializeError(result.payload.message);
+    }
+
+    if (result.payload.type === 'ThpError') {
+        throw ERRORS.serializeError(result.payload.message);
+    }
+
+    return result.payload as ThpCallResponse[T];
```
marekrjpolak, line 82: Is it ok that, unlike in `typedCall`, the payload is not typechecked in runtime? Or will it be added later?

marekrjpolak (approved):
I have just one question, seems good otherwise,

packages/connect/src/device/thp/thpCall.ts
szymonlesisz: it is runtime asserted on the lower (transport read) level see `expectedHeaders` in https://github.com/trezor/trezor-suite/pull/19522

writing additional assertion here seems to be useless

PR #19536, opened 2025-06-12
fix(suite-native): device onboarding e2e test firmware skip

PeKne:
## Description
- optionally skip the firmware update screen in device onboarding test if the firmware version is already latest (in that case the screen is not displayed in the flow)
  - this test was failing recently in CI because trezor-user-env was updated and the emulators used the latest firmware by default 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/onboarding-test-fw-install&lookback=7)


BrantalikP (approved):
🚀

PR #19514, opened 2025-06-11
Range type util

mroz22:
while reviewing #18821 (specifically this part https://github.com/trezor/trezor-suite/pull/18821#discussion_r2139763942) I thought that it would be nice to have this [kind of hammer](c87fd6a8ef8e188eb325cb229f4065e0d4378587).  And I also found a [nail](f35024b05d5e9be052fbf76f999583164f9733a1) that can be struck with it 

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=range-type-util&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=range-type-util&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=range-type-util&lookback=7)

PR #19539, opened 2025-06-12
feat: deterministic int in range from string

karliatto:
## Description

New utility to produce deterministic number in a range from 0 to 100 from a string.

This will be used in https://github.com/trezor/trezor-suite/pull/18821

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/util-string-to-int-in-range&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/util-string-to-int-in-range&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/util-string-to-int-in-range&lookback=7)

### DIFF
--- a/packages/utils/src/getIntegerInRangeFromString.ts
+++ b/packages/utils/src/getIntegerInRangeFromString.ts
@@ -1,0 +1,26 @@
+/**
+ *
+ * @param input - any string
+ * @param max - max number not included for the range, if 100 is provided the range will be 0 to 99, if 101, then range is 0 to 100
+ * @returns number
+ * @see https://stackoverflow.com/questions/6122571/simple-non-secure-hash-function-for-javascript
+ */
+export const getIntegerInRangeFromString = (input: string, max: number): number => {
+    let hash = 0;
+    if (input.length === 0) {
+        return 0;
+    }
+
+    // Non secure but fash hash function with browser compatibility
+    // https://stackoverflow.com/questions/6122571/simple-non-secure-hash-function-for-javascript
+    for (let i = 0; i < input.length; i++) {
+        const char = input.charCodeAt(i);
+        hash = (hash << 5) - hash + char;
+        // Convert to 32bit integer.
+        hash |= 0;
+    }
+
+    // The result of `x % n` is always in the range `-(n-1)` to `n-1`.
+    // Taking the absolute value ensures it's in the range 0 to max.
+    return Math.abs(hash % max);
+};
--- a/packages/utils/tests/getIntegerInRangeFromString.test.ts
+++ b/packages/utils/tests/getIntegerInRangeFromString.test.ts
@@ -1,0 +1,33 @@
+import { getIntegerInRangeFromString } from '../src/getIntegerInRangeFromString';
+
+describe('getIntegerInRangeFromString', () => {
+    it('should always produce the same number for the same string input', () => {
+        const input = '47A2176AC04E609FD0E9811A';
+        const result1 = getIntegerInRangeFromString(input, 101);
+        const result2 = getIntegerInRangeFromString(input, 101);
+        expect(result1).toEqual(result2);
+    });
+
+    it('should always be deterministic in producing a number from same input', () => {
+        const input = 'z';
+        const result1 = getIntegerInRangeFromString(input, 101);
+        expect(result1).toBe(21);
+    });
+
+    it('should return a number between 0 and 100 for various inputs', () => {
+        const inputs = [
+            'z',
+            '', // It should work with empty string as well.
+            'a-much-longer-string-to-test-just-to-test-that-something-like-this-will-also-work',
+            '47A2176AC04E609FD0E9811A',
+            '!@#$%^&*()-and-some-other-characters-ěščřžýáíé',
+            'another-device-id-456',
+        ];
+
+        inputs.forEach(input => {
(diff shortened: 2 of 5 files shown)


mroz22:
check pls this #19514 a similar utility already exist, so maybe they could be somehow unified

karliatto:
> check pls this #19514 a similar utility already exist, so maybe they could be somehow unified

That is what I am doing in this PR. I created the new utils `stringToIntegerInRange` and that new utils is used in  `getInclusionFromInstanceId`.

packages/utils/src/stringToIntegerInRange.ts
```diff
@@ -0,0 +1,17 @@
...
+    if (input.length === 0) {
+        return 0;
+    }
+
+    for (let i = 0; i < input.length; i++) {
+        const char = input.charCodeAt(i);
+        hash = (hash << 5) - hash + char;
+        // Convert to 32bit integer.
+        hash |= 0;
+    }
```
karliatto: We could use `createHash` from node but it would make this utils only work in node and like this will work in browser as well.

mroz22:
please please review @marekrjpolak

packages/utils/tests/getIntegerInRangeFromString.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        expect(result1).toEqual(result2);
+    });
+
+    it('should always be deterministic in producing a number from same input', () => {
+        const input = 'z';
+        const result1 = getIntegerInRangeFromString(input, 101);
+        expect(result1).toBe(21);
+    });
+
+    it('should produce different numbers for different string inputs', () => {
```
marekrjpolak: Well this may not be true in general since there's only a hundred possible outputs. I mean, it's a coincidence (high probability tho) that these different strings produce different numbers.

packages/utils/tests/getIntegerInRangeFromString.test.ts
```diff
@@ -0,0 +1,40 @@
...
+            'a-much-longer-string-to-test-just-to-test-that-something-like-this-will-also-work',
+            '47A2176AC04E609FD0E9811A',
+            '!@#$%^&*()-and-some-other-characters-ěščřžýáíé',
+            'another-device-id-456',
+        ];
+
+        inputs.forEach(input => {
+            const result = getIntegerInRangeFromString(input, 101);
+            expect(result).toBeGreaterThanOrEqual(0);
+            expect(result).toBeLessThan(100);
```
marekrjpolak: There should be `toBeLessThanOrEqual(100)` or `toBeLessThan(101)`, no?

packages/utils/src/getIntegerInRangeFromString.ts
```diff
@@ -0,0 +1,23 @@
...
+    if (input.length === 0) {
+        return 0;
+    }
+
+    for (let i = 0; i < input.length; i++) {
+        const char = input.charCodeAt(i);
+        hash = (hash << 5) - hash + char;
+        // Convert to 32bit integer.
+        hash |= 0;
+    }
```
marekrjpolak, line 21: I believe it works, but it would be nice to have a link to the source there.

karliatto: Added comment https://github.com/trezor/trezor-suite/pull/19539/commits/0b6868abedbdaff6d3ca745cdfdb77b8b457cf4a

packages/utils/tests/getIntegerInRangeFromString.test.ts
karliatto: Thanks! changed https://github.com/trezor/trezor-suite/pull/19539/commits/0b6868abedbdaff6d3ca745cdfdb77b8b457cf4a

packages/utils/tests/getIntegerInRangeFromString.test.ts
karliatto: You are right, I removed that test case.

karliatto:
I added the commit to update function `getInclusionFromInstanceId` in  suite-common/message-system/src/experimentUtils.ts to use the new util.

suite-common/message-system/src/experimentUtils.ts
```diff
@@ -20,9 +20,8 @@ type ExperimentGetGroupByInclusion = {
  */
 export const getInclusionFromInstanceId = (instanceId: string, experimentId: ExperimentId) => {
     const combinedId = `${instanceId}-${experimentId}`;
-    const hash = createHash('sha256').update(combinedId).digest('hex').slice(0, 8);
 
-    return parseInt(hash, 16) % 100;
+    return getIntegerInRangeFromString(combinedId, 100);
```
karliatto, line 24: @matejkriz do you know if there is an issue if we change this and some users that before were falling into group A now they will follow into group B because the combination of `instanceId` and `experimentalId` will produce different number with this change.

We are not sure if this is an issue.

matejkriz: There hasn’t been any experiment in production yet, so it’s fine. 👌

marekrjpolak (approved):
LGTM

PR #19484, opened 2025-06-10
fix(suite-native): prevent navigating to Connecting screen over Coin enabling

matejkriz:
## Description

Prevent navigating to Connecting screen over Coin enabling screen. 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/suite-native/connecting-screen-over-coin-enabling&lookback=7)

PR #19563, opened 2025-06-13
chore(suite-native): remove redirection with condition that should never happen

matejkriz:
## Description

- useOnDeviceReadyNavigation is only on ConnectingDeviceScreen, and we should never lead the user to that screen without Coin Enabling finished.

## Related Issue

Followup for https://github.com/trezor/trezor-suite/pull/19484

Notes for @trezor/qa:
- Try to break it by connecting trezor to fresh app (coin enabling not visited).
- It could make sense to test also opening app by connecting the trezor (kill app, connect trezor, open app by selecting it in system dialog)
- Test with btc-only FW too

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/remove-hopefully-redundat-redirection&lookback=7)

### DIFF
--- a/suite-native/module-authorize-device/src/hooks/useOnDeviceReadyNavigation.ts
+++ b/suite-native/module-authorize-device/src/hooks/useOnDeviceReadyNavigation.ts
@@ -56,14 +56,6 @@ export const useOnDeviceReadyNavigation = () => {
     useEffect(() => {
         if (isFirmwareInstallationRunning) return;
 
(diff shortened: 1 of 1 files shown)


PeKne (approved):

STew790:
**QA OK**
LGTM, didn't manage to break it, will monitor it throughout the RC testing as well though.

**Info**
25.7.1 65ef77d

PR #19568, opened 2025-06-13
Feat/connect handshake cancel

szymonlesisz:
## Description

Cleaning up in handshake Cancel aka preventive cancel, moving code to standalone module (workflow)

1. Cancel is called **always** as first message even before Initialize (currently ii is done only before GetFeatures)
2. not using Transport.call anymore, instead of call workflow is using send + receive loop until: receive meaningful response or timeout or attempt limit reached.
3. both send and receive are using timeout and abort signal
4. Cancel response can detect protocol v2/thp 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/connect-handshake-cancel&lookback=7)

### DIFF
--- a/packages/connect/src/device/workflow/handshake.ts
+++ b/packages/connect/src/device/workflow/handshake.ts
@@ -1,0 +1,83 @@
+import { PROTOCOL_MALFORMED } from '@trezor/protocol';
+
+import { DataManager } from '../../data/DataManager';
+import { WorkflowContext } from '../../types/workflow';
+import { Log } from '../../utils/debug';
+
+const CANCEL_TIMEOUT = 1_000;
+// Due to performance issues in suite-native during app start, original timeout is not sufficient.
+const CANCEL_TIMEOUT_REACT_NATIVE = 20_000;
+const ATTEMPTS_LIMIT = 10;
+
+type Context = {
+    device: WorkflowContext['device'];
+    signal: AbortSignal;
+    logger?: Log;
+};
+
+// note 1: clear communication with the device using Cancel message. This causes any remaining messages in its transport stack to get flushed.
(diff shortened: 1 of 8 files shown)


packages/transport/src/transports/bridge.ts
```diff
@@ -291,7 +292,7 @@ export class BridgeTransport extends AbstractTransport {
 
                 return this.success(undefined);
             },
-            { signal },
+            { signal, timeout },
```
marekrjpolak, line 295: This `timeout` (unlike the other ones), when not passed as a param, will nullify the default `ACTION_TIMEOUT` in `AbstractTransport`. Is it fine?

packages/connect/src/device/workflow/handshake.ts
```diff
@@ -0,0 +1,91 @@
...
+            signal,
+            timeout: cancelTimeout,
+        });
+
+        // Malformed protocol is thrown if received chunk:
+        // 1. is empty message (empty buffer)
+        // 2. has invalid protocol header
+        // 3. it is a continuation packet
+        // 4. message could not be encoded for some other reason (for example: missing definitions)
+        if (!result.success && result.error.message !== 'Malformed protocol format') {
```
szymonlesisz, line 72: get message from constants?

marekrjpolak (approved):
LGTM for now.
- works with HID T1's :heavy_check_mark: 
- I'll try to re-polish `DeviceCurrentSession` later when it'll be used for thp

packages/transport/src/transports/bridge.ts
szymonlesisz: its ok, discussed irl

PR #19574, opened 2025-06-13
fix(suite-native): initialize Connect after onboarding is finished

yanascz:
If an initialized Trezor is connected to a freshly installed Suite Lite app, Trezor Connect is initialized before onboarding is finished and thus may lead to a confusing prompt on the device. Mostly relevant to THP.

This is fixed by splitting the init code into two parts. 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/application-init&lookback=7)

### DIFF
--- a/suite-native/app/src/App.tsx
+++ b/suite-native/app/src/App.tsx
@@ -15,13 +15,14 @@ import { IntlProvider } from '@suite-native/intl';
 import { KillswitchMessageScreen } from '@suite-native/message-system';
 import { NavigationContainerWithAnalytics } from '@suite-native/navigation';
 import { initSentry } from '@suite-native/sentry';
-import { StoreProvider, selectIsAppReady, selectIsConnectInitialized } from '@suite-native/state';
+import { selectIsOnboardingFinished } from '@suite-native/settings';
+import { StoreProvider, selectIsAppReady } from '@suite-native/state';
 
 import { BannersRenderer } from './BannersRenderer';
 import { ModalsRenderer } from './ModalsRenderer';
 import { StylesProvider } from './StylesProvider';
 import { useReportAppInitToAnalytics } from './hooks/useReportAppInitToAnalytics';
-import { applicationInit } from './initActions';
+import { applicationInit, postOnboardingInit } from './initActions';
 import { RootStackNavigator } from './navigation/RootStackNavigator';
 import { disableRTL } from './rtl';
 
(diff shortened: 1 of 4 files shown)


suite-native/app/src/App.tsx
```diff
@@ -51,21 +52,32 @@ const AppComponent = () => {
...
+    const isOnboardingFinished = useSelector(selectIsOnboardingFinished);
+
+    const isApplicationInitDispatchedRef = useRef(false);
+    const isPostOnboardingInitDispatchedRef = useRef(false);
 
     useReportAppInitToAnalytics(APP_STARTED_TIMESTAMP);
 
     useEffect(() => {
-        if (!isConnectInitialized) {
+        if (!isApplicationInitDispatchedRef.current) {
```
juriczech: Why don't you use `!isAppReady` here? It might be useful to add a comment here to explain the reason.

suite-native/app/src/App.tsx
```diff
@@ -51,21 +52,32 @@ const AppComponent = () => {
...
         }
-    }, [dispatch, isConnectInitialized]);
+    }, [dispatch]);
 
     useEffect(() => {
         if (isAppReady) {
             SplashScreen.hideAsync();
         }
-    }, [isAppReady]);
+    }, [isAppReady, dispatch]);
```
juriczech: `dispatch` is redundant in the dependency array.

suite-native/app/src/initActions.ts
```diff
@@ -13,36 +13,35 @@ import {
 import { walletConnectInitThunk } from '@suite-common/walletconnect';
 import { initAnalyticsThunk } from '@suite-native/analytics';
 import { FeatureFlag, selectIsFeatureFlagEnabled } from '@suite-native/feature-flags';
-import { setIsAppReady, setIsConnectInitialized } from '@suite-native/state/src/appSlice';
+import { setIsAppReady } from '@suite-native/state/src/appSlice';
 
-let isAlreadyInitialized = false;
+const ACTION_PREFIX = '@suite-native/app';
```
juriczech, line 18: FYI: The reason for the global variable is described here: https://react.dev/learn/you-might-not-need-an-effect#initializing-the-application

It is according to the recommended approach from react docs and I wonder, does changing it to ref have any benefit?

suite-native/app/src/App.tsx
yanascz: Good catch, removed.

suite-native/app/src/App.tsx
yanascz: Because we want that code to be called exactly once.

suite-native/app/src/initActions.ts
yanascz: OK, used instead of refs.

juriczech (approved):

STew790:
**QA Status**
I did onboarding multiple times throughout past week(s) and haven't encountered any issue but I will leave this open and monitor through RC testing as well, just to be sure.

PR #19577, opened 2025-06-13
  Feat: update fee rate options regularly – attempt 2

### LINKED ISSUE #18406: Update fee rate regularly while being in send form

### LINKED ISSUE #19281: Fee refetch resets my passphrase input
**Describe the bug**

I think refetch should be disabled on `Review & send` button click

https://github.com/user-attachments/assets/844c0a30-f3e5-422d-b7de-78585446589a

### LINKED ISSUE #19282: Sending amount is changing under my hands because of fee refetching
**Describe the bug**

Amount I am sending has changed and nobody told me. I want to be notified, see some warning,... or do not modify it for me :/ 

- It undermines my trust in “what exactly am I sending?”
- I see that number changed unexpectedly and it terrifies me

https://github.com/user-attachments/assets/2e150ee2-bc23-4959-bfd4-929cc89e6fc5

- I send wrong amount and I lose my funds

Image

Lemonexe:
Reapply #19242 but a lot of changes were done.

Depends on #19721

:eyes: CR commit by commit recommended:

- 08cce95ca304f2f0292a6cdfb3dde94a836e188a – unchanged from [#19242](https://github.com/trezor/trezor-suite/pull/19242) except renaming [as per comment](https://github.com/trezor/trezor-suite/pull/19577#discussion_r2147494554)
- 61518189da7c2eaca491087facee2286b9c02d50 – **the main thing**, rewritten a lot, builds on [#19721](https://github.com/trezor/trezor-suite/pull/19721)
- de6b15db01eafda68deb62475d821316079eda60 – small UI tweak
- 91dafece6339fb4312dfa75bd02cad19e91250da – small UI tweak
- af9f99c05b48819318b685ec245914617465e4ec – fixes [#19282](https://github.com/trezor/trezor-suite/issues/19282)
- 7606f4d69a62e348b3a5b36d3271e7c66798a94f – fixes [#19281](https://github.com/trezor/trezor-suite/issues/19281)

## Description

Update fee rates regularly at all places where fee selection is available.
- periodicaly refetch fees every 60s
  - but not when modal is open
  - and not when send max is chosen
  - it works everywhere where the Fees levels options are displayed (not just send form)
- display skeletons when loading [as per figma](https://www.figma.com/design/qx5aOavcU7I2MW2PtnQtfz/Small-Things--DSK-?node-id=4048-978&t=MFI4WubrD2PkITeY-1)
- new `isApproxEqual` util for bignumbers

what I did different than specified in Figma:
- disabled the Review & send button consistently with the loaders around it
  - otherwise it would flick on/off, but with inconsistent timing (only as the draft is recomposed)
- the fetching is very quick, so if we displayed loaders only during the loading itself, they just blink on/off quickly.
  - that looks disruptive and it's not clear what is happening, just that _something blinked on the screen_
  - the point of the loaders is to visually signify to user that the fees are changing, so I extend the loader to at least one second. IMHO it works very well, see videos below :slightly_smiling_face:
- it also works in: Bump Fee, Stake, Unstake, Claim, Trading–sell
  - their confirm buttons are also loading with fees, consistently with Send Form (but no figma designs here)

## Related Issue

Resolve #18406 (again)
Resolve #19281
Resolve #19282

## Screenshots (new in this PR)

[Don't refetch when transaction review modal is open](https://github.com/user-attachments/assets/590a37bf-d698-43f3-93f1-746565c75104)
[Behavior in Bump Fee Modal](https://github.com/user-attachments/assets/523e3d99-f4fa-4d17-9dee-ff5f6ae33aef)
[Behavior in Sell page](https://github.com/user-attachments/assets/63bd5746-1363-48de-83d9-768716869632)
[Behavior in Stake modal](https://github.com/user-attachments/assets/47b15665-4b92-4862-89a1-2293c91af5e9)
[Behavior in Unstake modal](https://github.com/user-attachments/assets/9f866f04-1f79-4458-b18a-d670496057d2)
[Behavior in Claim modal](https://github.com/user-attachments/assets/de1e647e-5151-40bc-ab94-491678aaceeb)

## Screenshots (copied from old PR)

Video of the new behavior, when high fee value is selected and it's just changing: 

[demo.webm](https://github.com/user-attachments/assets/a0354d18-3949-464d-b0ff-0af6590d2573)

Playing around with various scenarios (it does not lose my progress, it does not reset my fee level choice, dust prevention works etc.), with update time lowered to 10 seconds

[playing around.webm](https://github.com/user-attachments/assets/ecd01cef-9a68-41eb-a82c-796b14fc5b40)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/send-form-update-fees-regularly-attempt-2&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/send-form-update-fees-regularly-attempt-2&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/send-form-update-fees-regularly-attempt-2&lookback=7)

### DIFF
--- a/packages/suite/src/components/wallet/Fees/DustPreventionNotice.tsx
+++ b/packages/suite/src/components/wallet/Fees/DustPreventionNotice.tsx
@@ -1,8 +1,13 @@
+import { NetworkSymbol } from '@suite-common/wallet-config';
+import { selectAreFeesLoading } from '@suite-common/wallet-core';
 import { Note } from '@trezor/components';
+import { isApproximatelyEqual } from '@trezor/utils';
 
+import { useSelector } from '../../../hooks/suite';
 import { Translation } from '../../suite';
 
 type DustPreventionNoticeProps = {
+    symbol: NetworkSymbol;
     chosenFeePerByte: string | undefined;
     composedFeePerByte: string | undefined;
     baseFee: number | undefined;
@@ -10,15 +15,20 @@ type DustPreventionNoticeProps = {
 };
 
 export const DustPreventionNotice = ({
+    symbol,
     chosenFeePerByte,
     composedFeePerByte,
     baseFee,
     feeUnits,
 }: DustPreventionNoticeProps) => {
+    const areFeesLoading = useSelector(state => selectAreFeesLoading(state, symbol));
+
+    const relativeTolerance = 1e-3;
     const isComposedFeeRateDifferent =
+        !areFeesLoading &&
         composedFeePerByte !== undefined &&
-        composedFeePerByte !== '' &&
-        chosenFeePerByte !== composedFeePerByte;
+        chosenFeePerByte !== undefined &&
+        !isApproximatelyEqual(composedFeePerByte, chosenFeePerByte, relativeTolerance);
 
     if (!isComposedFeeRateDifferent) return null;
 
--- a/packages/suite/src/components/wallet/Fees/Fees.tsx
+++ b/packages/suite/src/components/wallet/Fees/Fees.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from 'react';
+import { useMemo } from 'react';
 import {
     Control,
     FieldErrors,
@@ -13,7 +13,6 @@ import { useTheme } from 'styled-components';
 
 import { TranslationKey } from '@suite-common/intl-types';
 import { NetworkSymbol, NetworkType } from '@suite-common/wallet-config';
-import { updateFeeInfoThunk } from '@suite-common/wallet-core';
 import {
     FeeInfo,
     FormState,
@@ -27,7 +26,7 @@ import { spacings } from '@trezor/theme';
 import { HELP_CENTER_TRANSACTION_FEES_URL } from '@trezor/urls';
 
 import { Translation } from 'src/components/suite';
-import { useDispatch } from 'src/hooks/suite';
+import { useRefetchFees } from 'src/hooks/wallet/useRefetchFees';
 import { Account } from 'src/types/wallet';
 
 import { CustomFee } from './CustomFee/CustomFee';
@@ -144,7 +143,6 @@ export const Fees = <TFieldValues extends FormState>({
     // Type assertion allowing to make the component reusable, see https://stackoverflow.com/a/73624072.
     const { getValues, register, setValue, trigger } = props as unknown as UseFormReturn<FormState>;
     const theme = useTheme();
-    const dispatch = useDispatch();
 
     const selectedOption = getValues('selectedFee') || 'normal';
     const isCustomFee = selectedOption === 'custom';
@@ -158,9 +156,7 @@ export const Fees = <TFieldValues extends FormState>({
 
     const supportsCustomFee = networkType !== 'solana';
 
-    useEffect(() => {
-        dispatch(updateFeeInfoThunk({ networkSymbol: symbol }));
-    }, [dispatch, symbol]);
+    useRefetchFees({ networkSymbol: symbol });
 
     const feeLabelId = useMemo(() => {
         switch (networkType) {
--- a/packages/suite/src/components/wallet/Fees/Fees.tsx
+++ b/packages/suite/src/components/wallet/Fees/Fees.tsx
@@ -26,7 +26,7 @@ import { spacings } from '@trezor/theme';
 import { HELP_CENTER_TRANSACTION_FEES_URL } from '@trezor/urls';
 
 import { Translation } from 'src/components/suite';
-import { useRefetchFees } from 'src/hooks/wallet/useRefetchFees';
+import { useFetchFeesOnce, useRefetchFees } from 'src/hooks/wallet/useRefetchFees';
 import { Account } from 'src/types/wallet';
 
 import { CustomFee } from './CustomFee/CustomFee';
@@ -156,7 +156,10 @@ export const Fees = <TFieldValues extends FormState>({
 
     const supportsCustomFee = networkType !== 'solana';
 
-    useRefetchFees({ networkSymbol: symbol });
+    useFetchFeesOnce({ networkSymbol: symbol });
+    // this component is used under different contexts & form states, but `setMaxOutputId` will be compatible (see `FormState` type)
+    const isRefetchDisabled = getValues().setMaxOutputId !== undefined;
+    useRefetchFees({ networkSymbol: symbol, isDisabled: isRefetchDisabled });
 
     const feeLabelId = useMemo(() => {
         switch (networkType) {
--- a/packages/suite/src/components/wallet/Fees/StandardFee/FeeCard.tsx
+++ b/packages/suite/src/components/wallet/Fees/StandardFee/FeeCard.tsx
@@ -1,10 +1,11 @@
-import React from 'react';
+import { ReactNode } from 'react';
 
 import {
     Box,
     Column,
     RadioCard,
     Row,
+    SkeletonRectangle,
     TOOLTIP_DELAY_NORMAL,
     Text,
     Tooltip,
@@ -17,11 +18,12 @@ type FeeCardProps = {
     value: FeeLevel['label'];
     isSelected: boolean;
     changeFeeLevel: (level: FeeLevel['label']) => void;
-    topLeftChild: React.ReactNode;
-    topRightChild?: React.ReactNode;
-    bottomLeftChild: React.ReactNode;
-    bottomRightChild: React.ReactNode;
-    tooltipContent?: React.ReactNode;
+    topLeftChild: ReactNode;
+    topRightChild?: ReactNode;
+    bottomLeftChild: ReactNode;
+    bottomRightChild: ReactNode;
+    tooltipContent?: ReactNode;
+    isLoading?: boolean;
     'data-testid'?: string;
 };
 
@@ -34,6 +36,7 @@ export const FeeCard = ({
     bottomLeftChild,
     bottomRightChild,
     tooltipContent,
+    isLoading,
     'data-testid': dataTestId,
 }: FeeCardProps) => (
     <Box data-testid={dataTestId} minWidth={FEE_CARD_MIN_WIDTH}>
@@ -43,13 +46,15 @@ export const FeeCard = ({
                     <Row justifyContent="space-between">
                         <Text typographyStyle="highlight">{topLeftChild}</Text>
                         <Text variant="tertiary" typographyStyle="hint">
-                            {topRightChild}
+                            {isLoading ? <SkeletonRectangle animate={true} /> : topRightChild}
                         </Text>
                     </Row>
                     <Row justifyContent="space-between" height={24}>
-                        <Text>{bottomLeftChild}</Text>
+                        <Text>
+                            {isLoading ? <SkeletonRectangle animate={true} /> : bottomLeftChild}
+                        </Text>
                         <Text variant="tertiary" typographyStyle="hint">
-                            {bottomRightChild}
+                            {isLoading ? <SkeletonRectangle animate={true} /> : bottomRightChild}
                         </Text>
                     </Row>
                 </Column>
--- a/packages/suite/src/hooks/wallet/useSendForm.ts
+++ b/packages/suite/src/hooks/wallet/useSendForm.ts
@@ -74,7 +74,6 @@ const getStateFromProps = (props: UseSendFormProps) => {
     return {
         account,
         network,
-
         localCurrencyOption,
         online: props.online,
         metadataEnabled: props.metadataEnabled,
@@ -321,7 +320,7 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
         state.network.decimals,
     ]);
 
-    // load draft from reducer
+    // load draft from reducer and reset current form values, this should be only called once on mount
     useEffect(() => {
         const loadDraftValues = async () => {
             const storedState = await dispatch(getSendFormDraftThunk()).unwrap();
@@ -336,7 +335,9 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
             }
         };
         loadDraftValues();
-    }, [dispatch, getLoadedValues, reset, composeDraft]);
+        // composeDraft is excluded because its reference changes with each feeInfo update.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [dispatch, getLoadedValues, reset]);
 
     // register custom form fields (without HTMLElement)
     useEffect(() => {
@@ -349,7 +350,14 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
         if (!draft.current) return;
         composeDraft(draft.current);
         draft.current = undefined;
-    }, [draft, composeDraft]);
+        // composeDraft is excluded because its reference changes with each feeInfo update.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [draft]);
+
+    // update composedLevels when feeInfo changes
+    useEffect(() => {
+        composeDraft(getValues());
+    }, [composeDraft, getValues]);
 
     // handle draftSaveRequest
     useEffect(() => {
--- a/packages/suite/src/views/wallet/send/TotalSent/TotalSent.tsx
+++ b/packages/suite/src/views/wallet/send/TotalSent/TotalSent.tsx
@@ -1,15 +1,22 @@
-import { useMemo } from 'react';
+import { PropsWithChildren, useMemo } from 'react';
 
 import styled from 'styled-components';
 
+import { selectAreFeesLoading } from '@suite-common/wallet-core';
 import { formatAmount, formatNetworkAmount } from '@suite-common/wallet-utils';
-import { Card, Column, InfoItem } from '@trezor/components';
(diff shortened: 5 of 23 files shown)


packages/suite/src/components/wallet/Fees/DustPreventionNotice.tsx
```diff
@@ -15,10 +18,14 @@ export const DustPreventionNotice = ({
...
+    const areFeesLoading = useSelector(state => selectAreFeesLoading(state));
+
+    const relativeTolerance = 1e-3;
     const isComposedFeeRateDifferent =
+        !areFeesLoading &&
         composedFeePerByte !== undefined &&
-        composedFeePerByte !== '' &&
-        chosenFeePerByte !== composedFeePerByte;
+        chosenFeePerByte !== undefined &&
+        !isApproxEqual(composedFeePerByte, chosenFeePerByte, relativeTolerance);
```
Lemonexe: I found that the hard way that recomposing the transaction with unchanged feeLevels may produce different string representation of the same number: `'5.1'` vs `'5.10'`
and the `DustPreventionNotice` would display when it shouldn't.

packages/suite/src/hooks/wallet/useSendForm.ts
```diff
@@ -348,7 +349,14 @@ export const useSendForm = (props: UseSendFormProps): SendContextValues => {
...
         draft.current = undefined;
-    }, [draft, composeDraft]);
+        // composeDraft is excluded because its reference changes with each feeInfo update.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [draft]);
+
+    // update composedLevels when feeInfo changes
+    useEffect(() => {
+        composeDraft(getValues());
+    }, [composeDraft, getValues]);
```
Lemonexe, line 360: basically I extracted partial functionality from the hook above: when fees change, just recompose levels but do not reset the whole form..

Lemonexe: :thought_balloon:  I am not happy that Send Form relies so much on useEffects.

The lifecycle is really unpredictable – there was original assumption that the two useEffects above would run only once when mounted, but now `composeDraft` fn legitimately changes reference when fees are updated.. 

I'm not the first one to have this problem, maybe someone will rewrite it to OOP one day :sweat_smile:

packages/utils/src/isApproxEqual.ts
```diff
@@ -0,0 +1,24 @@
...
+    value2 = new BigNumber(value2);
+    relativeTolerance = new BigNumber(relativeTolerance);
+
+    // Cannot calculate relative difference if one value is zero, but then they must be equal
+    if (value1.eq(0)) return value1.eq(value2);
+
+    const relativeDifference = value2.minus(value1).abs().dividedBy(value1);
+
+    return relativeDifference.lte(relativeTolerance);
+};
```
Lemonexe: This util may seem like overkill, but it's methodically correct to do it when comparing floats.
There could be a lot of places in code where this should be used instead of `===`.
Also I had fun writing it :heart: 

Btw `bignumber.js` does not have this, not even simpler version with absolute tolerance.

suite-common/wallet-core/src/fees/feesThunks.ts
```diff
@@ -70,83 +75,83 @@ export const preloadFeeInfoThunk = createThunk(
...
+};
+
 export const updateFeeInfoThunk = createThunk(
     `${FEES_MODULE_PREFIX}/updateFeeInfoThunk`,
-    async ({ networkSymbol }: { networkSymbol: NetworkSymbol }, { dispatch, getState, extra }) => {
+    async ({ networkSymbol }: { networkSymbol: NetworkSymbol }, { dispatch, getState }) => {
         const network = getNetworkOptional(networkSymbol.toLowerCase());
         if (!network) return;
         const blockchainInfo = selectNetworkBlockchainInfo(getState(), network.symbol);
+        const device = selectSelectedDevice(getState());
```
Lemonexe, line 145: here I changed the `extra.selectors.selectDevice` to direct import from wallet-core `selectSelectedDevice`.

Hey that's another opportunity for random refactoring/cleanup: turns out we have extra selectors which do not have to be extra :smile:

suite-common/wallet-core/src/send/sendFormThunks.ts
```diff
@@ -607,3 +609,22 @@ export const enhancePrecomposedTransactionThunk = createThunk<
...
+        if (selectAreFeesLoading(getState())) return;
+
+        dispatch(sendFormActions.setAreFeesLoading({ areFeesLoading: true }));
+        await Promise.all([
+            dispatch(updateFeeInfoThunk({ networkSymbol })),
+            new Promise(resolve => setTimeout(resolve, FEE_UPDATE_DELAY_MILLISECONDS)),
+        ]);
+        dispatch(sendFormActions.setAreFeesLoading({ areFeesLoading: false }));
+    },
+);
```
Lemonexe: why did I put this into `sendFormReducer`?
Initially I started in `feesReducer`. But you can see that its state is just `Record<NetworkSymbol, FeeInfo>` so there would be no place to easily add it. I'd either have to append a veery hacky `& {commonState: {} }`, or refactor it completely `state` to `{ feesPerNetwork: state, ...otherThings }`, and that would be a very extensive refactoring :cry: 

But I belive that this actually makes some sense. The loading state is just an visual indication for send form that serves no relation to how the fees are actually fetched (they are already fetched and processed in reducer even when the artificial delay is still ticking). This "loading state" is just for send form..

suite-common/wallet-core/src/send/sendFormSelectors.ts
```diff
@@ -73,3 +73,5 @@ export const selectSendFormReviewButtonRequestsCount = (
 
     return isCardano ? sendFormReviewRequest.length - 1 : sendFormReviewRequest.length;
 };
+
+export const selectAreFeesLoading = (state: SendRootState) => state.wallet.send.areFeesLoading;
```
tomasklim: As we use fees for `send`, `stake`, `trade`, `wallet-connect` and we have `state.wallet.fees`. I would move this there.

Screenshot 2025-06-15 at 10 37 00

We can have isFeeLoading for each network, last updated time, error (using default),... params there

suite-common/wallet-core/src/send/sendFormThunks.ts
tomasklim: I would rather move it to the feesReducer and change how the fee state object looks like. To have everything on one place, nice and clear. 
Send form is our most problematic code, I would not make it even worse 😀

packages/utils/src/isApproxEqual.ts
```diff
@@ -0,0 +1,24 @@
+import { BigNumber, BigNumberValue } from './bigNumber';
+
+/**
+ * Check if two BigNumber values are approximately equal based on relative difference
+ * @param value1
+ * @param value2
+ * @param relativeTolerance maximum threshold for relative difference of values to declare them approx. equal
+ */
+export const isApproxEqual = (
```
tomasklim: nit: No need to shorten it

packages/suite/src/views/wallet/send/TotalSent/TotalSent.tsx
```diff
@@ -22,6 +29,7 @@ export const TotalSent = () => {
         composedLevels,
         getValues,
     } = useSendFormContext();
+    const areFeesLoading = useSelector(selectAreFeesLoading);
```
tomasklim: Can you please also add this to `stake`, `unstake`, `claim`, `sell`, `swap`? 😀

tomasklim (changes requested):

suite-common/wallet-core/src/send/sendFormSelectors.ts
tomasklim: Ah you put comment into the next file. So i commented it there as well

packages/suite/src/views/wallet/send/TotalSent/TotalSent.tsx
tomasklim: All places across the app should behave the same

tomasklim (changes requested):
I don't agree with closing and not resolving this https://github.com/trezor/trezor-suite/issues/19282 issue that the amount changes under my hands @hynek-jina @jirih-stsh @Hermez-cz @rojcyk @MiroslavProchazka 

I believe it definitely breaks some nielsen norman group rules. 

I believe this is not a good argument here.
> Don't use send max if you need specific amount.

- some exchanges have minimum for deposit, in case it changes in the last second, user can lose funds
- in case user is using fixed float, or something like that, he can copy the amount from the send form. Then he sends incorrect amount -> lose of funds or troubles on support
- ...

Easiest solution is just to not update fee when user is sending max 🤷‍♂️ or show some warning that amount changed.

packages/suite/src/components/wallet/Fees/StandardFee/FeeCard.tsx
```diff
@@ -43,13 +46,15 @@ export const FeeCard = ({
                     <Row justifyContent="space-between">
                         <Text typographyStyle="highlight">{topLeftChild}</Text>
                         <Text variant="tertiary" typographyStyle="hint">
-                            {topRightChild}
+                            {isLoading ? <SkeletonRectangle animate={true} /> : topRightChild}
```
tomasklim, line 49: Radio card should be disabled when loading?

suite-common/wallet-core/src/send/sendFormThunks.ts
```diff
@@ -607,3 +609,22 @@ export const enhancePrecomposedTransactionThunk = createThunk<
         return enhancedPrecomposedTransaction;
     },
 );
+
+const FEE_UPDATE_DELAY_MILLISECONDS = 1000;
+/**
+ * Delay updateFeeInfoThunk with an arbitrary timeout, because backend request is usually very quick.
+ * This can be used to display loader for a bit longer, to visually draw users attention to the fees which are changing.
+ */
+export const delayedUpdateFeeInfoThunk = createThunk(
```
tomasklim: Idea, maybe it does not make sense: Would it be possible to quietly refetch fees and if the fee changed, show loader for 1s and recalculate?

suite-common/wallet-core/src/send/sendFormThunks.ts
Lemonexe: Yes, that's what I did at first, but unfortunately encountered problems :disappointed: 
Refactoring `feesReducer` to respect the `isLoading` on each network is the simple part, but the fact that fee levels are referentially unstable even when the numbers don't change, created a lot unpredictability.
There are useEffects that recalculate when the network fee level changes – rightly so, but when you add isLoading on the network fee, it changes three times instead of once.

I will think about [alternative approach](https://github.com/trezor/trezor-suite/pull/19577#discussion_r2147502335).

suite-common/wallet-core/src/send/sendFormThunks.ts
Lemonexe: Yes.
Idea in nutshell:
- the loader does not have to live in redux at all.
  - before https://github.com/trezor/trezor-suite/pull/19577/commits/40b92d2ad53afa2fd62b3f44f06fcc292970cab4 it was local in useRefetchFees but that does not carry the information to components upstream where I need it
    - for example to indicate loading on the Review & Send button
- but it can be done with a new React Context. It's reusable, avoids modifying global redux logic
- mobile shall have the same functionality later, so where to put it so it can be shared?
  - `wallet-utils` sounds good, there is already some react in it

rojcyk:
> I don't agree with closing and not resolving this #19282 issue that the amount changes under my hands @hynek-jina @jirih-stsh @Hermez-cz @rojcyk @MiroslavProchazka
> 
> I believe it definitely breaks some nielsen norman group rules.
> 
> I believe this is not a good argument here.
> 
> > Don't use send max if you need specific amount.
> 
> * some exchanges have minimum for deposit, in case it changes in the last second, user can lose funds
> * in case user is using fixed float, or something like that, he can copy the amount from the send form. Then he sends incorrect amount -> lose of funds or troubles on support
> * ...
> 
> Easiest solution is just to not update fee when user is sending max 🤷‍♂️ or show some warning that amount changed.

Yeah, I agree with @tomasklim here. 

- From the ease-of-use POV it definitely makes sense that the value recalculates and updates automatically. It wouldn't make sense to use the feature without it.
- At the same time, users need to be aware that the value has changed, and there definitely needs to be some UI informing the user about the change. I'm sure we have plenty of users who are not aware of the second degree effects when using send max.

hynek-jina:
Fair point.. I've re-openned the mentioned issue:
https://github.com/trezor/trezor-suite/issues/19282#issuecomment-2975793668

packages/suite/src/components/wallet/Fees/Fees.tsx
```diff
@@ -158,9 +156,9 @@ export const Fees = <TFieldValues extends FormState>({
 
     const supportsCustomFee = networkType !== 'solana';
 
-    useEffect(() => {
-        dispatch(updateFeeInfoThunk({ networkSymbol: symbol }));
-    }, [dispatch, symbol]);
+    useFetchFeesOnce({ networkSymbol: symbol });
+    const isRefetchDisabled = getValues().setMaxOutputId !== undefined;
+    useRefetchFees({ networkSymbol: symbol, isDisabled: isRefetchDisabled });
```
Lemonexe, line 162: I uncoupled `useFetchFeesOnce` from `useRefetchFees`, it's nicer to have it atomic then one hook that does everything.. And I actually do need to disable only refetch, not initial fetch :slightly_smiling_face:

suite-common/wallet-core/src/fees/feesActions.ts
```diff
@@ -1,25 +1,17 @@
...
 export const FEES_MODULE_PREFIX = '@common/wallet-core/fees';
 
 const updateFee = createAction(
     `${FEES_MODULE_PREFIX}/updateFee`,
-    (payload: Partial<NetworksFees>) => ({
-        payload,
-    }),
-);
-
-const removeFee = createAction(
```
Lemonexe, line 15: this action was never used, so let's not support it. I don't see any case for it.
When you disable coins that's when this action maybe _could_ be used, but IMO it's useless, fees can stay in redux, no problem.

suite-common/wallet-core/src/send/sendFormThunks.ts
Lemonexe: Refactored. As part of #19721, fee object per network now has status.

And I think now I handled it quite elegantly with optional parameter, it ensures that loader is correctly displayed either during fetching itself (no parameter) or during the delayed time

packages/suite/src/views/wallet/send/TotalSent/TotalSent.tsx
Lemonexe: Done in https://github.com/trezor/trezor-suite/commit/95a4f67eb7a13d068cbfc91b04859894f7e0229b

If there are more UI tweaks to do, let's do it in separete PR; the scope of this one has blown up enough  :pray:

packages/suite/src/components/wallet/Fees/StandardFee/FeeCard.tsx
Lemonexe: What radio? In Send Form there is this, so I guess you mean another form?
image

packages/utils/src/isApproxEqual.ts
Lemonexe: Renamed in https://github.com/trezor/trezor-suite/commit/60cf4fa3099c17352efc28788cccfc051ec40512

Lemonexe:
Note: E2E tests ran manaully
[web](https://github.com/trezor/trezor-suite/actions/runs/15858279020)
[desktop](https://github.com/trezor/trezor-suite/actions/runs/15858277364)
because for some reason they do not show anymore (usually they are skipped so you can rerun them manually to actually start)

packages/suite/src/components/wallet/Fees/StandardFee/FeeCard.tsx
tomasklim: Screenshot 2025-06-25 at 7 51 54

Lemonexe: Oh sry I was looking at completely different component. Yes, this is how I understand [the figma specs](https://www.figma.com/design/qx5aOavcU7I2MW2PtnQtfz/Small-Things--DSK-?node-id=4048-978&t=r44B2LHJ4sIs9Nsh-0)

tomasklim (approved):

PR #19580, opened 2025-06-14
chore(env-utils): cleanup unused or duplicated env utils

Lemonexe:
## Description

Remove a deprecated env-util

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/cleanup-env-utils&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/cleanup-env-utils&lookback=7)

### DIFF
--- a/packages/env-utils/src/envUtils.native.ts
+++ b/packages/env-utils/src/envUtils.native.ts
@@ -26,7 +26,6 @@ const getDeviceType = () => '';
 
 // no need for async fn on native; only for the sake of consistency with Web/Desktop
 const getOsVersion = () => Promise.resolve(`${Platform.Version}`);
-const getDeprecatedOsVersion = () => `${Platform.Version}`;
 
 const getCpuArch = () => Promise.resolve('');
 
@@ -95,7 +94,6 @@ export const envUtils: EnvUtils = {
     getCommitHash,
     getDeviceType,
     getOsVersion,
-    getDeprecatedOsVersion,
     getCpuArch,
     getSuiteVersion,
     isFirefox,
--- a/packages/env-utils/src/envUtils.ts
+++ b/packages/env-utils/src/envUtils.ts
@@ -47,8 +47,6 @@ const getOsVersion = async () => {
 
     return version ?? '';
 };
-/** @deprecated: Use the async getOsVersion instead. */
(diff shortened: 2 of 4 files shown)


packages/env-utils/src/envUtils.ts
```diff
@@ -127,8 +125,6 @@ const getOsName = () => {
     return '';
 };
 
-const getOsNameWeb = () => getUserAgentParser().getOS().name?.replaceAll(' ', '');
```
Lemonexe, line 130: `getOsName` uses a different information source, it could work differently, but it's only used in the browser detection page only for iOS and I tested that on iOS :heavy_check_mark:

packages/env-utils/src/envUtils.native.ts
```diff
@@ -76,7 +75,7 @@ const getOsName = () => {
     return '';
 };
 
-const getOsNameWeb = () => '';
+const getOsNameWeb = () => 'iOS';
```
Lemonexe: consistency with the actual return for Suite Web

packages/env-utils/src/envUtils.ts
```diff
@@ -127,6 +125,7 @@ const getOsName = () => {
     return '';
 };
 
+// generally works the same as `getOsName`, just with different information source, but does not work in some specific iOS cases
```
Lemonexe, line 128: I tried to remove it in favor of `getOsName`. It is only used in [browser-detection only for iOS](https://github.com/trezor/trezor-suite/blob/75cc832eb3a3f54af4b298288e8831b8301232a1/packages/suite-data/src/browser-detection/index.ts#L225), nowhere else, and it worked when I tested it on an actual iphone, but fails in [E2E tests](https://github.com/trezor/trezor-suite/blob/b96c899ebbaa7d9b9c324356202758d5e5927630/packages/suite-desktop-core/e2e/tests/browser/ios.test.ts).. So let's keep it.

peter-sanderson (approved):

packages/env-utils/src/envUtils.native.ts
Lemonexe: Nah this is stupid, native can be Android too! :facepalm: 
→ reverted

PR #19589, opened 2025-06-16
fix(suite-native): replace selectConfiguration in native USB transport with NOOP

### LINKED ISSUE #18743: java.lang.ArrayIndexOutOfBoundsException in ReactNativeUsb.selectConfiguration
When a Trezor is connected via USB to an Android phone, the following error is logged:

```
 ERROR  usb: device.selectConfiguration error Error: Call to function 'ReactNativeUsb.selectConfiguration' has been rejected.
→ Caused by: java.lang.ArrayIndexOutOfBoundsException: length=1; index=1. device: {"productName":"TREZOR","manufacturerName":"SatoshiLabs","serialNumber":"5CEF44D0C198C85354855511","vendorId":4617,"productId":21441,"deviceVersionMajor":1,"deviceVersionMinor":0,"opened":true}
```

(To ensure you see the error, instantiate `UsbApi` in `NativeUsbTransport` with `logger: console`.)

The reason for the failure is that `UsbApi` (`packages/transport/src/api/usb.ts`) calls `selectConfiguration` with hardcoded value `CONFIGURATION_ID = 1`. And since https://developer.android.com/reference/android/hardware/usb/UsbDevice#getConfiguration(int) expects a zero-based index, `java.lang.ArrayIndexOutOfBoundsException` is thrown.

What is the reason for a hard-coded ID?
Is the native implementation expected to convert a 1-based index to 0-based index?
Could `selectConfiguration` in the native implementation be change to NOOP since USB connection works even if the method fails?

yanascz:
Make sure to initialize `NativeUsbTransport` with `logger: console` for testing.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18743

### DIFF
--- a/packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
+++ b/packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
@@ -80,10 +80,6 @@ class ReactNativeUsbModule : Module() {
             return@AsyncFunction closeDevice(deviceName)
         }
 
-        AsyncFunction("selectConfiguration") { deviceName: String, configurationIndex: Int ->
-            return@AsyncFunction selectConfiguration(deviceName, configurationIndex)
-        }
-
         AsyncFunction("claimInterface") { deviceName: String, interfaceNumber: Int ->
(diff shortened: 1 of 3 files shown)


matejkriz (approved):
It works as expected and makes sense - selectConfiguration is not needed at all in our use-case.

PR #19596, opened 2025-06-16
fix: lock auto-conneect during discovery, to prevent 'another call inprogress on device' error

peter-sanderson:
- When device is doing discovery, some it may be locked, so not other calls can be performed. 
- Respect this lock for auto-connect modal.

image

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix-lock-autoconnect&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix-lock-autoconnect&lookback=7)


Lemonexe (approved):
LGTM, though there could be a better indication.
Loader does not make it clear that you should shift your attention to the device.
If there is no design for this situation, you can let it be, but I think a a low hanging fruit would be to add a tooltip, maybe reuse "Connect & unlock your Trezor" copy?

PR #19625, opened 2025-06-17
refactor(suite-native): test-utils as dev dependency only

PeKne:
`@suite-native/test-utils` package was imported as main dependency at some places even though that it should be `devDependency` only (it is necessary only for tests). This PR fixes that.

Thanks for the edu session BTW ;)
 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/test-utils-as-dev-dependency&lookback=7)


jbazant (approved):
Good spot!

PR #19662, opened 2025-06-19
feat(trading): update amount for paymentRequest and tx preview modal

### LINKED ISSUE #19506: Trading - update trading fees for modal tx preview and payment requests
## Description
- update trading fees for modal tx preview and payment requests according to outputs

adderpositive:
## Description
- update the amount for the payment requests and the tx preview modal according to outputs

### Note
Initially, there was an idea to make fees persistent during Swap and Sell operations. However, this approach runs into issues due to the variable transaction size on Bitcoin-like networks:
1. Initial fee estimation is based on the assumption that a legacy address will be used, which results in calculating the maximum possible transaction size, because we do not know the address of the provider.
2. The actual recipient address (which could be SegWit or Native SegWit) is only known after the trade is confirmed. This affects the transaction size, as SegWit addresses typically result in smaller transactions.
3. This discrepancy leads to a mismatch between the expected trade amount (calculated at the beginning) and the final payment request, particularly for SLIP-24 compliance.

There is also the possibility to pad some additional bytes to balance the bytes between the initial fees and the sending transaction fees.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19506

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/update-trading-fees&lookback=7)

### DIFF
--- a/suite-common/trading/src/thunks/common/__tests__/createPaymentRequestsThunk.test.ts
+++ b/suite-common/trading/src/thunks/common/__tests__/createPaymentRequestsThunk.test.ts
@@ -259,6 +259,7 @@ describe('createPaymentRequestsThunk', () => {
                     type: 'exchange',
                     account: mockAccount,
                     composedLevels: mockComposedTransaction,
+                    formattedMaxAmount: mockExchangeQuote.sendStringAmount,
                 }),
             );
 
@@ -281,6 +282,7 @@ describe('createPaymentRequestsThunk', () => {
                     type: 'exchange',
                     account: mockAccount,
                     composedLevels: mockComposedTransaction,
+                    formattedMaxAmount: mockExchangeQuote.sendStringAmount,
                 }),
             );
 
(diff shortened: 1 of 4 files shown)


LukasRada (changes requested):
Can we support the change with unit test case?

packages/connect/src/data/config.ts
```diff
@@ -268,7 +268,7 @@ export const config = {
         {
             capabilities: ['slip24'],
             methods: ['getNonce'],
-            min: { T1B1: '0', T2T1: '2.9.1', T2B1: '2.9.1', T3B1: '2.9.1', T3T1: '2.9.1' },
+            min: { T1B1: '0', T2T1: '2.8.10', T2B1: '2.8.10', T3B1: '2.8.10', T3T1: '2.8.10' },
             comment: ['Since firmware 2.9.1 SLIP-24 is supported'],
```
tomasklim: wrong comment after change

tomasklim: aha, only for testing 😀

tomasklim (approved):

PR #19588, opened 2025-06-16
debugging discovery issue #19411 - repeated passphrase request

mroz22:
resolve #19411

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=discovery-nitpicks-12&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=discovery-nitpicks-12&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=discovery-nitpicks-12&lookback=7)

PR #19675, opened 2025-06-20
fix(suite): get rid of repeated passphrase requests for saved devices

matejkriz:
## Description

What about - as an ugly hotfix - updating state on every device change if the device is remembered and using passphrase? 

It seems to work fine on mobile. Don't know if extra device state change can break something 🤔 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/discovery-nitpicks-12-hotfix&lookback=7)


mroz22 (approved):

matejkriz:
merged to https://github.com/trezor/trezor-suite/pull/19588

PR #19683, opened 2025-06-20
fix(suite-native): submit coin enabling changes on button or leaving the screen

### LINKED ISSUE #19471: Quickly toggling network on/off discovers accounts for disabled network
**Describe the bug**
Accounts are discovered even if the network has been disabled since discovery start

**Info:**

- Mobile Suite Version: version/commit
- OS: android, develop as of b18d1e58

**How to reproduce**

1. Go to Enabled coins settings
1. Choose coin that is off, and quickly toggle it on and off again
1. Go Home
1. See accounts discovered for that coin, which is now off

### LINKED ISSUE #19628: Discovery should start only after leaving Coin Enabled screen
Discovery of newly enabled coins should not start before the user leaves the Coin Enabled screen in Settings. 

It used to work like that till 25.6.1. Now, the discovery of the newly enabled coin started right after clicking on the toggle on the Coin Enabling screen. 

It leads to an unresponsive feel from the toggle, and on a remembered passphrase wallet, it triggers the passphrase request too soon.

matejkriz:
## Description

- If we change enabled networks directly by clicking the toggle, the app can became easily unresponsive (new discovery is triggered) and toggle states are lagging.
- But the reason for this change was remembered passhrase wallet - it's not nice to ask user for passhprase at the moment he click the coin toggle.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19628
Resolve https://github.com/trezor/trezor-suite/issues/19471

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/coin-enabling-form&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/native/coin-enabling-form&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/native/coin-enabling-form&lookback=7)

### DIFF
--- a/suite-native/coin-enabling/src/components/CoinEnablingForm.tsx
+++ b/suite-native/coin-enabling/src/components/CoinEnablingForm.tsx
@@ -1,0 +1,86 @@
+import { useCallback, useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { useNavigation } from '@react-navigation/native';
+
+import { changeCoinVisibility } from '@suite-common/wallet-core';
+import { useAlert } from '@suite-native/alerts';
+import { EventType, analytics } from '@suite-native/analytics';
+import {
+    selectDeviceEnabledDiscoveryNetworkSymbols,
+    selectDiscoveryNetworkSymbols,
+} from '@suite-native/discovery';
+import { Form, useForm } from '@suite-native/forms';
+import { Translation } from '@suite-native/intl';
+
+import { DiscoveryCoinsFilter } from './DiscoveryCoinsFilter';
+import { CoinEnablingFormValues, coinEnablingFormValidationSchema } from '../coinEnablingSchema';
+
+export const CoinEnablingForm = () => {
+    const dispatch = useDispatch();
+    const navigation = useNavigation();
+
+    const enabledNetworkSymbols = useSelector(selectDeviceEnabledDiscoveryNetworkSymbols);
+    const networkSymbols = useSelector(selectDiscoveryNetworkSymbols);
+
+    const { showAlert } = useAlert();
+
+    const showLastNetworkAlert = useCallback(
+        () =>
+            showAlert({
+                title: <Translation id="moduleSettings.coinEnabling.oneNetworkSymbolAlert.title" />,
+                description: (
+                    <Translation id="moduleSettings.coinEnabling.oneNetworkSymbolAlert.description" />
+                ),
+                primaryButtonTitle: <Translation id="generic.buttons.gotIt" />,
+                primaryButtonVariant: 'redBold',
+            }),
+        [showAlert],
+    );
+
+    const form = useForm<CoinEnablingFormValues>({
+        defaultValues: {
+            enabledCoins: enabledNetworkSymbols,
+        },
+        validation: coinEnablingFormValidationSchema,
+    });
+
+    const handleSubmit = form.handleSubmit(values => {
+        const changedCoins = networkSymbols.filter(
+            symbol =>
+                enabledNetworkSymbols.includes(symbol) !== values.enabledCoins.includes(symbol),
+        );
+
+        if (changedCoins.length === 0) return;
+
+        changedCoins.forEach(symbol => {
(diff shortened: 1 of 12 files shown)


suite-native/coin-enabling/src/components/CoinEnablingForm.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    useEffect(
+        () =>
+            navigation.addListener('beforeRemove', () => {
+                if (form.formState.isDirty) {
+                    handleSubmit();
+                }
+            }),
+        [navigation, form.formState.isDirty, handleSubmit],
+    );
```
juriczech: image

Reference: https://react-hook-form.com/docs/useform/formstate

Ideally, destructure from form state like so: 
```
    const {
        handleSubmit,
        formState: { isDirty },
    } = useFormContext();
```

suite-native/coin-enabling/src/components/CoinEnablingForm.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    const filterProps = useMemo<ComponentProps<typeof DiscoveryCoinsFilter>>(
+        () => ({
+            allowDeselectLastCoin: false,
```
juriczech: This is unused if I'm not overlooking something. IMO, this const is not necessary and you could use a selector in `DiscoveryCoinsFilter.tsx` and only send `onDisablingLastCoin` as an optional prop.

juriczech: Or if its necessary due to rerendering, it's not super obvious from the code so maybe a comment might help?

juriczech (commented):
FYI: I've only reviewed the code, will test it later today 🙏

suite-native/coin-enabling/src/components/CoinEnablingForm.tsx
matejkriz: Thanks for the catch, I overlooked it during rebase, it is indeed not used anymore, since I only check if `onDisablingLastCoin` is defined.

3cb869e41fabe4aaca99d14bc7912181ad4a173f

juriczech: Question, is it necessary to create memoization for this and the callback with alert? If not, it's only making things complicated and I try to memoize only when it makes sense. 

It's not super important, just curious as to why it's there. If it should be memoized, the function who calls the callback should be memoized too, right? If it's not, I believe it's not doing anything useful...

Feel free to resolve not to waste time on unnecessary nitpicks. 😆

suite-native/coin-enabling/src/components/CoinEnablingForm.tsx
matejkriz: Thanks, I wasn't aware of that at all! I think the the check for dirtiness is redundant here anyway, checking for changed networks in handleSubmit should be cheap, so I removed the condition completely to simplify it.

d12f2e592cf8d15aae3d9f3cff3a8892be99ee55

suite-native/coin-enabling/src/components/CoinEnablingForm.tsx
juriczech: Edit: Now that I'm thinking about it, it is not necessary to have the toggle function memoized. It would be useless if it was the other way around.

matejkriz: No, the whole filterProps const is unnecesary complexity, thanks for pointing that out. But the callback makes sense to me since it's passed to child component.

1ec9d13a61c3f7cea651445e9b7158993ee47b4e

juriczech (approved):
LGTM 🥳 

Only thing I noticed during testing is that coin visibility doesn't update if you close the app. If it's only to background, form is persisted and when you leave the screen, it gets saved, but if you kill the app, the changes are not saved. Probably an edge case but could be easily implemented, if it's a bug.

PR #19696, opened 2025-06-23
fix(transport-bridge): allow read while call is running

szymonlesisz:
## Description

in DeviceCurrentSession file we have this comment which in my observation is not true or maybe it was true at the time writing.
node-bridge throws exacly the same error when there is a pending call (ButtonAck) and Cancel is sent 

```
/**
*Bridge version =< 2.0.28 throws "other call in progress" error....
```

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/bridge-call-in-progress&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/bridge-call-in-progress&lookback=7)


szymonlesisz:
turns out i was wrong about it, node-bridge does not check locks on `send` as @marekrjpolak pointed and tested :+1:

PR #19700, opened 2025-06-23
chore(suite): use suite version for storage

### LINKED ISSUE #19646: Start using Suite version for storage
Very often developers created more versions of database per one version of Suite. Let's make it easier and start using Suite versioning for database.

In case it is done before 25.7.0 release, you can change version `57` to version `25.7...` 

It has to be backwards compatible.

matusbalascak:
This PR aims to improve versioning and migrations for Suite IDB. 

## Description

There's a simple encoding used to transform semver *<major.minor.patch>* into a number that can be used as a version for IDB. Detailed explanation is in the `MIGRATION.md` file that is part of this PR.

In short, it transforms 25.7.0 into a 24-bit number where each 8 bits represent a part of the version (25 = 8 bits, 7 = 8 bits, 0 = 8 bits). The numbers are packed together creating a version.

So decimal value for `25.7.0`  would be `1 640 192`. That ensures that we don't break backwards compatibility with this change as the old versioning system stopped at `56`.

Every migration now has its own separate file. This makes testing migrations much easier as they can be tested in isolation.

There's also a scripts that helps you with scaffolding a new migration, just run:

```bash
yarn workspace @trezor/suite make:migration <major.minor.patch>
```

It creates a migration file, exports it and adds a new entry to the changelog. More info in `MIGRATION.md`

## Related Issue

Resolve [#19646](https://github.com/trezor/trezor-suite/issues/19646)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feature/suite-version-in-storage&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feature/suite-version-in-storage&lookback=7)


packages/suite/src/storage/index.ts
```diff
@@ -1,11 +1,30 @@
...
+const MIGRATIONS = Object.values(migrations).sort((a, b) => a.threshold - b.threshold);
+
+const runMigrations = async (
+    db: IDBPDatabase<SuiteDBSchema>,
+    oldVersion: number,
+    tx: IDBPTransaction<SuiteDBSchema, StoreNames<SuiteDBSchema>[], 'versionchange'>,
+) => {
+    for (const migration of MIGRATIONS) {
+        if (oldVersion < migration.threshold) {
+            console.log(`Running migration for version ${idbVersionToSemver(migration.threshold)}`);
```
tomasklim: In case a migration will be done, please also add console log of current version of DB

```
if (MIGRATIONS.length) {
	console.log(...)
	
	for (const migration of MIGRATIONS) {
		....
	}
}
```

tomasklim (commented):
LGTM, few comments

1. it would be great to have a separate package for this. On root in folder `suite-web` (does not exist yet) https://www.notion.so/satoshilabs/Code-improvements-9-breaking-the-monolith-528bf80c34ba4a6bba3f335e60ff29e8?source=copy_link#899b751a5cad4534b376d59c7d837998 (new folder)

2. ~In case we would need to introduce new migration for new RC (patch version). We will not be able to cherry pick it from develop as there will already be bumped minor version~ Solvable

scripts/createSuiteIDBMigration.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export default createMigration("${version}", (db, tx) => {
+  // ⚠️  TODO: implement migration logic for v${version}
+})
+`;
+
+    fs.writeFileSync(filePath, template, 'utf8');
+    console.log(`✅ Created ${filePath}`);
+
+    const exportLine = `export { default as ${exportIdentifier} } from './${version}';\n`;
```
tomasklim:
```suggestion
export default createMigration("${semver.version}", (db, tx) => {
  // ⚠️  TODO: implement migration logic for v${semver.version}
})
`;

    fs.writeFileSync(filePath, template, 'utf8');
    console.log(`✅ Created ${filePath}`);

    const exportLine = `export { default as ${exportIdentifier} } from './${semver.version}';\n`;
```

packages/suite/package.json
```diff
@@ -13,6 +13,7 @@
         "translations:backport-en": "yarn g:tsx ../suite-data/src/translations/backport-en.ts && yarn g:prettier --write \"./src/support/messages.ts\"",
         "translations:list-duplicates": "yarn g:tsx ../suite-data/src/translations/list-duplicates.ts",
         "translations:list-unused": "yarn g:tsx ../suite-data/src/translations/list-unused.ts",
+        "make:migration": "yarn g:tsx ../../scripts/createSuiteIDBMigration.ts",
```
tomasklim: `yarn workspace @trezor/suite make:migration <x.y.z>`

in case i would call this without <x.y.z>, maybe it should use default `suiteVersion`?

package.json
```diff
@@ -14,6 +14,7 @@
             "packages/connect-examples/*",
             "suite-native/*",
             "suite-common/*",
+            "suite-web/*",
```
Lemonexe: do you think it makes sense to create a new namespace just for this?
- IMO `idb-migrations` workspace can live in `packages`, as do other packages which are shared by Web and Desktop Suite but not Lite.
- And that leads me to the point that `suite-web` is not precise, because `idb` is used in Desktop as well (the electron renderer process also utilizes it).
- Also, I'd name it `idb-migration-utils`, because it does not contain migrations themselves. I think it fits among the many `packages/*-utils` that we have

suite-web/idb-migrations/MIGRATION.md
```diff
@@ -0,0 +1,69 @@
...
+### Layout
+
+| Bit range | Stored value |
+| --------- | ------------ |
+| 23 – 16   | `major`      |
+| 15 – 8    | `minor`      |
+| 7 – 0     | `patch`      |
+
+```typescript
+encoded = (major << 16) | (minor << 8) | patch;
```
Lemonexe: :thought_balloon:  this is an elegant way to do it :+1:
When I was thinking about this issue in past, my thought was to use `semver` for comparison of SemVer versions, but then the interoperability with old integer versioning would require logic that would have to be maintained forever (not difficult, but potential source of bugs).
This is a great way to avoid it  :+1:

suite-web/idb-migrations/MIGRATION.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+### Examples
+
+| SemVer   | Encoded (hex) | Encoded (dec) |
+| -------- | ------------- | ------------- |
+| `1.0.0`  | `0x010000`    | 65 536        |
+| `2.5.3`  | `0x020503`    | 132 611       |
+| `25.7.0` | `0x190700`    | 1 640 192     |
```
Lemonexe: for us slower ones, this will be more illustrative than just the footnote alone :pray:
```suggestion
| `25.7.0` | `0x190700`    | 1 640 192     |
| `255.255.255` | `0xFFFFFF`    | 16 777 215     |
```

packages/suite/src/storage/index.ts
```diff
@@ -15,15 +40,18 @@ const onUpgrade: OnUpgradeFunc<SuiteDBSchema> = async (db, oldVersion, newVersio
...
+            SuiteDB.removeStores(db);
         } catch (err) {
             console.error('Storage: Error during removing all stores', err);
             throw err;
         }
     }
 
-    // migrate functions
-    await migrate(db, oldVersion, newVersion, transaction);
+    if (oldVersion < 57) {
```
Lemonexe: to make it really clear I would extract this to a local constant
```suggestion
    if (oldVersion < LAST_LEGACY_VERSION) {
```

packages/suite/src/storage/migrations/index.ts
```diff
@@ -1283,10 +1283,4 @@ export const migrate: OnUpgradeFunc<SuiteDBSchema> = async (
     if (oldVersion < 56) {
         await migrateToV56(db, oldVersion, newVersion, transaction);
     }
-
-    if (oldVersion < 57) {
-        db.createObjectStore('bioAuth');
-        // @ts-expect-error
-        db.deleteObjectStore('discovery');
-    }
```
Lemonexe, line 1291: as a foolproof measure let's add a comment with nasty warning here, something like this.
People are used to the old system, so I'm afraid someone could go straight for adding a new legacy condition. 
Nothing prevents me from doing so (don't see how it'd be possible with TS), so at least a banner pls :slightly_smiling_face: 
```
    /*
     !!! DO NOT ADD ANY MORE MIGRATION CODE BELOW !!!

     These are legacy migrations, instead follow the instructions in
     idb-migrations/MIGRATION.md
    */
```

packages/suite/src/storage/migrations/versions/index.ts
```diff
@@ -0,0 +1 @@
+export { default as m25_7_0 } from './25.7.0';
```
Lemonexe, line 6: `packages/suite/src/storage/index.ts` relies on the fact that _all migrations_ are reexported in this barrel file when it does `import * as migrations`.
I know that the automatic script adds it, but it's not clear that this is autogenerated file.
As a foolproof measure, pls let's prepend the usual warning for autogen files:
```
/*
 !!! IMPORTANT: This file is autogenerated !!!
 make:migration script should be used to add migrations here.
*/
```

packages/suite/src/storage/migrations/index.ts
```diff
@@ -1283,10 +1283,4 @@ export const migrate: OnUpgradeFunc<SuiteDBSchema> = async (
     if (oldVersion < 56) {
         await migrateToV56(db, oldVersion, newVersion, transaction);
```
Lemonexe, line 1284: Could you please move the `migrateToV56` file from `./versions` to `./legacyVersions`?
I know you didn't touch that file, but you built the new migrations around it, and it's not clear at first glance that v56 doesn't belong with the new migrations :slightly_smiling_face:

suite-web/idb-migrations/src/encode.ts
```diff
@@ -0,0 +1,33 @@
+import * as semver from 'semver';
+
+export const semverToIDBVersion = (_version: string | undefined): number => {
```
Lemonexe: nit: rename `_version` to `rawVersion` or `versionString` or such.
It's a convention that params starting with `_` are unused params.

packages/suite/src/storage/migrations/versions/25.7.0.ts
```diff
@@ -0,0 +1,9 @@
+import { createMigration } from '@suite-web/idb-migrations';
+
+import { SuiteDBSchema } from 'src/storage/definitions';
+
+export default createMigration<SuiteDBSchema>('25.7.0', db => {
+    db.createObjectStore('bioAuth');
+    // @ts-expect-error
+    db.deleteObjectStore('discovery');
```
Lemonexe: These throw error when coming from develop, but that's ok.
That affects only developers :sweat_smile: 

Though if you want, you can make it idempotent, but up to you
```
    if (!db.objectStoreNames.contains('bioAuth')) {
        db.createObjectStore('bioAuth');
    }
    // @ts-expect-error
    if (db.objectStoreNames.contains('discovery')) {
        // @ts-expect-error
        db.deleteObjectStore('discovery');
    }
```

suite-web/idb-migrations/scripts/makeMigration.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    fs.writeFileSync(filePath, template, 'utf8');
+    console.log(`✅ Created ${filePath}`);
+
+    const exportLine = `export { default as ${exportIdentifier} } from './${semver.version}';\n`;
+    fs.appendFileSync(EXPORT_PATH, exportLine, 'utf8');
+    console.log(`✅ Appended export to ${EXPORT_PATH}`);
+}
+
+function updateChangelog(semver: SemVer) {
```
Lemonexe: nice touch with updating changelog.
Devs incl. me often forget to update it, but having an unstaged git diff is a good reminder to fill it before commiting :+1:

Lemonexe (commented):
Great work, thanks for the initiative!

Codewise LGTM overall, works well, but I have some comments:
- Mostly, I'd like to avoid creating a new namespace [as per this comment](https://github.com/trezor/trezor-suite/pull/19700#discussion_r2166332975) but I'd like to hear your thoughts on that :slightly_smiling_face: 
- Several of my comments are nits, and silly foolproof measures that may sound unnecessary, but I really believe that specifically for migrations code, it is warranted:
  - Different developers will be working on migrations for years to come, and they might not immediately understand the inner workings (or forget)
  - And mistakes in migrations have been costly in past (and still are)
  - So I suggest let's be as explicit as possible to reduce confusion opportunities
- Awesome `make:migration` script! :+1: 
  - tested, works as expected
    - ran at current state (migration already exists)
    - ran after incrementing to 25.8.0 in `suite/package.json` :heavy_check_mark:
      - here I also tested the migrations itself, the 25.8.0 migration script is called on first startup but not subsequently :heavy_check_mark: 
    - ran with arbitrary version param :heavy_check_mark:
- Great unit tests :+1:

package.json
Lemonexe: oh sry, only now I have noticed @tomasklim [comment](https://github.com/trezor/trezor-suite/pull/19700#pullrequestreview-2953761994)
Well ok, `suite-web` namespace makes sense as part of a wider reorg, but this workspace is not web-only

matusbalascak: Resolved in slack, we'll go with the name `suite/` instead of `suite-web/` and `idb-migrations` wlil be renamed to `idb-migration-utils` as suggested

tomasklim (approved):
Awesome!

Lemonexe (approved):
Thanks for implementing my nitpick comments :relieved: 
Looks LGTM to me :rocket:

PR #19751, opened 2025-06-25
feat(suite-native): firmware settings

### LINKED ISSUE #19711: Device settings firmware flow
In https://github.com/trezor/trezor-suite/pull/19638, firmware screen was refactored slightly but still needs couple of things that were left out to avoid too big of a PR:
- alert boxes above firmware version cards

Design: https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Settings--MOB-?node-id=1663-16667&t=2PRSVfc01VgxKpOK-4

juriczech:
## Description
- add alert for up to date / upgrade states in firmware screen
- handle edge case for when firmware update is not enabled (through message system) - simple toast message and disable user from accessing firmware update screen

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19711

## Screenshots:
Snímek obrazovky 2025-06-26 v 9 23 14
Snímek obrazovky 2025-06-26 v 9 23 21
Snímek obrazovky 2025-06-26 v 9 24 33
Snímek obrazovky 2025-06-26 v 9 24 39

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/native/device-settings-firmware&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/native/device-settings-firmware&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/native/device-settings-firmware&lookback=7)

### DIFF
--- a/suite-native/firmware/src/components/ConfirmFirmwareUpdateScreenContent.tsx
+++ b/suite-native/firmware/src/components/ConfirmFirmwareUpdateScreenContent.tsx
@@ -1,22 +1,87 @@
-import { Text, VStack } from '@suite-native/atoms';
-import { Translation } from '@suite-native/intl';
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+import {
+    selectDeviceFirmwareVersion,
+    selectDeviceUpdateFirmwareVersion,
+    selectHasBitcoinOnlyFirmware,
+    selectIsFirmwareUpgradable,
+} from '@suite-common/wallet-core';
+import { InlineAlertBox, Text, VStack } from '@suite-native/atoms';
+import { Translation, useTranslate } from '@suite-native/intl';
 
 import { FirmwareChangelogButton } from '../components/FirmwareChangelogButton';
-import { FirmwareUpdateVersionCard } from '../components/FirmwareVersionCard';
+import {
+    FirmwareUpdateVersionCard,
+    FirmwareVersionCard,
+    concatFirmwareVersion,
+} from '../components/FirmwareVersionCard';
 
-export const ConfirmFirmwareUpdateScreenContent = () => (
-    <VStack spacing="sp32">
-        <VStack>
-            <Text variant="titleMedium">
-                <Translation id="firmware.firmwareUpdateScreen.title" />
-            </Text>
-            <Text variant="body" color="textSubdued">
-                <Translation id="firmware.firmwareUpdateScreen.subtitle" />
-            </Text>
+const UpgradableContent = () => {
+    const updateFirmwareVersion = useSelector(selectDeviceUpdateFirmwareVersion);
+
+    return (
+        <>
+            <InlineAlertBox
+                title={
+                    <Translation
+                        id="firmware.updateCard.updateToVersionAvailable"
+                        values={{ firmwareVersion: updateFirmwareVersion }}
+                    />
+                }
+                variant="info"
+            />
+            <VStack>
+                <FirmwareUpdateVersionCard />
+                <FirmwareChangelogButton />
+            </VStack>
+        </>
+    );
+};
+
+const UpToDateContent = () => {
+    const firmwareVersion = useSelector(selectDeviceFirmwareVersion);
+    const isBtcOnly = useSelector(selectHasBitcoinOnlyFirmware);
+    const { translate } = useTranslate();
+
+    const firmwareTypeTranslationId = isBtcOnly
+        ? 'firmware.typeBitcoinOnly'
+        : 'firmware.typeUniversal';
+
+    return (
+        <>
+            <InlineAlertBox
+                title={<Translation id="firmware.updateCard.upToDate" />}
+                variant="success"
+            />
+            <FirmwareVersionCard
+                title={<Translation id="firmware.firmwareUpdateScreen.currentFirmware" />}
+                titleColor="textSubdued"
+                version={concatFirmwareVersion(firmwareVersion) ?? null}
+                fwType={translate(firmwareTypeTranslationId)}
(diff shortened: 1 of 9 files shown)


suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
```diff
@@ -44,6 +48,16 @@ export const DeviceFirmwareCard = () => {
     const firmwareVersion = getFirmwareVersion(device);
 
     const handleOnPress = () => {
+        if (!isFirmwareUpdateEnabled) {
+            showToast({
+                variant: 'warning',
+                message: <Translation id="firmware.updateNotAvailable" />,
```
PeKne, line 54: nit: With this you are setting non serializable ReactNode to the jotai state. I guess it will work, just something we should be cautious. Maybe let's use `useTranslate` in this case?

suite-native/intl/src/en.ts
```diff
@@ -2035,6 +2035,7 @@ export const en = {
         version: 'Version {firmwareVersion}',
         typeUniversal: 'Universal',
         typeBitcoinOnly: 'Bitcoin-only',
+        updateNotAvailable: 'Firmware update disabled',
```
PeKne, line 2038: nit: Where does this copy come from? Does not sound very user friendly. As a user I would start looking for a way how to "enable" it.

suite-native/firmware/src/components/ConfirmFirmwareUpdateScreenContent.tsx
```diff
@@ -1,22 +1,83 @@
...
+    return (
+        <>
+            <InlineAlertBox
+                title={<Translation id="firmware.updateCard.upToDate" />}
+                variant="success"
+            />
+            <FirmwareVersionCard
+                title="Current firmware"
+                titleColor="textSubdued"
+                version={firmwareVersion?.join('.') ?? '?.?.?'}
```
PeKne: nit: this nullish coalescing operator (`?? '?.?.?'`) is now exactly like this used at three places. Maybe it could be moved inside of the `FirmwareVersionCard` implementation to handle that in case that `version` prop is `null`?

suite-native/firmware/src/components/ConfirmFirmwareUpdateScreenContent.tsx
```diff
@@ -1,22 +1,83 @@
...
+        : 'firmware.typeUniversal';
+
+    return (
+        <>
+            <InlineAlertBox
+                title={<Translation id="firmware.updateCard.upToDate" />}
+                variant="success"
+            />
+            <FirmwareVersionCard
+                title="Current firmware"
```
PeKne: hardcoded string. I know this was already in place, but fix it please anyway.

PeKne (approved):
Some nits found, but nothing serious. Feel free to pick only those that make sense to you.

suite-native/module-device-settings/src/components/DeviceFirmwareCard.tsx
juriczech: We're doing it all over the app, so I wouldn't enforce it here TBH. If we want to enforce this, I suppose it shouldn't be typed as `ReactNode`. I will resolve this one, feel free to reopen if you strongly disagree.

suite-native/intl/src/en.ts
juriczech: This is an edge case in case the firmware update is disabled through message system. The user will in that case see a system message in the app. So this is just to "disable" the button in that case. I believe the button shouldn't be disabled with no message whatsoever but also the user will already have the system message there. 

I would not worry about this too much and if you see it as a problem, I would just remove the toast warning.

suite-native/firmware/src/components/ConfirmFirmwareUpdateScreenContent.tsx
juriczech: So the FirmwareVersionCard is used 3 times. 2 times with current version, 1 time with version to which the firmware can be updated. That means I don't want to move this logic to the component because it needs to be able to display other versions which might already be concatenated. 

So the join happens twice. I was thinking about this and decided it's not too important, but I will just create a helper function for it to reduce the repetition so that it only needs to be updated at one place in case of change.

suite-native/firmware/src/components/ConfirmFirmwareUpdateScreenContent.tsx
juriczech: https://github.com/trezor/trezor-suite/pull/19751/commits/9a797670a9901ed75685c3b0f1e9349090205c47

suite-native/firmware/src/components/ConfirmFirmwareUpdateScreenContent.tsx
juriczech: https://github.com/trezor/trezor-suite/pull/19751/commits/9a797670a9901ed75685c3b0f1e9349090205c47

PeKne: I was not talking about concatenating, i has to be done differently based on the incoming value per use case. But nullish coalescing operator operator might be moved (the `version` might have a default value). But this is total nit. Feel free to resolve.

juriczech: I misread it 🙈 Fixed, you're right!

PR #19823, opened 2025-06-30
chore(suite): nitpicks, renaming..

mroz22:
- tiny nitpick renaming. and also getting rid of 'bridge' in the codebase as a by product

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=nitpicks&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=nitpicks&lookback=7)

PR #19863, opened 2025-07-01
Switch bridge implementaion troubleshooting

mroz22:
based on #19823

- showing the "toggle bridge" troubleshooting tip not only for unreadable devices but also when no bridge is running.  I don't expect this much to happen. Even when you for example uninstall external bridge while suite-desktop is running, it should be able to detect it and turn on its onw bridge. Also I'd say that it would make sense to get rid of this screen completely and when suite-desktop can't start bridge for whatever reason, it should fallback to NodeUsbTransport. What do you think?
- image

- desktop + standalone bridge running. nothing interesting here. do we want to link status page similarly to web?
- image

- desktop + bundled bridge running. This is a escape hatch just in case some users have problems with the node-bridge. 
- image

- web, no bridge is running 
- image

- web, bridge running. note that I changed copy in the first bullet. no trezor-bridge mentions are present anymore.
- image

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=switch-bridge-implementaion-troubleshooting&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=switch-bridge-implementaion-troubleshooting&lookback=7)

### DIFF
--- a/packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
+++ b/packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
@@ -10,9 +10,11 @@ import {
     TROUBLESHOOTING_TIP_CABLE,
     TROUBLESHOOTING_TIP_DIFFERENT_COMPUTER,
     TROUBLESHOOTING_TIP_SUITE_DESKTOP,
+    TROUBLESHOOTING_TIP_SUITE_DESKTOP_TOGGLE_BRIDGE,
     TROUBLESHOOTING_TIP_UDEV,
     TROUBLESHOOTING_TIP_USB,
 } from 'src/components/suite/troubleshooting/tips';
+import { useBridgeDesktopApi } from 'src/hooks/suite/useBridgeDesktopApi';
 
 import { selectConnectingDevices } from '../../../actions/bluetooth/desktopBluetoothSelectors';
 import { useSelector } from '../../../hooks/suite';
@@ -31,20 +33,24 @@ export const DeviceConnect = ({ setIsBluetoothConnectOpen }: DeviceConnectProps)
     const { isBluetoothEnabled } = useSelector(selectSuiteFlags);
     const autoConnectingDeviceIds = useSelector(selectConnectingDevices);
 
+    const commonCableTips = [
+        TROUBLESHOOTING_TIP_UDEV,
+        TROUBLESHOOTING_TIP_CABLE,
+        TROUBLESHOOTING_TIP_USB,
+    ];
     const cableItem: TroubleshootingTipsItem[] = isWebUsbTransport
-        ? [
-              TROUBLESHOOTING_TIP_UDEV,
-              TROUBLESHOOTING_TIP_CABLE,
-              TROUBLESHOOTING_TIP_USB,
-              TROUBLESHOOTING_TIP_SUITE_DESKTOP,
-          ]
+        ? [...commonCableTips, TROUBLESHOOTING_TIP_SUITE_DESKTOP]
         : [
               TROUBLESHOOTING_TIP_BRIDGE_STATUS,
-              TROUBLESHOOTING_TIP_UDEV,
-              TROUBLESHOOTING_TIP_CABLE,
-              TROUBLESHOOTING_TIP_USB,
+              ...commonCableTips,
               TROUBLESHOOTING_TIP_DIFFERENT_COMPUTER,
           ];
+
+    if (useBridgeDesktopApi()?.bridgeProcess?.process) {
+        cableItem.push(TROUBLESHOOTING_TIP_SUITE_DESKTOP_TOGGLE_BRIDGE);
+    } else {
+        // todo: here we are going to put instruction to uninstall standalone bridge
+    }
 
     const getCallToActionButton = (): ReactNode => {
         if (isWebUsbTransport) {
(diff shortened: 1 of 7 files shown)


packages/transport/src/transports/bridge.ts
```diff
@@ -135,20 +135,13 @@ export class BridgeTransport extends AbstractTransport {
...
                 signal: this.abortController.signal,
             });
 
             if (!response.success) {
-                const time = Date.now() - listenTimestamp;
-                if (time <= 1100) {
-                    this.emit(TRANSPORT.ERROR, response.error);
-                    break;
-                }
-                await resolveAfter(1000);
```
mroz22, line 151: I had to get rid of this. For some reason, toggling from new to legacy did not emit transport error. Now both variants work (when we talk about emitting transport-error followed by transport-start after clicking the "toggle" button that I am adding here). cc @marekrjpolak

packages/suite/src/components/suite/troubleshooting/tips/BridgeTip.tsx
```diff
@@ -69,6 +73,16 @@ export const BridgeToggle = () => {
...
+
+                                // if bridge process is not running it might make sense to toggle it on and also
+                                // try to switch between legacy and node implementation (= who knows why it didn't start?)
+                                if (!bridgeProcess.process) {
+                                    // to get some sentry insights
+                                    console.error(
+                                        'Bridge process was not running and user toggled it in troubleshooting tips',
+                                    );
+                                    toggleBridge();
+                                }
```
karliatto, line 85: As far as I know if bridge process is not running in desktop is restarted automatically. But I understand what this is doing is to change to different bridge. The question is whether this way of managing the process is something that should be handled in the UI, it seams to me that should be part of the background process that is able to change the bridge type if for some reason the one using is not running.

karliatto: But I understand doing it on user action might just solve the issue.

karliatto (approved):
Looks good to me! 
Toggling between  bridges works nicely in my computer. 
Added some thoughts about doing it based on user action in UI or handle it in the background process.

packages/suite/src/components/suite/troubleshooting/tips/BridgeTip.tsx
mroz22: this is very very unlikely to render. there is already bridge auto-restart + bridge auto start after external bridge got killed.

packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
```diff
@@ -31,21 +33,25 @@ export const DeviceConnect = ({ setIsBluetoothConnectOpen }: DeviceConnectProps)
...
         : [
               TROUBLESHOOTING_TIP_BRIDGE_STATUS,
-              TROUBLESHOOTING_TIP_UDEV,
-              TROUBLESHOOTING_TIP_CABLE,
-              TROUBLESHOOTING_TIP_USB,
+              ...commonCableTips,
               TROUBLESHOOTING_TIP_DIFFERENT_COMPUTER,
           ];
 
+    if (useBridgeDesktopApi()?.bridgeProcess?.process) {
```
marekrjpolak, line 49: This probably causes these errors as it can be rendered outside desktop Suite:
image

packages/transport/src/transports/bridge.ts
marekrjpolak: I believe this should work (I've tried many possible combination but never encountered a legit timeout so every unsuccessful response could be emitted as an error), but it would be a bit safer to a) put at least the `resolveAfter` back so in case of unhandled error (which shouldn't happen) there isn't a shitload of them or b) set `stopped` after the first error so there are no more failed requests until transport is restarted from outside (which should always happen).

Anyway, even the current code should work just fine.

packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
mroz22: oh yes. Initially I had it like 
```
isDesktop() && useBridgeDesktopApi()..
```
and then I made an untested assumption that presence of isDesktop check is part of useBridgeDesktopApi and remove it. c1a08509f5994f6c8e9998eae1f26beaf4556dbc

packages/suite/src/components/suite/PrerequisitesGuide/DeviceConnect.tsx
```diff
@@ -31,21 +33,25 @@ export const DeviceConnect = ({ setIsBluetoothConnectOpen }: DeviceConnectProps)
...
         : [
               TROUBLESHOOTING_TIP_BRIDGE_STATUS,
-              TROUBLESHOOTING_TIP_UDEV,
-              TROUBLESHOOTING_TIP_CABLE,
-              TROUBLESHOOTING_TIP_USB,
+              ...commonCableTips,
               TROUBLESHOOTING_TIP_DIFFERENT_COMPUTER,
           ];
 
+    if (isDesktop() && useBridgeDesktopApi()?.bridgeProcess?.process) {
```
marekrjpolak: Unfortunately this won't work so moving `isDesktop` check inside `useBridgeDesktopApi` may be the right way after all. :man_shrugging:

mroz22: oh, exactly, this was the reason why I removed the check here 
```
  React Hook "useBridgeDesktopApi" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks
```

mroz22: or maybe 
image
🤷

PR #19891, opened 2025-07-01
feat(connect): add HyperEVM support

NelsonGaldeman:
## Description

Adds support for HyperEVM so you can successfully sign through Trezor Connect.


martykan:
Hi, unfortunately, this is not the right place to add it and it wouldn't actually help. 
Definitions for all EVM chains are specified in [trezor/definitions](https://github.com/trezor/definitions) repo, which takes [ethereum-lists/chains](https://github.com/ethereum-lists/chains/) as the source of data. The problem with HyperEVM is that according to the lists, chainID 999 is taken by [Wanchain Testnet](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-999.json) since 2021. 
We don't really want to get into politics of who gets which chainID, so it would be best to resolve the issue with ethereum-lists maintainers.

However we have discussed that we can lift the restriction on signing with ETH derivation paths on testnets - [related firmware issue](https://github.com/trezor/trezor-firmware/issues/5134), which would allow HyperEVM to be used normally in this case.

PR #19899, opened 2025-07-02
fix(suite): add compose methods to init blacklist too

mroz22:
might fix #19876 I did not test it personally, could you please @Ondra-Zik-SL 

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=blacklist-compose&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=blacklist-compose&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=blacklist-compose&lookback=7)

PR #19900, opened 2025-07-02
chore(suite): start using __info dry run param for connect methods

mroz22:
this PR exposes validation and other rules known by connect to the suite layer. At the moment, suite only uses information whether device will be used by a method it called and if yes, it applies its own UI lock (used for disabling buttons). 

followup for #19899

### QA 
- disabling buttons (eg. when getAddress is called) should work the same way as before
- nothing else should break 

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=use-connect-method-info&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=use-connect-method-info&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=use-connect-method-info&lookback=7)


suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -134,17 +134,25 @@ export const connectInitThunk = createThunk<void, ConnectInitHooks | void, void>
...
+                        const infoResult = await original({ ...params, __info: true });
+
+                        if (infoResult.success && infoResult.payload.useDevice) {
+                            dispatch(lockDevice(true));
+                        }
+                        return original(params).finally(() => {
+                            if (infoResult.success && infoResult.payload.useDevice) {
+                                dispatch(lockDevice(false));
+                                dispatch(
+                                    deviceActions.removeButtonRequests({
```
mroz22, line 159: I am thinking about registering listener for button requests here instead of doing it globaly to handle adding button requests. Only when method is expected to use device.

PR #19910, opened 2025-07-02
Fix: sell ethereum test

Lemonexe:
WIP followp to https://github.com/trezor/trezor-suite/pull/19909, after which the Normal/Advanced fee `SelectBar` is properly disabled, but the `sell-ethereum-test` does not know it yet, so it tries clicking on it right away.

- `disabled` attribute is missing on the `SelectBar` component, so add it
- in the test, await the option to be enabled, but that's not working yet :confused:

### DIFF
--- a/packages/components/src/components/form/SelectBar/SelectBar.tsx
+++ b/packages/components/src/components/form/SelectBar/SelectBar.tsx
@@ -213,6 +213,8 @@ export const SelectBar = <V extends ValueTypes>({
                                     <Option
                                         onClick={handleOptionClick(option)}
                                         $isDisabled={!!isDisabled}
+                                        // @ts-expect-error
+                                        disabled={!!isDisabled}
                                         $isSelected={isSelected}
                                         data-testid={`select-bar/${String(option.value)}`}
                                     >
--- a/packages/suite-desktop-core/e2e/tests/trading/sell-ethereum.test.ts
+++ b/packages/suite-desktop-core/e2e/tests/trading/sell-ethereum.test.ts
@@ -23,7 +23,7 @@ const formattedAddress = formatAddress(sellWatchEthereum.destinationAddress);
 const gasLimit = '26000';
 const maxFeePerGas = '2.67674454';
 const maxPriorityFeePerGas = '1.375641927';
-/*TODO: Uncomment once bug #19186 is resolved 
+/*TODO: Uncomment once bug #19186 is resolved
 + import BigNumber from '@trezor/utils' and localizeNumber from '@suite-common/wallet-utils'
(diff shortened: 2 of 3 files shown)


packages/components/src/components/form/SelectBar/SelectBar.tsx
```diff
@@ -213,6 +213,8 @@ export const SelectBar = <V extends ValueTypes>({
                                     <Option
                                         onClick={handleOptionClick(option)}
                                         $isDisabled={!!isDisabled}
+                                        // @ts-expect-error
+                                        disabled={!!isDisabled}
```
Lemonexe, line 217: `Option` is a `div` and `disabled` is not a standardized attribute on div, indeed I have to silence TS.
→ Maybe it should be a `button`?
It'd be semantically correct (better than div with onClick), so better accessibility etc.

packages/suite-desktop-core/e2e/tests/trading/sell-ethereum.test.ts
```diff
@@ -65,8 +65,12 @@ test.describe('Trading - Sell Ethereum', { tag: ['@group=trading', '@webOnly'] }
         },
     );
 
-    test('Sell Ethereum', async ({ tradingPage, devicePrompt }) => {
+    test('Sell Ethereum', async ({ page, tradingPage, devicePrompt }) => {
         await test.step('Fill in a sell request', async () => {
+            await expect(tradingPage.fees.switchModeButton('custom')).toBeVisible();
+            await expect(tradingPage.fees.switchModeButton('custom')).toBeEnabled();
+            // TODO: toBeEnabled() is not working, it immediately resolves despite disabled attribute being present
+            await page.waitForTimeout(1000);
```
Lemonexe, line 73: The component `SelectBar` did not pass `disabled` prop to the `Option`, so ofc it was not working.
But even when I did that, and I can actually see the `disabled` prop on the `div` rendered by the `SelectBar > Option`, playwright ignores it, and resolves `toBeEnabled` immediately.

Maybe because https://github.com/trezor/trezor-suite/pull/19910/files#r2179937934

But really, the attribute is there in DOM:
Screenshot from 2025-07-02 12-16-34

Vere-Grey: Alternatively, we could expect `await expect(<locator>).toHaveAttribute('disabled');`

Lemonexe: yes, I have thought of that. :slightly_smiling_face: 
But what you need is to **not** have attribute disabled, and playwright does not seem to have such method :confused:

PR #19925, opened 2025-07-02
chore(suite): remove systeminformation lib

### LINKED ISSUE #19787: Remove systeminformation dependency
- `systeminformation` is a library with wide array of features, but we use only [veeery few of them](https://github.com/trezor/trezor-suite/blob/b96c899ebbaa7d9b9c324356202758d5e5927630/packages/suite-desktop-core/src/libs/info.ts#L21-L26).
  - its executes a lot of shell commands (different per each platform) to query system info, so removing it will reduce the surface area between Suite and OS 🙂  
  - also one less Foundation team dependency to maintain 🙂 
- let's replace it with own code using just NodeJS API
  - the important info is easy to do. Minor details can be dropped if not feasible.

Lemonexe:
## Description

Remove `systeminformation` lib from Suite Desktop electron-main, because it is unnecessarily complex, and it does automatically many queries that we don't need.

I am partially replacing it with nodeJS api.

Please note these logs **are only generated [in debug mode](https://github.com/trezor/trezor-suite/blob/a188538e948d2a9e5a8a3e45de4e224c55b76736/packages/suite/src/components/suite/layouts/SuiteLayout/PageHeader/PageNames/SettingsName.tsx#L37-L46)** and this data **are not sent anywhere**.
Yeah, they are not even sent with the Suite logs generated by the UI (which are mostly redux logs).
These nodeJS logs stay in `suite app data/logs/*.txt` if you have debug mode on.

## Related Issue

Resolve #19787

## Logs

**Before**
```
Platform: linux x64
OS: Ubuntu 24.04.2 LTS
Manufacturer: Dell Inc.
Model: Latitude 5440
Virtual: No
CPU: Intel Gen Intel® Core™ i7-1365U
Cores: 1x10(+2) @ 2.23GHz
RAM: 38.8 GB
```

**After**
```
Platform: linux x64
CPU: 13th Gen Intel(R) Core(TM) i7-1365U
Cores: 12 @ 3.30 GHz
RAM: 38.8 GB
```

Missing:
- physical vs virtual cores, this only displays virtual cores
- manufacturer, model
- virtualization

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/remove-systeminformation&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/remove-systeminformation&lookback=7)

### DIFF
--- a/packages/suite-desktop-core/src/libs/info.ts
+++ b/packages/suite-desktop-core/src/libs/info.ts
@@ -1,10 +1,25 @@
 import { app } from 'electron';
-import si from 'systeminformation';
+import os from 'os';
 
 import { isDevEnv } from '@suite-common/suite-utils';
 import { bytesToHumanReadable } from '@trezor/utils';
 
 import { b2t } from './utils';
+
+const getCPUInfo = () => {
+    const cpus = os.cpus();
+    const cores = cpus.length;
+    const model = cpus[0]?.model.trim() ?? 'Unknown CPU';
+
+    const maxSpeedMHz = Math.max(...cpus.map(cpu => cpu.speed));
+    const speedGHz = isNaN(maxSpeedMHz) ? 'Unknown CPU speed' : (maxSpeedMHz / 1000).toFixed(2);
+
+    return {
+        model,
+        cores,
+        speedGHz,
+    };
+};
 
 export const getBuildInfo = () => [
     'Info:',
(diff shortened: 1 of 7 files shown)


packages/suite-desktop-core/src/libs/info.ts
```diff
@@ -1,11 +1,26 @@
...
 import { bytesToHumanReadable } from '@trezor/utils';
 
 import { b2t } from './utils';
 
+const getCPUInfo = () => {
+    const cpus = os.cpus();
+    const cores = cpus.length;
+    const model = cpus[0]?.model.trim() ?? 'Unknown CPU';
+
+    const maxSpeedHz = Math.max(...cpus.map(cpu => cpu.speed));
```
peter-sanderson:
```suggestion
    const maxSpeedMHz = Math.max(...cpus.map(cpu => cpu.speed));
```

Then `/1000` to GHz makes sense

peter-sanderson (approved):
Just those MHz, I was confused

martykan:
Might be useful to add `os.version()`. On Linux/Mac it will provide the kernel version instead of actual OS release, but we can at least estimate it from that.

packages/suite-desktop-core/src/libs/info.ts
Lemonexe: Oh yes sorry, I can't count :see_no_evil:

Lemonexe:
> Might be useful to add `os.version()`. On Linux/Mac it will provide the kernel version instead of actual OS release, but we can at least estimate it from that.

@martykan I added both release() and version().
This is what chatGPT says as examples. A bit chaotic :smile: 
But it's just debug info, not analytics, so the data does not have to be clean.

| Example           | `os.platform()` | `os.release()`           | `os.version()`                                       |
| ----------------- | --------------- | ------------------------ | -----------------------------------------------------|
| **Ubuntu 24.04**  | `'linux'`       | `'6.8.0-35-generic'`     | `'#123-Ubuntu SMP Thu Jun 6 12:34:56 UTC 2025'`      |
| **Windows 11**    | `'win32'`       | `'10.0.22631'`           | `'Windows 11 Pro 10.0.22631 Build 22631'`            |
| **macOS Sequoia** | `'darwin'`      | `'24.0.0'`               | `'Darwin Kernel Version 24.0.0: Wed May 15 08:00...` |

martykan (approved):
👍

PR #19941, opened 2025-07-03
feat(suite-native): contextual dynamic screen header

juriczech:
## Description
- add animation for screen header content
- create new component DynamicHeader so we can render top header when content is scrolled outside of the scroll view
- refactor multiple screens to reuse this header to avoid code duplicity

## Related Issue

Resolve [Dynamic screen header](https://github.com/trezor/trezor-suite/issues/19709)

## Screenshots:

https://github.com/user-attachments/assets/ea64e1f7-52f3-432f-bd4b-67066ca47658

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/native/dynamic-screen-header-with-context&lookback=7)

PR #19963, opened 2025-07-03
feat(suite-native): Check Backup - Unsupported Device Model Screen

### LINKED ISSUE #19842: Check backup - Model One unsupported modal
## Acceptance criteria
As a user with Model One connected to my phone, I would like to be informed that the check wallet backup feature is only available on Desktop

[Figma](https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Settings--MOB-?node-id=2595-4323&m=dev)

BrantalikP:
## Description

- Created a case for when a user wants to check the backup on an unsupported model (currently Model One only).
- e2e for Device Check Backup

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19842

## Screenshots:

https://github.com/user-attachments/assets/2de7f145-7c24-4820-b46d-83d5fe9c2ea3

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/native/checkup-backup-trezor-one&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/native/checkup-backup-trezor-one&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/native/checkup-backup-trezor-one&lookback=7)


suite-native/module-check-backup/src/screens/DeviceCheckBackupUnsupportedModelScreen.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    const redirectToWeb = () => {
+        openLink(SUITE_WEB_URL);
+    };
+
+    const redirectHome = () => {
+        navigation.navigate(DeviceSettingsStackRoutes.DeviceSettings);
+    };
+
+    return (
```
BrantalikP, line 44: Since a similar screen is already used in two cases (onboarding and this one)  https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Settings--MOB-?node-id=2595-4323&m=dev wondering whether to create a generic screen out of it. If so, where should it be located 🧐 A separate module feels a bit like overkill. cc: @PeKne @juriczech

PeKne: IMO it is worth it to crate a reusable component for this UI. What about moving it to the `@suite-native/device`? Both of the places already have this dependency.

juriczech: I wouldn't create a ScreenComponent for this. The first box will become a `DynamicScreenHeader` (see https://github.com/trezor/trezor-suite/pull/19941). I have a draft PR that will implement this and we'll be able to reduce the code here. 

What would make sense to me is to create a component for the list. If i'ts used in multiple places, we could create a component that would accept an object of icons and translations and would render the items using `IconListTextItem`. With this and the header gone, all you would have is that component and two buttons which I believe is okay and doesn't require a component wrapper for that.

juriczech: ...and the wrapper component would make sense in atoms, so it would be super easy to reuse.

BrantalikP: Hey guys, I appreciate the input from both of you. and thank you for taking the time to share your POV

I agreed with both of you at some point, but when I consider the benefits of having it as a reusable component vs the potential maintenance and scaling, it does not feel much valuable (If it doesn’t start appearing more frequently in designs, then we should treat it probably as atom component without any business logic included)

Regarding the `IconList` component. Although it can save some lines of code and reduce redundancy by passing fewer props I see here the same problem as the `Generic UnssuportedScreen`. Are we certain that we won’t need to modify individual `iconTextItem`'s props such as size, variant, or color in the future?

If not, we’re essentially just moving code from JSX into a modifiable config array/object, which might not bring much benefit.

As now I think the cleanest thing is actually keep it as is. 😅

BrantalikP: The reason I brought you into this conversation is that I spent quite a bit of time playing around with both of your approaches, and at some point, both felt a bit off to me. So I needed to validate it with someone :D

juriczech: To me, leaving it as it is, since header will be removed and the list items are an atom is perfectly fine and I'd be happy with this. Having the list would be nice-ish, IF it didn't bring more complexity than necessary. If it does, let's not complicate it and leave it as is. We can always do it if it starts appearing in more places and there is a lot of duplicity. 👍

suite-native/module-check-backup/src/components/CheckBackupTutorialStep2.tsx
```diff
@@ -27,15 +30,27 @@ type NavigationProps = StackNavigationProps<
...
     currentStepIndex,
 }: WalletBackupTutorialNumberedStepProps) => {
     const navigation = useNavigation<NavigationProps>();
+    const deviceModel = useSelector(selectDeviceModel);
 
     const { applyStyle } = useNativeStyles();
     const { showToast } = useToast();
 
     const navigateToCheckBackup = () => {
+        if (deviceModel && checkBackupUnsupportedDeviceModels.includes(deviceModel)) {
```
BrantalikP, line 45: waiting for @shenkys response if this is correct entry point https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU?node-id=2595-4323&m=dev#1326660712

suite-native/atoms/src/SwipeableWalkthrough/SwipeableWalkthroughStep.tsx
```diff
@@ -83,15 +83,15 @@ export const SwipeableWalkthroughStep = ({
                 walkthroughStepAnimatedStyle,
                 applyStyle(stepContainerStyle, { height: stepContainerHeight }),
             ]}
-            testID="@deviceOnboarding/SwipeableWalkthroughStep"
+            testID="swipeableWalkthroughStep"
```
juriczech: nitpick, just for consistency
```suggestion
            testID="@swipeableWalkthroughStep"
```

suite-native/atoms/src/SwipeableWalkthrough/SwipeableWalkthroughStep.tsx
```diff
@@ -83,15 +83,15 @@ export const SwipeableWalkthroughStep = ({
...
         >
             <ScrollView
                 bounces={false}
                 contentContainerStyle={applyStyle(scrollViewContentStyle, {
                     safeAreaInsetBottom: bottomSafeAreaInset,
                     height: stepContainerHeight,
                 })}
-                testID="@deviceOnboarding/SwipeableWalkthroughStep/scrollView"
+                testID="swipeableWalkthroughStep/scrollView"
```
juriczech:
```suggestion
                testID="@swipeableWalkthroughStep/scrollView"
```

suite-native/module-check-backup/src/components/CheckBackupTutorialStep2.tsx
```diff
@@ -27,15 +30,27 @@ type NavigationProps = StackNavigationProps<
     DeviceCheckBackupStackRoutes.CheckBackupTutorial
 >;
 
+const checkBackupUnsupportedDeviceModels: Array<DeviceModelInternal | null> = [
```
juriczech: no need for null. Am I missing something?
```suggestion
const checkBackupUnsupportedDeviceModels: Array<DeviceModelInternal> = [
```

suite-native/module-check-backup/src/screens/DeviceCheckBackupUnsupportedModelScreen.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const navigation = useNavigation<NavigationProps>();
+    const openLink = useOpenLink();
+
+    const redirectToWeb = () => {
+        openLink(SUITE_WEB_DEVICE_SETTINGS_URL);
+    };
+
+    const redirectHome = () => {
+        navigation.navigate(DeviceSettingsStackRoutes.DeviceSettings);
+    };
```
juriczech: Nitpick for naming - `redirectToDeviceSettings`. You could also use `navigateToInitialScreen` in case this would ever be used elsewhere

juriczech (commented):
Just some nits, otherwise LGTM

suite-native/module-check-backup/src/screens/DeviceCheckBackupUnsupportedModelScreen.tsx
BrantalikP: `navigateToInitialScreen` good point with this one, but It seems to work only for the Root Screen of The `Stack` meaning it will navigate me to the root of the DeviceCheckBackup Navigation rather than to settings

suite-native/atoms/src/SwipeableWalkthrough/SwipeableWalkthroughStep.tsx
BrantalikP: https://github.com/trezor/trezor-suite/pull/19963/commits/1b1a1219aca94a82c10fa5da8185455069f85627

suite-native/atoms/src/SwipeableWalkthrough/SwipeableWalkthroughStep.tsx
BrantalikP: https://github.com/trezor/trezor-suite/pull/19963/commits/1b1a1219aca94a82c10fa5da8185455069f85627

suite-native/module-check-backup/src/components/CheckBackupTutorialStep2.tsx
BrantalikP: https://github.com/trezor/trezor-suite/pull/19963/commits/1b1a1219aca94a82c10fa5da8185455069f85627 good catch

suite-native/module-check-backup/src/screens/DeviceCheckBackupUnsupportedModelScreen.tsx
BrantalikP: https://github.com/trezor/trezor-suite/pull/19963/commits/1b1a1219aca94a82c10fa5da8185455069f85627

juriczech: @BrantalikP Are you sure? That's weird. It should work that it will move you to previous stack. Something might be off...

If there is more than one route in the stack state, it will pop to top and then go back. If only one route, it will go back (so go to previous stack). 

But it's not critical, right now it is not necessary to go to this screen from other flows.

juriczech (approved):

suite-native/module-check-backup/src/screens/DeviceCheckBackupUnsupportedModelScreen.tsx
BrantalikP: Hmm, I think so, because I remembered that you actually suggested it to Peta somewhere, so I wanted to give it a shot,  but unfortunately, it didn’t work as expected. Let's investigate it later. anyway. Thanks 💪

juriczech (approved):
LGTM

PR #19976, opened 2025-07-04
fix(react-native-usb): do not close pending requests when Suite goes to background

### LINKED ISSUE #19556: Switching from Suite Lite to different app or locking phone when being asked to confirm address on device causes device to disconnect
**Describe the bug**
If I close the app or switch to a different app to check my address for example when sending a transaction, the Suite thinks the device gets disconnected. Same applies to locking your phone and also applies to different requests to device (changing PIN, showing receive address,...)

**Info:**

- Mobile Suite Version: version/commit
- OS: name, version

**How to reproduce**

1. navigate to send
2. enter address and amount
3. select fee
4. confirm the 2 steps on next screen
5. switch to different app/system home screen
6. go back to Suite
7. Suite thinks the device got disconnected

**Expected behavior**
I am able to switch to different app or lock my screen when I make request on device

**Screenshots**

https://github.com/user-attachments/assets/a4f66e8b-eae6-446d-97e4-4fadc37d1493

matejkriz:
## Description

We call `closeAllOpenedDevices` every time Suite Lite goes into the background, because this makes the Trezor accessible to other apps. Without closing the device, the user can’t use any third-party app after Suite has been opened, nor can they use Suite Web (which we need, for example, for T1 onboarding or checking the backup).

However, we should not cancel all pending requests at that point, because we want certain actions—like displaying a receive address—to remain active while the user checks it in the source app. We don’t want to cancel any pending requests just because the app went into the background. We only want to close the device when Suite itself is closed (killed).

I also added an easy way to turn on logging from transport-native-usb by switching ENV variable. I can put it in separate PR if needed.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19556

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/native/do-not-cancel-flow-on-background&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/native/do-not-cancel-flow-on-background&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/do-not-cancel-flow-on-background&lookback=7)

### DIFF
--- a/packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
+++ b/packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
@@ -192,6 +192,7 @@ class ReactNativeUsbModule : Module() {
             ReactNativeUsbAttachedReceiver.setOnDeviceConnectCallback(null)
             ReactNativeUsbDetachedReceiver.setOnDeviceDisconnectCallback(null)
 
+            cancelAllPendingRequests()
             closeAllOpenedDevices()
 
             try {
@@ -264,10 +265,14 @@ class ReactNativeUsbModule : Module() {
 
     private fun closeAllOpenedDevices() {
         Log.d(LOG_TAG, "Closing all devices")
-        pendingRequests.values.forEach { it.cancel() }
-        pendingRequests.clear()
+
         openedConnections.values.forEach { it.close() }
         openedConnections.clear()
+    }
+
+    private fun cancelAllPendingRequests() {
+        pendingRequests.values.forEach { it.cancel() }
+        pendingRequests.clear()
     }
 
     private fun claimInterface(deviceName: String, interfaceNumber: Int) {
--- a/packages/transport-native-usb/src/nativeUsb.ts
+++ b/packages/transport-native-usb/src/nativeUsb.ts
@@ -11,7 +11,10 @@ export class NativeUsbTransport extends AbstractApiTransport {
         super({
             api: new UsbApi({
                 usbInterface: new WebUSB(),
-                logger,
+                logger:
+                    process.env.EXPO_PUBLIC_IS_NATIVE_USB_LOGGER_ENABLED === 'true'
+                        ? console
+                        : logger,
             }),
             logger,
             ...rest,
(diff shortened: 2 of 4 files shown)


packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
```diff
@@ -194,6 +194,9 @@ class ReactNativeUsbModule : Module() {
 
             closeAllOpenedDevices()
 
+            pendingRequests.values.forEach { it.cancel() }
+            pendingRequests.clear()
```
yanascz: I believe this should be done before calling `closeAllOpenedDevices()`. I'd also hide it into a method called `cancelAllPendingRequests()`.

packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
```diff
@@ -264,8 +267,7 @@ class ReactNativeUsbModule : Module() {
 
     private fun closeAllOpenedDevices() {
         Log.d(LOG_TAG, "Closing all devices")
-        pendingRequests.values.forEach { it.cancel() }
-        pendingRequests.clear()
```
yanascz, line 268: This will break THP. Not sure how to fix it for the current FWs, though.

yanascz, line 268: Removing this will break THP. We'll have to figure out how ensure both use cases work correctly.

matejkriz: yes 🙉

packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
matejkriz: a3a33b2f1e87894ce71c440da0b6ad0af32d916a

yanascz (approved):

PR #19268, opened 2025-06-02
feat: follow-up fimware update for mobile

karliatto:
## Description
This is a follow-up of firmware update flow https://github.com/trezor/trezor-suite/pull/18821  for Suite mobile app.
* It makes mobile app to use the conditional firmware updates
* Add new dev setting in mobile app to select the firmware source "production" | "test-unsigned" | "test-signed"
* It refactors the way firmwareUpdateSource is stored in Suite web and desktop as well so it can play better with mobile.

## Related Issue

Related https://github.com/trezor/trezor-suite/issues/19763

## Screenshots:

image

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/firmware-update-wit-mobile&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/firmware-update-wit-mobile&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/firmware-update-wit-mobile&lookback=7)

PR #19997, opened 2025-07-07
TEST: for QA testing new FW update process

karliatto:
## Description
This is special build prepared for QA to test new FW update process `feat/firmware-update` .

## What to test

- test all the flows (from factory reset, from normal from bitcoin only to universal from universal to bitcoin only with all devices model
- test with and without internet
- test in web and desktop (for now only desktop since the testing server needs to be updated to provide header for CORS)
- test if downloading and storing and reusing FW works in desktop

### On-boarding flow
1. Reset device to factory so it can be used to install unsigned binaries
2. Open the provided testing version of suite and go throw on-boarding installing firmware (online and offline)
3. Firmware installation should complete properly and should be possible to complete on-boarding.

### Firmware update flow
1. Prepare device with older unsigned firmware inhaling it manually with `trezorctl` or Trezor Suite custom FW installation. 
2. If require do on-boarding in this device skipping the firmware update during on-boarding
3. Once on-boarding is completed firmware update should be offered, click on it and complete the FW update process


karliatto:
Done by https://github.com/trezor/trezor-suite/pull/19268

PR #20006, opened 2025-07-07
fix(connect): retry invalid state even without popup

martykan:
## Description

Previously, if the `sessionId` was invalid the flow would immediately fail with a `Device_InvalidState` error. 
Unless using Connect Popup, where it prompts user whether they want to retry. 

With this PR, we will now automatically retry once when encountering an invalid `sessionId`. If the retry also fails, we then throw the `Device_InvalidState` error.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/connect-retry-invalid-state&lookback=7)


martykan:
Probably not needed / not great idea, seems the real cause for our current problem was #20011

PR #20020, opened 2025-07-08
fix(suite-native): ui quirks in settings

juriczech:
## Description
Fix some UI quirks from design QA
https://www.figma.com/design/yY9Zl2jXLKXVV3CwuBmaMU/Settings--MOB-?node-id=2901-4353&p=f&t=K9LFhNEJiafVx5fr-0

## Related Issue

Resolve 

## Screenshots:
image
image
image
image
image
image
image

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/design-qa-settings&lookback=7)

### DIFF
--- a/suite-native/module-settings/src/components/CryptoUnitsSelector.tsx
+++ b/suite-native/module-settings/src/components/CryptoUnitsSelector.tsx
@@ -33,8 +33,10 @@ export const CryptoUnitsSelector = () => {
         >
             <Select<PROTO.AmountUnit>
                 selectValue={bitcoinUnit}
+                selectLabel={<Translation id="moduleSettings.preferences.bitcoinUnitsLabel" />}
                 items={bitcoinUnitsItems}
                 onSelectItem={handleSelectUnit}
+                shouldDisplayLabelInTrigger={false}
                 testID="@settings/localization/bitcoin-units-selector"
             />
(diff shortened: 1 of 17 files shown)


suite-native/module-settings/src/components/CryptoUnitsSelector.tsx
```diff
@@ -33,6 +33,7 @@ export const CryptoUnitsSelector = () => {
         >
             <Select<PROTO.AmountUnit>
                 selectValue={bitcoinUnit}
```
yanascz, line 35: Since it's related only to Bitcoin, what about renaming the component to `BitcoinUnitsSelector`? 🤔

juriczech: I think why we initially didn't name it Bitcoin name specifically is because it applies to bitcoin-like coins, not only to bitcoin. But I don't have a strong opinion on this and we can rename if you feel like it's better name.

yanascz: Selectors, translation keys, labels, ... All refer to Bitcoin. So I'd rename the component as well.

yanascz (approved):
Looks good! 👏

PR #20029, opened 2025-07-08
fix(connect): explicitly check button request name in ethereumSignTypedData

martykan:
## Description

Unfortunately in `ethereumSignTypedData` we get unrelated `ButtonRequest_Other` events, for example from passphrase confirmation before the signing, resulting in confusing UI. 
Not sure why FW doesn't use a different button request code for that...

This PR adds an explicit list of names to filter only the relevant button requests.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/ethereumsigntypeddata-buttonrequest-ignore-passphrase&lookback=7)

### DIFF
--- a/packages/connect/src/api/ethereum/api/ethereumSignTypedData.ts
+++ b/packages/connect/src/api/ethereum/api/ethereumSignTypedData.ts
@@ -118,8 +118,20 @@ export default class EthereumSignTypedData extends AbstractMethod<'ethereumSignT
         );
     }
 
-    getButtonRequestData(code: string) {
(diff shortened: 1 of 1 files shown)


mroz22 (approved):
unfortunately, these `name` fields are likely to be subject to changes not so far in the future in firmware :(

evgenysl:
Tested on https://connect.trezor.io/9/methods/ethereum/ethereumSignTypedData/ and in Suite with passphrase.
QA OK

Issue #5939, opened 2022-08-05
BTC as a currency pair for all coins


Hannsek:
Some users may find useful to have BTC as a currency pair for all coins.

Full specification and design links [below](https://github.com/trezor/trezor-suite/issues/5939#issuecomment-2090430995).

**proposed changes**

- [x] Put `BTC` ~~and `ETH`~~ as an option in the drop down menu in `Settings/Application/Localization/Fiat currency`
- [x] rename the _Fiat currency_ field to _Unit_ as there are other assets like `Gold`, `Silver`, `XDR` -> https://github.com/trezor/trezor-suite/pull/19956
   - [ ] make sure it works on mobile as well

-> BTC will be the currency pair as e.g. USD, EUR, … so at all places in suite, where USD (EUR) is now, BTC will replace it (except BTC accounts)

**examples**

<img width="214" alt="image" src="https://user-images.githubusercontent.com/57008159/186343202-c6101c7a-eac3-438f-bff8-a05a484eda36.png">

<img width="448" alt="image" src="https://user-images.githubusercontent.com/57008159/186343741-282677c0-d90c-4122-ac5b-e024a5bcb2dc.png">

sime:
Attempted to solve via #6547 

Testable: https://suite.corp.sldev.cz/suite-web/sime/add-btc-currency/web/

Hannsek:
Obsolete pair for btc accounts: 
<img width="276" alt="image" src="https://user-images.githubusercontent.com/57008159/199505573-a2c5ba05-04e3-4e06-bbd2-9e8981fd10fa.png">
_Fiat currency field_ not renamed

Hermez-cz:
1. [Wallet switcher](https://www.figma.com/file/T0DBX3enk49MAbGx7JjPpd/Suite-Desktop-PO-Workspace?type=design&node-id=498%3A74&mode=design&t=TyAJvWXhoRZL4tHc-1)
   1. ✅Replace the fiat values with values in BTC/sats according to the user's selection in the settings.
2. [Dashboard](https://www.figma.com/file/T0DBX3enk49MAbGx7JjPpd/Suite-Desktop-PO-Workspace?type=design&node-id=498%3A75&mode=design&t=TyAJvWXhoRZL4tHc-1)
   1. ✅Portfolio value: Replace the fiat value with values in BTC/sat according to the user's selection in the settings.
   2. ✅Graph Y-axis: Use the scale in BTC similar to how it is used in BTC accounts.
[...]

PR #20035, opened 2025-07-08
feat: better fiat/base currncy display

peter-sanderson:
Part of: https://github.com/trezor/trezor-suite/issues/5939

THIS MAY BREAK THINGS :exclamation: 

This removes the formatting responsibility from rate-conversion function. Formatting SHALL ALWAYS be a responsibility of the components/formatters all math shall be done in BigNumber in whole precision.

## Testing
- Staking, especially fiat input and limits validations and rounding
- SendForm, especially fiat input and limits validations and rounding
- Trading especially fiat input and limits validations and rounding
- Fiat Rates in general, all places where the fiat rates are visible
- Chart/Graph (fiat rates)

All off above is also relevant for mobile.

Extra focus to rounding and precision.

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=btc-as-base-currency3&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=btc-as-base-currency3&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=btc-as-base-currency3&lookback=7)


suite-common/wallet-utils/src/fiatConverterUtils.ts
```diff
@@ -1,6 +1,21 @@
...
-export const toFiatCurrency = (amount: string, rate?: number, decimals = 2) => {
+import { BaseCurrencyAmount, asBaseCurrencyAmount } from './baseCurrency';
+
+type ToFiatCurrencyParams = {
+    amount: string | BigNumber;
+    rate: number | undefined;
+};
+
+/**
+ * This function SHALL NEVER round or change precision. This is simply just a multiplication of two numbers.
```
Lemonexe: Instead of such stern comment, why don't you declare return type as `BigNumber | null` or such?
Then it will become clear that this function must return number and not something formatted (string)

suite-common/wallet-utils/src/fiatConverterUtils.ts
```diff
@@ -15,10 +30,20 @@ export const toFiatCurrency = (amount: string, rate?: number, decimals = 2) => {
...
+type FromFiatCurrencyParams = {
+    localAmount: string;
+    rate: number | undefined;
 };
 
-export const fromFiatCurrency = (localAmount: string, decimals: number, rate?: number) => {
+/**
+ * This function SHALL NEVER round or change precision. This is simply just a division of two numbers.
+ *
+ * Formatting MUST be handled only in formatters in the components, etc... NOT HERE!
```
Lemonexe: same here as above + add explicit return type

suite-common/wallet-utils/src/baseCurrency.ts
```diff
@@ -0,0 +1,10 @@
+import { Branded } from '@trezor/type-utils';
+import { BigNumber } from '@trezor/utils';
+
+/**
+ * Value in EUR, USD, ... but also it can be in BTC, currently the global BaseCurrency from the Settings is used.
+ */
+export type BaseCurrencyAmount = BigNumber & Branded<`base-currency-amount`>;
+export const asBaseCurrencyAmount = (value: BigNumber) => value as unknown as BaseCurrencyAmount;
```
Lemonexe: `as unknown` seems superfluous, isn't it a relict of some previous wip code?

packages/utils/tests/typedObjectKeys.test.ts
```diff
@@ -0,0 +1,6 @@
+import { typedObjectKeys } from '../src/typedObjectKeys';
+
+type AB = { a: number; b: number } | { b: string };
+const ab: AB = { b: 'B' };
+
+export const _test: 'b'[] = typedObjectKeys(ab);
```
Lemonexe: duplicate file with `typedObjectKeys.type-test.ts `

packages/utils/tests/typedObjectFromEntries.test.ts
```diff
@@ -0,0 +1,21 @@
...
+] as const);
+
+export const _test2: { [K in keyof typeof map]: number } = typedObjectFromEntries(
+    typedObjectKeys(map).map(k => [k, map[k] * 2]),
+);
+
+// @ts-expect-error String cannot be assigned to number as map-value
+export const _test3: { [K in keyof typeof map]: number } = typedObjectFromEntries(
+    typedObjectKeys(map).map(k => [k, '']),
+);
```
Lemonexe: duplicate file with `typedObjectFromEntries.type-test.ts`

Lemonexe (commented):
CR WIP, fajront.

packages/suite/src/components/suite/BaseCurrencyValue.tsx
```diff
@@ -78,7 +80,10 @@ export const BaseCurrencyValue = ({
     });
 
     const { BaseCurrencyAmountFormatter } = useFormatters();
-    const value = shouldConvert ? fiatAmount : amount;
+
+    // In case `shouldConvert === true` the `amount` is the Crypto-Value and will be converted.
+    // However, in case `shouldConvert === false`, the `amount` is already converted BaseCurrency (Fiat) value.
+    const value = shouldConvert ? fiatAmount : asBaseCurrencyAmount(new BigNumber(amount));
```
peter-sanderson, line 86: This is MEGA WTF stuff in the suite on many places. That conditionally do some stuff by some flag, that makes it really hard to maintain.

packages/suite/src/components/wallet/TransactionItem/CoinjoinBatchItem.tsx
```diff
@@ -154,7 +155,9 @@ export const CoinjoinBatchItem = ({
                                             <HiddenPlaceholder>
                                                 <BaseCurrencyAmountFormatter
                                                     currency={localCurrency}
-                                                    value={fiatAmount.absoluteValue().toFixed()}
+                                                    value={asBaseCurrencyAmount(
+                                                        fiatAmount.absoluteValue(),
```
peter-sanderson, line 159: This is the downside of this strict typing done by Branded. If you want to use method on BigNumber you get BigNumber and you have to cast it back.

Solution exists but it involves wrapping BigNumber to some generics wrapper.

suite-common/wallet-utils/src/fiatConverterUtils.ts
peter-sanderson: It is already declared ad `BaseCurrencyAmount` (branded BigNumber), so you suggest to drop the comment only?

suite-common/wallet-utils/src/baseCurrency.ts
peter-sanderson: fixed

suite-common/wallet-utils/src/fiatConverterUtils.ts
Lemonexe: Sry, `toFiatCurrency` is ok it has return type defined, only `fromFiatCurrency` has inferred return type.

And yes, I believe that explicit declaration of return type is totally sufficient to communicate "what this function should return" and the comment is superfluous. When you need to change type declaration, it is a sign you're changing the purpose of the function.

If you really think it is necessary to prevent someone in future from rewriting the return type to `string`, than the comment can be shorter and less caps, like:
`this function does only numerical operations, formatting is to be handled in formatters.`

peter-sanderson: fixed, 

sidenote: `fromFiatCurrency` shall be refactored in the future to return the `AmountXXX` branded type

suite-common/wallet-utils/src/fiatConverterUtils.ts
peter-sanderson: fixed

suite-common/wallet-utils/src/__tests__/fiatConverter.test.ts
```diff
@@ -1,49 +1,48 @@
...
+const rate = 3007.1079886708517;
 const rateString = '3007.1079886708517';
-const decimals = 18;
 
 describe('fiatConverter utils: toFiatCurrency', () => {
     it('to existing fiat currency', () => {
-        expect(toFiatCurrency('1', rateNumber)).toBe('3007.11');
-        expect(toFiatCurrency('0', rateNumber)).toBe('0.00');
-        expect(toFiatCurrency('1.00000000000', rateNumber)).toBe('3007.11');
+        expect(toFiatCurrency({ amount: '1', rate })?.toFixed(2)).toBe('3007.11');
```
Lemonexe, line 8: :thought_balloon:  wow, this is very smart! Idk, maybe you just moved the same code from the function to test without realizing it, but this is simple elegant solution to approx float equality!
I [made a util](https://github.com/trezor/trezor-suite/blob/24f4d182bce128c72c01a89e3d3824c3838fe9e9/packages/utils/src/isApproximatelyEqual.ts) for this, which is a very standard pattern in math code, but did not realize that we can simplify it by comparing toFixed. It effectively works like approx equality with absolute tolerance (although in my util I went a step further and used relative tolerance).
I will definitely use this in future, thx for inspiration :laughing: 

:thought_balloon:  Oh, one more thought – a reason why people might use it so much – using `toFixed` makes memoization easier, because the fixed string is referentially stable even when float differ just very slightly (e.g. float rounding error). This is purely theoretical, but I can see how some react memoization might rely on this, and rounding error might cause rerenders :thinking:

packages/suite/src/components/suite/graph/TransactionsGraph/GraphTooltipDashboard.tsx
```diff
@@ -36,11 +37,17 @@ export const GraphTooltipDashboard = ({
     const sentAmountString = sentValueFn(payload[0].payload);
 
     const receivedAmount = (
-        <BaseCurrencyAmountFormatter currency={localCurrency} value={receivedAmountString ?? '0'} />
+        <BaseCurrencyAmountFormatter
+            currency={localCurrency}
+            value={receivedAmountString ?? BASE_CURRENCY_ZERO}
```
Lemonexe, line 42: `BASE_CURRENCY_ZERO` is good thing :+1:

suite-common/wallet-utils/src/fiatConverterUtils.ts
```diff
@@ -15,10 +28,21 @@ export const toFiatCurrency = (amount: string, rate?: number, decimals = 2) => {
         return null;
     }
 
-    return decimals === -1 ? localAmount.toFixed() : localAmount.toFixed(decimals);
+    return asBaseCurrencyAmount(localAmount);
+};
+
+type FromFiatCurrencyParams = {
+    localAmount: string;
```
Lemonexe: Nit: I just realized, I don't understand the naming `localAmount`. Why local, what does that mean?
I'd name the parameter `amount` for brevity, and then change `amount` to `convertedAmount`.

In `toFiatCurrency` I'd change `localAmount` to `convertedAmount`

That way the code goes amount → convertedAmount in both functions.

suite-native/device/src/selectors.ts
```diff
@@ -118,7 +118,7 @@ const getTotalFiatBalanceNative = ({
         instanceBalance = instanceBalance.plus(accountFiatBalance);
     });
 
-    return instanceBalance.toFixed(2);
```
Lemonexe, line 121: :eyes:  this has a potential to break something if it needs referentially stable amount from selector..

suite-native/graph/src/hooks.ts
```diff
@@ -204,7 +204,7 @@ export const useGraphAtoms = <TGraphPoint extends FiatGraphPoint>({
     referencePointAtom: WritableAtom<TGraphPoint | null, [TGraphPoint | null], void>;
     selectedPointAtom: WritableAtom<TGraphPoint | null, [TGraphPoint | null], void>;
     graphPoints: TGraphPoint[];
-    totalFiatBalance: string;
+    totalFiatBalance: BaseCurrencyAmount;
```
Lemonexe, line 207: :memo: ad `suite-native/device/src/selectors.ts`: what used to be fixed string is now bignumber, and it's persisted as "the selected point on graph".
SO I tested, it works well :heavy_check_mark:

suite-native/device-manager/src/components/WalletItemBase.tsx
```diff
@@ -81,7 +83,7 @@ export const WalletItemBase = ({
                 <HStack alignItems="center" spacing="sp12">
                     {fiatBalance && (
                         <BaseCurrencyAmountFormatter
-                            value={fiatBalance}
+                            value={asBaseCurrencyAmount(new BigNumber(fiatBalance))}
```
Lemonexe, line 86: :memo: ad `suite-native/device/src/selectors.ts`: tested that formatting in WalletItems dropdown is not broken.

suite-native/module-trading/src/components/general/__tests__/FiatAmountBadge.comp.test.tsx
```diff
@@ -13,20 +15,18 @@ describe('FiatAmountBadge', () => {
...
+        });
 
         expect(getByText('$0.00')).toBeDefined();
     });
-
-    it('should display nothing for empty string value', () => {
-        const { toJSON } = renderFiatAmountBadge({ amount: '' });
-
-        expect(toJSON()).toBeNull();
-    });
```
Lemonexe, line 31: this assertion should be equivalent now: 

```
    it('should display nothing for empty string value', () => {
        const { toJSON } = renderFiatAmountBadge({
            amount: asBaseCurrencyAmount(new BigNumber('')),
        });

        expect(toJSON()).toBeNull();
    });
```

suite-native/module-trading/src/components/general/__tests__/FiatAmountBadge.comp.test.tsx
```diff
@@ -13,20 +15,18 @@ describe('FiatAmountBadge', () => {
...
+        });
 
         expect(getByText('$1,234.56')).toBeDefined();
     });
 
     it('should display 0 value', () => {
-        const { getByText } = renderFiatAmountBadge({ amount: '0' });
+        const { getByText } = renderFiatAmountBadge({
+            amount: asBaseCurrencyAmount(new BigNumber('0')),
+        });
```
Lemonexe, line 28: this test fails

packages/suite/src/components/suite/BaseCurrencyValue.tsx
Lemonexe: That's indeed horrible. Thanks for at least explaining it.

packages/suite/src/views/dashboard/AssetsView/AssetsView.tsx
```diff
@@ -80,10 +80,10 @@ const useAssetsFiatBalances = (
                 .reduce((balance, account) => balance + Number(account.formattedBalance), 0)
                 .toString() ?? '0';
 
-        const fiatBalance = toFiatCurrency(amount, fiatRate?.rate, 2) ?? '0';
+        const fiatBalance = toFiatCurrency({ amount, rate: fiatRate?.rate })?.toFixed(2) ?? '0';
 
         return [...acc, { fiatBalance, symbol: asset.network.symbol }];
-    }, []);
+    }, [] as AssetFiatBalance[]);
```
Lemonexe: unnecessary, the reduce generic takes care of the same thing.

packages/suite/src/views/dashboard/PortfolioCard/DashboardGraph.tsx
```diff
@@ -126,7 +127,7 @@ export const DashboardGraph = memo(({ accounts }: DashboardGraphProps) => {
                         isLoading={graph.isLoading || isProcessing}
                         localCurrency={localCurrency}
                         xTicks={xTicks}
-                        minMaxValues={minMaxValues}
+                        minMaxValues={[minMaxValues[0].toNumber(), minMaxValues[1].toNumber()]}
```
Lemonexe, line 130: nit: `TransactionsGraph` and all code downstream could work with `BigNumber`, converting to number or string at the very edge.
In MVC language, it'd be nice to have type consistency within the "model", and push conversions to the "view code".  But not do as you wish, BigNumber ↔ number interface is not a dealbreaker for me, string would be worse :D

Lemonexe (approved):
Looks good generally, but a lot of tests need to be fixed.

Overall amazing initiative :rocket: 

There are a lot of interfaces where conversions could be pushed further to the "edge of the code", I mean the rendering. But that's inevitable with refactoring of this size, so I do not request to make this PR even larger :smile: 

I tested briefly with mobile app, did not test Suite, I assume you did while developing.

suite-common/wallet-utils/src/fiatConverterUtils.ts
peter-sanderson: what about `fiatAmount`? currently for naming consistency, later we rename it to `amount` with proper Branded type

Lemonexe: Sure perfect. `fiatAmount` and `cryptoAmount` are more descriptive :+1:

suite-native/device/src/selectors.ts
peter-sanderson: Goo point, thanks: here I addressed it: https://github.com/trezor/trezor-suite/pull/20035/commits/8b39809894900384fd436e97cd850ec386017a95

peter-sanderson: It is used only in 2 selectors `selectSelectedDeviceTotalFiatBalance` and `selectDeviceTotalFiatBalanceByDeviceState`

- `selectDeviceTotalFiatBalanceByDeviceState` only one and I fixed that place to `2`, there was `String(..)` :scream: 
- `selectSelectedDeviceTotalFiatBalance` is used in `useGraphAtoms` that I already changed to get the Branded type, BUT.. WTF I found terrible (potential) bug! Ill send commit later

peter-sanderson: See: https://github.com/trezor/trezor-suite/pull/20035/commits/7001c67a005af354e983c0d686b27e23a834fd12

suite-native/module-trading/src/components/general/__tests__/FiatAmountBadge.comp.test.tsx
peter-sanderson: fixed

suite-common/formatters/src/formatters/tests/prepareBaseCurrencyAmountFormatter.test.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+describe(prepareBaseCurrencyAmountFormatter.name, () => {
+    let displaySymbolFormatter: ReturnType<typeof prepareBaseCurrencyAmountFormatter>;
+
+    beforeEach(() => {
+        displaySymbolFormatter = prepareBaseCurrencyAmountFormatter(formatterConfig);
+    });
+
+    const dataProvider: Array<{ it?: string; input: string; expected: string }> = [
```
Lemonexe, line 30: why is it called `dataProvider`? We usually call this `fixtures`.

Lemonexe (approved):
OK.
Some nits still stand, so resolve as per your discretion.

STew790:
**QA NOK**
Very probably related to this https://github.com/trezor/trezor-suite/issues/20594
On mobile, it also broke the fiat input (for some reason, only for satoshis):
image

STew790:
Also on mobile swap looks weird, there is no amount for balance on the BTC account when switched to satoshis
image

Lemonexe:
@STew790 this does not pertain base currencies (this is crypto to crypto, not fiat denominated as BTC).
I believe that this is not the screenshot you wanted to post, I think you meant this?
Anyway, pls new issue for mobile trends team.
I confirmed that https://github.com/trezor/trezor-suite/issues/20594 is not an issue at mobile.

Screenshot from 2025-08-07 13-30-15

PR #20047, opened 2025-07-08
Improve locktime field

### LINKED ISSUE #3746: Improve locktime field
**Is your feature request related to a problem? Please describe.**

When entering a locktime users need to use an external resource to either:
- calculate the timestamp of the date and time they want to enter, or
- look up the current block height.

Suite should provide a simple means for them to do so inside the application.

Image
Note to the screenshot: future blockheight and timestamp are not convertible, we could offer an estimate at best. Let's drop that. Instead, autofill current datetime/blockheight when the option is selected. Allow passing seconds as well, in format hh:mm:ss; when they are not provided, treat the seconds value as 00; the default value should not include seconds (format hh:mm, i.e. 00:00). Datetime will be the default variant. Below the input, on the right side, include info about current blockheight/datetime.

### LINKED ISSUE #5391: Do not try to broadcast transactions with a timelock
If user has created a transaction where `lock_time > current block height`, the Suite should not offer to "broadcast" the transaction = the broadcast button should be disabled. Ideally with an explanation that the transaction could not be broadcasted to the mempool and should rather be saved (copy-paste as hex?) and broadcasted later.

### LINKED ISSUE #19960: "Not enough funds selected" disappears when locktime is changed
**Describe the bug**
If send with not enough funds through coin control selected, `Not enough funds selceted` disappears after locktime is added.

**Info:**

- Suite Version: web, `c855d592ad`
- Browser: brave 138.1.80.115
- Browser Installation Method: AUR
- OS: linux, 6.15.4-arch2-1
- Bridge + Firmware Version: emulator (trezor-user-env)

**How to reproduce**

1. Open desktop Trezor Suite
2. Go to "Send"
3. Turn on Coin Control and select one UTXO
4. Set amount such that its higher then selected UTXO, but lower than wallet total
5. Add Locktime

**Expected behavior**

The error stays present.

**Screenshots**

With locktime:

Image

Without locktime:

Image

**Additional context**
Note that `Review and Send` is disabled, so there should be an error present.

5K1PY:
Improves the locktime field. 

There remains some work for selecting date via calendar interface.

## Description

Added locktime field, which allows two modes:
- In block mode, block hieght can be entered
- In datetime mode, timelock can be entered
This allows better validation of user input, because we know user's intention.

## Related Issue

Resolve #3746
Resolve #5391 (the datetime option)
Resolve #19960 

## Screenshots:

Before

locktime-before

After - Block mode

locktime-block

After - Datetime mode

locktime-time


suite-common/wallet-utils/src/bitcoinUtils.ts
```diff
@@ -0,0 +1,27 @@
+import { getUnixTime, parse } from 'date-fns';
+
+import { BTC_LOCKTIME_VALUE } from '@suite-common/wallet-constants';
+
+export const parseUTCdatetime = (input: string | undefined): Date | undefined => {
```
OriginalEveres: nit: parseUTCdatetime is needed in order to convert datetimeToLocktime, but does it make it bitcoin utility or date utility?

suite-common/wallet-utils/src/bitcoinUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (!input) return undefined;
```
OriginalEveres: is there any specific reason for this early return? or is it there to simplify the conditions of input validation rules?

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
OriginalEveres, line 1: nit: I feel like this component could be simplified, eg.: `LockBlockCount` or `LockDatetime` could be separate components or the logic could be moved to `useLocktime` hook. wdyt @komret

OriginalEveres:
I'm currently going through the code, but I have noticed that our [guides](https://trezor.io/guides/trezor-suite/trezor-suite-desktop/locktime-in-trezor-suite) do not match the current app design. Also `Broadcast is turned off by default when using Locktime, meaning you will receive raw text data to broadcast from a node. If Broadcast is on, the transaction will be sent over the current network.` does not seem to be true.

OriginalEveres (changes requested):
overall this looks good! I have a few nits

packages/suite/src/support/messages.ts
```diff
@@ -5815,7 +5815,11 @@ export default defineMessages({
     },
     LOCKTIME_IS_TOO_BIG: {
         id: 'LOCKTIME_IS_TOO_BIG',
-        defaultMessage: 'Timestamp is too big',
+        defaultMessage: 'Locktime is too big',
+    },
+    LOCKTIME_INVALID_DATETIME: {
+        id: 'LOCKTIME_INVALID_DATETIME',
+        defaultMessage: 'Invalid datetime',
```
komret: Let's humanize this a bit and make it more helpful, e.g. "invalid date format, use dd/mm/yyyy hh:mm:ss".

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -1,38 +1,52 @@
...
+import { BTC_LOCKTIME_VALUE } from '@suite-common/wallet-constants';
 import { selectNetworkBlockchainInfo } from '@suite-common/wallet-core';
-import { getInputState, isInteger } from '@suite-common/wallet-utils';
-import { Card, Icon, IconButton } from '@trezor/components';
+import {
+    datetimeToLocktime,
+    getInputState,
+    isInteger,
+    localizeNumber,
+    parseUTCdatetime,
```
komret: Should be imported from `@suite-common/suite-utils`

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -45,28 +59,46 @@ export const Locktime = ({ close }: LocktimeProps) => {
 
     const options = getDefaultValue('options', []);
     const broadcastEnabled = options.includes('broadcast');
-    const inputName = 'bitcoinLockTime';
-    const defaultInputValue = getDefaultValue(inputName) || '';
-    const error = errors[inputName];
 
-    const locktime = watch('bitcoinLockTime');
+    const blockCountInputName = 'bitcoinLockTimeBlockCount';
```
komret: Nit: It is preferable to move any code not depending on the components's props or hooks above the component.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -45,28 +59,46 @@ export const Locktime = ({ close }: LocktimeProps) => {
...
+    const blockCountDefaultInputValue = getDefaultValue(blockCountInputName) || '';
```
komret: Nit: Could you use `watch` instead? `getDefaultValue` is an old util which is redundant AFAIK.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -45,28 +59,46 @@ export const Locktime = ({ close }: LocktimeProps) => {
...
+    const blockCountError = errors[blockCountInputName];
+
+    const datetimeInputName = 'bitcoinLockTimeDatetime';
+    const datetimeDefaultInputValue = getDefaultValue(datetimeInputName) || '';
+    const datetimeError = errors[datetimeInputName];
+
+    const blockCountLocktime = watch(blockCountInputName);
+    const datetimeLocktime = watch(datetimeInputName);
+
+    const [currentLockValue, setCurrentLockValue] = useState(
```
komret: Nit: Better to have hooks on top of the component.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -45,28 +59,46 @@ export const Locktime = ({ close }: LocktimeProps) => {
...
+    const blockCountInputName = 'bitcoinLockTimeBlockCount';
+    const blockCountDefaultInputValue = getDefaultValue(blockCountInputName) || '';
+    const blockCountError = errors[blockCountInputName];
+
+    const datetimeInputName = 'bitcoinLockTimeDatetime';
+    const datetimeDefaultInputValue = getDefaultValue(datetimeInputName) || '';
+    const datetimeError = errors[datetimeInputName];
+
+    const blockCountLocktime = watch(blockCountInputName);
+    const datetimeLocktime = watch(datetimeInputName);
```
komret: Nit: You can do `const [blockCountLocktime, datetimeLocktime] = watch([blockCountInputName, datetimeInputName])`

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -45,28 +59,46 @@ export const Locktime = ({ close }: LocktimeProps) => {
...
+
+    const [currentLockValue, setCurrentLockValue] = useState(
+        blockCountDefaultInputValue ? 'block' : 'date',
+    );
 
     useEffect(() => {
         if (
-            error === undefined &&
```
komret, line 56: Why removing this line? I think it can stay, just look at the right error here.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -45,28 +59,46 @@ export const Locktime = ({ close }: LocktimeProps) => {
...
                 currentBlockHeight: blockchain.blockHeight,
             }) &&
             broadcastEnabled
         ) {
             toggleOption('broadcast');
         }
 
-        composeTransaction(inputName);
-    }, [locktime, blockchain, toggleOption, broadcastEnabled, composeTransaction, error]);
+        composeTransaction(blockCountInputName);
```
komret: The argument dictates which field is a validation error assigned to in case it is thrown during transaction composition. Not sure if this can happen in practice, however the correct implementation would be to assign the error to  `datetimeInputName ` in case date is selected.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -77,32 +109,126 @@ export const Locktime = ({ close }: LocktimeProps) => {
...
+        validate: {
+            valid: (value: string | undefined) => {
+                if (parseUTCdatetime(value) === undefined) {
+                    return translationString('LOCKTIME_INVALID_DATETIME');
+                }
+            },
+            too_low: (value: string | undefined) => {
+                if (getUnixTime(parseUTCdatetime(value)!)! < BTC_LOCKTIME_VALUE) {
+                    return translationString('LOCKTIME_IS_TOO_LOW');
+                }
```
komret: Safer than asserting with `!`:

```suggestion
                const parsedUTCdatetime = value === undefined ? undefined : parseUTCdatetime(value);
                if (parsedUTCdatetime && getUnixTime(parsedUTCdatetime) < BTC_LOCKTIME_VALUE) {
                    return translationString('LOCKTIME_IS_TOO_LOW');
                }
```

suite-common/wallet-core/src/send/sendFormBitcoinThunks.ts
```diff
@@ -262,9 +263,14 @@ export const signBitcoinSendFormTransactionThunk = createThunk<
...
 
-        if (formState.bitcoinLockTime) {
-            signEnhancement.locktime = new BigNumber(formState.bitcoinLockTime).toNumber();
+        if (formState.bitcoinLockTimeBlockCount) {
+            signEnhancement.locktime = new BigNumber(
+                formState.bitcoinLockTimeBlockCount!,
+            ).toNumber();
+        } else if (formState.bitcoinLockTimeDatetime) {
+            signEnhancement.locktime = datetimeToLocktime(formState.bitcoinLockTimeDatetime)!;
         }
```
komret: Redundant `!` assertions?

suite-common/wallet-utils/src/reviewTransactionUtils.ts
```diff
@@ -205,8 +205,13 @@ const constructOldFlow = ({
...
     }
 
-    if (hasBitcoinLockTime && precomposedForm.bitcoinLockTime) {
-        outputs.push({ type: 'locktime', value: precomposedForm.bitcoinLockTime });
+    if (precomposedForm.bitcoinLockTimeBlockCount) {
+        outputs.push({ type: 'locktime', value: precomposedForm.bitcoinLockTimeBlockCount! });
+    } else if (precomposedForm.bitcoinLockTimeDatetime) {
+        outputs.push({
+            type: 'locktime',
+            value: datetimeToLocktime(precomposedForm.bitcoinLockTimeDatetime)!.toString(),
```
komret: Remove `!` please.

suite-common/wallet-utils/src/reviewTransactionUtils.ts
```diff
@@ -387,8 +391,13 @@ const constructNewFlow = ({
...
```
komret: Remove `!` please.

komret (changes requested):

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
komret: I agree, splitting the components into 2 is probably the way to go here.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
komret: Or perhaps better - remove the call to `composeTransaction` altogether, it seems out of place. But needs to be tested.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -45,28 +59,46 @@ export const Locktime = ({ close }: LocktimeProps) => {
...
-    }, [locktime, blockchain, toggleOption, broadcastEnabled, composeTransaction, error]);
+        composeTransaction(blockCountInputName);
+    }, [
+        blockCountLocktime,
+        datetimeLocktime,
+        blockchain,
+        toggleOption,
+        broadcastEnabled,
+        composeTransaction,
+        blockCountError,
```
komret: Redundant dependency.

suite-common/wallet-core/src/send/sendFormBitcoinThunks.ts
```diff
@@ -263,12 +263,12 @@ export const signBitcoinSendFormTransactionThunk = createThunk<
         // transactionInfo needs some additional changes:
         const signEnhancement: Partial<SignTransaction> = {};
 
-        if (formState.bitcoinLockTimeBlockCount) {
+        if (formState.bitcoinLocktimeBlockCount) {
             signEnhancement.locktime = new BigNumber(
-                formState.bitcoinLockTimeBlockCount!,
+                formState.bitcoinLocktimeBlockCount!,
```
komret: redundant `!`

suite-common/wallet-utils/src/reviewTransactionUtils.ts
```diff
@@ -205,13 +205,15 @@ const constructOldFlow = ({
...
 
-    if (precomposedForm.bitcoinLockTimeBlockCount) {
-        outputs.push({ type: 'locktime', value: precomposedForm.bitcoinLockTimeBlockCount! });
-    } else if (precomposedForm.bitcoinLockTimeDatetime) {
-        outputs.push({
-            type: 'locktime',
-            value: datetimeToLocktime(precomposedForm.bitcoinLockTimeDatetime)!.toString(),
-        });
+    if (precomposedForm.bitcoinLocktimeBlockCount) {
+        outputs.push({ type: 'locktime', value: precomposedForm.bitcoinLocktimeBlockCount! });
```
komret: redundant `!`

suite-common/wallet-utils/src/reviewTransactionUtils.ts
```diff
@@ -391,13 +393,15 @@ const constructNewFlow = ({
...
```
komret: redundant `!`

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeBlockCount.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const error = errors[inputName];
+
+    const options = watch('options', []);
+    const broadcastEnabled = options.includes('broadcast');
+
+    const { translationString } = useTranslation();
+
+    const locale = useSelector(selectLanguage);
+
+    const blockchain = useSelector(state => selectNetworkBlockchainInfo(state, network.symbol));
```
komret: You can use more specific `selectBlockchainHeightBySymbol` instead.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeBlockCount.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import { useSelector, useTranslation } from 'src/hooks/suite';
+import { useSendFormContext } from 'src/hooks/wallet';
+import { selectLanguage } from 'src/reducers/suite/suiteReducer';
+
+import { canLocktimeTxBeBroadcast } from './canLocktimeTxBeBroadcast';
+
+const inputName = 'bitcoinLocktimeBlockCount';
+
+interface LocktimeBlockCountProps {
```
komret: nit: we prefer `type` over `interface`

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeBlockCount.tsx
```diff
@@ -0,0 +1,96 @@
...
+            !canLocktimeTxBeBroadcast({
+                locktimeBlockCount: Number(inputValue),
+                locktimeDatetime: undefined,
+                currentBlockHeight: blockchain.blockHeight,
+            }) &&
+            broadcastEnabled
+        ) {
+            toggleOption('broadcast');
+        }
+    }, [error, inputValue, blockchain, broadcastEnabled, toggleOption]);
```
komret: I'd prefer to see this effect in the `Locktime` component so that it is not duplicated. I realize this means passing some extra props or duplicating some other (less complex) logic, but it might be cleaner that way.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeBlockCount.tsx
komret, line 1: nit: I'd rename to `LocktimeBlockHeight`

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -39,184 +25,27 @@ interface LocktimeProps {
 }
 
 export const Locktime = ({ currentBlockHeight, close }: LocktimeProps) => {
```
komret: Looks like we're drilling `currentBlockHeight` from `BitcoinOptions` which isn't really necessary, you select it lower in the components anyway.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeBlockCount.tsx
```diff
@@ -0,0 +1,96 @@
...
+        formState: { errors },
+        network,
+        toggleOption,
+        watch,
+    } = useSendFormContext();
+
+    const inputValue = watch(inputName) || '';
+    const error = errors[inputName];
+
+    const options = watch('options', []);
```
komret: The default values are probably redundant.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeBlockCount.tsx
```diff
@@ -0,0 +1,96 @@
...
+        },
+    };
+
+    return (
+        <NumberInput
+            control={control}
+            name={inputName}
+            locale={locale}
+            inputState={getInputState(error)}
+            defaultValue={inputValue}
```
komret: This might also be redundant.

komret (changes requested):
Note that some of the comments relate to both locktime components.

Also, I noticed a bug - changing the locktime value doesn't store a draft.

komret (approved):

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeDatetime.tsx
```diff
@@ -0,0 +1,88 @@
...
+        const utcDate = new Date(
+            now.getUTCFullYear(),
+            now.getUTCMonth(),
+            now.getUTCDate(),
+            now.getUTCHours(),
+            now.getUTCMinutes(),
+        );
+
+        return format(utcDate, 'dd/MM/yyyy HH:mm');
+    };
```
OriginalEveres: nit: might be in date utils?

suite-common/suite-utils/src/date.ts
```diff
@@ -91,3 +92,26 @@ export const resetTime = (ts: number, resetDay?: boolean) => {
...
+ * Parses datetime in UTC from one of the following formats:
+ * - dd/MM/yyyy
+ * - dd/MM/yyyy HH:mm
+ * - dd/MM/yyyy HH:mm:ss
+ * If a component is omitted, it is assumed to be zero.
+ *
+ * @param {string | undefined} input
+ * @returns {Date | undefined}
+ */
+export const parseUTCdatetime = (input: string | undefined): Date | undefined => {
```
OriginalEveres: nit: I'm still not sure about accepting the undefined. If we know that the input is undefined, we don't need to call the function, because it would return undefined.

I assume it accepts the undefined because of the values coming from RHF watch? So I consider this as nit.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -1,108 +1,111 @@
...
-                }
-                labelRight={<IconButton icon="x" size="tiny" variant="tertiary" onClick={close} />}
-                bottomText={error?.message || null}
-                data-testid="locktime-input"
-            />
+            <Row justifyContent="space-between" alignItems="start">
+                <TextColumn
+                    title="Add locktime"
+                    description="Transition cannot be broadcast until:"
+                />
```
OriginalEveres: missing translations?

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeBlockHeight.tsx
```diff
@@ -0,0 +1,75 @@
...
+            name={inputName}
+            locale={locale}
+            inputState={getInputState(error)}
+            onChange={() => composeTransaction()}
+            rules={rules}
+            bottomText={
+                <Row justifyContent="space-between" width="100%">
+                    <Text>{error?.message || ''}</Text>
+                    <Text variant="tertiary">
+                        {'Current block height: ' + localizeNumber(blockchainHeight, locale)}
```
OriginalEveres: missing translations?

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/LocktimeDatetime.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    return (
+        <Input
+            inputState={getInputState(error)}
+            defaultValue={inputValue}
+            bottomText={
+                <Row justifyContent="space-between" width="100%">
+                    <Text>{error?.message || ''}</Text>
+                    <Text variant="tertiary">{'Current UTC datetime: ' + getUTCNow()}</Text>
```
OriginalEveres: missing translation?

OriginalEveres (changes requested):
Nice job!

I have 2 OCD nits that you can totally skip. (basic comments)

But there are missing translations - I marked them with a review comment.

packages/suite/src/views/wallet/send/Options/BitcoinOptions/Locktime/Locktime.tsx
```diff
@@ -1,108 +1,111 @@
...
-                        <Translation id="LOCKTIME_SCHEDULE_SEND" />
-                    </Label>
-                }
-                labelRight={<IconButton icon="x" size="tiny" variant="tertiary" onClick={close} />}
-                bottomText={error?.message || null}
-                data-testid="locktime-input"
-            />
+            <Row justifyContent="space-between" alignItems="start">
+                <TextColumn
+                    title={<Translation id="LOCKTIME_ADD" />}
```
OriginalEveres: nit: for passing translation to a prop you can use the `translate('LOCKTIME_ADD')` from `const { translate } = useTranslate();` in that case you dont have to pass a whole component

OriginalEveres (approved):
LGTM! (got one nit, but you don't have to address it)

PR #20081, opened 2025-07-09
Chore/native/handle redirect to connecting screen

### LINKED ISSUE #20156: Refactor navigating to connecting screen
Refactor the logic for navigating to Device Connecting screen so it's not a `useEffect` in a `useHandleDeviceConnection` 

### Acceptance criteria
- [x] Move navigational logic to middleware which is triggered by a simple condition - device connection
- [x] Handle cases for which this flow is aborted

### LINKED ISSUE #20754: Stuck on "Continue on your Trezor" screen after restarting device
**Describe the bug**
Stuck on "Continue on your Trezor" screen after restarting device
**Info:**

- Mobile Suite Version: 01ee4b2b6b4af8ec21ed4df788cbb5bdb54085f9
- OS: iOS

**How to reproduce**

1. Setup T3W1 device (Pair BLE, THP, …)
2. Turn off Trezor
3. Kill Suite lite app
4. Turn on Trezor
5. Turn on Suite lite app
6. See THP connection confirmation step on device
7. Click `allow` on Trezor
8. Get stuck on "Continue on your Trezor" screen

- after _some_ time (~2 mins), I was automatically brought back to homescreen (in Suite lite), saw device disconnected, then connected and from there, everything works as expected. In the meanwhile Trezor was connected to mobile all the time via bluetooth.

**Screenshots**

Image

juriczech:
## Description
- create a listener middleware that listens only to specific actions so we're able to create a single decision tree for device connection navigations
- create a hook that allows for route blacklisting so that specific flows which require it are able to ignore device connection

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20156
Resolve https://github.com/trezor/trezor-suite/issues/20754

## Screenshots:

https://github.com/user-attachments/assets/29e7c17d-6ea0-48ed-9426-e9286e0c6316

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/handle-redirect-to-connecting-screen&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/native/handle-redirect-to-connecting-screen&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/native/handle-redirect-to-connecting-screen&lookback=7)

### DIFF
--- a/suite-native/device/src/hooks/useDeviceChecks.ts
+++ b/suite-native/device/src/hooks/useDeviceChecks.ts
@@ -1,62 +1,16 @@
 import { useSelector } from 'react-redux';
 
-import { selectIsFirmwareAuthenticityCheckDismissed } from '@suite-common/wallet-core';
-import {
-    selectIsDeviceAuthenticityCheckEnabled,
-    selectIsOnboardingFinished,
-} from '@suite-native/settings';
-
-import {
-    selectHasFirmwareAuthenticityCheckHardFailed,
-    selectIsDeviceAuthenticityCheckFailed,
-    selectIsEntropyCheckEnabledAndFailed,
-} from '../selectors';
+import { selectIsEntropyCheckEnabledAndFailed } from '../selectors';
 
 export const useDeviceChecks = (isDeviceCompromisedModalFocused: boolean) => {
-    const isOnboardingFinished = useSelector(selectIsOnboardingFinished);
-    const hasFirmwareAuthenticityCheckHardFailed = useSelector(
-        selectHasFirmwareAuthenticityCheckHardFailed,
-    );
-    const isFirmwareAuthenticityCheckDismissed = useSelector(
-        selectIsFirmwareAuthenticityCheckDismissed,
-    );
-    const isDeviceAuthenticityCheckFailed = useSelector(selectIsDeviceAuthenticityCheckFailed);
-    const isDeviceAuthenticityCheckEnabled = useSelector(selectIsDeviceAuthenticityCheckEnabled);
-    const isDeviceAuthenticityEnabledAndFailed =
-        isDeviceAuthenticityCheckEnabled && isDeviceAuthenticityCheckFailed;
-
     const isEntropyCheckEnabledAndFailed = useSelector(selectIsEntropyCheckEnabledAndFailed);
-
-    const isFirmwareAuthenticityCheckHardFailedAndNotDismissed =
-        hasFirmwareAuthenticityCheckHardFailed && !isFirmwareAuthenticityCheckDismissed;
-
-    // any failing check should navigate to the DeviceCompromisedModal
-    const shouldNavigateToDeviceCompromisedModal =
-        isOnboardingFinished &&
-        (isDeviceAuthenticityEnabledAndFailed ||
-            isEntropyCheckEnabledAndFailed ||
-            isFirmwareAuthenticityCheckHardFailedAndNotDismissed);
 
     // but the DeviceCompromisedModal shall not be persistent for Entropy check, because you cannot exit the modal via normal means
     const shouldKeepDeviceCompromisedModal =
         isDeviceCompromisedModalFocused && !isEntropyCheckEnabledAndFailed;
 
-    const getFailedCheck = (): 'device-authenticity' | 'entropy' | 'firmware-authenticity' => {
-        if (isDeviceAuthenticityEnabledAndFailed) {
-            return 'device-authenticity';
-        }
-        if (isEntropyCheckEnabledAndFailed) {
-            return 'entropy';
-        }
-
-        return 'firmware-authenticity';
-    };
-
-    const failedCheck = getFailedCheck();
-
     return {
-        shouldNavigateToDeviceCompromisedModal,
+        // any failing check should navigate to the DeviceCompromisedModal
         shouldKeepDeviceCompromisedModal,
-        failedCheck,
     };
 };
--- a/suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
+++ b/suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
@@ -1,0 +1,89 @@
+import {
+    Dispatch,
+    ListenerEffectAPI,
+    UnknownAction,
+    createListenerMiddleware,
+} from '@reduxjs/toolkit';
+
+import {
+    selectIsDeviceConnectedAndAuthorized,
+    selectIsDeviceInitialized,
+    selectIsDeviceUsingPassphrase,
+} from '@suite-common/wallet-core';
+import { selectIsFirmwareInstallationRunning } from '@suite-native/firmware';
+import {
+    AuthorizeDeviceStackRoutes,
+    RootStackRoutes,
+    checkIsRouteAccessAllowed,
+    navigationContainerRef,
+} from '@suite-native/navigation';
+import { selectIsCoinEnablingInitFinished } from '@suite-native/settings';
+
+import { NativeDeviceRootState, selectIsDeviceCompromised } from '../selectors';
+import { isDeviceConnectAction } from '../utils';
+
+const DEVICE_CONNECTION_BLACKLISTED_ROUTES = [
+    RootStackRoutes.DeviceCompromisedModal,
+    RootStackRoutes.DeviceOnboardingStack,
+    RootStackRoutes.SendStack,
+];
+
+export const deviceConnectionMiddleware = createListenerMiddleware<NativeDeviceRootState>();
+
+const handleDeviceConnectNavigation = ({
+    isCoinEnablingInitFinished,
+}: {
+    isCoinEnablingInitFinished: boolean;
+}) => {
+    if (isCoinEnablingInitFinished) {
+        navigationContainerRef.navigate(RootStackRoutes.AuthorizeDeviceStack, {
+            screen: AuthorizeDeviceStackRoutes.ConnectingDevice,
+        });
+    } else {
+        navigationContainerRef.navigate(RootStackRoutes.CoinEnablingInit);
+    }
+};
+
+deviceConnectionMiddleware.startListening({
+    predicate: (action, currentState) =>
+        isDeviceConnectAction(action) &&
+        // TODO this should dissappear after we merge device onboarding redirect here as well
+        // https://github.com/trezor/trezor-suite/issues/20157
+        // If device is not initialized and is compromised, we display the modal (reason why this condition is here) and then want to redirect to uninitialized device landing.
+        (selectIsDeviceInitialized(currentState) || selectIsDeviceCompromised(currentState)),
+    effect: (
+        _action: UnknownAction,
+        { getState }: ListenerEffectAPI<NativeDeviceRootState, Dispatch<UnknownAction>>,
+    ) => {
+        const shouldNavigateToDeviceCompromisedModal = selectIsDeviceCompromised(getState());
+        const isCoinEnablingInitFinished = selectIsCoinEnablingInitFinished(getState());
+
+        if (
+            !checkIsRouteAccessAllowed({
+                blacklist: DEVICE_CONNECTION_BLACKLISTED_ROUTES,
+            })
+        )
+            return;
+
+        // During firmware installation, device restarts (disconnect + connect) and we want to ignore it.
+        if (selectIsFirmwareInstallationRunning(getState())) return;
+
+        if (shouldNavigateToDeviceCompromisedModal) {
+            // When the compromised modal is closed on first connection and no coins would be selected, we will need to redirect user
+            // to coin enabling so he can continue to the app with running discovery.
+            navigationContainerRef.navigate(RootStackRoutes.DeviceCompromisedModal);
+
+            return;
+        }
+
+        // If device is authorized already (usually in case of remembered device which has already been authorized)
+        const isDeviceConnectedAndAuthorized = selectIsDeviceConnectedAndAuthorized(getState());
+        // Passphrase protected devices are only connected through passphrase form
+        // The passphrase flow handles connection differently and redirect to connecting screen is not wanted.
+        const isDeviceUsingPassphrase = selectIsDeviceUsingPassphrase(getState());
+
+        if (isDeviceUsingPassphrase || isDeviceConnectedAndAuthorized) return;
+
+        handleDeviceConnectNavigation({ isCoinEnablingInitFinished });
+    },
+});
--- a/suite-native/device/src/selectors.ts
+++ b/suite-native/device/src/selectors.ts
@@ -31,6 +31,7 @@ import {
     selectIsDeviceInBootloader,
     selectIsDiscoveredDeviceAccountless,
     selectIsEntropyCheckFailed,
+    selectIsFirmwareAuthenticityCheckDismissed,
     selectIsUnacquiredDevice,
     selectSelectedDevice,
     selectSelectedDeviceAuthenticity,
@@ -46,20 +47,25 @@ import {
     FeatureFlagsRootState,
     selectIsFeatureFlagEnabled,
 } from '@suite-native/feature-flags';
-import { SettingsSliceRootState } from '@suite-native/settings';
+import { NativeFirmwareRootState } from '@suite-native/firmware';
+import {
+    SettingsSliceRootState,
+    selectIsDeviceAuthenticityCheckEnabled,
+} from '@suite-native/settings';
 import { doesCoinSupportStaking } from '@suite-native/staking';
 import type { BaseCurrencyCode } from '@trezor/blockchain-link-types';
 import { BigNumber } from '@trezor/utils';
 
 import { isDeviceSetupSupported, isFirmwareVersionSupported } from './utils';
 
-type NativeDeviceRootState = DeviceRootState &
+export type NativeDeviceRootState = DeviceRootState &
     AccountsRootState &
     DiscoveryRootState &
     SettingsSliceRootState &
     WalletSettingsRootState &
     FiatRatesRootState &
     FeatureFlagsRootState &
+    NativeFirmwareRootState &
     MessageSystemRootState;
 
 const createMemoizedSelector = createWeakMapSelector.withTypes<NativeDeviceRootState>();
@@ -254,4 +260,58 @@ export const selectShouldFactoryResetBeVisible = createMemoizedSelector(
     [selectIsDeviceInBootloader, selectHasDeviceFirmwareInstalled],
     (isDeviceInBootloader, hasDeviceFirmwareInstalled) =>
         isDeviceInBootloader && hasDeviceFirmwareInstalled,
+);
+
+export const selectIsDeviceCompromised = createMemoizedSelector(
+    [
+        selectIsDeviceAuthenticityCheckEnabled,
+        selectIsDeviceAuthenticityCheckFailed,
+        selectIsEntropyCheckEnabledAndFailed,
+        selectIsFirmwareAuthenticityCheckDismissed,
+        selectHasFirmwareAuthenticityCheckHardFailed,
+    ],
+    (
+        isDeviceAuthenticityCheckEnabled,
+        isDeviceAuthenticityCheckFailed,
+        isEntropyCheckEnabledAndFailed,
+        isFirmwareAuthenticityCheckDismissed,
+        hasFirmwareAuthenticityCheckHardFailed,
+    ) => {
+        const isDeviceAuthenticityEnabledAndFailed =
+            isDeviceAuthenticityCheckEnabled && isDeviceAuthenticityCheckFailed;
+
+        const isFirmwareAuthenticityCheckHardFailedAndNotDismissed =
+            hasFirmwareAuthenticityCheckHardFailed && !isFirmwareAuthenticityCheckDismissed;
+
+        return (
+            isDeviceAuthenticityEnabledAndFailed ||
+            isEntropyCheckEnabledAndFailed ||
+            isFirmwareAuthenticityCheckHardFailedAndNotDismissed
+        );
+    },
+);
+
+export const selectCompromisedDeviceFailedCheck = createMemoizedSelector(
+    [
+        selectIsDeviceAuthenticityCheckEnabled,
+        selectIsDeviceAuthenticityCheckFailed,
+        selectIsEntropyCheckEnabledAndFailed,
+    ],
+    (
+        isDeviceAuthenticityCheckEnabled,
+        isDeviceAuthenticityCheckFailed,
+        isEntropyCheckEnabledAndFailed,
+    ) => {
+        const isDeviceAuthenticityEnabledAndFailed =
+            isDeviceAuthenticityCheckEnabled && isDeviceAuthenticityCheckFailed;
+
+        if (isDeviceAuthenticityEnabledAndFailed) {
+            return 'device-authenticity';
+        }
+        if (isEntropyCheckEnabledAndFailed) {
+            return 'entropy';
+        }
+
+        return 'firmware-authenticity';
+    },
 );
--- a/suite-native/module-authenticity-checks/src/components/FirmwareAuthenticityCheckFailModalContent.tsx
+++ b/suite-native/module-authenticity-checks/src/components/FirmwareAuthenticityCheckFailModalContent.tsx
@@ -1,34 +1,40 @@
 import { useDispatch, useSelector } from 'react-redux';
 
-import { useNavigation } from '@react-navigation/native';
+import { useNavigation } from '@react-navigation/core';
 
-import { deviceActions, selectSelectedDevice } from '@suite-common/wallet-core';
+import {
+    deviceActions,
+    selectIsDeviceInitialized,
+    selectSelectedDevice,
+} from '@suite-common/wallet-core';
 import { Button } from '@suite-native/atoms';
 import { Translation } from '@suite-native/intl';
 import {
-    AppTabsRoutes,
-    HomeStackRoutes,
     RootStackParamList,
     RootStackRoutes,
     ScreenHeader,
     StackToStackCompositeNavigationProps,
 } from '@suite-native/navigation';
+import { selectIsCoinEnablingInitFinished } from '@suite-native/settings';
 import { TREZOR_SUPPORT_FW_REVISION_CHECK_FAILED_MOBILE_URL } from '@trezor/urls';
 
 import { DeviceCompromisedModalContent } from './DeviceCompromisedModalContent';
 
 const supportUrlWithChat = `${TREZOR_SUPPORT_FW_REVISION_CHECK_FAILED_MOBILE_URL}#open-chat`;
 
-type NavigationProp = StackToStackCompositeNavigationProps<
+type NavigationProps = StackToStackCompositeNavigationProps<
     RootStackParamList,
-    RootStackRoutes.AppTabs,
+    RootStackRoutes.DeviceCompromisedModal,
     RootStackParamList
 >;
 
 export const FirmwareAuthenticityCheckFailModalContent = () => {
-    const device = useSelector(selectSelectedDevice);
     const dispatch = useDispatch();
-    const navigation = useNavigation<NavigationProp>();
+    const navigation = useNavigation<NavigationProps>();
+
+    const isDeviceInitialized = useSelector(selectIsDeviceInitialized);
+    const isCoinEnablingInitFinished = useSelector(selectIsCoinEnablingInitFinished);
+    const device = useSelector(selectSelectedDevice);
 
     const dismissCheck = () => {
         if (device?.id) {
@@ -36,24 +42,18 @@ export const FirmwareAuthenticityCheckFailModalContent = () => {
         }
     };
 
-    // After dismissCheck, an effect could fire in useHandleDeviceConnection to navigate away, but it's not guaranteed!
-    // To be sure we don't lock user on on this screen, we navigate home.
     const handleClose = () => {
-        // the modal is most likely entered from OnboardingStack, ConnectingDevice or Home, so let's send user back
-        if (navigation.canGoBack()) {
-            navigation.goBack();
-        }
-        // Home screen set only as fallback if can't go back
-        else {
-            navigation.navigate(RootStackRoutes.AppTabs, {
-                screen: AppTabsRoutes.HomeStack,
-                params: { screen: HomeStackRoutes.Home },
-            });
-        }
         dismissCheck();
+
+        if (!isCoinEnablingInitFinished && isDeviceInitialized) {
+            navigation.navigate(RootStackRoutes.CoinEnablingInit);
+        } else {
+            if (navigation.canGoBack()) navigation.goBack();
+        }
     };
 
     const screenHeaderContent = <ScreenHeader closeActionType="close" closeAction={handleClose} />;
+
     const closeButtonContent = (
         <Button colorScheme="redElevation0" onPress={handleClose}>
             <Translation id="generic.buttons.close" />
--- a/suite-native/navigation/src/navigators.ts
+++ b/suite-native/navigation/src/navigators.ts
@@ -315,9 +315,7 @@ export type RootStackParamList = {
     [RootStackRoutes.WalletConnectPair]: undefined;
     [RootStackRoutes.SettingsScreenStack]: NavigatorScreenParams<SettingsStackParamList>;
     [RootStackRoutes.BackupFailedModal]: undefined;
-    [RootStackRoutes.DeviceCompromisedModal]: {
-        failedCheck: 'device-authenticity' | 'entropy' | 'firmware-authenticity';
-    };
+    [RootStackRoutes.DeviceCompromisedModal]: undefined;
     [RootStackRoutes.TradingWebView]: {
         closeCallbackUrl: string;
         source?: { uri?: string; html?: string };
--- a/suite-native/state/src/store.ts
+++ b/suite-native/state/src/store.ts
@@ -4,7 +4,7 @@ import { logger } from 'redux-logger';
 
 import { prepareFiatRatesMiddleware } from '@suite-common/wallet-core';
 import { blockchainMiddleware } from '@suite-native/blockchain';
-import { prepareDeviceMiddleware } from '@suite-native/device';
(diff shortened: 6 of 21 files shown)


suite-native/firmware/src/hooks/useFirmware.tsx
```diff
@@ -42,6 +46,11 @@ export const useFirmware = (
 
     const setIsFirmwareInstallationRunning = useCallback(
         (isRunning: boolean) => {
+            if (isRunning) {
+                stopDeviceConnectionListening();
+            } else {
+                startDeviceConnectionListening();
+            }
```
yanascz: Could this be triggered by the `nativeFirmwareActions.setIsFirmwareInstallationRunning` action? I've got two places in the THP branch where I call this action directy (to avoid including this huge hook).

suite-native/module-authenticity-checks/src/components/FirmwareAuthenticityCheckFailModalContent.tsx
```diff
@@ -1,56 +1,34 @@
...
 
 const supportUrlWithChat = `${TREZOR_SUPPORT_FW_REVISION_CHECK_FAILED_MOBILE_URL}#open-chat`;
 
-type NavigationProp = StackToStackCompositeNavigationProps<
-    RootStackParamList,
-    RootStackRoutes.AppTabs,
-    RootStackParamList
->;
+type FirmwareAuthenticityCheckFailModalContentProps = {
+    onCloseRedirect: () => void;
```
yanascz: Looking at the usages of this property, I don't think the naming is right. I'd call it just `onClose` since you cannot know what logic will the caller put there.

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
```diff
@@ -0,0 +1,100 @@
...
+                    // to coin enabling so he can continue to the app with running discovery.
+                    onCloseRedirect: () => {
+                        if (!isCoinEnablingInitFinished && selectIsDeviceInitialized(getState())) {
+                            connectDevice({
+                                isCoinEnablingInitFinished: false,
+                            });
+                        } else {
+                            if (navigationContainerRef.canGoBack()) navigationContainerRef.goBack();
+                        }
+                    },
```
yanascz: Would prefer to have this callback declared elsewhere and then just included here, in a similar manner as `compromisedDeviceFailedCheck`.

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
```diff
@@ -0,0 +1,100 @@
...
+} from '@suite-native/navigation';
+import { selectIsCoinEnablingInitFinished } from '@suite-native/settings';
+
+import {
+    NativeDeviceRootState,
+    selectCompromisedDeviceFailedCheck,
+    selectIsDeviceCompromised,
+} from '../selectors';
+
+export const deviceConnectionNavigationMiddleware =
```
yanascz: If I understand it correctly, this middleware might do also other stuff than just navigating. What about calling it `deviceConnectionMiddleware`? No strong opinion, though.

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
```diff
@@ -0,0 +1,100 @@
...
+                return;
+            }
+
+            connectDevice({ isCoinEnablingInitFinished });
+        },
+    });
+};
+
+export const stopDeviceConnectionListening = () => {
+    deviceConnectionNavigationMiddleware.clearListeners();
```
yanascz: Would it be possible to ditch this and just set e.g. `connectionMiddlewareEnabled = false` in state and then use it in the middleware's `predicated`? I would find it much cleaner.

suite-native/device/src/hooks/useDeviceConnectionRouteBlacklist.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const DEVICE_CONNECTION_BLACKLISTED_ROUTES: RootStackRoutes[] = [
+    RootStackRoutes.DeviceCompromisedModal,
+    RootStackRoutes.DeviceOnboardingStack,
+    RootStackRoutes.OnboardingStack,
+];
+
+const REMEMBERED_DEVICE_BLACKLISTED_ROUTES: RootStackRoutes[] = [RootStackRoutes.SendStack];
+
+export const useDisableDeviceConnectionOnRouteBlacklist = () => {
```
yanascz: Would it be possible to create a hook to disable/enable the middleware and then use it on relevant screens? What is the benefit of having this global?

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
```diff
@@ -0,0 +1,100 @@
...
+            // Reference https://github.com/trezor/trezor-suite/pull/11319/commits/a9152279fe6d70c57fa16ee0bf75dc9fd52bb930
+            // Double check this when device onboarding is refactored as well to make sure removing this will not break anything
+            // https://github.com/trezor/trezor-suite/issues/20157
+            if (isDeviceConnectedAndAuthorized) return;
+
+            const shouldNavigateToDeviceCompromisedModal = selectIsDeviceCompromised(getState());
+            const isCoinEnablingInitFinished = selectIsCoinEnablingInitFinished(getState());
+
+            if (shouldNavigateToDeviceCompromisedModal) {
+                const compromisedDeviceFailedCheck = selectCompromisedDeviceFailedCheck(getState());
```
yanascz: Why do we have to select the check here and pass via `navigate`? Wouldn't it be better to handle this in `DeviceCompromisedModalScreen`? 🤔

suite-native/device/src/hooks/useDeviceConnectionRouteBlacklist.ts
BrantalikP: you still can use `stopDeviceConnectionListening/startDeviceConnectionListening ` for some specific scenarios as @juriczech used for [biometrics](https://github.com/trezor/trezor-suite/pull/20081/files#diff-25139166443441528b55772c0d753a5e2c22f9063a127cd8f4439f7a96eab4d0R44)

Anyway, having a whitelist and blacklist in one place gives it more clarity and helps developers understand where the hook is prohibited ( much better than searching for all occurrences. :))

Of course, in the case of similar hooks that might come after this PR, it will be better to use them only on the screens/routes we want to listen to.

BrantalikP: Oh, I see what you mean now… sorry @yanascz , I didn’t have the context... now I get it.

Yeah, actually, having `useDisableDeviceConnectionOnRouteBlacklist` in `globalHooks` feels a bit strange.

Isn’t there a way to handle the `blacklisted` screens directly in the middleware?

@yanascz On the other hand If we would need to use the hook selectively in each screen, then I don’t think we need the `middleware` in the first place.

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
BrantalikP: Fix me if I am wrong here @juriczech but I think this is how     `createListenerMiddleware` works

suite-native/firmware/src/hooks/useFirmware.tsx
BrantalikP: We should have access to the state also in [startDeviceConnectionListening](https://github.com/trezor/trezor-suite/pull/20081/files#diff-037236299608a31cac2524674b876d02501fdb7bbd6d7c930305c10d2d0b723aR41) right? I would suggest handling it there if we want to avoid any action from the hook being triggered. simply by adding a `return` statement if the FW is running.

suite-native/biometrics/src/components/BiometricOverlay.tsx
```diff
@@ -34,6 +39,13 @@ export const BiometricOverlay = ({
 }: BiometricOverlayProps) => {
     const { applyStyle } = useNativeStyles();
 
+    // Ignore device connection in biometrics overlay (note for code review: is this actually necessary?)
+    useEffect(() => {
+        stopDeviceConnectionListening();
+
+        return () => startDeviceConnectionListening();
+    }, []);
```
BrantalikP: feel free to ignore the comment, just thinking loud :D 

but pattern 

```
  useEffect(() => {
        stopDeviceConnectionListening();

        return () => startDeviceConnectionListening();
    }, []);
```

seems to be pretty handy, and have feeling that it might be needed at more places in the future, maybe it could be a hook, smth like `useFocus(Something :D )`

BrantalikP: actually I see you are also using it in [useDisableDeviceConnectionOnRouteBlacklist](https://github.com/trezor/trezor-suite/pull/20081/files#diff-76c790c91a22bb3d0a60f3c8bedf447fe949ed4c6380ee5103020d509f3a7782R37)

```

    useFocusEffect(
        useCallback(() => {
            if (shouldDisableConnection) {
                stopDeviceConnectionListening();

                return () => {
                    startDeviceConnectionListening();
                };
            }
        }, [shouldDisableConnection]),
```

suite-native/module-authenticity-checks/src/components/FirmwareAuthenticityCheckFailModalContent.tsx
juriczech: https://github.com/trezor/trezor-suite/pull/20081/commits/40b17a6e17cc81a2ee0610dac1623797c2d7d5aa

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
juriczech: Great call. I moved it to selector for this exact purpose and then forgot to use it. Simplified even more 😍 
https://github.com/trezor/trezor-suite/pull/20081/commits/40b17a6e17cc81a2ee0610dac1623797c2d7d5aa

suite-native/biometrics/src/components/BiometricOverlay.tsx
matejkriz: I'm afraid this pattern is not working very well - imagine the same effect in multiple components. If one of those components unmount, it starts device connection listening even though other components are still mounted.

suite-native/device/src/hooks/useDeviceChecks.ts
```diff
@@ -1,62 +1,24 @@
...
-        }
-
-        return 'firmware-authenticity';
-    };
-
-    const failedCheck = getFailedCheck();
-
     return {
-        shouldNavigateToDeviceCompromisedModal,
+        shouldNavigateToDeviceCompromisedModal: isDeviceCompromised,
```
matejkriz: This is now only renaming isDeviceCompromised to shouldNavigateToDeviceCompromisedModal. Better to remove it from here or keep the name `isDeviceCompromised`

suite-native/device/src/hooks/useDeviceChecks.ts
```diff
@@ -1,62 +1,24 @@
...
-    );
-    const isFirmwareAuthenticityCheckDismissed = useSelector(
-        selectIsFirmwareAuthenticityCheckDismissed,
-    );
-    const isDeviceAuthenticityCheckFailed = useSelector(selectIsDeviceAuthenticityCheckFailed);
-    const isDeviceAuthenticityCheckEnabled = useSelector(selectIsDeviceAuthenticityCheckEnabled);
-    const isDeviceAuthenticityEnabledAndFailed =
-        isDeviceAuthenticityCheckEnabled && isDeviceAuthenticityCheckFailed;
-
+    const compromisedDeviceFailedCheck = useSelector(selectCompromisedDeviceFailedCheck);
```
matejkriz: `selectCompromisedDeviceFailedCheck` returns `'firmware-authenticity'` by default, is that ok here?

suite-native/device/src/hooks/useDeviceConnectionRouteBlacklist.ts
```diff
@@ -0,0 +1,48 @@
...
+    const lastRoute = useNavigationState(state => state?.routes.at(-1)?.name) as RootStackRoutes;
+    const isDeviceRemembered = useSelector(selectIsDeviceRemembered);
+
+    const shouldDisableConnection = useMemo(() => {
```
matejkriz: ?
```suggestion
    const shouldDisableConnectionListening = useMemo(() => {
```

suite-native/device/src/hooks/useDeviceConnectionRouteBlacklist.ts
```diff
@@ -0,0 +1,48 @@
...
+            if (shouldDisableConnection) {
+                stopDeviceConnectionListening();
+
+                return () => {
+                    startDeviceConnectionListening();
+                };
+            }
+        }, [shouldDisableConnection]),
+    );
+};
```
matejkriz: To be honest I'm not sure how this behaves, because the effect returns cleanup function conditionally. If you rewrite it like this, it would be a bit more clear to me:

```
useFocusEffect(
  useCallback(() => {
    if (shouldDisableConnection) {
      stopDeviceConnectionListening();
    }

    return () => {
      if (shouldDisableConnection) {
        startDeviceConnectionListening();
      }
    };
  }, [shouldDisableConnection]),
);
```

suite-native/state/src/store.ts
```diff
@@ -48,6 +48,8 @@ export const initStore = async (preloadedState?: PreloadedState) =>
                 },
                 serializableCheck: false,
                 immutableCheck: false,
-            }).concat(middlewares),
+            })
+                .prepend(deviceConnectionMiddleware.middleware)
```
matejkriz, line 52: Why prepend and not just add it to the list of middlewares?

matejkriz: I think I get it now, it is a listener middleware and needs to execute after actions have been processed.

matejkriz:
It breaks the ability to initiate Device Settings actions before connecting the Trezor. Instead of continuing the intended flow, it redirects to the home screen.

suite-native/state/src/store.ts
juriczech: This is the reason for prepending: 
image

suite-native/device/src/hooks/useDeviceConnectionRouteBlacklist.ts
juriczech: I believe it's the same. The cleanup function is recreated on every run of the effect callback so I believe behavior should be the same 🤔 and it's just stylistic issue of what is less confusing. I'm happy to change it to your proposed solution if you think that's cleaner? 

If it's all the same, I like that there is no cleanup function (nor actual effect callback) if the condition isn't satisfied but I don't have a strong opinion here.

suite-native/device/src/hooks/useDeviceChecks.ts
juriczech: The behaviour is maintained and it's as it was. Or do you see a problem with the default return that I'm not realising? 
image

suite-native/device/src/hooks/useDeviceConnectionRouteBlacklist.ts
juriczech: [73a7af9](https://github.com/trezor/trezor-suite/pull/20081/commits/73a7af95fca510434a4cd7116aedef01fc5e19b7)

suite-native/device/src/hooks/useDeviceChecks.ts
juriczech: [73a7af9](https://github.com/trezor/trezor-suite/pull/20081/commits/73a7af95fca510434a4cd7116aedef01fc5e19b7)

juriczech:
> It breaks the ability to initiate Device Settings actions before connecting the Trezor. Instead of continuing the intended flow, it redirects to the home screen.

@matejkriz Good catch, thanks! 💪 
Fixed in: https://github.com/trezor/trezor-suite/pull/20081/commits/2df91613f54f06ac88a5bd3c24cce2db2636870b

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
juriczech: Fixed

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
juriczech: I actually find this cleaner and it's how the middleware is designed to be used. Let's discuss further if you disagree.

suite-native/device/src/hooks/useDeviceChecks.ts
matejkriz: Now I see that `failedCheck` is no longer used from this useDeviceChecks hook at all...

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -0,0 +1,87 @@
...
+            // https://github.com/trezor/trezor-suite/issues/20157
+            // If device is not initialized and is compromised, we display the modal (reason why this condition is here) and then want to redirect to uninitialized device landing.
+            (selectIsDeviceInitialized(currentState) || selectIsDeviceCompromised(currentState)),
+        effect: (
+            _action,
+            { getState }: ListenerEffectAPI<NativeDeviceRootState, Dispatch<UnknownAction>>,
+        ) => {
+            // Passphrase protected devices are only connected through passphrase form
+            // The passphrase flow handles connection differently and redirect to connecting screen is not wanted.
+            const isDeviceUsingPassphrase = selectIsDeviceUsingPassphrase(getState());
```
matejkriz: I understand the confusion, `selectIsDeviceUsingPassphrase` is clearly a wrong name of the selector, it doesn't mean that the device is using passphrase. It only means that the passphrase feature is enabled on the trezor. Let's fix the name in separate PR.

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -0,0 +1,87 @@
...
+
+            // If device is authorized already (usually in case of remembered device which has already been authorized)
+            const isDeviceConnectedAndAuthorized = selectIsDeviceConnectedAndAuthorized(getState());
+
+            if (isDeviceUsingPassphrase || isDeviceConnectedAndAuthorized) return;
```
matejkriz: I'm not sure about the position of this condition, wouldn't be better to have it below shouldNavigateToDeviceCompromisedModal to not skip it?

matejkriz:
My biggest concern with this approach is the global handling of starting and stopping the device connection listener. If multiple components have stopped the listener, any one of them can restart it on unmount, even if other components are still mounted.
A related issue is that if we stop the listener (for example, on the biometric screen), connect the Trezor on that screen, and then leave the screen, navigation never happens, right? Starting the listener is not executing the effect, right?

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
juriczech: Good catch! 
0788c4bad841dbb7bd07bfe5ec0f6fa60b8c4f56

suite-native/device/src/hooks/useDeviceChecks.ts
juriczech: Yes, removed in favor of `selectCompromisedDeviceFailedCheck` because it wasn't necessary to be inside the hook anymore. My aim is to get rid of this hook completely as well. However I don't want to refactor more than necessary as part of this PR, so I kept it, even though it's not really all that useful anymore and could just be used in `useHandleDeviceConnection`

suite-native/device/src/hooks/useDeviceConnectionRouteBlacklist.ts
juriczech: Redundant, see https://github.com/trezor/trezor-suite/pull/20081/commits/0054b20586c7a0ef85bdd6b5f788ed003083a561

suite-native/biometrics/src/components/BiometricOverlay.tsx
juriczech: Redundant, stopping listening is not used anymore. See https://github.com/trezor/trezor-suite/pull/20081/commits/0054b20586c7a0ef85bdd6b5f788ed003083a561

suite-native/device/src/hooks/useDeviceConnectionRouteBlacklist.ts
juriczech: It's not implemented as part of the middleware to simplify route blacklisting.
https://github.com/trezor/trezor-suite/pull/20081/commits/0054b20586c7a0ef85bdd6b5f788ed003083a561

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
juriczech: Redundant, see https://github.com/trezor/trezor-suite/pull/20081/commits/0054b20586c7a0ef85bdd6b5f788ed003083a561

suite-native/device/src/middlewares/deviceConnectionNavigationMiddleware.ts
juriczech: I will revalidate it once we also add the uninitialized device redirect if you're okay with this. I want to wait and see how it will all fit together and then make it more elegant. I agree with your point though 👍

suite-native/firmware/src/hooks/useFirmware.tsx
juriczech: Redundant https://github.com/trezor/trezor-suite/pull/20081/commits/0054b20586c7a0ef85bdd6b5f788ed003083a561

juriczech:
@matejkriz @yanascz @BrantalikP Thanks for your reviews. Based on your comments, I removed the hook, that was blacklisting routes using the `useFocusEffect`. I think it made sense to flag this as unnecessary complexity and just do it in the middleware with simple util function.

Currently, the listener middleware will be active in all of the app, unless conditionally stopped in the middleware, but still be only triggered by device connection. 

I would now like to merge this pretty soon if possible and build on top of this. Let's make sure there is nothing critically wrong here and merge this so we can continue with this approach.

suite-native/device/src/hooks/useDeviceChecks.ts
matejkriz: It really makes no sense to return `failedCheck` from `useDeviceChecks` if it's not read anywhere.

juriczech: I misunderstood what you meant and thought it's not exporting it anymore 🙈 Obviously that's right, I removed it and amended to latest commit, sorry for confusion!

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -0,0 +1,96 @@
...
+import { NativeDeviceRootState, selectIsDeviceCompromised } from '../selectors';
+
+const DEVICE_CONNECTION_BLACKLISTED_ROUTES = [
+    RootStackRoutes.DeviceCompromisedModal,
+    RootStackRoutes.DeviceOnboardingStack,
+];
+
+export const deviceConnectionMiddleware = createListenerMiddleware<NativeDeviceRootState>();
+
+const connectDevice = ({ isCoinEnablingInitFinished }: { isCoinEnablingInitFinished: boolean }) => {
```
matejkriz: Misleading name, this function is not connecting the device.

suite-native/navigation/src/navigators.ts
```diff
@@ -310,7 +310,7 @@ export type RootStackParamList = {
     [RootStackRoutes.SettingsScreenStack]: NavigatorScreenParams<SettingsStackParamList>;
     [RootStackRoutes.BackupFailedModal]: undefined;
     [RootStackRoutes.DeviceCompromisedModal]: {
-        failedCheck: 'device-authenticity' | 'entropy' | 'firmware-authenticity';
+        onClose: () => void;
```
matejkriz: Route params should be serializable, no callbacks. It would work until we want state persistence or deep link for this page, but would be nice to refactor it so the callback is not needed or turn of the ["Non-serializable values were found in the navigation state"](https://reactnavigation.org/docs/troubleshooting#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state) warning so it's obvious that it's intentional.

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -0,0 +1,96 @@
...
+        (selectIsDeviceInitialized(currentState) || selectIsDeviceCompromised(currentState)),
+    effect: (
+        _action: UnknownAction,
+        { getState }: ListenerEffectAPI<NativeDeviceRootState, Dispatch<UnknownAction>>,
+    ) => {
+        const shouldNavigateToDeviceCompromisedModal = selectIsDeviceCompromised(getState());
+        const isCoinEnablingInitFinished = selectIsCoinEnablingInitFinished(getState());
+
+        const isDeviceConnectionNavigationAllowed = checkIsRouteAccessAllowed({
+            blacklist: selectIsDeviceRemembered(getState())
```
matejkriz: Why is there a condition for device.remember for SendStack?

juriczech: You can access send flow with view only device and start creating transaction without connected device. If you connect device, we block the navigation to connecting screen and stay on send stack. 

This is how it was in `useHandleDeviceConnection` so I kept the functionality. Do you think it's worth a comment here explaining why?

juriczech: FYI: I think why it's not showing connecting screen is because connecting screen **always** navigates to homescreen after finishing. I see this as incorrect and ideally we would redirect back from where the connecting screen started. If it was like this, we could remove this and simply redirect from send stack => connecting screen => back to send stack, which would be nicer IMO. 

I think this is a hack to avoid redirect to homescreen.

suite-native/navigation/src/navigators.ts
juriczech: Simplified here.
https://github.com/trezor/trezor-suite/pull/20081/commits/3095eba3b2dd4f4a39bace9604575ff331538aab

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
juriczech: https://github.com/trezor/trezor-suite/pull/20081/commits/3095eba3b2dd4f4a39bace9604575ff331538aab

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
juriczech: After discussing, we don't have to use the device.remember and just always add the route to the blacklist.
https://github.com/trezor/trezor-suite/pull/20081/commits/65276fbeb3905ea71ae2760b1fb436dea9c7ca47

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -0,0 +1,89 @@
...
+        // If device is not initialized and is compromised, we display the modal (reason why this condition is here) and then want to redirect to uninitialized device landing.
+        (selectIsDeviceInitialized(currentState) || selectIsDeviceCompromised(currentState)),
+    effect: (
+        _action: UnknownAction,
+        { getState }: ListenerEffectAPI<NativeDeviceRootState, Dispatch<UnknownAction>>,
+    ) => {
+        const shouldNavigateToDeviceCompromisedModal = selectIsDeviceCompromised(getState());
+        const isCoinEnablingInitFinished = selectIsCoinEnablingInitFinished(getState());
+
+        const isDeviceConnectionNavigationAllowed = checkIsRouteAccessAllowed({
```
matejkriz: One more case to handle - if you connect different trezor than the one associated with the selected device. We should not skip the navigation to connecting screen in that case.

juriczech: This has already been the case in develop version too. Only place where we check this is in device settings using the `useDeviceChangedCheck`. Good catch pointing it out, but preferably, I would tackle this in a separate issue so we don't prolong this one any longer. 🙏

matejkriz: Ok.

suite-native/device/src/selectors.ts
```diff
@@ -255,3 +261,57 @@ export const selectShouldFactoryResetBeVisible = createMemoizedSelector(
     (isDeviceInBootloader, hasDeviceFirmwareInstalled) =>
         isDeviceInBootloader && hasDeviceFirmwareInstalled,
 );
+
+export const selectIsDeviceCompromised = createMemoizedSelector(
```
yanascz, line 265: These two selectors are great candidates for unit tests.

suite-native/module-device-onboarding/src/navigation/DeviceOnboardingStackNavigator.tsx
```diff
@@ -61,13 +61,12 @@ export const DeviceOnboardingStackNavigator = () => (
...
         />
-        <DeviceOnboardingStack.Screen
-            name={DeviceOnboardingStackRoutes.FirmwareInstallation}
+        <DeviceOnboardingStack.Screen name={DeviceOnboardingStackRoutes.FirmwareInstallation}
             component={FirmwareInstallationScreen}
         />
         <DeviceOnboardingStack.Screen
             name={DeviceOnboardingStackRoutes.ThpPairingInfo}
-            component={ThpPairingInfoScreen}
+                component={ThpPairingInfoScreen}
```
yanascz: Why is this file changed? 🤔

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -0,0 +1,89 @@
...
+        navigationContainerRef.navigate(RootStackRoutes.AuthorizeDeviceStack, {
+            screen: AuthorizeDeviceStackRoutes.ConnectingDevice,
+        });
+    } else {
+        navigationContainerRef.navigate(RootStackRoutes.CoinEnablingInit);
+    }
+};
+
+deviceConnectionMiddleware.startListening({
+    predicate: (action, currentState) =>
```
yanascz, line 48: What about adding a unit test for this predicate? Might be handy to have it covered to prevent accidental regressions.

suite-native/module-device-onboarding/src/navigation/DeviceOnboardingStackNavigator.tsx
juriczech: [fixup! feat(suite-native): handle thp device connection](https://github.com/trezor/trezor-suite/pull/20081/commits/f78aef598cb52a1fee26247cd7b3e16d09ef6de3)

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
juriczech: [fixup! feat(suite-native): handle thp device connection](https://github.com/trezor/trezor-suite/pull/20081/commits/f78aef598cb52a1fee26247cd7b3e16d09ef6de3)

suite-native/device/src/selectors.ts
juriczech: As discussed, I'd avoid doing test for this at this point as it's just moving code from elsewhere. We can do it as a follow up if needed, but since it wasn't tested up to this point, I don't want to bloat the PR.

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -0,0 +1,89 @@
...
+];
+
+export const deviceConnectionMiddleware = createListenerMiddleware<NativeDeviceRootState>();
+
+const handleDeviceConnectNavigation = ({
+    isCoinEnablingInitFinished,
+}: {
+    isCoinEnablingInitFinished: boolean;
+}) => {
+    // If coin enabling is not finished, it takes priority over connecting screen
```
yanascz: This comment seems redundant as it's confusing – you first mention _not finished_ and then the condition is `if (isCoinEnablingInitFinished)`. I'd just drop it, I don't see any added value.

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -0,0 +1,89 @@
...
+        { getState }: ListenerEffectAPI<NativeDeviceRootState, Dispatch<UnknownAction>>,
+    ) => {
+        const shouldNavigateToDeviceCompromisedModal = selectIsDeviceCompromised(getState());
+        const isCoinEnablingInitFinished = selectIsCoinEnablingInitFinished(getState());
+
+        const isDeviceConnectionNavigationAllowed = checkIsRouteAccessAllowed({
+            blacklist: DEVICE_CONNECTION_BLACKLISTED_ROUTES,
+        });
+
+        if (!isDeviceConnectionNavigationAllowed) return;
```
yanascz: I'd personally inline the variable, since it's used just once and the function being called is descriptive enough.

yanascz (approved):
🚢🇮🇹

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
juriczech: True, I messed up the comment copy 🥶 Removing.

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
juriczech: 👌 fixed

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
juriczech: Fixed

PR #20085, opened 2025-07-09
feat(suite): throttle Solana RPC requests to avoid hitting rate limits

### LINKED ISSUE #19872: Solana account discovery error for token with large history
**Describe the bug**
Solana account discovery fails when a token has a large transaction history. The app is hitting rate limits on the RPC. It's doing an endless number of "getSignaturesForAddress" RPC calls for the tokens accounts. See screenshot below from the web app.

Image
Image

Behaviour is similar on the desktop app.

To confirm the bug, I built the app from sources and made a change here:
https://github.com/trezor/trezor-suite/blob/766356e7700d10ba024bc4abd8fa7622a7ce74b5/packages/blockchain-link/src/workers/solana/index.ts#L101

I break out of this loop to avoid endlessly loading transactions. After this change the account discovery succeeds. The history sized to be fetched needs to be capped. 

**Info:**

- Remembered Wallet: yes/no
- Suite Version: app/web, commit
- Browser: name, version
- Browser Installation Method: snap, flatpak, app store, windows store, downloaded installer, etc.
- OS: name, version
- Bridge: version
- Firmware Version: model, version/commit

**How to reproduce**

1. Click on '....'
1. Scroll down to '....'
1. See error

izmy:
## Description

Adds a throttling wrapper for Solana RPC transport, based on [anza-xyz/kit example](https://github.com/anza-xyz/kit/blob/main/examples/rpc-transport-throttled/src/example.ts).
Limits maximum requests per second with a token bucket approach to avoid RPC rate limits.

Current limit for Solana is 4 requests per 100ms

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19872

## Video

https://github.com/user-attachments/assets/f58780e6-43fb-4a8b-8e94-4fde7204e674

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/solana-throttled-transport&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/solana-throttled-transport&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/solana-throttled-transport&lookback=7)


mroz22:
would be great if @marekrjpolak could check on this a bit - we already have similar pieces of logic so it might make sense to consider sharing

packages/blockchain-link/src/workers/solana/utils/getThrottledTransport.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Throttled RPC transport for Solana.
+ *
+ * Based on https://github.com/anza-xyz/kit/blob/main/examples/rpc-transport-throttled/src/example.ts
+ *
+ * Adds a simple fixed-window queue (token bucket style) that limits requests per second (RPS).
+ *
+ * Safe for use with createSolanaRpcFromTransport.
+ */
+export function getThrottledTransport<TClusterUrl extends ClusterUrl>(
```
mroz22: for example stuff like this https://github.com/trezor/trezor-suite/blob/develop/suite-common/fiat-services/src/limiter.ts#L6

maybe we could haave some cleverer rate limiter somewhere in utils

packages/blockchain-link/src/workers/solana/index.ts
```diff
@@ -835,9 +837,15 @@ class SolanaWorker extends BaseWorker<SolanaAPI> {
             },
         });
 
+        const throttledTransport = getThrottledTransport(transport, {
```
tomasklim, line 840: would it be possible to throttle transport of all solana requests together?

tomasklim (changes requested):

packages/blockchain-link/src/workers/solana/index.ts
```diff
@@ -835,9 +837,15 @@ class SolanaWorker extends BaseWorker<SolanaAPI> {
...
+            interval: 300,
+            maxRps: 10,
```
tomasklim: so this limits it to 33 requests per second?

packages/blockchain-link/src/workers/solana/utils/getThrottledTransport.ts
izmy: I wouldn’t move this to a shared limiter, because it’s typed and made just for Solana

packages/blockchain-link/src/workers/solana/index.ts
izmy: I changed the limit to 4 requests per 100ms

tomasklim (approved):
LGTM, works good

PR #18146, opened 2025-04-04
Standalone bridge deprecation. Phase 3

mroz22:
### Phase 1 [already in production]

- suite: show banner with copy to uninstall non-canonicall standalone bridges
- stop offering standalone bridge installation as CTA

### Phase 2 [already in production]

- #18754 stop offering standalone bridge installation (escape hatch)
-  update copy 
- auto-starting built in variant in suite-desktop #17746 after the standalone one is gone.
- make suite-desktop background mode available in production ✅ 
- set node-bridge as default in CI for connect tests #19214
- give user an option to switch between implementation in troubleshooting tips. just in case #19863

### Phase 3 - oblivion [now]
- in old connect popup - present a deeplink to open suite-desktop instead so that users start running suite in the background more frequently

image
image

- get node-bridge rollout to 100%. #19209
- start using new port for bridge  #20245 
 -  mark all non non-built in bridges as outdated and show copy (banner, modal) with instructions to uninstall. 

image

image

### Phase 4 - death 
- kill public bridge (and Bluetooth) endpoint, start using a single trezor-connect websocket endpoint

Related issues
- All issues [labeled as trezord-2.0.27 ](https://github.com/trezor/trezor-suite/issues?q=is%3Aissue%20state%3Aopen%20label%3Atrezord-2.0.27) should go away

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=bridge-deprecated&lookback=7)

PR #20103, opened 2025-07-10
node-utils: http server port negotiation

mroz22:
we might want to instruct http server to chose from an array of possible ports to listen on. This is useful to get rid of rare edgecases when a certain user might have the ports we rely on occupied.

usecases:
- http receiver
- node-bridge once we decide to switch to a new port as part of the old bridge deprecation #18146

regarding the security benefits (ttps://github.com/trezor/trezor-suite-private/issues/9)  - I am afraid it won't bring any since we would need to use random port for this and this is not possible at least for services that need url whitelisting 

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=port-negotiation&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=port-negotiation&lookback=7)

### DIFF
--- a/packages/node-utils/src/http.ts
+++ b/packages/node-utils/src/http.ts
@@ -7,7 +7,6 @@ import type { RequiredKey } from '@trezor/type-utils';
 import { Log, TypedEmitter, arrayPartition } from '@trezor/utils';
 
 import { findProcessFromIncomingPort } from './findProcessFromIncomingPort';
-import { getFreePort } from './getFreePort';
 
 type Request = RequiredKey<http.IncomingMessage, 'url'>;
 const isRequest = (request: http.IncomingMessage): request is Request => request.url !== undefined;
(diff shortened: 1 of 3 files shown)


packages/suite-desktop-core/src/libs/http-receiver.ts
```diff
@@ -64,10 +64,12 @@ const applyTemplate = (content = 'You may now close this window.', options?: Tem
     return template;
 };
 
+const PORTS = [21335, 21336, 21337, 21338, 21339];
```
mroz22: at least some endpoints will need additional whitelisting (labeling). not sure how about invity features.

karliatto (approved):
Looks nice to me, I understand this is just preparation since I do not see that new functionality of using `ports` used anywhere.

PR #20111, opened 2025-07-10
chore(message-system): cleanup feature flag types

### LINKED ISSUE #20110: Add support for raw value in message-system feature
Currently, Message System `feature` only bears `flag: boolean` as payload (on|off).
But we need a way to remotely control a piece of data, that will be specifically handled by the consuming code.

Possible use cases:
- remotely controlled numerical parameter
- `A|B|C|off` toggle instead of just `on|off`

Lemonexe:
## Description

Message system `feature` can bear arbitrary payload already, so clean up the type definitions.

:thought_balloon:  the `flag: boolean` still has sense semantically when you use arbitrary payload: whether the remotely controlled value shall be heeded or discarded.

## Related Issue

Related to #20110, only to make things clearer

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/add-message-system-feature-value&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/add-message-system-feature-value&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/add-message-system-feature-value&lookback=7)

### DIFF
--- a/suite-common/message-system/schema/config.schema.v1.json
+++ b/suite-common/message-system/schema/config.schema.v1.json
@@ -121,7 +121,7 @@
                                 "minItems": 1,
                                 "items": {
                                     "title": "Feature",
-                                    "description": "Only used for 'feature' category. Feature flag can disable a feature for a specific version of a specific app.",
+                                    "description": "Only used for 'feature' category. Feature flag can disable a feature for a specific version of a specific app. Custom properties may be used to pass additional data to the feature flag.",
                                     "type": "object",
                                     "required": ["domain", "flag"],
                                     "properties": {
@@ -129,9 +129,6 @@
                                             "type": "string"
                                         },
                                         "flag": {
-                                            "type": "boolean"
-                                        },
-                                        "isPublic": {
                                             "type": "boolean"
                                         }
                                     },
--- a/suite-common/message-system/src/messageSystemSelectors.ts
+++ b/suite-common/message-system/src/messageSystemSelectors.ts
@@ -134,7 +134,7 @@ export const selectIsFeatureDisabled = (
 ) => {
     const featureFlag = selectFeatureConfig(state, domain)?.flag;
 
-    return typeof featureFlag === 'boolean' ? !featureFlag : (defaultValue ?? false);
+    return featureFlag !== undefined ? !featureFlag : (defaultValue ?? false);
 };
 
 const selectValidMessages = (state: MessageSystemRootState) => state.messageSystem.validMessages;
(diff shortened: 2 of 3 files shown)


suite-common/message-system/src/messageSystemSelectors.ts
```diff
@@ -134,9 +134,12 @@ export const selectIsFeatureDisabled = (
 ) => {
     const featureFlag = selectFeatureConfig(state, domain)?.flag;
 
-    return typeof featureFlag === 'boolean' ? !featureFlag : (defaultValue ?? false);
+    return featureFlag !== undefined ? !featureFlag : (defaultValue ?? false);
```
Lemonexe, line 137: incidental codestyle fix, I believe this is more idiomatic

suite-common/message-system/schema/config.schema.v1.json
```diff
@@ -130,9 +130,6 @@
                                         },
                                         "flag": {
                                             "type": "boolean"
-                                        },
-                                        "isPublic": {
-                                            "type": "boolean"
```
Lemonexe, line 135: IMHO it isn't right to define `isPublic` here. This property is specific **only** for [Feature.coinjoin](https://github.com/trezor/trezor-suite/blob/84b3a674429954068791f1f8b45a59021873a40c/packages/suite/src/reducers/wallet/coinjoinReducer.ts#L1056), but defining it in the schema makes it seem as if it was relevant for _every_ feature flag, which is not.

I consider it better to leave custom properties untyped, and make it responsibility of the consuming code to parse the expected values from `unknown`.
`isPublic` is already handled in a way so I didn't have to change anything, see link above :slightly_smiling_face:

suite-common/suite-types/src/messageSystem.ts
```diff
@@ -406,12 +406,11 @@ export interface Modal {
     image: string;
 }
 /**
- * Only used for 'feature' category. Feature flag can disable a feature for a specific version of a specific app.
+ * Only used for 'feature' category. Feature flag can disable a feature for a specific version of a specific app. Custom properties may be used to pass additional data to the feature flag.
  */
 export interface Feature {
     domain: string;
```
Lemonexe, line 412: :memo: mental note: this is autogenerated file, and that's why domain is `string`, although we have the type union available [in suite-common/message-system](https://github.com/trezor/trezor-suite/blob/f2b51776363ba3334476e7518e76e271d18fc077/suite-common/message-system/src/messageSystemTypes.ts#L71)

That's because the json config schema, being the source of truth, does not allow "a type union", only general declaration.

jvaclavik (approved):
LGTM

PR #20134, opened 2025-07-11
Hide the signature format switch if it doesn't matter

### LINKED ISSUE #7674: hide the signature format switch where it doesn't make a difference
coins where message signing is enabled display the Trezor/Electrum signature format switch. I'm unsure if it makes sense for non-Bitcoin coins with Electrum-fork wallets (Litecoin, Dash), but it's definitely meaningless for those without. e.g. for Ethereum the signature is the same even if I choose the Electrum format. I suggest removing the switch for coins where it doesn't make a difference.

Image

5K1PY:
The format switch in Sign & Verify is hidden if the signiture is same in both formats - Only shown for non-legacy bitcoin and litecoin

## Related Issue

Resolve #7674

## Screenshots

Bitcoin:
sign-bitcoin

Ethereum:
sign-ethereum

### DIFF
--- a/packages/suite/src/views/wallet/sign-verify/index.tsx
+++ b/packages/suite/src/views/wallet/sign-verify/index.tsx
@@ -104,6 +104,12 @@ const SignVerify = () => {
         }
     };
 
+    // Empty accountTypes means there is only 'normal' accountType and therefore the signatures are same.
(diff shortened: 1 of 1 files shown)


komret (approved):

komret:
Nice! Rebase please.

PR #18136, opened 2025-04-03
Feat: use UAParser2 for message system

Lemonexe:
## Description

:eyes: CR commit by commit recommended.

Followup to https://github.com/trezor/trezor-suite/pull/18023:
use the new async `getOsVersion` in message system, so that we can distinguish macOS > catalina (currently targetted as if it was catalina), and Windows 11 (currently targetted as if it was 10).

Also, cleanup message system test fixtures :broom: 

## Related Issue

Resolve #8674

## Screenshots:

You can test it [on this branch.](https://github.com/trezor/trezor-suite/compare/feat/use-uaparser2-for-message-system..feat/use-uaparser2-for-message-system-TESTING).

image

For a very quick simulation, paste this on page load in console and play around:
```
OS_NAME = 'windows';
OS_VERSION = '11';
```

```
OS_NAME = 'macos'
OS_VERSION = "15.3.2"
UA='Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36';
```

PR #20149, opened 2025-07-11
refactor(message-system): make message-system synchronous again by caching async osVersion

Lemonexe:
## Description

In #18136, message system evaluation [was made async](https://github.com/trezor/trezor-suite/blob/bf67a42a885270542bc63bcad8207ade3d4aaa6c/suite-common/message-system/src/messageSystemUtils.ts#L304), because the envData necessary to make the decisions became async, because of new `getOsVersion`.

This is partially a revert of 31c9d343, and a simple trick circumvent this issue:

- await osVersion only once at Suite/Suite Lite startup and cache it
- in message system decisions
  - use the cached osVersion
  - call the rest of the env utils (synchronous) whenver needed, as it was before 31c9d343
  - and make it all synchronous again, as it was before 31c9d343

The **overall goal:** have message system decions resolved [at this point](https://github.com/trezor/trezor-suite/pull/20149/files#r2205019020), it's needed for future work.

The async message-system redux makde things very complex, but at the time it didn't hurt anything.
However, the real problem is **[in the middleware](https://github.com/trezor/trezor-suite/blob/bf67a42a885270542bc63bcad8207ade3d4aaa6c/packages/suite/src/middlewares/suite/messageSystemMiddleware.ts#L48-L51)**, it cannot be awaited!
So awaiting dispatch initMessageSystemThunk is **not** sufficient, beacuse the middleware function runs async in parallel, unawaitable. Maybe it could be hacked by DFD but that would be even worse mess.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=refactor/rearrange-async-logic-in-message-system&lookback=7)


packages/suite/src/middlewares/suite/__tests__/messageSystemMiddleware.test.ts
Lemonexe, line 1: this is revert of 31c9d343 (no conflicts, no additional changes)

suite-native/message-system/src/messageSystemMiddleware.ts
```diff
@@ -20,37 +20,33 @@ const isAnyOfMessageSystemAffectingActions = isAnyOf(
...
+        const validExperimentIds = getValidExperimentIds(config, validationParams);
+
+        const categorizedValidMessages = categorizeMessages(validMessages);
+
+        dispatch(messageSystemActions.updateValidMessages(categorizedValidMessages));
+        dispatch(messageSystemActions.updateValidExperiments(validExperimentIds));
+    }
+
+    return action;
+});
```
Lemonexe, line 52: this is the same diff as [packages/suite/src/middlewares/suite/messageSystemMiddleware.ts](https://github.com/trezor/trezor-suite/pull/20149/files#diff-7d33be32b057462f2a4bc1ef8aa46207040349149135a2f3568bc3668078be1e)
but indentation changes make it hard to discern.

packages/suite/src/middlewares/suite/__tests__/messageSystemMiddleware.test.ts
```diff
@@ -75,7 +75,7 @@ const initStore = (preloadedState: State) => {
 };
 
 describe('Message system middleware', () => {
-    it('prepares valid messages for being displayed', async () => {
+    it('prepares valid messages for being displayed', () => {
```
Lemonexe, line 78: this **whole file** is revert of 31c9d343 (no conflicts, no additional changes)

packages/suite/src/middlewares/suite/__tests__/messageSystemMiddleware.test.ts
Lemonexe, line 1: this is revert of 31c9d343 (no conflicts, no additional changes)

packages/suite/src/actions/suite/__tests__/initAction.test.ts
```diff
@@ -88,25 +88,25 @@ const fixtures: Fixture[] = [
         actions: [
             SUITE.INIT,
             initDevices.pending.type,
+            initDevices.fulfilled.type,
             SUITE.SET_LANGUAGE,
             initMessageSystemThunk.pending.type,
             fetchConfigThunk.pending.type,
+            messageSystemActions.fetchSuccessUpdate.type,
+            fetchConfigThunk.fulfilled.type,
+            initMessageSystemThunk.fulfilled.type,
```
Lemonexe, line 97: this it may look like I'm just rearranging fixtures to fit reality, but if you look closely, it makes sense this way.
Between `initMessageSystemThunk.pending` and `initMessageSystemThunk.fulfilled` it's all related to this thunk, because it is awaited.

But previously, thunks were running async in parallel, and the order of actions in fixture **was not deterministic**

packages/suite/src/actions/suite/initAction.ts
```diff
@@ -56,7 +56,8 @@ export const init = () => async (dispatch: Dispatch, getState: GetState) => {
     dispatch(languageActions.setLanguage(language));
 
     // 3. fetch message system config
-    dispatch(initMessageSystemThunk());
+    await prepareCachedEnvData();
+    await dispatch(initMessageSystemThunk());
```
Lemonexe, line 60: awaiting `initMessageSystemThunk` to make sure it is ready before `connectInitThunk`, will be used for future work...

packages/suite/src/actions/suite/initAction.ts
```diff
@@ -40,7 +40,7 @@ export const init = () => async (dispatch: Dispatch, getState: GetState) => {
 
     dispatch({ type: SUITE.INIT });
 
-    dispatch(initDevices());
+    await dispatch(initDevices());
```
Lemonexe, line 43: main change below; this is just a tweak, because `initDevices` is a **synchronous** thunk, with no heavy calculations either, so it makes no sense to let it run async in parallel (and it did, see the changed test fixtures).
This simplifies the flow :slightly_smiling_face:

suite-common/message-system/src/__fixtures__/messageSystemUtils.ts
```diff
@@ -1349,9 +1349,7 @@ type GetValidMessagesFixture = {
     description: string;
     currentDate: string;
     userAgent: string;
-    // For some systems, userAgent may not be sufficient to determine osVersion (see envUtils.ts).
-    // To simulate those cases, instead of mocking userAgent, getOsVersion will be mocked directly
-    osVersion?: string;
+    osVersion: string;
```
Lemonexe, line 1352: I had a lot of dificulty with jest, [see below](https://github.com/trezor/trezor-suite/pull/20149/files#diff-26f1d8da7b9df7cb7e2e940e0338c3035afc65a7c396a887c866a6bac82e82f5)
So I explicitly defined osVersion, and userAgent is now used only for the other env utils functions.

suite-common/message-system/src/__tests__/messageSystemUtils.test.ts
```diff
@@ -116,26 +117,24 @@ describe('Message system utils', () => {
...
 
-                const { osVersion } = f;
-                if (osVersion) {
-                    jest.spyOn(envUtils, 'getOsVersion').mockImplementation(() =>
-                        Promise.resolve(osVersion),
-                    );
-                }
+                jest.spyOn(cachedEnvData, 'getCachedOsVersion').mockImplementation(
+                    () => f.osVersion,
+                );
```
Lemonexe, line 135: :warning: rabbit hole stuff: at first I wanted to keep using the User Agent mock by explicitly calling `await prepareCachedEnvData()` here, and let it populate the cache with return of real `getOsVersion`, using the mocked user agent from fixtures.
I gave up after an hour of really mysterious debugging. In the end I determined that _calling anything_ from file `cachedEnvData.ts ` in the test, actually ruins `getOsVersion`, so it returns undefined, although that lives in another file..?
It kinda looks like how jest behaves with circular deps, but that is definitely not the case here.
Just some dark magic of broken jest module resolution, which I warded off by mocking `getCachedOsVersion` with explicit value.

After all, we don't need to test a library function (no need to extend the test to cover `getOsVersion` parsing useragent). I just left the useragent for the `getOsName` to keep changes minimal..

suite-common/message-system/src/cachedEnvData.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Caches env data that has to be fetched async and is necessary for message-system decisions,
+ * so that message-system functions can be used synchronously.
+ * Currently it's only the OS version on Suite Web/Desktop (but async fn also used on mobile for consistent type).
+ */
+export const prepareCachedEnvData = async () => {
+    cachedOsVersion = await getOsVersion();
+};
+
+export const getCachedOsVersion = () => cachedOsVersion;
```
Lemonexe, line 14: this file it's little bit hacky, but a pattern we already do often.
I sincerely belive that this trick is worth, it because async message system thunks are a hard price to pay :sweat:

suite-common/message-system/src/messageSystemUtils.ts
```diff
@@ -215,30 +215,18 @@ export const isCountryCodeCompatible = (
...
-    osName: ReturnType<typeof getOsName>;
-    osVersion: ReturnType<typeof transformVersionToSemverFormat>;
-    browserName: ReturnType<typeof getBrowserName>;
-    browserVersion: ReturnType<typeof transformVersionToSemverFormat>;
-    environment: ReturnType<typeof getEnvironment>;
-    suiteVersion: ReturnType<typeof transformVersionToSemverFormat>;
-    commitHash: ReturnType<typeof getCommitHash>;
-};
+export const validateConditions = (condition: Condition, options: Options) => {
+    const { device, transports = [], settings, countryCode } = options;
```
Lemonexe, line 219: this is **partially** a revert of 31c9d343. Recreated manually due to conflicts; and the `osVersion` is changed.

Lot of changes have been done since 31c9d343, but you can see all the diffs in the relevant file betwen then and now:
```
git diff 31c9d343~1...refactor/rearrange-async-logic-in-message-system suite-common/message-system/src/messageSystemUtils.ts
```

suite-native/module-trading/src/hooks/buy/useBuyQuotes.ts
```diff
@@ -181,18 +181,16 @@ const useBuyQuotesThunk = (
...
                 quotesPromiseRef.current = requestPromise;
+                const action = await requestPromise;
+                if (isFulfilled(action) && (action.payload as BuyTrade[]).length > 0) {
+                    analytics.report({
+                        type: EventType.TradingQuoteReceived,
+                        payload: {
+                            type: 'buy',
+                        },
+                    });
+                }
```
Lemonexe, line 193: so this was another huge rabbit hole of debugging :facepalm: 


I was very surprised when a quite unrelated unit test `useBuyQuotes.test.ts` failed, with exception thrown from application code:
```
  ● useBuyQuotes › should query quotes once all required data is selected

    TypeError: requestPromise.then is not a function

      182 |                 };
      183 |                 const requestPromise = dispatch(buyThunks.handleRequestThunk(payload));
    > 184 |                 requestPromise.then(action => {
          |                                ^
      185 |                     if (isFulfilled(action) && (action.payload as BuyTrade[]).length > 0) {
      186 |                         analytics.report({
      187 |                             type: EventType.TradingQuoteReceived,

      at then (src/hooks/buy/useBuyQuotes.ts:184:32)
      at fn (src/hooks/buy/__tests__/useBuyQuotes.test.ts:28:51)
```
- **And this is where the fun begins:**
- I debugged that the critical part which breaks it is change of `suite-native messageSystemMiddleware` from async to sync.
- Doesn't matter the content, still broken. I can remove the middleware, still broken.
- Quite the contary, this fixes it: https://github.com/trezor/trezor-suite/commit/8861418b44d2af70d943d13eb69e377fb1204afb
- Yes, the test needed _at least one async middleware_...
- Here we have mocked [handleRequestThunkMock](https://github.com/trezor/trezor-suite/blob/445b290d687a6e521d39ee095193864341c2c45b/suite-native/module-trading/src/hooks/buy/__tests__/useBuyQuotes.test.ts#L36-L44)
  - :thought_balloon: note: it's not ideal that an async thunk is mocked as a mere `action` in a test, it's a slight deviation from reality but that's ok, for unit tests it's fine.
- But the test worked because of an async middleare, so that this line really returns a promise: `dispatch(buyThunks.handleRequestThunk(payload))`
- So in 7a41d4d6b65a6767b6fdfac5bd7b6617057661b9 I rewrote the runtime code from `.then` to async await. It's more idiomatic code, and it solves the issue because `await` passes through a non-promise value 
  - Though it would be ideal to mock, it's more difficult than it sounds. We did not suceed at it with @jbazant , and discussed privately that this solution is ok :heavy_check_mark:

packages/suite/src/middlewares/suite/__tests__/messageSystemMiddleware.test.ts
Lemonexe, line 78: this is revert of 31c9d343 (no conflicts, no additional changes))

suite-common/message-system/src/__fixtures__/messageSystemUtils.ts
peter-sanderson: :heart:

peter-sanderson (approved):

PR #20155, opened 2025-07-13
fix(native): fetch all transactions in SendUtxoScreen

OriginalEveres:
This PR fixes fetching additional data (mainly timestamp and data for `TransactionDetail` screen) in `SendUtxoScreen` (Coin Control feature) on mobile.

Unfortunatelly we have to fetch all account's transaction since connect is not capable of fetching specific `tx`. (see [docs](https://connect.trezor.io/9/methods/bitcoin/getAccountInfo/#get-account-info), but this behaviour is same on desktop) Ideal solution would be fetching details only for needed UTXOs.

**Changes**: 
- dispatch `fetchAllTransactionsForAccountThunk ` for selected `accountKey` when user opens a Coin Control

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/coin-control-fetch-all-transactions&lookback=7)

### DIFF
--- a/suite-native/module-send/src/screens/SendUtxoScreen.tsx
+++ b/suite-native/module-send/src/screens/SendUtxoScreen.tsx
@@ -1,10 +1,14 @@
-import { useCallback, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { TextInput } from 'react-native';
-import { useSelector } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 
 import { useNavigation } from '@react-navigation/native';
 
-import { AccountsRootState, selectAccountByKey } from '@suite-common/wallet-core';
+import {
+    AccountsRootState,
+    fetchAllTransactionsForAccountThunk,
+    selectAccountByKey,
+} from '@suite-common/wallet-core';
 import { useFilteredUtxos } from '@suite-common/wallet-utils';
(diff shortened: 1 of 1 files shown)


suite-native/module-send/src/screens/SendUtxoScreen.tsx
```diff
@@ -37,6 +42,21 @@ export const SendUtxoScreen = ({
...
+            fetchAllTransactionsForAccountThunk({
+                accountKey: account.key,
+                noLoading: true,
+            }),
+        );
+
+        return () => {
+            promise.abort();
+        };
+    }, [account, dispatch]);
```
vytick: nit: `account.key` might be sufficient?

vytick (approved):
looks good 💪 , just one nit

suite-native/module-send/src/screens/SendUtxoScreen.tsx
OriginalEveres: if I would include only `account.key` as dep - the linter would complain, because of this line `if (account === null) return;`

vytick: yes. and that key is only needed thing for that thunk. so no need to check whole account nor refetch on account object change

OriginalEveres: I did simplify it a bit more and I have used the `accountKey` prop.

vytick (approved):

PR #20179, opened 2025-07-14
refactor(suite-native): introduce firmware update stack navigator

yanascz:
Use a stack to simplify going back to the initial screen. It may either be the dashboard, or device settings.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=refactor/native/firmware-update-navigation&lookback=7)

### DIFF
--- a/suite-native/navigation/src/navigators.ts
+++ b/suite-native/navigation/src/navigators.ts
@@ -25,6 +25,7 @@ import {
     DeviceOnboardingStackRoutes,
     DevicePinProtectionStackRoutes,
     DeviceSettingsStackRoutes,
+    FirmwareUpdateStackRoutes,
     HomeStackRoutes,
     OnboardingStackRoutes,
     ReceiveStackRoutes,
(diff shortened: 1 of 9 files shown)


suite-native/navigation/src/navigators.ts
```diff
@@ -202,9 +203,8 @@ export type DeviceSettingsStackParamList = {
         type: PinActionType;
     };
     [DeviceSettingsStackRoutes.DeviceAuthenticity]: undefined;
+    [DeviceSettingsStackRoutes.FirmwareUpdateStack]: undefined;
```
PeKne: The type of the stack has to be provided otherwise the typing won't work as expected.

```suggestion
    [DeviceSettingsStackRoutes.FirmwareUpdateStack]: NavigatorScreenParams<FirmwareUpdateStackParamList>;
```

PeKne (commented):
one nit

juriczech (approved):
Aside from the type, it LGTM 👌 

I'm wondering about how we have multiple stacks in this module and whether they should be separated into multiple packages. But I guess it's just because we're interacting with device in this module on each flow and need this because of navigation so it kind of makes sense. Just thinking out loud.

suite-native/navigation/src/navigators.ts
yanascz: Good catch, thanks!

Issue #20129, opened 2025-07-11
Discovery runs on not initialized device


szymonlesisz:
i just noticed this warning in console while working with not initialized (factory reset) device.
i think discovery should not be be started in that case, no?
```
startDiscoveryThunk: discovery already in progress, cancelling start call
```

another issue that i've found is that discovery will never start if connected device is unacquired (or THP before pairing) and becomes acquired. i've fixed that in my branch but it needs some love:

https://github.com/trezor/trezor-suite/commit/2639d6ed9620e3ea04a5bdb8a6dacb5fc8e0b8f6

`discoveryMiddleware`:
```
            // device becomes acquired from unacquired in the middle of the process (acquireDevice)
            // device is still locked when it happens
            const theFix = becomesAcquired ? false : isDeviceLocked;
            if (device && device.connected && isDeviceAcquired(device) && !theFix) {
```

PR #20187, opened 2025-07-14
fix(suite): start discovery when device becomes acquired

szymonlesisz:
## Description
fixing second issue described in https://github.com/trezor/trezor-suite/issues/20129

## How to test it

1. Start suite with external (not build-in) bridge transport, go to account and show receive address.
2. refresh suite
3. connected device is unacquired (used in another window)
4. click on "use device here"

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/suite-discovery-unacuired&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/suite-discovery-unacuired&lookback=7)

### DIFF
--- a/packages/suite/src/components/suite/AcquireDeviceButton.tsx
+++ b/packages/suite/src/components/suite/AcquireDeviceButton.tsx
@@ -20,7 +20,7 @@ export const AcquireDeviceButton = ({ className, onClick }: AcquireButtonProps) 
 
     const handleClick: MouseEventHandler = e => {
         onClick?.(e);
-        dispatch(acquireDevice({ startDiscovery: true }));
+        dispatch(acquireDevice({}));
     };
 
     return (
--- a/packages/suite/src/components/suite/PrerequisitesGuide/DeviceAcquire.tsx
+++ b/packages/suite/src/components/suite/PrerequisitesGuide/DeviceAcquire.tsx
@@ -18,7 +18,7 @@ export const DeviceAcquire = () => {
 
     const handleClick: MouseEventHandler = e => {
         e.stopPropagation();
-        dispatch(acquireDevice({ startDiscovery: true }));
+        dispatch(acquireDevice({}));
     };
 
     const ctaButton = (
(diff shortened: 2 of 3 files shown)


mroz22:
@marekrjpolak now spent some time on fixing discovery issues, so he might have good insights to this

marekrjpolak (approved):
While I agree with the fix, I had problems reproducing the issue. The reason probably is that I used the wrong **Acquire** button: if the `acquireDevice` action is dispatched with `startDiscovery` param, it doesn't need the fix, but not all the buttons pass the param.

I'm not sure what the right approach is, but a) it may be possible to fix this issue by using the param instead and b) it's also possible that with your fix the param is no longer necessary and can be removed. Your call.

szymonlesisz:
ok so we have 2 overlapping approaches:

1. automatically starts the discovery in the middleware by handling `DEVICE.CONNECT` event which is emitted when device becomes acquired **but before it becomes unlocked** so basically `becomesAcuired` condition in the middleware is never fulfilled

2. manually (click) using `acquireDevice({ startDiscovery: true })`  if we want to use this approach everywhere then startDiscovery should be true by default  as there are multiple other buttons which should start the discovery after successful acquiring. im focusing mainly on DeviceTrezorHostProtocolPair component but NeedsAttentionBanner or AcquireDeviceButton looks like would need it also? 
the only component which should use startDiscovery: false is `FirmwareModal` i guess

packages/suite/src/components/suite/AcquireDeviceButton.tsx
```diff
@@ -20,7 +20,7 @@ export const AcquireDeviceButton = ({ className, onClick }: AcquireButtonProps)
 
     const handleClick: MouseEventHandler = e => {
         onClick?.(e);
-        dispatch(acquireDevice({ startDiscovery: true }));
+        dispatch(acquireDevice({}));
```
marekrjpolak, line 23: May the param be removed from `DeviceAcquire.tsx` as well?

szymonlesisz: sure i didnt noticetd that

PR #20049, opened 2025-07-08
fix(transport-bridge): error propagation

szymonlesisz:
## Description
fix transport-bridge error propagation.

errors from bridge server are returned to connect as TRANSPORT.ERROR:
```
@trezor/transport-bridge Handling request for POST /read/1
@trezor/transport-bridge core: readUtil protocol v1
@trezor/transport-bridge usb: device.transferIn
@trezor/transport-bridge usb: device.transferIn done. status: ok, byteLength: 64.
@trezor/transport-bridge core: readUtil catch: Malformed protocol format
```

however due to this bug it is always parsed to unexpected error. this was changed relatively not so long ago:
https://github.com/trezor/trezor-suite/commit/2dcaafd766e411ac9e19390ce09b244acf369df3
👀 👉 @mroz22 

```
Device sending a preventive cancel on the first encounter with the device
DeviceCommands Sending Cancel {}
@trezor/connect handleMessage device-changed
@trezor/transport BridgeTransport unexpected error:  unexpected error
DeviceCommands Received transport error unexpected error unexpected error
DeviceCommands Sending GetFeatures {}
@trezor/transport BridgeTransport unexpected error:  unexpected error
DeviceCommands Received transport error unexpected error unexpected error
Device Device._runInner error:  unexpected error
```

this is crucial fix to properly catch `THP` protocol during `Device.handshake` process while using NodeBridge

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/bridge-errors&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/bridge-errors&lookback=7)

PR #20191, opened 2025-07-14
fix(connect): transport error propagation

szymonlesisz:
## Description

another place with "unexpected error". continuation of https://github.com/trezor/trezor-suite/pull/20049

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/connect-transport-error&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/connect-transport-error&lookback=7)

### DIFF
--- a/packages/connect/src/device/DeviceList.ts
+++ b/packages/connect/src/device/DeviceList.ts
@@ -277,7 +277,7 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> implements IDevic
         const enumerateResult = await transport.enumerate({ signal });
 
         if (!enumerateResult.success) {
-            throw new Error(enumerateResult.error);
+            throw new Error(enumerateResult.message || enumerateResult.error);
         }
 
         const descriptors = enumerateResult.payload;


packages/connect/src/device/DeviceList.ts
```diff
@@ -224,7 +224,7 @@ export class DeviceList extends TypedEmitter<DeviceListEvents> implements IDevic
         const enumerateResult = await transport.enumerate({ signal });
 
         if (!enumerateResult.success) {
-            throw new Error(enumerateResult.error);
+            throw new Error(enumerateResult.message || enumerateResult.error);
```
mroz22, line 280: I see, if it is `unexpected error` then it should also have a message.

mroz22: and then this error is propagated here right https://github.com/trezor/trezor-suite/blob/fix/connect-transport-error/packages/connect/src/core/index.ts#L1268 ? and sent o user, and we belive that it is better to show details instead of unexpected-error? right?

szymonlesisz: exactly, same as `getErrorStr` from  https://github.com/trezor/trezor-suite/pull/20049.
 i think there is few more places like this

martykan (approved):

PR #20211, opened 2025-07-15
feat(transport-bluetooth): implement `BluetoothIpc` api

szymonlesisz:
## Description

i guess it is tested enough in `feat/bt-thp`

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/suite-bt-ipc&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/suite-bt-ipc&lookback=7)

### DIFF
--- a/packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
+++ b/packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
@@ -1,40 +1,184 @@
-import { TypedEmitter } from '@trezor/utils';
+import { TypedEmitter, resolveAfter } from '@trezor/utils';
 
-import type { BluetoothIpcApi, BluetoothIpcEvents, BluetoothIpcState } from './types';
-
-const noop = (..._args: any[]) =>
-    Promise.resolve({ success: false, error: 'BluetoothIpc.NOOP' } as const);
+import { TrezorBluetooth } from './trezor-bluetooth';
+import type {
+    BluetoothAdapterState,
+    BluetoothDevice,
+    BluetoothIpcApi,
+    BluetoothIpcEvents,
+    BluetoothIpcState,
+    IpcResponse,
+    TrezorBluetoothSettings,
+} from './types';
 
 /*
  * used in @trezor/suite-desktop-core nodejs (main) context
  * unavailable in browser (renderer) context
  */
 export class BluetoothIpc extends TypedEmitter<BluetoothIpcEvents> implements BluetoothIpcApi {
-    init(state?: BluetoothIpcState) {
-        return Promise.resolve({ success: true, payload: state } as const);
+    private api: TrezorBluetooth;
+    private state: BluetoothIpcState = { knownDevices: [] };
+    private isScanning = false;
+
+    constructor(settings: TrezorBluetoothSettings) {
+        super();
+        this.api = new TrezorBluetooth(settings);
+    }
+
+    // 1. suite knows the device but system may not. device could be removed manually from the system UI
+    // 2. system knows the device (boolean on linux and windows, unknown on macos)
+    // 3. device is in pairing mode (data[0] > 0)
+    private filterConnectableDevices(devices: BluetoothDevice[]) {
+        return devices.filter(
+            d =>
+                this.state.knownDevices.find(dev => d.macAddress === dev.macAddress) ||
+                d.paired ||
+                d.data[0],
+        );
+    }
+
+    private result(error?: string): IpcResponse {
+        return error ? { success: false, error } : { success: true };
+    }
+
+    private async connectApi() {
+        if (this.api.isConnected()) return;
+
+        try {
+            await this.api.connect();
+        } catch (error) {
+            return this.result(error.message);
+        }
+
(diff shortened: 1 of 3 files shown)


packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
```diff
@@ -1,40 +1,186 @@
...
+            }
+        };
+
+        this.api.on('device_discovered', emitListUpdate);
+        this.api.on('device_connected', emitListUpdate);
+        this.api.on('device_disconnected', emitListUpdate);
+        this.api.on('adapter_state_changed', emitAdapterState);
+        this.api.on('device_updated', ({ id, devices }) => {
+            emitListUpdate({ devices });
+            const device = this.filterConnectableDevices(devices).find(d => (d.id = id));
```
mroz22: this looks like a bug here? 
```
d.id = id
```
shouldn't it be
```
d.id === id 
```
?

packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
```diff
@@ -1,40 +1,186 @@
...
+            }
+        });
+
+        return this.result();
+    }
+
+    async init(state?: BluetoothIpcState) {
+        if (state) {
+            this.state = { knownDevices: state.knownDevices };
+        }
```
mroz22: ntipick
`this state` is typed as  `BluetoothIpcState` so mayebe it could be simplified to
```
if (state)
  this.state = state;
````

File: packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts [hunk 3]
```diff
@@ -1,40 +1,186 @@
...
+
+        if (this.state.knownDevices.length > 0) {
+            try {
+                await this.connectApi();
+                await this.api.send('set_state', {
+                    devices: this.state.knownDevices,
+                });
+                await this.api.send('get_info');
```
mroz22: I can see that get_info is never consumed at the moment - so I wonder whether it is needed to call it. Does it have some side effect on the backed side? Or maybe you intend to start consuming its result later?

packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
```diff
@@ -1,40 +1,186 @@
...
+
+                const scanResult = await this.api.send('start_scan');
+                if (scanResult.length === 0) {
+                    // wait. devices may not be returned immediately
+                    await resolveAfter(1000);
+                }
+                await this.api.send('enumerate');
```
mroz22: similar to previous question - not consuming this calls result. does this triggers some event?

packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
```diff
@@ -1,40 +1,186 @@
...
+            this.emit('device-list-update', this.filterConnectableDevices(devices));
+        };
+
+        const emitAdapterState = ({ state }: { state: BluetoothAdapterState }) => {
+            this.emit('adapter-event', state);
+            if (state === 'enabled' && this.isScanning) {
+                // auto restart scan if adapter becomes enabled
+                this.api.send('start_scan').catch(error => {
+                    console.warn('Auto start_scan error', error);
+                });
```
mroz22, line 63: not sure if I understand. If I am already scanning, I want to start scanning again?

szymonlesisz: suite isScanning but the server is not.

i wasnt (and im still not) sure which party should restart scanning when you enable/disable bluetooth using system UI.
see:
https://github.com/trezor/trezor-suite/blob/develop/packages/transport-bluetooth/src/server/methods/start_scan.rs#L85

so i decided to do it on host which should be easier to maintain/fix as it will not require bt binary build

packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
szymonlesisz: shame on me :) of course you are right, fixed

packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
szymonlesisz: done

packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
szymonlesisz: in the future i want to check if bluetooth is supported (BLE version 5.1+) - but currently it doesnt do nothing, removed for now, will be added later, together with corresponding suite action

packages/transport-bluetooth/src/client/bluetooth-ipc-main.ts
szymonlesisz: some leftover, removed

mroz22 (approved):

PR #20245, opened 2025-07-17
new bridge port 21328

mroz22:
when the first device using the protocol THP is delivered, users with the old standalone bridge are going to have a somewhat hard time - the reason is that this bridge version is a standalone installation and we have no means and no plans as well to update it. 

On the other hand, there is the "node bridge", which is a re-implementation of trezord-go rest api provided by trezor-suite and this version is capable of supporting THP protocol. 

We need to onboard maximum amount of users to start using the new version.  And there could be gotchas - if user has the expected port 21325 occupied by the old bridge we won't abe able to start the new one. 

So the proposed solution to this is as follows:
- make node-bridge available on a new port
- all our bridge clients (connect npm packages, suite-desktop) will start to prioritize the newly chosen port and fallback to the legacy one.
- our bridge clients stop searching for the legacy port. 
- we start prompting users to uninstall the legacy bridge #18146 

This will effectively kill the old bridge, even when it is a standalone installation. There might be a couple of implementations that still communicate with the legacy bridge (maybe they have custom implementation, maybe they did not update npm package). If such situation occurs, we might run into problems but my assesment is that this won't be that big deal.

And also the build-in node bridge is from now on available on 2 ports (21328 and 21325) to deal with situation where user has already updated his trezor-suite but he is still using it in the background (as bridge) with some outdated 3rd party implementaition.

Also cc @Hannsek we should implement this feature (new bridge port and fallback in trezorlib)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=bridge-port-negotiation&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=bridge-port-negotiation&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=bridge-port-negotiation&lookback=7)

### DIFF
--- a/packages/connect/src/device/TransportList.ts
+++ b/packages/connect/src/device/TransportList.ts
@@ -21,6 +21,15 @@ const getOrCreateTransport = (
     transportType: ConnectSettingsTransport,
     params: Params,
 ) => {
+    if (transportType === 'BridgeTransport') {
+        // Temporary handling of BridgeTransport which translates to two instances listening on ports 21328/21325
+        const existing = transports.filter(t => t.name === transportType);
+
+        return existing.length
+            ? existing
+            : [new BridgeTransport({ ...params, port: 21328 }), new BridgeTransport(params)];
+    }
+
     if (typeof transportType === 'string') {
         const existing = tryGetTransport(transports, transportType);
         if (existing) return existing;
@@ -30,8 +39,6 @@ const getOrCreateTransport = (
                 return new WebUsbTransport(params);
             case 'NodeUsbTransport':
                 return new NodeUsbTransport(params);
-            case 'BridgeTransport':
-                return new BridgeTransport(params);
             case 'UdpTransport':
                 return new UdpTransport(params);
         }
@@ -69,7 +76,7 @@ const createTransports = (
     // BridgeTransport is the ultimate fallback
     const transportTypes = transports?.length ? transports : ['BridgeTransport' as const];
 
-    return transportTypes.map(type => getOrCreateTransport(existing, type, params));
+    return transportTypes.flatMap(type => getOrCreateTransport(existing, type, params));
 };
 
 export const createTransportList =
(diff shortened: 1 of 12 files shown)


marekrjpolak:
It seems to work, but I tried to improve it a bit anyway. Accept or reject any of the fixups at will.

- https://github.com/trezor/trezor-suite/pull/20245/commits/301412db6cef8b195f6f6fd6caee7cdb6fc7b33b - this should mimic the multiport `BridgeTransport` without actually modifying it, simply by creating two of them on different ports (much less lines which can be simply removed in the future)
- https://github.com/trezor/trezor-suite/pull/20245/commits/ef8ca13499291331e150dd79ba4261ef54686d6f - Electron's bridge module enforced old 21325 port so the new default 21328 had never been started
- https://github.com/trezor/trezor-suite/pull/20245/commits/877481497354b7b269137a99d7688e9469fad1cb - Compatibility server is started only when the primary one's port is different from the compatibility one
- https://github.com/trezor/trezor-suite/pull/20245/commits/27cf4a37b67145bee5115f98632374daee9aa525 - `port` in node bridge was used for both requested one and actually running one, which could cause some issues imo, so I separated them

mroz22:
very nice. these fixups make sense, thanks a lot 🙏

marekrjpolak (approved):
Apart from one suggestion regarding protobuf messages, LGTM.

Ondra-Zik:
QA OK

bridge runs now on 21328 and 21325 ports when possible 

port 21328 is preferred even if 21325 runs legacy bridge meaning you can connect THP device.

 - Suite version: desktop 25.11.0 (e7455fe9e6be509223b22020c9b16e4cd8ca26f6)
 - Browser: Trezor Suite 25.11.0
 - OS: Linux x86_64
 - Screen: 2560x1440
 - Device: Trezor T3W1 2.9.4 universal (revision a491f75ab1603d536048fd2cb44ff147ff406b9f)
 - Transport: BluetoothTransport, BridgeTransport 3.1.0

PR #20260, opened 2025-07-17
feat(connect): DeviceStatus busy

szymonlesisz:
## Description

adding new DeviceStatus `busy`

Trezor will return protocol `v1` Failure: Failure_Busy for example after reboot to bootloader before user clicks on "connect to host device"

https://github.com/trezor/trezor-firmware/commit/54202b9f53d107abf2aae4fc371548887a848836

we can show some info during firmware update if device.status === 'busy' 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/connect-device-busy&lookback=7)

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -177,6 +177,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
     // DeviceState list [this.instance]: DeviceState | undefined
     private state: DeviceState[] = [];
     private stateStorage?: IStateStorage = undefined;
+    private busy?: boolean;
 
     private _unavailableCapabilities: UnavailableCapabilities = {};
     public get unavailableCapabilities(): Readonly<UnavailableCapabilities> {
@@ -508,8 +509,14 @@ export class Device extends TypedEmitter<DeviceEvents> {
                 } else {
                     await this.getFeatures();
                 }
+
+                this.busy = false;
             } catch (error) {
                 _log.warn('Device._runInner error: ', error.message);
+
+                if (error.code === 'Failure_Busy') {
+                    this.busy = true;
+                }
 
                 if (error.code === 'Device_ThpPairingTagInvalid') {
                     // return as TypedError
@@ -1065,6 +1072,14 @@ export class Device extends TypedEmitter<DeviceEvents> {
         return 'normal';
     }
 
+    private getStatus(): DeviceStatus {
+        if (this.isUsedElsewhere()) return 'occupied';
+        if (this.wasUsedElsewhere) return 'used';
+        if (this.busy) return 'busy';
+
+        return 'available';
+    }
+
     // simplified object to pass via postMessage
     toMessageObject(): DeviceTyped {
         const { name, uniquePath: path } = this;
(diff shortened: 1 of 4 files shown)


packages/connect/src/device/Device.ts
```diff
@@ -508,9 +509,15 @@ export class Device extends TypedEmitter<DeviceEvents> {
                 } else {
                     await this.getFeatures();
                 }
+
+                this.busy = true;
```
mroz22: isn't this some leftover? turning it on basically always and never setting to false?

szymonlesisz: ah, my bad of course it should be false :)  i dont know what i was thinking it was late and i was tired :)

szymonlesisz: a08a9476e391b75225941b6ff0e98a3871922cdd

packages/connect/src/device/Device.ts
```diff
@@ -1065,6 +1072,14 @@ export class Device extends TypedEmitter<DeviceEvents> {
         return 'normal';
     }
 
+    private getStatus(): DeviceStatus {
+        if (this.busy) return 'busy';
+        if (this.isUsedElsewhere()) return 'occupied';
+        if (this.wasUsedElsewhere) return 'used';
```
mroz22: The last one to think about - the order of returns here. Don't 'used' and 'occupied' take priority? Consider

1. device A used by client 1 sets to busy
2. device A is acquired by client 2
3. device A still appears as busy to client 1 however the information that it is used elsewhere might be mort important in that moment?

Or maybe 'occupied' and 'used' are sort of legacy protocol v1 and busy only applies for thp so it is spot on?

szymonlesisz: good point, "busy" should be last :+1: 

> busy only applies for thp so it is spot on?

this is bluetooth related change. it will happen if you are connected via bluetooth but application layer of Trezor is not ready (see example in the description)

error message is protocol `v1` hardcoded somewhere in nordic code (see linked firmware PR), so basically it is not returned by the Trezor firmware or bootloader itself but from the transport layer before the request is even processed (there is no one to process it to)

marekrjpolak (commented):
Apart from @mroz22's note (with which I agree), are there any changes needed in Suite even before adding new behavior? E.g. is it ok that `suite-utils/src/device/getStatus` will still return `connected`?

(Or maybe not? Can acquired device also be `busy` or is it only a status of unacquired devices while acquired ones can only be `available`/`occupied`/`used`?)

szymonlesisz:
@marekrjpolak 
> E.g. is it ok that suite-utils/src/device/getStatus will still return connected?

it is OK for now.  this is prerequisite needed to create implementation in suite (view, utils etc)

marekrjpolak (approved):

vojtatranta:
If I put the device to bootloader mode and refresh the app, there's isn't `status=busy`. I guess that's a bug?

mroz22:
well, it all depends on that whether bootloader retunrs 'error.code === 'Failure_Busy'' 🤷

PR #20261, opened 2025-07-17
refactor(suite): replace send max switch with a button in send form

### LINKED ISSUE #17340: Solana send max button no error but error when no max with same amount

izmy:
## Description

The current implementation uses a switch to toggle "send max" mode. This affects internal logic (fee estimation using minimal dummy amount), which becomes problematic when toggled off again - especially for Solana where fees and rent-exemption rules are strict.

The switch was replaced with a button, which fills the max value and triggers fee estimation.

Also fixed a bug where priority fee simulation sometimes failed, and cleaned up some code to avoid unnecessary React re-renders.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/17340

## Screenshots:

image

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=refactor/send-max-button&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=refactor/send-max-button&lookback=7)


packages/blockchain-link/src/workers/solana/index.ts
```diff
@@ -528,12 +529,23 @@ const estimateFee = async (request: Request<MessageTypes.EstimateFee>) => {
         api.rpc,
     );
 
-    const priorityFee = await getPriorityFee(
-        api.rpc,
-        decompiledTransactionMessage,
-        message,
-        transaction.signatures,
-    );
+    let priorityFee;
```
matusbalascak: Just a nitpick, but wouldn't it be better to write it like this?

```ts
   const DEFAULT_PRIORITY_FEE = {
        computeUnitPrice: '0',
        computeUnitLimit: '0',
        fee: '0',
    };

    const priorityFee = await getPriorityFee(
        api.rpc,
        decompiledTransactionMessage,
        message,
        transaction.signatures,
    ).catch(err => {
        console.warn('getPriorityFee failed:', err);

        return DEFAULT_PRIORITY_FEE;
    });
```

This ensures that `priorityFee` cannot be reassigned after it's been fetched. Also it would be nice to have a proper return type for `getPriorityFee` that can be used with `DEFAULT_PRIORITY_FEE`

matusbalascak (approved):

packages/blockchain-link/src/workers/solana/index.ts
izmy: yes, good point, thanks

packages/suite/src/views/wallet/send/Outputs/Amount/Amount.tsx
```diff
@@ -132,11 +131,9 @@ export const Amount = ({ output, outputId }: AmountProps) => {
     };
 
     const sendMaxSwitch = (
-        <SendMaxSwitch
-            isSetMaxActive={isSetMaxActive}
-            data-testid={maxSwitchId}
-            onChange={onSwitchChange}
-        />
+        <Button data-testid={maxButtonId} size="tiny" variant="tertiary" onClick={onSwitchChange}>
```
vytick: nit: 
```suggestion
        <Button data-testid={maxButtonId} size="tiny" variant="tertiary" onClick={onMaxClicked}>
```

izmy: changed

PR #19951, opened 2025-07-03
Chore: bump electron 37

Lemonexe:
## Description

Bump electron 35 → 37.
Previously it was downgraded from 36 to 35 in https://github.com/trezor/trezor-suite/pull/19483 because of broken web workers (`DashboardGraph`)

Webworkers are not functional since 36.. [see comment](https://github.com/trezor/trezor-suite/pull/19951#discussion_r2182716987).

:arrow_right: but here I am experimenting with moving the `DashboardGraph` from webworker to plain async function, seems to me it works very well and we can thus simplify our tech → remove headaches in future :slightly_smiling_face: 
_(this is the only webworker in our Electron renderer process)_

- [Web deploy here](https://dev.suite.sldev.cz/suite-web/chore/bump-electron-37/web/)
- [Desktop apps here](https://github.com/trezor/trezor-suite/actions/runs/16051373438)

## Related Issue

Part of #19008

## Screenshots:

No worker, wallet with rich history, you can see that it takes a while (~1000ms) but graph is correctly rendered in loading state and it does not block rendering.

[no worker.webm](https://github.com/user-attachments/assets/ce9792d1-bcd9-4d0b-b668-b07db1b6adcc)

PR #19681, opened 2025-06-20
fix(ipc-proxy): IPC response handling when rendered was destroyed

Lemonexe:
https://github.com/electron/electron/blob/3536d49/docs/api/structures/ipc-main-event.md
https://github.com/electron/electron/blob/3536d49/docs/breaking-changes.md#behavior-changed-frame-properties-may-retrieve-detached-webframemain-instances-or-none-at-all

## Related Issue

Resolve #19670

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/ipc-when-rendered-destroyed&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/ipc-when-rendered-destroyed&lookback=7)

PR #20318, opened 2025-07-22
fix(suite-desktop-core): preventive handling for destroyed renderer

Lemonexe:
## Description

Attempt to prevent https://satoshilabs.sentry.io/issues/6739270430/?project=5193825
```
electron: Frame property was accessed after it navigated or was destroyed.
```

but we don't really know what causes it :confused: 

## Notes

Previous attempt #19681

At first I identified a lot [async handlers like this one](https://github.com/trezor/trezor-suite/commit/e4cc2d172d973c3c6c930cab8212a17184c89e46#diff-20325fa663c3e5e2a5d24f2a3f149e62b4c1a7b4ead1d6fcb4d933abfac78fc5) but there the `isDestroyed()` fix is useless, because `mainWindowProxy.getInstance()` [will return undefined if destroyed](https://github.com/trezor/trezor-suite/blob/308b1a470d27b4863ca479e20a2beee33d26c7be/packages/suite-desktop-core/src/libs/main-window-proxy.ts#L27-L29) so that is already handled :heavy_check_mark: 

Possibly related issues for Electron 35: could be spontaneously resolved by bumping Electron in #19951
https://github.com/electron/electron/issues/45653
https://github.com/electron/electron/issues/45610
https://github.com/electron/electron/issues/45215
https://github.com/electron/electron/issues/45142

Read more [here](https://www.electronjs.org/docs/latest/api/web-contents)

---

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/sanitize-for-electron-frame-destroyed&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/sanitize-for-electron-frame-destroyed&lookback=7)

### DIFF
--- a/packages/ipc-proxy/src/__tests__/mockedElectron.ts
+++ b/packages/ipc-proxy/src/__tests__/mockedElectron.ts
@@ -5,6 +5,7 @@ import { ElectronIpcMainInvokeEvent } from '../proxy-handler';
 const ipcMainEvent: ElectronIpcMainInvokeEvent = {
     senderFrame: {
         url: 'http://localhost:8000/',
+        isDestroyed: () => false,
     },
 };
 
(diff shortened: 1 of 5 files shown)


peter-sanderson (approved):
:crossed_fingers:

PR #20281, opened 2025-07-21
refactor(suite): cleanup in Onboarding component structure

szymonlesisz:
## Description
grooming in suite `Onboarding` component, it was a little messy and kinda hard to understand.

this is not just some random PR :) i need to add new view: Device.status === 'busy' https://github.com/trezor/trezor-suite/pull/20260

- `IsSameDevice` renamed to `DeviceDifferent` as this is what it actually do.
- `UnexpectedState` simplified:
  removed unnecessary react memoization (isNotSameDevice, UnexpectedStateComponent)
  also moved ShowPinMatrix logic to separate component

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=refactor/onboarding-cleanup&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=refactor/onboarding-cleanup&lookback=7)

PR #20344, opened 2025-07-23
refactor(suite-common): useFirmwareInstallation uiEvents

szymonlesisz:
## Description

related to https://github.com/trezor/trezor-suite/pull/20281

`useFirmwareInstallation` provides `uiEvent` from the state. every time this event is validated it actually requires 3 conditions: uiEvent presence + uiEvent.type + uiEvent.payload.

by moving it to memoized const im reducing validation process to only one condition: event presence and i believe this cleans the code a little bit (additionally this fixes one ts-error in suite-native)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/useFirmwareInstallation&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/useFirmwareInstallation&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/useFirmwareInstallation&lookback=7)

### DIFF
--- a/packages/suite/src/components/firmware/ReconnectDevicePrompt.tsx
+++ b/packages/suite/src/components/firmware/ReconnectDevicePrompt.tsx
@@ -4,7 +4,7 @@ import { useFirmwareInstallation } from '@suite-common/firmware';
 import { TranslationKey } from '@suite-common/intl-types';
 import { selectSelectedDeviceLabelOrName } from '@suite-common/wallet-core';
 import { BulletList, Column, DeviceAnimation, H2, Modal, Paragraph } from '@trezor/components';
-import { DEVICE, Device, UI } from '@trezor/connect';
+import { Device } from '@trezor/connect';
 import { DeviceModelInternal, getFirmwareVersion } from '@trezor/device-utils';
 import { ConfirmOnDevice } from '@trezor/product-components';
 import { spacings } from '@trezor/theme';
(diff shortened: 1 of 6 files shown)


suite-common/firmware/src/hooks/useFirmwareInstallation.ts
```diff
@@ -65,27 +65,38 @@ export const useFirmwareInstallation = (
...
+            }
+            if (firmware.uiEvent.type === DEVICE.BUTTON) {
+                return [undefined, firmware.uiEvent.payload];
+            }
+            if (firmware.uiEvent.type === UI.FIRMWARE_PROGRESS) {
+                return [undefined, undefined, firmware.uiEvent.payload];
+            }
+        }
+
+        return [];
```
Lemonexe, line 81: nit: could be simplified by `firmware.uiEvent?.type === ...`
and removing the `if (firmware.uiEvent) {`

packages/suite/src/components/firmware/ReconnectDevicePrompt.tsx
```diff
@@ -123,13 +115,8 @@ export const ReconnectDevicePrompt = ({ onClose, onSuccess }: ReconnectDevicePro
...
         // internal_model cannot be read from features while in bootloader mode.
-        const deviceModelFromEvent = eventDevice?.features?.internal_model;
-
-        if (deviceModelFromEvent === undefined) {
-            // Fallback. This should never happen.
-            return 'TR_SWITCH_TO_BOOTLOADER_HOLD_LEFT_BUTTON';
-        }
-
+        const deviceModelFromEvent =
+            eventDevice?.features?.internal_model || DeviceModelInternal.UNKNOWN;
```
Lemonexe, line 119: nice  :+1:

Lemonexe (approved):
Very nice cleanup, thank you!

suite-common/firmware/src/hooks/useFirmwareInstallation.ts
szymonlesisz: well it could but `firmware.uiEvent?.type` is actually transpiled to runtime `if (firmware.uiEvent && firmware.uiEvent.type === "")` so it means 3 checks `if (firmware.uiEvent)` instead of 1

Lemonexe: yes, I know, it shortens the source code and I don't think it's a performance problem that in runtime it checks the same thing three times.
But feel free to ignore & resolve this :)

szymonlesisz: depends from what scale we are looking at. three times here times 5 components where it is used = 15.
almost each our component is using chaining operators which leads to hundred thousands conditions which im sure have an impact on the overall performance.

im aware that this particular case will not change anything but since im doing a refactor why not do it properly :)

PR #20357, opened 2025-07-23
Feat/bt server set state

szymonlesisz:
## Description
based on https://github.com/trezor/trezor-suite/pull/20356

add server `set_state` method, pass known devices from suite so it can be used to determine if discovered device is paired (macos doesnt share that information)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/bt-server-set-state&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/bt-server-set-state&lookback=7)

### DIFF
--- a/packages/transport-bluetooth/src/server/adapter_manager.rs
+++ b/packages/transport-bluetooth/src/server/adapter_manager.rs
@@ -17,7 +17,7 @@ use tokio::{
 
 use crate::server::{
     device::{DeviceConnectionStatus, TrezorDevice},
-    types::{AdapterState, ChannelMessage, NotificationEvent},
+    types::{AdapterState, ChannelMessage, KnownDevice, NotificationEvent},
     utils, ConnectionBroadcast,
 };
 
@@ -34,6 +34,7 @@ struct ManagerState {
     is_scanning: bool,
     listeners: Vec<ConnectionBroadcast>,
     peripherals: DashMap<String, TrezorDevice>,
+    known_peripherals: Vec<KnownDevice>,
 }
 
 impl Debug for AdapterManager {
@@ -67,6 +68,7 @@ impl AdapterManager {
             is_scanning: false,
             listeners: Vec::new(),
             peripherals: DashMap::new(),
+            known_peripherals: Vec::new(),
         }));
 
         Ok(Self {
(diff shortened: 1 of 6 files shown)


szymonlesisz:
@coderabbitai review

packages/transport-bluetooth/src/server/adapter_manager.rs
```diff
@@ -251,9 +253,22 @@ impl AdapterManager {
         state.is_scanning = value;
     }
 
+    pub async fn set_known_peripherals(&self, value: Vec<KnownDevice>) {
+        let mut state = self.manager_state.lock().await;
+        state.known_peripherals = value;
+    }
+
+    async fn is_known_peripheral(&self, id: String) -> bool {
```
obrusvit: `id: &str`

packages/transport-bluetooth/src/server/adapter_manager.rs
```diff
@@ -251,9 +253,22 @@ impl AdapterManager {
...
+
+        state.known_peripherals.iter().any(|x| x.id == id)
+    }
+
     async fn add_device(&self, id: &PeripheralId) -> Result<TrezorDevice, AdapterError> {
-        let peripheral = self.get_peripheral_or_die(&id.to_string()).await?;
-        let device = match TrezorDevice::new(peripheral).await {
+        let id = id.to_string();
+        let peripheral = self.get_peripheral_or_die(&id).await?;
+        let is_known = self.is_known_peripheral(id).await;
```
obrusvit: pass id as ref: `&id`

packages/transport-bluetooth/src/server/methods/set_state.rs
```diff
@@ -0,0 +1,13 @@
+use crate::server::{
+    adapter_manager::AdapterManager,
+    types::{MethodResult, SetStateParams, WsResponsePayload},
+};
+use log::info;
+
+pub async fn set_state(manager: AdapterManager, params: SetStateParams) -> MethodResult {
```
obrusvit, line 7: Shouldn't it be called something like `set_known_peripherals`? 

I find it quite confusing that `AdapterManager` has `adapter_state` and `manager_state` and this method is called `set_state` but it sets only one field in the `manager_state` 😕

obrusvit (commented):
Minor points.

packages/transport-bluetooth/src/server/methods/set_state.rs
szymonlesisz: yes, currently it sets only known peripherals but i was thinking more forward that it could be extended in the future with more options like for example `unpaired_peripherals` (those should be ignored by the discovery) or default timeouts etc. so i didnt want to narrow down the scope to only one name (set_known_peripherals)

server doesnt have own storage we need to pass some initial data at start to work with them later, so this is more like an "init" method however `init` is already misused across the whole codebase so i didnt want to call it like that.

any other suggestions except for "init" or "set_known_peripherals" ?

packages/transport-bluetooth/src/server/adapter_manager.rs
szymonlesisz: [b3bc6f3](https://github.com/trezor/trezor-suite/pull/20357/commits/b3bc6f378d8661e7b3fe992b570583710cbc2a07)

packages/transport-bluetooth/src/server/adapter_manager.rs
szymonlesisz: [b3bc6f3](https://github.com/trezor/trezor-suite/pull/20357/commits/b3bc6f378d8661e7b3fe992b570583710cbc2a07)

packages/transport-bluetooth/src/server/methods/set_state.rs
obrusvit: No other suggestions. I see that you have reasons to name it "set_state" and the websocket API is the most important in this project so I'm good.

obrusvit (approved):

PR #20368, opened 2025-07-24
fix: value for sats and BTC as base currency above Movile Graph

peter-sanderson:
image
image

image

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix-mobile&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix-mobile&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix-mobile&lookback=7)

### DIFF
--- a/suite-native/formatters/src/utils.ts
+++ b/suite-native/formatters/src/utils.ts
@@ -1,13 +1,25 @@
 import { BigNumber } from '@trezor/utils/src/bigNumber';
 
-/** Matches three groups: 1. currency symbol,  2. whole number part and 3. decimal part. */
+/** Matches three groups: 1. currency symbol, 2. whole number part and 3. decimal part. */
 const BALANCE_PARSING_REGEX = /^(\D+)([\d,]+)(?:\.(\d+))?$/u;
 
 export const convertTokenValueToDecimal = (value: string | number, decimals: number) =>
     BigNumber(value).div(BigNumber(10).exponentiatedBy(decimals));
 
+const normalizeValueForBtcAndSats = (value: string) => {
+    const [first, second] = value.split(' ');
+
+    if (!second) {
+        return first;
+    }
+
+    return second === 'sat' ? `${second}${first}` : `${first}${second}`;
+};
+
 export const parseBalanceAmount = (value: string) => {
-    const regexGroups = value.match(BALANCE_PARSING_REGEX);
+    const normalizedValue = normalizeValueForBtcAndSats(value);
+
+    const regexGroups = normalizedValue.match(BALANCE_PARSING_REGEX);
     const [_, currencySymbol, wholeNumberPart, decimalNumberPart] = regexGroups ?? [
         null,
         null,
--- a/suite-common/graph/src/types.ts
+++ b/suite-common/graph/src/types.ts
@@ -1,13 +1,14 @@
 import { type NetworkSymbol } from '@suite-common/wallet-config';
 import { AccountKey, TokenAddress } from '@suite-common/wallet-types';
+import { BaseCurrencyAmount } from '@suite-common/wallet-utils';
 import { BigNumber } from '@trezor/utils';
 
 export type FiatGraphPoint = {
     date: Date;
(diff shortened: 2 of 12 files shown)


suite-native/formatters/src/utils.ts
```diff
@@ -1,13 +1,25 @@
...
+    const [first, second] = value.split(' ');
+
+    if (!second) {
+        return first;
+    }
+
+    return second === 'sat' ? `${second}${first}` : `${first}${second}`;
+};
+
 export const parseBalanceAmount = (value: string) => {
```
peter-sanderson, line 19: This whole regex parsing of a :exclamation: formatted value is IMO really bad idea. But I do not have ambitions to refactor it. So, I slapped a patch to it. 🩹

suite-native/formatters/src/utils.ts
```diff
@@ -1,13 +1,25 @@
...
     BigNumber(value).div(BigNumber(10).exponentiatedBy(decimals));
 
+const normalizeValueForBtcAndSats = (value: string) => {
+    const [first, second] = value.split(' ');
+
+    if (!second) {
+        return first;
+    }
+
+    return second === 'sat' ? `${second}${first}` : `${first}${second}`;
```
matejkriz, line 16: It's weird to see the mix of `sat123` and `123 sat` on the same screen 🤔 Do we really  want to show BTC/sat before the number as fiat currency?

peter-sanderson: We use one formatter (`BaseCurrencyAmountFormatter`), it is uses `'react-intl'` and therefore we do not have real control over the formatting. 

It does not support BTC and Sats obiously, so it is hacked together. @hynek-jina  also wanted to use standard where `sats` are always after the number. 

So when we are parsing this formatted value (bad bad bad idea in the first place!) we have to unhack-it and hack it back together.

Lemonexe: but why only `=== 'sat'` and not BTC too?

hynek-jina: I agree that it's arbitrary and feel free to dispute.. I just never saw `sat 1 200`, but always `1 200 sat`

Lemonexe: Oh no sorry my bad, the raw `value` is different for sat and btc

matejkriz: I see 🙈

matejkriz: @hynek-jina `1 200 sat` would be nice, but it needs some more hacking of hacks or "rozebrat traktor"

Lemonexe (approved):
Codewise LGTM, works _as intended_

suite-native/formatters/src/utils.ts
matejkriz: current state in account detail, see the 24h change 😉  

You can also see the mixing of `123 sat` with `sat123` there

Image

peter-sanderson: LMK once conclusion is made, I solve it in different PR

hynek-jina: I propose to stick with `123 sat` and remove the bitcoin price section.

evgenysl:
I cannot select Satoshis on Android on the latest devel to check the graph.
~Related?~ https://github.com/trezor/trezor-suite/issues/20570

STew790:
**QA NOK**
Sats lgtm but BTC shows 0.00 which can definitely show more decimal spaces.
image
image
(Also for the accounts)

**Info**
25.8.1 0468da9

Lemonexe:
@STew790 yep. This is known, accepted for shipping and product needed for further refinement.
CC @hynek-jina ...I think?

PR #20375, opened 2025-07-24
Take locale into account when choosing CSV delimiter

### LINKED ISSUE #12995: CSV export does not use a comma
Trezor Suite CSV export (Comma Separated Value) format file doesn't follow the standard, it uses a semicolon delimiter, which is rare in the real world.

Could it be fixed to use the regular comma separator format?

5K1PY:
## Related Issue

Fixes #12995

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/csv-delimiter&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/csv-delimiter&lookback=7)


packages/suite/src/utils/wallet/exportTransactionsUtils.ts
```diff
@@ -333,17 +334,17 @@ const prepareCsv = (
...
     content.forEach(item => {
         line = [];
 
         fieldKeys.forEach(field => {
-            line.push(sanitizeCsvValue(item[field]));
+            line.push(sanitizeCsvValue(item[field], locale));
         });
 
-        lines.push(line.join(CSV_SEPARATOR));
+        lines.push(line.join(getCSVSeparator(locale)));
```
peter-sanderson, line 347: `getCSVSeparator` is called multiple times => 

```ts
const separator  = getCSVSeparator(locale)
```
```suggestion
        lines.push(line.join(separator));
```

packages/suite/src/utils/wallet/exportTransactionsUtils.ts
```diff
@@ -291,8 +291,8 @@ const prepareContent = (
         .filter(record => record !== null) as Fields[];
 };
 
-const sanitizeCsvValue = (value: string) => {
-    if (value.indexOf(CSV_SEPARATOR) !== -1) {
+const sanitizeCsvValue = (value: string, locale: string) => {
```
peter-sanderson, line 294: We have a type to cover locales `Locale` (from `packages/suite/src/config/suite/languages.ts`

packages/utils/src/getCSVSeparator.ts
```diff
@@ -0,0 +1,30 @@
+export const LOCALES_USING_SEMICOLON = [
```
peter-sanderson, line 1: I would prefer either:

1. Create a typesafe map to map the langages from `Locale` type onto a `bool` explicitly defining  the usage of semicolon/comma. When we add a new language Typesscript will force the developer to add it here.

2. Using some tricky-trick like `(new Intl.NumberFormat(locale).format(1.1)).includes(',')` to determine if language uses comma as separator

peter-sanderson: Something like this: 
```suggestion
export const LOCALES_USING_SEMICOLON: Record<Language, ";"|","> = [
```

peter-sanderson (changes requested):
If I understand correctly, we still uses `;` for languages that may have issues with `,` (Czech for example). 

The correct solution IMO shall be to use `,` everywhere and properly escape the characters. 

But I see we are already doing it here: `sanitizeCsvValue`

So why we ever need to use `;`?

5K1PY:
> If I understand correctly, we still uses `;` for languages that may have issues with `,` (Czech for example).
> 
> The correct solution IMO shall be to use `,` everywhere and properly escape the characters.

Ok, then the work here is in the wrong direction. Closing this in favor of #20644.

PR #20379, opened 2025-07-24
fix(native): use useNavigateToInitialScreen instead of plain goBack

### LINKED ISSUE #19725: Cancel transaction after TX details double-check on device
**Describe the bug**
Once you decides to cancel the transaction after you double-check the details on your device you can't go back and you're stuck in a navigation loop.

**Info:**

- Mobile Suite Version: develop/latest
- OS: android

**How to reproduce**

1. Choose a coin
1. Go to send screen
1. Fill out send details
1. Choose fees and cofirm
1. Double check the details on device, but do not sign
1. Navigate back in the mobile app and you'll end up in a loop

**Expected behavior**
It is expected to go back / cancel your transaction and return to fees/send screen.

https://github.com/user-attachments/assets/681aa7c8-add1-4190-b4a4-5b6a1496ae80

OriginalEveres:
Added `returnToInitialScreen` flag to `ScreenHeader` and `GoBackIcon` in order to support return to the first screen in Navigation Stack. I had to add the flag, because in some screens we need to really call `goBack` in order to not skip bottom sheets.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19725

## Screenshots:

https://github.com/user-attachments/assets/1109001d-432f-472d-8bd4-94531e7583a9

https://github.com/user-attachments/assets/5a0a7ef4-f9e7-4de7-bfca-b6345313085c

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/go-back-navigation&lookback=7)

### DIFF
--- a/suite-native/module-accounts-management/src/components/AccountDetailScreenHeader.tsx
+++ b/suite-native/module-accounts-management/src/components/AccountDetailScreenHeader.tsx
@@ -11,6 +11,7 @@ import {
     RootStackRoutes,
     ScreenHeader,
     StackToStackCompositeNavigationProps,
+    useNavigateToInitialScreen,
 } from '@suite-native/navigation';
 
 type AccountDetailScreenHeaderProps = {
@@ -51,6 +52,7 @@ export const AccountDetailScreenHeader = ({
     accountKey,
 }: AccountDetailScreenHeaderProps) => {
     const navigation = useNavigation<AccountDetailNavigationProps>();
+    const navigateToInitialScreen = useNavigateToInitialScreen();
     const route = useRoute<RouteProp<RootStackParamList, RootStackRoutes.AccountDetail>>();
     const { closeActionType } = route.params;
 
@@ -78,6 +80,7 @@ export const AccountDetailScreenHeader = ({
                 />
             }
             closeActionType={closeActionType}
+            closeAction={navigateToInitialScreen}
         />
     );
 };
--- a/suite-native/module-send/src/screens/SendAccountsScreen.tsx
+++ b/suite-native/module-send/src/screens/SendAccountsScreen.tsx
@@ -7,11 +7,14 @@ import {
     SendStackParamList,
     SendStackRoutes,
     StackProps,
+    useNavigateToInitialScreen,
 } from '@suite-native/navigation';
 
 export const SendAccountsScreen = ({
     navigation,
 }: StackProps<SendStackParamList, SendStackRoutes.SendAccounts>) => {
+    const navigateToInitialScreen = useNavigateToInitialScreen();
+
     const navigateToSendFormScreen: OnSelectAccount = ({ account, tokenAddress, tokenSymbol }) => {
         analytics.report({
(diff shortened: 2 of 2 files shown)


suite-native/navigation/src/components/GoBackIcon.tsx
```diff
@@ -5,22 +5,31 @@ import { useNavigation } from '@react-navigation/native';
 import { IconButton } from '@suite-native/atoms';
 
 import { CloseActionType } from '../navigators';
+import { useNavigateToInitialScreen } from '../useNavigateToInitialScreen';
 
 type GoBackIconProps = {
     closeActionType?: CloseActionType;
+    returnToInitialScreen?: boolean;
```
matejkriz: What is the benefit of adding new prop instead of using existing closeAction? 
`closeAction={navigateToInitialScreen}`

OriginalEveres: in this case I believe its much cleaner to pass a prop instead of importing and calling the hook in the component that it uses. Currently there are 2 usages of this flag, but I believe that there will be more of them.

@STew790 mentioned only the send form flow, but maybe he could also try to find some other cases?

matejkriz: I don’t think this makes the API cleaner — quite the opposite, I find it a bit confusing. Introducing a new boolean prop like returnToInitialScreen on top of the existing closeAction makes the component’s behavior less clear. What happens if both are defined? From the outside, it’s not obvious that returnToInitialScreen would simply be ignored.

Both props are meant to handle the same thing (defining what happens when the user goes back) but now it’s unclear how they relate or which one takes precedence.

 I also don’t think that using it in multiple places justifies introducing a new boolean prop, since we’re only forwarding a single function.

OriginalEveres: I agree with the fact that it is confusing when you have `closeAction` and `returnToInitialScreen` passed. 

`GoBackIcon.tsx` contains the `navigation.goBack()` as a default go back action and to me `navigateToInitialScreen` feels like it could be the default as well for certain cases, and if we have component for going back, it should contain the default behavior IMO. But I might be wrong with the thinking about `navigateToInitialScreen` as default.

juriczech: I agree with @matejkriz here. I believe adding another prop here makes the component API unclear.

If you see component `suite-native/receive/src/components/ReceiveScreenHeader.tsx`, there is same functionality that you're trying to achieve here. That IMO is cleaner. 

If we really wanted to change the API to be able to use `navigateToInitialScreen` from inside this component without having to pass it as a prop, it would need to be used with `RequireOneOrNone` from type-fest so that you cannot pass `closeAction` and `returnToInitialScreen` at the same time. But even then I would say that adding another prop for that is unnecessary.

juriczech (approved):

Issue #19763, opened 2025-06-26
Firmware release new process


karliatto:
This issue is to track all the related tasks to the new firmware release process that involves connect, suite-web, suite-desktop and suite-mobile.

- [x] Duplicate the releases.json configs in data repo - https://github.com/trezor/data/pull/140
- [x] Add the button to debug mode to change the testing env - https://github.com/trezor/trezor-suite/pull/19827
- [x] Change connect and suite to use the new Firmware Release Config as source to offer new firmware update - https://github.com/trezor/trezor-suite/pull/18821
- [x] Make connect use new releases.json format file
- [x] Prepare testing environments on our side
- [x] We merge Suite and Connect PRs to develop after Suite Mobile freeze
- [x] Instruct QA on how to test it

karliatto:
This issue is completed. Many of the related pull requests are linked to it.

STew790:
**QA OK**
Tested with production devices and also preproduction and with the release of 2.9.1 (and 2.9.2) fws. Working as intended. BTC only fw also installing correctly for BTC only devices.

**Info**
25.10.1 https://github.com/trezor/trezor-suite/commit/2f7cd99c9ded24a0b9031f87ffa9245a40dc26c2

STew790:
**QA OK**
Tested multiple times throughout the release process of TS7.

PR #20449, opened 2025-07-29
Feat: add codesign public key

karliatto:
## Description
Extracting pieces for new firmware release process. 

- 4952ae72243492140b33bce34b3065b70fae09a8 - public key for production and signed environments
- 7794222be356290020d26b20fe73e5eb69912a11 - adding the possibility to set the singing key  to codesign so we can use non-codesign builds like development to use codesign release config.

## Related Issue

https://github.com/trezor/trezor-suite/issues/19763

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/add-codesign-public-key&lookback=7)

### DIFF
--- a/packages/env-utils/src/envUtils.native.ts
+++ b/packages/env-utils/src/envUtils.native.ts
@@ -3,8 +3,8 @@ import { Dimensions, Platform } from 'react-native';
 import Constants from 'expo-constants';
 import { getLocales } from 'expo-localization';
 
-import { publicKey } from './jws';
-import { EnvUtils } from './types';
+import { firmwareConfigPublicKey, publicKey } from './jws';
+import { EnvUtils, JWSPublicKeyUse } from './types';
 
 const isWeb = () => false;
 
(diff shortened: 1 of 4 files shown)


packages/env-utils/src/envUtils.native.ts
```diff
@@ -79,7 +79,13 @@ const getOsNameWeb = () => '';
 
 const getOsFamily = (): 'Linux' => 'Linux';
 
-export const getJWSPublicKey = () => (isCodesignBuild() ? publicKey.codesign : publicKey.dev);
+export const getJWSPublicKey = (use: JWSPublicKeyUse) => {
```
martykan: Wouldn't it be better to also have `useCodeSignKey` for consistency?

karliatto: Sure, makes sense, and it will be used later on. So here it is: https://github.com/trezor/trezor-suite/pull/20449/commits/f1dd965e8fe9adfad717fa5622229d93d435e6f8

martykan (approved):

PR #20456, opened 2025-07-29
feat(suite): Update passphrase loading copy

### LINKED ISSUE #19428: New passphrase: Loader should have different copy.
When the new passphrase is entered for the **first time** *(the initial entry, not the confirmation*)*, the loader copy shouldn’t mention checking for balances.

Technically, we **are** checking *(in order to redirect the user to the “Opening previously used” flow if we detect any history)*, but the user explicitly asked to create a **new** one. So, until we learn otherwise, let’s respect that intent and avoid explaining what we’re doing in the background.

The loader text should simply be: **“Preparing your wallet…”**

**We can use the same message after the confirmation step too, as long as it’s distinct from the **Opening** flow — where it is appropriate to mention that we’re looking for balances and transactions (since the user opted to open a previously used passphrase wallet).*

Design:

Image

Implementation:

Image

---

Ideally, we could implement the new design right away if feasible instead: [Figma](https://www.figma.com/design/7g6aHZZ0RsIZ9JG1ZWO3N9/Passphrase?m=auto&node-id=8747-16303&t=8Q7DkeBiTi6pSRrr-1)

jvaclavik:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19428

## Screenshots:

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=update-passphrase-loading-copy&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=update-passphrase-loading-copy&lookback=7)

### DIFF
--- a/packages/suite/src/support/messages.ts
+++ b/packages/suite/src/support/messages.ts
@@ -2092,7 +2092,7 @@ export default defineMessages({
         id: 'TR_CLOSE_WINDOW',
     },
     TR_COIN_DISCOVERY_LOADER_DESCRIPTION: {
-        defaultMessage: 'Checking passphrase wallet for balances & transactions',
+        defaultMessage: 'Preparing your wallet…',
         id: 'TR_COIN_DISCOVERY_LOADER_DESCRIPTION',
     },
     TR_COINS: {


packages/suite/src/support/messages.ts
```diff
@@ -2063,7 +2063,7 @@ export default defineMessages({
         id: 'TR_CLOSE_WINDOW',
     },
     TR_COIN_DISCOVERY_LOADER_DESCRIPTION: {
-        defaultMessage: 'Checking passphrase wallet for balances & transactions',
+        defaultMessage: 'Preparing your wallet…',
```
vojtatranta, line 2095: will this change be enough?

This has already been translated through crowdin. I am not sure if crowdin has some mechanism to tell "hey, this string changed, we need to re-translate it"?

I am approving this, but asking just in case.

vojtatranta (approved):
will this change be enough?

This has already been translated through crowdin. I am not sure if crowdin has some mechanism to tell "hey, this string changed, we need to re-translate it"?

I am approving this, but asking just in case. Coz I obviously can't see it in the live app.

PR #20457, opened 2025-07-29
Message system testing

### LINKED ISSUE #8493: Simplify testing of messaging system update
Currently it is not reliable and it takes time to properly test message updates.

Ideas:
- always fetch also local config (right now we fetch local only if remote config fetch failed) and compare what has higher `currentSequence` number (so developer don't need to block remote fetch in order to get the local one) ❌ 
- for same `currentSequence` number as current one, compare also `config.timestamp` (so developer don't need to bump sequence for testing, just the timestamp) ❌ 
- add easy possibility to remove current config from state – button in debug menu? Or just document deleting from storage?
- describe nicely in documentation how to test it locally and how to test it on builds from CI

izmy:
## Description

- Added a toggle to switch between remote and local message system config
- Introduced tooling for manual message injection and preview

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/8493

## Screenshots:

image

image

image

## Videos:

https://github.com/user-attachments/assets/15187765-bddf-4fe1-936c-99a9610e61ab

https://github.com/user-attachments/assets/33870737-7d7a-43df-ade6-e6fb93ef5cdf

https://github.com/user-attachments/assets/25e3dff1-e7f5-438e-b995-e7fb9d2292f3

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/message-system-testing&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/message-system-testing&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/message-system-testing&lookback=7)


suite-common/message-system/src/messageSystemUtils.ts
```diff
@@ -327,3 +330,121 @@ export const resolveMessageContent = (localizedMessages: Localization, language:
...
+        case 'transport':
+            return { bridge: '*', webusbplugin: '*' };
+
+        case 'settings':
+            return [{ tor: false }];
+
+        case 'devices':
+            return [
+                {
+                    model: 'T',
```
tomasklim:
```suggestion
                    model: 'T3T1',
```

suite-common/message-system/src/messageSystemUtils.ts
```diff
@@ -327,3 +330,121 @@ export const resolveMessageContent = (localizedMessages: Localization, language:
...
+                    firmwareRevision: '',
+                    firmware: '*',
+                    bootloader: '*',
+                    variant: '*',
+                    vendor: 'trezor.io',
+                },
+            ];
+
+        case 'countryCodes':
+            return ['US'];
```
tomasklim:
```suggestion
            return ['CZ'];
```

so that it works without change on our VPN

suite-common/message-system/src/messageSystemUtils.ts
tomasklim: please change to internal model in tests

suite-common/message-system/src/messageSystemThunks.ts
```diff
@@ -17,17 +18,16 @@ import {
...
+    selectMessageSystemConfigSource,
     selectMessageSystemCurrentSequence,
     selectMessageSystemTimestamp,
 } from './messageSystemSelectors';
 import { jws as configJwsLocal } from '../files/config.v1';
 
-// Enable this for local development purposes:
-// set to true to always fetch local JWS
-const FORCE_LOCAL_JWS = false;
+export const messageSystemPolling = createPollingController();
```
tomasklim: What is this good for?

tomasklim (commented):
Epic! Please update message system readme

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemInfo.tsx
```diff
@@ -0,0 +1,194 @@
...
+        },
+        {
+            field: 'category',
+            required: <Icon name="checkCircleFilled" variant="primary" />,
+            description: 'Where the message will appear',
+            allowed: (
+                <>
+                    <Badge variant="tertiary">banner</Badge>
+                    <Badge variant="tertiary">context</Badge>
+                    <Badge variant="tertiary">modal</Badge>
```
tomasklim, line 42: not yet implemented

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemInfo.tsx
```diff
@@ -0,0 +1,194 @@
...
+                </>
+            ),
+        },
+        {
+            field: 'category',
+            required: <Icon name="checkCircleFilled" variant="primary" />,
+            description: 'Where the message will appear',
+            allowed: (
+                <>
+                    <Badge variant="tertiary">banner</Badge>
```
tomasklim, line 40: can you use somehow validate the values by types from `messageSystem.ts`? so that it yells in case of changes

suite-common/suite-utils/src/pollingController.ts
```diff
@@ -0,0 +1,35 @@
+type PollingCallback = () => void;
+
+export const createPollingController = () => {
+    let intervalId: ReturnType<typeof setInterval> | null = null;
+    let callback: PollingCallback | null = null;
```
matusbalascak: It doesn't seem like this variable is necessary, it is just assigned in `start`, but never used.

suite-common/message-system/src/messageSystemTypes.ts
```diff
@@ -58,6 +61,14 @@ export const Feature = {
     dashboardPromoBanner: 'dashboard.promoBanner',
 } as const;
 
+export const FEATURE_LIST = Object.values(Feature)
```
matusbalascak: It would be more fitting to put `FEATURE_LIST` and `CONTEXT_PATTERNS` into `messageSystemConstants.ts` instead of `messageSystemTypes.ts`.

suite-common/message-system/src/messageSystemTypes.ts
```diff
@@ -58,6 +61,14 @@ export const Feature = {
...
+    .flatMap(value =>
```
matusbalascak: It'd be better to use `.flatMap<string>` instead of `as string[]` type cast after `sort`.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+export const MessageSystemForm = () => {
+    const popoverRef = useRef<PopoverRef>(null);
+    const [showForm, setShowForm] = useState(false);
+    const [formData, setFormData] = useState<string>(DEFAULT_ACTION);
+    const [parsedData, setParsedData] = useState<Action | null>(null);
+    const [jsonError, setJsonError] = useState<string | null>(null);
+    const dispatch = useDispatch();
+
+    useEffect(() => {
```
matusbalascak: Nitpick, but I think it would be nice to limit effects usage where not needed. You can instead create a function `handleFormData` where you would put this logic and also call `setFormData` there. You can then use this function instead of calling `setFormData` directly.

tomasklim:
@coderabbitai review

pavelmario:
@izmy , I have tested this improvement and it looks really cool. However, it does not seem to be working correctly while changing the languages as the message always stays in English.

The message:

`{
  "conditions": [
    {
      "browser": {
        "firefox": "*",
        "chrome": "*",
        "chromium": "*"
      }
    }
  ],
  "message": {
    "id": "5bbbeb12-4492-4842-bcaf-7385ed9d02cf",
    "priority": 100,
    "dismissible": true,
    "variant": "critical",
    "category": "banner",
    "content": {
      "en": "EN Critical",
      "cs": "CS Critical",
      "es": "ES Critical",
      "de": "DE Critical",
      "fr": "FR Critical",
      "pt": "PT Critical"
    }
  }
}`

Screenshots:

03_cs
02_es
01_en

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemInfo.tsx
izmy: done... I did a complex validation using yup

suite-common/suite-utils/src/pollingController.ts
izmy: true... removed

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemForm.tsx
izmy: I reworked the entire `useEffect` for validation, so now I think it's necessary to use it without moving the logic into a function

izmy:
> @izmy , I have tested this improvement and it looks really cool. However, it does not seem to be working correctly while changing the languages as the message always stays in English.
> 
> The message:
> 
> `{ "conditions": [ { "browser": { "firefox": "*", "chrome": "*", "chromium": "*" } } ], "message": { "id": "5bbbeb12-4492-4842-bcaf-7385ed9d02cf", "priority": 100, "dismissible": true, "variant": "critical", "category": "banner", "content": { "en": "EN Critical", "cs": "CS Critical", "es": "ES Critical", "de": "DE Critical", "fr": "FR Critical", "pt": "PT Critical" } } }`
> 

It was a different bug unrelated to my PR: https://github.com/trezor/trezor-suite/pull/20875

suite-common/message-system/src/messageSystemThunks.ts
izmy: It’s a tiny wrapper around setInterval that gives us safe control over polling
- Ensures only one timer runs (no duplicate intervals after re-init).
- Provides start/stop/restart; restart does an immediate fetch before scheduling.

suite-common/message-system/src/messageSystemValidation.ts
```diff
@@ -0,0 +1,229 @@
+import type {
```
tomasklim, line 1: nice!

tomasklim (commented):
Please add info about this great tool into `docs/features/message-system.md`

packages/suite/package.json
```diff
@@ -124,6 +124,7 @@
         "redux-thunk": "^3.1.0",
         "semver": "^7.7.1",
         "styled-components": "^6.1.19",
+        "uuid": "^11.1.0",
```
Lemonexe, line 127: :memo: [10kB](https://bundlephobia.com/package/uuid@11.1.0), that's ok (just noting)

suite-common/suite-utils/src/pollingController.ts
```diff
@@ -0,0 +1,39 @@
+type PollingCallback = () => Promise<void> | void;
+
+export const createPollingController = () => {
+    let intervalId: ReturnType<typeof setInterval> | null = null;
```
Lemonexe: nit: use [this type](https://github.com/trezor/trezor-suite/blob/3f34981e5d2536fda9afc8616bb5647702d74898/packages/type-utils/src/timeout.ts) for consistency

suite-common/suite-utils/src/pollingController.ts
```diff
@@ -0,0 +1,39 @@
...
+    let running = false;
+
+    const start = (cb: PollingCallback, intervalMs: number) => {
```
Lemonexe: don't use unnecessary abbreviations please :pray: 
```suggestion
    const start = (callback: PollingCallback, intervalMs: number) => {
```

suite-common/suite-utils/src/pollingController.ts
```diff
@@ -0,0 +1,39 @@
...
+        start(cb, intervalMs);
+    };
+
+    return {
+        start,
+        stop,
+        restart,
+        isRunning: () => intervalId !== null,
+    };
+};
```
Lemonexe: overall: why is this not a class, instantiated with `new PollingController`?

Ok, it works to use just the function scope to  hold the state `intervalId` and `running`, but it's my opinion that using class (or constructor) would make it clearer that this is a factory for a bunch of stateful function. The object returned from this factory does not hold the state itself, so the fact that it relies on fn scope is a bit obscured.

(for examle `setInterval` itself is also a plain fn, but it returns handle, does not hold the state itself)

suite-common/suite-utils/src/pollingController.ts
```diff
@@ -0,0 +1,39 @@
...
+    let running = false;
+
+    const start = (cb: PollingCallback, intervalMs: number) => {
+        if (intervalId) return;
+        intervalId = setInterval(async () => {
+            if (running) return;
+            running = true;
+            try {
+                await cb();
+            } finally {
```
Lemonexe, line 19: Currently it silences errors. For debugging purposes, how about this? 
```suggestion
            } catch(err) { console.error(err); } finally {
```

suite-common/suite-utils/src/pollingController.ts
```diff
@@ -0,0 +1,39 @@
...
+    const stop = () => {
+        if (intervalId) {
+            clearInterval(intervalId);
+            intervalId = null;
+        }
+    };
+
+    const restart = async (cb: PollingCallback, intervalMs: number) => {
+        stop();
+        await cb();
```
Lemonexe: nit: stop & restart do not care if it might be `running` right now, but no need to implement it, it would be overkill

packages/suite/src/views/settings/SettingsDebug/MessageSystemConfigSourceSelect.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const onChange = useCallback(
+        (value: MessageSystemConfigSource) => {
+            dispatch(messageSystemActions.setConfigSource(value));
+            dispatch(initMessageSystemThunk());
+        },
+        [dispatch],
+    );
+
+    return (
+        <SettingsSectionItem anchorId={SettingsAnchor.AddressDisplay}>
```
Lemonexe: this is not the right anchor. In debug settings I think we don't need anchors, please remove.

suite-common/message-system/src/messageSystemUtils.ts
```diff
@@ -327,3 +330,121 @@ export const resolveMessageContent = (localizedMessages: Localization, language:
...
+                    feature: [
+                        {
+                            domain: '',
+                            flag: true,
+                        },
+                    ],
+                };
+            case 'banner':
+            default:
+                return {};
```
Lemonexe: this is not type exhaustive, please `return exhaustive(type)`.

:point_right:  But even better yet, you can rewrite the function with switch case to a POJO, and make it exhaustive simply by typing as `Record<Category, Partial<Message>>` :slightly_smiling_face:

suite-common/message-system/src/messageSystemUtils.ts
```diff
@@ -327,3 +330,121 @@ export const resolveMessageContent = (localizedMessages: Localization, language:
...
+    return {
+        message: {
+            ...baseMessage,
+            ...extraFields,
+        },
+        conditions: [{}],
+    };
+};
+
+export const getDefaultConditionValue = (key: string) => {
```
Lemonexe: Declarative typing would be appreciated :pray: I believe this would be appropriate?
```suggestion
export const getDefaultConditionValue = (key: keyof Condition): Condition[keyof Condition] => {
```

suite-common/message-system/src/messageSystemUtils.ts
```diff
@@ -327,3 +330,121 @@ export const resolveMessageContent = (localizedMessages: Localization, language:
...
+                    variant: '*',
+                    vendor: '*',
+                },
+            ];
+
+        case 'countryCodes':
+            return ['CZ'];
+
+        default:
+            return {};
```
Lemonexe: Once the fn is typed as I suggested above, `exhaustive(key)` would again be better, at which point I realized that this whole function can also be rewritten as object typed
as `Record<keyof Condition, Condition[keyof Condition]>`

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+                        <Column flex="1" gap={spacings.md}>
+                            <MessageDetail message={message} />
+                            <Divider color="backgroundNeutralBold" />
+                            <MessageConditions conditions={conditions} />
+                        </Column>
+                        <Column gap={spacings.xs}>
+                            <MessageInfo
+                                message={message}
+                                allValidMessages={allValidMessages}
+                                isInApp={allInAppIds[message.id]}
```
Lemonexe: nit: teeechnicaly this isn't type safe due to TS limitations.
`allInAppIds` is `Record<string, boolean>` but accessing allInAppIds[string] is not guaranteed to yield boolean, it can be undefined.
It'd be cleaner to simply use `!!`.

There will be no problem though as it's only tested for truthiness, so it's a smol nit.

suite-common/message-system/src/messageSystemSelectors.ts
```diff
@@ -137,6 +140,11 @@ export const selectIsFeatureDisabled = (
     return featureFlag !== undefined ? !featureFlag : (defaultValue ?? false);
 };
 
+export const selectAllMessages = (state: MessageSystemRootState) =>
+    state.messageSystem.config?.actions.map(a => a.message) ?? [];
```
Lemonexe: this is unused, remove pls

suite-common/message-system/src/messageSystemReducer.ts
```diff
@@ -52,9 +56,18 @@ export const prepareMessageSystemReducer = createReducerWithExtraDeps(
...
                 state.currentSequence = config.sequence;
+
+                const prevInAppActions =
+                    state.config?.actions?.filter(a => !!state.inAppIds[a.message.id]) ?? [];
+                const incomingFileActions = config.actions ?? [];
+
+                state.config = {
+                    ...config,
+                    actions: [...prevInAppActions, ...incomingFileActions],
+                };
```
Lemonexe, line 81: shouldn't you also update `state.inAppIds` with `incomingFileActions`?

suite-common/message-system/src/messageSystemReducer.ts
```diff
@@ -19,6 +19,10 @@ const initialState: MessageSystemState = {
     dismissedMessages: {},
 
     validExperiments: [],
+
+    configSource: 'remote',
+
+    inAppIds: {},
```
Lemonexe: name is not clear to me until I delve into the code implementation.
I'd suggest something like `loadedMessageIds`.

But really, does this have to be its own state variable? It's not a unique data, it's directly derived from messages.
IIUC it's only "caching", which could be solved easily by memoizedSelector from messages.
If you decide to go this way, ignore all my other comments pertaining `inAppIds`

:thought_balloon:  Also, performance is not really important in debug settings, is it?

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+                ))}
+            </Column>
+        </Modal>
+    );
+};
+
+export const MessageSystemDebugInfo = () => {
+    const config = useSelector(selectMessageSystemConfig);
+    const allValidMessages = useSelector(selectAllValidMessages);
+    const allInAppIds = useSelector(selectAllInAppIds);
```
Lemonexe: it's useless to select both of these here, only to pass them via props to `MessageModal`.
Select them in `MessageModal` instead.

Lemonexe (commented):
CR WIP, just dumping some comments.

suite-common/message-system/src/messageSystemReducer.ts
Lemonexe: Note: practically this means that you cannot locally delete messages that are loaded from your selected source.
You can only Remove those you added locally yourself.
When I load app, `state.inAppIds = {}`

EDIT:
now I see it [was probably intentional.](https://github.com/trezor/trezor-suite/blob/8c6f96b1920dfac0a8dc0bfddd2f9c7e3706ea66/packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx#L300) then the naming ([my other comment](https://github.com/trezor/trezor-suite/pull/20457#discussion_r2304086731)) should be different, like `manuallyAddedMessageIds`, but I'm not 100% sure what is the intention here.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+    onCategoryChange,
+}: MessageFiltersProps) => (
+    <Row alignItems="center" justifyContent="space-between" gap={spacings.sm}>
+        <SelectBar
+            selectedOption={selectedCategory}
+            options={[...CATEGORY_FILTER_OPTIONS]}
+            size="small"
+            onChange={value => {
+                onCategoryChange(value);
+            }}
```
Lemonexe: nit: simplify
```suggestion
            onChange={onCategoryChange}
```

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+            bottomContent={<MessageSystemForm />}
+        >
+            <Column gap={spacings.sm}>
+                <MessageFilters
+                    showActive={showActive}
+                    onToggleActive={() => setIsActive(prev => !prev)}
+                    selectedCategory={selectedCategory}
+                    onCategoryChange={category => {
+                        setSelectedCategory(category);
+                    }}
```
Lemonexe: nit: simplify
```suggestion
                    onCategoryChange={setSelectedCategory}
```

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+                iconName="translate"
+            >
+                <div>
+                    {Object.entries(messages)
+                        .filter(([lang]) => lang === 'en')
+                        .map(([lang, text]) => (
+                            <div key={lang}>
+                                <strong>{lang}:</strong> {text}
+                            </div>
+                        ))}
```
Lemonexe: I think this can be simplified with `.find`, which will make it clear that we only expect one of them.
`filter` and `map` sound like there will be more of them, but that is impossible, JS does not allow duplicate keys in an object.
:thought_balloon:  JSON does allow it, but parsing the JSON only keeps the last key.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+    return conditions.map(({ environment, os, devices, settings, countryCodes }, index) => (
+        <Column key={index} gap={spacings.sm}>
+            <Row gap={spacings.sm} alignItems="flex-start">
+                <InfoItem label="Environment" iconName="devices" variant="default">
+                    {environment ? (
+                        <ul>
+                            {Object.entries(environment).map(([key, value]) => (
+                                <InfoItem
+                                    key={key}
+                                    label={key.charAt(0).toUpperCase() + key.slice(1)}
```
Lemonexe: nit: `import { capitalizeFirstLetter } from '@trezor/utils';`

Same for the two subsequent blocks.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+                                    direction="row"
+                                >
+                                    {String(value)}
+                                </InfoItem>
+                            ))}
+                        </ul>
+                    ) : (
+                        '-'
+                    )}
```
Lemonexe: This block is almost identical to the previous Environment section, and the subsequent Setting section.
Opportunity to extract component `MessageConditionsPerCategory`, which will only need `os` as prop.

The only difference I notice between these blocks is that Settings are not wrapped in `<ul>`, but I think they can be too for consistency.

You could include the line 238 too `<InfoItem label="Operating System" iconName="browsers" variant="default">` but it would need more props, so not such a useful abstraction.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+                <MessageModal
+                    actions={config.actions}
+                    allValidMessages={allValidMessages}
+                    allInAppIds={allInAppIds}
+                    onCloseModal={handleCloseValidMessages}
+                />
+            )}
+        </>
+    );
+};
```
Lemonexe, line 44: Please split the subcomponents into separate files for easier navigation :pray:

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+    if (!devices || devices.length === 0) {
+        return (
+            <InfoItem label="Devices" iconName="devices" variant="default">
+                -
+            </InfoItem>
+        );
+    }
+
+    const firstDevice = devices[0];
+    const restDevices = devices.slice(1);
```
Lemonexe: I don't see the need to display `firstDevice` outside of the collapsible.
It makes sense for languages where `en` is default, but the listed device condition are equally important; this begets the question "why is this device condition more important than others?" but it isn't, so it's mildly confusing.
It's just my opinion that it will be clearer to list them equally in Collapsible. Plus it simplifies the code.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+                <div>
+                    <strong>{firstDevice.model}</strong> (firmware: {firstDevice.firmware},
+                    bootloader: {firstDevice.bootloader})
+                </div>
+
+                <Collapsible.Content>
+                    {restDevices.map((d, i) => (
+                        <div key={i}>
+                            <strong>{d.model}</strong> (firmware: {d.firmware}, bootloader:{' '}
+                            {d.bootloader})
```
Lemonexe: the condition has other properties, which are not listed here. I don't if it's important, but there is a lot of space so why not list them?
image

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
```diff
@@ -0,0 +1,492 @@
...
+                          )
+                        : '-'}
+                </InfoItem>
+
+                <InfoItem label="Country codes" iconName="globe" variant="default">
+                    {countryCodes ? countryCodes.join(', ') : '-'}
+                </InfoItem>
+            </Row>
+            <MessageDevices devices={devices} />
+        </Column>
```
Lemonexe: Some conditions are missing here: duration, browser, transport, settings
They are not used so often, so maybe there is no need to have them here, but at least write a TODO

packages/suite/src/actions/suite/storageActions.ts
```diff
@@ -503,14 +503,17 @@ export const forgetDeviceMetadataError =
...
+        getState().messageSystem;
 
     db.addItem(
         'messageSystem',
         {
             config,
             currentSequence,
             dismissedMessages,
+            configSource,
+            inAppIds,
```
Lemonexe: what value is there in persisting `inAppIds` if the messages themselves are not persisted?

suite-common/message-system/schema/config.schema.v1.json
```diff
@@ -130,6 +131,9 @@
                                         },
                                         "flag": {
                                             "type": "boolean"
+                                        },
+                                        "visibleBanner": {
+                                            "type": "string"
```
Lemonexe, line 136: please don't define this in schema, it makes it look like a generic property available for all feature flags.
But feature flags can have [an arbitrary payload](https://github.com/trezor/trezor-suite/blob/8c6f96b1920dfac0a8dc0bfddd2f9c7e3706ea66/suite-common/suite-types/src/messageSystem.ts#L409), which is typed as unknown, and has to be processed as such in  the consuming code: [DashboardPromoBanner](https://github.com/trezor/trezor-suite/blob/d2e036284e9cbc1c8ae4994346a177665b673f74/packages/suite/src/views/dashboard/DashboardPromoBanner/DashboardPromoBanner.tsx#L21).
But these are ad-hoc properties relevant only to specific feature flag domain , and it's responsibility of the consuming code to handle it.

packages/react-utils/src/hooks/useTextareaCursorPosition.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        cleanupRef.current = () => {
+            document.removeEventListener('selectionchange', onSelectionChange);
+        };
+    }, []);
+
+    useEffect(() => () => cleanupRef.current?.(), []);
+
+    return { textareaRef, position };
+};
```
Lemonexe, line 64: very nice :+1:

suite-common/message-system/src/messageSystemConstants.ts
```diff
@@ -27,3 +41,62 @@ export const CONFIG_URL_REMOTE = {
     stable: `${CONFIG_URL_REMOTE_BASE}/stable/${JWS_CONFIG_FILENAME_REMOTE}`,
     develop: `${CONFIG_URL_REMOTE_BASE}/develop/${JWS_CONFIG_FILENAME_REMOTE}`,
 };
+
+export const FEATURE_LIST = Object.values(Feature)
+    .flatMap<string>(value =>
+        typeof value === 'string'
+            ? value
+            : Object.values(value).flatMap(v => (typeof v === 'string' ? v : Object.values(v))),
```
Lemonexe: this is hardcoded level 2 recursion.
The reason why we need two levels is incidental – as that's how 
But there isn't any rule anywhere saying it can't be deeper, so writing recursive function would be better.

My main reason is that the code should convey clear meaning at first glance – why does it do what it does.

Lemonexe (commented):
Dumping more comments.
So far I have reviewed everything except `MessageSystemForm + validation`.
Overall looks good, but I have a lot of codestyle, naming comments. Only a few doubts about functionality.

Anyway I'll finish the CR later, need to do other stuff RN.

suite-common/suite-utils/src/pollingController.ts
izmy: Good point. I chose a closure for simplicity, but making the state explicit is clearer. I’ve refactored it to a class (new PollingController()), keeping the same API and behavior.

suite-common/suite-utils/src/pollingController.ts
izmy: Noted. I’ll keep stop/restart simple and not handle in-flight ticks... we can add a wait-for-idle later if needed.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
izmy: Good point.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
izmy: Yea, you are right, I added other properties.

suite-common/message-system/schema/config.schema.v1.json
izmy: Well, I’d like to keep this in the schema, because:

- Treat the schema as a contract, not just docs. Explicit optional props improve DX, codegen, and CI validation.
- It prevents drift/typos (unknown payload bags tend to accumulate ad-hoc keys).
- `visibleBanner` is optional, so it doesn’t force other flags to use it.
- If it were up to me, I would set `additionalProperties: false` 😀
- and I also discussed it with @jvaclavik - we agreed that a stricter scheme is better.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
izmy: heh, "en" is actually required, so I'll access it directly

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
izmy: Good catch!

suite-common/message-system/src/messageSystemReducer.ts
izmy: I changed it to `manuallyAddedMessageIds`.

suite-common/message-system/src/messageSystemReducer.ts
izmy: I keep `inAppIds`/`manuallyAddedMessageIds` because the message system polls new data and I need to persist it regularly, and I also want test messages to survive an app refresh.

packages/suite/src/actions/suite/storageActions.ts
izmy: Messages are inside the config.

tomasklim (approved):

jvaclavik:
This is awesome 😍 good job!

suite-common/message-system/src/messageSystemReducer.ts
Lemonexe: Right that makes sense! Thank you for renaming, this is much clearer :ok_hand:

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemDebugInfo.tsx
Lemonexe: Even better :slightly_smiling_face:

suite-common/message-system/schema/config.schema.v1.json
Lemonexe: Ok, my point was that it creates false expectation that `visibleBanner` is a property available for all feature flags, while the handling for it is specific only for `dashboardPromoBanner`.
In TS this would be easily solvable with a union of types per each feature flag, but JSON schema cannot do that. 
But I don't have a strong opinion about this.

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemForm/MessageSystemForm.tsx
```diff
@@ -0,0 +1,177 @@
...
+    const formatJSON = useCallback(() => {
+        setFormData(JSON.stringify(parsedData, null, 2));
+    }, [parsedData]);
+
+    const availableConditionOptions = useMemo(() => {
+        if (!Array.isArray(parsedData?.conditions) || !parsedData.conditions[0]) {
+            return CONDITION_OPTIONS;
+        }
+
+        const bucket = parsedData.conditions[0] as Record<keyof Condition, unknown>;
```
Lemonexe, line 96: why is this called `bucket`? :bucket: what does that mean?

Lemonexe (commented):
Finished CR ex-post, but have no more comments; the form was not as complex as I thought. It's great that you implemented all these features with only yup & JSON :slightly_smiling_face: 
Looking perfect :rocket:

packages/suite/src/views/settings/SettingsDebug/MessageSystem/MessageSystemForm/MessageSystemForm.tsx
izmy: That's probably AI noise 😄 I will rename it to `conditions` (because conditions in the message system is the array with an object and we (almost) always use only the first item from the array)

PR #20497, opened 2025-07-31
fix(suite): make sure that the positional params of the modal routes …

### LINKED ISSUE #20480: Route default params are nto populated at all
There is a bug introduced here `26f46e39451b8331ed57f7c14136e3ac43fbbacc`. 

Which causes that default values fro uri params like `cancelable` aren't applied at all.

So if you want to open a cancellable device switcher, you always need to pass `params: { cancelable: true }`. Without it, the params aren't defined at all and hence cancelable is always false, due to defaults aren't being applied.

There is certainly somewhere in the app this case

vojtatranta:
…are populated with defaults

## Description

Funny thing. If there were no params at all, the default `cancelable: true` hasn't been applied. There had to be something behind `#` in the URL.

Previous, before `26f46e39451b8331ed57f7c14136e3ac43fbbacc` that `SwitchDevice` modal didn't even cared about `cancelable=true/false` in its props, so it just worked as cancelable always.

So by "fixing" the switch device modal to care for `cancelable` prop, we broke other cases that relied on cancelable swich device modal and didn't pass the parameter explicitly in `goto()`

So from now on, you are forced by the TS to provide the default param values even if these are undefined.

## Related Issue

Resolve  https://github.com/trezor/trezor-suite/issues/20480

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=20480-no-default-params-routes&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=20480-no-default-params-routes&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=20480-no-default-params-routes&lookback=7)


vojtatranta:
wrf is dis

jvaclavik (approved):

vojtatranta:
ddammm!!!! :D i merged it! oh no! sorry guys

PR #20550, opened 2025-08-04
fix(suite): close the trezors ui window right when the passphrase is …

### LINKED ISSUE #20045: Passphrase confirmation modal during discovery shows for too long
When discovery is triggered on an existing wallet (eg. by changing coins) and it's necessary to enter passphrase again, the confirm action modal remains visible even after confirming, until the discovery finishes. 
It should close after confirming on device. 

https://github.com/user-attachments/assets/ab23998c-184b-493e-b0af-c9b2eed257ef

vojtatranta:
…confirmed when discovery requests it dynamically

## Description

This is IMHO a better solution. 

Here, I recognize that the additional discovery requested a passphrase and when this check passes, I manually hide the window. The weakness is that theoretically, there can be another window request and this can hide it. But the likelihood is small, and you see the dispatches in the console.

Second possible solution is to hide modal "as it renders" based on some state of the discovery. Guess that would require another flag in the discovery to make it reliable like "discovery requested passphrase dynamically" and there would be a condition that if discovery progresses, hide this modal without waiting for hte "window close" event.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20045#issuecomment-3149631542

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=20045-passphrase-confirmation-stuck&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=20045-passphrase-confirmation-stuck&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=20045-passphrase-confirmation-stuck&lookback=7)

### DIFF
--- a/suite-common/wallet-core/src/discovery/discoveryThunks.ts
+++ b/suite-common/wallet-core/src/discovery/discoveryThunks.ts
@@ -703,9 +703,32 @@ export const runAdditionalDiscoveryThunk = createThunk(
 
         TrezorConnect.on<DiscoverAccountsProgress>(UI.BUNDLE_PROGRESS, onBundleProgress);
 
-        const result = await TrezorConnect.discoverAccounts({
-            device,
+        // NOTE: prepare the device to the corresponding state eg. insert the passphrase
+        const stateResult = await TrezorConnect.getDeviceState({
+            device: {
+                path: device.path,
+                instance: device.instance,
+                state: device.state.staticSessionId,
+            },
             useEmptyPassphrase: device.useEmptyPassphrase,
+        });
+
+        if (stateResult.success) {
+            dispatch(
+                applyDeviceStatesThunk({
+                    newDeviceState: stateResult.payload._state,
+                    isAddingHiddenWallet: !device.useEmptyPassphrase,
+                    devicePath: device.path,
+                }),
+            );
+        }
+
+        // NOTE: keep here the previous device as default to prevent TS from screeming
+        const updatedDevice = selectDeviceByStaticSessionId(getState(), staticSessionId) ?? device;
+
+        const result = await TrezorConnect.discoverAccounts({
+            device: updatedDevice,
+            useEmptyPassphrase: updatedDevice.useEmptyPassphrase,
             accounts: accountsParam,
         });
 
@@ -721,14 +744,14 @@ export const runAdditionalDiscoveryThunk = createThunk(
                         error: result.payload.error,
                         errorCode: result.payload.code,
                     },
-                    device.path,
+                    updatedDevice.path,
                 ),
             );
 
             return;
         }
 
-        dispatch(discoveryActions.updateDiscovery({ status: 'complete' }, device.path));
+        dispatch(discoveryActions.updateDiscovery({ status: 'complete' }, updatedDevice.path));
     },
 );
 


mroz22:
doesn't look very nice. 
what about calling TrezorConnect.getDeviceState everytime right before TrezorConnect.discoverAccounts? It should do the trick - make sure that device is in the expected state. discoverAccounts checks this too, but takes long to resolve.

vojtatranta:
> doesn't look very nice. what about calling TrezorConnect.getDeviceState everytime right before TrezorConnect.discoverAccounts? It should do the trick - make sure that device is in the expected state. discoverAccounts checks this too, but takes long to resolve.

I can try it.

vojtatranta:
> doesn't look very nice. what about calling TrezorConnect.getDeviceState everytime right before TrezorConnect.discoverAccounts? It should do the trick - make sure that device is in the expected state. discoverAccounts checks this too, but takes long to resolve.

yeah, so I tried it and pushed it and it just requests the passphrase confirmation twice.

mroz22:
and are you sure that the subsequent TrezorConnect call does not pass  the old device state?

vojtatranta:
> and are you sure that the subsequent TrezorConnect call does not pass the old device state?

yeah, you are right, I didn't passed that, thanks

vojtatranta:
> > and are you sure that the subsequent TrezorConnect call does not pass the old device state?
> 
> yeah, you are right, I didn't passed that, thanks

Is this correct? https://github.com/trezor/trezor-suite/pull/20550/files#diff-eee8bbf1771fe0e239a278c32f649485df1acb663f95c54b08d139d9eac4f031R728

vojtatranta:
> > > and are you sure that the subsequent TrezorConnect call does not pass the old device state?
> > 
> > 
> > yeah, you are right, I didn't passed that, thanks
> 
> Is this correct? https://github.com/trezor/trezor-suite/pull/20550/files#diff-eee8bbf1771fe0e239a278c32f649485df1acb663f95c54b08d139d9eac4f031R728

Coz it still happens that it pops twice, weird

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -702,8 +702,34 @@ export const runAdditionalDiscoveryThunk = createThunk(
...
+                    isAddingHiddenWallet: !device.useEmptyPassphrase,
+                    devicePath: device.path,
+                }),
+            );
+        }
+
         const result = await TrezorConnect.discoverAccounts({
-            device,
+            device: {
+                ...device,
```
mroz22: I think you would need to call a selector again to get the freshly updated device.

vojtatranta: yup, cool, looks to be working

jvaclavik (approved):

suite-common/wallet-core/src/discovery/discoveryThunks.ts
```diff
@@ -703,9 +703,32 @@ export const runAdditionalDiscoveryThunk = createThunk(
...
+
+        if (stateResult.success) {
+            dispatch(
+                applyDeviceStatesThunk({
+                    newDeviceState: stateResult.payload._state,
+                    isAddingHiddenWallet: !device.useEmptyPassphrase,
+                    devicePath: device.path,
+                }),
+            );
+        }
```
mroz22, line 724: will this work? only for postive outcome - do something, else keep as it was before. I would expect here some early return if success.false

vojtatranta: If you fail to set proper one, it runs on ad request again. I  didn't consider it a mistake, You ca surely change it. This was mainly for the happy path

mroz22: It would be nice to handle error cases as well. I would suggest returning here with a toast 'wrong passphrase' or something like this. but I am not a product guy.

vojtatranta: https://github.com/trezor/trezor-suite/issues/20740

PR #20561, opened 2025-08-04
feat(suite-native): device manager refactor

### LINKED ISSUE #20116: Connect another device - improve Discoverability
**Description**
Improve the discoverability and intuitiveness of the "Connect another device" feature on both Android and iOS platforms. The goal is to make it easier for users to find and initiate the process of adding a new Trezor device within the app in both platforms. 

**Acceptance Criteria**
1. The entry point for connecting a new device is easily accessible on both Android and iOS
2. The UI uses clear, action-oriented language to guide users toward adding a new Trezor

**Measurement plan both for iOS and Android**
1. Amount of users that click on the header
2. Amount of users that click on "Connect your device"

BrantalikP:
## Description

enhance device discoverability

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20116

## Screenshots:

https://github.com/user-attachments/assets/c47da9de-0afb-4314-9a8c-763102192a13

https://github.com/user-attachments/assets/de58fab2-a7d1-4024-8c81-8c78c1c87c8b

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/native/enhance-discoverability&lookback=7)

### DIFF
--- a/suite-native/device-manager/src/components/ConnectButton.tsx
+++ b/suite-native/device-manager/src/components/ConnectButton.tsx
@@ -1,0 +1,118 @@
+import { useMemo } from 'react';
+import { Platform } from 'react-native';
+import { FadeInUp, FadeOutUp, LinearTransition } from 'react-native-reanimated';
+import { useSelector } from 'react-redux';
+
+import { useNavigation } from '@react-navigation/native';
+
+import { TrezorDevice } from '@suite-common/suite-types';
+import {
+    selectHasRunningDiscovery,
+    selectInstacelessUnselectedDevices,
+    selectIsNoPhysicalDeviceConnected,
+    selectSelectedDevice,
+} from '@suite-common/wallet-core';
+import { EventType, analytics } from '@suite-native/analytics';
+import { AnimatedBox, Button } from '@suite-native/atoms';
+import { FeatureFlag, useFeatureFlag } from '@suite-native/feature-flags';
+import { IconName } from '@suite-native/icons';
+import { Translation, TxKeyPath } from '@suite-native/intl';
+import {
+    AuthorizeDeviceStackRoutes,
+    RootStackParamList,
+    RootStackRoutes,
+    StackToStackCompositeNavigationProps,
+} from '@suite-native/navigation';
+import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
+
+import { useDeviceManager } from '../hooks/useDeviceManager';
+
+type ConnectButtonProps = {
+    onSelectDevice: (device: TrezorDevice) => void;
+};
+
+const buttonWrapperStyle = prepareNativeStyle(utils => ({
+    paddingTop: utils.spacings.sp32,
+    paddingHorizontal: utils.spacings.sp16,
+}));
+
+type NavigationProp = StackToStackCompositeNavigationProps<
+    RootStackParamList,
+    RootStackRoutes.AppTabs,
+    RootStackParamList
+>;
+
+export const ConnectButton = ({ onSelectDevice }: ConnectButtonProps) => {
+    const { applyStyle } = useNativeStyles();
+    const hasDiscovery = useSelector(selectHasRunningDiscovery);
+    const isNoPhysicalDeviceConnected = useSelector(selectIsNoPhysicalDeviceConnected);
+    const navigation = useNavigation<NavigationProp>();
+    const { setIsDeviceManagerVisible } = useDeviceManager();
+    const isBluetoothEnabled = useFeatureFlag(FeatureFlag.IsBluetoothEnabled);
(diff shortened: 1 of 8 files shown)


suite-native/device-manager/src/components/DeviceManagerModal.tsx
matejkriz, line 1: There is an extra (almost invisible) line on top the Device Switcher on my android phone.
Image

matejkriz (commented):
Except for the extra line on top on Android it looks good code-wise, and the animations are nicely improved 👏 

(I don’t like the design, but we discussed that elsewhere.)

suite-native/device-manager/src/components/ConnectButton.tsx
```diff
@@ -0,0 +1,125 @@
...
+        if (isOnlyBluetoothSupported) {
+            // make sure the device manager is already hidden before navigating to prevent app freezing
+            // TODO: might be fixed by https://github.com/trezor/trezor-suite/issues/17968
+            setTimeout(
+                () =>
+                    navigation.navigate(RootStackRoutes.AuthorizeDeviceStack, {
+                        screen: AuthorizeDeviceStackRoutes.ConnectBluetoothDevice,
+                    }),
+                2 * ANIMATION_DURATION,
+            );
```
juriczech: I know you're not the author of this code, but is this necessary? I wonder how is this different from the else statement where we also redirect. IMO this should be removed.

suite-native/device-manager/src/components/ConnectButton.tsx
```diff
@@ -0,0 +1,125 @@
...
+type NavigationProp = StackToStackCompositeNavigationProps<
+    RootStackParamList,
+    RootStackRoutes.AppTabs,
+    RootStackParamList
+>;
+
+export const ConnectButton = ({ onSelectDevice }: ConnectButtonProps) => {
+    const { applyStyle } = useNativeStyles();
+    const hasDiscovery = useSelector(selectHasRunningDiscovery);
+    const isDeviceConnected = useSelector(selectIsDeviceConnected);
```
matejkriz: nitpick: `selectIsNoPhysicalDeviceConnected` would be slightly better here if you have multiple trezors. Otherwise, we show the "Connect a different Trezor" button if you have the Trezor connected, but select a different view-only trezor.

suite-native/device-manager/src/components/DeviceManagerModal.tsx
BrantalikP: Haha, amazing catch @matejkriz 🙌

The main culprit here is my Monitor setup :D On The laptop one its visible :)

IMG_0140
IMG_0141
IMG_0143

suite-native/device-manager/src/components/ConnectButton.tsx
BrantalikP: You mean the `timeout` no sure about that now it works just fine even without it. but other than that the reason is that this redirects to the `ConnectBluetoothDevice` and the else statement to `ConnectAndUnlockDevice`

suite-native/device-manager/src/components/DeviceManagerModal.tsx
BrantalikP: https://github.com/trezor/trezor-suite/pull/20561/commits/c353251404bb2720fbd55afd0e372a141972c489

suite-native/device-manager/src/components/ConnectButton.tsx
BrantalikP: https://github.com/trezor/trezor-suite/pull/20561/commits/c353251404bb2720fbd55afd0e372a141972c489 like this?

suite-native/device-manager/src/components/ConnectButton.tsx
juriczech: Yes, I understand the difference between these two statements is the target screen 😄 But I think behavior wise technically it should work same and if the timeout isn't required in one, it shouldn't be required in the other one (which you now say that it works even without the timeout?)

If so, I would like to remove it if it's unnecessary complexity please 🙏

BrantalikP: Oki, makes sense, I was worried that there could be some super edge case when a bluetooth device is connected and since mine devkit is flaky I did not want to take a chance, but it seems to be fine. here is a polished version https://github.com/trezor/trezor-suite/pull/20561/commits/51cd51e0863c4b927bb471de574974b3f4d3798c

suite-native/device-manager/src/components/ConnectButton.tsx
matejkriz: yes 👍

matejkriz (approved):

PR #20535, opened 2025-08-01
fix: Temporary downgrade of tenv

HajekOndrej:
## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix-downgrade-tenv&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix-downgrade-tenv&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix-downgrade-tenv&lookback=7)

PR #20533, opened 2025-08-01
Message system 99 - phishing campaigns 

MiroslavProchazka:
Adding message alerting users about ongoing phishing campaigns. 
Tracked in Notion [here](https://www.notion.so/satoshilabs/add-A-B-test-on-trading-remove-Polygon-sync-issues-change-message-variant-of-the-sync-issues-tem-240dc526060680018f1ad752999dd684?source=copy_link)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/msg-system-phishing-alert&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/msg-system-phishing-alert&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/msg-system-phishing-alert&lookback=7)

PR #20563, opened 2025-08-05
Fix Suite Web/Desktop E2E

Lemonexe:
## Description

Fix two unrelated E2E crashes caused by:
- message system interference after https://github.com/trezor/trezor-suite/pull/20533
- FW hash check interference after https://github.com/trezor/trezor-suite/pull/20535
  - (probably, not proven) downgrading tenv was necessary to fix other problems, but changed version of the device, so it went from FW hash check skipped → FW hash check performed → and in E2E it needs to be disabled always when it's done

### Cause

As for message system:
for example the `safety-checks-warning.test.ts` expects this:
image
but gets this and fails because it cannot find [the CTA on the SafetyChecksBanner](https://github.com/trezor/trezor-suite/blob/307890a70774536167d23915178535506d87ede8/packages/suite/src/components/suite/banners/SuiteBanners/SafetyChecksBanner.tsx#L39):
image (1)

Similar case for `backup/t2t1-misc.test.ts` and `backup/t2t1-success.test.ts` which expect [NoBackupBanner](https://github.com/trezor/trezor-suite/blob/307890a70774536167d23915178535506d87ede8/packages/suite/src/components/suite/banners/SuiteBanners/NoBackupBanner.tsx)

### Notes

`page.route -> route.fulfill` is broken on playwright https://github.com/microsoft/playwright/issues/30495
(we use `onBeforeRequest` in electron-main, and it breaks it)

I tried to circumvent it [at here](https://github.com/trezor/trezor-suite/blob/e484640ea2e618d3f78e6e3104f0857520c4cf0a/packages/suite-desktop-core/e2e/support/testExtends/suiteBaseFixture.ts#L51), but it's no good, the fetch is intercepted but promise freezes pending:
```
    const encoded = encodeURIComponent(validJws);
    const redirectURL = `data:text/plain,${encoded}`;
    await suite.electronApp.evaluate(async ({ session }) => {
        session.defaultSession.webRequest.onBeforeRequest(
            { urls: ['https://data.trezor.io/config/develop/config.v1.jws'] },
            (_details, callback) => {
                callback({ redirectURL });
            },
        );
    });
```

I also tried this hack but did not finish it, the `.catch` does not suppress the error
```
/**
 * Dismiss all dismissible message system banners if any are present.
 */
export const optionallyDismissMessageSystemBanners = (page: Page) => {
    page.getByTestId(/@message-system\/.+\/dismiss/)
        .click()
        // check if any more banners are present behind the one we just dismissed
        .then(() => optionallyDismissMessageSystemBanners(page))
        // Nothing matched; suppress error.
        .catch();
};
```

---

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/e2e&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/e2e&lookback=7)

### DIFF
--- a/packages/suite-desktop-core/e2e/support/common.ts
+++ b/packages/suite-desktop-core/e2e/support/common.ts
@@ -1,8 +1,9 @@
-import test, { Locator, TestInfo } from '@playwright/test';
+import test, { Locator, Page, TestInfo } from '@playwright/test';
 import { isEqual, omit } from 'lodash';
 import { readdirSync } from 'node:fs';
 import path from 'node:path';
 
+import { validJws } from '@suite-common/message-system/src/__fixtures__/messageSystemActions';
 import { TradingCountryCode, regional } from '@suite-common/trading';
 import { getAccountDecimals, localizeNumber } from '@suite-common/wallet-utils';
 import { TrezorUserEnvLink } from '@trezor/trezor-user-env-link';
@@ -157,3 +158,11 @@ export const getBigNumberFromBalance = async (locator: Locator) => {
 
     return { originalBalance, hasEllipsis };
 };
+
+/**
+ * Mocks remote message-system with an empty JWS config signed by develop key.
+ */
+export const mockRemoteMessageSystem = async (page: Page): Promise<void> =>
+    await page.route('**/config.v1.jws', async route => {
+        await route.fulfill({ status: 200, body: validJws });
+    });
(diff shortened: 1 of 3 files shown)


OriginalEveres (approved):
lgtm

Vere-Grey (approved):
:+1:

Issue #20320, opened 2025-07-22
View only by default - Analytics


fang314:
We would like the following analytics for view only by default: 

1. Amount of users that disable view-only by default **in initial prompt**
2. ~Amount of users that decide on automatically eject wallets~
3. Amount of users that enable or disable view-only by default **in settings** page

seibei-iguchi:
I don't understand the difference between the three variants. They are the same thing, at least in the desktop app, and that is already passed along to analytics.

juriczech:
Implement for device disconnect: https://github.com/trezor/trezor-suite/issues/20547#issuecomment-3149968718

PR #20579, opened 2025-08-05
Chore/native/auto eject message on disconnect during discovery

### LINKED ISSUE #20547: Trezor Unplug During Discovery - For new view-only by default behavior
**Context of the problem**
The issue is that when a user unplugs their Trezor during the initial discovery, the process halts, leaving the app in an empty state with no assets remembered. For subsequent discoveries, unplugging within approximately 5 seconds prevents completion, but otherwise, the process may still finalize

**Proposed Solution**
To address this and given we will remember wallets by default, we will implement an error message to inform users that their balances may be incomplete or inaccurate if the device is unplugged during discovery ([meeting notes](https://www.notion.so/satoshilabs/View-only-by-default-unplug-use-case-242dc52606068076a9ddfa2cfa502eb7?source=copy_link))

[Figma Design](https://www.figma.com/design/aybSBSnUWPebHVCpFvVtrj/Auto-eject?node-id=2410-6831&m=dev)

juriczech:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20547

## Screenshots:

https://github.com/user-attachments/assets/8d6c5bb0-b9b3-4cdb-945d-53298a920665

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/native/auto-eject-message-on-disconnect&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/native/auto-eject-message-on-disconnect&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/auto-eject-message-on-disconnect&lookback=7)

### DIFF
--- a/suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
+++ b/suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
@@ -24,7 +24,7 @@ export const EjectedDuringDiscoveryWarning = () => {
                 <InlineAlertBox
                     title={<Translation id="assets.dashboard.ejectedDuringDiscovery" />}
                     variant="warning"
-                    iconName="trezorDevices"
+                    customIcon={<DeviceItemIcon deviceId={device?.id} iconSize="mediumLarge" />}
                 />
             </Box>
         </Animated.View>
--- a/suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
+++ b/suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
@@ -10,7 +10,13 @@ import {
     selectHasDeviceBeenEjectedDuringDiscovery,
 } from '@suite-native/settings';
 
-export const EjectedDuringDiscoveryWarning = () => {
+type EjectedDuringDiscoveryWarningProps = {
+    hasPadding?: boolean;
+};
+
+export const EjectedDuringDiscoveryWarning = ({
+    hasPadding = false,
+}: EjectedDuringDiscoveryWarningProps) => {
     const device = useSelector(selectSelectedDevice);
     const hasBeenEjectedDuringDiscovery = useSelector((state: SettingsSliceRootState) =>
         selectHasDeviceBeenEjectedDuringDiscovery(state, device?.id),
@@ -20,7 +26,7 @@ export const EjectedDuringDiscoveryWarning = () => {
 
     return (
         <Animated.View>
-            <Box padding="sp8">
+            <Box padding={hasPadding ? 'sp8' : undefined}>
                 <InlineAlertBox
                     title={<Translation id="assets.dashboard.ejectedDuringDiscovery" />}
                     variant="warning"
--- a/suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
+++ b/suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
@@ -1,0 +1,32 @@
+import Animated from 'react-native-reanimated';
+import { useSelector } from 'react-redux';
+
+import { selectSelectedDevice } from '@suite-common/wallet-core';
+import { Box, InlineAlertBox } from '@suite-native/atoms';
+import { DeviceItemIcon } from '@suite-native/device';
+import { Translation } from '@suite-native/intl';
+import {
+    SettingsSliceRootState,
+    selectHasDeviceBeenEjectedDuringDiscovery,
+} from '@suite-native/settings';
+
+export const EjectedDuringDiscoveryWarning = () => {
+    const device = useSelector(selectSelectedDevice);
+    const hasBeenEjectedDuringDiscovery = useSelector((state: SettingsSliceRootState) =>
+        selectHasDeviceBeenEjectedDuringDiscovery(state, device?.id),
+    );
+
+    if (!hasBeenEjectedDuringDiscovery || device?.connected) return null;
+
+    return (
+        <Animated.View>
+            <Box padding="sp8">
+                <InlineAlertBox
+                    title={<Translation id="assets.dashboard.ejectedDuringDiscovery" />}
+                    variant="warning"
+                    iconName="trezorDevices"
+                />
+            </Box>
+        </Animated.View>
+    );
+};
--- a/suite-native/discovery/src/discoveryMiddleware.ts
+++ b/suite-native/discovery/src/discoveryMiddleware.ts
@@ -8,16 +8,20 @@ import {
     deviceActions,
     discoveryActions,
     restartDiscoveryThunk,
+    runAdditionalDiscoveryThunk,
     selectSelectedDevice,
     selectShouldRediscoverNetworks,
     startDiscoveryThunk,
 } from '@suite-common/wallet-core';
 import {
     createAndBackupWalletThunk,
+    isDeviceEventAction,
     recoverWalletThunk,
     selectIsDeviceFirmwareSupported,
 } from '@suite-native/device';
 import {
+    removeDeviceFromEjectedList,
+    selectHasDeviceBeenEjectedDuringDiscovery,
     selectIsCoinEnablingInitFinished,
     setIsCoinEnablingInitFinished,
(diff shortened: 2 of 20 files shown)


suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    console.warn('ejected during discovery');
+
+    return (
+        <Animated.View>
+            <Box padding="sp8">
+                <InlineAlertBox
+                    title="Reconnect your trezor to load all assets."
+                    variant="warning"
+                    iconName="trezorDevices"
```
juriczech: TODO: change icon

suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const EjectedDuringDiscoveryWarning = () => {
+    const device = useSelector(selectSelectedDevice);
+    const hasBeenEjectedDuringDiscovery = useSelector((state: SettingsSliceRootState) =>
+        selectHasDeviceBeenEjectedDuringDiscovery(state, device?.id),
+    );
+
+    if (!hasBeenEjectedDuringDiscovery) return null;
+
+    console.warn('ejected during discovery');
```
matejkriz: TODO: delete

suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    return (
+        <Animated.View>
+            <Box padding="sp8">
+                <InlineAlertBox
+                    title="Reconnect your trezor to load all assets."
```
matejkriz: TODO: make translatable

matejkriz:
You also mentioned analytics for this here https://github.com/trezor/trezor-suite/issues/20547#issuecomment-3149968718, but it's not implemented here, right? Do you count that as a separate issue/PR?

juriczech:
@matejkriz Analytics are [here](https://github.com/trezor/trezor-suite/pull/20383). Still in draft but after this PR it will be ready for rebase and finishing.

suite-native/settings/src/settingsSlice.ts
```diff
@@ -12,6 +12,7 @@ export interface AppSettingsState {
     isFirmwareHashCheckEnabled: boolean;
     areTestnetsEnabled: boolean;
     shouldShowAutoEjectAlert: boolean;
+    disconnectedDevicesDuringDiscovery: Device['id'][];
```
matejkriz, line 15: Is it necessary to introduce a state for this? What about using the discovery state? Checking if it's failed with errorCode `Device_Disconnected`? 

If we really need the list of devices disconnected during discovery in the state, it doesn't belongs to settings slice and we need to handle deleting the device reference on "eject"

juriczech:
> @matejkriz Analytics are [here](https://github.com/trezor/trezor-suite/pull/20383). Still in draft but after this PR it will be ready for rebase and finishing.

I also added the comment there: https://github.com/trezor/trezor-suite/issues/20320#issuecomment-3167027090

suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
juriczech: https://github.com/trezor/trezor-suite/pull/20579/commits/db7bdeb6ec75c9e97b47b5d7f3e124862ea8484a

suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
```diff
@@ -0,0 +1,33 @@
...
+    SettingsSliceRootState,
+    selectHasDeviceBeenEjectedDuringDiscovery,
+} from '@suite-native/settings';
+
+export const EjectedDuringDiscoveryWarning = () => {
+    const device = useSelector(selectSelectedDevice);
+    const hasBeenEjectedDuringDiscovery = useSelector((state: SettingsSliceRootState) =>
+        selectHasDeviceBeenEjectedDuringDiscovery(state, device?.id),
+    );
+
```
matejkriz, line 24: One of the conditions here should be if the device is connected, I guess, since I can see the warning to Reconnect your trezor even though the trezor is connected.

matejkriz (commented):
I reconnect the trezor, but the discovery never resumed 🤔

suite-native/settings/src/settingsSlice.ts
matejkriz: Maybe even better condition would be if selected device is not connected and the discovery is not completed? (That would work even for the case when you enable more coins while the device is disconnected. EDIT: It would not work for that case without more changes. If you enable more coins while the device is disconnected, discovery state remain in complete state until the device is connected)

suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
juriczech: Fixed

suite-native/settings/src/settingsSlice.ts
juriczech: Hmmm, initially, I wanted to do it like that, but the key value for the discovery object is device path, which is not present when device is disconnected. So after device is disconnected, we can look if there was a discovery that failed with `Device_Disconnected` error code, but we want be able to match it to persisted device 🤔 

Yeah, I guess ideally, I would add it to discoveryReducer, but not sure if it's a good idea since it's platform specific for mobile app and extending the discoveryReducer seemed like an overkill. I will look into it whether it would make sense to create ejectWallets reducer...

suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
juriczech: https://github.com/trezor/trezor-suite/pull/20579/commits/92a1912441fe7ab6c94f75512755443aa39cc9ab

suite-native/settings/src/settingsSlice.ts
matejkriz: I see now, discovery is not even persistent so it can't be used like that. But still we have to be able to check somehow that discovery is not complete and needs to re-run.

suite-native/discovery/src/discoveryMiddleware.ts
```diff
@@ -80,7 +92,12 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
                         device.state.staticSessionId,
                     );
 
-                    if (shouldRediscover) {
+                    // If device is reconnected after ejecting during discovery, restart the discovery process
+                    const isConnectActionAndHasBeenEjectedDuringDiscovery =
+                        isDeviceEventAction(action, DEVICE.CONNECT) &&
+                        selectHasDeviceBeenEjectedDuringDiscovery(getState(), action.payload.id);
+
+                    if (shouldRediscover || isConnectActionAndHasBeenEjectedDuringDiscovery) {
```
matejkriz, line 100: Why `shouldRediscover` is not sufficient here?

matejkriz: I see, the `selectHasDeviceBeenEjectedDuringDiscovery` is checking only if there is any account with enabled network, but it doesn't mean the all of them were discovered.

juriczech: Exactly! I didn't want to change this at the moment, but I think we should think again about the triggers of discovery and figure out a way to optimize it.

suite-native/assets/src/components/EjectedDuringDiscoveryWarning.tsx
juriczech: Fixed

PR #20621, opened 2025-08-06
20525 windows hello thread

### LINKED ISSUE #20525: Run Windows Hello binary in a separate thread
The windows hello binary uses `process.dlopen` which cannot be easily cleared from the memory once it is no longer needed.

The better solution is to run the windows hello binary in a child process which can be killed once the interaction is finished.

vojtatranta:
## Description

As promised, I am moving the binary dlOpen in the web worker

But I found this https://www.electronjs.org/docs/latest/tutorial/multithreading#native-nodejs-modules That's sad...

So I didnt use the web worker, I used child process.

## Related Issue

Resolve  https://github.com/trezor/trezor-suite/issues/20525

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=20525-windows-hello-thread&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=20525-windows-hello-thread&lookback=7)


packages/suite-desktop/package.json
```diff
@@ -7,7 +7,7 @@
     "homepage": "https://trezor.io/",
     "main": "dist/app.js",
     "scripts": {
-        "dev": "yarn clean && LAUNCH_ELECTRON=true yarn run dev:local",
+        "dev": "yarn clean && yarn build:core && LAUNCH_ELECTRON=true yarn run dev:local",
```
Lemonexe: sus, this shouldn't be needed.
So I started investigating, and hey `build:core` is never explicitly called in any package.json script in this flow, so why did it always work?
It's because `dev:prepare` is called by a webpack plugin [here](https://github.com/trezor/trezor-suite/blob/85c704203a147dbe03b0d291dccfc10aaa125bac/packages/suite-build/configs/desktop.webpack.config.ts#L122). Wow.

TL;DR: revert this? It's not needed.

packages/suite-desktop-native/src/winHelloChildProcess.ts
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        const binaryPath = path.join(resourcesPath, 'bin/win_hello.node');
+
+        try {
+            const localRequiredBinary = require('../../suite-data/files/bin/win_hello.node');
+            this.winHello = localRequiredBinary as typeof WinHelloTypes | null;
+        } catch (err) {
+            console.warn(
+                `[WinHelloChildProcess] Failed to load local Windows Hello native module: ${(err as Error).message}`,
```
Lemonexe: Shilling my utils :slightly_smiling_face: 
But rly, you caught error is to be considered `unknown`. Could be `string`, and then this would throw.
```suggestion
                `[WinHelloChildProcess] Failed to load local Windows Hello native module: ${serializeError(err)}`,
```

packages/suite-desktop-native/src/winHelloChildProcess.ts
```diff
@@ -0,0 +1,175 @@
...
+            }
+            const module = { exports: {} };
+            try {
+                // NOTE: using dlopen instead of require() to avoid problems with Electron that threw module not found error
+                // even when the file existed in the production build
+                process.dlopen(module, binaryPath);
+                this.winHello = module.exports as typeof WinHelloTypes | null;
+            } catch (err) {
+                console.warn(
+                    `[WinHelloChildProcess] Failed to load Windows Hello native module with dlopen: ${(err as Error).message}`,
```
Lemonexe: same
```suggestion
                    `[WinHelloChildProcess] Failed to load Windows Hello native module with dlopen: ${serializeError(err)}`,
```

packages/suite-desktop-native/src/winHelloChildProcess.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        this.initialize(resourcesPath);
+    }
+
+    private initialize(resourcesPath: string) {
+        if (!this.isWindows) {
+            return;
+        }
+
+        const binaryPath = path.join(resourcesPath, 'bin/win_hello.node');
```
Lemonexe: didn't notice it before, but I suggest this is safer on windows
```suggestion
        const binaryPath = path.join(resourcesPath, 'bin', 'win_hello.node');
```

packages/suite-desktop-native/src/types.ts
```diff
@@ -0,0 +1,42 @@
...
+    params?: {
+        message?: string;
+        windowHandle?: Buffer | null;
+    };
+}
+
+export interface IPCResponse {
+    id: string;
+    success: boolean;
+    result?: any;
```
Lemonexe: this is a substantial downgrade.
Previously, both functions created by `createWinHello` had exact return types.
Now both are mixed into `any` 
Pls split into union of the two responses.

packages/suite-desktop-native/src/winHelloChildProcess.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    public start() {
+        if (!process.send) {
+            throw new Error('This script must be run as a child process');
+        }
+
+        process.on('message', async (message: IPCMessage) => {
+            if ('method' in message) {
+                const response = await this.handleRequest(message);
+                process.send!(response);
```
Lemonexe: nit: assign process.send into variable at the beginning of this method, then you won't need `!`

packages/suite-desktop-native/src/winHelloChildProcess.ts
```diff
@@ -0,0 +1,175 @@
...
+            };
+        }
+    }
+
+    public start() {
+        if (!process.send) {
+            throw new Error('This script must be run as a child process');
+        }
+
+        process.on('message', async (message: IPCMessage) => {
```
Lemonexe: doesn't make sense, when can this be `IPCResponse`?? This should be `IPCRequest`, and the `IPCMessage` type seems useless.

packages/suite-desktop-native/src/winHelloChildProcess.ts
```diff
@@ -0,0 +1,175 @@
...
+                return result;
+            } else {
+                throw new Error(errorMessageToStandardError(result));
+            }
+        });
+    }
+
+    private async handleRequest(request: IPCRequest): Promise<IPCResponse> {
+        try {
+            let result: any;
```
Lemonexe, line 114: ad types.ts: the `any` has to be removed also here

EDIT: now it can be : `string | boolean` with no problem :)

packages/suite-desktop-native/src/winHelloManager.ts
```diff
@@ -0,0 +1,248 @@
...
+     */
+    public create({
+        resourcesPath,
+        logger,
+    }: {
+        resourcesPath: string;
+        logger: {
+            info: (topic: string, message: string) => void;
+        };
+    }): Promise<void> {
```
Lemonexe: pls extract type for better readability :pray:

packages/suite-desktop-native/src/winHelloManager.ts
```diff
@@ -0,0 +1,248 @@
...
+                    },
+                });
+                logger.info('win-hello', 'Child process created');
+
+                this.childProcess.on('message', (message: any) => {
+                    logger.info(
+                        'win-hello',
+                        `Received message from child: ${JSON.stringify(message)}`,
+                    );
+                    if (message.ready) {
```
Lemonexe: you sure `message.ready` exists? It's typed as `any` so who knows

packages/suite-desktop-native/src/winHelloManager.ts
```diff
@@ -0,0 +1,248 @@
...
+        };
+
+        return new Promise((resolve, reject) => {
+            this.pendingRequests.set(id, { resolve, reject });
+
+            // Set timeout for the request
+            const timeout = setTimeout(() => {
+                this.pendingRequests.delete(id);
+                reject(new Error(`Request timeout for method: ${method}`));
+            }, 30000); // 30 second timeout
```
Lemonexe: pls extract to constant in module top scope

packages/suite-desktop-native/src/winHelloManager.ts
```diff
@@ -0,0 +1,248 @@
...
+
+export async function createWinHelloManager({
+    resourcesPath,
+    logger,
+}: {
+    resourcesPath: string;
+    logger: {
+        info: (topic: string, message: string) => void;
+    };
+}): Promise<WinHelloManager> {
```
Lemonexe: pls extract type for better readability :pray:

packages/suite-desktop-native/src/winHelloManager.ts
```diff
@@ -0,0 +1,248 @@
...
+            const killed = this.childProcess!.kill();
+            if (!killed) {
+                // If kill failed, force cleanup
+                cleanup();
+            }
+
+            // Force cleanup after timeout
+            setTimeout(() => {
+                cleanup();
+            }, 5000);
```
Lemonexe: pls extract to constant in module top scope

packages/suite-desktop-native/src/winHelloManager.ts
```diff
@@ -0,0 +1,248 @@
...
+                );
+            }
+
+            // Set a timeout for initialization
+            setTimeout(() => {
+                if (!this.isReady) {
+                    this.destroy();
+                    reject(new Error('Child process initialization timeout'));
+                }
+            }, 30000);
```
Lemonexe: pls extract to constant in module top scope

packages/suite-desktop-native/src/winHelloManager.ts
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        return new Promise(resolve => {
+            const cleanup = () => {
+                this.cleanup();
+                resolve();
+            };
+
+            // Try graceful shutdown first
+            this.childProcess!.once('exit', cleanup);
```
Lemonexe: nit: assign `this.childProcess` into variable at the beginning of this method, then you won't need `!`

packages/suite-desktop-native/src/winHelloManager.ts
```diff
@@ -0,0 +1,248 @@
...
+     * Requests Windows Hello authentication
+     */
+    public requestHello(
+        message: string = 'Verify your identity',
+        windowHandle: Buffer | null = null,
+    ): Promise<string> {
+        return this.sendRequest('requestHello', { message, windowHandle });
+    }
+
+    private sendRequest(method: 'isHelloAvailable' | 'requestHello', params?: any): Promise<any> {
```
Lemonexe: again, any any

Lemonexe (commented):
Nice refactoring lifecycle to class.
But pls fix the types, this refactoring makes type-safety a lot worse.
+ many comments.
Did not test I'm not on windows. Did you?

packages/suite-desktop/package.json
vojtatranta: yup thanks, I thought so

vojtatranta:
> Nice refactoring lifecycle to class. But pls fix the types, this refactoring makes type-safety a lot worse.
> 
> * many comments.
>   Did not test I'm not on windows. Did you?

sure, tested on win, will test again with build, dev mode etc

vojtatranta:
Damn it!!!

The win hello alert window is not in the foreground when it is run in child process omg.

vojtatranta:
> Damn it!!!
> 
> The win hello alert window is not in the foreground when it is run in child process omg.

Cool, AI was able to fix it and simplified the code

vojtatranta:
Tested build, tested dev, works well.

packages/suite-desktop-native/src/winHelloChildProcess.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+};
+
+const serializeError = (error: unknown): string => {
+    if (error instanceof Error) {
+        return error.message;
+    }
+
+    return String(error);
+};
```
Lemonexe, line 23: Well, I meant `import { serializeError } from '@trezor/utils';` but fine it can be here too xD
I mean without irony, it's a small util, here we don't have to count with exotic types so maybe it's better readable to have it here, idc.

packages/suite-desktop-native/src/types.ts
```diff
@@ -0,0 +1,62 @@
...
+    error: string;
+}
+
+export type IPCResponse = IPCSuccessResponse | IPCErrorResponse;
+
+export interface IPCReadyMessage {
+    ready: true;
+}
+
+export type IPCMessage = IPCRequest | IPCResponse | IPCReadyMessage;
```
Lemonexe: unused, delet this

packages/suite-desktop-native/src/types.ts
```diff
@@ -0,0 +1,62 @@
...
+    method: 'isHelloAvailable' | 'requestHello';
+    params?: {
+        message?: string;
+    };
+}
+
+export interface IPCSuccessResponse {
+    id: string;
+    success: true;
+    result: boolean | string; // isHelloAvailable returns boolean, requestHello returns string
```
Lemonexe: tiny nit: extracting types for both methods would remove the need for comment

packages/suite-desktop-native/src/win-hello.cpp
```diff
@@ -93,39 +95,90 @@ Napi::Value isHelloAvailable(const Napi::CallbackInfo& info) {
     }
 }
 
+// Helper function to find and elevate Windows Hello dialog
+HWND FindAndElevateHelloWindow() {
+    HWND helloWindow = FindWindowW(L"Credential Dialog Xaml Host", NULL);
```
Lemonexe, line 100: can this be extracted to a constant? And comment that it's a hardcoded reference to a class name, and this will break if the class is renamed in future windows.
That will make future debugging easier (for our artificial friends)

packages/suite-desktop-native/src/win-hello.cpp
```diff
@@ -93,39 +95,90 @@ Napi::Value isHelloAvailable(const Napi::CallbackInfo& info) {
...
+            EnableWindow(helloWindow, TRUE);
+            
+            // 6. Try to flash the window to get user attention
+            FLASHWINFO fi;
+            fi.cbSize = sizeof(FLASHWINFO);
+            fi.hwnd = helloWindow;
+            fi.dwFlags = FLASHW_ALL | FLASHW_TIMERNOFG;
+            fi.uCount = 3;
+            fi.dwTimeout = 0;
+            FlashWindowEx(&fi);
```
Lemonexe, line 161: seems to me that here are 6 alternatives how to use window api to focus the window. But this code does all of them. Seems unnecessary, but idk, maybe it's ok.
:thinking: but maybe it could modified, that after each method, it could check again if success:
if GetForegroundWindow() is helloWindow, then return.

Lemonexe (commented):
TS part looks codewise great, just some final nits :ok_hand: 
C++ good I guess? Idk. I had it explained to me by copilot, + some ideas of my own, but those could be BS.

I'll test and then approve :slightly_smiling_face:

Lemonexe (approved):
LGTM, works as supposed to :rocket:

packages/suite-desktop-native/src/win-hello.cpp
vojtatranta: AI says there isnt a variable for that.. SO yeah, it might break...

packages/suite-desktop-native/src/win-hello.cpp
vojtatranta: fxd

packages/suite-desktop-native/src/win-hello.cpp
Lemonexe: yeah I meant literally just extracting a local constant for code clarity
something like `const xamlHostClassName = 'Credential Dialog Xaml Host'` but in c++
but whatever, that's a nit

PR #20646, opened 2025-08-07
refactor(suite): extract suiteReducer selectors

Lemonexe:
## Description

`suiteReducer` file is very long, because it contains a complex reducer and a lot of selectors
→ extract to selectors files, like it has been done in most `suite-common/wallet-core` reducers.

I have creates `suiteSelectors.ts`, but also `suiteAuthenticityChecksSelectors.ts` because they are long boilerplate, depend on the former, and are kind of a self-contained topic.

### Notes

:information_source:  The reasoning in wallet-core was not just splitting files, but also circular dependencies. There might be some here, but who knows :shrug: 

:thought_balloon: I know that this is not an ideal solution. The `suite` package is a yuuge monolith that should be split into subpackages sorted by concern, not into the grand directories like "selectors" and "reducers"  and "actions"  (sorted by technology instead of concern). But that'd be a long road ahead. This is low hanging fruit, kinda.

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=refactor/extract-suite-selectors&lookback=7)

### DIFF
--- a/packages/suite-desktop-ui/src/support/DesktopUpdater/Available.tsx
+++ b/packages/suite-desktop-ui/src/support/DesktopUpdater/Available.tsx
@@ -6,7 +6,7 @@ import { download } from 'src/actions/suite/desktopUpdateActions';
 import { setFlag } from 'src/actions/suite/suiteActions';
 import { MarkdownWithComponents, Translation } from 'src/components/suite';
 import { useDispatch, useSelector } from 'src/hooks/suite';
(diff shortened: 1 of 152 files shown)


OriginalEveres (approved):
looks good to me! This is improvement!

There are some selectors like: `selectFirmwareRevisionCheckError ` that could be in the common package, but thats for a bigger investigation

PR #20650, opened 2025-08-07
fix(suite): fix amount recalculation on currency switch

### LINKED ISSUE #19895: Send form: Amount Not Recalculated on Currency Switch
**Description of the bug**
When entering an amount and then switching the fiat currency, the Bitcoin amount does not recalculate according to the newly selected currency

**Steps to Reproduce:**
1. Start a new transaction.
2. Enter an amount (e.g., 5) in the default fiat currency (e.g., USD).
3. Switch the fiat currency to another (e.g., CZK).
4. Observe the displayed Bitcoin amount.

**Info:**
1. Trezor Suite Version: 25.6.1 (& 25.7 develop)

**Expected behavior:**
The Bitcoin amount should be recalculated based on the newly selected fiat currency and the entered amount

**Attachments**

Image

OriginalEveres:
Fixes the amount recalculation when user switches currency.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19895

## Screenshots:
**BEFORE**
461386807-2a954447-3833-4e54-b20b-d5af5cbf8f05

**AFTER**
https://github.com/user-attachments/assets/b3bd0a2b-8d3f-4c37-92d9-155af124a2b1

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/currency-switch-recalculate-amount&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/currency-switch-recalculate-amount&lookback=7)

### DIFF
--- a/packages/suite/src/views/wallet/send/Outputs/Amount/Amount.tsx
+++ b/packages/suite/src/views/wallet/send/Outputs/Amount/Amount.tsx
@@ -60,12 +60,12 @@ export const Amount = ({ output, outputId }: AmountProps) => {
     const amountName = `outputs.${outputId}.amount` as const;
     const tokenInputName = `outputs.${outputId}.token`;
     const maxOutputId = getDefaultValue('setMaxOutputId');
-    const isSetMaxActive = maxOutputId === outputId;
+    const isSendMaxActive = maxOutputId === outputId;
     const outputError = errors.outputs ? errors.outputs[outputId] : undefined;
     const error = outputError ? outputError.amount : undefined;
 
     // corner-case: do not display "setMax" button if FormState got ANY error (setMax probably cannot be calculated)
-    const isSetMaxVisible = isSetMaxActive && !error && !Object.keys(errors).length;
+    const isSendMaxVisible = isSendMaxActive && !error && !Object.keys(errors).length;
     const maxSwitchId = `outputs.${outputId}.setMax`;
 
     const amountValue = getDefaultValue(amountName, output.amount || '');
@@ -136,13 +136,13 @@ export const Amount = ({ output, outputId }: AmountProps) => {
     const onSwitchChange = () => {
         const clearInput = network.networkType === 'solana';
 
-        setMax(outputId, isSetMaxActive, clearInput);
+        setMax(outputId, isSendMaxActive, clearInput);
         composeTransaction(amountName);
     };
 
     const sendMaxSwitch = (
         <SendMaxSwitch
-            isSetMaxActive={isSetMaxActive}
+            isSendMaxActive={isSendMaxActive}
             data-testid={maxSwitchId}
             onChange={onSwitchChange}
         />
@@ -159,10 +159,10 @@ export const Amount = ({ output, outputId }: AmountProps) => {
                     inputState={inputState}
                     locale={locale}
                     labelHoverRight={
-                        !isSetMaxVisible && (!isWithBaseCurrency || isBelowLaptop) && sendMaxSwitch
+                        !isSendMaxVisible && (!isWithBaseCurrency || isBelowLaptop) && sendMaxSwitch
                     }
                     labelRight={
-                        isSetMaxVisible && (!isWithBaseCurrency || isBelowLaptop) && sendMaxSwitch
+                        isSendMaxVisible && (!isWithBaseCurrency || isBelowLaptop) && sendMaxSwitch
                     }
                     labelLeft={
                         <Row>
@@ -198,6 +198,7 @@ export const Amount = ({ output, outputId }: AmountProps) => {
                                     <BaseCurrencyInput
                                         output={output}
                                         outputId={outputId}
+                                        isSendMaxActive={isSendMaxActive}
                                         // To fix alignment with the other input
                                         labelLeft={isBelowLaptop ? undefined : <>&nbsp;</>}
                                         labelRight={!isBelowLaptop && sendMaxSwitch}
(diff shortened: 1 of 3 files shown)


OriginalEveres:
just noticed that... @fang314 there was a previous implementation that got broken, but it was behaving differently:

**broken(now fixed) implementation**: you wanted to send 10 USD worth of BTC, but you decided to switch the USD to EUR -> it did not recalculate the BTC amount, but it converted 10 USD to 8.5 EUR. This is nice when you want to see the value of your BTC when sending using send max. 

**your expected result**: you want to send 10 USD of BTC, then you decide to switch to EUR -> 10 USD goes to 10 EUR, but the BTC is recalculated.

from UX perspective I believe that, if send max is active, we should just show what's the BTC value for different currencies, but when send max is not active, we should just keep the amount of currency and recalculate the BTC amount. wdyt?

komret (changes requested):
As @fang314, I'd expect it to recalculate the crypto amount, not fiat. I think it is more intuitive and a more common use case is probably typing the amount, then wanting to change the fiat currency.

Also, as this seems prone to become broken, can you add some test? Perhaps it can be integrated into some e2e test @evgenysl can help with that.

fang314:
hi guys, sorry for being late to the party

**Send max = Active**: umm, I think if send max is active there is no other choice but to recalculate the Fiat currency right? given that you can't send more e.g. Bitcoin than what you have in the balance

**Send max = NOT active**: yes, I think it would make sense to recalculate the amount in crypto (and not fiat) that users are trying to send. This is because, a little bit more than half of all our users think in terms of fiat when sending an amount ([data](https://satoshilabs.slack.com/archives/C011BRDBEAZ/p1755857724436549?thread_ts=1755768592.488479&cid=C011BRDBEAZ))

evgenysl:
Something is not right with BTC calculation.
1. Enter 10 USD
2. Observe error "Enter a whole number"

EDIT: bugged only with Satoshi units

Screenshot 2025-08-25 at 16 32 27
Screenshot 2025-08-25 at 16 35 09

packages/suite/src/views/wallet/send/Outputs/Amount/BaseCurrencyInput.tsx
```diff
@@ -84,23 +87,40 @@ export const BaseCurrencyInput = ({
...
+                : parseBaseCurrencyToFormattedCrypto({
+                      ...baseFormatOptions,
+                      value: fiatValue,
+                      isCryptoInSats: shouldSendInSats === true,
+                      cryptoDecimals,
+                  });
+
+            if (formattedAmount !== null) {
+                setValue(
+                    isSendMaxActive ? baseCurrencyInputName : amountInputName,
```
peter-sanderson: I really dont like this. I know this is dreaded sendform, but I think we can do a bit better here

```ts
if (isSendMaxActive) {
    const bla = parseCryptoToFormattedBaseCurrency({
                      ...baseFormatOptions,
                      value: cryptoValue,
                      baseCurrencyCode,
                      baseCurrencyToSats: shouldSendInSats === true,
                  })
                  
                  setValue(baseCurrencyInputName, bla)
} else {
    const ble = parseBaseCurrencyToFormattedCrypto({
                      ...baseFormatOptions,
                      value: fiatValue,
                      isCryptoInSats: shouldSendInSats === true,
                      cryptoDecimals,
                  })
    setValue(amountInputName, ble)
}
```

Basically split the flow just once, instead of twice

OriginalEveres: good point, I have already wrote a todo for refactoring of BaseCurrencyInput, but we can make it better now.

packages/suite/src/views/wallet/send/Outputs/Amount/Amount.tsx
```diff
@@ -136,13 +136,13 @@ export const Amount = ({ output, outputId }: AmountProps) => {
...
 
-        setMax(outputId, isSetMaxActive, clearInput);
+        setMax(outputId, isSendMaxActive, clearInput);
         composeTransaction(amountName);
     };
 
     const sendMaxSwitch = (
         <SendMaxSwitch
-            isSetMaxActive={isSetMaxActive}
+            isSetMaxActive={isSendMaxActive}
```
5K1PY: What about renaming this prop as well?

5K1PY (approved):
:shipit:

peter-sanderson (approved):

packages/suite/src/views/wallet/send/Outputs/Amount/BaseCurrencyInput.tsx
```diff
@@ -84,26 +87,42 @@ export const BaseCurrencyInput = ({
...
+                    cryptoDecimals,
+                });
+                setValue(amountInputName, formattedAmount || '', {
                     shouldValidate: true,
                 });
-                // call compose to store draft, precomposedTx should be the same
-                composeTransaction(amountInputName);
             }
+
+            // call compose to store draft, precomposedTx should be the same
```
komret: If `isSendMaxActive`, we should in fact recompose the transaction because crypto amount changes, so the comment is off.
If it is not active, crypto amount does not change, so there is no need to recompose. We should store draft, but that is already handled by a `useEffect` in `useSendForm`, so we should not do anything here.

Side quest - you could investigate why `storeDraft` is triggered twice when base currency changes, this is some unrelated bug.

komret:
I observed similar bug as the one found by @evgenysl. Switching between base currencies is broken when the unit is switched to satoshis.

[Screencast from 2025-09-03 15-46-00.webm](https://github.com/user-attachments/assets/54af03fb-d77b-45b1-99aa-9b9df83f2406)

komret (approved):

PR #20660, opened 2025-08-08
fix(connect): remove call to method.confirmation since it might use m…

mroz22:
…ethod.device which is not defined at that moment

it would crash on methods like this - https://github.com/trezor/trezor-suite/blob/develop/packages/connect/src/api/wipeDevice.ts#L21

found while debugging tests in #19900 which are still failing :( (on something else) 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix-method-info&lookback=7)


mroz22:
hmm no, I am planning more changes to this.

PR #20672, opened 2025-08-08
fix(e2e): expect correct device status selector

Lemonexe:
## Description

Reason:

`@menu/switch-device` is the container, and it contains either `@deviceStatus-connected` or `@deviceStatus-disconnected`

But depending on timing, when you load suite it most usually **first** shows Disconnected, then Connected.
So if you await the container and expect it to contain Connected, you'll be disappointed because the element is found, but at the time, it contains different text (for a fraction of second, but it fails the test).

:point_right: Need to await the Connected element.

Also, there may be multiple devices with remembered wallets. That's why I use `.first()` because `toBeVisible` is strict and fails otherwise ( :thought_balloon: could it happen that the first one is _not_ visible but the 2nd one is, and we should expect that one? Idk, this works)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/e2e-device-status-assertions&lookback=7)

### DIFF
--- a/packages/suite-desktop-core/e2e/tests/suite/initial-run.test.ts
+++ b/packages/suite-desktop-core/e2e/tests/suite/initial-run.test.ts
@@ -40,11 +40,12 @@ test.describe('Suite initial run', { tag: ['@group=suite'] }, () => {
 
     test('Once user passed trough, skips initial run and shows connect-device modal', async ({
         page,
-        dashboardPage,
         onboardingPage,
     }) => {
         await onboardingPage.completeOnboarding();
         await page.reload();
-        await expect(dashboardPage.deviceSwitchingOpenButton).toContainText('Connected');
+        await expect(page.getByTestId('@deviceStatus-connected').first()).toBeVisible({
+            timeout: 30_000,
+        });
     });
 });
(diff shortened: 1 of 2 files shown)


packages/suite-desktop-core/e2e/tests/suite/initial-run.test.ts
```diff
@@ -40,11 +40,12 @@ test.describe('Suite initial run', { tag: ['@group=suite'] }, () => {
...
     test('Once user passed trough, skips initial run and shows connect-device modal', async ({
         page,
-        dashboardPage,
         onboardingPage,
     }) => {
         await onboardingPage.completeOnboarding();
         await page.reload();
-        await expect(dashboardPage.deviceSwitchingOpenButton).toContainText('Connected');
+        await expect(page.getByTestId('@deviceStatus-connected').first()).toBeVisible({
+            timeout: 30_000,
```
Lemonexe, line 48: both timeouts 30_000 are because this is the first locator after a page reload, and that takes a long time

HajekOndrej (approved):

PR #20709, opened 2025-08-12
chore(suite-native): auto eject e2e tests

### LINKED ISSUE #20683: E2E tests for view only by default

juriczech:
## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20683

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/eject-wallets-e2e&lookback=7)

### DIFF
--- a/suite-native/module-settings/src/components/EjectWallets/WalletRememberModeIconButton.tsx
+++ b/suite-native/module-settings/src/components/EjectWallets/WalletRememberModeIconButton.tsx
@@ -58,6 +58,7 @@ export const WalletRememberModeIconButton = ({ device }: { device: TrezorDevice 
                 onPress={handleEjectWallet}
                 colorScheme="tertiaryElevation1"
                 size="extraSmall"
+                testID={`@settings/eject-single-wallet/${device.features?.label}/${device.walletNumber ?? 0}`}
             />
         </Animated.View>
     );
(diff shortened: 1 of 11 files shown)


suite-native/module-settings/src/components/EjectWallets/WalletRememberModeIconButton.tsx
```diff
@@ -58,6 +58,7 @@ export const WalletRememberModeIconButton = ({ device }: { device: TrezorDevice
                 onPress={handleEjectWallet}
                 colorScheme="tertiaryElevation1"
                 size="extraSmall"
+                testID={`@settings/eject-single-wallet/${device.features?.label}/${device.walletNumber ?? 0}`}
```
matejkriz: Please don't use device label, better to use staticSessionId. Change it ideally in WalletRememberModeIconButton too. Device label is user input and can be sensitive.

PeKne (approved):
Apart from minor code style discrepancies (discussed already), I see no issue 👍.

suite-native/module-settings/src/components/EjectWallets/WalletRememberModeIconButton.tsx
juriczech: That's how it was even before this PR, but to the point - this wouldn't be exposed unless the user runs E2E tests which is not possible unless he would build an e2e tests android build and there would not be any of the users sensitive data. Other than that we just run this on the CI with mocked wallets.

Using staticSessionId would require more work as I would need to mock the staticSessionId for the device so that I have it available and can use it for selecting correct wallet for targeting. 

IMO, *if* the sensitivity of label is a concern, easiest solution will be to just keep the testId as `@settings/eject-single-wallet` as I test this with just one device without passphrase for now and we don't have to target specific wallet since there will be just one. I just left it as is in case we extend these tests in the future.

juriczech: Fixed and changed the testId to `@settings/eject-single-wallet`

PR #20716, opened 2025-08-12
fix(suite): force 1sat/vb for RBF

### LINKED ISSUE #20710: Force 1sat/vb for RBF
Lot of nodes does accept `minrelaytxfee` as `0.1 sat/vb`. But they forgot to lower the incremental fee (used for RBF).

Bcs of that we need to force `1 sat/vb` for RBF transactions, otherwise the RBF tx would not be propagated and it would be required to do the RBF of the former tx manually.

OriginalEveres:
Forces 1 sat/vb for RFB.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20710

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/btc-rbf-for-min-1sat-per-vb&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/btc-rbf-for-min-1sat-per-vb&lookback=7)

### DIFF
--- a/packages/suite/src/hooks/wallet/useRbfForm.ts
+++ b/packages/suite/src/hooks/wallet/useRbfForm.ts
@@ -33,6 +33,8 @@ import { useCompose } from './form/useCompose';
 import { useFees } from './form/useFees';
 import { useBitcoinAmountUnit } from './useBitcoinAmountUnit';
 
+const MIN_FEE_RATE = 1; // minimum fee rate in sat/vB, introduced because nodes lowered min relay tx fee, but not incremental fee
+
 export type UseRbfProps = {
     selectedAccount: SelectedAccountLoaded;
     rbfParams: RbfTransactionParams;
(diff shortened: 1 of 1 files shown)


packages/suite/src/hooks/wallet/useRbfForm.ts
```diff
@@ -54,7 +56,7 @@ const getBitcoinFeeInfo = (info: FeeInfo, rbfParams: RbfTransactionParamsBitcoin
     return {
         ...feeInfo,
         levels,
-        minFee: new BigNumber(feeRate).plus(feeInfo.minFee).toNumber(), // increase required minFee rate
+        minFee: Math.max(new BigNumber(feeRate).plus(feeInfo.minFee).toNumber(), MIN_FEE_RATE),
```
Lemonexe, line 59: Nit: you could use `BigNumber.maximum` and then cast the final bignumber toNumber() but can be done later..

Lemonexe (approved):
LGTM.
vorks.
image

packages/suite/src/hooks/wallet/useRbfForm.ts
OriginalEveres: good to know, will improve later on

PR #20717, opened 2025-08-12
chore(suite-native): simplify PassphraseStackNavigator

mroz22:
iz diz same? @vojtatranta 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=suite-native-passphrase-simplification&lookback=7)

### DIFF
--- a/suite-native/module-authorize-device/src/navigation/PassphraseStackNavigator.tsx
+++ b/suite-native/module-authorize-device/src/navigation/PassphraseStackNavigator.tsx
@@ -96,10 +96,18 @@ export const PassphraseStackNavigator = () => {
             )}
 
             {passphraseState === 'confirm-empty-passphrase' && (
-                <PassphraseStack.Screen
-                    name={AuthorizeDeviceStackRoutes.PassphraseEmptyWallet}
-                    component={PassphraseEmptyWalletScreen}
-                />
+                <>
+                    <PassphraseStack.Screen
+                        name={AuthorizeDeviceStackRoutes.PassphraseEmptyWallet}
+                        component={PassphraseEmptyWalletScreen}
+                    />
+                    {/* The PassphraseVerifyEmptyWallet screen is shown when user confirms they want to use an empty passphrase */}
+
+                    <PassphraseStack.Screen
+                        name={AuthorizeDeviceStackRoutes.PassphraseVerifyEmptyWallet}
+                        component={PassphraseVerifyEmptyWalletScreen}
+                    />
+                </>
             )}
 
             {passphraseState === 'passphrase-mismatch' && (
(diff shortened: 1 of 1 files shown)


vojtatranta:
I hope you made it the same.

The flow is slightly different on native, so I created a helper function for it.

suite-native/module-authorize-device/src/navigation/PassphraseStackNavigator.tsx
```diff
@@ -40,14 +40,6 @@ const determineNativePassphraseFlowState = (
...
     }
 
-    if (discovery.status === 'failed') {
-        return 'passphrase-redirecting';
-    }
-
-    if (discovery.status === 'complete') {
-        return 'passphrase-complete';
-    }
-
```
mroz22, line 50: these were never rendered but maybe it was intented render the "catch all route" for these cases?

mroz22:
I did split it into commits to make the changes easier to review. 
Also I added the [8a91555](https://github.com/trezor/trezor-suite/pull/20717/commits/8a915554703383dc63b8bdd75355284da030c94c) which I believe is what we want - router should only render one view, shoudln't it? In the previous implementation, it was not guaranteed on the level of rendering condition that there won't be two screens rendered at the same time

and linter found that there were indeed two screens rendered before https://github.com/trezor/trezor-suite/actions/runs/16934168294/job/47986831877?pr=20717#step:6:11  which of them was the intended one @vojtatranta ?

vojtatranta:
> I did split it into commits to make the changes easier to review. Also I added the [8a91555](https://github.com/trezor/trezor-suite/pull/20717/commits/8a915554703383dc63b8bdd75355284da030c94c) which I believe is what we want - router should only render one view, shoudln't it? In the previous implementation, it was not guaranteed on the level of rendering condition that there won't be two screens rendered at the same time
> 
> and linter found that there were indeed two screens rendered before https://github.com/trezor/trezor-suite/actions/runs/16934168294/job/47986831877?pr=20717#step:6:11 which of them was the intended one @vojtatranta ?

no, router doesnt need to have only one. IT can render multiple screens behind each other and the displayed one is one first in the order.

There are other stacks that render in the same way. I am not sure if there are some effetive differences when only was is rendered. Obviously, it can lose its state. BUt I guess that stack handles the rendering properly on its own. That's what it is there for.

mroz22:
what do you guys think? may I rebase and maybe squash it into a single commit? @matejkriz

suite-native/module-authorize-device/src/navigation/PassphraseStackNavigator.tsx
mroz22, line 50: these are not rendered at all but the fact is that these statement could have resulted into this component rendering nothing. was this the real intention? cc @vojtatranta

vojtatranta: You have to redirect from the component. So if it is empty, it throws an error. You shouldn't get to that state.

matejkriz (approved):
shipit

PR #20728, opened 2025-08-13
fix(suite): don't redirect back to provider when not necessary

### LINKED ISSUE #20670: Returning from sell partner causes infinite looping and redirect
**Describe the bug**
Returning from a sell partner results in infinite looping and redirecting back to the partner page instead of staying to the Suite.

https://github.com/user-attachments/assets/cc748476-84d6-4ae3-84b1-228dd200a31b

**Info:**
 - Suite version: web 25.9.0 (9791cc749df683648dcbf19bdc99de5e10e99301)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 2560x1440
 - Device: Trezor T3T1 2.9.0 universal (revision a09e6f04d4204607d821d9b742dcf7ffe9c4e2f5)
 - Transport: WebUsbTransport

matusbalascak:
## Description

This PR fixes an issue where user with a lot of tokens would be stuck in infinite redirect loop between the Suite and provider when selling tokens.

### Root cause 

`useDebounce` would be called after the redirect back from the provider while `sellInfo` was still `undefined` and therefore `trade` request was not fired so the flow stopped [here](https://github.com/trezor/trezor-suite/blob/d17806cd044adf12476fb410745860461ca9574f/packages/suite/src/hooks/wallet/trading/form/useTradingSellForm.ts#L274-L276) because `commonFunctions` availability depends on availability of `sellInfo`.

This however happened only for users with not many assets.

For users with a lot of different assets `sellInfo` managed to load before it got to the `useDebounce` that made the flow continue and `trade` request was fired with outdated data causing the redirect.

### Result

Now, `doSellTrade` is fired only when user submits the form. After the redirect from the provider back to the app, we don't call `doSellTrade` again and instead just use already setup subscription to the trade changes via `useTradingWatchTrade`

## Related Issue

Resolve #20670

---

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/sell-infinite-redirect&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/sell-infinite-redirect&lookback=7)

### DIFF
--- a/packages/suite/src/hooks/wallet/trading/form/useTradingSellForm.ts
+++ b/packages/suite/src/hooks/wallet/trading/form/useTradingSellForm.ts
@@ -384,6 +384,14 @@ export const useTradingSellForm = ({
             });
 
             navigateToSellConfirm();
+
+            // empty quoteId means the partner requests login first, requestTrade to get login screen
+            if (
+                (sellInfo && sellUtils.needToRegisterOrVerifyBankAccount({ quote, sellInfo })) ||
+                !quote.quoteId
+            ) {
+                doSellTrade(quote);
+            }
         };
 
         const onCancel = () => {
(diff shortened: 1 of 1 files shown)


adderpositive (approved):
Looks good. GJ

FreeWall (approved):
LGTM

PR #20732, opened 2025-08-13
native send form can pre-fill invalid feePerUnit decimals

### LINKED ISSUE #20734: Its possible to have 3 decimals in fee when only 2 is allowed
**Describe the bug**
Sometimes fee can be automatically set to 3 decimals instead of 2 and user can't proceed with sending the transactions.

**How to reproduce**

Don't have any 100% steps unfortunately.

**Screenshots**

https://github.com/user-attachments/assets/794c10e5-a108-4e4b-ba56-6c41b3caaf78

vytick:
We didn't previously make sure we pre-fill fee data with correct max amount of decimals than what we later validate. It caused issues at least with Litecoin, where we pre-filed 3 decimals, but fee form validation rejected this because 2 decimals were expected. Unless user set custom fee with less decimals, he then wasn't allowed to send LTC

This fix introduces sanitising for the form before we insert pre-filed data. 

https://github.com/user-attachments/assets/0f355802-a47e-4a72-b8b5-6d8de28e9bfe

Resolve #20734

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/send-fees-form-decimal-validation&lookback=7)

### DIFF
--- a/suite-native/module-send/src/sendFeesFormSchema.ts
+++ b/suite-native/module-send/src/sendFeesFormSchema.ts
@@ -1,13 +1,14 @@
 import { yup } from '@suite-common/validators';
-import { NetworkType } from '@suite-common/wallet-config';
+import { NetworkSymbol, getNetwork } from '@suite-common/wallet-config';
 import { FeeInfo } from '@suite-common/wallet-types';
 import { isDecimalsValid } from '@suite-common/wallet-utils';
 import { BigNumber } from '@trezor/utils';
 
 import { NativeSupportedFeeLevel } from './types';
+import { getFeeDecimals } from './utils';
 
 type SendFeesFormContext = {
-    networkType?: NetworkType;
+    symbol?: NetworkSymbol;
     networkFeeInfo?: FeeInfo;
     minimalFeeLimit?: string;
 };
@@ -30,17 +31,17 @@ export const sendFeesFormValidationSchema = yup.object({
             (value, { options: { context } }: yup.TestContext<SendFeesFormContext>) => {
                 if (!value) return true;
 
-                const { networkFeeInfo, networkType } = context!;
+                const { networkFeeInfo, symbol } = context!;
+
+                if (!symbol) return false;
+
+                const { networkType } = getNetwork(symbol);
+
                 if (!networkFeeInfo || !networkType) return false;
 
-                if (networkType !== 'bitcoin' && networkType !== 'ethereum') return true;
+                const maxDecimals = getFeeDecimals({ symbol });
 
-                let maxDecimals = 0;
-                if (networkType === 'bitcoin') {
-                    maxDecimals = 2;
-                } else if (networkType === 'ethereum') {
-                    maxDecimals = 9;
-                }
+                if (maxDecimals === null) return true;
 
                 return isDecimalsValid(value, maxDecimals);
             },
@@ -80,7 +81,10 @@ export const sendFeesFormValidationSchema = yup.object({
             'fee-limit-too-low',
             'Value is too low.',
             (value, { options: { context } }: yup.TestContext<SendFeesFormContext>) => {
-                const { networkType, minimalFeeLimit } = context!;
+                const { symbol, minimalFeeLimit } = context!;
+                if (!symbol) return true;
+
+                const { networkType } = getNetwork(symbol);
 
                 // Fee limit is used only for Ethereum, pass this validation for other networks.
                 if (networkType !== 'ethereum') return true;
--- a/suite-native/module-send/src/utils.ts
+++ b/suite-native/module-send/src/utils.ts
@@ -1,6 +1,8 @@
+import { NetworkSymbol, getNetwork } from '@suite-common/wallet-config';
 import { FormState, TokenAddress } from '@suite-common/wallet-types';
 import { Utxo } from '@trezor/blockchain-link-types';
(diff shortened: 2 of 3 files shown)


suite-native/module-send/src/utils.ts
```diff
@@ -40,3 +42,65 @@ export const constructFormDraft = ({
...
+        default:
+            return null;
+    }
+};
+
+const getValueAsString = (value: string | BigNumber) =>
+    typeof value === 'string' ? value : value.toString();
+
+const getValueAsBigNumber = (value: string | BigNumber) =>
+    typeof typeof value === 'string' ? new BigNumber(value) : (value as BigNumber);
```
PeKne: duplicated typeof

PeKne: Also why is the `as BigNumber` needed? could not we just call BigNumber(value) and remove this util function completely?

suite-native/module-send/src/utils.ts
```diff
@@ -40,3 +42,65 @@ export const constructFormDraft = ({
 
 export const isSameUtxo = (utxo1: Utxo, utxo2: Utxo): boolean =>
     utxo1.txid === utxo2.txid && utxo1.vout === utxo2.vout;
+
+export const getFeeDecimals = ({ symbol }: { symbol: NetworkSymbol }) => {
+    const network = getNetwork(symbol);
+
+    switch (network.networkType) {
+        case 'ethereum': {
+            return hasNetworkSettlementLayer(symbol) ? 4 : 2;
```
PeKne: there was `9` originally. Is it intentional change?

suite-native/module-send/src/sendFeesFormSchema.ts
```diff
@@ -30,17 +31,19 @@ export const sendFeesFormValidationSchema = yup.object({
...
 
-                let maxDecimals = 0;
-                if (networkType === 'bitcoin') {
-                    maxDecimals = 2;
-                } else if (networkType === 'ethereum') {
-                    maxDecimals = 9;
-                }
+                const maxDecimals = getFeeDecimals({ symbol });
+
+                if (!maxDecimals) return true;
```
PeKne: nit: this if statement should be never true, few lines above there is a check that ensures that

``` 
if (networkType !== 'bitcoin' && networkType !== 'ethereum') return true;

```

Feel free to resolve this comment without further change.

suite-native/module-send/src/utils.ts
```diff
@@ -40,3 +42,65 @@ export const constructFormDraft = ({
...
+            const multiplier = Math.pow(10, decimals);
+            const value = Math.ceil(Number(feeRate) * multiplier) / multiplier;
+
+            return value.toFixed(decimals);
+        }
+
+        case 'bitcoin': {
+            const feeBn = getValueAsBigNumber(feeRate);
+
+            return feeBn.toFixed(decimals);
```
PeKne:
```suggestion
            return String(feeRate)
```
should be enough.

suite-native/module-send/src/utils.ts
```diff
@@ -40,3 +42,65 @@ export const constructFormDraft = ({
...
+            return 2;
+        }
+
+        default:
+            return null;
+    }
+};
+
+const getValueAsString = (value: string | BigNumber) =>
+    typeof value === 'string' ? value : value.toString();
```
PeKne: this util is IMO unnecessary. String(value) should do the trick.

PeKne (approved):
Minor code style issues.

suite-native/module-send/src/utils.ts
vytick: This is what they have in desktop and I took from [here](https://github.com/trezor/trezor-suite/tree/develop/packages/product-components/src/components/FeeRate/FeeRate.tsx#L18)

suite-native/module-send/src/utils.ts
vytick: String() does  not trim decimals

suite-native/module-send/src/utils.ts
vytick: Looking into it once more and 2 and 4 digits are used for normal fees, not for custom fees. Reverting back to 9. Good call 👍

vytick: updated in 729c7b60bb

suite-native/module-send/src/sendFeesFormSchema.ts
vytick: changed in 729c7b60bb

suite-native/module-send/src/utils.ts
vytick: 729c7b60bb

suite-native/module-send/src/utils.ts
vytick: simplified in 729c7b60bb

suite-native/module-send/src/utils.ts
vytick: simplified in 729c7b60bb

PeKne (approved):

STew790:
**QA OK**
I am not able to reproduce this.

**Info**
25.8.1 (⁠[5be07bb](https://github.com/trezor/trezor-suite/commit/5be07bba057d4088e46bde3dc773019537ce3461))

PR #20775, opened 2025-08-14
test(connect): make it possible to run tests against model that has no …

mroz22:
…other but main release yet

@karliatto this is similar to what we talked about today but in tenv basically. It would be probably better to solve this on the backend side but I did it quickly like this.

## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=connect-tests-withouth-officially-unreleased-fw&lookback=7)


karliatto (approved):
Ok.

PR #20787, opened 2025-08-15
Fix/allow undefined when no fw updates

karliatto:
## Description

We want to allow device models with no FW release to be supported instead of throwing error, so if device model does not have firmware release config we ignore except for essential that is FW update.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/allow-undefined-when-no-fw-updates&lookback=7)

### DIFF
--- a/packages/connect/src/data/firmwareInfo.ts
+++ b/packages/connect/src/data/firmwareInfo.ts
@@ -149,7 +149,7 @@ export const getReleaseConfig = (features: Features, firmwareType: FirmwareType)
     }
     const deviceMessageRelease = firmwareReleaseConfig[internal_model];
     if (!deviceMessageRelease) {
-        throw new Error(`No firmware release config found for device model ${internal_model}`);
+        return;
     }
 
     return deviceMessageRelease[firmwareType];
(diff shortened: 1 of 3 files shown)


packages/connect/src/data/firmwareInfo.ts
```diff
@@ -149,7 +149,8 @@ export const getReleaseConfig = (features: Features, firmwareType: FirmwareType)
     }
     const deviceMessageRelease = firmwareReleaseConfig[internal_model];
     if (!deviceMessageRelease) {
-        throw new Error(`No firmware release config found for device model ${internal_model}`);
+        // throw new Error(`No firmware release config found for device model ${internal_model}`);
```
szymonlesisz: forgotten comment

szymonlesisz (approved):
works for me in `feat/thp-pairing` branch :+1:

packages/connect/src/data/firmwareInfo.ts
karliatto: ohh, thanks! fixed.

PR #20811, opened 2025-08-18
Try to fix roouting

### LINKED ISSUE #20669: Staking tab flashes 'staking is not available' when switching to account without staking
**Describe the bug**
If you are in the staking tab in one account and then switch to another account where staking is not available, the message 'staking is not available' briefly flashes on your screen.

This is similar to issue #20668, where you cannot traverse directly from one account's tokens tab to another account without tokens. In this case, the bug affects the staking tab instead of tokens.

https://github.com/user-attachments/assets/a40191e2-7153-4707-80e7-189f68b4910a

**Info:**
 - Suite version: desktop 25.8.1 (c9e29dcb0c730bd79daadcbbb7db11397298b273)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.8.1 Chrome/138.0.7204.97 Electron/37.2.0 Safari/537.36
 - OS: Linux x86_64
 - Screen: 2560x1440
 - Device: Trezor T3T1 2.9.0 universal (revision a09e6f04d4204607d821d9b742dcf7ffe9c4e2f5)
 - Transport: BridgeTransport 3.1.0
 - 
**How to reproduce**
1. Go to an account that has staking available.
2. Click on the staking tab.
3. Switch to another account where staking is not available.
4. Notice that 'staking is not available' flashes on your screen.

**Expected behavior**
You should be able to switch between accounts without visual glitches, and the staking tab should not display a flash message if staking is not available in the target account.

**QA instructions**
- check that tor works on web and desktop
- check that connect popup works
- check that the trading redirects works (if you buy btc via invity)

**Additional context**
Similar to issue #20668 but with the staking tab instead of tokens.

vojtatranta:
## Description

The problem of the flashing describe in the issue is likely due to the fact that the `navigate()` and `onLocationChange` are kinda out of sync. 

I changed the code so that there is one central history object, that is shared as the source for redux as well as target for react-router to listen for URL changes.

And while in that, the Main web/desktop were nearly the same, so I unified them.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20669#issuecomment-3191458319

## QA instructions

For both suite Web & Desktop, test:
- navigation within app
- trading flows with redirects 
- external links
- TOR links

For Suite Web, test:
- direct links to pages

## Screenshots:

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=trying-to-fix-routing&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=trying-to-fix-routing&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=trying-to-fix-routing&lookback=7)


package.json
```diff
@@ -150,5 +150,8 @@
             "built": true
         }
     },
-    "packageManager": "yarn@4.5.3"
+    "packageManager": "yarn@4.5.3",
+    "dependencies": {
+        "history": "^5.3.0"
```
vojtatranta: this is depedency of react-router, will put it where it is needed

packages/suite-web/src/MainWeb.tsx
```diff
@@ -26,29 +25,35 @@ import { useDebugLanguageShortcut, useFormattersConfig } from 'src/hooks/suite';
...
+import { Router } from 'react-router';
+import { createBrowserHistory } from 'history';
 
 const MainWeb = () => {
     usePlaywright();
     useTor();
     useDebugLanguageShortcut();
     useConnectPopupModals();
     const formattersConfig = useFormattersConfig();
+    const browserHistoryRef = useRef(createBrowserHistory());
```
vojtatranta: IMHO: this should be a parameter of the `MainWeb/Desktop`.

vojtatranta: As `<ReduxProvider>` renders this component, we can pass the `history` object to the store as `extra` argument. 

And `navigate()` can be then just `history.push()`. This is how we can get rid of that nasty `navigationService`.

packages/suite-desktop-ui/src/MainDesktop.tsx
```diff
@@ -43,8 +43,10 @@ import { TorLoadingScreen } from './support/screens/TorLoadingScreen';
 import { ResponsiveContextProvider } from 'src/support/suite/ResponsiveContext';
 import { BioAuthGuard } from '../../suite/src/components/suite/BioAuthGuard/BioAuthGuard';
 import { desktopComponents } from './support/desktopComponents';
+import type { History } from 'history';
+import { extraDependencies } from 'src/support/extraDependencies';
 
-const MainDesktop = () => {
+const MainDesktop = ({ history }: { history: History }) => {
     useTor();
```
vojtatranta, line 40: seems like MainDestkpo and MainWeb are now almost identical...

packages/suite-desktop-ui/src/MainDesktop.tsx
```diff
@@ -96,7 +98,7 @@ export const init = async (container: HTMLElement) => {
 
     const preloadAction = await preloadStore();
     const { statePatch } = await desktopApi.handshake();
-    const store = initStore(preloadAction, statePatch);
+    const store = initStore(preloadAction, { statePatch });
```
Lemonexe: :thought_balloon: Currently you define
`routerServices: createRouterServices(browserHistory)`
for Web, and here nothing, instead a default
`routerServices: createRouterServices(createMemoryHistory())` 
is in `extraDependencies.ts` with effective only for Desktop, because for Web it's overriden.
Wouldn't it be clearer if both MainDesktop and MainWeb had their own explicit declaration, and nothing or a mock would be in the default suite `extraDependencies`?

It also adds complexity that in Web, the same browserHistory is passed to `extras` and to `MainWeb history`, but here the flow is reveresed, instead _taking_ from `extras` for `MainDesktop history`

vojtatranta: Don't think it adds complexity. 

But okay, if you wish so. Nothing basically changes. Desktop uses memory history.

packages/suite-desktop-ui/src/MainDesktop.tsx
Lemonexe: That's a good thing! :slightly_smiling_face:

packages/suite-desktop-ui/src/MainDesktop.tsx
Lemonexe: Yeah, I meant just for consistency :slightly_smiling_face:

vojtatranta: key key

packages/suite/src/actions/suite/routerActions.ts
```diff
@@ -87,10 +88,10 @@ export const onAnchorChange = (anchor?: AnchorType) => (dispatch: Dispatch, _get
  * Dispatch initial url
  * Called from `@suite-middlewares/suiteMiddleware`
  */
-export const init = () => (dispatch: Dispatch, getState: GetState) => {
+export const init = () => (dispatch: Dispatch, getState: GetState, extra: ExtraDependencies) => {
```
Lemonexe, line 91: nit: since you're changing it, I realized this is actually a thunk, you can probably use `createThunk` to take care of typing.

packages/suite/src/actions/suite/routerActions.ts
```diff
@@ -130,26 +131,27 @@ export const goto =
...
-        const location = getLocation();
+        const { location } = extra.routerServices.history;
         const newUrl = `${urlBase}${preserveParams ? location.hash : ''}`;
-        dispatch(onLocationChange(newUrl, anchor));
 
         const route = findRouteByName(routeName);
 
         if (route?.isForegroundApp) {
+            // NOTE: when we are in foreground apps, the "main" history doesn't have an effect, url needs to be switched manually
+            dispatch(onLocationChange(newUrl, anchor));
```
Lemonexe: Nice, so the `onLocationChange` is dispatched automatically in RouterHandler.
But the previous implementation did not work I guess, which is why the `dispatch(onLocationChange` was called even outside the if block, to fix it?.
Now it's only relevant for foreground apps, because those do not fire RouterHandler. 
Do I understand this right?

packages/suite/src/actions/wallet/selectedAccountActions.ts
```diff
@@ -152,69 +153,49 @@ export const getAccountState = (state: AppState): SelectedAccountStatus => {
...
+                'visible',
+                'utxo',
+                'status',
+                'syncing',
+            ],
+        });
+
+        if (stateChanged) {
+            dispatch(accountsActions.disposeAccount());
+            dispatch(accountsActions.updateSelectedAccount(newState));
```
Lemonexe: looking at `selectedAccountReducer`, it seems useless to me to dispose and then immeidately update. You can just update. 
What do you think?

packages/suite/src/support/suite/RouterHandler.tsx
```diff
@@ -1,38 +1,57 @@
...
-        setNavigate(navigate);
-    }, [navigate]);
+    useMount(() => {
+        handleLocationChange();
+    });
 
-    // Make location available globally (useful in actions)
     useEffect(() => {
-        setLocation(location);
-    }, [location]);
```
Lemonexe, line 35: good riddance. Together with the `navigationService`
I approved that previous PR because I did not see any easier way :smile:

Lemonexe (commented):
Great work!!
Tested with cache disabled, and after initial load I throttled network connection.
Now loading lazy-loaded pages in in sync with redux,  :tada: 

Generally LGTM, LMK when there is the final version :slightly_smiling_face:

packages/suite-desktop-ui/src/MainDesktop.tsx
vojtatranta: so i guess I can parametrize it.

packages/suite/src/actions/suite/routerActions.ts
vojtatranta: yes

There is also a big change in the sencet that route state and navigate are update in different router.

Previously, first was router state updated and then the router navigate was applied. Currently, the first is changed the UI and then the router is updated.
I think this is correct. The syncing is done witht that routeHandler.
Although I think this is also incorrect. This syncing should be outside of React environment. I think these handlers should be hanged in like `useRef((() => { / *  here */  return unsubscribe}))` or it should be like in some `initAction()`. The reason is that not to way for the react updates / callbacks for `useEffect` so that we know for sure that when `history.listen` is fired, the UI and redux states are updated "in sync".

packages/suite/src/actions/wallet/selectedAccountActions.ts
vojtatranta: I don't know what was that for

packages/suite/src/support/suite/RouterHandler.tsx
vojtatranta: 😱

packages/suite/src/actions/wallet/selectedAccountActions.ts
Lemonexe: ok, then I would delete `dispatch(disposeAccount` leave only the `dispatch(updateSelectedAccount`

vojtatranta: I know why this is. This is useful when you go to dashbaord, then there should be no account selected.

But like, there should be a different solution for it, IMHO.

vojtatranta:
I have absolutely no idae, why this test fails https://github.com/trezor/trezor-suite/actions/runs/17150706367/job/48656609301?pr=20811

I have no clue at all. It fails even locally when run on electron. When switched to develop, it doesn't fail at all. So the problem is in the code for sure.

But I really have no idea why.
This is digged up somwhere, I was checking the stake middlew are and the component and I realyl don't know what is going on.

vojtatranta:
I am guessing like there is no some refresh or something so that the value is visible imeddaitly and so it takes the older value as result.

vojtatranta:
Fixed E2E tests: the problem apparantly was that the switching account didn't work peoprly. I removed it from middleware, but there are apparently other states when it should be called.

packages/suite/src/views/wallet/trading/redirect/TradingRedirect.tsx
```diff
@@ -29,13 +30,13 @@ export const TradingRedirect = () => {
...
     } = useTradingRedirect();
-    const router = useSelector(state => state.router);
+    const location = useLocation();
 
     const dispatch = useDispatch();
 
     useEffect(() => {
         // get rid of parameters appended by some partners to url which we pass to them
-        const params = router?.hash?.split('?')[0].split('/');
+        const params = location?.hash?.replace('#', '').split('?')[0].split('/');
```
vojtatranta: the location isn't yet propagated to the router properly

packages/suite-desktop-ui/src/MainDesktop.tsx
vojtatranta: did it, hopefuyll it works

packages/suite/src/support/suite/RouterHandler.tsx
```diff
@@ -1,51 +1,27 @@
...
-
-export const RouterHandler = () => {
-    const routerLoaded = useSelector(state => state.router.loaded);
+export const RouterHandler = memo(({ history }: { history: History }) => {
     const dispatch = useDispatch();
 
-    const location = useLocation();
-    const navigate = useNavigate();
-
-    useDidUpdate(() => {
```
vojtatranta, line 18: ha, forgot this...

vojtatranta: So, given this dumb duality of react-reoute + custom routing, i cannot jsut do `history.listen()` coz I also call the `history.push` manually.

So, I guess I will need to keep this.

What this does is that it listens basically to the history when it changes it calls that `onLocationChange` which only updatees the router state. It doesnt actually "go to" coz thats what react-router does.

vojtatranta: But like what does this do exactly? location is changed als owhen the history changes... Does this even work peorply or is it an accident?

vojtatranta: Aha, heh, it doesn't work. `onLocationChanged ` is called twice, from the `goto` and from here as well. hehe.

So history.listen is okay

vojtatranta:
NOTE @vojtatranta check the rebased main web / desktop for the different code

packages/suite/src/support/suite/RouterHandler.tsx
```diff
@@ -1,51 +1,35 @@
...
-        // Let router to be initialized properly
-        if (routerLoaded) {
```
vojtatranta, line 20: I am not sure if this is still needed or not, isn't it hte intial redirection?

packages/suite/src/support/suite/RouterHandler.tsx
```diff
@@ -1,51 +1,35 @@
...
-            // Handle browser navigation (back button)
```
vojtatranta, line 21: back button is popstate unalded lower in the code?

vojtatranta:
https://github.com/coderabbitai review, please make sure that the main refactoring is the Main web and desktop files is correct

vojtatranta:
@coderabbitai review

packages/suite/src/support/suite/Main.tsx
```diff
@@ -0,0 +1,52 @@
...
+                                </FormatterProvider>
+                            </ConnectedIntlProvider>
+                        </ErrorBoundary>
+                    </ResponsiveContextProvider>
+                </Router>
+            </ConnectedThemeProvider>
+        </HelmetProvider>
+        // </StrictMode>
+    );
+};
```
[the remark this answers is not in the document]
vojtatranta: Are you sure that this component is properly reused as before there were separete maindesktop and mainweb

packages/suite/src/support/extraDependencies.ts
```diff
@@ -56,6 +59,11 @@ const connectInitSettings = {
     firmwareHashCheckTimeouts: FW_HASH_CHECK_DEFAULT_TIMEOUTS,
 };
 
+export const createRouterServices = (history: History) => ({
+    history,
+    navigate: (to: To, options?: NavigateOptions) => history.push(to, options),
+});
```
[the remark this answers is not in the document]
vojtatranta: Are you sure about the API of teh options? I think tohse options should be ocmpletle matchin, right? why just state?

suite-common/redux-utils/src/extraDependenciesType.ts
```diff
@@ -1,8 +1,11 @@
+import type { NavigateOptions, To } from 'react-router';
```
Lemonexe: I'd be very careful with this – deps leaking into Suite Lite runtime has proven to be a problem in the past :disappointed: 
 I assume it's fine, and `import type` is removed during TS compilation (before bundling).
But are you sure about it?

suite-common/test-utils/src/extraDependenciesMock.ts
```diff
@@ -1,4 +1,7 @@
+import type { NavigateOptions } from 'react-router';
+
 import { createAction } from '@reduxjs/toolkit';
+import { To, createMemoryHistory } from 'history';
```
Lemonexe: Ad my comment in `suite-common/redux-utils/src/extraDependenciesType.ts`:
this will **definitely** leak into Suite Lite runtime, because the mocks are unfortunately used in the real Suite Lite [extras](https://github.com/trezor/trezor-suite/blob/ccc434d6a4567a94c4414dfe63acd01e4a8e6d76/suite-native/state/src/extraDependencies.ts#L32).
Can you go without it in test? i.e. leave only a plain object stub for history and () => void for navigate?

Or @trezor/suite-mobile is this OK?

packages/suite/src/support/suite/RouterHandler.tsx
```diff
@@ -1,51 +1,35 @@
...
-
-    // Make navigate available globally (useful in actions)
-    useEffect(() => {
-        setNavigate(navigate);
-    }, [navigate]);
-
-    // Make location available globally (useful in actions)
-    useEffect(() => {
-        setLocation(location);
-    }, [location]);
```
Lemonexe, line 35: good riddance :+1:

packages/suite/src/support/suite/Main.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export const Main = ({ history, children }: { history: History; children: React.ReactNode }) => {
+    useConnectPopupModals();
+    const formattersConfig = useFormattersConfig();
+
+    return (
+        // Todo: Enable when issues are fixed (ReactTruncate & BumpFee)
+        // <StrictMode>
+        <HelmetProvider>
+            <TrafficLightDraggableWindowHeader />
```
Lemonexe: This should be only in Desktop, not in Web

Lemonexe (approved):
Codewise mostly LGTM, some small things to fix.

Tested web & desktop rebased on develop, works **much better** :heavy_check_mark: :rocket:
(navigation smoother, deeplinks incl. hash still works)

I understand that `extras` are now used to hold the history state, and love the idea! Much better than the awkward `navigationService`, redux extra is IMO the proper place for such state to live  :+1: :heart: 

But review from @trezor/suite-mobile needed concerning the extras in Lite.

suite-common/test-utils/src/extraDependenciesMock.ts
vojtatranta: Okay, I can write own type for it

packages/suite/src/support/suite/Main.tsx
vojtatranta: oh, yes, awesome, thanks

vojtatranta:
@mroz22 @peter-sanderson have a s econd to check it out?

suite-common/test-utils/src/extraDependenciesMock.ts
juriczech: Thanks for catching this @Lemonexe ! Yeah I agree that we shouldn't have this in suite-common ideally

vojtatranta: Jiří killed it again! GOAT

packages/suite-desktop-ui/src/MainDesktop.tsx
```diff
@@ -21,70 +18,41 @@ import {
...
-                                    </DesktopUpdater>
-                                </FormatterProvider>
-                            </ConnectedIntlProvider>
-                        </ErrorBoundary>
-                    </ResponsiveContextProvider>
-                </MemoryRouter>
-            </ConnectedThemeProvider>
-        </HelmetProvider>
-        // </StrictMode>
+        <Main history={history} trafficLightOffset={<TrafficLightDraggableWindowHeader />}>
```
vojtatranta, line 45: This is not brilliant, but the TrafficLight should be up there for desktkop only

packages/suite/src/support/suite/RouterHandler.tsx
vojtatranta: cc @Lemonexe any idea?

suite-common/redux-utils/src/extraDependenciesType.ts
```diff
@@ -3,6 +3,7 @@ import {
     ActionCreatorWithPreparedPayload,
     ActionCreatorWithoutPayload,
 } from '@reduxjs/toolkit';
+import type { History, Path } from 'history';
```
vojtatranta: This dep can be here or not? @Lemonexe I bet it shouldn't?

vojtatranta: If not, I'd have to make some "desktop dependencies"

vojtatranta: cc @Lemonexe

Lemonexe: I mean, `import type` _should_ be stripped by typeScript. At leat that's done by webpack TS loader, but I have no clue how it works in metro.

I really liked the idea of holding the `history` in redux for `suite`, but since it is partially shared in suite-common, maybe it's best to return back to basics:  something like it was in the deleted `navigationService.ts`.

where you could move `createRouterServices`. Instead of preparing the stuff for redux, you would set/get it in a module top-scope variable.
It's not pretty but it's very straightforward :slightly_smiling_face:  TBH the extras are anything but straightforward :smiling_face_with_tear:

packages/suite/src/support/suite/RouterHandler.tsx
Lemonexe: What was the purpose of this hook?
Seems to me that on any component update but the initial render, it sends the redux thunk.
So basically it's a reactivity to `location.pathname + location.hash`
So IDK, seems to me you have replaced it adequately with `history.listen` ...

packages/suite/src/support/suite/RouterHandler.tsx
```diff
@@ -1,51 +1,35 @@
...
+        () =>
+            history.listen(() =>
+                dispatch(onLocationChange(history.location.pathname + history.location.hash)),
+            ),
+        [dispatch, history],
+    );
 
     useEffect(() => {
         const onPopState = () => {
             const canGoBack = dispatch(onBeforePopState());
```
Lemonexe, line 21: 💭 unrelated rant: why the hell is `onBeforePopState` a thunk? It can easily be a selector 🤦

suite-common/redux-utils/src/extraDependenciesType.ts
vojtatranta: Hmmm.... So we'd have to have a different thunks . Those that are shared for RN and thos that are not. Those would have the extra type like `DesktopExtraDependencies` and the RN shared would have like `UniversalExtraDependencies` as extra type.

That woudl IMHO pretty good from the "architecture" perspective. But I get that people would be sad.

Making anohther navigation service isn't great.

What about making a custom type that would satisfiy the `History`?

packages/suite/src/support/suite/RouterHandler.tsx
vojtatranta: lol yeah, good point

suite-common/redux-utils/src/extraDependenciesType.ts
Lemonexe: Yeah I meant to take it out of redux extras, and put it back to a module variable, a new `navigationService`.
Special `DesktopExtraDependencies` would work, but it's further complicating an already complex thing.
If you can do a minimal type that satisfies `History` I think it's good enough, even if you might have to do some `as` typecasting in Suite.

IDK what's the best solution. I'd go with `navigationService` for simplest readability..

vojtatranta: But like, well, where is the n`goto` thunk? This thunk is completely desktop-only .
Hehe, what about other extra deps there?EE
```ts
 utils: {
        saveAs: (data: Blob, fileName: string) => void;
        connectInitSettings: ConnectInitSettings;
        reportSecurityCheck: (props: ReportSecurityCheckProps) => void;
    };
```
SaveAs is available on mobile? Or? 

So the problem isn't with the methods in `extra` but with some code leaking to the mobile...

Lemonexe: Sorry for misleading, I thought `goto` was one of the extras, but it's not.

vojtatranta: The cleanest and the best solution would be separateion of hte interface for the `Extra`.

`DesktopExtraDependencies` woudl extend `UniversalDependencies`

Eg. those `ExtraDependenciesType` be it `DesktopDependencies` would be only able to import in place where it makes sense - in desktop code (controlled by the location in packages). 

That would be cool like hell, coz Native could have their own `NativeExtraDependencies` taht would also extends `UniversalDependencies` and you couldn't import the `NativeExtraDeps...` anywhere else but in Native.

Yeah, I'd like that

vojtatranta: Those "common" code would then use the `UniversalExtraDependencies` all the time.

When creating a store on Mobile you'd pass your own implementation of `extra` on desktop you would pass different one.

And super cool, you could have liek `DesktopWebExtraDependnecies` that would be web-only. And surely like `ElectronDesktopExtraDependencies` could be the once correspoding with the stuf avaiaible in electron, like BT etc.

You would just swap the implementaion in `MainWeb/Desktop`

Lemonexe: Alright. Yeah the only downside I wrote is that it complicates it, will be a lot of diffs but why not.
Looking again at `extraDependencies`, I'm starting to like it, because there are **already** some `extraDependencies` that are never dispatched/called outside of `suite`:
- `storageLoadBioAuth`
- `saveAs`
For these two, the `ExtraDependencies` type & mock is useless in suite-common & suite-native.

→ so agreed, it'd be best to separate it :+1:

vojtatranta: I guess this could improve the architecture a lot, because it would force us to thing if the code is universal, desktop, web or web.

Anyways, won't do it today, will do a simple type for history.

Lemonexe: huh, now I see you already [created it](https://github.com/trezor/trezor-suite/blob/c00ef1febd08bd4fb5abd36da6b911c81ea3c4f6/packages/suite/src/types/suite/extraDependencies.ts) in June.
It's just not used yet.

vojtatranta: willl delete, unused

packages/suite/src/support/suite/RouterHandler.tsx
vojtatranta: I kinda guess why, because the code combines 2 states - router and modal.

suite-common/redux-utils/src/extraDependenciesType.ts
vojtatranta: @Lemonexe updated the types

suite-common/test-utils/src/extraDependenciesMock.ts
```diff
@@ -144,4 +149,12 @@ export const extraDependenciesMock: ExtraDependencies = {
...
             console.warn(`Mock reporting ${checkType} check ${level} to Sentry.`),
     },
+    routerServices: {
+        getLocation: () => ({
+            pathname: 'mocked_path',
+            hash: 'mocked_hash',
+        }),
+        navigate: (to: To, state?: LocationPushState) =>
+            console.warn(`Mock navigating to ${to} with state`, state),
+    },
```
Lemonexe, line 159: Now with c1dce68799845e8c264745114e4958ba205b40c1 it won't leak `history` into react-native.
This was imho the important case (definitely leaking), while the `extraDependenciesType` was only theoretically problem.

Lemonexe (approved):
LGTM! I made a fixup, now it does not affect mobile anymore → ready to be merged.

STew790:
**QA OK**
Tested here https://github.com/trezor/trezor-suite/issues/20669#issuecomment-3371020658

PR #20825, opened 2025-08-18
feat(suite-data): Add workflow for auto-updating TOR

5K1PY:
Added workflow file that creates a PR for updating TOR to new version if available.

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/update-tor-ci&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/update-tor-ci&lookback=7)

### DIFF
--- a/.github/workflows/update-tor.yml
+++ b/.github/workflows/update-tor.yml
@@ -1,0 +1,109 @@
+name: "[Bot] update TOR binary"
+
+on:
+  schedule:
+    - cron: "0 0 1 * *"
+  workflow_dispatch:
+
+jobs:
+  update-tor:
+    runs-on: ubuntu-latest
+    env:
+      UPDATE_FILE: packages/suite-data/files/bin/tor/update.sh
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Get current TOR versions
+        id: versions
+        run: |
+          content=$(cat $UPDATE_FILE)
+
+          newer_version () {
+            local version=$(echo "$content" | grep "^$1=" | cut -d= -f2)
+            local major=$(echo "$version" | cut -d_ -f1-2)_
+            local patch=$(echo "$version" | cut -d_ -f3)
+            local next=$((patch+1))
+            echo "${major}${next}"
+          }
+
+          crx_next=$(newer_version CRX_VER)
+          arm_next=$(newer_version CRX_LINUX_ARM_VER)
+
+          echo "crx_next=$crx_next" >> $GITHUB_OUTPUT
+          echo "arm_next=$arm_next" >> $GITHUB_OUTPUT
+
+      - name: Check if new CRX binary exists
+        id: check_crx
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/update-tor.yml
```diff
@@ -0,0 +1,110 @@
+name: Update TOR binaries
```
vdovhanych: Adhere to the other gh actions naming convention, this should most likely be `[Bot] update TOR binary"

.github/workflows/update-tor.yml
```diff
@@ -0,0 +1,110 @@
...
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
```
vdovhanych: why run it every night?

.github/workflows/update-tor.yml
```diff
@@ -0,0 +1,110 @@
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Get current TOR versions
```
vdovhanych, line 22:
```suggestion
      - name: Get current TOR versions
        id: versions
        run: |
          crx_next=$(grep "^CRX_VER=" $UPDATE_FILE | cut -d'=' -f2)
          arm_next=$(grep "^CRX_LINUX_ARM_VER=" $UPDATE_FILE | cut -d'=' -f2)
          
          echo "crx_next=$crx_next" >> $GITHUB_OUTPUT
          echo "arm_next=$arm_next" >> $GITHUB_OUTPUT
      
```
If we want to extract the version number, we can simply do it like this.

vdovhanych (commented):
I just reviewed a few things, but the whole process is unnecessarily overcomplicated if we want to update the TOR binary. Why do we check versions anyway? We can just download the latest every single time, print the binary version in the PR, and then merge it. We don't version the Tor binary anywhere in the code besides the download script.

5K1PY:
> I just reviewed a few things, but the whole process is unnecessarily overcomplicated if we want to update the TOR binary. Why do we check versions anyway? We can just download the latest every single time, print the binary version in the PR, and then merge it. We don't version the Tor binary anywhere in the code besides the download script.

I can make the workflow download the binaries, but we still need to check if there is newer version to construct the download url. At least I couldn't find one that always points to the latest version.

.github/workflows/update-tor.yml
5K1PY: Fixed 3144118

.github/workflows/update-tor.yml
5K1PY: Changed 3144118

.github/workflows/update-tor.yml
5K1PY: This doesn't increase the patch number by one. (Fair, non-descriptive function name by me.)

vdovhanych: Okay, I understand now. I don't see any versioning that would coincide with the CRX versions we use. Brave basically increments it with the Dynamo DB table upon uploading the built CRX package.

vdovhanych (approved):

PR #20842, opened 2025-08-19
Feat: update calculate firmware hash to support TS7

karliatto:
## Description

This is updating `calculateFirmwareHash` to work with TS7.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/update-calculateFirmwareHash-to-support-ts7&lookback=7)

### DIFF
--- a/packages/connect/src/api/firmware/calculateFirmwareHash.ts
+++ b/packages/connect/src/api/firmware/calculateFirmwareHash.ts
@@ -1,10 +1,35 @@
 import { blake2sHex } from 'blakejs';
 
-const SIZE_T1 = (7 * 128 + 64) * 1024;
-const SIZE_TT = 13 * 128 * 1024;
+import { DeviceModelInternal } from '@trezor/device-utils';
 
-export const calculateFirmwareHash = (major_version: number, fw: ArrayBuffer, key?: Buffer) => {
-    const size = major_version === 1 ? SIZE_T1 : SIZE_TT;
+// Size values are taken from
+// https://github.com/trezor/trezor-firmware/blob/56f9490c01b40e99d94fe5d8a283955800d86562/core/embed/models/T3T1/memory.h#L62
+// and the calculation reflects the internal memory layout.
+// We keep size values expanded even though the result is the same for all core models except T3W1.
+const SIZE_T1B1 = (7 * 128 + 64) * 1024; // 960 kB
+const SIZE_T2T1 = 13 * 128 * 1024; // 1664 kB
+const SIZE_T2B1 = 13 * 128 * 1024; // 1664 kB
+const SIZE_T3B1 = 208 * 8 * 1024; // 1664 kB
+const SIZE_T3T1 = 208 * 8 * 1024; // 1664 kB
+const SIZE_T3W1 = 417 * 8 * 1024; // 3336 kB
+
+const firmwareSizeMap: Partial<Record<DeviceModelInternal, number>> = {
+    [DeviceModelInternal.T1B1]: SIZE_T1B1,
+    [DeviceModelInternal.T2T1]: SIZE_T2T1,
+    [DeviceModelInternal.T2B1]: SIZE_T2B1,
+    [DeviceModelInternal.T3B1]: SIZE_T3B1,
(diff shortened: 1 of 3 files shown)


packages/connect/src/api/firmware/calculateFirmwareHash.ts
```diff
@@ -1,10 +1,24 @@
...
-export const calculateFirmwareHash = (major_version: number, fw: ArrayBuffer, key?: Buffer) => {
-    const size = major_version === 1 ? SIZE_T1 : SIZE_TT;
+export const calculateFirmwareHash = (
+    major_version: number,
+    internal_model: DeviceModelInternal,
+    fw: ArrayBuffer,
+    key?: Buffer,
+) => {
+    let size = SIZE_T1;
+    // We could only use `internal_model` but htere might be older T1 model without `internal_model`, so this is safer.
```
obrusvit: Nit: typo
```suggestion
    // We could only use `internal_model` but there might be older T1 model without `internal_model`, so this is safer.
```

packages/connect/src/api/firmware/calculateFirmwareHash.ts
```diff
@@ -1,10 +1,24 @@
 import { blake2sHex } from 'blakejs';
 
+import { DeviceModelInternal } from '@trezor/device-utils';
+
 const SIZE_T1 = (7 * 128 + 64) * 1024;
 const SIZE_TT = 13 * 128 * 1024;
+const SIZE_TS7 = (16 * 128 + 65) * 1024;
```
obrusvit:
```suggestion
const SIZE_T1B1 = (7 * 128 + 64) * 1024;  // 960 kB
const SIZE_T2T1 = 13 * 128 * 1024; // 1664 kB
const SIZE_T2B1 = 13 * 128 * 1024; // 1664 kB
const SIZE_T3B1 = 208 * 8 * 1024;  // 1664 kB
const SIZE_T3T1 = 208 * 8 * 1024;  // 1664 kB
const SIZE_T3W1 = 417 * 8 * 1024;  // 3336 kB
```

obrusvit (commented):
Suggested the correct sizes and also I suggest to distinguish according to the internal name. Distinction just by the version number won't work for TS7.

packages/connect/src/api/firmware/calculateFirmwareHash.ts
karliatto: I am just afraid that some old T1 do not report internal name `T1B1` so I would still keep the check when `major_version === 1` then use `SIZE_T1B1` as it was. WDYT?

obrusvit: Yes, it's sensible to do that for `T1`. For the core models, use the internal name.

Btw. for completeness, the values are taken e.g. from here https://github.com/trezor/trezor-firmware/blob/56f9490c01b40e99d94fe5d8a283955800d86562/core/embed/models/T3T1/memory.h#L62 and the calculation reflects the internal memory layout. That's why I kept it expanded even though the result is the same for all core models except T3W1

karliatto: I added part of your message as a comment to the code so we have a reference in the future.

https://github.com/trezor/trezor-suite/pull/20842/commits/34f9bf768f6799f3922c4c83deef02cc61c41ef7

packages/connect/src/api/firmware/calculateFirmwareHash.ts
karliatto: Thanks, I deleted the whole comment and replace it with https://github.com/trezor/trezor-suite/pull/20842/commits/34f9bf768f6799f3922c4c83deef02cc61c41ef7#diff-7bba71d1379cd1423163a57c49d92cfa1318a48dd0757a33313a78c6264fe662R30

packages/connect/src/api/firmware/calculateFirmwareHash.ts
obrusvit: LGTM

Just a small nit that not all firmware max sizes are taken from that file. Each model has its own memory.h file.

mroz22 (approved):

PR #20854, opened 2025-08-19
Chore: deduplicate screen names

matejkriz:
## Description

Deduplicate screen names. Reusing a screen name can cause incorrect navigation and makes analytics harder to interpret (it becomes almost impossible to distinguish between duplicated screens; the only workaround is to check them in combination with the previous screen).

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/deduplicate-screen-names&lookback=7)

### DIFF
--- a/suite-native/navigation/src/routes.ts
+++ b/suite-native/navigation/src/routes.ts
@@ -40,7 +40,7 @@ export enum OnboardingStackRoutes {
 }
 
 export enum DeviceOnboardingStackRoutes {
-    ConnectAndUnlockDevice = 'ConnectAndUnlockDevice',
+    ConnectAndUnlockDeviceOnboarding = 'ConnectAndUnlockDeviceOnboarding',
     UninitializedDeviceLanding = 'UninitializedDeviceLanding',
     SuspiciousDevice = 'SuspiciousDevice',
     SecurityCheck = 'SecurityCheck',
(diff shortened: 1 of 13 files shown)


suite-native/navigation/src/routes.ts
```diff
@@ -110,19 +110,19 @@ export enum DeviceAuthenticityStackRoutes {
...
 export enum DeviceNameStackRoutes {
     DeviceName = 'DeviceName',
-    ContinueOnTrezor = 'ContinueOnTrezor',
+    ContinueOnTrezor = 'ContinueOnTrezor', // FIXME
     DeviceNameLoadingScreen = 'DeviceNameLoadingScreen',
 }
 
 export enum AuthorizeDeviceStackRoutes {
-    ConnectAndUnlockDevice = 'ConnectAndUnlockDevice',
+    ConnectAndUnlockDeviceAuthorize = 'ConnectAndUnlockDeviceAuthorize',
```
juriczech, line 125: This seems like a better option to me. Alternatively, we could keep the key same as before and just change the actual string, which would enable us to minimize the amount of refactoring needed 🤔 Probably not the best approach though? 

I don't like the repeating of authorize name in the whole name though...
```suggestion
    AuthorizeConnectAndUnlock = 'AuthorizeConnectAndUnlock',
```

matejkriz: current proposal is to change only the value, not the key, and change it like: `AuthorizeDevice_ConnectAndUnlockDevice`

juriczech: Yeah, that sounds good to me!

Issue #20765, opened 2025-08-14
Bump mobile deps (2025.08)


PeKne:
## Developer checklist

- [x] run `yarn list-outdated` for your team name (or run this to list all team names)
- [x] one by one, update all listed outdated deps
    - [x] commit separately for easier reverting (individually or grouped by related packages)
    - [x] make sure the versions are consistent across all yarn workspaces
    - [x] run `yarn dedupe`, consider impact carefully
    - [x] pin exact versions when appropriate (use your best judgement)
- [x] check the source repository for code changes if feasible, or changelog _at the very least_
    - [x] carefully check any external code flagged by Socket Security
- [x] ensure TS and CI tests are ✅
- [x] besides CI tests, manually test affected parts of Trezor Suite, at least superficially

(See [notion page](https://www.notion.so/satoshilabs/Dependency-Management-1b5bf845aa1f4ca7b9d57ea9ccd3fe63) for more details.)

### Upgraded ✅:
 - https://github.com/trezor/trezor-suite/pull/20858 
    - upgrade of React navigation to 7.x.x ([changelog/breaking changes](https://reactnavigation.org/docs/upgrading-from-6.x/))
 -   https://github.com/trezor/trezor-suite/pull/20882
    - upgrade of all purely native dependencies that have no breaking changes and do not require new arch

### Won't upgrade in this PR 🚫:
- requires new architecture:
  - react-native-reanimated
  - react-native-mmkv
  - @shopify/flash-list
- requires new SDK (Expo SDK 54):
  - @react-native/metro-config
  - @react-native/babel-preset
  - react
  - react-native
  - react-dom

## QA instructions
Major version of react navigation was upgraded within this issue. Please go through the whole app and try to find some incosistency in navigation. In case that after any action you end up at some unexpected screen (eg. connecting screens instead of home screen) please report it.

STew790:
Will be definitely tested throughout the whole testing period.

STew790:
**QA OK**
There wasn't any issue with this.

PR #20882, opened 2025-08-20
chore(suite-native): bump mobile dependencies (2025.08)

PeKne:
## Description
This PR bumps all those mobile dependencies to a versions that are compatible to the old RN architecture (that we are still using). And do not have any breaking changes. More in linked issue.

The PR is distributed to multiple commits for easier revert in case of fuckup. 

## Related Issue

Related to #20765 

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/native/bump-mobile-deps-2025-08&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/native/bump-mobile-deps-2025-08&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/bump-mobile-deps-2025-08&lookback=7)


PeKne:
@SocketSecurity ignore @sentry/react-native@6.20.0

PeKne:
@SocketSecurity ignore react-native-screens@4.14.1

matejkriz:
If you remove all the `^` and `~` and use fixed versions instead, `yarn outdated` will report that fixed versions as Wanted even if there are newer versions, right? Will these dependencies still be reported by `yarn list-outdated`, or does this silence them forever?

matejkriz:
It would be nice to have `yarn dedupe` step as separate commit (next time).

suite-native/app/package.json
```diff
@@ -95,47 +95,48 @@
...
+        "lottie-react-native": "7.3.2",
         "node-libs-browser": "^2.2.1",
         "react": "19.0.0",
         "react-intl": "^7.1.11",
         "react-native": "0.79.3",
         "react-native-ble-plx": "3.5.0",
-        "react-native-edge-to-edge": "1.6.0",
-        "react-native-gesture-handler": "~2.24.0",
-        "react-native-keyboard-controller": "1.16.5",
+        "react-native-edge-to-edge": "1.6.2",
```
matejkriz, line 127: I see this warning in console:
` WARN  `setBackgroundColorAsync` is not supported with edge-to-edge enabled.`

And this white line at the bottom:
Image

Will double check on the develop. I'm not saying it's necessarily because of this update

matejkriz: It is OK on the develop so most probably caused by this update.

PeKne: Fixed by fixup a2cae79499. The problem was present already on develop, but minor version upgrade of [expo-navigation-bar](https://docs.expo.dev/versions/latest/sdk/navigation-bar/#methods) started printing this waring.

Problem was that `setBackgroundColorAsync()` is completely redundant call with edge-to-edge enabled, since there is no background.

I also found in [react-native-edge-to-edge](https://github.com/zoontek/react-native-edge-to-edge#systembars) documentation that it is not recommended use `NavigationBar` from `expo-navigation-bar`, but rather `SystemBars` from `react-native-edge-to-edge`. Also fixed in this fixup.

PeKne:
> If you remove all the `^` and `~` and use fixed versions instead, `yarn outdated` will report that fixed versions as Wanted even if there are newer versions, right? Will these dependencies still be reported by `yarn list-outdated`, or does this silence them forever?

@matejkriz `yarn outdated` does print out also outdated  fixed version, so this change should be no problem. I removed all the `^` and `~` symbols, because we ignore them while calling `yarn install`, am I right?

suite-native/app/package.json
PeKne: And it seems that it was the only usage of `expo-navigation-bar`, so this dependency was completely removed from our repo :tada: fixup d73c0dcf1b

matejkriz:
> > If you remove all the `^` and `~` and use fixed versions instead, `yarn outdated` will report that fixed versions as Wanted even if there are newer versions, right? Will these dependencies still be reported by `yarn list-outdated`, or does this silence them forever?
> 
> @matejkriz `yarn outdated` does print out also outdated fixed version, so this change should be no problem. I removed all the `^` and `~` symbols, because we ignore them while calling `yarn install`, am I right?

`yarn install` will always use the version fixed by yarn.lock. But yarn resolutions during `yarn add` and `yarn dedupe` will be affected. When the same dependency appears multiple times — especially if we list it directly in package.json but it’s also pulled in by another dependency — Yarn can install just one version as long as the version ranges overlap. But if we pin everything to exact versions, and one of those doesn’t fall into the sub-dependency’s range, Yarn will end up installing two separate versions.

We don't use the ranges consistently so I'm not against merging this, but I belive we should keep using the ranges where it makes sense.

matejkriz (approved):
👏

PR #20895, opened 2025-08-20
fix(suite-native): keep fiat value as null before loaded so skeleton is visible instead of zero

### LINKED ISSUE #20884: Assets show 0 instead of skeleton before rates load
**What**
When we load assets, we should be loading a skeleton to indicate the value loading is in progress. But now it shows zero instead of loader and then updates once we have a value to be shown.

**Info:**

- Mobile Suite Version: 25.8.1
- OS: android

**Steps to reproduce:**
1. Connect device and load assets
2. Disconnect 
3. Close and reopen the app
4. Assets will show 0 before loading rates and then changing value (should be skeleton instead)

**Expected behavior**
We should show the loading skeleton until we're certain that zero  is the correct value 

Image

https://github.com/user-attachments/assets/cd792cee-6fc2-4fff-884e-5927a32721f5

matejkriz:
## Description

To be honest, I don’t know how it used to work before, but this change is relatively isolated and makes sense to me, so hopefully it didn’t break something else in an unexpected place 🙈 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20884

## Screenshots:
Screenshot_2025-08-20-12-56-49-217_io trezor suite debug

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/null-fiat-value-skeleton&lookback=7)

### DIFF
--- a/suite-native/assets/src/assetsSelectors.ts
+++ b/suite-native/assets/src/assetsSelectors.ts
@@ -138,18 +138,24 @@ const selectDeviceAssetsWithBalances = createMemoizedSelector(
                 (sum, { cryptoValue }) => sum.plus(new BigNumber(cryptoValue)),
                 new BigNumber(0),
             );
-            const fiatBalance = networkAccounts.reduce(
-                (sum, { fiatValue }) => (fiatValue ? sum.plus(fiatValue) : sum),
-                new BigNumber(0),
-            );
+            const fiatBalance = networkAccounts.reduce<BigNumber | null>((sum, { fiatValue }) => {
+                // If any account has null fiat data, set the network fiat to null.
+                // This prevents showing partial/incomplete values to users - we show loading state until all data is available.
+                if (sum === null) return null;
+                if (fiatValue == null) return null;
 
-            totalFiatBalance = asBaseCurrencyAmount(totalFiatBalance.plus(fiatBalance));
+                return sum.plus(fiatValue);
+            }, new BigNumber(0));
+
+            if (fiatBalance) {
+                totalFiatBalance = asBaseCurrencyAmount(totalFiatBalance.plus(fiatBalance));
+            }
 
             const asset: AssetType = {
                 symbol,
                 // For assets we should always only 8 decimals to save space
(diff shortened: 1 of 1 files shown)


juriczech (approved):
LGTM

juriczech:
Quickly tested on android device, seems to be fixed 💪

suite-native/assets/src/assetsSelectors.ts
```diff
@@ -138,18 +138,22 @@ const selectDeviceAssetsWithBalances = createMemoizedSelector(
                 (sum, { cryptoValue }) => sum.plus(new BigNumber(cryptoValue)),
                 new BigNumber(0),
             );
-            const fiatBalance = networkAccounts.reduce(
-                (sum, { fiatValue }) => (fiatValue ? sum.plus(fiatValue) : sum),
-                new BigNumber(0),
-            );
+            const fiatBalance = networkAccounts.reduce<BigNumber | null>((sum, { fiatValue }) => {
+                if (sum === null) return null;
```
peter-sanderson, line 144: Do I understand that in case there is **at least one** account with `null` data, we want to make totalSum `null` as well as we assume that all accounts are not not loaded yet?

I would maybe add a comment here to explain.

peter-sanderson (approved):

suite-native/assets/src/assetsSelectors.ts
matejkriz: Yes, we don't want to present partial value to the user, we should present loading state until we have the final number. I'll add a comment that it is intentional.

matejkriz: [`15546b9` (#20895)](https://github.com/trezor/trezor-suite/pull/20895/commits/15546b9825076f9795d11bf7c65370fdb9dd517b#diff-1c8311e2d32ab0d88991a27b03c2e10cf05a7e684b127fe3f3691172d3154f85R142-R143)

PR #20913, opened 2025-08-20
chore(connect): unify handshake cancel timeouts

yanascz:
No need for different timeouts anymore because of https://github.com/trezor/trezor-suite/pull/19574.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/connect/handshake-timeout&lookback=7)

### DIFF
--- a/packages/connect/src/device/workflow/handshake.ts
+++ b/packages/connect/src/device/workflow/handshake.ts
@@ -3,13 +3,10 @@ import { TRANSPORT_ERROR } from '@trezor/transport';
 import { resolveAfter, versionUtils } from '@trezor/utils';
 
 import { TypedError } from '../../constants/errors';
-import { DataManager } from '../../data/DataManager';
 import { WorkflowContext } from '../../types/workflow';
 import { Log } from '../../utils/debug';
 
 const CANCEL_TIMEOUT = 1_000;
-// Due to performance issues in suite-native during app start, original timeout is not sufficient.
-const CANCEL_TIMEOUT_REACT_NATIVE = 20_000;
 const ATTEMPTS_LIMIT = 10;
 
 type Context = {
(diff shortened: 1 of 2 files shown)


packages/connect/src/device/workflow/handshake.ts
```diff
@@ -36,15 +33,12 @@ export const handshakeCancel = async ({ device, logger, signal }: Context) => {
         return;
     }
 
-    const isNative = DataManager.getSettings('env') === 'react-native';
-    const cancelTimeout = isNative ? CANCEL_TIMEOUT_REACT_NATIVE : CANCEL_TIMEOUT;
-
```
szymonlesisz, line 41: there is similar reference in `packages/connect/src/device/DeviceCurrentSession.ts`:

```
const getFeaturesTimeout = () =>
    // Due to performance issues in suite-native during app start, original timeout is not sufficient.
    DataManager.getSettings('env') === 'react-native' ? 20_000 : 3_000;
```

suite-native/state/src/extraDependencies.ts
```diff
@@ -57,7 +57,6 @@ export const extraDependencies: ExtraDependencies = mergeDeepObject(extraDepende
             lazyLoad: false,
             transportReconnect: false,
             debug: false,
-            env: 'react-native',
             popup: false,
```
szymonlesisz: i would keep it, connect may expect this in the future

packages/connect/src/device/workflow/handshake.ts
yanascz: Fixed.

suite-native/state/src/extraDependencies.ts
yanascz: OK, reverting.

szymonlesisz (approved):

PR #20931, opened 2025-08-21
test(suite-native): fix device settings element text matching

PeKne:
## Description
The end-to-end device settings tests were performing text matching without waiting for the screen content to render, which caused some tests to be unreliable. This fix addresses this issue.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=test/native/fix-device-settings-e2e-text-match&lookback=7)

### DIFF
--- a/suite-native/app/e2e/tests/deviceSettings.test.ts
+++ b/suite-native/app/e2e/tests/deviceSettings.test.ts
@@ -94,7 +94,9 @@ conditionalDescribe(device.getPlatform() === 'android', 'Device settings', () =>
 
             await onDeviceSettings.waitForSettingsScreen();
 
-            expect(element(by.label('new name'))).toBeVisible();
+            await waitFor(element(by.text('new name')))
+                .toBeVisible()
+                .withTimeout(10000);
         });
 
         test('Wipe device', async () => {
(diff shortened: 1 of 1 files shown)


suite-native/app/e2e/tests/deviceSettings.test.ts
```diff
@@ -114,7 +116,9 @@ conditionalDescribe(device.getPlatform() === 'android', 'Device settings', () =>
 
             await onDeviceSettings.passCheckBackupFlow();
 
-            expect(element(by.label('Your backup is valid'))).toBeVisible();
+            await waitFor(element(by.text('Your backup is valid')))
+                .toBeVisible()
+                .withTimeout(10000);
```
BrantalikP, line 121: 🥲

PeKne: What do you mean? 😁

BrantalikP: that we need to extend the timeout to make this test pass :D

BrantalikP (approved):

suite-native/app/e2e/tests/deviceSettings.test.ts
PeKne: There were multiple problems:
1. there was no timeout
2. the matching was done based on `by.label` not `by.text`.So it was looking for an `accessibilityLabel`
3. it was calling `Jest.expect()` instead of `detox.expect()`

So I am amazed it even worked sometimes :D

BrantalikP: aah, I missed the function swap. Thanks!

PR #20935, opened 2025-08-21
chore(connect): optimize cache local data

karliatto:
## Description
This PR adds couple of performance improvements to performance in the firmware release config decoding and verification. 

* Now the firmware release configs are only verified 2 times for each of them (local and remote) by caching data.
* Avoid unnecessary decode of firmware release config.

This should be a performance improvement specially in mobile since verify JWS function is very inefficient in that environment. Further improvemnts could be made by using a native library for mobile jWS verification.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/optimize-local-data-cache&lookback=7)

### DIFF
--- a/packages/connect/src/data/DataManager.ts
+++ b/packages/connect/src/data/DataManager.ts
@@ -5,6 +5,7 @@ import coins from '@trezor/connect-common/files/coins.json';
 import {
     ConditionalRelease,
     DeviceModelInternal,
+    FirmwareReleaseConfig,
     FirmwareType,
     IntermediaryReleaseConfig,
     ReleasesConfig,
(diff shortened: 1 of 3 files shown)


matejkriz (approved):
I can confirm that with this change the mobile performance during connect init is much much better. I haven't tested FW installation with this change.

STew790:
**QA OK**
Connection to mobile is indeed better. Installation of FW is OK as well.

**Info**
25.9.1 https://github.com/trezor/trezor-suite/commit/8f0c4dfe4b10c95461fca72f95b1de1fb6d72eb5

Issue #20866, opened 2025-08-19
Broken screen padding on FirmwareInstallationScreen


yanascz:
When installing FW via BT:

<img width="330"  alt="Image" src="https://github.com/user-attachments/assets/1004016e-a0bc-4bb5-945e-4fe96e45b1d9" />

<img width="330" height="854" alt="Image" src="https://github.com/user-attachments/assets/7521505c-3da1-4bad-aa85-f39c4f20c5d6" />

Possible to reproduce with TS7 and `feat/native/thp` (without the last hotfix).

## Expected behaviour

The _Continue_ button is placed correctly as it would be on any other Screen. I assume the logic from the `Screen` component has to be replicated/reused.

## Note for QA:
Please test it with all possible system navigation options 🙏

honzauher:
**QA OK**

Working as expected

**Info:**
- develop -25.9.1, commit 57533c
- OS: android & iOS

PR #20863, opened 2025-08-19
chore(suite-native): replace Bottom Sheet

BrantalikP:
## Description

Removal of Legacy Bottom Sheet and Replacement with the New Bottom Sheet Modal

### List of Affected Sheets:
	•	Account Rename Sheet
	•	Transaction Detail Sheet
	•	XPub Hint Sheet
	•	QR Scan Sheet
	•	Unverified Address Device Hint Sheet
	•	Unavailable Staking Sheet
	•	Connect Popup: Confirm on Trezor
	•	Unverified Address Sheet
	•	Wallet Backup: Select Type Sheet
	•	Firmware Backup Info Sheet
	•	Connect Trezor Help Sheet
	•	XPub QR Code Sheet
	•	Firmware Changelog
	•	Security Check – Seal Sheet
	•	WalletConnect Session Sheet

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/18907
## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/replace-bottom-sheet&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/native/replace-bottom-sheet&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/native/replace-bottom-sheet&lookback=7)

PR #20936, opened 2025-08-21
feat(suite-native): THP implementation for mobile

### LINKED ISSUE #19964: Trezor ↔ Host Protocol (THP) for mobile
[Figma with THP flows](https://www.figma.com/design/0L0g7OXJIALm0YHdyclXCN/Bluetooth---THP?node-id=5805-28162&p=f&t=vFLe5qE9kyjggtE3-0)

### What needs to be implemented for an MVP

- [x] THP pairing as part of FW installation
  - no NFC, just security code entry
  - UI adapted accordingly (temporarily)
- [x] THP connection once FW is already installed
- [x] autoconnect confirmation on 3rd attempt
- [x] works via both BT and USB

### Known issues

- [x] FW installation via BT during device onboarding stuck after device restart
- [x] FW installation on iOS via BT during device onboarding stuck for 20 seconds after device restart
  - might be caused by the fix of https://github.com/trezor/trezor-suite/issues/13456 which might already be obsolete because of https://github.com/trezor/trezor-suite/pull/19574
- [x] FW update during device onboarding fails if FW is already installed
  - either the device restarts in bootloader and there's no option to confirm FW installation
  - or the device screen goes dark and Suite thinks FW installation completed successfully
- [x] FW update from dashboard / device settings doesn't work, most likely the same issue as ☝️
- [x] Cancelling of THP pairing (and most likely other actions) in Suite does not clear the device screen
  - this is most likely caused by a FW issue

Follow-up for https://github.com/trezor/trezor-suite/issues/11751.

yanascz:
## Testing scenarios

- Device onboarding with a clean device
  - via BT (both platforms)
  - via USB (Android only)
- Connecting a device with FW installed
  - via BT (both platforms)
  - via USB (Android only)
- Updating FW via Device Settings
  - via BT (both platforms)
  - via USB (Android only)

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19964

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/native/thp&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/native/thp&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/native/thp&lookback=7)


suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
```diff
@@ -98,6 +99,12 @@ export const FirmwareInstallationScreenContent = ({
     const deviceRevision = originalDevice?.features?.revision;
     const deviceFirmwareVendor = originalDevice?.features?.fw_vendor;
 
+    useEffect(() => {
+        // On the first render, set the firmware installation status to 'initial'. A firmware update might have
+        // happened before, and we want to avoid showing any previous result when user returns to this screen.
+        dispatch(firmwareActions.setStatus('initial'));
+    }, [dispatch]);
```
juriczech, line 116: This is moved from the screen just because you need to trigger this in device settings as well?

suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
```diff
@@ -33,7 +34,7 @@ const bottomButtonsContainerStyle = prepareNativeStyle<{ isConfirmOnDeviceShown:
         position: 'absolute',
         left: utils.spacings.sp16,
         right: utils.spacings.sp16,
-        bottom: isConfirmOnDeviceShown ? 180 : 0,
+        bottom: isConfirmOnDeviceShown ? 180 : utils.spacings.sp52,
```
BrantalikP, line 37: Is this the fix for https://github.com/trezor/trezor-suite/issues/20866?

suite-native/device/src/hooks/useHandleDeviceConnection.ts
```diff
@@ -103,24 +112,47 @@ export const useHandleDeviceConnection = () => {
...
             !shouldNavigateToDeviceCompromisedModal
         ) {
-            navigation.navigate(RootStackRoutes.DeviceOnboardingStack, {
-                screen: DeviceOnboardingStackRoutes.UninitializedDeviceLanding,
-            });
+            if (!wasDeviceOnboardingCancelled) {
+                navigation.navigate(RootStackRoutes.DeviceOnboardingStack, {
+                    screen: DeviceOnboardingStackRoutes.UninitializedDeviceLanding,
+                });
+            } else if (isAuthorizeDeviceStackFocused) {
```
juriczech, line 148: Please add a comment why is this condition for AuthorizeDeviceStackFocused present here. It will be refactored pretty soon so just making sure it's understandable please 🙏

suite-native/module-device-onboarding/src/hooks/useInitiateThpConnection.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    const thpStep = useSelector(selectThpStep);
+
+    const initiateThpConnection = () => {
+        // Device is acquired by the FW installation hook, just respond as expected.
+        TrezorConnect.uiResponse({ type: 'ui-receive_confirmation', payload: true });
+    };
+
+    useEffect(() => {
+        if (thpStep === 'ConfirmConnectionBeforePairing' || thpStep === 'ConfirmOnlyConnection') {
```
BrantalikP, line 32: we are duplicating the logic from [useHandleDeviceconnection](https://github.com/trezor/trezor-suite/pull/20936/files#diff-2c42ad1964a723c2be0af81a1caacafe51f7c0aaabc5bcbf253830598d96311cR120) is possible to unify it, maybe creating a dedicated global hook? The reason is whether the condition or logic would change we would need to be also aware of the other locations of this call

suite-native/device/src/hooks/useDetectDeviceError.tsx
```diff
@@ -87,7 +89,12 @@ export const useDetectDeviceError = () => {
     // If device is unacquired (restarted app, another app fetched device session, ...),
     // we cannot work with device anymore. Shouldn't happen on mobile app but just in case.
     useEffect(() => {
-        if (isUnacquiredDevice && isOnboardingFinished) {
+        if (
+            isOnboardingFinished &&
+            isUnacquiredDevice &&
+            !isDeviceThpRequired &&
+            !isFirmwareInstallationRunning
```
juriczech, line 96: Why is it necessary to check firmware installation too? I'm just curious how it works and why THP check wouldn't be sufficient 🤔

packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
```diff
@@ -266,13 +265,10 @@ class ReactNativeUsbModule : Module() {
     private fun closeAllOpenedDevices() {
         Log.d(LOG_TAG, "Closing all devices")
 
-        openedConnections.values.forEach { it.close() }
-        openedConnections.clear()
-    }
-
-    private fun cancelAllPendingRequests() {
         pendingRequests.values.forEach { it.cancel() }
```
matejkriz: I don't think we can cancel pending requests in closeAllOpenedDevices because we call it every time the app goes on background - that would cancel any pending request even if there is an incoming phone call for example.

suite-native/module-authorize-device/src/hooks/useInitiateThpConnection.ts
```diff
@@ -0,0 +1,17 @@
+import { useDispatch, useSelector } from 'react-redux';
+
+import { acquireDevice, selectSelectedDevice } from '@suite-common/wallet-core';
+
+export const useInitiateThpConnection = () => {
```
juriczech, line 5: Nit: It seems weird that there is same name of a hook in `@suite-native/module-device-onboarding` and both do different things. Can we apply better name for either of them to avoid duplicated name?

suite-native/device/src/components/ContinueOnTrezorScreenContent.tsx
```diff
@@ -61,7 +61,7 @@ export const ContinueOnTrezorScreenContent = ({
             </VStack>
             <Box flex={1} alignItems="center" justifyContent="flex-end">
                 <DeviceImage
-                    deviceModel={deviceModel || DeviceModelInternal.T3T1}
+                    deviceModel={deviceModel || DeviceModelInternal.T3W1}
```
matejkriz: We should not use T3W1 image as a default here because that image claims `Trezor Safe 7`, and we would show it to the users in upcoming releases. Let's stay with T3T1 here, please. Or even better - use the exact copy of T3T1 image as T3W1 for now.

suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
```diff
@@ -277,16 +286,19 @@ export const FirmwareInstallationScreenContent = ({
...
                                     <Translation id="firmware.firmwareUpdateProgress.dontCloseAppMessage" />
                                 }
                             />
-                        </Box>
-                    )}
+                        ) : (
+                            // Blank space to prevent layout shift when done
+                            <Text variant="hint"> </Text>
+                        )}
+                    </Box>
```
matejkriz, line 303: It would be better to have it as a separate PR, but let's leave it here now.

suite-native/thp/src/components/SecurityCodeInput.tsx
```diff
@@ -0,0 +1,109 @@
...
+    onSubmit: (code: string) => void;
+};
+
+const textInputStyle = prepareNativeStyle(_ => ({
+    position: 'absolute',
+    top: -9999,
+    opacity: 0,
+}));
+
+// TODO: Move to atoms maybe?
```
matejkriz: I don't think it belongs in atoms if it's not reused on multiple places in multiple contexts.

suite-native/thp/src/components/SecurityCodeInput.tsx
```diff
@@ -0,0 +1,109 @@
...
+            if (newCode.length === length) {
+                onSubmit(newCode);
+            }
+        } else if (key === 'Backspace' && code.length > 0) {
+            setCode(code.slice(0, -1));
+        }
+    };
+
+    useEffect(() => {
+        setTimeout(focusInput, 1);
```
matejkriz: add cleanup, please

suite-native/thp/src/components/ThpPairingInfoHelpButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    return (
+        <Box>
+            <IconButton
+                colorScheme="tertiaryElevation0"
+                size="medium"
+                iconName="question"
+                onPress={toggleBottomSheet}
+            />
+            <BottomSheet
```
matejkriz: Please use BottomSheetModal instead, we are trying to get rid off the BottomSheet in https://github.com/trezor/trezor-suite/pull/20863

suite-native/thp/src/hooks/useThpAlert.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const dispatch = useDispatch();
+    const { showAlert } = useAlert();
+
+    const device = useSelector(selectSelectedDevice);
+
+    const turnOnAutoconnect = useCallback(async () => {
+        const response = await TrezorConnect.thpGetCredentials({ device });
+        if (response.success) {
+            dispatch(thpActions.addCredential({ credential: response.payload }));
+        }
```
juriczech, line 21: How is the failure handled?

packages/transport-native-bluetooth/src/api/bluetoothManager.ts
```diff
@@ -12,22 +13,16 @@ import {
 
 import { EventEmitter } from 'events';
 
-import type { IntervalId } from '@trezor/type-utils';
+import { readMessageBuffer } from '@trezor/transport/src/utils/readMessageBuffer';
```
juriczech, line 16: import { readMessageBuffer } from '@trezor/transport';

packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
matejkriz: I'll test it.

suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
yanascz: No, it's a hotfix for the page to not look broken. I haven't checked all scenarios on both platforms.

packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
matejkriz: Not able to confirm receive address after leaving the app and getting back again.
image

Please revert in this PR if possible and let's discuss separately in different PR 🙏

suite-native/thp/src/components/ThpPairingInfoHelpButton.tsx
matejkriz: can be followup

suite-native/firmware/src/components/FirmwareInstallationScreenContent.tsx
yanascz: Yep, I believe it was for this reason. I noticed incorrect behaviour without this last time i tried.

packages/react-native-usb/android/src/main/java/io/trezor/rnusb/ReactNativeUsbModule.kt
yanascz: When I last tested, it worked correctly with this change. Will revert since it's not necessary for BT.

suite-native/device/src/components/ContinueOnTrezorScreenContent.tsx
yanascz: Good point, it made sense only while it was in a branch. Reverted.

suite-native/thp/src/components/SecurityCodeInput.tsx
yanascz: Resolved.

packages/transport-native-bluetooth/src/api/bluetoothManager.ts
yanascz: Can't do, it's not exported. I discussed with the Connect team and they're happy with it. 🤷‍♂️ It's imported like this also in `packages/transport-bluetooth/src/client/bluetooth-api.ts`.

suite-native/thp/src/hooks/useThpAlert.tsx
yanascz: It's ignored. What would you expect to happen? I don't think this was even considered by the UX team.

suite-native/module-authorize-device/src/hooks/useInitiateThpConnection.ts
yanascz: Any suggestions? It does different things because each module has different flows.

suite-native/device/src/hooks/useDetectDeviceError.tsx
yanascz: Since THP is a part of FW installation. And FW installation needs to handle THP stuff on its own otherwise the FW installation breaks.

suite-native/device/src/hooks/useHandleDeviceConnection.ts
yanascz: Done.

suite-native/thp/src/components/SecurityCodeInput.tsx
yanascz: Added.

suite-native/module-device-onboarding/src/hooks/useInitiateThpConnection.ts
yanascz: No, we're not, we navigate to a different screen.

suite-native/thp/src/hooks/useThpAlert.tsx
juriczech: If it hasn't been a requirement, we can tackle this separately if needs be.

suite-native/module-authorize-device/src/hooks/useInitiateThpConnection.ts
juriczech: Let's not block this PR, we can rename later

juriczech (approved):
LGTM from my side & checked that comments from @matejkriz were also fixed, so I think let's go ahead with this and fix the bottom sheet and other issues in a follow up PR if needed.

PR #20939, opened 2025-08-21
Chore/native/disconnect navigation middleware

### LINKED ISSUE #20601: Send fails silently for view only devices
**Describe the bug**
If you try to send some coins, it fails silently after going though the send form and selecting fees.

**Info:**

- Mobile Suite Version: 25.8.1
- OS: name, version

**How to reproduce**

1. Connect device
2. Disconnect it and have it as view only
3. Navigate to send form of any coin
4. Input address and amount
5. Select fee
6. Now send fails silently and goes back to dashboard

**Expected behavior**
Suite shoul prompt user to connect a device.

### LINKED ISSUE #20938: Refactor navigation on device disconnect
In a similar fashion to refactoring navigation to connecting screens, we should refactor [this](https://github.com/trezor/trezor-suite/blob/f74f43213c0a7d0d76be7d25500cfb6ee72aaf04/suite-native/device/src/hooks/useHandleDeviceConnection.ts#L190) `useEffect` as well and share some functionality.

**Why**
We're trying to simplify the navigational logic and make this more maintainable, but also, this part of code can become problematic because it relies on only one device being connected. This will potentially cause unexpected behavior in the future when connecting multiple devices to the app and disconnecting one of them. This effect would not be triggered even though it should have been. Changing the behavior to react to device events will be better.

**Acceptance criteria**
- [x] app maintains same functionality which is correct, but the effect is only triggered by device disconnection
- [x] trigger navigation on device disconnect rather than checking for 0 connected devices

juriczech:
## Description
- move disconnection logic from `useHandleDeviceConnection` to `deviceConnectionMiddleware`
- Breaking change: `selectIsNoPhysicalDeviceConnected` was used to trigger disconnection, which was flawed and was replaced by `deviceActions.deviceDisconnect`. This is IMO better approach to handle disconnection with multiple devices present in the app, which is nearing with BT & THP connections possible. I tested that nothing broke, but it's possible I overlooked something
- tried to do some small refactors to the files, but don't be too strict yet please. I still want to get rid of the `useHandleDeviceConnection` hooks first and then possibly clean this up too and avoid as many repeated code as possible and also better organization of these utilities

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20938
Resolve https://github.com/trezor/trezor-suite/issues/20601

## Screenshots:

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=chore/native/disconnect-navigation-middleware&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=chore/native/disconnect-navigation-middleware&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/disconnect-navigation-middleware&lookback=7)


suite-native/navigation/src/checkIsActiveRouteBlacklisted.ts
```diff
@@ -0,0 +1,19 @@
+import { navigationContainerRef } from './components/NavigationContainerWithAnalytics';
+import { DeviceOnboardingStackRoutes, RootStackRoutes } from './routes';
+
+type RouteType = RootStackRoutes | DeviceOnboardingStackRoutes;
+type Routes = RouteType[];
+
+const getActiveRouteName = () =>
+    navigationContainerRef.getState()?.routes.at(-1)?.name as RouteType;
+
+export const checkActiveRouteMatch = ({ routeList }: { routeList: Routes }) => {
```
yanascz: The name is confusing... What about `isActiveRouteOneOf`, `isActiveRouteIncluded` or similar?

suite-native/navigation/src/checkIsActiveRouteBlacklisted.ts
```diff
@@ -0,0 +1,19 @@
...
+    const activeRouteName = getActiveRouteName();
+
+    if (!activeRouteName) return false;
+
+    return routeList.includes(activeRouteName);
+};
+
+export const checkIsActiveRouteBlacklisted = ({ blacklist }: { blacklist: Routes }) =>
```
yanascz: I'd be OK with keeping the argument as a simple array, might make the usage more readable.
```suggestion
export const checkIsActiveRouteBlacklisted = (blacklist: Routes) =>
```

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
```diff
@@ -87,3 +95,38 @@ deviceConnectionMiddleware.startListening({
         handleDeviceConnectNavigation({ isCoinEnablingInitFinished });
     },
 });
+
+deviceConnectionMiddleware.startListening({
+    predicate: action => deviceActions.deviceDisconnect.match(action),
```
yanascz, line 95: Would be great to cover with tests.

suite-native/navigation/src/checkIsActiveRouteBlacklisted.ts
juriczech: In this case, I'd prefer to keep the named param. I think it adds to the explanation of what it's doing rather than sending the array as param and going inside the function to check how it's doing it. But maybe it's a sign of bad name...no sure. But I think the formatting is a small price to pay for the named param.

suite-native/device/src/middlewares/deviceConnectionMiddleware.ts
juriczech: I am preparing e2e tests for this in a separate PR, incoming soon 🤞

juriczech: Then I plan to also test the middleware effects in a unit tests, but I will do that when we delete useHandleDeviceConnection probably

suite-native/navigation/src/checkIsActiveRouteBlacklisted.ts
juriczech: Good point. Renamed to `isActiveRouteAnyOf `

suite-native/navigation/src/checkIsActiveRouteBlacklisted.ts
yanascz: Well, it's like with `includes` used in the function. You just pass in `element`, not `{element: T}`. E.g. with `isActiveRouteAnyOf`, I'd expect an array to be used as argument.

suite-native/navigation/src/checkIsActiveRouteBlacklisted.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const isActiveRouteAnyOf = ({ routeList }: { routeList: Routes }) => {
+    const activeRouteName = getActiveRouteName();
+
+    if (!activeRouteName) return false;
+
+    return routeList.includes(activeRouteName);
+};
+
+export const checkIsActiveRouteBlacklisted = ({ blacklist }: { blacklist: Routes }) =>
```
yanascz: Would now make sense to rename?
```suggestion
export const isActiveRouteBlacklisted = ({ blacklist }: { blacklist: Routes }) =>
```

juriczech: isActiveRouteBlacklisted doesn't evoke function name in my obinion, but variable name.

juriczech: But there was inconsistency in naming between those 2 functions. I prefixed both with `check..` now.

suite-native/navigation/src/checkIsActiveRouteBlacklisted.ts
juriczech: Fixed

suite-native/navigation/src/checkIsActiveRouteBlacklisted.ts
yanascz: Yeah, was referring to the inconsistency. 👍

yanascz (approved):
Hope this helps! 🤞

PR #20944, opened 2025-08-21
BT/THP NAPI for MacOS

martykan:
## Description

There has been a persistent issue on certain Macs, where the paring PIN dialog never shows up. 
It seems there are two criteria that need to be met:
1. The process must belong to a signed .app bundle
2. The process must not be running in background, at minimum it must have an icon in the dock

This PR adds [NAPI-RS](https://napi.rs) bindings to build the Rust code as a library, so it can be loaded in Node directly.
Suite's Electron process already meets both of the requirements, so with the bluetooth server running inside of it, it actually works and fixes the pairing issue. 

Considerations: 
1. Fatal errors in Rust could now crash Electron as well
2. Stopping the server is currently not implemented
3. The issue is still present on CI builds, because they are not signed


packages/transport-bluetooth/scripts/Dockerfile
```diff
@@ -2,42 +2,52 @@
...
 # Build for Linux x64
-RUN cargo build --target x86_64-unknown-linux-gnu --release && \
+RUN cargo build --target x86_64-unknown-linux-gnu --release --bins && \
   mkdir -p /output/linux-x64 && \
   cp ./target/x86_64-unknown-linux-gnu/release/trezor-bluetooth /output/linux-x64/trezor-bluetooth && \
   strip /output/linux-x64/trezor-bluetooth
+
+# Build NAPI for macOS
+COPY ./package.json ./package.json
+RUN yarn global add @napi-rs/cli && yarn build:napi
```
szymonlesisz: missing empty line

packages/transport-bluetooth/scripts/Dockerfile
```diff
@@ -2,42 +2,52 @@
 
 FROM joseluisq/rust-linux-darwin-builder:1.85.1
 
-RUN apt-get update && apt-get install -y mingw-w64
+RUN apt-get update && apt-get install -y mingw-w64 curl ca-certificates
+
+# Node.js
```
szymonlesisz, line 7: maybe we can add it below the cargo builds? this way we will have all node coxtext/jobs next to eachother

packages/transport-bluetooth/Cargo.toml
```diff
@@ -30,3 +32,13 @@ bluez-async = "0.7.2"
...
+
+[features]
+default = []
+napi = ["dep:napi", "dep:napi-derive", "dep:napi-build"]
+
+[lib]
+crate-type = ["cdylib"]
+
+[build-dependencies]
+napi-build = { version = "2", optional = true }
```
szymonlesisz: missing empty line

.prettierignore
```diff
@@ -141,3 +141,6 @@ packages/suite-data/files/translations/master.json
 packages/suite-data/files/videos/lottie/*.json
 
 /.nx/cache
+
+# NAPI generated files 
+packages/transport-bluetooth/napi
```
szymonlesisz, line 146: it this intended? 
if yes then it's not working because files are included anyway, but i think we want to track those builds, no?

or if we dont want to track it then maybe
```
yarn global add @napi-rs/cli && yarn build:napi
```

should be done not in the Dockerfile but as a part of transport-bluetooth build:lib ?

szymonlesisz: ah, its priettier not git :)

packages/transport-bluetooth/package.json
```diff
@@ -20,10 +20,26 @@
...
         "build:server": "./scripts/build.sh",
+        "build:napi:apple-arm": "napi build -t aarch64-apple-darwin --release --platform --js index.js  --output-dir ./napi --features napi",
+        "build:napi:apple-intel": "napi build -t x86_64-apple-darwin --release --platform --js index.js  --output-dir ./napi --features napi",
+        "build:napi": "yarn build:napi:apple-arm && yarn build:napi:apple-intel",
         "build:ui": "webpack --config ./webpack/webpack.build.js"
     },
+    "napi": {
+        "binaryName": "trezor_bluetooth",
+        "targets": [
+            "x86_64-pc-windows-msvc",
```
szymonlesisz: we should use the same targets as in dockerfile

packages/transport-bluetooth/package.json
```diff
@@ -32,6 +48,7 @@
         "@trezor/websocket-client": "workspace:*"
     },
     "devDependencies": {
+        "@napi-rs/cli": "^3.1.5",
```
szymonlesisz: since this is needed only in docker and it is explicitly added there maybe it shoudnt be a part of monorepo dependency as it's never used by a local context

packages/transport-bluetooth/scripts/Dockerfile
martykan: It's better to install the dependencies first before copying files, this way the layer gets cached and doesn't have to rerun every time. In CI it doesn't matter, since we don't persist the cache, but it's nice to have for local development

PR #20971, opened 2025-08-22
Feat/update new button component

seibei-iguchi:
🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/update-new-button-component&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/update-new-button-component&lookback=7)

### DIFF
--- a/packages/components/src/components/buttons/NewButton/NewButton.tsx
+++ b/packages/components/src/components/buttons/NewButton/NewButton.tsx
@@ -2,10 +2,9 @@ import { ButtonHTMLAttributes } from 'react';
 
 import styled, { useTheme } from 'styled-components';
 
-import { spacings } from '@trezor/theme';
-
 import { NewButtonIntent, NewButtonPriority, NewButtonSize } from './types';
 import {
+    addAlphaToHex,
     mapPropsToCSS,
     mapPropsToColor,
     mapSizeToBorderRadius,
@@ -20,6 +19,7 @@ import {
     withFrameProps,
 } from '../../../utils/frameProps';
 import { TransientProps } from '../../../utils/transientProps';
+import { Box } from '../../Box/Box';
 import { Row } from '../../Flex/Flex';
 import { Icon, IconName } from '../../Icon/Icon';
 import { Spinner } from '../../loaders/Spinner/Spinner';
@@ -100,6 +100,7 @@ export type NewButtonProps = SelectedHTMLButtonProps &
         'data-testid'?: string;
         intent?: NewButtonIntent;
         priority?: NewButtonPriority;
+        shortcut?: string[];
     };
 
 export const NewButton = ({
@@ -113,6 +114,7 @@ export const NewButton = ({
     isInverse = false,
     isLoading = false,
     onClick,
+    shortcut,
     size = 'medium',
     tabIndex,
     target,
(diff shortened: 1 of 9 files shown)


packages/components/src/components/buttons/NewButton/NewButton.tsx
```diff
@@ -160,16 +157,37 @@ export const NewButton = ({
...
+                        color={color}
+                        ellipsisLineCount={1}
+                    >
+                        {children}
+                    </Text>
+                </Box>
                 {iconRight && !isLoading && <Icon name={iconRight} {...iconProps} />}
+                {shortcut && (
+                    <Row gap={2}>
+                        {shortcut.split('').map((char, index) => (
```
jvaclavik: I think this is not correct, what if I want to use `CMD` + something? How should I do it?

packages/components/src/components/buttons/NewButton/NewButton.tsx
```diff
@@ -160,16 +157,37 @@ export const NewButton = ({
...
+                            <Box
+                                backgroundColor={addAlphaToHex(
+                                    theme.baseFillElementNeutralDark,
+                                    0.09,
+                                )}
+                                padding={{ horizontal: 4 }}
+                                borderRadius="4px"
```
jvaclavik: btw just for discussion - what about using it also without `px`? Shouldn't we allow also % (probably not)?

seibei-iguchi: Yep, I just thought about that. Will update it.

packages/components/src/components/buttons/NewButton/NewButton.tsx
seibei-iguchi: I wanted to keep it as simple as possible. We don't use CMD but the icon/sign for command or alt in the app, so should be fine.

jvaclavik: and what symbol do we use for CTRL on Windows?

seibei-iguchi: I don't think we actually differentiate that:D but the symbol should be ^, no?

seibei-iguchi: I'd say, if we need it, let's transform it into an array, but until then yolo.

seibei-iguchi: AFAIK, we only use the shortcut in one place in the app, in the open previously used passphrase button and that uses alt.

jvaclavik: I would prefer an array a bit, because string doesn't say anything about format how it should look like

This is what we use `packages/components/src/components/HotkeyBadge/keyboardKeys.ts`, so on windows we have `Alt` for Windows:

```
ALT: { code: 18, value: 'Alt', valueMac: '⌥' },
```

seibei-iguchi: Ok then, will do:)

jvaclavik (approved):

PR #20973, opened 2025-08-22
feat(suite-native): prepare T3W1 device color images

yanascz:
To be replaced with proper renders before launch.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/native/t3w1-backplates&lookback=7)


suite-native/bluetooth/src/components/BluetoothDeviceCard.tsx
```diff
@@ -45,7 +44,7 @@ export const BluetoothDeviceCard = ({
     return (
         <Card style={applyStyle(cardStyle)}>
             <Box alignItems="center">
-                <DeviceImage deviceModel={deviceModel} size="small" />
+                <DeviceBackplateImage color={deviceColor} />
```
BrantalikP: Wouldn’t it be better to keep it in the `DeviceImage` component and just add the images to the array, like `T3W1-color`, `T3W1-anotherColor`, etc.? It could be usefull if we need backplate images for other models.

yanascz: It's not a device image but a device **backplate** image. Using `DeviceImage` in this package leads to circular dependency, see TODO in `BluetoothDeviceCard.tsx`. We can always move it elsewhere once necessary.

BrantalikP: Oh, I missed the imports—I see. I was thinking about `reusability`, so that the `DeviceImage` would also support `backplate` colors. The new `DeviceBackplateImage` doesn’t support other models in case we release another Bluetooth-supported device in the future.

yanascz: It doesn't support any other models because we don't have images for any other models at the moment. Once we have some other BT devices, we can extend the component and/or move it elsewhere. I don't see any point in implementing support for something that might never occur.

BrantalikP: Seeing the glass half empty huh? :D But yeh, you have solid points. Thanks for the discussion 💪

BrantalikP (approved):

suite-native/bluetooth/src/components/BluetoothDeviceCard.tsx
yanascz: No, years of coding experience. 😉 I've seen too many lines of code written just because "we might use it one day" and no one did after 10 or 15 years. The only outcome was more complex codebase, nothing more.

suite-native/bluetooth/src/components/DeviceBackplateImage.tsx
```diff
@@ -0,0 +1,15 @@
+import { Image } from '@suite-native/atoms';
+
+const deviceImageMap: Record<number, string> = {
+    1: require('../assets/backplate-t3w1-1.webp'),
```
matejkriz: nitpick, feel free to ignore, but this naming pattern would be more consistent with what we have for T3T1 and T3B1 color variants:

`trezor_t3w1_color_1.webp `

yanascz: Fixed in ac4edde071.

PR #21012, opened 2025-08-25
feat(suite-native): enable Connect device and Bluetooth FFs for adhoc builds

yanascz:
🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/native/adhoc-build-ffs&lookback=7)

### DIFF
--- a/suite-native/app/eas.json
+++ b/suite-native/app/eas.json
@@ -37,6 +37,10 @@
         },
         "adhoc": {
             "extends": "preview",
+            "env": {
+                "EXPO_PUBLIC_FF_IS_DEVICE_CONNECT_ENABLED": "true",
+                "EXPO_PUBLIC_FF_IS_BLUETOOTH_ENABLED": "true"
+            },
             "developmentClient": false
         },
         "production": {


BrantalikP (approved):

suite-native/app/eas.json
```diff
@@ -37,6 +37,10 @@
         },
         "adhoc": {
             "extends": "preview",
+            "env": {
```
BrantalikP, line 40: qq: do not we want to allow it also for the parent env `preview`? If so I would move it there

suite-native/app/eas.json
```diff
@@ -37,6 +37,10 @@
...
+                "EXPO_PUBLIC_FF_IS_DEVICE_CONNECT_ENABLED": "true",
```
matejkriz, line 41: I'm not sure how `extends` works exactly for object values but I can imagine this will override it and so EXPO_PUBLIC_ENVIRONMENT is missing here

yanascz: All the profiles already extend "base" and it seems to work correctly. 🤔

STew790:
**QA OK**
FFs for adhoc builds are enabled.

PR #21037, opened 2025-08-26
fix(suite-native): prevent crash on ios by removing extra navigation

matejkriz:
## Description

We have to double check that it's really needed and doesn't break any other flow.

## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/prevent-ios-crash-thp&lookback=7)


matejkriz:
Let's close this, we don't experience any crashing from navigation right now.

PR #20920, opened 2025-08-20
feat(connect): firmware update via BluetoothTransport

szymonlesisz:
## Description
last missing piece for @yanascz + bt-mobile implementation.

After the reboot from/to bootloader device stays connected to bluetooth, to keep current flow we need to request the UI (suite) to disconnect the device and then auto reconnect it.

**Note**: current FW build [have a bug](https://github.com/trezor/trezor-firmware/issues/5584) - this is why in `feat/bt-thp` the second condition needs to be commented out to work properly

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/bt-fw-update&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/bt-fw-update&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/bt-fw-update&lookback=7)

PR #21041, opened 2025-08-26
fix(connect): onCallFirmwareUpdate bluetooth and device-disconnect race condition

szymonlesisz:
## Description

Fix for the issue mentioned here: https://github.com/trezor/trezor-suite/pull/20920

after successful FW update device reboots from bootloader to normal mode. 

suite expects that device gets disconnected **or** waits 4 seconds (countdown on device) a tries to force disconnection via suite > bluetoothIpc (UI.FIRMWARE_DISCONNECT)

in case 1, we dont expect antoher device-disconnect event so we should use `method = 'auto'` of `waitForReconnectedDevice`

in case 2 or not bluetooth device (default) we expect disconnection so is  `method = 'wait'`

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/thp-fw-update-reboot&lookback=7)

### DIFF
--- a/packages/connect/src/core/onCallFirmwareUpdate.ts
+++ b/packages/connect/src/core/onCallFirmwareUpdate.ts
@@ -180,6 +180,66 @@ const waitForReconnectedDevice = async (
     return reconnectedDevice;
 };
 
+const getRebootMethod = async ({
+    deviceList,
+    device,
+    log,
+}: {
+    deviceList: DeviceList;
+    device: Device;
+    log: Log;
+}) => {
+    let method: ReconnectParams['method'] = 'wait';
+
+    // not a bluetooth device
+    if (!device.bluetoothProps) {
+        return method;
+    }
+
+    const ctrl = new AbortController();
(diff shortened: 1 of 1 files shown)


packages/connect/src/core/onCallFirmwareUpdate.ts
```diff
@@ -180,6 +180,66 @@ const waitForReconnectedDevice = async (
...
+    log,
+}: {
+    deviceList: DeviceList;
+    device: Device;
+    log: Log;
+}) => {
+    let method: ReconnectParams['method'] = 'wait';
+
+    // not a bluetooth device
+    if (device.bluetoothProps) {
```
Lemonexe: This may be unwanted.
```suggestion
    if (!device.bluetoothProps) {
```

Lemonexe (approved):
LGTM, just the inverted condition :pray:

packages/connect/src/core/onCallFirmwareUpdate.ts
szymonlesisz: cool, thx, not intentional

PR #21071, opened 2025-08-27
chore(suite-native): add possibility to disable device and FW checks via ENV var

### LINKED ISSUE #21029: Turn off device auth checks and FW checks for adhoc builds by default
To simplify testing of TS7

matejkriz:
## Description

This PR add possibility to change the default state for FW and device authenticity checks. Those are enabled by default. It is possible to set it via environment variables only in build-time, not runtime. User can still change the value in settings and his choice is persistent. 

There is no migration, so if you have the app already installed and checks enabled, an update to this version will not disable it.

We are disabling those checks for ad-hoc builds temporarily (until T3W1 can pass those checks) to make internal testing easier.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/21029

## Screenshots:
image

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/disable-device-checks-on-adhoc-builds&lookback=7)


yanascz (approved):
LGTM

PR #21109, opened 2025-08-28
chore(suite-native): fill in missing translations

### LINKED ISSUE #21103: Localization: Fill in all the missing translations
App historically did not use `react-intl`, so there are some hardcoded english string in the codebase. Found all of them and replace them with a translation key.

PeKne:
## Description
- [script](https://gist.github.com/PeKne/fe0fa1d8c70d088d2d748d09b043cefb) for printing out all the potentially translatable strings was created
  - this script uses babel parser to convert the code to the abstract syntax tree (in JSON format), then only JSX strings and general string literals were printed
- I went through all the string and replaced those that are user facing

## Related Issue

Resolve #21103 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/fill-in-missing-translations&lookback=7)

### DIFF
--- a/suite-native/module-trading/src/components/history/TradeStatusBadge.tsx
+++ b/suite-native/module-trading/src/components/history/TradeStatusBadge.tsx
@@ -1,7 +1,7 @@
 import { TradingTransactionStatus } from '@suite-common/trading';
 import { Badge, BadgeVariant } from '@suite-native/atoms';
 import { IconName } from '@suite-native/icons';
-import { Translation } from '@suite-native/intl';
+import { Translation, useTranslate } from '@suite-native/intl';
 import { exhaustive } from '@trezor/type-utils';
 
 export type TransactionStatusProps = {
@@ -98,6 +98,7 @@ const getLabel = (status: TradingTransactionStatus) => {
 };
 
 export const TradeStatusBadge = ({ status }: TransactionStatusProps) => {
+    const { translate } = useTranslate();
     if (!status) {
         return null;
     }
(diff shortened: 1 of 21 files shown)


suite-native/module-trading/src/components/history/TradeStatusBadge.tsx
```diff
@@ -108,7 +109,7 @@ export const TradeStatusBadge = ({ status }: TransactionStatusProps) => {
             size="small"
             variant={getBadgeVariant(status)}
             icon={getBadgeIconName(status)}
-            accessibilityHint="Trade status badge"
+            accessibilityHint={translate('moduleTrading.tradeHistory.status.badge')}
```
BrantalikP, line 112: why is it not possible to use <Translation here?

suite-native/notifications/src/components/TransactionNotification.tsx
```diff
@@ -111,7 +113,7 @@ export const TransactionNotification = ({
             isHiddenAutomatically={isHiddenAutomatically}
             onHide={handleRemoveNotification}
             onPress={navigateToTransactionDetail}
-            title={title}
+            title={translate(titleTxKey)}
```
BrantalikP, line 116: why not to use the <Translation ? JSX is already used below

BrantalikP (approved):
Great Idea with the Script! 🔥I am curious if we could not make an eslint rule for it 🤔

PeKne:
> Great Idea with the Script! 🔥I am curious if we could not make an eslint rule for it 🤔


Should be pretty easy for JSXStrings with a eslint rule, but very hard to filter the string literals (places where we pass the string as a variable or using useTranslate()).

PR #21161, opened 2025-08-29
test(connect): use only local firmware release config

karliatto:
## Description
There is no need to use remote FW release config in tests and it might make it more unpredictable and flaky, so for tests using it only local.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=test/use-only-local-fw-release-config-in-tests&lookback=7)


matejkriz (approved):
Makes sense.

PR #21164, opened 2025-08-29
chore(connect): removing unecessary TODO comment

karliatto:
## Description

When I wrote this comment I thought that we could use utility that gets firmware revision from local or remote but I realized there is actual value in knowing if it is from remote or not since the error is different and some errors are retry-able  and other are not.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/remove-comment-since-I-realized-it-does-not-make-sense&lookback=7)

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -691,8 +691,6 @@ export class Device extends TypedEmitter<DeviceEvents> {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


mroz22 (approved):

Lemonexe (approved):
Yes, makes sense.
This `release` at this point, shall be only local, looking at the new `getReleaseAsset` it only gets the already loaded release or nothing.
It is then responsibility of `checkFirmwareRevision` to fetch online if `!expectedRevision`

PR #21175, opened 2025-08-29
chore(connect): firmware URL commentary & types

Lemonexe:
## Description

Add some commentary to clarify different URLs for firmware releases, because it's otherwise difficult to understand how does the URL look like from the code.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/FW-binary-URL-commentary&lookback=7)

### DIFF
--- a/packages/connect/src/data/firmwareInfo.ts
+++ b/packages/connect/src/data/firmwareInfo.ts
@@ -59,7 +59,16 @@ const FIRMWARE_REMOTE_BASE_URLS: Record<FirmwareUpdateSource, RemoteBaseInfo> = 
     'localhost-signed': SIGNED_LOCALHOST,
 };
 
-export const getOnlineFirmwareBaseUrl = () => {
+type OnlineFirmwareBaseUrl = RemoteBaseInfo & { env: FirmwareUpdateSource };
+
+/**
+ * Obtains the base URL and middle path where to find firmware releases, based on the current settings.
+ * Examples:
+ *   { BASE_URL: 'https://data.trezor.io', MIDDLE_PATH: 'firmware', env: 'production' }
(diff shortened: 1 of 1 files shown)


packages/connect/src/data/firmwareInfo.ts
```diff
@@ -124,6 +124,13 @@ const getOnlineReleaseByPath = async (releasePath: string) => {
...
 };
 
+/**
+ * Returns the base URL and middle path for firmware updates based on the current settings.
+ * Examples:
+ * - production (default) https://data.trezor.io/firmware/t3t1/universal/t3t1-2.8.10-universal.json
+ * - test-unsigned https://data.trezor.io/dev/firmware/releases/unsigned/t3w1/universal/t3w1-2.9.1-universal.json
+ * - localhost-unsigned http://localhost:3000/firmware/unsigned/t3w1/universal/t3w1-2.9.1-universal.json
+ */
 export const getOnlineReleaseByVersion = async (
```
karliatto: This is not true `getOnlineReleaseByVersion` is returning online release as it names says, no base URL and middle path.

karliatto (changes requested):

packages/connect/src/data/firmwareInfo.ts
Lemonexe: Yes, thanks, I got it confused. That's really the reason why I'm doing this – I was quite lost in the new logic, and trying to make sense of the new concepts and flows.
I fixed it, and added more comments and explicit types. Hopefully correct now :slightly_smiling_face: 

In this specific case, I extracted `getOnlineReleasePath` so I could better document it.

Lemonexe:
[Connect-popup tests broken](https://github.com/trezor/trezor-suite/actions/workflows/test-connect-popup.yml), it's not related

karliatto (approved):
:+1:

karliatto:
It looks right now. :+1:

PR #21190, opened 2025-08-30
Entropy check check

mroz22:
problems solved

- `device already initialised` - conflict between 2 tabs - in general, the problem is that onboarding does not handle unacquired/occupied device at all. 
- `ui-device_bootloader_mode` - this is probably solved, needs some testing, I would suspect again the same cuplrit using suite in 2 tabs.
- `cancelled`, `cancelado` etc. - imho we simply need to filter out this kind of expected error (when cancelled by user)

problems remaining 

-  `device disconnected during action` - could be caused by conflict of 2 tabs, tab A started reset device flow but hasn't finished it. Tab B fired a call to device. 
- `unexpected error` 
- `Malformed protocol format`

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=entropy-check-check&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=entropy-check-check&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=entropy-check-check&lookback=7)

### DIFF
--- a/packages/suite/src/actions/settings/deviceSettingsActions.ts
+++ b/packages/suite/src/actions/settings/deviceSettingsActions.ts
@@ -98,7 +98,41 @@ export const resetDevice =
             getState(),
             Feature.entropyCheck,
         );
+
+        // todo: this should be handled most likely in a component somewhere above
+        if (device?.status === 'used' || device?.status === 'occupied') {
+            const features = await TrezorConnect.getFeatures({ device: { path: device.path } });
+            if (!features.success) {
+                dispatch(
+                    notificationsActions.addToast({
+                        type: 'error',
+                        error: 'Device is unreadable',
+                    }),
+                );
+
+                return;
+            }
+            if (features.payload.initialized) {
+                // todo: decide what should happen next UX wise. At the moment, user only gets an error toast
+                // and stays stuck on the 'create new wallet' screen;
+                dispatch(
+                    notificationsActions.addToast({
+                        type: 'error',
+                        error: 'This device has already been initialized',
+                    }),
+                );
+
+                return;
+            }
+        }
+
         if (!device || !device.features) return;
+
+        if (device.mode !== 'initialize') {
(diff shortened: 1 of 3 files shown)


packages/suite/src/actions/settings/deviceSettingsActions.ts
```diff
@@ -98,26 +98,54 @@ export const resetDevice =
             getState(),
             Feature.entropyCheck,
         );
+
+        // todo: this should be handled most likely in a component somewhere above
+        if (device?.status === 'used' || device?.status === 'occupied') {
+            const features = await TrezorConnect.getFeatures({ device: { path: device.path } });
+            if (!features.success) {
+                // todo: not sure how to handle? some toast?
```
Lemonexe: How about this:
```suggestion
            dispatch(
                    notificationsActions.addToast({
                        type: 'error',
                        error: 'Device is unreadable',
                    }),
                );
```

packages/suite/src/actions/settings/deviceSettingsActions.ts
```diff
@@ -98,26 +98,54 @@ export const resetDevice =
...
+                return;
+            }
+            if (features.payload.initialized) {
+                // todo: decide what should happen next UX wise. At the moment, user only gets an error toast
```
Lemonexe, line 116: yes, that shoud happen upstream in the onboarding screen – should not call `resetDevice` but rather goto next step.
I will start working on it (among other cases).

In that case, the toast can stay, because it is not expected it will come up.

Lemonexe (approved):
LGTM, just one suggestion.

packages/suite/src/actions/settings/deviceSettingsActions.ts
Lemonexe: Done.

Issue #21173, opened 2025-08-29
Can't change/remove pin in mobile app


MiroslavProchazka:
https://www.notion.so/satoshilabs/Can-t-change-remove-pin-in-mobile-app-25edc5260606801dbb74e070fab7855c?source=copy_link

hynek-jina:
I wasn't able to reproduce that on the newest build

STew790:
**QA OK**
Cant reproduce. as well

PR #21191, opened 2025-08-30
slice suite-native authorization slice

mroz22:
I have a long-term plan to unify pin/passphrase logic between native and suite as much as possible. Unfortunatelly, native guys were forced into creating something they call `deviceAuthorization` module https://github.com/trezor/trezor-suite/blob/develop/suite-native/device-authorization/src/deviceAuthorizationSlice.ts#L26 because passphrase/pin logic in suite is tightly bundled with its UI representation -> the infamous [modalReducer](https://github.com/trezor/trezor-suite/blob/develop/packages/suite/src/reducers/suite/modalReducer.ts#L69)

I believe that this modalReducer should be decoupled from storing device.events coming from trezor-connect. I am not there yet, but this is a small first step. Not 100% pure refactoring.

## Possibly related 
#21173

## QA notes
- affects both desktop and mobile
- affects anything related to the passphrase and PIN

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=native-slice-authorization-slice&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=native-slice-authorization-slice&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=native-slice-authorization-slice&lookback=7)

### DIFF
--- a/suite-common/connect-init/src/connectInitThunks.ts
+++ b/suite-common/connect-init/src/connectInitThunks.ts
@@ -95,6 +95,7 @@ export const connectInitThunk = createThunk<void, ConnectInitHooks | void, void>
             switch (action.type) {
                 case UI.REQUEST_PIN:
                 case UI.INVALID_PIN:
+                case UI.REQUEST_PASSPHRASE:
                     dispatch(
                         deviceActions.addButtonRequest({
                             // todo: note that this is not 'threadsafe', currently selected device is not necessarily the device
--- a/suite-native/device-authorization/src/deviceAuthorizationSlice.ts
+++ b/suite-native/device-authorization/src/deviceAuthorizationSlice.ts
@@ -38,8 +38,6 @@ export const deviceAuthorizationSlice = createSlice({
         builder
             .addCase(UI.REQUEST_PASSPHRASE, state => {
                 state.hasDeviceRequestedPassphrase = true;
-                // this is not fully substituted by button requests, since REQUEST_PASSPHRASE is not added into device.buttonRequests
-                // state.hasDeviceRequestedPin = false;
             })
             .addCase(UI.REQUEST_BUTTON, (state, action) => {
                 // @ts-expect-error Actions are not typed properly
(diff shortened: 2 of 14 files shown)


suite-common/connect-init/src/connectInitThunks.ts
```diff
@@ -96,6 +96,7 @@ export const connectInitThunk = createThunk<void, ConnectInitHooks | void, void>
             switch (action.type) {
                 case UI.REQUEST_PIN:
                 case UI.INVALID_PIN:
+                case UI.REQUEST_PASSPHRASE:
```
juriczech, line 98: I am not sure whether I like the buttonRequest array and how it's handled. It's very loose implementation and doesn't have good & unified management across the repo (resetting the array for example). We can remove button requests and add them, but you have no guarantee of this logic at the place where you're selecting, which is where I prefer to react to the connect events rather than button requests selectors.

suite-native/device-authorization/src/deviceAuthorizationSlice.ts
```diff
@@ -67,7 +53,6 @@ export const deviceAuthorizationSlice = createSlice({
                 }
             })
             .addCase(UI.CLOSE_UI_WINDOW, state => {
-                state.hasDeviceRequestedPin = false;
```
juriczech, line 70: Touching on what I wrote in my comment earlier, I'm afraid that since we implemented this without the need for button requests, native code will now lack the removal of button requests after success / error and the selector `selectDeviceRequestedPin ` might return true even though it shouldn't. 

Going into the future, I'd consider some kind of "resetting" of this array when reacting `UI.CLOSE_UI_WINDOW` so it's not possible to have a hanging button request, that wasn't deleted manually.

PR #21197, opened 2025-08-31
chore(suite-native): useOverrideBackNavigation refactor

### LINKED ISSUE #19838: useOverrideBackNavigation hook usage
There is a new hook for overriding the default back navigation (system ui back button, screen swipe from the left edge...):  https://github.com/trezor/trezor-suite/blob/ae5fc3209c9d812822cfd624c1d4a59b77d3c4c7/suite-native/navigation/src/hooks/useOverrideBackNavigation.tsx 

Go throug all the places where is a same useEffect as implemented in this hook and replace all the occurences with this new reusable hook.

BrantalikP:
## Description

_note: I omitted https://github.com/trezor/trezor-suite/blob/661a8df9fdc3eed03c56790adeb5a6849a391e56/suite-native/module-send/src/hooks/useHandleOnDeviceTransactionReview.tsx#L61 because the logic is a bit advanced, and I didn’t want to complicate the hook for just one use case._

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/19838

## Screenshots:

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=chore/native/override-back-navigation&lookback=7)

### DIFF
--- a/suite-native/module-device-settings/src/screens/WipeDeviceContinueOnTrezorScreen.tsx
+++ b/suite-native/module-device-settings/src/screens/WipeDeviceContinueOnTrezorScreen.tsx
@@ -16,8 +16,7 @@ export const WipeDeviceContinueOnTrezorScreen = () => {
     const shouldFactoryResetBeVisible = useSelector(selectShouldFactoryResetBeVisible);
 
     useDeviceConnectionGuard();
-
-    useHandleHardwareBackNavigation(() => TrezorConnect.cancel());
+    useHandleHardwareBackNavigation(TrezorConnect.cancel);
 
     const device = useSelector(selectSelectedDevice);
 
--- a/suite-native/module-send/src/screens/SendOutputsReviewScreen.tsx
+++ b/suite-native/module-send/src/screens/SendOutputsReviewScreen.tsx
@@ -1,7 +1,5 @@
-import { useEffect } from 'react';
+import { useCallback, useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
-
-import { useNavigation } from '@react-navigation/native';
 
 import {
     AccountsRootState,
@@ -12,13 +10,12 @@ import { Box, VStack } from '@suite-native/atoms';
 import { ConfirmOnTrezorWrapper, useConfirmOnTrezorController } from '@suite-native/device';
 import { Translation } from '@suite-native/intl';
 import {
-    RootStackParamList,
     ScreenHeader,
     SendStackParamList,
     SendStackRoutes,
     StackProps,
-    StackToStackCompositeNavigationProps,
     useNavigateToInitialScreen,
+    useOverrideBackNavigation,
 } from '@suite-native/navigation';
 import { prepareNativeStyle, useNativeStyles } from '@trezor/styles';
 
@@ -27,11 +24,6 @@ import { ReviewOutputItemList } from '../components/ReviewOutputItemList';
 import { useShowReviewCancellationAlert } from '../hooks/useShowReviewCancellationAlert';
 import { selectIsTransactionAlreadySigned } from '../selectors';
 
-type NavigationProps = StackToStackCompositeNavigationProps<
-    SendStackParamList,
-    SendStackRoutes.SendOutputsReview,
-    RootStackParamList
->;
 const spacerStyle = prepareNativeStyle(_ => ({
     height: 150,
 }));
@@ -40,7 +32,6 @@ export const SendOutputsReviewScreen = ({
     route,
 }: StackProps<SendStackParamList, SendStackRoutes.SendOutputsReview>) => {
     const { accountKey, tokenContract } = route.params;
-    const navigation = useNavigation<NavigationProps>();
     const showReviewCancellationAlert = useShowReviewCancellationAlert();
     const navigateToInitialScreen = useNavigateToInitialScreen();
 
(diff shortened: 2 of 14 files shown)


juriczech (approved):

suite-native/navigation/src/hooks/useOverrideBackNavigation.tsx
```diff
@@ -2,7 +2,9 @@ import { useEffect } from 'react';
 
 import { useNavigation } from '@react-navigation/native';
 
-export const useOverrideBackNavigation = ({ onNavigateBack }: { onNavigateBack?: () => void }) => {
+export const useOverrideBackNavigation = ({
+    onNavigateBack,
+}: { onNavigateBack?: () => void } = {}) => {
```
juriczech: Actually I'm thinking whether it would make sense to use route matching for this and if the screen that it is intended for is focused. Since some of our screens don't unmount, it would be nice that this would only work on the actual screen where this hook is initialized?

juriczech (changes requested):

suite-native/navigation/src/hooks/useOverrideBackNavigation.tsx
BrantalikP: I’m not the creator of this hook, but I think it should be the responsibility of the developer who decides to use it, and they should be aware of the behavior. If we need a fully mistake-proof solution, I’d actually rather refrain from using it or what do you think? @PeKne

juriczech: If I'm not mistaken, once it mounts, the back button will not work for the rest of the app's lifecycle. Can you give me an example of when this might be a good idea and wanted? 

I can't think of anything and it feels like unwanted functionality to have elsewhere than the screen where it's used. Regardless of who wrote it, I think it would be a nice improvement. Saying it's developers responsibility is one thing, of course we can look at it this way, but I'd much rather make our code to be as easy to use as possible and this will IMO make it simpler and more mistake proof.

juriczech:
We should also remove `useHandleHardwareBackNavigation` and replace it with this hook since we're refactoring this. `useOverrideBackNavigation` takes care of any `navigation.goBack` / swipe back / back button handler and that's what we want in all places that use `useHandleHardwareBackNavigation`

suite-native/module-device-settings/src/screens/WipeDeviceContinueOnTrezorScreen.tsx
```diff
@@ -7,13 +7,13 @@ import {
...
 } from '@suite-native/device';
-import { Screen, ScreenHeader, useHandleHardwareBackNavigation } from '@suite-native/navigation';
+import { Screen, ScreenHeader, useOverrideBackNavigation } from '@suite-native/navigation';
 import TrezorConnect from '@trezor/connect';
 
 export const WipeDeviceContinueOnTrezorScreen = () => {
     const shouldFactoryResetBeVisible = useSelector(selectShouldFactoryResetBeVisible);
 
-    useHandleHardwareBackNavigation(() => TrezorConnect.cancel());
+    useOverrideBackNavigation({ onNavigateBack: () => TrezorConnect.cancel() });
```
juriczech: Total nitpick 🙈 

```suggestion
    useOverrideBackNavigation({ onNavigateBack: TrezorConnect.cancel });
```

juriczech (commented):
Looks great, thanks for the refactor! 🫶 

There's few other screens that could use this hook too:
- SendOutputsReviewScreen
- ReceiveScreenHeader

suite-native/module-device-settings/src/screens/WipeDeviceContinueOnTrezorScreen.tsx
BrantalikP: good catch, next time I will read it more carefully https://github.com/trezor/trezor-suite/pull/21197/commits/831641c9926c601ef27d7dd57f2d9e18fa9f2634

BrantalikP:
> Looks great, thanks for the refactor! 🫶
> 
> There's few other screens that could use this hook too:
> 
> * SendOutputsReviewScreen
> * ReceiveScreenHeader

Nice! 

https://github.com/trezor/trezor-suite/pull/21197/commits/831641c9926c601ef27d7dd57f2d9e18fa9f2634

Applied for `SendOutputsReviewScreen`


regarding `ReceiveScreenHeader` it does something a bit different instead of overwriting the navigation it just cancel Trezor actions when the screen is unfocused, it could be written differently and simplified but this approach is not wrong as well, then I am keeping it as is if its ok 👍

suite-native/module-send/src/screens/SendOutputsReviewScreen.tsx
```diff
@@ -57,24 +48,16 @@ export const SendOutputsReviewScreen = ({
 
     const showOutputsReviewFooter = isTransactionAlreadySigned && account;
 
-    useEffect(() => {
-        const unsubscribe = navigation.addListener('beforeRemove', async e => {
-            // We want to modify only behavior of back button actions.
-
-            if (e.data.action.type !== 'GO_BACK') return;
+    const onNavigateBack = async () => {
```
juriczech: This should be callback (`useCallback`) because it's used as a dependency in `useEffect`

BrantalikP: you are right! Although I am thinking rather remove the deps from the useEffect and instead of `[navigation, onNavigateBack, gestureEnabled]` having just `[]`. I see no point of having it as deps. WDYT

BrantalikP: we want to run it just once in the end.

juriczech: I would keep the deps there and just make the memoization work as expected. That's better IMO.

BrantalikP: I prefer to believe my intuition rather than eslint but here you it is https://github.com/trezor/trezor-suite/pull/21197/commits/f054d5faf93996ccb4cd3748fac824f9a1f8ead0 🙏 thanks for review!

juriczech: I think since the callbacks are being called within useFocusEffect & useCallback combination, which should, IMO, behave predictably, it's best to memoize it as it's described in the docs 🤷 😄

juriczech (approved):
LGTM, thank you for doing the refactor! 💪

PR #21204, opened 2025-09-01
Native connect calls race

mroz22:
connect calls fired from here https://github.com/trezor/trezor-suite/blob/develop/suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts#L12
could unexpectedly interfere with app by removing button requests here http://github.com/trezor/trezor-suite/blob/develop/suite-common/connect-init/src/connectInitThunks.ts#L155
and possibly also elsewhere. what is the point of using mutex if not used for all connect calls? 🤔 

I think I got hit by this in https://github.com/trezor/trezor-suite/pull/21191 where I am trying to unify some logic between suite ant native

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=native-connect-calls-race&lookback=7)

### DIFF
--- a/suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
+++ b/suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
@@ -1,6 +1,7 @@
 import { useEffect } from 'react';
 import { useSelector } from 'react-redux';
 
+import { requestDeviceAccess } from '@suite-native/device-mutex';
 import TrezorConnect, { FIRMWARE } from '@trezor/connect';
 import { TimerId } from '@trezor/type-utils';
 import { isArrayMember } from '@trezor/utils';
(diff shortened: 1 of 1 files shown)


suite-native/device/src/hooks/useRetryFwAuthenticityChecks.ts
```diff
@@ -19,10 +21,16 @@ export const useRetryFwAuthenticityChecks = () => {
     useEffect(() => {
         let timeoutHandle: TimerId;
         const recheckFwRevision = () => {
+            if (deviceAccessMutex.taskQueue.length > 0) {
+                // refrain from scheduling multiple tasks (since this runs in a loop)
+                return;
+            }
```
Lemonexe: But this means it will never be rechecked just because it is busy at this point.
I think that the task queue is busy, it should only skip this check and schedule next one, like so:

```javascript
        const recheckFwRevision = () => {
            if (isRetriableError) {
                // refrain from scheduling multiple tasks (since this runs in a loop)
                if (deviceAccessMutex.taskQueue.length === 0) {
                    // any device call will cause the tests to be rerun, so getFeatures is used as the most basic one
                    // it'd be useless to await the result; what interests us is the Device state that updates, and gets propagated into redux
                    requestDeviceAccess({
                        deviceCallback: () =>
                            TrezorConnect.getFeatures({ useEmptyPassphrase: true }),
                    });
                }
                timeoutHandle = setTimeout(recheckFwRevision, REFRESH_INTERVAL);
            }
        };
```

mroz22: oh exactly!

Lemonexe (approved):
:upside_down_face:

PR #21218, opened 2025-09-01
fix(suite-desktop-core): ignore missing connect process info for linux

martykan:
## Description

@Ondra-Zik-SL found that it's not possible to load process info under AppImage/FlatPak, causing Connect in Suite to not work
This PR allows skipping the check for Linux. 

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/linux-process-info&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/linux-process-info&lookback=7)

### DIFF
--- a/packages/suite-desktop-core/src/libs/connect-ws.ts
+++ b/packages/suite-desktop-core/src/libs/connect-ws.ts
@@ -10,7 +10,7 @@ import {
     PopupHandshake,
     parseConnectSettings,
 } from '@trezor/connect';
-import { isMacOs, isWindows } from '@trezor/env-utils';
+import { isLinux, isMacOs, isWindows } from '@trezor/env-utils';
 import { ProcessInfo, findProcessFromIncomingPort } from '@trezor/node-utils';
 import { ConnectPopupResponse } from '@trezor/suite-desktop-api/src/messages';
 import { Deferred, createDeferred, resolveAfter } from '@trezor/utils';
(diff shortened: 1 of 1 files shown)


packages/suite-desktop-core/src/libs/connect-ws.ts
```diff
@@ -210,12 +213,14 @@ export const exposeConnectWs = ({
...
-                        icon: await getProcessIcon(processOnPort.fullPath),
-                    },
+                    process: processOnPort
+                        ? {
+                              name: processOnPort.name,
+                              fullPath: processOnPort.fullPath,
+                              warning: !!processOnPort.warning,
+                              icon: await getProcessIcon(processOnPort.fullPath),
+                          }
+                        : undefined,
```
mroz22, line 223: wasn't this a 'security' feature? to present users with process ? 🤔

martykan: Yes, but unfortunately it's not possible to get info on the other process with Flatpak sandboxing.

mroz22: and this is solved in UI in some clever way right? no info -> warning, or something like this.

mroz22 (approved):

packages/suite-desktop-core/src/libs/connect-ws.ts
martykan: Yeah there is a warning label, no extra detailed explanation, but I would say it's ok, it doesn't affect that many users. For example Ubuntu should probably be fine

PR #21220, opened 2025-09-01
fix(suite-desktop-core): getComputerName fallback

martykan:
## Description

@Ondra-Zik-SL found that `hostnamectl` called from `getComputerName` might not be available on all Linux distros. 
We need to catch the error and provide a fallback. 

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/getcomputername-fallback&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/getcomputername-fallback&lookback=7)

### DIFF
--- a/packages/suite-desktop-core/src/libs/info.ts
+++ b/packages/suite-desktop-core/src/libs/info.ts
@@ -3,7 +3,7 @@ import { app } from 'electron';
 import os from 'os';
 
 import { isDevEnv } from '@suite-common/suite-utils';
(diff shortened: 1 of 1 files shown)


Ondra-Zik:
Also not all systems have `Pretty Hostname` so `hostnamectl --pretty` can return empty string in some cases

mroz22 (approved):

PR #21232, opened 2025-09-02
Fix: reset device edge cases

Lemonexe:
## Description

- ignore entropy check failure case `'Device is used in another window'`
  - that can only come from transport AFAIK
- cleanup interface of resetDevice, split into two thunks, proper notifications, commentary 

## QA instructions

Try going through device onboarding with two sessions at once: see the situations described in Screenshots.
If you find a new situation that's not handled well please create a new issue :pray: 

## Screenshots:

Case A:
- In session "left", go through onboarding up to creating a wallet but don't do it.
- In session "right", create a wallet.
 - Now try to continue with creating wallet in session "left":
- [Before and after](https://github.com/user-attachments/assets/46c27e56-4e5a-48c7-88fa-b7f59e3101ab) (unchanged behavior)

Case B:

- In session "left", start creating a wallet but don't confirm in device.
- In session "right", start creating a wallet (cannot, device is occupied).
- In session "left", the creation of wallet has failed, but entropy check has not failed!
- **Before**: [entropy check fail](https://github.com/user-attachments/assets/27e88d26-029f-4148-94e3-cf37276897a3) when interrupting resetDevice on-device confirmation with another request
- **After**: [no entropy check fail](https://github.com/user-attachments/assets/a317de37-3348-4587-be79-e180d3d2c573).

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/resetDevice-edge-cases&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/resetDevice-edge-cases&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/resetDevice-edge-cases&lookback=7)

### DIFF
--- a/packages/suite/src/actions/settings/deviceSettingsActions.ts
+++ b/packages/suite/src/actions/settings/deviceSettingsActions.ts
@@ -8,7 +8,7 @@ import TrezorConnect, { ERRORS } from '@trezor/connect';
 import * as modalActions from 'src/actions/suite/modalActions';
 import * as DEVICE from 'src/constants/suite/device';
 import { selectIsEntropyCheckEnabled } from 'src/selectors/suite/suiteSelectors';
-import { Dispatch, GetState } from 'src/types/suite';
+import { AcquiredDevice, Dispatch, GetState } from 'src/types/suite';
 
 export const applySettings =
     (params: Parameters<typeof TrezorConnect.applySettings>[0]) =>
@@ -89,50 +89,21 @@ export const changeWipeCode =
         }
     };
 
+export type ResetDeviceParams = Parameters<typeof TrezorConnect.resetDevice>[0];
+
+type ResetDeviceProps = {
+    params: ResetDeviceParams;
+    device: AcquiredDevice;
+};
+
 export const resetDevice =
-    (params: Parameters<typeof TrezorConnect.resetDevice>[0] = {}) =>
+    ({ params, device }: ResetDeviceProps) =>
     async (dispatch: Dispatch, getState: GetState) => {
-        const device = selectSelectedDevice(getState());
         const isEntropyCheckEnabled = selectIsEntropyCheckEnabled(getState());
         const isEntropyCheckDisabledByMessageSystem = selectIsFeatureDisabled(
             getState(),
             Feature.entropyCheck,
         );
-
-        // todo: this should be handled most likely in a component somewhere above
-        if (device?.status === 'used' || device?.status === 'occupied') {
-            const features = await TrezorConnect.getFeatures({ device: { path: device.path } });
-            if (!features.success) {
-                dispatch(
-                    notificationsActions.addToast({
-                        type: 'error',
-                        error: 'Device is unreadable',
-                    }),
-                );
-
-                return;
-            }
-            if (features.payload.initialized) {
-                // todo: decide what should happen next UX wise. At the moment, user only gets an error toast
-                // and stays stuck on the 'create new wallet' screen;
-                dispatch(
-                    notificationsActions.addToast({
-                        type: 'error',
-                        error: 'This device has already been initialized',
-                    }),
-                );
-
-                return;
-            }
-        }
-
-        if (!device || !device.features) return;
-
-        if (device.mode !== 'initialize') {
-            console.error('resetDevice: device in invalid mode', device.mode);
-
-            return;
-        }
 
         const defaults = {
             strength: DEVICE.DEFAULT_STRENGTH[device.features.internal_model],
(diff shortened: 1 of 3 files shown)


packages/suite/src/actions/settings/deviceSettingsActions.ts
```diff
@@ -89,51 +89,22 @@ export const changeWipeCode =
...
+    ({ params, device }: ResetDeviceProps) =>
     async (dispatch: Dispatch, getState: GetState) => {
-        const device = selectSelectedDevice(getState());
         const isEntropyCheckEnabled = selectIsEntropyCheckEnabled(getState());
         const isEntropyCheckDisabledByMessageSystem = selectIsFeatureDisabled(
             getState(),
             Feature.entropyCheck,
         );
 
-        // todo: this should be handled most likely in a component somewhere above
```
Lemonexe, line 102: At first I wanted to redo this functionality in `ResetDevice.tsx > onResetDevice`, but then realized that @mroz22  implemented it already, but without the skip functionality. Which I did with a callback.
I think it's better to have this live in thunk rather than a component useCallback. But I at least split it into two thunks for clarity.

packages/suite/src/actions/settings/deviceSettingsActions.ts
```diff
@@ -89,51 +89,22 @@ export const changeWipeCode =
...
-            }
-            if (features.payload.initialized) {
-                // todo: decide what should happen next UX wise. At the moment, user only gets an error toast
-                // and stays stuck on the 'create new wallet' screen;
-                dispatch(
-                    notificationsActions.addToast({
-                        type: 'error',
-                        error: 'This device has already been initialized',
-                    }),
-                );
```
Lemonexe: this is the part that is changed, rest of the code is moved.

packages/suite/src/actions/settings/deviceSettingsActions.ts
```diff
@@ -157,6 +128,45 @@ export const resetDevice =
...
+            }
+            if (response.payload.initialized) {
+                alreadyInitializedCallback();
+
+                return;
+            }
+        }
+
+        // This should never occurr because it is handled by prerequisites:
+        // Onboarding renders the proper UI instead of rendering ResetDevice UI
```
Lemonexe, line 163: comment added here: this is already handled by the `device-unaquired` prerequisite.

packages/suite/src/actions/settings/deviceSettingsActions.ts
```diff
@@ -157,6 +128,45 @@ export const resetDevice =
...
+            const response = await TrezorConnect.getFeatures({ device: { path: device.path } });
+            if (!response.success) {
+                dispatch(
+                    notificationsActions.addToast({ type: 'error', error: 'Device is unreadable' }),
+                );
+
+                return;
+            }
+            if (response.payload.initialized) {
+                alreadyInitializedCallback();
```
komret: I don't think we should do this. Getting stuck on this screen in this edge case is tolerable IMO. The problem with this solution is that the behaviour is quite unexpected and it makes one think that a new wallet was created while it is in fact an old wallet. Also, it just shifts the problem further because maybe they already did the backup/pin step in another window and then we'd have to skip those steps, too. So I'd simply just show the toast.

packages/suite/src/actions/settings/deviceSettingsActions.ts
```diff
@@ -157,6 +128,45 @@ export const resetDevice =
...
+                return;
+            }
+        }
+
+        // This should never occurr because it is handled by prerequisites:
+        // Onboarding renders the proper UI instead of rendering ResetDevice UI
+        if (!device || !device.features) return;
+
+        if (device.mode !== 'initialize') {
+            console.error('resetDevice: device in invalid mode', device.mode);
```
komret, line 174: We should also show the toast here, otherwise nothing happens on a second click on the button.

suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -527,6 +527,7 @@ export const failEntropyCheckThunk = createThunk(
             'A transfer error has occurred',
             // This one was not in Sentry but can be triggered by manually disconnecting the device. Investigate. https://github.com/trezor/trezor-suite-private/issues/135
             'device disconnected during action',
+            'Device is used in another window',
```
komret, line 530: Why is this needed? This error does not trigger entropy check failure AFAIK.

komret (changes requested):

packages/suite/src/actions/settings/deviceSettingsActions.ts
Lemonexe: Ok. Looks like my PR is basically useless after all. ~I will close it and create another one just for a new toast (comment below) and code reorganization.~ EDIT: I'll just remove the `alreadyInitializedCallback`

suite-common/wallet-core/src/device/deviceThunks.ts
Lemonexe: Ah, yes, I see now that we have not received such error in Sentry. But I reproduced Entropy check failure with this message and thought it was one of the cases we are investigating. Nevertheless, I think we should handle it this way.
[entropy check fail interrupting RESET DEVICE itself.webm](https://github.com/user-attachments/assets/a317de37-3348-4587-be79-e180d3d2c573)

packages/suite/src/actions/settings/deviceSettingsActions.ts
Lemonexe: Done.

komret:
Sorry for butchering this, @Lemonexe :( I am closing it in favour of https://github.com/trezor/trezor-suite/pull/21325. I did not split the thunk there because I think it is no longer needed now that we do not skip anymore.

Issue #19262, opened 2025-06-02
Evolu Lableing


peter-sanderson:
Figma: https://www.figma.com/design/wYZKr3GKb9JFpgAefmuwfb/Labels?node-id=201-2&p=f&t=XTGZtuFEAVLpu3xN-0

- [x] Every passphrase encrypts data with own secret
 - [x] secret from Trezor to Evolu (from the right one)
-----------------
- [x] Wallets Labels
- [x] Account Labels
- [x] Address Labels 
- [x] Output Labels 
-----------------
- [x] debug settings feature-flag to switch labeling to Evolu
    - [x] use it
- [ ] data migration from old labeling to Evolu (no way back)
- [x] Settings: custom Evolu relay URL
    - [x] use it
- [ ] Adminkos: Run our own Evolu relay 

-----------------
### Crypto audit
- [x] NanoId conversion, will it work this way?
- [x] derive mnemonic, will it work this way?
- [x] do we need derive key for every account? It was done this way in original labeling, but to me it seems unnecessary => whole wallet instance (device+passphrase) is encrypted with one key

hynek-jina:
Let's track the rest in dedicated issues...

STew790:
Will be tracked in each issue.

PR #21245, opened 2025-09-02
feat: Local First Storage (Evolu) for mobile [6]

peter-sanderson:
Part of: https://github.com/trezor/trezor-suite/issues/19262

Implement labeling (in Evolu) for Native

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=evolu_native&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=evolu_native&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=evolu_native&lookback=7)


suite-native/labeling/src/components/AccountLabel.tsx
```diff
@@ -0,0 +1,32 @@
+import { useSelector } from 'react-redux';
+
+import { WithLabelingState, selectAccountLabel } from '@suite-common/local-first-storage';
+import { AccountKey, WalletDescriptor } from '@suite-common/wallet-types';
+import { Text } from '@suite-native/atoms';
+
+import { useIsLabelingEnabled } from './useIsLabelingEnabled';
+
+type AccountLabelEProps = {
```
5K1PY, line 9: Is the `E` intentional?

suite-native/module-send/src/sendFormThunks.ts
```diff
@@ -272,3 +281,91 @@ export const calculateCustomFeeLevelThunk = createThunk(
...
+>(
+    `${SEND_MODULE_PREFIX}/sendTransactionThunk`,
+    async (
+        { selectedAccount, wasAppLeftDuringReview, tokenContract },
+        { dispatch, getState, rejectWithValue, fulfillWithValue },
+    ) => {
+        const sendResponse = await dispatch(pushSendFormTransactionThunk({ selectedAccount }));
+
+        if (sendResponse.payload === undefined) {
+            throw new Error('Todo: wtf? shall not be undefined');
```
5K1PY: TODO

suite-native/transactions/src/components/TransactionDetail/TransactionDetailAddressesSection.tsx
```diff
@@ -146,11 +153,17 @@ export const TransactionDetailAddressesSection = ({
...
-                                        value={address}
-                                        style={applyStyle(addressTextStyle)}
+                                {changeAddresses.map(({ address, outputIndex }) => (
+                                    <TransactionUtxoAddress
+                                        key={`change:${outputIndex}`}
+                                        address={address}
+                                        outputIndex={outputIndex}
+                                        deviceStaticSessionId={transaction.deviceState}
+                                        txId={transaction.txid}
+                                        // Todo: input not implemented yet. The idea is, that transaction input is just output
```
5K1PY, line 161: TODO

suite-native/transactions/src/components/TransactionDetail/TransactionDetailAddressesSection.tsx
```diff
@@ -105,12 +104,20 @@ export const TransactionDetailAddressesSection = ({
...
-                            />
+                        {targetAddresses.slice(0, 2).map(({ address, outputIndex }) => (
+                            <>
+                                <TransactionUtxoAddress
+                                    key={`target:${outputIndex}`}
+                                    address={address}
+                                    outputIndex={outputIndex}
+                                    deviceStaticSessionId={transaction.deviceState}
+                                    txId={transaction.txid}
+                                    // Todo: input not implemented yet. The idea is, that transaction input is just output
```
5K1PY: TODO

peter-sanderson:
@SocketSecurity ignore npm/@evolu/react-native@12.0.1-preview.4

.depcheckrc.json
```diff
@@ -13,7 +13,9 @@
         "protobufjs-cli",
         "buffer",
         "node-gyp",
-        "version-bump-prompt"
+        "version-bump-prompt",
+        "expo",
```
peter-sanderson: Peer dependency of Evolu

suite-native/app/src/initActions.ts
```diff
@@ -58,5 +59,8 @@ export const postOnboardingInit = createThunk(
         if (selectIsFeatureFlagEnabled(getState(), FeatureFlag.IsWalletConnectEnabled)) {
             dispatch(walletConnectInitThunk());
         }
+
+        // This needs to be initialized to subscribe to all Remembered Wallets
```
peter-sanderson: `Todo:` no longer true, is called in `selectDeviceThunk`, make sure it works same as on desktop

peter-sanderson:
@SocketSecurity ignore npm/expo-sqlite@15.2.14

peter-sanderson:
@SocketSecurity ignore npm/expo-sqlite@15.2.14

suite-native/module-trading/src/components/general/AccountList/__tests__/AccountListAddressItem.test.tsx
```diff
@@ -14,6 +13,39 @@ jest.mock('@suite-common/wallet-core', () => {
     };
 });
 
+const createAccount = (
+    values: Pick<Account, 'key' | 'symbol' | 'accountLabel' | 'availableBalance'>,
+): Account => ({
+    ...values,
```
vytick: `...values` probably should be after the rest of params

peter-sanderson: fixed

suite-native/feature-flags/src/useToggleFeatureFlag.ts
```diff
@@ -1,9 +1,27 @@
...
 
-    return () => dispatch(toggleFeatureFlag({ featureFlag }));
+    // const originalFlagState = useSelector((state: FeatureFlagsRootState) =>
+    //     selectIsFeatureFlagEnabled(state, FeatureFlag.IsLocalFirstStorageEnabled),
+    // );
+
+    return () => {
+        dispatch(toggleFeatureFlag({ featureFlag }));
+
+        // if (featureFlag === FeatureFlag.IsLocalFirstStorageEnabled) {
```
peter-sanderson: `Todo`: Solve this mess

suite-native/transactions/src/utils.ts
```diff
@@ -39,11 +40,12 @@ export const mapTransactionInputsOutputsToAddresses = ({
 export const sortTargetAddressesToBeginning = (
     addresses: readonly VinVoutAddress[],
     targetAddresses: readonly VinVoutAddress[],
-) =>
-    A.concat(
+) => [
+    ...A.concat(
```
peter-sanderson: This removes the need for `as`. The goal is to remove the `readonly` constraint.

suite-native/transactions/src/tests/utils.test.ts
```diff
@@ -100,8 +116,16 @@ describe('sortTargetAddressesToBeginning', () => {
         });
 
         const expectedResult: VinVoutAddress[] = [
-            { address: '3QpCQP3A2q7kCr8QgsWuqG1Bg1P6RySonw', isChangeAddress: false },
-            { address: '3BcXPstZ4ZHhvLxPFkjFocuFySKt8nsGgs', isChangeAddress: false },
+            {
+                address: '3BcXPstZ4ZHhvLxPFkjFocuFySKt8nsGgs',
```
peter-sanderson, line 120: Note: here is a change of the order of the addresses.

suite-native/transactions/src/components/TransactionDetail/NetworkTransactionDetailSummary.tsx
```diff
@@ -54,37 +51,26 @@ export const VerticalSeparator = ({ inputsCount }: VerticalSeparatorProps) => {
...
-            selectTransactionAddresses(state, accountKey, txid, 'inputs'),
+            selectTransactionAddresses(state, accountKey, transaction.txid, 'inputs'),
     );
     const transactionOutputAddresses = useSelector(
         (state: TransactionsRootState & TokenDefinitionsRootState) =>
-            selectTransactionAddresses(state, accountKey, txid, 'outputs'),
+            selectTransactionAddresses(state, accountKey, transaction.txid, 'outputs'),
     );
 
-    if (!transaction) {
```
peter-sanderson, line 78: Transaction is now required by type in Props, this shall not happen now

suite-native/module-trading/src/components/general/AccountList/__tests__/AccountListAddressItem.test.tsx
```diff
@@ -91,50 +135,36 @@ describe('AccountListAddressItem', () => {
...
+                received: '',
+            },
         };
         const { getByLabelText } = await renderAccountListAddressItem(receiveAccount);
 
         expect(getByLabelText('Balance in fiat')).toHaveTextContent('$0.00');
         expect(getByLabelText('Balance in crypto')).toHaveTextContent('0 BTC');
     });
 
-    it('should not display balance when address has no balance', async () => {
```
peter-sanderson, line 111: Balance is required parameter of the `address`. In case no `address` is provided -> there is another test covering it. If Address is provided, type ensures that `balance` is there.

suite-native/module-send/src/sendOutputsFormSchema.ts
```diff
@@ -104,125 +104,120 @@ const hasEnoughBalanceForFees = (context?: SendFormFormContext) => {
     return amountBigNumber.gt(networkFeeInfo.minFee);
 };
 
-export const sendOutputsFormValidationSchema = yup.object({
```
peter-sanderson, line 107: I have only extracted the `outputs` fields into separate definition to get its type. 

Here it quite messy as the form logic requires to have same types as our data model, see the type-check below.

(btw its a madness, .... I hate forms)

suite-native/module-send/src/components/OutputsReviewFooter.tsx
```diff
@@ -143,29 +128,15 @@ export const OutputsReviewFooter = ({
         setIsSendInProgress(true);
 
         const sendResponse = await dispatch(
-            pushSendFormTransactionThunk({
```
peter-sanderson, line 146: This is shared code for broadcasting transaction. In the desktop, we have another thunk wrapping this to do additional stuff. 

Nothing like this existed here on mobile, so I have to introduce it => behold: `sendTransactionThunk` :grin:

suite-native/module-send/src/sendFormThunks.ts
peter-sanderson: fixed

peter-sanderson:
@SocketSecurity ignore npm/@evolu/react@9.0.1-preview.4

suite-native/transactions/src/components/TransactionDetail/TransactionDetailAddressesSection.tsx
peter-sanderson: This todo is meant to stay. Currently we are not labeling the the UTXO inputs (only outputs) but we will. So this is here to explain that.

suite-native/transactions/src/components/TransactionDetail/TransactionDetailAddressesSection.tsx
peter-sanderson: This todo is meant to stay. Currently we are not labeling the the UTXO inputs (only outputs) but we will. So this is here to explain that.

suite-native/module-dev-utils/package.json
```diff
@@ -36,7 +36,7 @@
         "@trezor/styles": "workspace:*",
         "@trezor/theme": "workspace:*",
         "@trezor/urls": "workspace:*",
-        "expo": "53.0.20",
+        "expo": "^53.0.22",
```
karliatto, line 39: Why adding `^` if it was not there? Is not better to keep to fixed version?

packages/connect-examples/mobile-expo/package.json
```diff
@@ -10,7 +10,7 @@
     },
     "dependencies": {
         "@trezor/connect-mobile": "workspace:*",
-        "expo": "53.0.20",
+        "expo": "^53.0.22",
```
karliatto, line 13: Why adding ^ if it was not there? Is not better to keep to fixed version?

suite-native/video-assets/package.json
```diff
@@ -13,7 +13,7 @@
     },
     "dependencies": {
         "@trezor/styles": "workspace:*",
-        "expo": "53.0.20",
+        "expo": "^53.0.22",
```
karliatto, line 16: Why adding ^ if it was not there? Is not better to keep to fixed version?

.depcheckrc.json
```diff
@@ -13,7 +13,9 @@
...
+        "expo-sqlite"
```
PeKne: This will ignore these packages globally for all the packages. Please look for some better solution.

packages/connect-examples/mobile-expo/package.json
peter-sanderson: I am simply used to use `^`. I have asked @matejkriz why we don't have this and got the answer that `^` is ok.

The issue was, that `evolu` uses `53.0.22` and by requiring exact version, the yarn could not update it and hoist it in the root `node_modules`. This actually caused an ugly error with imports. (Because I had to do this shit -> https://github.com/trezor/trezor-suite/blob/develop/suite-native/app/metro.config.js#L75)

suite-native/transactions/src/components/TransactionName.tsx
```diff
@@ -57,8 +58,14 @@ export const TransactionName = ({ transaction, isPending }: TransactionNameProps
 
     // use name of eth txns, but not for recv or sent Transfer
     if (ethName) {
-        return ethName;
+        return <Text variant="body">{ethName}</Text>;
     }
 
-    return <Translation id={getTransactionName(transaction, isPending)} />;
+    return (
```
PeKne, line 61: The header is broken. Why was wrapping in Text and HStack necessary?
Screenshot 2025-09-12 at 12 49 43 PM

packages/connect-examples/mobile-expo/package.json
karliatto: Ok, good enough. :diki:

suite-native/app/src/initActions.ts
```diff
@@ -58,5 +59,7 @@ export const postOnboardingInit = createThunk(
         if (selectIsFeatureFlagEnabled(getState(), FeatureFlag.IsWalletConnectEnabled)) {
             dispatch(walletConnectInitThunk());
         }
+
```
PeKne, line 62: It is not change of this PR, but why do you store the localFirstStorageProvider to the redux store? Redux is designed to store serializable data only, you are storing instance of class there. IMO it would be enough to create the instance as a global `const` and read the reference from there.

suite-native/app/src/initActions.ts
```diff
@@ -58,5 +59,7 @@ export const postOnboardingInit = createThunk(
...
+        await dispatch(initNativeLocalFirstStorageThunk());
```
PeKne: Does this need to be awaited? In case it would be taking longer for any reason (e.g. timeout) the app would not load until then. I suggest to remove the await. Is there any async call used inside of the thunk anyway?

suite-native/app/src/initActions.ts
PeKne: more here: https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions

suite-native/device-manager/src/components/WalletItem.tsx
```diff
@@ -1,19 +1,18 @@
...
 type WalletItemProps = {
     onPress: () => void;
-    deviceState: NonNullable<TrezorDevice['state']>;
+    device: TrezorDevice;
     isSelectable?: boolean;
 };
 
-export const WalletItem = ({ onPress, deviceState, isSelectable = true }: WalletItemProps) => {
-    const device = useSelector((state: any) => selectDeviceByState(state, deviceState));
+export const WalletItem = ({ onPress, device, isSelectable = true }: WalletItemProps) => {
```
PeKne, line 15: what is the benefit of passing the whole device instead of deviceState?

suite-native/feature-flags/src/useToggleFeatureFlag.ts
```diff
@@ -1,9 +1,27 @@
...
+import { useDispatch, useSelector } from 'react-redux';
 
-import { FeatureFlag, toggleFeatureFlag } from './featureFlagsSlice';
+import { disposeAllLocalFirstStorageThunk } from '@suite-common/local-first-storage';
+import { initNativeLocalFirstStorageThunk } from '@suite-native/local-first-storage';
+
+import { selectIsFeatureFlagEnabled } from './featureFlagsSelectors';
+import { FeatureFlag, FeatureFlagsRootState, toggleFeatureFlag } from './featureFlagsSlice';
 
 export const useToggleFeatureFlag = (featureFlag: FeatureFlag): (() => void) => {
```
PeKne, line 5: This hook is used for all the feature flags. Please remove all local-first-storage related logic from here.

suite-native/labeling/src/components/AccountLabel.tsx
PeKne: typo?
```suggestion
type AccountLabelProps = {
```

suite-native/labeling/src/components/AccountLabel.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export const AccountLabel = ({
+    accountKey,
+    walletDescriptor,
+    fallbackLabel,
+}: AccountLabelEProps) => {
+    const isLabelingEnabled = useIsLabelingEnabled();
+
+    const label = useSelector((state: WithLabelingState) =>
```
PeKne, line 23: nit: please create a selector that will return accountLabel.label specifically

suite-native/labeling/src/components/AddLabelButtonText.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { Icon, iconSizes } from '@suite-native/icons';
+import { Translation } from '@suite-native/intl';
+
+type AddLabelButtonTextProps = {
+    onPress: () => void;
+    justifyContent?: FlexStyle['justifyContent'];
+    label?: ReactNode;
+};
+
+export const AddLabelButtonText = ({ onPress, justifyContent, label }: AddLabelButtonTextProps) => (
```
PeKne: This is IMO duplication of TextButton component. Use that instead please https://github.com/trezor/trezor-suite/blob/1e4e8307b917a356a66aab9b54e857c27791e95e/suite-native/atoms/src/Button/TextButton.tsx

suite-native/labeling/src/components/AddressLabelEditable.tsx
```diff
@@ -0,0 +1,65 @@
...
+        (state: WithLabelingState) =>
+            (address !== undefined
+                ? selectAddressLabel({
+                      state,
+                      address,
+                      deviceStaticSessionId,
+                  })
+                : null
+            )?.label ?? null,
+    );
```
PeKne: this is not a good practice. Please create new selector where you pass `address` and `deviceStaticSessionId` as an argument and it calls `selectAddressLabel` inside.

suite-native/labeling/src/components/AddressLabelEditable.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <EditableLabelLayout label={label} justifyContent="center">
+            {({ onClose }) => (
+                <LabelEditForm
+                    label={label ?? ''}
+                    onSubmit={newLabel => {
+                        onSubmit(newLabel);
+                        onClose();
+                    }}
+                />
+            )}
```
PeKne, line 58: How does this work? Where does the `onClose` come from? Would not be simplier to create another component of this?

PeKne: calling children as `children()` is probably not necessary.

suite-native/labeling/src/components/LabelEditForm.tsx
```diff
@@ -0,0 +1,47 @@
...
+        onSubmit(formValues.label);
+    });
+
+    return (
+        <VStack spacing="sp16">
+            <Form form={form}>
+                <VStack spacing="sp8">
+                    <TextInputField ref={inputRef} name="label" asBottomSheetInput />
+                    <Button onPress={onConfirm} size="large" isDisabled={!isValid}>
+                        Confirm
```
PeKne: use generic.buttons.confirm translation.

suite-native/labeling/src/components/TransactionOutputLabel.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+    const label =
+        useSelector((state: WithLabelingState) =>
+            selectOutputLabel({
+                state,
+                txId,
+                outputIndex,
+                deviceStaticSessionId,
+            }),
+        )?.label ?? null;
```
PeKne: again, please create specific selector please.

suite-native/module-accounts-management/src/components/AccountDetailScreenHeader.tsx
```diff
@@ -25,51 +24,38 @@ type AccountDetailNavigationProps = StackToStackCompositeNavigationProps<
...
             </Text>
         </HStack>
     );
 };
 
-export const AccountDetailScreenHeader = ({
-    accountLabel,
-    accountKey,
-}: AccountDetailScreenHeaderProps) => {
+export const AccountDetailScreenHeader = ({ account }: AccountDetailScreenHeaderProps) => {
```
PeKne, line 44: nit: again, why is necessary to pass the whole account object? It was selected from selector before.

suite-native/labeling/src/components/TransactionOutputLabelEditable.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    if (!isLabelingEnabled) {
+        return null;
+    }
+
+    return (
+        <EditableLabelLayout label={label}>
+            {({ onClose }) => (
+                <LabelEditForm
+                    label={label ?? ''}
```
PeKne, line 45:
```suggestion
                    label={label}
```

and return '' from selector in case that label do not exists.

suite-native/module-send/src/components/AddressInput.tsx
```diff
@@ -73,6 +77,12 @@ export const AddressInput = ({ index, accountKey }: AddressInputProps) => {
                         DEV: self address
                     </Button>
                 )}
+                <SendFormLabelEditable
+                    label={utxoLabel ?? null}
```
PeKne, line 81: could be default value `null`, so you do not have to assign it here?

suite-native/module-send/src/sendFormThunks.ts
```diff
@@ -272,3 +281,104 @@ export const calculateCustomFeeLevelThunk = createThunk(
...
+            });
+        }
+
+        if (!('success' in sendResponse.payload)) {
+            return rejectWithValue(sendResponse.payload);
+        }
+
+        const formValues = selectSendFormDraftByKey(getState(), selectedAccount.key, tokenContract);
+
+        // Todo: This code below is kinda copy-paste from `applySendFormMetadataLabelsThunk` in Desktop.
```
PeKne: what about extracting this "copy-pasted" code to the specific labeling thunk that will be called here? It would improve the readability a lot. Because now is here more labeling code than code related to the sending the transaction itself.

suite-native/module-trading/src/components/general/AccountList/AccountListAddressItem.tsx
```diff
@@ -22,14 +25,24 @@ export const AccountListAddressItem = ({
     const { applyStyle } = useNativeStyles();
     const { address } = receiveAccount;
 
+    const addressLabel = useSelector((state: WithLabelingState) =>
```
PeKne, line 28: again, create specific selector for this

suite-native/transactions/src/utils.ts
PeKne: F.toMutable is what you are looking for:

```
F.toMutable(
        A.concat(
            A.intersection(addresses, targetAddresses),
            A.difference(addresses, targetAddresses),
        ),
    );
```

suite-native/transactions/src/components/TransactionDetail/TransactionUtxoAddress.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    fallback={
+                        <AccountAddressFormatter
+                            key={address}
+                            value={address}
+                            style={applyStyle(addressTextStyle)}
+                        />
+                    }
+                />
+
+                {isDebugEnv() && <Text>[{outputIndex}]</Text>}
```
PeKne: Does this gonna stay?

suite-native/transactions/src/components/TransactionDetail/TransactionDetailAddressesSection.tsx
```diff
@@ -105,12 +104,20 @@ export const TransactionDetailAddressesSection = ({
...
                             />
                         </Text>
-                        {targetAddresses.slice(0, 2).map(({ address }) => (
-                            <AccountAddressFormatter
-                                key={address}
-                                value={address}
-                                style={applyStyle(addressTextStyle)}
-                            />
+                        {targetAddresses.slice(0, 2).map(({ address, outputIndex }) => (
+                            <>
```
PeKne: Probably redundant React fragment (<></>)

PeKne (commented):
The connection between desktop and mobile works flawlessly! Well done. Most of my comments are requests for styling and code quality improvements. Feel free to ignore those that will be resolved in a follow-up.

.depcheckrc.json
peter-sanderson: :grin: see https://github.com/trezor/trezor-suite/pull/21520

PeKne:
There are some overflow issue
IMG_7563
IMG_7561

suite-native/transactions/src/utils.ts
peter-sanderson: :+1: Fixed

.depcheckrc.json
peter-sanderson: Fixed

suite-native/transactions/src/components/TransactionName.tsx
peter-sanderson: Dunno, some leftover from the past I guess. Nice catch.

`Text` is however IMO needed `Translation` without it cannot be used in the native code without Text wrapper so I unified it.

peter-sanderson: I have removed the `Text` here: https://github.com/trezor/trezor-suite/pull/21245/files#diff-2543216142c2519eba3ab8adc1fa0e1f93c2d19bd0f635495621e24bc911d533L157

I would intuitively expect that component can be rendered everywhere. Component that fails (because it returns the `string` not wrapped in `Text` seems bad to me.

suite-native/app/src/initActions.ts
peter-sanderson: Godd catch, true, I think I can remove `await` from desktop as well. I think there used to be a async code that I have removed and this is leftover of that

suite-native/device-manager/src/components/WalletItem.tsx
peter-sanderson: - Technically because the `WalletItemBase` needs a `device?.walletNumber`

- Philosophically, I kinda feel we shall pass `device` more often. It is a value that we already have in memory all the time and it makes code much more straight-forward. Of course, not to pass it to component where we need one parameter to render some `name` or something. But In higher level component (Wallet/Layout/Item/...) I think we shall keep it simple. Refactoring, UI changes and everything is then much simpler when we have a relevant `device` on hand.

peter-sanderson: Yes, I can pass just the `walletNumber` and the `state` as 2 params. But for the component named `WalletItem` that represents whole Wallet .. I kinda (see philosophical part) feel it shall get whole `device` (which is in fact the wallet entity on this context).

suite-native/feature-flags/src/useToggleFeatureFlag.ts
peter-sanderson: Fixed: https://github.com/trezor/trezor-suite/pull/21245/commits/6756986d69559489200a365d08aaf706c97c869c

suite-native/labeling/src/components/AccountLabel.tsx
peter-sanderson: fixed

suite-native/labeling/src/components/AddLabelButtonText.tsx
peter-sanderson: Oh, nice, thanks! Fixed

peter-sanderson: https://github.com/trezor/trezor-suite/pull/21245/commits/a6e6fba35543e416896a363a46e91cbf488d4f3e

suite-native/labeling/src/components/TransactionOutputLabel.tsx
peter-sanderson: fixed

suite-native/transactions/src/components/TransactionDetail/TransactionDetailAddressesSection.tsx
peter-sanderson: fixed: https://github.com/trezor/trezor-suite/pull/21245/commits/bd95525cad5293e8ef4efacce71c325ef9a1c72f

(solves the key prop unique issue)

suite-native/labeling/src/components/TransactionOutputLabel.tsx
peter-sanderson: I changed all the label selectors to return `string|null` and those issues are gone. No need for special selectors after all

suite-native/labeling/src/components/AddressLabelEditable.tsx
peter-sanderson: I changed all the label selectors to return `string|null` and those issues are gone. No need for special selectors after all

suite-native/labeling/src/components/AddressLabelEditable.tsx
peter-sanderson: This is a simple higher-order component. It allows to pass a `onClose` from shared layout code (originally from the `useBottomSheetModal` hook in that  `EditableLabelLayout`.

suite-native/labeling/src/components/LabelEditForm.tsx
peter-sanderson: fixed

suite-native/module-accounts-management/src/components/AccountDetailScreenHeader.tsx
peter-sanderson: As before there was this re-select just to get the symbol:
image

This is used only in `AccountDetailContentScreen` -> `AccountDetailScreen`. For such a component tree (Account Detail) it makes sense to me just select the account on the higher (top) level and pass it down. This reduces the complexity of the code, does not require re-selects here-and-there and is much more future-proof. 

Yes it is kinda philosophical but it makes much more sense to me. We can call a discuss this more.

suite-native/labeling/src/components/TransactionOutputLabelEditable.tsx
peter-sanderson: I kinda don't like this. I designed the labeling (evolu) in a way, that `<no label>` is represented by a `null`. There is explicit conversion (see the `suite-common/local-first-storage/src/labeling/normalizeLabel.ts`).

Here I need to put a default value into "HTML" `input` so this is IMO the correct place to convert `null` (no label) to `''` for the input component.

suite-native/module-send/src/components/AddressInput.tsx
peter-sanderson: image

I have no idea where are the `defaults` defined. This is dreaded react-hook-form, I have no clue. Also, it is not type-safe somehow, so I really fear touching anything there.

suite-native/module-trading/src/components/general/AccountList/AccountListAddressItem.tsx
peter-sanderson: fixed by changing the `selectAddressLabel` so it accepts undefined for address

suite-native/module-send/src/sendFormThunks.ts
peter-sanderson: Extracted to solo thunk.

suite-native/transactions/src/components/TransactionDetail/TransactionUtxoAddress.tsx
peter-sanderson: Id like to keep it (at least for a while, it helps with debugging and it is visible only in `isDebugEnv()`

suite-native/labeling/src/components/AccountLabel.tsx
peter-sanderson: fixed by normalizing selectors to work with null

suite-native/transactions/src/components/TransactionName.tsx
PeKne: this is still not correct. The word `transaction` is styled as `<Text variant="highlight"...` but the dynamic prefix is now styled as variant `body` (It comes from the ScreenHeader > ScreenHeaderContent component) . Because of that is highlighted only part of the text. 

Image

suite-native/transactions/src/components/TransactionDetail/TransactionUtxoAddress.tsx
PeKne: Ok got it. Could be the condition changed in a way, that this is rendered only if the labeling FF is enabled? 🙏

suite-native/labeling/src/components/TransactionOutputLabelEditable.tsx
PeKne: ok, lets keep it as it is.

suite-native/module-accounts-management/src/components/AccountDetailScreenHeader.tsx
PeKne: It might lead to performance issues. If you are using only part of of account, e.g. `account.symbol`, then the component will rerender only if this specific prop will change. The other changes of account will be ignored. 

In this case it does not matter, because the whole `account` object is selected in the parent. So the `AccountDetailScreenHeader` would need to be memoized. But I believe that in general it is a good practice to always select only the relevant part of the state.  What do you think?

PeKne: Same applies for WalletItem conversation. I closed that thread.

suite-native/transactions/src/components/TransactionName.tsx
peter-sanderson: fixed

suite-native/app/src/initActions.ts
peter-sanderson: discussed in person, not in redux but gloval variable

suite-native/labeling/src/components/AddressLabel.tsx
```diff
@@ -0,0 +1,35 @@
...
+type AddressLabelEProps = {
+    address: string;
+    deviceStaticSessionId: StaticSessionId;
+    fallback: ReactNode;
+};
+
+export const AddressLabel = ({ address, deviceStaticSessionId, fallback }: AddressLabelEProps) => {
+    const isLabelingEnabled = useIsLabelingEnabled();
+
+    const label = useSelector(
```
PeKne: This selector is still not refactored.

PeKne (approved):
One more nit. Please fix this as well. Otherwise looks good. Approving :+1:

suite-native/labeling/src/components/AddressLabel.tsx
peter-sanderson: fixed

PR #21251, opened 2025-09-02
fix(connect): isCompatible should be true when same version

karliatto:
## Description
The check that compares versions to be compatible `isCompatible` should be valid if they are newer or equal not only newer.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/is-compatible-release-should-accept-compatible-as-equal&lookback=7)

### DIFF
--- a/packages/connect/src/data/firmwareInfo.ts
+++ b/packages/connect/src/data/firmwareInfo.ts
@@ -303,12 +303,12 @@ export const getLanguage = (languageBinPath: string) => {
     return httpRequest(url, 'binary');
 };
 
-const getCurrentVersion = (
-    features: Features,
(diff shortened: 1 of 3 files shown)


martykan (approved):

zako05:
Seems to be OK 👍 

* [X] TS5 fresh: FW - install latest vresion (universal 2.8.10) [web]
* [X] TS5 (with already installed FW): FW - update latest version  (universal 2.8.1) [web, desktop: macOS]
* [X] TS3 fresh: FW - install latest version (universal 2.8.10)  [web, desktop: mackOS]

PR #21310, opened 2025-09-04
fix(suite): show only connected selected devices

### LINKED ISSUE #21297: Unexpected pairing modal after device list modal is closed
Unexpected behaviour when pairing modal is closed. It closes the pairing list but reopens the connection modal. I think the tall connection modal should never reopen when the list is closed.

OriginalEveres:
For selected device connect modal - show only not connected or connecting devices.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/21297

## Screenshots:

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/connected-device-do-not-show&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/connected-device-do-not-show&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/connected-device-do-not-show&lookback=7)

### DIFF
--- a/packages/suite/src/components/connection/BluetoothConnectionModal.tsx
+++ b/packages/suite/src/components/connection/BluetoothConnectionModal.tsx
@@ -18,7 +18,10 @@ type BluetoothConnectionModalProps = {
     onRescanClick: () => void;
     onConnect: (deviceId: string) => Promise<void>;
     onCancel: () => void;
+    onClose: () => void;
 };
+
+const selectedDeviceConnectionTypes = ['connecting', 'pairing'];
 
 export const BluetoothConnectionModal = ({
     devices,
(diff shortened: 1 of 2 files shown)


komret (changes requested):
I think we can use this but it does not fully address the problem. If there is another Trezor in pairing mode nearby, you'll run into the same unexpected behaviour - closing the modal will open another modal.

As for filtering out the connected Trezor, it makes sense for now, but we might want to return it in future and enable disconnecting via this view.

packages/suite/src/components/connection/BluetoothConnectionModal.tsx
komret, line 1: I wonder why you did not use it on line 125 as well? When does that scenario occur?

OriginalEveres: because it is a state in which you can't find you desired device, thus I believe you should be able to go back, so that you can click on the "Dont see your trezor bubble"

komret (approved):
Check the failing test pls.

PR #21320, opened 2025-09-05
fix(suite): Show base currency amount even for bitcoin base currency

### LINKED ISSUE #21288: Send form: When sending BTC and base currency is BTC, fiat amout is not shown
Current state:

Image

Expected behaviour: Show fiat amount (and probably use defaultly USD?)

5K1PY:
Display base currecny amount even when current currency == base curency. This allows users to see value in fiat/other valuables if they want to.

## Related Issue

Resolve #21288 

## Screenshots:

Before:

Image

After:
image

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/bitcoin-show-base-currency&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/bitcoin-show-base-currency&lookback=7)

### DIFF
--- a/packages/suite/src/views/wallet/send/Outputs/Amount/Amount.tsx
+++ b/packages/suite/src/views/wallet/send/Outputs/Amount/Amount.tsx
@@ -1,6 +1,5 @@
 import { formInputsMaxLength } from '@suite-common/validators';
 import { getNetworkDisplaySymbol } from '@suite-common/wallet-config';
-import { useDisplayBaseCurrency } from '@suite-common/wallet-core';
 import { Output, TokenAddress } from '@suite-common/wallet-types';
 import {
     convertAmountUnitsToSubunits,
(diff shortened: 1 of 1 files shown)


Lemonexe (approved):
Does what the issue specifies, but I'd wait for @hynek-jina if this is really desirable.
I think it'd be better to switch it: remove "crypto input" and leave only "fiat input" when `baseCurrency === BTC`

hynek-jina:
Good point @Lemonexe , I agree that it's even better to show only baseCurrency field (that can be changed to any currency) and when user chose different currency than is the account currency, then show both.

Lemonexe (approved):
Perfect :rocket:

PR #21323, opened 2025-09-05
fix(suite): runaway cycle of THP acquire errors

### LINKED ISSUE #21303: THP reconnection deadlock
suite keeps reconnecting THP device when cancelled

https://github.com/user-attachments/assets/b30ded9b-6b6b-4428-a8b5-6467cc9e4f18

Lemonexe:
## Description

Fixes an endless loop of acquire errors caused by missing dependency list in `useEffect`. 

## Related Issue

Resolve #21303

## Screenshots:

Replicating the steps I described [in the issue coment](https://github.com/trezor/trezor-suite/issues/21303#issuecomment-3258622306), and now there is only one error toast as it should:

[AFTER.webm](https://github.com/user-attachments/assets/ff565459-6c8b-46c0-8fda-34324a2cf9ad)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/runaway-device-acquire-errors&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/runaway-device-acquire-errors&lookback=7)

### DIFF
--- a/packages/suite/src/components/suite/PrerequisitesGuide/DeviceTrezorHostProtocolPair.tsx
+++ b/packages/suite/src/components/suite/PrerequisitesGuide/DeviceTrezorHostProtocolPair.tsx
@@ -1,4 +1,4 @@
-import { MouseEventHandler, useEffect } from 'react';
+import { MouseEventHandler } from 'react';
 
 import { acquireDevice } from '@suite-common/wallet-core';
 import { Button } from '@trezor/components';
@@ -31,13 +31,6 @@ export const DeviceTrezorHostProtocolPair = () => {
     );
 
     const tips = [TROUBLESHOOTING_TIP_CLOSE_ALL_TABS, TROUBLESHOOTING_TIP_RECONNECT];
-
-    // if possible, prompt for THP pairing right away
-    useEffect(() => {
-        if (!isDeviceLocked) {
-            dispatch(acquireDevice({ requestedDevice: device }));
-        }
-    });
 
     return (
         <TroubleshootingTips


packages/suite/src/components/suite/PrerequisitesGuide/DeviceTrezorHostProtocolPair.tsx
```diff
@@ -32,13 +32,6 @@ export const DeviceTrezorHostProtocolPair = () => {
 
     const tips = [TROUBLESHOOTING_TIP_CLOSE_ALL_TABS, TROUBLESHOOTING_TIP_RECONNECT];
 
-    // if possible, prompt for THP pairing right away
-    useEffect(() => {
-        if (!isDeviceLocked) {
-            dispatch(acquireDevice({ requestedDevice: device }));
-        }
-    });
```
Lemonexe, line 40: Eslint react-hooks doesn't care if deps are missing entirely.  

At first I thought I would just add deps here, well except the `device`, because it's precisely the device change that is causing the component to rerender :warning: , so I'd add this: 
```
        // eslint-disable-next-line react-hooks/exhaustive-deps
    }, [isDeviceLocked, dispatch]);
```
But the useEffect is actually useless.
I traced this when you connect TS7:
1. [connectInitThunks](https://github.com/trezor/trezor-suite/blob/1c6b8cc4ed9f240cb9703f2dc612f49c6f14f56a/suite-common/connect-init/src/connectInitThunks.ts#L65)
2. [deviceConnectThunks](https://github.com/trezor/trezor-suite/blob/75298784b7537d954657d880af9c9aabe2a0625c/suite-common/wallet-core/src/device/deviceThunks.ts#L420)
3. [autoInitThpAfterDeviceConnectionThunk](https://github.com/trezor/trezor-suite/blob/fea37d02c30e18fdcda21a6779db9c80294b96b6/suite-common/thp/src/autoInitThpAfterDeviceConnectionThunk.ts#L27)
4. `acquireDevice`

And until this happens, the device is not yet in redux, so we don't know it's a THP device, so the `DeviceTrezorHostProtocolPair` component is not rendered! Instead it's `ConnectDevicePrompt`.
→ this component renders only _after_ the first acquireDevice, so a second call is useless.

Also I don't like that important stuff like this should be autonomously fired from a react component useEffect. Important redux stuff that happens without user interaction should be controlled by redux.

Lemonexe:
unstake test [confirmed to be broken](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1757096669575989)

packages/suite/src/components/suite/PrerequisitesGuide/DeviceTrezorHostProtocolPair.tsx
OriginalEveres: good catch, I did add it there (not sure, why without deps, shame on me) because it triggers the THP modal. Looking at it again, it does not make sense to dispatch the aquireDevice, so I approve this PR and I'll create a new one right away to open the THP modal in the thunk.

OriginalEveres (approved):
good catch, I did add it there (not sure, why without deps, shame on me) because it triggers the THP modal. Looking at it again, it does not make sense to dispatch the aquireDevice, so I approve this PR and I'll create a new one right away to open the THP modal in the thunk.

PR #21330, opened 2025-09-07
Fix: intermediary and compatible

### LINKED ISSUE #21267: Cannot update FW on devices with BL older then 1.12.0
Cannot update FW on devices with BL older then 1.12.0

**Info:**
 - Suite version: desktop 25.9.1 (3e5133fe89674211936afd7db19e62914598b471)
 - Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuite/25.9.1 Chrome/138.0.7204.185 Electron/37.2.6 Safari/537.36
 - OS: Linux x86_64
 - Screen: 3440x1440
 - Device: Trezor T1B1 1.11.2 universal (revision 0d87b55ba4fed7eecc72bf2a94ee473830b095e9)
 - Transport: BridgeTransport 3.1.0

karliatto:
## Description

It should properly select intermediaries and it should offer to update to T1B1 devices with FW version that requires intermediary.

I also add some test cases to make sure this stays solid and refactor it a bit for better readability.

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/21267

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/intermediary-and-compatible&lookback=7)

### DIFF
--- a/packages/connect/src/data/firmwareInfo.ts
+++ b/packages/connect/src/data/firmwareInfo.ts
@@ -351,37 +351,32 @@ const getCurrentVersion = (features: Features): CurrentVersion => {
 };
 
 const getIntermediaryMessageRelease = (features: Features) => {
-    const { internal_model } = features;
-    const firmwareIntermediaryReleasesConfig = DataManager.getFirmwareIntermediaryReleaseConfig();
-
-    if (!firmwareIntermediaryReleasesConfig) {
+    const config = DataManager.getFirmwareIntermediaryReleaseConfig();
+    if (!config) {
         throw new Error('Firmware release config not loaded.');
     }
 
-    const deviceIntermediaryReleases = firmwareIntermediaryReleasesConfig[internal_model];
-    if (!deviceIntermediaryReleases) {
-        // There are not intermediary releases for this model.
+    const deviceIntermediaryReleases = config[features.internal_model];
+    if (!deviceIntermediaryReleases || deviceIntermediaryReleases.length === 0) {
+        // No intermediary releases are defined for this model.
+        return;
+    }
+
+    const { bootloaderVersion, firmwareVersion } = getCurrentVersion(features);
+
+    const currentVersion = features.bootloader_mode ? bootloaderVersion : firmwareVersion;
+    const minVersionKey = features.bootloader_mode
+        ? 'min_bootloader_version'
+        : 'min_firmware_version';
+
+    if (!currentVersion) {
         return undefined;
     }
-    let intermediary;
 
-    if (features.bootloader_mode) {
-        const { bootloaderVersion } = getCurrentVersion(features);
-        intermediary = deviceIntermediaryReleases.find(
-            inter =>
-                bootloaderVersion &&
-                versionUtils.isNewer(inter.min_bootloader_version, bootloaderVersion),
-        );
-    } else {
-        const { firmwareVersion } = getCurrentVersion(features);
-        intermediary = firmwareVersion
-            ? deviceIntermediaryReleases.find(inter =>
-                  versionUtils.isNewer(inter.min_firmware_version, firmwareVersion),
-              )
-            : undefined;
-    }
-
-    return intermediary || undefined;
+    // Find the first intermediary release that requires a newer version than the current one.
+    return deviceIntermediaryReleases.find(release =>
+        versionUtils.isNewer(release[minVersionKey], currentVersion),
+    );
 };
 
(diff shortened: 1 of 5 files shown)


packages/connect/src/data/firmwareInfo.ts
```diff
@@ -351,37 +351,32 @@ const getCurrentVersion = (features: Features): CurrentVersion => {
...
-                  versionUtils.isNewer(inter.min_firmware_version, firmwareVersion),
-              )
-            : undefined;
+
+    const { bootloaderVersion, firmwareVersion } = getCurrentVersion(features);
+
+    const currentVersion = features.bootloader_mode ? bootloaderVersion : firmwareVersion;
+    const minVersionKey = features.bootloader_mode
+        ? 'min_bootloader_version'
+        : 'min_firmware_version';
```
marekrjpolak, line 370: Definitely better readable than before :+1:

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -589,6 +597,13 @@ export const getFirmwareLocation = ({
     const isRealBundled = !bundledBaseUrl.includes('data.trezor.io');
     const bundledVersion = getBundledFirmwareVersion(deviceModel, firmwareType);
 
+    if (intermediaryVersion) {
+        return {
+            baseUrl: bundledBaseUrl,
+            path: `firmware/${deviceModel.toLowerCase()}/${firmwareName}`,
+        };
+    }
+
```
marekrjpolak: From what I understand, shouldn't this be part of the condition below? 

I mean, the return value is the same, `bundledBaseUrl && isRealBundled` part of the condition could be valid here as well, and the last part can be simplified to something like `(intermediaryVersion || (bundledVersion === versionString)` ? Also, `versionString` can be always `firmwareVersion.join('.')` as in intermediary case it's unused.

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -497,40 +496,49 @@ export const getReleaseInfo = ({
...
+        );
+        // If an alternative is found, use it. Otherwise, we proceed with the original.
+        if (alternativeRelease) {
+            suitableRelease = alternativeRelease;
+        }
+    }
 
     const intermediary = getIntermediaryMessageRelease(features);
-    const isBitcoinOnlyAvailable = getIsBitcoinOnlyAvailable(features);
+    const finalTargetRelease = intermediary ? release : suitableRelease;
```
marekrjpolak, line 536: This ternary condition can be probably moved above, as in `if (!isCompatible && !intermediary) {...` ? Then, `suitableRelease` can be used directly instead of `finalTargetRelease`.

marekrjpolak: Also, to be sure: This means that when intermediary firmware is needed to install, the last known release is always advertised even if incompatible with the current version?

packages/connect/src/data/firmwareInfo.ts
```diff
@@ -480,6 +475,10 @@ export const getReleaseInfo = ({
     const { rollout_probability } = conditions;
     const shouldBeOffered = calculateShouldOfferRelease(rollout_probability, features.device_id);
 
+    if (requiresIntermediary && intermediary) {
+        isNewer = true;
+    }
```
marekrjpolak, line 480: To be sure: This means that whenever an intermediary firmware is needed to install, it is signalled that there's update available?

karliatto: Yes.

packages/connect/src/data/firmwareInfo.ts
karliatto: I see,you are right, what about this https://github.com/trezor/trezor-suite/pull/21330/commits/1ec33d267528014f485706a209de1629aebb2a6f ?

packages/connect/src/data/firmwareInfo.ts
karliatto: Yes, that is what I understand, when `intermediary` FW we use latest available since the intermediary should allow to install to latest FW.

karliatto: I think that should be the case for T1B1, but if we use intermediary FWs more often in the future we might need to add more information to those intermediary FW like if there is any limit to what they make possible to update.

marekrjpolak (approved):
Seems to work :+1:

PR #21331, opened 2025-09-07
feat: firmware release config bundled in JSON

karliatto:
## Description

This PR is changing the FW release config that is bundled to be in clear JSON instead of a JSON that contains a JWS
 that has to be decoded and we can verify.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/local-relase-config-in-json&lookback=7)

### DIFF
--- a/packages/connect/src/utils/firmwareReleaseConfigUtils.ts
+++ b/packages/connect/src/utils/firmwareReleaseConfigUtils.ts
@@ -15,120 +15,95 @@ const JWS_CONFIG = {
 
 type JwsInfo = {
     jws: string;
-    source: 'remote' | 'local';
     env: FirmwareUpdateSource;
 };
 
-const fetchRemoteJWS = async (): Promise<JwsInfo> => {
+const fetchRemoteJws = async (): Promise<JwsInfo> => {
     const { BASE_URL, MIDDLE_PATH, env } = getOnlineFirmwareBaseUrl();
-
-    const remoteReleasesUrl = new URL(BASE_URL);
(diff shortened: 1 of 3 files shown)


packages/connect/src/utils/firmwareReleaseConfigUtils.ts
```diff
@@ -15,120 +15,93 @@ const JWS_CONFIG = {
...
+        throw new Error('Invalid JWS algorithm');
+    }
 
-const verifyJwsSignature = (jws: string, publicKey: string): void => {
-    const decoded = decode(jws);
-    if (decoded?.header.alg !== JWS_CONFIG.SIGN_ALGORITHM) {
-        throw new Error(
-            `Invalid JWS algorithm: expected ${JWS_CONFIG.SIGN_ALGORITHM}, got ${decoded?.header.alg}`,
-        );
+    if (decoded.payload.version !== JWS_CONFIG.VERSION) {
```
matejkriz: decoded.payload is not parsed here, so it will always fail here

matejkriz: easy fixup [`fcad851` (#21331)](https://github.com/trezor/trezor-suite/pull/21331/commits/fcad85189e5d1ce4ab45de42891c48390e531f65) (but I screw it and fixup wrong commit, would be nice to fix it during rebase)

matejkriz (approved):

PR #21349, opened 2025-09-08
fix(connect): current release for only remote fws

karliatto:
## Description

We were getting current release in Device only from bundled ones and we need it also for FWs that are only remote for example in case that you want to change language of current FW we need the information of the available languages and the path to obtain them.

## Related Issue

Resolve 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/current-release-for-only-remote-fws&lookback=7)

### DIFF
--- a/packages/connect/src/device/Device.ts
+++ b/packages/connect/src/device/Device.ts
@@ -19,7 +19,12 @@ import { checkFirmwareRevision } from './checkFirmwareRevision';
 import { abortThpWorkflow, getThpChannel } from './thp';
 import { checkFirmwareHashWithRetries } from './workflow/checkFirmwareHashWithRetries';
 import { getAllNetworks } from '../data/coinInfo';
-import { getFirmwareReleaseConfigInfo, getFirmwareStatus, getLanguage } from '../data/firmwareInfo';
+import {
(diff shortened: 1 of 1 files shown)


packages/connect/src/device/Device.ts
```diff
@@ -656,6 +662,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
     async getFeatures() {
         const { message } = await this.getCurrentSession().typedCall('GetFeatures', 'Features', {});
+        await this._updateCurrentRelease(message);
```
karliatto: Basically main reason we need to have it here is because in `_updateFeatures` we do:

```
            this.availableTranslations = this._currentRelease?.translations ?? {};

```

And as it was done before it was geting only releases of bundled ones, but if we use it for a FW which is only provided in a remote way we need to fetch it. 

So `getReleaseByVersion` that is used inside `_updateCurrentRelease` gets the bundled one if present so it does not have to fetch anything in case of desktop but if it is not present in the bundled then it fetches it.

marekrjpolak (approved):
I believe it's safe. :+1:

PR #21351, opened 2025-09-08
chore(connect-common): remove unused releases 2.9.0

karliatto:
## Description

Remove unused releases 2.9.0, it is not an issue since they are not included in the index https://github.com/trezor/trezor-suite/blob/develop/packages/connect-common/files/firmware/index.ts#L49  But exactly because of that is not necessary to have them there and they could even create a potential issue in the future.


MiroslavProchazka:
@karliatto  Keep in mind that some users have Trezor with this firmware installed. Since it was officially released, even though it was rolled back after a few days, we need to keep it public for transparency and for all security checks that rely on it. Just FYI.

karliatto:
> @karliatto Keep in mind that some users have Trezor with this firmware installed. Since it was officially released, even though it was rolled back after a few days, we need to keep it public for transparency and for all security checks that rely on it. Just FYI.

It will still be in data.trezor.io but it was not included in the bundled before I did the migration to new format of releases JSON so I consider it as a misatake. But I have not problem to add them there, just right now they are there but never used because they are not indexed, so I would remove them or add them properly.

mroz22 (approved):

PR #21353, opened 2025-09-08
ci(suite-native): crowdin translations tooling

### LINKED ISSUE #21146: Localization - integration of Crowdin to suite-native
Setup tooling around Crowdin to work the same way on mobile as it does for desktop.

PeKne:
## Description
- [Crowdin CI workflow](https://github.com/trezor/trezor-suite/pull/21353/files#diff-89281be8697fc5f061aa80d9c0334985bf723b04add08b9c83f9f64a5bdda1fd) was extended to support mobile
  - There is a new manual input that tells if the crowdin synchronization should be executed for a mobile or a desktop project
  - Both desktop and mobile have separate projects in the Crowdin backend and also use different translation source files in the codebase, so the translations of one platform should never influence the other platform
- a few utility scripts that help to extract and synchronize source translations were added

## Issue
closes #21146 

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/native/crowdin-translations-tooling&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/native/crowdin-translations-tooling&lookback=7)


.github/workflows/bot-crowdin-sync.yml
```diff
@@ -3,14 +3,21 @@ name: "[Bot] Crowdin translations update"
...
+      platform:
+        type: choice
+        options:
+          - desktop
+          - mobile
+        description: Should be synced desktop or mobile crowdin project?
+        required: true
       remove_unused_translations:
-        description: "Confirm removal of unused translations. Check them first in misc tests pipeline."
+        description: "Confirm removal of unused translations. Check them first in misc tests pipeline. (works for desktop only)"
```
PeKne, line 14: this will be resolved in the followup #21380

suite-native/intl/src/utils.ts
```diff
@@ -13,3 +13,45 @@ export const flatten = (obj: Record<string, any>, prefix = '') => {
...
+                }
+                current = current[key];
+            }
+        });
+    }
+
+    return result;
+};
+
+export const deepMerge = (target: Record<string, any>, source: Record<string, any>) => {
```
matejkriz: What about using mergeDeepObject from @trezor/utils? Are there any differences?

BrantalikP (approved):

suite-native/intl/scripts/backportEnglish.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * This script takes latest en-US.json and creates new file messages.ts
+ *
+ * Problem ?
+ * - messages.ts is created by developers using "developers English" which might be miles away from proper English
```
BrantalikP: Nit: I’d rephrase this a bit, since some of the translations are also created by our Design and Product teams, so it might not be accurate. I’d suggest something like:

"messages.ts might not reflect the final version of the translations. The en-us.json file contains polished versions approved by our Product and Marketing teams.... When meaning ... "

suite-native/intl/src/utils.ts
PeKne: good point, fixed. Already squashed.

suite-native/intl/scripts/backportEnglish.ts
PeKne: fixed

PR #21401, opened 2025-09-09
Fix: delay discovery during THP autoconnect modal

### LINKED ISSUE #20957: Auto-connect modal opens and the button is disabled (because of discovery)
**Describe the bug**
Auto-connect modal opens and the primary button is disabled (because of discovery). As a user this confuses me

**Steps to reproduce:**
1. Connect a device
2. After the secure pairing, the auto-connect modal should appear
3. The button for approving is disabled while the discovery is going on

**Info:**
 - Suite version: desktop 25.9.0 (0791bf0d705d55a87a016228c366ca30d9e089a1)
 - Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) TrezorSuiteDev/25.9.0 Chrome/138.0.7204.185 Electron/37.2.6 Safari/537.36
 - OS: MacIntel
 - Screen: 1512x982
 - Device: Trezor T3W1 2.9.1 universal (revision ec4123667f852d52caae693e14235e48c9bbc29c)
 - Transport: BluetoothTransport, BridgeTransport 3.1.0

**Screenshots:**

Image

Lemonexe:
## Description

CR commit by commit recommended:

- slight refactoring in discovery to unify some common default functionality to `restartDiscoveryThunk`, now renamed to `startOrRestartDiscoveryThunk`
- defer discovery when THP Autoconnect Modal 

## Related Issue

Resolve https://github.com/trezor/trezor-suite/issues/20957

## Screenshots:

Discovery starts after:
- [succesful THP Autoconnect](https://github.com/user-attachments/assets/24a30d2f-84a4-4604-a610-e9b69c85d3c8)
- [THP Autoconnect cancelled on-device](https://github.com/user-attachments/assets/fe093e2e-c64a-4150-a668-2d8073202a9f)
- [canceled modal](https://github.com/user-attachments/assets/c2aa73ac-d015-4bb0-b0f4-ff1179a695f1)

Note: I tried playing around, if I can fire autoconnect during discovery. But the request gets in queue and again, [you just have a loader on the autoconnect modal](https://github.com/user-attachments/assets/9740a85e-d6f4-41ab-93fc-b8105f80b0dc), and it actually calls after discovery is finished (locks are released). So no.

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/autoconnect-modal-during-discovery&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/autoconnect-modal-during-discovery&lookback=7)

### DIFF
--- a/packages/suite/src/components/connection/thp/ThpAutoconnectInfoModal.tsx
+++ b/packages/suite/src/components/connection/thp/ThpAutoconnectInfoModal.tsx
@@ -1,6 +1,6 @@
 import { useState } from 'react';
 
-import { startThpAutoconnectThunk, thpActions } from '@suite-common/thp';
+import { finishThpAutoconnectThunk, startThpAutoconnectThunk } from '@suite-common/thp';
 import { Column, H3, Modal, Paragraph } from '@trezor/components';
 
 import { useDevice, useDispatch } from '../../../hooks/suite';
@@ -18,7 +18,7 @@ export const ThpAutoconnectInfoModal = () => {
     };
 
     const onCancel = () => {
-        dispatch(thpActions.finishThpFlow());
+        dispatch(finishThpAutoconnectThunk());
     };
 
     return (
--- a/packages/suite/src/middlewares/wallet/discoveryMiddleware.ts
+++ b/packages/suite/src/middlewares/wallet/discoveryMiddleware.ts
@@ -1,6 +1,7 @@
 import { connectPopupCallThunkInner } from '@suite-common/connect-popup';
 import { createMiddlewareWithExtraDeps } from '@suite-common/redux-utils';
 import { isDeviceAcquired } from '@suite-common/suite-utils';
+import { selectThpStep } from '@suite-common/thp';
 import {
     accountsActions,
     changeNetworks,
@@ -49,6 +50,11 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
         const isDeviceReady =
             device?.connected && isDeviceAcquired(device) && (!isDeviceLocked || becomesAcquired);
 
+        // delay discovery if THP Autoconnect modal is open (discovery executed by the modal), as it is the only
+        // THP step that takes place *after* device acquisition, and also needs device interaction to complete.
+        const isTHPAutoconnectModal = selectThpStep(getState()) === 'AutoconnectInfo';
+        const isUIReady = !isTHPAutoconnectModal;
+
         if (
             becomesAcquired ||
             becomesConnected ||
@@ -58,7 +64,7 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
             changeNetworks.match(action) ||
             accountsActions.changeAccountVisibility.match(action)
         ) {
-            if (isDeviceReady) {
+            if (isDeviceReady && isUIReady) {
                 const shouldRediscover = selectShouldRediscover(getState(), device);
                 const noDiscoveryYet = device?.state?.staticSessionId === undefined;
                 if (noDiscoveryYet || shouldRediscover) {
--- a/packages/suite/src/middlewares/wallet/discoveryMiddleware.ts
+++ b/packages/suite/src/middlewares/wallet/discoveryMiddleware.ts
@@ -5,10 +5,9 @@ import {
     accountsActions,
     changeNetworks,
     deviceActions,
-    runAdditionalDiscoveryThunk,
     selectSelectedDevice,
     selectShouldRediscover,
-    startDiscoveryThunk,
+    startOrRestartDiscoveryThunk,
 } from '@suite-common/wallet-core';
 
 import { SUITE } from 'src/actions/suite/constants';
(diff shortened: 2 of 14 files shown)


packages/suite/src/components/connection/thp/ThpAutoconnectInfoModal.tsx
```diff
@@ -20,6 +21,8 @@ export const ThpAutoconnectInfoModal = () => {
 
     const onCancel = () => {
         dispatch(thpActions.finishThpFlow());
+        // see discoveryMiddleware, discovery is delayed until THP AutoConnect is closed either way
+        dispatch(startOrRestartDiscoveryThunk());
```
Lemonexe: in order to delay the discovery, I blocked it in `discoveryMiddleware`, but then I have to manually execute it when "the thing we are waiting for" is settled (either cancelled or finished THP autoconnect).

An alternative would be to use a deferred promise, and have the flow defined in middleware (in middleware, wait for the promise, then call the discovery we originally wanted). I don't like bloating middlewares, especially with async logic..

suite-native/discovery/src/discoveryMiddleware.ts
```diff
@@ -72,12 +71,10 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
...
                 isDeviceAcquired(device)
             ) {
-                if (!device.state) {
-                    dispatch(startDiscoveryThunk({}));
-                } else if (device.state.staticSessionId) {
-                    if (selectShouldRediscover(getState(), device)) {
-                        dispatch(restartDiscoveryThunk());
-                    }
+                const shouldRediscover = selectShouldRediscover(getState(), device);
+                const noDiscoveryYet = device?.state?.staticSessionId === undefined;
```
juriczech, line 75: `selectShouldRediscoverHelper` blocks rediscovery for devices without `staticSessionId` (https://github.com/trezor/trezor-suite/blob/9d9513f6498f495962e7b9da4b4597b2f4ba0625/suite-common/wallet-core/src/selectors.ts#L127) 

If we're changing the behavior, wouldn't it be cleaner to remove that condition? 

cc @marekrjpolak

packages/suite/src/components/connection/thp/ThpAutoconnectInfoModal.tsx
juriczech: How about creating a thunk for cancellation which would dispatch `finishThpFlow()` as well as `startOrRestartDiscoveryThunk()` so we could use it on both platforms? 🙏

Lemonexe: Good idea. Though the middlewares are still different, and the condition to delay discovery would need to get implemented in mobile, the middlewares are getting closer and closer together, hopefully we can get it united someday.

suite-native/discovery/src/discoveryMiddleware.ts
Lemonexe: We think alike, I also wanted to do just that :laughing: 
But I discussed with @marekrjpolak that we should keep the should rediscover selectors to only **re**discovery.
They are used in other situations, such as manual refresh buttons, which would be undesired to render before discovery.   For example if you are in the settings, discovery has not started yet, but on the side panel there would be visible refresh button.

juriczech: Makes sense.

packages/suite/src/components/connection/thp/ThpAutoconnectInfoModal.tsx
juriczech: In ideal scenario, I think the autoconnect thunk state (pending / fulfilled / rejected) could handle this delay and UI rendering IMO. But no need to implement that now I guess so we don't get too stuck on this as it's a small fix and would require some refactoring to achieve it.

Lemonexe: Cleaned up in 60ed2560

And yes, I could trigger the discovery on thunk fulfillment, but it'd still not handle the Cancel flow, I'd have to dispatch an arbitrary action just to trigger a middleware → which sounds to me that middleware is not the place to handle it.

By the way, is this relevant at all on mobile?

juriczech: I'm not sure whether it's a problem on mobile, but I assume it should behave similarly. That was why I was asking but I'm not sure TBH.

juriczech (commented):
Suite-native changes LGTM 🚀 Guess someone from desktop might do better job at code review since it's mostly work done on desktop platform.

packages/suite/src/middlewares/wallet/discoveryMiddleware.ts
```diff
@@ -50,6 +50,11 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
         const isDeviceReady =
             device?.connected && isDeviceAcquired(device) && (!isDeviceLocked || becomesAcquired);
 
+        // delay discovery if THP Autoconnect modal is open (discovery executed by the modal), as it is the only
+        // THP step that takes place *after* device acquisition, and also needs device interaction to complete.
+        const isTHPAutoconnectModal = selectThpStep(getState()) === 'AutoconnectInfo';
+        const isUIReady = !isTHPAutoconnectModal;
```
peter-sanderson, line 56: I love this naming! :heart:

peter-sanderson (approved):

Lemonexe:
7 tests from Suite Web E2E [confirmed broken](https://satoshilabs.slack.com/archives/CKYFBT2C9/p1757661174740719), merging regardless.

Issue #21374, opened 2025-09-08
Endless loading in THP pairing


komret:
Related to https://github.com/trezor/trezor-suite/issues/21375

1. Device is connected but THP not yet paired
2. Cancel THP pairing in Suite
3. Click the button to pair again
4. Endless loading

https://github.com/user-attachments/assets/1dfc8266-cc74-4186-aacc-efc40ab2f3a4

The pill is broken here but that should be fixed with the final renders.

OriginalEveres:
I'm upvoting, this happens in the onboarding flow as well

szymonlesisz:
i think this is FW THP issue,

this scenario works with emulator but doesnt work with TS7 build regardless of transport BT  and/or USB

OriginalEveres:
@szymonlesisz nah, I already have a fix for this in BT mode, im fixing it now for cable mode. 

I called with @martykan and he will check one thing in connect in which I think its buggy.

PR #21416, opened 2025-09-10
Fix/thp abort workflow

szymonlesisz:
## Description
Partial revert of https://github.com/trezor/trezor-suite/commit/0789c968622dcc14171e480018a198b189e7f28b

Bug discovered while investigating [this issue](https://github.com/trezor/trezor-suite/issues/21374)

Turns out i was wrong [here](https://github.com/trezor/trezor-suite/pull/20951)

> while doing the tests i also noticed that abortThpWorkflow didnt reject/abort pending device.currentRun promise and hungs forever.

This claim is (was) true only in the tests because `ui-event-*` is dispatched before `ButtonAck` message is actually sent to the device, thats why abort was effective - but not exactly correct.

communication workflow should be:

1. Host: -> GetAddress
1. Trezor: <- ThpAck
1. Trezor: <- ButtonRequest
1. Host: -> ThpAck
1. Host:  -> ButtonAck + emit ui-event at the same time
1. Trezor: <- ThpAck
1. Host: in read state ....
1. Host: -> `Cancel`
1. Trezor: <- `Cancelled`
1. Host: -> `ThpAck`

currently last 2 points are not done in connect because `device.currentRun` is aborted, so Trezor is leaved in retransmission state (sends Cancelled until receives ThpAck)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/thp-abort-workflow&lookback=7)

### DIFF
--- a/packages/connect/e2e/tests/device/thpPairing.test.ts
+++ b/packages/connect/e2e/tests/device/thpPairing.test.ts
@@ -14,7 +14,6 @@ describe('THP pairing', () => {
 
     const waitForDevice = async (settings: Partial<ConnectSettings['thp']>) => {
         await initTrezorConnect(controller, {
-            debug: true,
             pendingTransportEvent: false,
             thp: {
                 appName: 'TrezorConnect',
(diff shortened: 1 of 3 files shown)


mroz22 (approved):
tested and it works for me 👍

PR #21434, opened 2025-09-10
Feat/bt napi connect module

szymonlesisz:
## Description
Followup to https://github.com/trezor/trezor-suite/pull/20982

use NAPI bindings only to call  `connectDevice` in the main thread while the rest of the logic is running on the background in rust binary.

NAPI module contains only connect device function without all other logic and state management.

Its worth to consider rewrite other functions/methods like this

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=feat/bt-napi-connect-module&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=feat/bt-napi-connect-module&lookback=7)

### DIFF
--- a/packages/suite-desktop-core/src/libs/processes/BluetoothProcess.ts
+++ b/packages/suite-desktop-core/src/libs/processes/BluetoothProcess.ts
@@ -71,22 +71,35 @@ export class BluetoothProcess extends BaseProcess {
         };
     }
 
-    async start() {
+    start() {
         if (this.debug) {
             process.env.RUST_LOG = 'debug';
             process.env.RUST_BACKTRACE = '1';
         }
 
-        if (isMacOs()) {
+        return super.start();
+    }
+
+    // workaround for macos pairing issue
+    async connectAndSubscribeInMainThread(
+        id: string,
+    ): Promise<{ success: true } | { success: false; error: string }> {
+        if (!isMacOs()) {
+            return Promise.resolve({ success: true });
+        }
+
+        try {
             const processPath = path.join(super.getProcessDir(), `index.js`);
             this.logger.info(this.logTopic, `Loading bluetooth native module from ${processPath}`);
 
-            const { trezorBluetoothRun } = await import(/*webpackIgnore: true */ processPath);
-            trezorBluetoothRun(this.port);
+            const { connectDevice } = await import(/*webpackIgnore: true */ processPath);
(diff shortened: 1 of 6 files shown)


packages/suite-desktop-core/src/libs/processes/BluetoothProcess.ts
```diff
@@ -71,22 +71,35 @@ export class BluetoothProcess extends BaseProcess {
...
+        }
+
+        try {
             const processPath = path.join(super.getProcessDir(), `index.js`);
             this.logger.info(this.logTopic, `Loading bluetooth native module from ${processPath}`);
 
-            const { trezorBluetoothRun } = await import(/*webpackIgnore: true */ processPath);
-            trezorBluetoothRun(this.port);
+            const { connectDevice } = await import(/*webpackIgnore: true */ processPath);
+            await connectDevice(id, 30000, (...args: any[]) => {
```
martykan, line 96: Looks like callback is not used anymore and can be removed

packages/transport-bluetooth/scripts/Dockerfile
```diff
@@ -38,6 +38,19 @@ RUN cargo build --target x86_64-unknown-linux-gnu --release --bins && \
   cp ./target/x86_64-unknown-linux-gnu/release/trezor-bluetooth /output/linux-x64/trezor-bluetooth && \
   strip /output/linux-x64/trezor-bluetooth
 
+# Build for macOS ARM64
```
martykan, line 41: Dockerfile is updated but binaries are not included in this PR.
Not sure if intentional?

martykan (approved):
Works well (after building binaries on top of `feat/bt-server-src`)

packages/suite-desktop-core/src/libs/processes/BluetoothProcess.ts
szymonlesisz: will be used to pass notifications (device updates) in following PR

packages/transport-bluetooth/scripts/Dockerfile
szymonlesisz: intentional as this PR was not based on top of `feat/bt-server-src` but since it's already merged i will run dev build and include binaries here

PR #21435, opened 2025-09-10
feat: try to forget BT in OS when wiping or forgetting device

### LINKED ISSUE #21387: `Forget this device` button does not disconnect BT device thus it just reconnects
https://github.com/user-attachments/assets/04597823-b120-411c-a1b8-0538e441f50e

### LINKED ISSUE #21395: Instruct users to remove the BT pairing from OS after wipe
Remove BT pairing from system if possible after wipe (Windows, Android, Linux).
Or instruct user to remove it by themself:

[Figma](https://www.figma.com/design/yl2g0XLU9iIgJsrxSOqkS9/Bluetooth---THP--Final-?node-id=3084-11190&t=HRx3RgIJQcxWUrj2-1)

Whenever device is wiped (normal or bootloader), remove also BT & THP pairing.

Lemonexe:
## Description

CR commit by commit recommended:

- move code within file for `no-use-before-define`, I extracted a commit for transparency of the next one: 
- try to remove BT device from OS when `forgetSingleDevicePersistentDataThunk`
  - this really works only on Windows but on macOS | Linux it will at least yield the Manual Device Removal modal
  - and call `forgetSingleDevicePersistentDataThunk` when wiping device
    - BTW this also removes THP credentials and device persistent data, which was omitted before 

## Related Issue

Resolve #21395
Resolve #21387

## Screenshots:

Wiping device on Linux. The Bluetooth removal works the same for forgetting a device.
→ leads to the Manual Device Removal modal on next pairing (AFAIK needed only on macOS | Linux, while Windows is the only platform where we can successfully unpair device in OS, and do not need this modal)  

[Screencast from 2025-09-10 19-03-19.webm](https://github.com/user-attachments/assets/66c2547a-07b0-4965-8519-b8f843b6560c)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/forget-bt-also-in-os-also-on-wipe&lookback=7)

### DIFF
--- a/suite-common/wallet-core/src/device/deviceThunks.ts
+++ b/suite-common/wallet-core/src/device/deviceThunks.ts
@@ -1,4 +1,4 @@
-import { bluetoothActions } from '@suite-common/bluetooth';
+import { bluetoothActions, selectKnownDevices } from '@suite-common/bluetooth';
 import { createThunk } from '@suite-common/redux-utils';
 import { AcquiredDevice, TrezorDevice } from '@suite-common/suite-types';
 import {
@@ -425,12 +425,14 @@ type ForgetAllDeviceDataThunkParams = {
  */
 export const forgetSingleDevicePersistentDataThunk = createThunk(
     `${DEVICE_MODULE_PREFIX}/forgetSingleDevicePersistentDataThunk`,
-    ({ device }: ForgetAllDeviceDataThunkParams, { dispatch }) => {
+    ({ device }: ForgetAllDeviceDataThunkParams, { dispatch, extra }) => {
         if (typeof device.id === 'string') {
             dispatch(deviceActions.forgetDevicePersistentData({ deviceId: device.id }));
         }
         if (device.bluetoothProps !== undefined) {
             dispatch(bluetoothActions.removeKnownDeviceAction({ id: device.bluetoothProps.id }));
+            // try to remove OS-level Bluetooth bonds, if supported by the platform
+            dispatch(extra.thunks.forgetBluetoothDevice({ bluetoothId: device.bluetoothProps.id }));
         }
         if (isThpDevice(device)) {
             dispatch(thpActions.removeCredentials({ credentials: device.thp.credentials }));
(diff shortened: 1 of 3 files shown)


OriginalEveres (approved):
LGTM, not tested tho

suite-common/wallet-core/src/device/deviceThunks.ts
```diff
@@ -464,11 +464,12 @@ type ForgetAllDeviceDataThunkParams = {
  */
 export const forgetSingleDevicePersistentDataThunk = createThunk(
     `${DEVICE_MODULE_PREFIX}/forgetSingleDevicePersistentDataThunk`,
-    ({ device }: ForgetAllDeviceDataThunkParams, { dispatch }) => {
+    ({ device }: ForgetAllDeviceDataThunkParams, { dispatch, extra }) => {
         if (typeof device.id === 'string') {
             dispatch(deviceActions.forgetDevicePersistentData({ deviceId: device.id }));
         }
         if (device.bluetoothProps !== undefined) {
+            dispatch(extra.thunks.forgetBluetoothDevice({ bluetoothId: device.bluetoothProps.id }));
```
peter-sanderson: In case, you are forgetting all devices and setting the array to `[]`, I think this shall be there as well for bluetooth devices.

peter-sanderson: Maybe we shall reopen our last discussion about this `forgetAllDevicesPersistentDataThunk` 

What if we:
- really run through the array and do `forgetSingleDevicePersistentDataThunk` one-by-obe.
- set the arrays (plural) to `[]` just to be sure same as you do in `forgetAllDevicesPersistentDataThunk`

OriginalEveres: I did not see the `forgetAllDevicesPersistentDataThunk` and I agree that it should run the `forgetSingleDevicePersistentDataThunk ` one-by-one.

Lemonexe: @peter-sanderson  I gave it some more thought, and realized that iterating `forgetSingleDevicePersistentDataThunk` on all devices  would be principially wrong – doesn't do what it's supposed to:

If you have AutoEject on, but not AutoForget, or if you eject individual remembered disconnected wallets,
then there are orphaned data (BT, THP, devicePersistentData) that should be cleared if you turn on AutoForget.
Those are not referenced by any `devices` (can't iterate through them). This is not an edge-case, but an expected situation that there will be orphaned data.

See 54fd4cd9, I think iterating through devices is fit only for the OS-level BT removal – no possibility to remove all devices, only to imperatively remove one device through bluetoothIpc (and that only works on Windows :laughing: )

peter-sanderson (approved):

suite-common/wallet-core/src/device/deviceThunks.ts
Lemonexe: As discussed I agree it's better to refactor it so that BT and THP are on `persistentDeviceData`, and then it's viable (and sensible) to refactor `forgetAllDevicesPersistentDataThunk` to iterate through `forgetSingleDevicePersistentDataThunk`

I will merge this to resolve the issues though, because it does do the job. The refactoring is extensive.

PR #21445, opened 2025-09-11
feat(suite-native): remove is coin enabling init finished

juriczech:
## Description
- `isCoinEnablingInitFinished` flag is redundant. We can simply detect this by the length of array with enabled coins. Removing this flag simplifies dependencies in packages and some unnecessary screen logic.

Notes for @trezor/qa:
- please retest existing coin enabling logic remains unchanged
- coin enabling init after app wipe and device connection + app settings coin enabling

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=fix/native/remove-isCoinEnablingInitFinished&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=fix/native/remove-isCoinEnablingInitFinished&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=fix/native/remove-isCoinEnablingInitFinished&lookback=7)


suite-native/discovery/src/discoveryMiddleware.ts
```diff
@@ -49,29 +46,25 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
...
                 }
-                if (!isCoinEnablingInitFinished) {
-                    dispatch(setIsCoinEnablingInitFinished(true));
-                }
             }
         }
 
         // if we changed enabled networks, check for token definitions right away
-        if (changeNetworks.match(action) && isCoinEnablingInitFinished) {
+        if (changeNetworks.match(action) && isCoinEnablingFinished) {
```
matejkriz: Here I see a risk of race condition 👀

juriczech: Good point! 

I think the second part of condition can be removed altogether from here 🤔

matejkriz: good point, looks redundant

suite-native/discovery/src/discoveryMiddleware.ts
```diff
@@ -49,29 +46,25 @@ export const prepareDiscoveryMiddleware = createMiddlewareWithExtraDeps(
...
             changeCoinVisibility.fulfilled.match(action) ||
-            setIsCoinEnablingInitFinished.match(action) ||
             accountsActions.changeAccountVisibility.match(action) ||
             createAndBackupWalletThunk.fulfilled.match(action) ||
             recoverWalletThunk.fulfilled.match(action)
         ) {
             const device = selectSelectedDevice(getState());
             if (
-                isCoinEnablingInitFinished &&
+                isCoinEnablingFinished &&
```
matejkriz: Here can be a race condition too.

juriczech: Thinking if this can be removed as well or not. I guess unless we ever set enabled coins as empty array or implement some kind of reset mechanism, this shouldn't be necessary, right?

juriczech: Tested this on develop & this branch and works exactly same. There's no race condition. However on both branches, this gets triggered n times, where n is the number of coins you enable in the form. 

I think next improvement could be to enable enabling (😆) multiple coins within one dispatch upon submitting the form so this doesn't get unnecessarily triggered more than once, which is what we need when submitting the form. 

However I'd keep that as a follow up in the future and mark this as "safe" in a sense that it behaves same as on desktop.

suite-native/discovery/src/discoveryMiddleware.ts
juriczech: Removed 👍

matejkriz (approved):
Works well! And I think I finally realized why we originally had a separate isCoinEnablingInitFinished state — it’s because we used to enable the coins as soon as the user toggled them. But now we have it as a form that gets confirmed all at once, so it’s no longer necessary.

PR #21472, opened 2025-09-12
chore: add a madge check for circular imports

peter-sanderson:
Circular import/dependencies are an long-standing issue we have. This PR tries to help it not getting worse by preventing creation of the new circular imports.

1. It simply makes a diff against hard-coded snapshot.
2. Every time we fix a circular import => we update snapshot until its empty
3. Profit :tada: 

Part of https://github.com/trezor/trezor-suite/issues/19072

image
image

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=circular_dep&lookback=7)

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=circular_dep&lookback=7)

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=circular_dep&lookback=7)


scripts/circular-dependencies/madge-snapshot.txt
Lemonexe, line 1: :see_no_evil: :smiling_face_with_tear:  :hear_no_evil: :facepalm: :clown_face:

scripts/circular-dependencies/check-circular-dependencies.sh
```diff
@@ -0,0 +1,25 @@
...
+TMP_FILE="$(mktemp)"
+
+# Run madge, allow it to fail
+if ! npx madge --circular --extensions ts,tsx,js,jsx --exclude "node_modules|lib|libDev" packages suite suite-native suite-common | tail -n +2 > "$TMP_FILE"; then
+  echo "⚠️ madge exited with non-zero status, continuing to compare results..."
+  echo
+fi
+
+# Compare results
+if ! git diff --no-index --color=always "$SNAPSHOT_FILE" "$TMP_FILE"; then
```
Lemonexe, line 17: :thought_balloon: note: We know that the madge output is deterministic, but how chaotic is it? It's just a theoretical concern that if the order is chaotic, even a small status quo PR (e.g. file reorganization) significantly changes the order of the lines, or order of the files in each line that represents the cyclic graph. And then it would not be easy to tell from the diff it the cycles topology is the same or isn't.
That'd be annyoing a slow us down a lot, but hopefully not, we can only see :grin:

Lemonexe (approved):
Perfect! Tried locally with a simple file rename, and the changes are very clear to approve or dismiss. Let's see how well it will work :crossed_fingers:

scripts/circular-dependencies/check-circular-dependencies.sh
peter-sanderson: see https://github.com/trezor/trezor-suite/pull/21472/commits/834e9bc5854ee0df052228b14d63be139ab80f8e :wink:

Lemonexe: Nice!!!

scripts/circular-dependencies/check-circular-dependencies.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+set -uo pipefail   # no -e here, we’ll handle errors manually
+export GIT_PAGER=cat  # prevent diffs in vim, ...
+
+SNAPSHOT_FILE="scripts/circular-dependencies/madge-snapshot.txt"
+TMP_FILE="$(mktemp)"
+
+# Run madge, allow it to fail
+if ! npx madge --circular --extensions ts,tsx,js,jsx --exclude "node_modules|lib|libDev" packages suite suite-native suite-common | tail -n +2 | sed 's/^[0-9]\+)//' | sort > "$TMP_FILE"; then
```
Lemonexe: Small nit: `sed 's/^[0-9]\+) *//'`  to remove the space that is now leading each line
(also below in the echo)

scripts/circular-dependencies/check-circular-dependencies.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+set -uo pipefail   # no -e here, we’ll handle errors manually
+export GIT_PAGER=cat  # prevent diffs in vim, ...
```
Lemonexe:
```suggestion
export GIT_PAGER=cat  # cat used so that raw output is compared (prevent pagination or diff tools such as vim)
```

peter-sanderson: fixed

scripts/circular-dependencies/check-circular-dependencies.sh
peter-sanderson: fixed

martykan (changes requested):
Not getting consistent output on MacOS due to older BSD vs GNU sed implementation.
Can you try `sed -E 's/^[0-9]+) *//'` or `sed 's/^[0-9][0-9]*) *//'` on Linux?

martykan (approved):

peter-sanderson:
Those 2 failing tests are failing on develop as well => merging

PR #21482, opened 2025-09-12
refactor(e2e-suite): Disable debug mode in tests after onboarding

Vere-Grey:
Disabling debug mode will bring our tests closer to production like state. There won't be any hidden or not-finished features.

I had to introduce redux store exposing in desktop tho. Alternative we could decouple debug mode and fw / authenticity checks. But I assume that would be more work.
Please let me know any idea how to improve my suggested implementation or whether this could introduce problems or risks. 

## Related Issue

Resolve 

## Screenshots:

🔍🖥️ **Suite desktop test results:** [View in Currents](https://app.currents.dev/projects/4ytF0E/runs?branch=refactor-e2e-tests-without-debug&lookback=7)

🔍🖥️ **Suite web test results:** [View in Currents](https://app.currents.dev/projects/Og0NOQ/runs?branch=refactor-e2e-tests-without-debug&lookback=7)

### DIFF
--- a/packages/suite-desktop-core/e2e/tests/settings/electrum.test.ts
+++ b/packages/suite-desktop-core/e2e/tests/settings/electrum.test.ts
@@ -9,7 +9,7 @@ test.describe(
     { tag: ['@group=settings', '@desktopOnly'] },
     () => {
         test.beforeEach(async ({ onboardingPage }) => {
-            await onboardingPage.completeOnboarding();
+            await onboardingPage.completeOnboarding({ keepDebugModeEnabled: true });
         });
 
         test(
--- a/packages/suite-desktop-core/src/app.ts
+++ b/packages/suite-desktop-core/src/app.ts
@@ -65,6 +65,7 @@ const createMainWindow = (winBounds: WinBounds, store: Store) => {
             webSecurity: !isDevEnv,
             allowRunningInsecureContent: isDevEnv,
             preload: path.join(__dirname, 'preload.js'),
+            additionalArguments: hasSwitch('expose-store') ? ['--expose-store'] : [],
         },
         icon: path.join(global.resourcesPath, 'images', 'icons', '512x512.png'),
         backgroundColor: colorVariants[darkTheme ? 'dark' : 'standard'].backgroundSurfaceElevation0,
--- a/packages/suite-desktop-core/src/preload.ts
+++ b/packages/suite-desktop-core/src/preload.ts
@@ -4,6 +4,7 @@ import { exposeIpcProxy } from '@trezor/ipc-proxy';
 import { getDesktopApi } from '@trezor/suite-desktop-api';
 
 import '@sentry/electron/preload'; // With this only IPCMode.Classic is ever taken into account
+import { hasSwitch } from './libs/process-switches';
 
 contextBridge.exposeInMainWorld(
     ...exposeIpcProxy(ipcRenderer, [
@@ -16,3 +17,6 @@ contextBridge.exposeInMainWorld(
 
 const desktopApi = getDesktopApi(ipcRenderer);
 contextBridge.exposeInMainWorld('desktopApi', desktopApi);
+contextBridge.exposeInMainWorld('desktopFlags', {
+    exposeStore: hasSwitch('expose-store'),
+});
--- a/packages/suite-desktop-ui/src/MainDesktop.tsx
+++ b/packages/suite-desktop-ui/src/MainDesktop.tsx
@@ -99,6 +99,11 @@ export const init = async (container: HTMLElement) => {
     const { statePatch } = await desktopApi.handshake();
     const store = initStore(preloadAction, statePatch);
 
+    // Expose Redux store for Playwright/e2e tests
+    if (typeof window !== 'undefined' && window.desktopFlags?.exposeStore) {
+        (window as any).store = store;
+    }
+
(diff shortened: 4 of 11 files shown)


packages/suite-desktop-ui/src/MainDesktop.tsx
```diff
@@ -99,6 +99,11 @@ export const init = async (container: HTMLElement) => {
     const { statePatch } = await desktopApi.handshake();
     const store = initStore(preloadAction, statePatch);
 
+    // Expose Redux store for Playwright/e2e tests
+    if (typeof window !== 'undefined' && window.desktopFlags?.exposeStore) {
+        (window as any).store = store;
```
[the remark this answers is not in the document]
Vere-Grey: I could define it as any in global.d.ts but that's it.

packages/suite-desktop-core/src/app.ts
```diff
@@ -65,6 +65,7 @@ const createMainWindow = (winBounds: WinBounds, store: Store) => {
             webSecurity: !isDevEnv,
             allowRunningInsecureContent: isDevEnv,
             preload: path.join(__dirname, 'preload.js'),
+            additionalArguments: hasSwitch('expose-store') ? ['--expose-store'] : [],
```
Vere-Grey, line 68: This would not scale well with other args in future. I will refactor this if you confirm this is the way we want to go.

packages/suite-desktop-core/e2e/tests/settings/electrum.test.ts
```diff
@@ -9,7 +9,7 @@ test.describe(
     { tag: ['@group=settings', '@desktopOnly'] },
     () => {
         test.beforeEach(async ({ onboardingPage }) => {
-            await onboardingPage.completeOnboarding();
+            await onboardingPage.completeOnboarding({ keepDebugModeEnabled: true });
```
Vere-Grey, line 12: Test uses Regtest which is available only in debug mode.

packages/suite-desktop-core/src/app.ts
Lemonexe: Currently there is already a way to pass information from electron-main to electron-renderer: the `statePatch`. But it's not as straightforward, because it's passed through the handshake Ipc api.
This implements a new way of passing information from main to renderer.

~Alternative proposed here https://github.com/trezor/trezor-suite/pull/21516~ but does not work.
I think passing it via `exposeInMainWorld` is cleaner and more reliable than passing it via handshake.
So maybe even the `statePatch` shall be moved there if this works well.

packages/suite-desktop-ui/src/MainDesktop.tsx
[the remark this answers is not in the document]
Lemonexe: It would be better to do like on Web using [global.d.t.s](https://github.com/trezor/trezor-suite/blob/e0c565d78c933f3c9a39c7bcc09ea96520f9e958/packages/suite-web/src/global.d.ts) but this is outdated now..

packages/suite-desktop-core/src/preload.ts
```diff
@@ -16,3 +17,6 @@ contextBridge.exposeInMainWorld(
 
 const desktopApi = getDesktopApi(ipcRenderer);
 contextBridge.exposeInMainWorld('desktopApi', desktopApi);
+contextBridge.exposeInMainWorld('desktopParams', {
```
Lemonexe: I suggest renaming to `desktopParams`, as there might be params other than boolean in the future.

packages/suite-desktop-ui/src/MainDesktop.tsx
```diff
@@ -89,7 +91,9 @@ const MainDesktop = () => {
 };
 
 export const init = async (container: HTMLElement) => {
-    initSentry(SENTRY_CONFIG);
+    if (!isRunningWithinPlaywright) {
+        initSentry(SENTRY_CONFIG);
+    }
```
Lemonexe: For consistency with Web

packages/suite-desktop-ui/src/global.d.ts
```diff
@@ -1,5 +1,6 @@
-type TrezorConnectIpcChannel = (method: string, ...params: any[]) => Promise<any>;
-
 interface Window {
-    TrezorConnectIpcChannel?: TrezorConnectIpcChannel; // Electron API
```
Lemonexe, line 4: obsolete, this is never referenced anywhere in repo

packages/suite-web/src/support/usePlaywright.ts
```diff
@@ -1,24 +1,23 @@
...
+ * Used to augment window object with redux store instance to make it accessible in tests.
  */
 export const usePlaywright = () => {
     const store = useStore();
 
     useEffect(() => {
-        if (typeof window !== 'undefined' && window.Playwright) {
+        if (isRunningWithinPlaywright) {
             window.store = store;
-            window.TrezorConnect = TrezorConnect;
```
Lemonexe: not used anymore, this was probably for Cypress, so I deleted it.

Lemonexe (approved):
:rocket: 

EDIT: I tested manually that sentry is correctly initialized in normal run of Suite Web and Suite Desktop :heavy_check_mark:

HajekOndrej (approved):

PR #21499, opened 2025-09-12
feat: store releases JSON once downloaded for FW installation in desktop

karliatto:
## Description
Adding improvement to the stored downloaded FW in desktop. It will also store the releases JSON when storing the FW binary so in following uses of it, it can use the downloaded one and don't download it again.

## Related Issue

🔍🖥️ **Suite native android test results:** [View in Currents](https://app.currents.dev/projects/iUe1Y4/runs?branch=feat/store-also-jsons-of-downloaded-bins&lookback=7)


marekrjpolak (approved):
Makes sense to me :+1:

